PR #9597, opened 2016-04-10
VideoPlayer: some rework after b37c8565e080b89fb51e2f5963a38cd6360d021c

FernetMenta:
see title
restores some still needed behaviour that got lost


FernetMenta:
jenkins build this please

PR #7367, opened 2015-06-30
[epg] Sync epg update from client / db on startup

ghost:
On Kodi startup two threads (1: client -> db, 2: db -> EpgContainer) accessing the same database. 1 is checking for new epg entries, 2 is retrieving valid epg's -> crash.

This first approach locks the dbaccess in PersistXXX functions, wich runs but could lead to freezes in the gui when reading Epg's (Timeline window).
So there must be a better solution for this - this PR is just to not forget the issue, it is an major ticket as kodi don't startup anymore after epg db is filled,

Workaround: "Don't use DB for Epg (in TV::EPG settings dialog)

PR #9599, opened 2016-04-10
Revert "[epg] avoid PVR thread to persist epg tables directly (sync i…

FernetMenta:
…ssue with epg thread)"

this change makes no sense. It triggers Persist for systems that want to ignore DB

This reverts commit be24ceb3211aa5b1d0cc6a8395ba29a0b005f0be.

reverts https://github.com/xbmc/xbmc/pull/7367

fixes very slow EPG loading for VNSI that does nor require DB.


FernetMenta:
added 93a45352c607b0a2b841b26bcf7b4495a5ef9e1c

jenkins build this please

PR #9600, opened 2016-04-10
fix CGUIWindowMedia::OnClick() overrides

Montellese:
Someone forgot to fix some of the overrides and some of the calls related to `CGUIWindowMedia::OnClick()` when the signature of the method was changed.

Right now the overrides in `CGUIWindowMusicNav` and `xbmcgui::WindowXMLInterceptor` don't work at all resulting in wrong behaviour due to the overrides not being called.


FernetMenta:
likely that "someone" was me. thanks!
jenkins build this please

FernetMenta:
ok to merge?

PR #9602, opened 2016-04-11
[guiinfo] add ListItem.EndTimeResume

tamland:
Same as EndTime, but if the video is resumed instead of played from the beginning.


tamland:
jenkins build this please

PR #9603, opened 2016-04-11
JSON fix song reported when playing cuesheet ticket #16615

DaveTBlake:
This fixes ticket #16615, a longstanding issue with playing cuesheet. If individual songs from a cuesheet are added to a playlist using JSON then when one of these songs is currently playing `Playlist.GetItems` only ever returns the info for the first song on the cuesheet rather than the actual one playing.  See http://trac.kodi.tv/ticket/16615 for full discussion.

The issue does NOT happen when using Playlist.Add on the whole Album object, only individual songs, nor does it happen when the playlist is created using the GUI rather than JSON.
Example JSON API commands, given songs are on a cuesheet (single file)

{"jsonrpc":"2.0","id":1,"method":"Playlist.Clear","params":{"playlistid":0}}
{"jsonrpc":"2.0","id":1,"method":"Playlist.Add","params":{"playlistid":0,"item":{"songid":141}}}
{"jsonrpc":"2.0","id":1,"method":"Playlist.Add","params":{"playlistid":0,"item":{"songid":142}}}
{"jsonrpc":"2.0","id":1,"method":"Playlist.Add","params":{"playlistid":0,"item":{"songid":143}}}
{"jsonrpc":"2.0","id":1,"method":"Player.Open","params":{"item":{"playlistid":0,"position":0}}}
{"jsonrpc":"2.0","id":1,"method":"Player.GetItem","params":{"properties":["album", "albumartist", "artist", "playcount"], "playerid":1}}

In the long run a redesign of playback commands is needed to simplify things. In the meanwhile this modification will enable cuesheet playback to fucntion correctly from remotes and web interfaces to Kodi.

The solution is that when songs are added to the playlist by JSON they need to pass the songid, just as the GUI does, not just the filename as for songs on a cuesheet the filename does not identify the song. This means CFileitem.m_strPath needs to be a virtual file indicating songid.


DaveTBlake:
Oh yes, I think this should be backported to Jarvis. What is the policy on what gets backported?

arigit:
Tested the change on the 16.1/jarvis stable (manually modifying FileItem.cpp per the diff)  in addition to testing in krypton - worked perfectly. The issue is gone.

DaveTBlake:
@Montellese no changes to API code here only core, but it does fix a bug experienced when using JSON commands so should I bump the JSON version a minor step?

Thanks for the real world testing @arigit, and applying it To Jarvis. @MartijnKaijser do we backport?

@Tolriq like to check this out, as I know you have had other users complain to you about it. I will get a Windows build up if it helps, but it has a Python problem at the moment so may not work for you without messing.

Montellese:
@DaveTBlake: First of all thanks for looking into this and tracking it down.

Concerning the JSON-RPC API version please bump the path number to let devs know that something has been fixed.

Tolriq:
@DaveTBlake no need for test build, fix is simple and the user that reported the problem have confirmed the fix with Yatse.

I do not have necessary cue files to test anyway. For python problem it's just a matter of copying a dll not a big deal :p

I just need to find time to install VS 2015 now :(

DaveTBlake:
Oh phooey! Bump the  JSON-RPC API version path number and now I have "merge conflicts"

DaveTBlake:
jenkins build this please

PR #9595, opened 2016-04-10
fix plugin context menus

tamland:
Fixes the ordering of context items from plugins which got messed up in recent refactoring.

I have also removed the broken 'replaceItems' option as I don't why plugins should be allowed to destroy menus coming from core and other addons, or what this would be useful for.

PR #9605, opened 2016-04-12
[python-api] - do not use core context menus for WindowXML containers

phil65:
@tamland  
This fixes the issues I mentioned in https://github.com/xbmc/xbmc/pull/9595 
Core context menu doesnt really work with listitems added to WindowXML instances, so let script authors deal with that via onAction().

### DIFF
--- a/xbmc/interfaces/legacy/WindowXML.cpp
+++ b/xbmc/interfaces/legacy/WindowXML.cpp
@@ -374,6 +374,9 @@ namespace XBMCAddon
                 PulseActionEvent();
                 return true;
               }
+              // the core context menu can lead to all sort of issues right now when used with WindowXMLs, so lets intercept the corresponding message
+              else if (controlClicked->IsContainer() && message.GetParam1() == ACTION_CONTEXT_MENU)
+                return true;
             }
           }
         }
--- a/xbmc/interfaces/legacy/WindowXML.cpp
+++ b/xbmc/interfaces/legacy/WindowXML.cpp
@@ -319,6 +319,10 @@ namespace XBMCAddon
         }
         break;
 
+      case GUI_MSG_NOTIFY_ALL:
+        // GUI_MSG_NOTIFY_ALL breaks container content, so intercept it.
+        return true;
+
       case GUI_MSG_CLICKED:
         {
           int iControl=message.GetSenderId();


tamland:
im guessing this has something to do with the CGUIMediaWindow 'Interceptor'. So every time media window is changed, it breaks..

phil65:
yup, correct, anything which updates the container of the mediawindow breaks it.
I guess to fix all issues, I also have to intercept MSG_GUI_UPDATE ? That one also breaks the current listing. (happens when addons get updated, or by manually triggering the builltin "Container.Refresh").
I know that this is all quite hacky, but I dont see any better way to get that stuff into a usable shape (I guess I am the only one who uses these parts at all :) )

tamland:
Yep. I'll trust you with this. Not touching that class with a ten foot pole:)

MartijnKaijser:
/me hands over an 11 foot pole ;)

phil65:
I now completely intercept GUI_MSG_NOTIFY_ALL, which fixes the remaining issues I had.
@tamland , does that look fine enough from 10 foot away? :)

xbmc/interfaces/legacy/WindowXML.cpp
```diff
@@ -319,6 +319,13 @@ namespace XBMCAddon
         }
         break;
 
+      case GUI_MSG_NOTIFY_ALL:
+        {
+          // GUI_MSG_NOTIFY_ALL breaks container content, so intercept it.
+          return true;
+        }
+        break;
```
tamland: no need for break and the brackets

xbmc/interfaces/legacy/WindowXML.cpp
```diff
@@ -370,6 +377,9 @@ namespace XBMCAddon
                 PulseActionEvent();
                 return true;
               }
+              // the core context menu can lead to all sort of issues right now when used with WindowXMLs, so lets intercept the corresponding message
+              else if (controlClicked->IsContainer() && message.GetParam1() == ACTION_CONTEXT_MENU)
+                  return true;
```
tamland: double indent

tamland:
Other than cosmetics, it looks ok ..from a distance

phil65:
took care of cosmetics, thx for review.
jenkins build this please.

PR #9608, opened 2016-04-12
VideoPlayer: some cleanup to interfaces

FernetMenta:
see title

@popcornmix please adapt OMX. I could not find this methods used when searching in Xcode. That's a good indicator that something was wrong.


popcornmix:
I'm not spotting any issue with this PR, seems to behave as before.
There is an issue with omxplayer - I am seeing a hang occasionally after seeking or EOF, but it was there prior to this change.

I can't quite remember why we needed SubmittedEOS (it's still there on Jarvis) so I suspect it was added for some reason, but we are not currently doing anything with it.

FernetMenta:
@popcornmix thanks
jenkins build this please

PR #9610, opened 2016-04-12
fix description label in select dialog

tamland:
Removes the 'addon.summary' properties abused by non-addon items in the select dialog which (surprise!) broke in a recent skin refactor. @phil65 fyi.

### DIFF
--- a/addons/skin.estuary/1080i/DialogSelect.xml
+++ b/addons/skin.estuary/1080i/DialogSelect.xml
@@ -83,7 +83,7 @@
 						<height>60</height>
 						<font>font12</font>
 						<textcolor>grey</textcolor>
-						<label>$INFO[ListItem.AddonSummary]</label>
+						<label>$INFO[ListItem.Label2]</label>
 					</control>
 				</itemlayout>
 				<focusedlayout height="125" width="880">
@@ -128,7 +128,7 @@
 						<height>60</height>
 						<font>font12</font>
 						<textcolor>grey</textcolor>
-						<label>$INFO[ListItem.AddonSummary]</label>
+						<label>$INFO[ListItem.Label2]</label>
 					</control>
 				</focusedlayout>
 			</control>
--- a/xbmc/filesystem/AddonsDirectory.cpp
+++ b/xbmc/filesystem/AddonsDirectory.cpp
@@ -656,8 +656,6 @@ CFileItemPtr CAddonsDirectory::FileItemFromAddon(const AddonPtr &addon,
   //TODO: fix hacks that depends on these
   item->SetProperty("Addon.ID", addon->ID());
   item->SetProperty("Addon.Name", addon->Name());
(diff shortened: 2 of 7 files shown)


addons/skin.estuary/1080i/DialogSelect.xml
```diff
@@ -83,7 +83,7 @@
 						<height>60</height>
 						<font>font12</font>
 						<textcolor>grey</textcolor>
-						<label>$INFO[ListItem.AddonSummary]</label>
+						<label>$INFO[ListItem.Property(Description)]</label>
```
Razzeee: This also needs to be PRed against the estuary repo

phil65: Not needed,  will take care of this.

phil65:
+1. Alternative would be to use "listitem.label2" for that sublabel, that is perhaps more in line with the rest of the skinning world.

ronie:
+1 and what @phil65 said :-)

tamland:
ok, changed it to use label2
jenkins build this please

phil65:
For reference: https://github.com/xbmc/skin.estuary/commit/8db76d407dfebb6b9ca78754584de9b31ad74056

xbmc/filesystem/BlurayDirectory.cpp
```diff
@@ -74,7 +74,7 @@ CFileItemPtr CBlurayDirectory::GetTitle(const BLURAY_TITLE_INFO* title, const st
   item->m_strTitle = buf;
   item->SetLabel(buf);
   chap = StringUtils::Format(g_localizeStrings.Get(25007).c_str(), title->chapter_count, StringUtils::SecondsToTimeString(duration).c_str());
-  item->SetProperty("Addon.Summary", chap);
+  item->SetProperty("Description", chap);
```
phil65: you missed this one here.

xbmc/filesystem/AddonsDirectory.cpp
```diff
@@ -656,8 +656,6 @@ CFileItemPtr CAddonsDirectory::FileItemFromAddon(const AddonPtr &addon,
   //TODO: fix hacks that depends on these
   item->SetProperty("Addon.ID", addon->ID());
   item->SetProperty("Addon.Name", addon->Name());
-  item->SetProperty("Addon.Version", addon->Version().asString());
-  item->SetProperty("Addon.Summary", addon->Summary());
```
phil65: ...and here, too? or is there already a label2?

xbmc/filesystem/BlurayDirectory.cpp
tamland: thanks! i've pushed a fix to master

xbmc/filesystem/AddonsDirectory.cpp
tamland: That one should be correct. this was the last thing using those properties.

PR #9612, opened 2016-04-12
Add XTimeUtils to OpenMaxVideo.cpp

zpon:
XTimeUtils.h was removed from PlatformInclude.h in 1cadfe6a768c79b3942c881023888950be90adab

To clarify: Without this change the OpenMaxVideo.cpp fails to compile as it is unable to find the Sleep function used on line 424/427.


akva2:
Whoops.  Not covered by jenkins builds apparently.

PR #9613, opened 2016-04-12
fix stretch 16x9 view mode handling and its label

notspiff:
i'm not sure if adding the view mode to the list of cycle-able modes was on purpose but let's do it correctly atleast. most important: quells my ocd twitch - there is no 'replaygain' view mode!


FernetMenta:
jenkins build and merge

PR #9616, opened 2016-04-13
[android] fix delay when opening programs window

tamland:
Often block for several seconds. There is no need to check for success before adding this source.


tamland:
jenkins build this please

PR #9617, opened 2016-04-13
[chg] add guisetting to configure number of volume control steps

stefansaraev:
lets try again.

### DIFF
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -2599,6 +2599,16 @@
           </constraints>
           <control type="list" format="string" />
         </setting>
+        <setting id="audiooutput.volumesteps" type="integer" label="1302" help="36442">
+          <level>1</level>
+          <default>90</default>
+          <constraints>
+            <minimum>10</minimum>
+            <step>5</step>
+            <maximum>90</maximum>
+          </constraints>
+          <control type="list" format="integer" />
+        </setting>
       </group>
       <group id="3">
         <setting id="audiooutput.passthrough" type="boolean" label="348" help="36368">
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -2296,11 +2296,16 @@ bool CApplication::OnAction(const CAction &action)
       if (m_muted)
         UnMute();
       float volume = m_volumeLevel;
+      int volumesteps = CSettings::GetInstance().GetInt(CSettings::SETTING_AUDIOOUTPUT_VOLUMESTEPS);
+      // sanity check
+      if (volumesteps == 0)
+        volumesteps = 90;
+
 // Android has steps based on the max available volume level
 #if defined(TARGET_ANDROID)
       float step = (VOLUME_MAXIMUM - VOLUME_MINIMUM) / CXBMCApp::GetMaxSystemVolume();
 #else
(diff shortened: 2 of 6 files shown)


zag2me:
Thanks for this. A huge improvement to customization of Kodi.

zag2me:
Sorry if I am bike-shedding and you go round in circles ;) but it seems a GUI setting would be nice.
We already have settings for skip steps so I think this could be added under expert mode?
Again apologies if this is already discussed, I get the feeling you guys already went through this. I'm just speaking from the users point of view here.

stefansaraev:
@zag2me done. may need better help hint. I am not a native speaker so suggestions welcome.

@Montellese could you please check for stupids?

@Montellese side question: if I understand right, settings system ensures that a min/max value is guaranteed. but  as distros may override via appliance.xml and pass invalid values (<10 || > 90), should I do some sanity check?

zag2me:
English string sounds good to me :+1:

xbmc/Application.cpp
```diff
@@ -2296,11 +2296,18 @@ bool CApplication::OnAction(const CAction &action)
       if (m_muted)
         UnMute();
       float volume = m_volumeLevel;
+      int volumesteps = CSettings::GetInstance().GetInt(CSettings::SETTING_AUDIOOUTPUT_VOLUMESTEPS);
+      // appliance.xml may contain invalid values. sanity check
```
FernetMenta: this should not be required. the settings systems takes care for limits.

stefansaraev: not really. hence the appliance.xml comment.

stefansaraev: try this appliance.xml and see what happens.
```
<settings>
  <section id="system">
    <category id="audiooutput">
      <group id="3">
        <setting id="audiooutput.volumesteps">
          <default>0</default>
          <constraints>
            <minimum>-10</minimum>
            <step>1</step>
            <maximum>200</maximum>
          </constraints>
        </setting>
      </group>
    </category>
  </section>
</settings>
```

system/settings/settings.xml
```diff
@@ -2599,6 +2599,16 @@
           </constraints>
           <control type="list" format="string" />
         </setting>
+        <setting id="audiooutput.volumesteps" type="integer" label="1302" help="36442">
+          <level>1</level>
+          <default>90</default>
+          <constraints>
+            <minimum>10</minimum>
+            <step>5</step>
+            <maximum>90</maximum>
```
Montellese: Is 90 the current default or is it 100? If it's 100 I would allow 100 as well.

xbmc/Application.cpp
Montellese: Do we really need to guard against that? An appliance doing something like this basically criples its own system which is certainly not in their interest. The only check really necessary would be for `volumesteps` to not be `0` to avoid a division by zero problem.

Montellese:
settings stuff looks ok.

system/settings/settings.xml
stefansaraev: current is 90.

stefansaraev:
updated.

FernetMenta:
+1

stefansaraev:
jenkins build this please

fritsch:
WIN32 machine seems to have hung ... builds since 10 hours. Good to go :-)

PR #9618, opened 2016-04-13
Fixes deadlock when closing Videoplayer

ghost:
When stopping a running video using the inputstreamaddon the lock removed here leads to an dedlock of kodis mainthread. With advise of @FernetMenta


FernetMenta:
thanks

PR #9620, opened 2016-04-13
Stop scanning directory if scan is cancelled.

tgagneret:
If the list of paths are remote directories which are unreachable it won't enter the DoScan function (because the directories will be seen as non-existent) and it continues the scan of the directories. So we allow the possibility to exit the directory scan loop at each iteration.

For example if there is 10 remote directories and we are stopping the scan after 6, it would avoid to have 4 connections timeout before exiting the loop.

### DIFF
--- a/xbmc/video/VideoInfoScanner.cpp
+++ b/xbmc/video/VideoInfoScanner.cpp
@@ -139,7 +139,11 @@ namespace VIDEO
          * occurs.
          */
         std::string directory = *m_pathsToScan.begin();
-        if (!CDirectory::Exists(directory))
+        if (m_bStop)
+        {
+          bCancelled = true;
+        }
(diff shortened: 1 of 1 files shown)


tamland:
This doesn't look complete. If you just exit the loop, it will proceed with cleaning/compression the db below.

tgagneret:
Hi,
I hope I correctly understood what was the problem, and did the correct fix with this new commit.

da-anda:
@tamland mind having a look again?

tamland:
@tgagneret Can you please format according to [Code guidelines](https://github.com/xbmc/xbmc/blob/master/CODING_GUIDELINES.dox) and squash down to one commit?  Logic-wise it should be good.

tgagneret:
@tamland Done

tamland:
thanks!

PR #9621, opened 2016-04-13
Remove HTTPFile / CURLAddOption PostData added

ghost:
see title


hudokkow:
Please drop https://github.com/xbmc/xbmc/blob/master/xbmc/filesystem/CMakeLists.txt#L25 and https://github.com/xbmc/xbmc/blob/master/xbmc/filesystem/CMakeLists.txt#L107

mapfau:
@hudokkow thx!

mapfau:
Needs some more work - it is currently not working because 
`bool CFile::Open(const CURL& file, const unsigned int flags)`
makes interesting things with CURL

xbmc/filesystem/FileFactory.cpp
```diff
@@ -148,8 +147,9 @@ IFile* CFileFactory::CreateLoader(const CURL& url)
   {
     if (url.IsProtocol("ftp")
     ||  url.IsProtocol("ftps")
-    ||  url.IsProtocol("rss")) return new CCurlFile();
-    else if (url.IsProtocol("http") ||  url.IsProtocol("https")) return new CHTTPFile();
+    ||  url.IsProtocol("rss")
+    ||  url.IsProtocol("http") 
+    ||  url.IsProtocol("https"))return new CCurlFile();
```
arnova: space missing between )) and return

xbmc/URL.cpp
```diff
@@ -412,6 +412,11 @@ void CURL::SetProtocolOptions(const std::string& strOptions)
   }
 }
 
+void CURL::SetPayload(const std::string& strPayload)
+{
+  m_strPayload = strPayload;
+}
```
arnova: Is there no other way we can do this? I'm not sure I like having this here. It's not something that belongs to the URL itself(!) does it?

mapfau: @arnova Does a Header belongs to the URL itself? Its already part of the URL structure.
Sure, I can also push it int the URLOptions Variant, Is it more straightforward in your eyes?

If you have an better idea, pls. let me know. I only want to finish it soon because its needed by the new inputstream features. Thanx!

mapfau: Just thought over it . I'll place it into CURLOptions but:

- Will remove CURL::get wich forms the URL with separators
- Will only support substitution of URL path in the direction URL with separators -> URL + filled URL members,
(for backwards compatibility

Then postdata is completely hidden for you.

OK?

mapfau:
Changed that I pass base64 encoded postdata using CURL File options.
Its only a very small operation and should not have any impact on existing code.

Thanks for reviewing (and merging :-)

xbmc/filesystem/File.cpp
```diff
@@ -261,39 +258,32 @@ bool CFile::Open(const CURL& file, const unsigned int flags)
   try
   {
     bool bPathInCache;
-
-    CURL url(URIUtils::SubstitutePath(file)), url2(url);
```
arnova: Why did you move this?

xbmc/filesystem/File.cpp
```diff
@@ -243,10 +243,7 @@ bool CFile::CURLAddOption(XFILE::CURLOPTIONTYPE type, const char* name, const ch
 
 bool CFile::CURLOpen(unsigned int flags)
 {
-  if (flags & XFILE::READ_AFTER_WRITE)
-    return OpenForWrite(m_curl, true);
-  else
-    return Open(m_curl, flags);
+  return Open(m_curl, flags);
```
arnova: Mind explaining why we don't need this anymore?

arnova:
For the rest seems to look good. Since you're touching rather important areas, I'd like someone else to review it as well. Perhaps @Paxxi or @mkortstiege  ?

arnova:
I think the substitute thing changes behavior so I really like to know why you changed that.

xbmc/filesystem/File.cpp
mapfau: save resources, save time. There is no need to make 2 copies od URL (beside all the things happen in the constructor (Concatting strings from a, barsing this string in b)

xbmc/filesystem/File.cpp
mapfau: The flag was introduced in one of my PR's some weeks ago. It was necessary to wait for the CURL action because writing of POSTDATA has to wait.
This is obsolete now: PostData is prepared before any open call is done.
I also removed the XFILE::READ_AFTER_WRITE flag. No need for this anymore.

mapfau:
The substitute() move changes the situation in exactly one situation: If a CURL is build up not from string. Means, you put all things separately into it (SetProtocol / SetDomain / SetFilename)

In all cases where URL is buid up from a std::string it should work like before

arnova:
The problem is that for any (derived) CURL we will perform the substitute now, where it previously only did it for stuff inside CFile. In theory we could substitute a path, and later on do it again (due to multiple rules matching). This can't happen the way it is now.

Bottomline is: I have a bad feeling about this, and I'm not confident enough this won't introduce any regressions. Back then when I fixed/refactored the substitute stuff it took a lot of work to get things right, so I'd rather keep that as is, if possible. So unless another team member can OK that substitute-change, I think it should be dropped.

xbmc/filesystem/File.cpp
```diff
@@ -261,39 +258,32 @@ bool CFile::Open(const CURL& file, const unsigned int flags)
   try
   {
     bool bPathInCache;
-
-    CURL url(URIUtils::SubstitutePath(file)), url2(url);
-
-    if (url2.IsProtocol("apk") || url2.IsProtocol("zip") )
-      url2.SetOptions("");
```
arnova: Same thing here. Are you sure moving this won't introduce any regressions?

arnova:
Btw. split the 3 changes into separate commits as there's no way telling now why the other changes are there.

mapfau:
Removed moifications in CURL.[h/cpp]

arnova:
Thanks. Looking good now (and simple) :-) 

jenkins build this please.

mapfau:
thank you, @arnova !

mapfau:
clang: error: no such file or directory: '/Users/Shared/jenkins/workspace/IOS/xbmc/filesystem/HTTPFile.cpp'

for android, do you know where it is located?

mapfau:
@arnova Could you pls start an OSX build?

hudokkow:
http://jenkins.kodi.tv/job/OSX-64/8007/

mapfau:
looks good now, thx

arnova:
@MartijnKaijser : Your button, I assume this is fine?

MartijnKaijser:
my button? this is master branch :)

arnova:
@MartijnKaijser : I know ;-) I'm not sure about its use, that's why :-)

mapfau:
below an usecase snippet:

```
void* file = host->CURLCreate(blocks[0].c_str());

//Set our std headers
host->CURLAddOption(file, SSD_HOST::OPTION_PROTOCOL, "acceptencoding", "gzip, deflate");
host->CURLAddOption(file, SSD_HOST::OPTION_PROTOCOL, "seekable", "0");
host->CURLAddOption(file, SSD_HOST::OPTION_HEADER, "Expect", "");
(...)
std::string decoded = b64_encode(reinterpret_cast<const unsigned char*>(blocks[2].data()), blocks[2].size(), false);
host->CURLAddOption(file, SSD_HOST::OPTION_PROTOCOL, "postdata", decoded.c_str());
(...)
if (!host->CURLOpen(file))
```

FernetMenta:
jenkins build this please

MilhouseVH:
It looks like this PR is causing a problem with `pvr.argustv` although I don't know if the problem is with Kodi or the addon. Reported on the forum: http://forum.kodi.tv/showthread.php?tid=269815&pid=2313577#pid2313577 - reverting this PR fixes the issue with the addon.

FernetMenta:
yes we know and we informed the original author of this change and former maintainer of the addon. I don't know who maintains this addon now. The addon needs adaption by whoever cares for this code.

FernetMenta:
@MilhouseVH note that you break maintained inputstream addons when you revert this in your build.

MilhouseVH:
> yes we know and we informed the original author of this change and former maintainer of the addon.

Thanks, I'll open an issue against pvr.argustv for what it's worth.

> @MilhouseVH note that you break maintained inputstream addons when you revert this in your build.

I only reverted it as a one-off to get confirmation it was the reason for the change with respect to pvr.argustv, I won't be reverting this PR in regular builds.

mapfau:
I'll have a look on it

PR #9623, opened 2016-04-14
Use aligned_malloc when using buffers with sws_scale

fritsch:
Since ffmpeg 3 we heavily segfault under certain conditions as ffmpeg sometimes asumes 16 byte aligned memory adresses, which are normally guaranteed when using ffmpeg's alloc methods. As we supply external buffers we need to take great care.

Let's see which includes are missing for Windows, OSX and so on.

Edit: To add, I talked with the ffmpeg guys and they highly, highly suggest to use proper alignment as it will be totally slow when not doing so.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/DVDCodecUtils.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/DVDCodecUtils.cpp
@@ -76,7 +76,7 @@ DVDVideoPicture* CDVDCodecUtils::AllocatePicture(int iWidth, int iHeight)
 
 void CDVDCodecUtils::FreePicture(DVDVideoPicture* pPicture)
 {
-  delete[] pPicture->data[0];
+  av_free(pPicture->data[0]);
   delete pPicture;
 }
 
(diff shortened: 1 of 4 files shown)


fritsch:
@afedchin please verify WinRenderer.cpp as the memory comes from "somewhere" else.

The first commit fixes, e.g. this: http://forum.kodi.tv/showthread.php?tid=269232

fritsch:
jenkins build this please to see which includes are still missing

fritsch:
@FernetMenta should we use av_malloc / av_free instead to be on the safe side for the future? At least in the above  touched files we have a dependency to av_malloc anyways?

FernetMenta:
meh, this crappy ms function spread over the app like cancer. Defining a Windows function for the other platforms is rather hacky and usage of this should not continue. 
Why don't you define a function for all platforms with maybe different implementations. Saves a lot of those ugly ifdefs in the include section.

fritsch:
Yeah ... in this concrete case ... we can just use av_malloc as we have the dependency to sws_scale anyways ...

fritsch:
jenkins build this please with our new and shiny av_malloc / av_free usage

fritsch:
Acked by @FernetMenta via slack.

PR #8904, opened 2016-01-18
Add consistency checks to use of musicbrianz id tags

DaveTBlake:
Some users have difficulty tagging their music files consistently. A common problem, even when tagging using Picard, leaves Kodi with a mis-match in the number or artist names compared to the  number of musizbrainz artist Ids. This results in a mess in the artists held in the library that has given Kodi musicbrainz integration a bad reputation (users want to disable it or don't use mbids at all).

Things have improved with the introduction of the ARTISTS and ALBUMARTISTS tag, but there are still issues and older tagged files may not have these tags. Part of the problem is that Kodi only supports a single item separator, and does not accomodate the standard output from Picard (where the delimiter can vary). Some users know how to apply Picard scripts to fix that, but others don't, and they don't understand why music collections are fine in other players (that don't use the mbids) but go wrong in Kodi.

Common example of difficulties where the are multiple mb Ids but Kodi sees only a single artist include: "Artist1 feat. Artist2" or "Composer; Conductor, Orchestra, Soloist". 

Rather than quietly making a mess, it is a simple matter for Kodi to log in debug when a mis-match has occurred and then attempt to rectify the situation and identify the right number of artist names. There will be times when even this fails, but it will catch a majority of the tagging issues.

I am tempted to suggest we back-port this into Javis even at RC1.

@Paxxi check my stringutils addition. @Scott967,@Razzeee and @zag2me  hope this makes sense to you guys. @jjd_uk maybe you will consider using mbids now? Anyway you know the users this will help.

PR #9625, opened 2016-04-14
Better handling of mis-matching musicbrianz album artist id and hints tags

DaveTBlake:
A follow on from #8904 with more improved handling of mis-matching numbers of Musicbrainz (Album) Artist Ids and (Album) Artist tags when creating songs and albums.

Primarily this was to fix an issue where missing hints (ALBUMARTISTS tag) was causing the album artist tag to be ignored (even when matched mbid) and artist hints and names to be used instead to get album artist names to match the number of mbid. If there was  a mis-match of artist names and mbid then the results were quite confusing. 
For example:
ARTIST = "artist1 feat. artist2"
MUSICBRAINZ ARTISTID = mbid1
MUSICBRAINZ ARTISTID = mbid2
ALBUMARTIST = "artist1"
MUSICBRAINZ ALBUMARTISTID = mbid1

resulted in an artist named "artist1 feat. artist2" with mbid1 for the album even though the number of album artist names and mbids matched, and the lack of ARTISTS tag and mis-match of mbid <-> name count would have been repaired for the song.

Also a user managed to have ARTISTS and ALBUMARTISTS tags present but as single string with an unexpected separator. If one user can do it so can others, so a more flexible approach to parsing both (ALBUM)ARTIST and (ALBUM)ARTISTS tags when the number of names does not match the number of Musicbrainz ids seems worthwhile.

Hence when an inconsistency in the number of artist names to mbid occurrs this applies improved parsing to both hints and names, using multiple item separators progressively, in an attempt to correctly interpret the way music files have been tagged.

@Razzeee got some time to look at this? The sooner it is out there getting more real user data through it the better

### DIFF
--- a/xbmc/music/Album.cpp
+++ b/xbmc/music/Album.cpp
@@ -52,70 +52,102 @@ CAlbum::CAlbum(const CFileItem& item)
   strAlbum = tag.GetAlbum();
   strMusicBrainzAlbumID = tag.GetMusicBrainzAlbumID();
   genre = tag.GetGenre();
-  std::vector<std::string> musicBrainAlbumArtistHints = tag.GetMusicBrainzAlbumArtistHints();
+  std::vector<std::string> musicBrainzAlbumArtistHints = tag.GetMusicBrainzAlbumArtistHints();
   strArtistDesc = tag.GetAlbumArtistString();
 
   if (!tag.GetMusicBrainzAlbumArtistID().empty())
-  { // have musicbrainz artist info, so use it
-    // Establish tag consistency - do the number of musicbrainz ids and number of names in hints or artist match
-    if (tag.GetMusicBrainzAlbumArtistID().size() != musicBrainAlbumArtistHints.size() &&
-      tag.GetMusicBrainzAlbumArtistID().size() != tag.GetAlbumArtist().size())
+  { // Have musicbrainz artist info, so use it
+    
+    // Vector of possible separators in the order least likely to be part of artist name
+    const std::vector<std::string> separators{ " feat. ", ";", ":", "|", "#", "/", ",", "&" };
+
+    // Establish tag consistency
+    // Do the number of musicbrainz ids and number of names in hints and artist mis-match?
+    if (musicBrainzAlbumArtistHints.size() != tag.GetMusicBrainzAlbumArtistID().size() &&
+      tag.GetAlbumArtist().size() != tag.GetMusicBrainzAlbumArtistID().size())
     {
       // Tags mis-match - report it and then try to fix
       CLog::Log(LOGDEBUG, "Mis-match in song file albumartist tags: %i mbid %i name album: %s %s",
-        (int)tag.GetMusicBrainzAlbumArtistID().size(), (int)tag.GetAlbumArtist().size(), strAlbum.c_str(), strArtistDesc.c_str());
+        (int)tag.GetMusicBrainzAlbumArtistID().size(),
+        (int)tag.GetAlbumArtist().size(),
+        strAlbum.c_str(), strArtistDesc.c_str());
       /*
       Most likey we have no hints and a single artist name like "Artist1 feat. Artist2"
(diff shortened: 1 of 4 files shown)


xbmc/music/Album.cpp
```diff
@@ -52,70 +52,103 @@ CAlbum::CAlbum(const CFileItem& item)
...
   if (!tag.GetMusicBrainzAlbumArtistID().empty())
-  { // have musicbrainz artist info, so use it
-    // Establish tag consistency - do the number of musicbrainz ids and number of names in hints or artist match
-    if (tag.GetMusicBrainzAlbumArtistID().size() != musicBrainAlbumArtistHints.size() &&
-      tag.GetMusicBrainzAlbumArtistID().size() != tag.GetAlbumArtist().size())
+  { // Have musicbrainz artist info, so use it
+    
+    // Vector of possible separators in the order least likely to be part of artist name
+    const char *sepinit[] = { " feat. ", ";", ":", "|", "#", "/", ",", "&" };
+    std::vector<std::string> separators(sepinit, std::end(sepinit));
```
phate89: Is sepinit used only to initialize the vector? in c++11 you should be able to do
`std::vector<std::string> separators{ " feat. ", ";", ":", "|", "#", "/", ",", "&" };`
Same thing in Song.cpp

xbmc/utils/StringUtils.cpp
```diff
@@ -759,6 +759,61 @@ std::vector<std::string> StringUtils::Split(const std::string& input, const char
   return results;
 }
 
+std::vector<std::string> StringUtils::Split(const std::vector<std::string> &input, const std::vector<std::string> &delimiters, unsigned int iMaxStrings /* = 0 */)
+{
+  std::vector<std::string> results;
+  if (input.empty())
+    return results;
+
+  for (std::vector<std::string>::const_iterator it = input.begin(); it != input.end(); ++it)
```
phate89: it's the same thing but imo it's cleaner to use range based iterators to iterate in vectors like this:
```for (const auto& item: input)
  results.push_back(item);
```

xbmc/utils/StringUtils.cpp
```diff
@@ -759,6 +759,61 @@ std::vector<std::string> StringUtils::Split(const std::string& input, const char
   return results;
 }
 
+std::vector<std::string> StringUtils::Split(const std::vector<std::string> &input, const std::vector<std::string> &delimiters, unsigned int iMaxStrings /* = 0 */)
```
phate89: Isn't SplitMulti a better name? There's already a SplitMulti and it's similar to that since it takes more than one separator

DaveTBlake:
Thanks @phate89 for the code suggestions. My dev env is in pieces at the moment, attempting to back out of VS2015 (seems incompatible with my old PC) which has brought me to a halt.

xbmc/utils/StringUtils.cpp
DaveTBlake: Not sure. SplitMulti had to have a different name to differentiate between split with a string that should be applied as a single delimter, and a string where each character is a deliimter. SplitMulti is also unused now, I wrote it to do the initial version of tag repair, so do I remove it? Or do I replace it?

phate89: IF it's unused you should remove it.. Imho is better to have a name distinction between a regular split and this complex split..

xbmc/utils/StringUtils.cpp
```diff
@@ -759,6 +759,61 @@ std::vector<std::string> StringUtils::Split(const std::string& input, const char
...
+{
+  std::vector<std::string> results;
+  if (input.empty())
+    return results;
+
+  for (const auto& item : input)
+    results.push_back(item);
```
phate89: You can avoid the loop initializing result with the current vector:
```
if (input.empty())
  return vector<std::string>();

std::vector<std::string> results(input);
```

DaveTBlake:
Thanks for the improvements @phate89 

jenkins build this please

DaveTBlake:
Compile error not related to this PR

PR #9626, opened 2016-04-14
[Jarvis] Fix song reported on playback of songs from cuesheet

DaveTBlake:
Backport of #9603  that fixes ticket #16615, a longstanding issue with playing songs from a cuesheet uisng JSON-RPC API. If individual songs from a cuesheet are added to a playlist using JSON then when one of these songs is currently playing `Playlist.GetItems` only ever returns the info for the first song on the cuesheet rather than the actual one playing. See http://trac.kodi.tv/ticket/16615 for full discussion.

The issue does NOT happen when using Playlist.Add on the whole Album object, only individual songs, nor does it happen when the playlist is created using the GUI rather than JSON.

@arigit has already tested this patch applied to Jarvis, and it solves the bug.


DaveTBlake:
Jenkins build this please

DaveTBlake:
jenkins build this please

PR #9627, opened 2016-04-15
[python-api] - several additions

phil65:
- getCurrentContainerId() allows to properly use the core viewtype management for WindowXML instances.
- setInfo("dbid", xxx) allows to properly emulate database items. This is especially important for stuff like the dataprovider script from @BigNoid. Several context item add-ons depend on listitem.dbid for example.
- getArt() should also be self-explaining.  

@tamland

### DIFF
--- a/xbmc/interfaces/legacy/ListItem.h
+++ b/xbmc/interfaces/legacy/ListItem.h
@@ -333,6 +333,27 @@ namespace XBMCAddon
       String getProperty(const char* key);
 
       /**
+       * getArt(key) -- Returns a listitem art path as a string, similar to an infolabel.\n
+       * \n
+       * key            : string - art name.\n
+       * \n
+       *
+       * - Some default art values (any string possible):
+       *     - thumb         : string - image path
+       *     - poster        : string - image path
+       *     - banner        : string - image path
+       *     - fanart        : string - image path
+       *     - clearart      : string - image path
+       *     - clearlogo     : string - image path
+       *     - landscape     : string - image path
+       *     - icon          : string - image path
+       *
+       * example:
+       *   - poster = self.list.getSelectedItem().getArt('poster')
+       */
+      String getArt(const char* key);
+
+      /**
        * setPath(path) -- Sets the listitem's path.\n
        * \n
        * path           : string or unicode - path, activated when item is clicked.\n
--- a/xbmc/interfaces/legacy/ListItem.h
+++ b/xbmc/interfaces/legacy/ListItem.h
@@ -230,6 +230,7 @@ namespace XBMCAddon
        *     - trailer       : string (/home/user/trailer.avi)
        *     - dateadded     : string (%Y-%m-%d %h:%m:%s = 2009-04-05 23:16:04)
        *     - mediatype     : string - "video", "movie", "tvshow", "season", "episode" or "musicvideo"
+       *     - dbid          : integer (23) - Only add this for items which are part of the local db. You also need to set the correct 'mediatype'!
        * - Music Values:
        *     - tracknumber   : integer (8)
        *     - discnumber    : integer (2)
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -10090,11 +10090,15 @@ std::string CGUIInfoManager::GetItemLabel(const CFileItem *item, int info, std::
   case LISTITEM_DBID:
     if (item->HasVideoInfoTag())
       {
-        return StringUtils::Format("%i", item->GetVideoInfoTag()->m_iDbId);
+        int dbId = item->GetVideoInfoTag()->m_iDbId;
+        if (dbId > -1)
+          return StringUtils::Format("%i", dbId);
       }
     if (item->HasMusicInfoTag())
       {
-        return StringUtils::Format("%i", item->GetMusicInfoTag()->GetDatabaseId());
+        int dbId = item->GetMusicInfoTag()->GetDatabaseId();
+        if (dbId > -1)
+          return StringUtils::Format("%i", dbId);
       }
     break;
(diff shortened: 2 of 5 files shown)


ronie:
> allows to properly use the core viewtype management for WindowXML instances.

sounds cool... though i have no idea what it means. ;-)
care to elaborate?

phil65:
Sure. Most of the time people add content to WindowXML lists via Control.addItems(), but there is also another way to do this, namely WindowXML.addItem(). When using that, you basically get MediaWindow behaviour for that list (example: you can reference to the listitems via listitem.xxx from outside of the container).
The problem is that the script author does not know which viewtype is currently active. By adding getCurrentContainerId(), that will become possible.

phil65:
I appended another commit. While dealing with the new dbid setters and getters i noticed that items with a VideoinfoTag return -1 as dbid in case they dont have a dbid assigned. I think for skinners it is more convenient to check with IsEmpty() instead of isGreater(-1) for dbids. Agreed, @ronie ?

Montellese:
Doesn't the last commit break any skin comparing against `-1`?

Concerning the possibility for plugins to set the database id this might have much further implications than you might think. A lot of code uses the fact that an item has a database id to differentiate between library and plugin items. With this change that is not always true anymore.
Could you explain the exact use case how this would be used? A plugin must not set an arbitrary database id on an item that is not stored in the video or music database or some core logic ends up thinking that the item is an actual library item and overwrites the information in the database with the one from the plugin which might not be wanted.

ronie:
> I think for skinners it is more convenient to check with IsEmpty() instead of isGreater(-1) for dbids. Agreed, @ronie ?

for sure, if no info is available it should return empty.

phil65:
> Doesn't the last commit break any skin comparing against -1?

In theory yes, but breakage will be minimal. We have much bigger breakers in v17. :)
For me issues mainly came up when doing script work (I am enriching remote metadata with local metadata if available in my add-ons)

> Could you explain the exact use case how this would be used? A plugin must not set an arbitrary database id on an item that is not stored in the video or music database or some core logic ends up thinking that the item is an actual library item and overwrites the information in the database with the one from the plugin which might not be wanted.

A valid usecase for plugins (esp. for those which get integrated into skins via dynamic content) is to imitate library listitems. A good example probably is perhaps service.library.data.provider, which provides listitems for several widget types. Those scripts have to do hacks like https://github.com/BigNoid/service.library.data.provider/blob/master/default.py#L269 for example in order to allow skinners or other scripts to have the correct dbid at hand. (example: https://github.com/phil65/context.extendedinfo.dialog/blob/krypton/addon.py#L28-L29 )
This PR tries to get rid of those hacks.

Montellese:
@phil65: I've never used `service.library.data.provider` or any of the other add-ons you mentioned so excuse my ignorance. Are the imitated library items based on real library items just with additional information? Or ar they fake items that don't exist in the database?

phil65:
They are based on real library items. The aim for those plugins basically is to create listitems via python which are identical to the ones delivered via core, all infolabels should be identical.

xbmc/interfaces/legacy/ListItem.h
```diff
@@ -333,6 +333,17 @@ namespace XBMCAddon
...
       /**
+       * getArt(key) -- Returns a listitem art value as a string, similar to an infolabel.\n
+       * \n
+       * key            : string - art name.\n
+       * \n
+       *
+       * example:
+       *   - poster = self.list.getSelectedItem().getArt('poster')
+       */
+      String getArt(const char* key);
```
tamland: Could use some better documentation. 'It returns a value'  A value of what? And list the available types please, as setArt does.

Montellese:
How do we protect core against add-ons using wrong/arbitrary database ids which could either overwrite data from existing library items or try to write library items that don't exist?

phil65:
In my opinion a warning in docs should be fine that this should only be used for local items and in combination with the correct mediatype value, I can add that. Would that be sufficient for you?

xbmc/interfaces/legacy/ListItem.h
phil65: Ok.

Montellese:
Ideally there would be a way to prevent plugins from using invalid database ids in the python interface implementation but that's probably not possible right now or would be too much work so a warning is certainly nice.
Maybe we should change the method to set the id and the type to force plugin authors to set both at the same time.

phil65:
@Montellese : I updated the docs, is it fine enough now?

Montellese:
Yes, better.

phil65:
jenkins build this please.

mapfau:
This PR leads to an issue commented in:

https://github.com/xbmc/xbmc/commit/65ce5ef4753e3a3d12577bd487a089adf9912479

phil65:
Already fixed on estuary repository.

phil65:
https://github.com/xbmc/skin.estuary/commit/d13f4c935c9dea4525c7bea9a8393b4e8ea1fc69

tamland:
@phil65 you should push things like that to _this_ repo because otherwise it will be left broken in nightly and for everyone doing development. Also, that's a private repo mapfau can't access.

phil65:
We´re in pre-alpha, I noticed it after the merge, and there are also a lot of other small fixes I dont push on a daily basis....

tamland:
Sure, you don't have to push every small fix daily. But you can see how it's not ideal that issues are known and fixes exists, but can't be acquired yet right?

phil65:
Then the thing to do is probably to discuss whether the estuary repo should get public, and not instantly attacking someone for their workflow (which was agreed on by the team).

phil65:
https://github.com/xbmc/xbmc/pull/9665

tamland:
It was never my intention to attack anyone. sorry for bringing it up

PR #9629, opened 2016-04-15
VideoPlayer: consider offset to vsync

FernetMenta:
PTS of a video frame refers to the time the frames comes to display. But this time depends on vsync. This change determines the offset from video player's clock to vsync and adjusts audio sync.

Render thread looks at the middle of the the timespan of a frame. This reduces the chance that it doesn't pick a frame because it is not due by i.e. a 1ms. This resulted in rendering a frames twice and having to drop /skip next frames.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDClock.cpp
+++ b/xbmc/cores/VideoPlayer/DVDClock.cpp
@@ -39,6 +39,8 @@ CDVDClock::CDVDClock()
   m_systemAdjust = 0;
   m_speedAdjust = 0;
   m_startClock = 0;
+  m_vSyncAdjust = 0;
+  m_frameTime = DVD_TIME_BASE / 60;
 
   m_videoRefClock.reset(new CVideoReferenceClock());
   m_lastSystemTime = m_videoRefClock->GetTime();
@@ -86,6 +88,18 @@ double CDVDClock::GetClock(double& absolute, bool interpolated /*= true*/)
   return SystemToPlaying(current);
 }
 
+void CDVDClock::SetVsyncAdjust(double adjustment)
+{
+  CSingleLock lock(m_critSection);
+  m_vSyncAdjust = adjustment;
+}
+
+double CDVDClock::GetVsyncAdjust()
+{
+  CSingleLock lock(m_critSection);
+  return m_vSyncAdjust;
+}
+
 void CDVDClock::SetSpeed(int iSpeed)
 {
   // this will sometimes be a little bit of due to rounding errors, ie clock might jump abit when changing speed
@@ -124,38 +138,40 @@ double CDVDClock::GetSpeedAdjust()
   return m_speedAdjust;
 }
 
-bool CDVDClock::Update(double clock, double absolute, double limit, const char* log)
+double CDVDClock::ErrorAdjust(double error, const char* log)
 {
-  double was_absolute;
-  double was_clock;
+  CSingleLock lock(m_critSection);
 
-  {
-    CSingleLock lock(m_critSection);
-    was_absolute = SystemToAbsolute(m_startClock);
-    was_clock = m_iDisc + absolute - was_absolute;
-  }
-
-  double error = std::abs(clock - was_clock);
+  double clock, absolute, adjustment;
+  clock = GetClock(absolute);
 
   // skip minor updates while speed adjust is active
   // -> adjusting buffer levels
   if (m_speedAdjust != 0 && error < DVD_MSEC_TO_TIME(100))
   {
-    return false;
+    return 0;
   }
-  else if (error > limit)
-  {
-    Discontinuity(clock, absolute);
 
-    CLog::Log(LOGDEBUG, "CDVDClock::Discontinuity - %s - was:%f, should be:%f, error:%f"
-                      , log
-                      , was_clock
-                      , clock
-                      , clock - was_clock);
-    return true;
+  adjustment = error;
+
+  if (m_vSyncAdjust != 0)
+  {
+    if (error > 0.5 * m_frameTime)
+      adjustment = m_frameTime;
+    else if (error < -0.5 * m_frameTime)
+      adjustment = -m_frameTime;
+    else
+      adjustment = 0;
   }
-  else
-    return false;
+
+  if (adjustment == 0)
+    return 0;
+
+  Discontinuity(clock+adjustment, absolute);
+
+  CLog::Log(LOGDEBUG, "CDVDClock::ErrorAdjust - %s - error:%f, adjusted:%f",
+                      log, error, adjustment);
+  return adjustment;
 }
 
 void CDVDClock::Discontinuity(double clock, double absolute)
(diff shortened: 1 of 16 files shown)


FernetMenta:
@MilhouseVH could you test if this fixes the issue on your ION system?

xbmc/cores/VideoPlayer/DVDClock.cpp
```diff
@@ -39,6 +39,8 @@ CDVDClock::CDVDClock()
   m_systemAdjust = 0;
   m_speedAdjust = 0;
   m_startClock = 0;
+  m_vSyncAdjust = 0;
+  m_frameTime = 1/60 * DVD_TIME_BASE;
```
popcornmix: Integer division will always set m_frameTime to zero

popcornmix: Should it know about the actual framerate?

popcornmix:
Just an observation. 

We will never present a frame early, but may present a frame late (due to cpu being busy). The greater the ime we can tolerate being late the better (e.g. when rendering the first subtitle of a new font or bringing up OSD we'd prefer not to skip if possible).
Scheduling frames at mid-point of vsyncs allows 8ms of lateness (at 60Hz).
Scheduling frames just after vsync allows 16ms of lateness (at 60Hz).

xbmc/cores/VideoPlayer/DVDClock.cpp
FernetMenta: oh, thanks. yes it should know about frametime because DISCONT corrections by audio are only allowed in intervals of frametime

FernetMenta:
> We will never present a frame early,

this is not correct on all platforms. if drivers buffer internally, you have to take case that frames are not presented early.

> Scheduling frames at mid-point of vsyncs allows 8ms of lateness (at 60Hz).

Frames are not scheduled at mid-point. It does not matter if lateness is 8ms or 16ms. Either we are late by an entire frame or not. If we are late by a frame the system  allows 10 frames to recover until player will request a drop.

FernetMenta:
I added another commit that simplifies lateness detection. Lateness is calculated in frames now.

MilhouseVH:
Thanks @FernetMenta, it does fix the issue, mostly. Playback is now looking normal, although I am seeing occasional frame skips and/or drops - perhaps 1 skip or drop every 3-4 minutes (23.976 h264 with "Adjust refresh rate" enabled). The VsyncOff value is increasing slowly (0.2-0.3-ish every 1-2 seconds) until it seems to reach some sort of limit and a skip or drop occurs, then it resets (or even starts to count down).

FernetMenta:
@MilhouseVH increasing VSyncOff means that video fps does not exactly match refresh rate. You know that with a custom modeline NVidia systems switch to 23.971 instead of 23.976, don't you? Did you try with "sync playback to display"?

fritsch:
Please post /usr/lib/kodi/kodi-xrandr | pastebinit please - here we will
see if you have the correct nvidia Modelines, which are _still_ after 5+
years to be set manually :-(

2016-04-16 8:51 GMT+02:00 Rainer Hochecker <notifications@github.com>:

> @MilhouseVH <https://github.com/MilhouseVH> increasing VSyncOff means
> that video fps does not exactly match refresh rate. You know that with a
> custom modeline NVidia systems switch to 23.971 instead of 23.976, don't
> you? Did you try with "sync playback to display"?
>
> —
> <https://github.com/xbmc/xbmc/pull/9629#issuecomment-210748820>
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

MilhouseVH:
> @MilhouseVH increasing VSyncOff means that video fps does not exactly match refresh rate. You know that with a custom modeline NVidia systems switch to 23.971 instead of 23.976, don't you?

Ummm... no I didn't actually! I've never knowingly needed it until the changes in these recent builds after #0409.

> Did you try with "sync playback to display"?

This is with "Sync playback to display" and "Adjust display refresh rate" both enabled:

s1

then a few minutes later:

s2

> Please post /usr/lib/kodi/kodi-xrandr | pastebinit please - here we will see if you have the correct nvidia Modelines, which are _still_ after 5+ years to be set manually :-(

`/usr/lib/kodi/kodi-xrandr`: http://sprunge.us/OGNK

I'm using a custom xorg.conf, mainly to allow the Revo3700 to find the display when it powers on before the Amp (Onkyo TX-NR828):

`/storage/.config/xorg.conf`: http://sprunge.us/WRHb
`/storage/.config/edid.bin`: http://sprunge.us/giYN (dumped using RPi `edidparser`)

Thanks both.

fritsch:
See:
>    <mode id="0x1c9" name="1920x1080" w="1920" h="1080" hz="24.00000" current="false" preferred="false"/>
    <mode id="0x1ca" name="1920x1080" w="1920" h="1080" hz="23.97091" current="true" preferred="false"/>

You miss the proper 23.9768 mode, e.g.:
>     ModeLine       "1920x1080_24" 74.250 1920 2558 2602 2750 1080 1084 1089 1125 +hsync +vsync
    ModeLine       "1920x1080_23.976" 74.175 1920 2558 2602 2750 1080 1084 1089 1125 +hsync +vsync

See the OpenELEC wiki for more information.

koenkooi:
As a data point: disabling vsync on Jarvis fixed the staggering amount of framedrops when using matching refreshrates and fps.

See http://forum.kodi.tv/showthread.php?tid=227199&pid=2312398#pid2312398 for more info.

fritsch:
That's not a data point, that's just insane forcing TearFree ... and Jarvis has absolutely nothing (zero) to do with that part of the code, which is completely rewritten.

MilhouseVH:
Thanks @fritsch.

After adding the missing modelines, it still wasn't working - turns out the edid.bin from my Onkyo TX-NR828 amp is faulty:
```
[518502.451] (WW) NVIDIA(0): No valid modes for "DFP-0:1920x1080_23.976"; removing.
```
So I'm now using an edid.bin from a [Marantz amp](https://www.dropbox.com/s/qngzoiuegapj615/edid-samsung-marantz.bin?dl=0) which seems to be working better now.

OE Wiki page for reference: http://wiki.openelec.tv/index.php?title=Configuring_a_Custom_xorg.conf

FernetMenta:
jenkins build this please

xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp
```diff
@@ -1370,12 +1378,25 @@ void CRenderManager::DiscardBuffer()
...
+
+void CRenderManager::CheckEnableClockSync()
+{
+  if (fabs(m_fps - g_graphicsContext.GetFPS()) < 0.01)
+  {
+    m_clockSync.m_enabled = true;
+  }
+  else
+  {
+    m_clockSync.m_enabled = true;
```
popcornmix: Should this be false?

FernetMenta: oops. thanks

PR #9631, opened 2016-04-16
[linux] support for simple xselection pasting

sergiomb2:
From: Matthias Kortstiege mkortstiege@kodi.tv
Date: Sat, 4 Jul 2015 16:51:06 +0200
Subject: [PATCH] [linux] support for simple xselection pasting

in http://forum.kodi.tv/showthread.php?tid=231227 says to patch kodi with 
https://github.com/mkortstiege/xbmc/commit/62e56b7e8f0d944439a3f60eaa6a2bc41cdce67b 

I test it works and no regressions noted . Kodi don't have copy and paste working on Linux !

### DIFF
--- a/xbmc/windowing/X11/WinSystemX11.cpp
+++ b/xbmc/windowing/X11/WinSystemX11.cpp
@@ -48,6 +48,8 @@
 
 using namespace KODI::MESSAGING;
 
+#include <limits.h>
+
 #define EGL_NO_CONFIG (EGLConfig)0
 
 CWinSystemX11::CWinSystemX11() : CWinSystemBase()
@@ -1068,6 +1070,62 @@ void CWinSystemX11::UpdateCrtc()
 
   m_crtc = g_xrandr.GetCrtc(posx+winattr.width/2, posy+winattr.height/2, fps);
   g_graphicsContext.SetFPS(fps);
+}
+
+std::string CWinSystemX11::GetClipboardText(void)
+{
+  if (!m_dpy)
+     return "";
+
+  Atom utf8_string = XInternAtom(m_dpy, "UTF8_STRING", False);
+  if (utf8_string == None)
+    return "";
+
+  Atom clipboard = XInternAtom(m_dpy, "CLIPBOARD", False);
+  if (clipboard == None)
+    return "";
+
+  // if no selection or there is no owner for the selection (ancient X ?!)
+  if (XGetSelectionOwner(m_dpy, clipboard) == None)
+    return "";
+
+  std::string result = "";
+  XConvertSelection(m_dpy, clipboard, utf8_string, None, m_glWindow, CurrentTime);
+//  Wait for clipboard event (SelectionNotify) sync, code copied from
+//  https://github.com/vinniefalco/SimpleDJ/blob/master/Extern/JUCE/modules/juce_gui_basics/native/juce_linux_Clipboard.cpp
+//  clipboard content requesting is inherently slow on x11, it
+//  often takes 50ms or more so...
+  int count = 20; // will wait at most for 200 ms
+  while (--count >= 0)
+  {
+    CLog::Log(LOGNOTICE, "GetClipboardText calls MessagePump count = %d ", count);
+    if (CWinEventsX11Imp::MessagePump())
+      break;
+    Sleep(10);
+  }
+
+  Atom real_type = None;
+  int real_format = 0;
+  for (unsigned long offset = 0;;)
+  {
+    unsigned long items_read = 0;
+    unsigned long items_left = 0;
+    unsigned char *data;
+
+    if (XGetWindowProperty(m_dpy, m_glWindow, utf8_string, offset, INT_MAX, 
+      False, AnyPropertyType, &real_type, &real_format, &items_read,
+      &items_left, &data) == Success) 
+    {
+      result.append(reinterpret_cast<const char*>(data), items_read);
+      offset += items_read;
+      XFree(data);
+    } else 
+      break;
+    if (!items_left)
+      break;
+  }
+
+  return result;
 }
 
 #endif
--- a/xbmc/windowing/WinEventsX11.cpp
+++ b/xbmc/windowing/WinEventsX11.cpp
@@ -344,6 +344,10 @@ bool CWinEventsX11Imp::MessagePump()
 
     switch (xevent.type)
     {
+      case SelectionNotify:
+        ret = true;
+        break;
+
       case MapNotify:
       {
         g_application.SetRenderGUI(true);
--- a/xbmc/windowing/X11/WinSystemX11.h
+++ b/xbmc/windowing/X11/WinSystemX11.h
@@ -68,6 +68,7 @@ public:
   bool IsCurrentOutput(std::string output);
   void RecreateWindow();
   int GetCrtc() { return m_crtc; }
+  std::string GetClipboardText(void);
 
 protected:
   virtual bool SetWindow(int width, int height, bool fullscreen, const std::string &output, int *winstate = NULL) = 0;


FernetMenta:
X11 does already poll XNextEvent. Makes no sense to do the same at a different location. I also has a main window, no need to create an additional one.

sergiomb2:
man I just want paste clipboard to kodi , it is very useful and this patch works , I will make a better patch , and suggestions on coding are very welcome . 
Thanks.

FernetMenta:
check out https://github.com/xbmc/xbmc/blob/master/xbmc/windowing/WinEventsX11.cpp#L322
there is no need for an addition window

sergiomb2:
patch updated , I'm testing it now ,

FernetMenta:
hmm, you seem to have ignored my comments: no additional window, no additional XNextEvent

sergiomb2:
Hi, I done another patch https://github.com/sergiomb2/xbmc/commit/cb0a44a645c8d2dd6d9471eeed96140f27a69f64  not yet the final patch 

where should I put this 2 lines : 

    Atom utf8_string = XInternAtom(m_dpy, "UTF8_STRING", False);
    XConvertSelection(m_dpy, XA_PRIMARY, utf8_string, None, m_glWindow, CurrentTime);

xbmc/windowing/X11/WinSystemX11.cpp
```diff
@@ -521,6 +523,11 @@ void CWinSystemX11::NotifyAppFocusChange(bool bGaining)
   }
   if (!bGaining)
     m_bIgnoreNextFocusMessage = false;
+  else
+  {
+    Atom utf8_string = XInternAtom(m_dpy, "UTF8_STRING", False);
+    XConvertSelection(m_dpy, XA_PRIMARY, utf8_string, None, m_glWindow, CurrentTime);
```
FernetMenta: I don't think we should blindly try to convert anything on AppFocucChange. Further I think we want to copy from clipboard, not primary. We may also copy from Kodi to Kodi.

FernetMenta:
check out what SDL does: http://www.libsdl.org/tmp/SDL/src/video/x11/SDL_x11clipboard.c
this can be easily ported to Kodi.

sergiomb2:
I will try change and test XA_PRIMARY to XA_CLIPBOARD. 
The feature that is *very* important to have is "simple xselection pasting", have full feature of copy and paste working seems much more complicated and I haven't time or acknowledgment to do that , also don't think even that is practical, after analyze one textbox in kodi under Linux I see that can't select the text in the box , so I also can't copy or cut the text , but paste text from clipboard is possible and the unique feature that I like to have and might be useful for kodi users , so the question is, one simple patch like this can be upstreamed or I just use it in my packages ?

sergiomb2:
Hello 
I take a look in projects: 
https://github.com/astrand/xclip
https://github.com/edrosten/x_clipboard
http://www.vergenet.net/~conrad/software/xsel
specs:
https://specifications.freedesktop.org/clipboards-spec/clipboards-latest.txt
and some code:
https://github.com/ccxvii/snippets/blob/master/x11clipboard.c
https://github.com/myint/pygame/blob/master/src/scrap_x11.c
https://git.enlightenment.org/apps/eterm.git/tree/src/startup.c

I found one method (EnableTextInput) where I could put the XConvertSelection , I also could read just CLIPBOARD instead XA_PRIMARY, so now for me my patch looks good to be applied, all code seems in right place. 
And it is working applied on Jarvis 16.1 .

FernetMenta:
This code may work, somehow, but it is rather hacky. We had too many of those "it works somehow hacks" in the past that made architecture decay. There are many locations in this PR that need rework. 
I already told you that we won't blindly call XConvertSelection. Now you moved it from one location that was wrong to another that is wrong too. Why don't you follow my advice any look into SDL code?

sergiomb2:
It looks good to me have XConvertSelection in EnableTextInput ... , today I understand better SDL code (after yesterday read much documentation) , SDL code do all in same function , after XConvertSelection, do something that I don't know how to do, which is : 

        waitStart = SDL_GetTicks();
        videodata->selection_waiting = SDL_TRUE;
        while (videodata->selection_waiting) {

SDL code waits for selection ... how we may do this ( I will google it) .

About "it works somehow hacks" , I understand you very well , but know I don't think my patch is a hack , when X enables a text input we call XConvertSelection , looked good to me . 
Thanks .

FernetMenta:
> About "it works somehow hacks" , I understand you very well , but know I don't think my patch is a hack , when X enables a text input we call XConvertSelection 

g_Windowing.EnableTextInput is called by Kod's edit control when it gets focus. Nothing related to X.

FernetMenta:
> SDL code waits for selection ... how we may do this ( I will google it) .

exactly the same way. Call MessagePump for a given time.

sergiomb2:
> >     SDL code waits for selection ... how we may do this ( I will google it) .
> 
> exactly the same way. Call MessagePump for a given time.

Done ! , now I think it it by the book. 

Thanks.

FernetMenta:
why don't you call MessagePump as I suggested?

sergiomb2:
> why don't you call MessagePump as I suggested?
OK , Using MessagePump it is the correct way , what do you think now ?

xbmc/windowing/X11/WinSystemX11.cpp
```diff
@@ -1070,4 +1072,51 @@ void CWinSystemX11::UpdateCrtc()
...
+  std::string result;
+  Atom utf8_string = XInternAtom(m_dpy, "UTF8_STRING", False);
+  Atom clipboard = XInternAtom(m_dpy, "CLIPBOARD", False);
+  XConvertSelection(m_dpy, clipboard, utf8_string, None, m_glWindow, CurrentTime);
+//  Wait for clipboard event (SelectionNotify) sync, code copied from
+//  https://github.com/vinniefalco/SimpleDJ/blob/master/Extern/JUCE/modules/juce_gui_basics/native/juce_linux_Clipboard.cpp
+//  clipboard content requesting is inherently slow on x11, it
+//  often takes 50ms or more so...
+  int count = 20; // will wait at most for 200 ms
+  while (--count >= 0)
```
FernetMenta: check SDL, this is only required if we are not the owner and owner != none

xbmc/windowing/X11/WinSystemX11.cpp
```diff
@@ -1070,4 +1072,51 @@ void CWinSystemX11::UpdateCrtc()
...
 }
 
+std::string CWinSystemX11::GetClipboardText(void)
+{
+  if (!m_dpy)
+     return "";
+
+  std::string result;
+  Atom utf8_string = XInternAtom(m_dpy, "UTF8_STRING", False);
+  Atom clipboard = XInternAtom(m_dpy, "CLIPBOARD", False);
```
FernetMenta: this requires check, if atom is available or the follow call results in error

xbmc/windowing/X11/WinSystemX11.cpp
```diff
@@ -1070,4 +1072,51 @@ void CWinSystemX11::UpdateCrtc()
...
+//  often takes 50ms or more so...
+  int count = 20; // will wait at most for 200 ms
+  while (--count >= 0)
+  {
+    if (CWinEventsX11Imp::MessagePump())
+      break;
+    Sleep(10);
+  }
+
+  for(unsigned long offset = 0;;)
```
FernetMenta: why do you think this loop is required? SDL code has none

xbmc/windowing/X11/WinSystemX11.cpp
```diff
@@ -1070,4 +1072,51 @@ void CWinSystemX11::UpdateCrtc()
...
+  {
+    Atom real_type = None;
+    unsigned char *data;
+    unsigned long items_read, items_left = 0;
+    int real_format = 0;
+
+    Atom utf8_string = XInternAtom(m_dpy, "UTF8_STRING", False);
```
FernetMenta: this atom may not be available, needs check

xbmc/windowing/X11/WinSystemX11.cpp
```diff
@@ -1070,4 +1072,51 @@ void CWinSystemX11::UpdateCrtc()
...
+  Atom clipboard = XInternAtom(m_dpy, "CLIPBOARD", False);
+  XConvertSelection(m_dpy, clipboard, utf8_string, None, m_glWindow, CurrentTime);
+//  Wait for clipboard event (SelectionNotify) sync, code copied from
+//  https://github.com/vinniefalco/SimpleDJ/blob/master/Extern/JUCE/modules/juce_gui_basics/native/juce_linux_Clipboard.cpp
+//  clipboard content requesting is inherently slow on x11, it
+//  often takes 50ms or more so...
+  int count = 20; // will wait at most for 200 ms
+  while (--count >= 0)
+  {
+    if (CWinEventsX11Imp::MessagePump())
```
FernetMenta: of course you need to adapt MessagePump that is listens to the event triggered by XConvertSelection

sergiomb2:
https://www.jwz.org/doc/x-cut-and-paste.html
in  2002 Cut buffers are officially obsolete.

I don't know how translate :  ` selection = X11_GetSDLCutBufferClipboardType(display);` and 
selection = X11_XInternAtom(display, "SDL_SELECTION", False);

I can add checks to atom like this:
```
   Atom utf8_string = XInternAtom(m_dpy, "UTF8_STRING", False);
+  if (utf8_string == None) {
+    return "";
+  }
   Atom clipboard = XInternAtom(m_dpy, "CLIPBOARD", False);
+  if (clipboard == None) {
+    return "";
+  }
```

owner = X11_XGetSelectionOwner(display, XA_CLIPBOARD);
if owner is null instead fall back to X11 of 2000 or return ""  (which means we don not support it)
if owner is window I don't get it, X11_GetSDLCutBufferClipboardType !? , CutBuffer in 2002 was already obsoleted.
if not set owner = window and do X11_XConvertSelection, so for me we we have atoms we do just this case. 

> 
of course you need to adapt MessagePump that is listens to the event triggered by XConvertSelection

Also I'm not seeing what we need , MessagePump returns true or false , if we process one event ...

FernetMenta:
you don't want to wait for any event but for a specific one. hence you need to modify MessagePump to signal this specific event. while waiting for this event MessagePump will deliver other events to application

sergiomb2:
Keep it simple, we just need call MessagePump once , when MessagePump returns all events are processed, I think. 
if XGetWindowProperty returns Success we can copy result and loop until retrieve all content .
if XGetSelectionOwner returns null  if no selection or there is no owner for the selection (ancient X ?!)
Now patch looks that works with synchronized code and we verify if X11 support Atoms and other stuff like XGetSelectionOwner  , if not we simply return an empty string like happens before this patch.
Best regards.

xbmc/windowing/X11/WinSystemX11.cpp
```diff
@@ -1070,4 +1072,49 @@ void CWinSystemX11::UpdateCrtc()
...
+
+  // if no selection or there is no owner for the selection (ancient X ?!)
+  if (XGetSelectionOwner(m_dpy, clipboard) == None)
+    return "";
+
+  std::string result = "";
+  XConvertSelection(m_dpy, clipboard, utf8_string, None, m_glWindow, CurrentTime);
+  CWinEventsX11Imp::MessagePump();
+
+  for(unsigned long offset = 0;;)
```
FernetMenta: no chance for this loop, as I said: wait for the specific event for a maximum time.
and again: follow the example of SDL

FernetMenta:
keep is simple is good, but first it needs to be correct.

sergiomb2:
> we just need call MessagePump once , when MessagePump returns all events are processed, I think.  

After test it , I saw that we need wait for one MessagePump positive.  
If by a race condition, which is process one event, empty the event queue and SelectionNotify was not called, selection will not be update , but don't do any harm . 
About X11_XGetSelectionOwner and when owner == window , for what I understood this is just applicable on SDL or other cases where windows have selection saved as one property  , which is not applicable to X11.

xbmc/windowing/X11/WinSystemX11.cpp
sergiomb2: hum , this is the correct way, XGetWindowProperty, may send just partial content

FernetMenta: oh :) this is really a funny style of a loop
if you did:
while (itemsLeft)
it would be much clearer

and the used parameters should be outside of the loop

FernetMenta:
You don't get around modifying MessagePump. It does not return true an any event and you current code just times out every time.

sergiomb2:
Now patch verify: 

1. if X11 not support Atoms and other stuff like XGetSelectionOwner, we simply return an empty string like happens before this patch.
2. if XGetWindowProperty returns Success, we copy result and loop until retrieve all content .
3. if XGetSelectionOwner returns null, means: no selection or there is no owner for the selection (ancient X ?!), we return an empty string.
4. We need wait for MessagePump return true , I add a log line to see and count how many times MessagePump is called [1] 

Adding catch SelectionNotify and return true like this : 

      case SelectionNotify:
        ret = true;
        break;


Without catch SelectionNotify, MessagePump was called 10 or 11 times , with SelectionNotify catch, MessagePump is called only 2 times ! 
I personally would like *not* change CWinEventsX11Imp::MessagePump, because this function can be called in other situations that I don't control. But set return true when SelectionNotify is called, seems to me, harmless, this is the only doubt I have now , I can't find a way to have the code in a  synchronized way.
Anyway, I checked where MessagePump is called in kodi and return code is always ignored. so seems be safe the change in MessagePump function.

Best regards.
[1] CLog::Log(LOGNOTICE, "GetClipboardText calls MessagePump count = %d ", count);

xbmc/windowing/X11/WinSystemX11.cpp
sergiomb2: About X11_XGetSelectionOwner and when owner == window , for what I understood this is just applicable on SDL or other cases where windows have selection saved as one SDL specific property , which, looks to me, is not applicable to X11.

FernetMenta:
Getting closer :) but I see no need for guessing ind praying that things go right now and in future. We can set an event/flag on SelectionNotify and wait on that event in WinSystemX11

PR #9633, opened 2016-04-16
fixed: Compile warnings

arnova:


arnova:
@FernetMenta : Your button

jenkins build this please

fritsch:
@arnova size_t would be correct for the comparison with deque's size() right?

arnova:
@fritsch: Yes didn't realise that. Can change it if you like but all it will ever become is either 1 or 5 so it didn't seem that important.

fritsch:
@arnova if you are already touch it - please do if you can.

arnova:
@fritsch : Updated

fritsch:
Thanks much. Push at will :-)

arnova:
jenkins build and merge

PR #9636, opened 2016-04-16
[PVR] EPG data import performance tuning.

ksooo:
This patch increases the performance of (initial) epg data import by refraining from expensive eager evaluation of the epg tag that might belong to a timer tag.

Of course I double checked that the epg tag will be created later when it is actually needed.

@xhaggi @Jalle19 mind taking a look.

### DIFF
--- a/xbmc/pvr/timers/PVRTimerInfoTag.cpp
+++ b/xbmc/pvr/timers/PVRTimerInfoTag.cpp
@@ -976,36 +976,35 @@ std::string CPVRTimerInfoTag::GetDeletedNotificationText() const
   return StringUtils::Format("%s: '%s'", g_localizeStrings.Get(stringID).c_str(), m_strTitle.c_str());
 }
 
-CEpgInfoTagPtr CPVRTimerInfoTag::GetEpgInfoTag(void) const
+CEpgInfoTagPtr CPVRTimerInfoTag::GetEpgInfoTag(bool bCreate /* = true */) const
 {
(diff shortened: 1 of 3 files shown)


xbmc/pvr/timers/PVRTimerInfoTag.cpp
```diff
@@ -976,36 +976,35 @@ std::string CPVRTimerInfoTag::GetDeletedNotificationText() const
...
       const CEpgPtr epg(channel->GetEPG());
       if (epg)
       {
-        if (m_iEpgUid != EPG_TAG_INVALID_UID)
-        {
-          m_epgTag = epg->GetTagByBroadcastId(m_iEpgUid);
-        }
-        else if (!m_bStartAnyTime && !m_bEndAnyTime)
+        CSingleLock lock(m_critSection);
+        if (!m_epgTag)
```
Jalle19: This check seems unnecessary, it's already checked earlier.

ksooo: Hehe we had that already. This is the double checked locking pattern. First check needs to be done without muted locked. Between first and second check (this time with the lock) the member value might have changed.

ksooo:
jenkins build this please

ksooo:
If nobody objects I will merge this PR tomorrow

ksooo:
AppVeyor seems to be broken, not related to this PR.

PR #9637, opened 2016-04-17
[IMX/RefClock] refactor refclock IMX core engine. 

mk01:
create global provider, adapt vsyncIMX

### DIFF
--- a/xbmc/linux/Makefile.in
+++ b/xbmc/linux/Makefile.in
@@ -15,6 +15,8 @@ SRCS += XTimeUtils.cpp
 
 SRCS += RBP.cpp
 
+SRCS += $(wildcard @CORE_SYSTEM_VARIANT@/*.cpp)
+
 ifeq (@USE_OMXLIB@,1)
 SRCS += OMXClock.cpp
 SRCS += OMXCore.cpp
--- a/xbmc/linux/imx/CMakeLists.txt
+++ b/xbmc/linux/imx/CMakeLists.txt
@@ -1,0 +1,6 @@
+set(SOURCES IMX.cpp
+            GlobalsIMX.cpp)
+
+set(HEADERS IMX.h)
+
+core_add_library(linuxsupport_imx)
--- a/xbmc/linux/imx/GlobalsIMX.cpp
+++ b/xbmc/linux/imx/GlobalsIMX.cpp
@@ -1,6 +1,5 @@
-#pragma once
 /*
- *      Copyright (C) 2005-2014 Team XBMC
+ *      Copyright (C) 2005-2013 Team XBMC
  *      http://xbmc.org
  *
  *  This Program is free software; you can redistribute it and/or modify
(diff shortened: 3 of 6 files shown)


mk01:
@FernetMenta 
can you please recheck and merge?

this PR alone has neutrality - so functionality is not added nor removed.
(this is rework as discussed here https://github.com/xbmc/xbmc/pull/9253)

FernetMenta:
I know that other platforms like Pi don't do better but we need to start somewhere. I makes zero sense adding IBM.cpp to all platforms with a ifdef in the code.
Please adapt cmake and makefile to only include this specific files for IMX. Using systemGlobals is also a bad example given by RBP. The remaining lifetime of SystemGlobals is very limited.

FernetMenta:
@mk01 I would create a separate subfolder IMX in adapt make and cmake to use source files in this folder when compiling for this platform. Similar to https://github.com/xbmc/xbmc/tree/master/xbmc/cores/VideoPlayer/Process (but no overrides in this case)

mk01:
@FernetMenta 
ok, will look into it

mk01:
@FernetMenta 
something like this ?

FernetMenta:
@mk01 looks good as far as I can judge. @wsnipex @fetzerch could you please have a look at cmake, make

xbmc/linux/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+file(GLOB VARIANTS_CPP ${CORE_SYSTEM_VARIANT}/*.cpp)
+file(GLOB VARIANTS_H ${CORE_SYSTEM_VARIANT}/*.h)
```
fetzerch: There's one problem with globbing of source files: When you add/remove a file but don't touch any CMake files, then CMake doesn't know that it needs to rerun and you'll get some difficult to understand compile/link errors. Also if I'm not mistaken, CORE_SYSTEM_VARIANT is currently not userd/set for CMake.

@wsnipex, @notspiff: Do you already have a solution for those variants in mind? It should probably be an own lib that gets added through the treedata structure, right? Then we have to decide if we want to handle it like rbpi (with it's own platform config in project/cmake/platform) or we add variant support to linux.

akva2: globbing is never okay. i haven't thought much about this variant stuff, i thought it was all covered by the treedata construct.
if there are sub-variants needed for linux, we should add that there imo. ie a dedicated directory for variants and conditionally processing of those text files using something ala CORE_SYSTEM_VARIANT.

mk01:
@fetzerch @akva2 

how can this commit be changed - that it can go in?

xbmc/linux/CMakeLists.txt
fetzerch: To not hold this back, please create a `CMakeLists.txt` in the imx directory with the following content. Then I can do the necessary changes to the build system independently when I find the time. 
```
set(SOURCES IMX.cpp
            GlobalsIMX.cpp)

set(HEADERS IMX.h)

core_add_library(linuxsupport_imx)
```

mk01:
@fetzerch , @FernetMenta 
thanks. changed accordingly.

FernetMenta:
looks good to me. @fetzerch @wsnipex ok?

fetzerch:
CMake part looks good to me, not much experience with autotools though.

FernetMenta:
This change is required for fixing broken IMX. IMO this can go in outside merge window.
Any objections?

fritsch:
Perfectly fine. Thanks @mk01 

let's see if we broke something else: 

jenkins build this please

PR #9639, opened 2016-04-17
fixed: Wrong parent path comparisons + cleanup

arnova:
There are several ways we use to look for a certain base/parent path. Not all of these methods are correct. When we use StringUtils::StartsWith() and URIUtils:PathStarts(), one can only reliably check for a certain parent path if the parent path has a trailing slash, else when looking for e.g. "/parent", "/parent2" will also match which is obviously wrong. The only exception here are our  internal (special/pvr/addon) paths which only have a fixed non-overlapping set of parent paths, for those comparisons we can safely use StartsWith(). Furthermore we have URIUtils::IsInpath() and URIUtils::PathStarts() which almost do the same thing, so I have merged those into the (imho) better named and fixed "PathHasParent()".

### DIFF
--- a/xbmc/utils/URIUtils.cpp
+++ b/xbmc/utils/URIUtils.cpp
@@ -36,14 +36,6 @@
 
 using namespace XFILE;
 
-bool URIUtils::IsInPath(const std::string &uri, const std::string &baseURI)
-{
-  std::string uriPath = CSpecialProtocol::TranslatePath(uri);
-  std::string basePath = CSpecialProtocol::TranslatePath(baseURI);
-
-  return !basePath.empty() && StringUtils::StartsWith(uriPath, basePath);
-}
-
 /* returns filename extension including period of filename */
 std::string URIUtils::GetExtension(const CURL& url)
 {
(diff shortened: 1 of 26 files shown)


arnova:
@Montellese / @mkortstiege : Please review

jenkins build this please

xbmc/utils/URIUtils.cpp
```diff
@@ -517,32 +509,57 @@ bool URIUtils::IsProtocol(const std::string& url, const std::string &type)
   return StringUtils::StartsWithNoCase(url, type + "://");
 }
 
-bool URIUtils::PathStarts(const std::string& url, const char *start)
+bool URIUtils::PathHasParent(const std::string& path, const std::string& parent, bool translate /* = false */)
```
Montellese: Just make `path` and `parent` values instead of const references because you need to copy them into internal strings anyway.

Montellese:
Looks good.

xbmc/utils/URIUtils.cpp
arnova: @Montellese : Initially I had those as copies (not refs), but some how the "path = CSpecialProtocol::TranslatePath(path)"/"parent = CSpecialProtocol::TranslatePath(parent)" caused crashes. That's why I made it the way it is now. You understand why that doesn't work, one should be able to pass something by ref and at the same return into that same variable the result or is that illegal?

Montellese: That shouldn't crash at all.

arnova: Strange, just did a full recompile and now the crash is gone :-) Must be a glitch on my system.

arnova:
jenkins build this please

PR #9641, opened 2016-04-17
CDebugRenderer: fix debug info overlays alignment

wiromare:
just a cosmetic thing for the debug info lines alignment that have a strange behavior when kodi is resized while is windowed

before the the PR

after PR

@FernetMenta


FernetMenta:
thanks!

PR #9642, opened 2016-04-17
[addons] database improvements

tamland:
see commits

### DIFF
--- a/xbmc/addons/AddonDatabase.cpp
+++ b/xbmc/addons/AddonDatabase.cpp
@@ -244,7 +244,11 @@ void CAddonDatabase::SyncInstalled(const std::set<std::string>& ids, const std::
           "VALUES('%s', 0, '%s')", id.c_str(), now.c_str()));
 
     for (const auto& id : removed)
+    {
       m_pDS->exec(PrepareSQL("DELETE FROM installed WHERE addonID='%s'", id.c_str()));
+      RemoveAddonFromBlacklist(id);
+      DeleteRepository(id);
+    }
 
     for (const auto& id : enabled)
       m_pDS->exec(PrepareSQL("UPDATE installed SET enabled=1 WHERE addonID='%s'", id.c_str()));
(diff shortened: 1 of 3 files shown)


tamland:
@MilhouseVH Should fix your repository issues fyi. Mind giving it a test?

MilhouseVH:
@tamland - many thanks for this! As far as I can tell it appears to be working as intended - after upgrading from an OE build to LibreELEC, the now non-existent OE repository is being dropped from the Addons23 database, which is perfect. Thanks again.

DaveTBlake:
Where you have added "EXISTS (SELECT * FROM..." subqueries would it not be more efficient to use a field name or 1 rather than "*"?

Razzeee:
Yes, but more important would be a limit 1

Montellese:
You don't need a limit 1 if you use `1` as the field. Any decent SQL optimizer will realise that it can only return a single result and will stop as soon as it finds one.

tamland:
Any sane optimizer should be able to eliminate everything from an `EXISTS` clause. Besides, this is the least expensive query in that method, taking ~0ms;)

DaveTBlake:
Does SQLite have a "sane" optimiser? I know the guys on the SQLite mailing list recommend not using SELECT *, so I have to wonder.  I would agree with the LIMIT 1 too, but I think that MySQL has a problem with limits in subqueries.

Surly, allowing for known product limits, isn't it a matter of principle to write good SQL regardless how fast this particular query or how clever the optimser?

tamland:
addonID is unique so any limit is redundant. it will only be 0 or 1.

>Surly, allowing for known product limits, isn't it a matter of principle to write good SQL regardless how fast this particular query or how clever the optimser?

That goes for clever, premature, hand optimized sql as well. If you have any suggestions that makes the query simpler and intent clearer, shoot.;)

stefansaraev:
jenkins build this please

PR #9643, opened 2016-04-17
[addons] ask for version on install

tamland:
This changes 'install' to ask for version/reposiotry when there are multiple available; similar to the current update dialog.

This pr is branched off #9642


tamland:
jenkins build this please

PR #9644, opened 2016-04-17
[win32][dependencies] New build of openssl that's not linked against debug vcruntime

Paxxi:
This should fix the libcurl issues seen in latest nightlies


Paxxi:
Oh, I was thinking of testing first but nightly should work as well for that

MartijnKaijser:
Worked for me :)

da3dsoul:
That's why it hasn't been working. I was confused as to why it was throwing VC errors when I know I have it. Thanks a ton

PR #9645, opened 2016-04-17
Implemented Base64Encoding for Addon Interface

ghost:


akva2:
Not against the code per se, but for general consideration.  this callback creep is what i tried to stop by introducing kodiplatform. This is not a callback, this is kodi acting like a generic toolbox.

Montellese:
I agree with @akva2. Anything that doesn't depend on Kodi core internal state i.e. simple utility stuff should be part of `kodiplatform`. And this is a perfect example for that.

Paxxi:
I also agree with @akva2. I did look at yanking out most of the utils folder we have to make it usable by addons, doable but some work.

mapfau:
To be able to use CURL POSTDATA, you need in the current implementation B64 encoding.
I think it MUST be available through the interface, it's not very nice to users to let them all copy the B64 code inside their addon.

If you support addon functionality, you should also support a toolkit.

Or did I understand your objections wrong?

I have no problem copying my personal B64 code into pvr.august, but I think its wrong

akva2:
yes, you misunderstood. kodiplatform is the toolbox. kodi should not be the toolbox. and i didn't object to this code, i totally understand it is needed.

mapfau:
Then pls. let me know where it should be. A complete new container? Or is there already an exiting one where string ops should be in?

I'm not deep enough in what to place where, I need your help for this.

akva2:
i don't think you are the correct person to address my concerns then. but https://github.com/xbmc/kodi-platform/

akva2:
and please read my comment in the pr referenced further up. it should explain things.

mapfau:
:-) No, I'm for sure not the correct person for your concerns in this area.

I read the PR above, well witten and give alot of insight whats planned.
Thanx for the link!

kodi-addon-sdk is imo exactly whats missing. I'll close this PR here.
Because pvr.augustv is not my main interest, I'll get it simply run (with copying b64).

mapfau:
Last but not least: Why are so many people of you work on a similar solution?
Should be better to sit an hour together and agree to an concept and start with exactly one wich is fine fo all of you.

PR #9646, opened 2016-04-18
ffmpeg: reduce log spam

FernetMenta:
see title

@afedchin this should fix the issue


afedchin:
thanks. I can confirm that this fixes the issue.

FernetMenta:
jenkins build and merge

PR #9647, opened 2016-04-19
[gui] Use SelectDialog for SortMethod selection

phil65:
screenshot005

This exchanges the default action of the sortmethod button in MediaWindows from "choose next" to "open selectdialog". Feels more convenient esp. when the amount of sortmethods is high. There are content types with more than 10 sortmethods ("songs" has 12 for example), handling those with a single togglebutton is quite painful.

What might be worth considering is to choose between one of these two methods automatically based on the amount of available viewtypes. (let´s say SelectDialog for >= 3 sortmethods, old behaviour for less).

@ronie @da-anda @BigNoid for opinions and @tamland for code.    

EDIT: header is only wrong in screenshot, PR contains correct one.

### DIFF
--- a/xbmc/view/GUIViewState.cpp
+++ b/xbmc/view/GUIViewState.cpp
@@ -33,6 +33,7 @@
 #include "GUIPassword.h"
 #include "ViewDatabase.h"
 #include "AutoSwitch.h"
+#include "dialogs/GUIDialogSelect.h"
 #include "guilib/GUIWindowManager.h"
 #include "addons/Addon.h"
 #include "addons/AddonManager.h"
@@ -342,6 +343,32 @@ void CGUIViewState::SetSortMethod(SortBy sortBy, SortOrder sortOrder /* = SortOr
 void CGUIViewState::SetSortMethod(SortDescription sortDescription)
 {
   return SetSortMethod(sortDescription.sortBy, sortDescription.sortOrder);
+}
+
+bool CGUIViewState::ChooseSortMethod()
+{
+  
+  CGUIDialogSelect *dialog = static_cast<CGUIDialogSelect *>(g_windowManager.GetWindow(WINDOW_DIALOG_SELECT));
+  if (dialog)
+  {
+    dialog->Reset();
+    dialog->SetHeading(CVariant{ 32104 }); // Label "Sort by"
+    for (int i = 0; i < m_sortMethods.size(); ++i)
+    {
+      dialog->Add(g_localizeStrings.Get(m_sortMethods[i].m_buttonLabel));
+    }
+    dialog->SetSelected(m_currentSortMethod);
+    dialog->Open();
+    int newSelected = dialog->GetSelectedItem();
+    // check if selection has changed
+    if (!dialog->IsConfirmed() || newSelected < 0 || newSelected == m_currentSortMethod)
+      return false;
+
+    m_currentSortMethod = newSelected;
+
+    SaveViewState();
+  }
+  return true;
 }
 
 SortDescription CGUIViewState::SetNextSortMethod(int direction /* = 1 */)
--- a/xbmc/view/GUIViewState.cpp
+++ b/xbmc/view/GUIViewState.cpp
@@ -349,25 +349,21 @@ bool CGUIViewState::ChooseSortMethod()
 {
   
   CGUIDialogSelect *dialog = static_cast<CGUIDialogSelect *>(g_windowManager.GetWindow(WINDOW_DIALOG_SELECT));
-  if (dialog)
-  {
-    dialog->Reset();
-    dialog->SetHeading(CVariant{ 32104 }); // Label "Sort by"
-    for (int i = 0; i < m_sortMethods.size(); ++i)
-    {
-      dialog->Add(g_localizeStrings.Get(m_sortMethods[i].m_buttonLabel));
-    }
-    dialog->SetSelected(m_currentSortMethod);
-    dialog->Open();
-    int newSelected = dialog->GetSelectedItem();
-    // check if selection has changed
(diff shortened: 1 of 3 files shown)


BigNoid:
Nice improvement imo. I would say to also show the dialog if there are few choices. Its instant recognisable that there are few choices with selectdialog whereas it might take a few clicks to realise there's only a few choices with a selectbutton.

MartijnKaijser:
Yep, really nice

da-anda:
looks good. I'd also always show the dialog

ronie:
no objections from my end.

tamland:
Improvement for sure, but I always envisioned the side menu to be similar to the context menu where all select-like actions takes you to a new 'page' within the same menu.

xbmc/windows/GUIMediaWindow.cpp
```diff
@@ -282,7 +282,7 @@ bool CGUIMediaWindow::OnMessage(CGUIMessage& message)
       else if (iControl == CONTROL_BTNSORTBY) // sort by
       {
         if (m_guiState.get())
-          m_guiState->SetNextSortMethod();
+          m_guiState->ChooseSortMethod();
```
tamland: You don't use the result for anything. You should change this to return bool and only call UpdateFileList if the sort method was changed instead.

xbmc/view/GUIViewState.cpp
```diff
@@ -344,6 +345,37 @@ void CGUIViewState::SetSortMethod(SortDescription sortDescription)
   return SetSortMethod(sortDescription.sortBy, sortDescription.sortOrder);
 }
 
+SortDescription CGUIViewState::ChooseSortMethod()
+{
+  
+  CGUIDialogSelect *dialog = (CGUIDialogSelect *)g_windowManager.GetWindow(WINDOW_DIALOG_SELECT);
```
tamland: use `static_cast`

xbmc/view/GUIViewState.cpp
```diff
@@ -344,6 +345,37 @@ void CGUIViewState::SetSortMethod(SortDescription sortDescription)
...
+  if (dialog)
+  {
```
tamland: call `dialog->Reset()` here

xbmc/view/GUIViewState.cpp
```diff
@@ -344,6 +345,37 @@ void CGUIViewState::SetSortMethod(SortDescription sortDescription)
...
+    dialog->SetHeading(CVariant{ 32104 }); // Label "Sort by"
+    int iIndex = 0;
+    int iSelected = 0;
+    for (int i = 0; i < m_sortMethods.size(); ++i)
+    {
+      dialog->Add(g_localizeStrings.Get(m_sortMethods[i].m_buttonLabel));
+      if (i == m_currentSortMethod)
+        iSelected = iIndex;
+      iIndex++;
+    }
```
tamland: All the indices here seems unnecessary.. Simply pass m_currentSortMethod to SetSelected

phil65: oh, ouch :) not sure what I was thinkin there..

phil65:
> Improvement for sure, but I always envisioned the side menu to be similar to the context menu where all select-like actions takes you to a new 'page' within the same menu.

We also already have the filterdialog for example, so I dont  think that this introduces inconsistency.

phil65:
@tamland updated and took care of your comments. Thx for review.
Sorry for the quality of the initial PR, quite some brainfarts in there. I probably shouldnt work at night. :)

xbmc/view/GUIViewState.cpp
```diff
@@ -344,6 +345,32 @@ void CGUIViewState::SetSortMethod(SortDescription sortDescription)
...
+    int newSelected = dialog->GetSelectedItem();
+    // check if selection has changed
+    if (!dialog->IsConfirmed() || newSelected < 0 || newSelected == m_currentSortMethod)
+      return false;
+
+    m_currentSortMethod = newSelected;
+
+    SaveViewState();
+  }
+  return true;
```
tamland: now it returns success if dialog is null and wasn't opened. probably best to swap the two return statements (such that the success case is the inner-most, failure last)

xbmc/view/GUIViewState.cpp
```diff
@@ -344,6 +345,32 @@ void CGUIViewState::SetSortMethod(SortDescription sortDescription)
...
 
+bool CGUIViewState::ChooseSortMethod()
+{
+  
+  CGUIDialogSelect *dialog = static_cast<CGUIDialogSelect *>(g_windowManager.GetWindow(WINDOW_DIALOG_SELECT));
+  if (dialog)
+  {
+    dialog->Reset();
+    dialog->SetHeading(CVariant{ 32104 }); // Label "Sort by"
+    for (int i = 0; i < m_sortMethods.size(); ++i)
```
tamland: please use ranged-based loops, and no brackets for single statements;)

xbmc/windows/GUIMediaWindow.cpp
```diff
@@ -282,8 +282,8 @@ bool CGUIMediaWindow::OnMessage(CGUIMessage& message)
       else if (iControl == CONTROL_BTNSORTBY) // sort by
       {
         if (m_guiState.get())
-          m_guiState->SetNextSortMethod();
-        UpdateFileList();
+        if (m_guiState->ChooseSortMethod())
```
tamland: merge to one `if`. is nesting them like this even correct? I can never remember:\

phil65:
@tamland updated, and refactored to exit early in case dialog = NULL.

phil65:
jenkins build this please.

PR #9648, opened 2016-04-19
Disable autoscrolling while on screensaver and while opening streams.

mcaptur:
Taken From: anaconda anaconda@menakite.eu Date: Thu, 11 Sep 2014 21:30:43 +0200
Been used for over a year in RPi (newclock)

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -5146,3 +5146,13 @@ bool CApplication::NotifyActionListeners(const CAction &action) const
   
   return false;
 }
+
+bool CApplication::ScreenSaverDisablesAutoScrolling()
+{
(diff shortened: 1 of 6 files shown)


FernetMenta:
The implementation is not very good because of circular dependencies. GUI is owned by application and should not use those crappy globals to poll application status. Instead application should instruct GUI how to behave.

anaconda:
The reason this is only in newclock5 branch is because it was rejected upstream (#5363).

alan-ml:
It would be great if this makes it to the next stable release, I for one will enable my screensaver once this is in.
Its not just about reducing CPU usage, but considering the number of Kodi users, this little change will save many watts worldwide.

MartijnKaijser:
It's safe to say this patch won't get back in

PR #9650, opened 2016-04-19
[ios] - fix missing song metadata in control center when music is scanned into the db

Memphiz:
What the topic says...

@DaveTBlake for review (i think you are the guru when it comes to the music db)

### DIFF
--- a/xbmc/platform/darwin/ios-common/AnnounceReceiver.mm
+++ b/xbmc/platform/darwin/ios-common/AnnounceReceiver.mm
@@ -28,6 +28,7 @@
 #import "Application.h"
 #import "FileItem.h"
 #import "music/tags/MusicInfoTag.h"
+#import "music/MusicDatabase.h"
 #import "TextureCache.h"
 #import "filesystem/SpecialProtocol.h"
 #import "playlists/PlayList.h"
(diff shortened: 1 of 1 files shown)


xbmc/platform/darwin/ios-common/AnnounceReceiver.mm
```diff
@@ -93,10 +94,52 @@ id objectFromVariant(const CVariant &data)
...
+        CMusicDatabase db;
+        if (db.Open())
+        {
+          CSong song;
+          if (db.GetSong(item_id, song))
+          {
+            nonConstData["item"]["title"] = song.strTitle;
+            nonConstData["item"]["track"] = song.iTrack;
+            nonConstData["item"]["album"] = song.strAlbum;
+            nonConstData["item"]["artist"] = song.GetArtist();
```
DaveTBlake: You probably want song.GetArtistString() here, the display value rather than the vector of artists. If we have Musicbrainz IDs then one is not always the simple union of the other e.g. "B. B. King & Eric Clapton" rather than "B. B.King", "Eric Clapton"

DaveTBlake:
I don't know much about the rest of Kodi to know how this fits in with everything else though, but music DB I get :)

xbmc/platform/darwin/ios-common/AnnounceReceiver.mm
Memphiz: I need the array here - thats what is in there when we don't have the item scanned into the db - the rest of the code relies on it. I simply try to mimic the data as if it was announced from a non scanned item (where all that metadata is already filled in the data param) - its only when the item is scanned into the db that the announce caller only puts the db id into that data dictionary...

DaveTBlake: Fair enough, I'm not familiar with the announcer at all.

PR #9651, opened 2016-04-19
[RFC] [ffmpeg] Fix memleaks

Kwiboo:
This PR should fix some ffmpeg memleaks I discovered while doing some debugging, grateful for feedback from someone with more ffmpeg knowledge.

Changes to use `av_frame_free` (ffmpeg 2.2) and `avcodec_free_context` (ffmpeg 2.3) to free resources instead of `av_free`/`av_freep`.
Note: `avcodec_free_context` will call `avcodec_close` if the codec is open, free extradata and any other internal resources used by the context.

The change to use `avsubtitle_free` changes behavior and keeps the subtitle images in memory until next Decode call.

### DIFF
--- a/xbmc/cdrip/EncoderFFmpeg.cpp
+++ b/xbmc/cdrip/EncoderFFmpeg.cpp
@@ -159,8 +159,8 @@ bool CEncoderFFmpeg::Init(audioenc_callbacks &callbacks)
   if(!m_BufferFrame || !m_Buffer)
   {
     CLog::Log(LOGERROR, "CEncoderFFmpeg::Init - Failed to allocate necessary buffers");
-    if(m_BufferFrame) av_frame_free(&m_BufferFrame);
-    if(m_Buffer) av_freep(&m_Buffer);
+    av_frame_free(&m_BufferFrame);
+    av_freep(&m_Buffer);
     av_freep(&m_Stream);
     av_freep(&m_Format->pb);
     av_freep(&m_Format);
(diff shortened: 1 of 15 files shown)


FernetMenta:
Thanks!
jenkins build this please

fritsch:
@Kwiboo Thanks very much, could you also have a look in FFmpegPicture.cpp while at it?

Kwiboo:
@fritsch Did a quick glance and there could be a `av_free` call that should be changed, see https://github.com/xbmc/xbmc/blob/master/xbmc/guilib/FFmpegImage.cpp#L462, there is also some unneeded nullptr resets and unneeded `avcodec_close` calls but will look more closely and under a debugger later

fritsch:
av_free is used to free something that was alloced with av_malloc earlier. Why do you want to change it?

Kwiboo:
`pictureRGB` is a AVFrame allocated with `av_frame_alloc` and that should be freed with `av_frame_free` not `av_free`/`av_freep`

`av_frame_free` calls `av_frame_unref` (https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/frame.c#L481) on the frame and that method takes care of cleaning up internal allocations, in this case I do not think anything have been allocated so it probably does not matter.

fritsch:
@Kwiboo obviously correct, yes - thanks. I overlooked the "frame" twice ...

fritsch:
@Kwiboo are you still looking into PictureFFmpeg? Don't want to push something and interrupt / highjack your work, that's why I am asking.

Kwiboo:
@fritsch I have not started anything yet and wont have time until tomorrow, so please push :)

fritsch:
No hurry - Currently gathering the findings here: https://github.com/fritsch/xbmc/commits/ffmpegpicture - just that we don't forget it. Thanks in advance.

PR #2250, opened 2013-02-16
addons: Add Announce function and register addons for announcements

fetzerch:
This allows addons to receive announcements and react properly on
all kind of events and system state transitions.

For example pvr addons might want to close and reopen their backend
connections when the system goes to or resumes from standby.

PR #9654, opened 2016-04-20
ADDON: revert announcement interface, not thread safe

FernetMenta:
The addon dll interface is not designed for thread safety. Only the owning object is supposed to use this interface.
reverts: https://github.com/xbmc/xbmc/pull/2250

@fetzerch @ksooo @Jalle19 
afaik only a few pvr addons make use of this. PVRManager implements IAnnouncer already. No need for the short cut.

### DIFF
--- a/xbmc/addons/AddonDll.h
+++ b/xbmc/addons/AddonDll.h
@@ -31,8 +31,6 @@
 #include "filesystem/SpecialProtocol.h"
 #include "filesystem/Directory.h"
 #include "utils/log.h"
-#include "interfaces/IAnnouncer.h"
-#include "interfaces/AnnouncementManager.h"
 #include "utils/XMLUtils.h"
 #include "utils/Variant.h"
 #include "Util.h"
@@ -40,7 +38,7 @@
 namespace ADDON
 {
   template<class TheDll, typename TheStruct, typename TheProps>
-  class CAddonDll : public CAddon, public ANNOUNCEMENT::IAnnouncer
+  class CAddonDll : public CAddon
   {
   public:
     CAddonDll(AddonProps props);
@@ -60,8 +58,6 @@ namespace ADDON
     void Destroy();
 
     bool DllLoaded(void) const;
-
-    void Announce(ANNOUNCEMENT::AnnouncementFlag flag, const char *sender, const char *message, const CVariant &data);
 
   protected:
     void HandleException(std::exception &e, const char* context);
@@ -235,7 +231,6 @@ ADDON_STATUS CAddonDll<TheDll, TheStruct, TheProps>::Create()
     if (status == ADDON_STATUS_OK)
     {
       m_initialized = true;
-      ANNOUNCEMENT::CAnnouncementManager::GetInstance().AddAnnouncer(this);
     }
     else if ((status == ADDON_STATUS_NEED_SETTINGS) || (status == ADDON_STATUS_NEED_SAVEDSETTINGS))
     {
@@ -298,8 +293,6 @@ void CAddonDll<TheDll, TheStruct, TheProps>::Stop()
 template<class TheDll, typename TheStruct, typename TheProps>
 void CAddonDll<TheDll, TheStruct, TheProps>::Destroy()
 {
-  ANNOUNCEMENT::CAnnouncementManager::GetInstance().RemoveAnnouncer(this);
-
   /* Unload library file */
   try
   {
@@ -535,19 +528,6 @@ ADDON_STATUS CAddonDll<TheDll, TheStruct, TheProps>::TransferSettings()
   }
 
   return ADDON_STATUS_OK;
-}
-
-template<class TheDll, typename TheStruct, typename TheProps>
-void CAddonDll<TheDll, TheStruct, TheProps>::Announce(ANNOUNCEMENT::AnnouncementFlag flag, const char *sender, const char *message, const CVariant &data)
-{
-  try
-  {
-    m_pDll->Announce(ANNOUNCEMENT::AnnouncementFlagToString(flag), sender, message, &data);
-  }
-  catch (std::exception &e)
-  {
-    HandleException(e, "m_pDll->Announce()");
-  }
 }
 
 template<class TheDll, typename TheStruct, typename TheProps>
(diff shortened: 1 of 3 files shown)


ksooo:
@FernetMenta pvr.hts is affected. It "listens" for Kodi suspend and resume events. https://github.com/kodi-pvr/pvr.hts/blob/master/src/client.cpp#L175 What is the replacement for this after removal of ADDON_Announce?

Jalle19:
@ksooo we could probably manage without our wake/sleep handlers, AFAICT we're only using it to prevent some log spam.

FernetMenta:
> What is the replacement for this after removal of ADDON_Announce?

I delegate this to the authors who implemented it. If this is really required, I suggest to route the event via CPVRManager::Announce to the addon.

ksooo:
> AFAICT we're only using it to prevent some log spam.

 Wrong. We're for instance preventing to reconnect to tvh server while suspending/not fully resumed which results in all kind of errors (which are logged, then).

> I suggest to route the event via CPVRManager::Announce to the addon

 If this is a not yet existing interface I strictly vote for first implementing this new approach and only after that removing the old functionality. I'm a friend of removing broken stuff, but just removing something without providing adequate replacement is not fair.

FernetMenta:
@ksooo fixing crashes is for sure ranked higher than a broken feature. This feature was implemented incorrectly and it is causing segfaults.

ksooo:
-1 from my side as long as there is no alternative available.

FernetMenta:
@ksooo you can't really think that other have to suffer from crashes that a few can have their feature. Merging this functionality was wrong and I will revert it for the sake of the majority. My long todo list won't allow me to provide an alternative. If you want an alternative, you have to implement it yourself.

fritsch:
In case someone did not get, what this causes: https://dl.dropboxusercontent.com/u/53913326/kodi_crashlog_20160420104333.log it broke most of our viz ... so it's not fun at all ...

fetzerch:
Well sorry for introducing this issue (and provoking yet another fight). I do think that it's important to fix the crash but I fear I won't have time to look into it at the moment. If no one else is interested or has time I don't see a better way than reverting it for the time being.

For MythTV it's gonna cause the issue that you have to restart Kodi after suspend/resume.
Ping @janbar just as a heads-up.

janbar:
Hi all. Yes the MythTV addon is affected by this and It needs the feature to handle properly the suspend/wakeup. So it seems to be not so hard to make the alternative. I think pvr-manager could push "announce" to addon in function https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/PVRManager.cpp#L114. If nobody want make the alternative I will make it.

ksooo:
> If nobody want make the alternative I will make it.

 Many thanks. Go ahead. And please don't forget, this will be a major PVR addon API bump, because all(!) PVR addons need to implement the new API function.

janbar:
@ksooo, implementing the alternative won't change the API. It will be nice if all addons keep their current interface, including announce(), avoiding me to remake existing code in addons.

janbar:
I will rename ADDON_Announce() to Announce() too.

FernetMenta:
@janbar If you change the API, I think it makes sense to rename this function to something meaningful. If there's is more than a single event, create more functions.

ksooo:
> If you change the API, I think it makes sense to rename this function to something meaningful. If there's is more than a single event, create more functions.

I would grep over the pvr addons looking for ADDON_Announce and what events the addons are actually interested in and make separate functions for that events or if it is always a combination of suspend/wake maybe one function for "power management" and the like. Depends on what is actually used.

ksooo:
@FernetMenta what comes to my mind is the question whether generic stuff like power management events should really be part of the pvr addon api. Wouldn't it be better to have such generic stuff in a generic addon api? But wait, that's where we are coming from, right?

FernetMenta:
>  But wait, that's where we are coming from, right?

Right. The interface is not thread safe and only the object that calls Create() is supposed to talk to the addon. "Create" is ofc misleading because it does not create the instance but loads the shared lib.

I am not sure if power management is something generic in regard to addons. pvr addons are unique because they communicate with a backend. We started with pvr addons and have grown the number of types. As a result we need to demote things from the generic parts back into the specific addons. Another example is ADDON_STATUS_LOST_CONNECTION that we should finally finally drop.

ksooo:
> pvr addons are unique because they communicate with a backend.

 yeah, at least atm. But this can change at every point in time. Do you suggest that the "other" addon type then also introduces its own power management api interface or because there are then two of that kind to introduce a generic interface, drop the pvr interface and switch pvr addons to the generic interface, then? Both are valid approaches. Just asking.

FernetMenta:
I prefer the latter: implement new functions, understand the system, and identify thinks like hierarchy and common methods.

ksooo:
> I prefer the latter: implement new functions, understand the system, and identify thinks like hierarchy and common methods.

 Okay. I'm fine with this.

janbar:
Implementing one function per message will break the flexibility. Looking in addons code I see not only event about power management. But "OnPlay" too. Also announcement seems to be useful to receive event from other addon too. For my part I use only suspend/wakeup and screensaverActivated/Deactivated. So current interface using "Announce()" is well enough for me and I haven't so many time to design a big stuff :). I Plan to create a pending job in pvr-manager which push the announcement, not more.
What you need exactly ?

FernetMenta:
In most cases flexibility equals abuse and decay of design. The less strict an interface is the worse it gets. Devs tend to be very creative in putting the funniest things into such generic methods.
Just look at the current situation. We don't know exactly for what purposes Announce is used which makes it hard to change anything.

> Also announcement seems to be useful to receive event from other addon too

Creating dependencies between addon instances is a bad idea. This needs to be handles by the parent objects.

FernetMenta:
> What do you need exactly ?

suspend/wakeup for now

FernetMenta:
btw: announcements are asynchronous messages and there is no guarantee that the message is delivers before the system goes down. with an extra job in PVRManager the probability the the info hit the addon is even less.

janbar:
Do you mean I should call addon interface synchronously ?

FernetMenta:
> Do you mean I should call addon interface synchronously 

If you want to make sure that the addon gets the info, announce is not the way to go.

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

PR #9655, opened 2016-04-20
[PVR] CPVRClients::GetTimers: Improve error handling.

ksooo:
Ever wondered about strange timer notifications - namely "deleted timer blah" and the like - and you are sure that actually no timer was deleted? 

So far, Kodi effectively ignores errors that occurred when timers should be obtained from clients. Especially, if on first call timers could be obtained successfully and on second call not, thus list of returned timers will be empty, and the timer sync algo here https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/timers/PVRTimers.cpp#L189 erroneously will treat the empty list as "all timers got deleted on pvr client".

All other "wrong" notifications are followups of the problem just described.

@Jalle19 mind taking a look. btw, for pvr.hts I discovered an additional problem that can also trigger wrong timer notifications. Will come up with a PR for this as well.

### DIFF
--- a/xbmc/pvr/addons/PVRClients.cpp
+++ b/xbmc/pvr/addons/PVRClients.cpp
@@ -458,9 +458,9 @@ bool CPVRClients::HasTimerSupport(int iClientId)
   return false;
 }
 
-PVR_ERROR CPVRClients::GetTimers(CPVRTimers *timers)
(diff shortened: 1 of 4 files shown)


ksooo:
jenkins build this please

ksooo:
if nobody objects I will merge this tomorrow

da-anda:
nice one, thanks. This bugged me since ages.

PR #9656, opened 2016-04-21
[OSX] Fix non working apple remote / universal ir remote / harmony

Memphiz:
This finally brings back remote control support on osx which was lost during joystick refactor.

@garbear fyi

This needs to go into 17 - there are a lot of macmini users relying on using the small silver apple remote for example (or other ir remotes) which would not work without this.

I verified that those key ids don't conflict with normal keyboard codes (which are vkeys and have a much higher value).

Any concerns?

### DIFF
--- a/system/keymaps/universalremote.AppleRemote.xml
+++ b/system/keymaps/universalremote.AppleRemote.xml
@@ -1,0 +1,200 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- This file contains the mapping of keys (gamepad,remote and keyboard) to actions within XBMC -->
+<!-- The <global> section is a fall through - they will only be used if the button is not        -->
+<!-- used in the current window's section.  Note that there is only handling                     -->
+<!-- for a single action per button at this stage.                                               -->
+<!-- For joystick/gamepad configuration under linux/win32, see below as it differs from xbox     -->
+<!-- gamepads.                                                                                   -->
+
+<!-- The format is:                                                                              -->
+<!--    <device>                                                                                 -->
+<!--      <obc#>xbmc action</obc#>                                                    -->
+<!--    </device>                                                                                -->
+
+<!-- To map keys from other remotes using the RCA protocol, you may add <universalremote> blocks -->
+<!-- In this case, the tags used are <obc#> where # is the original button code (OBC) of the key -->
+<!-- You set it up by adding a <universalremote> block to the window or <global> section:        -->
+<!--    <universalremote>                                                                        -->
+<!--       <obc45>Stop</obc45>                                                                   -->
+<!--    </universalremote>                                                                       -->
+
+<!-- Note that the action can be a built-in function.                                            -->
+<!--            eg <obc6>ActivateWindow(Favourites)</obc6>                       -->
+<!-- would bring up Favourites when the button with the id of 6 is press. In this case, "Menu"   -->
+
+<!--                                                                                             -->
+<!--  Button Ids:                                                                                -->
+<!--  'id' is the button ID used by SDL. The key ids recognized from your remote appears      -->
+<!--  in /var/log/syslog on the ATV2 for each button pressed and when debug mode is enabled      -->
+<!--  Use your log to discover and map custom buttons to actions.                                -->
+
+<keymap>
+  <global>
+    <universalremote>
+      <!-- plus       -->      <obc1>Up</obc1>
+      <!-- minus      -->      <obc2>Down</obc2>
+      <!-- left       -->      <obc3>Left</obc3>
+      <!-- right      -->      <obc4>Right</obc4>
+      <!-- center     -->      <obc5>Select</obc5>
+      <!-- menu       -->      <obc6>ParentDir</obc6>
+      <!-- hold center-->      <obc7>Fullscreen</obc7>
+      <!-- hold menu  -->      <obc8>ContextMenu</obc8>
+
+      <!-- old buttons for ATV1 <2.2, used on OSX  -->
+      <!-- hold left  -->      <obc9>Left</obc9>
+      <!-- hold right -->      <obc10>Right</obc10>
+
+      <!-- new aluminium remote buttons  -->
+      <!-- play/pause -->      <obc12>PlayPause</obc12>
+
+      <!-- Additional buttons via Harmony Apple TV remote profile - these are also the learned buttons on Apple TV 2gen-->
+      <!-- pageup     -->      <obc13>PageUp</obc13>
+      <!-- pagedown   -->      <obc14>PageDown</obc14>
+      <!-- pause      -->      <obc15>Pause</obc15><!-- work? -->
+      <!-- play2      -->      <obc16>Play</obc16>
+      <!-- stop       -->      <obc17>Stop</obc17>
+      <!-- fast fwd   -->      <obc18>FastForward</obc18><!-- work? -->
+      <!-- rewind     -->      <obc19>Rewind</obc19>
+      <!-- skip fwd   -->      <obc20>SkipNext</obc20>
+      <!-- skip back  -->      <obc21>SkipPrevious</obc21>
+
+      <!-- Learned remote buttons (ATV1 >2.3) -->
+      <!-- Play       -->      <obc70>Play</obc70>
+      <!-- Pause      -->      <obc71>Pause</obc71>
+      <!-- Stop       -->      <obc72>Stop</obc72>
+      <!-- Previous   -->      <obc73>SkipPrevious</obc73>
+      <!-- Next       -->      <obc74>SkipNext</obc74>
+      <!-- Rewind     -->      <obc75>Rewind</obc75>
+      <!-- Forward    -->      <obc76>FastForward</obc76>
+      <!-- Return     -->      <obc77>OSD</obc77>
+      <!-- Enter      -->      <obc78>ShowVideoMenu</obc78>
+
+      <!-- few gestures from Apple's iPhone Remote (ATV1 > 2.3 ?) -->
+      <!-- SwipeLeft  -->      <obc80>Left</obc80>
+      <!-- SwipeRight -->      <obc81>Right</obc81>
+      <!-- SwipeUp    -->      <obc82>Up</obc82>
+      <!-- SwipeDown  -->      <obc83>Down</obc83>
+
+      <!-- FlickLeft  -->      <obc85>Left</obc85>
+      <!-- FlickRight -->      <obc86>Right</obc86>
+      <!-- FlickUp    -->      <obc87>Up</obc87>
+      <!-- FlickDown  -->      <obc88>Down</obc88>
+
+    </universalremote>
+  </global>
+  <Home>
+    <universalremote>
+      <obc6>ActivateWindow(Favourites)</obc6>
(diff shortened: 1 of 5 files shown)


Memphiz:
@NedScott you've got a bunch of remotes iirc - are you able to test?

testbuild http://mirrors.kodi.tv/test-builds/osx/x86_64/kodi-20160419-91d79ce-osxremote_as_keyboard-x86_64.dmg

Tolriq:
@Memphiz I think that this will clash with event server sending keyboard keys as unicode.

You can easily test if you have an Android device and Yatse.

garbear:
As a last ditch effort, EventServer has a "gamepad" API that uses [gamepad.xml](https://github.com/xbmc/xbmc/blob/master/system/keymaps/gamepad.xml), it only has ~20 keys, but it could be useful if you run out of key IDs or need analog values

Memphiz:
harmony remote has around 100 mappable buttons - i need a way to pass arbitrary IDs. - how exactly can i test that possible clash?

Memphiz:
@tolriq are you sure yatse comes out on the eventserver? pretty sure it comes in via json...

Montellese:
Input events from Yatse come in through EventServer. The rest comes in through JSON-RPC.

Tolriq:
@Memphiz I wrote each line of Yatse, so yes I'm pretty sure ;) + I wrote the EventServer patch that added the Unicode support + restored EventServer in @garbear patch ;)

From Yatse click the keyboard button in left menu or in remote screen.

The paste or the dialog that auto shows on input request do use JSON.

If you need a solution that does not use the correct joystick API, adding a new mapping can be done or you can use the R1 mapping https://github.com/xbmc/xbmc/blob/master/xbmc/input/ButtonTranslator.cpp#L1162

Memphiz:
1. i can't use the joystick api and rely on the eventserver - the client is a daemon which also allows to start kodi via remote (so it mus run when kodi is not there yet)
2. I can't use any predefined key mappings because i need to pass through some button codes. I don't know which button sends which because the harmony remote can be freely programmed.

Memphiz:
ATM we even support some key sequences which are no single button presses but something like "menu" + "up" - see https://github.com/xbmc/xbmc/blob/master/tools/EventClients/Clients/OSXRemote/xbmcclientwrapper.mm#L303

NedScott:
I was under the impression that the Harmony remotes can only send a small number of buttons with an Apple IR receiver? For example, I've never heard of anyone getting the numberpad to work on a Harmony remote with Apple's built-in receiver under OS X.

Memphiz:
I only can tell from the code - it supports 10x10 keys - maybe @davilla can light me up?

Tolriq:
IMO the main question is what is the long term plan.

If Event Server is still considered as a core component like WebServer and is not to be moved to a controller binary addon, then proper solution would be to have the same kind of lirc handling.

Adding support for a mapping like CS:XXX (CS for custom) that will load a use a custom.XXX.xml mapping file.

This allows support for anything without the need for core change.

Memphiz:
Switched over to use the universalremote mapping now (obc ids). I didn't try if the former approach influences with yatse though. All happy with this one now?

Memphiz:
I am going to merge this in the next couple of days if no objections - @NedScott do you think you could adapt the related wiki pages? There is also a testbuild on the mirrors if you are interested in it...

NedScott:
I'll be sure to update the wiki pages.

Sadly, I can only test with the Apple remote, as I don't have a Harmony.

Memphiz:
@NedScott merged it :)

MilhouseVH:
After this PR, with LibreELEC and an Ortek VRC-1100 remote (on both RPi and x86) I get the following incorrect button mappings, where RIGHT button action is Up, and LEFT button action is Right:
```
23:05:26  83.924530 T:1962168320   DEBUG: LIRC: Update - NEW at 72246:6a 0 KEY_RIGHT devinput (KEY_RIGHT)
23:05:26  83.924690 T:1962168320   DEBUG: OnKey: 168 (0xa8, obc87) pressed, action is Up
23:05:27  84.207924 T:1962168320   DEBUG: LIRC: Update - NEW at 72529:6a 0 KEY_RIGHT_UP devinput (KEY_RIGHT_UP)
23:05:28  85.110397 T:1962168320   DEBUG: LIRC: Update - NEW at 73432:69 0 KEY_LEFT devinput (KEY_LEFT)
23:05:28  85.110512 T:1962168320   DEBUG: OnKey: 169 (0xa9, obc86) pressed, action is Right
23:05:28  85.370789 T:1962168320   DEBUG: LIRC: Update - NEW at 73692:69 0 KEY_LEFT_UP devinput (KEY_LEFT_UP)
```
The mappings in `system/keymaps/universalremote.AppleRemote.xml` appear to be used (`obc86` is Right, and `obc87` is Up), overriding the normal default button mapping.

Any suggestions on why this is happening? Reverting this PR for now fixes the issue. Not shipping AppleRemote.xml in LibreELEC is another option...

Full debug log: http://sprunge.us/TVGb

popcornmix:
Same issue with CEC remote and LibreELEC on Pi.

Memphiz:
Mhhh are those even using obc codes or "universalremote" maps? Sounds like those maps shouldn't even be Applied in the lirc and cec situation no? - are only the cursor keys a problem here? If in doubt we Could fix the Map and change XBMCHelper to send other Codes ... But it more Sounds like a Bug in the Map Handling?

Memphiz:
In think i simply will change it to only distribute that keymap in osx an keep it somewhere osx specific in our Source tree...

MilhouseVH:
> are only the cursor keys a problem here

I didn't test extensively, so there may be other button conflicts, but only KEY_LEFT and KEY_RIGHT appeared to be affected as the obc code for these buttons is mapped to `Right` and `Up` in the Apple xml. The KEY_DOWN obc code is 88 and the same as Apple so works as expected, however KEY_UP is obc 89 which isn't mapped in either of the new Apple or Harmony maps so presumably this falls back to the default mapping.

So really the fact Up and Down are working is probably more by luck than anything else!
```
23:04:18  15.136905 T:1962168320  NOTICE: eventlircd: keys:0 ext:0 but:1 rel:2 abs:0 m_deviceType:1 bits:110000
23:04:53  50.853397 T:1962168320   DEBUG: LIRC: Update - NEW at 39175:6c 0 KEY_DOWN devinput (KEY_DOWN)
23:04:53  50.853527 T:1962168320   DEBUG: OnKey: 167 (0xa7, obc88) pressed, action is Down
23:04:53  51.041183 T:1962168320   DEBUG: LIRC: Update - NEW at 39363:6c 0 KEY_DOWN_UP devinput (KEY_DOWN_UP)
23:04:54  51.607845 T:1962168320   DEBUG: LIRC: Update - NEW at 39929:67 0 KEY_UP devinput (KEY_UP)
23:04:54  51.607983 T:1962168320   DEBUG: OnKey: 166 (0xa6, obc89) pressed, action is Up
23:04:54  51.807888 T:1962168320   DEBUG: LIRC: Update - NEW at 40129:67 0 KEY_UP_UP devinput (KEY_UP_UP)
```
> In think i simply will change it to only distribute that keymap in osx an keep it somewhere osx specific in our Source tree...

If there's no other way to prevent loading the Apple xml on non-Apple distros, that maybe the best solution.

Tolriq:
Just passing again for the bad guy, but my last comment is actually the real question leading to the proper solution.

Mapping files are all loaded and merged based on the type declared in it.

Those files could impact also on MacOS. If both your mapping files had overlapping keys, result would be unknown.

Normal solution would be that your application that use EventServer does send the known default codes that are in default keymaps to have the correct options.

Or as proposed extends EventServer, but providing keymaps that override default / other more global commands for a specific remote is clearly not the way to use what is currently in Kodi core.

Memphiz:
As already said - i don't know the keys which are pressed by the harmony remote for example - thats why we send key codes and not key names ... - i of course can add just another mapping to the eventserver but am pretty sure that someone other will reje t it because we already have a bunch of maps available. Technically all remotes that send keyboard ley codes will influence with our keyboard map aswell. The only real thing is what is done with the peripherals iirc - those can be mapped by usb id or something like that.

Memphiz:
Thats what i really hate about regressions. The maintainer who spots it has to clean up the mess :(

Tolriq:
I tried to warn about the impacts on the original PR :) I did ask for a global communication so that more people see the impacts.
I fought to at least recover the minimal EventServer functions.

Original PR was the same problem, in the fact that your keymaps override default ones globally :( So would have impacted lot's of things using EventServer.

All remotes that send keyboard codes, sends keyboard code that have default action and that the user can override. Having a file named apple that override default keymap won't be easy for end user to spot.

Event server did support LI:XXX code to achieve what you want to but it was removed too in original PR. My proposal is to bring back a CS:XXXX to achieve the same purpose but extended to not only LIRC but any EventServer client.

About the part about keycodes and names I'm not sure to understand the problem, in every case you receive data that you convert to the obc code no ?

I doubt any one would refuse a PR to add CS:XXX as there's no impact and you can still provide only the keymap for MacOS, but another solution would be that you provide the mapping in your tool no ?

bas-t:
This sure messes up my mceusb bigtime.
I'm not a coder so I don't know how. But reverting these changes brings back my mceusb to it's normal fantastic behaviour.

garbear:
Sorry for causing this headache. What if we restore the eventserver code I removed? it was an abuse to call this a "joystick" in the first place, but what if we bring back the idea of a many-buttoned device and call it a "remote" or something? We could restore the code from bbbc903 but change `TranslateJoystickString()` to `TranslateRemoteString()`, `MapJoystickActions()` to `MapRemoteActions()`, etc. We don't even need the axis or hat code, just the code for the buttons. The keymap would stay the same as the old one, but start with `<remote>` instead of `<joystick>`

Tolriq:
Well it was really made to make joystick things as it did allow axis and lot's of other things.

But what you propose is exactly the CS:XXXXX, since we need a new identifier to be split to get a mapping name.

If it's no more joystick it does not seems correct to call it JS:XXXXX, if you do not like custom then why not RX:XXXX to follow R1 and R2, X for custom ?

Specially if there's other things that use this legacy things that would lead to unknown result for them. Better to deprecate / start fresh.

Using already used remote mapping may have some impacts due to merging and everything, I do not know enough that part of the code to know how all cases could be handled but if someone can validate how things woks for all edge cases (what heritage from what / where when a map is not declared somewhere) then why not.

Memphiz:
Does it need to be a 2 char code before the colon? I think i will do exactly that - bring back the custom keymapnames - i like the naming proposal from garbear (calling it remote), how about RC:keymapname (rc - remote control or remote custom...)

Tolriq:
No just at least 2 from current code. Having 2 limits the needed changes.

About the naming, remote without name is already used. And as I said I doubt anyone can validate all the possible edge cases as no one have seen most of the problems of previous PRs :(.

IMO avoiding risk and starting fresh / proper sounds better.

Memphiz:
Let me rephrase - i will do it in a way that doesn't influence anything that we have now ...

bas-t:
I hope that you mean:
"Let me rephrase - i will do it in a way that doesn't influence anything that we have now ..., and by now I mean the situation we had prior to comiting this pr"

Memphiz:
Of course :)

PR #9658, opened 2016-04-21
PVR code cleanup

MaxKellermann:


ksooo:
jenkins build this please

ksooo:
Thanks, @MaxKellermann

FernetMenta:
@ksooo we are not in a merge window and this is not a fix (at least it is labelled cleanup)

ksooo:
Ooops, sorry.

FernetMenta:
np. I just wanted to mention it because I think more members and contributors may have missed this info

MartijnKaijser:
I think we should allow these minor "cleanups" like includes and dead code. If it was wrong it wouldn't even compile.

PR #9660, opened 2016-04-21
SubtitleDialog, remember last service/search.

raptorjr:
My first attempt to make a little change to Kodi that I think is a improvement.
This makes the SubtitleDialog remember the last service used, and also the last search instead of always using, and searching, the first service in the list.

The intent is that last search should only be remembered for the current video playing, and if the playback is stopped and the SubtitleDialog is opened for another video only the last service is remembered.

But I do think that what I do in StopPlaying isn't the right way to do it, or if I even use CGUIMessage the right way?

### DIFF
--- a/xbmc/video/dialogs/GUIDialogSubtitles.cpp
+++ b/xbmc/video/dialogs/GUIDialogSubtitles.cpp
@@ -162,8 +162,14 @@ bool CGUIDialogSubtitles::OnMessage(CGUIMessage& message)
 
     CGUIDialog::OnMessage(message);
 
+    return true;
+  }
+  else if (message.GetMessage() == GUI_MSG_WINDOW_RESET)
+  {
+    CGUIDialog::OnMessage(message);
+    
+    // Clear saved subtitles from any previous subtitle search
     ClearSubtitles();
-    ClearServices();
     return true;
   }
   return CGUIDialog::OnMessage(message);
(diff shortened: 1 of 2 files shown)


phate89:
There's an option in settings to chose the preferred subtitle provider. Isn't this overriding that option?

raptorjr:
No, the first time the dialog opens the preferred subtitle provider is used, if one is selected in the settings. And if that provider gives you the subtitles you need this patch does nothing besides remember the search so there is no need for a new search the next time you open the dialog, if the first subtitle you download wasn't good.

But if your preferred provider don't have any subtitles and you choose to search another one, and download the first from that provider and that subtitle is bad also. This patch will let you continue to find a good subtitle from the last provider used, instead of always defaulting to a provider you already tried and that didn't give you good subtitles.

FernetMenta:
@phil65 @Voyager1 @ace20022  could you please have a look at this

MartijnKaijser:
Ping

phil65:
That change of behaviour makes sense to me, so +1

MartijnKaijser:
jenkins build this please

xbmc/video/dialogs/GUIDialogSubtitles.cpp
```diff
@@ -168,6 +168,7 @@ bool CGUIDialogSubtitles::OnMessage(CGUIMessage& message)
   {
     CGUIDialog::OnMessage(message);
     
+    // Clerar saved subtitles from any previos subtitle search
```
garbear: "Clear" and "previous"

ronie:
you might want to drop that merge commit ;-)

also, please squash your changes into a single commit.

raptorjr:
Sorry, I'm out on some deep water now, learning by doing =\
Didn't even know I did a merge commit, much less how to remove it. Googling about it now.
Will also find out how to make all changes into one commit.

raptorjr:
It just gets worse =( This is hard to understand.

MartijnKaijser:
another ping for @phil65 @Voyager1 @ace20022

MartijnKaijser:
Jenkins build this please 
Any objections?

arnova:
+1

PR #9661, opened 2016-04-21
FFmpegImage: Fix memory leak

fritsch:
After recent ffmpeg fixes, there was also an issue found in our Picture implementation.


fritsch:
jenkins build this please

da-anda:
is this an older memleak? could that be the fix for Kodi going crazy on RAM when browsing large picture folders?

fritsch:
No - the other one is no memleak, it is by design :-) if you navigate into a folder with a huge amount of images in there.

PR #9662, opened 2016-04-22
[PVR] Alternative pvr announce

janbar:
Alternative to handle announcements to pvr addon. See PR #9654.

@ksooo @Jalle19 @FernetMenta

### DIFF
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h
@@ -77,10 +77,10 @@ struct DemuxPacket;
 #define PVR_STREAM_MAX_STREAMS 20
 
 /* current PVR API version */
-#define XBMC_PVR_API_VERSION "5.1.1"
+#define XBMC_PVR_API_VERSION "5.2.0"
 
 /* min. PVR API version */
-#define XBMC_PVR_MIN_API_VERSION "5.1.1"
+#define XBMC_PVR_MIN_API_VERSION "5.2.0"
 
 #ifdef __cplusplus
 extern "C" {
@@ -602,6 +602,10 @@ extern "C" {
     bool         (__cdecl* IsTimeshifting)(void);
     bool         (__cdecl* IsRealTimeStream)(void);
     PVR_ERROR    (__cdecl* SetEPGTimeFrame)(int);
+    void         (__cdecl* OnSystemSleep)(void);
+    void         (__cdecl* OnSystemWake)(void);
+    void         (__cdecl* OnPowerSavingActivated)(void);
+    void         (__cdecl* OnPowerSavingDeactivated)(void);
   } PVRClient;
 
 #ifdef __cplusplus
--- a/xbmc/pvr/PVRManager.cpp
+++ b/xbmc/pvr/PVRManager.cpp
@@ -113,23 +113,34 @@ CPVRManager::~CPVRManager(void)
 
 void CPVRManager::Announce(AnnouncementFlag flag, const char *sender, const char *message, const CVariant &data)
 {
-  if (!IsStarted() || (flag & (System)) == 0)
+  if (!IsStarted())
     return;
 
-  if (strcmp(message, "OnWake") == 0)
+  if ((flag & (ANNOUNCEMENT::System)))
   {
-    /* start job to search for missing channel icons */
-    TriggerSearchMissingChannelIcons();
+    if (strcmp(message, "OnWake") == 0)
+    {
+      /* start job to search for missing channel icons */
+      TriggerSearchMissingChannelIcons();
 
-    /* continue last watched channel */
-    ContinueLastChannel();
+      /* continue last watched channel */
+      ContinueLastChannel();
 
-    /* trigger PVR data updates */
-    TriggerChannelGroupsUpdate();
-    TriggerChannelsUpdate();
-    TriggerRecordingsUpdate();
-    TriggerEpgsCreate();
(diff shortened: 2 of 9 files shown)


janbar:
It works with addon pvr.mythtv, and it will work too with others after adaptation.

xbmc/pvr/PVRManager.cpp
```diff
@@ -113,23 +113,38 @@ CPVRManager::~CPVRManager(void)
...
 void CPVRManager::Announce(AnnouncementFlag flag, const char *sender, const char *message, const CVariant &data)
 {
-  if (!IsStarted() || (flag & (System)) == 0)
+  if (!IsStarted())
     return;
 
-  if (strcmp(message, "OnWake") == 0)
+  if ((flag & (ANNOUNCEMENT::System)))
+  {
+    if (strcmp(message, "OnSleep") == 0)
```
FernetMenta: this is not reliable. there is no guarantee that async announce messages get delivered before system gets down. you may get this message on wakeup.
that why I said the the IAnnounce interface was abused for those purposes. System behaviour should not depend on messages coming through announce.

Better do something like g_PVRManager.OnSleep/OnWake/etc

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h
```diff
@@ -77,10 +77,10 @@ struct DemuxPacket;
 #define PVR_STREAM_MAX_STREAMS 20
 
 /* current PVR API version */
-#define XBMC_PVR_API_VERSION "5.1.1"
+#define XBMC_PVR_API_VERSION "6.0.0"
```
FernetMenta: no sure if we should bump major

MartijnKaijser: I'd say 5.2.0 should be enough as it's not a total breaking change

ksooo: I suggested this, but I think I was wrong. Minor bump should be the right thing to do for new api functions, that are actually called by Kodi, right?

xbmc/pvr/PVRManager.cpp
ksooo: @janbar Yeah, unfortunately, changing it the way you did would have exactly the problem we are trying to solve with removal if the old api function.

janbar:
When I have a couple of minutes , I will update it regarding your comments.

xbmc/pvr/PVRManager.cpp
janbar: Not exactly. Here addon functions are called by the owner.

xbmc/pvr/PVRManager.cpp
```diff
@@ -113,23 +113,34 @@ CPVRManager::~CPVRManager(void)
...
-    /* continue last watched channel */
-    ContinueLastChannel();
-
-    /* trigger PVR data updates */
-    TriggerChannelGroupsUpdate();
-    TriggerChannelsUpdate();
-    TriggerRecordingsUpdate();
-    TriggerEpgsCreate();
-    TriggerTimersUpdate();
+    if (strcmp(message, "OnWake") == 0)
```
Jalle19: Shouldn't this be moved to the newly introduced `OnWake()` method?

xbmc/pvr/addons/PVRClients.cpp
```diff
@@ -1615,3 +1615,43 @@ void CPVRClients::ConnectionStateChange(int clientId, std::string &strConnection
...
   }
 }
+
+void CPVRClients::OnSystemSleep()
+{
+  PVR_CLIENTMAP clients;
+  GetCreatedClients(clients);
+
+  /* propagate announcement to each client */
+  for (PVR_CLIENTMAP_CITR itrClients = clients.begin(); itrClients != clients.end(); itrClients++)
```
Jalle19: `for (auto &client : clients)`

xbmc/pvr/PVRManager.cpp
```diff
@@ -466,6 +477,22 @@ bool CPVRManager::SetWakeupCommand(void)
   return false;
 }
 
+void CPVRManager::OnSleep()
+{
+  if (!IsStarted())
```
FernetMenta: IsStarted is wrong. PVRmanager does not own the addons anymore

janbar: Here the test checks is the PVRManager has fully started, not addon. What do you think to check instead ?

xbmc/pvr/PVRManager.cpp
janbar: Depends. It designed like this for some reason I don't know.

xbmc/pvr/PVRManager.cpp
FernetMenta: nothing to check here. PVRManager's state is of no interest.

xbmc/pvr/PVRManager.cpp
FernetMenta: I would leave this here. It is something that can be executed async.

janbar:
Still long story with pvr addons. I will prepare their PR in the wake.

janbar:
So now all is ready to be triggered.

FernetMenta:
Thank you very much!
jenkins build this please

ksooo:
Thanks @janbar for this and the monkey work with the add-ons.

stefansaraev:
it seems you forgot to update addons/xbmc.pvr/addon.xml

PR #9663, opened 2016-04-22
Haptics and Controller power-off

garbear:
This PR activates controller rumble motors when the user receives a notification. Also, thanks to Montellese, we have the ability to power off controllers on exit.

Two settings are added:
- "Test rumble"
- "Power off controllers on exit"

This PR currently relies on the [retroplayer](https://github.com/kodi-game/peripheral.joystick/tree/retroplayer) branch of peripheral.joystick. These changes will be merged to master when this PR is merged.

ATM, rumble and power-off-on-exit are only supported via the XInput API on windows. Additional APIs can be added at a later date independent of Kodi.

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -15950,7 +15950,19 @@ msgctxt "#35050"
 msgid "Controller profiles"
 msgstr ""
 
-#empty strings from id 35051 to 35054
+#. Setting to test rumble
+#: system/settings/settings.xml
+msgctxt "#35051"
+msgid "Test rumble"
+msgstr ""
+
+#. Description of setting to test rumble
+#: system/settings/settings.xml
+msgctxt "#35052"
+msgid "Activate the rumble motors of all controllers."
+msgstr ""
+
+#empty strings from id 35053 to 35054
 
 #. Help text for controller window
 #: xbmc/games/controllers/windows/GUIControllerWindow.cpp
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -16024,7 +16024,19 @@ msgctxt "#35064"
 msgid "Pair your controllers with the various input devices of different game systems."
 msgstr ""
 
-#empty strings from id 35065 to 35089
+#empty strings from id 35065 to 35087
+
+#. Name of the setting that enables/disables powering off controllers on exit
+#: system/settings/settings.xml
+msgctxt "#35088"
+msgid "Power off controllers on exit"
+msgstr ""
+
+#. Description for the setting that enables/disables powering off controllers on exit
+#: system/settings/settings.xml
+msgctxt "#35089"
+msgid "Powers off any controller that supports it on exit."
+msgstr ""
 
 #. Button prompt without timeout. %s - button name
 #: xbmc/games/controllers/guicontrols/GUIScalarFeatureButton.cpp
--- a/xbmc/dialogs/GUIDialogKaiToast.cpp
+++ b/xbmc/dialogs/GUIDialogKaiToast.cpp
@@ -19,6 +19,7 @@
  */
 
 #include "GUIDialogKaiToast.h"
+#include "peripherals/Peripherals.h"
 #include "threads/SingleLock.h"
 #include "utils/TimeUtils.h"
 
@@ -133,6 +134,9 @@ bool CGUIDialogKaiToast::DoWork()
 
     //  Play the window specific init sound for each notification queued
     SetSound(toast.withSound);
+
+    // Activate haptics for this notification
+    PERIPHERALS::g_peripherals.OnUserNotification();
 
     ResetTimer();
     return true;
--- a/xbmc/input/joysticks/RumbleGenerator.cpp
+++ b/xbmc/input/joysticks/RumbleGenerator.cpp
@@ -31,10 +31,23 @@ using namespace JOYSTICK;
 CRumbleGenerator::CRumbleGenerator(const std::string& controllerId) :
   CThread("RumbleGenerator"),
   m_motors(GetMotors(controllerId)),
-  m_receiver(nullptr)
+  m_receiver(nullptr),
+  m_type(RUMBLE_UNKNOWN)
 {
 }
 
+void CRumbleGenerator::NotifyUser(IInputReceiver* receiver)
+{
+  if (receiver && !m_motors.empty())
+  {
+    if (IsRunning())
+      StopThread(true);
+
+    m_receiver = receiver;
+    m_type = RUMBLE_NOTIFICATION;
+    Create();
+  }
+}
 bool CRumbleGenerator::DoTest(IInputReceiver* receiver)
 {
   if (receiver && !m_motors.empty())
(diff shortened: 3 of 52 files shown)


xbmc/peripherals/addons/AddonInputHandling.cpp
```diff
@@ -39,15 +42,25 @@ CAddonInputHandling::CAddonInputHandling(CPeripheral* peripheral, IInputHandler*
   {
     m_buttonMap.reset(new CAddonButtonMap(peripheral, addon, handler->ControllerID()));
     if (m_buttonMap->Load())
-      m_driverHandler.reset(new CInputHandling(handler, m_buttonMap.get()));
-    else
-      m_buttonMap.reset();
```
Montellese: Did you drop this reset on purpose?

garbear: it may have been a rebase error. lemme check it out

garbear: Indeed, a rebase error. Will fix

xbmc/input/joysticks/RumbleGenerator.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  std::vector<std::string> motors;
+
+  AddonPtr addon;
+  if (CAddonMgr::GetInstance().GetAddon(controllerId, addon, ADDON_GAME_CONTROLLER))
+  {
+    ControllerPtr controller = std::static_pointer_cast<CController>(addon);
+    if (controller->LoadLayout())
+    {
+      for (const CControllerFeature& feature : controller->Layout().Features())
```
Montellese: `for (const auto& feature ...)`

xbmc/dialogs/GUIDialogKaiToast.cpp
```diff
@@ -134,6 +135,9 @@ bool CGUIDialogKaiToast::DoWork()
     //  Play the window specific init sound for each notification queued
     SetSound(toast.withSound);
 
+    // Activate haptics for this notification
+    PERIPHERALS::g_peripherals.OnUserNotification();
```
Montellese: Maybe this should only be done when the notification doesn't follow right after a previous notification. Otherwise when e.g. updating Kodi to a new version and it start updating all add-ons the controller won't stop rumbling for quite a while which certainly also affects its battery.

xbmc/input/joysticks/RumbleGenerator.cpp
garbear: `auto` is a c++11 feature I tend to use, but on the rare occurrence it actually replaces a single bare class, I prefer the class for clarity

xbmc/dialogs/GUIDialogKaiToast.cpp
garbear: where is a better place to place this fn call?

"Add rumble on Kia Toast notification " where is the typo?

Montellese: I'd have to take a closer look at the code myself.

It's called "Kai Toast" and not "Kia Toast"

Montellese:
BTW the Android API supports vibration patterns, see http://developer.android.com/reference/android/os/Vibrator.html. Maybe that might be something that would be useful for other stuff as well? I don't really know how libretro provides rumble events though.

Hedda:
Will haptics (rumble) have GUI options to be disabled and enabled? Maybe even have GUI settings to enable or disable rumble for different things? I know that at least I would not want to have rumble enabled for notifications by default but I would like to have it enabled in games by default.

Montellese:
@garbear is there a branch in `peripheral.joystick` with just the rumble stuff in it? I wanted to update my "turn off controller on shutdown" work and thought it best to do it on top of this work but I also need to update `peripheral.joystick`.

Montellese:
nvm I pulled together the necessary commits, squashed them as good as possible and added mine on top, see https://github.com/Montellese/peripheral.joystick/commits/joystick_poweroff in case you are interested in the squashed commits. And https://github.com/Montellese/xbmc/commits/joystick_shutdown has my commits on top of this PR.

garbear:
@Hedda I'm wary of settings creep, but I can see your point, it might be annoying if all 4 controllers rumble every time a notification is shown. What if we drop the commit and only rumble in games? I added rumble on notifications as more of a proof of concept anyway.

@Montellese you're right, the rumble command should go somewhere else in case multiple notifications are queued. Thanks for the power off support, should I tack that on and make this a dual-feature PR?

Montellese:
You can include it here or I can provide it as a separate PR once this has gone in, I don't really care.

garbear:
OK, I'll save you the trouble and add it here.

So what's the consensus? Do we drop rumble-on-notification for now? Once rumble is in it'll be trivial to add it later. Also, rumble can be made less annoying by shortening the duration or lowering the intensity (motors support values from 0.0 to 1.0)

Montellese:
I'd either make the rumble on notifications configurable or remove it completely.

xbmc/peripherals/addons/PeripheralAddon.cpp
```diff
@@ -71,7 +72,20 @@ CPeripheralAddon::~CPeripheralAddon(void)
...
   for (const auto& peripheral : m_peripherals)
+  {
+    if (CSettings::GetInstance().GetBool(CSettings::SETTING_INPUT_CONTROLLERPOWEROFF))
+    {
+      // shutdown the joystick if it is supported
+      if (peripheral.second->Type() == PERIPHERAL_JOYSTICK)
+      {
+        CPeripheralJoystick* joystick = static_cast<CPeripheralJoystick*>(peripheral.second);
+        if (joystick->SupportsPowerOff())
+          PowerOffJoystick(joystick->RequestedPort());
```
garbear: Should be `PowerOffJoystick(peripheral.first);` - RequestedPort() can be -1. Will fix and squash into PR

garbear:
I'll add a setting. The input page in system settings is getting a little crowded, but this is fixed in my RetroPlayer work - all controller settings are moved to a "game input" page

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -15568,7 +15568,19 @@ msgctxt "#35050"
 msgid "Controller profiles"
 msgstr ""
 
-#empty strings from id 35051 to 35054
+#. Setting to test rumble
```
da-anda: not sure that it's obvious to translators that read this description out of context. They neither see the settings category this setting is in, nor that it's related to input devices. So I think the settings description should be a bit more descriptive and also incluude that "rumble" is related to input devices.

Maybe: `Label of a button to test the rumble feature of input devices like game controllers` ?

garbear:
jenkins build this please (to test if I rebased on alpha 2 properly)

MartijnKaijser:
@garbear nudge

garbear:
The major thing stopping this was a crash if you disconnected the controller shortly after connecting it. My wireless 360 controller had low batteries, and when it connected Kodi told it to rumble and the rumble drew too much power and killed the controller and crashed Kodi :)

Fixing this will require a bit of refactoring. I rebased this patch on master last night, I'll try to look into the crash soon.

MartijnKaijser:
@garbear well v16 crashed when even a controller was connected. 
Also you can now drop the visual Studio files as windows now uses cmake

garbear:
Rebased on master. I'll look into the crash

garbear:
Fixed the crash. jenkins build this please

garbear:
@MartijnKaijser ok to merge?

nille02:
Is there a way to disable the rumble? its horrible for each notification.

PR #9664, opened 2016-04-22
guilib: only dialogs are allowed to call ProcessRenderLoop

FernetMenta:
Fix and another step in splitting the render thread. Only guilib is allowed to call ProcessRenderLoop. This change enforces this. It also reverts https://github.com/xbmc/xbmc/commit/fcc109044eef551aee72a50c5320fd8847e9455b

@mapfau could you test this please?

Some explanation:
The render loop has a sequence: FrameMove, Render, Flip, FrameMove, ...

If we allow calling the loop outside of guilib, i.e. by a method called from FrameMove, something like this may happen:

FrameMove (partly), FrameMove, Render, ...


mapfau:
works perfect for my issue! Thanks, @FernetMenta

FernetMenta:
jenkins build this please

xbmc/pvr/windows/GUIWindowPVRBase.cpp
```diff
@@ -139,37 +139,7 @@ void CGUIWindowPVRBase::OnInitWindow(void)
...
-          g_windowManager.PreviousWindow();
-          return;
-        }
-
-        g_windowManager.ProcessRenderLoop(false);
-      }
-
-      dialog->Close();
-    }
+    return;
```
Jalle19: What happens to the logic removed here?

FernetMenta: it was obsolete anyway.

FernetMenta:
@Memphiz iOS fails in IOSKeyboardView. Could you have a look please. Why does it not use a BusyDialog?

tamland:
Will something break from this? I'm assuming there was a reason for adding fcc1090

FernetMenta:
Possible but hard to tell. This change was done 4 years ago. If anybody notices an issues, please let me know. We'll find a better solution.

xbmc/pvr/windows/GUIWindowPVRBase.cpp
Jalle19: Do you mean obsolete as in not used anymore?

tamland:
lets ping @jimfcarroll then

xbmc/pvr/windows/GUIWindowPVRBase.cpp
FernetMenta: this is dead code

Memphiz:
may have time tomorrow to look at it - gui keyboard ios needs to call renderloop because its like a modal dialog and if you for example enter something in a live filter you want kodi to react on it while keyboard is still open...

Memphiz:
there are also some comments in the code there https://github.com/xbmc/xbmc/blob/master/xbmc/platform/darwin/ios/IOSKeyboardView.mm#L252

FernetMenta:
@Memphiz is this required to run on the main/render thread?

phil65:
can confirm that this fixes the visual glitches related to plugins. Will keep this in my branch and see if I notice any regressions.
Thx!

Memphiz:
@FernetMenta well its by design i think - see

https://github.com/xbmc/xbmc/blob/master/xbmc/guilib/GUIKeyboardFactory.cpp#L114

our generic keyboard dialog is a modal dialog in our gui thread aswell - that GetInput method is meant to be blocking...

FernetMenta:
@Memphiz then I think we should create something like CGUIDialogKeyboardIOS instead of emulating a dialog outside guilib. This dialog could embed OSKeyboard. As far as I can tell this should be doable. What do you think?

Memphiz:
@FernetMenta wouldn't this mean we need to add a skin xml file for this "fake" window? Not sure if this is what we want really? Not even sure if what i would need to skin in that xml - actually we don't want to see any window from the guilib with this do we? (maybe a 1x1 pixel window which is covered by the native keyboard then - not sure).

FernetMenta:
@Memphiz it should be a modal dialog because this is the behaviour we want. The skin file does not need to have any controls, means that we don't show anything but the IOS keyboard.
I think the current implementation of the ios keyboard has flaws because it isn't a modal dialog. That means other windows can be opened behind the keyboard triggered by API calls from outside.

Memphiz:
@FernetMenta am i right that you would call this approach here abusive?

https://github.com/Memphiz/xbmc/commit/f71043f1c57d2aced09960d42487fff42258ab8f

FernetMenta:
@Memphiz those changes are in the context of splitting the render thread from the main thread. The outside world should not block or drive the render loop.

Memphiz:
args time is over already ... please simply comment out the call to processrenderloop in case you need to go forward quick here. Keyboard should still work like before. Only the live filtering will not work anymore. (but it will update after keyboard is closed)

mapfau:
Line168, ProcessRenderLoop, leads to glitches commented in various posts.
I don't think simply enabling this line again is the solution here.

Enabling this line again will lead to development needs in GUI (opening VideoNav for example), Removing this line will lead to development in keyboard things. If I understand right the ProcessRenderLoop call at this place is really part of a concept, its more a workaround, right?

Isn't there a solution for the ios keyboard wich follows the concept @FernetMenta described?

da-anda:
as for the "native keyboard overlay" used by iOS atm, if this is moved to some GUI component, I think it would be nice to implement it (or at least name it) in some generic way so that other OSes (like Android) could also make use of it. Just an idea.

Memphiz:
Well it already was designed in a generic way. Thats why we have the CGUIKeyboard class (which needs to be derived and implemented for each natvie keyboard platform). Moving the ios native keyboard over the a GUIDialog won't change things that much - you still need to implement the CGUIKeyboard interface for android ...

FernetMenta:
@Memphiz I embedded the iOS keyboard into a dialog.
jenkins build this please

FernetMenta:
jenkins build this please

Memphiz:
That looks pretty clean - thx!

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

MilhouseVH:
Would this change be responsible deadlocking the Kodi GUI when clicking on "Search..." > "Search YouTube" from Estuary Home screen?

http://forum.kodi.tv/showthread.php?tid=269815&pid=2328176#pid2328176

MilhouseVH:
This is a "crashlog" obtained by executing `kill -SIGSEGV $(pidof kodi.bin)` on the deadlocked process: http://sprunge.us/dgJW

PR #9667, opened 2016-04-23
[pvr] - fix deadlock in epg

FernetMenta:
see title


ksooo:
Sigh, have seen this anti pattern (notify observers with locked mutex) so many times in our code base and fixed a couple of those already myself. Thanks, Rainer. Definitely +1.

PR #9671, opened 2016-04-23
[cmake/linux] CMake buildsystem & binary addons

fetzerch:
This simplifies developing binary addons on linux. It allows to build an addon against the local kodi build directory (which contains a specially prepared `kodi-config.cmake`) and copies the compiled addon to the build directory, so that it can directly be loaded. 

In addition, it fixes the generation of the installed `kodi-config.cmake` which had incorrect paths and compiler flags before. (ping @hudokkow since you touched that area in your clean up work)

This has been written by @notspiff and been discussed in https://github.com/fetzerch/xbmc/pull/2. I've done some cleanups and testing. PR is possibly interesting for @wsnipex.

@notspiff: I've done some testing with the this [script](https://gist.github.com/fetzerch/073339a41654e32ca9f81b573e2e10ab). The [log](https://www.zerobin.net/?8b9230be247e333e#9UI6MvYT890uOtHazVaCtv0INs9npyyIpaEUsGLkdRU=) shows that some addons fail to include `kodi/libXBMC_addon.h` because the include dir is set to `BUILDDIR/include/kodi/`. This looks like a bug in the addons to me which just happens to work with a `/usr`-prefix where the compiler uses the default `/usr/include`+`kodi/libXBMC_addon.h`. Would it make sense to add both paths to `kodi-config.cmake`? `BUILDDIR|PREFIX/include/kodi` and `BUILDDIR|PREFIX/include` or should we have it fixed it the failing addons?


project/cmake/scripts/linux/install.cmake
```diff
@@ -14,35 +14,36 @@ else()
...
                ${CORE_BUILD_DIR}/scripts/${APP_NAME_LC}-standalone @ONLY)
 
-install(TARGETS ${APP_NAME_LC} DESTINATION ${libdir}/kodi)
+# cmake config
+set(APP_LIB_DIR ${CMAKE_INSTALL_PREFIX}/lib/${APP_NAME_LC})
+set(APP_PREFIX ${CMAKE_INSTALL_PREFIX})
+set(APP_INCLUDE_DIR ${CMAKE_INSTALL_PREFIX}/include/${APP_NAME_LC})
+set(CXX11_SWITCH "-std=c++11")
+configure_file(${PROJECT_SOURCE_DIR}/kodi-config.cmake.in
+               ${CORE_BUILD_DIR}/scripts/${APP_NAME_LC}-config.cmake @ONLY)
```
fetzerch: The original version had the config generated to `BUILDDIR/kodi-config.cmake`. I decided to hide it in the scripts folder since it's only used for installation. When building a addon with the build dir, the user has to set `-DCMAKE_PREFIX_PATH` to `BUILDDIR/build`. It's just too easy to just to forget the `build` which would lead to the wrong config being picked up and weird errors.

hudokkow:
:+1: from me.

Question: what's the goal with core vs addon build systems? Keep them separated? I ask because right now we're duplicating logic in a bunch of places. See https://github.com/xbmc/xbmc/blob/master/project/cmake/scripts/common/prepare-env.cmake#L1 vs https://github.com/xbmc/xbmc/blob/master/project/cmake/scripts/common/macros.cmake#L365 for instance. And platform specific https://github.com/xbmc/xbmc/blob/master/project/cmake/scripts/linux/archsetup.cmake vs https://github.com/xbmc/xbmc/blob/master/project/cmake/addons/CMakeLists.txt#L148. And the list goes on.

[This] (https://github.com/hudokkow/xbmc/commits/cmake) is what I have been working on. Trying to unify core/addons build systems *and* keep autotools system happy at the same time (until we ditch it). Not an easy task. ;)

Autotools and cmake are building fine and generating config correct config files. However, @notspiff's solution to generate kodi-config.cmake is much simpler but we continue to have the same problem: two build systems with differing variable names that are set to same (APP_ROOT vs CORE_SOURCE_DIR, for instance) and duplicate logic all over the place.

How shall we proceed?

Non-intrusive fixes PR in https://github.com/xbmc/xbmc/pull/9676

FernetMenta:
I always build my binary addons against my local sources of kodi. The output of the addon build is copied     to my local addons folder. What would change for me? Why would I need this?

fetzerch:
@hudokkow: Good question. Personally I think it'll make it easier for us in the longrun if we align both build systems. Avoiding code duplication and calling the same vars by the same names is always a good idea. I've seen your changes, and that's a very welcomed start. It's probably also interesting for you to read https://github.com/fetzerch/xbmc/pull/2 (or maybe you knew the history already).

@FernetMenta: Idea here is that you build kodi with cmake and then build an addon against the local build (where the kodi-config.cmake is generated to). You use `cmake -DKODI_BUILD_DIR=<kodi build dir> -DCMAKE_PREFIX_PATH=<kodi build dir>/build <myaddon source dir>` and the addon build will then automatically install the addon into the kodi tree. When you say that you build against the local sources, does that work without having kodi or the depends compiled? If so, how? :)

FernetMenta:
> When you say that you build against the local sources, does that work without having kodi or the depends compiled? If so, how? :)

On Linux there is no need to compile depends, on OSX I have never tried without proper depends in place. There is no reason to compile an addon against a dev tree without the intent to run this build.

As a developer I don't want a copy of my local kodi/system folder in any other tree but my source tree. Are you saying that cmake creates a copy of this folder any installs the addon to the copy?

notspiff:
A few advantages;
1) no need to relate to the build system in project/cmake/addons and its
complexity (setting ADDON_SRC_ROOT, misusing -DPACKAGE_ZIP and all that).
this is a major source of confusion and has made everybody believe things
are much more complex than necessary. this build system is complex because
it has to work on all platforms, it's not a development tool. if i had to
develop my 40 add-ons using this i'd go bonkers in no time.

you just have to relate to 1 build system for the add-on, and one build
system for kodi. no wrappers, each of the build systems "just work" and
does their thing.

2) developers only rely on common cmake/unix knowledge.

what copies are you talking about? there is no copying. all this does it
prepare an add-on development environment in the kodi build folder.

FernetMenta:
> what copies are you talking about?

currently my addon build systems installs the output to kodi/system/addon folder that maps at runtime to special://system. When I change i.e. settings.xml in the source tree, the executable will notice this change. Same with language or other resources in the system folder. 
I need a dev environment for Kodi and addons at the same time. If this is given, I am fine.

akva2:
this does exactly the same except it doesn't do stupid things like write stuff to the source tree. it instead puts in the build directory as it should. the cmake builds mirrors everything in the build tree, so you can just execute from the top of the build tree. the only slight disadvantage is that if you change a resource file, you have to run make in the kodi build to re-mirror the updated file. the good thing is that you get every advantage of out-of-source trees (see .gitignore to start getting an idea why in-source builds is horrible).

this is how it works:

- i build kodi in $KODIDIR
- i build the addon in $ADDONROOT/build passing -DCMAKE_PREFIX_PATH=$KODIDIR/build -DDKODI_BUILD_DIR=$KODIDIR
- i run 'make' in $ADDONROOT/build. it builds, and mirrors the results in $KODIDIR/addons
- i execute kodi from $KODIDIR and test shit.
- i change something in the addon. i re-run make. it is re-mirrored. i can just re-launch the add-on in kodi and it will use the new copy. no fuss.

in all cases nothing here breaks your old way of doing stuff in any way.

FernetMenta:
Thanks for clarification. I am fine with that.

I would have more experience with cmake if my primary dev machine was not a mac. Unfortunately the generated XCode project is not really usable. Would it be possible that the project tree looked the same as the directory structure?

akva2:
everything's possible. currently the cmake build system is modeled on the autotools layout with separate .a files for each directory. it's not hard to change this at all as such but it would mean a less modular layout than what was there earlier for the makefile build.

FernetMenta:
If I had the skills and my list of todos was not that long, I would take the task and make cmake output the same project tree as the file system for XCode and VS. Once this is done, sunset autotools.

akva2:
well i think IDEs are a giant waste of time so i won't put in any effort either ;)

hudokkow:
@fetzerch, thanks for the clarification and yes, I've read https://github.com/fetzerch/xbmc/pull/2. Interesting read, btw.
Unless there's a reason to keep both systems separate, that's also my opinion.

What's the deadline to retire autotools to sunny Florida? L? L+1?

wsnipex:
there is no deadline. Its done when we feels cmake works at least as good as autotools on all platforms.
Imho there is not much missing and lately many more devs have picked it up, so we're moving at a good pace atm.

fetzerch:
I've got it on my todo list to look into better ide support, once packaging is done for all platforms.
Merging this as there are no more objections.

PR #9672, opened 2016-04-24
Add support for musl-based toolchains

bkuhls:
https://www.musl-libc.org/faq.html
`musl is a “libc”, an implementation of the standard library functionality described in the ISO C and POSIX standards, plus common extensions, intended for use on Linux-based systems.`

This PR contains patches downloaded from
https://github.com/voidlinux/void-packages/tree/master/srcpkgs/kodi/patches
which I tried to port to the current master branch. Using a buildroot-compiled
musl toolchain I could compile current Kodi master with these patches.
Unfortunately I am unable to do runtime testing.

Please note that I did not code the patches so I can not judge their technical
quality. Therefore I consider these patches, especially the fileemu patch,
as WIP to start a discussion.

### DIFF
--- a/xbmc/cores/DllLoader/exports/emu_msvcrt.cpp
+++ b/xbmc/cores/DllLoader/exports/emu_msvcrt.cpp
@@ -22,6 +22,7 @@
 #include <stdio.h>
 #include <stdarg.h>
 #include <math.h>
+#include <paths.h>
 #ifndef TARGET_POSIX
 #include <io.h>
 #include <direct.h>
@@ -477,13 +478,10 @@ extern "C"
     EmuFileObject* o = g_emuFileWrapper.GetFileObjectByDescriptor(fd);
     if (o)
     {
-      if(!o->used)
-        return NULL;
-
       int nmode = convert_fmode(mode);
       if( (o->mode & nmode) != nmode)
         CLog::Log(LOGWARNING, "dll_fdopen - mode 0x%x differs from fd mode 0x%x", nmode, o->mode);
-      return &o->file_emu;
+      return g_emuFileWrapper.GetStreamByFileObject(o);
     }
     else if (!IS_STD_DESCRIPTOR(fd))
     {
@@ -546,7 +544,7 @@ extern "C"
         return -1;
       }
       object->mode = iMode;
-      return g_emuFileWrapper.GetDescriptorByStream(&object->file_emu);
+      return g_emuFileWrapper.GetDescriptorByFileObject(object);
     }
     delete pFile;
     return -1;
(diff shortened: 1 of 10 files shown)


stefansaraev:
thanks!

@Karlson2k @notspiff any chance you can look at dlloader/emu stuff when you have time?

xbmc/cores/DllLoader/exports/emu_msvcrt.cpp
```diff
@@ -1633,7 +1631,7 @@ extern "C"
     int ret;
 
     ret = dll_fgetpos64(stream, &tmpPos);
-#if !defined(TARGET_POSIX) || defined(TARGET_DARWIN) || defined(TARGET_FREEBSD) || defined(TARGET_ANDROID)
+#if !defined(__GLIBC__) || defined(TARGET_DARWIN) || defined(TARGET_FREEBSD) || defined(TARGET_ANDROID)
```
wsnipex: does this mean that musl is not posix? Or that the else branch here is actually posix instead?

bkuhls: I hope my following assumptions are correct, my understand of the code is limited.

Musl follows the POSIX standard, the patch I posted tinkers around with the internal representation of different archs. TARGET_POSIX is defined by Kodi configure in m4/xbmc_arch.m4, based on this patchset this will also the case for musl toolchains but the patch author seems to want to use the non-POSIX code path in emu_msvcrt.cpp for whatever reason.

The problem is that Musl does not provide a macro like \__MUSL\__ to identify it:
http://wiki.musl-libc.org/wiki/FAQ#Q:_why_is_there_no_MUSL_macro_.3F
so other ways need to be found to treat Musl differently from glibc/uClibc. What about adding an internal define to m4/xbmc_arch.m4 like we do for FreeBSD, something like -D_MUSL
```
  i*86*-linux-gnu*|i*86*-*-linux-uclibc*|i*86*-*-linux-musl*)
     AC_SUBST(ARCH_DEFINES, "-DTARGET_POSIX -DTARGET_LINUX -D_LINUX -D_MUSL")
```
and patch emu_msvcrt.cpp like this?
```
-#if !defined(TARGET_POSIX) || defined(TARGET_DARWIN) || defined(TARGET_FREEBSD) || defined(TARGET_ANDROID)
+#if !defined(TARGET_POSIX) || defined(TARGET_DARWIN) || defined(TARGET_FREEBSD) || defined(TARGET_ANDROID) || defined(_MUSL)
```

wsnipex: ok, thats what I suspected. Lets use a the define as you suggested, cause I don't like tinkering with our TARGET_POSIX

Karlson2k:
@stefansaraev, will do.

Karlson2k:
Musl-based void Linux is not very stable yet.
Failed to run properly on VirtualBox and VMware.
Glibc-versions works fine.

bkuhls:
How to prevent spamming this PR while being able to mention it in my local patch file?

MartijnKaijser:
you just don't mention it al all or call it "PR 9672" or whatever. Have you seen how how times it already spammed here?

bkuhls:
@MartijnKaijser: Yes :( Sorry, I did not notice it to its full extend. Just found https\://gist.github.com/alexpeattie/4729247, I will try to use that from now on.

bkuhls:
Closed due to lack of interest.

PR #9659, opened 2016-04-21
[addon] binary add-on library rework for API level system

AlwinEsch:
**This request include a complete rework about the binary add-on library interface from them to kodi.**

It has as mean features and goal:
- Use only C++ on add-on itself to make it more easy and to prevent mem leaks
- Remove of all currently present shared libraries of add-ons
- Handle all calls from add-on to kodi only with "C" handled from independent over function table
- Reduce the dlopen call amount to only two, `Register` and `Unregister`
- Bring in a nearly equal function access to Kodi as on Python
- Add a complete good usable doxygen documentation for C++ and Python
- **And as main feature:** Support about API levels to allow also add-on with older versions

There is still a lot todo and maybe a lot to fix, but by the very big size of this request I want to bring it in and do a improvement with "Fix requests"

---

This request is the first step to change the interface from add-on to Kodi, as next steps in a new request comes rework about the Kodi to Add-on interface to use also the API level design.

---

As last request, if everything is in new level design, comes the add of a executable add-on system to use for everything like on Python.

---

Changes on add-on side for testing are visible here:
- pvr.vdr.vnsi: https://github.com/kodi-pvr/pvr.vdr.vnsi/pull/74
- adsp.freesurround: https://github.com/kodi-adsp/adsp.freesurround/pull/9
- pvr.demo - https://github.com/kodi-pvr/pvr.demo/pull/34
- pvr.iptvsimple - https://github.com/kodi-pvr/pvr.iptvsimple/pull/73

---

**The current link on http://esmasol.de/addon-dev-kit/ is for testing purpose only!!!**

PR #9652, opened 2016-04-20
Use CodeDocs for Doxygen documentation

paulnovo:
This PR adds a configuration file for [CodeDocs.xyz](https://codedocs.xyz), a GitHub integration that generates and hosts Doxygen documentation (Disclosure, I am one of the CodeDocs creators, and longtime Kodi user). Like Travis-CI it will regenerate after each update to the repo. I matched the configuration to Kodi's [Doxyfile.doxy](https://github.com/xbmc/xbmc/blob/master/doxygen_resources/Doxyfile.doxy).

I have this setup on my fork of Kodi now, and you can see the results on CodeDocs [here](https://codedocs.xyz/CodeDocs/xbmc/). The PR also adds a badge to the README.md, which you can see an example of, again, in my [Kodi fork](https://github.com/CodeDocs/xbmc/blob/master/README.md).

PR #9674, opened 2016-04-24
[addon] Reworked python docs

AlwinEsch:
Here the reworked documentation to use doxygen complete for Python.
Is taken from #9659 to separate it.

The generated Website for test purpose visible here: http://alwinesch.github.io/group__python.html


MartijnKaijser:
please remove the images

AlwinEsch:
I remove the screenshot, but if you want to remove also the Python logo, no this I want to let in. Language logos on start are for me required.

MartijnKaijser:
only meant the Kodi screenshots

AlwinEsch:
Is updated, has seen the table on addControl was not complete on old docs and added the missing ones there.

But has seen on http://alwinesch.github.io/group__python__xbmcgui__window.html#gafb8b45333ce0fbed3b7d88f4ed97cc68 `getControl(controlId)` a note who I not understand.
Original is here: https://github.com/xbmc/xbmc/blob/master/xbmc/interfaces/legacy/Window.h#L470-471
Have someone a alternative for them?

AlwinEsch:
A note:
If someone see a 404 on test website, is not a fault on Kodi. Whatever the reason,  it supports file name e.g. _addon__ rendering control.html not on github.io.

AlwinEsch:
A question has seen libraries are named modules in some docs.
Better change e.g. `Library - xbmc` to `Module - xbmc`?

MartijnKaijser:
Python is in general called module

MartijnKaijser:
Let's just merge this when window opens and fix any typos and unclear text afterwards.

AlwinEsch:
Okey Dokey

MartijnKaijser:
Current master comes out like this
https://codedocs.xyz/xbmc/xbmc/pages.html
with this PR it ends up like
https://codedocs.xyz/MartijnKaijser/xbmc/pages.html

AlwinEsch:
See my last comment here: https://github.com/xbmc/xbmc/pull/9652

CodeDocs does not use the most of the doxygen settings!

AlwinEsch:
Further use the doxygen for add-on documents this https://github.com/AlwinEsch/kodi/blob/reworked-python-docs/xbmc/addons/kodi-addon-dev-kit/doxygen/DoxygenLayout.xml

Is not major, but it correct and remove not wanted parts on the html.

xbmc/interfaces/legacy/Addon.h
```diff
@@ -56,78 +75,145 @@ namespace XBMCAddon
...
 
-      /**
-       * getLocalizedString(id) -- Returns an addon's localized 'unicode string'.
-       * 
-       * id        : integer - id# for string you want to localize.
-       * 
-       * example:
-       *   - locstr = self.Addon.getLocalizedString(32000)
-       */
+#ifdef DOXYGEN_SHOULD_USE_THIS
```
tamland: Why are these ifdefs still here? Should be unnecessary now that the old docs are removed, right?

tamland:
Remind me: why is everything using `///`? According to doxygen docs, it supports `/** ..*/` and it appeared to work fine with the old doxygen config.

AlwinEsch:
See here about `///` http://alwinesch.github.io/a00004.html.

Is to know in Kodi what is for them itself and with `///`what is also related to add-ons. There comes much inside and if something like GUIInfoManger.cpp has so much comments is good to see direct for where it is needed.

xbmc/interfaces/legacy/Addon.h
AlwinEsch: Is needed to become function names in a better Python style.

Without this use the newest doxygen always `String getLocalizedString(int id);` and with this `ifdef`s becomes it forded to `getLocalizedString(...)`

tamland: doxygen can't be adjusted to hide it? `getLocalizedString(...);` is not valid c++, so I'm suspecting this is a hack and that it may break in a future doxygen update.

AlwinEsch: Without them does it use `String getLocalizedString(int id);` what is not Python conform. To use a setting for the old style make no sense, there comes also documentation for C++ where the complete function is required.

tamland: I understand why it's happening. As was said in the other PR this is because you're trying to create python documentation from a c++ header.

Right now the entire file is in a weird dual mode where all code is ifdeffed out from doxygen. Is there even the point of having the docs in the same file now when code isn't used or attached to anything?

AlwinEsch: Yes but then it becomes not so good sorted with the function names and becomes complicated for updates. Currently it use the Module System from doxygen who is optimized for functions. The other way can be the Page System, but with a so big list of functions then very bad.

AlwinEsch:
Hello all, changes acceptable to bring in?

MartijnKaijser:
i'm fine with it if @tamland is. Typos and text improvements can be done later

tamland:
As I said the I really don't like the ifdef solution and the signature 'hack'. It all seems to work because of something that looks like a bug in doxygen. From a quick look there is several alternative standard ways to solve this problem: http://www.stack.nl/~dimitri/doxygen/manual/docblocks.html#structuralcommands

Also, doxygen is used by the generator to insert docstring into the python module. What happens with this? Will it be able to handle the `///` change?

AlwinEsch:
The problem for python docs is not related to the `///`, after one of the Standard doxygen interpreters e.g. `@param` becomes used, does it not generate the xml in a style compatible to python.

With the use of `\fn` does not prevent the problem about function names for python. Has also tried it now but not usable.

AlwinEsch:
Has checked it now.

As one of the xml problems for swig, if there is on docs as example a `@note` becomes it set on xml as <simplesect kind="note">...</simplesect>` but as them it becomes ignored from swig!

All values with a `@` or `\` ends as `simplesect` and if the first character on start of documentation is one of them becomes it complete lost on function.


Has tried also here as test with `xsltproc` is different but not better :(

AlwinEsch:
Hi @MartijnKaijser , @tamland an @phil65 need to know if acceptable.

I'm not a Python programmer and not know how much the before from swig inserted documents are used. Has only seen that it becomes hard to bring it detailed to them.

I know with this `#ifdef` is not good, but has not found another way to remove the C++ function (who confuse a Python programmer) on website. So long the documentation is on a `*.h` becomes it always from doxygen seen as `C` or `C++`

If not acceptable then I close this and continue only on the C++ add-on parts.

AlwinEsch:
jenkins build this please

phil65:
@tamland, perhaps a comment if this is acceptable?
I dont mind as long as we dont have to maintain several docs.

phil65:
too late. :)

AlwinEsch:
Hi @tamland, was one second before your comment was here :)

A main goal from me is to have for add-on development everything documented inside Kodi. Python was already there only not in a good style. 

With them in Kodi is it much more easy to change on code changes direct the documents from them like on #9704.

With the use also for C++ add-ons comes again a much. Also for GUI is a much. To have them only on Wiki makes it then more hard and hard.

PR #5957, opened 2014-12-16
Avoid compiling cdrip when build with --disable-optical-drive

dteteruk:
Avoid compiling unnecessary files (EncoderFFmpeg, CDDARipJob) when configure with --disable-optical-drive option.

PR #9675, opened 2016-04-24
Avoid compiling cdrip when build with --disable-optical-drive

MartijnKaijser:
Replaces https://github.com/xbmc/xbmc/pull/5957


MartijnKaijser:
jenkins build this please

PR #9676, opened 2016-04-24
[cmake] Some more fixes

hudokkow:
@fetzerch, some non intrusive fixes referenced in https://github.com/xbmc/xbmc/pull/9671


fetzerch:
Oh nice! I've been working on OSX packaging today and when I wanted to check why the git revision is not generated I saw your PR :)

https://github.com/xbmc/xbmc/compare/master...fetzerch:cmake_osx_packaging
The first commit 07a4789 also affects Linux, without it estuary's curial theme is not installed. If you want, you could cherry-pick it into your PR. I'll drop the second one because your solution of installing the xsd files is better.

Review wise, all your changes look sane to me.

hudokkow:
Cherry-picked your commit. Your button if you feel this is ready!

fetzerch:
Thanks, I've verified also the git rev change with my osx packaging changes. Works fine.

PR #9681, opened 2016-04-25
File cache improvements

arnova:
This PR brings a couple of improvements concerning our filecache. The most important change is the move of cache related settings in as.xml from the network-section to its own cache-section + proper (re)naming of the settings. The way it has evolved no longer made any sense (cache is not only used by network) and the names were rather cumbersome. I also added an additional caching mode ("remote") and did some cleaning/refactoring.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamFile.cpp
+++ b/xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamFile.cpp
@@ -67,14 +67,11 @@ bool CDVDInputStreamFile::Open()
    */
   if (!URIUtils::IsOnDVD(m_item.GetPath()) && !URIUtils::IsBluray(m_item.GetPath())) // Never cache these
   {
-    if (g_advancedSettings.m_networkBufferMode == 0 || g_advancedSettings.m_networkBufferMode == 2)
+    if ((g_advancedSettings.m_cacheBufferMode == CACHE_BUFFER_MODE_INTERNET && URIUtils::IsInternetStream(m_item.GetPath(), false))
+     || (g_advancedSettings.m_cacheBufferMode == CACHE_BUFFER_MODE_TRUE_INTERNET && URIUtils::IsInternetStream(m_item.GetPath(), true))
+     || (g_advancedSettings.m_cacheBufferMode == CACHE_BUFFER_MODE_ALL))
     {
-      if (URIUtils::IsInternetStream(CURL(m_item.GetPath()), (g_advancedSettings.m_networkBufferMode == 0) ) )
-        flags |= READ_CACHED;
-    }
-    else if (g_advancedSettings.m_networkBufferMode == 1)
-    {
-      flags |= READ_CACHED; // In buffer mode 1 force cache for (almost) all files
+      flags |= READ_CACHED;
     }
   }
 
--- a/xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamFile.cpp
+++ b/xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamFile.cpp
@@ -69,6 +69,7 @@ bool CDVDInputStreamFile::Open()
   {
     if ((g_advancedSettings.m_cacheBufferMode == CACHE_BUFFER_MODE_INTERNET && URIUtils::IsInternetStream(m_item.GetPath(), false))
      || (g_advancedSettings.m_cacheBufferMode == CACHE_BUFFER_MODE_TRUE_INTERNET && URIUtils::IsInternetStream(m_item.GetPath(), true))
+     || (g_advancedSettings.m_cacheBufferMode == CACHE_BUFFER_MODE_REMOTE && URIUtils::IsRemote(m_item.GetPath()))
      || (g_advancedSettings.m_cacheBufferMode == CACHE_BUFFER_MODE_ALL))
     {
       flags |= READ_CACHED;
--- a/xbmc/filesystem/FileCache.cpp
+++ b/xbmc/filesystem/FileCache.cpp
@@ -179,7 +179,7 @@ bool CFileCache::Open(const CURL& url)
 
   if (!m_pCache)
   {
-    if (g_advancedSettings.m_cacheMemBufferSize == 0)
+    if (g_advancedSettings.m_cacheMemSize == 0)
     {
       // Use cache on disk
       m_pCache = new CSimpleFileCache();
@@ -188,14 +188,14 @@ bool CFileCache::Open(const CURL& url)
     else
     {
       size_t cacheSize;
-      if (m_fileSize > 0 && m_fileSize < g_advancedSettings.m_cacheMemBufferSize && !(m_flags & READ_AUDIO_VIDEO))
+      if (m_fileSize > 0 && m_fileSize < g_advancedSettings.m_cacheMemSize && !(m_flags & READ_AUDIO_VIDEO))
       {
         // NOTE: We don't need to take into account READ_MULTI_STREAM here as it's only used for audio/video
         cacheSize = m_fileSize;
       }
       else
       {
-        cacheSize = g_advancedSettings.m_cacheMemBufferSize;
+        cacheSize = g_advancedSettings.m_cacheMemSize;
       }
 
       size_t back = cacheSize / 4;
@@ -296,13 +296,13 @@ void CFileCache::Process()
 
     while (m_writeRate)
     {
-      if (m_writePos - m_readPos < m_writeRate * g_advancedSettings.m_readBufferFactor)
+      if (m_writePos - m_readPos < m_writeRate * g_advancedSettings.m_cacheReadFactor)
       {
         limiter.Reset(m_writePos);
         break;
       }
 
-      if (limiter.Rate(m_writePos) < m_writeRate * g_advancedSettings.m_readBufferFactor)
+      if (limiter.Rate(m_writePos) < m_writeRate * g_advancedSettings.m_cacheReadFactor)
         break;
 
       if (m_seekEvent.WaitMSec(100))
--- a/xbmc/settings/AdvancedSettings.cpp
+++ b/xbmc/settings/AdvancedSettings.cpp
@@ -751,7 +751,7 @@ void CAdvancedSettings::ParseSettingsFile(const std::string &file)
   if (pElement)
   {
     XMLUtils::GetUInt(pElement, "memorysize", m_cacheMemSize);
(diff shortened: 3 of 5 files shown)


arnova:
jenkins build this please

arnova:
jenkins build this please

arnova:
@MartijnKaijser : You (or any others) against the change of the as.xml settings which obviously forces users to update their as.xml ?

MartijnKaijser:
Nope

NedScott:
The change actually has a very good positive in making people change their as.xml file, in that it makes many "bad guides" about cache settings outdated. While not the intention, it's a nice bonus ;)

I doubt this would be supported, but IMO this should be in the GUI, where the different settings can be explained. Ironically, keeping cache settings out of the GUI has actually caused more confusion because of the rise of various configuration "wizards", which don't tell the user which settings are safe. I also think it's used enough to be justified in the GUI. However, that's outside the scope of this PR, but I mention it in hopes that the idea will be more welcomed in the future :)

Any thoughts about changing the default to cache LAN traffic by default (SMB, NFS, etc)?

popcornmix:
> Any thoughts about changing the default to cache LAN traffic by default (SMB, NFS, etc)?

I believe that enabling the cache makes seeking a little slower, and as the cache isn't usually needed for wired ethernet that doesn't sound worthwhile. If wireless could be reliably detected it might make more sense there, but I suspect that is not something kodi can easily determine.

MartijnKaijser:
jenkins build this please

arnova:
@popcornmix / @NedScott : I already mentioned in the past that I'm all for enabling cache by default for all protocols except local. As an alternative we could move buffer-mode to GUI, but I would certainly not move mem-size and read-factor to GUI. @MartijnKaijser / others: Your take on this?

@popcornmix : It could be that seeks used to be slow due to the readratefactor hurting performance but since I upped it to 10 now it shouldn't be an issue anymore. Could you test whether it's still in fact true? We could perhaps even disable the limiter by default when we use CircularCache and just let it fill as fast as possible. It's a remnant of the Xbox-days anyways where hardware was limited and cache was on hdd.

arnova:
ps. Take into account that I believe most of our users use wireless nowadays, but perhaps I'm wrong here?

fritsch:
A factor of 10 drives the "average read" nuts. That basically means: Don't
throttle at all. Also reading with 100 Mbit/s an e.g. 10 Mbit/s stream also
works in combination with a large enough buffer. So it might be useful to
align the cache with average bitrate and the readbufferfactor.

2016-04-28 7:18 GMT+02:00 arnova <notifications@github.com>:

> ps. Take into account that I believe most of our users use wireless
> nowadays, but perhaps I'm wrong here?
>
> —
> <https://github.com/xbmc/xbmc/pull/9681#issuecomment-215312666>
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

popcornmix:
The cache does significantly increase memory usage and cpu (an extra two memcpy's on any data read from file I imagine).
I wonder if promoting this to gui settings would be useful? Then the default behaviour could be overridden by platform settings files, but it's still easy for users to switch?

arnova:
@fritsch : Mind elaborating that? I'm not sure what you mean.

fritsch:
How long does it need for a connection to get up to 100 MBit/s speed. How long can you download with that speed to fill a buffer of 5MB?

With a 10 Mbit/s movie - how large must the cache be to hold 10 seconds of a movie? And how long does this 100 Mbit/s need to run to fill those 10 seconds?

fritsch:
default_read_factor
(default setting)

factor10

(Factor 10)

You see what happens? The primary peak gets much larger, afterwards the buffer is full and nothing happens anymore. So what I said: higher speed is only helpful as long as you cache it.

This was playing Lord of the Rings via wireless network.

fritsch:
cache10largebuffer

(Factor 10 with ~ 200 MB Buffer via local network)

fritsch:
And here a 2gb cache:
2gb cache

It's filling with full speed. But after it was full once, it just continues to fill concerning the cache fill state:
2gb cache2nd

fritsch:
And finally to also show what tuning helps most (my wireless can do 30 MB/s while kodi just got 16). When increase the ChunkSize - in my case in NFS VFS implementation I can easily get these 30 MB/s.

So let's not optimize at the wrong position, please.

See for example: https://github.com/popcornmix/xbmc/commit/e79ab7879aa23ef81d46021973fe239035617c61 which made raw bluray via nfs possible on the PI2 back at that time. I will upload new figures with that patch applied tonight.

fritsch:
And here with the above VFS tuning, so if it is about throughtput - our VFS needs tuning and then we can think about caching, currently we loose nearly 50% with our read chunks logic on wireless connections, while the above tuning - proposed in this thread - has literally no impact in practice.

vfstuning

arnova:
@fritsch: Ok, I'll drop the readfactor change. We'll look into that part later on then.

arnova:
@fritsch / @MartijnKaijser / @popcornmix : What do we do with the buffer-mode setting? Keep it in as.xml or move it to GUI?

MartijnKaijser:
imo buffermode is something too advanced and that user shouldn't even have to deal with such a setting at all to start with. I am still of the opinion that if we need this setting we are doing something wrong

NedScott:
It's far less advanced than many settings we already have, and something being advanced should not be a reason to keep it out of the GUI now that we have advanced and expert GUI settings levels.

Like I said before, leaving it in as.xml has actually caused more problems because users apply settings blindly or from some "wizard". The cache settings are probably the most used as.xml setting right now, and moving this one setting to the GUI will help both the users and help them use the correct setting. 

This is not one of those times when you can say that we don't need a setting because that's doing it wrong. This is exactly the time when you do use a setting.  People have different networks, different needs, and Kodi has no practical way of detecting what they need (certainly not from a cross platform standpoint). Settings are not "bad".

MartijnKaijser:
That's your opinion.

arnova:
@MartijnKaijser : I partially agree with you but if we don't move this this to GUI (and again note I don't have a strong opinion about whether we should), we should work torwards enabling cache by default for any non-local filesystem. As an alternative we could add an option to the dialog when adding a new source. I've already looked at his before, but the problem with that it's non-trivial, but perhaps the most viable option?

NedScott:
@MartijnKaijser I think there is some confusion here. This is about buffermode (what to buffer: internet only, internet + LAN, nothing). This isn't about the fill rate or the cache size. The network speed can't be used to figure out what the user wishes to have cached, especially for issues such as wifi interference, which won't be detected until it's too late for video playback.

FernetMenta:
Those settings are beyond the skills of the average user, hence no gui setting. An advanced setting is not the way to go either. Having a setting only results in some try and error game.

arnova:
Well we don't have a proper alternative it seems, so for now we might as well stick to as.xml . Btw. I don't see how this results in an "try and error game". It's pretty clear what it does IMHO..

FernetMenta:
> Btw. I don't see how this results in an "try and error game". It's pretty clear what it does IMHO..

if it's pretty clear, why you we need a setting. Just auto set it to the best option.

arnova:
We can't since we don't know about the environment we're operating in, only the user knows.

arnova:
So you're proposing to get rid of this completely now?

NedScott:
Consider PVR settings, which are far much more advanced, but still necessary, because you cannot "auto" this kind of thing. It's not hard. It's basically asking the user if they want to:

Buffer everything

Buffer only internet

Buffer LAN + internet

Buffer nothing

How is that hard to understand? How would you automatically decide which is best for a user?


Not everyone has to agree or have the same viewpoint, but it's pretty clear that this is in both high need and high demand from the userbase. It's the same as when there was a debate about "play next video automatically", which some people thought had no place in the GUI (that feature was a "fan favorite" for that release, by the way).

Every other media player on my desktop and my tablet has cache settings, by the way. Claiming this is too advanced for users is nonsense. It's almost like this idea won't be considered out of spite, either for the user, or for myself for being involved in this discussion.

FernetMenta:
We can detect the environment and adapt to. The problem is that up to know not much devs did care which resulted in too many settings. Some even pretend AUTO but only get mapped to some other settings. I already told you about the new ProcessInfo object. The system can measure and learn about the environment. And an expert can do this much better than an average user.

@NedScott please get lost

NedScott:
@FernetMenta If you believe I have said something inappropriate for this discussion then I apologize. I'm still in charge of the Kodi wiki, a member of the community, a member of LibreELEC, and now an employee of WeTek. As such, I will be involved in this discussion and many more in the future. I can assure you that any communications will be relevant and appropriate for the situation.

That being said, unless you are volunteering to replace the current cache system with something better, then I don't understand your objection to what is being discussed here. It appears that the GUI setting is already too disputed, so all that is left is an improvement on the current cache settings (as.xml and defaults) that are already in place. Is there an objection to _that_?

FernetMenta:
@NedScott I don't care who ever made the big mistake to employ you. I don't consider you as a member of the community worth talking to. Github is a dev space and you are obviously no dev. Go trolling in the forums.

arnova:
@FernetMenta : He does have a valid question: will we include this until we have something better? I think this could take a while.... ?

FernetMenta:
@arnova I am not against this change here. I was just saying that we should not make this dependent on settings.
btw. paplayer would need this for local content because unlike VideoPlayer it has no 8 sec demuxer queue.

arnova:
@FernetMenta : What do you mean exactly by "this change" ? And didn't we previously always cache files for PAPlayer? Or did this get lost when we moved to DVDPlayer-codec?

FernetMenta:
> What do you mean exactly by "this change"

this PR

> And didn't we previously always cache files for PAPlayer

Some codecs did, others didn't. Same story here: there is no one-fits-all cache with a couple of parameters that should be configured by the user. Depending on the situation there are different requirements, that even change dynamically.

Instead of giving the controls to the user, we should give them to player, the components that use it. When going rw caching requirement are totally different from that when doing normal playback.

arnova:
@FernetMenta : We'll sort that out later on, right? Seems non-trivial.

Good to go?

FernetMenta:
yes

username155:
After this was merged, davs streams are no longer cached, please see this thread: http://forum.kodi.tv/showthread.php?tid=272973

arnova:
@username155 : Spotted the issue, fix will arrive soon.

arnova:
Fix here: https://github.com/xbmc/xbmc/pull/9795

In the meantime you could set buffermode to "2" to workaround the issue.

PR #9682, opened 2016-04-25
configure.ac: Add target_platform=target_linux for powerpc

bkuhls:
Checking for libuuid only occurs with target_platform set to a sane value. Without libuuid being added to LIBS, the check for libcrossguid fails because libcrossguid is linked to libuuid.


stefansaraev:
cc @wsnipex

wsnipex:
looks good

stefansaraev:
jenkins does not build this, so I'll just push it.

PR #9683, opened 2016-04-25
[cleanup] remove not need ogg/vorbis check from configure.ac

stefansaraev:
kodi has no direct dependency on libogg / libvorbis anymore. makes no sense to check those headers in configure

a side note: ffmpeg has native ogg decoder (my samples play fine with ffmpeg compiled without libogg/libvorbis). we dont use vorbisenc in ffmpeg anymore (audioencoder.vorbis exists), so it seems libogg/libvorbis can be removed from depends, too.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -963,8 +963,6 @@ if test "$use_mysql" = "yes"; then
   fi
 fi
 AC_CHECK_HEADER([ass/ass.h],, AC_MSG_ERROR($missing_library))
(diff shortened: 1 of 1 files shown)


bkuhls:
fyi: a cross-compile test using Kodi git master without ogg/vorbis being compiled before was successful.

stefansaraev:
closing in favour of #9683

PR #9590, opened 2016-04-09
Add xboxdrv name for xbox keymap

HerbFargus:
Driver is from this repo: https://github.com/xboxdrv/xboxdrv

PR #9684, opened 2016-04-26
Revert "Add xboxdrv name for xbox keymap"

MartijnKaijser:
Reverts xbmc/xbmc#9590


garbear:
I just suggested that the user from the other joystick keymap PR should send his keymap to the Jarvis branch. Is this not the case?

MartijnKaijser:
Was too late for 16.1 and won't be a 16.2

PR #9685, opened 2016-04-26
[doxygen] Fix and improve doxygen parts

AlwinEsch:


AlwinEsch:
OK to merge after travis is done now or better wait until 1.5.? There are no changes on code.

MartijnKaijser:
doc changes are always good for merge

MartijnKaijser:
@NedScott now this is doxygen we could remove the wiki page or wiki could generate it from code. What's best as having it in two places isn't good

NedScott:
For the short term, the best is to probably replace the wiki pages with a notice and link to the doxygen pages, so that old links still work and to help people find the new doxygen pages. There are some options for the long term that can be explored, but I'm hesitant to install anything new on our wiki until the server and mediawiki can be updated.

PR #9687, opened 2016-04-26
[doxygen] add code guidelines

AlwinEsch:
This add the guidelines from http://kodi.wiki/view/Official:Code_guidelines_and_formatting_conventions to Kodi itself.

Also becomes my guideline for Add-on related documents added.

Plan from me is to use for code from Kodi itself a `/*!` and for Code related also for add-on developers a `///`. The last mark it as field to know also needed alone on add-on development kit.

### DIFF
--- a/CODING_GUIDELINES.dox
+++ b/CODING_GUIDELINES.dox
@@ -1,0 +1,372 @@
+/*!
+
+\page code_guidelines Code guidelines and formatting conventions
+
+@brief \doc_header{ Code guidelines and formatting conventions }
+
+\tableofcontents
+
+These are conventions which we try to follow when writing code for Kodi. They
+are this way mainly for reasons of taste, however, sticking to a common set of
+formatting rules also makes it slightly easier to read through our sources. If
+you want to submit patches, please try to follow these rules.
+
+As such we don't follow these rules slavishly, in certain cases it is ok (and
+in fact favorable) to stray from them.
+
+================================================================================
+\section code_guidelines_1 Indentation
+
+Use spaces as tab policy with an indentation size of 2
+
+--------------------------------------------------------------------------------
+\subsection code_guidelines_1_1 Statements
+
+No multiple statements on a single line, like this:
+~~~~~~~~~~~~~
+std::vector<std::string> test; test.push_back("foobar"); // This is the bad way
+~~~~~~~~~~~~~
+
+Always use a new line for a new statement:
+~~~~~~~~~~~~~
+std::vector<std::string> test;
+test.push_back("foobar");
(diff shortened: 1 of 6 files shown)


AlwinEsch:
The uploaded test website visible here: http://alwinesch.github.io/a00006.html

AlwinEsch:
Is updated, @FernetMenta @Montellese @MartijnKaijser OK to merge in?

MartijnKaijser:
can you update the page?

CODING_GUIDELINES.dox
```diff
@@ -0,0 +1,335 @@
...
+--------------------------------------------------------------------------------
+\subsection code_guidelines_5_7 Variables
+
+We use CamelCase for variables. Type prefixing is optional.
+\subsubsection code_guidelines_5_7_1 Global Variables
+
+Prefix global variables with g_
+
+~~~~~~~~~~~~~
+int g_globalVariableA;
```
FernetMenta: no globals at all

FernetMenta:
maybe a native speaker volunteers for proof-reading

MartijnKaijser:
@garbear for proof read?

CODING_GUIDELINES.dox
AlwinEsch: Better to remove complete?

For my is this also a bad example, to use a editable integer as global is a worst way to use.
Only for me a safe example can be a class with thread safe functions.

AlwinEsch:
Test website is updated and a example about spaces added http://alwinesch.github.io/a00006.html#code_guidelines_3_1

CODING_GUIDELINES.dox
MartijnKaijser: I think it's better to include the statement that globals are not allowed just to be clear. Should any one touch part of the code they could even try to remove them

AlwinEsch: What think you about:
```
Global editable variables are not thread safe and not allowed to use!

If you need a minor global class should be the creation of them inserted in `./xbmc/ServiceManager.h` and `./xbmc/ServiceBroker.h`.

For a major class needed on early start of Kodi can be the Prefix global variables with g_ used.
```

FernetMenta: How about:

> Use of globals reduces the chance of submitted code to be accepted to a minimum

garbear:
I can proof read. When is this going to be merged? If I get a chance I'll send a PR to alwin before, otherwise I'll PR against master

AlwinEsch:
jenkins build this please

Not really needed but why not for safe. ;)

PR #9688, opened 2016-04-26
Chop out anything before Android 5 and don't build with amlogic

MartijnKaijser:
Reasoning:
- Android only started maturing after Android 5.0 as well as the devices
- AMLogic was the main codec used in the initial Android port however now it's now time to follow the official Android API. AMLogic doesn't completely follow this and that causes a lot of IFDEF / hacks to support both.
- We are too low on manpower to keep supporting hacks on this platform and the only sane way forward is following the official Android API.

The code is still there for now. This patch will simply stop compiling AMLogic for the Android platform. Should no one pick this up it will be completely removed for Android but could still be used for Linux. AMLogic code needs a complete rewrite anyway.


MartijnKaijser:
fyi we should include this in Alpha1

PR #9690, opened 2016-04-26
[droid] [refclock] do not stop callback when focus is lost. 

MartijnKaijser:


MartijnKaijser:
jenkins build this please 
@fritsch ok for merge?

fritsch:
As we got no complains before that patch - let's see if someone complains after it :-)

PR #9691, opened 2016-04-27
cpluff: Build ExternalProject in source

smspillaz:
CPluff doesn't like to be built out of source.

Fixes #16707

### DIFF
--- a/project/cmake/modules/FindCpluff.cmake
+++ b/project/cmake/modules/FindCpluff.cmake
@@ -11,6 +11,7 @@ if(NOT WIN32)
   get_filename_component(expat_dir ${EXPAT_LIBRARY} PATH)
   set(ldflags "-L${expat_dir}")
   ExternalProject_Add(libcpluff SOURCE_DIR ${CORE_SOURCE_DIR}/lib/cpluff
+                      BUILD_IN_SOURCE 1
                       PREFIX ${CORE_BUILD_DIR}/cpluff
                       PATCH_COMMAND rm -f config.status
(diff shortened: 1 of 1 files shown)


MartijnKaijser:
you don't have to create trac tickets if you're going to directly fix it :)

smspillaz:
@MartijnKaijser Ah, good to know. Saves me some time :)

hudokkow:
I've seen the error you describe in the ticket when source tree isn't clean, i.e. you build with autotools and than you launch a cmake build without performing a <code>git clean -xfd</code>

smspillaz:
Hm, that could be the reason. I'll give it a try without this fix and see
what happens. Thanks for the pointer.

wsnipex:
yes, autotools artifacts definitely break external_project builds. Are you saying that BUILD_IN_SOURCE fixes that?
That'd be awsome

wsnipex:
its more likely that we don't delete all autotools cruft in the PATCH_COMMAND, although I would have though that removing config.status and running autoreconf should be enough. Maybe we have to delete eventual binaries as well

smspillaz:
@wsnipex Yes, this fixes the issue I was having. Its not exactly the most proper solution (out of source builds are great!), though I suppose in this case its mostly inconsequential since the in-source build of cpluff would happen in the build directory anyway.

wsnipex:
fine for me.
@notspiff any comments?

akva2:
in-tree builds are evil and should never be done by anything. in the name of pragmatism this should go in though.

PR #9692, opened 2016-04-27
cmake: Handle generated files in core_add_test correctly

### LINKED ISSUE #16708: Context-Menu of Music-Addons contains wrong entries
### Describe the bug
Opening the context menu of a music addon shows entries like "Queue item", "Play next" or "Play" which in most cases lead to a neverending busy-dialog aca crash as you can't queue an addon etc. 

## Expected Behavior
Context menu should only contain "Information", "Disable", "Settings" and "Add to favourites". 

## Actual Behavior
neverending busy-dialog

### To Reproduce
1. Go to addons / music-addons
2. Select an addon 
3. Press 'c' 
4. Select one of the mentioned entries

### Debuglog
The debuglog can be found here:  not necessary

### Screenshots 
[https://i.imgur.com/QrRFIhY.jpg](https://i.imgur.com/QrRFIhY.jpg)

### Your Environment
Used Operating system:
 - [ ] Android
 - [ ] iOS
 - [ ] Linux
 - [ ] OSX
 - [ ] Raspberry-Pi
 - [x] Windows
 - [ ] Windows UWP

 - Operating system version/name: Win 8.1 Pro
 - Kodi version:
18.4

smspillaz:
Fixes #16708

### DIFF
--- a/project/cmake/scripts/common/macros.cmake
+++ b/project/cmake/scripts/common/macros.cmake
@@ -41,9 +41,12 @@ endfunction()
 
 # Add a test library, and add sources to list for gtest integration macros
 function(core_add_test_library name)
+  # Backup the old SOURCES variable, since we'll append SUPPORT_SOURCES to it
+  set(TEST_ONLY_SOURCES ${SOURCES})
+  set(SOURCES ${SOURCES} ${SUPPORT_SOURCES})
   core_add_library(${name} NO_MAIN_DEPENDS)
   set_target_properties(${name} PROPERTIES EXCLUDE_FROM_ALL 1)
-  foreach(src ${SOURCES})
+  foreach(src ${TEST_ONLY_SOURCES})
     # This will prepend CMAKE_CURRENT_SOURCE_DIR if the path is relative,
     # otherwise use the absolute path.
     get_filename_component(src_path "${src}" ABSOLUTE)
--- a/project/cmake/scripts/common/macros.cmake
+++ b/project/cmake/scripts/common/macros.cmake
@@ -44,7 +44,10 @@ function(core_add_test_library name)
   core_add_library(${name} NO_MAIN_DEPENDS)
   set_target_properties(${name} PROPERTIES EXCLUDE_FROM_ALL 1)
   foreach(src ${SOURCES})
-    set(test_sources ${CMAKE_CURRENT_SOURCE_DIR}/${src} ${test_sources} CACHE STRING "" FORCE)
+    # This will prepend CMAKE_CURRENT_SOURCE_DIR if the path is relative,
+    # otherwise use the absolute path.
+    get_filename_component(src_path "${src}" ABSOLUTE)
+    set(test_sources "${src_path}" ${test_sources} CACHE STRING "" FORCE)
   endforeach()
   set(test_archives ${test_archives} ${name} CACHE STRING "" FORCE)
 endfunction()
--- a/project/cmake/scripts/common/projectmacros.cmake
+++ b/project/cmake/scripts/common/projectmacros.cmake
@@ -54,6 +54,9 @@ function(GTEST_ADD_TESTS executable extra_args)
         message(FATAL_ERROR "Missing ARGN: Read the documentation for GTEST_ADD_TESTS")
     endif()
     foreach(source ${ARGN})
+        # This assumes that every source file passed in exists. Consider using
+        # SUPPORT_SOURCES for source files which do not contain tests and might
+        # have to be generated.
         file(READ "${source}" contents)
         string(REGEX MATCHALL "TEST_?[F]?\\(([A-Za-z_0-9 ,]+)\\)" found_tests ${contents})
         foreach(hit ${found_tests})
--- a/project/cmake/README.md
+++ b/project/cmake/README.md
@@ -135,6 +135,17 @@ cmake -DCMAKE_TOOLCHAIN_FILE=<KODI_SRC>/tools/depends/target/Toolchain.cmake <KO
 cmake --build . -- VERBOSE=1 -j$(nproc)  # or: make VERBOSE=1 -j$(nproc)
 ```
 
+## Tests
+
+Kodi uses Google Test as its testing framework. Each test file is scanned for tests and these
+are added to CTest, which is the native test driver for CMake.
+
+This scanning happens at configuration time. If tests depend on generated support files which
+should not be scanned, then those support files should be added to the SUPPORT_SOURCES
+variable as opposed to SOURCES before calling core_add_test. You might want to do this where
+the generated support files would not exist at configure time, or if they are so large that
+scanning them would take up an unreasonable amount of configure time.
+
 ## Extra targets
 
 When using the makefile builds a few extra targets are defined:


project/cmake/scripts/common/macros.cmake
```diff
@@ -44,7 +44,13 @@ function(core_add_test_library name)
   core_add_library(${name} NO_MAIN_DEPENDS)
   set_target_properties(${name} PROPERTIES EXCLUDE_FROM_ALL 1)
   foreach(src ${SOURCES})
-    set(test_sources ${CMAKE_CURRENT_SOURCE_DIR}/${src} ${test_sources} CACHE STRING "" FORCE)
+    # This will prepend CMAKE_CURRENT_SOURCE_DIR if the path is relative,
+    # otherwise use the absolute path. We can't just prepend
+    # CMAKE_CURRENT_SOURCE_DIR as this source file might be genertaed
+    # in the build directory (in which case it would have been passed
+    # to SOURCES as an absolute path).
```
wsnipex: since the context of this comment will be mostly lost after this commit(unless one looks at the history, simply saying:
"resolve absolute path" should be good enough

project/cmake/scripts/common/projectmacros.cmake
```diff
@@ -50,10 +50,29 @@ endfunction()
...
     foreach(source ${ARGN})
+        get_property(GENERATED SOURCE "${source}" PROPERTY GENERATED)
+        get_property(SKIP_SCAN_TESTS SOURCE "${source}" PROPERTY SKIP_SCAN_TESTS)
+        if (NOT EXISTS "${source}")
+          # We can't rely on the GENERATED property here since it is likely that
+          # the file was added in a separate subdirectory. According to the CMake
+          # documentation, properties on source files can only be read by other
+          # listfiles in the same directory.
+          list(FIND XBMC_GENERATED_TEST_FILES "${source}" INDEX)
+          if ("${INDEX}" EQUAL -1)
```
wsnipex: INDEX EQUAL -1
should work

wsnipex: also, mind pointing to the docs that say the property cannot be accessed arbitrarily? I cannot find it

smspillaz: I've added a note about `set_source_files_properties` (which seems to be what is used internally)

project/cmake/scripts/common/macros.cmake
smspillaz: Fixed.

smspillaz:
@wsnipex Let me know if there's additional things that need to be addressed here.

wsnipex:
fine for me. Maybe @fetzerch or @notspiff have an idea how to handle such generated files

akva2:
trivial bit; variable should be CORE prefixed (yes i know not entirely consistent). 

i don't quite get the solution. while some of the contents of these tests are generated, the things we snoop for shouldn't be so we can as well snoop the template files afaik.

smspillaz:
@akva2 I was thinking more of the case where you have a "support" file which is generated. That file might not contain any tests, but is used by the tests. This happens in a PR that I am working on. Something has to be done with the generated file and in this case ignoring it makes the most sense. I gather that there might be templates which contain tests - I suppose the best way to handle that is to add them to the `SOURCES` of the test target (they will be automatically ignored by the build system, but still available for use by things like `gtest_add_tests`)

I don't prefer the approach taken by `FindGTest.cmake` in regexing files for tests. In practice I've found it to be [quite fragile](http://smspillaz.github.io/2012/07/05/unit-test-autodiscovery-with-cmake-and-google-test/). That discussion is probably out of scope for the purposes of this PR, but maybe worth looking into in the future?

akva2:
the problem is that the regex'ing is the only way to map gtest to ctest. otherwise you just get a single test under ctest. i'm all ears for a better approach but it's the only one i have found.

i don't see why we need to bother with generated resources here at all. you just make the test target depend on the resource file and it will be generated before the test is executed.

smspillaz:
@akva2 Ah, that's not really the problem I was referring to.

The problem is this:

    add_custom_command (OUTPUT my_generated_support_file.cpp
                                           COMMAND ...)

    set (SOURCES ...
                              my_generated_support_file.cpp)
    core_add_test (...)

In this case, the test uses symbols from `my_generated_support_file.cpp`. However, `core_add_test` calls `gtest_add_tests` which does something like this:

    get_property (TEST_SOURCES TARGET test_target PROPERTY SOURCES)
    foreach (SOURCE ${TEST_SOURCES})
        file (READ ${SOURCE} ...)
    endforeach ()

However, at that point, `${SOURCE}` might not actually exist on-disk, because it hasn't been generated yet (only happens at build time, not configure time). Because of that, we have to skip it.

akva2:
aha. now i get it. it's an unfortunate assumption on my part, that all sources going into the test executable will hold tests. then this should be fine, but the variable name is slightly confusing. this is not only for generated sources, but test support sources (or whatever name you want to apply, but generated is misleading).

akva2:
what i'm suggesting is to add a second list of sources to be linked into the test application which is not passed to gtest_add_tests, then give core_add_test another parameter for such sources. can you take care of that? i can, but it won't be now for sure, likely in a few days.

project/cmake/scripts/common/projectmacros.cmake
```diff
@@ -50,10 +50,32 @@ endfunction()
 # Copied from FindGTest.cmake 
 # Thanks to Daniel Blezek <blezek@gmail.com> for the GTEST_ADD_TESTS code
 function(GTEST_ADD_TESTS executable extra_args)
+    get_property(XBMC_GENERATED_TEST_FILES GLOBAL PROPERTY XBMC_GENERATED_TEST_FILES)
+
     if(NOT ARGN)
         message(FATAL_ERROR "Missing ARGN: Read the documentation for GTEST_ADD_TESTS")
     endif()
     foreach(source ${ARGN})
+        get_property(GENERATED SOURCE "${source}" PROPERTY GENERATED)
```
smspillaz: This is a leftover that I should probably remove.

project/cmake/scripts/common/projectmacros.cmake
```diff
@@ -50,10 +50,32 @@ endfunction()
 # Copied from FindGTest.cmake 
 # Thanks to Daniel Blezek <blezek@gmail.com> for the GTEST_ADD_TESTS code
 function(GTEST_ADD_TESTS executable extra_args)
+    get_property(XBMC_GENERATED_TEST_FILES GLOBAL PROPERTY XBMC_GENERATED_TEST_FILES)
```
smspillaz: @akva2 Was this the variable name you were referring to? I can probably rename it to something like `CORE_TEST_SUPPORT_FILES`

akva2: yeah. that's the one i was referring to. but if you read my other comment, i think that suggestion is a better one. then we don't have to hack around in gtest_add_tests at all. if we distinguish between sources going into the test .a and things added to test_SOURCES in core_add_test, we can avoid this completely.

smspillaz: Hmm, so I'm looking into this and it looks like such a change would either impact quite a lot of other code or would be kind of weird.

The first solution would be to have to have two variables, `SOURCES` and `SUPPORT_SOURCES`. Both of them are used by `add_library`, except that only the former is used by `gtest_add_tests`. The downside here is that it isn't immediately obvious that `SOURCES` is supposed to exclude "support" files.

The second solution would be to have `TEST_SOURCES` and `SUPPORT_SOURCES`, which are both merged into `SOURCES` at the time `core_add_test` is run. That would require all the tests' CMakeLists to be updated.

Since its the least invasive, I'll implement the first so we can see how it looks.

akva2: yes, there will be an impact. consequence of getting things wrong in the first place, but let's fix it properly.
the first is what i had in mind. whoever creates the CMakeListst.txt for the test library simply needs to be aware.

wsnipex:
this looks much cleaner now.
Could you please add a note in the README about support_sources usage in tests?

smspillaz:
Done :)

akva2:
Good stuff thx for patience.

wsnipex:
great, thanks!

PR #9693, opened 2016-04-27
cmake: Handle ENABLE_VAAPI and ENABLE_VDPAU both being disabled correctly

smspillaz:


FernetMenta:
@fetzerch @wsnipex ok to merge?

PR #9592, opened 2016-04-10
VideoPlayer: InputStream

FernetMenta:
- fix use of multiple types of InputStream addons
- use parent, if not busy. safes copying the dll every time on open

@mapfau could you test this with your second addon?

PR #9694, opened 2016-04-27
[VideoPlayer] fixup! stalls after seek.

afedchin:
see title

@FernetMenta ping

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayerVideo.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayerVideo.cpp
@@ -853,7 +853,7 @@ int CVideoPlayerVideo::OutputPicture(const DVDVideoPicture* src, double pts)
 
   int timeToDisplay = DVD_TIME_TO_MSEC(pts - iPlayingClock);
   // make sure waiting time is not negative
-  int maxWaitTime = std::max(timeToDisplay + 500, 50);
+  int maxWaitTime = std::min(std::max(timeToDisplay + 500, 50), 500);
   // don't wait when going ff
   if (m_speed > DVD_PLAYSPEED_NORMAL)
     maxWaitTime = std::max(timeToDisplay, 0);


FernetMenta:
+1
thanks!

popcornmix:
I'm not sure this is a complete fix. For many files a couple of seconds after seeking I now get a ~500ms stall then video continues as normal.
That is obviously better than without this PR where stalls could last for several seconds, but it's not as good as it used to be with no stalls (I believe the stalls appeared with inputstream PR).
Ping @FernetMenta

FernetMenta:
how far is pts off after seek?

afedchin:
pst off equals to amount of seek. as an example: before seek pts was 4206756746.333336 after seek forward for 30 sec the pts became 4236536500.000000

FernetMenta:
I meant offset related to clock
timeToDisplay = DVD_TIME_TO_MSEC(pts - iPlayingClock);

afedchin:
timeToDisplay: 29476 just after seeking forward 30 sec.
timeToDisplay: 184887 just after randon seeking forward about 3 min.

afedchin:
timeToDisplay: -176686 just after random seeking backward.

FernetMenta:
I can't reproduce here. Does it happen with only some material? With SW decodeing?

afedchin:
try with bluray.

FernetMenta:
I don't see how this matters. After seek the renderbuffers get flushed. There should be a buffer available and maxWaitTime is meaningless. 
What do I miss?

afedchin:
If I understand correctly CRenderManager::DiscardBuffer() doesn't make all buffers as free and only marks them to discard. The actual freeing of frames occurs on FrameMove. So CRenderManager::WaitForBuffer will wait till timeout or till FrameMove

FernetMenta:
correct but this happens all within one render cycle that is max 41ms.
looks like that the event is not set when buffer are pushed to queue of free buffers.

popcornmix:
Actually my stall seems to follow a "timeout waiting for buffer" message:
http://paste.ubuntu.com/16363857/

I get one of these a couple of seconds after most seeks (but not during normal playback).
Can't see an obvious reason for this. The last couple of calls to Decode returned pictures.
The same picture was passed to RenderUpdate many (17) times before the timeout.

FernetMenta:
> Actually my stall seems to follow a "timeout waiting for buffer" message:

That's 7 seconds after the seek. During this timeframe many pictures have passed. What relationship do you see to this PR?

popcornmix:
Yes, I think I'm wrong with relating it to this PR.

I attempted to bisect to find the start of this stall.
The bisect found "Merge pull request #9592 from FernetMenta/inputstream" which actually caused very long stalls (of many seconds). I believe this PR fixed the long stalls, but there was still a shorter stall remaining.
So there are possibly two unrelated stall bugs that overlapped in time.

FernetMenta:
bisect is of no interest. things that never worked properly or only by chance can't get broken. better investigate why pts jumps forward.

PR #9695, opened 2016-04-27
VideoPlayer: add setting for double/triple buffers

afedchin:
see title

@FernetMenta ping

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -16238,7 +16238,10 @@ msgctxt "#36042"
 msgid "Use limited colour range (16-235)"
 msgstr ""
 
-#empty string with id 36043
+#: system/settings/settings.xml
+msgctxt "#36043"
+msgid "Number of buffers used by graphics driver"
+msgstr ""
 
 #. Name of an action to perform on a specific event, like changing the CEC source away from Kodi
 #: xbmc/peripherals/devices/PeripheralCecAdapter.cpp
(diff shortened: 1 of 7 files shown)


zag2me:
Better String:
"Defines the number of presentation buffers that are used by the graphics driver. Select 2 if the driver uses double buffering or 3 for triple buffering"

afedchin:
@zag2me thanks.
updated

FernetMenta:
+1
thanks for continuing work on this

FernetMenta:
jenkins build this please

FernetMenta:
@afedchin could you rebase and resolve the merge conflict please

afedchin:
will do asap

afedchin:
jenkins build this please

MartijnKaijser:
Does this work on all platforms? If not this setting needs to be hidden on those.

afedchin:
this is base implementation. maintainers of those platforms may hide setting if it's not needed.

FernetMenta:
@MartijnKaijser your question does not match 100% to the problem :)
For a/v sync VideoPlayer needs to know display latency. This is dependent on the number of buffers used be the driver. Prior to this patch we assumed double buffering but in the meantime many systems have triple buffers. Unfortunately this is not detectable on many systems, hence the setting. If a platform or appliance knows the number, it can hide the setting from the user.

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -18444,7 +18447,13 @@ msgctxt "#36551"
 msgid "Defines the strength of the stereoscopic 3D effect in the GUI. This is done by controlling the depth of perception within the GUI, so the higher the value, the more elements will pop out of the screen. [Zero] Disables the stereoscopic 3D effect of the GUI.[CR]For a good visual experience, the value should be higher for small screens and lower for large screens. Note: this is not supported by all skins."
 msgstr ""
 
-#empty strings from id 36552 to 36599
+#. Description of setting "System -> Video output -> NoOfBuffers" with label #36043
+#: system/settings/settings.xml
+msgctxt "#36552"
+msgid "Defines the number of presentation buffers used by the graphics driver. Select 2 if the driver uses double buffering or 3 for triple buffering."
```
FernetMenta: "defines" sounds misleading to me. This value does not affect the number of buffers used by the driver.

afedchin: any suggestions?

PR #9696, opened 2016-04-27
[skin] save skin settings early on exit

stefansaraev:
this should fix the issue where skin settings may not be saved on application exit / reboot / shutdown.

cc @Montellese this is trivial but second look would be nice.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -2797,6 +2797,13 @@ void CApplication::Stop(int exitCode)
     else
       CLog::Log(LOGNOTICE, "Not saving settings (settings.xml is not present)");
 
+    // kodi may crash or deadlock during exit (shutdown / reboot) due to
+    // either a bug in core or misbehaving addons. so try saving
+    // skin settings early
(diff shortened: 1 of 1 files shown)


Montellese:
You should rather mention that the problem only happens when something crashes on exit / reboot / shutdown. When everything goes as it should, the skin settings should also be fine right?

Apart from that it looks ok.

stefansaraev:
> When everything goes as it should, the skin settings should also be fine right

thats right. I'll update the comment.

MartijnKaijser:
jenkins build this please

piotrasd:
Im using this patch and work great ! even if device get crash
https://github.com/wrxtasy/LibreELEC.tv.7.0/blob/libreelec-7.0/projects/Odroid_C2/patches/kodi/g10-Fix.Save.Skin.settings.changes.immediately.patch

and thats secod
https://github.com/wrxtasy/LibreELEC.tv.7.0/blob/libreelec-7.0/projects/Odroid_C2/patches/kodi/k22-allow-settings-saved-from-python.patch

PR #9697, opened 2016-04-27
[python-api] - deprecate listitem.getduration(), .getfilename(), .getdescription()

phil65:
- getdescription() is basically getLabel(), just without locking, but with confusing naming instead. 
- getduration() should be done via InfoTagMusic
- no idea what getfilename() should be used for at all (docs from setPath() tell me that this is not the filename, and InfoTagVideo already has .getPath() and .getFile() available)

Also, all use wrong upper/lowercasing.

From my side we could remove all those directly. I guess @tamland wants them depracated for now?

### DIFF
--- a/xbmc/interfaces/legacy/ListItem.cpp
+++ b/xbmc/interfaces/legacy/ListItem.cpp
@@ -277,6 +277,12 @@ namespace XBMCAddon
       return item->GetPath();
     }
 
+    String ListItem::getPath()
+    {
+      LOCKGUI;
+      return item->GetPath();
+    }
+
     void ListItem::setInfo(const char* type, const InfoLabelDict& infoLabels)
     {
       LOCKGUI;
--- a/xbmc/interfaces/legacy/ListItem.h
+++ b/xbmc/interfaces/legacy/ListItem.h
@@ -620,7 +620,7 @@ namespace XBMCAddon
       /// **Example:**
       /// ~~~~~~~~~~~~~{.py}
       /// ...
-      /// self.list.getSelectedItem().setPath(path='ActivateWindow(Weather)')
+      /// self.list.getSelectedItem().setPath(path='/path/to/some/file.ext')
       /// ...
       /// ~~~~~~~~~~~~~
       ///
@@ -691,9 +691,8 @@ namespace XBMCAddon
       /// \ingroup python_xbmcgui_listitem
       /// @brief \python_func{ getdescription() }
       ///-----------------------------------------------------------------------
-      /// Returns the description of this PlayListItem.
+      /// @warning Deprecated.
       ///
-      /// @return Description string of play list item
       ///
       getdescription();
 #else
@@ -705,9 +704,8 @@ namespace XBMCAddon
       /// \ingroup python_xbmcgui_listitem
       /// @brief \python_func{ getduration() }
       ///-----------------------------------------------------------------------
-      /// Returns the duration of this PlayListItem
+      /// @warning Deprecated.
       ///
-      /// @return duration as string
       ///
       getduration();
 #else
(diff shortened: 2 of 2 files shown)


xbmc/interfaces/legacy/ListItem.h
```diff
@@ -393,17 +393,17 @@ namespace XBMCAddon
...
-       * getduration() -- Returns the duration of this PlayListItem\n
+       * getduration() -- Deprecated.\n
        */
       String getduration();
 
       /**
-       * getfilename() -- Returns the filename of this PlayListItem.\n
+       * getfilename() -- Depracated.\n
        */
       String getfilename();
```
AlwinEsch: From my side for `String getfilename();` is good to have a `String getPath();` instead.

To remove them makes for me no sense, not sure about python, but for c++ can be the before found path only be stored in list item and if add-on need it for something again can it be loaded from there.

garbear:
p.s. spelling is "deprecated"

phil65:
@garbear Hehe I know, got it right in 2 of 3 occasions :)
@AlwinEsch ok, I could introduce .getPath() then instead if it is needed.

phil65:
@AlwinEsch are you sure that this is related to the filename at all?
The docs for .setPath() say:

* example:
*   - self.list.getSelectedItem().setPath(path='ActivateWindow(Weather)')

AlwinEsch:
Mmmmmmmhhhh, has never tried it like this but has looked over the FileItem and it does a much with path but not seen somewhere code for a call like this.

Is it possible that this example is strongly wrong?
The Python call for `setPath` is a call from `SetPath`from CFileitem

phil65:
Yes, the example might be wrong.
So what about:
1) deprecating .getfilename()
2) adding .getPath() instead
3) fixing the example
Does that sound good to you, @AlwinEsch ?

AlwinEsch:
Perfect :+1:

phil65:
Updated.

tamland:
>From my side we could remove all those directly. I guess @tamland wants them depracated for now? 

Fine by me to remove them If it isn't used by anything in repo. getdescription looks rather broken.. it doesnt have anything to do with 'description'.

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -277,6 +277,11 @@ namespace XBMCAddon
       return item->GetPath();
     }
 
+    String ListItem::getPath()
+    {
+      return item->GetPath();
```
tamland: needs lock

phil65:
updated and rebased.

xbmc/interfaces/legacy/ListItem.h
```diff
@@ -691,9 +691,8 @@ namespace XBMCAddon
       /// \ingroup python_xbmcgui_listitem
       /// @brief \python_func{ getdescription() }
       ///-----------------------------------------------------------------------
-      /// Returns the description of this PlayListItem.
+      /// Deprecated.
```
AlwinEsch: I know a bit new with the full doxygen usage :)

Maybe add it as `@warning Deprecated`, it bring then a red line before, as example visible here http://alwinesch.github.io/group__python__monitor.html#ga5ee33edfc21fb06cfe99f892da9b234c.

xbmc/interfaces/legacy/ListItem.h
```diff
@@ -731,6 +728,20 @@ namespace XBMCAddon
 #ifdef DOXYGEN_SHOULD_USE_THIS
       ///
       /// \ingroup python_xbmcgui_listitem
+      /// @brief \python_func{ getfilename() }
```
AlwinEsch: A small c&p fault :) should be `getPath()`

phil65:
Updated.

phil65:
jenkins build this please.

phil65:
hello jenkins? jenkins build this please.

phil65:
jenkins build this please.

MartijnKaijser:
jenkins build this please

phil65:
jenkins build this please.

stefansaraev:
jenkins build this please

phil65:
jenkins build this please

verybadsoldier:
I am just trying to port my addons to v19 and stumbled upon that `getfilename` was removed. I have problems finding the new way to get the physical file path. Could someone help me out please?

PR #9698, opened 2016-04-27
[gui] - change lock for CGUIWindowManager::Render() to CSingleLock

phil65:
While playing with MediaWindow functionality of our Python interface, I noticed quite some regular crashing when using WindowXML.addItem(). 
@mapfau helped me finding the issue and it seems to be a wrong lock for our WindowManager Render() method. (LOCKGUI in WindowXML.cpp doesnt seem to kick in without this change)
@jimfcarroll and @FernetMenta may be able to provide valuable input here if this is a correct fix.
I guess the issue can happen for lot more methods, addItem() just makes it more likely for it to appear.

DISCLAIMER: I dont really have experience in this area, so I am completely dependent on the help of the experienced guys here. :) I opened this PR anyways to get a discussion going, would be great to get this fixed.

The 2 conflicting callstacks can be found here: http://pastebin.com/UQM2gNuL

### DIFF
--- a/xbmc/guilib/GUIWindowManager.cpp
+++ b/xbmc/guilib/GUIWindowManager.cpp
@@ -1063,7 +1063,7 @@ void CGUIWindowManager::RenderEx() const
 bool CGUIWindowManager::Render()
 {
   assert(g_application.IsCurrentThread());
-  CSingleExit lock(g_graphicsContext);
+  CSingleLock lock(g_graphicsContext);
 
   CDirtyRegionList dirtyRegions = m_tracker.GetDirtyRegions();
 


FernetMenta:
This is wrong. gfx lock must not be hold during render. If there are any item manipulating functions in Render(), those must be moved to FrameMove

FernetMenta:
We can't allow the outside world to lock Kodi's rendering thread. Hence LOCKGUI should work as follows:

- every interface call that wants to manipulate gui signals this to Kodi
- the call is locked by a special critical section, not gfx lock
- CApplication::FrameMove unlocks the door
- the python call does its job and leaves the lock
- FrameMove locks the door again

mapfau:
I'll give it a try

phil65:
Thanks!

phil65:
Dropped in favour of https://github.com/xbmc/xbmc/pull/9748

PR #9699, opened 2016-04-27
[python-api] remove checkmark control

phil65:
Removes the checkmark control stuff from our Python interface.
Checkmark controls are deprecated since a long time and should get ripped from core completely (Most skinners also dont define them in defaults.xml) . This is a first small step.
@ronie @tamland


AlwinEsch:
Good way to start remove of them.

ronie:
yup, this can go.

tamland:
jenkins build this please

PR #9110, opened 2016-02-11
[addons] fix 'system addons'

tamland:
This removes various workarounds and fixes the broken path check for determining which are system addons (the ones per-installed and required to operate). They're now loaded from a manifest file.

Note that this flips the default state for addons installed via cpluff to disabled. For users there will be no change, but if you're a distribution maintainer you'll need to patch the manifest file if you want to have other addons part of the system addon. <-@stefansaraev?

PR #9700, opened 2016-04-28
[addondb] Enable addons by type when syncing the database with the filesystem

garbear:
Kodi keeps the add-on database in sync with the filesystem. This means that when an add-on is copied manually, Kodi adds it to the database. This also means that when the database is deleted or Kodi is first run, all add-ons are added to the database.

The current behavior is that Kodi sets all add-ons to disabled when they are added to the database. Then, if the add-on is a system add-on (appears in [addon-manifest.xml](https://github.com/xbmc/xbmc/blob/master/system/addon-manifest.xml)), it is enabled.

This PR causes Peripheral add-ons (and harmless controller profiles) to be enabled when syncing to the database.

Going forward, an "optional" parameter should be added to addon-manifest.xml. Optional add-ons shouldn't prevent Kodi from starting, and the user should be allowed to disable them.

Once that's possible, we can remove the peripheral add-ons special case added here.

### DIFF
--- a/xbmc/addons/AddonDatabase.cpp
+++ b/xbmc/addons/AddonDatabase.cpp
@@ -240,8 +240,11 @@ void CAddonDatabase::SyncInstalled(const std::set<std::string>& ids, const std::
     std::string now = CDateTime::GetCurrentDateTime().GetAsDBDateTime();
     BeginTransaction();
     for (const auto& id : added)
+    {
+      int enable = CAddonMgr::GetInstance().EnabledOnSync(id) ? 1 : 0;
       m_pDS->exec(PrepareSQL("INSERT INTO installed(addonID, enabled, installDate) "
-          "VALUES('%s', 0, '%s')", id.c_str(), now.c_str()));
+          "VALUES('%s', %d, '%s')", id.c_str(), enable, now.c_str()));
+    }
 
     for (const auto& id : removed)
     {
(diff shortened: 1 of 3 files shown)


garbear:
btw the comments aren't very elaborate... can someone say why pvr and audio decoder add-ons shouldn't be enabled on startup?

garbear:
I updated the PR to only enable peripheral add-ons and controller profiles when syncing.

@MartijnKaijser Can this go in to fix silence our joystick complaints?

The next course of action is to add an "optional" parameter to the manifest from #9110 that doesn't prevent Kodi from starting when missing and also allows the add-on to be disabled.

Once that's done, we can add peripheral.joystick as an optional system add-on and remove the peripheral add-ons special case from this commit.

tamland:
>Going forward, an "optional" parameter should be added to addon-manifest.xml. Optional add-ons shouldn't prevent Kodi from starting, and the user should be allowed to disable them.

Why aren't you doing that?

garbear:
I said it "should" be done, I didn't say I was going to do it ;) This was a quick fix that took little effort to fix a crash-and-burn issue.

Is the optional parameter something you could look into? I won't have time today, but I could invest the effort when I get a chance in the future

garbear:
once I put some thought into it, optional add-ons turned out to be really easy. 15 lines max. Closing this in favor of #9706

PR #9702, opened 2016-04-28
VideoPlayer: OSX/iOS - use ffmpeg videotoolbox

FernetMenta:
see title
@Memphiz works on OSX, next is iOS
also fixes the issue of playing 720p channels


tools/darwin/Configurations/App-OSX.xcconfig
```diff
@@ -24,7 +24,7 @@ ARCHS = i386 x86_64
 ONLY_ACTIVE_ARCH = YES
 
 SDKROOT = macosx
-MACOSX_DEPLOYMENT_TARGET = 10.7
+MACOSX_DEPLOYMENT_TARGET = 10.8
```
Memphiz: needs a bump here too

https://github.com/xbmc/xbmc/blob/master/tools/depends/configure.ac#L257

not sure if anywhere else for cmake

xbmc/cores/VideoPlayer/DVDCodecs/Video/VTB.cpp
```diff
@@ -0,0 +1,124 @@
+/*
+ *      Copyright (C) 2005-2013 Team XBMC
```
Memphiz: use current gpl header with kodi and 2016?

Memphiz:
That looks pretty cool as of the changeset statistics - will give it a shot ln my mini

Memphiz:
It still links against the VideoDecodeAcceleration.framework - not sure if this is by intention?

system/settings/settings.xml
```diff
@@ -817,8 +817,8 @@
           <default>true</default>
           <control type="toggle" />
         </setting>
-        <setting id="videoplayer.usevda" type="boolean" label="13429" help="36160">
-          <requirement>HasVDA</requirement>
+        <setting id="videoplayer.usevtb" type="boolean" label="13429" help="36160">
+          <requirement>HasVTB</requirement>
```
Memphiz: The referenced help string still mentions VDA (needs a change to VTB).

Memphiz:
Mmhhh doesn't seem to work for my macmini with intel i5 3210m.

I just get:

16:11:09 T:123145302839296  NOTICE: Creating video codec with codec id: 28
16:11:09 T:123145302839296   DEBUG: FactoryCodec - Video:  - Opening
16:11:09 T:123145302839296  NOTICE: CDVDVideoCodecFFmpeg::Open() Using codec: H.264 / AVC / MPEG-4 AVC / MPEG-4 part 10
16:11:09 T:123145302839296   DEBUG: FactoryCodec - Video: ff-h264 - Opened

Even if i have ffmpeg debugging enabled. It seems like "setup_hwaccel" in ff_get_format fails. Whats the easiest way to compile ffmpeg with -O0 and -g? (so i can be sure)

FernetMenta:
@Memphiz why don't you set a bp on if (av_videotoolbox_default_init(avctx) < 0) and step into this function?

Memphiz:
well somehow there is always a -O3 sneaking into the ffmpeg compile which doesn't allow me to debug properly due to optimisation ...

wm4:
FFmpeg's videotoolbox has some issues currently. There is work going on in ffmpeg to improve the decoder. It will be reimplemented as full decoder instead of hwaccel, which should improve its performance and format support.

>well somehow there is always a -O3 sneaking into the ffmpeg compile which doesn't allow me to debug properly due to optimisation ...

You need something like ` --disable-optimizations --enable-debug --disable-stripping`

Memphiz:
I think i got it working. I see a drop in cpu load between software decode and vtb usage from 150% to 20% ... i somehow just expected it to show ff-vtb in the log instead of ff-h264 ...

FernetMenta:
@Memphiz now with iOS support. fixes the flickering issue.

Memphiz:
Nice will try it out in some minutes ... - no deployment target change for ios?

FernetMenta:
> no deployment target change for ios?

not that I am aware of :)

Memphiz:
Then it has to crash on ios 5 i am pretty sure ... will try it out ...

FernetMenta:
@Memphiz could you submit the change if required? my skills in this area are limited

Memphiz:
will do once i got the needed info.

Memphiz:
Works on my ios9.0 device, doesn't work on ios 6.1.2 (debugging atm ...) - not verified on ios5.1.1 yet

Memphiz:
@wm4 what is the lowest ios version this is supposed to run? Or do you also dyload the VTB dylib (and use the privateframeworkpath for < ios 6?)

wm4:
Personally I have no idea about iOS and never cared about it in any way.

MrMC:
VTB was moved public in iOS8, everything below that needs to dyload the private lib which has a different path. Also note that as being private, the API has slight changes from iOS version to version.

Memphiz:
I know that ios8 was the version which had it public in the headers ... but i could swear they moved the binary into the normal frameworks folder already in ios6 on the devices ...

MrMC:
yes, that is correct. 'load from public framework path -> available for >= ios 6.0'
love those code comments, saves from having to grep about :)

this implementation would be safe back to iOS6. or should be.

Memphiz:
ok - need some printouts then why vtsessioncreate failes ... i guess the prototype changed at one point ...

MrMC:
Nice to see CVOpenGLESTextureCache come in, now if ffmpeg sets up the surface correctly, it is the fastest way to decode and render on iOS/tvOS as the surface stays entirely on the GPU side. 

Even via decode sw FFMpeg, if you get the decoded frame into a GPU surface right away, it is a performance boost.

MrMC fork was planning to bring in the rendering restructure just to get add in CVOpenGLESTextureCache handling to recover more CPU for a decoding boost.

Memphiz:
well - looking at the cpu load and looking at the ffmpeg code (which suggests memcpys on buffers) it is not as efficient as one might think ... sw decode seems even worse now (not sure if it is even touched in this PR).

MrMC:
Humm, yea thinking on it, 150 to 20 percent under VTB. Something is sucking CPU :) Should dive much lower than that.

Memphiz:
@FernetMenta is there a sane way to use your vtbgles renderer but our old videotoolbox decoder?

Memphiz:
mhh problem on ios6 is "kVTCouldNotFindVideoDecoderErr" status -12906

FernetMenta:
@Memphiz we have out own ffmpeg fork and can easily patch ffmpeg code. shouldn't be a big deal, it's all in this file: https://github.com/FFmpeg/FFmpeg/blob/eae2d89bf715bc3edff478174b43e1f388e768bf/libavcodec/videotoolbox.c

FernetMenta:
I agreed with @Memphiz that we merge this and fix broken VTB on iOS. We will investigate iOS 6 issues later. 
jenkins build this please

ksooo:
@FernetMenta cool. Just did a quick check with pvr.hts. First time 720p channels work (to some extend). "Das Erste HD" seems to work just fine. "ZDF HD" does not. I remember from a former discussion with the tvheadend guys that "ZDF HD" is "somwwhat different". What ever this means. Just wanted to let you know. Do you want a log file for both "Das Erste HD" and "ZDF HD"?

FernetMenta:
@ksooo vtb is currently reworked in ffmpeg. Once this is completed it, it is suppoed to work on interlaced h264 content as well.

ksooo:
@FernetMenta okay. thanks for the info.

fritsch:
@ksooo http://solidrun.maltegrosse.de/~fritsch/720p50_h264.mp4 <- does this ZDF recording work?

ksooo:
@fritsch this works, but no wonder, because it is an "Das Erste HD" recording, not "ZDF HD". ;-)

fritsch:
Jep - only uploaded via ssh ... will provide you a sample later tonight.

fritsch:
@ksooo http://fritsch.fruehberger.net/samples/zdf_raw_128.ts <- here, a raw dump from dvblink of my brother in law ... play it via the file dialogue, please.

ksooo:
@fritsch no, does not work. Same behavior as with live tv here.

PR #9703, opened 2016-04-28
lib/cximage-6.0: fix compilation with gcc6

bkuhls:
For a quick fix I renamed min() to cxmin() and max() to cxmax() to
prevent the conflict with the gcc definition.

Forum thread for reference:
http://forum.kodi.tv/showthread.php?tid=263884

### DIFF
--- a/lib/cximage-6.0/CxImage/ximabmp.cpp
+++ b/lib/cximage-6.0/CxImage/ximabmp.cpp
@@ -46,7 +46,7 @@ bool CxImageBMP::Encode(CxFile * hFile)
 		bihtoh(&infohdr);
 
 		// Write the file header
-		hFile->Write(&hdr,min(14,sizeof(BITMAPFILEHEADER)),1);
+		hFile->Write(&hdr,cxmin(14,sizeof(BITMAPFILEHEADER)),1);
 		hFile->Write(&infohdr,sizeof(BITMAPINFOHEADER),1);
 		 //and DIB+ALPHA interlaced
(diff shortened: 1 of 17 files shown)


fritsch:
If that was the only thing that is problematic with gcc6.

cximage was removed from master branch ans as of now - Jarvis won't get another point release. So not sure what to do with the PR. Let's ask the RM  @MartijnKaijser what do we want to do?

MartijnKaijser:
Indeed I don't forsee another point release. Master diverged way too much already for doing backports anyway. 
Thank you anyway for this PR but we're gonna close it. It should still be helpful for people running into the same problem.

stefansaraev:
I think this should be open for some time, so its easier for people to find it. with comments locked to minimize the noise

there be some distros in next few months, compiling jarvis with gcc6

fritsch:
Yes, let's keep it open - as the work is already done.

MartijnKaijser:
Open or closed doesn't matter they can still find it 
I was trying to bring down our open PR list with stuff that won't get added or is outdated

Razzeee:
I also feel like @MartijnKaijser it shouldn't be open if it's not considered (anymore) to be merged at any point.

stefansaraev:
it is obviously easier to find it in open PRs, if something is broken for me, I never search closed PRs. most people wont. we have over 9000! and it is near impossible to find something there at github :)

I dont mind closing it, however, making other peoples lives easier does not hurt.

PR #9704, opened 2016-04-28
[GUI] add orientation support to slider controls

ronie:
allow skins to create a vertical slider, as requested on the forum:
http://forum.kodi.tv/showthread.php?tid=218756&pid=2318996#pid2318996

ping @AchimTuran 

@tamland i tried to add it to the xbmcgui.ControlSlider() class as well for python addons,
but i'm pretty uncomfortable messing in this code area, please review.

*note this will mess up existing sliders as the default orientation defined globally in kodi is vertical.
skins/addons will need to adopt.


xbmc/interfaces/legacy/Control.h
```diff
@@ -1517,7 +1519,8 @@ namespace XBMCAddon
 #ifndef SWIG
       std::string strTextureBack;
       std::string strTexture;
-      std::string strTextureFoc;    
+      std::string strTextureFoc;
+      int iOrientation;
```
tamland: you can use the enum here I think

xbmc/interfaces/legacy/Control.h
```diff
@@ -1496,7 +1497,8 @@ namespace XBMCAddon
       ControlSlider(long x, long y, long width, long height, 
                     const char* textureback = NULL, 
                     const char* texture = NULL,
-                    const char* texturefocus = NULL);
+                    const char* texturefocus = NULL,
+                    long orientation = (VERTICAL | HORIZONTAL));
```
tamland: Not sure what you're trying to do here. This is the same as 0|1 which is the same as 1 and VERTICAL. And it can be int, no need for long;)

xbmc/interfaces/legacy/Control.h
```diff
@@ -1473,15 +1473,16 @@ namespace XBMCAddon
...
      * 
      * x              : integer - x coordinate of control.\n
      * y              : integer - y coordinate of control.\n
      * width          : integer - width of control.\n
      * height         : integer - height of control.\n
      * textureback    : [opt] string - image filename.\n
      * texture        : [opt] string - image filename.\n
-     * texturefocus   : [opt] string - image filename.n"
+     * texturefocus   : [opt] string - image filename.\n
+     * orientation    : [opt] integer - orientation of slider (VERTICAL / HORIZONTAL).n"
```
tamland: n at the end

tamland: does VERTICAL and HORIZONTAL exist in the python api? if not, it should be mapped, or docs changed to '0 or 1'

AchimTuran:
@ronie 
Awesome work, which is perfect for adsp.biquad.filters and it's build in EQ. This will improve the usability :-)

AchimTuran:
> *note this will mess up existing sliders as the default orientation defined globally in kodi is vertical.
skins/addons will need to adopt.

Is it not possible to use existing skin code? If not would it be possible to set the slide as vertical if the orientation tag is missing?

xbmc/interfaces/legacy/Control.cpp
```diff
@@ -533,12 +533,14 @@ namespace XBMCAddon
     ControlSlider::ControlSlider(long x, long y, long width, long height,
                                  const char* textureback,
                                  const char* texture,
-                                 const char* texturefocus)
+                                 const char* texturefocus,
+                                 long orientation)
```
AchimTuran: What @tamland wrote. If it is possible you could use the enum here.

AchimTuran:
Except the things what @tamland mentioned I can't see any issues.

Many thanks for the patch, can't wait to see it in our master branch. I guess Rjsachse will be very happy about this.

rjsachse:
Wouldn't a skinner just need to change the default.xml so it don't break the rest of the skin. Should not be a huge deal breaker since krypton has so many changes.

ronie:
> Wouldn't a skinner just need to change the default.xml so it don't break the rest of the skin. Should not be a huge deal breaker since krypton has so many changes.

correct, it's exactly 1 line of code skinners need to add to  defaults.xml

ronie:
@tamland updated the python implementation.
compile & runtime tested.

xbmc/interfaces/legacy/Control.h
```diff
@@ -1473,15 +1473,16 @@ namespace XBMCAddon
...
      * 
      * x              : integer - x coordinate of control.\n
      * y              : integer - y coordinate of control.\n
      * width          : integer - width of control.\n
      * height         : integer - height of control.\n
      * textureback    : [opt] string - image filename.\n
      * texture        : [opt] string - image filename.\n
-     * texturefocus   : [opt] string - image filename.n"
+     * texturefocus   : [opt] string - image filename.\n
+     * orientation    : [opt] integer - orientation of slider (0=horizontal / 1=vertical(default)).
```
tamland: @ronie can you map these to enums as I probably too subtly suggested?

tamland: All you have to do is add

 ````
SWIG_CONSTANT(int, HORIZONTAL);
SWIG_CONSTANT(int, VERTICAL);
````
in ModuleXbmcgui.h. In python people can then write `ControlSlider(.. orientation=xbmcgui.VERTICAL)` instead of `orientation=0` and trying to remember what 0 means.

ronie:
@tamland thx! i need a bit of hand holding now and then :-)
please check if last commit does the trick.

jenkins build this please

ronie:
squashed & added skin needed changes

ronie:
rebased after https://github.com/xbmc/xbmc/issues/9674

@tamland good to go?

tamland:
yeah, if it builds:)

ronie:
jenkins build this please

PR #9706, opened 2016-04-29
Optional addons

garbear:
Adds optional add-ons to addon-manifest.xml

Optional add-ons are only used to set the enabled bit when add-ons are synced from the filesystem to the database. They do not participate in the other functions of system add-ons, such as preventing Kodi from starting or blocking being disabled in the GUI.

Replaces #9700

### DIFF
--- a/xbmc/addons/AddonDatabase.cpp
+++ b/xbmc/addons/AddonDatabase.cpp
@@ -210,7 +210,9 @@ void CAddonDatabase::UpdateTables(int version)
   }
 }
 
-void CAddonDatabase::SyncInstalled(const std::set<std::string>& ids, const std::set<std::string>& enabled)
+void CAddonDatabase::SyncInstalled(const std::set<std::string>& ids,
+                                   const std::set<std::string>& system,
+                                   const std::set<std::string>& optional)
 {
   try
   {
@@ -240,8 +242,15 @@ void CAddonDatabase::SyncInstalled(const std::set<std::string>& ids, const std::
     std::string now = CDateTime::GetCurrentDateTime().GetAsDBDateTime();
     BeginTransaction();
     for (const auto& id : added)
+    {
+      int enable = 0;
+
+      if (system.find(id) != system.end() || optional.find(id) != optional.end())
+        enable = 1;
+
       m_pDS->exec(PrepareSQL("INSERT INTO installed(addonID, enabled, installDate) "
-          "VALUES('%s', 0, '%s')", id.c_str(), now.c_str()));
+        "VALUES('%s', %d, '%s')", id.c_str(), enable, now.c_str()));
+    }
 
     for (const auto& id : removed)
     {
@@ -250,7 +259,7 @@ void CAddonDatabase::SyncInstalled(const std::set<std::string>& ids, const std::
       DeleteRepository(id);
     }
 
-    for (const auto& id : enabled)
+    for (const auto& id : system)
       m_pDS->exec(PrepareSQL("UPDATE installed SET enabled=1 WHERE addonID='%s'", id.c_str()));
 
     CommitTransaction();
--- a/xbmc/addons/AddonDatabase.h
+++ b/xbmc/addons/AddonDatabase.h
@@ -129,7 +129,9 @@ public:
   /*! Clear internal fields that shouldn't be kept around indefinitely */
   void OnPostUnInstall(const std::string& addonId);
 
-  void SyncInstalled(const std::set<std::string>& ids, const std::set<std::string>& enabled);
+  void SyncInstalled(const std::set<std::string>& ids,
+                     const std::set<std::string>& system,
+                     const std::set<std::string>& optional);
 
   void GetInstalled(std::vector<ADDON::CAddonBuilder>& addons);
 
--- a/xbmc/addons/AddonManager.cpp
+++ b/xbmc/addons/AddonManager.cpp
@@ -54,6 +54,7 @@
 #include "utils/log.h"
 #include "utils/StringUtils.h"
 #include "utils/XBMCTinyXML.h"
+#include "utils/XMLUtils.h"
 #include "ServiceBroker.h"
 
 using namespace XFILE;
@@ -190,7 +191,7 @@ void CAddonMgr::FillCpluffMetadata(const cp_plugin_info_t* plugin, CAddonBuilder
   }
 }
 
-static bool LoadManifest(std::set<std::string>& addons)
+static bool LoadManifest(std::set<std::string>& system, std::set<std::string>& optional)
 {
   CXBMCTinyXML doc;
   if (!doc.LoadFile("special://xbmc/system/addon-manifest.xml"))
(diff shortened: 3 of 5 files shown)


stefansaraev:
stupid question: what happens on further upgrade, if user has disabled one or more optional addons, but those are set to non-optional next release ? is it forced enabled in db or left as-is ?

garbear:
they are left as is. the only time Kodi considers the optional add-ons is when syncing new add-ons to the database. If the add-on is already in the database, it doesn't matter if it's optional or not.

stefansaraev:
then that will bite us back soon or later

garbear:
I might be wrong. I don't know this system as well as @tamland

xbmc/addons/AddonManager.cpp
```diff
@@ -671,7 +677,11 @@ bool CAddonMgr::FindAddons()
       for (int i = 0; i < n; ++i)
         installed.insert(cp_addons[i]->identifier);
       m_cpluff->release_info(m_cp_context, cp_addons);
-      m_database.SyncInstalled(installed, m_systemAddons);
+
+      std::set<std::string> addonsToEnable(m_systemAddons);
+      addonsToEnable.insert(m_optionalAddons.begin(), m_optionalAddons.end());
+
+      m_database.SyncInstalled(installed, addonsToEnable);
```
tamland: This will set them to enabled in db. I.e your disabled optional addons will be re-enabled on every sync.

garbear: You're right. I'll fix this

garbear:
@tamland I updated the PR. optional add-ons will stay disabled now, I have tested this

xbmc/addons/AddonDatabase.cpp
```diff
@@ -240,8 +242,15 @@ void CAddonDatabase::SyncInstalled(const std::set<std::string>& ids, const std::
     std::string now = CDateTime::GetCurrentDateTime().GetAsDBDateTime();
     BeginTransaction();
     for (const auto& id : added)
+    {
+      int enable = 0;
+
+      if (system.find(id) != system.end() || optional.find(id) != optional.end())
```
garbear: system add-ons are enabled below, so should this be just `if (optional.find(id) != optional.end())`?

garbear: I guess this should stay, let's avoid having system add-ons in a disabled state to begin with

garbear:
@tamland ok to merge? we should get this in before 17 alpha 1 drops

xbmc/addons/AddonDatabase.cpp
tamland: It's covered below, in the same transaction, so it doesn't really do anything.

xbmc/addons/AddonDatabase.h
```diff
@@ -129,7 +129,9 @@ class CAddonDatabase : public CDatabase
   /*! Clear internal fields that shouldn't be kept around indefinitely */
   void OnPostUnInstall(const std::string& addonId);
 
-  void SyncInstalled(const std::set<std::string>& ids, const std::set<std::string>& enabled);
+  void SyncInstalled(const std::set<std::string>& ids,
+                     const std::set<std::string>& system,
+                     const std::set<std::string>& optional);
```
tamland: Please try avoid vertical alignment in the future as it breaks formatting and vcs on every change. it's in our guidelines as well now.

tamland:
@garbear Fine with me if you want it in alpha 1, but let's be perfectly clear here: the way the joystick addon is currently handled really needs to be rethinked. We now have a dependency on a 3rd party addon in core and special handling for one addon.

tamland:
jenkins build this please

garbear:
peripheral.joystick isn't a third party add-on. it is hosted by the organization https://github.com/kodi-game which, like https://github.com/kodi-pvr, is a kodi organization to handle game-related extensions. it is perfectly acceptable to have peripheral.joystick be an optional add-on.

garbear:
should the stewardship of this add-on be moved to the xbmc org? it kinda feels out of place at kodi-game, because although controllers are used for games they can be used without games as well

tamland:
The point wasn't where it's hosted but that it's external, not part of the build/packaging or tree like other addons. I.e there's now a dependency/special handling in core code for it because it's in this weird state where it's optional, yet at the same time it's sort of required (hence why you wanted it to be excepted from the disable rule right?).

garbear:
I understand. This PR was meant to address the issues that the OE/LE packagers brought up, where the system vs non-system add-on dichotomy left no room for shipping add-ons "in between" these two extremes, peripheral.joystick being one but they had many others such add-ons. You suggested rethinking this PR, what approach would you take?

MilhouseVH:
Just to say this PR doesn't really address those issues - in fact `peripheral.joystick` was never really an issue for OE/LE as OE/LE builders can and do include it in the `addon-manifest.xml` because they know they're including ("shipping") the add-on within the build. Consequently when Addons*.db is built from scratch `peripheral.joystick` is correctly enabled, with or without this PR, at least in OE/LE (but thanks anyway!)

The issue that still exists even after this PR is that of non-system addons that the builder includes/ships for convenience, but which he/she may not want to be enabled by default (eg. PVR/AudioDecoder addons). Or addons that the builder does want enabled by default, but which can still be disabled by the user (eg. inputstream.*).

Right now there's no way for the builder to have this fine grained control over non-system addons that are shipped as part of a distribution, unless the builder adds the addons to `addon-manifest.xml` in which case the addons are always and permanently enabled, because they're now treated as "system addons" even though they are nothing of the sort - abusing `addon-manifest.xml`.

And with the current addons system, all non-system addons (shipped with the build, or user installed) will be disabled by default when rebuilding the Addons*.db (ie. first boot, or to resolve an issue). One could argue this is the correct status for the user installed addons, but for those built-in addons perhaps the builder might want the final word.

Actually, another nice feature (in addition to being able to control the default enabled/disabled state) would be for the builder to be able to configure a built-in addon such that it never updates from an upstream repository - ie. locking the built-in addon version for the lifetime of the build in which it shipped. Or maybe that's just a feature I'd like to see, from a nightly testing perspective... :)

Imagine an `appliance-manifest.xml` where two PVR and two inputstream addons have been shipped as part of a build:
```
<addons>
  <addon allowupdate=false enabled=false>pvr.argustv</addon>
  <addon allowupdate=false enabled=false>pvr.demo</addon>
  <addon allowupdate=false>inputstream.mpd</addon>
  <addon allowupdate=false>inputstream.smoothstream</addon>
</addons>  
```
Defaults for `allowupdate` and `enabled` would both be `true`.

None of these addons would be able to update to any newer version that may be found out on the internet. The PVR addons would be disabled at first boot (or db rebuild), while the inputstream addons would both be enabled.

Sorry, this turned into a bit of a ramble... :)

Razzeee:
Btw has this been documented or does this need the documentation needed tag?

MartijnKaijser:
@tamland @garbear not sure if this PR or another however on a fresh install on Android every single PVR addon is enabled. This doesn'y happen on a fresh Windows install.

PR #9707, opened 2016-04-29
[pvr] - make running instance aware of changed settings

FernetMenta:
fix, this got broken with rework of PVRManager

@ksooo, thanks for bringing this up


ksooo:
Thanks, Rainer. What is not clear to me is how this is supposed to work. When gets the new method called and for what purpose?

FernetMenta:
If GetRunningInstance is not overridden by an addon, AddonManager and the dialog operate on a new uninitialized instance of an addon. If an addon is not initialized, TransferSettings is not called and nobody gets notified.

PR #9708, opened 2016-04-29
[adsp] remove usage of list item entries on skins

AlwinEsch:
This remove the set of list entries on skin itself and do this on code side.

Further becomes with this way only the Mode type shown who is present.

Here screenshots about the changes
screenshot

screenshot1

screenshot2

### DIFF
--- a/xbmc/settings/dialogs/GUIDialogAudioDSPManager.cpp
+++ b/xbmc/settings/dialogs/GUIDialogAudioDSPManager.cpp
@@ -55,15 +55,17 @@ typedef struct
 {
   const char* sModeType;
   int iModeType;
+  int iName;
+  int iDescription;
 } DSP_MODE_TYPES;
 
 static const DSP_MODE_TYPES dsp_mode_types[] = {
-  { "preprocessing",    AE_DSP_MODE_TYPE_PRE_PROCESS },
-  { "inputresampling",  AE_DSP_MODE_TYPE_INPUT_RESAMPLE },
-  { "masterprocessing", AE_DSP_MODE_TYPE_MASTER_PROCESS },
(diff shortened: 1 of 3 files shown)


AlwinEsch:
@phil65 @roni @AchimTuran can check them.

ronie:
nice, skin changes are ok.

phil65:
Nice, thx for goin after my request that fast. :)

xbmc/settings/dialogs/GUIDialogAudioDSPManager.cpp
```diff
@@ -704,14 +706,24 @@ void CGUIDialogAudioDSPManager::Update()
...
     modes.clear();
     db.GetModes(modes, iModeType);
 
     // No modes available, nothing to do.
     if (!modes.empty())
     {
+      CFileItemPtr item(new CFileItem());
+      item->SetLabel(g_localizeStrings.Get(dsp_mode_types[i].iName));
+      item->SetLabel2(g_localizeStrings.Get(dsp_mode_types[i].iDescription));
+      item->SetProperty("currentMode", dsp_mode_types[i].sModeType);
```
AchimTuran: Great idea @AlwinEsch, it's much better to do set this in our core codebase instead in every skin.

AchimTuran:
I guess something similiar like `[estuary] remove no more needed adsp manager items` would also be needed for the skin confluence, because now the current mode property is set at the C/C++ side and not in the skin right?

AlwinEsch:
Already a pull request on confluence: https://github.com/xbmc/skin.confluence/pull/9 :D

If I'm not wrong, works it also if the set is still is still in xml and the values from Kodi becomes overwritten.

AlwinEsch:
jenkins build this please

PR #9715, opened 2016-04-30
[guilib] - fix for multiimage controls in case imagepath switches to empty

phil65:
At the moment the content of multiimage controls does not clear when switching to an empty path, so you still get the image(s) of the previous path. 
I think @BigNoid also stumbled over this one already.  
@ronie for confirmation and perhaps @mkortstiege for code in case you have some minutes :)


ronie:
confirmed. works as expected.

mkortstiege:
You might want to invalidate m_image in order to properly proceed with the new image width. Might not be needed though. Other than that, it looks good to me.

phil65:
As discussed via Slack, that shouldnt be needed. Thx for reviewing, jenkins build this please.

PR #9719, opened 2016-05-01
VideoPlayer: drop render method SW

FernetMenta:
see title
Doing YUV to RGB conversion by CPU is rather outdated and I see no need to maintain support for this.

@afedchin could you have a look at the Windows changes?

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoRenderers/BaseRenderer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/BaseRenderer.cpp
@@ -567,6 +567,5 @@ void CBaseRenderer::SettingOptionsRenderMethodsFiller(const CSetting *setting, s
 #ifdef HAS_GL
   list.push_back(make_pair(g_localizeStrings.Get(13417), RENDER_METHOD_ARB));
   list.push_back(make_pair(g_localizeStrings.Get(13418), RENDER_METHOD_GLSL));
-  list.push_back(make_pair(g_localizeStrings.Get(13419), RENDER_METHOD_SOFTWARE));
 #endif
 }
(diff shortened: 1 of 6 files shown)


afedchin:
changes look good. but I would keep SW for at least for windows. it solves rendering issues on some bad drivers.

FernetMenta:
I can leave SW for Windows but I seriously doubt that nowadays any driver or hardware has an issue with shaders. More likely that we do something wrong in our implementation.

afedchin:
I thought also what there are no problems with shaders, but there are several reports at the forum which show that some drivers have an issue with shaders, the issue looks like yuv2rgb shader ignores V surface.

FernetMenta:
@afedchin I can't imagine that a driver fails in such very basic behaviour. IMO more likes that our code misses some condition.
We keep render_sw for Windows until this issue can be resolved, driver or Kodi.

jenkins build this please

FernetMenta:
jenkins build this please

arnova:
Isn't it handy to keep this for debugging/isolating issues ?

FernetMenta:
I never used it in all the years. If anybody wants to use it for what purpose ever, git won't forget it.

xbmc/cores/VideoPlayer/VideoRenderers/BaseRenderer.cpp
```diff
@@ -567,6 +567,5 @@ void CBaseRenderer::SettingOptionsRenderMethodsFiller(const CSetting *setting, s
 #ifdef HAS_GL
   list.push_back(make_pair(g_localizeStrings.Get(13417), RENDER_METHOD_ARB));
   list.push_back(make_pair(g_localizeStrings.Get(13418), RENDER_METHOD_GLSL));
-  list.push_back(make_pair(g_localizeStrings.Get(13419), RENDER_METHOD_SOFTWARE));
```
MartijnKaijser: please also drop the string from language as it doesn't seem used anymore

FernetMenta: still needed for Windows

PR #9720, opened 2016-05-01
VideoPlayer: bump framebuffer object to 16 bit

FernetMenta:
@laurimyllari please have a look at this
Can we move yuv2rgb conversion back to pass #1 with this?


FernetMenta:
jenkins build this please

laurimyllari:
I tested this both with and without the 16bit bump using the NTSC Digital Video Essentials. The 16bit FBO fixes banding in full range output with advanced scaling.

Looks good to me.

FernetMenta:
@laurimyllari does this mean limited/full range conversion can be moved to colour conversion matrix?

laurimyllari:
Yes, I think it can be moved. I would like to have an override for the 3dlut case though, but I can work on that.

FernetMenta:
@laurimyllari I would just set a flag like for the single pass method: https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp#L909
3dlut can of course handle it differently

laurimyllari:
That looks convenient

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

PR #8807, opened 2016-01-08
Controller input

garbear:
Here it is... the first half of RetroPlayer!

Relavant links:
- Guide
  (pics!): [Guide to configure controllers in Kodi](http://forum.kodi.tv/showthread.php?tid=252465)
- Documentation: [Controller Input for Emulator Development](http://forum.kodi.tv/showthread.php?tid=257764)
- Code (github can't get the commit order right): [commits](https://github.com/garbear/xbmc/commits/controller-input)
- Forum thread for non-code discussion: [Input](http://forum.kodi.tv/showthread.php?tid=211138)
- Some fancy flow charts: [The lifetime of a button press for peripherals with input](http://forum.kodi.tv/showthread.php?tid=230276)
- Evolution of the controller dialog: [Evolution of the controller window (pics!)](http://forum.kodi.tv/showthread.php?tid=258827)

Controller window

This PR contains a new system for controller and keyboard input used by RetroPlayer. Games aren't included in this PR, so you'll have to use a RetroPlayer build to test the full extent of the input system. Still, the input system offers many improvements in Kodi outside of games, and lets us start collecting button map data for when games hit.

This PR is accompanied by a binary add-on, [peripheral.joystick](https://github.com/kodi-game/peripheral.joystick). It warrants review alongside the PR. This add-on was created to contain all of Kodi's platform-specific joystick code and keymap data.

Here are the blocking issues:
- [x] Compilation on all platforms (RPi needs an `#ifdef HAS_LIBUDEV` around the touchscreen fix)
- [x] Android support ([thread](http://forum.kodi.tv/showthread.php?tid=246680)) (thanks montellese)
- [x] The controller window's "get more" button fails silently ([report](http://forum.kodi.tv/showthread.php?tid=173361&pid=2202355#pid2202355))
- [x] Breaks some EventServer clients (Apple IR Remotes, some Yatse buttons, etc) ([report](http://forum.kodi.tv/showthread.php?tid=173361&pid=2196239#pid2196239))
- [x] Erratic button presses while mapping buttons ([report](http://forum.kodi.tv/showthread.php?tid=173361&pid=2178873#pid2178873))
- [x] Empty keyboard settings breaks GUI ([report](https://github.com/xbmc/xbmc/pull/8807#issuecomment-171599339)) (thanks montelllese)
- [x] Missing scroll bars in controller dialog (thanks hitcher)
- [x] Broken volume commands ([report](https://github.com/xbmc/xbmc/pull/8807#issuecomment-172169117))
[...]

PR #9723, opened 2016-05-01
[OSX] - second try to properly fix apple remote and harmony support, aka - fix the regression of the regression

Memphiz:
@Tolriq fyi

@MilhouseVH  fyi

This should truly seperate the event server keymaps which we use on osx via XBMCHelper EventClient from the rest of the keymappig. It introduces the "concept" of CustomControllers which are a bunch of buttons with IDs which can be mapped to actions.


Memphiz:
@NedScott fyi for you too - hope you didn't do the wiki thing yet - turns out to be a more bigger issue with those remote thingys. Lets hope we can come to a conclusion in this PR ...

MilhouseVH:
@Memphiz: Seems to be working fine so far - many thanks.

xbmc/network/EventClient.cpp
```diff
@@ -115,6 +115,12 @@ void CEventButtonState::Load()
         - (unsigned char)'0'; // convert <num> to int
       m_joystickName = m_joystickName.substr(2); // extract joyname
     }
+    
+    if (m_mapName.length() > 3 &&
+        (StringUtils::StartsWith(m_mapName, "CS")) ) // custom map - CS:<controlelrName>
```
Tolriq: Small typo, and maybe use CC to be consistent with customcontroller map name ?

Tolriq:
Looks good, no sure there's a reason to keep the isjoystick, all that part could be dropped + the JS handling could be removed as no more used.

When I restored the ES in previous PR it was a quick and dirty rollback to restore function, not a proper cleanup as I had no idea what would be the communication, long term decision.

I guess decision it's full removal now :)

Memphiz:
@Tolriq last 2 commits would remove joystick via eventserver - not sure if i went over the line with this (axis is pretty sure joystick only - not sure about that amount thingy though) - @garbear fyi rfc

Tolriq:
@Memphiz yeah not sure about amount too :( I do not know if someone could use it to send like 10 times same commands.

Never tested so not even sure it worked before for something else than joystick.

bas-t:
I'm building this now, to verify that my mceusb will still work as expected in linux.
(without using ccache, so it'll take some time...)
Will report back shortly.

bas-t:
Looking good to me, it does not affect my mceusb.
Thanks for the fix!

xbmc/input/InputManager.cpp
```diff
@@ -190,22 +190,42 @@ bool CInputManager::ProcessEventServer(int windowId, float frameTime)
...
-  unsigned int wKeyID = es->GetButtonCode(joystickName, isAxis, fAmount);
+  unsigned int wKeyID = es->GetButtonCode(strMapName, isAxis, fAmount, isJoystick);
 
   if (wKeyID)
   {
-    if (joystickName.length() > 0)
+    if (strMapName.length() > 0)
     {
-      return false;
+      // joysticks ar not supported via eventserver
```
garbear: small typo. and point of clarification - eventserver used to support two kinds of joystick, the first was a 360-style that used our deceased joystick system and the second is an original-xbox style that is handled directly below this code.

The original xbox style is still supported, it uses the [gamepad.xml](https://github.com/xbmc/xbmc/blob/master/system/keymaps/gamepad.xml) keymap so I guess it would be called "gamepad" instead of "joystick".  You might want to mention that joysticks are still supported, but as xbox-style "gamepads".

garbear:
you should drop your last commit removing anything that sounds like joystick. trust me, that was my exact strategy that i used for #8807 :) I was purposefully over-aggressive, and that was how I discovered that EventServer had two kinds of joysticks - the 360-style "joystick" and the xbox-style "gamepad". Clearly someone tacked on a new API without bothering to deprecate the old one.

The remaining "gamepad" api is still used for some smartphone remotes. It took a couple tries, but eventually I was able to restore just enough joystick code to fix gamepads. So everything I left behind is strictly necessary to not break smartphone remotes.

MartijnKaijser:
@Memphiz there seems to be something wrong with inputstream addon so can't release first alpha. Might give you some time to get this in.

Memphiz:
ok - removed the removal of everything that sounds like joystick ... and fixed the typo - i am stell waiting on test feedback from the users... (the ones WITH apple remote, not the ones who confirm that its not broken for others anymore ;) ).

bas-t:
Regardless of the feedback you are awaiting, this should be merged.
If only to unbreak all other remotes.

NedScott:
I tested the latest test build from Memphiz ( http://mirrors.kodi.tv/test-builds/osx/x86_64/kodi-20160503-97e0a6f-osxremote_as_keyboard-x86_64.dmg ) and my Apple remote (silver) worked perfectly.

Memphiz:
Some issue with Volume control left - will investigate when i am back home on the Weekend - good for now - jenkins build and merge

PR #9724, opened 2016-05-01
rbp: Fix up settings after layout changes

popcornmix:
The 3D settings got lost in the refactor.
Also fix a couple of cosmetics.

### DIFF
--- a/system/settings/rbp.xml
+++ b/system/settings/rbp.xml
@@ -28,11 +28,18 @@
           <control type="edit" format="integer" />
         </setting>
       </group>
+      <group id="5">
+        <setting id="videoplayer.supportmvc" type="boolean" label="38027" help="38028">
+          <level>2</level>
+          <default>true</default>
+          <control type="toggle" />
+        </setting>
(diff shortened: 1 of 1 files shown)


popcornmix:
ping @jjd-uk

system/settings/rbp.xml
```diff
@@ -27,12 +27,17 @@
           </control>
           <control type="edit" format="integer" />
         </setting>
+        <setting id="videoplayer.supportmvc" type="boolean" label="38027" help="38028">
```
jjd-uk: I think this belongs with the other 3D settings under:

 group id="5" label="14232"

As a follow up I will then look at whether the label for that group needs to change as it's currently "Stereoscopic 3D".

jjd-uk:
Oops sorry about that, not sure how that happened. 
Other than change the group "videoplayer.supportmvc" belongs to, it fits with what I've done so +1 to shove straight in.

popcornmix:
Moved the section. No need for jenkins.

PR #9727, opened 2016-05-01
[python] - whitelist some GUI_MSG_NOTIFY_ALL messages for WindowXML windows

phil65:
Turns out that intercepting all GUI_MSG_NOTIFY_ALL  messages , as done in https://github.com/xbmc/xbmc/pull/9605, was a bit too restrictive, since we need GUI_MSG_PAGE_CHANGE  for scrollbar navigation. This fixes it by whitelisting some workin messages ( think that´s easier than blacklisting in this case).
@ronie and @tamland


ronie:
scrollbars are happy again :-)

phil65:
jenkins build this please.

PR #9730, opened 2016-05-01
[cmake] Sync VTB changes

fetzerch:
Fix broken build after https://github.com/xbmc/xbmc/pull/9702/


FernetMenta:
@fetzerch Sorry for this but I think those mistakes will continue to happen if we don't sunset XCode and VS project.
Can you make the cmake generated project tree look like the file system?

fetzerch:
@FernetMenta: Unfortunately not so easy with the current project setup of building small static libs. But it's one of the next things on my todo list after I finish with packaging. (osx is done, ios and android should be doable in the same way, but windows will take a bit).

PR #9731, opened 2016-05-02
3dlut and simple ICC linking support for color correction

laurimyllari:
This PR adds a ColorManager class that supports loading madVR 3DLUT files, and loading ICC display profiles, creating source profiles and linking them.

The end user can use this to correct their display response with a 3DLUT file, or emulate other displays (with whitepoint, primaries and gamma selectable at runtime) with an ICC profile of their own display.

Code changes are in two commits. The first one implements a (hopefully) platform independent ColorManager class. If lcms2 is not available, only 3DLUT file format is supported. The second commit adds the necessary bits to the Linux OpenGL renderer, building on the previously added output stage.

The ICC profile linking is still work in progress, but I included it as it is functional and shows what can be done with it. It will also be needed if/when color management is implemented for the photo viewer. The 3DLUT files provide the best accuracy (at the cost of runtime flexibility) - ArgyllCMS is very smart and the linker here is quite simple.

Slight adjustments may be needed for https://github.com/xbmc/xbmc/pull/9720 - I'll update if/when that is merged to master.

One significant improvement I would like help with: it would be nice to keep the 3D lookup texture around as it takes a while to upload. It is most noticeable when running many short video clips back to back. @FernetMenta, any thoughts?

I'm still pretty new to the Kodi codebase. Any feedback is welcome.

Pinging @Memphiz and @popcornmix since my previous work broke gles - would you mind checking this out? Are there any gles platforms that could be supported, or better to just disable this and make sure not to break anything?


system/shaders/output.glsl
```diff
@@ -3,11 +3,20 @@ uniform sampler2D m_dither;
...
+uniform float     m_CLUTsize;
+uniform sampler3D m_CLUT;
+#endif
 
 void main()
 {
   vec4 rgb        = process();
 
+#if defined(XBMC_3DLUT)
+  // FIXME: can this be optimized?
```
Memphiz: KODI_3DLUT please

xbmc/ColorManager.cpp
```diff
@@ -0,0 +1,531 @@
+#include <boost/algorithm/clamp.hpp>
```
Memphiz: missing gpl header

Memphiz: and we don't have boost anymore iirc

xbmc/video/dialogs/GUIDialogCMSSettings.cpp
```diff
@@ -0,0 +1,243 @@
+/*
+ *      Copyright (C) 2005-2014 Team XBMC
```
Memphiz: should by kodified

Memphiz:
Well if it can't be done for gles with the human ressources this pr has available ( you ;) ) - then it should be as defensive as possible for gles. As of the looks it doesn't touch gles - so theoretically this should be ok.

xbmc/ColorManager.cpp
FernetMenta: Looks like VideoRederer is the only client of this class. Then it should be moved to this component.

FernetMenta:
nice!

laurimyllari:
I updated the code to address the feedback above, and to apply to current master.

xbmc/cores/VideoPlayer/VideoRenderers/ColorManager.cpp
```diff
@@ -0,0 +1,550 @@
...
+bool CColorManager::IsEnabled()
+{
+  //TODO: check that the configuration is valid here (files exist etc)
+
+  return CSettings::GetInstance().GetBool("videoscreen.cmsenabled");
+}
+
+CMS_PRIMARIES videoFlagsToPrimaries(int flags)
+{
+  if (flags & CONF_FLAGS_COLPRI_BT709) return CMS_PRIMARIES_BT709;
```
FernetMenta: return statement on new line please

xbmc/cores/VideoPlayer/VideoRenderers/ColorManager.cpp
```diff
@@ -0,0 +1,550 @@
...
+      cmsSetAdaptationState(0.0);
+#ifdef _DEBUG
+      int64_t start;
+      start = CurrentHostCounter();
+#endif
+      cmsHTRANSFORM deviceLink =
+        cmsCreateTransform(sourceProfile, TYPE_RGB_FLT,
+            m_hProfile, TYPE_RGB_FLT,
+            INTENT_ABSOLUTE_COLORIMETRIC, 0);
+#ifdef _DEBUG
```
FernetMenta: please drop those blocks. this won't ever get used. devs who are interested can add the measurements

xbmc/cores/VideoPlayer/VideoRenderers/ColorManager.cpp
```diff
@@ -0,0 +1,550 @@
...
+};
+
+bool CColorManager::Probe3dLut(const std::string filename)
+{
+    struct H3DLUT header;
+    CFile lutFile;
+
+    if (!lutFile.Open(filename))
+    {
+        CLog::Log(LOGERROR, "%s: Could not open 3DLUT file: %s", __FUNCTION__, filename.c_str());
```
FernetMenta: wrong indent

xbmc/cores/VideoPlayer/VideoRenderers/ColorManager.h
```diff
@@ -0,0 +1,173 @@
...
+};
+
+class CColorManager
+{
+public:
+  /*!
+   \brief Access the global color management system
+   \return the global instance
+   */
+  static CColorManager& Get();
```
FernetMenta: no singletons, this is an anti pattern. embed the object where you want to use it.

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/GLSLOutput.cpp
```diff
@@ -25,26 +25,36 @@
...
   m_hDitherSize  = -1;
+  m_hCLUT        = -1;
+  m_hCLUTSize    = -1;
 
   m_dither = useDithering;
   m_ditherDepth = ditherDepth;
   m_fullRange = fullrange;
+  // make sure CMS is enabled - this allows us to keep the texture
+  // around to quickly switch between CMS on and off
+  m_3DLUT = CColorManager::Get().IsEnabled() && (clutTex > 0);
```
FernetMenta: pass a reference to CM instead of using a global

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/GLSLOutput.cpp
```diff
@@ -69,6 +81,11 @@ void GLSLOutput::OnCompiledAndLinked(GLuint programHandle)
     m_hDitherQuant = glGetUniformLocation(programHandle, "m_ditherquant");
     m_hDitherSize = glGetUniformLocation(programHandle, "m_dithersize");
   }
+  //   3DLUT
+  if (m_3DLUT) {
```
FernetMenta: braces at new line

xbmc/settings/DisplaySettings.cpp
```diff
@@ -767,3 +798,59 @@ void CDisplaySettings::ClearCustomResolutions()
...
+}
+
+void CDisplaySettings::SettingOptionsCmsGammaModesFiller(const CSetting *setting, std::vector< std::pair<std::string, int> > &list, int &current, void *data)
+{
+  const static std::string cmsGammaModeLabels[] = {
+    "BT.1886",
+    "Input offset",
+    "Output offset",
+    "Absolute",
+  }; // FIXME: should be moved to ColorManager.h?
```
FernetMenta: who is supposed to answer this?

xbmc/video/dialogs/GUIDialogCMSSettings.h
```diff
@@ -0,0 +1,52 @@
...
+
+class CGUIDialogCMSSettings : public CGUIDialogSettingsManualBase
+{
+public:
+  CGUIDialogCMSSettings();
+  virtual ~CGUIDialogCMSSettings();
+
+protected:
+  // implementations of ISettingCallback
+  virtual void OnSettingChanged(const CSetting *setting);
```
FernetMenta: use override key word for overrides

xbmc/video/dialogs/GUIDialogCMSSettings.cpp
laurimyllari: @Memphiz what's the right way to do this? I'm hesitant to touch the copyright notices for others :)

laurimyllari:
Updated to remove the singleton. @FernetMenta - would you mind giving this another look?

xbmc/video/dialogs/GUIDialogCMSSettings.cpp
Memphiz: Copyright (C) 2016 - your name - then exchange everything that points to XBMC in the rest of the header to Kodi and kodi.tv - in case you don't want to put your name in there you can also set copyright to Team Kodi

laurimyllari: Thank you! I just set it to Team Kodi since most of the file is copied directly from another dialog.

FernetMenta:
@laurimyllari sorry, I almost missed this. I will have a look this weekend.

xbmc/cores/VideoPlayer/VideoRenderers/ColorManager.h
```diff
@@ -0,0 +1,165 @@
...
+  bool Create3dLut(cmsHTRANSFORM transform, uint16_t **clutData, int *clutSize);
+
+  // keep current display profile loaded here
+  cmsHPROFILE m_hProfile;
+  cmsCIEXYZ   m_blackPoint = { 0, 0, 0 };
+
+#endif // HAVE_LCMS2
+
+  // current configuration:
+  CMS_PRIMARIES curVideoPrimaries;
```
FernetMenta: we prefix members with m_

xbmc/cores/VideoPlayer/VideoRenderers/ColorManager.cpp
```diff
@@ -0,0 +1,511 @@
...
+#include "cores/VideoPlayer/VideoRenderers/RenderFlags.h"
+#include "filesystem/File.h"
+#include "settings/Settings.h"
+#include "utils/log.h"
+#include "utils/TimeUtils.h"
+
+using namespace XFILE;
+
+CColorManager::CColorManager() :
+  m_hProfile(NULL)
```
FernetMenta: this member is only available with LCSM2, right? did you try to compile with LCSM2 disabled?

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
```diff
@@ -186,6 +193,11 @@ CLinuxRendererGL::~CLinuxRendererGL()
     delete m_pYUVShader;
     m_pYUVShader = NULL;
   }
+
+  if (m_ColorManager)
```
FernetMenta: why do you check this being not zero?

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
```diff
@@ -713,6 +741,8 @@ void CLinuxRendererGL::UpdateVideoFilter()
   bool pixelRatioChanged    = (CDisplaySettings::GetInstance().GetPixelRatio() > 1.001f || CDisplaySettings::GetInstance().GetPixelRatio() < 0.999f) !=
                               (m_pixelRatio > 1.001f || m_pixelRatio < 0.999f);
   bool nonLinStretchChanged = false;
+  bool cmsChanged           = (m_cmsOn != m_ColorManager->IsEnabled())
```
FernetMenta: that results in a call so settings every cycle
if ColorManager is not enabled, I see not need to check its configuration

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.h
```diff
@@ -278,6 +279,17 @@ class CLinuxRendererGL : public CBaseRenderer
   bool  m_nonLinStretch;
   bool  m_nonLinStretchGui;
   float m_pixelRatio;
+
+  // color management
+  CColorManager *m_ColorManager;
```
FernetMenta: please use unique_ptr, are stopped making use unmanaged pointers.

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.h
```diff
@@ -278,6 +279,17 @@ class CLinuxRendererGL : public CBaseRenderer
...
+  GLuint    m_tCLUTTex;
```
FernetMenta: please no vertical alignment

xbmc/video/dialogs/GUIDialogCMSSettings.cpp
```diff
@@ -0,0 +1,243 @@
...
+ */
+
+// FIXME: clean up includes
+#include "system.h"
+#include "cores/VideoPlayer/VideoRenderers/ColorManager.h"
+#include "FileItem.h"
+#include "GUIDialogCMSSettings.h"
+#include "GUIPassword.h"
+#include "addons/Skin.h"
+#ifdef HAS_VIDEO_PLAYBACK
```
FernetMenta: HAS_VIDEO_PLAYBACK is outdated, please drop

FernetMenta:
apart from some minors (see comments) look ok. first round for jenkins

jenkins build this please

FernetMenta:
@laurimyllari check jenkins log, build failed on all platforms.

laurimyllari:
@FernetMenta thanks for the review. I made a mistake that broke the build without lcms2. I'll fix that and the other issues and will ping you again.

xbmc/cores/VideoPlayer/VideoRenderers/ColorManager.h
laurimyllari: Fixed it. Thanks for your patience with these style issues - hopefully these become natural for me soon.

xbmc/cores/VideoPlayer/VideoRenderers/ColorManager.cpp
laurimyllari: I missed these while working with the lcms2 features and previous fixes. I found a few similar issues - I'll make sure to test with lcms2 disabled before requesting another review.

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
laurimyllari: Out of habit. I'll take it out as unnecessary.

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
laurimyllari: My thinking here is that we need to check if cms has been toggled on/off or its configuration has changed. If cms is not on, CheckConfiguration is not called.

The second check misses the case where cms was just switched on and skips the call to CheckConfiguration, but that's ok since it will be checked later to see if a new CLUT texture needs to be loaded.

I did it this way to avoid reloading the CLUT texture when cms is toggled on/off repeatedly.

laurimyllari:
I found an additional issue - the settings were merged into the wrong settings group.

MilhouseVH:
You probably know this already but this won't compile for Raspberry Pi so may need some if-deffery?

```
CPP     xbmc/addons/BinaryAddonCache.o
CPP     xbmc/cores/paplayer/CodecFactory.o
CPP     xbmc/dbwrappers/dataset.o
CPP     xbmc/cores/AudioEngine/Sinks/AESinkNULL.o
GLSLOutput.cpp: In member function 'bool Shaders::GLSLOutput::OnEnabled()':
GLSLOutput.cpp:155:19: error: 'GL_TEXTURE_3D' was not declared in this scope
     glBindTexture(GL_TEXTURE_3D, m_tCLUTTex);
                   ^
GLSLOutput.cpp: In member function 'void Shaders::GLSLOutput::OnDisabled()':
GLSLOutput.cpp:175:15: error: 'GL_TEXTURE_3D' was not declared in this scope
     glDisable(GL_TEXTURE_3D);
               ^
../../../../../Makefile.include:97: recipe for target 'GLSLOutput.o' failed
make[2]: *** [GLSLOutput.o] Error 1
make[2]: Leaving directory '/home/neil/projects/LibreELEC.tv/build.LibreELEC-RPi2.arm-8.0-devel/kodi-17.0-alpha2-e31bee7/xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders'
Makefile:420: recipe for target 'xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoShaders.a' failed
make[1]: *** [xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoShaders.a] Error 2
```

laurimyllari:
Thanks for letting me know, I was not aware of that yet. I'll look for a solution, hopefully something cleaner than #ifndef HAS_GLES

laurimyllari:
I tried to fix the RPi and Windows build issues, and rebased to current master.

MilhouseVH:
Thanks, this PR builds on the RPi now.

FernetMenta:
> Maybe merging the master to my branch would've been the cleaner solution?

downstream always does a rebase

jenkins build this please

MartijnKaijser:
jenkins build this please

MartijnKaijser:
android/osx failed

FernetMenta:
XCode fix: https://github.com/FernetMenta/xbmc/commit/024950c0716f1985ba114520dabb086b0a56d037

laurimyllari:
Can I access the android build log somewhere?

MartijnKaijser:
@laurimyllari http://jenkins.kodi.tv/job/Android-ARM/8421/console

laurimyllari:
Thanks!

Looks like std::to_string is not available for some reason.

fritsch:
Yeah, Android does not have it. They ship a cripled C++11 compiler missing
some basics. If you want to workaround that you need to use the old
std::stringstream approach :-(

2016-06-13 19:46 GMT+02:00 Lauri Myllari <notifications@github.com>:

> Thanks!
>
> Looks like std::to_string is not available for some reason.
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/xbmc/xbmc/pull/9731#issuecomment-225655807>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ABCfHdoz-OKDOYL2LOgX9PsmnoMXvMDYks5qLZdmgaJpZM4IVHQR>
> .
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

stefansaraev:
StringUtils::Format() ?

fritsch:
As this issue appears again and again it might be worth it to add a toString method to StringUtils that just uses to_string everywhere and something else on the other platforms

FernetMenta:
just define to_string for Android (and all the crappy other platforms that come without it)

hudokkow:
If you're going that route, it might be worth to do the same for a couple more missing bits. std::stoi and std::stol come to mind.

FernetMenta:
@laurimyllari do you have an update? Merge window ends tomorrow for this months. Next chance to get this in is July 1st - 20th

laurimyllari:
@FernetMenta I haven't had a chance to look into the fixes for Android builds. Should I #ifdef out the problematic part, or wait until we can fix the string issues for Android?

FernetMenta:
do it the other way round. include only if defined HAS_GL

laurimyllari:
I think xbmc/video/dialogs/GUIDialogCMSSettings.cpp should be excluded from the build if OpenGL is not available. I'm not sure how to do that properly though.

FernetMenta:
That's not trival with Kodi's gazillion build systems. We can do this after we have reduced those to cmake only. For now I would place an ifdef around the body like we do in other files: https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp#L28

laurimyllari:
Thanks! Would you please trigger a jenkins build if this looks ok?

fritsch:
jenkins build this please

MilhouseVH:
This PR still needs rebasing - there are problems with `xbmc/guilib/WindowIDs.h`, `xbmc/input/ButtonTranslator.cpp` and `Kodi.xcodeproj/project.pbxproj`.

laurimyllari:
I rebased to current master and didn't see any obvious issues, but I am unable to test at the moment.

MilhouseVH:
Thanks, builds OK for RPi and x86 (LibreELEC).

FernetMenta:
jenkins build this please

MilhouseVH:
This PR is causing an issue on RPi/RPi2: http://forum.kodi.tv/showthread.php?tid=269814&pid=2362626#pid2362626

The "Color management" option is present in the Video OSD menu - it doesn't do anything when clicked, and the two options above it cannot be accessed.

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -18502,7 +18502,188 @@ msgctxt "#36552"
 msgid "Defines the number of presentation buffers used by the graphics driver. Select 2 if the driver uses double buffering or 3 for triple buffering."
 msgstr ""
 
-#empty strings from id 36553 to 36597
+#: system/settings/settings.xml
+msgctxt "#36560"
+msgid "Color management"
```
MilhouseVH: In `en_gb`, shouldn't this should be `Colour management`? Same for all other uses of `color/colors` => `colour/colours`.

MartijnKaijser: @jjd-uk ping for grammar check for proper British ;)

popcornmix: It's colour:
```
$ grep -i colour addons/resource.language.en_gb/resources/strings.po | wc -l
13
$ grep -i color addons/resource.language.en_gb/resources/strings.po | wc -l
0
```

laurimyllari:
@MilhouseVH I tried to hide the option from video OSD menu, but I don't have a RPi to test.

MilhouseVH:
> @MilhouseVH I tried to hide the option from video OSD menu, but I don't have a RPi to test.

Thanks, it's better but unfortunately not entirely fixed.

With this updated PR I no longer see the CMS menu option, and I can move the menu selection up to the top menu item (ie. "Audio and subtitle settings") but when moving the menu selection back down I cannot move it past "Video Settings" to access "Toggle audio stream" - it's stuck on "Video settings" (although I can move back up, to the audio option). So however CMS is being disabled or prevented from being shown, it's breaking the menu navigation.

s1

MilhouseVH:
> Thanks, it's better but unfortunately not entirely fixed.

I've been looking at this again, and there's an interaction between this PR and other ADSP commits in my build (specifically, https://github.com/popcornmix/xbmc/commit/3b8d6158d92a620160dd308f85e4191e21f1a964 from https://github.com/popcornmix/xbmc/commits/newclock5) so it's quite likely there's no issue with this PR, but maybe the issue is with the ADSP commit(s). Certainly they're not playing well together.

@popcornmix, @AchimTuran ?

MilhouseVH:
LEibreELEC on x86 has the same Video OSD menu issue, and the x86 build is using the following commits: https://github.com/xbmc/xbmc/compare/xbmc:d9fd7743066445b78ae8bb1b64a0ab3ab079f2aa...AchimTuran:6c265b99e71e32f0b8d3d0dce4ec5bd3ddc8b728

It's not actually necessary to enable ADSP, simply including both this PR and the above ADSP commits will result in broken Video OSD menu navigation.

If I leave out this PR but include the ADSP commits then the OSD menu works normally. If I leave out the ADSP commits but include this PR the OSD menu works normally. But when I include the ADSP commits and also this PR, the OSD menu navigation is broken and I don't know where the problem lies.

Perhaps it's simply a consequence of my naive inclusion of two changes that both touch the same skin code without any consideration of each other... once this PR is merged perhaps the ADSP changes can be rebased to take the CMS menu into consideration, but that's just a guess.

For now I'll drop this PR from my builds - if anyone wishes to test it there's several builds with it included (last is #0624). Hopefully this PR will be merged sooner than later.

MilhouseVH:
OK, this is almost certainly just a patching issue, as the VideoOSD.xml file when using both this PR and the ADSP commits will look like this: https://gist.github.com/MilhouseVH/8ebda3b5e74edf2e534f4bf190152c49 - notice the repeated button id `11105` in lines [313-326](https://gist.github.com/MilhouseVH/8ebda3b5e74edf2e534f4bf190152c49#file-gistfile1-txt-L313-L326)

The reason for the merge wierdness is that I'm using standard `patch` which allows a degree of fuzz when applying the ADSP commits. `git apply` on the other hand will refuse to apply the ADSP commits, which to be honest is what I would have expected to happen when uising `patch` but it tries to be clever and succeeds in applying the patch even though it would have been preferable if it had not. Sometimes fuzz is useful, sometimes not (definitely not in this case).

Argh. Very sorry for all this noise...

FernetMenta:
jenkins build this please

laurimyllari:
@MilhouseVH Thanks for looking into it!

Should I change the IDs to make room for ADSP?

MilhouseVH:
You could, but I think it would be best for this PR to be merged as-is (which will hopefully happen soon), and then for ADSP to rebase as necessary (which will have to happen anyway).

AchimTuran:
@laurimyllari 
> it would be best for this PR to be merged as-is
+1

I will rebase adsp anyway :smile:

MartijnKaijser:
@bkuhls STOP linking this PR please!!!

MartijnKaijser:
@laurimyllari could you please update the TODO strings as we are going to beta it's best that we have proper strings.

laurimyllari:
@MartijnKaijser I've created pull request #10296 with updated strings.

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
```diff
@@ -300,6 +307,22 @@ bool CLinuxRendererGL::Configure(unsigned int width, unsigned int height, unsign
...
 #endif
 
+  // load 3DLUT
+  if (m_ColorManager->IsEnabled())
+  {
+    if (!m_ColorManager->CheckConfiguration(m_cmsToken, m_iFlags))
+    {
+      CLog::Log(LOGDEBUG, "CMS configuration changed, reload LUT");
+      if (!LoadCLUT())
+        return false;
```
FernetMenta: @laurimyllari ColorManager needs to be disabled in case or error or the system won't recover.
I enabled ColorManager without a 3dlut file. After this I had to manually edit guisettings to disable colorManager again

laurimyllari: Do you think it's ok to disable it here?

A user can go in the settings later, re-enable ColorManager and set the 3dlut file.

FernetMenta: here it is error handling last resort. if LoadCLUT fails, there is no chance to recover.
that does not mean that other error handling is not needed

PR #9734, opened 2016-05-02
[settings] cleanup obsolete setting after reshuffle

MartijnKaijser:
@FernetMenta

### DIFF
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -46,17 +46,6 @@
           </constraints>
           <control type="list" format="string" />
         </setting>
-        <setting id="videoplayer.pauseafterrefreshchange" type="integer" parent="videoplayer.adjustrefreshrate" label="13550" help="36165">
-          <level>1</level>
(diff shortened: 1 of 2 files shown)


addons/resource.language.en_gb/resources/strings.po
```diff
@@ -6811,12 +6811,7 @@ msgctxt "#13516"
 msgid "Add art"
 msgstr ""
 
-#empty strings from id 13517 to 13549
-
-#: system/settings/settings.xml
-msgctxt "#13550"
```
FernetMenta: this is still needed: https://github.com/MartijnKaijser/xbmc/blob/setting_cleanup/system/settings/settings.xml#L1882

this is a videoscreen setting now

MartijnKaijser:
updated PR

FernetMenta:
looks good, thanks.

@afedchin are you aware that this is now a videoscreen setting and should be implemented by Windowing if required.

PR #9736, opened 2016-05-02
guilib: update items of GUIMediaWindow after adding a media source

FernetMenta:
see title

### DIFF
--- a/xbmc/addons/GUIWindowAddonBrowser.cpp
+++ b/xbmc/addons/GUIWindowAddonBrowser.cpp
@@ -184,14 +184,17 @@ bool CGUIWindowAddonBrowser::OnClick(int iItem, const std::string &player)
       g_mediaManager.GetNetworkLocations(shares);
       std::string path;
       if (CGUIDialogFileBrowser::ShowAndGetFile(shares, "*.zip", g_localizeStrings.Get(24041), path))
+      {
         CAddonInstaller::GetInstance().InstallFromZip(path);
+      }
     }
     return true;
   }
   if (item->GetPath() == "addons://update_all/")
   {
     UpdateAddons updater;
-    return CGUIDialogBusy::Wait(&updater);
+    CGUIDialogBusy::Wait(&updater);
+    return true;
   }
   if (!item->m_bIsFolder)
   {
@@ -210,7 +213,10 @@ bool CGUIWindowAddonBrowser::OnClick(int iItem, const std::string &player)
     return true;
   }
   if (item->IsPath("addons://search/"))
-    return Update(item->GetPath());
+  {
+    Update(item->GetPath());
+    return true;
+  }
 
   return CGUIMediaWindow::OnClick(iItem, player);
 }
--- a/xbmc/windows/GUIMediaWindow.cpp
+++ b/xbmc/windows/GUIMediaWindow.cpp
@@ -900,7 +900,10 @@ bool CGUIMediaWindow::OnClick(int iItem, const std::string &player)
     else if (!CProfilesManager::GetInstance().GetCurrentProfile().canWriteSources() && !g_passwordManager.IsProfileLockUnlocked())
       return false;
 
-    return OnAddMediaSource();
+    if (OnAddMediaSource())
+      Refresh(true);
+
+    return true;
   }
 
   if (!pItem->m_bIsFolder && pItem->IsFileFolder(EFILEFOLDER_MASK_ONCLICK))


xbmc/windows/GUIMediaWindow.cpp
```diff
@@ -900,7 +900,10 @@ bool CGUIMediaWindow::OnClick(int iItem, const std::string &player)
     else if (!CProfilesManager::GetInstance().GetCurrentProfile().canWriteSources() && !g_passwordManager.IsProfileLockUnlocked())
       return false;
 
-    return OnAddMediaSource();
+    if (OnAddMediaSource())
+      Refresh();
+
+    return true;
```
Tolriq: Should it not return the real result ? (BTW thanks for fixing this irritating bug)

FernetMenta: It is supposed to return true if the action was taken. The result of the action is irrelevant regarding the return value. I am 100% sure that this is done incorrectly at other location because I recently fixed one of those already.
When you return false, the parent window or some other client may try to take the action again.

FernetMenta:
@MartijnKaijser @tamland 

added a fix for GUIWindowAddonBrowser

xbmc/windows/GUIMediaWindow.cpp
DaveTBlake: Wondering if that should be `Refresh(true)` to clear the cache too.

When I had problems with the GUI not updating when adding a new music source it was cache related. I had been debugging, so the list got cached because of time taken, less likely in real life, but could happen with a server delay I guess. Anyway once cached that cache was never cleared or updated despite `GUI_MSG_UPDATE_SOURCES`  causing a refesh of GUIMediaWindow

FernetMenta: Why should we clear the cache after adding things?

DaveTBlake: To see those things you added. Either clear it or update it I guess, I really don't have an opinion which. But if this change is to ensure that the GUI is updated with the __current__ list of items when a source is added then the cache needs to be considered too. At the moment it works by accident in that the source list isn't always cached.

DaveTBlake: Maybe @arnova has a view on the cache side of things, I know he has been working on that recently.

xbmc/addons/GUIWindowAddonBrowser.cpp
```diff
@@ -184,14 +184,18 @@ bool CGUIWindowAddonBrowser::OnClick(int iItem, const std::string &player)
       g_mediaManager.GetNetworkLocations(shares);
       std::string path;
       if (CGUIDialogFileBrowser::ShowAndGetFile(shares, "*.zip", g_localizeStrings.Get(24041), path))
+      {
         CAddonInstaller::GetInstance().InstallFromZip(path);
+        Refresh();
```
tamland: why is this needed?

FernetMenta: the name of the method implies that an addon gets installed and should appear in the list. is this not the case? if vecItems is changed, Refresh needs to,be called

xbmc/windows/GUIMediaWindow.cpp
FernetMenta: this is something completely diffent than the filecache arnova is working on. a cache does not need to be cleared after an item was added. if this was the case, the cache design would be completely wrong.

Montellese: @FernetMenta: it is. It's a list based cache. Whenever a call to `CDirectory::GetDirectory()` takes longer than X milliseconds (where X is 2000 or something) the whole result is cached based on it's path. All (well only the ones to "cacheable paths) calls to `CDirectory::GetDirectory()` check whether there's a cached list for the path being retrieved. If there is, the whole list is loaded from the cache. Whenever an item in the database is added, removed or changed the whole video/music db cache needs to be cleared because we don't know which item is cached for which paths.
So you need to call `Refresh(true)` here.

I've always hated this caching because it's a huge PITA especially when debugging with breakpoints (in which case the limit of X milliseconds is easily reached) but never bothered to replace it with something better.

FernetMenta: In what cases we call Refresh(false) then?

xbmc/addons/GUIWindowAddonBrowser.cpp
tamland: Nah, not needed then. The install happens in the background and it will message back and refresh for that after it's complete.

FernetMenta: I recommend making this a synchronous user interaction. Currently a user don't get proper confirmation on whether the install succeeded or not. It happened several times to me that I missed the short kai toast indicating a failure.

FernetMenta:
updated
jenkins build this please

Montellese:
@FernetMenta: sorry I missed your question concerning when `CGUIMediaWindow::Refresh()` is called without clearing the cache. This happens when e.g. the watched setting is changing from showing all movies to only showing unwatched movies. The path of the items being displayed stays the same but the filtering of the items changed. There are quite a few of these cases (also in the addon browser etc).

FernetMenta:
@Montellese thanks for the update

PR #9737, opened 2016-05-02
cosmetics - display correct window name

ronie:
the first name assigned to WINDOW_SETTINGS_SERVICE is shown in the GUI (when debugging is enabled). it should be servicesettings, not the deprecated networksettings.

@HitcherUK


HitcherUK:
Thanks

PR #9738, opened 2016-05-03
Update settings.xml

Guilouz:
Label must be "Audio" not "Picture" for audio accessibility.


jjd-uk:
Confirmed that string should indeed be Audio which is id 14221, thanks for the PR.

My +1 to merge.

PR #9740, opened 2016-05-03
fix rating / votes infolabels

ronie:
this (accidentally?) slipped in with the settings shuffle and breaks rating / votes in the gui.
see: http://forum.kodi.tv/showthread.php?tid=272322

@jjd-uk


jjd-uk:
Not sure how that happened, maybe another rebase error although I did a diff on the strings in my branch & master, so it's certainly not something I intentionally changed.

PR #9741, opened 2016-05-03
[cmake] Fix *.cmake files install after 3e19e40

hudokkow:
@fetzerch ^^


project/cmake/scripts/linux/install.cmake
```diff
@@ -53,6 +53,8 @@ if(NOT EXISTS ${includedir}/xbmc)
 install(CODE "execute_process (COMMAND ln -sf ${APP_NAME_LC}/ xbmc WORKING_DIRECTORY ${includedir})")
 endif()
 
+install(FILES ${cmake_files} DESTINATION ${libdir}/${APP_NAME_LC})
```
fetzerch: ups, sry. change looks good, but i'd move it to line 41 so that it's together with the other cmake files

hudokkow:
Moved as requested.

PR #9742, opened 2016-05-03
[cmake] Reuse version.txt parsing across buildsystems

hudokkow:
@fetzerch, @wsnipex


wsnipex:
do we really need a new file for this? Whats wrong with macros?

hudokkow:
Nope, no need for a new file.
Not unless we want to follow CMake's *apparent* philosophy of having one file per purpose. Kinda how upstreamed modules are coded and named.

Personally I would prefer one file per purpose but that's just me. It's easier to find what's doing what, easier to reuse, easier to maintain and easier to grasp the concept for a newcomer. Just compare scripts/common and modules/ readability. :wink:
modules/ naming convention instantly conveys what each file code does.

On another note, my slight OCD finds annoying that we have *.cmake files with 3 naming conventions:

- foobar
- foo-bar
- foo_bar

Can we agree on one? Perhaps inline with cmake? FooBar.cmake? Easier to read too...
Anyway, whatever we decide, now is the time to start enforcing it.

Oh, and nothing wrong with macros.

wsnipex:
my personal preference would be to leave the version parsing as macro, since it is totally kodi specific. But if others vote for its own file I'll be ok with it.

As for file naming: Lets make it CamelCase like the Find modules.

hudokkow:
Closed in favour of https://github.com/xbmc/xbmc/pull/9845

PR #9744, opened 2016-05-03
[keymap] remove invalid actions

ronie:
there's no such thing as JUMPSMS1

```
22:28:08 T:139935491103104    INFO: Loading special://xbmc/system/keymaps/universalremote.Harmony.xml
22:28:08 T:139935491103104   ERROR: Keymapping error: no such action 'jumpsms1' defined
22:28:08 T:139935491103104   ERROR: Previous line repeats 3 times.
```

@Memphiz

### DIFF
--- a/system/keymaps/universalremote.Harmony.xml
+++ b/system/keymaps/universalremote.Harmony.xml
@@ -136,7 +136,6 @@
   <MyMusicFiles>
     <universalremote>
       <!-- * clear 	-->      <obc145>Delete</obc145>
-      <!-- 1 		-->      <obc111>JumpSMS1</obc111>
       <!-- 2 		-->      <obc112>JumpSMS2</obc112>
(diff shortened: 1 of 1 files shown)


Memphiz:
i will integrate it in the other pr that is pending because i rename those map files there - is there a replacement for jumpsms?

Memphiz:
i wonder why jumpsms1 is not valid but the other (jumpsms2 and so on) are?

Memphiz:
have it integrated into #9656 (made it number1 so a user who wants to alter it sees that button id 111 is valid) ... thx

Montellese:
@Memphiz: There's probably no `jumpsms1` because on any T9 keyboard the button `1` is used for special characters like dot, comma etc. and we don't support jumping to special characters. `ABC` is on button `2` and so on.

PR #9745, opened 2016-05-03
[aml] Fix includes in amcodec/codec.h to make them relative

codesnake:
Fix includes in amcodec/codec.h to make them relative. This will allow to drop the patches like https://github.com/OpenELEC/OpenELEC.tv/blob/master/projects/WeTek_Core/patches/kodi/0005-aml-Fix-compiler-badness-when-compiling-with-amcodec.patch in OpenELEC while still have successful Android builds with amcodec.


stefansaraev:
manual jenkins android/arm build with this PR + 1015c480966d786eb050ce3d3701cdbf5231d1f5 reverted was succesful. so this is good to go

PR #9746, opened 2016-05-04
Fix CEC not being detected in master tree builds

mdeguzis:
libcec3 3.0.1 would not work.
once I upgraded the build to libcec 3.1.0 it was fine.

If this is coincidence due to this new version using p8-platform as a build-dep (name rebrand only I thought), please just close the PR.

### DIFF
--- a/project/cmake/modules/FindCEC.cmake
+++ b/project/cmake/modules/FindCEC.cmake
@@ -15,7 +15,7 @@
 #   CEC::CEC   - The libCEC library
 
 if(PKG_CONFIG_FOUND)
-  pkg_check_modules(PC_CEC libCEC>=3.0.0 QUIET)
+  pkg_check_modules(PC_CEC libCEC>=3.1.0 QUIET)
 endif()
 
 find_path(CEC_INCLUDE_DIR libCEC/CEC.h


wsnipex:
@opdenkamp ping

stefansaraev:
this makes sense. indeed, libcec 3.1.0 is the minimum required version to build kodi, due to platform -> p8-platform changes.

jenkins build this please

mk01:
@ProfessorKaos64 
3.0.1 runs with krypton perfectly fine - even look here
https://github.com/xbmc/xbmc/blob/master/xbmc/peripherals/devices/PeripheralCecAdapter.cpp#L46

also, between 3.0.1 and 3.1.0 there is no functional change (beside the change platform -> p8-platform). all changes are docs corrections (+marginal update to auto build setup for RPI).

popcornmix:
There are bug fixes and functional changes (including my support for repeating buttons) between 3.0.1 and 3.1.0
https://github.com/Pulse-Eight/libcec/compare/libcec-3.0.1...libcec-3.1.0

mk01:
@popcornmix 
buttons won't make 3.0.1 not running, work again.
(no functional changes affecting sub-version change compatibility)

ProfessorKaos64:
I found what my issue was, though. In mirror Debian packaging from the nightly PPA, I now know that previously libcec3 was the build dep, now it is libcec-dev. In my repository, I had built the libcec packages >= 3.0.x. The kicker here is [libcec3](https://packages.debian.org/stretch/libcec3) is not in jessie/jessie-backports, but libcec-dev >= 3 *is* and installs libcec3v4 as a dep. The default pbuilder apt behavior does not respect apt-prefs/pinning (yes I know I could use pbuilder's gdebi apt-satisfy config, but apt is better IMHO). cec has a binary depends, libcec3 (= ${binary:Version}), for instance, so I suspect that is where the conflict happened, and where CEC was not being detected right.

Sorry for the false alarm.

PR #9747, opened 2016-05-04
VideoPlayer: drop old fps(reconfigure) workaround

FernetMenta:
After RenderManager has got the capability to drain, this old workaround is not needed anymore.

@popcornmix brought this up as a fix for Pi.


popcornmix:
+1
A similar patch was in last night's Milhouse build without obvious complaint.
And it was reported to fix the issue of half rate deinterlace due to renderer having the wrong refresh rate. This PR also fixes the issue.

PR #9748, opened 2016-05-04
Pylock

ghost:
This change implements a processing guard that only allows processing of external calls (e.g. python) during CApplication::FrameMove. This change is necessary because of syncronization issues between Renderer (mainthread) and modifications in GUI wich are done from python thread:

Pls. refer to:
https://github.com/xbmc/xbmc/pull/9698

for an example of the issue.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -262,6 +262,8 @@ CApplication::CApplication(void)
   , m_progressTrackingItem(new CFileItem)
   , m_musicInfoScanner(new CMusicInfoScanner)
   , m_fallbackLanguageLoaded(false)
+  , m_WaitingExternalCalls(0)
+  , m_ProcessedExternalCalls(0)
 {
   m_network = NULL;
   TiXmlBase::SetCondenseWhiteSpace(false);
@@ -305,6 +307,8 @@ CApplication::CApplication(void)
 
   m_muted = false;
   m_volumeLevel = VOLUME_MAXIMUM;
+
+  m_frameMoveGuard.lock();
 }
 
 CApplication::~CApplication(void)
@@ -2622,6 +2626,21 @@ void CApplication::HandleShutdownMessage()
   }
 }
 
+void CApplication::LockFrameMoveGuard()
+{
+  ++m_WaitingExternalCalls;
+  m_frameMoveGuard.lock();
+  ++m_ProcessedExternalCalls;
+  g_graphicsContext.Lock();
+};
+
+void CApplication::UnlockFrameMoveGuard()
+{
+  --m_WaitingExternalCalls;
+  g_graphicsContext.Unlock();
+  m_frameMoveGuard.unlock();
+};
+
 void CApplication::FrameMove(bool processEvents, bool processGUI)
 {
   MEASURE_FUNCTION;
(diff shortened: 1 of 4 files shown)


xbmc/Application.cpp
```diff
@@ -2657,7 +2678,21 @@ void CApplication::FrameMove(bool processEvents, bool processGUI)
...
+
+    // Open the door for external calls e.g python exactly here.
+    // We try to close the door latest after 10 ms.
+    unsigned int maxLoops = 5;
+    m_ProcessedExternalCalls = 0;
+    while ((m_WaitingExternalCalls || m_ProcessedExternalCalls) && maxLoops--)
+    {
+      m_ProcessedExternalCalls = 0;
+      m_frameMoveGuard.unlock();
+      Sleep(2);
```
FernetMenta: this window should start at zero. assume there is only a single python call waiting, that only takes some micro secs to be completed. we can expand the window if there are threads waiting for x times in a row when calling FrameMove.

This code blocks the main thread at least for 4ms  :(

xbmc/Application.cpp
```diff
@@ -2622,6 +2628,21 @@ void CApplication::HandleShutdownMessage()
   }
 }
 
+void CApplication::LockFrameMoveGuard()
+{
+  AtomicIncrement(&m_WaitingExternalCalls);
```
FernetMenta: please drop this ancient method. c++ supports atomics

xbmc/Application.h
```diff
@@ -509,7 +519,11 @@ class CApplication : public CXBApplicationEx, public IPlayerCallback, public IMs
   bool m_fallbackLanguageLoaded;
   
 private:
-  CCriticalSection                m_critSection;                 /*!< critical section for all changes to this class, except for changes to triggers */
+  CCriticalSection m_critSection;                 /*!< critical section for all changes to this class, except for changes to triggers */
+  
+  CCriticalSection m_frameMoveGuard;              /*!< critical section for synchronizing GUI actions from inside and outside (python) */
+  long m_WaitingExternalCalls;                    /*!< counts threads wich are waiting to be processed in FrameMove */
```
FernetMenta: std::atomic_int

FernetMenta:
jenkins build this please

FernetMenta:
jenkins was almost fine: http://jenkins.kodi.tv/view/Linux/job/BuildMulti-PR/4985/
I had to kill Windows build due to usual error: cannot stat ...

MartijnKaijser:
Looks like this PR makes Kodi hang on startup and doesn't go past the splash screen. At for the droid builds i've based on this PR

mapfau:
@MartijnKaijser could you pls provide a log? And best possible something like a backtrace?
Important are the two threads "languageInvoker" and the main thread.

MartijnKaijser:
Tried on two devices and both lock up on splash screen
http://xbmclogs.com/pc0jiiwcx

MartijnKaijser:
Doesn't seem to happen on windows from testing however it for sure does on android as I've made a build before and after.

mapfau:
Thanks, just looked into log but there is no chance to see anything.
I know that there could be places where mainthread and languageInvoker could dead lock, to really see the place I need some backtrace of the two threads.

Are you able to attach wich an debugger, then toggle to both threads and make an back-trace of them? With this information I could easily see whats the reason....

FernetMenta:
looks like Android only issue. in this case we certainly do not revert

MartijnKaijser:
What? If it was linux it would be reverted instantly.

FernetMenta:
we don't revert for a decaying platform without maintainer. this pr is a very important. i suggest to fix that Android crap instead doing harm to the maintained platforms

ronie:
this one is pretty spammy:
https://github.com/xbmc/xbmc/pull/9748/commits/82218b570c8d5e7d2ace70cbe71f24e101e72163#diff-63246197ac0f0ac89358c1155ad21d5bR2689

it's hitting my log ~40 times every minute

mapfau:
If there are no known issues with this "FrameMoveGuard", we can for sure remove the line.
@FernetMenta : anything popped up regarding this?

FernetMenta:
Yes, I think we should drop the this log

mapfau:
I'll push an PR tomorrow

PR #9749, opened 2016-05-04
Fix screensaver issue with JSON and pictures

Tolriq:
Opening a picture or a slideshow from JSON should stop any active screensaver like any other player action.

@Montellese for review or suggesting a better place to handle that.

### DIFF
--- a/xbmc/interfaces/json-rpc/PlayerOperations.cpp
+++ b/xbmc/interfaces/json-rpc/PlayerOperations.cpp
@@ -1151,6 +1151,9 @@ JSONRPC_STATUS CPlayerOperations::StartSlideshow(const std::string& path, bool r
   if (!firstPicturePath.empty())
     params.push_back(firstPicturePath);
 
+  // Reset screensaver when started from JSON only to avoid potential conflict with slideshow screensavers
+  g_application.ResetScreenSaver();
+  g_application.WakeUpScreenSaverAndDPMS();
   CGUIMessage msg(GUI_MSG_START_SLIDESHOW, 0, 0, flags);
   msg.SetStringParams(params);
   CApplicationMessenger::GetInstance().SendGUIMessage(msg, WINDOW_SLIDESHOW, true);


Tolriq:
Ping @Montellese and maybe @FernetMenta since it's player related and there's maybe a proper fix at player level.

Just to know if this can be envisaged, or remote should send fake keypress to leave screensaver.

Montellese:
For other players this is done when handling the `TMSG_MEDIA_PLAY` in `CPlayListPlayer`. Maybe this should also be done in `CGUIWindowSlideShow::OnMessage()` when handling the `GUI_MSG_START_SLIDESHOW` message?

But apart from that I'm fine with doing that as we do the same for video and music player as well.

Tolriq:
PR updated according to your request.

Montellese:
Fine with me. jenkins build this please

Tolriq:
@Montellese just had a light up when sleeping, I have checked the calls from the code and it seems the message is not sent in those case but I remembered slideshows can be used as screensavers.

So can someone confirm how that part works and that it's not broken anymore ?

MartijnKaijser:
ping @Montellese

Montellese:
I've never used a slideshow as a screensaver so I'm not sure how it works. But it probably has to run through that code so it might get broken.

Tolriq:
So should I get back to first implementation to avoid problems ? (Inside JSON) 

Would be cool to have a fix / workaround for v17.

Montellese:
I checked and couldn't find a script that uses the builtin slideshow functionality but maybe I missed it.
Moving it back into JSON is probably the safest course of action. Sorry for that.

MartijnKaijser:
@ronie maybe knows?

Tolriq:
@Montellese no problems, Yes I've checked too and can't find a trigger of that code from script but not an addon expert at all :(

@MartijnKaijser since this is a fix, I guess even if you go direct beta after alpha 2 this would still be possible to merge ? If so then I guess we can wait if someone knows.

If not I'll revert back to first solution to be sure to have fix in v17.

ronie:
> @ronie maybe knows?

no idea about other screensaver addons, but i'm not using the built-in slideshow functionality in the picture slideshow screensaver.

arnova:
@ronie: Why not using the builtin slideshow? Would be great if we could, since the current implementation still has lockup issues on slow hardware which are hard to fix (I've tried).

ronie:
it lacks the functionality the picture slideshow screensaver provides.

i just did a quick test with starting the builtin slideshow from a screensaver and the results are pretty odd. i had to hit escape like at least ten times before the slideshow would stop.

MartijnKaijser:
if it's a fix it can be merged yes

arnova:
@ronie: Odd, will have a look at that then.

ronie:
python, and it's probably an issue in the code i used....
will have another look at it tonight

ronie:
> will have another look at it tonight

changed my mind, lack of time and interest. sorry.

arnova:
@ronie: No worries, will put it on my own todo-list

Tolriq:
So I should fix the issue at pure JSON level in case you switch back to that later ?

Since I now stream my pictures to Kodi instead of Chromecast this fix would be highly appreciated :)

Montellese:
Yeah put it back into JSON and maybe add a comment explaining why it's there and not in the slideshow logic.

Tolriq:
PR updated.

Montellese:
Thanks. Good for me. jenkins build this please

PR #9754, opened 2016-05-04
RenderManager: Fix use of unitialized variables

garbear:
RenderManager relies on VideoPlayer to initialize m_videoDelay, but this caused loss of video in my RetroPlayer branch. Also, m_width and m_height are first used uninitialized in a `!=` comparison, so uninitialized values are unlikely to cause an issue.

 Still, better to be correct than lucky.


FernetMenta:
thanks

PR #9755, opened 2016-05-05
load libkodi_inputstream.dll on android from temp path

ghost:
This PR should enable loading inputstream addons on android devices
(Completely untested - I don't have android device running here)


FernetMenta:
CAddonStatusHandler is a design flaw. This class needs a redesign or should be trashed completely. It's kicked off by an instance with the id of a class. Then the handler creates another instance and acts upon that. This makes no sense.

MartijnKaijser:
jenkins build and merge

mazeyx:
this has fixed http://trac.kodi.tv/ticket/16712 and now kodi is launching and running fine it seems, i dont run any addons personally, but if you like me to test some i can. :) thank you.

PR #9756, opened 2016-05-05
[pvr] improvement for independent API levels

AlwinEsch:
This is a small part to bring in binary add-ons independent.

The goal on end is to handle parts from add-on headers only on PVRClient.h and for all other binary add-ons the same way.

With this are changes of levels only be needed on the add-on classes with `ADDON::CAddonDll`.

@ksooo @FernetMenta a way like this here OK.

Has not made it now to all to have a better overview.

### DIFF
--- a/xbmc/addons/PVRClient.cpp
+++ b/xbmc/addons/PVRClient.cpp
@@ -417,8 +417,8 @@ bool CPVRClient::GetAddonProperties(void)
   try
   {
     memset(&addonCapabilities, 0, sizeof(addonCapabilities));
-    PVR_ERROR retVal = m_pStruct->GetAddonCapabilities(&addonCapabilities);
-    if (retVal != PVR_ERROR_NO_ERROR)
+    PVRError retVal = TranslateError(m_pStruct->GetAddonCapabilities(&addonCapabilities));
+    if (retVal != PVRError_NO_ERROR)
     {
       CLog::Log(LOGERROR, "PVR - couldn't get the capabilities for add-on '%s'. Please contact the developer of this add-on: %s", GetFriendlyName().c_str(), Author().c_str());
       return false;
@@ -453,9 +453,9 @@ bool CPVRClient::GetAddonProperties(void)
       std::unique_ptr<PVR_TIMER_TYPE[]> types_array(new PVR_TIMER_TYPE[PVR_ADDON_TIMERTYPE_ARRAY_SIZE]);
       int size = PVR_ADDON_TIMERTYPE_ARRAY_SIZE;
 
-      PVR_ERROR retval = m_pStruct->GetTimerTypes(types_array.get(), &size);
+      PVRError retval = TranslateError(m_pStruct->GetTimerTypes(types_array.get(), &size));
 
-      if (retval == PVR_ERROR_NOT_IMPLEMENTED)
+      if (retval == PVRError_NOT_IMPLEMENTED)
       {
         // begin compat section
         CLog::Log(LOGWARNING, "%s - Addon %s does not support timer types. It will work, but not benefit from the timer features introduced with PVR Addon API 2.0.0", __FUNCTION__, strFriendlyName.c_str());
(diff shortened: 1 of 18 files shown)


ksooo:
I understand the intention, but mapping all kind of possible PVR_ERRORs to just a bool imo is not good as it reduces the possibilities for the application to react differently on the different errors. Wouldn't it be better to introduce errors "of interest" on the kodi side and to map the errors at a central place? I'm not sure whether from the architectural point of view it is a good idea to do gui stuff (OK dialog) at the level of PVRClient class.

xbmc/addons/PVRClient.cpp
```diff
@@ -2093,3 +2093,42 @@ void CPVRClient::OnPowerSavingDeactivated(void)
...
+    break;
+  case PVR_ERROR_FAILED:
+    error = PVRError_FAILED;
+    break;
+  case PVR_ERROR_UNKNOWN:
+  default:
+    error = PVRError_UNKNOWN;
+  }
+  return error;
+}
```
AlwinEsch: Has added a way as example to bring error codes from add-on headers to a one used on Kodi.

As long it is only one CPVRClient class present looks it crazy ;)

To use different levels comes on end this on `CPVRClientBase` and the the level class looks like:
```cpp
namespace V3
{
namespace KodiAPI
{
class CPVRClient
  : public CPVRClientBase
```

xbmc/addons/PVRClient.cpp
```diff
@@ -2071,3 +2093,42 @@ void CPVRClient::OnPowerSavingDeactivated(void)
   }
   catch (std::exception &e) { LogException(e, __FUNCTION__); }
 }
+
+PVRError CPVRClient::TranslateError(PVR_ERROR error)
```
FernetMenta: this is dead code. I don't see this method used anywhere

AlwinEsch: Was only added to see a way and to ask about. Not wanted to do all changes and then to bad for you :)

xbmc/addons/binary/interfaces/PVRClientBase.h
```diff
@@ -0,0 +1,51 @@
...
+  class CPVRClientBase
+  {
+  public:
+    CPVRClientBase();
+    virtual ~CPVRClientBase();
+
+    static const char *ToString(const PVRError error);
+  };
+
+}
```
AlwinEsch: The error code translation becomes now used with the `PVRError CPVRClient::TranslateError(PVR_ERROR addonError)`

Also added this class who becomes later for everything as standartized base PVRClient used as parent inside Kodi.

xbmc/addons/PVRClient.cpp
```diff
@@ -601,50 +601,50 @@ const std::string& CPVRClient::GetFriendlyName(void) const
...
 
-  PVR_ERROR retVal(PVR_ERROR_UNKNOWN);
+  PVRError retVal(PVRError_UNKNOWN);
   try
   {
     PVR_CHANNEL addonChannel;
     WriteClientChannelInfo(channel, addonChannel);
 
-    retVal = m_pStruct->OpenDialogChannelAdd(addonChannel);
+    retVal = TranslateError(m_pStruct->OpenDialogChannelAdd(addonChannel));
```
FernetMenta: from a performance pov this wrapping is not acceptable. in 99.9% of all cases the result is success and caling TranslateError is just overhead

AlwinEsch: Thats also one of the main reasons why I has used on begin only `bool` returns. Becomes in a `try`, `catch` not so easy to prevent it.

Has seen on most of all calls ends with `if (CPVRClient::... == PVR_ERROR_NO_ERROR)` and nothing else. Thats why I thought for me `bool`is the easiest.

ksooo: Is performance really critical here? I doubt this.

FernetMenta: don't forget that this serves as a template for other addons to. I see no reason for doubling the interface calls. from a logical pov it makes no sense either. why should we translate a return code "no error" by a method called TranslateError?

PR #9757, opened 2016-05-05
[Coverity]  Misc. Coverity issues

ksooo:
Just a bunch of random Coverity issue fixes.

@Jalle19 @Paxxi ... mind taking a look.

### DIFF
--- a/lib/addons/library.kodi.adsp/libKODI_adsp.cpp
+++ b/lib/addons/library.kodi.adsp/libKODI_adsp.cpp
@@ -104,15 +104,15 @@ DLLEXPORT void ADSP_release_sound_play(CAddonSoundPlay* p)
 }
 
 CAddonSoundPlay::CAddonSoundPlay(void *hdl, void *cb, const char *filename)
- : m_Filename(filename)
+ : m_Filename(filename),
+   m_Handle(hdl),
+   m_cb(cb)
 {
   m_PlayHandle = NULL;
   if (!hdl || !cb)
     fprintf(stderr, "libKODI_adsp-ERROR: ADSP_get_sound_play is called with NULL handle !!!\n");
   else
   {
-    m_Handle     = hdl;
-    m_cb         = cb;
     m_PlayHandle = ((CB_ADSPLib*)m_cb)->SoundPlay_GetHandle(((AddonCB*)m_Handle)->addonData, m_Filename.c_str());
     if (!m_PlayHandle)
       fprintf(stderr, "libKODI_adsp-ERROR: ADSP_get_sound_play can't get callback table from KODI !!!\n");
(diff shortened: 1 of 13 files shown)


ksooo:
jenkins build this please

xbmc/utils/ActorProtocol.h
```diff
@@ -51,7 +51,9 @@ class Message
   bool Reply(int sig, void *data = NULL, int size = 0);
 
 private:
-  Message() {isSync = false; data = NULL; event = NULL; replyMessage = NULL;};
+  Message()
+  : signal(0), isSync(false), isSyncFini(false), isOut(false), isSyncTimeout(false),
```
FernetMenta: I am not a fan of initializing something for the sake of initializing. before this change the state after construction was obvious, now it isn't anymore. Why should a message defined as not being isOut after construction?

ksooo: My intention was ofc not to initialize instances with random values. ;-) For sure the right initial values should be chosen. Well change isOut imo value. Thanks for bringing this up.

FernetMenta: this is more like a tri-state, undefined until the message is sent. would be helpful if we can indicate this somehow.

Paxxi: if you want three states then switch to an int or better yet an enum. As long as this is a bool initializing it is valid. It's not like we can see that it's not initialized anyway since it'll just randomly be true or false.

Paxxi:
looks good to me, didnt verify that the order of initialization matches order of definition but I assume you've taken care of that?

xbmc/utils/ActorProtocol.h
FernetMenta: you missed the point. members are dependent on state and the value of. i.e. isOut is meaningless when the object gets constructed. This is important information if you want to understand the behaviour. initializing all those members is meaningless to the behaviour and just confuses the reader if not stated explicitly. Some may think it is important that those values are initialized with those values that were taken arbitrarily.

Paxxi: I see your point but I don't agree that leaving them uninitialized is the solution. Initializing objects with safe defaults are the norm, if that makes the design harder to reason about it can be solved with comments as to what the fields does and what the values mean. 
A second option would be to not have parameter less constructors if there's no valid default values.

ksooo:
> didnt verify that the order of initialization matches order of definition but I assume you've taken care of that?

yes, I did.

xbmc/utils/ActorProtocol.h
ksooo: I will make it an enum.

xbmc/utils/ActorProtocol.h
```diff
@@ -33,12 +33,19 @@ class Protocol;
 
 class Message
 {
+  typedef enum
+  {
+    TYPE_UNDEFINED,
```
FernetMenta: no way, please drop

FernetMenta:
@ksooo please drop ActorProtocol from this pr. I don't want it drift in this direction.

ksooo:
> please drop ActorProtocol from this pr.

 Will do. I'm not gonna fight you or discuss this back and forth. I need my nerves for other, more important things. ;-) (Please note the smiley, no offense here)

ksooo:
jenkins build this please

PR #9758, opened 2016-05-05
addons: enable/disable callback interface for pvr

FernetMenta:
fix crash on exit when PVR addons still want to use the callback interface

@AlwinEsch other interface suffer from the same issue.

### DIFF
--- a/xbmc/ServiceManager.cpp
+++ b/xbmc/ServiceManager.cpp
@@ -23,9 +23,12 @@
 #include "utils/log.h"
 #include "interfaces/AnnouncementManager.h"
 #include "interfaces/generic/ScriptInvocationManager.h"
+#include "addons/binary/interfaces/api1/PVR/AddonCallbacksPVR.h"
 
 bool CServiceManager::Init1()
 {
+  V1::KodiAPI::PVR::CAddonCallbacksPVR::EnableInterface(false);
+
   m_announcementManager.reset(new ANNOUNCEMENT::CAnnouncementManager());
   m_announcementManager->Start();
 
@@ -46,6 +49,7 @@ bool CServiceManager::Init2()
 
   m_ADSPManager.reset(new ActiveAE::CActiveAEDSP());
   m_PVRManager.reset(new PVR::CPVRManager());
+  V1::KodiAPI::PVR::CAddonCallbacksPVR::EnableInterface(true);
 
   m_binaryAddonCache.reset( new ADDON::CBinaryAddonCache());
   m_binaryAddonCache->Init();
@@ -63,6 +67,8 @@ bool CServiceManager::Init3()
 
 void CServiceManager::Deinit()
 {
+  V1::KodiAPI::PVR::CAddonCallbacksPVR::EnableInterface(false);
+
   m_binaryAddonCache.reset();
   m_PVRManager.reset();
   m_ADSPManager.reset();
--- a/xbmc/addons/binary/interfaces/api1/PVR/AddonCallbacksPVR.cpp
+++ b/xbmc/addons/binary/interfaces/api1/PVR/AddonCallbacksPVR.cpp
@@ -50,6 +50,8 @@ namespace KodiAPI
 namespace PVR
 {
 
+std::atomic_bool CAddonCallbacksPVR::m_enabled;
+
 CAddonCallbacksPVR::CAddonCallbacksPVR(CAddon* addon)
   : ADDON::IAddonInterface(addon, 1, XBMC_PVR_API_VERSION),
     m_callbacks(new CB_PVRLib)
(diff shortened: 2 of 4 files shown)


xbmc/addons/binary/interfaces/api1/PVR/AddonCallbacksPVR.cpp
```diff
@@ -365,6 +418,9 @@ typedef struct EpgEventStateChange
 
 void CAddonCallbacksPVR::UpdateEpgEvent(const EpgEventStateChange &ch, bool bQueued)
 {
+  if (!m_enabled)
+    return;
+
```
ksooo: <code>CAddonCallbacksPVR::UpdateEpgEvent</code> is not a callback function. It is a private member function, only called by <code>CAddonCallbacksPVR::PVREpgEventStateChange</code>. => https://github.com/xbmc/xbmc/blob/master/xbmc/addons/binary/interfaces/api1/PVR/AddonCallbacksPVR.h#L237 So, no need for the enabled check here.

FernetMenta: The interface is not protected by a lock and can be disabled while in PVREpgEventStateChange. So better to check here too.

xbmc/addons/binary/interfaces/api1/PVR/AddonCallbacksPVR.h
```diff
@@ -237,6 +240,7 @@ class CAddonCallbacksPVR : public ADDON::IAddonInterface
   static void UpdateEpgEvent(const EpgEventStateChange &ch, bool bQueued);
 
   CB_PVRLib    *m_callbacks; /*!< callback addresses */
+  static std::atomic_bool m_enabled;
```
AlwinEsch: Can you add it to `IAddonInterface` as `static std::atomic_bool m_pvrEnabled;` to hold in future independent from levels, as `IAddonInterface::EnablePVRInterface(true);`?

FernetMenta: adding a specific variable like pvrEnabled to a generic base class violates the concept of hierarchy.

xbmc/ServiceManager.cpp
```diff
@@ -23,9 +23,12 @@
 #include "utils/log.h"
 #include "interfaces/AnnouncementManager.h"
 #include "interfaces/generic/ScriptInvocationManager.h"
+#include "addons/binary/interfaces/api1/PVR/AddonCallbacksPVR.h"
 
 bool CServiceManager::Init1()
 {
+  V1::KodiAPI::PVR::CAddonCallbacksPVR::EnableInterface(false);
```
AlwinEsch: Hi @FernetMenta I understand you about the hierarchy, but make for me no sense to have something like the follow in future:
```cpp
V1::KodiAPI::PVR::CAddonCallbacksPVR::EnableInterface(false);
V2::KodiAPI::PVR::CAddonCallbacksPVR::EnableInterface(false);
V3::KodiAPI::PVR::CAddonCallbacksPVR::EnableInterface(false);
V4::KodiAPI::PVR::CAddonCallbacksPVR::EnableInterface(false);
V5::KodiAPI::PVR::CAddonCallbacksPVR::EnableInterface(false);
```
better to have only one call on a part used by all.

FernetMenta: Not sure what you are trying to say. It does not change the fact, that having a specialized variable in a base class is wrong.

AlwinEsch: I mean that then for every new API level it need to call the same independent for his class. This bring a list of same calls and if for some reason it becomes more and the `EnableInterface`becomes for some reason also called on other place a easy source for errors.

AlwinEsch: To prevent things like this was also my idea about `IAddonInterface`, but with the functions it have better a name as `CAddonInterfaceBase`, with the `I` on begin then really not right.

FernetMenta: both variants are not ideal. this tells us that the design of the interface has not stabilized yet.

AlwinEsch: Is a double use like
```cpp
V1::KodiAPI::PVR::CAddonCallbacksPVR::EnableInterface(false);
V2::KodiAPI::PVR::CAddonCallbacksPVR::EnableInterface(false);
```
after level 2 comes in OK for you?

I think this faults becomes fixed with level 3 where everything on add-on's becomes performed with classes instead of `C` functions.

For all who not know, see: https://github.com/AlwinEsch/pvr.demo/commit/a213319e9c267ced38cccacecce25e9c51235bde

FernetMenta:
@AlwinEsch looking forward to a better solution with v3. In the meantime I close this.

PR #9759, opened 2016-05-06
Revert "load libkodi_inputstream.dll on android from temp path"

FernetMenta:
Reverts xbmc/xbmc#9755

not that fast. I didn't ack

### DIFF
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/libKODI_inputstream.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/libKODI_inputstream.h
@@ -40,10 +40,6 @@
 #define INPUTSTREAM_HELPER_DLL "/library.kodi.inputstream/" INPUTSTREAM_HELPER_DLL_NAME
 #endif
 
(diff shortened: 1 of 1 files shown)


xbmc/addons/kodi-addon-dev-kit/include/kodi/libKODI_inputstream.h
```diff
@@ -78,15 +74,6 @@ class CHelper_libKODI_inputstream
     libBasePath  = ((cb_array*)m_Handle)->libPath;
     libBasePath += INPUTSTREAM_HELPER_DLL;
 
-#if defined(ANDROID)
```
FernetMenta: apart from the fact that we use TARGET_ANDROID, the ifdef at this location is wrong

MartijnKaijser:
Then you point out exactly where it should be or create a PR for it. Currently android doesn't start and i call that huge deal

FernetMenta:
as a first step I recommend to disable inputstream for Android. it won't work properly on that platform anyway @mapfau please update the xml file for the addon

PR #9760, opened 2016-05-06
[pvr] destroy all created clients on unload

FernetMenta:
this also fixes the crash reported in https://github.com/xbmc/xbmc/pull/9758


FernetMenta:
travis 3 out of 4 is prof enough

PR #9761, opened 2016-05-06
addons: fix death spiral of recursion

FernetMenta:
fix crash on startup if load of addon fails. having status handler operate on the class instead of the object was wrong.

credits to @mapfau for tracking this down

@MartijnKaijser this may also cure the crash you observed on Android

### DIFF
--- a/xbmc/addons/AddonDll.h
+++ b/xbmc/addons/AddonDll.h
@@ -243,7 +243,6 @@ ADDON_STATUS CAddonDll<TheDll, TheStruct, TheProps>::Create()
     else
     { // Addon failed initialization
       CLog::Log(LOGERROR, "ADDON: Dll %s - Client returned bad status (%i) from Create and is not usable", Name().c_str(), status);
-      new CAddonStatusHandler(shared_from_this(), status, "", false);
     }
   }
   catch (std::exception &e)
--- a/xbmc/addons/AddonDll.h
+++ b/xbmc/addons/AddonDll.h
@@ -191,7 +191,7 @@ bool CAddonDll<TheDll, TheStruct, TheProps>::LoadDll()
   {
     delete m_pDll;
     m_pDll = NULL;
-    new CAddonStatusHandler(ID(), ADDON_STATUS_UNKNOWN, "Can't load Dll", false);
+    new CAddonStatusHandler(shared_from_this(), ADDON_STATUS_UNKNOWN, "Can't load Dll", false);
     return false;
   }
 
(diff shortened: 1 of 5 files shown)


MartijnKaijser:
will test

MartijnKaijser:
Still closes Kodi on start.
#9759 reverted and this one on top and including both inputstream addons.

http://pastebin.com/xpxcp7EL

FernetMenta:
ERROR: AddOnLog: ARGUS TV client: Ping failed... No connection to Argus TV.

please disable all PVR addons and try again. you can lease VNSI enable, for this I take responsibility :)

MartijnKaijser:
There's no way to disable those as they are all default enabled on a fresh install. See my comment here https://github.com/xbmc/xbmc/pull/9706#issuecomment-217126520

I have already build it previously without **any** PVR addons and the result was exactly the same.

FernetMenta:
looking into the log and taking into account the time it started to fail, my guess is this:
https://github.com/kodi-pvr/pvr.argustv/pull/51/files

that's why I always suggest to let unmaintained components die. argustv is one of those.

FernetMenta:
> There's no way to disable those as they are all default enabled on a fresh install

who did change this. PVR addons and decoders MUST be disabled by default

MartijnKaijser:
They are disabled by default if i do a fresh windows install but they are not on a fresh android install. No idea at what point that became borked. Installing a master build nothing crashes and although the PVR addons do complain Kodi keeps running

FernetMenta:
@MartijnKaijser this should finally fix it

MartijnKaijser:
rebuilding now including all binary addons

MartijnKaijser:
new logcat
http://pastebin.com/amzzzE5c

FernetMenta:
jenkins build this please

MartijnKaijser:
Needs rebase

FernetMenta:
jenkins build this please

tamland:
Please don't use shared_from_this. Needing that is a major code smell making it even harder to get rid of the shared_ptr coupling in this code

FernetMenta:
without shared_from_this we have to trash AddonStatusHandler. for this kind od async operation a shared pointer is required.

tamland:
Well, shared_from_this requires an external shared_ptr to exist which tightly couple the implementations together with users of it. Any other use will result in a runtime exception. Which is why i think the whole `enable_shared_from_this` is clusterfuck and should be avoided. Using it in interfaces such as `IAddon` and `AddonDll` ties the entire interface to shared_ptr because one implementation requires it.

FernetMenta:
I agree, it's ugly. I'll have a look into AddonStatusHandler, maybe it can solved differently.

PR #9762, opened 2016-05-06
[win32][depends] Updated taglib to v1.11

Paxxi:


Paxxi:
@Razzeee @DaveTBlake here you go, files are uploaded and should be available in 30 minutes or so

MartijnKaijser:
Also uploaded for non-win32 platforms
http://mirrors.kodi.tv/build-deps/sources/taglib-1.11.tar.gz

MrMC:
does this fix the issue see with currnet taglib ?

MartijnKaijser:
That's what we'll find out. First needed a compiled win32 version

PR #9763, opened 2016-05-06
[EPG] Fix crash on opening epg grid window directly after channel(s) got deleted.

ksooo:
Difficult to reproduce, but fix is obvious, I guess.

@Jalle19 mind taking a look.

### DIFF
--- a/xbmc/epg/GUIEPGGridContainerModel.cpp
+++ b/xbmc/epg/GUIEPGGridContainerModel.cpp
@@ -82,6 +82,9 @@ void CGUIEPGGridContainerModel::Refresh(const std::unique_ptr<CFileItemList> &it
     m_programmeItems.emplace_back(fileItem);
 
     channel = fileItem->GetEPGInfoTag()->ChannelTag();
+    if (!channel)
+      continue;
+
     int iCurrentChannelID = channel->ChannelID();
     if (iCurrentChannelID != iLastChannelID)
     {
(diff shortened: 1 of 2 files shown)


xbmc/epg/GUIEPGGridContainerModel.cpp
```diff
@@ -258,8 +261,9 @@ void CGUIEPGGridContainerModel::FindChannelAndBlockIndex(int channelUid, unsigne
     CDateTime gridCursor(m_gridStart); //reset cursor for new channel
     unsigned long progIdx = m_epgItemsPtr[channel].start;
     unsigned long lastIdx = m_epgItemsPtr[channel].stop;
-    int iEpgId            = m_programmeItems[progIdx]->GetEPGInfoTag()->EpgID();
+    int iEpgId = m_programmeItems[progIdx]->GetEPGInfoTag()->EpgID();
     CEpgInfoTagPtr tag;
+    CPVRChannelPtr chan;
```
Jalle19: Why not call it `channel`?

Jalle19: Could be `const` too I suppose?

ksooo: channel is already taken for another variable. no const, because used inside for loop and reassigned while iterating.

ksooo:
jenkins build this please

xbmc/epg/GUIEPGGridContainerModel.cpp
Jalle19: In that case no complaints from here :-)

PR #9765, opened 2016-05-07
configure: add support for aarch64

bkuhls:
Downloaded from:
https://github.com/OpenELEC/OpenELEC.tv/blob/master/packages/mediacenter/kodi/patches/kodi-999.10-aarch64-support.patch

and added uClibc support.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -622,6 +622,16 @@ case $host in
      use_gl=no
      USE_STATIC_FFMPEG=1
      ;;
+  aarch64*-*-linux-gnu*|aarch64*-*-linux-uclibc*)
+     target_platform=target_linux
+     ARCH="aarch64"
+     use_arch="aarch64"
+     use_neon=yes
+     use_gles=yes
+     use_gl=no
+     use_wayland=no
+     USE_STATIC_FFMPEG=1
+     ;;
   arm*-*linux-android*)
      target_platform=target_android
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -622,6 +622,18 @@ case $host in
      use_gl=no
      USE_STATIC_FFMPEG=1
      ;;
+  aarch64*-*-linux-gnu*|aarch64*-*-linux-uclibc*)
+     target_platform=target_linux
+     ARCH="aarch64"
+     use_arch="aarch64"
+     use_cpu=cortex-a53
+     use_joystick=no
```
stefansaraev: use_joystick is no more. you can remove it

configure.ac
```diff
@@ -622,6 +622,18 @@ case $host in
      use_gl=no
      USE_STATIC_FFMPEG=1
      ;;
+  aarch64*-*-linux-gnu*|aarch64*-*-linux-uclibc*)
+     target_platform=target_linux
+     ARCH="aarch64"
+     use_arch="aarch64"
+     use_cpu=cortex-a53
```
koenkooi: The use_cpu should be dropped, kodi builds & runs fine without it on a dragonboard-410c. Which incidentally works best with GL, not with GLES.

bkuhls:
@stefansaraev, @koenkooi: I updated the PR according to your comments.

PR #9766, opened 2016-05-07
addon fix status handler

FernetMenta:
alternative to https://github.com/xbmc/xbmc/pull/9761

@MartijnKaijser could you test this please?

### DIFF
--- a/xbmc/addons/AddonDll.h
+++ b/xbmc/addons/AddonDll.h
@@ -26,11 +26,14 @@
 #include "AddonManager.h"
 #include "AddonStatusHandler.h"
 #include "addons/binary/interfaces/AddonInterfaces.h"
+#include "guilib/GUIWindowManager.h"
+#include "dialogs/GUIDialogOK.h"
 #include "utils/URIUtils.h"
 #include "filesystem/File.h"
 #include "filesystem/SpecialProtocol.h"
 #include "filesystem/Directory.h"
 #include "utils/log.h"
+#include "utils/StringUtils.h"
 #include "utils/XMLUtils.h"
 #include "utils/Variant.h"
 #include "Util.h"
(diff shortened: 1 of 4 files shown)


MartijnKaijser:
@FernetMenta it doesn't "crash"  anymore though i now get the OK dialog for each inputstream addon 3 times.
Entering add-on browser Kodi just locks up.

MartijnKaijser:
hereś the logcat http://xbmclogs.com/p3m034qir

FernetMenta:
@MartijnKaijser updated. could you try again?

MartijnKaijser:
No crashes or lockups with this build. Also i do not get any OK dialog (if that was supposed to happen) about non working inputstream addons. Doesn't matter if they are disabled or enabled.
[logcat.txt](https://github.com/xbmc/xbmc/files/253503/logcat.txt)

FernetMenta:
this works as expected. AddonManager is started before WindowManager and as a result the first error window is not displayed.

jenkins build this please

FernetMenta:
jenkins build this please

MartijnKaijser:
Great. Thanks.

MartijnKaijser:
jenkins build this please

da-anda:
was just wondering if a kaitoast + eventlog wouldn't be a better approach to the OK dialogs. Opinions?

FernetMenta:
IMO a kai toast is completely useless. we already have the same crap on installation of addons. if you miss to read it during that short while displayed, you won't know that installation has failed.

da-anda:
for that we have the event log now, but I agree that this is a bit too hidden atm and we would need some more prominent notification center.

FernetMenta:
event log is not appropriate here. you can't expect a user to browse the event log after every task. while working with addons it happened quite often to me that I was wondering that an addon was not available after install because the crappy kaitoast only displays the error for a very short time.
there is zero reason for an asynchronous user feedback. you won't watch a movie while waiting for the addon install to be finished, right?

ksooo:
I see this exactly like @FernetMenta does.

Montellese:
You can still put it into the event log so that when the user goes there he can see that something went wrong after it happened.
From what I understand right now there's no user feedback at all because the ok dialog gets eaten by the fact the the window manager isn't initialized yet when the error happens.

FernetMenta:
@Montellese that's correct for startup. For addon installation we should drop kaitoast in favour of DialogOK.

MartijnKaijser:
I would agree that when installing an add-on manually an ok dialog would fit better regarding feedback.

PR #9769, opened 2016-05-07
[droid] clean binary addons to prevent them from repackaging even if …

MartijnKaijser:
…they should not be build

While testing the inpustream addons i found out that when you disable certain add-ons in binary repo they are still packaged in the final APK although they are not build. So they are likely floating around and then moved to final packaging again.

thx to @Memphiz for the pointer. I gave it a quick test and it indeed seems to solve the problem.

@wsnipex @Montellese for comments if this is correct. Not sure if the comment line also has to be removed.


Memphiz:
Mhhh i didn't meant to have this as a generic solution - i meant it to solve your particular problem - hence i suggested that you add this change to your branch while poking at the addon repo ...

PR #9772, opened 2016-05-08
[aml] add gui setting to enable/disable mpeg2 hw decoding

stefansaraev:
amlogic's mpeg2 decoder does not always work good for all videos, let's expose a gui option to enable/disable it to the user

on faster platforms (m8/gxbb) users may want this option disabled, while on m6 decoding in software may be slow. distros are free to ship their defaults via appliance.xml

fyi @codesnake @koying

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -6741,7 +6741,18 @@ msgctxt "#13452"
 msgid "Enable this option to use hardware acceleration for VC-1 based codecs. If disabled the CPU will be used instead. Especially VC-1 interlaced fails hard on Intel hardware."
 msgstr ""
 
-#empty strings from id 13453 to 13456
+#: system/settings/settings.xml
+msgctxt "#13453"
+msgid "Use amcodec for SD content"
+msgstr ""
+
+#. Description of setting with label #13453 "Use amcodec for SD content"
+#: system/settings/settings.xml
+msgctxt "#13454"
+msgid "Enable this option to use hardware acceleration for MPEG-(1 / 2) SD content. If disabled the CPU will be used instead."
+msgstr ""
+
+#empty strings from id 13455 to 13456
 
 #. Option for video related setting #13454: sw filter
 #: system/settings/settings.xml
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -6743,16 +6743,25 @@ msgstr ""
 
 #: system/settings/settings.xml
 msgctxt "#13453"
-msgid "Use amcodec for SD content"
+msgid "Use amcodec for MPEG-2 SD content"
 msgstr ""
 
-#. Description of setting with label #13453 "Use amcodec for SD content"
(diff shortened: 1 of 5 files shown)


koying:
I generalized the concept in SPMC
https://github.com/koying/SPMC/commit/0b3e9ecd0cf850b37a2923242eccd312942a21d7

stefansaraev:
thanks @koying but I am not sure I want to make t like in spmc, as it's totally off with what we have for vdpau/va-api. @codesnake @fritsch what's your opinion ?

koying:
Not sure what you mean. You cannot pick it as it is, of course, as it's for 16
My point is rather: Why limit it to mpeg2 (and amlcodec, but that's for another day ;) and just on/off. 

Know fact that xvid/divx/mpeg4 and their infinite variations can also be problematic for some h/w decoder + s/w decoding SD and s/w decoding HD is for different class of devices.

stefansaraev:
uh, right. I have a 1080i59 mpeg-2 sample that works just fine hw decoded. and plays like cr** in sw. so it makes sense to use amcodec for mpeg-2 hd (width > 800). maybe I add another option for this.

fritsch:
My opinion: if possible, please fix the decoder. If it's really not
possible fall back to SW for width <= 800.

2016-05-08 11:56 GMT+02:00 Stefan Saraev <notifications@github.com>:

> uh, right. I have a 1080i59 mpeg-2 sample that works just fine hw decoded.
> and plays like cr** in sw. so it makes sense to use amcodec for mpeg-2 hd
> (width > 800). maybe I add another option for this.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/xbmc/xbmc/pull/9772#issuecomment-217706933>
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

codesnake:
@stefansaraev Maybe is better to use something like "Use amcodec for SD content", because SW decoding is already used for MPEG-4 SD as well.

stefansaraev:
updated label and help hint.

stefansaraev:
does anyone have to say something in favor / against this ?

jenkins build this please

codesnake:
Would be good this setting to affect MPEG-4 SD as well.
Currently it always use SW decoding for now: https://github.com/stefansaraev/xbmc/blob/65a73c71733e66339188a678c8eb2880c15e5024/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAmlogic.cpp#L129-L131

stefansaraev:
@codesnake this will be change in behaviour for existing users. are you sure we want that? or do we want a separate option for mpeg-4 ?

stefansaraev:
commit added for mpeg-4, defaults to false (no change in behaviour), I have yet to runtime test that.

koying:
Let me re-iterate my opinion that just an on/off switch, not taking account resolution (and thus resources needed to s/w decode), is not good enough. A multi-value switch and removing the hardcoded SD test is the proper way. 

Having a on/off switch + a hardcoded test is very ugly ;)
Just my opinion, of course.

stefansaraev:
I'll look at your commit, @koying

PR #9773, opened 2016-05-08
fix Compile Kodi for Windows

primatic:
buildhelpers.sh expected archive to be always .tar.gz, but for some projects github does not provide .tar.gz, but does provide .zip. This remains undetected until somebody does a fresh checkout following the instructions on http://kodi.wiki/view/HOW-TO:Compile_Kodi_for_Windows

### DIFF
--- a/tools/buildsteps/win32/buildhelpers.sh
+++ b/tools/buildsteps/win32/buildhelpers.sh
@@ -32,10 +32,11 @@ fi
 do_wget() {
   local URL="$1"
   local archive="$2"
-
   if [[ -z $archive ]]; then
+    do_print_status  "wget --tries=5 --retry-connrefused --waitretry=2 --no-check-certificate -c $URL" "$green_color" "wget"
     wget --tries=5 --retry-connrefused --waitretry=2 --no-check-certificate -c $URL
   else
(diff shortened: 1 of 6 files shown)


afedchin:
> for some projects github does not provide .tar.gz, but does provide .zip.

provide an example please

In my oppinion your changes has bad design:
1. now to change version of ffmpeg need to change the following `VERSION`, `ARCHIVE` and `REMOTE` intead of changing only `VERSION`
2. `REMOTE` is actually `BASE_URL` + `ARCHIVE`
3. I don't understand why `LOCAL` is needed.

MartijnKaijser:
agree with @afedchin as i'm not sure why you want this. Github also provides tar.gz file and if you want to define local files there a different method for that

wsnipex:
-1 
there is no reason for this change as github provides tarballs for every repo.

primatic:
I agree that this fix is not needed if there is a tarball for every project, and I don't understand why that was not the case for libdvdcss, libdvdnav and  libdvdread. For those projects github returned a 404 when running the original script. Today, I reverted these changes locally and rerun the script, and to my surprise, it now works.

PR #9778, opened 2016-05-08
fix PluginDirectory

FernetMenta:
@MilhouseVH this should fix http://trac.kodi.tv/ticket/16718

### DIFF
--- a/xbmc/windows/GUIMediaWindow.cpp
+++ b/xbmc/windows/GUIMediaWindow.cpp
@@ -85,6 +85,8 @@
 #define PROPERTY_SORT_ORDER         "sort.order"
 #define PROPERTY_SORT_ASCENDING     "sort.ascending"
 
+#define PLUGIN_REFRESH_DELAY 200
+
 using namespace ADDON;
 using namespace KODI::MESSAGING;
 
@@ -506,6 +508,11 @@ bool CGUIMediaWindow::OnMessage(CGUIMessage& message)
       if (message.GetParam1() != WINDOW_INVALID)
       { // first time to this window - make sure we set the root path
         m_startDirectory = returning ? dir : GetRootPath();
+      }
+      if (message.GetParam2() == PLUGIN_REFRESH_DELAY)
+      {
+        Refresh();
+        return true;
       }
     }
     break;
@@ -1486,7 +1493,21 @@ void CGUIMediaWindow::OnInitWindow()
 
   // the start directory may change during Refresh
   bool updateStartDirectory = URIUtils::PathEquals(m_vecItems->GetPath(), m_startDirectory, true);
-  Refresh();
+
+  // we have python scripts hooked in everywhere :(
+  // those scripts may open windows and we can't open a window
+  // while opening this one.
+  // for plugin sources delay call to Refresh
+  if (!URIUtils::IsPlugin(m_vecItems->GetPath()))
+  {
+    Refresh();
+  }
+  else
+  {
+    CGUIMessage msg(GUI_MSG_WINDOW_INIT, 0, 0, 0, PLUGIN_REFRESH_DELAY);
+    g_windowManager.SendThreadMessage(msg, m_controlID);
+  }
+
   if (updateStartDirectory)
   {
     // reset the start directory to the path of the items
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -2681,6 +2681,7 @@ void CApplication::FrameMove(bool processEvents, bool processGUI)
     // Window size can be between 2 and 10ms and depends on number of continuous requests
     if (m_WaitingExternalCalls)
     {
+      CSingleExit ex(g_graphicsContext);
       m_frameMoveGuard.unlock();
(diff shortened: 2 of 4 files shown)


FernetMenta:
@MartijnKaijser this pr has a slight rework for https://github.com/xbmc/xbmc/pull/9748
maybe you are lucky and it fixes the Android issue.

MilhouseVH:
http://trac.kodi.tv/ticket/16718 appears to be fixed by this PR - many thanks!

FernetMenta:
jenkins build this please

xbmc/windows/GUIMediaWindow.cpp
```diff
@@ -1486,7 +1492,21 @@ void CGUIMediaWindow::OnInitWindow()
 
   // the start directory may change during Refresh
   bool updateStartDirectory = URIUtils::PathEquals(m_vecItems->GetPath(), m_startDirectory, true);
-  Refresh();
+
+  // we have python scripts hooked in everywhere :(
+  // those scripts may open windows and we can't open a window
+  // while opening this one.
+  // for pligin sources delay call to Refresh
```
ksooo: typo here. pligin -> plugin

MartijnKaijser:
No change

MilhouseVH:
Although the Search YouTube hang appeared to be resolved, it's possible this PR is causing other odd behaviour when starting addons (or it's another hangover from the original problem, #9664). 

With PR9778, sometimes the user interface for the addon will "flash" (appear briefly then disappear), other times the addon will simply show a blank background (but the GUI hasn't hung - "Back" will eventually return to the Estuary Home menu).

I can reproduce this with latest master + PR9778 and YouTube addon in Ubuntu when running Estuary Home > Addons > YouTube.

Another addon showing similar problems is the Amazon Prime Instant addon.

In fact when you run the Amazon Prime Instant addon you see only a blank background. Click "Back" and you see a constantly spinning "Busy" icon. Click "Back" again and you're returned to the Estuary Home menu. Now click on Search > Search YouTube and you'll be shown the Amazon Prime Instant addon screen overlaid with a "Busy" dialog...

xbmc/windows/GUIMediaWindow.cpp
FernetMenta: thanks, corrected

FernetMenta:
@MilhouseVH updated. could you test again please

MilhouseVH:
Sorry for the delay, updates look good @fernetmenta have tested all the add-ons on my RPi3 system and all are working as expected. Thanks again!

FernetMenta:
thanks for feedback. there is one minor fix I am going to add to this PR before merge. currently you have to cancel twice when the keyboard dialog is displayed.

MilhouseVH:
> currently you have to cancel twice when the keyboard dialog is displayed.

For example the YouTube addon? Yes I see that, first "Back" (or Cancel) shows a busy dialog. Strangely I don't get this issue with Global Search, so I assumed it was specific to the YouTube addon (ie. searching for `""`).

FernetMenta:
> For example the YouTube addon? Yes I see that, first "Back" (or Cancel) shows a busy dialog

Exactly this. The old code did observe script termination. I will add this back but slightly different.

FernetMenta:
jenkins build this please

HitcherUK:
Same problem here with script filled dynamic lists triggering the busy dialog.

@ronie @phil can you confirm the same?

FernetMenta:
scripts are not supposed to hijack the main thread for a long time. if you see the busy dialog, you are doing something wrong

Ignoble61:
I don't believe this is a case of plugins 'hijacking' the main thread, but rather when a list is filled directly from a plugin (`<content>plugin://plugin.name/</content>`) this now triggers the busy dialog where it didn't previously. It causes major usability issues with skins which directly fill lists from plugins as widgets, and definite usability issues with all skins which use @phil65's autocompletion plugin with the virtual keyboard, including Estuary.

If plugins are now returning their contents on the main thread, then this would appear to be a change of behaviour leading to a regression in terms or usability, and not the result of any change in plugin behaviour - and would affect all plugins I have tested.

ronie:
> @ronie @phil can you confirm the same?

yup, can confirm.

perhaps we can grant plugins a little more time to provide a listing before popping up the busy dialog?
if i read the changes correctly, it used to be 1500 msecs before a progressbar was shown.

phil65:
I think for dynamic content included via `<content>plugin://plugin.name/</content>` there never was a progress/busydialog. Those dialogs should only appear if the list which is loading content is our main media list (Video/Audio plug-ins for example)

FernetMenta:
1)
the old design was wrong and caused issues. that was the reason why I fixed it. 

2)
GetDirectory must not silently hijack the main thread. 

3)
if the current behaviour does not fit your requirements, write those down in the dev section of the forum

phil65:
just fyi, i had a talk with @FernetMenta today and explained the "problem" in more detail. He is aware of all the details now and will come up with a sane fix for this issue soon I think. :)

PR #9780, opened 2016-05-08
Fix some compiler warnings

ksooo:
@Paxxi mind taking a look?


ksooo:
@Paxxi good to go?

Paxxi:
Personally I would switch to auto instead of size_t. But it's fine as is

ksooo:
jenkins build this please

PR #9781, opened 2016-05-08
[iOS] Add "Copy to Kodi" functionality

DanTheMan827:
This adds the option to the iOS share sheet of apps to "Open In" Kodi

I also added special://unixhome which resolves to the home path of the user ($HOME) and changed the path in xbmc/storage/osx/DarwinStorageProvider.cpp since on a non-jailbroken device $HOME changes with every upgrade.


MrMC:
unixhome ?

Memphiz:
maybe call it userhome instead?

DanTheMan827:
there appears to already be special://userhome...

According to SpecialProtocol.h is should resolve to ~/.kodi but I don't see SetUserHomePath called anywhere...

It looks like it may have been partially removed at one point since the log notice for it is commented out

Memphiz:
well as you also use it on windows i would like to change the wording - either userhome or envhome or so - adn then all getenv("HOME") in the other code should use the special path instead... (darwinstorageprovider on osx has still the getenv in there.

MrMC:
envhome makes more sense if it maps to getenv("HOME") . trying to keep track of all the 'homes' is a pain.

DanTheMan827:
What about tilde expansion?

DanTheMan827:
I mean you COULD point ~/ to the windows user profile path as well...

But that could just complicate things even more...

DanTheMan827:
I changed AdvancedSettings.cpp to substitute ~ to getenv("HOME")

If there is a better way please let me know

Memphiz:
mhh - @davilla what do you think?

MrMC:
I like it :)

rebase out that merge :)

DanTheMan827:
So here's an interesting bug....

If there is a DVD iso inside the home folder and you browse into a folder inside that ISO and then go back up outside the ISO you won't be at the substitute path anymore but rather the absolute path...

On a sandboxed install you can't go past the app container so that means it is then impossible to get to the root file manager window

I'm pretty sure this is a general bug with path substitution / file manager... it just happens to be a bigger bug on sandboxed apps...

DanTheMan827:
@Memphiz is there something I need to change on this or did it just get lost in the long list of PRs?

It does seem there is a bug with file manager and going into an ISO with path substitution... An easy fix? or would it be better to make special://envhome?

Memphiz:
That bug you figured out is due to path substitution is a one way ticket - there is no way to do a reverse lookups from absolute path to substitution and there most likely never will be one (due to possible conflictes where multiple substitutions are pointing to the same or partly same directory - you can't reverse that substitution in an unique way then) - well if this bug doesn't make this feature totally useless i consider merging it in the next merge window (starting on june 1st) - needs a rebase first.

DanTheMan827:
I changed it from tilde expansion just to using special://envhome

~/ looks much cleaner than special://envhome/ but the latter works...

Memphiz:
jenkins build and merge

DanTheMan827:
@Memphiz I guess this never actually got merged for some reason...

I rebased it on the current code and it passes appveyor testing (still waiting on travis-ci)

Memphiz:
Stupid Jenkins Build this please

Memphiz:
@Paxxi just for double checking (i am pretty sure this PR doesn't touch anything for windows) - could you have a look at the win32 jenkins log of this PR that failed to build and confirm that this is some unrelated glitch?

MartijnKaijser:
Unrelated as it fails in ffmpeg building. Likely workspace is stuck

Memphiz:
@MartijnKaijser this was ment to get merged 1 month ago already - i simply lost track of it because i thought jenkins would merge it ... any chance we can get this in even that we are not in the merge window anymore?

Memphiz:
thx :)

Tolriq:
This PR triggers assert at CSpecialProtocol::GetPath as envhome is not defined on Windows. (From : void CSpecialProtocol::LogPaths())

xbmc/Application.cpp
```diff
@@ -469,7 +474,12 @@ bool CApplication::Create()
   win32_exception::install_handler();
 
 #endif
-
+  
+  #if defined(TARGET_POSIX)
```
Memphiz: I think we need to do this for Windows aswell then @Tolriq @DanTheMan827

Memphiz:
Sorry for that. @DanTheMan827 welcome to multiplatform development. Could you craft a quick fix for this?

Paxxi:
don't worry, fetzerch patched it #10175

PR #9782, opened 2016-05-09
fix frameMoveGuard intruduced here by 82218b570c8d5e7d2ace70cbe71f24e…

FernetMenta:
…101e72163

@mapfau grabbing the lock in the constructor is wrong, we want to lock the render thread.

@MartijnKaijser I am almost sure this fixes the issue on Android


MartijnKaijser:
problem solved. thx

mazeyx:
nice work :) thx

PR #9784, opened 2016-05-09
Add parameter to set action on video click

deblockt:
The new parameter "Action on video click" have two values :
    - Show informations : Show video informations when click on item
    - Play video : Play video when click on item

The parameter is visible :
- Skin > file list settings menu
- "Basic" parameter mode


MartijnKaijser:
This option is already available. "Default select action".

PR #9785, opened 2016-05-09
Add plugin on android:scheme to open emby video

deblockt:


Jalle19:
@koying

koying:
How does it work, i.e. how is the "plugin:" url generated?

deblockt:
The "plugin:" URL is generated by Emby plugin. Emby uses it to play video (with smb or http).

URL is saved on files.strFileName column.
this is an URL exemple : 
```
plugin://plugin.video.emby.movies/?dbid=1&mode=play&id=e88aeee4c4f9171cceab2bb2e5a31856&filename=4+Bassets+pour+un+danois.avi
```

koying:
Ok, but adding "plugin:" to the manifest is for opening "plugin://..." url's from *outside* kodi.
So it'd eg means a web page with a "plugin://" url that, when clicked, would open Kodi to play it.

Is it what you're trying to achieve?

deblockt:
I have wrote an external app to show recomendation tiles on android TV.
When we click on recomendation tiles, i open kodi intent with file URL as parameter.

When file URL start by plugin://, a blue screen is displayed because plugin:// is not on android:scheme and intent can not be started.

koying:
You don't need this, then.
Just create an intent targeted at Kodi:

```
Intent intent = new Intent();
intent.setComponent(new ComponentName("org.xbmc.kodi","org.xbmc.kodi.Splash"));
intent.setAction(Intent.ACTION_VIEW);
intent.setData(Uri.parse("plugin://..."));
```

See also https://github.com/Razzeee/xbmc/pull/2

MartijnKaijser:
i guess this can be closed?

PR #9788, opened 2016-05-09
Cleanup taglibtagloader a bit and fix null padded tags

Paxxi:
Hide taglib includes from rest of Kodi code and move some helper methods into cpp file

### DIFF
--- a/xbmc/music/tags/TagLoaderTagLib.cpp
+++ b/xbmc/music/tags/TagLoaderTagLib.cpp
@@ -28,6 +28,28 @@
 #include <taglib/asftag.h>
 #include <taglib/xiphcomment.h>
 #include <taglib/id3v1genres.h>
+#include <taglib/aifffile.h>
+#include <taglib/apefile.h>
+#include <taglib/asffile.h>
+#include <taglib/modfile.h>
+#include <taglib/mp4file.h>
+#include <taglib/mpegfile.h>
+#include <taglib/oggfile.h>
+#include <taglib/oggflacfile.h>
+#include <taglib/rifffile.h>
+#include <taglib/speexfile.h>
+#include <taglib/s3mfile.h>
+#include <taglib/trueaudiofile.h>
+#include <taglib/vorbisfile.h>
+#include <taglib/wavfile.h>
+#include <taglib/wavpackfile.h>
+#include <taglib/xmfile.h>
+#include <taglib/flacfile.h>
+#include <taglib/itfile.h>
+#include <taglib/mpcfile.h>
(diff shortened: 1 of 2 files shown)


Paxxi:
@Razzeee @DaveTBlake review, maybe not the cleanest changeset but changes should be good

DaveTBlake:
Thanks @Paxxi, I've been off-line a couple of days and won't be able to check this fully until Friday. It is odd how the null padding issue related to our string handler didn't show up in 15 or 16. Do you understand why that was?

Razzeee:
Can you split the fix from the cleanup - commit wise

DaveTBlake:
As I understand it the fix to nul padded ID3v1 tags problem is the removeal of TagStringHandler, a separate commit would make that clearer. :)

But what I don't understand is why exactly the same TagStringHandler code works in Jarvis. What else changed with 17 that made this an issue? The Taglib guys think our code would have had the same issue with 1.9.1, so the change to 1.10 is not the cause. Was it the VS2015 change? I guess I would need to compile Taglib 1.9.1 in VS2015 to know. @Paxxi do you understand it?

Paxxi:
I'll make a separate commit. No I haven't dug that deep into it but there's not any reason to use our string handling as taglib does conversions itself so I just dropped it

DaveTBlake:
No need to dig @Paxxi I found my explanation and can be at peace. There was a change in taglib 1.10 to the way tstring.cpp String::copyFromUFT8 worked, hence how the nul padding issue appeared.

Good to drop the unnecessary and broken string handling, and thanks for the tidy-up. I will give it a bit more testing tomorrow (managed to break my dev setup).

DaveTBlake:
Thanks for splitting the commits.
The latter fixes the singles (music files with ID3v1 tags and not album title) appearing as albums issue reported in http://trac.kodi.tv/ticket/16454 caused by those nul padded strings.

DaveTBlake:
jenkins build this please

DaveTBlake:
@Paxxi nice crop of errors on all platforms except Win32, care to fix them please.

Paxxi:
jenkins build this please

DaveTBlake:
@Paxxi jenkins finally attempted this build, but still fails on multiple platforms.

Paxxi:
jenkins build this please
I don't think it picked up the changes judging by the errors

Paxxi:
@DaveTBlake there we go, all good

DaveTBlake:
Thanks @Paxxi. Tests fine for me too, do you want to hit the button or shall I?

PR #9791, opened 2016-05-10
[depends] libzip 1.1.2

MartijnKaijser:


MartijnKaijser:
Tested ok with installing addons on Android

PR #9792, opened 2016-05-11
VideoPlayer: drop old workaround in ffmpeg demuxer

FernetMenta:
no evidence that this is needed. setting NOPTS for 0 is wrong


popcornmix:
+1
This hack has caused me grief long ago as VideoCore doesn't like an unknown timestamp on first packet.

PR #9793, opened 2016-05-11
VideoPlayer: fix flush after seek for inputstreams implementing IPosTime

FernetMenta:
follow-up on https://github.com/xbmc/xbmc/pull/9694

@afedchin the problem was that DVDClock was not set for i.e. blurays


afedchin:
confirm, this fixes the issue with timeToDisplay. thanks.

PR #9794, opened 2016-05-11
Use separators from language addon

phate89:
Since some system doesn't have a locale I changed the number formatting to use an option in the language addon.
If there's not that option in the language addon it will fall back to en-us (like 100,000.45)
@stefansaraev @MilhouseVH This should make it work also for Open/LibreElec

I'm not sure if we want an option to let the user choose like for dates (and what other options there are, I know only 1,000.4 and 1.000,4)

### DIFF
--- a/addons/resource.language.en_gb/addon.xml
+++ b/addons/resource.language.en_gb/addon.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="resource.language.en_gb"
-  version="2.0.0"
+  version="2.0.1"
   name="English"
   provider-name="Team Kodi">
   <requires>
--- a/addons/resource.language.en_gb/resources/langinfo.xml
+++ b/addons/resource.language.en_gb/resources/langinfo.xml
@@ -54,6 +54,8 @@
       <time symbolAM="AM" symbolPM="PM">h:mm:ss xx</time>
       <tempunit>F</tempunit>
       <speedunit>mph</speedunit>
+      <thousandsseparator groupingformat="\3">,</thousandsseparator>
+      <decimalseparator>.</decimalseparator>
     </region>
 
     <region name="USA (24h)" locale="US">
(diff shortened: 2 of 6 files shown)


stefansaraev:
thanks! I dont remember what exacly was wrong with separators (how to reproduce), so I'll let @MilhouseVH test this ;)

btw, all posix systems should have POSIX (and C, which is guaranteed to be same as POSIX), not sure if this info may be useful to you.

MilhouseVH:
Thanks for this, it certainly seem to be working. I'm setup for English UK (24h), and now see the comma grouping for movie votes etc. Thanks.

Presumably `thousandsseparator` and `decimalseparator` will eventually need to be added to for all regions in all langinfo.xml that use different values to `en-US`?

phate89:
oops I forgot to set the fallback also for grouping. Without that it wasn't working. Now the fallback works.
Right now I set it only for USA (12h) because I'm not sure all the regions use that method (and they all fall back to that format anyways). Anyone knows formats for that regions?

MilhouseVH:
Thanks, looks good - I've got the comma group separator for English UK (24h) now.

I'd have thought all native English-speaking regions would use comma as the group (thousands) separator and period as the decimal point, so as you have it now should be fine for the English regions. Although perhaps Central Europe and India could be different.

Obviously this isn't going to work so well for those languages/regions that use period/comma as group/decimal separator as they will now fall back to comma/period. This will need localising for all the regions that need it, maybe that should be handled by subsequent PR(s)?

phate89:
I checked manually with windows locales (in english addon) and seems they're all the same except for india where they group first 3 numbers and then by 2 (something like 10,00,00,000).
I added all the values in the shipped english addon. All other addons needs to be updated but in the repository..

stefansaraev:
jenkins build this please

stefansaraev:
@MilhouseVH @phate89 ready to go ?

MilhouseVH:
I'm happy, although other non-English languages will need to their defaults added once this merges.

PR #9795, opened 2016-05-11
fixed: Caching of quasi-internet streams like webdav was broken

arnova:
Fixes issue mentioned here http://forum.kodi.tv/showthread.php?tid=272973


arnova:
jenkins build and merge

PR #7472, opened 2015-07-10
[fix][libexif] Various fixes for exif processing

ace20022:

PR #9796, opened 2016-05-11
[exif] Fix for out-of-memory errors with large numbers of jpegs

popcornmix:
This reverts part of https://github.com/xbmc/xbmc/pull/7472

Basically the commit made the 4 comments in the exif block increase
from 2K to 64K each, so you now need 256K per photo.

When opening a folder exif information for all photos is extracted.
So, for a folder of 5000 jpegs, 1.2GB of RAM is needed just for
the comments.

As a 64K comment string is of no use to kodi, just truncate them to 2K
like we used to.

See:
http://trac.kodi.tv/ticket/16193
http://forum.kodi.tv/showthread.php?tid=251908

### DIFF
--- a/lib/libexif/libexif.h
+++ b/lib/libexif/libexif.h
@@ -81,7 +81,7 @@ typedef struct {
 #define EXIF_COMMENT_CHARSET_UNICODE    3 // Exif: Unicode (UTF-16)
 #define EXIF_COMMENT_CHARSET_JIS        4 // Exif: JIS X208-1990
 
-#define MAX_COMMENT 65533 // 2 bytes - 2 for the length param
+#define MAX_COMMENT 2000
 #define MAX_DATE_COPIES 10
 
 typedef struct {


ace20022:
I wouldn't call it a fix, but +1 anyway

bkuhls:
Please wait with committing this patch, I tried it shortly yesterday on Jarvis and kodi segfaulted when starting the picture viewer, the crash happened before the folder list appeared on the screen. In the coming days I will check the backtrace.

popcornmix:
@bkuhls did you rebuild lib/libexif as well as kodi? They both use this define.

bkuhls:
I cross-compiled Kodi from scratch, the source code was extracted into an empty directory. But I forgot to update libexif-i486-linux.so on the target machine, thanks for the hint. After updating both libexif-i486-linux.so and kodi.bin I can confirm that your patch fixes indeed the crash of Kodi 16.1-Jarvis when opening directories with many picture files! :)
Does this fix justify a 16.2 release (just hoping ;))? Maybe also including the gcc6 build fix  https://github.com/xbmc/xbmc/pull/9703 ?

stefansaraev:
travis build result is suficient

da-anda:
thanks popcornmix. But why do we extract exif comments for listings upfront? It makes no sense at all. Would there be a way to lazyload the info once requested by the skin?

popcornmix:
Lazy loading is absolutely the best solution. The same applies to music, where kodi currently adds perhaps 50K songs to a playlist before playing the song you want from the songs node.

But it is not trivial, and it's not code that I am familiar with. Volunteers for this job would be welcome.

Paxxi:
Lazy loading is a pretty big task as none of the ui code is really designed for it. I do agree it's the best solution for it though and should be done sooner or later

theincogtion:
Is the discussed problem here the reason for these crashes:
https://discourse.osmc.tv/t/osmc-crashing-when-opening-photos/16511/2

?

markst:
This fix does prevent crashing. However is very very slow at loading for the first time.
Is it necessary for the slideshow to process every single recursive image before presenting?

da-anda:
Kodi does not yet support lazy loading of data, so yes, currently all items need to be processed upfront. If you have the skills, any help is welcome in this regard.

markst:
understood. @da-anda I'd love to contribute. unfortunately I have a lack any cpp knowledge!

PR #9797, opened 2016-05-11
mmalrenderer: Avoid creating presenting thread when it is not required

popcornmix:
This is mostly to avoid debug logging spam when it is not being used.


popcornmix:
jenkins build this please

PR #9798, opened 2016-05-11
VideoPlayer: Different accurate value for omxplayer_mode has no effect

popcornmix:


popcornmix:
Ping @FernetMenta 
We actually ignore the accurate parameter inside FlushBuffers in omxplayer mode, so may as well remove a couple of omxplayer specific lines.

FernetMenta:
+1

popcornmix:
jenkins build and merge

PR #9800, opened 2016-05-11
rbp: Fix possible leak in rendercapture interface

popcornmix:
See: http://forum.kodi.tv/showthread.php?tid=269814&pid=2332575#pid2332575


popcornmix:
Fix confirmed here: http://forum.kodi.tv/showthread.php?tid=269814&pid=2333344#pid2333344
jenkins build this please

PR #9801, opened 2016-05-11
mmalcodec: Set initial aligned height explicitly to avoid multiple reconfigures in renderer

popcornmix:


popcornmix:
jenkins build this please

PR #9802, opened 2016-05-11
[cleanup] drop apkfile/directory and libzip dependency

stefansaraev:
this drops android specific apkfile/apkdirectory, handling apks is now done via zipfile/zipdirectory, just like normal zips. apk:// protocol is still there

this also drops libzip from depends.


stefansaraev:
what do android people think about this.

@koying if I am not mistaken, you picked this for spmc (or was it a testing branch?)

jenkins build this please

PR #9666, opened 2016-04-22
[python-api] - add preselect params to Dialog.multiselect() and Dialog().select()

phil65:
Requested here for example: http://forum.kodi.tv/showthread.php?tid=270814
I also have some usecases for this.

PR #9807, opened 2016-05-12
[python] fix seq fault if optional list is leaved empty

AlwinEsch:
This is a fix related to request #9666.

If the optional list entry (vector) was leaved empty was a `nullptr` given to `AddonModuleXbmcgui.cpp` who has created the crash.

This fix add a check of the value to confirm it is present.


phil65:
@tamland , could you review, please?

PR #9808, opened 2016-05-13
RenderSystem: Load correct identity

fritsch:
Someone on IRC asked me if that would look correct. It does not. Somebody knows that code as the change / fix also seems to have no visual influence.

### DIFF
--- a/xbmc/rendering/gl/RenderSystemGL.cpp
+++ b/xbmc/rendering/gl/RenderSystemGL.cpp
@@ -186,7 +186,7 @@ bool CRenderSystemGL::ResetRenderSystem(int width, int height, bool fullScreen, 
   glEnable(GL_SCISSOR_TEST);
 
   glMatrixProject.Clear();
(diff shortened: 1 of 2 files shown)


fritsch:
@popcornmix I don't fully understand what we do here - as this seems to have no influence. Any idea?

popcornmix:
The patched code looks more like what was intended, both from a grouping and by removing duplication - glMatrixModview->LoadIdentity() is called twice with nothing reading/writing in between.
LoadIdentity just sets the matrix to identity matrix (all zeros with ones on leading diagonal).

I don't really know this code well, so assuming no one can give a more definitive answer, I'll add it to my tree and we'll see anyone spots a difference in the Milhouse builds.

If no differences occur, I'd vote for merging this - it looks like a c/p error.

popcornmix:
No complaints from milhouse builds, so it's probably safe to merge.

MartijnKaijser:
ping for merge?

PR #9809, opened 2016-05-14
Fix dateadded sort method in AudioLibrary.GetSongs

DaveTBlake:
AudioLibrary.GetSongs is broken when using sort method dateadded, see
http://forum.kodi.tv/showthread.php?tid=273321

FieldDateAdded had been omitted from a case statement when this sort method was added to the JSON-RPC API for Jarvis.

Thanks for spotting this @Millencolin007
@Montellese, @Tolriq for interest.


DaveTBlake:
Anyone mind if I just merge this?

PR #9811, opened 2016-05-14
[python][addon] Add nullptr checks to parameter handling

Paxxi:
This should solve the crash introduced with #9666
@phil65 for review, works with your sample addon


Paxxi:
jenkins build this please

AlwinEsch:
Hi @Paxxi #9807 :)

I close my then.

Paxxi:
Sorry @AlwinEsch didn't see you had already looked into it, both work so whichever is fine by me

phil65:
Thx to both of you for investigating. :) Choose whatever you think is the better fix.

PR #9812, opened 2016-05-14
Music artist tag processing improvements

DaveTBlake:
Improve individual artist identification from music file tags, for both songs and albums,  by making processing more flexible and able to accomodate both how popular tagging software like Musicbrainz Picard tags music files by default, and how older files may have already been tagged.

Although FLAC/Vorbis/Xiph and ID3v2.4 allow for tags to have multiple values, there are many occassions where multiple artist names are combined in a single tag string. Kodi currently allows for a single item separator (space-slash-space by default), so recognises that "artist1 / artist2" is 2 artists, and requires users to always use the same separator in all music files. Inevitably not all music files are that consistent. For sometime Kodi has made use of Musicbrainz Id tags when present to help identify artists, but there are still many people that have not re-tagged their music this way. 

The aim is that Kodi will correctly identify all the individual artists from a string when artist or albumartist tags are like "artist1 ft. artist2" or "artist1; artist2"  or "artist1    / artist2" (trim white space). Parsing artist names is non-trivial because so many characters could be part of a name on one ocassion and a join phrase linking individuals on another. Some users also want to treat the artist tag as a display string, so don't want to replace all the join phrases with a fixed separator like space-slash-space.

The solution proposed here is to allow for multiple separator values e.g. " feat. ", " ft. ", ";" etc. in addition to the default single item separator, and make them user configurable via advancedsettings.xml.  

Secondly is to make use of hints (ARTISTS and ALBUMARTISTS tag) to help identify individual artist names even when musicbrainz artist id tag is not present. This provides a simple way for a user to add a tag that separate individual artists without having to use mbids, or change the way the artist credit is displayed.

@zag2me step by step moving to the robust tag processing we have discussed elsewhere.
@Razzeee or @phate89  maybe have some time to check this over?

### DIFF
--- a/xbmc/settings/AdvancedSettings.cpp
+++ b/xbmc/settings/AdvancedSettings.cpp
@@ -280,6 +280,7 @@ void CAdvancedSettings::Initialize()
   m_strMusicLibraryAlbumFormat = "";
   m_prioritiseAPEv2tags = false;
   m_musicItemSeparator = " / ";
+  m_musicArtistSeparators = { ";", ":", "|", " feat. ", " ft. " };
   m_videoItemSeparator = " / ";
   m_iMusicLibraryDateAdded = 1; // prefer mtime over ctime and current time
 
@@ -702,6 +703,19 @@ void CAdvancedSettings::ParseSettingsFile(const std::string &file)
     XMLUtils::GetString(pElement, "albumformat", m_strMusicLibraryAlbumFormat);
     XMLUtils::GetString(pElement, "itemseparator", m_musicItemSeparator);
     XMLUtils::GetInt(pElement, "dateadded", m_iMusicLibraryDateAdded);
+    //Music artist name separators
+    TiXmlElement* separators = pElement->FirstChildElement("artistseparators");
+    if (separators)
+    {
+      m_musicArtistSeparators.clear();
+      TiXmlNode* separator = separators->FirstChild("separator");
+      while (separator)
+      {
+        if (separator->FirstChild())
+          m_musicArtistSeparators.push_back(separator->FirstChild()->ValueStr());
+        separator = separator->NextSibling("separator");
+      }
+    }
   }
 
   pElement = pRootElement->FirstChildElement("videolibrary");
--- a/xbmc/music/Album.cpp
+++ b/xbmc/music/Album.cpp
@@ -155,10 +155,20 @@ CAlbum::CAlbum(const CFileItem& item)
     }
   }
   else
-  { // no musicbrainz info, so fill in directly
-    for (std::vector<std::string>::const_iterator it = tag.GetAlbumArtist().begin(); it != tag.GetAlbumArtist().end(); ++it)
+  { // No musicbrainz album artist ids so fill artist names directly.
+    // This method only called when there is a musicbrainz album id, so means mbid tags are incomplete.
+    // Try to separate album artist names further, and trim blank space.
+    std::vector<std::string> albumArtists = tag.GetAlbumArtist();
+    if (musicBrainzAlbumArtistHints.size() > albumArtists.size())
+      // Make use of hints (ALBUMARTISTS tag), when present, to separate artist names
+      albumArtists = musicBrainzAlbumArtistHints;
+    else
+      // Split album artist names further using multiple possible delimiters, over single separator applied in Tag loader
+      albumArtists = StringUtils::SplitMulti(albumArtists, g_advancedSettings.m_musicArtistSeparators);
+
+    for (auto artistname : albumArtists)
     {
-      artistCredits.emplace_back(*it);
+      artistCredits.emplace_back(StringUtils::Trim(artistname));
     }
   }
 
--- a/xbmc/music/Song.cpp
+++ b/xbmc/music/Song.cpp
@@ -43,7 +43,7 @@ CSong::CSong(CFileItem& item)
   { // Have musicbrainz artist info, so use it
 
     // Vector of possible separators in the order least likely to be part of artist name
-    const std::vector<std::string> separators{ " feat. ", ";", ":", "|", "#", "/", ",", "&" };
+    const std::vector<std::string> separators{ " feat. ", " ft. ", " Feat. "," Ft. ", ";", ":", "|", "#", "/", " with ", ",", "&" };
 
     // Establish tag consistency - do the number of musicbrainz ids and number of names in hints or artist match
     if (tag.GetMusicBrainzArtistID().size() != musicBrainzArtistHints.size() &&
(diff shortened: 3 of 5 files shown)


xbmc/settings/AdvancedSettings.cpp
```diff
@@ -280,6 +280,7 @@ void CAdvancedSettings::Initialize()
   m_strMusicLibraryAlbumFormat = "";
   m_prioritiseAPEv2tags = false;
   m_musicItemSeparator = " / ";
+  m_musicArtistSeparators = { ";", ":", "|", "#", " feat. ", " ft. " };
```
garbear: On spotify I found an artist named 10 Ft. Ganja Plant

DaveTBlake: Thanks for taking an interest @garbear, there are some crazy artist names out there!

This is one reason why the music artist separators are configurable. If you have "10 Ft. Ganja Plant" in your music collection then you can drop " ft " as a music artist separator, or use ARTISTS tag or tag with Musicbrainz IDs (having added it to their database) in those songs. 

Many users have "artist1 ft. artist2" in tags, so I think it reasonable to have " ft. " as a default setting.

zag2me: Could be made case sensitive to avoid that maybe? Ft. = foot and ft. = featuring? who knows!

zag2me:
Yep the logic looks good. I'll need a test build but not sure how useful I will be with a perfect picard library these days. This will need user testing I guess on different libraries.

DaveTBlake:
Thanks @zag2me, I'll set a Win32 build going for you. Practically the best approach for testing with a diverse range ot real-life less than perfectly tagged music files, is to get it into the nightlies. We are still alpha, and I am on hand to deal with any unexpected consequences.

DaveTBlake:
OK, build on the mirrors at http://mirrors.kodi.tv/test-builds/win32/KodiSetup-20160514-13c4904-HEAD.exe or appropriate

Tolriq:
@DaveTBlake not a lot available but if you want some tests with a very very less than perfectly tagged library but with large data just tell :) (I just need to know what to seek before / after)

zag2me:
OK tested and all looks good on my library. Every artist seems to be scanned correctly. My library is 100% picard named.

I did notice lots of debug messages (50mb on full scan), here's a snippet.

http://xbmclogs.com/pgdy1ppwl

DaveTBlake:
Thanks @zag2me that is reassuring, confirms that I have not broken mbid tag handling. 

The log messages are a reminder that Picard adds many tags that Kodi still doesn't use e.g.  ACOUSTID_ID, ALBUMARTISTSORT, ARTISTSORT, ASIN, BARCODE, CATALOGNUMBER, DISCSUBTITLE, DISCTOTAL, MEDIA, MUSICBRAINZ_RELEASEGROUPID, MUSICBRAINZ_RELEASETRACKID, ORIGINALDATE, ORIGINALYEAR, RELEASECOUNTRY, RELEASESTATUS, RELEASETYPE, SCRIPT, TOTALDISCS, TOTALTRACKS, TRACKTOTAL

I can't think of a use for these for now! Not sure we need the messages bloating the log either, but that is not for here.

DaveTBlake:
@Tolriq testing is always welcome. What to look for....

The artist list should better identify the individual artists without messing up band names etc.

Enable display of all artists and check on the artists list. Where you originally have "artist1 feat. artist2"  or "artist1; artist2" as a single artist because Kodi did not identify individuals (and no mbids), or space padding infront of artist name (shows when sorted), on a re-scan these artists should be separated and trimmmed. But other names like "thing1 & thing2" or "thing1, thing2 and thing3" will still be taken as a single artist, unless you add ARTISTS/ALBUMARTISTS tags as hints, or of course tag fully with Picard adding mbids.

Additionally this extra spiltting is configurable, so if there is a join phrase in French e.g.  a language version of " feat. " that you want always used to split artists then add that to advancedsettings.xml for example my list of spearators is
```
 <musiclibrary>
    <artistseparators>
	<separator> feat. </separator>
	<separator> with </separator>
	<separator>;</separator>
	<separator>#</separator>
	<separator> ft. </separator>
    </artistseparators>
</musiclibrary>
```

Tolriq:
Full scrape started, will report in a few hours.

Good things for the trim, I was doing it at Yatse side since ages as it's a frequent problem (don't know if you cover every case even when no split but maybe worth a check).

In France most of the time ft. is used, other separator could be also used as normal band name.
But I've just checked and I have a lot of XXX & YYY. So I suppose **&** could be a standard default separator ?

And sending by PM an MP3 that trigger an artist that only shows ?????? (Did not have such ???? artist in 16.x full scrape).

Tolriq:
Well it was way faster than before (about 40mins for 22100 songs)

First Kodi find 68 more artists with this patch. (3516 artists now)

1) It seems separators are case sensitive Feat. and Ft. are not handled.
2) Remaining common separators found : 
- XXX, YYY, XXX (very very common)
- XXX & YYY (very very common)
- XXX featuring YYY (not common)
- XXX Feat YYY (no dot) (rare)
- XXX (feat. YYY) (rare)

3) I now have a lot's of duplicates "arious Artists" and "Various Artists" I guess the arious Artists can't be handled due to very bad tagging and missing utf8 BOM maybe.
But the Various Artists case should need special attention, I suppose this is a common case for users to have some MBID tags that where not correctly handled before so to manage compilation in previous Kodi have just changed the artist to Various Artists without touching other tags. But now this will triggers tons of Various Artists associated with MBIDs. MBIDs that could be correct for other artist on other songs but will be locked to Various Artists if this is what was scrapped first.
IMO Various Artists value should prevail over the rest as a long legacy heritage.

Not 100% related but @phil65 it seems there's no way to get the song full path in estuary ? It's quite hard to find the file associated with the bad tags to edit them and fix them.

DaveTBlake:
@Tolriq thanks for the test feedback.

This trims all names that get written to the database (not just split), so re-scanning will tidy up but just upgrading database version will not. This is a common problem in user music files and makes a mess of sorting, so nice to solve.

The "????" artist was from a file with no title or artist tags, I'll take a closer look at how that is handled.

Common separators: on side of caution yes case sensitive, I could add " Ft. " and " Feat. " to the default set. As separators "&" and "," are in band names e.g. "Earth, wind & Fire", so not in default but a user could add them if they really wanted all "A & B" separated. Similarly with "featuring".

Processing brackets is more involved, for another PR maybe? I prefer to chip away at this than have a big-bang!

Various artists: This needs a separate approach (and PR). I'm game to tackle it sometime, after all how painful can it be? I best start a forum thread about what is required and the issues.

zag2me:
Artists that may cause issues:

"#" - http://www.theaudiodb.com/artist/120171
";" - none
"," - http://www.theaudiodb.com/artist/132015 
"," - http://www.theaudiodb.com/artist/139814 
"," - http://www.theaudiodb.com/artist/121676 
"," - http://www.theaudiodb.com/artist/133482 
"," - http://www.theaudiodb.com/artist/125257
"ft." - http://www.theaudiodb.com/artist/113831
"|" - none

Basically I think we are pretty good, that's actually very little artist names. I'm totally going off to listen to "10 ft. ganja plant" now :palm_tree: :palm_tree:

Tolriq:
> Common separators: on side of caution yes case sensitive, I could add " Ft. " and " Feat. " to the default set. As separators "&" and "," are in band names e.g. "Earth, wind & Fire", so not in default but a user could add them if they really wanted all "A & B" separated. Similarly with "featuring".

Not for this PR but another one but if there's risks of collision for some separators, what about some exclusion list ? Like how could I handle & for 99% of my cases but not 1.
One solution could maybe be : First check if artist is present in DB ? Then when later you add support for the special nfo folder for artists you handle this one first so special cases are in DB before the actual scrape ?

DaveTBlake:
Thanks for the artist name data check @zag2me 

I'm not sure how far (in another PR) Kodi needs to go with name parsing and collision avoidence @Tolriq.  For those odd cases the user could tag albums from those artists with mbids or ARTISTS tag. The multi separators can be reduced or turned off completely if a user doesn't like the results.

Let's see how users get on with this improvement, I figure just " ft. " will enhance Kodi's handling of quite a few user music collections.

DaveTBlake:
jenkins build this please

PR #9501, opened 2016-03-30
[cmake][addons] Raise an error when ADDON_SRC_PREFIX is an relative path

AchimTuran:
@wsnipex @ksooo @AlwinEsch @FernetMenta @Jalle19 @mapfau 

After having a chat with @FernetMenta we decided that `ADDON_SRC_PREFIX` has to be an absolute path, because relative paths can easily break and you can search the issue for hours.

Furthermore I adapted the README to mention that we have to use absolute paths in the future.

Hopefully this will keep it simple and stop confusing our devs with relative paths. But we have to adapt our Travis CI and AppVeyor scripts.

PR #9813, opened 2016-05-15
[cmake/addons] Make relative ADDON_SRC_PREFIX relative to the build dir

fetzerch:
This is a counter proposal for https://github.com/xbmc/xbmc/pull/9501 as the discussion stalled.
As discussed in that PR, instead of forcing the user to provide an absolute directory, we make ADDON_SRC_PREFIX relative to the root build directory, aka the directory from where you issue the cmake command. This is both consistent with CMake and with any other shell commands where path parameters are usually taken relative to the current working directory.

I'm not aware of further issues this might cause. But if there are, please let me know and let's fix the root causes. The addon system is already difficult enough for developers. They shouldn't have to deal with restrictions, that we can fix in the system. 

pings: @notspiff, @wsnipex, @FernetMenta, @AchimTuran, @Montellese, @hudokkow

### DIFF
--- a/project/cmake/addons/CMakeLists.txt
+++ b/project/cmake/addons/CMakeLists.txt
@@ -138,7 +138,9 @@ endif()
 get_filename_component(ADDONS_DEFINITION_DIR "${ADDONS_DEFINITION_DIR}" ABSOLUTE)
 
 if(ADDON_SRC_PREFIX)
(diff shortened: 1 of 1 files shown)


FernetMenta:
> we make ADDON_SRC_PREFIX relative to the root build directory

+1, this is what most people think how it should work

fetzerch:
Thanks, will merge it now. Just wanted to mention that I've tested this with building retroplayer addons in the last week and it worked as expected. But as mentioned, let me know if there are further issues with the addon build system.

PR #9814, opened 2016-05-15
[cleanup] remove DEFAULT_SKIN from system.h

stefansaraev:
not used after Gotham


ronie:
it's used by one of the binary addons.
perhaps we should add a note to the code as this is the fifth time someone wants to remove it...

stefansaraev:
ouch?

why would a binary addon care what's the default skin..

stefansaraev:
https://github.com/notspiff/audiodecoder.timidity/blob/master/lib/timidity/interface/xskin_i.c#L937-L939

I dont see how and where our system.h is used in audiodecoder.timidity or am I blind? cc @notspiff

akva2:
Heh. It is not a very unique define name. That code refers to another define and we never hit that code in the timidity addon.

PR #9815, opened 2016-05-15
[cmake/win32] Fix syncing cpluff.dll to build tree

fetzerch:
A clean build with CMake will cause an error when trying to run Kodi,
because cpluff.dll is not found in the synced tree. Reason is that the
list of files that are synced (from installdata directory) is done at
configure time but cpluff.dll is only compiled at build time.

These lines would be obsolete if we could properly compile cpluff in the
build tree.

for @AchimTuran


AchimTuran:
> These lines would be obsolete if we could properly compile cpluff in the
build tree.

Should we create a new jira ticket for this? So that we don't forget about that.

PR #9816, opened 2016-05-15
[cmake] Small binary-addon build system improvements

fetzerch:
Let's try to make the addon build system a bit more user friendly.

For the moment, issue an error when the addon config file couldn't be found. I think it would be also convenient that if `ADDON_SRC_PREFIX` contains the source directory of the addon specified in `ADDONS_TO_BUILD`, then just use it even without config. That makes the bootstrap process for addon devs a bit easier.

ping: @notspiff, @wsnipex, @hudokkow, @AchimTuran

### DIFF
--- a/project/cmake/addons/CMakeLists.txt
+++ b/project/cmake/addons/CMakeLists.txt
@@ -224,6 +224,10 @@ if(NOT addons)
   endif()
 endif()
 
+# Track if at least one addon has been found. Everything else is likely an
+# error either in ADDONS_TO_BUILD or in the directory configuration.
+set(SUPPORTED_ADDON_FOUND FALSE)
+
 foreach(addon ${addons})
   if(NOT (addon MATCHES platforms.txt))
     file(STRINGS ${addon} def)
@@ -243,6 +247,7 @@ foreach(addon ${addons})
 
     if(ADDON_FOUND)
       message(STATUS "\n-- ---- Configuring addon ${addon} ----")
+      set(SUPPORTED_ADDON_FOUND TRUE)
 
       get_filename_component(dir ${addon} PATH)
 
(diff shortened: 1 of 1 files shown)


project/cmake/addons/CMakeLists.txt
```diff
@@ -416,6 +425,12 @@ if(NEED_SUDO)
                     COMMAND sudo -k)
 endif()
 
+if(NOT SUPPORTED_ADDON_FOUND)
+  message(FATAL_ERROR "${ADDONS_TO_BUILD} did not match any of the supported addons. \
```
AchimTuran: This is great and will help new users to setup our build system.

AchimTuran:
> I think it would be also convenient that if ADDON_SRC_PREFIX contains the source directory of the addon specified in ADDONS_TO_BUILD, then just use it even without config. That makes the bootstrap process for addon devs a bit easier.

Definitly a step in the right direction.

wsnipex:
looks good.

>I think it would be also convenient that if ADDON_SRC_PREFIX contains the source directory of the addon specified in ADDONS_TO_BUILD, then just use it even without config.

That would break if ADDONS_TO_BUILD contains more then one addon (or a regex)

AchimTuran:
> That would break if ADDONS_TO_BUILD contains more then one addon (or a regex)

When I set `ADDON_SRC_PREFIX` and `ADDONS_TO_BUILD` contains more than one add-on. I assume that the path in `ADDON_SRC_PREFIX` contains all required <addon-id> folders and they have to contain the required sources.
What should break?
I meanly use this option for developing an add-on and not for building a list of add-ons.

wsnipex:
maybe its a misunderstand on what is meant here. If you mean simply build the addon even if no addon.foo.txt file exists - then yes, that should work.

If you mean that ADDON_SRC_PREFIX can/should contain the full path to /some/example/addon.foo (so including the addon src dir itself), then it will break.

fetzerch:
Thanks. I'm aware that it won't work with regexs, I meant just for convinience of addon devs that either want to start a new addon, or simply want to build only one specific addon that they are developing and have local sources. But that's either way material for another PR.

fetzerch:
Addon build error on linux seems unrelated, and appeared already in an earlier build. Merging.

PR #9819, opened 2016-05-16
AddonsDirectory.cpp: Fix root folder name

Guilouz:
Root folder name must be "Addon Browser" and not "Install from repository".

See : http://forum.kodi.tv/showthread.php?tid=274235


tamland:
Thanks!

PR #9821, opened 2016-05-16
fix 'unknown sources' setting visibility when opened from dialog

tamland:
This make sure the setting is actually visible and focues when going from the warning dialog.

### DIFF
--- a/xbmc/settings/windows/GUIWindowSettingsCategory.cpp
+++ b/xbmc/settings/windows/GUIWindowSettingsCategory.cpp
@@ -27,6 +27,7 @@
 #include "settings/DisplaySettings.h"
 #include "settings/Settings.h"
 #include "settings/lib/SettingSection.h"
+#include "settings/windows/GUIControlSettings.h"
 #include "utils/log.h"
 #include "view/ViewStateSettings.h"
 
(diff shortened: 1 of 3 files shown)


stefansaraev:
addons.unknownsources should be moved to level 0....

xbmc/settings/windows/GUIWindowSettingsCategory.cpp
```diff
@@ -81,12 +82,38 @@ bool CGUIWindowSettingsCategory::OnMessage(CGUIMessage &message)
...
+
+      // Ensure the setting we are about to focus will be visible.
+      auto focusSetting = message.GetStringParam(1);
+      if (!focusSetting.empty())
+      {
+        auto setting = GetSetting(focusSetting);
+        if (setting)
+        {
+          if (CViewStateSettings::GetInstance().GetSettingLevel() < setting->GetLevel())
+            CViewStateSettings::GetInstance().SetSettingLevel(setting->GetLevel());
```
stefansaraev: you should not do this, because:

1. we do support "lock settings" in profiles (+ master lock enabled). 
2. there are people using kodi in "kiosk mode" with level pre-set to 0 or 1 + skin that does not support swtiching settings levels. by upgrading current settingslevel you effectively broke those.
3. it makes no sense to force show a setting we previously wanted to hide from user, so you should focus it if accessible in current level and noop if not

Montellese:
I'm not really sure what the circumstances are as the description is not very verbose and I have no idea what warning dialog we are talking about but we should not just change the visible setting level. This is something that only users should do. If there is a setting that should be visible on a certain level but isn't, we should change the level of the setting in the setting's definition.

`CGUIWindowSettingsCategory` makes sure to only display settings matching the current level so there always is a setting to focus and if a category doesn't have any settings the category isn't shown either.

stefansaraev:
@Montellese this PR "upgrades" the current settings level. asume your is 1 but a setting is available at 2, your settinglevel is automagicaly changed to 2.

focusing this particular setting is really good thing, imo, I'd just prefer moving it into level0 so it is always visible and not touch settingslevel at all.

tamland:
The level can be automatically reverted on deinit. Don't get caught up in such pointless details...
It's fine with me to set it to 0 (not me who changed it) but that means the focusing feature will only ever be usable for settings that are level 0 and categories that have at least one level 0 setting.

Montellese:
@tamland: It will not only work for settings with level 0 but for settings with at least the level that the user has chosen. I agree that it's not ideal but programatically changing the setting (even if you change it back afterwards) is not ideal either. As @stefansaraev has pointed out there are users who might be confused by suddenly seeing a lot more settings than they are used to. It's not like they can only change the focused setting but they might also change other settings on the level that they didn't see before and maybe they are not supposed to.

Is there any reason why we don't change the current dialog to allow enabling the "unknown sources" setting directly from the dialog? That way we could also immediately continue installing the addon if the user chose to enable the setting. Currently they try to install an addon, it fails, they go to settings, enable the setting, go back and then have to manually install the addon again.

Razzeee:
I think it's mostly trying to not make it to easy to enable it. As it's something that shouldn't be done lightheartedly.

tamland:
Yeah, the point is to have some hoops to go though. If it's just a yes/no dialog nobody will read it and just click yes. Basically the same as done in android.

But it shouldn't be a puzzle figuring out how to change it. Right now it just opens setting window, and you have 'known' that the setting you're supposed to change is hidden behind a different level. This is far more confusing than the level changing imo. To avoid this situation you have to either temporarily change the level, or you can't use focusing for anything but level 0 settings.

PR #9822, opened 2016-05-16
[cleanup] - remove some unused stuff from ControlFactory

phil65:
remove some unused stuff from ControlFactory.
@ronie "imagefolder" and "imagefolderfocus" is some outdated garbage, right?


ronie:
> @ronie "imagefolder" and "imagefolderfocus" is some outdated garbage, right?

never heard of 'm. it's not used anywhere, so safe to remove.

spinInfo is used by the settings spincontrol though.

phil65:
settings spincontrol uses labelInfo instead of spinInfo 
https://github.com/xbmc/xbmc/blob/master/xbmc/guilib/GUIControlFactory.cpp#L1474

ronie:
ah, ok. good to go then.

phil65:
jenkins build this please.

PR #9823, opened 2016-05-16
Fix joysticks on android

garbear:
The android peripheral bus doesn't rely on polling, so no thread is created. The thread scans for joysticks on creation, but if no thread is created, no initial scan is done.

This change runs an initial scan in the main thread when Kodi initializes the android bus. This should allow joysticks to work on startup.

-- untested --

### DIFF
--- a/xbmc/peripherals/bus/PeripheralBus.cpp
+++ b/xbmc/peripherals/bus/PeripheralBus.cpp
@@ -249,6 +249,11 @@ void CPeripheralBus::Initialise(void)
     Create();
     SetPriority(-1);
   }
+  else
+  {
+    lock.Leave();
+    ScanForDevices();
+  }
 }
 
 void CPeripheralBus::Register(CPeripheral *peripheral)


garbear:
After testing, it appears this causes the deadlock fixed by #9825

Montellese:
I'm a bit confused. The android peripheral bus already gets all input devices in its constructor, see https://github.com/xbmc/xbmc/blob/master/xbmc/peripherals/bus/android/PeripheralBusAndroid.cpp#L51. Is that not enough?

garbear:
You're right. This PR probably won't affect the android problem. I'll start looking for other options

mk01:
@garbear 

wouldn't be better to use proven semantics - that means needspolling would be set to true in constructor, then Bus::Process() is created, calls Android::ScanForDevices() which would just set needspolling to false.

this will pass input devices for initialisation + and disable polling too. I also think the lock here is one lock too much.

like this: https://github.com/mk01/xbmc/commit/694241ef5400d508c00aa168be008b3e474d116a

Montellese:
@mk01: I'm pretty sure there are different "proven" semantics. At work we have the guideline that in derived classes anything that influences the behaviour of the base class needs to be initialized/set in the constructor which is basically what I've done.

mk01:
@Montellese 
oh, sorry for confusion. I was speaking about the other derived classes within peripherals and nothing else.

garbear:
I'm fairly certain I should close this. The deadlock fixes in #9825 are still valid.

PR #9825, opened 2016-05-17
[peripherals] Fix another deadlock

garbear:
HasPeripheral() is thread-safe so locking is not needed here.

A deadlock occurs during bus initialization when g_peripherals.CreatePeripheral()
tries to lock the mutex of its inherited Observer class while this lock is held.

### DIFF
--- a/xbmc/peripherals/Peripherals.cpp
+++ b/xbmc/peripherals/Peripherals.cpp
@@ -727,7 +727,7 @@ void CPeripherals::ProcessEvents(void)
 {
   std::vector<PeripheralBusPtr> busses;
   {
-    CSingleLock lock(m_critSection);
+    CSingleLock lock(m_critSectionBusses);
     busses = m_busses;
   }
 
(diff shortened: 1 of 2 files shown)


garbear:
I've also pushed a fix for locking the wrong mutex, though this may be incorrect.

garbear:
And a fix for another deadlock

Montellese:
Concerning your last fix what happens if one of the bus initializations (which might contain creating peripheral instances) call back into `CPeripherals` and try to get the bus which has not been added to the list of available busses yet? Or do we pass the bus instance to the created peripheral instances now (don't remember anymore)?

garbear:
> Concerning your last fix what happens if one of the bus initializations (which might contain creating peripheral instances) call back into `CPeripherals` and try to get the bus which has not been added to the list of available busses yet? Or do we pass the bus instance to the created peripheral instances now (don't remember anymore)?

I believe we pass the bus instance to the created peripheral (see https://github.com/xbmc/xbmc/blob/93dcfc0/xbmc/peripherals/bus/PeripheralBus.cpp#L122) so this should be safe

Montellese:
You're right and I actually think I added that myself to avoid deadlocks but I wasn't sure anymore...

garbear:
ok to merge then?

PR #9826, opened 2016-05-17
Player fixes

garbear:
Here's a collection of random improvements to the video and audio players I've done over the past month.

The `GUIInfoManager` and `SeekHandler` commits deal with casting time amounts into ints. The commits replace an int with a double and use proper rounding for casts of time amounts.

### DIFF
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -6478,16 +6478,16 @@ bool CGUIInfoManager::GetInt(int &value, int info, int contextWindow, const CGUI
             {
               const CEpgInfoTagPtr tag(GetEpgInfoTag());
               if (tag)
-                value = static_cast<int>(tag->ProgressPercentage());
+                value = MathUtils::round_int(tag->ProgressPercentage());
               else
-                value = static_cast<int>(g_application.GetPercentage());
+                value = MathUtils::round_int(g_application.GetPercentage());
               break;
             }
           case PLAYER_PROGRESS_CACHE:
-            value = (int)(g_application.GetCachePercentage());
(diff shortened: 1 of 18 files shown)


xbmc/GUIInfoManager.cpp
```diff
@@ -6478,16 +6478,16 @@ bool CGUIInfoManager::GetInt(int &value, int info, int contextWindow, const CGUI
             {
               const CEpgInfoTagPtr tag(GetEpgInfoTag());
               if (tag)
-                value = static_cast<int>(tag->ProgressPercentage());
+                value = MathUtils::round_int(tag->ProgressPercentage());
```
tamland: std::round?

garbear: didn't know this was added in c++11. will change to std::lround()

FernetMenta:
+1 looks good as far as I can tell

garbear:
Updated MathUtils::round_int() to std::roundl(). jenkins build this please

garbear:
oops, std::lround() instead of std::roundl(). jenkins build this please

garbear:
hm, android (exclusively) is failing with 'std::lround() not found'. Any ideas?

hudokkow:
I assume Android doesn't fully support C++11 so std::lround() is missing.
For instance, std::to_string() is missing, so is std::stoi(), etc.

garbear:
k, I'll change back to MathUtils::round_int() then

garbear:
jenkins build this please

ksooo:
> I assume Android doesn't fully support C++11 so std::lround() is missing.

 Yep, unfortunately. Stumbled myself over exactly this some time ago.

PR #9827, opened 2016-05-17
FFmpeg: Make usable from add-ons

notspiff:
This contains two alternative ways to enable the use of internal ffmpeg in add-ons, ref https://github.com/notspiff/screensaver.electricsheep
- either by always enabling PIC so the .a files can be linked into the .so file
- or by making sure the entirety of ffmpeg is included in the kodi binary, accompanied by lazy linking in the add-on. as-is lazy linking cannot work since the lnker will prune unused symbols, and if the add-on use other functionality than kodi core, the code won't be around (happens with electric sheep).

choose your poison. note autotools is lacking, i didn't want to do that before the preferred solution is agreed on.

### DIFF
--- a/tools/depends/target/ffmpeg/CMakeLists.txt
+++ b/tools/depends/target/ffmpeg/CMakeLists.txt
@@ -16,7 +16,7 @@ if(CROSSCOMPILING)
 endif()
 
 if(CORE_SYSTEM_NAME STREQUAL linux)
-  list(APPEND ffmpeg_conf --enable-vdpau --enable-vaapi)
+  list(APPEND ffmpeg_conf --enable-vdpau --enable-vaapi --enable-pic)
 elseif(CORE_SYSTEM_NAME STREQUAL android)
   if(CPU MATCHES arm)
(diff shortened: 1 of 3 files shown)


notspiff:
for the record my vote is on the PIC approach.

notspiff:
.. and the second approach does not work as-is, it needs some more work, was just to show the concept.

wsnipex:
+1 for PIC

fetzerch:
strong +1 for PIC

notspiff:
ping.

stefansaraev:
jenkins build this please

stefansaraev:
unrelated ios jenkins make -jXX race build failure, ref #9938

hudokkow:
http://jenkins.kodi.tv/job/IOS/7518/
:wink:

stefansaraev:
missed that, jenkins cycle wasted.

hudokkow:
Not wasted. I saw that it iOS failed and started a new job for it.

PR #9828, opened 2016-05-17
change labelenum setting type

notspiff:
earlier you got the translated label as value which is not very useful.
now you get the index of the chosen entry.


ronie:
if you want the index, you need to use the 'enum' setting type.

labelenum will return the chosen value or the translated label, depending on whether 'values' or 'lvalues' are used.

notspiff:
right you are.

PR #9829, opened 2016-05-17
Add-on helper: add foreign buildsystem support

notspiff:
Allow taking advantage of foreign buildsystems. Make the CMake target copy a binary built by the external project. This is used for libretro-based add-ons where the add-on binary is not specific to Kodi.

This has been sitting in retroplayer for a few years. I have extended it with support for a custom data directory. It's split so @garbear doesn't have to deal with a conflict.


project/cmake/scripts/common/addon-helpers.cmake
```diff
@@ -103,22 +114,46 @@ macro (build_addon target prefix libs)
...
-              COMPONENT ${target}-${${prefix}_VERSION})
+      if(${prefix}_SOURCES)
+        INSTALL(TARGETS ${target} DESTINATION ${target}
+                COMPONENT ${target}-${${prefix}_VERSION})
+      endif()
+      if (${prefix}_CUSTOM_BINARY)
+        list(GET ${prefix}_CUSTOM_BINARY 0 FROM_BINARY)
+        list(GET ${prefix}_CUSTOM_BINARY 1 TO_BINARY)
+        if(OS STREQUAL "android")
+          set(TO_BINARY "lib${TO_BINARY}")
```
wsnipex: although its not likely to happen at the moment, but what about non-shared libs?

notspiff: Not sure I understand. This is only to be used for the final addon binary
which will always be shared unless you plan to add linker steps to addon
loading in kodi.
17. mai 2016 15:14 skrev "Wolfgang Schupp" <notifications@github.com>:

wsnipex: right. As long as addons are libs this won't ever be an issue.
Not sure why, but I remember someone (maybe @AlwinEsch ) thinking/working on possibly making addons exes.
Might have imagined it :)
Anyway, not important here, sorry for the noise.

fetzerch:
+1 for the change, verifed couple of times that it works with libretro cores.
Just since you anyways touch those lines: mind changing the commands to lowercase? (if/else/install, ...)

hudokkow:
Don't bother changing to lowercase. I have a branch with that done.
Although not perfect, [this] (https://cmake.org/pipermail/cmake/2008-May/021752.html) helps. Rest is grep and s/old/new/

fetzerch:
Ok, otherwise good to go?

hudokkow:
:+1:
Build is fine. The usual suspects fail on Linux and peripheral.joystick didn't liked the includes change but I'm waiting on garbear's input.

Needs a rebase.

notspiff:
rebased.

hudokkow:
Thanks!

hudokkow:
Thanks!

PR #9830, opened 2016-05-17
fix 'unknown sources' setting visibility when opened from dialog

tamland:
alternative to #9821

### DIFF
--- a/xbmc/settings/windows/GUIWindowSettingsCategory.cpp
+++ b/xbmc/settings/windows/GUIWindowSettingsCategory.cpp
@@ -27,6 +27,7 @@
 #include "settings/DisplaySettings.h"
 #include "settings/Settings.h"
 #include "settings/lib/SettingSection.h"
+#include "settings/windows/GUIControlSettings.h"
 #include "utils/log.h"
 #include "view/ViewStateSettings.h"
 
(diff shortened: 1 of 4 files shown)


tamland:
@Montellese this ok with you?

xbmc/settings/windows/GUIWindowSettingsCategory.cpp
```diff
@@ -201,15 +202,32 @@ void CGUIWindowSettingsCategory::Save()
...
+      SET_CONTROL_FOCUS(CONTROL_SETTINGS_START_BUTTONS + i, 0);
+      return;
+    }
+    for (const auto& group: m_categories[i]->GetGroups())
+    {
+      for (const auto& setting : group->GetSettings())
+      {
+        if (setting->GetId() == elementId)
+        {
+          SET_CONTROL_FOCUS(CONTROL_SETTINGS_START_BUTTONS + i, 0);
```
Montellese: I'm pretty sure this should be `CONTROL_SETTINGS_START_CONTROL`

Montellese:
Not a fan of the indexed for loop but no simple way around it. Apart from that it looks good.

stefansaraev:
jenkins build this please

PR #9833, opened 2016-05-18
[cmake] Cosmetics: change commands to lowercase

hudokkow:
Cosmetics only and low priority.
Script used to automate it can be found [here](https://github.com/hudokkow/scripts/tree/master/cmake). Might be useful for others.

@fetzerch, @wsnipex for review


fetzerch:
Personally I'm quite in favor of strict rules. One thing that would additionally make sense:
`if("${OS}")` is simply `if(OS)`.

For the files that are only used in the main CMake build system, I'd say: go for it. (there's not much git  history anyway). For the binary addon buildsystem it still has my +1, even though I see a small disadvantage of `git blame` being a bit more difficult to use.

fetzerch:
One more thing might be to replace "if( " and " )" with "if(" and ")"?

hudokkow:
If no one objects, this is going in tomorrow.

PR #9834, opened 2016-05-18
[aml] add gui settings to enable/disable sd material hw decoding

stefansaraev:
alternative of #9772

cc @koying @codesnake


codesnake:
For me this looks like more elegant solution. Compiled and tried on several videos and it seems works without problems.

stefansaraev:
jenkins build this please

mazeyx:
these options are greyed out on android arm tv and have no effect, anyway this could be enabled id like to use software decode on sd content (would fix a zoom issue im having with media codec surface enabled)

stefansaraev:
@mazeyx those are aml specific, but I somehow forgot to make them conditional argh. will fix. thanks!

PR #9081, opened 2016-02-07
Fix limited songs query (for MySQL) and improve efficiency of GetSongs

DaveTBlake:
This is a follow on from #8993 that fixed a problem with AudioLibrary.GetSongs when using limits not returning the correct number of songs.  The SQL used produced an error in MySQL.  A quick fix was proposed in #9034 (that has been backported and works but is a little ugly). Testing this revealed that cuesheet related queries were being made per song to return replaygain, something often not required.

This changes the query syntax when limit on songs is applied to the join to this format
`SELECT sv.*, songartistview.* FROM (SELECT songview.* FROM songview WHERE  ..... LIMIT 1, 1000) AS sv LEFT JOIN songartistview on songartistview.idsong = sv.idsong`

A parameter is added to `CMusicDatabase::GetSongsFullWhere` to allow the attempted loading of song replaygain from cuesheets to be skipped. 

As replaygain is not returned by JSON API,  `AudioLibrary.GetSongs` is altered to take advantage of this parameter and avoid the unnecessary cuesheet processing. Similarly it also now uses the artistData parameter to only query song_artist table when some artist properties are wanted. Overall, only using a join when artist data is to be returned  by the API should speed up GetSongs when minimal data wanted. It is not so much the query speed that effects this, as the time taken transfering the full dataset results to CFileItem when they are subsequently not passed back through the API.

@phate89 I have gone with a variation of your suggestion. It seems faster to me than the alternative, and more reable, but I would appreciate you having a look at it.

@MilhouseVH some MySQL testing would be appreciated. Particularly want to convince myself that it works when we have a where clause and limit e.g. "filter":{"artistid":554}

@Montellese I have bumped API version because the implementation of GetSongs changes. Need to make sure my other API work merges correctly with this. And sorry but I have fixed some odd white spaces too.

PR #8993, opened 2016-01-30
Fix AudioLibrary.GetSongs missing songs when using limits

DaveTBlake:
Problems have been reported in recent builds with AudioLibrary.GetSongs when using limits to return songs in sections. The total number of songs returned not matching the number displayed in Kodi.

The problem was in `CMusicDatabase::GetSongsFullByWhere` introduced by #8306 to return accurate artist data for the songs by querying a join of songview and songartistview. This was not correctly applying the limit criteria (to just songs not the join), and needed to be a left join to cover those songs that do not have artists.

Thanks @Millencolin007 for doing the leg work and pinning this issue down so well for me to fix.

I think this is safe enough to be back ported to Jarvis, but would like some testing first obviously. It is a music library issue, but only shows up when using JSON to get songs from a full library.

PR #8306, opened 2015-10-27
[Fix]jsonrpc getting consistent song and album artist data - names, IDs and MusicbrainzIDs

DaveTBlake:
This fixes the issue that JSON may in some circumstances return inconsistent artist name and ID arrays for song and albums. It also correctly provides Musicbrainz IDs, that would never have been working properly before.

Since adding processing for ARTISTS tag #7486 it is possible that the splitting the artists string may not give the same artists as held in the song_artist or album_artist tables, so could lead to discrepencies. In particular `audioLibrary::GetAlbums` and `GetSongs` the array of artists for an album may not match the array of artistIds.  This was discussed under #8268 

It ensures that when JSON gets albums or songs the underlying query uses song_artist or album_artist tables to fully populate artist credits. Hence artist names and id arrays will be consistent. As a bonus now gets MBID (as a vector to match names and ids)

It bumps JSON version for now artist data returned is consistent.

This is a fix, not an improvement, but does make changes to underlying queries so could be considered major. I have deliberately avoided impacting the main music navigation SQL, but longer term we should be getting artist data from the song_artist table in more places.

Even if it does not make Jarvis alpha since it is JSON related @Montellese I am hoping you can have a look at this before you are away for a month, and at least approve the approach. It needs some more testing with live data, so I am hoping a team member will build it so this can happen.

@Tolriq see I did as I promised. Hope that you can test it with some Yatse data.
@Razzeee and @evilhamster have a look please since it is music.

PR #9836, opened 2016-05-19
[JSON]GetAlbums and GetSongs with limits returning too few records

DaveTBlake:
When some albums have multiple artists `AudioLibrary.GetAlbums` with limits, whether sorted or not, can result in too few albums being returned.

Similarly when some songs have multiple artists `AudioLibrary.GetSongs` with limits and a sort order, and when returning artist data (such as "artist", "artistid", or "musicbrainzartistid"), can result in too few songs being returned.

This issue was introduced in Jarvis when #8306 corrected the return of inconsistent artist data by querying with a join to the artist table. The limits issue was noticed but not fully resolved for GetSongs by subsequent fixes such as #8993 to #9081 etc., while the prolbems with limits in GetAlbums was overlooked.

@Tolriq I will get a build going if you would kindly test this out. @Razzeee for a sanity check perhaps.
@Montellese have bumped JSON minor rest is all `musicdatabase.cpp`


DaveTBlake:
@MilhouseVH given previous experience could you (or someone) possibly checkout the changed query in GetAlbums **with limit but not sort order** on a MySQL database?

MilhouseVH:
Seems to be working as expected...
```
08:45:09 T:140304061667072   DEBUG: GetAlbumsByWhere query: SELECT av.*, albumartistview.* FROM (SELECT albumview.* FROM albumview  WHERE albumview.strReleaseType = 'album' LIMIT 400) AS av JOIN albumartistview ON albumartistview.idalbum = av.idalbum
08:45:10 T:140304061667072   DEBUG: GetAlbumsByWhere - query took 187 ms
```

:+1:

DaveTBlake:
Thanks @MilhouseVH good to know

@Tolriq for testing a build is here http://mirrors.kodi.tv/test-builds/win32/KodiSetup-20160519-0753949-HEAD.exe

Tolriq:
Will test tomorrow. I hope performance is not too much degraded.

DaveTBlake:
>Will test tomorrow. I hope performance is not too much degraded.

Should be OK. Obviously requesting the 2000th song out of 32k of songs when sorted by artist name and including its artist data, is not going to be the quickest activity. But then it never has been. I suspect that most times when getting songs or albums in batches there is no sorting so the slowest case is avoided anyway.

I was a little surprized that the GetAlbums with limit issue had been overlooked, both by myself and in use. The broader problem was spotted in GetSongs, and somehow I didn't extend the thinking to albums.

Tolriq:
Well there's http://trac.kodi.tv/ticket/15735 and the same for albums for mysql (I asked user to report with logs but well they do not really care about bug being fixed)

So while I did refuse to add a sort in Yatse to fix this Kodi bug that would slow down sync for 99% of other users for no reason, there's a high chance that others do add a sort order to bypass this problem.

About the problem, it's a little more complicated it seems, as Yatse does have a smarter way to sync data, it seems requesting data in batch will still returns all the data, the problem is more that when requesting 20 albums only 15 could be returned needing to do more queries. So some that do not count correctly or handle the limits values could end up without having all the data.

DaveTBlake:
I have to admit that all of Kodi's queries without ORDER BY statements have bugged me, didn't know they lead to an actual bug on MySQL. Is it MySQL specific, as seems Ok in the testing I have done on SQLite? Bug reports are so easily overlooked on Trac, and too many open and ancient to browse.

Anyway it sounds like I need to test more of the batched/paginated approach (based on using LIMIT in SQL) to reading songs, albums and artists on MySQL.

Tolriq:
Yes all the report I have about that problem are MySQL only.

Tolriq:
Ok so tested from inside Yatse, always get the same number of albums (but was not affected by the problem). So no regression :)

No sorting = 4.7 secs, sorting on title or artist = 6.5 secs. For 4365 albums.

DaveTBlake:
Thanks @Tolriq, no regression is good.

Yatse obviously works around it, but I'm a bit bothered that no one else seems to have noticed that the albums weren't returning properly.  I don't know if someone can use some direct JSON calls onto say Jarvis and compare with this this patched version (same music loaded into both databases) just to repeat the issue and see the fix works (on SQLite at least).

With the issue of LIMIT clauses in MySQL seeming to need an ORDER BY statement to produce sequential/consistent results, I have to wonder if it is still a problem or has MySQL changed? Yet to set up MySQL and test myself.

DaveTBlake:
Rebased for JSON version bump.

I have tested the order issue with MySQL and limits  reported 16 months ago in  http://trac.kodi.tv/ticket/15735 and I can't reproduce it. Possibly MySQL version dependent? I see no reason to add an ORDER BY statement to the query, especially if it impacts speed.

The implication is that consumers fetching data in batches may be using a sort paramater as well as limit in the JSON call just to avoid the limit being applied in the SQL query (and thus any MySQL issue). When sorting too the limit is applied to the resulting dataset.  I can not see that as necessary with the lastest version of MySQL, but if they are then sorting will be more common used and this PR is even more important.

But I have been unable to find out if other JSON consumers use GetAlbums or GetSongs with a sort order or not, and still not had anyone confirm experiencing the problem that this PR fixes. Maybe everyone does something clever to ensure they get all the albums or songs?

Regardless this fix still needs to be applied.

Tolriq:
Well there's not a lot of people that follow dev and try to report / fix bugs :)

Let's see if that user do answer, it's the only one I had made accept to report problem with logs since he was able to reproduce and seemed tech compliant :)

DaveTBlake:
>Well there's not a lot of people that follow dev and try to report / fix bugs :)

True, lack of motivation for that all round, except us of course.
Meanwhile it seems reasonable to merge this PR, OK with you?

Tolriq:
Yep if you have tested with Mysql fix does not harm Sqlite. so should be merged.

DaveTBlake:
jenkins build this please.

PR #9838, opened 2016-05-19
[skinning-api] - remove "yes" / "no" as accepted values for booleans

phil65:
When workin on my skinning editor plugin, I noticed that we are quite inconsistent in terms of when we allow just "true" / "false" or also "yes" / "no" for certain node content / attributes.
I think @ronie  already tried to remove them once and @BigNoid  wasnt a big fan (cannot remember the reason), I will push again for this now though since it causes more confusion than being useful . :)

### DIFF
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -5026,9 +5026,9 @@ int CGUIInfoManager::TranslateSingleString(const std::string &strCondition, bool
   const Property &cat = info[0];
   if (info.size() == 1)
   { // single category
-    if (cat.name == "false" || cat.name == "no")
+    if (cat.name == "false")
       return SYSTEM_ALWAYS_FALSE;
-    else if (cat.name == "true" || cat.name == "yes")
+    else if (cat.name == "true")
       return SYSTEM_ALWAYS_TRUE;
 
     // deprecated begin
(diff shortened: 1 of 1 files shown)


BigNoid:
I dont think removing stuff like this has any benefit and I use it quite a lot myself.
But as I said in the other pr, i wont stand in the way if it must be removed.

phil65:
There is no real need for it, but imo things become easier to understand when they are consistent and when there are not several different ways to achieve the same. Some skinners might try "yes" / "no" for flipx for example, and there it wont work.

BigNoid:
you have a point there. Ofc an argument could be made to add yes/no to the other ones, but I wont go there :)

ronie:
for several tags, the accepted values are defined elsewhere:
https://github.com/xbmc/xbmc/blob/master/xbmc/utils/XMLUtils.cpp#L109-L124

ideally that needs to be cleaned up as well, as removing it in one place and not the other would lead to more confusion instead of less ;-)

phil65:
Thx for the heads-up, will add those.

phil65:
@Montellese Since XMLUtils::GetBoolean() is used all over the place, would that change be safe?

Montellese:
It might break NFOs and scrapers relying on these values being valid.

phil65:
Ok, that´s what I expected :) 
Will close this then, it was just a minor detail I noticed while writing some skin checks.

PR #9840, opened 2016-05-19
Reinstate the correct behaviour for 'menu' (single press)

joelittlejohn:
This behaviour was broken in 779c7c4c and an attempt was made to fix it in b702e6b4. The fix didn't reinstate the original button function though, causing the menu button to change the behaviour it has had for the last decade. This change should reinstate the correct behaviour.

### DIFF
--- a/system/keymaps/remote.xml
+++ b/system/keymaps/remote.xml
@@ -57,7 +57,7 @@
       <pageplus>PageUp</pageplus>
       <pageminus>PageDown</pageminus>
       <back>Back</back>
-      <menu>ContextMenu</menu>
+      <menu>PreviousMenu</menu>
       <menu mod="longpress">Menu</menu>
       <contentsmenu>PreviousMenu</contentsmenu>
       <rootmenu>PreviousMenu</rootmenu>


NedScott:
Are you sure that's the correct behavior? If the physical button is labeled "menu" then something like "back" or "previousmenu" doesn't seem logical, no matter how long it has been around.

joelittlejohn:
By way of a justification here:

* I guess it's being used to mean 'Main Menu', so I think it does make sense. 
* PreviousMenu is a fantastically useful command, ContextMenu really isn't.
* It's been this way for 10(?) years (not a good reason on it's own, but it helps)

stefansaraev:
-1 to changing default behaviour of "menu" buttons to previousmenu command. contextmenu is what it does on keyboard-like remotes. too.

joelittlejohn:
@stefansaraev Um, no, it was PreviusMenu before 779c7c4:

https://github.com/xbmc/xbmc/commit/779c7c4c710168a70cd8f25a0e1f380781f2cb7a#diff-e2d96b43a04d54676899256861d747a3L55

The behaviour on keyboard-like remotes seems irrelevant to me. Consistency across different remotes is less important than consistency of operation for remotes over time.

stefansaraev:
> @stefansaraev Um, no, it was PreviusMenu before 779c7c4:

not really.

https://github.com/xbmc/xbmc/blob/Jarvis/system/keymaps/keyboard.xml#L55
https://github.com/xbmc/xbmc/blob/Gotham/system/keymaps/keyboard.xml#L62

> The behaviour on keyboard-like remotes seems irrelevant to me. Consistency across different remotes is less important than consistency of operation for remotes over time.

I disagree, while you dont care about keyboard like remotes, I (and many more) think consistency is important. 

and a button should fire the action that makes most sense, in this case, PreviousMenu makes no snese, IMO.

those, who do not like it, can always remap it. default kodi mappings are not always to my lining and I have a not-so-small keyboard.xml override here.,

stefansaraev:
@NedScott @xhaggi up to you. what you think ?

NedScott:
What I really care about is how the actual button appears to the end-user, or a rough guess at what the average button is. Does it say "menu"? Most of the MCE remotes that I'm familiar with don't actually have a button that says "menu". Some have a "DVD menu" button, some have an "info" button that sometimes says "menu" on it as well, etc. It's a bit confusing to me.

What does this button look like to the end-user? I'll try to research this up myself, but there really isn't a pure standard. If you have specific examples then it would help me if you could link to them.

xhaggi:
The wetek remote hold a hamburger button which is mapped to KEY_MENU IIRC. My remote has a menu button.

NedScott:
@xhaggi none of the wetek remotes use remote.xml. They're all seen as "keyboard" type remotes, which is not changed in this PR.

stefansaraev:
I dont know how it looks on mce remotes. but in the past, we got countless of questions "how to .. context menu with mce remote?"

joelittlejohn:
Happy to close this if there is no appetite to stick with the old behaviour. This issue will appear on Google for anyone that starts looking and that's probably enough to put most people back on the right track.

NedScott:
I have no problem with this if someone would just tell me what the damn button looks like to the end user, as I asked here: https://github.com/xbmc/xbmc/pull/9840#issuecomment-222884561

Take a picture with your phone and show me. It shouldn't be this hard to communicate online.

joelittlejohn:
@nedscott Sorry, I don't actually have access to an mce remote as I use a Harmony One. The Harmony One has a button that says 'Menu'.

I originally started using XBMC with the Xbox DVD Remote kit (on the original Xbox). This remote (the original XBMC remote I suppose 😃) had a menu button to, and the command sent was PreviousMenu:

http://media.engadget.com/img/product/19/eyk/microsoft-xbox-dvd-remote-1fsp-800.jpg

joelittlejohn:
Probably two of the most popular mce designs are this:

http://www.windowsdevcenter.com/2006/11/28/graphics/fig-21.jpg

Where the button says 'DVD Menu', and this:

http://www.oneweb.co.uk/mythtv/wp-content/uploads/2009/02/mce-remote.jpg

Where the button is an icon representing 'DVD (Menu)'

joelittlejohn:
To me, the context menu is a totally different kind of thing to the DVD Menu. A 'DVD Menu' button is a way of stopping what you're doing and jumping straight back to the 'top' of the system. A 'context menu' a way of saying you want to drill down, and do something more specific that relates to your current task - it's the opposite in some ways.

NedScott:
The Xbox remote has always used menu for context menu, and all three examples have their own back buttons. Granted, previousmenu is a different command, it still doesn't seem like "DVD menu" should be doing that. 

I don't object to this, though. The button seems to be labeled "DVD menu" for most remotes, so previousmenu makes as much sense as context menu.

Razzeee:
I'm closing this, as the discussion here seems to have died off.

@joelittlejohn 
if you want to revisit this, just let us know and we will reopen the PR, thank you!

PR #9841, opened 2016-05-19
Fix API pkgconfig install issue

mackal:
When $(DESTDIR)$(libdir)/pkgconfig did not exist, the install command
would just rename the files.


mackal:
This shouldn't be needed with #9843 being merged. However, when ever the API changes do happen (if they do) they need to make sure this issue is resolved.

AlwinEsch:
Thanks @mackal I include your fix to my lib rework (was a bit to fast with the merge of them)!

PR #9843, opened 2016-05-20
Revert "[addon] binary add-on library rework for API level system"

AlwinEsch:
@FernetMenta @ksooo this revert commits and remove my bad merged pull request #9659

Sorry to all!


AlwinEsch:
jenkins build this please

AchimTuran:
>  bad merged pull request #9659
no one said that it was a bad pull request :-D

@FernetMenta @ksooo 
Is a full revert required? Why we're not going further from here?

It would be nice if we could discuss on how to handle so huge PRs like #9659 in future to prevent reverts or anything similiar.

ksooo:
I did not request the revert, although I'm convinced that it was wrong to merge it without a sign off of at least one Kodi developer with senior reputation.

ksooo:
Ah, and personally I'm far away from being Kodi senior developer. :-) 

Not saying that the PR is in bad shape. With my current Kodi code knowledge I cannot even judge on most of the things Alwin did here. It's just my feeling that it could have needed some more love/conceptions discussion/code nitpicking before merge.

AchimTuran:
> Ah, and personally I'm far away from being Kodi senior developer. :-)
Me too :-b

@AlwinEsch @FernetMenta @ksooo 
I did some quick fixes for Windows see https://github.com/xbmc/xbmc/pull/9844/

ksooo:
> I did some quick fixes for Windows

 you see what I ment with "could have needed some more love"?

AlwinEsch:
Has brought this with my much to fast merge from the other, I like to have my level 2 in but without the OK was a complete wrong way. There was a stupid minute inside me where I has clicked the "Merge pull request" button.

The Kodi team is for me the best on world! If my api2 stays in OK, but absolut also no problem for me to revert it and I bring it again as request to allow the right way.

hudokkow:
No need to apologize and don't feel bad about it.
Good news is, it works. Atleast on linux. I can't comment on most code, not enough code knowledge to do it but great stuff and great work.

fritsch:
@AlwinEsch All fine. No need to feel bad. Revert is a method of software development. And pushing a button too early won't kill anyone. Thanks for coming back with the revert. Please wait until tomorrow (Saturday / Sunday) when @FernetMenta is back home. So thumbs up and start feeling good again :-)

Paxxi:
You made a mistake, that happens. We don't want to revert just because, it'll be messy. Let's review as  is and save a revert only if we find something really broken or badly designed that can't be fixed going forward

AchimTuran:
Sorry guys I have to write this. It's great how the team handles this sitauation. I can remember some fights on Github.
It's great that the team spirit comes back. :-b

Paxxi:
So I decided to poke around some and do a quick review of the merged PR and so far it looks fine. 

- There's some design choices that I don't really agree with but so far I haven't found one that wouldn't be fixable through additional PR.
- There's a fair amount of code duplication because of the versioning but that should also be solvable without causing issues if we find a better solution (first thought would be looking at inline namespaces but that's another discussion)

FernetMenta:
Mistakes happen and this one is no big deal. Nobody blames anyone. Like @fritsch mentioned we have appropriate methods to correct mistakes. In this case it is a revert. 
Before merging such a big change that affects all binary addons we should discuss the new design and test a prototype.
Another issue is lost history. We need to discuss whether we really need to maintain backwards compatibility for the api. It really required it might be more appropriate to (git) move v1 to v2.

Thanks @AlwinEsch please go ahead with this revert.

AchimTuran:
@FernetMenta 
> Another issue is lost history. We need to discuss whether we really need to maintain backwards compatibility for the api. It really required it might be more appropriate to (git) move v1 to v2.

It would be nice if we could support more than one API level at the same time. Consequently the add-on developers would have more time to port their functionality to the newer API level and we could introduce new add-on features faster. In my view it is a step in the right direction.
So far I only see benefits with the drawback that maintaining will require more effort. But according to @AlwinEsch it is very easy with his implementation to introduce new API levels.

@AlwinEsch it would be nice if you can create a new branch in your repository and we can try to improve things there. Like @FernetMenta did with his VideoPlayer development.

@FernetMenta 
Do you have the time and the motivation in helping to improve @AlwinEsch's new binary add-on system?

FernetMenta:
> Do you have the time and the motivation in helping to improve @AlwinEsch's new binary add-on system?

Sure

> It would be nice if we could support more than one API level at the same time. Consequently the add-on developers would have more time to port their functionality to the newer API level and we could introduce new add-on features faster.

This is not trivial. In general APIs are versioned in regard to features but now we have a v1 and v2 with identical feature sets. What if we add a new feature to pvr or inputstream? Add v3, v4, v5? Or v2.1 v3.1?

As the maintainer of inputstream I don't want to support multiple APIs for this addon in the near future. Why should v17 be released with multiple versions of inputstream API? This extension point is going to be introduced with v17.

AchimTuran:
@FernetMenta 
Great looking forward to working with you.

> This is not trivial. In general APIs are versioned in regard to features but now we have a v1 and v2 with identical feature sets. What if we add a new feature to pvr or inputstream? Add v3, v4, v5? Or v2.1 v3.1?

I guess Alwin introduced v1 and v2 because v2 is a major change to give the devs the time to adapt to it. Furthermore we can support both old add-ons and new add-ons. Until now I can't say much about versioning add-on API's, because I'm not familiar enough with @AlwinEsch's implementation. Maybe we can talk about this at DevCon or create a meeting with all involved devs.

> As the maintainer of inputstream I don't want to support multiple APIs for this addon in the near future. Why should v17 be released with multiple versions of inputstream API? This extension point is going to be introduced with v17.

Sorry I can't answer this question. I hope @AlwinEsch could explain why he introduced two API's.

AlwinEsch:
Currently is the use of two levels (API1 and API2) to have time to check reworked add-ons without the force to add directly to all (which now a lot add-ons) like it was before on version changes.

After all add-ons are using Level 2 is the level 1 complete removed. Then it starts the way on the calls from Addon to Kodi defined with his API level function table, as example about changes for level 3 is it visible here: https://github.com/xbmc/xbmc/blob/master/xbmc/addons/binary/interfaces/api3/PVR/Addon_PVR.cpp#L39-L55

If now a new function is needed for this level is it only be added there and included in his function table which use also the already present functions from Level 2.

With this is for Addon to Kodi calls the amount of changes in Kodi very small.

With the calls from Kodi to add-on becomes is it a little more difficult, but also in a acceptable way.

Also in the future is my main goal of them to allow changes on add-on interfaces without the force to bring all add-ons to this changes. As example, active in release is level 2, during development becomes then the changes added to level 3. All add-ons are still usable with level 2 and them with 3 can be reworked until the next release. And if the developer of a add-on has no time to change them on 3 are no problems on next release present.

After a time of 2-4 Release can then the old level e.g. 2 complete removed. If we reach a amount of over 100 add-ons is this the for me best way. Otherwise comes a lot of time on new releases the problem with not working add-ons.

AlwinEsch:
I make on the DevCon a detailed description about this system. With this lot of code is it a bit more easy to explain with words and to give answers to questions direct.

Problematic becomes it after the Kodi to add-on interface becomes changes, there are changes on all add-on functions needed, this can be done as smaller requests to allow usage of old style together with new style.

As example for the complete use of the new level system is visible here for pvr demo: https://github.com/AlwinEsch/pvr.demo/commit/a213319e9c267ced38cccacecce25e9c51235bde

This use a function table defined on the add-on library headers with https://github.com/AlwinEsch/kodi/blob/binary-addon-level3/xbmc/addons/kodi-addon-dev-kit/include/kodi/api3/pvr/.internal/internalHead.hpp#L37-L107 there is for everything this new `void *hdl,` needed to identify the class on Kodi. Also on goal with this `hdl` is in future to create more as one class on add-on to create several connections.

PR #9845, opened 2016-05-20
[cmake] Reuse version.txt and add-on API parsing across buildsystems

hudokkow:
Alternative to https://github.com/xbmc/xbmc/pull/9742

Do we want Macros.cmake copied to APP_LIB_DIR?
@fetzerch, @wsnipex

### DIFF
--- a/project/cmake/scripts/common/Macros.cmake
+++ b/project/cmake/scripts/common/Macros.cmake
@@ -238,7 +238,7 @@ function(core_file_read_filtered result filepattern)
     file(STRINGS ${filename} fstrings REGEX "^[^#//]")
     foreach(fstring ${fstrings})
       string(REGEX REPLACE "^(.*)#(.*)" "\\1" fstring ${fstring})
-      string(REGEX REPLACE "//.*" "" fstring ${fstring})
+      string(REGEX REPLACE "[ \n\r\t]//.*" "" fstring ${fstring})
       string(STRIP ${fstring} fstring)
       list(APPEND filename_strings ${fstring})
     endforeach()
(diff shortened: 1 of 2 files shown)


wsnipex:
Nice, I prefer this one.

hudokkow:
Yeah, a lot less shuffling around.

project/cmake/scripts/common/Macros.cmake
```diff
@@ -367,13 +367,32 @@ function(core_find_git_rev)
...
+#   APP_ADDON_API - the addon API version in the form of 16.9.702
+#   FILE_VERSION - file version in the form of 16,9,702, Windows only
+#
+# The following variables are set from libKODI_guilib.h:
+#   guilib_version - current ADDONGUI API version
+#   guilib_version_min - minimal ADDONGUI API version
 macro(core_find_versions)
   include(CMakeParseArguments)
   core_file_read_filtered(version_list ${CORE_SOURCE_DIR}/version.txt)
   string(REPLACE " " ";" version_list "${version_list}")
```
fetzerch: unrelated, but since you're at it: since the macro will put all vars in the parent scope, mind using unset(version_list) as that one is local. 

(hope these kind of requests are ok, but it's probably less effort to include in that commit instead of starting a new one)

Nice work with the docs, btw!

hudokkow:
@fetzerch, done.
Requests are fine, of course! We're here to improve stuff, correct? :wink:

hudokkow:
what? jenkins build this with addons please

hudokkow:
Just noticed cmake_parse_arguments doesn't like URLs.
Parsing <code>http://kodi.tv </code> results in the variable being set to <code>http:</code> only.

Any ideas? cmake_parse_arguments documentation is scarce and unhelpful.

fetzerch:
@hudokkow: it's caused by this: https://github.com/xbmc/xbmc/blob/master/project/cmake/scripts/common/Macros.cmake#L241 It could probably be fixed by prepending a space the the regexp (or `[ \t]` (not sure if cmake understands `\s`)).

hudokkow:
Completely missed that. Thanks.
Corrected. Unfortunately CMake's regex support is a bit lacking.

hudokkow:

hudokkow:
Build is fine. IOS is acting up.

PR #9850, opened 2016-05-21
configure.ac: do not check for libpng

stefansaraev:
libpng is no direct dependency for kodi anymore, no
point linking against it.

cc @wsnipex


stefansaraev:
jenkins build this please

bkuhls:
wfm, thx

PR #9851, opened 2016-05-21
optimize png images losslessly using zopflipng

PeterDaveHello:
Using Google's zopflipng to recompress the png images losslessly.


stefansaraev:
just fyi, this will have no effect on generated Textures.xbt size.

cc @phi2039 @ronie

ronie:
> just fyi, this will have no effect on generated Textures.xbt size.

just wondering, what's the benefit then?

stefansaraev:
perhaps @PeterDaveHello knows more than me

I dont know his zopflipng used to optimize those pngs, but in general, when no texturepacker is used, pngs will be ~ 10-15% smaller. no idea how much this will be faster (or how less memory is used) at runtime.

when using texturepacker (which we, and most distros do by default) there is no benefit, at all.

PeterDaveHello:
hmmm ... actually I don't know about texturepacker, so I can't not comment about it, would you guys mind to explain about using texturepacker caused no difference? 

If I need to talk a bout other benefits, for a huge repo like xbmc has about 40k commits, to clone it, I'll like to use shallow clone instead of cloning the whole repo history, so, with shallow clone, the speed of cloning can be faster, the disk space usage will be smaller.

stefansaraev:
http://kodi.wiki/view/TexturePacker

PeterDaveHello:
hmmm ... I wonder if that will really make no any difference before and after this PR, will somebody like to give it a try?

stefansaraev:
> will somebody like to give it a try?

sure, I did
before this pr: Textures.xbt size: 16694174 
after this pr: Textures.xbt size: 16698563

when doing loseless compression, there is no reason to see a (noticeable) difference. it's how texturepacker works.

PeterDaveHello:
Thanks @stefansaraev, should I close this PR now?

stefansaraev:
I dont think it will be merged unless the people I pinged here say "go ahead" :)

PeterDaveHello:
Okay, close first.

PR #9853, opened 2016-05-21
[webinterface.default] tvshows: default to poster view and use sorttitle

chewitt:
The TVShows view has a "banner - poster - landscape" toggle at the top of the page. The toggle does not retrieve different images it simply resizes the same poster image to different sizes. Removing the toggle completely is beyond my code fu so this PR simply sets the default to poster so the page displays correctly. It also sorts using sorttitle (same as the movie page) else tvshows appear in the unsorted order they were added to the library which makes no sense.

### DIFF
--- a/addons/webinterface.default/js/MediaLibrary.js
+++ b/addons/webinterface.default/js/MediaLibrary.js
@@ -867,7 +867,11 @@ MediaLibrary.prototype = {
             'studio',
             'mpaa',
             'premiered'
-          ]
+          ],
+          'sort': {
+            'method': 'sorttitle',
+            'ignorearticle': true
+          }
         },
         'success': function (data) {
           if (data && data.result && data.result.tvshows) {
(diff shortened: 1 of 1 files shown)


Montellese:
Can you split the artwork change and the sorting change into two commits? Personally I use banners for tvshows as well but I'm not sure if there's a "universal" standard artwork for tvshows.

What would be really nice is if we would retrieve all available artwork (which is supported by the JSON-RPC API) and then show the proper artwork depending on the chosen artwork type.

chewitt:
@Montellese - I've split the PR into two commits as requested.

stefansaraev:
good to go ?

Montellese:
Fine with me.

stefansaraev:
@chewitt I just tested this (with latest chrome and firefox). but default view is still banner. TVView cookie is correctly set but after refresh it defaults back to banner view. (seems click is not triggered on a toggle somewhere, js is sorta black magic to me)
not sure if it's my browsers fault

chewitt:
It's not the browser's fault as it does the same here (but didn't when I originally tested the tweak). I guess there's an underlying issue in the JavaScript as "remember view preference" has never worked (or not that I can remember - this has bugged me for a couple of years). I tried to remove the toggle entirely so tvshows matched the movies view but I couldn't figure it out. The sort tweak is fine so I can either drop the banner change or you can regard it as no less broken than it already is and merge both. Let me know what you'd prefer? :)

PR #9860, opened 2016-05-22
CppCheck performance fixes

tobbi:
This uses prefix operators instead of postfix operators where appropriate. Also: Pass by ref instead of value where appropriate.

### DIFF
--- a/xbmc/FileItem.cpp
+++ b/xbmc/FileItem.cpp
@@ -675,7 +675,7 @@ void CFileItem::ToSortable(SortItem &sortable, Field field) const
 void CFileItem::ToSortable(SortItem &sortable, const Fields &fields) const
 {
   Fields::const_iterator it;
-  for (it = fields.begin(); it != fields.end(); it++)
+  for (it = fields.begin(); it != fields.end(); ++it)
     ToSortable(sortable, *it);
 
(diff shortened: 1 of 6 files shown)


xbmc/FileItem.cpp
```diff
@@ -675,7 +675,7 @@ void CFileItem::ToSortable(SortItem &sortable, Field field) const
 void CFileItem::ToSortable(SortItem &sortable, const Fields &fields) const
 {
   Fields::const_iterator it;
-  for (it = fields.begin(); it != fields.end(); it++)
+  for (it = fields.begin(); it != fields.end(); ++it)
```
FernetMenta: we switched to newer style of loops: 
for (auto &field : fields)

tobbi: I assume I need to change all of the iterators?

FernetMenta: That would be really nice but we can't demand this. I would accept as is because it is already an improvement.

MartijnKaijser:
@tobbi could you please split off the whitespace fixes?

tobbi:
I can try, though this is something that Atom, the editor I am using, did automatically.

MartijnKaijser:
@tobbi maybe you can update if you have some time?

PR #9862, opened 2016-05-22
PluginDirectory - avoid BusyDialog if not called by main thread

FernetMenta:
follow up from https://github.com/xbmc/xbmc/pull/9778

### DIFF
--- a/xbmc/filesystem/PluginDirectory.cpp
+++ b/xbmc/filesystem/PluginDirectory.cpp
@@ -18,7 +18,7 @@
  *
  */
 
-
+#include "Application.h"
 #include "threads/SystemClock.h"
 #include "system.h"
 #include "PluginDirectory.h"
@@ -476,14 +476,28 @@ bool CPluginDirectory::WaitOnScriptResult(const std::string &scriptPath, int scr
 {
   bool cancelled = false;
 
-  if (!m_fetchComplete.WaitMSec(20))
+  // CPluginDirectory::GetDirectory can be called from the main and other threads.
+  // If called form the main thread, we need to bring up the BusyDialog in order to
+  // keep the render loop alive
+  if (g_application.IsCurrentThread())
   {
-    CScriptObserver scriptObs(scriptId, m_fetchComplete);
-    if (!CGUIDialogBusy::WaitOnEvent(m_fetchComplete, 200))
+    if (!m_fetchComplete.WaitMSec(20))
+    {
+      CScriptObserver scriptObs(scriptId, m_fetchComplete);
+      if (!CGUIDialogBusy::WaitOnEvent(m_fetchComplete, 200))
+      {
+        cancelled = true;
+      }
+      scriptObs.Abort();
+    }
+  }
+  else
+  {
+    // kill the script if it does not return within 30 seconds
+    if (!m_fetchComplete.WaitMSec(30000))
     {
       cancelled = true;
     }
-    scriptObs.Abort();
   }
 
   if (cancelled)


Montellese:
Ideally there shouldn't be any GUI related code in the VFS part (no dialogs or display related stuff like the "All albums" nodes etc.). That logic should be part of `CGUIMediaWindow` or any of its derived classes. But obviously that's a major task and out of scope for this fix.

xbmc/filesystem/PluginDirectory.cpp
```diff
@@ -476,14 +476,28 @@ bool CPluginDirectory::WaitOnScriptResult(const std::string &scriptPath, int scr
...
+      }
+      scriptObs.Abort();
+    }
+  }
+  else
   {
-    CScriptObserver scriptObs(scriptId, m_fetchComplete);
-    if (!CGUIDialogBusy::WaitOnEvent(m_fetchComplete, 200))
+    // kill the script if it does not return within 10 seconds
+    if (!m_fetchComplete.WaitMSec(10000))
```
Montellese: Where are the 10 seconds coming from? That might be too little on a raspberry pi or for something that retrieves data from the internet.

FernetMenta: arbitrarily chosen. what value do you suggest?

Montellese: I don't know but it just stood out as a "magic number" which is why I asked. I'm no expert on plugin response times.

FernetMenta: IMO no synchronous call should take more than a couple of seconds to complete.

Tolriq: Yatse use JSON to browse VFS from addons and timeout was at 15 second at start and there's was thousand of complains, it's now 30 sec and I still receive regularly complains. But IMO more than 30 sec is not user friendly, addon should manage data differently and request less data / use chunks.

10 seconds sounds really too short for most of the large addons like Youtube, Internet is slow by nature.

FernetMenta: Thanks @Tolriq I will change it to 30 then

popcornmix: Will we get an obvious log message when the timeout is hit?
Seems for debugging it would be useful to distinguish timeouts (that may have eventually succeeded) from a clear error response which never would have succeeded.

FernetMenta: > Will we get an obvious log message when the timeout is hit?

yes, the script will be killed and this will appear in the log

FernetMenta:
updated
jenkins build this please

FernetMenta:
Android build failure unrelated

BmdOnline:
Hi,
like I said in kodi forum (http://forum.kodi.tv/showthread.php?tid=269814&pid=2345031#pid2345031), some plugins aren't working since Milhouse 0522 build.

For example, with Youtube 5.2.1 plugin, I have this message in kodi.log
> 23:09:36 12917.675781 T:1962225664   ERROR: GetDirectory - Error getting plugin://plugin.video.youtube/channel/UCMkOKU7usWmqcpLDcWG9PtQ/
23:09:36 12917.719727 T:1962225664   ERROR: CGUIMediaWindow::GetDirectory(plugin://plugin.video.youtube/channel/UCMkOKU7usWmqcpLDcWG9PtQ/) failed

And, with SuprFavourites 1.0.43 plugin, I have same kind of error :
> 23:12:00 13061.955078 T:1962225664   ERROR: GetDirectory - Error getting plugin://plugin.program.super.favourites/?fanart=%2Fstorage%2F.kodi%2Faddons%2Fplugin.program.super.favourites%2Ffanart.j​pg&image=DefaultFolder.png&label=Explore%20Kodi%20favourites&mode=300
23:12:00 13061.968750 T:1962225664   ERROR: CGUIMediaWindow::GetDirectory(plugin://plugin.program.super.favourites/?fanart=%2Fstorage%2F.kodi%2Faddons%2Fplugin.program.super.favourites%2Ffanart.j​pg&image=DefaultFolder.png&label=Explore%20Kodi%20favourites&mode=300) failed

Everything is OK with 0421 build.

PR #9863, opened 2016-05-22
make libxml2 / libxslt optional

stefansaraev:
support for xslt scrapers was added in #3758 more than 2 years agom but if I am not mistaken, atm there is no working xslt scraper in kodi repo (I installed all and they all use regexps), there was an attempt to make one (http://forum.kodi.tv/showthread.php?tid=188738) but it has gone nowhere.

this PR makes libxml/libxslt optional, for use in distributions that prefer less dependencies. it is enabled (and required) by default, so no change to our official builds.

@fetzerch I am not sure I did cmake right..

comments ?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -399,6 +399,12 @@ AC_ARG_ENABLE([libbluray],
   [use_libbluray=$enableval],
   [use_libbluray=auto])
 
+AC_ARG_ENABLE([libxslt],
+  [AS_HELP_STRING([--disable-libxslt],
+  [enable libxslt (XSLT scrapers) support (default is enabled)])],
+  [use_libxslt=$enableval],
+  [use_libxslt=yes])
+
 AC_ARG_ENABLE([texturepacker],
   [AS_HELP_STRING([--enable-texturepacker],
   [enable texturepacker support (default is auto)])],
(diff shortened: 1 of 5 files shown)


wsnipex:
you have to add XSLT to optional_deps for cmake

part from that it looks good

Razzeee:
Well it's not correct that it's gone nowhere. The official test version of that scraper is far superior to our tvdb scraper at the moment.

bkuhls:
wfm, thx

phate89:
@Razzeee  sure but tvdb just switched to json api that are the official ones and will fully phase out xml:
https://forums.thetvdb.com/viewtopic.php?f=3&t=29643

jjd-uk:
I thought the Music scrapers use XSLT which is why it was introduced when @night199uk redid the music database for the introduction of Musicbrainz ID support.

Razzeee:
@phate89 I am not sure but I remember @scudlee saying he looked into them, so maybe he already uses that?

ScudLee:
XSLT is currently in use in my AniDb [Mod] scrapers - specifically in [parsing the mapping lists](https://github.com/ScudLee/scudlee-xbmc-addons/commit/8cab763908d11d16f3ee02ad30a7d79dfd8a9249) and in [converting the AniDb tags into genres](https://github.com/ScudLee/scudlee-xbmc-addons/commit/e60d9c8b4a767a4754d1eb5eb781b95ef75de8e5), both of which would be impossible using just regexes alone.  They're not in the official repository, so it would be easy to miss.

As for the change to thetvdb api, like I said in the [testing thread](http://forum.kodi.tv/showthread.php?tid=259457), I've already (mostly) written a version with the same features that uses the new api, by using regexes to convert the JSON input to an intermediate xml format and then using essentially the same xslt to generate the final output to Kodi.

I've just updated it from using the beta-api (and fixed a couple of broken things), so here's a download for that version to test: [metadata.tvdb.com.json](https://mega.nz/#!p9BjEL5B!-PaNbXrlyzJEEDPrC8--jzK0KiVUtje5Os49BcYEncw)

It's still very rough and the non-english language handling is almost certainly broke, but you should get the idea.

(I haven't been very active lately because real life has been a little...)

phate89:
@scudlee beta api are now official api :)

ScudLee:
That's what I meant when I said I updated it.  The version I originally wrote used the beta api URL.

zag2me:
Just for background, Musicbrainz is also going towards JSON API in their next web service release

stefansaraev:
please excuse my ignorance, but how exactly parsing json via regex to convert to xml is good idea ?

fritsch:
Python has dictionaries, no single need to construct an intermediate xml.
Cause json can be mapped 1:1 and then easily processed.

2016-05-23 11:45 GMT+02:00 Stefan Saraev <notifications@github.com>:

> please excuse my ignorance, but how exactly parsing json via regex to
> convert to xml is good idea ?
>
> —
> <https://github.com/xbmc/xbmc/pull/9863#issuecomment-220934964>
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

ScudLee:
If only we had Python scrapers.

Or do we?  I've not been around much lately.

fritsch:
Yeah :-) in deed. Though producing intermediate xml from json just to use
xslt is wrong.

2016-05-23 11:59 GMT+02:00 ScudLee <notifications@github.com>:

> If only we had Python scrapers.
>
> Or do we? I've not been around much lately.
>
> —
> You are receiving this because you commented.
> <https://github.com/xbmc/xbmc/pull/9863#issuecomment-220938024>
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

ScudLee:
Since the point of the scraper is to convert the JSON input to XML anyway, it's not that much of a stretch to add a middle step so the XSLT can do the heavy lifting the regexes can't handle.  

Especially when that's the only tool available.

stefansaraev:
can we agree to merge this, for those embedded people who prefer less dependencies? this is no way changes a behaviour for official builds / regular distributors.

Paxxi:
Using xslt + json +xml is a mess but we don't provide anything better so we really should fix that, should be possible to fix something better for json in the short term and possibly go the python route in the long-term

stefansaraev:
that was not the point, @Paxxi 

jenkins build this please

Paxxi:
@stefansaraev I know :) just wanted to defend @ScudLee merging is fine by me

stefansaraev:
jenkins build this please again

PR #9864, opened 2016-05-22
[tests] Fix some of the issues with the test suite on win32

Paxxi:
Delay loading wasn't properly initialized during the tests
Changed how rarmanager global is handled to get rid of a crash during shutdown because of ordering.
Still crashes during shutdown but at least now all the tests run fine

### DIFF
--- a/xbmc/SystemGlobals.cpp
+++ b/xbmc/SystemGlobals.cpp
@@ -70,9 +70,6 @@ std::map<std::string, std::string> CSpecialProtocol::m_pathMap;
   CRBP               g_RBP;
 #endif
 
-#ifdef HAS_FILESYSTEM_RAR
-  CRarManager g_RarManager;
-#endif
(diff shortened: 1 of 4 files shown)


Paxxi:
@Razzeee review please

Razzeee:
Doesn't seem to work as the test result xml is not written? 
Also seems to break Linux :/
http://jenkins.kodi.tv/job/WIN-32/8937/console

Paxxi:
Probably crashes before the xml is written, try locally and see, should give you the test results before crashing

Razzeee:
Yep, seems to work.
Maybe this helps to track down the rest:
image

Paxxi:
Yeah I get the same, will continue to look into it but not sure when the next time will be so I decided to pr the results so far as it improves the current brokenness

Razzeee:
jenkins build and merge

PR #9866, opened 2016-05-23
Changing android ndk directory existence test

vadikml:
Starting ndk 11 ndk_dir/RELEASE.TXT was removed.
The new file to test is ndk_dir/source.properties according to a proposal at
http://forum.kodi.tv/showthread.php?tid=274621&pid=2337598#pid2337598

### DIFF
--- a/tools/depends/configure.ac
+++ b/tools/depends/configure.ac
@@ -456,7 +456,7 @@ if test "$platform_os" == "android"; then
     AC_MSG_ERROR("NDK path is required for android")
   fi
 
-  if ! test -f "$use_ndk/RELEASE.TXT" ; then
+  if ! test -f "$use_ndk/source.properties" ; then
     AC_MSG_ERROR("$use_ndk is not an NDK directory")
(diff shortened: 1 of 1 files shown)


Razzeee:
@koying

koying:
You must test both, for those having an older ndk

vadikml:
Yes, you are perfectly right.
I guess a simple 'or' between the older and newer implementation is enough
right ?

Or maybe something more complex like if ndk < 11 check for README.TXT if
ndk>=11 check for source.properties is needed ?

Vadim.

koying:
Nah. "or" will do. That's just a simple sanity check

vadikml:
adding "or" check for pre and post 11 ndk version and re-submitting.

koying:
You don't need to close the PR.
You can amend your branch with the updated code, then do "force push" of it and the PR will update by itself.

vadikml:
got it.
tnx.
in progress.

PR #9867, opened 2016-05-23
VideoPlayer: fix audio for wmapro (and may some other codecs)

FernetMenta:
see title

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDMessageQueue.cpp
+++ b/xbmc/cores/VideoPlayer/DVDMessageQueue.cpp
@@ -56,9 +56,13 @@ void CDVDMessageQueue::Flush(CDVDMsg::Message type)
 {
   CSingleLock lock(m_section);
 
-  m_list.remove_if([type](const DVDMessageListItem &item){
-      return type == CDVDMsg::NONE || item.message->IsType(type);
-    });
+  m_messages.remove_if([type](const DVDMessageListItem &item){
+    return type == CDVDMsg::NONE || item.message->IsType(type);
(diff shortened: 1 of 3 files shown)


MrMC:
startup maybe fixed, but seeking is even worse than before.

FernetMenta:
it's working now
jenkins build this please

MrMC:
much better now, thxs.

xbmc/cores/VideoPlayer/DVDMessageQueue.cpp
```diff
@@ -150,9 +169,11 @@ MsgQueueReturnCode CDVDMessageQueue::Get(CDVDMsg** pMsg, unsigned int iTimeoutIn
 
   while (!m_bAbortRequest)
   {
-    if (!m_list.empty() && m_list.back().priority >= priority)
+    std::list<DVDMessageListItem> &msgs = (priority > 0 || !m_prioMessages.empty()) ? m_prioMessages : m_messages;
```
popcornmix: This looks a bit strange to me. Using ``&&`` rather than ``||`` here would make more sense to me.
I may well be missing something, so feel free to ignore if you are happy with it.

MrMC:
passes my test suite of wmapro files.

xbmc/cores/VideoPlayer/DVDMessageQueue.cpp
FernetMenta: prio messages are always considered first, even it player requests a message with prio 0

PR #9871, opened 2016-05-23
[PVR] Remove CPVRTimerInfoTag::Has[Series]EpgInfoTag().

ksooo:
No longer needed. Implementation was broken (no lazy eval) and has/get pattern does not work in multithreaded environment.

Broken implementation (regression) led to PVR timer window missing EPG information for the epg-based timers.

@Jalle19 mind taking a look.


ksooo:
jenkins build this please

Jalle19:
Looks fine at a glance

PR #9876, opened 2016-05-24
[cmake] Make CMake aware of textfiles that configure the build

fetzerch:
The CMake based buildsystem is based on reading the input of textfiles at configure-time. As CMake does not track files that are just accessed using "file(READ)" building does not automatically retrigger a reconfiguration when those files are changes. This commit fixes it with the [CMAKE_CONFIGURE_DEPENDS](https://cmake.org/cmake/help/latest/prop_dir/CMAKE_CONFIGURE_DEPENDS.html) variable.

This should remove the necessity to manually rerun CMake when files are changed during development. (Due to globbing in some places, CMake still has to be rerun when textfiles are added or removed!)

pings: @hudokkow, @wsnipex, @notspiff
Not sure if I covered all cases though. Please also let me know if you've seen other cases, where a manual reconfigure was necessary.


akva2:
Lets kill the glob for text files

fetzerch:
Yes, definitively!

PR #9878, opened 2016-05-25
[cmake] get rid of ManageString.cmake

hudokkow:
Only one of the provided macros is used (STRING_SPLIT) and only
in one place and we can achieve the same with a simple string(REPLACE ...

@wsnipex, @fetzerch


fetzerch:
Very much +1, I've seen this already some time ago, but was busy with something else.

hudokkow:
Yeah, I stumbled into it looking at something completely different.

PR #9879, opened 2016-05-25
[PVR] add recording id to recording path

Glenn-1990:
The recording path should be unique.

This fixes the following issue when using a pvr client with demuxer (tested pvr.hts)
1) start recording show A
2) abort recording show A
3) resume recording show A
4) show A is shown 2 times in the recordings view, but only 1 in playable...

There is a small regression as bookmarks aren't migrated (only when using a pvr client with internal demuxer and no backend support for played position). Migration isn't easy without introducing a hack into the code... The path was changed in the past without migrating the bookmarks, so I don't think this is a big deal?

@ksooo @Jalle19

### DIFF
--- a/xbmc/pvr/recordings/PVRRecording.cpp
+++ b/xbmc/pvr/recordings/PVRRecording.cpp
@@ -448,7 +448,7 @@ void CPVRRecording::UpdatePath(void)
   else
   {
     m_strFileNameAndPath = CPVRRecordingsPath(
-      m_bIsDeleted, m_bRadio, m_strDirectory, m_strTitle, m_iSeason, m_iEpisode, m_iYear, m_strShowTitle, m_strChannelName, m_recordingTime);
+      m_bIsDeleted, m_bRadio, m_strDirectory, m_strTitle, m_iSeason, m_iEpisode, m_iYear, m_strShowTitle, m_strChannelName, m_recordingTime, m_strRecordingId);
   }
 }
 
--- a/xbmc/pvr/recordings/PVRRecordingsPath.cpp
+++ b/xbmc/pvr/recordings/PVRRecordingsPath.cpp
@@ -78,7 +78,7 @@ CPVRRecordingsPath::CPVRRecordingsPath(bool bDeleted, bool bRadio,
                        const std::string &strDirectory, const std::string &strTitle,
                        int iSeason, int iEpisode, int iYear,
                        const std::string &strSubtitle, const std::string &strChannelName,
-                       const CDateTime &recordingTime)
+                       const CDateTime &recordingTime, const std::string &strId)
 : m_bValid(true),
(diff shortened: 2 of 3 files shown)


ksooo:
Is there really no clean way to handle old bookmarks? I'm not talking about migrating those, but about handling them correctly at runtime. What's the problem with this? There must be some code that picks the old uri from the bookmark and dispatched it to pvr, right? Why can't pvr support reading both old and new recording uris?

Glenn-1990:
@ksooo Something like this could work, but I would call it a hack.
`    if (!supportsLastPlayed)
    {
      if (!db.GetResumeBookMark(m_strFileNameAndPath, m_resumePoint))
        db.GetResumeBookMark(m_strFileNameAndPathOld, m_resumePoint); // check for old path URl
    }`

ksooo:
@Glenn-1990 yep, thats ugly. Let's go without migration. We will see what will happen. ;-)

ksooo:
jenkins build this please

Jalle19:
We're still in pre-alpha so there's time to fix potential user complaints later on.

ksooo:
Yep. That's what I ment with "we will see what will happen."

Jalle19:
Yes I know, I was just agreeing with you :-)

ksooo:
failures of Travis and AppVeyor are unrelated

ksooo:
@Glenn-1990 Almost overlooked this: your change breaks <code>CPVRRecordingPath::GetTitle()</code>. You need to fix the regexp here: https://github.com/Glenn-1990/xbmc/blob/30c1d64607eb3ca15dbbda13cbe656abe587474e/xbmc/pvr/recordings/PVRRecordingsPath.cpp#L151

Glenn-1990:
@ksooo Updated the regexp.
Looking to the code, I guess that the regexp will not work when episode info is present, not related to this PR of course.. 

The regexp wasn't changed with the season info PR:
https://github.com/xbmc/xbmc/commit/57d27388802d3b632e3fbfc142ec12158164dd77

ksooo:
> I guess that the regexp will not work when episode info is present

 I think it is okay as it is now. Epsode stuff is contained in the 'path' part of the URI. Directory, title, episode, year, subtitle are simply concatenated and as long as the resulting string has at least one character (.*) should match.

ksooo:
thx @Glenn-1990

metaron-uk:
Well pvr.mythtv is one of those 'broken' clients.
I guess it means bookmarks will be reset by this PR, but I don't see this as a major issue myself. 
There was a bug in bookmarks caused by pvr.mythtv / mythbackend until very recently anyway introduced at the same as the season info PR and this has only just been reported / fixed (​https://github.com/janbar/pvr.mythtv/issues/37)

PR #9881, opened 2016-05-27
VideoPlayer: fixes

FernetMenta:
VideoPlayer fixes an improvements

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp
@@ -1343,17 +1343,19 @@ void CRenderManager::PrepareNextRender()
     ++iter;
     while (iter != m_queued.end())
     {
-      // the slot for rendering in time is [pts .. (pts + frametime)]
+      // the slot for rendering in time is [pts .. (pts +  x * frametime)]
       // renderer/drivers have internal queues, being slightliy late here does not mean that
-      // we are really late. If we don't recover here, player will take action
-      if (renderPts < m_Queue[*iter].pts + 0.98 * frametime)
+      // we are really late. The likelihood that we recover decreases the greater m_lateframes
+      // get. Skipping a frame is easier than having decoder dropping one (lateframes > 10)
+      double x = (m_lateframes <= 6) ? 0.98 : 0;
+      if (renderPts < m_Queue[*iter].pts + x * frametime)
         break;
       idx = *iter;
       ++iter;
     }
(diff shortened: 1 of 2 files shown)


FernetMenta:
jenkins build this please

xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp
```diff
@@ -1398,7 +1398,14 @@ bool CRenderManager::GetStats(int &lateframes, double &pts, int &queued, int &di
 
 void CRenderManager::CheckEnableClockSync()
 {
-  if (fabs(m_fps - g_graphicsContext.GetFPS()) < 0.01)
+  // refresh rate can be a multiple of video fps
+  double diff;
+  if (g_graphicsContext.GetFPS() >= m_fps)
+    diff = fmod(g_graphicsContext.GetFPS(), m_fps);
```
popcornmix: This doesn't look quite right. Suppose:
``g_graphicsContext.GetFPS() = 59.94``  ``m_fps=30``
``diff = .998`` so clocksync is not enabled, but I think it should be.

fritsch: fmod(59.94, 30.0) =  29.94

An additional condition if wanted could be || abs(m_fps - diff) < threshold

popcornmix: Does this work in all cases?
```
int ratio = g_graphicsContext.GetFPS() / m_fps + 0.5;
if (fabs(m_fps * ratio - g_graphicsContext.GetFPS()) < 0.01)
```
May need to protect against m_fps == 0.0.

FernetMenta: this is fine adjustment for a/v sync
you can't calculate an offset for pts/vblank, if fps does not exactly match refresh rate.
video pts relates to the time a frame is displayed (vblank) and ideally audio pts is synced exactly to this time.

popcornmix: Okay - I was assuming you wanted the 1001/1000 variants to be enabled - thinking the 0.01 was a 1% match, but it's actually more like a 0.01/30 = 0.03% match which excludes the 1001/1000 variants.
In that case the code does work for my example, so no problem.

FernetMenta:
added check for m_fps == 0
jenkins build this please

PR #9861, opened 2016-05-22
VideoPlayerVideo: Need to clear pictures before switching decoder

popcornmix:
The new decoder may not know how clear a picture initialised from a different decoder.
For example MMALCodec wants Release() called on clear, but VideoCodecFFMPEG does not.

PR #9882, opened 2016-05-28
[CodecIMX] correction after 449758b82b8ff92590c420f213c918b94495eb0f  …

mk01:
…(remove Dispose API)


mk01:
@FernetMenta @popcornmix 
can you please click-in this too ? #9861 by mistake removed Dispose() from different class.

PR #9884, opened 2016-05-28
SendClick() should operate on the focused window/dialog

ronie:
when SendClick(controlID) is run from a dialog, the message is send to the underlying window instead of to the dialog itself. this commit fixes that.

as reported on the forum: http://forum.kodi.tv/showthread.php?tid=277216

@BigNoid / @HitcherUK / @phil65 if you can think of any possible regressions, let me know


BigNoid:
No regressions found. When sending a click to anything other than the focused window, the window id should be added as parameter anyway imo.

stefansaraev:
jenkins build this please

PR #9888, opened 2016-05-29
[pvr] add instance pointer value to all add-on calls

AlwinEsch:
This is also one part needed for reworked binary PVR add-ons, it add a handle pointer to all calls from Kodi to add-on. This handle is also usable for the current version and can be set during add-on create with `RegisterHandle` call on libPVR.

Goal from them is to identify a from add-on headers defined parent class (comes in future) on calls to Kodi, further is another coming goal to allow more as one client connection on a add-on.

If this way is Ok becomes from me for all add-ons a request created. Also becomes this way added to other types.

### DIFF
--- a/addons/xbmc.pvr/addon.xml
+++ b/addons/xbmc.pvr/addon.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<addon id="xbmc.pvr" version="5.2.0" provider-name="Team-Kodi">
-  <backwards-compatibility abi="5.2.0"/>
+<addon id="xbmc.pvr" version="5.3.0" provider-name="Team-Kodi">
+  <backwards-compatibility abi="5.3.0"/>
   <requires>
     <import addon="xbmc.core" version="0.1.0"/>
   </requires>
--- a/lib/addons/library.xbmc.pvr/libXBMC_pvr.cpp
+++ b/lib/addons/library.xbmc.pvr/libXBMC_pvr.cpp
@@ -197,4 +197,12 @@ DLLEXPORT void PVR_transfer_channel_group_member(void *hdl, void* cb, const ADDO
   ((CB_PVRLib*)cb)->TransferChannelGroupMember(((AddonCB*)hdl)->addonData, handle, member);
 }
 
+DLLEXPORT void PVR_register_addon_instance(void *hdl, void* cb, void* addonInstance)
+{
+  if (cb == nullptr)
+    return;
+
+  ((CB_PVRLib*)cb)->RegisterAddonInstance(((AddonCB*)hdl)->addonData, addonInstance);
+}
+
 };
--- a/xbmc/addons/PVRClient.cpp
+++ b/xbmc/addons/PVRClient.cpp
@@ -64,7 +64,8 @@ std::unique_ptr<CPVRClient> CPVRClient::FromExtension(AddonProps props, const cp
 CPVRClient::CPVRClient(AddonProps props)
   : CAddonDll<DllPVRClient, PVRClient, PVR_PROPERTIES>(std::move(props)),
     m_apiVersion("0.0.0"),
-    m_bAvahiServiceAdded(false)
+    m_bAvahiServiceAdded(false),
+    m_addonInstance(nullptr)
 {
   ResetProperties();
 }
@@ -76,7 +77,8 @@ CPVRClient::CPVRClient(AddonProps props, const std::string& strAvahiType, const 
     m_strAvahiIpSetting(strAvahiIpSetting),
     m_strAvahiPortSetting(strAvahiPortSetting),
     m_apiVersion("0.0.0"),
-    m_bAvahiServiceAdded(false)
+    m_bAvahiServiceAdded(false),
+    m_addonInstance(nullptr)
 {
   ResetProperties();
 }
@@ -125,6 +127,11 @@ ADDON::AddonPtr CPVRClient::GetRunningInstance() const
   ADDON::AddonPtr addon;
   CPVRManager::GetInstance().Clients()->GetClient(ID(), addon);
   return addon;
+}
+
+void CPVRClient::RegisterAddonInstance(void* addonInstance)
+{
+  m_addonInstance = addonInstance;
 }
 
 void CPVRClient::ResetProperties(int iClientId /* = PVR_INVALID_CLIENT_ID */)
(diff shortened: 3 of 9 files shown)


FernetMenta:
This does not make much sense to me. Why do you want to pass a pointer with every function call?

AlwinEsch:
Is to identify on reworked systems a on coming add-on dev-kit used parent class for add-ons.

As example see here:
Internal code to communicate with Kodi: https://github.com/AlwinEsch/kodi/blob/binary-addon-level3/xbmc/addons/kodi-addon-dev-kit/include/kodi/api3/pvr/.internal/internalTail.hpp
The for add-on needed header with parent class: https://github.com/AlwinEsch/kodi/blob/binary-addon-level3/xbmc/addons/kodi-addon-dev-kit/include/kodi/api3/pvr/Addon.hpp

FernetMenta:
@AlwinEsch and I had a telephone call and he explained it to me. This PR tries to solve a fundamental issue of our current binary addons design.

1) Currently addons instantiate objects as globals. As a result we can only have a single addon instance of a given type.
2) An ugly hack that copies/renames the shared lib, works around 1)

Now I see the need for passing the handle with every call. This is the object (addon instance). Not sure if there are better alternatives. @Paxxi @Montellese do yo have ideas?

Jalle19:
Wouldn't it be enough to pass the handle to `ADDON_Create()` and let the addon store the handle, that way every API call wouldn't have to include this pointer?

ksooo:
As I understand the idea it is sort of other way round, @Jalle19. Addon_Create internally creates an instance and returns a handle to that instance to Kodi. Kodi stores all the handles and passes the right one back with every call to the addon c api. This way an add-on can identify the internal instance belonging to reach particular call. With this concept, an add-on internally can manage multiple instances. @Alwinus, correct me if I got it wrong.

AlwinEsch:
Unfortunately not, there are two reasons. One is that with reworked add-on system the handle is used on the dev-kit to identify his related class and the other (bigger) `ADDON_Create()` becomes only called one time, but on end are more as one possible to use, thats why it is needed on every call to know what is the from add-on opened class (if more as one becomes used).

Jalle19:
@AlwinEsch so is the ultimate goal like @ksooo said to be able to handle multiple instances in the same addon?

AlwinEsch:
Hi @ksooo you are right, this is first step to bring it in and to prevent conflicts between new and old add-ons.

@Jalle19 correct.

AlwinEsch:
In future if everything becomes complete is the `ADDON_Create()` no more related to his add-on type and more like a `main()` on exe and Kodi say what it want. The other is then that `ADDON_Create()` then can also be as add-on for video (like now with python).

ksooo:
@AlwinEsch what does an adoon do with the handle? Do you have an example?

Jalle19:
It sounds like every addon that wants to support multiple instances of itself has to handle it internally, which is not that ideal. Couldn't this be handled in Kodi somehow? Maybe I'm misunderstanding something.

AlwinEsch:
Hi @ksooo see as example here the [pvr.demo](https://github.com/AlwinEsch/pvr.demo/blob/change-to-api3/src/PVRDemoData.h#L99) where the handle is required for `KodiAPI::PVR::CAddonInterface`. Everything what was before on client.cpp is gone with this.

The handle itself becomes on begin used by this [add-on header](https://github.com/AlwinEsch/kodi/blob/binary-addon-level3/xbmc/addons/kodi-addon-dev-kit/include/kodi/api3/pvr/.internal/internalTail.hpp#L133) what the handle use to get the on first link defined class to call.

Thats the first part of the use from them. The other who comes later is that the with `public KodiAPI::PVR::CAddonInterface` defined part can be created more as one time. 

With the to do everything in kodi I'm not sure if it can be easy and not sure if `dlopen` (to open add-on can be done more as one time on exact the same `dll`.

AlwinEsch:
The for me very nice part is the [client.cpp](https://github.com/AlwinEsch/pvr.demo/blob/change-to-api3/src/client.cpp) if this becomes used. There nearly nothing needed anymore.

AlwinEsch:
For the with [KodiAPI::PVR::CAddonInterface](https://github.com/AlwinEsch/kodi/blob/binary-addon-level3/xbmc/addons/kodi-addon-dev-kit/include/kodi/api3/pvr/Addon.hpp#L45-L1302) defined class can be this used as [docs](http://alwinesch.github.io/group___c_p_p___kodi_a_p_i___p_v_r___c_addon_interface.html) if everything is done.

FernetMenta:
@AlwinEsch the linked client.cpp does still has globals that would not allow multiple instances.

AlwinEsch:
Yes thats right, the multiple instances comes not in one step. 

First I want to have the needed class for them on add-on itself who make it much more clean on add-on. To bring in the complete support is a much need to do on Kodi itself. But if the add-on function is already in needed style and the add-on itself use this parent class are only smaller parts needed.

AlwinEsch:
If everything is done comes this `new PVRDemoData;` on another place (with a new add-on function). This ADDON_Create should be called only one to inform them that it is enabled from user.

FernetMenta:
In order to get a better view if this goes into right direction, it would be great to have an example/prototype that shows final concept.

AlwinEsch:
Prototype is comming, but need a bit time ;)

AlwinEsch:
Is updated and `handle` renamed to `addonInstance`.

About the final (need a lot more) can you see here pvr.demo who using complete my wanted binary add-on way (there is really much gone away :D ):
https://github.com/AlwinEsch/pvr.demo/tree/change-to-api3/src
There is as new addon functions the `CreateInstance` and `DestroyInstance`, used to allow Kodi to create them several times.

The related kodi headers for them are here (still a lot todo but pvr.demo is working with the new way):
https://github.com/AlwinEsch/kodi/blob/binary-addon-rework-complete/xbmc/addons/kodi-addon-dev-kit/include/kodi/api2/pvr/Addon.hpp
https://github.com/AlwinEsch/kodi/blob/binary-addon-rework-complete/xbmc/addons/kodi-addon-dev-kit/include/kodi/api2/Addon.hpp

About the PVR client is the change especially, it is no more related to `AddonDll` who allows the several times usage:
https://github.com/AlwinEsch/kodi/blob/binary-addon-rework-complete/xbmc/addons/PVRClient.h#L67

If this is OK then I create requests to all PVR add-ons and the first step of rework comes in.

FernetMenta:
I have looked at the prototype and think @AlwinEsch did a great job. It is not an easy task and still a long way to got but we have to start somewhere. 
+1

PR #9892, opened 2016-05-31
[python] add support for image border

ronie:
scripts using WindowXML would crash if they use an image control with a bordertexture and/or bordersize tag.
`RuntimeError: Unknown control type for python`

let's fix that.

@tamland


ronie:
ping @tamland

tamland:
I guess it make sense.. I think you know the WindowXML stuff way better than I do;)

ronie:
thx!

PR #9806, opened 2016-05-12
[cmake/ios] Packaging

fetzerch:
@Memphiz: This is the packaging work for IOS. This time I fear I'd need your help. I cannot get it to work because fakeroot fails to set the permissions on the package contents. Could that be an issue with El Capitan? Maybe you could try it out on your machine. Next problem is that I don't have an ios device, so I can't test the generated deb.

Other than that it's fairly similar to osx, except that we have to take care of these bundle resources. I'll see if I find a better way to handle them once we got it working.

Jenkins seems to be fine already: [job](http://jenkins.kodi.tv/job/IOS/7344/) / [deb](http://mirrors.kodi.tv/test-builds/darwin/ios/kodi-20160512-040c238-cmake_ios_packaging-ios.deb)

PR #9895, opened 2016-06-01
[cmake] Android Packaging

fetzerch:
This is the packaging work for android (`make apk`). It's implemented using the already existing makefile, in a similar way as for [OSX](https://github.com/xbmc/xbmc/pull/9716) and [iOS](https://github.com/xbmc/xbmc/pull/9806). CPack can be done at a later point.

Jenkins is able to build packages for both arm and x86. I was able to install the arm package on my phone (http://mirrors.kodi.tv/test-builds/android/arm/kodi-20160601-0cf692d-cmake_android_packaging-armeabi-v7a.apk).

Unfortunately Kodi fails to start at the moment. Anyone interested to test and support me here?


fetzerch:
A bit of progress:
- The "failed to download obb" error was caused by the apk being < than 40 MB, because FFMPEG was not linked.
- When loading libcpluff, `dll_dlopen` couldn't be resolved because the wrapper libs didn't link `libkodi.so` on android.

Unfortunately libcpluff still cannot be loaded. This time I get:
```
ERROR: Unable to load libcpluff-arm.so, reason: (null)
```
Help would still be appreciated.

MrMC:
You need to package it exactly like how the existing Makefile does it. Anything else is epic fail. To test, apk's are just zips, so change the extension to zip and extract. This will let you compare the existing packaging to the new one.

fetzerch:
Yes, that's what I do. The problem was caused by libs not having the correct arch suffix. The error I got before (about not being able to resolve `dll_dlopen`) tricked me into believing that the library was found. It's fixed now and Kodi finally starts up and is working.

I notice still one small difference in the package. The drawable directories are missing the `-v4` suffix.
Looking at the Makefile (https://github.com/xbmc/xbmc/blob/master/tools/android/packaging/Makefile.in#L104) I can't see where that is coming from. Do you know by chance?

MrMC:
"-v4 suffix" eh ?

fetzerch:
The latest nightly from Jenkins has for example this:
`res\drawable-hdpi-v4\ic_launcher.png`
The apk built with CMake:
`res\drawable-hdpi\ic_launcher.png`
(same for all res\drawable-* dirs).

Apparently the v4 is the Plarform Version. It's described here https://developer.android.com/guide/topics/resources/providing-resources.html. But since the original Makefile doesn't add that suffix, I was wondering where in the build steps it gets added.

MrMC:
might want to check how Jenkins is set up. 

http://stackoverflow.com/questions/35700272/android-build-tool-adds-v4-qualifier-to-drawable-folders-by-default-in-generated

fetzerch:
Thanks! That was indeed just due to different sdk/ndk versions on jenkins and my dev machine.

@hudokkow: Mind reviewing the cmake parts?

hudokkow:
Code looks good. Great work! :+1:

PR #9897, opened 2016-06-02
[audiodsp] Use label/label2/icon instead of properties

BigNoid:
To be consistent I changed the list properties to common label/label2/icon.
Also added a check if the icon is empty to use `DefaultAddonAudioDSP.png` instead so no fallbacks have to be used in skin code.

@phil65 @AchimTuran

### DIFF
--- a/xbmc/settings/dialogs/GUIDialogAudioDSPManager.cpp
+++ b/xbmc/settings/dialogs/GUIDialogAudioDSPManager.cpp
@@ -1026,14 +1026,17 @@ CFileItem *CGUIDialogAudioDSPManager::helper_CreateModeListItem(CActiveAEDSPMode
   // set list item properties
   pItem->SetProperty("ActiveMode", isActive);
   pItem->SetProperty("Number", number);
-  pItem->SetProperty("Name", modeName);
+  pItem->SetLabel(modeName);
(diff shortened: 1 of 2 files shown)


xbmc/settings/dialogs/GUIDialogAudioDSPManager.cpp
```diff
@@ -1026,14 +1026,17 @@ CFileItem *CGUIDialogAudioDSPManager::helper_CreateModeListItem(CActiveAEDSPMode
   // set list item properties
   pItem->SetProperty("ActiveMode", isActive);
   pItem->SetProperty("Number", number);
-  pItem->SetProperty("Name", modeName);
+  pItem->SetLabel2(modeName);
```
AlwinEsch: Looks good, but maybe swap Label2 with Label, the Mode name is more the primary in DSP Manager.

BigNoid:
@AlwinEsch Ah okay. I just followed what Estuary used as primary as I am not that into audio dsp.
Updated.

phil65:
Makes sense. Thx!

AchimTuran:
Look good thanks.

BigNoid:
jenkins build this please

PR #9898, opened 2016-06-02
[guilib] Remove cache before opening plugin urls from shortcut - Fixes #16560

### LINKED ISSUE #16560: feature_python3 - not compatible with Python 3.8
### Describe the bug

xbmc/interfaces/python/PythonInvoker.cpp

There are a three references in the file above to the private interfaces of PyInterpreterState.

Example:

    s = state->interp->tstate_head;

Python 3.8 has finally removed the public interfaces to PyInterpreterState. The C documentation has always stated that there were never any public interfaces but projects such as Kodi used them anyway.

Announcement:
https://mail.python.org/pipermail/python-dev/2019-February/156344.html

Bug tracker to request a public interface:
https://bugs.python.org/issue35886

## Possible Fix
Work with upstream to create a real public interface.

Workaround:

Adding the following to the top of PythonInvoker.cpp will expose the structs again.

    #if PY_VERSION_HEX >= 0x03080000
    # define Py_BUILD_CORE
    # undef HAVE_STD_ATOMIC
    /* for access to the fields of PyInterpreterState */
    #  include "internal/pycore_pystate.h"
    # undef Py_BUILD_CORE
    # define HAVE_STD_ATOMIC
    #endif

### Your Environment
Used Operating system:

 - [ ] Android
 - [ ] iOS
 - [X] Linux
 - [ ] OSX
 - [ ] Raspberry-Pi
 - [ ] Windows
 - [ ] Windows UWP

 - Kodi version: 18.4 / 19.x

BigNoid:
As a plugin is mostly online content the cache should be removed before opening to make sure we get latest content.

### DIFF
--- a/xbmc/windows/GUIMediaWindow.cpp
+++ b/xbmc/windows/GUIMediaWindow.cpp
@@ -656,8 +656,9 @@ bool CGUIMediaWindow::GetDirectory(const std::string &strDirectory, CFileItemLis
     CAddonMgr::GetInstance().UpdateLastUsed(pathToUrl.GetHostName());
 
   // see if we can load a previously cached folder
+  // skip in case of plugin content as this needs the latest content
   CFileItemList cachedItems(strDirectory);
-  if (!strDirectory.empty() && cachedItems.Load(GetID()))
+  if (!strDirectory.empty() && cachedItems.Load(GetID()) && !pathToUrl.IsProtocol("plugin"))
   {
     items.Assign(cachedItems);
   }


Montellese:
The question is if the lists should be cached at all. It might change at any minute and only showing changes after closing/re-opening the plugin might be confusing as well.

BigNoid:
Agreed, but if you stay idle on the list it doesn't get refreshed at all even if there was no cache I think?
Any other way of entering a plugin list triggers the refresh/update routine which clears the cache first.

phil65:
Just FYI, there already is a cache param ("cacheToDisc" ) for the xbmcplugin.endOfDirectory() call which can be used by add-on authors. Not 100% sure if that is related though.

BigNoid:
I did some testing and seems that urls are only cached if slow. Specifically it is checked [here](https://github.com/xbmc/xbmc/blob/master/xbmc/windows/GUIMediaWindow.cpp#L688-L689).
I think the best way is either current implementation or adding a check there to never cache plugin urls. If I am not mistaken it will affect directory provider if CacheToDiscIfSlow() is set to false for plugin urls, which will make widgets loading slower.

MartijnKaijser:
still needed?

coolllman:
Yes this need, i remove cache from temp folder everyday

phil65:
I guess @tamland knows best how to deal with this?

tamland:
Also wondering what's the point of caching in the first place when it's always removed?

BigNoid:
Updated now to don't look for a cached folder in case of plugin content. If we need to remove caching completely for plugin content than I am in over my head.

PR #9899, opened 2016-06-03
VideoPlayer: init ffmpeg threads like LAVFilters does

FernetMenta:
see title

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp
@@ -302,9 +302,9 @@ bool CDVDVideoCodecFFmpeg::Open(CDVDStreamInfo &hints, CDVDCodecOptions &options
     }
     else
     {
-      int num_threads = std::min(8 /*MAX_THREADS*/, g_cpuInfo.getCPUCount());
-      if( num_threads > 1)
-        m_pCodecContext->thread_count = num_threads;
+      int num_threads = av_cpu_count() * 3 / 2;
+      num_threads = std::max(1, std::min(num_threads, 16));
+      m_pCodecContext->thread_count = num_threads;
       m_pCodecContext->thread_safe_callbacks = 1;
       m_decoderState = STATE_SW_MULTI;
       CLog::Log(LOGDEBUG, "CDVDVideoCodecFFmpeg - open frame threaded with %d threads", num_threads);


afedchin:
is this related to http://forum.kodi.tv/showthread.php?tid=277724?

fritsch:
@afedchin yes - that should allow up to 16 threads, while av_cpu_count returns the number of logical cores. E.g. 8 on a i7 quad core. 

The *3 / 2 most likely was found by cpuload testing as it seems the decoder does not fully use the cpu cores. So running 50% more most likely has increased performance.

afedchin:
@fritsch thanks for clarification.

FernetMenta:
jenkins build this please

MrMC:
can remove 

#if defined(TARGET_POSIX) || defined(TARGET_WINDOWS)
#include "utils/CPUInfo.h"
#endif

MrMC:
fixed for android
https://github.com/MrMC/mrmc/commit/602d6d90a07682ff6233bb573ead3523db084770

fritsch:
@MrMC av_cpu_count <- does not work correctly or does it ignore sleeping CPUs which are not usable afterwards anyways?

MrMC:
av_cpu_count only counts active cpus. With the change, sleeping cpus will be woken and used. Tested under a quad.

stefansaraev:
that means, not quite working on linux if hotplug cpu scaling governor is used, too :)

MrMC:
depends on if sched_getaffinity is present on linux and others.

stefansaraev:
uh. it is available of course, but it wont count "inactive" cpus (hotplug, as said..) ;)

however. I dont think it will be a big deal

MrMC:
It can be a big deal depending on cpus active at time of call. Leads to unpredictable behavior for sw decoding that depends now on how many cpus are active at the time of the call. If you have all spun up, then can handle resource hungry formats, if only one then same media will max decode and fall behind.

stefansaraev:
ookay. then I guess hotplug will not be good choice anymore, I'll advice people switch back to ondemand.

btw, I just tested it. while /proc/cpuingo says "4 cpus", sched_getaffinity sees 3 right now (most of the time  it sees 1, if kodi is idle) :)

FernetMenta:
let's switch back to g_cpuInfo.getCPUCount() and multiply with 3 / 2

stefansaraev:
@FernetMenta most linux hw/distributions do not use hotplug, it's just wetek (and maybe some chineses) who do afaik. hotplug is not even available on mainline kernels. so up to you :)

PR #8018, opened 2015-09-11
refactor builtins

FernetMenta:
credits to @notspiff for this excellent work
- makes CBuiltins a singleton
- refactors the methods into several files

PR #9901, opened 2016-06-03
[builtins] fix build without non-free

stefansaraev:
ref #8018


stefansaraev:
jenkins build and merge

PR #9902, opened 2016-06-03
fixed fps rate on startup

ghost:
This fixes the fps rate change for inputstream video playback


FernetMenta:
that is wrong. please drop irFpsScale and assign the correct variable here
https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxClient.cpp#L419

PR #9903, opened 2016-06-04
Path rule for music "artists" type smart playlists

DaveTBlake:
Addition of "path" as a rule for (music) "artists" type smart playlists.

Already available for "songs" and "albums" type playlists, his will filter the artists with songs with path that satisfy the rule.


zag2me:
Definitely a good idea, I just noticed this missing as well

DaveTBlake:
Thanks @zag2me, it already did this in my imagination.

jenkins build this please

PR #9908, opened 2016-06-04
[cmake] Make sanitizers easy to use

neo1973:
To use a sanitizer with clang or gcc add -DECM_ENABLE_SANITIZERS='san1;asn2;...'
as an option to your CMake command line. More documentation is provided in the
CMake file.

Note: This file is from the "Extra CMake Modules" package: https://api.kde.org/ecm/module/ECMEnableSanitizers.html

### DIFF
--- a/project/cmake/CMakeLists.txt
+++ b/project/cmake/CMakeLists.txt
@@ -61,6 +61,7 @@ set(CORE_BUILD_DIR build)
 message(STATUS "Source directory: ${CORE_SOURCE_DIR}")
 message(STATUS "Build directory: ${CMAKE_BINARY_DIR}")
 
+include(modules/extra/ECMEnableSanitizers.cmake)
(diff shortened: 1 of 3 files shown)


fetzerch:
Nice, that's definitively something I want to play with.
@hudokkow: What do you think about putting in external modules? It should probably go into the modules folder, should we have a subfolder for these?

hudokkow:
Fine by me. Name subfolder "extra"?

fetzerch:
Yes sounds good, @neo1973 mind moving it to modules/extra?

If you want, you could also mention it in https://github.com/xbmc/xbmc/blob/master/project/cmake/README.md. Maybe in a section before "Debugging the build".

neo1973:
Done. I hope everything is OK now?

fetzerch:
Yes, looks good now. Thanks much!

PR #9914, opened 2016-06-05
fix failing activation of startup window

FernetMenta:
@MilhouseVH 
This cures the problem mentioned in this ticket http://trac.kodi.tv/ticket/16754

@ksooo @Jalle19 
But the actual problem is that setting PVR (TV) window as startup window is wrong by design. A window that is not available at startup and that depends on other services can't be a startup window. I see those options:

1) Fix PVR window and allow activation (as empty window) if there are no connected PVR clients. Currently it would crash.

2) Add a setting to PVR that makes Kodi switch to PVR window as soon the first PVR clients becomes available. Remove TV window form the list of startup windows.

3) Do nothing and just remove PVR from the list of startup windows.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -1153,6 +1153,9 @@ bool CApplication::Initialize()
       int firstWindow = g_SkinInfo->GetFirstWindow();
       g_windowManager.ActivateWindow(firstWindow);
 
+      if (g_windowManager.GetActiveWindowID() == WINDOW_STARTUP_ANIM)
+        g_windowManager.ActivateWindow(WINDOW_HOME);
+
       // the startup window is considered part of the initialization as it most likely switches to the final window
       uiInitializationFinished = firstWindow != WINDOW_STARTUP_ANIM;
 


FernetMenta:
jenkins build this please

MilhouseVH:
Many thanks - that fixes the issue.

FernetMenta:
Linux build error unrelated.

Jalle19:
@FernetMenta how is the current logic different from 2)?

FernetMenta:
@Jalle19 
- currently there is no such PVR setting
- there is no code that would activate PVR window

ronie:
many skins run one or more scripts in Startup.xml that need some time to finish before opening the Home window.
this breaks skins that keep WINDOW_STARTUP_ANIM active for a while to display a startup animation for instance.

FernetMenta:
we can't exit init in an unknown state. if the script fails to activate the desired window, the application is in an unrecoverable state. those scripts need to run synchronous at startup.

FernetMenta:
@ronie would you have detailed steps to reproduce this issue?

ronie:
it's not just the use of scripts that's now broken in Startup.xml, it's plain skin code as well. 

to test, set your startup window to pictures, for instance.
then use the attached Startup.xml in Estuary.

without this PR, it opens the pictures window on startup (as expected)
with this PR applied, it opens the Home window instead...

[startup.zip](https://github.com/xbmc/xbmc/files/299300/startup.zip)

FernetMenta:
unfortunately this is really bad design of Kodi. it hands off control to an addon (skin) and prays that this addon would activate another window. if this fails, rien ne va plus
I'll think of some solution.

PR #9915, opened 2016-06-05
Default Scale for HQ scaling above

fritsch:
This setting is fully confusing for users as hq scaling is a waste of resources if source and target nearly have the very same dimensions. It puts immense load on the GPU while visual quality improvement is not given at all. There is no need to process a 1080p file with a 6x6 convolution shader if the output image is still ~ 1080p.

Also the barely existing documentation is confusing to the user, too.

Example:
1080p Output resolution.

Video: 1280x720
(1920 - 1280) / 1920 = 33%
(1080 - 720) / 1080 = 33 %

Video: 1920x1080
(1920 - 1920) / 1920 = 0 %
(1080 - 1080) / 1080 = 0%

In code the minimum of both values is chosen.

In short a default value of 20% will use HQ scaling when scaling 720p to 1080p or when scaling 576i to 1080p. It won't use hq scalers starting with (1536 width) for this 1080p example.

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -16647,7 +16647,7 @@ msgstr ""
 #. Description of setting with label #13435 "Enable HQ Scalers for scaling above"
 #: system/settings/settings.xml
 msgctxt "#36154"
-msgid "Use high quality scalers when upscaling a video by at least this percentage."
+msgid "Use high quality scalers when upscaling a video by at least this percentage. A value below 5% makes little sense as video is processed with high GPU load without any visible picture quality improvements."
 msgstr ""
 
 #. Description of setting with label #13425 "Allow hardware acceleration (VDPAU)"
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -100,7 +100,7 @@
         </setting>
         <setting id="videoplayer.hqscalers" type="integer" parent="videoplayer.rendermethod" label="13435" help="36154">
           <level>2</level>
-          <default>0</default>
+          <default>20</default>
           <constraints>
             <minimum>0</minimum>
             <step>10</step>


addons/resource.language.en_gb/resources/strings.po
```diff
@@ -16647,7 +16647,7 @@ msgstr ""
 #. Description of setting with label #13435 "Enable HQ Scalers for scaling above"
 #: system/settings/settings.xml
 msgctxt "#36154"
-msgid "Use high quality scalers when upscaling a video by at least this percentage."
+msgid "Use high quality scalers when upscaling a video by at least this percentage. A value below 5% makes limited sense as video is processed with high load whithout any visible picture quality improvements."
```
MartijnKaijser: .... high CPU load ....

fritsch: It's GPU load - will add.

FernetMenta:
makes sense

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -16647,7 +16647,7 @@ msgstr ""
 #. Description of setting with label #13435 "Enable HQ Scalers for scaling above"
 #: system/settings/settings.xml
 msgctxt "#36154"
-msgid "Use high quality scalers when upscaling a video by at least this percentage."
+msgid "Use high quality scalers when upscaling a video by at least this percentage. A value below 5% makes limited sense as video is processed with high GPU load whithout any visible picture quality improvements."
```
MrMC: whithout -> without

MrMC: also 'limited sense' is strange english :) 'little sense' is more common among mortals.

kq01526:
Related to:

http://forum.kodi.tv/showthread.php?tid=231955&pid=2352246#pid2352246

@fritsch :

Thank you.

I see that [you have changed the description to say "A value below 5% makes limited sense as video is processed with high GPU load whithout any visible picture quality improvements."](https://github.com/xbmc/xbmc/pull/9915/commits/9b740df329846408f3be5664d0c4a4f1fd372200).

But at the same time [you have left the minimum value at 0%](https://github.com/xbmc/xbmc/pull/9915/commits/a9bb82ae82cfce000fcfc17bd2f90efccde2cd42).

Why not raise the minimum value to 5% then?

Also, why are the steps set to "10"?

Why not change the steps to "1"?

Regards

da-anda:
minimum value of 0 is required to turn it off atm, but do we still need he "off" state with all the enumeration etc we're having in place these days? And platforms that don't support HQ scalers will hide the entire setting anyways.

fritsch:
@kq01526 Here is a math work for you:

Which value do you have to setup in per cent, so that on a 1400x1050 resolution a video of 1366x1024 is upscaled via HQ scalers. If you can answer this correctly - you are ready to answer me, why we have 10% steps and not 5% and not 1%.

Awaiting your answer.

kq01526:
> @fritsch wrote:
> 
> @kq01526 Here is a math work for you:
> 
> Which value do you have to setup in per cent, so that on a 1400x1050 resolution a video of 1366x1024 is upscaled via HQ scalers.

Whooops, maybe you should have removed your math from [the first post](https://github.com/xbmc/xbmc/pull/9915#issue-158551259) prior to asking me.

Because now I can simply be lazy and just replace the numbers:

> @fritsch wrote:
> 
> If you can answer this correctly

I guess I can:

1050p Output resolution.

Video: 1366x1024
(1400 - 1366) / 1400 = 2.43%
(1050 - 1024) / 1050 = 2.48%

> @fritsch wrote:
> 
> you are ready to answer me, why we have 10% steps and not 5% and not 1%.
> 
> Awaiting your answer.

Sorry, I am not ready to answer you.

I still don't get why we have 10% steps and not 5% nor 1% steps.

Please explain.

Awaiting your answer.

Regards

fritsch:
The question is as follows:

Given an amount of N default CEA resolutions M. Which value p of stepsize is needed so that one can seperate all elements of M from each other while uniting resolution that have the same width?

In short: 10 is enough to do so.

Asked differently: Find two CAE modes that you cannot separate with a difference of 10 %.

fritsch:
jenkins build this please

PR #9918, opened 2016-06-05
[pictures] remove few raw formats

stefansaraev:
we do not support those raw image formats, anyway

related bug report: http://forum.kodi.tv/showthread.php?tid=278594

I could not reproduce, but I see no point trying to display those as pictures or generating thumbs.

### DIFF
--- a/xbmc/settings/AdvancedSettings.cpp
+++ b/xbmc/settings/AdvancedSettings.cpp
@@ -382,7 +382,7 @@ void CAdvancedSettings::Initialize()
   m_databaseMusic.Reset();
   m_databaseVideo.Reset();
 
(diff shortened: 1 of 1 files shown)


stefansaraev:
@notspiff I know you have support for imageencoder addons in your fork. how does that work? each imageencocder (well, decoder..) addon registers a list of extensions or those still depend on m_pictureExtensions ?

notspiff:
It should but iirc i never added that part. Easy enough to add, just monkey
how its done for audio decoders.
5. jun. 2016 16:48 skrev "Stefan Saraev" <notifications@github.com>:

stefansaraev:
asking, because I did look at imageencoder.raw and could not see anything that tells kodi "I handle this and that extension". are you ok with this PR ?

notspiff:
Yeh, it was a poc to show how it could be done, not a complete solution.
Removing unsupported extensions makes fully sense to me.
5. jun. 2016 16:52 skrev "Stefan Saraev" <notifications@github.com>:

garbear:
FWIW, game extensions are entirely determined by aggregating the <extensions> tags of addon.xml files. I even aggregate extensions from remote repos, which causes games to show up even if the add-ons isn't installed. As a result, the user can launch a game without an add-on, and the add-on is installed "just-in-time" and used to launch the game. I think this would be a cool feature for image encoder add-ons.

PR #9920, opened 2016-06-05
[cmake] Get rid of scripts/common/PathSetup

hudokkow:
It's only an include(...) placeholder.


hudokkow:

PR #9923, opened 2016-06-05
improve current code of unused db fields

phate89:
This change fixes some warning due to my incorrect use of NULL..
@fritsch thanks for your diff. I changed from DEPRECATED to UNUSED because imho is more rapresentative (if someone wants to use it's possible)
My idea is (when I find time) to kill this cXX stuff..

### DIFF
--- a/xbmc/video/VideoDatabase.cpp
+++ b/xbmc/video/VideoDatabase.cpp
@@ -2072,8 +2072,6 @@ std::string CVideoDatabase::GetValueString(const CVideoInfoTag &details, int min
   std::vector<std::string> conditions;
   for (int i = min + 1; i < max; ++i)
   {
-    if (offsets[i].type == NULL || offsets[i].offset == NULL)
-      continue;
     switch (offsets[i].type)
     {
     case VIDEODB_TYPE_STRING:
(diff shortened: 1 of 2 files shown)


xbmc/video/VideoDatabase.cpp
```diff
@@ -2072,8 +2072,6 @@ std::string CVideoDatabase::GetValueString(const CVideoInfoTag &details, int min
   std::vector<std::string> conditions;
   for (int i = min + 1; i < max; ++i)
   {
-    if (offsets[i].type == NULL || offsets[i].offset == NULL)
```
fritsch: the new type is not handled here and the offset (which is 0 for the UNUSED_TYPE) is also not checked. This worked by accident with the == NULL check.

How does that work now?

phate89: Before the compiler replaced nulls with 0 so if it was null (or 0 after the compiling) it will hit "continue" and pass it to the next field. 
But now that's a valid value for switch a specific check is pointless.. now it will enter in the switch and since there's not the unused type it won't find a match and it will continue automatically to the next one

fritsch: default:
continue;

please

fritsch: or we run into the next coverity report :-)

phate89:
@fritsch fixed it

fritsch:
@phate89 fine with that PR - but I am not the maintainer of this code. Just came into it through my compile errors. The maintainers shall decide.

phate89:
so I guess it's @Montellese call?

MartijnKaijser:
jenkins build this please

fritsch:
@MartijnKaijser you can press this button. Without that change the build log is spammed as hell and real errors cannot be found easily. If there are minor changes later on, like a default case or something - they can be appended.

Montellese:
IMO it's ok like this.

fritsch:
@Montellese thx!

PR #9924, opened 2016-06-06
[cmake/linux] Increase link performance (GOLD, DebugFission)

fetzerch:
Increase the link performance with GCC by
- defaulting to GNU Gold Linker if available
- defaulting to Debug Fission if available (https://gcc.gnu.org/wiki/DebugFission)

Some measurements on Ubuntu Xenial (GNU GCC 5.3.1, Binutils 2.26)
with incremental linking (touch xbmc/FileItem.cpp) of a debug build:
- GNU ld:   16.5s
- GNU gold:  6.5s

Not sure if someone had experimented with that already, but the results look promising. Will have to check if that's an option also for android.
Ping: @hudokkow, @wsnipex, @notspiff.


stefansaraev:
not sure I understand the cmake stuff (does cmake_dependent_option really takes care here?) right, but. -gsplit-dwarf existence in compiler does not guarantee -Wl,--gdb-index is available in linker. ld.bfd has no -Wl,--gdb-index, only gold has.

wsnipex:
he's checking for gold first, so should be fine

akva2:
neat. can we turn lto with gold, thus killing off the speed advantage again, i wonder.

hudokkow:
Just tested. Files are a *lot* smaller: kodi.bin - 167M vs 253M

Unfortunately it [crashed](http://paste.ubuntu.com/17060373/) in seconds:
> kodi.bin: ../nptl/pthread_mutex_lock.c:349: __pthread_mutex_lock_full: Assertion `INTERNAL_SYSCALL_ERRNO (e, __err) != EDEADLK || (kind != PTHREAD_MUTEX_ERRORCHECK_NP && kind != PTHREAD_MUTEX_RECURSIVE_NP)' failed.
Aborted (core dumped)
./kodi: 118: [: |/usr/share/apport: unexpected operator

Built with this PR on top of master. Master runs fine.

akva2:
strip the binary and be amazed ;)

fetzerch:
@hudokkow: Interesting. It works flawlessly here, but I'll do more testing.
Files being smaller is because the debug symbols are moved to separate files. It's explained nicely on the wiki page linked above.

project/cmake/modules/LDGOLD.cmake
```diff
@@ -0,0 +1,41 @@
...
+if(ENABLE_LDGOLD)
+  execute_process(COMMAND ${CMAKE_C_COMPILER} -fuse-ld=gold -Wl,--version ERROR_QUIET OUTPUT_VARIABLE LD_VERSION)
+  if(LD_VERSION MATCHES "GNU gold")
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fuse-ld=gold")
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fuse-ld=gold")
+    set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fuse-ld=gold")
+    message(STATUS "Linker: GNU gold")
+  else()
+    message(WARNING "GNU gold linker is not available, falling back to default system linker")
+    set(ENABLE_LDGOLD OFF)
```
akva2: personally i prefer the LDGOLD_FOUND idiom.

fetzerch: makes sense, done

hudokkow:
@fetzerch, I can't reproduce the crash either with this PR as is or [rebased](https://github.com/hudokkow/xbmc/tree/PR9924_rebased) on master.

Sorry for wasting your time. I say shove it in.

fetzerch:
@hudokkow: No need to worry, these kind of changes can also introduce issues.
@FernetMenta: [This](https://github.com/xbmc/xbmc/pull/9924#issuecomment-223950351) was the crash I asked you about yesterday.

I'd also say we merge it - and if it reappears, fix the root cause.

PR #6772, opened 2015-03-20
[musicdb] enable the library via xml

mkortstiege:
This enables the the music library nodes to be defined via xml just as we're already doing for the video library.

@Montellese for review please. Btw, I could need some smart playlist help so we can group by first letter. It's not really related to this PR but would come handy to have something like a flat and full blown library to choose from ;)

PR #9926, opened 2016-06-06
add MusicAddons as a library xml node

ronie:
not sure why one was forgotten when we moved to library xml nodes for music in https://github.com/xbmc/xbmc/pull/6772

ping @mkortstiege


mkortstiege:
Got lost somewhat. Thanks.

PR #9927, opened 2016-06-06
[upnp] fix TransportPlaySpeed state variable in CUPnPRenderer

Montellese:
These are two fixes for the `TransportPlaySpeed` state variable in the `CUPnPRenderer` implementation.
- The first commit fixes the speed value taken from an announcement which was trying to read a non-existing value.
- The second commit makes sure that the value of `TransportPlaySpeed` is never `0` which is forbidden by the UPnP AVTransport specification. When paused the value must be `1` as the `TransportState` state variable is used to specify that playback is paused.


MartijnKaijser:
jenkins build this please

Montellese:
Any objections?

PR #9930, opened 2016-06-07
Using TagLib for opus file tags.

beedaddy:
Thus gathering opus tags is on par with, for example, ogg vorbis tags.


fritsch:
@beedaddy which information are you missing from FFmpeg Tagloader?

DaveTBlake:
@fritsch what is the FFmpeg equivalent of TagLoaderTagLib.cpp?
Does it populate MusicInfoTag for opus files with all the tag data that say FLAC files get from Vorbis comments? I have no opus files to test with, but guessing that is the reason for this PR

fritsch:
@DaveTBlake https://github.com/xbmc/xbmc/blob/master/xbmc/music/tags/MusicInfoTagLoaderFFmpeg.cpp#L87

beedaddy:
@fritsch I was wondering... Please correct me if I'm wrong but the ffmpeg based tag reader only collects title, album, artist and tracknumber. But the taglib based tag reader already collects a lot more tags. So I thought switching to taglib is only a small change and thus it's on par with other file types which use xiphcomments.

beedaddy:
(But I'm very (!) new to Kodi, so I might have missed the reason why not to use the taglib approach.)

fritsch:
@beedaddy the ffmpeg tag reader was implemented when taglib did not yet support opus, which gave us the advantage of having metadata also when taglib has no support for it.

Do we need to bump taglib dependency for opus? Which version was the first version that supported it?

beedaddy:
@fritsch I think it was version 1.9.

fritsch:
Yeah - that way we loose support for Ubuntu 14.04. Technically your commit is fine, but I fear we break on different platforms.

@wsnipex do you have an overview where we have taglib 1.9.x ready and where not?

beedaddy:
@fritsch Ah, I haven't thought about that... Perhaps because taglib 1.9 is from 2013...

fritsch:
@beedaddy some people run debian stable :p don't mess with them.

beedaddy:
@fritsch But even they (Jessie) use 1.9.1, don't they? :)

fritsch:
@beedaddy it's on our chief packager to decide, you know as most likely also bumps in our depends need to happen, that's why I pinged him. Personally I also don't see an issue with bumping to 1.9.x but we need to ask before doing so.

stefansaraev:
taglib is already 1.11 in depends (so that's used for android / windows / osx)

we have taglib 1.9.1 in trusty and vivid stable ppas.

fritsch:
> configure.ac:PKG_CHECK_MODULES([TAGLIB],  [taglib >= 1.8]

fritsch:
@stefansaraev okay, then just bumping the configure should be good enough, right?

wsnipex:
yes, I backported taglib 1.9.1 to all ppas, so just bump min version in configure.ac and project/cmake/modules/FindTagLib.cmake

fritsch:
@beedaddy can you add an additional commit (Depends: Bump taglib min required version to 1.9) to your PR? Then it is ready to go.

stefansaraev:
jenkins build this please

DaveTBlake:
Beedaddy thanks for this, I have been vaguely aware that opus wasn't getting as much tag data as othe formats but then forgot all about it. 

But can anyone let me have some fully tagged opus files to test this with please?

beedaddy:
DaveTBlake: You could perhaps just convert a flac file with opusenc (from opus-tools). It transfers most of the tags (but perhaps not the cover art as I experienced just now).

DaveTBlake:
+1

Once I had an opus file I could generate more tagged variations for testing. All looks fine to me.

PR #9934, opened 2016-06-09
Fix: Fetching cover art from xiph comments.

beedaddy:
TagLib::Ogg::XiphComment now handles METADATA_BLOCK_PICTURE separately from other fields. Thus, we have to use XiphComment::pictureList() in order to get cover art.


beedaddy:
Sorry, "now" means TagLib 1.11. So this is a little more complicated to not breaking compatibility.

PR #9935, opened 2016-06-09
Fix: Fetching cover art from xiph comments for taglib >= 1.11

beedaddy:
TagLib::Ogg::XiphComment now handles METADATA_BLOCK_PICTURE separately from other fields. Thus, we have to use XiphComment::pictureList() in order to get cover art.

### DIFF
--- a/xbmc/music/tags/TagLoaderTagLib.cpp
+++ b/xbmc/music/tags/TagLoaderTagLib.cpp
@@ -650,6 +650,7 @@ bool CTagLoaderTagLib::ParseTag(Ogg::XiphComment *xiph, EmbeddedArt *art, CMusic
       if (iUserrating > 0 && iUserrating <= 100)
         tag.SetUserrating((iUserrating / 10));
     }
+#if TAGLIB_MAJOR_VERSION <= 1 && TAGLIB_MINOR_VERSION < 11
     else if (it->first == "METADATA_BLOCK_PICTURE")
     {
       const char* b64 = it->second.front().toCString();
@@ -676,10 +677,12 @@ bool CTagLoaderTagLib::ParseTag(Ogg::XiphComment *xiph, EmbeddedArt *art, CMusic
     {
       pictures[2].setMimeType(it->second.front());
     }
+#endif
     else if (g_advancedSettings.m_logLevel == LOG_LEVEL_MAX)
       CLog::Log(LOGDEBUG, "unrecognized XipComment name: %s", it->first.toCString(true));
   }
 
+#if TAGLIB_MAJOR_VERSION <= 1 && TAGLIB_MINOR_VERSION < 11
   // Process the extracted picture frames; 0 = CoverArt, 1 = Other, 2 = COVERART/COVERARTMIME
   for (int i = 0; i < 3; ++i)
     if (pictures[i].data().size())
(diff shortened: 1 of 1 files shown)


xbmc/music/tags/TagLoaderTagLib.cpp
```diff
@@ -694,6 +697,29 @@ bool CTagLoaderTagLib::ParseTag(Ogg::XiphComment *xiph, EmbeddedArt *art, CMusic
 
       break;
     }
+#else
+  auto pictureList = xiph->pictureList();
+  FLAC::Picture *cover[2] = {};
+
+  for (List<FLAC::Picture *>::ConstIterator i = pictureList.begin(); i != pictureList.end(); ++i)
```
fritsch: for auto i : xiph->pictureList() also works?

beedaddy: Thanks for the comment. Yes, I changed that...

DaveTBlake:
Thanks for this @beedaddy. Feeling a bit amazed that embedded art had stopped working with Taglib 1.11 and somehow I had not noticed! Does this impact all formats using Vorbis comments?

I'll do some testing once I get back to my dev env

fritsch:
Please squash the two commits.as @DaveTBlake seems to be fine and you @beedaddy don't go anywhere also fine with me :-)

beedaddy:
@DaveTBlake Yes, and I wondered (giving kodi on my Arch Linux a try, which uses taglib 1.11) why no one complains about seeing no cover art in ogg files. ;)

I think it impacts all formats with xiph comments, yes. Flac is an exception because it already handled cover art seperately before.

DaveTBlake:
Ah well I'm all Flac so would not see the issue. Why I need some more diverse format files with xiph comments for test purposes :)

DaveTBlake:
Thanks for the test file @beedaddy , I'll possibly create a few more.

Am I right in thinking that the new processing in Taglib 1.11 for all XiphComment format files is now the same has SetFlacArt? So would another approach be rather than excluding the depricated  "METADATA_BLOCK_PICTURE", "COVERART" and "COVERARTMIME" tags for 1.11 and above in ParseTag(Ogg::XiphComment...) be to call SetFlacArt for all formats?

It would remove the repeated code this PR creates by making it common.

beedaddy:
@DaveTBlake First, I would prefer to fetch the cover art in ParseTag() as this is done also in other tag formats like ID3v2 and it seems (for me) to be the right place. 
For some reason (hexdump shows no METADATA_BLOCK_PICTURE in my flac files but they use xiph comments) XiphComment::pictureList() for flac (!) files returns an empty list and you have to use FLAC::File::pictureList(). Because of this difference I left SetFlacArt() as it is but for all other Xiph-cases handle it inside ParseTag().

DaveTBlake:
@beedaddy I agree fetching of cover art from within ParseTag, like other formats, is best place.  I was just hoping  `XiphComment::pictureList()` would now work for flac too. Perhaps it is worth raising with taglib?

What you have done works fine for ogg files I have tested, so I will look to merge this next week (giving others the chance for an opinion)

As an aside I am getting different results for cover art with FLAC files between Jarvis (taglib 1.10) and master 1.11. Seems to be limited to FLAC files where I have replaced the embedded art using Mp3tag. That means `FLAC::File::pictureList()` behaving differently too.

DaveTBlake:
jenkins build this please

PR #9937, opened 2016-06-09
reduce impact of touching VideoDatabase.h, PVRRecordings.h

notspiff:
touch xbmc/video/VideoDatabase.h ->

Before PR: 99 objects rebuilt
After PR: 75 objects rebuilt

touch xbmc/pvr/recordings/PVRRecordings.h

Before PR: 36 object rebuilt
After PR:  15 objects rebuilt

real-life impact is probably small but fake dependency is fake. needs a jenkins run as header shuffles tend to bite on other platforms.

motivated by the ugly warning gcc 5.3 throws due to the hacky offset mapping stuff.

### DIFF
--- a/xbmc/ThumbnailCache.cpp
+++ b/xbmc/ThumbnailCache.cpp
@@ -23,7 +23,6 @@
 #include "threads/SingleLock.h"
 
 #include "FileItem.h"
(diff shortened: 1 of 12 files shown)


MartijnKaijser:
jenkins build this please

PR #9939, opened 2016-06-09
use numeric 0, not NULL

notspiff:
quells a bunch of warnings of the type

'/home/jenkins/workspace/LINUX-64/xbmc/video/VideoDatabase.h:339:1:
warning: converting to non-pointer type ‘size_t {aka long unsigned int}’
from NULL [-Wconversion-null]'


hudokkow:
jenkins build this please

DaveTBlake:
Isn't #9923 doing a similar change already? @phate89 you may need to be aware of this change

phate89:
Yes it's the same..I also changed the checks in VideoDatabase.cpp

notspiff:
yours is better.

PR #9940, opened 2016-06-09
changed: get rid of FileReaderFile

notspiff:
this was an earlier version of the VFS cache.
it was kept around as backwards compatiblity
wrapper for the READ_CACHE support as it was
used in some audio decoders.

the audio decoders have since been moved to add-ons and rewritten to
use the cache through the new mechanisms, so there is no longer
any legacy to take into consideration.


hudokkow:
jenkins build this please

MartijnKaijser:
jenkins build this please

PR #9944, opened 2016-06-10
call SetVideoResolution before streams are Enabled

ghost:
To select the best matching stream in inputstream.mpd's EnableStream(), I need the video resolution before EnableStream() is called.

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -895,6 +895,13 @@ void CVideoPlayer::OpenDefaultStreams(bool reset)
 
   // open video stream
   valid   = false;
+  
+  if (m_pDemuxer)
+  {
+    // TODO desired resolution needs to come from somewhere else
+    RESOLUTION_INFO res = g_graphicsContext.GetResInfo();
+    m_pDemuxer->SetVideoResolution(res.iWidth, res.iHeight);
+  }
 
   for (const auto &stream : m_SelectionStreams.Get(STREAM_VIDEO, PredicateVideoPriority))
   {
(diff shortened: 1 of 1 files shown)


xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3528,6 +3528,13 @@ bool CVideoPlayer::OpenStream(CCurrentStream& current, int64_t demuxerId, int iS
     if(!stream || stream->disabled)
       return false;
 
+    if (current.type == STREAM_VIDEO)
```
FernetMenta: you have to admit that this is rather ugly. it won't solve the problem either because player can call SetVideoResolution at any time.

mapfau:
sure, I'm aware of it, IMO it's not much more ugly than before (except the additional if(), wich I'm also not a friend of).

Fact is: I need this information at start time of the stream.

FernetMenta:
it is much more ugly than before because it pulls the call to a higher level where it does not fit in and places an if around it. its addon logic that you trying to fix. doing this here is wrong. think in objects and state. set the state on enable and finish the action when you got the desired resolution

FernetMenta:
what does the addon do if player calls this midsream whenthe user changed window size?

mapfau:
Currently I'm only saving the values (and use them to find an matching video stream at Enable).
In (near) future I'll always check these values on fragment change (~all 2 seconds) and maybe adjust the stream to a higher / lower resolution.

FernetMenta:
from a logical pov it makes most sense to move this to OpenDefaultStreams right before video stream is selected.

mapfau:
OK, I'll give it another try tomorrow, today there is unfortuantely no time left

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -893,6 +893,13 @@ void CVideoPlayer::OpenDefaultStreams(bool reset)
 
   bool valid;
 
+  if (m_pDemuxer && m_SelectionStreams.Count(STREAM_VIDEO) > 0)
```
FernetMenta: i would drop the 2nd condition. demuxer may have video at a later time and the call does not harm if there are no video streams.

mapfau: ok, will push this evening

FernetMenta:
thanks
jenkins build this please

FernetMenta:
jenkins build this please

MartijnKaijser:
needs rebase at least

mapfau:
rebased...

PR #9945, opened 2016-06-11
added: allow input streams to register for protocol types

notspiff:
useful for e.g. https://github.com/notspiff/inputstream.rtmp


FernetMenta:
jenkins build this please

@mapfau do yoir addons require an update?

mapfau:
doesn't seem so, empty protcol list is valid, therefore everthing fine

PR #9954, opened 2016-06-11
[cmake] Allow building with Ninja generator

fetzerch:
Small fixes to enable building with Ninja. Tested on OSX and Linux.
The build with Ninja is only a little faster (couple of seconds for a full build), but Ninja has a few nice things like generating a dependency graph.

Might be interesting for: @hudokkow, @notspiff, @wsnipex.

### DIFF
--- a/project/cmake/modules/FindCrossGUID.cmake
+++ b/project/cmake/modules/FindCrossGUID.cmake
@@ -5,6 +5,7 @@ if(ENABLE_INTERNAL_CROSSGUID)
   list(GET CGUID_VER 0 CGUID_VER)
   string(SUBSTRING "${CGUID_VER}" 8 -1 CGUID_VER)
 
+  set(CROSSGUID_LIBRARY ${CMAKE_BINARY_DIR}/${CORE_BUILD_DIR}/lib/libcrossguid.a)
   externalproject_add(crossguid
                       URL http://mirrors.kodi.tv/build-deps/sources/crossguid-${CGUID_VER}.tar.gz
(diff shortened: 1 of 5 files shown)


fetzerch:
@AchimTuran: Works also on Windows now. I remember that you once tried that and ran into some issues. If you like, give it another try :).

project/cmake/modules/FindCrossGUID.cmake
```diff
@@ -18,7 +18,8 @@ if(ENABLE_INTERNAL_CROSSGUID)
                                     <SOURCE_DIR> &&
                                     ${CMAKE_COMMAND} -E copy
                                     ${CORE_SOURCE_DIR}/tools/depends/target/crossguid/FindCXX11.cmake
-                                    <SOURCE_DIR>)
+                                    <SOURCE_DIR>
+                      BUILD_BYPRODUCTS ${CMAKE_BINARY_DIR}/${CORE_BUILD_DIR}/lib/libcrossguid.a)
```
wsnipex: maybe use a var for the lib here and in CROSSGUID_LIBRARIES below?
Same for dvd libs

fetzerch: Oh sure, indeed.

AchimTuran:
Awesome. Thanks I will test it during the week. I guess it could speed up our jenkins builds.

fetzerch:
Thanks @AchimTuran for testing, I forgot the Effects11 lib (didn't clean my source dir, hence I didn't get any error).
Now it's working also on Windows.

PR #9959, opened 2016-06-12
[depends][win32] bump libmicrohttpd to 0.9.50

Paxxi:
this is to match #9956 
Files are uploaded just now so wait a while before triggering a build


Paxxi:
jenkins build this please

PR #9960, opened 2016-06-12
[xbmc][cleanup] Some minor cleanup of texture handling

Paxxi:
moved some includes into the cpp files.
Got rid of the Cleanup methods as feel ugly, might make sense for an object that's costly to create and can handle a partial reset but these classes have none of this so lets rely on the destructor instead and recreate when needed.

### DIFF
--- a/xbmc/guilib/TextureBundle.cpp
+++ b/xbmc/guilib/TextureBundle.cpp
@@ -18,15 +18,10 @@
  *
  */
 
-#include "system.h"
 #include "TextureBundle.h"
 
-CTextureBundle::CTextureBundle(void)
-{
-  m_useXBT = false;
-}
-
-CTextureBundle::~CTextureBundle(void)
+CTextureBundle::CTextureBundle()
+  : m_useXBT{false}
 {
 }
 
@@ -36,15 +31,14 @@ bool CTextureBundle::HasFile(const std::string& Filename)
   {
     return m_tbXBT.HasFile(Filename);
   }
-  else if (m_tbXBT.HasFile(Filename))
+
+  if (m_tbXBT.HasFile(Filename))
   {
     m_useXBT = true;
     return true;
   }
-  else
-  {
-    return false;
-  }
+
+  return false;
 }
 
 void CTextureBundle::GetTexturesFromPath(const std::string &path, std::vector<std::string> &textures)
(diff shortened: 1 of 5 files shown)


Paxxi:
jenkins build this please

MartijnKaijser:
jenkins build this please

MartijnKaijser:
jenkins build this please

Paxxi:
jenkins build this please
DO NOT MERGE, will squish once jenkins is happy

MartijnKaijser:
http://jenkins.kodi.tv/job/BuildMulti-PR/5146/

MartijnKaijser:
@Paxxi jenkins is happy

MilhouseVH:
Not sure if this is a skin issue, but this PR appears to be causing an issue with the Titan skin, see here: http://forum.kodi.tv/showthread.php?tid=269814&pid=2369497#pid2369497 - note the lack of "rounded corners" once this PR is merged. Later in the thread I produced a test build with this PR reverted which resolved the issue (rounded corners visible again).

xbmc/guilib/TextureBundle.cpp
```diff
@@ -18,15 +18,10 @@
...
 #include "TextureBundle.h"
 
-CTextureBundle::CTextureBundle(void)
-{
-  m_useXBT = false;
-}
-
-CTextureBundle::~CTextureBundle(void)
+CTextureBundle::CTextureBundle()
+  : m_useXBT{false}
```
FernetMenta: this looks not right

Paxxi: Whats not right? The destructor is in the header file

Paxxi:
Thanks @MilhouseVH, I'll investigate tomorrow

xbmc/guilib/TextureBundle.cpp
fritsch: Check {} vs ()

Paxxi: That is correct since c++11, also preferred since it breaks the build if you accidentally perform implicit conversion during init

fritsch: Nice. Cares for narrowing.
Am 06.07.2016 9:48 vorm. schrieb "Pär Björklund" <notifications@github.com>:

Paxxi:
This is a nasty one, managed to repro and the cause seems to be that it tries loading the texture bundle before CApplication::LoadSkin is called so no skin path is set which causes it to fail the load and revert back to using textures.xbt. Not sure why this PR triggered the issue though as nothing here should cause any changes to the startup. Possibly a race condition that's been waiting to happen and now triggered by these minor changes.
I'll ping you once I figure out a solution.

PR #9963, opened 2016-06-12
[xbmc][threads][cleanup] Get rid of homegrown thread local and use the built in language support

Paxxi:
probably also needs some xcode sync, need to see if this will build or fail on android as so many things have done previously


Paxxi:
jenkins build this please

Paxxi:
ignoring the stupid mistakes what do you think @FernetMenta ?

MrMC:
http://stackoverflow.com/questions/28094794/why-does-apple-clang-disallow-c11-thread-local-when-official-clang-supports/29929949#29929949

Paxxi:
arghh, why for the love of god do they mess with clang?
I don't like this new world where stuff works in VS but not other places, was much easier when you knew that if it worked in VS all other compilers had supported it for a few years.
thanks for the link @MrMC

FernetMenta:
if I interpret the docs right, thread_local makes every single thread we create have a copy of this variable. even those threads not involved get a copy. not sure if I like this.

Paxxi:
@FernetMenta we use it in 3 places for pointers so it's about 24bytes overhead per thread and as far as I can tell no different from current situation.
On Windows the TLS get's pre-allocated on thread creation so we pay no extra storage penalty there.
according to http://linux.die.net/man/3/pthread_key_create it MUST create the key in all active threads.

Thinking about it, it's really the only way it could work as the compiler/pthread can't know which thread might execute which part of the program so it can't create the key/slots for a specific thread as that could lead to situations where randomly a key would disappear and not just change value if a method is scheduled on a different thread.

Reading some more it might be possible to use __thread instead of thread_local on os X but I think it defeats the purpose of this PR so might as well close it until Apple reconsiders

MrMC:
iOS does not support __thread so buzz there too. There was a reason threads\ThreadLocal.h was added, that reason still holds true.

Paxxi:
MrMC yeah I incorrectly assumed that it would have been made unnecessary by now but apparently not

PR #9970, opened 2016-06-12
VideoPlayer: stop clock while display is lost

FernetMenta:
see title

@popcornmix ping

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDClock.cpp
+++ b/xbmc/cores/VideoPlayer/DVDClock.cpp
@@ -227,11 +227,6 @@ int CDVDClock::UpdateFramerate(double fps, double* interval /*= NULL*/)
   return rate;
 }
 
-double CDVDClock::GetRefreshRate()
-{
-  return m_videoRefClock->GetRefreshRate();
-}
-
 bool CDVDClock::GetClockInfo(int& MissedVblanks, double& ClockSpeed, double& RefreshRate) const
 {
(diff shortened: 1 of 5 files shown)


popcornmix:
This works for first video I play. Second video just hangs on start and I get these errors:
  ERROR: CDVDAudio::AddPacketsRenderer - timeout adding data to renderer

http://paste.ubuntu.com/17271029/

FernetMenta:
your log doesn't seem to be related to this change here. where do you see a relationship?

popcornmix:
I don't have an explanation. I have double checked - this PR stops me being able to play a second file.

In fact with "delay after refresh rate change" disabled, it hangs even on first play of the file:
http://paste.ubuntu.com/17273264/

xbmc/cores/VideoPlayer/DVDClock.cpp
```diff
@@ -100,14 +101,41 @@ double CDVDClock::GetVsyncAdjust()
...
+    else
+      m_speedAfterPause = DVD_PLAYSPEED_PAUSE;
+
+    SetSpeed(DVD_PLAYSPEED_PAUSE);
+    m_paused = true;
+  }
+  else if (!pause && m_paused)
+  {
+    SetSpeed(m_speedAfterPause);
+    m_paused = false;
```
FernetMenta: @popcornmix those 2 statements are in wrong order. SetSpeed won't start the clock in m_paused is true. I will correct it

FernetMenta:
jenkins build this please

popcornmix:
Updated PR looks good now.

PR #9973, opened 2016-06-13
mmalffmpeg: Add lock on buffer release and log message

popcornmix:
There was an occasional crash with software decode in
AllocateBuffer near end of file that appears to be due to
the missing lock in ReleaseBuffer function.

Also add a debug message for the (unusual) case of a buffer
being freed due to a change in resolution of the stream.


popcornmix:
jenkins build this please

popcornmix:
jenkins build this please

PR #9974, opened 2016-06-13
mmalrenderer: Add support for more formats

popcornmix:
Currently mmalrenderer only supports YUV420.
This adds support for additional formats, like RGB.
The new formats are not currently used,
but will be useful for retroplayer in the future.

PR #9975, opened 2016-06-13
 mmal_renderer: Use shared pointers to maintain the pool until the decoder has been destroyed

popcornmix:
This fixes an issue when PVR fast channel switch changes between a SW decoded channel
(e.g. MPEG-2 without a codec licence) and a HW decoded channel (e.g. H.264).

If the pool is destroyed on the reconfigure there may still be pictures from it in the render
queue which can crash if the pool has gone when they are processed.


popcornmix:
The first 3 commits are from #9974 and will disappear when that is merged.

popcornmix:
Jenkins is happy: http://jenkins.kodi.tv/view/All/job/BuildMulti-PR/5145/

PR #9976, opened 2016-06-13
VideoPlayer: expose stream player info to GUI

FernetMenta:
player info data for new a/v stream player info dialog. 
info exposed:

video:
- decoder name (string)
- pixel format of content, i.e. NV12 (string)
- currently active deinterlacing method (string)
- video dimensions (width and height) used by decoder (non cropped), i.e. 1920x1088 (int)
- video fps steam player video has detected (string)
- isHwDecoder (bool)
- DAR (display aspect ratio)

audio:
- decoder name
- channels as string, (FL,FL,RL,RR)
- sample rate
- bits per sample

@ronie @phil65 could you create an dialog for this?


phil65:
Sure sure. Any preference how to do it? A button in the VideOSD which opens a dialog containing the info?

FernetMenta:
> A button in the VideOSD which opens a dialog containing the info?

that would be nice

da-anda:
what about the keymap "o" we have atm? Should this be dropped then, as it's of no use for mortals anymore and devs can easily add it back via custom keymap?

@phil65 I don't think we need a new dialog, simply show the info in "metadata" OSD tab you already have in Estuary

FernetMenta:
showing it metadata is wrong simply because this info is no metadata. this is process information of streamplayers. in future there will be also process info for AE.

da-anda:
width, height and fps could be considered metadata. Also, that tab is currently almost empty, so adding this info wouldn't hurt. It could also contain signal strength, transponder etc for PVR (wherever that is atm, never used that). The label "meta data" might not suite anymore in that case, but we could also change the label.
If no additional info is added to the current "meta data" tab, I vote to remove it, as that tiny bit of info currently shown would easily fit in the regular info tab as well. But that's a different story.

Just my 2c, up to you guys.

akva2:
Shall we plan for the future and make it player (1).foo or the likes? For pip, etc.

phil65:
If this is just intended for debugging, it might also be an option to completely format all that info in core so that skinners only have to display an infolabel by using a  textbox control?

stefansaraev:
maybe look how was it was done in the old debug overlay (part of dialogfullscreen), there we still have some important pvr info (signal quality and so on, lost after debug overlay removal) that perhaps belong to this new dialog ?

FernetMenta:
@phil65 users want this info for happiness, not for debugging. I am not falling back to creating the same crappy infolabel dialog we had before. I see a new clean dialog provided by skins as the only option for moving forward.

phil65:
ok, was just an idea.

ksooo:
> in the old debug overlay (part of dialogfullscreen), there we still have some important pvr info (signal quality and so on, lost after debug overlay removal)

@FernetMenta seems that we have lost the pvr infos? Tried with both Estuary and Confluence - thus, not a skinning problem. Or how do I access the "old debug overlay (part of dialogfullscreen)" these days?

FernetMenta:
> thus, not a skinning problem

it is because skins need to provide the new dialog

ksooo:
> it is because skins need to provide the new dialog

 not sure we're talking about the same thing. I'm talking about pvr signal strength and the like.

screenshot from 2016-06-18 18-58-54

What speaks against showing this information like we always did when hitting "O" while in fullscreen video window?

ksooo:
@phil65 the code to display the pvr information is in Estuary's VideoFullScreen.xml. Do you have an idea why this does not work anymore? Cannot see how this should be related to @FernetMenta's removal of codec info...

FernetMenta:
@ksooo the dialog you referenced doesn't exist anymore. a new one has to be created

stefansaraev:
the (videofullscreen) dialog exists, it's just the codec info part that is forced hidden now @ksooo

ksooo:
> the (videofullscreen) dialog exists, it's just the codec info part that is forced hidden now @ksooo

 that's what i thought.  => https://github.com/xbmc/xbmc/blob/master/addons/skin.estuary/1080i/VideoFullScreen.xml#L40 unfortunately i don't have an idea why the pvr info is not shown anymore.

stefansaraev:
http://sprunge.us/eLbA will give you an idea what's going on.

ksooo:
@stefansaraev thanks. now things are clear to me.
@phil65 how about bringing pvr info back like this?

screenshot000

stefansaraev:
@FernetMenta excuse my ignorance, but not all of this is "static info". for example, at least video width/height may change during commercial breaks. no ?

does this PR guarantee that the info will be updated on the fly, or users have to close/re-open thew new dialog we are supposed to have ?

FernetMenta:
@stefansaraev compared to the other variables we have on the debug screen, that change every couple of ms, I consider this info as static. It will be updated by player when it changes.
Users can watch this dialog instead of the movie, wait for a stream change, and finally enjoy the changing values :)

stefansaraev:
well. that's not the point. also, looking at pvr stream quality info in realtime is important (no, users dont want this for happines, they want it for debugging!) 

so, it is a valid question IMO, if all moved to new dialog, will pvr stream info be realtime as before?

FernetMenta:
pvr signal quality is not related to this work here. it is something exposed by PVR, not VidoePlayer. Skinners are free to place this info where they think it fits best.

stefansaraev:
sure, sure.

ksooo:
@stefansaraev 

> but that's completely unrelated pvr issue (one of the many) ;)

 please file trac issues or it never happened.

ksooo:
> pvr signal quality is not related to this work here. it is something exposed by PVR, not VidoePlayer. Skinners are free to place this info where they think it fits best.

 Exactly.

stefansaraev:
> please file trac issues or it never happened.

trac is broken. so no chance. it was reported few times at least year ago on trac or forum, I dont remember. search it..

however, lets not spam here.

FernetMenta:
jenkins build this please

FernetMenta:
@popcornmix could you please make this work for RPi?

popcornmix:
Yes, will do tomorrow.

FernetMenta:
@popcornmix PAR and video dimensions have to be set by decoder

jenkins build this please

FernetMenta:
jenkins build this please

xbmc/GUIInfoManager.cpp
```diff
@@ -2108,6 +2108,21 @@ const infomap videoplayer[] =    {{ "title",            VIDEOPLAYER_TITLE },
...
+  { "videowidth", PLAYER_PROCESS_VIDEOWIDTH },
+  { "videoheight", PLAYER_PROCESS_VIDEOHEIGHT },
+  { "videofps", PLAYER_PROCESS_VIDEOFPS },
+  { "videohwdecoder", PLAYER_PROCESS_VIDEOHWDECODER },
+  { "audiodecoder", PLAYER_PROCESS_AUDIODECODER },
+  { "audiochannels", PLAYER_PROCESS_AUDIOCHANNELS },
+  { "audiosamplerate", PLAYER_PROCESS_AUDIOSAMPLERATE },
+  { "audiobitspersample", PLAYER_PROCESS_AUDIOBITSPERSAMPLE }
+};
+
```
popcornmix: Shouldn't PLAYER_PROCESS_VIDEODAR be in here?

FernetMenta: right, I missed this. thanks

FernetMenta:
jenkins build this please

If no objections, I will merge this next merge window. Then skinners can take over.

xbmc/GUIInfoManager.cpp
```diff
@@ -5320,6 +5336,14 @@ int CGUIInfoManager::TranslateSingleString(const std::string &strCondition, bool
           return videoplayer[i].val;
       }
     }
+    else if (cat.name == "player_process")
+    {
+      for (size_t i = 0; i < sizeof(player_process) / sizeof(infomap); i++)
+      {
+        if (prop.name == player_process[i].str)
+          return videoplayer[i].val;
```
Memphiz: that should read player_process[i].val - thats why skinners get the wrong info i guess ...

Memphiz: will do a pr in a min

Paxxi:
@FernetMenta this PR seems to have broken the VS test suite, trying to compile gives this nice error
```
5>C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\memory(1193): error C2027: use of undefined type 'CDataCacheCore'
5>  c:\code\kodi\xbmc\ServiceBroker.h(43): note: see declaration of 'CDataCacheCore'
5>  C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\memory(1192): note: while compiling class template member function 'void std::default_delete<_Ty>::operator ()(_Ty *) noexcept const'
5>          with
5>          [
5>              _Ty=CDataCacheCore
5>          ]
5>  C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\memory(1397): note: see reference to function template instantiation 'void std::default_delete<_Ty>::operator ()(_Ty *) noexcept const' being compiled
5>          with
5>          [
5>              _Ty=CDataCacheCore
5>          ]
5>  C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\memory(1227): note: see reference to class template instantiation 'std::default_delete<_Ty>' being compiled
5>          with
5>          [
5>              _Ty=CDataCacheCore
5>          ]
5>  C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\memory(1236): note: see reference to class template instantiation 'std::_Get_deleter_pointer_type<_Ty,std::default_delete<_Ty>>' being compiled
5>          with
5>          [
5>              _Ty=CDataCacheCore
5>          ]
5>  C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\memory(1279): note: see reference to class template instantiation 'std::_Unique_ptr_base<_Ty,_Dx>' being compiled
5>          with
5>          [
5>              _Ty=CDataCacheCore,
5>              _Dx=std::default_delete<CDataCacheCore>
5>          ]
5>  c:\code\kodi\xbmc\ServiceManager.h(69): note: see reference to class template instantiation 'std::unique_ptr<CDataCacheCore,std::default_delete<_Ty>>' being compiled
5>          with
5>          [
5>              _Ty=CDataCacheCore
5>          ]
5>C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\memory(1194): error C2338: can't delete an incomplete type
5>C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\memory(1195): warning C4150: deletion of pointer to incomplete type 'CDataCacheCore'; no destructor called
5>  c:\code\kodi\xbmc\ServiceBroker.h(43): note: see declaration of 'CDataCacheCore'
```

FernetMenta:
hmm. how should I know? it passed jenkins build.
what is the purpose of this test suite and why does it deal with CDataCacheCore?

hudokkow:
It seems it also broke linux: https://s3.amazonaws.com/archive.travis-ci.org/jobs/143012613/log.txt

Paxxi:
Jenkins doesn't run the test suite on every build so it's easy to break the tests. 
The issue is in initialization of the app for testing, it tries to create a CServiceManager and fails, it fails with the same error during normal build if removing the datacachcore include from application.cpp

MartijnKaijser:
There's one last hurdle that prevents from auto building testsuite on each PR as well which is possibly solved by 

https://github.com/xbmc/xbmc/pull/10070

FernetMenta:
confused. I'd be happy to fix, if anybody tells me what's wrong with test suite.

akva2:
The test suite bring up some services to test stuff. There is a missing include for DataCacheCore in ServiceManager.h which materializes in building the test application. Add the missing include or add an (empty) destructor with its code in the .cpp file (it is deleting the unique_ptr in the generated destructor and fails cause it doesnt have the class definition).-------- Opprinnelig melding --------
Fra: Rainer Hochecker <notifications@github.com>
Dato: 08.07.2016 19:49 (GMT+01:00)
Til: xbmc/xbmc <xbmc@noreply.github.com>
Ko: Arne Morten Kvarving <Arne.Morten.Kvarving@sintef.no>, Comment <comment@noreply.github.com>
Emne: Re: [xbmc/xbmc] VideoPlayer: expose stream player info to GUI (#9976)


confused. I'd be happy to fix, if anybody tells me what's wrong with test suite.

FernetMenta:
this may fix it: https://github.com/xbmc/xbmc/pull/10090

xbmc/GUIInfoManager.cpp
```diff
@@ -5320,6 +5336,14 @@ int CGUIInfoManager::TranslateSingleString(const std::string &strCondition, bool
           return videoplayer[i].val;
       }
     }
+    else if (cat.name == "player_process")
+    {
+      for (size_t i = 0; i < sizeof(player_process) / sizeof(infomap); i++)
+      {
+        if (prop.name == player_process[i].str)
```
KenV99: Perhaps prop.name here needs to be changed to lower case to allow it to be case insensitive as mentioned here: [http://forum.kodi.tv/showthread.php?tid=285695&pid=2390071#pid2390071](url)

FernetMenta: here we go: https://github.com/xbmc/xbmc/pull/10270

PR #9424, opened 2016-03-22
python: use kodi provided cert if available

FernetMenta:
fix ssl on systems having no OpenSLL, i.e. Pi and OSX

PR #9979, opened 2016-06-14
[python] use our provided CA cert if building with unified deps

wsnipex:
follow up from https://github.com/xbmc/xbmc/pull/9424

Not really well tested, specially the cmake side is untested due to lack of time.
@FernetMenta @hudokkow @fetzerch @ksooo


wsnipex:
jenkins build this please

MartijnKaijser:
jenkins build this please

FernetMenta:
imo too complex and does not fit every case. it is much easier to set the environment variable, if desired, before Kodi is started

wsnipex:
which you are still free to do.
Which case is not covered?

FernetMenta:
building kodi with depends is not sufficient for using kodi provided certificate. one may still wand to use her/his own cert. further, setting environment variables in code is an abuse of those.

wsnipex:
ok. You're welcome to come up with a "proper" solution.

PR #9983, opened 2016-06-15
overlayrenderer: Avoid null access

popcornmix:


popcornmix:
Fairly minor - I hit this after switching between different build methods.
I think perhaps the font hadn't been installed.
It was resolved with a clean build, but seems worth protecting against a null access.

@FernetMenta okay?

jenkins build this please

xbmc/cores/VideoPlayer/VideoRenderers/OverlayRendererGUI.cpp
```diff
@@ -159,6 +159,9 @@ void COverlayText::PrepareRender(const std::string &font, int color, int height,
   if (!m_layout)
     m_layout = GetFontLayout(font, color, height, style, fontcache, fontbordercache);
 
+  if (m_layout == NULL)
+    return;
```
FernetMenta: this should not happen, right? why not logging an error?

popcornmix:
@FernetMenta Added log message for failure.

FernetMenta:
thanks

PR #9984, opened 2016-06-15
Add python scrapers

notspiff:
I saw this mentioned in some other PR so i figured i would re-offer my implementation.

This adds python based scrapers. I have rebased it and slightly modernized it and it appears to still work fine on the surface at least.

Sure, this is far from optimal. But it has one very obvious strength: it works within the limits set by the current system. There are of course a bunch of things that can be done more sane if XML support is removed (batch episode processing comes to mind, and converting to XML so some classes can parse is kinda stupid) . The idea was that this allows a seamless transition.

### DIFF
--- a/addons/metadata.demo.movies/demo.py
+++ b/addons/metadata.demo.movies/demo.py
@@ -1,0 +1,116 @@
+#!/usr/bin/env python
+# -*- coding: UTF-8 -*-
+
+import xbmcplugin,xbmcgui,xbmc,xbmcaddon
+import os,sys,urllib
+
+def get_params():
+        param=[]
+        paramstring=sys.argv[2]
+        if len(paramstring)>=2:
+                params=sys.argv[2]
+                cleanedparams=params.replace('?','')
+                if (params[len(params)-1]=='/'):
+                        params=params[0:len(params)-2]
+                pairsofparams=cleanedparams.split('&')
+                param={}
+                for i in range(len(pairsofparams)):
+                        splitparams={}
+                        splitparams=pairsofparams[i].split('=')
+                        if (len(splitparams))==2:
+                                param[splitparams[0]]=splitparams[1]
+                                
+        return param
+
+
+params=get_params()
+
+action=urllib.unquote_plus(params["action"])
+
+if action == 'find':
+    year = 0
+    title=urllib.unquote_plus(params["title"])
+    try:
+        year=int(urllib.unquote_plus(params["year"]))
+    except:
+        pass
+
+    print 'Find movie with title %s from year %i' %(title, int(year))
+    liz=xbmcgui.ListItem('Demo movie 1', thumbnailImage='DefaultVideo.png', offscreen=True)
+    liz.setProperty('relevance', '0.5')
+    xbmcplugin.addDirectoryItem(handle=int(sys.argv[1]), url="/path/to/movie", listitem=liz, isFolder=True)
+    liz=xbmcgui.ListItem('Demo movie 2', thumbnailImage='DefaultVideo.png', offscreen=True)
+    liz.setProperty('relevance', '0.3')
+    xbmcplugin.addDirectoryItem(handle=int(sys.argv[1]), url="/path/to/movie2", listitem=liz, isFolder=True)
+elif action == 'getdetails':
+    url=urllib.unquote_plus(params["url"])
+    if url == '/path/to/movie':
+        liz=xbmcgui.ListItem('Demo movie 1', offscreen=True)
+        liz.setProperty('video.original_title', 'Demo måvie 1')
+        liz.setProperty('video.sort_title', '2')
+        liz.setProperty('video.ratings', '1')
+        liz.setProperty('video.rating1.value', '5')
+        liz.setProperty('video.rating1.votes', '100')
+        liz.setProperty('video.user_rating', '5')
+        liz.setProperty('video.top250', '3')
+        liz.setProperty('video.unique_id', '123')
+        liz.setProperty('video.imdb_id', '456')
+        liz.setProperty('video.plot_outline', 'Outline yo')
+        liz.setProperty('video.plot', 'Plot yo')
+        liz.setProperty('video.tag_line', 'Tag yo')
+        liz.setProperty('video.duration_minutes', '110')
+        liz.setProperty('video.mpaa', 'T')
+        liz.setProperty('video.trailer', '/home/akva/Videos/porn/bukkake.mkv')
+        liz.setProperty('video.thumbs', '2')
+        liz.setProperty('video.thumb1.url', 'DefaultBackFanart.png')
+        liz.setProperty('video.thumb1.aspect', 'poster')
+        liz.setProperty('video.thumb2.url', '/home/akva/Pictures/hawaii-shirt.png')
+        liz.setProperty('video.thumb2.aspect', 'banner')
+        liz.setProperty('video.genre','Action / Comedy')
+        liz.setProperty('video.country', 'Norway / Sweden / China')
+        liz.setProperty('video.writing_credits', 'None / Want / To Admit It')
+        liz.setProperty('video.director', 'spiff / spiff2')
+        liz.setProperty('video.tvshow_links' ,'Demo show 1')
+        liz.setProperty('video.actors', '2')
+        liz.setProperty('video.actor1.name', 'spiff')
+        liz.setProperty('video.actor1.role', 'himself')
+        liz.setProperty('video.actor1.sort_order', '2')
+        liz.setProperty('video.actor1.thumb', '/home/akva/Pictures/fish.jpg')
+        liz.setProperty('video.actor1.thumb_aspect', 'banner')
+        liz.setProperty('video.actor2.name', 'monkey')
+        liz.setProperty('video.actor2.role', 'orange')
+        liz.setProperty('video.actor2.sort_order', '1')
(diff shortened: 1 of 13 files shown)


stefansaraev:
nice

libxml2 stays, as long libplist (requirment for airplay) stays, libxslt is another story

passing pure json data around (as json seems to be the future of scrapers) is next step, eventualy, I guess.

notspiff:
uhm, why would you do that. the whole point of the scraper is to translate to a common format.

stefansaraev:
right. slap me for talking too soon before reading

Razzeee:
Without looking at the code, oh how much I would love that feature! Thank you.

addons/metadata.demo.movies/demo.py
```diff
@@ -0,0 +1,116 @@
...
+        liz.setProperty('video.actor2.role', 'orange')
+        liz.setProperty('video.actor2.sort_order', '1')
+        liz.setProperty('video.actor1.thumb_aspect', 'poster')
+        liz.setProperty('video.actor2.thumb', '/home/akva/Pictures/coffee.jpg')
+        liz.setProperty('video.set_name', 'Spiffy creations')
+        liz.setProperty('video.set_overview', 'Horrors created by spiff')
+        liz.setProperty('video.tags', 'Very / Bad')
+        liz.setProperty('video.studio', 'Studio1 / Studio2')
+        liz.setProperty('video.fanarts', '2')
+        liz.setProperty('video.fanart1.url', 'DefaultBackFanart.png')
```
phil65: Wouldnt we need a 2nd prefix in order to indicate that something should get written into the art table? (so sth like "video.art.fanart" in this case)

notspiff: Dont follow.  All fanarts are stored based on the cfanart member afaik.
16. jun. 2016 09:37 skrev "Philipp Temminghoff" <notifications@github.com>:

garbear:
If this needs testing, I can cherry pick into my RetroPlayer branch. This excites me, because [Heimdall](https://github.com/topfs2/heimdall) is now a possibility. It'd be awesome if, when RetroPlayer drops in v18, we could include a full blown game library populated using Heimdall and this PR.

notspiff:
please give it a good run.

akva2:
i need a volunteer pythonista to do the scrapers. even though i could probably spit em out i do not want to maintain those on top of all the other addons in my ballpark...

ronie:
i wouldn't mind to create a music scraper, sounds like fun :-)

perhaps others can chime in too, iirc @phil65 showed some interest as well?

phate89:
I can help with tvshow/movie scraper addons for sure.. I think @phil65 already wrote some code for themoviedb for his extendedinfo addon.. I whink we need this mainly because noone knows in the  team anymore how old scrapers even work..

akva2:
great, thanks guys! branch is rebased, i tested the movie bit and it appeared to still worked fine. shout if anything's unclear from the examples, you see things that could be done better or whatever. i know the (new) unique_id stuff for video tags needs more work, just did the quickie for now.

akva2:
some easy whoopsies: 
1) the demo scrapers are not put into the build dir. copy them manually
2) remember to enable the scrapers using the browser (my favorite new functionality, it's so fun)

jjd-uk:
Maybe @olympia might have an interest in this as well.

Paxxi:
Looks good to me. Want to hear about the tests and then we can likely merge this and start cleaning up the scraper system internally.

phate89:
@notspiff I was looking into implementing a first scraper when I got a question.
With setinfo https://github.com/xbmc/xbmc/blob/master/xbmc/interfaces/legacy/ListItem.cpp#L333 we can already set all the infotag variables directly from python (at least with video files). you added new fileitem  properties that we need to import one by one. There's a reason to do this instead of using directly the infotag we can set directly from Python?

akva2:
i opted for info labels to keep one common interface. you can use the tags directly, but then you need to mix it with info labels for the stuffs that doesn't directly apply to the tag (thumbs et al).  no strong opinion on the approach as such, but that was the rationale.

phate89:
The main problem I found is that it becomes hard with complex fields and we have already specific functions. For example you can add  more than one rating but you can give them a name and set it to default too.
a thing like this is easier with a "setrating" function than with separate strings..
Some other opinion on this?

akva2:
well, it's very likely that i'd make the other choice if i were to develop the code today. none of those existed back when i took the decision, but i certainly see why you want to use them now.

phate89:
I understand that and it's not for sure a complaint :) it's something needed from a long time.. but since we just started I think we should start from the best place possible.
If everyone agree I can do the changes needed. But I just want to see if everyone think it's a positive change or not and worth the trouble.. @ronie @phil65 any opinion on this?

ronie:
sounds good. python offers multiple methods we could use

- setInfo()
- setRating()
- setCast()
- setArt()

setProperty() would still be needed for specific artist/album infolabels though...

phate89:
Set property is needed also video thumbs/fanarts right now since SetArt does a different thing (if I'm not wrong). Anyway could be a good moment to fill the Python API gaps..

MartijnKaijser:
setArt should exactly that and is supposed to replace setProperty
https://codedocs.xyz/xbmc/xbmc/group__python__xbmcgui__listitem.html#gad3f9b9befa5f3d2f4683f9957264dbbe

phate89:
I was referring to https://github.com/xbmc/xbmc/blob/master/xbmc/video/VideoInfoTag.h#L266
It's a xml that's separate than usual artworks. Not sure if it's used from something though..

phil65:
I would also be in favour of properly using VideoInfoTag etc instead of just using listitem properties. Ideally listitem creation would look the same as for plug-ins.
For most parts that shouldnt be hard to change, not sure what to do with fanartX.dim, fanartX.preview properties (and some more) though.. That stuff isnt part of our python API. Perhaps @notspiff could give us some pointers here.

notspiff:
well at its core the problem is that setArt() is meant to set the *active* artwork for an item, while we want to set the *available* artwork for an item. so either you have to extend setArt to allow the latter, or you have to add an additional binding.

Montellese:
Shouldn't the last commit have been removed before merging? ;-)

ronie:
@notspiff 

i'm working on python artist and album scrapers...
if the album scraper returns a musicbrainz artist id, kodi crashes

i *think* kodi tries to parse the artist scraper .py file as an xml file,
trying to find out if the artist scraper can search for an artist by musicbrainz id.  

http://paste.ubuntu.com/23921807/

notspiff:
Yip analysis is correct. Mb stuff was added after i did this code. Will fix
it up asap

4. feb. 2017 01:52 skrev "ronie" <notifications@github.com>:

ronie:
@DaveTBlake i'll open a thread on the forum where we can discuss to future of music scraping.
i only started looking into it a day ago and have a few questions myself.

would certainly love to hear your ideas/opinions as well.

xbmc/interfaces/legacy/AddonUtils.h
```diff
@@ -68,6 +69,9 @@ namespace XBMCAddonUtils
   };
 
 #define LOCKGUI XBMCAddonUtils::GuiLock __gl
+#define LOCKGUIIF(cond) std::unique_ptr<XBMCAddonUtils::GuiLock> __gl; \
```
FernetMenta: I really hate those crappy marcros. sorry, I couldn't resist :)
The problem here is that this crappy marco does not only was its name suggests. It also exits the GIL.

akva2: how is that a problem ? the code is right there to read, one time instead of many. the code would be just as crappy, only pasted all over, it if was not put in a macro.

FernetMenta: LOCKGUIIF(cond)
vs
XBMCAddonUtils::GuiLock _gl(cond)

I really don't see an advantage of a macro here. 

But this is not the actual problem. The ctor of GuiLock also exits the GIL. If you passed cond as an argument to the ctor, you could leave the gui lock but still exit gil.

akva2: it was 2 lines of code that had to be pasted all over. macro use was a no brainer. the whole point of this macro IS to lock the GIL. we have an on-screen listitem and it needs to not be accessed by any thread in the the python interpreter while we update it on the c++ side.

FernetMenta: LOCKGUI unlocks GIL as we do in every other python interface funtion. Now with offscreen use, GIL is not unlocked. Calling into Kodi without unlocking GIL can lead to deadlocks.

FernetMenta: another problem is that LOCKGUI on list items seems to be broken in general because there is no language hook in thread local storage.

Paxxi: Afaik all our python hooks call unblock_threads before calling into Kodi itself, trying to lock the Gil again here might be problematic. Testing will show i guess

FernetMenta: GIL is held here and should be unlocked, not locked. I will submit a PR soon.

jeryuni:
"liz.setProperty('video.date_added', '2016-01-01')" does not yet work in 'metadata.demo.movies / demo.py'.
Is this a bug?

PR #9986, opened 2016-06-16
[JSON] JSONServiceDescription is not thread safe

Tolriq:
This commit is not for merge as for the moment does not fix anything.
It's a base commit to have someone help me fix the problem as it's a little (totally) out of my competence.

Took me a lot of time to figure out and reproduce the issue but JSONServiceDescription is not thread safe during the build of it's definition cache.

If 2 JSON queries reach Kodi at the same time and tries to initialize the same description like list.fields.all then Kodi will crash.

Crash occurs in CVariant cleanup that tries to cleanup something that is already cleaned up.
So maybe CVariant should be thread safe I don't know, but for this big issue the problem is easy to trigger by JSON.

Since it's a cache initialization problem, crash can only occurs after a fresh start of Kodi then send the following query 2 or more times at the exact same time.

`{"jsonrpc":"2.0","id":1,"method":"Playlist.GetItems","params":{"playlistid":0,"properties":["title","thumbnail","fanart","rating","genre","artist","track","season","episode","year","duration","album","showtitle","playcount","file"]}}`

Happy to fix myself but I have 0 knowledge of C++ threads locks / mutex or whatever it is. So will need some assistance on this.

@Montellese @notspiff as behind the bear hides knowledge ;)

### DIFF
--- a/xbmc/interfaces/json-rpc/JSONServiceDescription.cpp
+++ b/xbmc/interfaces/json-rpc/JSONServiceDescription.cpp
@@ -40,6 +40,7 @@
 #include "FavouritesOperations.h"
 #include "TextureOperations.h"
 #include "SettingsOperations.h"
+#include <mutex>
 
 using namespace JSONRPC;
 
@@ -1154,11 +1155,13 @@ void JSONSchemaTypeDefinition::Print(bool isParameter, bool isGlobal, bool print
   }
 }
 
+std::mutex m;
+
 void JSONSchemaTypeDefinition::Set(const JSONSchemaTypeDefinitionPtr typeDefinition)
 {
   if (typeDefinition.get() == NULL)
     return;
-
+  m.lock();
   std::string origName = name;
   std::string origDescription = description;
   bool origOptional = optional;
@@ -1185,6 +1188,7 @@ void JSONSchemaTypeDefinition::Set(const JSONSchemaTypeDefinitionPtr typeDefinit
     referencedType = referencedTypeDef;
 
   referencedTypeSet = true;
+  m.unlock();
 }
 
 JSONSchemaTypeDefinition::CJsonSchemaPropertiesMap::CJsonSchemaPropertiesMap() :


Montellese:
Thanks. This has been reported before. I'll have to take a close look when this happens. I was under the impression that the whole JSONServiceDescription object/state is setup once during startup and then is only ever read from but it's been a long time since I wrote that code so I might not remember that correctly.

Tolriq:
I can help for some details if you want / need, just C++ multi thread is unknown to me :(
And most of my quick reading talks about boost that you try to remove from Kodi seeing some other PRs.

You can generate the crash relatively easily with Yatse in fact as there's a missing return on my side that triggers the 2 queries at the same time.

Start Kodi fresh, On fresh Yatse start an album then open left menu and click relatively quick on the image to get to the now playing, this case triggers the multi request at same time.

It's quite hard to reproduce by hand.


The Check does call Set to update for each unreferenced type set.
Sounds like just a matter of adding a lock to secure the writes.
Maybe just converting referencedTypeSet to atomic bool if that makes sense in C++.

Montellese:
This looks like a hack to me which is there because referenced types might not yet be fully defined when parsing a reference. But that might be solvable by going through the whole tree and fixing that after everything has been added to `CJSONServiceDescription` instead of doing it when needed during a method call. It will take more time during startup but it will then not require any locks during concurrent JSON-RPC requests (which might slow down the requests).

Tolriq:
There was the problem about order definition / recursivity.

But when reading log of V17 startup :

```
16:01:05 T:20548   DEBUG: JSONRPC: JSON schema type definition references an unknown type Setting.Details.Setting
16:01:05 T:20548 WARNING: JSONRPC: Could not parse type "Setting.Details.SettingList"
16:01:05 T:20548    INFO: JSONRPC: Adding type "Setting.Details.SettingList" to list of incomplete definitions (waiting for "Setting.Details.Setting")
16:01:05 T:20548    INFO: JSONRPC: Resolving incomplete types/methods referencing Setting.Details.Setting
16:01:05 T:20548    INFO: JSONRPC v7.15.1: Successfully initialized
```

This sounds like it's already correctly handled. Maybe it's now just a missing cleanup of that previous runtime definition update ?

Montellese:
I don't think so. This code for handling recursive type definitions was the last functionality added to `CJSONServiceDescription`.

What happens in above example is: first we try to parse a reference to a type (in above case `Setting.Details.Setting`) that hasn't been parsed yet. So we remember that we couldn't parse `Setting.Details.SettingList` because of the missing reference. Then when we parse `Setting.Details.Setting` we realize that it caused `Setting.Details.SettingList` to not be properly parsed so after parsing `Setting.Details.Setting` we try parsing `Setting.Details.SettingList` again which now succeeds.

I'll have to check how that behaves for recursive type definitions. It feels like running an extra loop over all types and executing the call to `JSONSchemaTypeDefinition::Set()` where necessary should allow us to remove that call from `JSONSchemaTypeDefinition::Check()` which should also get rid of the concurrency issues.

Tolriq:
I have tested to remove the call to Set in Check and all still works, but I have not tested all possible JSON commands.

And from the PR and VS says the Set call is only made from that Check.
So I'm not sure what this really address.

Anyway this is your part maybe throwing idea is not that useful.

If you need any test or anything just tell. 
The fact that the issue is now found and clear allows me to workaround on remote side to mitigate it a lot that's already cool seeing how long this one lasted without being able to reproduce / understand.

arnova:
@Montellese : i think this is the same as my issue mentioned in http://trac.kodi.tv/ticket/16656

Tolriq:
@arnova I was looking for someone to be able to reproduce and give logs so many times that I gave up on users :( Too bad you did not ping me so I had a look :(

This is the same issue, due to a missing return in Yatse, the crash case was much more often possible with it but it can happens with any JSON consumer during the first "second" build of the descriptions.

arnova:
@Tolriq : Sorry I didn't bother informing you. Will do next time. The reason I didn't is because it was pretty obvious to me the issue was on "our" side.

arnova:
@Montellese : Any update on this? Would be nice to fix this for v17.

Montellese:
@arnova: Unfortunately something is broken with my Visual Studio build environment. In 90% of the cases when starting Kodi through the debugger it crashes during startup with a memory access violation. The location where it crashes changes with every start so something must be wrong with my setup (a clean build didn't help). That makes it very tedious to hunt bugs because I have to restart the debugger several times before Kodi manages to start without an exception.

PR #9991, opened 2016-06-17
Use year in label2 when sorting by year

phate89:
Since there's not always a  valid premiered date (at least for now) it's better to use year for label2 when sorting to get a nicer result. skinners can always use premiered field if they want to


phate89:
jenkins build this please

stefansaraev:
jenkins build this please

PR #9400, opened 2016-03-19
Only use surface rendering for content > 1080p

razzeee:
This changes behavior for android codec handling.
For content smaller than 800 in width we will use software decoding.

And use surface rendering for anything HD, if wanted by the user.
It builds fine on each platform, http://jenkins.kodi.tv/view/Helpers/job/BuildMulti-All/1259/

Thanks to @fritsch for showing me the ropes.

@mazeyx please test http://mirrors.kodi.tv/test-builds/android/arm/kodi-20160319-24736ff-android-surface-rendering-armeabi-v7a.apk and let us know if it fixes your problem.

PR #9993, opened 2016-06-17
[aml] amcodec is now linux only

stefansaraev:
shame and blame on me, Ive been asked to.

fyi @codesnake @wrxtasy @mapfau 

EDIT: and I also share the general opinion that android should be mediacodec only!


codesnake:
Very sad, but I doubt I can do anything.

MartijnKaijser:
jenkins build this please 

That means that amlogic should make its firmware work according to MediaCodec standards.

MartijnKaijser:
jenkins build this please

codesnake:
@MartijnKaijser This is not related to MediaCodec implementation, but the way OSD and video display is implemented on Amlogic.

da-anda:
@codesnake sorry for not knowing, but what does it mean @codesnake? A 720p movie will be upscaled to 1080p on a 1080p GUI? Or what did you mean with "always upscaled"

codesnake:
@da-anda On Amlogic's Android GUI is always upscaled for highest supported resolutions. 1080p-capable SoCs ususally have 1280x720 framebuffer for 1080p/i resolutions, and 4K-capable SoCs have 1920x1080 framebuffer for 4K resolutions.

da-anda:
@codesnake and I assume with Mediacodec the videos are rendered to GUI framebuffer? So first downscaled and then again upscaled? That's ofc not nice, but I suppose all other apps (Youtube, Netflix, ...) suffer the same issue

MrMC:
Mediacodec can go two ways, a surface of the same size as GUI, or one that matches the actual display (S). Mediacodec(S) is the preferred method and will render 4k content on 4k displays without intermediate scaling. Mediacodec(S) does have certain device quirks, see the other PR that refs these quirks.

MartijnKaijser:
So in short https://github.com/xbmc/xbmc/pull/9400 should be updated and merged and that would "fix" several cases. @codesnake it would certainly be great if some one could help get that finished as razzeee is not really at home on android.

wrxtasy:
Observations:

Just tested mediacodec only on Android AML, Interlaced TV viewing defaults to ff-mpeg2. ff-h264 has no chance of playing properly for 1080i stuff. Interlacing is pretty stuffed as well, hundreds of dropped frames.
This breaks core Kodi TV viewing functionality on AML Android with libamcodec removed.

[b]@codesnake[/b] Do AMLogic themselves need to fix this for their AML Android implementation of mediacodec ?

MartijnKaijser:
@wrxtasy is that something amlogic should fix or is it in our code? I'm still very much in favour on dropping amlogic specific code for Android. It's currently already build without.

MartijnKaijser:
Another reason to just nuke it 
http://forum.kodi.tv/showthread.php?tid=280482&pid=2362698#pid2362698

codesnake:
@da-anda, @MrMC MediCodec on Amlogic uses EGL to render video frames, so it renders to framebuffer. And because on Amlogic's Android framebuffer is upscaled for 4K resolutions, 4K video will also be upscaled. Correct me if I'm wrong.

fritsch:
There is two modes. EGL and Surface. The latter does not use the EGL path.

MartijnKaijser:
tik tok.... clocks ticking. Is there any way we can still detect if device is using amlogic passthrough method and keep that part while still dropping the rest?

codesnake:
@MartijnKaijser I didn't quite understood what you mean in particular "detect if device is using amlogic passthrough".

MartijnKaijser:
amlogic chips currently use IEC passthrough method while standard android uses raw packages. In Android N standard android will switch to IEC as well (for us at least). So if we could keep current IEC passthrough working for amlogic by some hardware detection without the rest of the code that be be "nice".

stefansaraev:
mediacodec (and android, in general) support is not in good shape (even on devices where mediacodec is supposed to be doing right) at all. so I changed my mind, I dont want to be part of this removal, we broke too much in past one year, we always tell others what they are supposed to do, but neither we are doing it right..

I'll close this for now, if you guys want, pick the commits and re-submit this PR

wrxtasy:
I have to agree with you stefan, amcodec removal would break high quality deinterlacing, 4K video and likely more I have not even discovered yet for AML on Krypton. Not to mention it would very likely be an extremely unpopular move for the AML Android Kodi usersbase, regardless of mediacodec support status and dev. progress by AMLogic themselves.

MartijnKaijser:
I also don't see the companies that profit with AMLogic getting of their ass and improve default MediaCodec (or even AML)

stefansaraev:
my opinion that android should be mediacodec only is not changed. I just say we broke too much, and we dare to break even more without even having an android maintainer right now. companies that profit is another story. as already said, we wont get any help from MXQ people, this feels like punishing the wrong people (legit companies, that, btw, can profit even without kodi)

NedScott:
Don't blame AMLogic for what eBay sellers do. It's probably the best SoC for getting the best picture quality on Android. AMLogic hardware video decoding was there before mediacodec was a realistic option. The plan was always to wait for mediacodec to mature before removing other options. Stagefright got retired when mediacodec caught up with it. However, it has not caught up with amlcodec.

It will eventually be removed. Rushing this only makes Kodi look foolish, breaks the good boxes and companies who *do* help the Kodi project and contribute code, and does nothing to stop crappy box sellers. So just wait until the program is ready., and then everybody wins.

stefansaraev:
@wrxtasy @NedScott this still does not change the fact, that official kodi v17 on play store will be mediacodec only, just fyi.

MartijnKaijser:
Fact: no company is allowed to release a modified Kodi version still having the Kodi name without our explicit approval.
Fact: MediaCodec is mature and even more so in Android 6 (and 7). It just lacks adjustments/fixes on our side and proper firmware from manufacturers. Heck nvidia broke theirs appearently with their 3.1 > 3.2 upgrade judging from the user's posts. 
Fact: You got an old device then stick to v16 cause no one forces you to upgrade. Remember the no strings attache clause that always holds.
Fact: AML did their own hacks and doesn't follow the API standards. Should they do that then there's absolutely no problem.
Fact: Patches only land towards OpenELEC/LibreELEC and get dragged along forever and very rarely they get send to the one and only upstream which is here.

@wrxtasy you are wrong cause 4K does work on MediaCodec.

NedScott:
To clarify, I don't see this as much of a big deal, since forks like SPMC and WeTek's and Minix's forks will be able to cover these bumps in the road. Most videos still play fine on mediacodec for amlogic hardware.

However, I do think it only serves to damage Kodi's reputation in a functional sense to remove this this before MediaCodec is up to par. This will punish PVR users and 4K users, but the clone boxes won't really be hurt. Pirate streams are using 720/1080 h.264, and mediacodec on amlogic boxes play h.264 very well.

The good users, the ones we want, are the ones that are being pushed away to forks. Don't cut off your nose to spite your face.

NedScott:
@MartijnKaijser

I'm sorry if I gave you the wrong impression. Everyone knows that mediacodec will be the codec path for the future (of Android), but they also knew that XBMC/Kodi needed help before it was fully ready. That's why we had amcodec and libstagefright. It's okay to retire amcodec, but it feels like dropping amcodec is being rushed. @stefansaraev said it better than I could, and I completely agree with what he said.

If you're talking about sponsors with "no strings attached", then I'm a bit confused. There's no issue that I can see with sponsors. I can't speak for Minix, but WeTek has nothing to worry about. WeTek has their own (always renamed) fork that is pre-installed. Our PVR users typically don't use Android, because there's no Kodi-compatible DVR software for Android. Everyone else will most likely be fine with what vanilla Kodi + MediaCodec can do. 

If anything, WeTek and Minix have a lot to gain from amcodec being dropped. They are more likely to get the eventual firmware updates that make mediacodec run smoothly, while the crappy boxes won't :)

I get a little fired up, as you know, but my comments here are just honest concern for Kodi. However, it's probably not a big deal in the long run, so I'm sorry for the noise.

PR #9995, opened 2016-06-18
[depends] remove libmodplug as it's now a binary add-on

MartijnKaijser:


MartijnKaijser:
@FernetMenta @fritsch ping

FernetMenta:
looks good to me

PR #9998, opened 2016-06-18
Fix misc compiler warnings.

ksooo:
Namely:

screen shot 2016-06-18 at 11 05 37

@paxxi for review.


Paxxi:
Looks good.

xbmc/network/WebServer.cpp
```diff
@@ -974,7 +974,11 @@ int CWebServer::ContentReaderCallback(void *cls, size_t pos, char *buf, int max)
     return -1;
 
   if (g_advancedSettings.CanLogComponent(LOGWEBSERVER))
+#if (MHD_VERSION >= 0x00090200)
```
Paxxi: Semi related, @wsnipex are there any platform that has mhd older than this? Got me thinking if we can get rid of a few ifdefs related to mhd

ksooo:
jenkins build this please

ksooo:
OSX jenkins error is unrelated.

xbmc/network/WebServer.cpp
wsnipex: you can drop support for the older versions

MartijnKaijser: @wsnipex what should be minimum version? Our depends currently has 0.9.50

wsnipex: I'd keep the one defined here

PR #9208, opened 2016-02-25
Settings layout changes improvements

jjd-uk:
This is a refactor of the layout of Settings inspired by the discussions on http://forum.kodi.tv/showthread.php?tid=189031

For Team members I've open a thread in our forums so discussion of the details can be made there to avoid people getting bombarded with too much github spam, as I imagine this will generate a lot of discussion.

To make sure I've not messed up the platform specific xmls (as this has only been tested on Windows) and any other comments @Memphiz @popcornmix @koying @Paxxi @FernetMenta 

Also for UI apsects @da-anda @ronie @phil65 @HitcherUK

PR #9999, opened 2016-06-18
[CHG] Settings - Library: groupings

jjd-uk:
Change the labelling of groupings within the Settings - Library sections and adjust the allocated settings in each group to suit the new labels.

Follow up to https://github.com/xbmc/xbmc/pull/9208

This is the 1st of 3 follow up PR's I want to make, the next 2 rely on this.

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -6544,7 +6544,7 @@ msgstr ""
 
 #: system/settings/settings.xml
 msgctxt "#13414"
-msgid "Include artists who appear only on compilations"
+msgid "Show song and album artists"
 msgstr ""
 
 #: system/settings/settings.xml
@@ -17183,10 +17183,10 @@ msgctxt "#36254"
 msgid "Action to perform when pressing the record button. [Record current show] will record the current show from \"now\" to the end of the show. If no TV guide data is currently available a fixed length recording starting \"now\", with the value set for \"Instant recording duration\" will be scheduled. [Record for a fixed period of time] will schedule a fixed length recording starting \"now\", with the value set for \"Instant recording duration\". [Ask what to do] will open a dialogue containing different recording actions to choose from, like \"Record current show\", \"Record next show\" and some fixed duration recordings."
 msgstr ""
 
-#. Description of setting with label #13414 "Include artists who appear only on compilations"
+#. Description of setting with label #13414 "Show song and album artists"
 #: system/settings/settings.xml
 msgctxt "#36255"
-msgid "Determine if artists that appear only on compilations are shown in the library artist view."
+msgid "When enabled, both song and album artists are shown. When disabled, only album artists are shown and artists that appear only on individual songs from an album are excluded."
 msgstr ""
 
 #. Description of setting with label #20192 "Fetch additional information during updates"
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -848,18 +848,20 @@
   </section>
   <section id="library" label="14202" help="38101">
     <category id="filelists" label="16000" help="36121">
-      <group id="1" label="14240">
+      <group id="1" label="593">
         <setting id="filelists.showparentdiritems" type="boolean" label="13306" help="36122">
           <level>1</level>
           <default>true</default>
           <control type="toggle" />
         </setting>
-        <setting id="filelists.showextensions" type="boolean" label="497" help="36123">
+        <setting id="filelists.ignorethewhensorting" type="boolean" label="13399" help="36124">
           <level>1</level>
           <default>true</default>
           <control type="toggle" />
-        </setting>
-        <setting id="filelists.ignorethewhensorting" type="boolean" label="13399" help="36124">
+        </setting>		
+      </group>
+      <group id="2" label="744">
+        <setting id="filelists.showextensions" type="boolean" label="497" help="36123">
           <level>1</level>
           <default>true</default>
           <control type="toggle" />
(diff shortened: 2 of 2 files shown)


jjd-uk:
@MartijnKaijser some minor changes to the groupings following feedback

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -6544,7 +6544,7 @@ msgstr ""
 
 #: system/settings/settings.xml
 msgctxt "#13414"
-msgid "Include artists who appear only on compilations"
+msgid "Show song and album artists"
```
jjd-uk: This was requested by @DaveTBlake

jjd-uk:
jenkins build this please

system/settings/settings.xml
```diff
@@ -100,7 +100,7 @@
         </setting>
         <setting id="videoplayer.hqscalers" type="integer" parent="videoplayer.rendermethod" label="13435" help="36154">
           <level>2</level>
-          <default>20</default>
+          <default>0</default>
```
MartijnKaijser: Wasn't there something with this @fritsch?

DaveTBlake:
Thanks for this, it makes me so happy see this corrected.

I have had one tiny second though, does "When enabled, both song & album artists are shown. When disabled, only album artists are shown and artists that appear only on individual songs are excluded" sound like singles artists are not shown?

Singles get treated like a fake albums, hence the singles artists do get listsed when albumartistsonly is enabled. So maybe we need to say
"When enabled, both song & album artists are shown. When disabled, only album artists are shown and artists that appear only on individual songs *from an album* are excluded"

system/settings/settings.xml
jjd-uk: just spotted that correcting now

fritsch: Yeah 0 makes no sense, please keep 20.

MartijnKaijser: Perhaps we can add a minimum of 20?

jjd-uk:
@DaveTBlake updated help string

jjd-uk:
These are hopefully minor uncontroversial changes that I'd like to get in for Alpha 2.

I have another set of changes lined up will dealing with changing setting defaults, labels etc, those will be PR'd separately since those changes have more potential to cause some discussion, and didn't want these changes held up.

jjd-uk:
jenkins build this please

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -17183,10 +17183,10 @@ msgctxt "#36254"
 msgid "Action to perform when pressing the record button. [Record current show] will record the current show from \"now\" to the end of the show. If no TV guide data is currently available a fixed length recording starting \"now\", with the value set for \"Instant recording duration\" will be scheduled. [Record for a fixed period of time] will schedule a fixed length recording starting \"now\", with the value set for \"Instant recording duration\". [Ask what to do] will open a dialogue containing different recording actions to choose from, like \"Record current show\", \"Record next show\" and some fixed duration recordings."
 msgstr ""
 
-#. Description of setting with label #13414 "Include artists who appear only on compilations"
+#. Description of setting with label #13414 "Show song and album artists"
 #: system/settings/settings.xml
 msgctxt "#36255"
-msgid "Determine if artists that appear only on compilations are shown in the library artist view."
+msgid "When enabled, both song & album artists are shown. When disabled, only album artists are shown and artists that appear only on individual songs from an album are excluded."
```
MartijnKaijser: Might just wanna use  ....song and album artists....  instead of the & 
Not sure if that works in all languages

jjd-uk: sure

system/settings/settings.xml
```diff
@@ -1019,17 +1021,7 @@
...
-          <level>2</level>
-          <default>true</default>
-          <control type="toggle" />
-        </setting>
-        <setting id="musiclibrary.showcompilationartists" type="boolean" label="13414" help="36255">
-          <level>0</level>
-          <default>true</default>
-          <control type="toggle" />
-        </setting>
+	  <group id="1" label="593">
```
MartijnKaijser: fix indenting which was already wrong before

system/settings/settings.xml
```diff
@@ -2249,7 +2253,7 @@
         </setting>
         <setting id="audiooutput.ac3transcode" type="boolean" parent="audiooutput.ac3passthrough" label="667" help="36429">
           <level>2</level>
-          <default>false</default>
+          <default>true</default>
```
MartijnKaijser: @fritsch @FernetMenta ping

fritsch: What's the intention behind it?

It can't overrule multichannel setting, it would transcode to ac3 when user has 2 channel layout and passthrough enabled though. Is this the intention?

jjd-uk: sorry that was supposed in the next PR

jjd-uk: Yes that was the intention.

When a user has 2 channel enabled this and dobly digital equipment this ensure they get 5.1

fritsch: Yes. Makes sense.

jjd-uk: Can drop it from this PR if it requires discussion and move to follow as I was planning.

fritsch: @FernetMenta what do you think? @popcornmix this needs a special xml for the Pi1, right out of performance reasons?

jjd-uk: That was the one thing I wasn't sure on, is what performance impact if any would be on RPi or any other low power platforms, other than that couldn't see any issues in my testing.

MartijnKaijser:
The rest are all shuffles correct? Maybe drop the ac3 change and just merge this to start with so you can start the next.

jjd-uk:
Dropped transcoding change, this will go in a follow up as it should have been.

@MartijnKaijser Yes the rest is just shuffles

MartijnKaijser:
@jjd-uk seems at least rpi doesn't like this shuffle as it had set other values. Maybe check the rest as well

jjd-uk:
Bugger, a conflict wasn't thrown up but I see problem. Doubled checked the other and it's just the RPi affected by the shuffle.

jjd-uk:
Fix https://github.com/xbmc/xbmc/pull/10003

PR #10004, opened 2016-06-19
[addons] - fix folderpath in case only one repo is available

phil65:
Folderpath seems to be wrong atm when only one repo is installed.
@tamland is this the correct place to fix it or should it be done inside Browse()?


tamland:
> or should it be done inside Browse()?

would fit better there, yes, as this is just a redirect

phil65:
oops.
@tamland, updated.

PR #10006, opened 2016-06-19
[fileutils] allow remote access for addons://

phil65:
This would allow some new nice script features like "addons by same author".
Not sure if this is considered insecure because of addons://install/ ? @Montellese might know. :)

### DIFF
--- a/xbmc/utils/FileUtils.cpp
+++ b/xbmc/utils/FileUtils.cpp
@@ -140,7 +140,7 @@ bool CFileUtils::RemoteAccessAllowed(const std::string &strPath)
     return true;
   else if (StringUtils::StartsWithNoCase(realPath, "special://profile/addon_data"))
     return true;
-  else if (StringUtils::StartsWithNoCase(realPath, "addons://sources"))
+  else if (StringUtils::StartsWithNoCase(realPath, "addons://"))
     return true;
   else if (StringUtils::StartsWithNoCase(realPath, "upnp://"))
     return true;


Tolriq:
At the time I udpated this function it was said that this could give access to some addons details like api keys or things that should not be exposed.

Does not JSON allows listing those already ?

phil65:
> Does not JSON allows listing those already ?

No, that's exactly the reason why I put up this PR :)

Tolriq:
Well it was more a rhetorical question ;) 

JSON have "Addons.GetAddons" and "Addons.GetAddonDetails" this is how remotes list all addons and authors / properties and everything.

Question is more what is missing in those ?

phil65:
That's only for local addons right? I'd like to get a list of addons from the repo.

MartijnKaijser:
Shouldn't that be made a separate json call to get available Addons?

Tolriq:
Ho yes :(

But that would make a great JSON addition :) Specially if can be used to install addons without popups on Kodi. But then there's security matters :(

Anyway about addons:// I remember that there was big security concerns at the time I added that function.
And for addons://sources it needed arguments just to have sources added.

@MartijnKaijser do you think allowing remote install of Addons would be a security concerns or a feature ? (Considering there's currently no API level security on JSON side so activating web server would open this)

wsnipex:
imho remote addon install in general is a no go, especially without any user approval

MartijnKaijser:
Yes should certainly be a user approval before installing. Either with some pincode that is generated onscreen and when authorized it will allow remote to install within a certain period of time. Other option would be to permanently give a certain remote approval with a certain key that is stored on both sides. 
There are certainly ways to do it "secure" enough.

phil65:
Would it be fine to just whitelist "addons://repository.xbmc.org/" ?

Tolriq:
@MartijnKaijser yes there was discussion with @Montellese about those but never ended.

If it's just about listing those addons, what about just adding a param to getAddons like includeNotInstalled (or any better name ?) 

That I could add without problem I suppose.

MartijnKaijser:
Get available addons as extra parameter should be no issue imo

phil65:
Ok @Tolriq, will you take care of this? I will close then.

Tolriq:
See https://github.com/xbmc/xbmc/pull/10007 ;)

phil65:
Thank you! :+1:

PR #10007, opened 2016-06-19
[JSON] Add new parameter to Addons.GetAddons to retrieve repositories addons.

Tolriq:
Add new option installed to Addons.GetAddons.

Add installed field to Addon.Fields / Addon.Details to indicate installation status.


Tolriq:
@phil65 @Montellese for namings and concept validation.

phil65:
Naming seems fine to me, and +1 for concept of course. ;)

Tolriq:
Code formatting done, ready for merge if / when validated.

xbmc/interfaces/json-rpc/AddonsOperations.cpp
```diff
@@ -40,6 +40,7 @@ JSONRPC_STATUS CAddonsOperations::GetAddons(const std::string &method, ITranspor
   TYPE addonType = TranslateType(parameterObject["type"].asString());
   CPluginSource::Content content = CPluginSource::Translate(parameterObject["content"].asString());
   CVariant enabled = parameterObject["enabled"];
+  CVariant includeInstallable = parameterObject["includeInstallable"];
```
Montellese: We use lower case only property/parameter names.

Montellese:
I don't really know the use case so @phil65 should probably comment on this question: Is it ok that the available options are "only installed addons" and "nstalled and installable addons" and that there's no option for "only installable addons"?

@Tolriq: Have you checked if the resulting output contains installed addons twice because they are available as installed addons and as installable addons?

Tolriq:
@Montellese history of request : https://github.com/xbmc/xbmc/pull/10006

How would you see the last option ? Another option installed default true ? Leading to having to request installed false and includeInstallable true ?

Or a installed param boolean + all value like enabled ? ( I suppose this is better)

I did not thought of the others needs, just went fast to avoid the security problems.



When I did my tests GetInstallableAddons only returned non installed addons, but I will triple check later today.

MartijnKaijser:
By default it should give the same result as before: only return installed add-ons
With the extra option it should give installed+available for install add-ons. The only install-able add-ons could be filtered out using the "installed" property. No sure if the "only install-able addons" option has a big use case. If you see Play Store you also see all apps and the ones that you have installed have a checkmark or whatever.

Tolriq:
PR updated, went the installed road as it's more logical and consistent with the rest of that function (enabled)

Default result is of course same as before for compatibility.

And triple checked there's no duplicates as GetInstallableAddons does not returns installed ones.

phil65:
So to get all addons (remote + local), I need to make two separate JSON calls now, right?

Tolriq:
No no, you just set the installed to "all"

If set to false then only installable, if set to true only installed. It's the same as enabled.

All = all, true = enabled, false = disabled.

phil65:
ok, great. That sounds good to me.

MartijnKaijser:
jenkins build this please

@Montellese you ok with this implementation? If so we can get it still merged today.

phil65:
nitpick: commit message doesnt reflect latest changes.

xbmc/interfaces/json-rpc/AddonsOperations.cpp
```diff
@@ -79,19 +80,29 @@ JSONRPC_STATUS CAddonsOperations::GetAddons(const std::string &method, ITranspor
...
+      {
+        if (!installed.isBoolean() || installed.asBoolean())
+          CAddonMgr::GetInstance().GetInstalledAddons(typeAddons);
+        if (!installed.isBoolean() || (installed.isBoolean() && !installed.asBoolean()))
+          CAddonMgr::GetInstance().GetInstallableAddons(typeAddons);
+      }
+      else if (enabled.asBoolean() && (!installed.isBoolean() || installed.asBoolean())) //Enabled
         CAddonMgr::GetInstance().GetAddons(typeAddons);
-      else
+      else if (!installed.isBoolean() || installed.asBoolean())
```
Montellese: I'm not sure if this makes sense. There are no not-installed disabled add-ons so the question is whether the "installed" parameter makes any sense when "enabled" is false. But this way it is consistent.

Montellese:
Yup, I like the new approach better.

xbmc/interfaces/json-rpc/AddonsOperations.cpp
Tolriq: Yes this makes no sense to use that combination, but I choose to be respectful of the non sense and as such respect the params. This will obviously return no addons.

MartijnKaijser:
@Montellese can this be merged or do any additional changes need to be made?

Tolriq:
Bump on this one too. I'm out of computer in a few hours for a week, if changes are needed for this last merge window please tell fast, I'll be able to do fix for beta later of course if this is merged.

@Montellese @MartijnKaijser

MartijnKaijser:
jenkins build and merge

Montellese:
This would be fine to go in BTW.

PR #10008, opened 2016-06-19
[JSON] Fix JSON playlist handling

Tolriq:
That could result in duplicates added due to item path resolved to internal path.

See http://forum.kodi.tv/showthread.php?tid=279797

@DaveTBlake


DaveTBlake:
That looks nice and straight forward. Unable to test at the moment, but perhaps Colin will feedback. Does the JSON API need an increment?

Tolriq:
Tested here and fix the reported issue. Can't think of side effects.

About bump since I was not able to figure when the change / break occured it's hard to tell if it's needed.

But can bump if needed.

DaveTBlake:
jenkins build this please

DaveTBlake:
I'm not worried about bump if you aren't :)

Montellese:
I would bump the bugfix version anyway. The change looks ok. We really need to have a better concept concerning virtual paths and real paths.

Tolriq:
Version bumped

PR #10009, opened 2016-06-19
[xbmc][fix] Fix dll hooking for VS 2015 update 3 as there's some changes incoming

Paxxi:
From the header file
Prior to Visual Studio 2015 Update 3, these hooks were non-const.  They were
made const to improve security (global, writable function pointers are bad).
If for backwards compatibility you require the hooks to be writable, define
the macro DELAYIMP_INSECURE_WRITABLE_HOOKS prior to including this header and
provide your own non-const definition of the hooks.


Paxxi:

PR #10011, opened 2016-06-19
[settings] Adjustments to the location of settings and defaults

jjd-uk:
Some more setting shuffles and changing of defaults, also includes some minor cosmetic cleanup.

@MartijnKaijser these are the ones I thought might generate a bit more debate, no rush for these.

If anyone can think of any other setting defaults they think might need changing then ping me please.

Note - changing "Default select action" to "show information" was originally part of these changing, has been removed since @tamland drew to attention to the fact it can cause problems in File views. See http://forum.kodi.tv/showthread.php?tid=261542&pid=2360192#pid2360192

### DIFF
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -858,7 +858,7 @@
           <level>1</level>
           <default>true</default>
           <control type="toggle" />
-        </setting>		
+        </setting>
       </group>
       <group id="2" label="744">
         <setting id="filelists.showextensions" type="boolean" label="497" help="36123">
@@ -900,7 +900,7 @@
       </group>
     </category>
     <category id="video" label="14215" help="38107">
-	  <group id="1" label="593">
+      <group id="1" label="593">
         <setting id="myvideos.selectaction" type="integer" label="22079" help="36177">
           <level>0</level>
           <default>1</default> <!-- SELECT_ACTION_PLAY_OR_RESUME -->
@@ -929,8 +929,8 @@
           <default>true</default>
           <control type="toggle" />
         </setting>
-      </group>		
-	  <group id="2" label="744">
+      </group>
+      <group id="2" label="744">
         <setting id="myvideos.stackvideos" type="boolean" label="20435" help="36182">
           <level>1</level>
           <default>false</default>
@@ -941,7 +941,7 @@
           <default>true</default>
           <control type="toggle" />
         </setting>
-      </group>	
+      </group>
       <group id="3" label="14022">
         <setting id="videolibrary.showallitems" type="boolean" label="38011" help="38012">
           <level>2</level>
@@ -951,6 +951,16 @@
         <setting id="videolibrary.showunwatchedplots" type="boolean" label="20369" help="36141">
           <level>0</level>
           <default>true</default>
+          <control type="toggle" />
+        </setting>
+        <setting id="videolibrary.groupmoviesets" type="boolean" label="20458" help="36145">
+          <level>1</level>
+          <default>false</default>
+          <control type="toggle" />
+        </setting>
+        <setting id="videolibrary.groupsingleitemsets" type="boolean" parent="videolibrary.groupmoviesets" label="20470" help="36157">
+          <level>1</level>
+          <default>false</default>
           <control type="toggle" />
         </setting>
         <setting id="myvideos.flatten" type="boolean" label="20456" help="36183">
@@ -1003,23 +1013,13 @@
           </dependencies>
           <control type="spinner" format="string" />
         </setting>
-        <setting id="videolibrary.groupmoviesets" type="boolean" label="20458" help="36145">
-          <level>1</level>
-          <default>false</default>
-          <control type="toggle" />
-        </setting>
-        <setting id="videolibrary.groupsingleitemsets" type="boolean" label="20470" help="36157">
-          <level>1</level>
-          <default>false</default>
-          <control type="toggle" />
-        </setting>
         <setting id="videolibrary.actorthumbs" type="boolean" label="20402" help="36143">
           <level>2</level>
           <default>true</default>
           <control type="toggle" />
         </setting>
       </group>
-    </category>	
+    </category>
     <category id="music" label="14216" help="38108">
       <group id="1" label="593">
         <setting id="musicfiles.trackformat" type="string" label="13307" help="36275">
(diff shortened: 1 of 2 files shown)


system/settings/settings.xml
```diff
@@ -1008,18 +1008,21 @@
...
         </setting>
-        <setting id="videolibrary.groupsingleitemsets" type="boolean" label="20470" help="36157">
+        <setting id="videolibrary.groupsingleitemsets" type="boolean" parent="videolibrary.groupmoviesets" label="20470" help="36157">
           <level>1</level>
           <default>false</default>
           <control type="toggle" />
+          <dependencies>
+            <dependency type="enable" setting="videolibrary.groupmoviesets" operator="is">true</dependency>
+          </dependencies>
         </setting>
```
jjd-uk: I've attempted to simplify the wording here and as part of that made the "videolibrary.groupsingleitemsets" dependant on "videolibrary.groupmoviesets"

@Montellese  Are you ok with this since you introduced the setting? want to be sure there wasn't a reason for you to not do it this way in the first place.

system/settings/settings.xml
```diff
@@ -2173,43 +2178,20 @@
...
+            <allowempty>true</allowempty>
+          </constraints>
+          <updates>
+            <update type="change" />
+          </updates>
+          <control type="button" format="addon">
+            <show more="true" details="true">installed</show>
+          </control>
+        </setting>      
       </group>
```
jjd-uk: GUI sounds settings totally removed from "Interface -> Skin" and now appear in a new grouping in "System - Audio" since it's seems better to have all options governing sound in the one place. At the moment have left "GUI Sounds" enabled by default, can change to disabled if wanted.

system/settings/settings.xml
```diff
@@ -2322,6 +2304,43 @@
...
+          <dependencies>
+            <dependency type="visible">
+              <or>
+                <condition on="property" name="aesettingvisible" setting="audiooutput.dspaddonsenabled">audiooutput.dspresetdb</condition>
+              </or>
+            </dependency>
+          </dependencies>
+        </setting>
+      </group>
     </category>
```
jjd-uk: In System -> Audio, to accommodate the "GUI Sounds" grouping the "ADSP" grouping has been moved to the bottom since it's the most advanced of the audio features hence won't see as much widespread use.

system/settings/settings.xml
```diff
@@ -3188,7 +3181,7 @@
         </setting>
         <setting id="lookandfeel.enablerssfeeds" type="boolean" label="13305" help="36111">
           <level>0</level>
-          <default>true</default>
+          <default>false</default>
           <control type="toggle" />
```
jjd-uk: RSS changed to be disabled by default, this was from a conseus from a Team channel Slack/IRC conversion on Thur 16/06/2016.

system/settings/settings.xml
```diff
@@ -2253,7 +2235,7 @@
         </setting>
         <setting id="audiooutput.ac3transcode" type="boolean" parent="audiooutput.ac3passthrough" label="667" help="36429">
           <level>2</level>
-          <default>false</default>
+          <default>true</default>
           <dependencies>
```
jjd-uk: @fritsch yesterday's inclusion was a mistake, so here it is again.

@FernetMenta a ping to see if you are ok with this?
 
The intention of this is to make getting 5.1 audio easier for users of 2.0 channel configurations. At the moment when passthrough is enabled then only the Dolby Digital format is enabled, leaving the others such as DTS disabled. This means in a 2.0 configuration even though a file being played may contain 5.1 the user will only get PCM 2.0 audio for DTS or any of the non-Dolby Digital formats. With Transcoding set to be enabled by default this will mean DTS etc will be output as Dolby Digital 5.1 so the user gets the 5.1 audio with no other user selections other than just enabling passthrough and configuring the output device.

For RPi I've set AC3 transcoding to be off as I noticed the RPi only help statement about CPU resource. @popcornmix  can you confirm this can still cause CPU resource issues if enabled, and does it apply to both RPi and RPi2.

system/settings/settings.xml
jjd-uk: How it looks

image

jjd-uk: Reminder to self, must update string to be "Include sets containing a single movie"

jjd-uk: Hadn't noticed before that ""videolibrary.groupsingleitemsets" applied to sets node as well, so will remove the dependancy, so question is does this

image

or this

image

make more sense.

system/settings/settings.xml
zag2me: Hurray!

system/settings/settings.xml
jjd-uk: @FernetMenta would like to get your +1 on the transcode settings change.

system/settings/rbp.xml
```diff
@@ -95,6 +95,7 @@
       </group>
       <group id="3">
         <setting id="audiooutput.ac3transcode" help="37024">
+          <default>false</default>
         </setting>
```
jjd-uk: @popcornmix with enabling transcoding by default in the main settings.xml I've set it to be disabled in the appliance xml's for both RPi & RPi2, just wanted to confirm this is needed assuming the change for the setting.xml is accepted.

FernetMenta:
@jjd-uk I don't stop you from changing it but here my thoughts. I think there are many users who connect their HTPC via HDMI to a 2.0 setup. They choose passthrough because they are misguided and will suffer from lower quality by transcoding. I think the users with 5.1 setups are more inclined to read through the settings.

system/settings/settings.xml
popcornmix: Yes, AC3 transcode is too heavy for Pi1.
It is usable on Pi2/Pi3. Not sure about it being enabled by default.

system/settings/rbp.xml
popcornmix: Definitely want ac3 transcode disabled on Pi1.
I'm happy for Pi2/Pi3 to match whatever the default is chosen to be.

MartijnKaijser:
I think must agree with @fernetmenta on this.

jjd-uk:
The AC3 transcode change was mainly meant to simplify things for the SPDIF users, as at the moment the "receiver capable" settings are hidden at the Standard setting level, so enabling passthrough would automatically give them 5.1 without having uncovering the "receiver capable" settings. However I respect the view that it might not be a good idea to enable a settings that would result in a degradation of quality, so I'll remove this change.

PR #8405, opened 2015-11-18
Cleanup music ratings

razzeee:
This tries to do the following:
- Separate rating and userrating for music
- Give albums and songs both rating types
- Move everything to 1-10 ratings

The tricky part about this, was that albums only had scraped ratings (called "rating") and songs only had userratings. Both used the rating field of the music tag. So I tried to unravel this with this PR.

This will also have the side effect, that rating now is a float. And that some addons might think rating is a userrating. They need to be updated.

@DaveTBlake as this is music
@Montellese as it has json rpc changes
@zag2me as you might like this
Special thx to @ronie 

screenshot 2015-11-18 20 17 56
screenshot 2015-11-18 20 17 47
screenshot 2015-11-18 20 18 37
screenshot 2015-11-18 20 18 17
screenshot 2015-11-18 20 18 10

PR #10014, opened 2016-06-20
fixed: i think this was supposed to be 10. cut-and-paste ftw

notspiff:
I tried to ping the original author but no response so here's my take.


Razzeee:
Thx, looking much better

DaveTBlake:
Helps to tie this back to the original discussion #8405

PR #10015, opened 2016-06-20
added: FFmpeg-supplied RTMP support

notspiff:
This enables FFmpeg supplied RTMP.

It is not as capable as librtmp, but does fine in no-crypto situations. The add-on overrides.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -600,6 +600,8 @@ bool CApplication::Create()
   av_register_all();
   // register avfilter
   avfilter_register_all();
+  // initialize network protocols
+  avformat_network_init();
   // set avutil callback
   av_log_set_callback(ff_avutil_log);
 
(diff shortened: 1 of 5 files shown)


notspiff:
note this is incomplete, we need to change the build options for ffmpeg. will do if this is welcome.

fritsch:
It is welcome. Please go ahead. Thanks much in advance.

notspiff:
turns out ffmpeg build options have changed (?) and we now build all protocols so no buildsystem changes ncessary. the rebased branch should now build. @FernetMenta please review

notspiff:
i have not yet runtime tested the rebased code so that should be done before the green is hit.

notspiff:
lightly runtime tested with some streams. works as expected afaict.

notspiff:
added a second commit which calls the global network init function of avformat. quells a runtime warning and adds forward compatibility.

MartijnKaijser:
jenkins build this please

MartijnKaijser:
Is the depends version now obsolete with this and should be removed?

FernetMenta:
As far as I understood @notspiff the addon inputstream.rtmp still requires it in this phase.

PR #10019, opened 2016-06-20
Enablestream

ghost:
This PR is mainly to not forget things I figured out when starting a stream not from start.
I'm quite sure that the second commit is not OK, so pls. don't bash me :-)

1.) Before the inputstream is open, kodi does not know the startpts of the video.
Its a general question how seek should be take place: relativ or absolute.
Relative to video startpts would enable the possibility to call EnableStreamAtPTS with correct value, if absolute there must be implemented an option to let kodi know th startpts before a stream is started,

2.) If inputstream returns DMX_SPECAL_STREAMCHANGE on first packet (maybe also on later packets), kodi rebuilds selectionstreams and calls OpenDefaultStreams(). OpenDefaultStreams call EnableStream() without Disabling the streams before. Either the call to EnableStream is not necessary, or streams has to be disabled before.
I'm also wondering, that both streams are completely rebuild, as only the audio stream has changed. It would be enough here to only reopen the audio stream.

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -1237,6 +1237,9 @@ void CVideoPlayer::Process()
     m_OmxPlayerState.av_clock.OMXPause();
   }
 
+  if (m_PlayerOptions.starttime > 0)
+    m_State.time = m_PlayerOptions.starttime * 1000;
+
   OpenDefaultStreams();
 
   // look for any EDL files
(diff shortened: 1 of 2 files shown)


FernetMenta:
> 1.) Before the inputstream is open, kodi does not know the startpts of the video

Kodi does not want to know startpts because a stream can start with what pts ever. PVR (mpeg-ts) stream for example start with arbitrary pts. pts wraps every 23 hours.

> 2.) If inputstream returns DMX_SPECAL_STREAMCHANGE on first packet (maybe also on later packets), kodi rebuilds selectionstreams and calls OpenDefaultStreams().

OpenDefaultStream can be called as often as you like. It does not do anything if nothing has changed.

PR #9300, opened 2016-03-08
Improvements to AudioLibrary.GetArtists, GetAlbums and GetSongs 

DaveTBlake:
API improvements to AudioLibrary.GetArtists, GetAlbums and GetSongs. This follows on from  #8015 and the addition of artist roles to the music library, and adresses some flaws that were in the initial API implementation for accessing artist role data.

It also attempts to fix some longer standing issues: GetArtists now returns the song genres and genreids of the songs of each artist, and a flag if the artist is an "album artist". Thus returning the data values that correspond to the filters. It  makes it clearer in the documentation that the only genre filtering currently offered is song genre (not the artist genre value that can be scraped, stored and displayed but is not used in queries). Extending filtering to all kinds of genre (artist and album) is something for another PR.

@Montellese you have seen all of this before, so hope that all is well.
@JoeTheFox finally catching up with the iOS remote!
@Tolriq I think this covers the many things you pointed out with roles, but also improves GetArtists generally in the way I said I would. More work to do of course, especially with genre, but this seems enough for one PR.

PR #10020, opened 2016-06-21
Change ratings to reflect json art (and future uniqueid) structure

phate89:
Ratings are exactly like uniqueids and this brings the same structure uniqueids will have.
@Montellese @Tolriq Is this api change ok even if it was already added differently? I don't think anyone implemented it yet


Tolriq:
I have not yet implemented those but it's on Todo :)

From quick reading definition looks ok with just one caveat maybe I have not read all code.
How do you handle the default ? If the default is deleted should it move to something else ?
If you add a new rating with default true, should it set false to previous true ?

During v17 alpha it was decided to ease JSON consumers, that breaking changes are handled with a little details.

The update to version.txt should be in a separated commit with title containing [Breaking Change] see https://github.com/xbmc/xbmc/pull/9300 for history.

But there's no problems for breaking changes before Beta 1.


Since this is related to something in my TODO I ask my question about it here :)
How is all that managed in the GUI ? (Coupled with userrating ?)

Even if there's a rating with default true, if there's a user rating it shows the user rating ?
Or is there's a global Kodi settings somewhere that toggle between ratings and userratings ?

MartijnKaijser:
A separate commit is plain stupid to do. Period

Tolriq:
Well maybe you should read the full discussion in https://github.com/xbmc/xbmc/pull/9300
 
This was decided and acted, because breaking changes should bump major version, but during alphas it was also decided to not bump version to avoid a too big number.

version.txt is the file that any API consumer have to look to know what PR made changes.

You can't ask someone to read 60 or 70 PR to know when the breaking change was made. Specially when JSON change is hidden in the middle of other changes.

Not indicating a breaking change in a way for API consumers to detect it is plain stupid. Period.

Please stop endless debating on acted things.

Either major version bump, or commit title on version.txt that was the choice and it was made.
If you have a better idea that would revolution the problem then previous discussion would have been the right timing for that.

phate89:
userratings are a separate value in a separate field.
If you delete the default one without setting a new one it will use the first one available as default (to avoid .
If you set a new default the other one is set to false.. There's only one default value 
I can add [breaking change] to the commit breaker but one specific only for the version change seems unnecessary..

Tolriq:
As long as version.txt is contained in a commit that have that tag in the tittle it's OK.

A separate commit was easier to avoid polluting the global git history to not fear everyone specially skinners.
But if the whole change only impact JSON adding  [breaking change] and [json] as @Montellese loves is perfect.


About ratings, how do skin choose between userrating and ratings ?

phate89:
I'm not sure but they can do what they want. If they think it's better to replace rating with userrating they can, if they think it's better to show them separate they have both

xbmc/interfaces/json-rpc/schema/types.json
```diff
@@ -392,6 +392,22 @@
...
+      "default": { "type": "boolean" }
+    }
+  },
+  "Video.Ratings": {
+    "type": "object",
+    "additionalProperties": { "type": "$ref": "Video.Rating" }
+  },
+  "Video.Ratings.Set": {
+    "type": "object",
+    "additionalProperties": { "type": [ "null", { "$ref": "Video.Rating", "required": true } ] }
```
MilhouseVH: Should this be `Video.Rating` or `Video.Ratings`? As it is, the `Video.Ratings` type (403:406) suggests to be redundant.

phate89: If I didn't something wrong they're not redundant.
Video.Rating should be:
```
{ "rating": 1, "votes": 2 }
```
Video.Ratings should be:
```
{ "imdb": (Video.Rating), "tvdb": (Video.Rating) }
```

MilhouseVH: In that case, I don't see where the `Video.Ratings` (plural) type is used or referenced by a field so it appears (with my limited understanding of this JSON syntax) to be redundant/unnecessary as only `Video.Rating` is used by `Video.Ratings.Set`, which in turn is used only by the `ratings` field/property.

phate89: That's because the "Video.Ratings" was already there before this change. I just moved and changed the definition. So all the times it was used (except 3 I changed now to Video.Ratings.Set) are still there. You just can't see it from the diff ;)
Look at the full file and you will find them
https://github.com/phate89/xbmc/blob/4b9fb67521205adf2c48a675d13db0f8e96200e5/xbmc/interfaces/json-rpc/schema/types.json

MilhouseVH: Doh, sorry for the noise!

xbmc/interfaces/json-rpc/schema/types.json
```diff
@@ -392,6 +392,22 @@
...
+    "type": "object",
+    "properties": {
+      "rating": { "type": "number", "required": true },
+      "votes": { "type": "integer" },
+      "default": { "type": "boolean" }
+    }
+  },
+  "Video.Ratings": {
+    "type": "object",
+    "additionalProperties": { "type": "$ref": "Video.Rating" }
```
Montellese: This looks wrong. It should be
```
"additionalProperties": { "$ref": "Video.Rating" }
```

xbmc/video/VideoInfoTag.cpp
```diff
@@ -556,16 +556,13 @@ void CVideoInfoTag::Serialize(CVariant& value) const
   value["episode"] = m_iEpisode;
   value["uniqueid"]["unknown"] = m_strUniqueId;
   value["rating"] = GetRating().rating;
-  value["ratings"] = CVariant(CVariant::VariantTypeArray);
   for (const auto& i : m_ratings)
   {
-    CVariant rating;
```
Montellese: I would change this a bit to reduce having to type all the indices multiple times to something like
```
CVariant ratings = CVariant(CVariant::VariantTypeObject);
for (const auto& i : m_ratings)
{
  CVariant rating;
  rating["rating"] = i.second.rating;
  rating["votes"] = i.second.votes;
  rating["default"] = i.first == m_strDefaultRating;

  ratings[i.first] = rating;
}
value["ratings"] = ratings;
```
This way you only have to use every index once.

Montellese:
Looks good apart from my comments. Thanks for thinking of this (based on the discussion in the other PR) and taking care of it.

phate89:
@Montellese updated according to your comments

Montellese:
Looks good.

phate89:
jenkins build this please

sualfred:
@phate89 @Montellese @MartijnKaijser 

I'm no dev and I'm just guessing so please be gentle with me:
Artwork Downloader, and other addons which are going to update the Art value via JSON, will break/remove the cast of the updated TV show. Movies are fine.
The nightly before this commit is working, so I just guess it has something to do with this change?

See: http://forum.kodi.tv/showthread.php?tid=282596

PR #10023, opened 2016-06-23
Quell some warnings

notspiff:
- redefinition of HAVE_OPENSSL with cmake builds (comes from command line not config.h)
- avoid deprecated symbol usage in mhd with newer versions

@Montellese for the last one i guess.

### DIFF
--- a/xbmc/network/WebServer.cpp
+++ b/xbmc/network/WebServer.cpp
@@ -102,9 +102,23 @@ CWebServer::CWebServer()
 #endif
 }
 
+static MHD_Response* create_response(size_t size, void* data, int free, int copy)
+{
+#if (MHD_VERSION >= 0x00094001)
+  MHD_ResponseMemoryMode mode = MHD_RESPMEM_PERSISTENT;
+  if (copy)
+    mode = MHD_RESPMEM_MUST_COPY;
+  else if (free)
+    mode = MHD_RESPMEM_MUST_FREE;
+  return MHD_create_response_from_buffer(0, nullptr, mode);
+#else
+  return MHD_create_response_from_data(0, nullptr, free, copy);
+#endif
+}
+
 int CWebServer::AskForAuthentication(struct MHD_Connection *connection) const
 {
-  struct MHD_Response *response = MHD_create_response_from_data(0, nullptr, MHD_NO, MHD_NO);
+  struct MHD_Response *response = create_response(0, nullptr, MHD_NO, MHD_NO);
   if (!response)
   {
     CLog::Log(LOGERROR, "CWebServer[%hu]: unable to create HTTP Unauthorized response", m_port);
@@ -284,7 +298,7 @@ int CWebServer::HandlePartialRequest(struct MHD_Connection *connection, Connecti
                 ifModifiedSinceDate.SetFromRFC1123DateTime(ifModifiedSince) &&
                 lastModified.GetAsUTCDateTime() <= ifModifiedSinceDate)
               {
-                struct MHD_Response *response = MHD_create_response_from_data(0, nullptr, MHD_NO, MHD_NO);
+                struct MHD_Response *response = create_response(0, nullptr, MHD_NO, MHD_NO);
                 if (response == nullptr)
                 {
                   CLog::Log(LOGERROR, "CWebServer[%hu]: failed to create a HTTP 304 response", m_port);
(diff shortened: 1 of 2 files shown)


MartijnKaijser:
or @Paxxi ping

Our depends/win32 all have MHD 0.9.50 so this is just for the distros that have a lower version correct?

notspiff:
Well the old code is kept for that reason. it quells warnings which
appeared in 0.9.4
23. jun. 2016 08:50 skrev "Martijn Kaijser" <notifications@github.com>:

xbmc/network/WebServer.cpp
```diff
@@ -102,9 +102,25 @@ CWebServer::CWebServer()
...
 
+static MHD_Response* create_response(size_t size, void* data, int free, int copy)
+{
+#if (MHD_VERSION >= 0x00094001)
+  MHD_ResponseMemoryMode mode = MHD_RESPMEM_PERSISTENT;
+  if (copy)
+    mode = MHD_RESPMEM_MUST_COPY;
+  else if (free)
+    mode = MHD_RESPMEM_MUST_FREE;
+  struct MHD_Response *response = MHD_create_response_from_buffer(0, nullptr, mode);
```
Montellese: I would just return here (and two lines down as well) because otherwise we run the danger of someone changing the name in the `#if` part and not in the `#else` part.

Montellese:
I'm not really up-to-date with the development of libmicrohttpd but the changes look ok.

xbmc/network/WebServer.cpp
```diff
@@ -102,9 +102,25 @@ CWebServer::CWebServer()
 #endif
 }
 
+static MHD_Response* create_response(size_t size, void* data, int free, int copy)
```
Paxxi: I would prefer having an enum here instead of free and copy as it would make it clearer at the call sites what's going on instead of mostly passing `MDH_NO, MHD_NO`.
I realize this is a bit about personal taste and probably out of scope for this PR but I thought I'd mention it

Paxxi:
good for merge imo unless @notspiff agrees about the enum and want to change it :)

akva2:
If you think investing work in the legacy code path is worth it.. it has been ints for years due to mhd api. Basically your comment is what was adressed upstream.

notspiff:
early returns added. i will do the enumeration only if you insist. when we drop the legacy code path we can directly use the mhd enum so it feels kinda unnnecessary to put work into it.

Paxxi:
@notspiff not needed, as I mentioned in previous comment I left the decision to you as it's your time and a minor thing.

L-S-D:
FYI, this breaks at least the IOS Remote and in general the WebIF for me

notspiff:
if you are going to FYI me, you better give me some information ;)

L-S-D:
yeah you are absolutely right, just made my wife happy in finding the PR that broke it. For now I just can say that IOS Remote cannot connect to kodi anymore. I'm running on non standard ports. I won't be able to make logs and dig into that deeper until in the evening.

L-S-D:
working fine with libmicrohttpd 0.9.37 on debian (default for jessie) and fails with 0.9.44 from stretch

xbmc/network/WebServer.cpp
```diff
@@ -102,9 +102,23 @@ CWebServer::CWebServer()
...
+{
+#if (MHD_VERSION >= 0x00094001)
+  MHD_ResponseMemoryMode mode = MHD_RESPMEM_PERSISTENT;
+  if (copy)
+    mode = MHD_RESPMEM_MUST_COPY;
+  else if (free)
+    mode = MHD_RESPMEM_MUST_FREE;
+  return MHD_create_response_from_buffer(0, nullptr, mode);
+#else
+  return MHD_create_response_from_data(0, nullptr, free, copy);
```
popcornmix: This should be:
```return MHD_create_response_from_data(size, data, free, copy);```

Comment (notspiff) [inline xbmc/network/WebServer.cpp]: Doh. Please fix i am not around a puter for a few days..
24. jun. 2016 17:16 skrev "popcornmix" <notifications@github.com>:

**Comment by MilhouseVH** [comment]
Hi, just some additional detail on what is not working.

iOS and Yatse remote control apps no longer work after this change (Ubuntu x86_64 16.04, LibreELEC #0623 RPi and x86).

Connecting to the webgui on port 8080 with a browser will result in "Connection to server lost" messages.

The following JSON query will never return a result, it simply hangs - it should return the JSON schema more or less instantly:
```
curl -sG -H 'content-type: application/json;' http://192.168.0.18:8080/jsonrpc
```

iOS and Yatse applications can't create connections to clients because the webserver does not respond with the above query.

Reverting this change restores functionality.

I'll try a build with the change suggested by @popcornmix.

xbmc/network/WebServer.cpp
popcornmix: Sure. I'll PR when @MilhouseVH confirms it is good (I suggested both calls are fixed).

PR #10025, opened 2016-06-23
[depends] bump mDNSResponder to 625.41.2

Rechi:


tools/depends/target/mdnsresponder/Makefile
```diff
@@ -1,13 +1,14 @@
 include ../../Makefile.include
-DEPS= ../../Makefile.include Makefile
+DEPS= ../../Makefile.include Makefile makefile.internal 01-android-embedded.patch
```
MartijnKaijser: why is this line changed as it's already added on line 24

Rechi: other depends like python or libssh also have there patches added to ```DEPS```.
This also enables rebuild of the package if only the patch file gets changed.

MartijnKaijser:
@Rechi i uploaded the file to our server so you can remove the url change

@Memphiz ping as you added this lib in the first place :)

Rechi:
@MartijnKaijser removed url change

Memphiz:
anything we will gain by bumping this? (just curious) - jenkins build this please

Rechi:
I'm not sure if we are affected by the issue of [mDNSResponder](https://support.apple.com/en-us/HT206846), but I thought it would be a good idea to include the security update.

Paxxi:
I'm all for keeping up to date, if it might fix security issues all the better. 
This needs bumping on windows as well so I guess I'll add it to my todo-list :)

PR #10031, opened 2016-06-25
Timer icons in guide view, fixes #16621

### LINKED ISSUE #16621: [REQUEST] play m3u8 content with "EXT-X-MEDIA-SEQUENCE"
### Describe the bug

The content m3u8 with these headers not play in Kodi 18.4
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-TARGETDURATION:10
#EXT-X-MEDIA-SEQUENCE:268998

## Expected Behavior
Here is a clear and concise description of what was expected to happen:

To be able to play these type of contents
VLC can play this content

## Actual Behavior

Not playable

### To Reproduce

1. Use this file M3u file : [1501.zip](https://github.com/xbmc/xbmc/files/3612599/1501.zip)
2. Play in Kodi

### Debuglog

The debuglog can be found here:
Log : https://pastebin.com/0aLjaTs9
Full Log: [kodi.log](https://github.com/xbmc/xbmc/files/3612613/kodi.log)

### Screenshots 
Here are some links or screenshots to help explain the problem:

No screenshot to provide

## Additional context or screenshots (if appropriate)
Here is some additional context or explanation that might help:

No more context

### Your Environment
Used Operating system:

 - [ ] Android
 - [ ] iOS
 - [x] Linux
 - [ ] OSX
 - [ ] Raspberry-Pi
 - [ ] Windows
 - [ ] Windows UWP

 - Kodi version: 18.4

*note: Once the issue is made we require you to update it with new information or Kodi versions should that be required.
Team Kodi will consider your problem report however, we will not make any promises the problem will be solved.*

bas-t:
Patch by @dnairb, all credits to him/her.

### DIFF
--- a/addons/skin.estuary/1080i/Includes_PVR.xml
+++ b/addons/skin.estuary/1080i/Includes_PVR.xml
@@ -206,7 +206,32 @@
 							<top>15</top>
 							<width>35</width>
 							<height>35</height>
-							<texture>$VAR[PVRStatusImageVar]</texture>
+							<texture>windows/pvr/record.png</texture>
+							<visible>ListItem.IsRecording</visible>
+						</control>
+						<control type="image">
+							<left>1000</left>
+							<top>15</top>
+							<width>35</width>
+							<height>35</height>
+							<texture>windows/pvr/timer.png</texture>
+							<visible>ListItem.HasTimer + !ListItem.IsRecording + !ListItem.HasTimerSchedule</visible>
+						</control>
+						<control type="image">
+							<left>1000</left>
+							<top>15</top>
+							<width>35</width>
+							<height>35</height>
+							<texture>windows/pvr/timer.png</texture>
+							<visible>ListItem.HasTimerSchedule + !ListItem.IsRecording</visible>
+						</control>
+						<control type="image">
+							<left>1000</left>
+							<top>15</top>
+							<width>35</width>
+							<height>35</height>
+							<texture>windows/pvr/timer.png</texture>
+							<visible>ListItem.HasRecording + !ListItem.IsRecording + !ListItem.HasTimer</visible>
 						</control>
 					</itemlayout>
 					<focusedlayout height="58">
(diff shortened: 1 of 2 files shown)


ksooo:
Not a skin expert but I'm not sure this fixes anything. For me, without this patch and latest master (which includes #10024 - intended to fix the icon problems) the different icons are working now, without any skin changes. The problem with the not (immediately) appearing status icons was a bug in kodi pve core, not in the skin.

@bas-t @dnairb can you check latest kodi master without your skin patch and report back, please?

bas-t:
@ksooo I did, and your latest commits did not fix anything regarding my ticket, as opposed to the fix that @dnairb presented.

ksooo:
Hmmm. Just to be sure, you really tested a build the includes #10024? This PR got merged some hours ago! You need a really very recent kodi master build, I guess atm you have to build yourself. Did you?

bas-t:
@ksooo I allways build kodi myself, don't worry, your latest commits are included.

ksooo:
What I do not get is the the trac ticket talks about Confluence not working and Estuary working. How can an Estuary fix do something good here? For me this reads exactly like what I found out the root cause being a bug in kodi pvr core.

bas-t:
That comments are outdated. Currently both estuary and confluence skins are affected.
However, confluence is not part of the master branch anymore, so no update to that is submitted atm.

ksooo:
This is really strange because it works here. Using latest krypton master on MacOS, without additional patches plus tvheadend 4.1.something-some-weeks-old running on Raspi2.

bas-t:
Ok, it might work on MacOS, it does not on linux.

ksooo:
I thought that you would say this. ;-) Which Linux?

ksooo:
Libreelec master (with latest kodi master patched in) works for me.

bas-t:
Well, Ubuntu Xenial and Debian Stretch don't work for me.

ksooo:
O well, seems I mixed up something. Sorry. The problem you describe indeed might be platform depedent (though I'm not an expert on this). I think some caching stuff around kodi guilib is involved here. My fix approach was to invalidate the epg grid controls in case of changes. Obviously this works fine on MacOS, but not on for instance Ubuntu.

@Montelese do you have an idea what could be wrong here, if on MacOS calls to SetInvalid() help, but not on Linux? Mind taking a look at trac ticket 16621 and #10024 for more details as well?

Glenn-1990:
Stumbled into this problem when working on the timer icon stuff. Removing the include var doesn't seems a good idea as this will result in much duplicate code (which is sensitive for inconsistencies). The caching/delay problem should be fixed. And it was never more than a few seconds?

ksooo:
> The caching/delay problem should be fixed.

 yep. Unfortunately I'm completely lost on this, because i have absolutely no knowledge about guiilib (or even platform graphics stuff). @Montellese ?

ksooo:
> And it was never more than a few seconds?

A small delay is okay as kodi will paint the new icons only *after* the pvr backend (e.g. tvheadend) has confirmed the change (e.g. timer added, deleted). "a few seconds" would mean that the confirmation for the action from the tvheadend server takes that long. Possible (if slow server connection or heavy-loaded server) but not the usual behavior, I'd say.

Glenn-1990:
@ksooo sometimes it took longer, but in that case the program info was also wrong (record instead of delete timer was shown) so it seems a combination of caching and your other PR. This was some months ago, not tested with recent master.

ksooo:
> so it seems a combination of caching and your other PR

 yeah, that's my take on this, too. And I'm quite convinced that  "my other PR" (#10024) fixes the epg event context menu entries not matching the actual state.

bas-t:
This PR is superseeded by a real fix (as opposed to this workaround) that @ksooo has in https://github.com/ksooo/xbmc/commits/pvr-fix-timer-icon-issues

Closing this PR.

PR #10032, opened 2016-06-25
[webserver] CHTTPImageHandler: add GetLastModifiedDate implementation

Rechi:
Posters for movies, tv shows and music can be cached by the browser.

### DIFF
--- a/xbmc/network/httprequesthandler/HTTPImageHandler.cpp
+++ b/xbmc/network/httprequesthandler/HTTPImageHandler.cpp
@@ -37,7 +37,15 @@ CHTTPImageHandler::CHTTPImageHandler(const HTTPRequest &request)
     XFILE::CImageFile imageFile;
     const CURL pathToUrl(file);
     if (imageFile.Exists(pathToUrl))
+    {
       responseStatus = MHD_HTTP_OK;
+      struct __stat64 statBuffer;
+      if (imageFile.Stat(pathToUrl, &statBuffer) == 0)
+      {
+        SetLastModifiedDate(&statBuffer);
+        SetCanBeCached(true);
+      }
+    }
     else
       responseStatus = MHD_HTTP_NOT_FOUND;
   }
--- a/xbmc/network/httprequesthandler/HTTPImageHandler.h
+++ b/xbmc/network/httprequesthandler/HTTPImageHandler.h
@@ -33,6 +33,7 @@ public:
   virtual bool CanHandleRequest(const HTTPRequest &request);
 
   virtual int GetPriority() const { return 5; }
+  virtual int GetMaximumAgeForCaching() const { return 60 * 60 * 24 * 7; }
 
 protected:
   explicit CHTTPImageHandler(const HTTPRequest &request);
--- a/xbmc/network/httprequesthandler/HTTPFileHandler.cpp
+++ b/xbmc/network/httprequesthandler/HTTPFileHandler.cpp
@@ -20,7 +20,6 @@
 
 #include "system.h"
 #include "HTTPFileHandler.h"
-#include "filesystem/File.h"
 #include "utils/Mime.h"
 #include "utils/StringUtils.h"
 #include "utils/URIUtils.h"
@@ -89,17 +88,7 @@ void CHTTPFileHandler::SetFile(const std::string& file, int responseStatus)
     {
       struct __stat64 statBuffer;
       if (fileObj.Stat(&statBuffer) == 0)
-      {
-        struct tm *time;
-#ifdef HAVE_LOCALTIME_R
-        struct tm result = { };
-        time = localtime_r((time_t*)&statBuffer.st_mtime, &result);
-#else
-        time = localtime((time_t *)&statBuffer.st_mtime);
-#endif
-        if (time != NULL)
-          m_lastModified = *time;
-      }
+        SetLastModifiedDate(&statBuffer);
     }
   }
 
@@ -113,4 +102,17 @@ void CHTTPFileHandler::SetFile(const std::string& file, int responseStatus)
   // disable caching if the last modified date couldn't be read
   if (!m_lastModified.IsValid())
     m_canBeCached = false;
+}
+
+void CHTTPFileHandler::SetLastModifiedDate(const struct __stat64 *statBuffer)
+{
+  struct tm *time;
+#ifdef HAVE_LOCALTIME_R
+  struct tm result = { };
+  time = localtime_r((time_t*)&statBuffer->st_mtime, &result);
+#else
+  time = localtime((time_t *)&statBuffer->st_mtime);
+#endif
+  if (time != NULL)
+    m_lastModified = *time;
 }
--- a/xbmc/network/httprequesthandler/HTTPFileHandler.h
+++ b/xbmc/network/httprequesthandler/HTTPFileHandler.h
@@ -22,6 +22,7 @@
 #include <string>
 
 #include "XBDateTime.h"
+#include "filesystem/File.h"
 #include "network/httprequesthandler/IHTTPRequestHandler.h"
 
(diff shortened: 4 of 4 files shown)


xbmc/network/httprequesthandler/HTTPImageHandler.cpp
```diff
@@ -36,8 +37,13 @@ CHTTPImageHandler::CHTTPImageHandler(const HTTPRequest &request)
 
     XFILE::CImageFile imageFile;
     const CURL pathToUrl(file);
-    if (imageFile.Exists(pathToUrl))
+    if (imageFile.Exists(pathToUrl)) {
```
popcornmix: Open bracket should be on a new line

Rechi:
@popcornmix updated

MartijnKaijser:
@Montellese ping

Tolriq:
@Rechi what headers does it update ? 

I have not checked since a long time but there was also an etag problem at some point.

Rechi:
Only ```Last-Modified``` get added to the response header.

Tolriq:
@Rechi Ok no I've tested the problem was not the etag (it's not returned on Windows at least) but the Cache-Control header

Cache-Control: private, max-age=0, no-cache

You need to also have this header to be removed for /image but not the rest.

I've  tested your patch it only adds : Last-Modified: Thu, 19 May 2016 10:12:55 GMT

But leave the other header that will forbid the cache on any client that respect those (meaning nearly all)

xbmc/network/httprequesthandler/HTTPImageHandler.h
```diff
@@ -32,8 +32,14 @@ class CHTTPImageHandler : public CHTTPFileHandler
   virtual IHTTPRequestHandler* Create(const HTTPRequest &request) { return new CHTTPImageHandler(request); }
   virtual bool CanHandleRequest(const HTTPRequest &request);
 
+  virtual bool CanBeCached() const { return true; }
+  virtual bool GetLastModifiedDate(CDateTime &lastModified) const;
```
Montellese: Please add the `override` keyword on both methods.

xbmc/network/httprequesthandler/HTTPImageHandler.cpp
```diff
@@ -46,6 +53,15 @@ CHTTPImageHandler::CHTTPImageHandler(const HTTPRequest &request)
   SetFile(file, responseStatus);
 }
 
+bool CHTTPImageHandler::GetLastModifiedDate(CDateTime &lastModified) const
```
Montellese: The base class `CHTTPFileHandler` already offers helper methods to handle this. You just need to call `SetCanBeCached()` in the constructor and `SetLastModifiedDate()` instead of using your own private `m_lastModified` variable.

xbmc/network/httprequesthandler/HTTPImageHandler.cpp
```diff
@@ -37,7 +38,13 @@ CHTTPImageHandler::CHTTPImageHandler(const HTTPRequest &request)
     XFILE::CImageFile imageFile;
     const CURL pathToUrl(file);
     if (imageFile.Exists(pathToUrl))
+    {
       responseStatus = MHD_HTTP_OK;
+      struct __stat64 buffer;
```
Montellese: Please see https://github.com/Rechi/xbmc/blob/92cd69039cc531a9b193e1636f1df480d6c76fea/xbmc/network/httprequesthandler/HTTPFileHandler.cpp#L90 for how to do this properly. Maybe we can refactor that logic into a helper method in `CHTTPFileHandler` which can be used by the base and any derived classes.

Montellese:
To fix the issue @Tolriq reported you can either implement/override `GetMaximumAgeForCaching()` or you can manually set the `Cache-Control` header. But I always tried not to manually set any of these "generic" headers and tried to implement it through the `IHTTPRequestHandler` interface. Maybe the `Cache-Control` logic in `CWebServer::FinalizeRequest()` needs some tweaking as well.

Tolriq:
For images correct header should be : 

Cache-Control: public, max-age=XXXXX 

For XXXX the value is debatable, but since there's no etag I guess at least 1 week but @Montellese call.

xbmc/network/httprequesthandler/HTTPImageHandler.cpp
```diff
@@ -37,7 +37,25 @@ CHTTPImageHandler::CHTTPImageHandler(const HTTPRequest &request)
     XFILE::CImageFile imageFile;
     const CURL pathToUrl(file);
     if (imageFile.Exists(pathToUrl))
+    {
       responseStatus = MHD_HTTP_OK;
+      struct __stat64 statBuffer;
+      if (imageFile.Stat(pathToUrl, &statBuffer) == 0)
+      {
+        struct tm *lastChange;
+#ifdef HAVE_LOCALTIME_R
```
akva2: this will need a probe in the cmake buildsystem if you are to use the define.

Rechi: that is already used in https://github.com/xbmc/xbmc/blob/master/xbmc/network/httprequesthandler/HTTPFileHandler.cpp#L94

akva2: sure. that's also broken in that case :)

Rechi: is [that](https://github.com/xbmc/xbmc/blob/master/project/cmake/scripts/common/ArchSetup.cmake#L96) the probe you want?

akva2: aha. somebody has already adressed the issue. that's the best fixes ever don't you agree? :)

xbmc/network/httprequesthandler/HTTPImageHandler.cpp
Rechi: What parameter should the helper method have?
maybe ```__stat64``` or ```XFILE::IFile```?

Rechi:
@Montellese is the second commit a good way for the helper method?

Montellese:
Looks good. But you still might have to override `GetMaximumAgeForCaching()` in `CHTTPImageHandler` to get a proper `Cache-Control` header.

Rechi:
@Montellese what will happen if `GetMaximumAgeForCaching()` is set and the picture changes? Are the browsers always checking if `Last-Modified` has changed, or will the old picture be shown until `max-age` expires?

Montellese:
I'd have to re-read the RFC(s) but IIRC `max-age` says that the browser should re-cache the image at least after that period but he's free to do so before that. And if `Last-Modified` changed it should re-cache anyway.

Rechi:
@Montellese have added `GetMaximumAgeForCaching`.
In Firefox the pictures get refreshed if they have changed, also if max-age haven't expired.

xbmc/network/httprequesthandler/HTTPImageHandler.h
```diff
@@ -33,6 +33,7 @@ class CHTTPImageHandler : public CHTTPFileHandler
   virtual bool CanHandleRequest(const HTTPRequest &request);
 
   virtual int GetPriority() const { return 5; }
+  virtual int GetMaximumAgeForCaching() const { return 60 * 60 * 24 * 7; }
```
Montellese: Either adjust the commit message or move this into it's own commit.

Montellese:
@Rechi thanks for checking and implementing `GetMaximumAgeForCaching`. This is good to go as soon as my last comment has been resolved.

Rechi:
@Montellese splitted `GetMaximumAgeForCaching` into a separate commit.

Montellese:
Thanks, this is fine with me now. jenkins build this please

Montellese (approved):

PR #10036, opened 2016-06-27
[PVR] Guide window: Fix timer icon issues. Fixes #16621

### LINKED ISSUE #16621: [REQUEST] play m3u8 content with "EXT-X-MEDIA-SEQUENCE"

ksooo:
Fixes issues with not appearing/disappearing timer state icons in PVR Guide window, like discussed here: https://github.com/xbmc/xbmc/pull/10024#issuecomment-228530671 ff. 

Fixes http://trac.kodi.tv/ticket/16621

Code changes are strait forward.

Reported to actually fix the problems by @bas-t and @seo

### DIFF
--- a/xbmc/pvr/windows/GUIWindowPVRGuide.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRGuide.cpp
@@ -98,10 +98,12 @@ void CGUIWindowPVRGuide::StopRefreshTimelineItemsThread()
 void CGUIWindowPVRGuide::RegisterObservers(void)
 {
   g_EpgContainer.RegisterObserver(this);
+  g_PVRTimers->RegisterObserver(this);
 }
 
 void CGUIWindowPVRGuide::UnregisterObservers(void)
 {
+  g_PVRTimers->UnregisterObserver(this);
   g_EpgContainer.UnregisterObserver(this);
 }
 
@@ -403,6 +405,12 @@ bool CGUIWindowPVRGuide::OnMessage(CGUIMessage& message)
           Refresh(true);
           break;
         }
+        case ObservableMessageTimersReset:
+        case ObservableMessageTimers:
+        {
+          SetInvalid();
+          break;
+        }
         case ObservableMessageEpgActiveItem:
         {
           if (m_viewControl.GetCurrentControl() != GUIDE_VIEW_TIMELINE)
@@ -423,8 +431,7 @@ bool CGUIWindowPVRGuide::OnContextButton(int itemNumber, CONTEXT_BUTTON button)
     return false;
   CFileItemPtr pItem = m_vecItems->Get(itemNumber);
 
-  bool bReturn
-    = OnContextButtonPlay(pItem.get(), button) ||
+  return OnContextButtonPlay(pItem.get(), button) ||
       OnContextButtonInfo(pItem.get(), button) ||
       OnContextButtonStartRecord(pItem.get(), button) ||
       OnContextButtonStopRecord(pItem.get(), button) ||
@@ -436,11 +443,6 @@ bool CGUIWindowPVRGuide::OnContextButton(int itemNumber, CONTEXT_BUTTON button)
       OnContextButtonEnd(pItem.get(), button) ||
       OnContextButtonNow(pItem.get(), button) ||
       CGUIWindowPVRBase::OnContextButton(itemNumber, button);
-
-  if (bReturn)
-    SetInvalid();
-
-  return bReturn;
 }
 
 void CGUIWindowPVRGuide::GetViewChannelItems(CFileItemList &items)
--- a/xbmc/pvr/windows/GUIWindowPVRGuide.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRGuide.cpp
@@ -97,13 +97,19 @@ void CGUIWindowPVRGuide::StopRefreshTimelineItemsThread()
 
 void CGUIWindowPVRGuide::RegisterObservers(void)
 {
+  CSingleLock lock(m_critSection);
   g_EpgContainer.RegisterObserver(this);
-  g_PVRTimers->RegisterObserver(this);
+  if (g_PVRTimers)
+    g_PVRTimers->RegisterObserver(this);
+  CGUIWindowPVRBase::RegisterObservers();
 }
 
 void CGUIWindowPVRGuide::UnregisterObservers(void)
 {
-  g_PVRTimers->UnregisterObserver(this);
+  CSingleLock lock(m_critSection);
+  CGUIWindowPVRBase::UnregisterObservers();
+  if (g_PVRTimers)
+    g_PVRTimers->UnregisterObserver(this);
   g_EpgContainer.UnregisterObserver(this);
 }
 
@@ -112,6 +118,7 @@ void CGUIWindowPVRGuide::Notify(const Observable &obs, const ObservableMessage m
   if (m_viewControl.GetCurrentControl() == GUIDE_VIEW_TIMELINE &&
       (msg == ObservableMessageEpg ||
        msg == ObservableMessageEpgContainer ||
+       msg == ObservableMessageChannelGroupReset ||
        msg == ObservableMessageChannelGroup))
   {
     CSingleLock lock(m_critSection);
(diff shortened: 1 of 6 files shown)


ksooo:
jenkins build this please

ksooo:
@MilhouseVH ping

bas-t:
@ksooo thanks for your effort, this works like a charm for me.

Is marking the trac ticket as fixed (referencing this PR)  and closing it an automated process? It's not done yet and I don't know how to do it.

ksooo:
> Is marking the trac ticket as fixed (referencing this PR) and closing it an automated process?

@bas-t  no, no automated process, but cannot login to trac currently. no idea what's broken. if trac is working again i will close the ticket.

BigNoid:
If you add ```Fixes #1662``` to the PR title it should automatically close the ticket when merged.

ksooo:
@BigNoid does this work for trac tickets? I thought this is only for github issues, no?

BigNoid:
There's some trac integration afaik. Github Janitor will close the ticket.
See #4115 and http://trac.kodi.tv/ticket/14885 for an example.

ksooo:
Cool. Thanks. Will change PR title, then.

ksooo:
Will changing the PR title do the trick or do I have to change one if the commit messages?

BigNoid:
I am not sure tbh. Let's see what happens :) 
We can always close it manually.

ksooo:
@BigNoid changing PR title seems not to be enough. Ticket got not closed automatically on merge. One needs to put 'fixes #xxxxx' into one of the involved commit messages.

xbmc/pvr/windows/GUIWindowPVRGuide.cpp
```diff
@@ -97,11 +97,19 @@ void CGUIWindowPVRGuide::StopRefreshTimelineItemsThread()
...
+  if (g_PVRTimers)
+    g_PVRTimers->RegisterObserver(this);
+  CGUIWindowPVRBase::RegisterObservers();
 }
 
 void CGUIWindowPVRGuide::UnregisterObservers(void)
 {
+  CSingleLock lock(m_critSection);
+  CGUIWindowPVRBase::UnregisterObservers();
+  if (g_PVRTimers)
```
FernetMenta: this will cause the system crash when PVRManager is restarted because i.e. another clients gets connects. There is no guarantee the the lifetime of Timers is longer than that of this window.

FernetMenta: IMO the entire observes should go away from GUIWindows. This pattern is not suited for this purpose. Why don't you use window messages?

ksooo: Just to confirm that we are on the same page: problem here is that g_PVRTimers (which is a CPVRTimers *) can get invalid between call to <code>if (g_PVR_Timers)</code> and <code>g_VRTimers->...</code>, right? OMG, we have this pattern all over the place...

> Why don't you use window messages?

 I'm not the original author of this code, but valid point.

 Would it be okay to use window messages inside the PVR core classes, like CPVRTimers? I think so.

FernetMenta: Your first statement is correct. If it does not access an invalid pointer, it may unregister at a different instance which is not correct either. I don't think we should switch to shared pointers as a solution, the problem is the observer pattern that is an anti pattern here.
I would use window messages, and maybe route them through a proxy object.

ksooo:
ok. what would be the purpose of the proxy object?

xbmc/pvr/windows/GUIWindowPVRGuide.cpp
ksooo: ok. what would be the purpose of the proxy object?

FernetMenta: Not having a dependency from every pvr child component to windowmanager. maybe pvrmanager can act as the proxy but not sure if this makes sense without looking into the details.
which is quite hard atm because Telekom killed my internet connection :(

ksooo:
> Not having a dependency from every pvr child component to windowmanager

 Absolutely. That was also my take on you suggestion.

> pvrmanager can act as the proxy

 I thought about this too, but wasn't pvrmanager getting restarted part of the original story? Does 'restarted' mean a new pvrmanager instance gets created, then?

xbmc/pvr/windows/GUIWindowPVRGuide.cpp
ksooo: > Not having a dependency from every pvr child component to windowmanager

 Absolutely. That was also my take on your suggestion.

> pvrmanager can act as the proxy

 I thought about this too, but wasn't pvrmanager getting restarted part of the original story? Does 'restarted' mean a new pvrmanager instance gets created, then?

FernetMenta: PVRManager instance is owned by ServiceManager which is owned by CApplication. A restart of PVRManager won't create a new instance.

ksooo: > PVRManager instance is owned by ServiceManager which is owned by CApplication. A restart of PVRManager won't create a new instance.

 yeah. looked this up myself meanwhile. good that you confirm. so, no lifecycle issues when using g_PVRManager as proxy for messaging.

using window messages as replacement for current observer approach relies on the fact that there can only be max one active window at a time, right? messaging is 1:1, observer is 1:n communication, right? So, what if currently there are non-window observers (have not checked this yet) the messaging approach wont work, no.

FernetMenta: > using window messages as replacement for current observer approach relies on the fact that there can only be max one active window at a time, right?

not correct. 

regarding non-window observers, I think we should treat this as a separate topic.

ksooo: never dealt with kodi window messaging before. How does broasdcasting messages work?

> regarding non-window observers, I think we should treat this as a separate topic.
 
ok

ksooo: what do you think about this approach, which does not need window messages:

pvr components will be no longer Observables, instead of calling NotifyObservers() directly like they do now, they will call g_PVRManager.NotifyObservers(), passing the same observable message like they do today. the pvr windows no longer observe different pvr components, but the pvr manager. problem solved.

FernetMenta: sounds reasonable :)

ksooo:
Okay, I will see what I can make out of this.

xbmc/pvr/windows/GUIWindowPVRGuide.cpp
ksooo: Okay, I will see what I can make out of this.

PR #10037, opened 2016-06-28
[addons] Add Window.Property(Addon.ID) to addon settings dialog

BigNoid:
Can be used to show the add-on icon for example in that dialog.

@phil65


MartijnKaijser:
@ronie ?

ronie:
fine with me

PR #10038, opened 2016-06-28
FFmpeg deprecate old API methods

fritsch:
This is a placeholder, so that no one needs to start from scratch again. DVDVideoCodecFFmpeg and DVDDemuxFFmpeg I left out, cause these need major rework, including the hw decoder to react on EAGAIN.

### DIFF
--- a/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp
+++ b/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp
@@ -3036,17 +3036,7 @@ IAESound *CActiveAE::MakeSound(const std::string& file)
       else
         sent = -1;
 
-      if (sent != -1)
-      {
-        receive_frame_error = avcodec_receive_frame(dec_ctx, decoded_frame);
-
-        if (receive_frame_error >= 0)
-          got_frame = 1;
-        else if (receive_frame_error != AVERROR(EAGAIN))
-          received = -1;
-      }
-
-      if (sent == -1 || received == -1)
+      if (sent == -1)
       {
         av_frame_free(&decoded_frame);
         avcodec_free_context(&dec_ctx);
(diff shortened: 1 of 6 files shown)


FernetMenta:
please bump ffmpeg in a separate pr

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp
```diff
@@ -3021,12 +3020,19 @@ IAESound *CActiveAE::MakeSound(const std::string& file)
...
     // decode until eof
     av_init_packet(&avpkt);
-    int len;
+    int ret = 0;
     while (av_read_frame(fmt_ctx, &avpkt) >= 0)
     {
       int got_frame = 0;
-      len = avcodec_decode_audio4(dec_ctx, decoded_frame, &got_frame, &avpkt);
-      if (len < 0)
+      ret = avcodec_send_packet(dec_ctx, &avpkt);
```
FernetMenta: I don't like this. a return value of a function should be checked for error right after the call, not 10 lines after.

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp
```diff
@@ -3021,12 +3020,19 @@ IAESound *CActiveAE::MakeSound(const std::string& file)
...
+
+      if (ret >= 0)
+        ret = avcodec_receive_frame(dec_ctx, decoded_frame);
+
+      if (ret >= 0)
+        got_frame = 1;
+
+      if (ret < 0)
```
FernetMenta: that is still wrong btw. AVERROR(EGAIN) should not be treated as an error. that hapens becuase you try to reuse the same ret code for 2 different functions. keep it simple

fritsch:
Bump separated: https://github.com/xbmc/xbmc/pull/10039

FernetMenta:
> DVDVideoCodecFFmpeg and DVDDemuxFFmpeg I left out, cause these need major rework, including the hw decoder to react on EAGAIN.

this is btw not correct. only minimal adjustments are required.

fritsch:
Okay - then I am thankful for a PR for these two.

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp
```diff
@@ -3059,6 +3049,30 @@ IAESound *CActiveAE::MakeSound(const std::string& file)
         delete sound;
         return NULL;
       }
+
+      if (sent != -1)
+      {
+        receive_frame_error = avcodec_receive_frame(dec_ctx, decoded_frame);
```
FernetMenta: I think this this method could be simplified a lot if it handled read and send in a loop, then receive in a separate one

fritsch:
Obsolete - WIP

PR #10042, opened 2016-06-29
[PVR] Guide window: Fix timer icons for manual (not epg-tag based) timers

ksooo:
A manual timer (with a fixed start and end time) can cover multiple broadcasts (epg events). With this PR, all epg events covered by a manual timer get the right timer association. Visible effect in PVR Guide window: all epg events completely covered by a manual timer now have a timer icon. Before this PR, no or only the first covered epg event had a timer icon.

Second commit is just non-functional code cleanup.

@Jalle19 for review?

### DIFF
--- a/xbmc/pvr/timers/PVRTimerInfoTag.cpp
+++ b/xbmc/pvr/timers/PVRTimerInfoTag.cpp
@@ -988,17 +988,22 @@ CEpgInfoTagPtr CPVRTimerInfoTag::GetEpgInfoTag(bool bCreate /* = true */) const
   return m_epgTag;
 }
 
-void CPVRTimerInfoTag::ClearEpgTag(void)
+void CPVRTimerInfoTag::SetEpgTag(const CEpgInfoTagPtr &tag)
 {
-  CEpgInfoTagPtr deletedTag;
+  CEpgInfoTagPtr previousTag;
   {
     CSingleLock lock(m_critSection);
-    deletedTag = m_epgTag;
-    m_epgTag.reset();
+    previousTag = m_epgTag;
+    m_epgTag = tag;
   }
 
-  if (deletedTag)
-    deletedTag->ClearTimer();
+  if (previousTag)
+    previousTag->ClearTimer();
+}
+
+void CPVRTimerInfoTag::ClearEpgTag(void)
+{
+  SetEpgTag(CEpgInfoTagPtr());
 }
 
 CPVRChannelPtr CPVRTimerInfoTag::ChannelTag(void) const
@@ -1006,10 +1011,11 @@ CPVRChannelPtr CPVRTimerInfoTag::ChannelTag(void) const
   return m_channel;
 }
 
-void CPVRTimerInfoTag::UpdateChannel(void)
+CPVRChannelPtr CPVRTimerInfoTag::UpdateChannel(void)
 {
   CSingleLock lock(m_critSection);
   m_channel = g_PVRChannelGroups->Get(m_bIsRadio)->GetGroupAll()->GetByUniqueID(m_iClientChannelUid, m_iClientId);
+  return m_channel;
 }
 
 const std::string& CPVRTimerInfoTag::Title(void) const
--- a/xbmc/pvr/timers/PVRTimers.cpp
+++ b/xbmc/pvr/timers/PVRTimers.cpp
@@ -105,24 +105,38 @@ bool CPVRTimers::IsRecording(void) const
 
 bool CPVRTimers::SetEpgTagTimer(const CPVRTimerInfoTagPtr &timer)
 {
-  const CEpgInfoTagPtr epgTag(timer->GetEpgInfoTag());
-  if (epgTag)
-  {
-    epgTag->SetTimer(timer);
-    return true;
-  }
-  return false;
+  if (timer->IsRepeating() || timer->m_bStartAnyTime || timer->m_bEndAnyTime)
+    return false;
+
+  std::vector<CEpgInfoTagPtr> tags(g_EpgContainer.GetEpgTagsForTimer(timer));
+
+  if (tags.empty())
+    return false;
+
+  // assign first matching epg tag to the timer.
+  timer->SetEpgTag(tags.front());
+
+  // assign timer to every matching epg tag.
+  for (const auto &tag : tags)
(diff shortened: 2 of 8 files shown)


xbmc/pvr/timers/PVRTimerInfoTag.cpp
```diff
@@ -988,6 +988,13 @@ CEpgInfoTagPtr CPVRTimerInfoTag::GetEpgInfoTag(bool bCreate /* = true */) const
   return m_epgTag;
 }
 
+void CPVRTimerInfoTag::SetEpgTag(const CEpgInfoTagPtr &tag)
+{
+  CSingleLock lock(m_critSection);
```
Jalle19: A double lock will be introduced here (`ClearEpgTaga()` also locks)

xbmc/pvr/timers/PVRTimers.cpp
```diff
@@ -105,24 +105,38 @@ bool CPVRTimers::IsRecording(void) const
...
+  if (timer->m_bStartAnyTime || timer->m_bEndAnyTime)
+    return false;
+
+  std::vector<CEpgInfoTagPtr> tags(g_EpgContainer.GetEpgTagsForTimer(timer));
+
+  if (tags.empty())
+    return false;
+
+  // assign first matching epg tag to the timer.
+  timer->SetEpgTag(tags.front());
```
Jalle19: Does the timer have to have an EPG tag associated? If we could skip this we could avoid the potential locking issues introduced by `SetEpgTag()`.

xbmc/pvr/windows/GUIWindowPVRBase.cpp
```diff
@@ -587,8 +587,12 @@ bool CGUIWindowPVRBase::EditTimer(CFileItem *item)
...
   }
 
-  if (ShowTimerSettings(timer) && !timer->GetTimerType()->IsReadOnly())
-    return g_PVRTimers->UpdateTimer(timer);
+  // clone the timer.
+  const CPVRTimerInfoTagPtr newTimer(new CPVRTimerInfoTag);
+  newTimer->UpdateEntry(timer);
+
+  if (ShowTimerSettings(newTimer) && !timer->GetTimerType()->IsReadOnly())
+    return g_PVRTimers->UpdateTimer(newTimer);
```
Jalle19: What is the purpose of this change?

ksooo: This is actually an ancient bug that pops up in this context and not really easy to explain. Took me a while to understand.

Creating timers works the way that a new timer tag not known to CPVRTimers singleton is created which will be send to the addon. Addon signals timer changes. Kodi calls GetTimers at the addon and syncs local timer list with new list obtained via GetTimers. During sync the new timer will be identified inserted at CPVRTimers singleton. https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/timers/PVRTimers.cpp#L186 All good.

Updating timers before this change works different. There the timer tag known to the CPVRTimers singleton will be updated directly and immediately, even before timer update has been confirmed by the backend. This is inconstent compared to new timer handling and leads to other problems.

Example: I create a manual timer, lets say with start/end 14:00/18:00  that spans 3 epg events on a certain channel. <code>CPVRTimers::UpdateEntries</code> inserts a corresponding timer tag at CPVRTimers singleton. This tag gets associated with the 3 abovementioned epg tags. PVR Guide window reflects this (3 timer icons). Now I change the end time for the timer from 18:00 to 16:00. let's assume this leads to the timer no longer covering the 3rd epg event. Kodi calls UpdateTimer at addon, then. Due to the fact that the origin timer tag is changed directly, <code>CPVRTimers::UpdateEntries</code> triggerd by UpdateTimer has no performant way to determine the 3rd epg tag with the timer that must be removed. Remember, the associated tag already has the new end time set. The information that the timer before was set to end at 18:00 is lost. With this change, this is no longer a problem.

Another problem with existing implementation is that an updated event might not get correctly identifies as updated by <code>CPVRTimers::UpdateEntries</code>, because the item in existing timer list is logically identically with the one delivered with new timer list.

Uff. HTH?

Jalle19: Sounds complicated but sane. Does this mean timee updates will only be reflected in the UI once the backend has processes the request and triggered Kodi to update timers?

ksooo: > Does this mean timer updates will only be reflected in the UI once the backend has processes the request and triggered Kodi to update timers?

Exactly. My take is that this was always the intention of the original authors of the code, but it was not correctly implemented or it broke at some point in time (maybe even by changes I did).

Jalle19: Sounds good

xbmc/pvr/timers/PVRTimers.cpp
ksooo: <code>CPVRTimerInfoTag::GetEpgTag</code> is called a gazillion of times. Obtaining the tag on demand is quite expensive as the epg tag must be found among potentially tens of thousands of epg events. So, if you found it once, hold it (or set it from outside if you already have it).

On my long list of things to improve at some later point in time is removal of the circular dependency between CPVRTimerInfoTag and CEpgInfoTag instances, though. Tried it once, but it quickly turned out that this is not a no-brainer.

xbmc/pvr/timers/PVRTimerInfoTag.cpp
ksooo: Is this really an issue? SetEpgTag is not called that often (once per timer per timer list update) and is not performance critical at all.

Jalle19: Probably not but it tends to lead to obscure bugs further down the line.

ksooo:
jenkins build this please

ksooo:
jenkins build this please

ksooo:
jenkins build this please

ksooo:
@Jalle19 no more double lock in <code>CPVRTimerInfoTag::SetEpgTag</code>. Good to go now?

ksooo:
jenkins build this please

Jalle19:
+1

PR #10043, opened 2016-06-29
DVDAudioCodecFFmpeg: Do not access private AVFrame fields

fritsch:
We access some fields that we should not access. Keep it open I think I can find some more.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Audio/DVDAudioCodecFFmpeg.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Audio/DVDAudioCodecFFmpeg.cpp
@@ -218,11 +218,11 @@ int CDVDAudioCodecFFmpeg::GetData(uint8_t** dst)
 {
   if(m_gotFrame)
   {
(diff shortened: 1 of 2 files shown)


jamrial:
https://github.com/xbmc/xbmc/blob/master/xbmc/guilib/FFmpegImage.cpp#L287

frame->pkt_duration should be av_frame_get_pkt_duration(frame)

fritsch:
Thanks much, added a commit.

FernetMenta:
this is for v18

FernetMenta:
reason: this is api cosmetic that has zero advantage on behavior. on the other hand it introduces a risk of breaking anything. Makes the decision very simple. This won't get considered for v17. Keep in mind that Kodi does not support shared ffmpeg libs hence ABI compatibility is not an issue.

fritsch:
As you wish as you decide. My point on this: Broken API usage is broken and wrong no matter what we support and what we don't support. I will label it v18.

fritsch:
Obsolete (soon).

PR #10044, opened 2016-06-30
Advancedsettings: Add a setting to disable resolution up switching

fritsch:
I have no strong feeling about this. This workarounds:
- 4k TV behind 1080p only AVR (upswitch leads to blackscreen)
- Broken modelines
- 3D SBS issues with Sony TVs (http://forum.kodi.tv/showthread.php?tid=241359&page=4)
- Broken and old intel drivers on windows (chances > 0 that it works with latest drivers)

It only affects a small number of users, so a GUI setting is not needed. Do what you want with it.

Usage:

```
<advancedsettings>
<video>
<adjustrefreshrate>
<allowresolutionchange>false</allowresolutionchange>
</adjustrefreshrate>
</video>
</advancedsettings>
```

### DIFF
--- a/xbmc/guilib/Resolution.cpp
+++ b/xbmc/guilib/Resolution.cpp
@@ -228,7 +228,8 @@ RESOLUTION CResolutionUtils::FindClosestResolution(float fps, int width, bool is
          (info.iScreenWidth < orig.iScreenWidth) || // new res is smaller
          (info.iScreenHeight < orig.iScreenHeight) || // new height would be smaller
          (info.dwFlags & D3DPRESENTFLAG_MODEMASK) != (curr.dwFlags & D3DPRESENTFLAG_MODEMASK) || // don't switch to interlaced modes
-         (info.iScreen != curr.iScreen)) // skip not current displays
+         (info.iScreen != curr.iScreen) || // skip not current displays
+         (!g_advancedSettings.m_allowResolutionChange)) // user does not want us to switch resolution
       {
         continue;
       }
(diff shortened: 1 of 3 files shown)


FernetMenta:
The purpose of advanced settings is not to implement workarounds for small groups of users. In this case the known amount of affected users is 1?
What is the root cause. Did you try to fix it?

afedchin:
> Broken intel drivers on windows

what do you mean?

fritsch:
You remember the intel devices that switched to 3840x2160 with AdjustRefreshrate enabled, we discussed in the forum? This can be workarounded with this, too.

afedchin:
I don't remember. Can you give me a link to the thread?

fritsch:
Here: http://forum.kodi.tv/showthread.php?tid=255772&pid=2246685#pid2246685 (iirc)

afedchin:
the thread is about half year old and perhaps the latest drivers has no the issue.

fritsch:
@afedchin is that a reason to not ship that option at all?

fritsch:
@FernetMenta 
Users have 1080p AVRs, but 4k TVs. AVR adds its audio EDID information to TV's edid but is not able to passthrough a 4k image -> hardware failure, not fixable via software. So when user plays a 4K video and has Adjust Refreshrate to Match Video enable, we switch to 4k -> blackscreen for user.

As said - I don't care a single cent about that feature or workaround - I am not affected. If it is considered useful, we take it - if not - we just close it.

afedchin:
I have no objection. It just  was interesting

FernetMenta:
This is a workaround at the wrong place. The root cause seems to be invalid or not wanted resolutions. You could create a dialog for blacklisting resolutions. This would also be the first step in long demanded native res output. IMO much better then dead end advancedsettings.

fritsch:
Yes. You are right here.
Am 30.06.2016 4:42 nachm. schrieb "Rainer Hochecker" <
notifications@github.com>:

fritsch:
Settings harm ... as users don't know what to set. Be more apple like ...

PR #10049, opened 2016-07-01
[eventclients] Nuke J2ME

hudokkow:
It's broken and if no one noticed before, no one is using it.

J2ME debian package doesn't even contain the jar file and not much going on since [2009](http://forum.kodi.tv/showthread.php?tid=32396)

Not sure its worth keeping java stuff under lib/ and examples/ in https://github.com/xbmc/xbmc/tree/master/tools/EventClients

@wsnipex, package kodi-eventclients-j2me can go.

@topfs2, since you're the original author, thoughts?


MartijnKaijser:
jenkins build this please

hudokkow:
I'm pulling the trigger tonight if no one objects.
Needs wiki work but I'll take care of it.

PR #8810, opened 2016-01-08
[Win32] Fix for rotated videos.

afedchin:
proper fix for rotated videos. 

@da-anda @phate89 can you test this?

PR #10050, opened 2016-07-01
rbp: Update transposed video scaling to match other platforms

popcornmix:


popcornmix:
This should match change in https://github.com/xbmc/xbmc/pull/8810

jenkins build this please

PR #10051, opened 2016-07-01
mmal_codec: Use EOS through codec to determine drain is complete

popcornmix:
Rather than relying on a timeout from codec, feed an EOS through to ensure all frames have been returned


popcornmix:
jenkins build this please

PR #10052, opened 2016-07-01
mmalcodec: Add another buffer when deinterlacing

popcornmix:
See: http://forum.kodi.tv/showthread.php?tid=276372


popcornmix:
jenkins build this please

PR #10053, opened 2016-07-01
omxvideo: Remove call to AutoInterlaceMethod. Treat auto as advanced

popcornmix:
There used to be performance issues in firmware so we only used advanced deinterlace for SD, but that is now resolved so make auto choose advanced deinterlace.


popcornmix:
jenkins build this please

PR #10054, opened 2016-07-01
[chg] do not allow minimize if runing in standalone mode

stefansaraev:
imo, it makes no sense to allow this if runing with --standalone, as in some situations (egl, or x11 without windowmanager), it may not be possible at all to bring kodi back to front

we already check if we are runing in standalone mode and hide "minimize" in powersaving options on non-darwin.

{Open|Libre}ELEC uses a similar patch to disable minimize. this one should be a bit more acceeptable.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -4922,7 +4922,10 @@ bool CApplication::SwitchToFullScreen(bool force /* = false */)
 
 void CApplication::Minimize()
 {
-  g_Windowing.Minimize();
+#if !defined(TARGET_DARWIN_IOS)
+  if (!g_application.IsStandAlone())
+#endif
+    g_Windowing.Minimize();
 }
 
 std::string CApplication::GetCurrentPlayer()


stefansaraev:
@wsnipex any clue why we have --standalone in tools/Linux/kodi.desktop? this feels wrong to me...

FernetMenta:
standalone is something platform specific that btw. was never implemented properly. Application should not bother with this. please move the check to windowing.

stefansaraev:
sure, will do, once it is properly implemented

FernetMenta:
WinsystemX11 tests for existing windowmanager. That is already there and more appropriate then standalone. It would be a simple change in WinsystemX11::mimimize

stefansaraev:
as said, it is not only x11 affected.

FernetMenta:
then affected implemetations of Windowing need to take care for this as well. x11 already provides a method, Wondows, OSX, iOS are not affected. If you wnat to improve thing, I suggest that you start with X11.

stefansaraev:
if you looked at the code you would know why it's in application, no idea what standalone should be in your view, but to me it means "kodi is the only app I want to use, keep it always fullscreen, on top, and let it allow me sleep/suspend/shutdown and so on"

however, forget this, distros shipped the patch for years, they certainly will do so for another few years.

and sorry, but I wont start improving anything for a platform (linux/x11) that I dont use, and cant test myself

FernetMenta:
> and sorry, but I wont start improving anything for a platform (linux/x11) that I dont use, and cant test myself

fair enough, but you shouldn't make them worse either by those ugly ifdefs

PR #10057, opened 2016-07-01
[PVR] Timers: Allow to change timer type for existing timers.

ksooo:
This feature was requested several time sin the forum. With this PR it s possible in the timer settings dialog to edit the type of an _existing_ timer or timer rule, for instance to convert a one time timer to a series recording.

### DIFF
--- a/xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
+++ b/xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
@@ -114,36 +114,32 @@ void CGUIDialogPVRTimerSettings::SetTimer(const CPVRTimerInfoTagPtr &timer)
 
   m_timerInfoTag = timer;
 
-  if (!m_timerInfoTag)
-  {
-    CLog::Log(LOGERROR, "CGUIDialogPVRTimerSettings::SetTimer - No timer info tag");
-    return;
-  }
-
   // Copy data we need from tag. Do not modify the tag itself until Save()!
   m_timerType     = m_timerInfoTag->GetTimerType();
   m_bIsRadio      = m_timerInfoTag->m_bIsRadio;
   m_bIsNewTimer   = m_timerInfoTag->m_iClientIndex == PVR_TIMER_NO_CLIENT_INDEX;
-  m_bTimerActive  = m_bIsNewTimer || m_timerInfoTag->IsActive();
-  m_bStartAnyTime = m_bIsNewTimer || m_timerInfoTag->m_bStartAnyTime;
-  m_bEndAnyTime   = m_bIsNewTimer || m_timerInfoTag->m_bEndAnyTime;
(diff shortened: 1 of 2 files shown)


Tolriq:
@Ksooo how does that impact JSON ? Does it need API change to allow that ?
I've seen your comment on a previous PR, is the naming rewrite planned for v17 ?

ksooo:
@Tolriq no and no.

Tolriq:
@ksooo Anyway to have the just the JSON chance for v17 ? 
There's already a lot of breaking  chance in JSON API for v17.
Timer API was added in JSON in v17 alpha.
Currently no one use that API.

That would simplify a lot the future if v17 and v18 are not incompatible on a new feature.

ksooo:
Good point, not sure whether I find the time for this, though. PR welcome.

Tolriq:
Tell me what to rename to what :) I have plans to include Timers in Yatse but for the moment I have no idea how it works as I do not use PVR in Kodi. So I do not even understand your naming changes :)

ksooo:
"repeating timer" -> "timer rule". That's it.

Tolriq:
Ok sounds pretty simple :)

Will try to find time before moving to do that.

Tolriq:
@ksooo I've checked quickly current definition. Can you explain me what the new description should looks like for the setters ?

Currently it's "the add/delete/edit should be done for all occurances" for repeating, but I'm not sure to understand the difference of this new timer rule definition, so need your description as I suppose it's a little different behavior.

ksooo:
Come up with a suggestion in a PR and I will be happy to review.

ksooo:
@Jalle19 Good to go? 

@Tolriq thanks for the PR. Will have a look the next days.

ksooo:
Ooops.

xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
```diff
@@ -121,29 +121,38 @@ void CGUIDialogPVRTimerSettings::SetTimer(const CPVRTimerInfoTagPtr &timer)
   }
 
   // Copy data we need from tag. Do not modify the tag itself until Save()!
-  m_timerType     = m_timerInfoTag->GetTimerType();
+  m_timerType = m_timerInfoTag->GetTimerType();
+
+  if (!m_timerType)
```
Jalle19: Are there really any situations where a timer tag could be created and stored without a timer type?

Jalle19:
Apart from my previous comment, +1

xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
ksooo: In theory that's possible, yes. But you may be right, it's just theory and of no real value here. I will drop this.

Jalle19: Thanks

ksooo:
jenkins build this please

ksooo:
travis-ci errors are unrelated.

PR #10046, opened 2016-06-30
[PVR] Cleanup: Make timer internal terminology consistent, step 2

ksooo:
This is a followup to https://github.com/xbmc/xbmc/pull/8713/commits/df2604aa1339ebc4529e784c76f05143e1c73ce0 (Make timer UI terminology consistent: 'repeating timer' / 'timer schedule' -> 'timer rule) and https://github.com/xbmc/xbmc/pull/8713/commits/814b1277891043c1cd74aa18ebadc2456b605c4f (Make timer internal terminology consistent, step 1: 'timer schedule' -> 'timer rule)

Last step in a followup PR will be renaming at API level (PVR addon API, json, guiinfo).

No functional changes here.

@Jalle19 good to go?

PR #10058, opened 2016-07-02
[JSON] [PVR] Update JSON PVR Timer API 

Tolriq:
To match new naming (See #10046)

@ksooo / @Jalle19 

So it seems current implementation was broken and did not even serialize the repeating field, so not so much breaking change.

I'm not in PVR at all, but when looking at the serialization problem I saw the "type" with possible values that are not exposed to JSON, I suppose this should be exposed to JSON too, or at least the IsManual field.

And seeing the add/delete/toggle for timers on JSON that only takes broadcastid / timerrule option, maybe there's the need for someone more in PVR to validate that the correct timers are created with correct values / types. I test with demo PVR addon so hard to be sure of the values.

And to finish I'm not sure the description text for the timerrule param of add/delete/toggle timers is accurate, but I have no real idea of the global concept of this timerrule / type thing.

Edit : Seems there's also a concept of read only timers that are not exposed / serialized. But I guess you can't delete a readonly one so it should be exposed too.


Tolriq:
@ksooo the  [breaking change]  is mandatory for the PR that update version.txt and include breaking change please do not remove it, see other PRs for history

Tolriq:
Replaced by #10074

PR #10060, opened 2016-07-02
[addons] screeshot support

tamland:
See 2c4681c for usage example.

Required repository generator changes here:  https://github.com/tamland/xbmc-repo-gen/commits/master

To make screenshots access in repositories this PR adds a new `<assets>` element which is basically a manifest of the asset files the addon contains. The same is done for icon/fanart deprecating the use of the implicit icon.png/fanart.jpg location. This means icon, fanart and screenshot locations in the addon package **must now be explicitly listed in the `<assets>` element** when present. This also makes the `noicon` and `nofanart` elements obsolete.

To keep compatibility with older versions, keep the icon/fanart name and location unchanged and add:

``` xml
<assets>
    <icon>icon.png</icon>
    <fanart>fanart.jpg</fanart>
</assets>
```

to the `xbmc.addon.metadata` or `kodi.addon.metadata` extension point.

Todos/issues:
- [x] skinning
- [x] graphical issues from making slideshow window a dialog (transparent background, flickering, doesn't clear on close)

~~Note: this pr is branched off  #10059~~


BigNoid:
@tamland maybe I can help with the skinning todo/issue. What do you need?

tamland:
It's actual skinning:) I'm thinking we could have a thumb list in the addon info dialog (which opens the slideshow window). [Currenty](https://github.com/xbmc/xbmc/pull/10060/commits/73f8ee7345b7a38593d99082d797c420cca1043f) I just copy-pasted something on top of the info dialog to have something to demo. So, make that into something presentable. Or we can probably do something else if you have any ideas?

BigNoid:
Something like this perhaps:
screenshot002

Only thing that sucks is that the screenshots take a prominent place and it will take quite some time before all addons have the screenshots added and that there are a lot of addons where it doesn't make sense to have a preview.
We could use the old and new layout based on whether container ```50``` has any items imo.

tamland:
Looks awesome if you ask me:)

MartijnKaijser:
When will these images be visible? Only when opening the infoscreen or always? Will they be loaded from our server as well which will become bogged down again?
I do think it's great and all but i am worried about server load. Not sure if we can do cloudflare for just the images (maybe we do already??).

BigNoid:
Okay great :) The list is now hidden when there are no screenshots.  That should be changed if we want to show fallback images. 
@Martijn: Only in info screen.  Not sure if they are cached though.
@phil65 what do you think of the layout? The disclaimer now scrolls, not sure if that's an issue.

BigNoid:
Nvm can show fallback images when the list is hidden ofc doh

tamland:
@MartijnKaijser I've made it as lazy as it possibly can be: They are loaded when info screen is opened and thumb list shown. They're cached the same way icons/fanart currently are. Ie. only invalidated on version bump. How much load this will amount to, I'm not sure.

The images doesn't have to be *in* the addon repository, so what we can do is move all images to some different site and edit the locations in addons.xml. Then it wouldn't go though the redirector. But having cloudflare handle it would ofc be the easier solution..

MartijnKaijser:
@tamland ok sounds good to me.

BigNoid:
See https://github.com/BigNoid/xbmc/commit/11d15aeb79c075b8ab9cb6c2e1aae50119d33d03

tamland:
jenkins build this please

tamland:
@MartijnKaijser I'm soon done with this; could you have a look at deploying/testing the repo script?

MartijnKaijser:
should no impact on normal repo workings correct?

tamland:
Yes, it's the same and should be compatible with the old repos. The only change is that it will now extract the files mentioned in `<assets>`. If that doesn't exists it will fallback the the old way.

MartijnKaijser:
@tamland there's a bug in the script as it now removes all older add-on versions.
image
image

tamland:
Well that's a stupid mistake: https://github.com/tamland/xbmc-repo-gen/commit/9313b7e574560c5be1ce8cb68bf9759b4911f2f1 :\ Hopefully it only caused some minor annoyance.

MartijnKaijser:
only one plugin was affected afaik

tamland:
jenkins build this please

MartijnKaijser:
@tamland seems it still deletes all except 1 version

tamland:
@MartijnKaijser so.. all good? It looks ok in my test mirror at least: https://github.com/tamland/test-mirror/tree/master/krypton

MartijnKaijser:
seems to work ok again. thx

HitcherUK:
@tamland Does this only work for skins in a repo and not local ones?

gade01:
> @tamland Does this only work for skins in a repo and not local ones?

@HitcherUK I made this work locally without the updated feature in the repo version. I moved the files (icon, fanart, screenshots) to a new resources folder.

HitcherUK:
@gade01 Got a skin I can check with because it's not working for me in fTV or Estuary?

gade01:
@HitcherUK Yup, Rapier 10.3.2 supports the screenshot feature: https://github.com/gade01/repository.gade/blob/master/krypton/skin.rapier/skin.rapier-10.3.2.zip?raw=true

tamland:
It should work with local ones. e7ac9c1, which is included in this pr is local only.

addons/skin.estuary/1080i/DialogAddonInfo.xml
```diff
@@ -8,45 +8,101 @@
...
+				<width>1215</width>
+				<height>233</height>
+				<ondown>9000</ondown>
+				<onup>9000</onup>
+				<orientation>horizontal</orientation>
+				<itemlayout height="225" width="405">
+					<control type="image">
+						<width>385</width>
+						<height>220</height>
+						<texture border="2" fallback="DefaultAddonNone.png" background="true">$INFO[ListItem.Icon]</texture>
```
tamland: @BigNoid shouldn't this be DefaultNoPreview.png?
Not sure it it's a skin problem or not but the fallback image isn't showing while the screenshot thumbs are being loaded. It's just blank. Any ideas?

BigNoid: Yeah should be DefaultNoPreview.png. But that's not the issue. If the screenshots are there than the  icon is not empty (has an url) and the fallback will not kick in and you just have to wait for the image to be downloaded. The fallback can be removed as it's not used here. What I can do is show DefaultNoPreview.png always behind the screenshots so that it shows that in case of slow connection.

tamland: Ah, makes sense. I see that's what the actor list does. Would be great if you could pr that:) Otherwise I can probably copy-paste the actor one

zag2me:
@tamland apologies for the bump, but how would you suggest I target both older versions of Kodi and Krypton by while using this new <asset> tag. I've written an Add-on and would like to use this new method, but it of course won't show an icon on older Kodi versions.

Is it acceptable to add a 2nd icon to the root? Or do you suggest another way?

PR #10062, opened 2016-07-02
[addons] fix default sort order in recently updated

tamland:
should have been descending.
@MartijnKaijser


MartijnKaijser:
Thx. 
Is sort order still available or just removed in that view? Doesn't matter to though. 
jenkins build this please

tamland:
You can still toggle it. It was just being overridden by SetSortOrder (it's set to descending above);)

PR #10063, opened 2016-07-02
Random fixes found with clang/gcc sanitizers

neo1973:
This fixes are pretty simple, but they maybe have side effects I'm not aware of. If you are interested in the sanitizers error messages which lead to these patches, I can try to reproduce them.

### DIFF
--- a/xbmc/AutoSwitch.cpp
+++ b/xbmc/AutoSwitch.cpp
@@ -238,5 +238,5 @@ float CAutoSwitch::MetadataPercentage(const CFileItemList &vecItems)
     if(item->IsParentFolder())
       total--;
   }
-  return (float)count / total;
+  return (total != 0) ? ((float)count / total) : 0.0f;
 }
--- a/xbmc/cores/VideoPlayer/DVDMessage.cpp
+++ b/xbmc/cores/VideoPlayer/DVDMessage.cpp
@@ -88,13 +88,9 @@ bool CDVDMsgGeneralSynchronize::Wait(unsigned int milliseconds, unsigned int sou
   return true;
 }
 
-void CDVDMsgGeneralSynchronize::Wait(volatile bool *abort, unsigned int source)
+void CDVDMsgGeneralSynchronize::Wait(std::atomic<bool>& abort, unsigned int source)
 {
-  while(!Wait(100, source))
-  {
-    if(abort && *abort)
-      return;
-  }
+  while(!Wait(100, source) && !abort);
 }
 
 long CDVDMsgGeneralSynchronize::Release()
--- a/xbmc/guilib/GUIFontCache.cpp
+++ b/xbmc/guilib/GUIFontCache.cpp
@@ -32,6 +32,10 @@ class CGUIFontCacheImpl
     using HashIter = typename HashMap::iterator;
     using AgeMap = std::multimap<size_t, HashIter>;
 
+    ~EntryList()
+    {
+      Flush();
+    }
     HashIter Insert(size_t hash, CGUIFontCacheEntry<Position, Value> *v)
     {
       auto r (hashMap.insert(typename HashMap::value_type(hash, v)));
--- a/xbmc/input/KeyboardLayout.cpp
+++ b/xbmc/input/KeyboardLayout.cpp
@@ -69,7 +69,7 @@ bool CKeyboardLayout::Load(const TiXmlElement* element)
 
   const TiXmlElement *keyboard = element->FirstChildElement("keyboard");
   if (element->Attribute("codingtable"))
-    m_codingtable = CInputCodingTableFactory::CreateCodingTable(element->Attribute("codingtable"), element);
+    m_codingtable = IInputCodingTablePtr(CInputCodingTableFactory::CreateCodingTable(element->Attribute("codingtable"), element));
   else
     m_codingtable = NULL;
   while (keyboard != NULL)
--- a/xbmc/threads/Event.cpp
+++ b/xbmc/threads/Event.cpp
@@ -120,6 +120,7 @@ namespace XbmcThreads
          signaled == NULL && iter != events.end(); ++iter)
     {
       CEvent* cur = *iter;
+      CSingleLock lock2(cur->mutex);
       if (cur->signaled) 
         signaled = cur;
     }
--- a/xbmc/cores/VideoPlayer/DVDMessage.h
+++ b/xbmc/cores/VideoPlayer/DVDMessage.h
@@ -30,6 +30,7 @@
 
 // include as less is possible to prevent dependencies
 #include "DVDResource.h"
+#include <atomic>
 #include <string>
 #include <string.h>
 
@@ -148,8 +149,8 @@ public:
 
   // waits until all threads waiting, released the object
   // if abort is set somehow
-  bool Wait(unsigned int   ms   , unsigned int source);
-  void Wait(volatile bool *abort, unsigned int source);
+  bool Wait(unsigned int ms         , unsigned int source);
+  void Wait(std::atomic<bool>& abort, unsigned int source);
 private:
   class CDVDMsgGeneralSynchronizePriv* m_p;
 };
--- a/xbmc/cores/VideoPlayer/DVDMessageQueue.cpp
+++ b/xbmc/cores/VideoPlayer/DVDMessageQueue.cpp
@@ -246,7 +246,7 @@ void CDVDMessageQueue::WaitUntilEmpty()
   CLog::Log(LOGNOTICE, "CDVDMessageQueue(%s)::WaitUntilEmpty", m_owner.c_str());
   CDVDMsgGeneralSynchronize* msg = new CDVDMsgGeneralSynchronize(40000, 0);
   Put(msg->Acquire());
-  msg->Wait(&m_bAbortRequest, 0);
+  msg->Wait(m_bAbortRequest, 0);
   msg->Release();
 }
 
--- a/xbmc/cores/VideoPlayer/DVDMessageQueue.h
+++ b/xbmc/cores/VideoPlayer/DVDMessageQueue.h
@@ -21,6 +21,7 @@
  */
 
 #include "DVDMessage.h"
+#include <atomic>
 #include <string>
 #include <list>
 #include <algorithm>
@@ -111,7 +112,7 @@ private:
   CEvent m_hEvent;
   mutable CCriticalSection m_section;
 
-  bool m_bAbortRequest;
+  std::atomic<bool> m_bAbortRequest;
   bool m_bInitialized;
 
   int m_iDataSize;
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -2368,7 +2368,7 @@ void CVideoPlayer::SynchronizeDemuxer(unsigned int timeout)
 
   CDVDMsgGeneralSynchronize* message = new CDVDMsgGeneralSynchronize(timeout, 0);
   m_messenger.Put(message->Acquire());
-  message->Wait(&m_bStop, 0);
+  message->Wait(m_bStop, 0);
   message->Release();
 }
 
@@ -3938,7 +3938,7 @@ void CVideoPlayer::FlushBuffers(bool queued, double pts, bool accurate, bool syn
       CDVDMsgGeneralSynchronize* msg = new CDVDMsgGeneralSynchronize(1000, 0);
       m_VideoPlayerAudio->SendMessage(msg->Acquire(), 1);
       m_VideoPlayerVideo->SendMessage(msg->Acquire(), 1);
-      msg->Wait(&m_bStop, 0);
+      msg->Wait(m_bStop, 0);
       msg->Release();
 
       // purge any pending PLAYER_STARTED messages
--- a/xbmc/dialogs/GUIDialogKeyboardGeneric.h
+++ b/xbmc/dialogs/GUIDialogKeyboardGeneric.h
@@ -87,7 +87,7 @@ class CGUIDialogKeyboardGeneric : public CGUIDialog, public CGUIKeyboard
     std::string m_strHeading;
     std::string m_text;       ///< current text
 
-    IInputCodingTable *m_codingtable;
+    IInputCodingTablePtr m_codingtable;
     std::vector<std::wstring> m_words;
     std::string m_hzcode;
     int         m_pos;
--- a/xbmc/input/InputCodingTable.h
+++ b/xbmc/input/InputCodingTable.h
@@ -21,6 +21,7 @@
 */
 
 #include <string>
+#include <memory>
 #include <vector>
 
 class IInputCodingTable
@@ -58,3 +59,5 @@ public:
 protected:
   std::string m_codechars;
 };
+
+typedef std::shared_ptr<IInputCodingTable> IInputCodingTablePtr;
--- a/xbmc/input/KeyboardLayout.h
+++ b/xbmc/input/KeyboardLayout.h
@@ -23,15 +23,16 @@
 #include <string>
 #include <vector>
 
+#include "InputCodingTable.h"
+
 class TiXmlElement;
-class IInputCodingTable;
 
 class CKeyboardLayout
 {
 public:
   CKeyboardLayout();
   virtual ~CKeyboardLayout();
-  IInputCodingTable* GetCodingTable() { return m_codingtable; }
+  IInputCodingTablePtr GetCodingTable() { return m_codingtable; }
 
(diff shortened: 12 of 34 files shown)


MartijnKaijser:
jenkins build this please

xbmc/cores/VideoPlayer/DVDMessage.cpp
```diff
@@ -88,11 +88,11 @@ bool CDVDMsgGeneralSynchronize::Wait(unsigned int milliseconds, unsigned int sou
   return true;
 }
 
-void CDVDMsgGeneralSynchronize::Wait(volatile bool *abort, unsigned int source)
+void CDVDMsgGeneralSynchronize::Wait(std::atomic<bool>& abort, unsigned int source)
 {
   while(!Wait(100, source))
   {
-    if(abort && *abort)
+    if(abort && abort)
```
fritsch: look again here. the old code checked the ptr for existence and then its value - this is not needed with the reference here. Remove one of the abort

xbmc/threads/Event.cpp
```diff
@@ -120,6 +120,7 @@ namespace XbmcThreads
          signaled == NULL && iter != events.end(); ++iter)
     {
       CEvent* cur = *iter;
+      CSingleLock lock2( cur->mutex);
```
fritsch: too much whitespace

xbmc/filesystem/NFSFile.cpp
```diff
@@ -351,11 +351,11 @@ void CNfsConnection::Deinit()
 /* This is called from CApplication::ProcessSlow() and is used to tell if nfs have been idle for too long */
 void CNfsConnection::CheckIfIdle()
 {
+  CSingleLock lock(*this);
   /* We check if there are open connections. This is done without a lock to not halt the mainthread. It should be thread safe as
    worst case scenario is that m_OpenConnections could read 0 and then changed to 1 if this happens it will enter the if wich will lead to another check, wich is locked.  */
   if (m_OpenConnections == 0 && m_pNfsContext != NULL)
   { /* I've set the the maxiumum IDLE time to be 1 min and 30 sec. */
-    CSingleLock lock(*this);
```
fritsch: See the comment above that change. It was ugly and not thread safe before, but the author knew what he did.

fritsch: Something else: the scope of that lock changed and now two locks are taken in that method - be very careful of additional deadlocks.

Let's ping the author: @Memphiz could you take a "lock" ;-) ?

xbmc/linux/LinuxTimezone.h
```diff
@@ -59,6 +60,7 @@ class CLinuxTimezone : public ISettingCallback, public ISettingsHandler
    std::map<std::string, std::string> m_countriesByTimezoneName;
 };
 
-extern CLinuxTimezone g_timezone;
+XBMC_GLOBAL_REF(CLinuxTimezone,g_timezone);
+#define g_timezone XBMC_GLOBAL_USE(CLinuxTimezone)
```
fritsch: Much more important here: Do you have a better idea to get rid of this global nogoes completely? They will kick us in the balls later on again and a proper design would help use a lot here.

xbmc/network/EventServer.cpp
```diff
@@ -203,7 +203,10 @@ void CEventServer::Run()
   // add our socket to the 'select' listener
   listener.AddSocket(m_pSocket);
 
-  m_bRunning = true;
+  {
+    CSingleLock lock(m_critSection);
```
fritsch: The way it was used before (e.g. not caring about parallel codeblocks) indicates that an atomic bool might be enough. Locking the critSection might have side effects on the way the class is used.

fritsch:
@FernetMenta https://github.com/xbmc/xbmc/pull/10063/commits/c2239ebe1a703f374a2763f2a9edaae4d0b5e61d - could you look at that one. Programmatically he is right of course, though our classes have a non general usage, so that might be overkill.

xbmc/guilib/GUIFontCache.cpp
```diff
@@ -32,6 +32,10 @@ class CGUIFontCacheImpl
     using HashIter = typename HashMap::iterator;
     using AgeMap = std::multimap<size_t, HashIter>;
 
+    ~EntryList()
+    {
+      Flush();
+    }
```
fritsch: While obviously right. Do we have a code exit that does not call Flush() manually?

xbmc/AutoSwitch.cpp
```diff
@@ -238,5 +238,5 @@ float CAutoSwitch::MetadataPercentage(const CFileItemList &vecItems)
     if(item->IsParentFolder())
       total--;
   }
-  return (float)count / total;
+  return (total != 0) ? ((float)count / total) : 0;
```
fritsch: a) Can a a vector with size n have n parentFolders()? 
b) Is it enough to early return if size == 0?

xbmc/linux/LinuxTimezone.h
FernetMenta: GlobalHandling is not supposed to be used anymore. every object is supposed to have an owner.

xbmc/threads/Thread.cpp
```diff
@@ -102,6 +102,7 @@ void CThread::Create(bool bAutoDelete, unsigned stacksize)
 
 bool CThread::IsRunning() const
 {
+  CSingleLock lock(m_CriticalSection);
```
FernetMenta: this is not needed because this method is not interested in the ID itself.

xbmc/guilib/GUIFontCache.cpp
neo1973: According to the AddressSanitizer/LeakSanitizer there was non deleted memory after closing Kodi. So a code path not calling Flush() is very likely.

neo1973:
Thanks for the review! I will address the issues and resubmit the pull request.

FernetMenta:
> I will address the issues and resubmit the pull request.

no need to resubmit. pull requests do observe changes of you branch

xbmc/threads/Thread.cpp
neo1973: Sorry, I don't understand. If the ID gets changed from zero to non zero or the other way round, the result of the expression changes. So the the value of m_ThreadId is important and should be synchronized, I believe.

And I noticed, that m_ThreadId isn't synchronized in the ThreadId() method too. I think this should be changed also. If a mutex is to heavy, maybe m_ThreadId could be changed to std::atomic<int>?

FernetMenta: we are not interest in the value of m_ThreadId here, only if it is null. if the value is only half written, won't change anything in bahavior

tamland: @neo1973 is correct; the lock/atomic is required here. Whether or not you are interested in the id have nothing to do with it. Without synchronization the value may not even be visible to readers.

xbmc/filesystem/NFSFile.cpp
Memphiz: this is fine as is and this commit doesn't fix anything - please drop it...

Memphiz: the intention here is to not hold the lock on idle check while nfs files are open ... just to be clear

xbmc/threads/Thread.cpp
FernetMenta: again m_ThreadId is of no interest. I have written this code.

FernetMenta: you are wrong

xbmc/linux/LinuxTimezone.h
neo1973: Ok. Then how about a method like `static CLinuxTimezone& CLinuxTimezone::getInstance()` which creates an instance of `CLinuxTimezone` on the first call?

FernetMenta: In order to have control every object needs an owner. There is an ongoing process to kill this globals and singletons one by one and we don't open the backdoor here.

xbmc/threads/Thread.cpp
tamland: I took the time to give an actual reason for why your augment that it's correct, is wrong. so waiting on you;)

FernetMenta: If you looked at the code, you would see that there is no sync wanted. Assume one thread calls IsRunning() and the result is true. Task switch, thread stops, thread 1 continues, now IsRunning is not true anymore.
The way this code is written and used, the result of IsRunning may not be valid for the next operation. Given this fact, it does not matter if there is a task switch when while m_ThreadId.

You won't be able to construct a scenario where this change here would have any influence to current behaviour.

If you wanted to do something like this:
"Perform an operation only if thread is running and make running thread aware of this (i.e. prevent from terminating)"
you would need a lock over more than a single statement.

neo1973: Ok, thanks for the explanation. Then how about removing the lock and making m_ThreadId atomic? This will make IsRunning() behave like most people expect it to while the impact on performance won't be noticeable. But if you think the code is correct as it is, I will drop this commit from the pull request. Your choice.

FernetMenta: making it atomic won't change behavior a bit.

fritsch: Perhaps to make it clearer:
T1:
1 Calls isrunning of t2
2 does sth with assumption t2 is there

T2:
1 join

If t2's join happens directly after t1's call to isrunning neither the
threadid nor the running status is worth anything without a more central
lock t1 and t2 share.
Am 04.07.2016 5:27 vorm. schrieb "Rainer Hochecker" <
notifications@github.com>:

tamland: I think you are talking about something entirely different.. The only important part is here:

>If you looked at the code, you would see that there is no sync wanted. Assume one thread calls IsRunning() and the result is true. Task switch, thread stops

Is it guaranteed that IsRunning is never called from a different thread when m_ThreadId is being written to in stop? If no, it's undefined behaviour.

FernetMenta: No, it is not undefined behaviour because the value of m_ThreadId is not of any interest. Lets assume it is only half written and as a result not null. The result of IsRunning would be true. A few processor cycles later the result would be false. As already mentioned several times here: The result of InRunning might not be valid anymore after the call. Neither a lock nor an atomic would change this fact.

Please construct a scenario where a lock would change behaviour. I bet you can't.

tamland: >No, it is not undefined behaviour

I recommend you to read http://en.cppreference.com/w/cpp/language/memory_model Specifically
"When an evaluation of an expression writes to a memory location and another evaluation reads or modifies the same memory location, the expressions are said to conflict. A program that has two conflicting evaluations has a data race unless either

both conflicting evaluations are atomic operations (see std::atomic)
one of the conflicting evaluations happens-before another (see std::memory_order) 

If a data race occurs, the behavior of the program is undefined.
(in particular, release of a std::mutex is synchronized-with, and therefore, happens-before acquisition of the same mutex by another thread, which makes it possible to use mutex locks to guard against data races) "

Thus, the way I see it, scenario 1:
T1: executes pThread->m_ThreadId = 0;
T2: executes pThread->m_ThreadId;  =undefined behaviour

scenario 2: change staticThread to this:
````
pThread->m_StartEvent.Set();
//use m_ThreadId store some temporary here that happens to be 0. to any other threads reading it it will be 0 but it's sequentially consistent and a perfectly valid optimization.
pThread->m_ThreadId = 0;
pThread->Action();
// lock during termination
CSingleLock lock(pThread->m_CriticalSection);
pThread->m_ThreadId = 0;
````
adding a lock or making it atomic ensures scenario 1 & 2 can't happen.

Comment (FernetMenta) [inline xbmc/threads/Thread.cpp]: Sorry I don't follow. I would have expected a scenario that shows how Kodi fails in any of its functions.

I agree that there is a race but this has no influence on Kodi's behavior.

Note that m_CriticalSection is private so your example above won't work

Comment (neo1973) [inline xbmc/threads/Thread.cpp]: Just for clarification: I didn't make this patch because I saw an actual error happen. But someone not familiar with the code may expects `IsRunning()` (and also `ThreadId()` btw.) to always return the actual state of the thread at any time.

Comment (FernetMenta) [inline xbmc/threads/Thread.cpp]: > to always return the actual state of the thread at any time

And I am trying to say that this does not matter regardless if there is a lock or not. Example, lets assume there is a lock as you suggested:

if (thread->IsRunning())
{
  - we expect the thread as running because IsRunning returned true, but
  - here comes the task switch and thread terminates
  - now we are in a block that is supposed to run if thread is active but it already terminated
}

Even with a lock or atomic the calling thread can't trust the return value if the thread we are querying can terminate.

Comment (neo1973) [inline xbmc/threads/Thread.cpp]: We still don't talk about the same thing. You are right that you can't assume the thread to still be running after `IsRunning()` returned true. But sometimes you are interested in the state at exact that one moment. This is for example the case in `CTimer::GetElapsedMilliseconds()` I believe (maybe I'm wrong, I haven't looked at it in detail).

Comment (FernetMenta) [inline xbmc/threads/Thread.cpp]: CTimer::GetElapsedMilliseconds() is a good example that underlines my statement. It won't work reliable regardless the lock. It implements exactly the scenario I gave above:

```
float CTimer::GetElapsedMilliseconds() const
{
  if (!IsRunning())
    return 0.0f;
 
 !!! no guarantees that we are still running. thread could have terminated already
  return (float)(XbmcThreads::SystemClockMillis() - (m_endTime - m_timeout));
}
```

Comment (neo1973) [inline xbmc/threads/Thread.cpp]: But see it the other way round: The thread could already be ended, but `IsRunning()` is still returning true. I really don't know if this could happen here, but this scenario is possible in an other context.

Regardless, you admitted that there is a data race. It maybe isn't relevant in the places where it is used today. But at some point someone could rely on the "correct" behaviour of `IsRunning()` and of `ThreadId()`. I think making `m_ThreadId` a std::atomic\<int> doesn't hurt and saves us from possible errors in the future.

Anyway, I think this discussion reached a dead end. If you still think there should be no synchronization, I remove this commit. The change can then be discussed in a separate  pull request if someone is interested in it.

Comment (FernetMenta) [inline xbmc/threads/Thread.cpp]: All I wanted to say is that a lock or atomic won't change anything in current behaviour but I am not against introducing an atomic per se. ThreadIdentifier is maybe not a good candidate for an atomic.

Comment (tamland) [inline xbmc/threads/Thread.cpp]: As explained FernetMenta all those things you are saying have nothing to do with why the lock/atomic is needed here. To re-cite: "When an evaluation of an expression writes to a memory location and another evaluation reads [the program has a data race]. If a data race occurs, the behavior of the program is undefined". Ie. both your examples have undefined behaviour at the 'if' where IsRunning is called. it does not matter what comes after.

**Comment by neo1973** [comment]
I did a rebase against master because of a merge conflict. I also removed the commits which where wrong and fixed up small mistakes in others. @fritsch about  a850e05: I'm not sure if a folder with n elements could have n parents, but during debugging this never happened.

**Comment by fritsch** [comment]
@neo1973 thx much - that question was an academic question, as when we would know - it would make the code much more easy, but fine as is.

**Comment by MartijnKaijser** [comment]
jenkins build this please

**Comment by fritsch** [comment]
@FernetMenta your concerns adressed?

File: xbmc/AutoSwitch.cpp (hunk 7 above)
Comment (Paxxi) [inline xbmc/AutoSwitch.cpp]: Should be 0.0f and not 0,will be an implicit conversion

**Comment by Paxxi** [comment]
I'm apparently a bit late to the party so feel free to ignore my comments

File: xbmc/input/KeyboardLayout.cpp [hunk 9]
```diff
@@ -69,7 +69,7 @@ bool CKeyboardLayout::Load(const TiXmlElement* element)
 
   const TiXmlElement *keyboard = element->FirstChildElement("keyboard");
   if (element->Attribute("codingtable"))
-    m_codingtable = CInputCodingTableFactory::CreateCodingTable(element->Attribute("codingtable"), element);
+    m_codingtable = IInputCodingTablePtr(CInputCodingTableFactory::CreateCodingTable(element->Attribute("codingtable"), element));
```
Paxxi: Should use `std::make_shared`

neo1973:
No problem @Paxxi, I did the changes you suggested.

xbmc/input/KeyboardLayout.cpp
neo1973: `std::make_shared` won't work here because this isn't a call to a constructor but a factory.

Paxxi: Oh right, my bad reading diffs on the phone

FernetMenta:
> FernetMenta your concerns adressed?

looks ok

MartijnKaijser:
jenkins build and merge

PR #10064, opened 2016-07-03
Added navigation controls for EDL (commercials) in PVR playback

thardie:
Added functionality so that when playing from a PVR if there is an EDL, NEXT will go to the end of the next commercial block, and PREV will go to the beginning of the previous commercial block. If no EDL is present, old behavior is preserved.

### DIFF
--- a/xbmc/cores/VideoPlayer/Edl.cpp
+++ b/xbmc/cores/VideoPlayer/Edl.cpp
@@ -840,6 +840,51 @@ bool CEdl::InCut(const int iSeek, Cut *pCut) const
   return false;
 }
 
+bool CEdl::GetNearestCut(bool bPlus, const int iSeek, Cut *pCut) const
+{
+  if (bPlus) 
+  {
+    // Searching forwards
+    for (auto &cut : m_vecCuts)
+    {
+      if (iSeek >= cut.start && iSeek <= cut.end) // Inside cut.
+      {
+        if (pCut)
+          *pCut = cut;
+        return true;
+      }
+      else if (iSeek < cut.start) // before this cut
+      {
+        if (pCut)
+          *pCut = cut;
+        return true;
+      }
+    }
+    return false;
+  } 
+  else
+  {
+    // Searching backwards
+    for (int i = (int)m_vecCuts.size() - 1; i >= 0; i--)
+    {
+      if (iSeek - 20000 >= m_vecCuts[i].start && iSeek <= m_vecCuts[i].end) 
+        // Inside cut. We ignore if we're closer to 20 seconds inside
+      {
+        if (pCut)
+          *pCut = m_vecCuts[i];
+        return true;
+      }
+      else if (iSeek > m_vecCuts[i].end) // after this cut
+      {
+        if (pCut)
+          *pCut = m_vecCuts[i];
+        return true;
+      }
+    }
+    return false;
+  }
+}
+
 bool CEdl::GetNextSceneMarker(bool bPlus, const int iClock, int *iSceneMarker) const
 {
   if (!HasSceneMarker())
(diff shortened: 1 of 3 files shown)


FernetMenta:
why did you limit it for pvr only?

thardie:
@akva2 The second commit fixes the indenting and braces - Did I miss something
@FernetMenta I haven't seen EDLs for any other kind of videos, so can't test it, and didn't want to break old behavior. Is it safe to remove the PVR restriction?

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -4355,6 +4355,21 @@ bool CVideoPlayer::OnAction(const CAction &action)
       case ACTION_NEXT_ITEM:
       case ACTION_CHANNEL_UP:
       {
+        if (m_pInputStream->IsStreamType(DVDSTREAM_TYPE_PVRMANAGER))
```
FernetMenta: In v18 there won't be a streamtype PVR anymore. I would just drop this limitation.

thardie:
@FernetMenta Removed the restriction as requested.

xbmc/cores/VideoPlayer/Edl.cpp
```diff
@@ -840,6 +840,50 @@ bool CEdl::InCut(const int iSeek, Cut *pCut) const
   return false;
 }
 
+bool CEdl::GetNearestCut(bool bPlus, const int iSeek, Cut *pCut) const
+{
+  if (bPlus) {
```
FernetMenta: curly braces next line please

xbmc/cores/VideoPlayer/Edl.cpp
```diff
@@ -840,6 +840,50 @@ bool CEdl::InCut(const int iSeek, Cut *pCut) const
...
+    // Searching forwards
+    for (int i = 0; i < (int)m_vecCuts.size(); i++)
```
FernetMenta: we moved to c++ 11 style loops: for (auto &cut : m_vecCuts)

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -4355,6 +4355,18 @@ bool CVideoPlayer::OnAction(const CAction &action)
       case ACTION_NEXT_ITEM:
       case ACTION_CHANNEL_UP:
       {
+        if (m_Edl.HasCut()) 
+        {
+          // For PVR streams, we'll do EDL searching, since they can't have next/prev
```
FernetMenta: I would drop reference to PVR in this comment

FernetMenta:
thanks, apart from the minors (see comments) I am ok with this.

xbmc/cores/VideoPlayer/Edl.cpp
thardie: There's another loop below this where I go through the list backwards - Do you want me to re-implement that in C++ 11 style as well?

FernetMenta: hmm, there is no revers range based loop, right? then might be better keeping it as is.

FernetMenta:
@thardie could you please squash the 4 commits into a single one. then we are good to go

xbmc/cores/VideoPlayer/Edl.cpp
Montellese: But you could use iterators instead of indices. They exist for both directions (`begin()` and `rbegin()` etc.)

thardie:
@FernetMenta Done

FernetMenta:
there is a "merge remote tracking" commit in this branch. always rebase on upstream, never merge upstream

thardie:
Oops, sorry. Will fix it this evening (approx 8 hours from now)

FernetMenta:
thanks!

thardie:
@FernetMenta Should be good now.

FernetMenta:
jenkins build this please

PR #10066, opened 2016-07-03
[PeripheralCecAdapter] Fix: Slovak language code (slo) considered as Slovenian (slv)

vzahradnik:
Hello,
current implementation of PeripheralCecAdapter seems to have a bug in CPeripheralCecAdapter::SetMenuLanguage implementation. As a result, when I set menu language in my TV to Slovak, it is reported via HDMI CEC to device and Kodi user interface is actually set to Slovenian.

According to [HDMI specification](http://www.microprocessor.org/HDMISpecification13a.pdf), ISO 639-2 shall be used. For language code table please [look here](http://www.loc.gov/standards/iso639-2/php/code_list.php).

Slovak language is using ISO code **slo** and Slovenian is using ISO code **slv**. I have updated the code for both Slovak and Slovenian.

I have not yet tested this change locally, but I might do that if required. I think that because the change is actually simple, if the code builds, we should be OK.

### DIFF
--- a/xbmc/peripherals/devices/PeripheralCecAdapter.cpp
+++ b/xbmc/peripherals/devices/PeripheralCecAdapter.cpp
@@ -594,6 +594,8 @@ void CPeripheralCecAdapter::SetMenuLanguage(const char *strLanguage)
   else if (!strcmp(strLanguage, "srp"))
     strGuiLanguage = "sr_rs@latin";
   else if (!strcmp(strLanguage, "slo"))
+    strGuiLanguage = "sk_sk";
+  else if (!strcmp(strLanguage, "slv"))
     strGuiLanguage = "sl_si";
   else if (!strcmp(strLanguage, "spa"))
     strGuiLanguage = "es_es";


popcornmix:
ping @opdenkamp

stefansaraev:
@vladimir-zahradnik change looks ok to me (no cec here, so cant runtime test), but you obviously should runtime test it

vladimir-zahradnik:
@popcornmix : OK, I'll try to test the change. I will update this thread once I test my changes.

vladimir-zahradnik:
@popcornmix Could you advise how can I ping @opdenkamp? Thanks.

popcornmix:
You need to do nothing. The mention should mean he gets alerted through github notifications. He then has the option to comment here.

opdenkamp:
+1

vladimir-zahradnik:
I just tested my changes on RPi with OpenELEC (custom build). Slovak language is now set correctly based on TV language settings. I didn't have a chance to test Slovenian, because this language is not present in my TV.

Here is except from debug log:
DEBUG: SetMenuLanguage - language set to 'resource.language.sk_sk'

popcornmix:
jenkins build and merge

PR #10067, opened 2016-07-03
[AE/OSXSink] - bundle audio device list changes into one final CAEFac…

Memphiz:
…tory::DeviceChange() call by using a timer. Should prevent multiple audio dropouts during hdmi enumeration and stuff like that (which results in multiple events fired by the os but only needs to inform AudioEngine once after all os events are fired).

This is in response to trac ticket http://trac.kodi.tv/ticket/16785

@fritsch something like that?

### DIFF
--- a/xbmc/cores/AudioEngine/Sinks/AESinkDARWINOSX.cpp
+++ b/xbmc/cores/AudioEngine/Sinks/AESinkDARWINOSX.cpp
@@ -93,6 +93,48 @@ static CADeviceList GetDevices()
   return list;
 }
 
+//---------- DEVICE CHANGED EVENT HANDLING
+// when device list change callbacks are received
+// inform the ae engine this amount of time after
+// the last callback was received (bundle multiple callbacks)
+// into one final call
+#define DEVICE_LIST_CHANGED_TIMER_MS 1500
+
+CDeviceChangeListTimeout::CDeviceChangeListTimeout()
+{
+  m_pDeviceListChangedTimer = new CTimer(this);
+}
+
+CDeviceChangeListTimeout::~CDeviceChangeListTimeout()
+{
+  delete m_pDeviceListChangedTimer;
+}
+
+// ITimerCallback
+// inform AE that the device list changed
+void CDeviceChangeListTimeout::OnTimeout()
+{
+  CLog::Log(LOGDEBUG, "CoreAudio: audiodevicelist delayed timer fired");
+  CLog::Log(LOGDEBUG, "CoreAudio: audiodevicelist changed - reenumerating");
+  CAEFactory::DeviceChange();
+  CLog::Log(LOGDEBUG, "CoreAudio: audiodevicelist changed - done");
+}
+
+// reset the timer to delay the timeout a bit more  
+void CDeviceChangeListTimeout::ResetDeviceListChangedTimer()
+{
+  if (m_pDeviceListChangedTimer->IsRunning())
+  {
+    m_pDeviceListChangedTimer->Restart();
+    CLog::Log(LOGDEBUG, "CoreAudio: audiodevicelist delayed timer reset");
+  }
+  else
+  {
+    m_pDeviceListChangedTimer->Start(DEVICE_LIST_CHANGED_TIMER_MS);
+    CLog::Log(LOGDEBUG, "CoreAudio: audiodevicelist delayed timer armed");
+  }
+}
+
 OSStatus deviceChangedCB(AudioObjectID                       inObjectID,
                          UInt32                              inNumberAddresses,
                          const AudioObjectPropertyAddress    inAddresses[],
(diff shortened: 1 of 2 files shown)


xbmc/cores/AudioEngine/Sinks/AESinkDARWINOSX.cpp
```diff
@@ -93,13 +93,56 @@ static CADeviceList GetDevices()
...
+//---------- DEVICE CHANGED EVENT HANDLING
+// when device list change callbacks are received
+// inform the ae engine this amount of time after
+// the last callback was received (bundle multiple callbacks)
+// into one final call
+#define DEVICE_LIST_CHANGED_TIMER_MS 1500
+
+CDeviceChangeListTimeout::CDeviceChangeListTimeout()
+{
+  m_pDeviceListChangedTimer = new CTimer(this);
```
fritsch: Personall I had chosen something like that: https://github.com/xbmc/xbmc/blob/master/xbmc/cores/AudioEngine/Sinks/AESinkAUDIOTRACK.h#L70 on the sack and just checking if it is experied before calling the Factory Method? Basically a "two liner"?

fritsch: This: http://sprunge.us/SYEU or something like that?

Memphiz: That would fire on the first event but not on the last one ... consider the case where a device gets removed and readded during 2 secs ... in your case we would enumerate while the device is missing and would skip the second event which would signal that the device was added back ... AE would have a list of audio devices then with this very device missing ...

Memphiz: Basically your code suppresses callbacks for 1,5 secs after the first callback. My code delays the final callback until no events occured for 1,5seconds ...

Memphiz: Also that timer can't be a member of the sink class because the sink object can be deleted by AE due to the fact that the device that sink has currently open was removed (i had this approach checked and wondered why my timer didn't fire - it was because AE timedout while adding packets to the sink and destroyed the sink object - which also destroyed the timer which was supposed to fire and notify AE of the changed device list).

fritsch: The latter case would just construct a new sink, with a new member on stack, initialized to expired. Yes - you are right with the other arguments. Though, with a little tweaking it can be extended to also match this usecase (another member during AddPackets (m_deviceChanged = true after first event + Set(1500) for every consequent event), if m_deviceChanged and Timer elapsed -> Rescan.

If the sink is deleted while a timer is running, the timer should (!) go away, cause AE is reiniting itself already and will create new sinks anyways. The timer is not aware of this situation, it would just falsely trigger AE again, but AE just did a new enumeration by itself and is uptodate.

Whatever you prefer.

Memphiz: Yep you are right - that would be the use case where the audio device the sink is using gets removed (in my case i turn off the bluetooth speaker i play audio through for example). You are right that AE gets a timeout in addpackets, destroys the sink and reenumerates the devices. The timer is still active and fires the delayed device changed event. Well this is assuming, that the OS is already aware of the lost device and AE really gets the up to date list (where the bluetooth speaker is already removed).
As i said in the ticket ... at least some use cases to think off and some assumptions to make. Maybe i simply let it be as is. It works and not much to gain from this anyway i think.

Memphiz: One theoretical question though - in your last statement you said:

"if m_deviceChanged and Timer elapsed -> Rescan"

With your approach i don't have a place to do that check ... there is no thread running where i could put that if into - or did you suggest to check for rescan in addpackets of the sink?

fritsch: In AddPackets - not nice - but that's the only thing that does something (at least if the sink is active). If it is not -> it is suspended and unloaded anyways.

fritsch: On the other hand as the OSX api is more capable and most likely also can have some user ptr in its callback method. You can add this information dynamically with a reference to a local member (hopefully still existent, proper locking :-) ...

And check the timer / results in this callback thread and fire if needed.

fritsch: See the PulseAudio implementation. It only uses the lock, which takes care that the object is not destroyed while being callbacked, but it could maintain another local member.

Though it would still be AddPackets which asks if the timer is gone as your use case: "Do it 1.5 seconds later as the last callbacks" means: It can not be fired by such callback.

Memphiz:
doesn't make sense in the end ...

PR #8271, opened 2015-10-21
[addons] Preserve strings which are not string ids in enum lvalues

LS80:
Sometimes with addon setting spinners some of the values may be untranslatable, e.g. a proper noun or some kind of identifier. Currently it is impossible to mix string ids to be localized and actual strings because the strings are converted to the string "Programs". This is a result of a lookup of the string id 0 as returned by atoi when it can't convert to an integer. 

This PR fixes that so that the fallback is to preserve the original string.

The original thread with some more detail is at http://forum.kodi.tv/showthread.php?tid=244607 in case that's better for discussion.

PR #10068, opened 2016-07-03
[addons] Preserve strings which are not string ids in enum lvalues

LS80:
Fix for https://github.com/xbmc/xbmc/pull/8271.

> Sometimes with addon setting spinners some of the values may be untranslatable, e.g. a proper noun or some kind of identifier. Currently it is impossible to mix string ids to be localized and actual strings because the strings are converted to the string "Programs". This is a result of a lookup of the string id 0 as returned by atoi when it can't convert to an integer.
> 
> This PR fixes that so that the fallback is to preserve the original string.


tamland:
thanks. lets try again then:)
jenkins build this please

PR #10070, opened 2016-07-04
Various micro-optimizations in the URL code

MaxKellermann:


fritsch:
jenkins build this please

MrMC:
you could make the in-lines all on one line to match code convention used in other classes. Otherwise, the code convention is "{" and "}" on separate lines :)

xbmc/music/MusicDatabase.cpp
```diff
@@ -4043,9 +4033,9 @@ bool CMusicDatabase::GetAlbumsByWhere(const std::string &baseDir, const Filter &
     // get data from returned rows
     items.Reserve(results.size());
     const dbiplus::query_data &data = m_pDS->get_result_set().records;
-    for (DatabaseResults::const_iterator it = results.begin(); it != results.end(); ++it)
+    for (const auto &i : results)
     {
-      unsigned int targetRow = (unsigned int)it->at(FieldRow).asInteger();
+      unsigned int targetRow = (unsigned int)i.at(FieldRow).asInteger();
```
Paxxi: Would you mind also updating lines you've touched to use c++ style casts? Can be painful with several commits so this is me asking nicely, not demanding :)

Paxxi:
Nice work @MaxKellerman, always nice getting some modernization done. The inlining of methods will probably be a noop, compiler should figure this out better than us. It doesn't hurt either, but if it's to be merged it needs to follow coding style as @davilla mentioned

MaxKellermann:
On 2016/07/04 21:46, Pär Björklund <notifications@github.com> wrote:
> Nice work @MaxKellerman, always nice getting some modernization
> done. The inlining of methods will probably be a noop, compiler
> should figure this out better than us.

No, that's impossible.  The compiler can't optimize across compilation
units.

(That is, unless you use LTO, which is a workaround for the
"compilation unit" thing.)

MaxKellermann:
Second round:
- fixed the `TestBasicEnvironment` build failure, which was not caused, only triggered by my include cleanups;
- dropped the `std::to_string()` patch (for now), because `std::to_string()` is broken on the old NDK your Jenkins server is using

MaxKellermann:
Third round: another attempt for the `TestBasicEnvironment` build failure, which turns out to be a gcc 4.8 bug (I don't have such an old buggy version, but your build server does). Workaround: add an include dependency which should not be necessary ...

MartijnKaijser:
jenkins build this please

MartijnKaijser:
Any objections?

MartijnKaijser:
@MaxKellermann here's a run with test enabled http://jenkins.kodi.tv/job/TestMulti-All/668/

Paxxi:
👍

PR #10071, opened 2016-07-04
CRenderManager: fix GetAspectRatio at playback start

wiromare:
at playbackstart in CGUIInfoManager::UpdateAVInfo() the variables m_videoInfo and m_audioinfo are filled with the info from the player but when these info are requested the rendermanager is not already configured or at least it's not configured on Windows

so when CBaseRenderer::GetAspectRatio()  is called m_sourceWidth and m_sourceHeight are not set with the right value but with 720x480

without this PR i have always the tag with the aspect ratio at 1.33:1

I don't know if this is a proper solution

@FernetMenta @afedchin

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp
@@ -324,6 +324,7 @@ bool CRenderManager::Configure()
     m_renderState = STATE_UNCONFIGURED;
 
   m_stateEvent.Set();
+  m_playerPort->VideoParamsChange();
   return result;
 }
 


FernetMenta:
are you saying that AR is gets not set at all or just for a small period of time after start?

wiromare:
after that rendermanager it's configured GetAspectRatio will return the proper value but it's too late because at this point in CGUIInfoManager::UpdateAVInfo()
 if (CServiceBroker::GetDataCacheCore().HasAVInfoChanges()) will return always false so m_videoInfo and m_audioInfo (needed for the skin tags i suppose) cannot be updated, and the gui show the wrong AR tag for ever

so why i added m_playerPort->VideoParamsChange() after the configure in rendermanager to make possible update m_videoInfo and m_audioInfo

FernetMenta:
That does not really answer my question. Background:

- CRenderManager::Configure sets m_bTriggerUpdateResolution
- then CRenderManager::UpdateResolution is called
- CRenderManager::UpdateResolution() calls m_playerPort->VideoParamsChange()

Moving m_playerPort->VideoParamsChange should not change anything. maybe an additional update some milliseconds earlier.

What goes different on your system?

wiromare:
ok i missed the trigger m_bTriggerUpdateResolution, now i know where is the problem

the problem come out only when the kodi is windowed because this

if (g_graphicsContext.IsFullScreenVideo() && g_graphicsContext.IsFullScreenRoot()) in CRenderManager::UpdateResolution

(I was debugging other things when i noticed the tag issue)

FernetMenta:
> the problem come out only when the kodi is windowed because this

ahh, that is important info :)

this change is fine, thanks!

jenkins build this please

PR #10072, opened 2016-07-05
UPnP startup honors user settings

Rallymen007:
During some research I noticed Kodi would not respect the user's UPnP setting on initial launch.
http://forum.kodi.tv/showthread.php?tid=280683

This simple fix adds a routine setting check before initializing each UPnP subset on startup.


MartijnKaijser:
@montellese ping

Montellese:
Our coding style is two spaces indentation and curly braces on a separate line. Apart from that the fix is correct. Thanks for bringing it up.

xbmc/network/NetworkServices.cpp
```diff
@@ -826,10 +826,16 @@ bool CNetworkServices::StartUPnP()
 {
   bool ret = false;
 #ifdef HAS_UPNP
-  ret |= StartUPnPClient();
-  ret |= StartUPnPServer();
-  ret |= StartUPnPController();
-  ret |= StartUPnPRenderer();
+  if (CSettings::GetInstance().GetBool(CSettings::SETTING_SERVICES_UPNPSERVER)) {
+	  ret |= StartUPnPClient();
```
Montellese: The client should always be started because it is required for browsing UPnP sources independent of whether the user wants to share its library or not.

Rallymen007:
The rebase was half a good idea as it turns out, but I fixed the indentation (my mistake, VS added tabs and I wasn't paying attention) and moved the UPnP client initialization.

In my use case I wanted to completely turn off the UPnP portion as it seemed to interfere with my custom UPnP client, hence why StartClient was there. I ended up just recompiling Kodi with -disable_upnp and it was all good.

MartijnKaijser:
@Rallymen007 sadly the rebase went wrong. Best it to always make a branch of our master and name it different. A quick guide can be found here
https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md

Could you create a new pull rquest please in which you do all changes in a single commit?

PR #10073, opened 2016-07-06
[xbmc][fix] Fix theme xbt loading that broke during cleanup.in #9960

Paxxi:
Found a race condition that seems to have been hidden and a screwup during cleanup clobbering state.

The race condition is that g_graphicsContext.GetMediaDir() isn't set to anything when first called, it's only later on during skin loading it's actually populated. If this happens now we fall back to the currently set skin.

@MilhouseVH I've tested with the Titan beta skin, I can't see any difference on the corners when using modern_rounded but works fine setting it to classic so I'm thinking it might be my eyes deceiving me or some change in the beta I tested. Would be great if you or the author can test the changes.

### DIFF
--- a/xbmc/guilib/TextureBundle.cpp
+++ b/xbmc/guilib/TextureBundle.cpp
@@ -21,7 +21,14 @@
 #include "TextureBundle.h"
 
 CTextureBundle::CTextureBundle()
-  : m_useXBT{false}
+  : m_tbXBT{false}
+	, m_useXBT{false}
+{
+}
+
+CTextureBundle::CTextureBundle(bool useXBT)
+  : m_tbXBT{useXBT}
+	, m_useXBT{useXBT}
 {
 }
 
--- a/xbmc/guilib/TextureBundleXBT.cpp
+++ b/xbmc/guilib/TextureBundleXBT.cpp
@@ -41,9 +41,15 @@
 #endif
 #endif
 
-CTextureBundleXBT::CTextureBundleXBT(void)
+CTextureBundleXBT::CTextureBundleXBT()
   : m_TimeStamp{0}
   , m_themeBundle{false}
+{
+}
+
+CTextureBundleXBT::CTextureBundleXBT(bool themeBundle)
+  : m_TimeStamp{0}
+  , m_themeBundle{themeBundle}
 {
 }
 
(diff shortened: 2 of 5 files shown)


MilhouseVH:
Thanks, I'll include it in tonight's build (though a bit worried you can't see the difference rounded corners makes, it's fairly clear in the screenshots posted by the user).

Paxxi:
Yeah I saw it in the posted screenshots so that's why I'm doubting my own vision here. The classic theme option has rounded corners and switching between that and default it clearly changes the corners rounding so it could be something wrong with just the modern_rounded theme in the skin currently.

MilhouseVH:
I'm pretty sure there's still an issue, though as you say it might be skin related (I've really no idea).

This is how Titan "modern_rounded" looked before PR9960 (ie. build #0613):
s1

And now current kodi master (ie. with PR9960) and this PR:
s3

The strange thing is that build #0704x, which is reportedly working correctly (ie. with rounded corners), looks the same as current kodi master to me (ie. second screenshot) so I'm not convinced that reverting PR9960 is actually working as reported - I'll ask the user to test #0704x again.

Then again, if you suspect this might be some race condition, it could be that it works some of the time...

Paxxi:
maybe we can get @marcelveldt to comment, if I understand it correctly he's the maintainer of the skin.

Paxxi:
oh nvm, had a look at it again and there was one more thing I failed to fix, have updated the PR and now it should be good to go for tonight @MilhouseVH.
screenshot 10

xbmc/guilib/TextureBundleXBT.cpp
```diff
@@ -41,12 +41,18 @@
...
 
-CTextureBundleXBT::CTextureBundleXBT(void)
+CTextureBundleXBT::CTextureBundleXBT()
   : m_TimeStamp{0}
   , m_themeBundle{false}
 {
 }
 
+CTextureBundleXBT::CTextureBundleXBT(bool themeBundle)
+	: m_TimeStamp{0}
```
FernetMenta: is this a tab?

xbmc/guilib/TextureBundle.cpp
```diff
@@ -21,7 +21,14 @@
 #include "TextureBundle.h"
 
 CTextureBundle::CTextureBundle()
-  : m_useXBT{false}
+  : m_tbXBT{false}
+	, m_useXBT{false}
```
FernetMenta: tab

MilhouseVH:
Looks good now @paxxi - many thanks!

marcelveldt:
Great work, thanks for the quick fix. This issue was indeed reported on the forums

xbmc/guilib/TextureBundleXBT.cpp
Paxxi: Looks like it, no idea how that got in there, will fix before merging

Paxxi:

PR #8880, opened 2016-01-15
[json-rpc] add/delete/toggle timer functionality

stevegal:
updates #8816 which I can't seem to reopen

PR #10074, opened 2016-07-06
[PVR][json-rpc] change/fix timer API, sync pvr types fields with current implementation

ksooo:
This PR superseeds #10058. It changes/fixes the current json-rpc pvr timer methods.

Furthermore it synchronizes the json-rpc fields for the various pvr types (timers, epg tags, channels, recordings) with it's current implementation.

@Tolriq @joethefox : could you test this, please?


ksooo:
@stevegal sorry, I somehow missed your PR #8880  which added the timer methods to json-rpc API. I changed the original implementation slightly (and incomapatible!) now to be of more value. Basically I removed param "repeating" fro DeleteTimer as this is imo not needed.

joethefox:
sure, I will do \m/ But consider that at the moment nothing is implemented on the remote app side to manipulate the timers, they are only listed. It will take some time to test everything, but I will do.

Tolriq:
@ksooo not at home until next week so unable to test before that.

The PR that update the version.txt file must have [breaking change] in the title. See previous discussions and validations to skip the discussion on that part :)

And from previous PR question (That was in an edit) it seems there's a concept of readonly timers.
Is this still true ? If yes the field should be serialised / exposed too so that remote does not offer an option to remove them as it will certainly does not work.

And a quick question : What are the new channelnumber / subchannelnumber ?

ksooo:
> It will take some time to test everything, but I will do.

@joethefox sure. many thanks.

> The PR that update the version.txt file must have [breaking change] in the title. See previous discussions and validations to skip the discussion on that part :)

 Thought we have pr label "JSON-RPC API Change" for this, but anyway. Thanks for the heads-up. Changed.

> it seems there's a concept of readonly timers. Is this still true ? If yes the field should be serialised / exposed too

 Good point. will ad this.

>  What are the new channelnumber / subchannelnumber 

 No idea. This was not added by me. github file history will help you to find out, i guess. I just exposed to json-rpc what was already there in c++.

Tolriq:
@ksooo Not the PR title, the commit title for https://github.com/xbmc/xbmc/pull/10074/commits/19349e1f3ab6d61e11ec6f46a26281b25dd478ba

The concept is to be able with a git log on version.txt to find out breaking change during alphas since it was decided to not bump the major version that should indicate there's a breaking change.

We can't rely on PR title / github labels for that.

ksooo:
'readonly' field for a timer is now implemented.

ksooo:
@Tolriq 

>> The PR that update the version.txt file must have [breaking change] in the title.

> Not the PR title, the commit title

lol, will change that.

xbmc/interfaces/json-rpc/schema/methods.json
```diff
@@ -2042,34 +2042,33 @@
...
-      { "name": "timerid", "$ref": "Library.Id", "required": true, "description": "the id of the timer to delete" },
-      { "name": "repeating", "type": "boolean", "default": false, "description": "the delete should be done for all occurances" }
+      { "name": "timerid", "$ref": "Library.Id", "required": true, "description": "the id of the onetime timer or timer rule to delete" }
     ],
     "returns":  "string"
   },
   "PVR.ToggleTimer": {
     "type": "method",
-    "description": "Toggles a timer",
+    "description": "Creates or deletes a onetime timer or timer rule for a given show. If it exists, it will be deletedif not, it will be created",
```
Montellese: missing space at `deletedif`.

Tolriq:
@ksooo yeah sorry not at home on a small screen laptop :) (No relation but maybe it can pass as an excuse :p)

stevegal:
@ksooo thanks for the heads up on this change. Looks good to me! I've been using this for a while now and I've been wondering if we should return something that points to the timer ID on success for add. What do you think?

xbmc/interfaces/json-rpc/schema/methods.json
Jalle19: Missing a comma too, should be `If it exists, it will be deleted, if not, it will be created`

ksooo:
@Montellese typo fixed. thx for the review.
@Jalle19 damn. my force push killed your comments. What was it, again?
@stevegal returning timer id on add is not possible as the timer will be added asynchronously to the backed. Kodi gets an async ACK from the backend and then and only then the timer id will be created by Kodi.

ksooo:
@Jalle19 found the comment. ;-) Fixed. Thx for the review.

Tolriq:
@ksooo / @MartijnKaijser really hard to find time to test before my holidays :( will there be alpha 4 or is 20th July dead line for this to be merged ?

Tolriq:
Since no answers :( I tried to quickly test and all that I can test with demo PVR addon does works.
But I can't test add / delete / toggle as the demo addon does not support that.

So would be cool to have this merged if no other merge window, if there's fix needed then beta is here for that. That would prevent breaking change for no purpose in V18.

@ksooo I see that toggleTimer is based on broadcastid but PVR.Details.Timer does not return that id, do you think it could make sense to return it too as it could simplify timer displaying in some cases ?

ksooo:
jenkins Build this please

ksooo:
Okay, will merge now. Regression risk is nearly zero. If something is still missing we can add this later.

joethefox:
Tested all the changes and I can confirm that I haven't noticed any regressions so far.

The issue I've found is when calling PVR.SetTimer method with "timerrule" set to TRUE. 
The resulting timer is created but seems invalid: "Search guide for" is empty (default is the same as the "Name" field), the "Days of week" are empty, "Start any time" and "End any time" are not enabled by default with the result that no one broadcast is actually programmed to be recorded.

ksooo:
@joethefox thanks for the feedback. I will look into the issue you found.

ksooo:
> The issue I've found is when calling PVR.SetTimer method with "timerrule" set to TRUE.
The resulting timer is created but seems invalid: "Search guide for" is empty (default is the same as the "Name" field), the "Days of week" are empty, "Start any time" and "End any time" are not enabled by default with the result that no one broadcast is actually programmed to be recorded.

@joethefox what's the correct curl command to reproduce this?

joethefox:
`curl --user <username>:<password> --header 'Content-Type: application/json' --data-binary '{ "id": 1, "jsonrpc": "2.0", "method": "PVR.AddTimer", "params" : {"broadcastid":<broadcastid>, "timerrule":true} }' 'http://<kodi_ip>:<kodi_port>/jsonrpc'`

PR #10076, opened 2016-07-06
VideoPlayer: check for duplicate filenames when opening multi source …

FernetMenta:
…demuxer

fixes http://forum.kodi.tv/showthread.php?tid=281725

*wtv files can be audio and video. Checking for external audio ended up having the same file twice in the list.

@ace20022 /cc


FernetMenta:
jenkins build this please

FernetMenta:
confirmed fixed by user (forum thread)

PR #10077, opened 2016-07-06
player processinfo: set higher precision for FPS and DAR

FernetMenta:


gade01:
Thanks for adding this!

Works perfectly:

FernetMenta:
thanks for testing!

doctorsin:
@FernetMenta can I doubt kodi on my iPad, if so, how?

PR #10017, opened 2016-06-20
[android/packaging] Sync rtmp removal

fetzerch:
RTMP got removed in https://github.com/xbmc/xbmc/pull/9946.

This somehow broke the packaging with CMake even though it uses the same file as the autotools build.
CMake build of current master: http://jenkins.kodi.tv/job/Android-ARM/8496/
CMake build with this change: http://jenkins.kodi.tv/job/Android-ARM/8503/

PR #10079, opened 2016-07-07
Revert "[android/packaging] Sync rtmp removal"

fetzerch:
Reverts xbmc/xbmc#10017

The library seems to be still used (see comments in xbmc/xbmc#10017)


stefansaraev:
could you please re-test with #10086

fetzerch:
Ok, then let's merge this revert. I'll dig into the problem we had with building the packages with cmake.

PR #10080, opened 2016-07-07
guilib: do not kill plugin script if we are waiting for user input

FernetMenta:
see title

### DIFF
--- a/xbmc/filesystem/PluginDirectory.cpp
+++ b/xbmc/filesystem/PluginDirectory.cpp
@@ -494,9 +494,23 @@ bool CPluginDirectory::WaitOnScriptResult(const std::string &scriptPath, int scr
   else
   {
     // kill the script if it does not return within 30 seconds
-    if (!m_fetchComplete.WaitMSec(30000))
+    // and we are not waiting for user input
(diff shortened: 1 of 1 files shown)


Tolriq:
@FernetMenta do you think this is possible to reset the timer when the dialog is shown ?

Reading code, if the dialog is shown for 29sec this only leave 1 sec to get the data.

Typical use case still Youtube browsing from a remote get to search screen, enter the text. 30 sec timeout should start at that moment and not take in account the X seconds he took to enter the data.

FernetMenta:
@Tolriq good input. I updated it, please review.

Tolriq:
Well not at home to test and not fluent with Kodi threading / timers :)
I would have gone an easier way with your previous implementation, adding a simple boolean test in the while and updating the boolean in the loop if dialog was shown. 
Would not have generated a pure 30 sec wait after a dialog but a 30sec + 30sec - dialog duration. But was simple.

Anyway if that code works and address the issue I'm fine with it :)

FernetMenta:
There was a case not considered by the old code that is now fixed: script terminates without calling DirectoryComplete.

FernetMenta:
jenkins build this please

PR #10081, opened 2016-07-07
[PVR] Extract thumbnails for recordings, if not provided by pvr backend.

ksooo:
If a pvr backend does not provide thumbnails for recordings, the thumbnails now will be automatically generated by Kodi.

screenshot001
screenshot002

This feature has been requested several times in the forum.

Surprisingly, it was more than trivial to implement. ;-)

@Jalle19 for review?

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDFileInfo.cpp
+++ b/xbmc/cores/VideoPlayer/DVDFileInfo.cpp
@@ -103,8 +103,7 @@ bool CDVDFileInfo::ExtractThumb(const std::string &strPath,
   unsigned int nTime = XbmcThreads::SystemClockMillis();
   CFileItem item(strPath, false);
 
-  if (item.IsDiscImage() ||
-      item.IsPVR())
+  if (item.IsDiscImage() || URIUtils::IsLiveTV(strPath))
     return false;
 
   item.SetMimeTypeForInternetFile();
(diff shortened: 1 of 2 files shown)


FernetMenta:
I have intentionally disabled this because it results in segfault. The PVR api does not support multiple streaming connections that are required for this feature.
It will be possible once streaming functions are moved to inputstream ext point.

Jalle19:
@ksooo is the channel icon still visible in the recording list?

ksooo:
@Fernetmenta: Fair enough. Thanks for pointing this out.

ksooo:
@Jalle19 as long as the thumb is not extracted (this may take some time, because it's a low prio job) the channel icon is displayed, yes. If a thumb is available, no more channel icon. What is the behavior you expect?

Jalle19:
@ksooo personally I'd like the small icon next to the item name to have the channel logo (since the icon is so small you can't make much sense of a thumbnail anyway, especially on some skins) and the large image near the description to have the thumbnail.

ksooo:
@Jalle19 are you talking about Estuary or Confluence?

ksooo:
Or do you mean channel logo in recordings list )(screenshot 1) and thumb in recording details (screenshot 2)? This would have to be changed in the skin, then.

Jalle19:
@ksooo Confluence

FernetMenta:
@ksooo a scenario that won't work is .i.e. starting playback of a pvr recording while thumb extraction is active. Then two clients would access the same API functions.

ksooo:
@FernetMenta yeah. understood.
@Jalle19 Confluence would need to get adapted (no big deal, though). Currently it would look like this:
screenshot003

PR #10082, opened 2016-07-07
[VideoSyncD3D] CVideoSyncD3D should implement IDispResource interface…

afedchin:
… instead of ID3DResource.

ping @FernetMenta


afedchin:
jenkins build this please

FernetMenta:
this fixes the issue with not updated clock after change of refresh rate, right?
+1

afedchin:
yes, you are right this should fix the issue with not updated clock.

PR #10083, opened 2016-07-07
[D3D11] Hack to fix refresh rate issue in full screen exclusive mode.

afedchin:
There is an ugly "bug" in DX11 which causes DX system is unable switch to desired display mode in full screen exclusive mode

23.976 <-> 24.00
29.970 <-> 30.00
47.952 <-> 48.00
59.940 <-> 60.00

On some system DX is unable to switch to 24.00 on others to 23.976 mode. Some systems has no the issue. This trick fixes the issue.

This also requires:
- WDK 8.1 installed on build machine https://www.microsoft.com/en-us/download/details.aspx?id=42273. (need to edit wiki page as well)
- EasyHook package on our mirrors https://www.dropbox.com/s/y1z8kiqx4s5wll2/easyhook-2.7.5870.0-win32-vc140.7z?dl=0

ping @Voyager1 @Paxxi @FernetMenta

### DIFF
--- a/xbmc/rendering/dx/RenderSystemDX.cpp
+++ b/xbmc/rendering/dx/RenderSystemDX.cpp
@@ -34,13 +34,32 @@
 #include "utils/MathUtils.h"
 #include "utils/log.h"
 #include "win32/dxerr.h"
+#include "utils/SystemInfo.h"
+#pragma warning(disable: 4091)
+#include "d3d10umddi.h"
+#pragma warning(default: 4091)
+#include <algorithm>
 
+#ifndef _M_X64
+#pragma comment(lib, "EasyHook32.lib")
+#else
+#pragma comment(lib, "EasyHook64.lib")
+#endif
 #pragma comment(lib, "d3d11.lib")
 #pragma comment(lib, "dxgi.lib")
 #pragma comment(lib, "dxguid.lib")
 
 #define RATIONAL_TO_FLOAT(rational) ((rational.Denominator != 0) ? \
  static_cast<float>(rational.Numerator) / static_cast<float>(rational.Denominator) : 0.0f)
+
+// User Mode Driver hooks definitions
+void APIENTRY HookCreateResource(D3D10DDI_HDEVICE hDevice, const D3D10DDIARG_CREATERESOURCE* pResource, D3D10DDI_HRESOURCE hResource, D3D10DDI_HRTRESOURCE hRtResource);
+HRESULT APIENTRY HookCreateDevice(D3D10DDI_HADAPTER hAdapter, D3D10DDIARG_CREATEDEVICE* pCreateData);
+HRESULT APIENTRY HookOpenAdapter10_2(D3D10DDIARG_OPENADAPTER *pOpenData);
+static PFND3D10DDI_OPENADAPTER s_fnOpenAdapter10_2{ nullptr };
+static PFND3D10DDI_CREATEDEVICE s_fnCreateDeviceOrig{ nullptr };
+static PFND3D10DDI_CREATERESOURCE s_fnCreateResourceOrig{ nullptr };
+CRenderSystemDX* s_windowing{ nullptr };
 
 using namespace DirectX::PackedVector;
 
@@ -54,10 +73,12 @@ CRenderSystemDX::CRenderSystemDX() : CRenderSystemBase()
   ZeroMemory(&m_viewPort, sizeof(m_viewPort));
   ZeroMemory(&m_scissor, sizeof(CRect));
   ZeroMemory(&m_adapterDesc, sizeof(DXGI_ADAPTER_DESC));
+  s_windowing = this;
 }
 
 CRenderSystemDX::~CRenderSystemDX()
 {
+  s_windowing = nullptr;
 }
 
 bool CRenderSystemDX::InitRenderSystem()
@@ -98,8 +119,8 @@ void CRenderSystemDX::SetMonitor(HMONITOR monitor)
   IDXGIAdapter1*  pAdapter;
   for (unsigned i = 0; m_dxgiFactory->EnumAdapters1(i, &pAdapter) != DXGI_ERROR_NOT_FOUND; ++i)
   {
-    DXGI_ADAPTER_DESC1 adaperDesc;
-    pAdapter->GetDesc1(&adaperDesc);
+    DXGI_ADAPTER_DESC1 adapterDesc;
+    pAdapter->GetDesc1(&adapterDesc);
 
     IDXGIOutput* pOutput;
     for (unsigned j = 0; pAdapter->EnumOutputs(j, &pOutput) != DXGI_ERROR_NOT_FOUND; ++j)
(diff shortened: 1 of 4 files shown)


FernetMenta:
Is there still any advantage in running Kodi windowless an Windows?

afedchin:
on win10 there are no advantages I think, but on pre win10 systems FSE mode requires less resources. In my oppinion windowed mode is still a bit stuttery when compared to full screen.  also FSE exclusive supports 10bit output on all systems (we don't use 10b output, but this may change in future)

Jalle19:
@FernetMenta lots of people use Kodi in windowed mode, especially when looking at things on the computer while doing other things.

xbmc/rendering/dx/RenderSystemDX.cpp
```diff
@@ -1750,4 +1779,177 @@ void CRenderSystemDX::SetMaximumFrameLatency(uint8_t latency) const
...
+    // no entries in the registry
+    if (drivers.empty())
+      return;
+    // 3. we take only first three values (dx12 driver isn't needed if it exists ofc)
+    if (drivers.size() > 3)
+      drivers = std::vector<std::wstring>(drivers.begin(), drivers.begin() + 3);
+    // 4. and then reverse vector to start iterate with the best candidate for d3d11 driver
+    std::reverse(drivers.begin(), drivers.end());
+
+    for (auto it = drivers.begin(); it != drivers.end(); ++it)
```
Paxxi: instead of `std::reverse` above you can use `for (auto it = drivers.rbegin(); it != drivers.rend(); ++it)` to iterate in reverse order.

xbmc/rendering/dx/RenderSystemDX.cpp
```diff
@@ -1750,4 +1779,177 @@ void CRenderSystemDX::SetMaximumFrameLatency(uint8_t latency) const
...
+  if (hKey != 0)
+    RegCloseKey(hKey);
+}
+
+void CRenderSystemDX::FixRefreshRateIfNecessary(const D3D10DDIARG_CREATERESOURCE* pResource)
+{
+  if (pResource && pResource->pPrimaryDesc)
+  {
+    float refreshRate = RATIONAL_TO_FLOAT(pResource->pPrimaryDesc->ModeDesc.RefreshRate);
+    if (refreshRate > 10.0 && refreshRate < 300.0)
```
Paxxi: this will cast to a double, should be `10.0f` and `300.0f`

Voyager1:
I agree with @afedchin - windowed mode has some disadvantages (as you all know), but on a relatively low-powered APU (like my current A4-5000) used in a living-room HTPC setting it's slightly more fluent. Furthermore, I would suggest that we do one of three things:
1) implement this fix/hack, OR
2) document in the UI that full screen exclusive sometimes does not allow the fully matching refresh rate, OR
3) simply turn off the possibility to use full screen exclusive.
My preference is (1).

xbmc/rendering/dx/RenderSystemDX.cpp
afedchin: good observation. I'll update this.

Paxxi:
full screen windowed mode is my preference as it avoid the ugly switching that you get with full screen exclusive. Maybe not the one a hardcore movie-buff would choose but it's perfect for say having kodi up on a second screen(which my daughter use extensively) and ofc during debugging as it's so much faster when hitting a breakpoint.

Can't say I understand it all but codewise it looks fine so no objections to merging

AchimTuran:
@FernetMenta 
> Is there still any advantage in running Kodi windowless an Windows?

Sure, I use it daily for development :smile: or listening to music.

FernetMenta:
I was referring to true / faked fullscreen. On Linux and OSX for example there is only faked fullstreen, means Kodi is running in a window without any title bar or frame visible.

Jalle19:
Based on the level black arts involved in this hack I'd vote in favor of dropping exclusive fullscreen.

afedchin:
dropping FSE mode will cause a shit storm on windows forum

FernetMenta:
Hacky or not, this is fairly self contained. I support any decision the Windows devs take.

Sorien:
real fullscreen mode(Use fullscreen window = false) is almost useless at win10 and multi screen setup (huge flickering/crashing/no picture when system wakes up from suspend/hibernate sometimes even when you bring another app to top) and 50% more CPU usages 4-5% at idle compared to 1-2% when Use fullscreen window is is enabled

afedchin:
@Sorien why I have results opposite than yours, on win10? If you have an issue with fse mode this does not mean that fse mode is useless.

xbmc/rendering/dx/RenderSystemDX.cpp
```diff
@@ -57,8 +57,7 @@
 void APIENTRY HookCreateResource(D3D10DDI_HDEVICE hDevice, const D3D10DDIARG_CREATERESOURCE* pResource, D3D10DDI_HRESOURCE hResource, D3D10DDI_HRTRESOURCE hRtResource);
 HRESULT APIENTRY HookCreateDevice(D3D10DDI_HADAPTER hAdapter, D3D10DDIARG_CREATEDEVICE* pCreateData);
 HRESULT APIENTRY HookOpenAdapter10_2(D3D10DDIARG_OPENADAPTER *pOpenData);
-typedef HRESULT(APIENTRY *pfnOpenAdapter10_2)(_Inout_ D3D10DDIARG_OPENADAPTER *);
```
Voyager1: because of this compile error at line 1861 (cast to (pfnOpenAdapter10_2))

afedchin: thanks. fixed

afedchin:
hm.. appveyor is happy, jenkins build this please

Voyager1:
does jenkins have WDK8.1 installed?

afedchin:
I don't think so. It fails because has not found header file.

Jalle19:
@afedchin @FernetMenta okay, let's not drop it then :-)

Voyager1:
@afedchin - I would suggest the addition of another patch: https://github.com/Voyager1/xbmc/commit/5a5d35d5a1041e9d9644604670057029240d30ed - the reason being that Numerator is UINT and you assign it a float value. This is each time used for FindClosestMatchingMode, so if refreshrate to be matched is 23.976 or 59.940 the decimals are dropped. Not a big deal, because one can assume that the closest matching mode will return the one you're looking for anyway. Maybe this is another "cosmetic" fix.

afedchin:
jenkins build this please

afedchin:
@Voyager1 take a look please https://github.com/xbmc/xbmc/pull/10083/commits/8e5ab838967e08452a86cd6f4364d68d95f96e35

Voyager1:
> @Voyager1 take a look please 8e5ab83

nice!

xbmc/rendering/dx/RenderSystemDX.cpp
```diff
@@ -1915,6 +1914,13 @@ void CRenderSystemDX::FixRefreshRateIfNecessary(const D3D10DDIARG_CREATERESOURCE
   }
 }
 
+void CRenderSystemDX::GetRefreshRatio(uint32_t refresh, uint32_t * num, uint32_t * den)
```
Voyager1: minor comment: pass num and den by reference (uint32_t&).

afedchin: I think that it will not be apparent to the caller that the num and den will be changed in method. passing pointers is common practices for DX API. but if you insist I'll change.

afedchin:
jenkins build this please

afedchin:
if no objections I will merge this tonight
jenkins build this please

PR #10086, opened 2016-07-07
[depends] build curl without librtmp support

stefansaraev:
kodi now uses ffmpeg's own rtmp implementation, and optionaly inputstream.rtmp
rtmp support was picked automaticlay by libcurl configure if librtmp header/libs
are present in sysroot. no point linking our libcurl against librtmp now


wsnipex:
jenkins build this please

PR #9987, opened 2016-06-16
[cmake/cpack] Debian packages generation

hudokkow:
This is the work to generate debian packages with our shiny CMake/CPack system.

It currently generates deb packages following stable, unstable or nightly release conventions for name/version:
kodi-bin_16.1~git20160425.1001-final-0xenial_amd64.deb   / 2:16.1~git20160425.1001-final-0xenial
kodi-bin_17.0~git20160703.1014-alpha2-0xenial_amd64.deb  / 2:17.0~git20160703.1014-alpha2-0xenial
kodi-bin_17.0~git20160714.0200-b0970fe-0xenial_amd64.deb / 2:17.0~git20160714.0200-b0970fe-0xenial
##### Packages currently generated

kodi
kodi-addon-dev
kodi-audio-dev
kodi-bin
kodi-eventclients-common
kodi-eventclients-dev
kodi-eventclients-ps3
~~kodi-eventclients-j2me~~
kodi-eventclients-wiiremote
kodi-eventclients-xbmc-send
kodi-inputstream-dev
kodi-pvr-dev
kodi-screensaver-dev
kodi-tools-texturepacker
kodi-visualization-dev
~~xbmc~~ removed
### Design decisions
- ~~CPackConfig.cmake file contains variables and logic common to all generators (DEB/RPM/NSIS/etc.)~~
- CPackConfigDEB.cmake file contains variables and logic specific to DEB packaging system
- scripts/linux/Install.cmake handles install targets logic separated by components. It calls (includes) CPackConfigDEB.cmake for package generation. ~~In turn CPackConfigDEB.cmake is the only file calling CPackConfig.cmake~~
- Variables and logic are enclosed in if statements. Should provide an easy way to manipulate packages naming/version/etc. using command-line only.
- CMake/CPack >=3.6.0 is required for proper package naming generation because it introduced <code> CPACK_DEBIAN_component_name_FILE_NAME</code>. It will work with lower versions but resulting package names are meh.
### Todo
- [x] Add documentation to packages (copyright, license, etc.)
- [x] Generate debian changelog on-the-fly and add it to packages (ping @wsnipex for ideas)
- [x] ~~Generate xbmc -> kodi symlinks and add them to packages~~ symlinks removed. They mess cpack DEB generation and xbmc to kodi transition as lasted long enough.
[...]

PR #10087, opened 2016-07-08
Freebsd cmake

Fneufneu:
Some cmake fixes to make things works on FreeBSD.
I marked [cmake] commits that can affect other platforms.


hudokkow:
Looks good.
CMAKE_MAKE_PROGRAM change in https://github.com/xbmc/xbmc/pull/10087/commits/7aad68656c010849ccb65e8f084329edce31d5fb should be in https://github.com/xbmc/xbmc/pull/10087/commits/527cd070af8076fc254c8a12117bbd7742183248

Apart from that, only one other comment: so you're the **one** FreeBSD user we have, heheh.

If you have the time, can you test https://github.com/xbmc/xbmc/pull/9987 commenting https://github.com/xbmc/xbmc/pull/9987/commits/6f5e4a5ee3c34ec00d59312927e11344a50f40c9#diff-2cec9cf9f25dcb68763ba455c63ba01bR324 out?
I'm curious and haven't had the time to test FreeBSD using linux's Install.cmake

project/cmake/modules/FindSWIG.cmake
```diff
@@ -1,6 +1,9 @@
 if(NOT SWIG_EXECUTABLE)
   find_program(SWIG_EXECUTABLE NAMES swig2.0 swig PATH_SUFFIXES swig)
 endif()
+if(NOT SWIG_EXECUTABLE)
+  find_program(SWIG_EXECUTABLE NAMES swig3.0 swig PATH_SUFFIXES swig)
```
wsnipex: just adding swig3.0 to the NAMES list above should suffice

fetzerch:
Thanks much! LGTM (besides the small minor that wsnipex mentioned already).
If you want you can also mention it in https://github.com/xbmc/xbmc/blob/master/project/cmake/README.md (maybe that way we might even find a second bsd user ;))

Fneufneu:
thanks for the feedback. i will update the pr and look at #9987 next week,
but most of the time, linux cmake files are fine for FreeBSD.

Fneufneu:
PR updated and ready to be merged.
I really need to update the README one day :) and you are more than one user on FreeBSD !
Someone send me a patch to make kodi run on RPi 2 / FreeBSD :)

fetzerch:
Thanks, I have the readme updated together with a few more things. Will push that with my next PR.

project/cmake/modules/FindCpluff.cmake
```diff
@@ -22,7 +22,7 @@ if(NOT WIN32)
                                         --host=${ARCH}
                                         CFLAGS=${defines}
                                         LDFLAGS=${ldflags}
-                      BUILD_COMMAND make V=1)
+                      BUILD_COMMAND ${CMAKE_MAKE_PROGRAM} V=1)
```
fetzerch: @Fneufneu: I just figured out that this breaks the build if one uses the ninja generator. [CMAKE_MAKE_PROGRAM](https://cmake.org/cmake/help/latest/variable/CMAKE_MAKE_PROGRAM.html) is set to a generator specific tool. But here and also for ffmpeg, we need to use "make". I'll investigate in a fix.

fetzerch: Fixed in https://github.com/xbmc/xbmc/pull/10130 with 26290aa.

PR #10090, opened 2016-07-08
fix test suite by declaring a deleter for CDataCacheCore

FernetMenta:
see title

### DIFF
--- a/xbmc/ServiceManager.cpp
+++ b/xbmc/ServiceManager.cpp
@@ -111,3 +111,9 @@ CDataCacheCore& CServiceManager::GetDataCacheCore()
 {
   return *m_dataCacheCore;
 }
+
+// deleters for unique_ptr
+void CServiceManager::delete_dataCacheCore::operator()(CDataCacheCore *p) const
+{
+  delete p;
+}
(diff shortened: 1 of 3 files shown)


xbmc/ServiceManager.cpp
```diff
@@ -110,4 +110,10 @@ ActiveAE::CActiveAEDSP& CServiceManager::GetADSPManager()
 CDataCacheCore& CServiceManager::GetDataCacheCore()
 {
   return *m_dataCacheCore;
-}
\ No newline at end of file
+}
+
+// delters for unique_ptr
```
hudokkow: s/delters/deleters

FernetMenta:
jenkins build this please

Paxxi:
I have a feeling that this hides the real issue as we really shouldn't have to specify a deleter. Poking around with it I can't find anything that seems wrong that would cause the original issue though so we should be fine.

FernetMenta:
As @notspiff said, either include the header in ServiceManager.h or specifying a deleter.

Paxxi:
Ahh I missed that comment, anyway issue solved, thanks

PR #10091, opened 2016-07-08
[AML] CPU temperature readings for devices with aml_thermal interface

drieschel:
Adds correct cpu temp readings for Amlogic devices which have the aml_thermal (AML_PLATFORM_THERMAL) interface enabled in kernel.

It would be great to get some feedback.

### DIFF
--- a/xbmc/utils/CPUInfo.cpp
+++ b/xbmc/utils/CPUInfo.cpp
@@ -24,6 +24,7 @@
 #include "utils/Temperature.h"
 #include <string>
 #include <string.h>
+#include <sys/stat.h>
 
 #if defined(TARGET_DARWIN)
 #include <sys/types.h>
@@ -254,8 +255,17 @@ CCPUInfo::CCPUInfo(void)
     m_cores[core.m_id] = core;
   }
 #else
+  struct stat stBuf;
+  m_isAmlThermal = stat("/sys/bus/platform/devices/aml_thermal", &stBuf) == 0 && S_ISDIR(stBuf.st_mode);  
   m_fProcStat = fopen("/proc/stat", "r");
-  m_fProcTemperature = fopen("/proc/acpi/thermal_zone/THM0/temperature", "r");
+  if(m_isAmlThermal)
+  {
+    m_fProcTemperature = fopen("/sys/class/thermal/thermal_zone1/temp", "r");
+    if (m_fProcTemperature == NULL)   
+      m_fProcTemperature = fopen("/sys/class/thermal/thermal_zone0/temp", "r");  
+  }
+  if (m_fProcTemperature == NULL)
+    m_fProcTemperature = fopen("/proc/acpi/thermal_zone/THM0/temperature", "r");
   if (m_fProcTemperature == NULL)
     m_fProcTemperature = fopen("/proc/acpi/thermal_zone/THRM/temperature", "r");
   if (m_fProcTemperature == NULL)
@@ -625,7 +635,9 @@ bool CCPUInfo::getTemperature(CTemperature& temperature)
     if (!ret)
     {
       ret = fscanf(m_fProcTemperature, "%d", &value);
-      value = value / 1000;
+      if(!m_isAmlThermal)
+        value = value / 1000;
+
       scale = 'c';
       ret++;
     }
--- a/xbmc/utils/CPUInfo.h
+++ b/xbmc/utils/CPUInfo.h
@@ -111,6 +111,7 @@ private:
   FILE* m_fProcTemperature;
   FILE* m_fCPUFreq;
   bool m_cpuInfoForFreq;
+  bool m_isAmlThermal = false;
 #elif defined(TARGET_WINDOWS)
   PDH_HQUERY m_cpuQueryFreq;
   PDH_HQUERY m_cpuQueryLoad;


Jalle19:
cc @codesnake ?

drieschel:
Is there nobody @Kodi Team who can review this?

Paxxi:
There's nothing wrong with the code so I don't see any issues. But I'm not responsible for Linux really.

fritsch:
CPUInfo is one of the most poluted files we have, ifdeffery en masse and now with your change the normal linux path was moved into the else clause assuming that aml path is the happy path, which makes these changes run on every other system, be it AML or not.

There is a AMLOGIC define in the code to clean this code further.

The second commit needs squashing.

fritsch:
In short: 

- It works without the additional member
- if AML does not need the / 1000, just multiply the read value by 1000 and fine
- compact the checking code for thermal so that other platforms are not affected
- if you need an ifdef which is bad practice and sucks (but the whole file is polluted that way) - so be it.

drieschel:
Thanks for your feedback guys!

@fritsch
- I squashed the last commit. 
- The aml_thermal interface returns degree.
- Other platforms shouldn't be affected.
- There is no extra ifdef statetment needed for this PR.

So the only thing is missing, that someone can confirm the correct pathes for aml_thermal interface, right? 

I was thinking about that a part of this code could be moved into xbmc/utils/AMLutils.cpp as a function is_aml_thermal() which could be used instead of the m_isAmlThermal member property. Not sure if this is a good idea or not.

fritsch:
After your change every other linux needs to check m_isAmlThermal which it did not have to do before.

Two liner: http://sprunge.us/cRJY

^^ do I miss something?

drieschel:
My first approach was exactly your example. One problem is the order of the pathes: Sometimes /sys/class/thermal/thermal_zone0/temp and /sys/class/thermal/thermal_zone1/temp are existent but /sys/class/thermal/thermal_zone1/temp is the right one for reading the temperature. So first I put /sys/class/thermal/thermal_zone1/temp over /sys/class/thermal/thermal_zone0/temp. For having less problems I decided checking if the aml_thermal interface is existent (activated in kernel). I don't think that this one more filesystem read has a big performance or resource impact. And of course I also didn't want to use an extra if(n)?def statement. Would be a tempDivisor member more acceptable than the isAmlThermal?

fritsch:
Basically there are two things I want to avoid:

- code in generic code, that has nothing to do here, I don't like the ifdef myself
- workarounds for lazy kernel development
- basically every AML device out there uses its own set of patches with different behaviour. There is no reason thermal_zone1 should have priority if not documented anywhere otherwise

CPUInfo is full of #ifdefs anyways

stefansaraev:
my 2 cents on this. the whole class is a mess. there is no stable sysfs thermal api. it is same s*** on every new device with every next linux kernel (yes, including mainline kernels on x86). 

thermal readings (and cpufreq) is not worth fixing, sooner it is removed from kodi - the better.

fritsch:
Yeah - same opinion. So as less code as possible, here, please. If it only helps one device while not helping three others, distros need to patch themselves.

@stefansaraev do you have the entries of the wetek devices? Which are those? Is current code working on them?

stefansaraev:
I dont have CONFIG_AML_PLATFORM_THERMAL enabled on my core kernel config, and I am not sure it is supposed to work. I remember some known saradc (?) issue in the past. will test in 15min and update this post.

drieschel:
Of course this should be a more general solution, this was my intention at least. Otherwise it doesn't make sense merging this. This is also the reason why I am checking if the aml_thermal interface is activated in kernel. Interesting would be also to know, if the aml_thermal interface is working identical in aml 3.10 and 3.14 kernel. I only tested it with 3.10.

stefansaraev:
existence of /sys/bus/platform/devices/aml_thermal does not guarantee that CONFIG_AML_PLATFORM_THERMAL is compiled in, and in use. devices/aml_thermal/ exists as long it is declared in devicetree.

after enabling CONFIG_AML_PLATFORM_THERMAL (with or without CONFIG_AML_VIRTUAL_THERMAL) here, I still dont have /sys/class/thermal. I dont pretend my devicetree is correct. I never cared that much for thermal readings ;)

drieschel:
I own a WeTek Core which I can use for testing, I tested it with S802 devices so far. For the S802 devices I am using a fork of @codesnake's amlogic-3.10.y fork, where aml_thermal seems to work with.

Compiling for the WeTek_Core will take some time (Using LibreELEC). I can post my results tomorrow.

drieschel:
Okay, I tested it with WeTek Core: It's not working with the default Amlogic kernel for arm architecture in the LibreELEC project, which is probably https://github.com/codesnake/linux/tree/c8d5b2fc05adbed51d77942675b5268448abeb45. But it is working on WeTek Core and my MXIII 2GB S802 box with another kernel (https://github.com/kszaq/linux/tree/b09c8547cee45dabbf72abd2e4756f1e828f2d26). Probably @kszaq knows why it is working with his kernel.

mcaptur:
Tried it on WeTek Play using LibreELEC Master and Kodi Master. It doesn't work, tried patching kernel with kszaq/linux@7f12bb6 and it still does not work. (Play is Meson 6 whilst this patch seems to be for Meson 8b)

fritsch:
Then let's close this for now until the aml Interface, implementation settles. Dpecialized hacks can be considered by distribution shipping kodi.

drieschel:
I don't think that this patch is meson8b (S805) specific. It worked also on meson8 (S802) and meson8m2 (S812). I already created a pull request with this patch at @codesnake's amlogic-linux. It's a question of kernel configration I think.

mcaptur:
ok let me play around with kernel config.. if i get it to work will submit a pull request to LibreELEC

codesnake:
I will re-check thermal stuff in Amlogic kernel(s). There was always a mess.
@stefansaraev If I recall correctly saradc was working on WeTek Play, but reported wrong temperature values (too high).

PR #10093, opened 2016-07-08
PluginDirectory: dont time out while script is modal/not cancelled

tamland:
Complete fix for #10080. Nothing in foreground (not just modals) should time out and attempted stopped.

Also, since scripts can end at any time, it should not block the caller for such long time waiting on completion (as it did not do before, so this brings back the old behaviour).

### DIFF
--- a/xbmc/filesystem/PluginDirectory.cpp
+++ b/xbmc/filesystem/PluginDirectory.cpp
@@ -71,7 +71,8 @@ void CPluginDirectory::CScriptObserver::Abort()
 }
 
 CPluginDirectory::CPluginDirectory()
-  : m_cancelled(false)
+  : m_fetchComplete(true)
+  , m_cancelled(false)
   , m_success(false)
   , m_totalItems(0)
 {
@@ -474,8 +475,6 @@ bool CPluginDirectory::RunScriptWithParams(const std::string& strPath)
 
 bool CPluginDirectory::WaitOnScriptResult(const std::string &scriptPath, int scriptId, const std::string &scriptName, bool retrievingDir)
 {
-  bool cancelled = false;
-
   // CPluginDirectory::GetDirectory can be called from the main and other threads.
   // If called form the main thread, we need to bring up the BusyDialog in order to
   // keep the render loop alive
@@ -486,35 +485,26 @@ bool CPluginDirectory::WaitOnScriptResult(const std::string &scriptPath, int scr
       CScriptObserver scriptObs(scriptId, m_fetchComplete);
       if (!CGUIDialogBusy::WaitOnEvent(m_fetchComplete, 200))
       {
-        cancelled = true;
+        m_cancelled = true;
       }
       scriptObs.Abort();
     }
   }
   else
   {
-    // kill the script if it does not return within 30 seconds
-    // and we are not waiting for user input
+    // Wait for directory fetch to complete, end, or be cancelled
+    while (!m_cancelled
+        && CScriptInvocationManager::GetInstance().IsRunning(scriptId)
+        && !m_fetchComplete.WaitMSec(20));
+
+    // Give the script 30 seconds to exit before we attempt to stop it
     XbmcThreads::EndTime timer(30000);
-    while (CScriptInvocationManager::GetInstance().IsRunning(scriptId))
-    {
-      if (m_fetchComplete.WaitMSec(1000))
-      {
-        break;
-      }
-      if (g_windowManager.HasModalDialog( {DialogModalityType::MODAL }))
-      {
-        timer.Set(30000);
-      }
-      if (timer.IsTimePast())
-      {
-        cancelled = true;
-        break;
-      }
-    }
+    while (!timer.IsTimePast()
+          && CScriptInvocationManager::GetInstance().IsRunning(scriptId)
+          && !m_fetchComplete.WaitMSec(20));
   }
 
-  if (cancelled)
+  if (m_cancelled)
   { // cancel our script
     if (scriptId != -1 && CScriptInvocationManager::GetInstance().IsRunning(scriptId))
     {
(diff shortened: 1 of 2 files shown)


FernetMenta:
I don't see that this does fix anything. I strategy was that the scripts gets killed if it does not return withing a given timeout.

xbmc/filesystem/PluginDirectory.cpp
```diff
@@ -486,35 +485,26 @@ bool CPluginDirectory::WaitOnScriptResult(const std::string &scriptPath, int scr
       CScriptObserver scriptObs(scriptId, m_fetchComplete);
       if (!CGUIDialogBusy::WaitOnEvent(m_fetchComplete, 200))
       {
-        cancelled = true;
+        m_cancelled = true;
```
FernetMenta: this does not help if the plugin has opened another dialog like keyboard for input.

xbmc/filesystem/PluginDirectory.cpp
```diff
@@ -486,35 +485,26 @@ bool CPluginDirectory::WaitOnScriptResult(const std::string &scriptPath, int scr
...
       }
       scriptObs.Abort();
     }
   }
   else
   {
-    // kill the script if it does not return within 30 seconds
-    // and we are not waiting for user input
+    // Wait for directory fetch to complete, end, or be cancelled
+    while (!m_cancelled
```
FernetMenta: this can result in deadlock of the entire app, if something in the script goes wrong

xbmc/filesystem/PluginDirectory.cpp
```diff
@@ -486,35 +485,26 @@ bool CPluginDirectory::WaitOnScriptResult(const std::string &scriptPath, int scr
...
-      {
-        timer.Set(30000);
-      }
-      if (timer.IsTimePast())
-      {
-        cancelled = true;
-        break;
-      }
-    }
+    while (!timer.IsTimePast()
```
FernetMenta: waiting for another 30 sec after user has cancelled the script make no sense.

Tolriq:
@FernetMenta I did not test your last iteration but if 

> Also, since scripts can end at any time, it should not block the caller for such long time waiting on completion (as it did not do before, so this brings back the old behaviour).

Is true it should for sure be fixed, a 1 sec script should not take 30sec to return, specially since in my case I timeout at 30 so if there's too much data or slow network to return the JSON data it will fail 100% of the time :( Would be a big regression.

FernetMenta:
@Tolriq 
> Also, since scripts can end at any time, it should not block the caller for such long time

where do you see this in the current code? the code checks if the script is still running in a loop. As soon as the script terminated the caller is not blocked anymore.

Tolriq:
@FernetMenta small typo it was **if true**, I was quoting @tamland so I supposed he have tested and report that because it was true. As I said not at home so was not able to test your iteration :)

tamland:
Having a timeout like this is not a good idea because:
- There's a million different things plugins can do that are perfectly "fine" and should not time out, not just modal and dialogs. Including that it can just be naturally long running.

- When it's in the foreground user is already giving consent to continue and can cancel it at any moment. Timeout in this case have no use.

- Lastly, there seem to be some confusion around what 'stop' does. It does in no way reliably "kill" the script. There is only one scenario where it will work: when the script isn't doing any blocking operation. 

    In other cases it's only going to *cause* deadlocks and corruption. Therefore it should be avoided until you absolutely have to. That's why the 30 wait is there. Since directory fetch happens in the background there's no need to rush and stop it.

tamland:
>>Also, since scripts can end at any time, it should not block the caller for such long time waiting on completion (as it did not do before, so this brings back the old behaviour).

>Is true it should for sure be fixed, a 1 sec script should not take 30sec to return, specially since in my case I timeout at 30 so if there's too much data or slow network to return the JSON data it will fail 100% of the time :( Would be a big regression.

I was referring to the 1 second wait for completion in the loop, nothing as dramatic as that;). It if *ends* (as opposed to complete) it's a 1 second gui delay we don't need.

xbmc/filesystem/PluginDirectory.cpp
tamland: if something in the script goes wrong, the script ends and the loop exits

FernetMenta: consider a malfunctioning script that is stuck and won't end. there has to be way to terminate this script

FernetMenta:
> When it's in the foreground user is already giving consent to continue and can cancel it at any moment. Timeout in this case have no use.

If this is always the case, I agree with you. But I don't see that this is the case. If the script brings up an additional window, the user can't cancel the busy dialog.

xbmc/filesystem/PluginDirectory.cpp
tamland: it will be attempted stopped when kodi shuts down, then this will terminate. I don't see what else can be done about it

tamland:
>If this is always the case, I agree with you. But I don't see that this is the case. If the script brings up an additional window, the user can't cancel the busy dialog.

Example on how to produce that? I just tried with a script opening a dialog from a plugin and it seem to work just fine. The busy dialog can be cancelled both before and after.

xbmc/filesystem/PluginDirectory.cpp
FernetMenta: I have been thinking that malfunctioning scripts can be killed by the app. I did not observe any issues with killed scripts during my tests. What are you concerns?

FernetMenta:
I was thinking that plugins intercept user actions but this does not seem to be the case. Worst that can happen is that a user has to push back twice, first to close the dialog opened by the plugin, then to cancel the busy dialog. Is this correct? If so, you are right and an additional timeout makes indeed not much sense.

FernetMenta:
yes, please go ahead with this.

tamland:
jenkins build this please

PR #10096, opened 2016-07-09
Android: Insane hack - workaround CPU load for surface rendering

fritsch:
The title says it all. "Workarounds", "Dump fixes" for RenderUpdate after the change, that we ask for frames repeatadly driven by display refreshrate. The real fix here is to wait for next vblank, which currently is not possible on Droids's implementation.

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
@@ -179,6 +179,7 @@ bool CRendererMediaCodecSurface::RenderUpdateVideoHook(bool clear, DWORD flags, 
     }
 
     next_render = now + frame_time + sleep_time;
+    CLog::Log(LOGINFO, "Fps: %lf Now: %lf next_render: %lf sleep_time: %lf", (double) fps, (double) now, (double) next_render, (double) sleep_time);
   }
 
   usleep(sleep_time * 1000);
--- a/xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
@@ -25,6 +25,7 @@
 
 #include "DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.h"
 #include "utils/log.h"
+#include "threads/SystemClock.h"
 
 CRendererMediaCodecSurface::CRendererMediaCodecSurface()
 {
@@ -156,6 +157,31 @@ bool CRendererMediaCodecSurface::RenderUpdateVideoHook(bool clear, DWORD flags, 
     mci->RenderUpdate(srcRect, dstRect);
   }
 
+  /** @todo This is a really bad hack and not meant to be
+  * a solution. This needs proper fixing when android maintainer
+  * is back. Real solution: Block depending on next vblank,
+  * which is estimated here by assuming 2 buffers and frametime.
+  */
+  uint64_t sleep_time = 10; // 10 ms by default
+  double fps = g_graphicsContext.GetFPS();
+
+  if (fps > 0.0)
+  {
+    uint64_t now = XbmcThreads::SystemClockMillis();
+    uint64_t frame_time = static_cast<uint64_t>((1000.0 / fps + 0.5));
+    if (next_render <= now) // we are late don't sleep
+      sleep_time = 0;
+    else
+    {
+       double diff = next_render - now;
+       // leave 10% room
+       sleep_time = std::min(frame_time, static_cast<uint64_t>(diff * 0.9 + 0.5));
+    }
+
+    next_render = now + frame_time + sleep_time;
+  }
+
+  usleep(sleep_time * 1000);
   return true;
 }
 
--- a/xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.h
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.h
@@ -58,6 +58,9 @@ protected:
   virtual bool RenderHook(int index);  
   virtual int  GetImageHook(YV12Image *image, int source = AUTOSOURCE, bool readonly = false);
   virtual bool RenderUpdateVideoHook(bool clear, DWORD flags = 0, DWORD alpha = 255);
+
+private:
+  uint64_t next_render = 0; // abs time when next render is due
 };
 
 #endif


mazeyx:
ohh thank you so much for this, thought it would never be solved. its strange since with surface hw accelaration it uses more cpu then off.

mazeyx:
just tested it, cpu load went from 80% to 25% for me, amazing. :)

fritsch:
@mazeyx How much load do you get with surface rendering disabled and the very same sample? 

As we "fake" waiting for vblank we are not really allowed to block even longer, the 10000 might be too much in some situations, too.

mazeyx:
@fritsch around 50% with surface rendering disabled.

fritsch:
Thx much. And with v16 + surface?
Am 11.07.2016 12:45 vorm. schrieb "mazeyx" <notifications@github.com>:

mazeyx:
@fritsch i cant install it atm on my tv since i have v17 plus database. but i remember it being around 40% it was also around that point with v17 before the simplify render loop commit was added then it went upto 80%.. so its even better now, i have been watching stuff since yesterday with this test build kodi-20160709-edb6ca8-surface2-armeabi-v7a.apk and it has worked flawlessly, no stuttering at all before because of the 80% occasionally if any other load was added it would stutter, now it is all fine.

fritsch:
oO if that simple hack improves that situation that much in comparison to
v16, good that the render loop was changed to reveal such platform issues.

Not saying this code here is any good, just sad that it needs major general
changes so that these are fixed on such platforms. Same with other bypass
platgormd, people whine after that change not seeing, that what came to
light was a huge showstopper already before.
Am 11.07.2016 7:30 vorm. schrieb "mazeyx" <notifications@github.com>:

mazeyx:
@fritsch a reduction from 80% cpu load and occasional stuttering to 25% with no stuttering i can see no noticable differences at all in quality etc, i dont care how you did it haha just glad it works for now :) because for the past few months its been broken in my opinion now its fixed.

fritsch:
@mazeyx try the updated build, please - to see if load is further reduced

mazeyx:
@fritsch Its a little better, its around 22% cpu. it jumps around abit but im watching the lows and peaks and avarage and it is about 3% lower same sample.

fritsch:
Any regressions?
Am 12.07.2016 12:35 vorm. schrieb "mazeyx" <notifications@github.com>:

mazeyx:
@fritsch i did have 1 video stop for no reason whilst playing but this happens rarely and might of had nothing to do with this. i kept running stuff trying to get it to happen again by but it hasnt i even tried going back to xmb by pressing return then going back to video alot and seeking alot and all seems fine. the cpu is definitely using less now so it is improved just had no idea why that one video stopped half way through could it be related to this? it has happened before in the past its just rare. couldnt find anything else wrong with it.

mazeyx:
@fritsch still playing fine here hasn't cut out again tried alot of different types of files (think it was just that rare occurrence, i played the same file again and it worked fine), it really is amazing how little cpu it is now using if you take away the 10% cpu when its just idle, its using around 12% to play 720hd mkv stuff.  much better then the 80% cpu or so and occasional spike to 100% and stutter i been dealing with for months now. im more then happy with it. thanks so much.

fritsch:
jenkins build this please and may the source code one day tell future devs and users that it was meant as a method to lengthen the dying phase of a platform with hope at least somewhere in mind.

MrMC:
fyi, both SPMC and MrMC have support for android's Choreographer to give vsync feedback

fritsch:
As has v17, it was picked. The intention of the dumpfix is different - low
hanging fruit for new maintainer, perhaps :-)
Am 12.07.2016 6:59 nachm. schrieb "Scott Davilla" <notifications@github.com

koying:
For future maintainer reference, triggering a CEvent in CXBMCapp::doFrame should mimic the Linux "waitforvsync".

FernetMenta:
for the record: this "hack" has a major flaw because it does not consider past render times. If this is called after frametime + x, it will sleep again but it shouldn't.

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
```diff
@@ -156,6 +156,17 @@ bool CRendererMediaCodecSurface::RenderUpdateVideoHook(bool clear, DWORD flags,
     mci->RenderUpdate(srcRect, dstRect);
   }
 
+  /** @todo This is a really bad hack and not meant to be
+  * a solution. This needs proper fixing when android maintainer
+  * is back. Real solution: Block until your next vblank.
```
FernetMenta: this is not a good guidance. it certainly should not wait for next vblank if "next" vblank has just passed right before this method was called.

mazeyx:
well im +1 for merge, still running fine here and without this it runs at 80% cpu which is not good, perhaps down the road we can find a proper fix.

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
koying: @popcornmix this is how it's done for rpi, is it not ? cf. https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/MMALRenderer.cpp#L467
What's your definition of "if next vblank has just passed right before this method was called", and how are renderers suppose to know that?
Bottom-line: What are bypass renderers supposed to wait for?

FernetMenta: > Bottom-line: What are bypass renderers supposed to wait for?

Depending on how accurately you want to sync with audio and subtitles, you can

1) use this hack, results in poor sync and maybe skipped frames when subs are involved
2) throttle render loop to fps by using time only. This is rather trivial. All you need is a member that remembers last rendering times. Sync may be off because player has no info when a particular frame was rendered.
3) Proper sync by lettering player know when a particular frame was rendered to display.

koying: 2) is "wait((1/fps) - (current_time - m_lastRenderTime))", right?
3) Where / how do you let player know? This is actually trivial, because you can assume it's the time of ::RenderUpdate() with an error of 1 frame at worst

mazeyx: 1) Im testing now as i was curious with this patch and subtitles on seems in sync and it says drop:0 skip:0 pc:1 after a few minutes.

fritsch: Yes, you are right. Then what Pi / IMX want to do is not enough, as we need the interval, e.g. the clock to decided "how much" already has passed.

fritsch:
@mazeyx that's not the point - I will change the comment accordingly to state how it should work.
vblank ==2==1==3=vblank

2: don't wait for vblank, wait something shorter
3: Wait for vblank
1: decide according your metric

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
FernetMenta: > 2) is "wait((1/fps) - (current_time - m_lastRenderTime))", right?

Depends on the amount of internal buffers used. If triple buffers are used, it should also take m_lastRenderTime-1 into account. Such a system can recover even it was already late more than frametime.

> 3) Where / how do you let player know? This is actually trivial, because you can assume it's the time of ::RenderUpdate() with an error of 1 frame at worst

Due to the fact that current architecture is outgrown, the main loop has to serve input, gui rendering, video rendering (by-pass). VideoPlayer assumes that when a new cycle starts and last rendered frame was frame(n), 
frame(n - (g_Windowing.NoOfBuffers -1)) has made it to display. (additional display latency was omitted here)
In other words: If internal buffers are full, RenderUpdate should block until frame(n - (g_Windowing.NoOfBuffers -1)) has become visible.
This is the behaviour of swapBuffers and by-pass renderers have to emulate this behaviour if they want to stay in sync with the rest of the system.

mazeyx:
Im just saying without this "hack" we get extremely high cpu usage and occasional stutter, so it might not be a proper fix but its a lot better then what the current build does without this fix.

fritsch:
@mazeyx: quit clear - reason known. The discussion moved on to fix it
properly which is a good thing.
Am 13.07.2016 1:55 nachm. schrieb "mazeyx" <notifications@github.com>:

mazeyx:
just tested it, works nicely, slightly more cpu being used but still very low. :)

fritsch:
What is "more" in numbers? I will add some tempirary debug logging to see
if it does what I had in mind.
Am 13.07.2016 11:31 nachm. schrieb "mazeyx" <notifications@github.com>:

mazeyx:
The average is around 26% now so its up 4% from the last build.

fritsch:
We will see - currently building.

mazeyx:
just waiting for 96837cc test build and will get a debug log for you. :)

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
```diff
@@ -156,6 +157,32 @@ bool CRendererMediaCodecSurface::RenderUpdateVideoHook(bool clear, DWORD flags,
...
+  * a solution. This needs proper fixing when android maintainer
+  * is back. Real solution: Block depending on next vblank,
+  * which is estimated here by assuming 2 buffers and frametime.
+  */
+  uint64_t sleep_time = 10; // 10 ms by default
+  double fps = g_graphicsContext.GetFPS();
+
+  if (fps > 0.0)
+  {
+    uint64_t now = XbmcThreads::SystemClockMillis();
```
FernetMenta: this is not correct. SystemClockMillis is of type unsigned int and can wrap. If you don't take this into account, you may sleep here for a very long time.

mazeyx:
heres the logfile, sorry xbmclogs is currently down.

http://s000.tinyupload.com/index.php?file_id=64082561536032273809

fritsch:
Could you use paste.ubuntu.com - this one is above is blocked here, thx.

mazeyx:
http://paste.ubuntu.com/19344539/

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
fritsch: See the std::min in the fixup commits which adresses the issue.
Am 14.07.2016 8:01 vorm. schrieb "Rainer Hochecker" <
notifications@github.com>:

fritsch:
@mazeyx thx, works as expected. starts with less sleeps and converges to roughly 1000 / 60 * 0.9 which is around 14 to 15 ms sleep time.

mazeyx:
@fritsch yeah, running great here all day. :)

koying:
@mazeyx Mind testing http://mirrors.kodi.tv/test-builds/android/arm/kodi-20160714-1b4fe79-fixhighcpu-armeabi-v7a.apk ?
It's vblank-based, like rpi.

mazeyx:
@koying one step ahead of you, runs good i thought it was fritsch's build cant really tell the difference. very similiar cpu usage im seeing.

koying:
Thanks. Here is the commit, for reference: https://github.com/koying/xbmc/commit/1b4fe79b2ce35ee346a0fecbfc8e52466c464256

fritsch:
@koying's fix is to be preferred as it binds to the vsync coreographer. I will close mine here in favor of this code. Thx much.

koying:
It's just a commit for reference, not a PR.
Someone still has to bring it in...

ping @MartijnKaijser

fritsch:
It would be good to additionally remember the time of the event (vblanks) to see if we are already late and can perhaps skip a frame by extending the mci->RenderUpdate.

FernetMenta:
There is not need for an additional skipping logic, this is already done by RenderManager.

fritsch:
The bypass renderer needs to care though, which I currently don't see it does.

FernetMenta:
you are wrong. look at the code, it takes the surface given by a particular render buffer and renderes it.
https://github.com/koying/xbmc/blob/1b4fe79b2ce35ee346a0fecbfc8e52466c464256/xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp#L126

If RenderManager wants to skip a frame, it skips a frame just like GL renderers do as well.

FernetMenta:
A system that uses double buffers can recover from a render cycle that did last longer than frametime.
The self resetting event in @koying 's patch won't block if vblank has already occurred. This gives the renderer the chance to recover. In case of lateness, RenderManager will take action.

fritsch:
@FernetMenta ah, cool - thanks much. So we have a nice solution then.

PR #10098, opened 2016-07-10
Smart playlists: fix how song based rules applied to artists and albums lists

DaveTBlake:
Since adding role and path rules to "artists" type smart playlists I realised that all the rules based on song criteria e.g. genre, path and role, rather than simply a field in the artist table, need to be combined in SQL rather than left as spearate SQL clauses for each rule.

In logic terms the artists with songs satisfying criteria A and songs satisfying criteria B is not the always same as those artists with songs satisfying criteria A and B.

The effects of this can be seen in Jarvis with an "artists" smart playlist with 2 genre rules: genre = A AND genre = B - what is wanted is all artists with songs that have both genre A and B, what you get is artists that have both at least a song  with genre A and at least a song with genre B. This is not major, but once role and path rule combinations are possible as in v17 then the impact becomes more significant.

Also having selected a playlist of artists using song criteria when that list is played it is the songs that satisfy those criteraia that are wanted. Say for example the artist has some songs of genre Jazz and other songs are Rock. You make a playlist of Jazz artists i.e. artists with songs of genre = Jazz, then currently when you play that list you would get all the songs by that artist rather than just the Jazz ones. Hence song based rules need transfering to filter the subsequent list of albums or songs shown when navigating down from an artist on smart playlist or when playing that list.

This is particularly important when there are role rules e.g. "artist is composer", currently clicking on a composer from such an artist list can result in nothing being displayed despite the fact thay they do compose some songs because the role rule is not passed to the list of albums or artists.

### DIFF
--- a/xbmc/dbwrappers/DatabaseQuery.cpp
+++ b/xbmc/dbwrappers/DatabaseQuery.cpp
@@ -333,6 +333,14 @@ std::string CDatabaseQueryRule::GetOperatorString(SEARCH_OPERATOR op) const
   return operatorString;
 }
 
+std::string CDatabaseQueryRule::GetOperatorFormat(const std::string& strType) const
+{
+  SEARCH_OPERATOR op = GetOperator(strType);
+  std::string operatorString = GetOperatorString(op);
+
+  return operatorString;
+}
+
 std::string CDatabaseQueryRule::GetWhereClause(const CDatabase &db, const std::string& strType) const
 {
   SEARCH_OPERATOR op = GetOperator(strType);
--- a/xbmc/dbwrappers/DatabaseQuery.h
+++ b/xbmc/dbwrappers/DatabaseQuery.h
@@ -77,6 +77,7 @@ public:
   void                        SetParameter(const std::vector<std::string> &values);
 
   virtual std::string         GetWhereClause(const CDatabase &db, const std::string& strType) const;
+  virtual std::string         GetOperatorFormat (const std::string& strType) const;
 
   int                         m_field;
   SEARCH_OPERATOR             m_operator;
--- a/xbmc/dialogs/GUIDialogMediaFilter.cpp
+++ b/xbmc/dialogs/GUIDialogMediaFilter.cpp
@@ -100,7 +100,8 @@ static const CGUIDialogMediaFilter::Filter filterList[] = {
   { "artists",      FieldGenre,         515,    SettingTypeList,    "list",   "string",   CDatabaseQueryRule::OPERATOR_EQUALS },
 
   { "albums",       FieldAlbum,         556,    SettingTypeString,  "edit",   "string",   CDatabaseQueryRule::OPERATOR_CONTAINS },
-  { "albums",       FieldArtist,        557,    SettingTypeList,    "list",   "string",   CDatabaseQueryRule::OPERATOR_EQUALS },
+//  { "albums",       FieldArtist,        557,    SettingTypeList,    "list",   "string",   CDatabaseQueryRule::OPERATOR_EQUALS }, ####Blake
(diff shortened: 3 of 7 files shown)


DaveTBlake:
Remove `std::to_string()` that Android compiler didn't like despite being standard C++11

jenkins build this please

DaveTBlake:
Replace `std::find()` that Android compiler didn't like either. Seems we need to stick to eariler C++

jenkins build this please

DaveTBlake:
@phate89 don't know if you can help me move this forwards, merge window is looming?

DaveTBlake:
Since I have missed the merge window broaden this change to rework the "albums" smart playslists in a similar way. The Issues were less pronounced here as all songs from an album are often by the same artist and have the same genre. But still rules genre = A and genre = B should be applied together, finding albums that have one or more songs with both genre = A and B.

DaveTBlake:
Since it is much easier to see this new SQL in acton than read though the process of building the varied where clauses from rules/options there is a test build on the mirrors http://mirrors.kodi.tv/test-builds/win32/KodiSetup-20160725-996658a-HEAD.exe

Living in hope that someone will find time to test this.

tamland:
You know, (not that I care much about music) but the massive increate in query complexity this is adding is not good. MusicDatabase is already a horrible mess from long again, and to be frank, I don't see how this is moving it in the right direction..

Trust me, if you simplify things and take the complexities you must at a higher level (not sql) you'll achieve way better performance in the end.

DaveTBlake:
@tamland these queries really are not that complex, and they run efficiently enough on both SQLite and MySQL now. Unfortunately building the query clauses in bits looks ugly and complex, but that is pretty normal if you want to give users the power to create any combination of rules in a smart playlist.  It is just manipulation of some strings as building blocks, nothing technical.

There are horrible messes in Kodi, but actually MusicDatabase,cpp isn't that bad. It may have many lines of code, but what it does it pretty simple.  The performance issues for the music library are all to do with populating lists of CFileItem and the mysteries of GUIInfoManger. Refactoring those is more than one person's work.

As for moving in the right direction... for those with big music collections powerful smart playlists or custom nodes are essential. The point of a library is not just to show pretty pictures while listening, but to make navigation and selection of what to play a great experience too. We store the data, this is just about making that data useful for the user. 

But thank you for taking an interest to comment, even discent is better than silence.

MartijnKaijser:
any objections?

MartijnKaijser:
ping? Not sure who to review

jjd-uk:
@DaveTBlake do you still need someone to review this?

DaveTBlake:
>do you still need someone to review this?

Theoretically yes, in practice there seems little point until v17 is out. 

I have taken some of the work off into a separate PR that is waiting for approveal, and I really do want that in v17. But the navigation from smart playlists, or combining with filtering, remains flawed. 

In SQL terms it is not hard to fix, and as far as I am concerned the point of having a db is to exploit the data fully and flexibly. But to anyone not familiar with the schema, or with SQL, is just going to see it as "query complexity" as @tamland did. Here I also try to stay within the existing object model - SQL being constructed in two separate units MusicDatabase.cpp  and SmartPlaylist.cpp - but they need to interact at a lower level and the resulting fix is not the nicest design.  If I was starting from a clean sheet I would not have done it like this, as it was this was intended as an interim fix for a summer release.

So it remains a WIP instead. Input as to how we make full use of the data we have (in library navigation and playlists etc.), and how an implementation fits in with whatever architecture is planned for the future is very welcome. The SQL it produces is the SQL needed, how it does that in an acceptable way I am less clear.

Rechi:
@DaveTBlake this needs a rebase

phunkyfish:
@DaveTBlake 

Is this PR still relevant or can it be closed?

Is it's something you'd like to revisit we can add a backburner label.

DaveTBlake:
I hope one day I will get this finished :) 
 I have done more locally towards a v19 version, and find it helpful to have this sitting here in my open PR as a reminder. So marked backburner.

PR #10099, opened 2016-07-10
Duplicate include removed

drieschel:
The AMLUtils.h was included twice. Don't see a benefit for.

Also <string.h> and <string> are included. Could one of the two statements get removed?

### DIFF
--- a/xbmc/utils/AMLUtils.cpp
+++ b/xbmc/utils/AMLUtils.cpp
@@ -25,7 +25,6 @@
 #include <fcntl.h>
 #include <string>
 
-#include "AMLUtils.h"
 #include "utils/CPUInfo.h"
 #include "utils/log.h"
 #include "utils/SysfsUtils.h"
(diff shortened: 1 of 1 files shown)


Paxxi:
sorry to say this but you removed the wrong one :) That one should stay and the utils/AMLUtils.h should go. `<string>` and `<string.h>` are different files, the first is the c++ include for `std::string` and the second are for C string functions.

drieschel:
I think it doesn't matter which one have to get removed, because both versions link to the same folder. So should get the other "#include utils/*.h" statements changed to "#include *.h" ?

drieschel:
@Paxxi Now it should be okay.

Paxxi:
Thanks :)

There's a convention when it comes to `#include` and generally in a .cpp file the first include should be the corresponding header file without a path. Having it first catches many errors and not having a path means you can freely move the files and they'll find each other correctly

PR #10102, opened 2016-07-11
Handle Remote Typing via web interface

shali3:
There's a problem today while using the "Remote" functionality if the default web interface. When typing text remotely only one character is sent and the virtual keyboard is immediately closes.
This  fix handles continuous typing with support for space and backspace as well.
This follows @PhenX fix from 2 years ago #2033 

I hope you'll like it :)


shali3:
Yay!

PR #10103, opened 2016-07-12
IPlayer: properly sync playspeed with application

FernetMenta:
see title. There has been several workarounds in the past due to not properly synchronised playspeed. Having a variable in CApplicationPlayer assuming players do always have exactly the speed from the last command does not work. 
Further player wants to change its speed according to events like failed seek i.e. when hitting the end of the stream on ff.

Removed CApplication::CheckPlayingProgress
I could not see this method doing anything useful. During testing this method caused the system hang in an endless while loop if speed == 0

@popcornmix I also removed bApplicationMuted from SetPlaySpeed. It turned out that omx_player was the only client of this. Such kind of specialisations should not be visible at the application layer. Please move appropriate action to player.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -3406,12 +3406,8 @@ PlayBackRet CApplication::PlayFile(CFileItem item, const std::string& player, bo
 
   if (iResult == PLAYBACK_OK)
   {
-    // if player has volume control, set it.
-    if (m_pPlayer->ControlsVolume())
-    {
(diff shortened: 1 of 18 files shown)


popcornmix:
@FernetMenta I've tested omxplayer and this PR. Seems to build fine and I couldn't spot an obvious regression. Mute still works, and audio is muted during trickplay.

FernetMenta:
thanks for testing. seems it was some obsolete code.

FernetMenta:
jenkins build this please

PR #10104, opened 2016-07-12
Add ipv6 support to Event Server

Tolriq:
More than inspired by http://trac.kodi.tv/ticket/12491 with some fixes

Use dual stack when possible with fallback to ipv4 if any problem to not end with an ipv6 listener only.

Tested and working on Windows. Needs tests on other platforms.

CC @Montellese / @FernetMenta as I saw some feedback on other ipv6 patch from you.

Had a few report about that missing, and since everything else support ipv6 it should be enabled for event server too.

### DIFF
--- a/xbmc/network/EventServer.cpp
+++ b/xbmc/network/EventServer.cpp
@@ -153,14 +153,10 @@ void CEventServer::Process()
 
 void CEventServer::Run()
 {
-  CAddress any_addr;
   CSocketListener listener;
   int packetSize = 0;
 
-  if (!CSettings::GetInstance().GetBool(CSettings::SETTING_SERVICES_ESALLINTERFACES))
-    any_addr.SetAddress("127.0.0.1");  // only listen on localhost
-
-  CLog::Log(LOGNOTICE, "ES: Starting UDP Event server on %s:%d", any_addr.Address(), m_iPort);
+  CLog::Log(LOGNOTICE, "ES: Starting UDP Event server on port %d", m_iPort);
 
   Cleanup();
 
@@ -186,7 +182,7 @@ void CEventServer::Run()
     CLog::Log(LOGERROR, "ES: Invalid port range specified %d, defaulting to 10", port_range);
     port_range = 10;
   }
-  if (!m_pSocket->Bind(any_addr, m_iPort, port_range))
+  if (!m_pSocket->Bind(!CSettings::GetInstance().GetBool(CSettings::SETTING_SERVICES_ESALLINTERFACES), m_iPort, port_range))
   {
     CLog::Log(LOGERROR, "ES: Could not listen on port %d", m_iPort);
     return;
--- a/xbmc/network/Socket.cpp
+++ b/xbmc/network/Socket.cpp
@@ -38,13 +38,58 @@ using namespace SOCKETS;
 /* CPosixUDPSocket                                                    */
 /**********************************************************************/
 
-bool CPosixUDPSocket::Bind(CAddress& addr, int port, int range)
+bool CPosixUDPSocket::Bind(bool localOnly, int port, int range)
 {
   // close any existing sockets
   Close();
 
-  // create the socket
-  m_iSock = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
+  // If we can, create a socket that works with IPv6 and IPv4.
+  // If not, try an IPv4-only socket (we don't want to end up
+  // with an IPv6-only socket).
+  if (!localOnly) // Only bind loopback to ipv4. TODO : Implement dual bindinds.
+  {
+    m_iSock = socket(AF_INET6, SOCK_DGRAM, IPPROTO_UDP);
+    if (m_iSock != INVALID_SOCKET)
+    {
+#ifdef WINSOCK_VERSION
+      const char zero = 0;
+#else
+      int zero = 0;
+#endif
+      if (setsockopt(m_iSock, IPPROTO_IPV6, IPV6_V6ONLY, &zero, sizeof(&zero)) == -1)
+      {
+        close(m_iSock);
+        m_iSock = INVALID_SOCKET;
+      }
+      else
+      {
+        m_addr = CAddress("::");
+
+        SOCKET testSocket = socket(AF_INET6, SOCK_DGRAM, IPPROTO_UDP);
+        setsockopt(testSocket, IPPROTO_IPV6, IPV6_V6ONLY, &zero, sizeof(&zero));
+        // Try to bind a socket to validate ipv6 status
+        for (m_iPort = port; m_iPort <= port + range; ++m_iPort)
+        {
+          m_addr.saddr.saddr6.sin6_port = htons(port);
+          if (bind(testSocket, (struct sockaddr*)&m_addr.saddr, m_addr.size) >= 0)
+          {
+            closesocket(testSocket);
+            m_ipv6Socket = true;
+            break;
+          }
+        }
+        if (!m_ipv6Socket)
+        {
+          CLog::Log(LOGWARNING, "UDP: Unable to bind to advertised ipv6, fallback to ipv4");
+          close(m_iSock);
+          m_iSock = INVALID_SOCKET;
+        }
+      }
+    }
+  }
+
+  if (m_iSock == INVALID_SOCKET)
+    m_iSock = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
 
   if (m_iSock == INVALID_SOCKET)
   {
@@ -52,7 +97,7 @@ bool CPosixUDPSocket::Bind(CAddress& addr, int port, int range)
     int ierr = WSAGetLastError();
     CLog::Log(LOGERROR, "UDP: Could not create socket %d", ierr);
     // hack for broken third party libs
-    if(ierr == WSANOTINITIALISED)
+    if (ierr == WSANOTINITIALISED)
     {
       WSADATA wd;
       if (WSAStartup(MAKEWORD(2,2), &wd) != 0)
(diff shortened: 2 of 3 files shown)


xbmc/network/Socket.h
```diff
@@ -175,6 +213,9 @@ namespace SOCKETS
   protected:
     SOCKET   m_iSock;
     CAddress m_addr;
+
+  private:
+    boolean m_ipv6Socket;
```
stefansaraev: bool

Tolriq: Yes already fixed saw the travis log :) Too much java those times.

stefansaraev:
dual stack does not work on linux and loopback interfaces. you should bind on both ::1 and 127.0.0.1

Tolriq:
This is what is done for the rest, and will fallback to ipv4 in that case. If OS is ipv6 only then it would have failed before and will continue to fail now.

Dual binding requires much more changes that could prevent this simple working approach to be in V17.

stefansaraev:
it will failback to v4 if setsockopt(..., IPPROTO_IPV6,...) fails. as long ipv6 is enabled (which is the case with most major linux distros) on user's box, no matter if he has global v6 address or not, you'll bind on ::1 only.

before this pr:
```
udp        0      0 127.0.0.1:9777          0.0.0.0:*                           180/kodi.bin
```

after this pr:
```
udp        0      0 ::1:9777                :::*                                178/kodi.bin
```

clients, that are configured to use 127.0.0.1 by default (like kodi-send), will fail to connect now.

stefansaraev:
it is not a bug. it is how dual stack works on linux. and it does not work on lo, by design. as said, once you have v6 support enabled in your kernel, on the loopback interface you will have both ::1/128 and 127.0.0.1/8 addresses added, it does not fail to bind on ::1. but ::1 is different than 127.0.0.1. no bug

Tolriq:
PR updated, and lo specific to linux may not be a bug, but no RFC ;)

stefansaraev:
thanks. works for me.

and well, maybe I did not express myself right.

unlike :: (in6addr_any, everything) ::1 is a specific ipv6 address (ipv6 equivalent of 127.0.0.1). when you bind to a specific address, you can't expect that a socket would accept connections to any other address. this rule applies when you bind to non-local address, too. 

for example if you have 2a01:1234::1/48 and 192.168.0.1/24, if you bind to 2a01:1234::1, clients can't connect to 192.168.0.1. if you have a secondary address 2a01:1234::2, clients cant use that, too.

in6addr_any (:: - the v6 equivalent of INADDR_ANY - 0.0.0.0) is a different story. no idea how it is supposed to work on windows, but on linux, it will accept connections directed to any configured ip address (both v4 and v6) that's available on the system, from any client ip address (both v4 and v6).  you dont even need to check setsockopt(IPV6_V6ONLY), just bind and if bind succeeds, it works for both v4 and v6 clients.

::1 is a valid, always configured ip address when you run v6-enabled kernel. nothing special here. thats why you need to bind another socket if you want clients to connect to 127.0.0.1

xbmc/network/Socket.cpp
```diff
@@ -38,13 +38,38 @@ using namespace SOCKETS;
...
+  {
+    m_iSock = socket(AF_INET6, SOCK_DGRAM, IPPROTO_UDP);
+    if (m_iSock != INVALID_SOCKET)
+    {
+#ifdef WINSOCK_VERSION
+      const char zero = 0;
+#else
+      int zero = 0;
+#endif
+      if (setsockopt(m_iSock, IPPROTO_IPV6, IPV6_V6ONLY, &zero, sizeof(&zero)) == -1)
```
FernetMenta: this does not work for detecting if ipv6 is available or not. I already fixed this somewhere else some years ago:
https://github.com/xbmc/xbmc/pull/3398
you have to bind the socket for reliable detection

xbmc/network/Socket.cpp
```diff
@@ -77,24 +102,38 @@ bool CPosixUDPSocket::Bind(CAddress& addr, int port, int range)
...
+    else
+      m_addr = CAddress("0.0.0.0");
+  }
 
   // bind the socket ( try from port to port+range )
-  while (m_iPort <= port + range)
+  for (m_iPort = port; m_iPort <= port + range; ++m_iPort)
   {
-    if (bind(m_iSock, (struct sockaddr*)&m_addr.saddr, sizeof(m_addr.saddr)) != 0)
+    if (m_ipv6Socket) {
```
FernetMenta: wrong coding style, curly brackets new line

Tolriq:
PR updated to address @FernetMenta remarks.

And thanks @stefansaraev much more cleaner explanation ;) This is now logical :p

xbmc/network/Socket.h
```diff
@@ -70,21 +74,54 @@ namespace SOCKETS
...
+      }
+      else {
+        return inet_ntoa(saddr.saddr4.sin_addr);
+      }
     }
 
     unsigned long ULong()
     {
-      return (unsigned long)saddr.sin_addr.s_addr;
+      if (saddr.saddr_generic.sa_family == AF_INET6) {
```
FernetMenta: there are sill many of these coding style errors

Tolriq:
@FernetMenta should be ok now, I'm too much used to Android Studio that format on pasting according to my config :(

Is there any tools in VS to do that ?

Paxxi:
@Tolriq VS has built in support for it, tools->options->text editor->c++ and configure the style you want

xbmc/network/Socket.h
```diff
@@ -51,16 +51,20 @@ namespace SOCKETS
   class CAddress
   {
   public:
-    sockaddr_in saddr;
+    union {
```
FernetMenta: :)

FernetMenta:
looks ok to me. @Paxxi what do you think?

Tolriq:
Fixed last one :p

Thanks @Paxxi I looked at that before and only saw the tab / space things, but all is there :)

Do think a PR with some config like with https://visualstudiogallery.msdn.microsoft.com/410e9b9f-65f3-4495-b68e-15567e543c58 or with exported settings for the relevant part would be accepted to simplify this a lot for new comers. 

I lost so many time on that ...

Paxxi:
@fernetmenta I'm a bit rusty on sockets but something feels off, I'll try to review properly tonight or tomorrow.

xbmc/network/Socket.h
```diff
@@ -70,21 +75,56 @@ namespace SOCKETS
 
     void SetAddress(const char *address)
     {
+      in6_addr addr6;
       memset(&saddr, 0, sizeof(saddr));
-      saddr.sin_family = AF_INET;
-      saddr.sin_addr.s_addr = inet_addr(address);
-      size = sizeof(saddr);
+      if (inet_pton(PF_INET6, address, &addr6) == 1)
```
Paxxi: should be `AF_INET6` and why not use inet_pton for ipv4 as well? would make more sense to stick to one method.
Also there's no sanity checking here, it just assumes that not ipv6 is ipv4 which isn't true, it could very well be that the string is a bad ipv6 address.

xbmc/network/Socket.h
```diff
@@ -70,21 +75,56 @@ namespace SOCKETS
...
+        return inet_ntop(AF_INET6, &saddr.saddr6.sin6_addr, buf, size);
+      }
+      else
+        return inet_ntoa(saddr.saddr4.sin_addr);
     }
 
     unsigned long ULong()
     {
-      return (unsigned long)saddr.sin_addr.s_addr;
+      if (saddr.saddr_generic.sa_family == AF_INET6)
```
Paxxi: ULong seems to only be used by CEventServer, rather than doing tricks to fit into an uin32_t I would suggest scrapping this method entirely and just string notation of IP to be protocol agnostic.

Paxxi:
I'm thinking the reasonable thing to do would be to add dual-socket support here to simplify for all callers. On Windows no real work is needed but on other platforms it would make sense that any higher layer don't have to care.

Not sure I'm a fan of the special casing for local only for Bind either.

It's a valid attempt though and an improvement over what we have but it could be even better :)

Also @Tolriq adding something like that to the repo is perfectly fine, I've thought about it a few times but never gotten around to actually doing it so I welcome the idea.

xbmc/network/EventServer.cpp
```diff
@@ -186,7 +182,7 @@ void CEventServer::Run()
     CLog::Log(LOGERROR, "ES: Invalid port range specified %d, defaulting to 10", port_range);
     port_range = 10;
   }
-  if (!m_pSocket->Bind(any_addr, m_iPort, port_range))
+  if (!m_pSocket->Bind(!CSettings::GetInstance().GetBool(CSettings::SETTING_SERVICES_ESALLINTERFACES), m_iPort, port_range))
```
Paxxi: This predates this PR so not meant as critique. Are there any good reasons for wanting to bind to localhost only? It seems to complicate our lives and I have a hard time to see any benefit from it.

stefansaraev: yes. obviously we dont want random people from internet connecting to our services.

Paxxi: But is there a real use case for it? you obviously have a modified setup with other apps running locally. Regular users might not configure their firewall but do they make use of the local bind? Seems more power user oriented.

stefansaraev: yes. security (no, I am neither paranoid, nor a security fanatic). expose as less you can by default to the bad guys waiting to make a botnet out of your app :)

most users will never need a non-local eventserver and jsonserver

Paxxi: Isn't eventserver used by the various remotes like Kore which aren't local?

fritsch: @paxxi: Take a bit of sleep :-)

2016-07-12 22:26 GMT+02:00 Pär Björklund <notifications@github.com>:

> In xbmc/network/EventServer.cpp
> <https://github.com/xbmc/xbmc/pull/10104#discussion_r70515393>:
>
> > @@ -186,7 +182,7 @@ void CEventServer::Run()
> >      CLog::Log(LOGERROR, "ES: Invalid port range specified %d, defaulting to 10", port_range);
> >      port_range = 10;
> >    }
> > -  if (!m_pSocket->Bind(any_addr, m_iPort, port_range))
> > +  if (!m_pSocket->Bind(!CSettings::GetInstance().GetBool(CSettings::SETTING_SERVICES_ESALLINTERFACES), m_iPort, port_range))
>
> Isn't eventserver used by the various remotes like Kore which aren't
> local?
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/xbmc/xbmc/pull/10104/files/12e006dfee5da0f398b01ae317ea13c2a1780c4c#r70515393>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABCfHe8uTzqJltm3j958HjTTavU-3o1iks5qU_iDgaJpZM4JKWO1>
> .
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

Paxxi: Waiting on a build so no sleep yet :)

Tolriq: By default ES is localhost only, with a settings to open to all but is needed by remote :)

Maybe @Paxxi does not know but I'm the author of Yatse and currently 1 million monthly users do rely on ES accessible from LAN ;) Kore have added it recently also with limited things.

ES on local host is used too by some tools like on mac for some hardware remotes.

So nothing have to be changed on settings and global usage, it's just about adding support for IPV6 to something that already works and fit needs.

For information I fixed ES implementation recently, before there was a huge security problem as by default ES was on on all interfaces.

I have long history with ES ;)

xbmc/network/Socket.h
Tolriq: The goal of this PR is to address something reported by a few users with minimal changes to avoid discussions and allow merge for v17.

xbmc/network/Socket.h
Tolriq: As on other comment I chose the simplest path to limit changes to ensure quick validation.

Tolriq:
> I'm thinking the reasonable thing to do would be to add dual-socket support here to simplify for all callers. On Windows no real work is needed but on other platforms it would make sense that any higher layer don't have to care.
Not sure I'm a fan of the special casing for local only for Bind either.
It's a valid attempt though and an improvement over what we have but it could be even better :)

This PR is about fixing a problem without too much impact.
As I said things can be done better for v18, but it's not the point for v17.

If we go the only perfect road, no problem it will wait v18 if I keep motivation but rewrite is a different story than simple fixing. Specially when it would be cool to fix some other things reported like JSON RPC on TCP for localhost that could be useful for Android in future for recommendations as relying on webserver seems odd.

FernetMenta:
@Paxxi any objections for getting this in?

Paxxi:
No objections, it does improve the current state so you're free to merge

FernetMenta:
jenkins build this please

PR #10106, opened 2016-07-13
[wip][guilib] Vertical Settings Slider

AchimTuran:
This PR adds orientiation for settings sliders.

I need this for AudioDSP add-ons. Furthermore it is a first step to generic dialogs that are supported by almost all skins, which support e.g. OSD menus like audio settings. Only the gui control settings API needs to be forwared to binary add-ons.

### DIFF
--- a/addons/skin.estuary/1080i/Defaults.xml
+++ b/addons/skin.estuary/1080i/Defaults.xml
@@ -151,9 +151,10 @@
 		<aligny>center</aligny>
 		<texturefocus border="40" colordiffuse="button_focus">buttons/button-fo.png</texturefocus>
 		<texturenofocus border="40">buttons/button-nofo.png</texturenofocus>
+		<orientation>horizontal</orientation>
 		<sliderwidth>150</sliderwidth>
 		<sliderheight>28</sliderheight>
-		<texturesliderbar>buttons/slider-back.png</texturesliderbar>
+		<texturesliderbar border="10">buttons/slider-back.png</texturesliderbar>
 		<textureslidernib>buttons/slider-nib.png</textureslidernib>
 		<textureslidernibfocus colordiffuse="button_focus">buttons/slider-nib.png</textureslidernibfocus>
 		<font>font13</font>
--- a/xbmc/guilib/GUIButtonControl.h
+++ b/xbmc/guilib/GUIButtonControl.h
@@ -32,12 +32,15 @@
 #include "GUILabel.h"
 #include "GUIControl.h"
 
+class CGUISettingsSliderControl;
+
 /*!
  \ingroup controls
  \brief
  */
 class CGUIButtonControl : public CGUIControl
 {
+  friend class CGUISettingsSliderControl;
 public:
   CGUIButtonControl(int parentID, int controlID,
                     float posX, float posY, float width, float height,
--- a/xbmc/guilib/GUIControlFactory.cpp
+++ b/xbmc/guilib/GUIControlFactory.cpp
@@ -888,8 +888,6 @@ CGUIControl* CGUIControlFactory::Create(int parentID, const CRect &rect, TiXmlEl
 
   XMLUtils::GetFloat(pControlNode, "markwidth", checkWidth);
   XMLUtils::GetFloat(pControlNode, "markheight", checkHeight);
-  XMLUtils::GetFloat(pControlNode, "sliderwidth", sliderWidth);
-  XMLUtils::GetFloat(pControlNode, "sliderheight", sliderHeight);
   if (!GetTexture(pControlNode, "textureradioonfocus", textureRadioOnFocus) || !GetTexture(pControlNode, "textureradioonnofocus", textureRadioOnNoFocus))
   {
     GetTexture(pControlNode, "textureradiofocus", textureRadioOnFocus);    // backward compatibility
@@ -963,8 +961,18 @@ CGUIControl* CGUIControlFactory::Create(int parentID, const CRect &rect, TiXmlEl
   {
     StringUtils::ToLower(strTmp);
     if (strTmp == "horizontal")
+    {
       orientation = HORIZONTAL;
+      // default orientation for a settings slider is vertical
+      // consequently it has to changed here
+      float tmpVal = sliderWidth;
+      sliderWidth = sliderHeight;
+      sliderHeight = tmpVal;
+    }
   }
+
+  XMLUtils::GetFloat(pControlNode, "sliderwidth", sliderWidth);
+  XMLUtils::GetFloat(pControlNode, "sliderheight", sliderHeight);
   XMLUtils::GetFloat(pControlNode, "itemgap", buttonGap);
   XMLUtils::GetInt(pControlNode, "movement", iMovementRange);
   GetAspectRatio(pControlNode, "aspectratio", aspect);
@@ -1257,7 +1265,7 @@ CGUIControl* CGUIControlFactory::Create(int parentID, const CRect &rect, TiXmlEl
     {
       control = new CGUISettingsSliderControl(
         parentID, id, posX, posY, width, height, sliderWidth, sliderHeight, textureFocus, textureNoFocus,
-        textureBar, textureNib, textureNibFocus, labelInfo, SLIDER_CONTROL_TYPE_PERCENTAGE);
+        textureBar, textureNib, textureNibFocus, labelInfo, SLIDER_CONTROL_TYPE_PERCENTAGE, orientation);
 
       ((CGUISettingsSliderControl *)control)->SetText(strLabel);
       ((CGUISettingsSliderControl *)control)->SetInfo(singleInfo);
--- a/xbmc/guilib/GUISettingsSliderControl.cpp
+++ b/xbmc/guilib/GUISettingsSliderControl.cpp
@@ -20,12 +20,20 @@
 
 #include "GUISettingsSliderControl.h"
 
-CGUISettingsSliderControl::CGUISettingsSliderControl(int parentID, int controlID, float posX, float posY, float width, float height, float sliderWidth, float sliderHeight, const CTextureInfo &textureFocus, const CTextureInfo &textureNoFocus, const CTextureInfo& backGroundTexture, const CTextureInfo& nibTexture, const CTextureInfo& nibTextureFocus, const CLabelInfo &labelInfo, int iType)
-    : CGUISliderControl(parentID, controlID, posX, posY, sliderWidth, sliderHeight, backGroundTexture, nibTexture,nibTextureFocus, iType, HORIZONTAL)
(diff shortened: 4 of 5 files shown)


addons/skin.estuary/1080i/Defaults.xml
```diff
@@ -151,9 +151,10 @@
 		<aligny>center</aligny>
 		<texturefocus border="40" colordiffuse="button_focus">buttons/button-fo.png</texturefocus>
 		<texturenofocus border="40">buttons/button-nofo.png</texturenofocus>
+		<orientation>horizontal</orientation>
 		<sliderwidth>150</sliderwidth>
 		<sliderheight>28</sliderheight>
-		<texturesliderbar>buttons/slider-back.png</texturesliderbar>
+		<texturesliderbar border="10">buttons/slider-back.png</texturesliderbar>
```
AchimTuran: @phil65 or @ronie 
Do you know why this is needed for vertical sliders? If it is not present the slider look really ugly.

addons/skin.estuary/1080i/Defaults.xml
```diff
@@ -151,9 +151,10 @@
 		<aligny>center</aligny>
 		<texturefocus border="40" colordiffuse="button_focus">buttons/button-fo.png</texturefocus>
 		<texturenofocus border="40">buttons/button-nofo.png</texturenofocus>
+		<orientation>horizontal</orientation>
 		<sliderwidth>150</sliderwidth>
 		<sliderheight>28</sliderheight>
```
AchimTuran: @ronie @phil65 
How can we provide default values for `sliderwidth` and `sliderheight` for horizontal and vertical sliders? Or should we remove them like we did for GUISliders?

xbmc/guilib/GUIControlFactory.cpp
```diff
@@ -963,8 +961,18 @@ CGUIControl* CGUIControlFactory::Create(int parentID, const CRect &rect, TiXmlEl
   {
     StringUtils::ToLower(strTmp);
     if (strTmp == "horizontal")
+    {
       orientation = HORIZONTAL;
+      // default orientation for a settings slider is vertical
```
FernetMenta: this does not make much sense to me because we currently don't have any vertical sliders. further it would break all existing skins.

xbmc/guilib/GUIButtonControl.h
```diff
@@ -32,12 +32,15 @@
...
 
+class CGUISettingsSliderControl;
+
 /*!
  \ingroup controls
  \brief
  */
 class CGUIButtonControl : public CGUIControl
 {
+  friend class CGUISettingsSliderControl;
```
AchimTuran: @FernetMenta 
I needed quick acces to `m_Label`. Would it make sense to create a new public method to access m_Label. It is needed to properly place the labels at a correct position and size.

xbmc/guilib/GUIControlFactory.cpp
AchimTuran: sorry I meant horizontal sliders.

xbmc/guilib/GUIControlFactory.cpp
```diff
@@ -963,8 +961,18 @@ CGUIControl* CGUIControlFactory::Create(int parentID, const CRect &rect, TiXmlEl
...
+      // consequently it has to changed here
+      float tmpVal = sliderWidth;
```
FernetMenta: it is very unclear that you change the initial values set above. Took me quite a while to see this.

xbmc/guilib/GUIControlFactory.cpp
AchimTuran: It doesn't break horizontal sliders. If you look above the default orientation is `VERTICAL` [see](https://github.com/xbmc/xbmc/pull/10106/files#diff-d4fa1e4244e98591cf23ce6e5f167887R740).
Furthermore the values need to be changed otherwise the slider looks really ugly. It took very long to figure out this issue.

FernetMenta:
I am wondering why orientation is needed here. Isn't it clear that a slider is horizontal if width > height and vertical if width < height?

xbmc/guilib/GUIControlFactory.cpp
AchimTuran: I know. That's why I said guilib source code is very hard to read.

xbmc/guilib/GUIControlFactory.cpp
FernetMenta: it will break all skins but Estuary

AchimTuran:
@FernetMenta 
> I am wondering why orientation is needed here. Isn't it clear that a slider is horizontal if width > height and vertical if width < height?

I implemented it the same way like normal sliders. But yes it is easier to use this information.

What do the skinners say to this? Maybe @ronie or @phil65.

FernetMenta:
I doubt that this will work without rotating the texture image. Estuary uses a simple image for the slider that allows converting to vertical by stretching and compressing the image but this is an assumption not valid for all skins.

addons/skin.estuary/1080i/Defaults.xml
FernetMenta: that is most likely because the image is not rotated

AchimTuran:
> I doubt that this will work without rotating the texture image.

Do we already have functions to rotate textures?

addons/skin.estuary/1080i/Defaults.xml
ronie: indeed. you'll need a new set of textures for vertical settings sliders.
it's the same for slider and scrollbar controls, skins ship with separate textures for horizontal and vertical ones.

xbmc/guilib/GUIControlFactory.cpp
ronie: it's common practice to break skins during development cycles. we do it all the time.

addons/skin.estuary/1080i/Defaults.xml
ronie: you can't define default values for both.
i would keep it as is and define the width/height values on each individual vertical settings slider you use.

BigNoid:
I am not sure how this would work. The way the controls are defined are explicitly for either horizontal or vertical orientation. You can see this for the slider width and height tags for example.
Furthermore we use group lists for the default controls in settings. Having one control in there which has much more height and less width than all the others wouldn't look very good too.

FernetMenta:
> I am not sure how this would work

Assume you want a slider that has width of 20 and height of 200. This you would define the the skin. Kodi will render a vertical slider. Why shouldn't this work?

xbmc/guilib/GUIControlFactory.cpp
FernetMenta: sure but no need to do so if this could easily be avoided. you won't break a skin for the sake of breaking it, right?

BigNoid:
> Assume you want a slider that has width of 20 and height of 200. This you would define the the skin. Kodi will render a vertical slider. Why shouldn't this work?

I am sure it would work on a technical level :) My concerns are about design and navigation.

FernetMenta:
sounds like the way to go is to create a new GUIControl for the vertical slider.

BigNoid:
And a new group list for vertical orientated controls.

ronie:
we never had a need for a new guicontrol for vertical scrollbars or vertical slider controls.
i don't see why settings sliders should be different?

BigNoid:
Because we never had the same control in settings have two orientations.

ronie:
sorry but i don't get it....are we talking about DialogAddonSettings.xml here?

if so, a new grouplist for vertical controls makes sense, but why would you need a new control type if you can just add the orientation tag to get a vertical settings slider?

like:
```

			<control type="sliderex" id="8">
				...
			</control>
			<control type="sliderex" id="9">
				...
				<orientation>vertical</orientation>
			</control>
```

BigNoid:
Sure, if that's possible than there's no need for an extra control type
indeed.

FernetMenta:
@ronie confused :)

> indeed. you'll need a new set of textures for vertical settings sliders.
it's the same for slider and scrollbar controls, skins ship with separate textures for horizontal and vertical ones

didn't you suggest to have separate controls for vertical and horizontal?

ronie:
> didn't you suggest to have separate controls for vertical and horizontal?

nope, i didn't. ;-)

skins can handle vertical settings sliders the same way they handle vertical sliders and vertical scrollbars.

here's an example of how we handle it for scrollbars:
https://github.com/ronie/skin.transparency/blob/master/1080p/Includes.xml#L147-L180
each one uses their own set of textures and the orientation tag.

HitcherUK:
While we're here we've often discussed making these controls only move when you actually click on them, this way any movement controls (onleft, onright for horizontal sliders) will work correctly.

FernetMenta:
@HitcherUK can you elaborate on this? What do you think is wrong the way those work currently?

HitcherUK:
Currently if you had a horizontal slider you can't navigate left/right to a different control without adjusting the slider first. The idea was to be able to navigate left/right without actually moving the slider and to adjust it you'd press enter to activate it.

FernetMenta:
Yes, this makes sense. Thanks for explaining it to me.

AchimTuran:
> While we're here we've often discussed making these controls only move when you actually click on them, this way any movement controls (onleft, onright for horizontal sliders) will work correctly.

100+ on this, currently it's quite confusing to navigate with slider controls.

AchimTuran:
@FernetMenta @ronie 
> didn't you suggest to have separate controls for vertical and horizontal?

> > nope, i didn't. ;-)

Great. A new control would things only more complex. Orientation is an option and doesn't require a new control.
I don't know what is better for skinners use orientanation or by defining width > height for vertical sliders. But the second approach will reduce the posibilities.

AchimTuran:
@FernetMenta 
>    I doubt that this will work without rotating the texture image.

> > Do we already have functions to rotate textures?

^^ do you know an answer to this question?

BigNoid:
We don't have to rotate the the textures. Skins need to supply the textures for the vertical sliders.
Only add ```<orientation>horizontal</orientation>``` to sliderex control in defaults.xml as its designed for horizontal sliderex atm. 
Then when implementing it in addonsettings or settings category add a new control id for the vertical sliderex where skins can define the design for that. Plus imo it needs a new group list for vertical orientated controls as its really bad ux to have those controls in a group list that navigates vertically.

AchimTuran:
@BigNoid 
I don't speak from a skinner perspective, I meant we should rotate the texture with C/C++ and guilib to provide a better looking slider. Currently it is only stretched, which could break for other skins. Without this option `border="10"` the slider will look really ugly.

If you want you can have a look at my my latest work for [adsp.template](https://github.com/AchimTuran/adsp.template/blob/adsp.template_readd/adsp.template/resources/skins/skin.estuary/1080i/DialogGainMode.xml).
But currently it is not possible to create controls directly from binary add-ons through C/C++, which would allow generic dialogs for many skins for binary add-ons.
volume_dialog_preview

BigNoid:
Maybe I don't understand correctly but isn't the image stretched because it's designed for a horizontal slider? As said skins need to explicitly make textures for vertical sliders, like they do for scrollbars as per @ronie example

AchimTuran:
@BigNoid 
Sorry I miss understod the hints that @ronie mentioned.
Right the skin has to provide textures for vertical sliders, which is currently missing. Could someone do this? Or should I rotate them?

Rechi:
@AchimTuran this needs a rebase

MartijnKaijser:

PR #9225, opened 2016-02-27
Fix out of memory crash in CArchive

Paxxi:
This should fix this issue

```
Thread 1 (Thread 0x7f6369ff3700 (LWP 13304)):
#0  0x00007f63da306cc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007f63da30a0d8 in __GI_abort () at abort.c:89
#2  0x00007f63da90b535 in __gnu_cxx::__verbose_terminate_handler() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007f63da9096d6 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007f63da909703 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007f63da909922 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007f63da909e0d in operator new(unsigned long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007f63da909ea9 in operator new[](unsigned long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x0000000000ff265c in CArchive::operator>>(std::string&) ()
#9  0x0000000000e6fb6f in CFileItemList::Archive(CArchive&) ()
#10 0x0000000000ff19ff in CArchive::operator>>(IArchivable&) ()
#11 0x0000000000e67bb3 in CFileItemList::Load(int) ()
#12 0x0000000000f5942f in UPNP::CUPnPServer::OnBrowseDirectChildren(NPT_Reference<PLT_Action>&, char const*, char const*, unsigned int, unsigned int, char const*, PLT_HttpRequestContext const&) ()
```

size_t is variable sized and if someone copies userdata from a 32-bit to a 64-bit system some string sizes could be quite large as 4 bytes from the string are added onto the size.

I limited everything to 32bit as we likely won't need large items in our archives, also it seems safest compatibility wise.

PR #10107, opened 2016-07-13
Update usage of Crc32 after changing methods to static

Paxxi:
@MilhouseVH this should fix the breakage in #9225


Paxxi:
jenkins build this please

PR #10108, opened 2016-07-13
[darwin] - fix ssl access from python addons

Memphiz:
As pointed out here http://forum.kodi.tv/showthread.php?tid=282604 SSL urls are not working currently from python scripts on all darwin platforms (ios, osx, future tvos).

This PR corrects it by packaging the openssl/cacert.pem and setting the environment variable SSL_CERT_FILE (similar to what is done for android).

@FernetMenta fyi
@fetzerch no clue where to put that Makefile change in the new cmake build sys ... it simply needs to copy the cacert.pem from depends to system/certs ...

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -458,6 +458,7 @@ bool CApplication::Create()
     g_graphicsContext.ResetOverscan((RESOLUTION)i, CDisplaySettings::GetInstance().GetResolutionInfo(i).Overscan);
   }
 
+  //! @todo - move to CPlatformXXX
 #ifdef TARGET_POSIX
   tzset();   // Initialize timezone information variables
 #endif
@@ -465,6 +466,7 @@ bool CApplication::Create()
   // Grab a handle to our thread to be used later in identifying the render thread.
   m_threadID = CThread::GetCurrentThreadId();
 
+  //! @todo - move to CPlatformXXX
 #ifndef TARGET_POSIX
   //floating point precision to 24 bits (faster performance)
   _controlfp(_PC_24, _MCW_PC);
@@ -474,7 +476,8 @@ bool CApplication::Create()
   win32_exception::install_handler();
 
 #endif
-  
+
+  //! @todo - move to CPlatformXXX
   #if defined(TARGET_POSIX)
     // set special://envhome
     CSpecialProtocol::SetEnvHomePath(getenv("HOME"));
@@ -492,6 +495,7 @@ bool CApplication::Create()
   CopyUserDataIfNeeded("special://masterprofile/", "favourites.xml");
   CopyUserDataIfNeeded("special://masterprofile/", "Lircmap.xml");
   
+  //! @todo - move to CPlatformXXX
   #ifdef TARGET_DARWIN_IOS
     CopyUserDataIfNeeded("special://masterprofile/", "iOS/sources.xml", "sources.xml");
   #endif
@@ -520,6 +524,8 @@ bool CApplication::Create()
   buildType = "Unknown";
 #endif
   std::string specialVersion;
+  
+  //! @todo - move to CPlatformXXX
 #if defined(TARGET_RASPBERRY_PI)
   specialVersion = " (version for Raspberry Pi)";
 //#elif defined(some_ID) // uncomment for special version/fork
@@ -540,6 +546,7 @@ bool CApplication::Create()
     CLog::Log(LOGNOTICE, "Running on %s, kernel: %s %s %d-bit version %s", g_sysinfo.GetOsPrettyNameWithVersion().c_str(),
               g_sysinfo.GetKernelName().c_str(), g_sysinfo.GetKernelCpuFamily().c_str(), g_sysinfo.GetKernelBitness(), g_sysinfo.GetKernelVersionFull().c_str());
 
+  //! @todo - move to CPlatformXXX ???
 #if defined(TARGET_LINUX)
 #if USE_STATIC_FFMPEG
   CLog::Log(LOGNOTICE, "FFmpeg statically linked, version: %s", FFMPEG_VERSION);
@@ -560,6 +567,8 @@ bool CApplication::Create()
     CLog::Log(LOGNOTICE, "Host CPU: %s, %d core%s available", cpuModel.c_str(), g_cpuInfo.getCPUCount(), (g_cpuInfo.getCPUCount() == 1) ? "" : "s");
   else
     CLog::Log(LOGNOTICE, "%d CPU core%s available", g_cpuInfo.getCPUCount(), (g_cpuInfo.getCPUCount() == 1) ? "" : "s");
+  
+  //! @todo - move to CPlatformXXX ???
 #if defined(TARGET_WINDOWS)
   CLog::Log(LOGNOTICE, "%s", CWIN32Util::GetResInfoString().c_str());
   CLog::Log(LOGNOTICE, "Running with %s rights", (CWIN32Util::IsCurrentUserLocalAdministrator() == TRUE) ? "administrator" : "restricted");
@@ -597,6 +606,7 @@ bool CApplication::Create()
   std::string strExecutablePath = CUtil::GetHomePath();
 
   // for python scripts that check the OS
+  //! @todo - move to CPlatformXXX
 #if defined(TARGET_DARWIN)
   setenv("OS","OS X",true);
 #elif defined(TARGET_POSIX)
@@ -652,6 +662,7 @@ bool CApplication::Create()
 
   update_emu_environ();//apply the GUI settings
 
+  //! @todo - move to CPlatformXXX
 #ifdef TARGET_WINDOWS
   CWIN32Util::SetThreadLocalLocale(true); // enable independent locale for each thread, see https://connect.microsoft.com/VisualStudio/feedback/details/794122
 #endif // TARGET_WINDOWS
@@ -693,6 +704,7 @@ bool CApplication::Create()
     return false;
   }
 
+  //! @todo - move to CPlatformXXX
 #if defined(TARGET_DARWIN_OSX)
   // Configure and possible manually start the helper.
   XBMCHelper::GetInstance().Configure();
--- a/xbmc/interfaces/python/XBPython.cpp
+++ b/xbmc/interfaces/python/XBPython.cpp
@@ -593,9 +593,6 @@ bool XBPython::OnScriptInitialized(ILanguageInvoker *invoker)
     CEnvironment::putenv(buf);
     buf = "OS=win32";
     CEnvironment::putenv(buf);
-
-#elif defined(TARGET_ANDROID)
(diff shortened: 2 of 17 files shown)


xbmc/interfaces/python/XBPython.cpp
```diff
@@ -581,6 +581,10 @@ bool XBPython::OnScriptInitialized(ILanguageInvoker *invoker)
       CLog::Log(LOGDEBUG, "PYTHONHOME -> %s", CSpecialProtocol::TranslatePath("special://frameworks").c_str());
       CLog::Log(LOGDEBUG, "PYTHONPATH -> %s", CSpecialProtocol::TranslatePath("special://frameworks").c_str());
     }
+
+#if defined(TARGET_DARWIN)
```
FernetMenta: imo this should be set before the application is launched, not in the code. it should not overwrite user set environment variables (overwrite flag -> 0)

Memphiz: mhh on osx there is not even a place i know of to set environment variables for normal gui app bundles. The only thing i can think off is to check the env before setting it to have it as fallback. Though i am 99% sure that there is no easy way for the user to set an environment for kodi from extern. On iOS its even more unlikely that this var overwrites anything explicitly set by the user. Would a check before setting it in code acceptable?

FernetMenta: There is OSX specific startup code: SDLMain.mm. Why not putting it there?

> Would a check before setting it in code be acceptable?

Just set the overwrite flag to 0:
setenv("SSL_CERT_FILE", blabla, 0);

Memphiz: Coming back to this one. SDLMain.mm is not so good because i wanted a spot where it works for all darwin platforms. I also selected this spot because the very same thing is already done for android in this file. You want me to find a "better" spot? In that case the android part will still be in this file ...

FernetMenta: the fix for Android here was wrong and should go away. SSL_CERT_FILE is not a python thing, hence this is a bad spot to add it. Environment settings should all be in a single place. That is where app is started or before.

Memphiz: should i add such a spot in form of a new class and call some init method somewhere at the very beginning where platforms can add such code to?

FernetMenta: That would be great. Maybe some object that also collects runtime information about the system, like OS version, gfx vendor/model, etc. 
Could be owned by ServiceManager

fetzerch:
@Memphiz: Sorry this slipped through and I only noticed it now. I've implemented the necessary cmake bits in https://github.com/xbmc/xbmc/pull/10162/commits/a6af81b9ac8548ffa12f84b1e400d2fd715563b2 (as part of https://github.com/xbmc/xbmc/pull/10162)

Memphiz:
@fetzerch thx - does this also cover ios?

fetzerch:
Yes! `project/cmake/scripts/ios/Macros.cmake` is a symlink to `project/cmake/scripts/osx/Macros.cmake` because the macros are all the same (wrapping, finding so-names, and this).

dokavsoft2:
Hi folks,

Is this fix already available in latest night builds?

Keep up the good work!
Regards

Memphiz:
No its not - Lack of Time to do it properly atm

Memphiz:
@FernetMenta i have updated the PR with a reference implementation of the new platform class for darwin. Could you have a look and tell me if this is what we want, just before i start to create classes for the other platforms and hook it up into our various build systems?

xbmc/platform/PlatformFactory.cpp
```diff
@@ -0,0 +1,40 @@
...
+IPlatform *CPlatformFactory::GetPlatform()
+{
+#if defined(TARGET_DARWIN)
+  return new CPlatformDarwin();
+#endif
+}
+
+void CPlatformFactory::DeletePlatform(IPlatform *platform)
+{
+  if (platform != nullptr)
```
garbear: nullptr check is unnecessary

xbmc/platform/PlatformFactory.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include "PlatformFactory.h"
+#include "interfaces/IPlatform.h"
+#if defined(TARGET_DARWIN)
+#include "platform/darwin/PlatformDarwin.h"
+#endif
+
+IPlatform *CPlatformFactory::GetPlatform()
+{
+#if defined(TARGET_DARWIN)
```
FernetMenta: we could avoid the ifdefs by using this concept: https://github.com/xbmc/xbmc/tree/master/xbmc/cores/VideoPlayer/Process

this also allows to create new ports without touching mainline code.

FernetMenta:
nice. if you have time, please look at my comment

xbmc/Application.cpp
```diff
@@ -458,13 +458,15 @@ bool CApplication::Create()
     g_graphicsContext.ResetOverscan((RESOLUTION)i, CDisplaySettings::GetInstance().GetResolutionInfo(i).Overscan);
   }
 
+// TODO - move to CPlatformXXX
```
MartijnKaijser: can you make those doygen style?
`//! @todo move to CPlatformXXX`

xbmc/platform/PlatformFactory.cpp
Memphiz: Mhhh yes thats neat. But this depends on CORE_SYSTEM_NAME ... there is CORE_SYSTEM_NAME osx and ios ... but when doing the ifdef stuff i have TARGET_DARWIN, TARGET_DARWIN_OSX, TARGET_DARWIN_IOS (and TARGET_DARWIN_TVOS...) - i make use of TARGET_DARWIN here because it applies to all of them. With the linked approach i would need to have the same file there multiple times (one for osx and one for ios).

xbmc/Application.cpp
Memphiz: yep

xbmc/platform/PlatformFactory.cpp
Memphiz: yep thx

xbmc/platform/PlatformFactory.cpp
FernetMenta: You could also have a single override for darwin and do the ifdeffery for ios, tvos, osx in this file, if you don't want to have 3 different ones.

Memphiz: yes but the automagically setting of the needed define wouldn't workt then like done here for example:

https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/Process/CMakeLists.txt#L1

because it looks for the file in CORE_SYSTEM_NAME which would be darwin in that case

FernetMenta: I am sure this can be adapted to the new requirements. We should avoid another ifdeffery factory

Memphiz: yes i could abuse the CORE_SYSTEM_NAME osx ... or hardcode a search for darwin subfolder ...

Memphiz:
@FernetMenta ping  - i think i got it now - are you with me?

FernetMenta:
@Memphiz looks ok to me
thanks!

Memphiz:
jenkins build this please

Memphiz:
jenkins build this please with fixed buildsys

gl-sergei:
Here is how bundled `cacert.pem` looks like:

```
cat /Applications/Kodi.app/Contents/Resources/Kodi/system/certs/cacert.pem.bak
subject=/C=AU/SOP=QLD/O=Mincom Pty. Ltd./OU=CS/CN=SSLeay demo server
issuer= /C=AU/SOP=QLD/O=Mincom Pty. Ltd./OU=CS/CN=CA
-----BEGIN X509 CERTIFICATE-----

MIIBgjCCASwCAQQwDQYJKoZIhvcNAQEEBQAwODELMAkGA1UEBhMCQVUxDDAKBgNV
BAgTA1FMRDEbMBkGA1UEAxMSU1NMZWF5L3JzYSB0ZXN0IENBMB4XDTk1MTAwOTIz
MzIwNVoXDTk4MDcwNTIzMzIwNVowYDELMAkGA1UEBhMCQVUxDDAKBgNVBAgTA1FM
RDEZMBcGA1UEChMQTWluY29tIFB0eS4gTHRkLjELMAkGA1UECxMCQ1MxGzAZBgNV
BAMTElNTTGVheSBkZW1vIHNlcnZlcjBcMA0GCSqGSIb3DQEBAQUAA0sAMEgCQQC3
LCXcScWua0PFLkHBLm2VejqpA1F4RQ8q0VjRiPafjx/Z/aWH3ipdMVvuJGa/wFXb
/nDFLDlfWp+oCPwhBtVPAgMBAAEwDQYJKoZIhvcNAQEEBQADQQArNFsihWIjBzb0
DCsU0BvL2bvSwJrPEqFlkDq3F4M6EGutL9axEcANWgbbEdAvNJD1dmEmoWny27Pn
IMs6ZOZB
-----END X509 CERTIFICATE-----
```

Shouldn't it look more like this one: https://github.com/xbmc/xbmc/blob/master/tools/depends/target/openssl/cacert.pem

Memphiz:
@fetzerch yep - somehow cmake ends up with a wrong cacert.pem not the one we have in our tree in tools/depends/target/openssl/cacert.pem ... i have not the slightest idea how this is even possible.

fetzerch:
Omg. Thanks, indeed an interesting one :) When I tested this locally, I was using system wide installed depends. On Jenkins the build uses depends build into tools/depends. There we apparently have more `cacert.pem` files in subdirectories and the last one wins.

```
#From Jenkins build/ExportFiles.cmake
file(COPY "/Users/Shared/jenkins/workspace/IOS/tools/depends/target/openssl/cacert.pem" DESTINATION "/Users/Shared/jenkins/workspace/IOS/build/system/certs")
file(COPY "/Users/Shared/jenkins/workspace/IOS/tools/depends/target/openssl/iphoneos8.1_armv7-target/apps/demoCA/cacert.pem" DESTINATION "/Users/Shared/jenkins/workspace/IOS/build/system/certs")
file(COPY "/Users/Shared/jenkins/workspace/IOS/tools/depends/target/openssl/iphoneos8.1_armv7-target/demos/cms/cacert.pem" DESTINATION "/Users/Shared/jenkins/workspace/IOS/build/system/certs")
file(COPY "/Users/Shared/jenkins/workspace/IOS/tools/depends/target/openssl/iphoneos8.1_armv7-target/demos/smime/cacert.pem" DESTINATION "/Users/Shared/jenkins/workspace/IOS/build/system/certs")
```

Will work on a fix. Sry.

Memphiz:
So its magic that i didn't see in the cmake code ;)

fetzerch:
Let's not call it magic, but rather unintuitive weirdness in how file globbing works with cmake. Fixed in #10416.

PR #10111, opened 2016-07-14
FIX: [droid] high cpu with amc surface

fritsch:
Thanks to @koying.


fritsch:
jenkins build and merge (it's a fix)

MrMC:
yep, those are the missing bits that are already in SPMC/MrMC.

PR #10115, opened 2016-07-15
player: further cleanup of playing state

FernetMenta:
further cleanup and fixing.

IsPaused must not return a different value as GetPlaySpeed() == 0
That was the case before and as a result all those hacks in GUIInfoManager were needed:

```
     case PLAYER_PLAYING:
       bReturn = !g_application.m_pPlayer->IsPausedPlayback() && (g_application.m_pPlayer->GetPlaySpeed() == 1);
```

This is obviously nonsense because playback can't be paused if playspeed equals 1

### DIFF
--- a/xbmc/ApplicationPlayer.cpp
+++ b/xbmc/ApplicationPlayer.cpp
@@ -193,10 +193,9 @@ bool CApplicationPlayer::HasRDS() const
   return (player && player->HasRDS());
 }
 
-bool CApplicationPlayer::IsPaused() const
+bool CApplicationPlayer::IsPaused()
 {
-  std::shared_ptr<IPlayer> player = GetInternal();
-  return (player && player->IsPaused());
+  return (GetPlaySpeed() == 0);
 }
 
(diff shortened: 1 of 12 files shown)


FernetMenta:
jenkins build this please

Rechi:
@FernetMenta
`PAPlayer::GetSpeed()` returns `1` also if playback is paused.
So `Player.Paused` in the GUI doesn't work if PAPlayer is used.

MilhouseVH:
@rechi that would explain the issue I reported in the Estuary forum: http://forum.kodi.tv/showthread.php?tid=262603&pid=2384301#pid2384301

Rechi:
@MilhouseVH If I play audio with alternative player VideoPlayer it works as expected.

MilhouseVH:
> @MilhouseVH If I play audio with alternative player VideoPlayer it works as expected.

Yes agreed, problem is only with PAPlayer.

MilhouseVH:
@Rochi: I've tested #10277 and this fixes the PAPlayer pause issue.

PR #10119, opened 2016-07-16
[settings] New addons section in settings & move weather/adsp

jjd-uk:
Since the original Settings layout I've not been happy about having Weather in the Services section, in addition the Add-on's category didn't feel right sitting under System. This aims to solve this by introducing a new Add-ons Section in the Settings to provide a home for all core settings related to add-on's. As many skins link to the Add-on Browser in Settings, I've added a button to provide this access to this in the same way as tamland did in PR8607 for accessing dependencies and show running addons, thus skinners can remove any existing link to the Add-on browser in Settings as this will provide a replacement means to access.

I've also moved ADSP from System->Audio as it seems to me to be a natural fit for this new Section, and to my mind has the benefit of making System->Audio less busy in number of settings, plus ADSP having it's own category may aid discovery of the feature by users.

@MartijnKaijser if nothing too controversial I'd like to try and get this in during the current window.

@tamland to introduce the new button to the add-on browser I did a copy and paste of your code from PR8607 then made the necessary adjustments. I've run time tested on Windows and all seems ok, but I guess your +1 would help get this in and make sure I've done nothing wrong as I don't know much about coding.

@Ronie for the skin aspects and since you maintain weather add-ons.

@AlwinEsch @AchimTuran are you ok with me moving the ADSP settings?

As more stuff gets done via add-on's this provides a location for any settings required in Core. This maybe where core settings for Game add-on could go once retroplayer is merged.

Since pictures are sometimes worth more than words:

image

image

image

image


jjd-uk:
jenkins build this please

AchimTuran:
@jjd-uk 
Will this section also handle pvr add-ons?

MartijnKaijser:
not sure i really like this approach cause now you have to go all over the place to get your audio done. In the end this will also get crowded again

ksooo:
I also would like more a functional approach for categorizing the settings, not an approach based on the implementation of a feature (add-on vs. core). How should a a Kodi newbie know that weather and whatever is implemented as an add-on and therefore settings appears under add-ons?

jjd-uk:
Ok, so if we keep Weather and ADSP where they are for the moment and maybe discuss in Team forum an approach for v18, as bejng a percectionist I'd like to get this right. m assuming this might be last window before Beta. I did have different ready that I abandoned in preference

What remains then is it worthwhile keeping this new Addons section for the Manager and General categories, if not I can just PR the addition of the Addon Browser to the existing

da-anda:
I'm also not a fan of technical grouping. As ksooo said, users don't know which part is an add-on or core feature, and as the core code is moving more and more into add-ons itself, where to draw the line? If we think that the current settings system has some limitations as certain sections are getting too crowded, we probably should try to think outside the box and see what we can fix at the presentation layer, or if we would need more flexible nesting, along with a new navigation concept for the settings.

Tolriq:
A side note but completely related as settings are sometimes unclear.

Remote control settings are now moved to Interface, the problem is that Application control part, is 100% related to services.
It's about not only activating Event Server that can emulate keyboard (but much more too) but it also controls the JSON RPC over TCP service that is more or less the exact same thing as the Webserver settings.

Even with docs and screenshots users have problem finding new place.

NedScott:
For the wiki, will a new attempt be made or should I start making v17 screenshots of settings? I'm not in a hurry, but just to prevent myself or someone else from doing it in case there will be more changes.

jjd-uk:
@NedScott I planned to start updating the wiki this weekend to reflect the changes I've made.

PR #10120, opened 2016-07-16
MMAL: Move deinterlacer from decoder to renderer

popcornmix:
Previously we deinterlaced inside the hardware decoder. This has a few downsides:
-     It doesn't work with software decoded video (e.g. MPEG-2 without a codec licence)
-     The deinterlacer stops aspect ratio changes from being reported by decoder meaning they are missed when they occur mid-stream
-     Enabling/disabling deinterlace adjusts the framerate Kodi's renderer sees, and it takes a number of seconds to lock back on. This can cause issues with dropping/stuttering.
-     Interlaced content drives Kodi's GUI loop at a higher rate which increases CPU.

So the new scheme handles deinterlace in the mmal renderer, meaning kodi is unaware of the change in framerate. This also allows the higher quality gpu deinterlace to work with software decode.

I'd like to get this in this merge window. There will probably some tweaks before it is merged.

TODO: Report deinterlace state and the increased deinterlace framerate through processinfo

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/MMALCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/MMALCodec.cpp
@@ -821,7 +821,7 @@ int CMMALVideo::Decode(uint8_t* pData, int iSize, double dts, double pts)
        if (m_hints.ptsinvalid) buffer->pts = MMAL_TIME_UNKNOWN;
        buffer->length = (uint32_t)iSize > buffer->alloc_size ? buffer->alloc_size : (uint32_t)iSize;
        // set a flag so we can identify primary frames from generated frames (deinterlace)
-       buffer->flags = MMAL_BUFFER_HEADER_FLAG_USER0;
+       buffer->flags = 0;
        if (m_dropState)
          buffer->flags |= MMAL_BUFFER_HEADER_FLAG_USER3;
 
@@ -1078,6 +1078,7 @@ bool CMMALVideo::GetPicture(DVDVideoPicture* pDvdVideoPicture)
           pDvdVideoPicture->dts == DVD_NOPTS_VALUE ? 0.0 : pDvdVideoPicture->dts*1e-6, pDvdVideoPicture->pts == DVD_NOPTS_VALUE ? 0.0 : pDvdVideoPicture->pts*1e-6,
           pDvdVideoPicture->iFlags, buffer->mmal_buffer->flags, pDvdVideoPicture->MMALBuffer, pDvdVideoPicture->MMALBuffer->mmal_buffer);
     assert(!(buffer->mmal_buffer->flags & MMAL_BUFFER_HEADER_FLAG_DECODEONLY));
+    buffer->mmal_buffer->flags &= ~MMAL_BUFFER_HEADER_FLAG_USER3;
   }
   else
   {
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/MMALCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/MMALCodec.cpp
@@ -66,7 +66,6 @@ CMMALVideoBuffer::CMMALVideoBuffer(CMMALVideo *omv, std::shared_ptr<CMMALPool> p
   m_aligned_height = 0;
   m_encoding = MMAL_ENCODING_UNKNOWN;
   m_aspect_ratio = 0.0f;
-  m_refs = 0;
   m_rendered = false;
 }
 
@@ -276,7 +275,12 @@ void CMMALVideo::dec_output_port_cb(MMAL_PORT_T *port, MMAL_BUFFER_HEADER_T *buf
     PortSettingsChanged(port, buffer);
   }
   if (!kept)
-    mmal_buffer_header_release(buffer);
+  {
+    if (omvb)
+      omvb->Release();
+    else
+      mmal_buffer_header_release(buffer);
+  }
 }
 
 static void dec_output_port_cb_static(MMAL_PORT_T *port, MMAL_BUFFER_HEADER_T *buffer)
@@ -734,10 +738,7 @@ void CMMALVideo::Reset(void)
       m_output_cond.notifyAll();
     }
     if (buffer)
-    {
-      buffer->Acquire();
       buffer->Release();
-    }
     else
       break;
   }
@@ -806,7 +807,6 @@ bool CMMALVideo::GetPicture(DVDVideoPicture* pDvdVideoPicture)
     pDvdVideoPicture->dts = buffer->mmal_buffer->dts == MMAL_TIME_UNKNOWN ? DVD_NOPTS_VALUE : buffer->mmal_buffer->dts;
     pDvdVideoPicture->pts = buffer->mmal_buffer->pts == MMAL_TIME_UNKNOWN ? DVD_NOPTS_VALUE : buffer->mmal_buffer->pts;
 
-    pDvdVideoPicture->MMALBuffer->Acquire();
     pDvdVideoPicture->iFlags  = DVP_FLAG_ALLOCATED;
     if (buffer->mmal_buffer->flags & MMAL_BUFFER_HEADER_FLAG_USER3)
       pDvdVideoPicture->iFlags |= DVP_FLAG_DROPPED;
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/MMALCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/MMALCodec.cpp
@@ -53,8 +53,9 @@ using namespace KODI::MESSAGING;
 
 #define VERBOSE 0
 
-CMMALVideoBuffer::CMMALVideoBuffer(CMMALVideo *omv)
-    : m_omv(omv)
+
+CMMALVideoBuffer::CMMALVideoBuffer(CMMALVideo *omv, std::shared_ptr<CMMALPool> pool)
+    : m_omv(omv), m_pool(pool)
 {
   if (VERBOSE && g_advancedSettings.CanLogComponent(LOGVIDEO))
     CLog::Log(LOGDEBUG, "%s::%s %p", CLASSNAME, __func__, this);
@@ -63,8 +64,10 @@ CMMALVideoBuffer::CMMALVideoBuffer(CMMALVideo *omv)
   m_height = 0;
   m_aligned_width = 0;
   m_aligned_height = 0;
+  m_encoding = MMAL_ENCODING_UNKNOWN;
   m_aspect_ratio = 0.0f;
   m_refs = 0;
+  m_rendered = false;
 }
 
 CMMALVideoBuffer::~CMMALVideoBuffer()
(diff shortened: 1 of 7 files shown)


popcornmix:
ping @FernetMenta 
Doesn't affect any generic code, so shouldn't break anything non-pi, but feel free to comment.

xbmc/cores/VideoPlayer/DVDCodecs/Video/MMALFFmpeg.h
```diff
@@ -34,16 +34,50 @@ class CGPUMEM;
...
+class CGPUPool
+{
+public:
+  CGPUPool();
+  virtual ~CGPUPool();
+  static void AlignedSize(AVCodecContext *avctx, uint32_t &w, uint32_t &h);
+  CMMALYUVBuffer *GetBuffer(uint32_t mmal_format, uint32_t width, uint32_t height, uint32_t aligned_width, uint32_t aligned_height, uint32_t size, AVCodecContext *avctx);
+  CGPUMEM *AllocateBuffer(uint32_t numbytes);
+  void ReleaseBuffer(CGPUMEM *gmem);
+  unsigned sizeFree() { return m_freeBuffers.size(); }
```
Jalle19: Capital S

fritsch: And we are not the kernel :-) default int.

FernetMenta:
> So the new scheme handles deinterlace in the mmal renderer, meaning kodi is unaware of the change in framerate.

This is not correct and may lead to issues in the future. TV channels for example deliver 50 demuxer packets with a delta in pts of 20ms. Assume the stream has changing subtitles/overlays for every single pts. Player is supposed to deliver 50 frames/sec to renderer, not only half the frame rate. With only half the frame rate you may lose every other overlay.

popcornmix:
@FernetMenta we have always supported half-rate bob and advanced deinterlaced schemes. They don't double the input framerate.
Isn't that is exactly the behaviour you will see with this PR?

I'm not sure Pi would handle subs changing 50 times a second anyway. In the omxplayer case we limit overlay changes to [10 per second](https://github.com/xbmc/xbmc/blob/master/xbmc/cores/omxplayer/OMXPlayerVideo.cpp#L215) just because too many updates makes the video unwatchable.

FernetMenta:
> They don't double the input framerate.

That assumption is wrong (at least not correct for all cases). It is the other way round. Half-rate bob reduced input frame rate to 50% and by doing so it manipulates the stream. Happy path is deinterlacing and keeping input frame rate. IMO every implementation should support the happy path.

popcornmix:
We may be talking at cross purposes. Let me clarify what I am saying (I know you understand how all this works, but I'd like to check we are talking about the same thing).

For all interlaced streams I've been testing, the video decoder produces 25 frames per second, with each frame consisting of two interlaced fields woven together. The timestamp increases by 40ms each frame returned from hardware decoder.

If deinterlace is disabled, or you use one of the "half" deinterlacers, the frame rate is unchanged, and calls to GetPicture have timestamps increasing by 40ms each call.

The "full" deinterlace modes double the amount of data produced. This can be thought of as doubling the framerate or doubling the vertical resolution, depending on whether you think of the original frames as 1920x1080@25 or 1920x540@50. Either way the calls to GetPicture have timestamps increasing by 20ms each call.

Now all the files I've been testing have two fields per frame. I know it is possible to encode an isolated field in a frame. We have code in the hw decoder to clone these to produce a valid full frame. I imagine if this were in use we would be getting 50 frames per second from the decoder so nothing would change with this PR.

So, I believe there is no change in behaviour when deinterlace is disabled or using a half rate mode after this PR. There is a change in behaviour for the full rate modes, but the only effect it would have would be when subtitles/overlays are changing at 50 fps, which may be a case we can't handle anyway.

Do you have an sample file that may show the problem?

FernetMenta:
The question is not if a decoder does combine two field into a single frame or not. If an interlaced stream has i.e. 50 access units per sec or in other word is 50 fps, decoder should honour this on the happy path and submit 50 video pics to renderer. 

The correct behaviour would be to pass the same frame twice to renderer. One time with pts for one field, then with pts for other field.

popcornmix:
@FernetMenta I've added a commit (I originally had this when developing but dropped it).
It makes the RenderUpdate loop get called at 50Hz for a 1080i50 stream and doesn't affect a 24p stream.

mmalrender will ignore the second call (as m_rendered is true), but it makes the generic render code run twice as often.

Does that help?

FernetMenta:
The input frame rate is defined by the stream. In general an interlaced stream in PAL region is comprised of 50 access units per sec. Every access unit has distinct pts/dts. Renderer is supposed to be configured with this frame rate, not 50% of it. 
Currently renderer selects a 50 hz resolution for a config frame rate of 25. But this is based on assumptions and can easily break. When video player will be used for transcoding this assumption will break. Output frame rate will be defined by config frame rate set by player.

Your last commit does not what the commit message suggests. CVideoPlayerVideo::OutputPicture is supposed to be called once for every picture, not combined frame. For an interlaced stream with 50 access units per sec this is 50 times a second. Subtitles for video pictures are selected by this function.

popcornmix:
@FernetMenta I've added a commit to return two pictures for an interlaced frame.
It's not quite right currently, but I wanted to check if this is along the right lines.
GetPicture is now called 50 times a second for a 1080i50 stream.

xbmc/cores/VideoPlayer/DVDCodecs/Video/MMALCodec.cpp
```diff
@@ -849,6 +855,18 @@ bool CMMALVideo::GetPicture(DVDVideoPicture* pDvdVideoPicture)
...
+
+    if (buffer->mmal_buffer->flags & MMAL_BUFFER_HEADER_VIDEO_FLAG_INTERLACED)
+    {
+      pDvdVideoPicture->iFlags |= DVP_FLAG_INTERLACED;
+      if (buffer->mmal_buffer->flags & MMAL_BUFFER_HEADER_VIDEO_FLAG_TOP_FIELD_FIRST)
+        pDvdVideoPicture->iFlags |= DVP_FLAG_TOP_FIELD_FIRST;
+      if (!firstInterlacedFrame)
+      {
+        pDvdVideoPicture->dts = DVD_NOPTS_VALUE;
+        pDvdVideoPicture->pts = DVD_NOPTS_VALUE;
```
FernetMenta: why don't you set pts?

popcornmix: I was unsure if VideoPlayerVideo would produce a better value.

Are you suggesting using the framerate from hints? I'm not sure that is always valid (it is 0 from some PVR backends).

I could use the difference of last two timestamps returned, but I have to make a guess for the first deinterlaced frame, and there are complications with dropped/corrupt frames as well as frames with DVD_NOPTS_VALUE. Seems I would be duplicating code from VideoPlayerVideo.

FernetMenta: VideoPlayerVideo can only guess while decoder should know. Decoder gets 50 distinct access units with pts. Eating half of the pts values reduces information.
VideoPlayerVideo does not know that decoder combines two fields into a frame with only one pts. It has some fallback logic to set pts but this can't ever be as accurate than what decoder can do. Decoder has more information and is in charge of setting pts.

> Are you suggesting using the framerate from hints?

Definitely not. That would be worse than guessing.

popcornmix: Not sure I agree here. With my [1080i60 sample](https://dl.dropboxusercontent.com/u/3669512/temp/1620_20130917060600.mpg) (recording from live TV), I get 30 packets per second from demuxer. 30 packets with timestamps get passed to decoder. I get 30 decoded pictures (frames consisting of two fields) out of decoder.
Any timestamp created for the second deinterlace frames would have to be guessed.

FernetMenta: That is a mpeg2 sample. I get one packet for every field on my h264 channels.

15:59:34 T:123145327652864  NOTICE:  ----------- dts: 1006844.555555, pts: 1046844.555556
15:59:34 T:123145327652864  NOTICE:  ----------- dts: 1026844.555556, pts: 1226844.555556
15:59:34 T:123145327652864  NOTICE:  ----------- dts: 1046844.555556, pts: 1246844.555557
15:59:34 T:123145327652864  NOTICE:  ----------- dts: 1066844.555557, pts: 1146844.555555
15:59:34 T:123145327652864  NOTICE:  ----------- dts: 1086844.555557, pts: 1166844.555555
15:59:34 T:123145327652864  NOTICE:  ----------- dts: 1106844.555557, pts: 1106844.555557

popcornmix:
Seems to vary with h264 - this has 25 packets per second:
http://solidrun.maltegrosse.de/~fritsch/1080i50_h264_mbaff.mp4

This has 50 packets per second:
http://solidrun.maltegrosse.de/~fritsch/1080i50_h264-2.ts

I'll look into them.

FernetMenta:
This is how vaapi does it, rather simple:
https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp#L2952

popcornmix:
Yes, that's a simple solution. Not quite right for missing frames due to drops/corruption, but I guess smooth video is not possible then. I can add something like that.

FernetMenta:
vaapi does not drop frames when deinterlacing is active. It just skips deint cycles. lost frames due to corruption is a different story that would require more code to be handled correctly. the deinterlacer would need to look at frame number if it took past and future frames into account. but I doubt that this would be worth it.

fritsch:
The linked sample is mbaf style. A special way of interlacing.

Am 18.07.2016 6:22 nachm. schrieb "Rainer Hochecker" <
notifications@github.com>:

popcornmix:
@FernetMenta is it okay with you for this to be merged now, and sort out generating the extra packets and timestamps in a follow up? It's not quite simple enough to get that done this merge window.

The code has been in Milhouse builds for a few days and isn't causing any blatant issues for users.

FernetMenta:
+1

FernetMenta:
@popcornmix I was not my intention to delay this PR. I was just trying to give you all the information how player works.

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/MMALRenderer.cpp
```diff
@@ -449,25 +461,96 @@ void CMMALRenderer::Process()
...
+      MMAL_STATUS_T status = MMAL_SUCCESS;
       m_error -= 1.0;
-      MMAL_BUFFER_HEADER_T *buffer = mmal_queue_get(m_queue);
+      MMAL_BUFFER_HEADER_T *buffer = mmal_queue_get(m_queue_render);
       if (buffer == &m_quitpacket)
         bStop = true;
       else if (buffer)
-        mmal_port_send_buffer(m_vout_input, buffer);
+      {
+        CMMALBuffer *omvb = (CMMALBuffer *)buffer->user_data;
```
Jalle19: `reinterpret_cast` perhaps?

FernetMenta: reinterpret_cast is not required when casting from a void pointer

popcornmix:
@FernetMenta appreciated. Will try to bring behaviour closer to desired with a subsequent update.

popcornmix:
jenkins build this please

popcornmix:
Windows has an unrelated ffmpeg failure.

PR #10123, opened 2016-07-16
[xbmc][libs] Cleanup work that started with running as Win10 app

Paxxi:
UWP(project centennial) does not allow AddDllDirectory which
we don't really need anyway but it's use made the folder structure
pretty. Since we run perfectly fine without it both native and
as UWP I decided to drop it completely.

To make the build, installer and debugging experience easier this
led to plenty of changes to packages and build scripts to output
all dlls into /system/ instead of different folders.

While moving stuff around some dependencies got updated, it's possible
that we can use java menus in bluray movies now, not tested.

EasyHook just got repackaged to fit the new folder structure.

The rebuilding of dependencies led to us being able to drop old
vcredist files which also simplifies installer and packaging, especially
for UWP.

Some minor cleanup along the way as well


MartijnKaijser:
jenkins build this please

MartijnKaijser:
@paxxi this needs some fixes in testsuite

Paxxi:
@martijn yeah, might need some fixing for cmake as well but currently outside so sometime later today

xbmc/win32/Win32DelayedDllLoad.cpp
```diff
@@ -80,8 +80,8 @@ FARPROC WINAPI delayHookFailureFunc (unsigned dliNotify, PDelayLoadInfo pdli)
 #if !defined(DELAYIMP_INSECURE_WRITABLE_HOOKS)
 const
 #endif
-PfnDliHook __pfnDliNotifyHook2 = delayHookNotifyFunc;
+//PfnDliHook __pfnDliNotifyHook2 = delayHookNotifyFunc;
```
MartijnKaijser: leave these commented out?

Paxxi:
jenkins build this please

MediaBrasil:
@Paxxi, after commit, now we have same dlls in 2 places, in root together with .exe and in the folder system, please check.

PR #10126, opened 2016-07-17
[listprovider] respect select action setting

tamland:
Makes the select action setting work on home screen and other dynamic lists.


tamland:
jenkins build this please

PR #10127, opened 2016-07-17
OpenGL: cleanup vsync

FernetMenta:
fixes false positive log entry: "GL: Vertical Blank Syncing unsupported"


FernetMenta:
jenkins build this please

MartijnKaijser:
you can ignore the failure in testsuite (which is enable by default from now on). It's being fixed in different PR

PR #10129, opened 2016-07-17
CGUIWindowFullScreen: don't call SetDisplayAfterSeek on pause

wiromare:
Fix for the SeekLabel on pause in Estuary after the PR #10103

before the recent changes in CGUIWindowFullScreen::FrameMove() g_infoManager.SetDisplayAfterSeek() was never called on pause 

instead now with this different behavior Player.DisplayAfterSeek return true on pause and it is shown the label "Seeking" instead of "Paused"

maybe this could create some issue even with other skins

before:
err

after:
paused

@FernetMenta, @phil65

### DIFF
--- a/xbmc/video/windows/GUIWindowFullScreen.cpp
+++ b/xbmc/video/windows/GUIWindowFullScreen.cpp
@@ -361,7 +361,7 @@ EVENT_RESULT CGUIWindowFullScreen::OnMouseEvent(const CPoint &point, const CMous
 
 void CGUIWindowFullScreen::FrameMove()
 {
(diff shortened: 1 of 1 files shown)


FernetMenta:
the label "seeking" while paused is a skin issue and needs to be fixed there.

FernetMenta:
the label "seeking" while paused is a skin issue and needs to be fixed there.

wiromare:
ok even I was unsure where fix this, so why I also called phil65 just to point out the small issue, the only doubt was why Player.DisplayAfterSeek return true even on pause

closed

PR #10130, opened 2016-07-17
[cmake] Change the way how shared libraries are built

fetzerch:
This is the main prerequisites to make the VS and Xcode projects more usable. I haven't found a better way than getting rid of the static libs per directory and build kodi in one go. My plan is to change `core_add_library` for Xcode and VS to use `target_sources` and add all sources to the main executable here. In order to allow this `core_add_library` must not be used for anything else (for example for libraries).
I've put some info in http://jira.kodi.tv/browse/KODI-16651.

The main change is cbafd20. Even though it contains a few things that are independent, most of the things are somehow connected, so I put it in one commit for simplicity. @hudokkow, @notspiff: Could you review this please?

Tested for Linux, OSX (GNU Makefiles) and Android (GNU Makefiles). More testing needed for Windows (NMake and VS), OSX (Xcode) and Ninja generators.

The PR includes also a few minor changes and improvements. d0befd2 is certainly not the best way to do it, because the addon build system is also CMake based. I've mainly done it now for testing so that I can locally build osx and android packages that contain a few binary addons to see if the libraries still load fine after installation.


lib/UnrarXLib/CMakeLists.txt
```diff
@@ -41,12 +41,13 @@ set(SOURCES archive.cpp
             unpack.cpp
             volume.cpp)
 
-core_add_library(unrarxlib)
-
-if(CORE_SYSTEM_NAME STREQUAL windows)
+add_library(unrarxlib STATIC ${SOURCES} ${HEADERS})
+if(NOT CORE_SYSTEM_NAME STREQUAL windows)
```
Paxxi: isn't it easier to use `if(NOT WIN32)` ?

wsnipex: we prefer to use our own CORE_SYSTEM_NAME instead of the cmake builtins here.

wsnipex:
Very nice. Makes the whole buildsys easier to understand and maintain as well.

hudokkow:
@fetzerch, please pick https://github.com/hudokkow/xbmc/commit/4e458a46029d6dc86ee46257fcc231ca3cedb0e2

Rest looks OK. Great work!

fetzerch:
Thanks! I've squashed @hudokkow's fix and `make binary-addons ADDONS=visualization.spectrum` now works on linux as well. Will test the missing generators now.

fetzerch:
Tested on all platforms now. Windows works as well, due to @Paxxi's wdt/depends fixes. If there are no objections, I'll merge it tomorrow.

@hudokkow: as discussed, I've removed the generator expressions for the libs again because we cannot get the relative directories at configure time. That also makes the android bundle target a bit easier again.

hudokkow:
Merge today or window is closed...

PR #10131, opened 2016-07-17
[test][fix] Fix running of test suite on win32

Paxxi:
Forgot to update the path here to reflect the new dll location


Paxxi:
jenkins build this please

PR #10132, opened 2016-07-17
Improve const correctness and add missing override

neo1973:
This makes some of the more regular used interfaces of `xbmc/cores/` const-correct. I also added the override keyword at many places which, excuse my french, silences a shitload of warnings ;-)

### DIFF
--- a/xbmc/addons/AudioDecoder.cpp
+++ b/xbmc/addons/AudioDecoder.cpp
@@ -132,7 +132,7 @@ int CAudioDecoder::GetTrackCount(const std::string& strPath)
   return result;
 }
 
-CAEChannelInfo CAudioDecoder::GetChannelInfo()
+CAEChannelInfo CAudioDecoder::GetChannelInfo() const
 {
   return m_format.m_channelLayout;
 }
--- a/xbmc/addons/AudioDecoder.h
+++ b/xbmc/addons/AudioDecoder.h
@@ -60,16 +60,16 @@ namespace ADDON
 
     // Things that MUST be supplied by the child classes
     bool Init(const CFileItem& file, unsigned int filecache) override;
-    int ReadPCM(uint8_t* buffer, int size, int* actualsize);
-    bool Seek(int64_t time);
-    bool CanInit() { return true; }
-    void DeInit();
+    int ReadPCM(uint8_t* buffer, int size, int* actualsize) override;
+    bool Seek(int64_t time) override;
+    bool CanInit() const override { return true; }
+    void DeInit() override;
     void Destroy();
     bool Load(const std::string& strFileName,
               MUSIC_INFO::CMusicInfoTag& tag,
-              MUSIC_INFO::EmbeddedArt *art = NULL);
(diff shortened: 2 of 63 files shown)


FernetMenta:
Although most of this is correct there is limited benefit. The downside of those batch cleanups is that is ruins history. Currently core is heavily under construction and I rank history higher than those type of clean-ups.

garbear:
I think it would look cleanest if you dropped the component-derived prefix, and add a tag that stands out, e.g. `[improvement]`.

```
08af733 [improvement] Make interface const-correct
6a5120c player: further cleanup of playing state
a06df2b player: drop unneded method ControlsVolume
6fcee60 IPlayer: properly sync playspeed with application
ef6a155 Add missing vector includes
b1e8208 IPlayer: Remove unused function
353b0bc VideoPlayer: drop method AfterRender
b37c856 VideoPlayer: simplify video rednering
bfc4818 VideoPlayer: drop obsolete method GetRenderResolution
c436454 VideoPlayer: drop obsolte methods
1fae954 VideoPlayer: drop streamInfo
445441a VideoPlayer: revert 20e6f03 and fix possible segfault
20e6f03 VideoPlayer: drop HasMenu from interface
```

The clutter is minimized thanks to @FernetMenta's videoplayer labels, so I think these improvements would be beneficial.

neo1973:
@garbear Do you mean squashing all my commits together into one?

FernetMenta:
Code cosmetics in core won't get merged for the reason I gave.

garbear:
@neo1973 please reopen with only the const fixes in the public interfaces (AE.h, IPlayer.h and ICodec.h). Const-incorrectness in these pollutes the rest of the codebase with unneeded const_cast()s. I've been upstreaming const fixes on an individual basis when it affects my code, so it'd be nice if someone took care of this all at once.

FernetMenta:
changes in the public interfaces require adaption of the subclasses and this won't happen at this time. don't waste you time on this.

MartijnKaijser:
@FernetMenta perhaps you can give @neo1973 some pointers on some refactors or changes that would be acceptable ?

FernetMenta:
@garbear please point me to one of those unneeded const_sasts you did reference. AE and Player are comprised of active objects that communicate via message passing. In such a context const methods make not much sense because a value may get requested by a message or a value just gets read from a cache.

@neo1973 contribution is highly appreciated. I prefer changes that focus on behaviour. We have a backlog of bugs and feature requests. If you want to try your skills there, please let me know. I am happy to assist.

garbear:
@FernetMenta I haven't found any unneeded const_casts in the code in the last few years, but when I started RetroPlayer in 2012 there were several, and I added several more to RetroPlayer that I've been getting rid of over the last 4 years. My *final* const_cast is now gone with the const fix in #9826, so i'm happy with or without this PR :)

FernetMenta:
@garbear you know that the const keyword behind the method name means that the method does not change the instance's state, don't you? so these changes are meaningless for the outside world.

PR #10135, opened 2016-07-18
[addons] reinstate multidir repositories. partially reverts eec2e0aba…

tamland:
…cac1a32fd5a42fafc90336237f8bcf7


tamland:
jenkins build this please

MartijnKaijser:
@tamland forgot to thank you for bringing this back :)
If the mutiple addon versions in addons.xml could act as replacement we should certainly promote it and deprecate this over time.

PR #10137, opened 2016-07-18
fixed, plug some http exploits with a zebra list check

razzeee:
Picked from MrMC thx to @davilla 

Would be nice to get some feedback from @Montellese

### DIFF
--- a/xbmc/network/WebServer.cpp
+++ b/xbmc/network/WebServer.cpp
@@ -43,6 +43,7 @@
 #include "URL.h"
 #include "Util.h"
 #include "utils/Base64.h"
+#include "utils/FileUtils.h"
 #include "utils/log.h"
 #include "utils/Mime.h"
 #include "utils/StringUtils.h"
@@ -764,6 +765,10 @@ int CWebServer::CreateFileDownloadResponse(const std::shared_ptr<IHTTPRequestHan
 
   std::shared_ptr<XFILE::CFile> file = std::make_shared<XFILE::CFile>();
   std::string filePath = handler->GetResponseFile();
+
+  // white/black list access check
+  if (!CFileUtils::ZebraListAccessCheck(filePath))
+    return SendErrorResponse(request.connection, MHD_HTTP_NOT_FOUND, request.method);
 
   if (!file->Open(filePath, XFILE::READ_NO_CACHE))
   {
--- a/xbmc/utils/FileUtils.cpp
+++ b/xbmc/utils/FileUtils.cpp
@@ -35,6 +35,9 @@
 #include "URL.h"
 #include "settings/Settings.h"
 #include "utils/Variant.h"
+#if defined(TARGET_DARWIN)
+  #include "platform/darwin/DarwinUtils.h"
+#endif
 
 using namespace XFILE;
 
@@ -225,4 +228,49 @@ CDateTime CFileUtils::GetModificationDate(const std::string& strFileNameAndPath,
     CLog::Log(LOGERROR, "%s unable to extract modification date for file (%s)", __FUNCTION__, strFileNameAndPath.c_str());
   }
   return dateAdded;
+}
+
+bool CFileUtils::ZebraListAccessCheck(const std::string &filePath)
+{
+  // white/black list access checks, disallow exploits
+
+  // no access to the passwords.xml file,
+  // this can expose user/pass of remote servers
+  if (filePath.find("passwords.xml") != std::string::npos)
+  {
+    CLog::Log(LOGDEBUG,"http access denied");
+    return false;
+  }
+
+  // no access to the sources.xml file,
+  // this can expose user/pass of remote servers
+  if (filePath.find("sources.xml") != std::string::npos)
+  {
+    CLog::Log(LOGDEBUG,"http access denied");
+    return false;
+  }
+
+#if defined(TARGET_DARWIN)
+  char *fullpath = realpath(filePath.c_str(), nullptr);
+  if (fullpath)
+  {
+    const std::string testpath = fullpath;
+    free(fullpath);
+
+    // if this is a real path and accesses into user home, allow.
+    std::string userHome = CDarwinUtils::GetUserHomeDirectory();
+    if (testpath.find(userHome) != std::string::npos)
+      return true;
+
+    // if this is a real path and accesses outside app, deny.
+    std::string appRoot = CDarwinUtils::GetOSAppRootFolder();
+    if (testpath.find(appRoot) == std::string::npos)
+    {
+      CLog::Log(LOGDEBUG,"http access denied");
+      return false;
+    }
+  }
+#endif
+
+  return true;
 }
--- a/xbmc/utils/FileUtils.cpp
+++ b/xbmc/utils/FileUtils.cpp
@@ -234,20 +234,23 @@ bool CFileUtils::ZebraListAccessCheck(const std::string &filePath)
 {
   // white/black list access checks, disallow exploits
 
-  // no access to the passwords.xml file,
+  // no access to these files,
   // this can expose user/pass of remote servers
-  if (filePath.find("passwords.xml") != std::string::npos)
-  {
-    CLog::Log(LOGDEBUG,"http access denied");
-    return false;
-  }
+  static const char *blacklist_files[] = {
+    "passwords.xml",
+    "sources.xml",
+    "guisettings.xml",
+    "advancedsettings.xml",
+    NULL
+  };
 
-  // no access to the sources.xml file,
-  // this can expose user/pass of remote servers
-  if (filePath.find("sources.xml") != std::string::npos)
+  for (const char **ptr = blacklist_files; *ptr; ++ptr)
   {
-    CLog::Log(LOGDEBUG,"http access denied");
-    return false;
+    if (filePath.find(*ptr) != std::string::npos)
+    {
+      CLog::Log(LOGDEBUG,"http access denied");
+      return false;
+    }
   }
 
 #if defined(TARGET_DARWIN)
--- a/xbmc/utils/FileUtils.cpp
+++ b/xbmc/utils/FileUtils.cpp
@@ -239,14 +239,14 @@ bool CFileUtils::ZebraListAccessCheck(const std::string &filePath)
     "sources.xml",
     "guisettings.xml",
     "advancedsettings.xml",
-    NULL
+    nullptr
   };
 
-  for (const char **ptr = blacklist_files; *ptr; ++ptr)
+  for (auto ptr = blacklist_files; *ptr; ++ptr)
   {
     if (filePath.find(*ptr) != std::string::npos)
     {
-      CLog::Log(LOGDEBUG,"http access denied");
+      CLog::Log(LOGDEBUG, "%s access denied to %s", __FUNCTION__, filePath.c_str());
       return false;
     }
   }
@@ -268,7 +268,7 @@ bool CFileUtils::ZebraListAccessCheck(const std::string &filePath)
     free(fullpath);
 
     // if this is a real path and accesses into user home, allow.
-    std::string userHome = CSpecialProtocol::TranslatePath("special://home");
+    auto userHome = CSpecialProtocol::TranslatePath("special://home");
     if (testpath.find(userHome) != std::string::npos)
       return true;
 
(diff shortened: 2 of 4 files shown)


xbmc/utils/FileUtils.cpp
```diff
@@ -226,3 +229,48 @@ CDateTime CFileUtils::GetModificationDate(const std::string& strFileNameAndPath,
...
+
+#if defined(TARGET_DARWIN)
+  char *fullpath = realpath(filePath.c_str(), nullptr);
+  if (fullpath)
+  {
+    const std::string testpath = fullpath;
+    free(fullpath);
+
+    // if this is a real path and accesses into user home, allow.
+    std::string userHome = CDarwinUtils::GetUserHomeDirectory();
```
MrMC: replace

std::string userHome = CDarwinUtils::GetUserHomeDirectory();

with

std::string userHome = CSpecialProtocol::TranslatePath("special://home");

and you can remove the drawin ifdefs, also needs

#include "filesystem/SpecialProtocol.h"

MrMC: realpath is the key here, it will collapse any usage of "../../../" and hex encodings into a realpath.

xbmc/utils/FileUtils.cpp
```diff
@@ -226,3 +229,48 @@ CDateTime CFileUtils::GetModificationDate(const std::string& strFileNameAndPath,
...
+    if (testpath.find(userHome) != std::string::npos)
+      return true;
+
+    // if this is a real path and accesses outside app, deny.
+    std::string appRoot = CDarwinUtils::GetOSAppRootFolder();
```
Razzeee: What about this darwin specific then?

xbmc/utils/FileUtils.cpp
Razzeee: I'm not sure how to get realpath/fullpath

xbmc/utils/FileUtils.cpp
MrMC: std::string appRoot;
    CUtil::GetHomePath(appRoot);

another change I forgot about :)

xbmc/utils/FileUtils.cpp
MrMC: realpath() should exist on all posix based platforms, for windows platform, not a clue. Maybe something like the GetFullPathName function, best pray to google as usage of realpath() is pretty common for collapsing any usage of "../../../" and hex encodings into a realpath and for converting relative paths to a full path.

Razzeee:
jenkins build this please

Razzeee:
Tests on osx doesn't seem to like this change :/
http://jenkins.kodi.tv/view/Helpers/job/TestMulti-All/

MrMC:
yep, they fail because the zebra list does not permit access outside certain paths, It would have to be expanded to permit the test path.

xbmc/utils/FileUtils.cpp
```diff
@@ -226,3 +227,61 @@ CDateTime CFileUtils::GetModificationDate(const std::string& strFileNameAndPath,
...
+
+
+  char *fullpath;
+#if defined(TARGET_POSIX)
+  fullpath = realpath(filePath.c_str(), nullptr);
+#endif
+
+#if !defined(TARGET_POSIX)
+  char *namePtr;
+  GetFullPathName(filePath.c_str(), sizeof(fullpath), fullpath, &namePtr);
```
Tolriq: Seems like you forget to initialize fullpath before the call ;) You need to pass a valid char[]

Tolriq:
As tried to explain on the "locked" thread this is not the correct approach :(

This goes against all the previous work to allow remote access to allowed sources. 

The correct way is to extends CFileUtils::RemoteAccessAllowed and eventually add missing calls to that if there's some.

This patch will kills Kore file download, negge video server and many things.

See https://github.com/xbmc/xbmc/pull/2764 for some history about that.

MrMC:
I can craft an url that will never even hit CFileUtils::RemoteAccessAllowed :)

Bottom line, there are some places that the external world should not be allowed to see. This patch puts a stops to that nonsense. External apps just should not be allowed unrestricted and transparent access to internal user/pass items nor the underlying OS file system.

Tolriq:
@MrMC yes there's flaws, yes something needs to be done, as I said some calls to that function are missing. And the black list needs to be done there.

Currently this patch breaks all the previously done work to allow access to allowed sources.

Breaking Kore download function + negge video server + many many tools, because you do not take those in account.

Take time to read #2764 please as there's a long history and this patch will really have large unwanted side effects.

MrMC:
Yea, I read it. However I'm a bit more concerned with huge gaping exploit hole that exists in the current code base. Closing that ASAP out weighs past sins. Some other PR can get submitted to refactor things to fit other usages and restore what other apps need. But some thought needs to be given to this, for example, user/pass info is private to kodi. Other apps should never have unrestricted access to any user/pass info.

Tolriq:
Then add a call to it inside zebra or add the check if the file is inside sources with allowed access to true directly in zebra.

Personally Yatse have fallback to UPnP so I do not care at all, but this patch in current form break dozens of applications affecting millions of users, Kore / Video Server are applications from Team Kodi.

Since this security thing have started before my leaving, I'll continue to answer about history and details of why this must not be removed if information is needed / wanted. But merge in current status specially in beta state will leads to hacks of hacks to quick hacks to maybe restore functionality.
Since the breaking is known, either it's merged like that and removal of function for all users should be announced, but what would be the point of merging something that you know will need update to be fixed in hurry.

TBH I do not care that there's another function that does the same thing but not the same with some crossing all over if this is correct for you as I no more care about anything.

(And of course non posix needs quick fix but the crash will be quickly seen see CUtil::GetHomePath for proper usage at least on Windows)

> Closing that ASAP out weighs past sins
This is known for at least 6 month according to your MrMc patch, proper fix could have been done.

As a last note, I do agree at 3000% with the security problem and everything you say about that, but not on a quick bad fix that will break millions of users.

MrMC:
v17 is still in alpha, the purpose of alpha releases is to discover and fix these type of issues. Some big change is made, it breaks all sorts of things, the breakage gets fixed. If this was late beta or rc, it would be a different story. In this case, the breakage is already known and it is up to someone to PR as fix for it. Given that no one has cared to fix up this PR to handle such cases, I'd say that no one cares enough yet.

Ultimately, it is up to kodi devs to merge or not. I've done my part.

xbmc/network/WebServer.cpp
```diff
@@ -765,6 +766,10 @@ int CWebServer::CreateFileDownloadResponse(const std::shared_ptr<IHTTPRequestHan
   std::shared_ptr<XFILE::CFile> file = std::make_shared<XFILE::CFile>();
   std::string filePath = handler->GetResponseFile();
 
+  // white/black list access check
+  if (!CFileUtils::ZebraListAccessCheck(filePath))
+    return SendErrorResponse(request.connection, MHD_HTTP_NOT_FOUND, request.method);
```
Montellese: Why `MHD_HTTP_NOT_FOUND` (404) and not `MHD_HTTP_FORBIDDEN` (403) as the file exists (or we wouldn't make it until here)?

xbmc/utils/FileUtils.cpp
```diff
@@ -226,3 +227,61 @@ CDateTime CFileUtils::GetModificationDate(const std::string& strFileNameAndPath,
...
+    "guisettings.xml",
+    "advancedsettings.xml",
+    NULL
+  };
+
+  for (const char **ptr = blacklist_files; *ptr; ++ptr)
+  {
+    if (filePath.find(*ptr) != std::string::npos)
+    {
+      CLog::Log(LOGDEBUG,"http access denied");
```
Montellese: Space after the comma. And I would make the log message more generic as this method is not really HTTP specific (at least its name does not indicate that).

xbmc/utils/FileUtils.cpp
```diff
@@ -226,3 +227,61 @@ CDateTime CFileUtils::GetModificationDate(const std::string& strFileNameAndPath,
   }
   return dateAdded;
 }
+
+bool CFileUtils::ZebraListAccessCheck(const std::string &filePath)
```
Montellese: The first part of this method's implementation looks very much C-style (NULL terminated array, char pointers etc.). Could we C++1x-ify it a bit to match the rest of our code?

xbmc/utils/FileUtils.cpp
```diff
@@ -226,3 +227,61 @@ CDateTime CFileUtils::GetModificationDate(const std::string& strFileNameAndPath,
...
+    }
+  }
+
+
+  char *fullpath;
+#if defined(TARGET_POSIX)
+  fullpath = realpath(filePath.c_str(), nullptr);
+#endif
+
+#if !defined(TARGET_POSIX)
```
Montellese: The `#endif` and the `#if !defined(TARGET_POSIX)` could be combined into an `#else`. But I'm not sure if `GetFullPathName()` is available on anything else but win32?

xbmc/network/WebServer.cpp
MrMC: MHD_HTTP_FORBIDDEN implies a hit which gives a hint that it is there,
MHD_HTTP_NOT_FOUND gives no info about the presence or absence.
I figured, this is a security issues, why provide additional info. Could be either, dealers choice.

xbmc/utils/FileUtils.cpp
Razzeee: Well it did build fine on jenkins. With the exception of osx 64 I think.

xbmc/utils/FileUtils.cpp
Razzeee: Not really sure how to address this, I added some C++ 1x sugar now.

xbmc/utils/FileUtils.cpp
Razzeee: Added

xbmc/network/WebServer.cpp
Paxxi: I'm with @MrMC here, 403 could leak info to a would be attacker that the file is there so a 404 is safer in that regard.

xbmc/utils/FileUtils.cpp
Paxxi: for starters std::begin and std::end are supported for plain arrays so using a `for const auto ptr: blacklist_files)` should work fine or `for (auto ptr = std::begin(blacklist_files); ptr != std::end(blacklist_files); ++ptr)` last I checked stupid android didn't support cbegin and cend so not sure about those.

could use std::array with std::string like

xbmc/network/WebServer.cpp
Montellese: I'm fine with that reasoning but it would be nice to add a comment explaining it.

xbmc/utils/FileUtils.cpp
Paxxi: this should be
```
auto len = GetFullPathName(filePath.c_str(), 0, nullptr, nullptr);
if (len > 0)
{
  fullpath = reinterpret_cast<char *>(malloc(len));
  len = GetFullPathName(filePath.c_str(), len, fullpath, &namePtr);
  if (len != 0) //blow up, something went badly wrong
}
```

xbmc/utils/FileUtils.cpp
Paxxi: I think this should be `#if defined(TARGET_WINDOWS)` to not make any assumptions about platforms that may get added in the future and may blow up

xbmc/utils/FileUtils.cpp
```diff
@@ -226,3 +227,61 @@ CDateTime CFileUtils::GetModificationDate(const std::string& strFileNameAndPath,
...
+    free(fullpath);
+
+    // if this is a real path and accesses into user home, allow.
+    auto userHome = CSpecialProtocol::TranslatePath("special://home");
+    if (testpath.find(userHome) != std::string::npos)
+      return true;
+
+    // if this is a real path and accesses outside app, deny.
+    std::string appRoot;
+    CUtil::GetHomePath(appRoot);
```
Paxxi: this one has been changed and should be `auto appRoot = CUtil::GetHomePath();`

xbmc/network/WebServer.cpp
Paxxi: Agreed

Tolriq: Since I'm still notified for that PR.

Currently https://github.com/xbmc/xbmc/blob/master/xbmc/network/httprequesthandler/HTTPVfsHandler.cpp#L84 returns 401 for sources / files that are not allowed for remote access.

This is what is used by most remotes to help the user configuring correctly it's sources when having a problem.

Not talking again about the fact that this patch removes access to allowed sources (@Montellese in case you missed that side effect), but if everything returns 404, it may worth more than a comment in code but some wider communication for all users to be aware of the change.

And if this is merged and access to sources is added back later but still returns 404 then maybe extends the JSON getSources to publish the allowsharing so that there's enough information to help users.

MrMC:
maybe time to dust this off now -> http://seclists.org/fulldisclosure/2017/Feb/27

PR #10139, opened 2016-07-19
[PeripheralCecAdapter] fix broken standby after …

mk01:
6a5120c3789a7ad93a235467a4cb421dee643438

(cherry picked from commit adf8a26efe7538bd4cddb4016b6b9f2d03498a69)

### DIFF
--- a/xbmc/peripherals/devices/PeripheralCecAdapter.cpp
+++ b/xbmc/peripherals/devices/PeripheralCecAdapter.cpp
@@ -175,7 +175,7 @@ void CPeripheralCecAdapter::Announce(AnnouncementFlag flag, const char *sender, 
   else if (flag == GUI && !strcmp(sender, "xbmc") && !strcmp(message, "OnScreensaverActivated") && m_bIsReady)
   {
     // Don't put devices to standby if application is currently playing
-    if ((!g_application.m_pPlayer->IsPlaying() && !g_application.m_pPlayer->IsPaused()) && m_configuration.bPowerOffScreensaver == 1)
+    if (!g_application.m_pPlayer->IsPlaying() && m_configuration.bPowerOffScreensaver == 1)
     {
       // only power off when we're the active source
       if (m_cecAdapter->IsLibCECActiveSource())


mk01:
@FernetMenta @opdenkamp 

FernetMenta can you please review and merge ? CEC devices going standby functionality got broken with https://github.com/xbmc/xbmc/commit/6a5120c3789a7ad93a235467a4cb421dee643438

xbmc/peripherals/devices/PeripheralCecAdapter.cpp
```diff
@@ -175,7 +175,7 @@ void CPeripheralCecAdapter::Announce(AnnouncementFlag flag, const char *sender,
   else if (flag == GUI && !strcmp(sender, "xbmc") && !strcmp(message, "OnScreensaverActivated") && m_bIsReady)
   {
     // Don't put devices to standby if application is currently playing
-    if ((!g_application.m_pPlayer->IsPlaying() && !g_application.m_pPlayer->IsPaused()) && m_configuration.bPowerOffScreensaver == 1)
+    if (!g_application.m_pPlayer->IsPausedPlayback() && m_configuration.bPowerOffScreensaver == 1)
```
FernetMenta: this seems wrong.

FernetMenta: you would send a system that is playing a video into standby

opdenkamp: Sorry, haven't checked recent internal API changes. it should not send the computer to standby if video is "active" (playing or paused) because users don't expect that to happen. It should only do this when the system is actually idle.

mk01: @FernetMenta 

how you get "OnScreensaverActivated" if video is playing ?

mk01: @opdenkamp 

that is not completely true as statement. it is valid only until CECAdapter is active CEC device.

opdenkamp: ofcourse, don't change the rest

FernetMenta: > how you get "OnScreensaverActivated" if video playback is active (playing) and is not paused?

that is just an assumption. the code has to reflect what is written in the comment:
if ((!g_application.m_pPlayer->IsPlaying() && !g_application.m_pPlayer->IsPlaybackPaused()) ....

opdenkamp: so basically all that changed is that `IsPlaying()` stayed the same, but `IsPaused()` got changed to `IsPlaybackPaused()`? What does this new `IsPlaybackPaused()` do different from `IsPaused()` which is apparently still present or this wouldn't have compiled?

Don't we have an alternative in our beloved CApp that tells whether the player is active? Because that's all what it's supposed to do, don't put the TV to standby while the player is active.

FernetMenta: IsPlaybackPaused() is not new
This code here only worked by chance because IsPaused has ever been undefined in case of playback not being active.

> Don't we have an alternative in our beloved CApp that tells whether the player is active?

yes, we do:
g_application.m_pPlayer->IsPlaying()

IsPlaying does not give any information on whether playback is paused.

opdenkamp: `g_application.m_pPlayer->IsPlaying()` should be fine for this then

mk01:
@FernetMenta @opdenkamp 

changed accordingly

FernetMenta:
looks good
jenkins build this please

FernetMenta:
This change is what @opdenkamp proposed. I assume he does agree to merge.

PR #10141, opened 2016-07-19
set tvshow thumb on *all seasons node

BigNoid:
Can't find when this was lost, but this was default behaviour in previous versions. Without this PR it shows the default folder image.

Not sure how to fix and maybe unrelated but there's an issue when selecting `manage` in the context menu on this node.
It doesn't open the manage section where there should be an entry `Choose art`.

### DIFF
--- a/xbmc/video/windows/VideoFileItemListModifier.cpp
+++ b/xbmc/video/windows/VideoFileItemListModifier.cpp
@@ -77,6 +77,7 @@ void CVideoFileItemListModifier::AddQueuingFolder(CFileItemList& items)
     pItem.reset(new CFileItem(strLabel));  // "All Seasons"
     videoUrl.AppendPath("-1/");
     pItem->SetPath(videoUrl.ToString());
+    pItem->SetArt("thumb", items.GetArt("thumb"));
     // set the number of watched and unwatched items accordingly
     int watched = 0;
     int unwatched = 0;


axbmcuser:
I've been missing that. Works here. Thanks.

BigNoid:
jenkins build this please

MilhouseVH:
With this PR, I'm not seeing any "season all" posters, or tvshow fanart at all.

I now see this in Estuary:

Imgur

Notice the fanart is also missing.

Without this PR (ie. only this PR reverted) I see the following:

Imgur

ie. both season-all poster and tvshow fanart are present.

My season-all poster and fanart are already cached in the texture cache:
```
017717|8/87ff64f0.jpg|0000|0000|0002|2015-07-10 03:40:42|2015-07-10 04:40:29|nfs://192.168.0.3/mnt/share/media/Video/TVShows/Marvel's Daredevil/season-all-poster.jpg
017440|f/fbb112cd.jpg|1080|1920|0032|2016-07-19 16:33:46|2016-07-29 01:17:15|nfs://192.168.0.3/mnt/share/media/Video/TVShows/Marvel's Daredevil/fanart.jpg
```

These are the items in my tvshow folder:

Imgur

I'm seeing identical behaviour on LibreELEC and Ubuntu. Reverting just this PR restores `season-all-poster.jpg` poster and tvshow fanart visibility. I'm not seeing any errors logged to the debug log.

To be honest, I'm not entirely sure what problem this PR is trying to solve, as (for me) `season-all` posters were being displayed correctly without any issues, but now they (and fanart) are not being displayed.

BigNoid:
I see that ```ListItem.Art(fanart)``` is indeed now missing. Will revert.

axbmcuser:
The problem which remains is the following - as far as i can see:

- The "All seasons" entry can't access the path to the seasons-all poster from withihn "itemlayout" or "focusedlayout" in the list itself. The working estuary listview @MilhouseVH posted has the poster outside of the list, which is the reason why it works.
Once you switch to a view type which uses the poster inside "itemlayout" or "focusedlayout" (using a "panel" control for example), the season-all poster is not shown.

BigNoid:
This problem needs a proper fix, hence why I reverted it. It has to do with the parent folder item, once that is disabled all is working as it should.

axbmcuser:
I did not mean to disagree with reverting. It was just the answer to @MilhouseVH asking 
"I'm not entirely sure what problem this PR is trying to solve" :)

MilhouseVH:
@axbmcuser: as BigNoid says, you also experience different behaviour (without this PR) when enabling "Show Parent folder items":

Parent folder ENABLED:

Imgur

Parent folder DISABLED:

Imgur

Since I normally have the Parent folder disabled I didn't see a problem that needed solving! But thanks for explaining.

PR #10142, opened 2016-07-19
Fix retrieve all the information before update

phate89:
I assumed with update everything was handled as movies (update only if changes occurs) but tv shows doesn't. For now I'll restore the previous behavior (retrieve everything).. I'll try to improve it when I find time..


phate89:
jenkins build this please

MilhouseVH:
Fix looks good - thanks.

PR #10143, opened 2016-07-19
[depends][xbmc][fix] Get rid of wdk requirement

Paxxi:
Bump sdk to Win 10 as cmake defaults to that for addons anyway. Add the few headers we need from wdk as a package to simplify setup

### DIFF
--- a/lib/addons/library.kodi.adsp/project/VS2010Express/libKODI_adsp.vcxproj
+++ b/lib/addons/library.kodi.adsp/project/VS2010Express/libKODI_adsp.vcxproj
@@ -14,6 +14,7 @@
     <ProjectGuid>{44F93C4D-85DD-4452-99BB-F1D196174024}</ProjectGuid>
     <RootNamespace>KODI_ADSP</RootNamespace>
     <Keyword>Win32Proj</Keyword>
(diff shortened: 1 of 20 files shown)


afedchin:
is there any benefits from sdk10? and why cmake uses it as default?

Paxxi:
Not for us, no. I assume cmake just picks the newest, or runs vcvars.bat which defaults to newest. There's also no downside to using it. I was a bit confused when doing the 2015 upgrade and thought it's best to target the lowest possible. In hindsight this has only caused trouble with having to install an extra sdk.

afedchin:
thanks for clarifications

Paxxi:
jenkins build this please

garbear:
I still need WDK 8.1 installed to compile Kodi on Windows 7. The headers in mini_wdk-win32.7z seem to use structs and typedefs that aren't available on windows 7

Paxxi:
If you have sdk 10.x installed it should be fine

garbear:
Can you make a note of this on the wiki?

Paxxi:
I updated the wiki when this went in I think or maybe it was for another change but it does say to install sdk 10

garbear:
lol you mean the big image that takes up the entire screen? my powers of observation are incredible :)

Paxxi:
haha yeah that one :)

PR #10144, opened 2016-07-19
mmalrender: Allow a frame of slop when waiting for vsync

popcornmix:
Previously we would always wait an extra vsync (even if one had been missed) in RenderUpdate.
Now if a single vsync is missed, we won't wait.


popcornmix:
Ping @FernetMenta 
This is based on comment: https://github.com/xbmc/xbmc/pull/10096#issuecomment-232654359

FernetMenta:
+1

popcornmix:
jenkins build this please

PR #9120, opened 2016-02-12
[win32] Update msys build system + move libdvd to depends

afedchin:
This is update of our mingw build system to modern msys2.

pros
- ffmpeg dlls depends only on msvcrt.dll
- posibility to build ffmpeg in debug with pdb via msvc compiler (without gnutls and dcadec).
- possibility to build x64 version of libs.

To build ffmpeg in debug just run `make-mingwlibs.bat msvc`

@FernetMenta @Paxxi ping

PR #10145, opened 2016-07-20
[depends] Readd option to build without dvdcss after #9120

Rechi:
Maybe cmake needs also to be adopted.

### DIFF
--- a/tools/depends/target/libdvdread/Makefile
+++ b/tools/depends/target/libdvdread/Makefile
@@ -16,14 +16,21 @@ else
   ifeq ($(PLATFORM),)
     PLATFORM = native
     TARBALLS_LOCATION = $(ROOT_DIR)
-    PKGCONFIGPATH = PKG_CONFIG_PATH="$(ROOT_DIR)/../libdvdcss/dvdcss-install/lib/pkgconfig"
-    EXTRA_CFLAGS += -I$(ROOT_DIR)/../libdvdcss/dvdcss-install/include
+    ifneq ($(BUILD_DVDCSS),0)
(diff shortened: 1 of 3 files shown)


tools/depends/target/libdvdread/Makefile
```diff
@@ -16,14 +16,21 @@ else
   ifeq ($(PLATFORM),)
     PLATFORM = native
     TARBALLS_LOCATION = $(ROOT_DIR)
-    PKGCONFIGPATH = PKG_CONFIG_PATH="$(ROOT_DIR)/../libdvdcss/dvdcss-install/lib/pkgconfig"
-    EXTRA_CFLAGS += -I$(ROOT_DIR)/../libdvdcss/dvdcss-install/include
+    ifneq ($(BUILD_DVDCSS),0)
```
wsnipex: this disables DVDCSS support globally for every platform that builds with depends as BUILD_DVDCSS is not defined in depends.

wsnipex:
I removed this stuff, because it is pretty difficult to get right for all platforms and imho not worth the hassle.
Since the only reason to build without dvdcss is licensing/copyright, distros are free to patch it out.

tools/depends/target/libdvdread/Makefile
Rechi: I tested again on osx and linux and if `BUILD_DVDCSS`is not set, it will build with with libdvdcss.

FernetMenta:
I agree with @wsnipex 
We don't build without dvdcss, hence no reason for the extra code.

MartijnKaijser:
Guess we can one this then

PR #10146, opened 2016-07-20
[estuary] Show fallback behind thumb in addoninfo preview list

BigNoid:
ping @tamland 

Follow up for #10060


tamland:
great. thanks!

PR #10147, opened 2016-07-20
Add python functions to set and get ratings

phate89:
@tamland there aren't better ways to check boolean in dictionary since it's all converted to string right?

### DIFF
--- a/xbmc/interfaces/legacy/ListItem.h
+++ b/xbmc/interfaces/legacy/ListItem.h
@@ -291,6 +291,42 @@ namespace XBMCAddon
 #ifdef DOXYGEN_SHOULD_USE_THIS
       ///
       /// \ingroup python_xbmcgui_listitem
+      /// @brief \python_func{ setRating(type, rating, votes = 0, default = True) }
+      ///-----------------------------------------------------------------------
+      /// Sets a listitem's rating. It needs at least type and rating param
+      ///
+      /// @param type       string - the type of the rating. Any string.
+      /// @param rating     float - the value of the rating.
+      /// @param votes      int - the number of votes. Default 0.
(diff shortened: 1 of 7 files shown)


xbmc/interfaces/legacy/ListItem.h
```diff
@@ -291,6 +291,40 @@ namespace XBMCAddon
...
+      ///
+      ///
+      ///
+      ///-----------------------------------------------------------------------
+      ///
+      /// **Example:**
+      /// ~~~~~~~~~~~~~{.py}
+      /// ...
+      /// # setRating(values)
+      /// self.list.getSelectedItem().setRating({ 'default':True, 'rating':9.5, 'votes':4}, 'imdb')
```
tamland: Documentation and implementation says different things. If there's only 3 possible keys though, im not sure it makes sense to have a dictionary for that. Just have them as regular parameters.

phate89: I implemented with dict because (AFAIK for what I tried) we can't use optional fields.. default and votes are optional...

tamland: it should work. see https://github.com/phate89/xbmc/blob/2c84b648a36c27127b356effd53499c5e30728b5/xbmc/interfaces/legacy/ModuleXbmcplugin.h#L80 for example

phate89:
@tamland You are right of course... I don't know what I did wrong the first time.. I updated the pr with a function with 4 args

tamland:
Do you want this in? Should be fine imo since it's just a simple getter and setter

phate89:
Better to add it I guess..
Jenkins build this please

PR #10148, opened 2016-07-20
Remove pointless check in python setUniqueIDs (and fix description)

phate89:
I'm not sure why I added it in the first place.. If there is no videoinfotag we create one..


phate89:
jenkins build this please

phate89:
jenkins build this please

PR #10151, opened 2016-07-21
AESinkPULSE: Disable remixing if no server downmix is needed

fritsch:
@btbn @notspiff what do you think.

~~This is a workaround for PA by default remixing and therefore up and downmixing. I acquire the real device's channel map and tell AE to mute (send silence) the channels after AE does an ResolveChannels internally.~~

~~That way 5.1 will output as 5.1 on a 6 channel PA configuration, but Stereo will also output without upmixing as channel != FR or channel != FL will get silence by AE.~~

Just disable remixing if server has enough channels for us

### DIFF
--- a/xbmc/cores/AudioEngine/Sinks/AESinkPULSE.cpp
+++ b/xbmc/cores/AudioEngine/Sinks/AESinkPULSE.cpp
@@ -267,6 +267,7 @@ struct SinkInfoStruct
   bool device_found;
   pa_threaded_mainloop *mainloop;
   int samplerate;
+  pa_channel_map map;
   SinkInfoStruct()
   {
     list = NULL;
@@ -274,6 +275,7 @@ struct SinkInfoStruct
     device_found = true;
     mainloop = NULL;
     samplerate = 0;
+    pa_channel_map_init(&map);
   }
 };
 
@@ -290,6 +292,7 @@ static void SinkInfoCallback(pa_context *c, const pa_sink_info *i, int eol, void
 
     sinkStruct->samplerate = i->sample_spec.rate;
     sinkStruct->device_found = true;
+    sinkStruct->map = i->channel_map;
   }
   pa_threaded_mainloop_signal(sinkStruct->mainloop, 0);
 }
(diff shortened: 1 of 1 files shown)


FernetMenta:
This strategy looks weird. I doubt that any other desktop application behaves like that.

Jalle19:
Personally I think this upmixing that Pulseaudio is doing is a huge anti-feature. In fact it's the sole reason why I force Kodi to use ALSA even though other apps use Pulseaudio. Hacking around this just gives credibility to their bad decisions.

fritsch:
We have an alternative. I will explain in a minute.

Am 22.07.2016 5:27 nachm. schrieb "Sam Stenvall" <notifications@github.com>:

fritsch:
@Jalle19 can you test? This will disable PA's remixing to "related channels". We only enable it whenever we open e.g. 8 channels but PA server only has 2 to get it properly downmixed for the user.

fritsch:
Test it, works nicely. If one selects a special device nothing is altered by PA. When using the default device it does its normal job.

fritsch:
jenkins build this please

Jalle19:
@fritsch can't test for a few days, don't have the right setup here. So what does this patch actually do now? Will stereo content be sent as stereo to the receiver even when 5.1 channels is selected in Pulseaudio's settings?

fritsch:
If you have pulseaudio setup 5.1 - there is no way that it signals 2.0 to the AVR as it keeps a 5.1 out, no matter what you do. But with the patch above we will send 2.0 channels only and tell PA to not touch it -> Stereo is not mangled.

If a user chooses 7.1 channels in kodi but his hardware only has 2.0 channels, this REMIX is not turned off to let pulse do its job.

Also the above only works if you don't use the "default pulse device" as this remixes / mangles the way it wants.

In short: choose your HDMI out or your Analog out and the number of channels kodi sends will appear on your AVR - the rest of the channels are muted by pulseaudio.

fritsch:
jenkins build this please

fritsch:
jenkins looks fine. I consider this as a FIX - if issues arise I will care for them as maintainer of Pulse Sink.

PR #9542, opened 2016-04-03
VideoPlayer: drop WaitPresentTime + simplify render loop

FernetMenta:
this only causes issues on most systems. not needed anymore

@afedchin @popcornmix /cc

PR #10153, opened 2016-07-22
Codecinfo, Drain and ResetDisplay fixes

razzeee:
Big thanks to @koying

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.cpp
@@ -490,9 +490,6 @@ bool CDVDVideoCodecAndroidMediaCodec::Open(CDVDStreamInfo &hints, CDVDCodecOptio
       return false;
   }
 
-  if (m_render_surface)
-    m_formatname += "(S)";
-
   // CJNIMediaCodec::createDecoderByXXX doesn't handle errors nicely,
   // it crashes if the codec isn't found. This is fixed in latest AOSP,
   // but not in current 4.1 devices. So 1st search for a matching codec, then create it.
@@ -595,6 +592,11 @@ bool CDVDVideoCodecAndroidMediaCodec::Open(CDVDStreamInfo &hints, CDVDCodecOptio
 
   m_opened = true;
   memset(&m_demux_pkt, 0, sizeof(m_demux_pkt));
+  
+  m_processInfo.SetVideoDecoderName(m_formatname, true );
+  m_processInfo.SetVideoPixelFormat(m_render_surface ? "Surface" : (m_render_sw ? "YUV" : "EGL"));
+  m_processInfo.SetVideoDimensions(m_hints.width, m_hints.height);
+  m_processInfo.SetVideoDeintMethod("hardware");
 
   return m_opened;
 }
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.cpp
@@ -309,7 +309,7 @@ void CDVDMediaCodecInfo::RenderUpdate(const CRect &SrcRect, const CRect &DestRec
   {
     CXBMCApp::get()->setVideoViewSurfaceRect(DestRect.x1, DestRect.y1, DestRect.x2, DestRect.y2);
     cur_rect = DestRect;
-    
+
     // setVideoViewSurfaceRect is async, so skip rendering this frame
     ReleaseOutputBuffer(false);
   }
@@ -355,6 +355,7 @@ bool CDVDVideoCodecAndroidMediaCodec::Open(CDVDStreamInfo &hints, CDVDCodecOptio
 
   m_render_surface = CSettings::GetInstance().GetBool(CSettings::SETTING_VIDEOPLAYER_USEMEDIACODECSURFACE);
   m_drop = false;
+  m_codecControlFlags = 0;
   m_hints = hints;
 
   switch(m_hints.codec)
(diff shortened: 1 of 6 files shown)


xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.cpp
```diff
@@ -846,6 +848,14 @@ void CDVDVideoCodecAndroidMediaCodec::SetDropState(bool bDrop)
     m_videobuffer.iFlags &= ~DVP_FLAG_DROPPED;
 }
 
+void CDVDVideoCodecAndroidMediaCodec::SetCodecControl(int flags)
+{
+  if (m_codecControlFlags != flags)
+    if (g_advancedSettings.CanLogComponent(LOGVIDEO))
```
Jalle19: Double if, could be simplified.

MrMC:
Looks good as is. Nice tweaks koying.

koying:
Mmm... The DVD_CODEC_CTRL_DRAIN part is not ready ;)

popcornmix:
I believe the rule is when DVD_CODEC_CTRL_DRAIN is set is to return VC_PICTURE (without VC_BUFFER) until all buffered pictures from codec are returned, then return VC_BUFFER.

xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.cpp
Razzeee: @koying 
maybe you fooled yourself with this? Is the first if missing backets by any chance?

koying: That's a c/p from rpi, but yeah, not great now that I look at it, but working @popcornmix

Razzeee:
Arm build can be found here http://mirrors.kodi.tv/test-builds/android/arm/kodi-20160722-03100f3-feed-info-armeabi-v7a.apk

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodec.cpp
```diff
@@ -116,8 +116,8 @@ CRenderInfo CRendererMediaCodec::GetRenderInfo()
 {
   CRenderInfo info;
   info.formats = m_formats;
-  info.max_buffer_size = NUM_BUFFERS;
-  info.optimal_buffer_size = 2;
+  info.max_buffer_size = 4;
```
Jalle19: Magic numbers?

koying: Those are hardware buffers locked by Kodi, so the lowest, the best.
3 seems an actual minimum for proper working, nowadays, and "max" doesn't seem to be actually used, but 6 is too much.

FernetMenta: not setting optimal_buffer_size results in actual buffers used = 2, which is far from ideal. should be at least 4 or skipping won't work properly

koying: > not setting optimal_buffer_size results in actual buffers used = 2

Actually, I understand from https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp#L284 that its the lowest value between the codec requirements, the renderer requirements (optimal and max) and "6".
If nothing is implemented codec and hw-rendrer wise, the value would be 4, from CLinuxRendererGLES, but the absolute minimum is indeed "2".

De facto, "2" causes problems after a refresh rate change, but is fine otherwise. That might point out to some deadlocked buffers in the render loop after a refresh change, though, so a different root cause.

koying: I'd add that max_buffer_size has no purpose, thus, IMO, unless you accept that optimal might be > max?

FernetMenta: RenderManager won't skip a frame immediately if it detects a lateness of less than frametime. In order to have frame skipping work properly, i.e. play some 60fps content on a 30Hz screen, you should have at least 4 buffers.

FernetMenta: > I'd add that max_buffer_size has no purpose, thus, IMO, unless you accept that optimal might be > max?

right, @popcornmix do you remember the purpose of having max_buffer_size?

koying: > RenderManager won't skip a frame immediately if it detects a lateness of less than frametime

Mmm... Call me thick, but I don't see immediately in what having an additional buffer would influence that...
Or do you need at least fps/Hz*2 buffers? What about 60fps on 24Hz, then...

FernetMenta: https://github.com/xbmc/xbmc/pull/9542#issuecomment-207771279

I remember that skipping was not smooth anymore and a/v desync was to high. 

> Or do you need at least fps/Hz*2 buffers? 

I don't think so. iirc I tested with TV set to 24/23.975 Hz.

MrMC:
an area rip for documentation...

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodec.cpp
Razzeee: Any problem if I leave it as it is and we adjust it afterwards, if needed.

FernetMenta: go ahead, it is already an improvement.

Razzeee:
let's see how happy jenkins is

jenkins build this please

PR #10155, opened 2016-07-22
[xbmc][fix] Fix on demand dll loading to point to the correct files

Paxxi:
I had obviously missed these causing some issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -577,3 +577,4 @@ lib/addons/library.kodi.inputstream/project/VS2010Express/Release
 tools/depends/native/TexturePacker/src/Win32/Debug/
 tools/depends/native/TexturePacker/src/Win32/Release/
 tools/depends/native/TexturePacker/src/Win32/TexturePacker.VC.db
+exclude_dll.txt
--- a/project/Win32BuildSetup/BuildSetup.bat
+++ b/project/Win32BuildSetup/BuildSetup.bat
@@ -237,6 +237,9 @@ set WORKSPACE=%CD%\..\..
   rem Exclude skins as they're copied by their own script
   Echo addons\skin.estuary\>>exclude.txt
   Echo addons\skin.estouchy\>>exclude.txt
+
+  rem Exclude dlls from system to avoid duplicates
+  Echo .dll>>exclude_dll.txt
   
   md BUILD_WIN32\application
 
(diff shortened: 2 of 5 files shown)


MartijnKaijser:
Installer needs adjustment as well regarding copying of files?

Paxxi:
it should already copy all files into the root or did you change that?

MartijnKaijser:
Looks good so can be merged i guess (no jenkins build needed)

Tolriq:
@Paxxi don't know if caused by this one, but when debugging on Windows when building to a custom path, cpluff.dll is tried to be loaded at the wrong place. 

Loading all the rest works and load dlls from the main kodi directory but cpluff.dll seems to be loaded from portable_directory/..

Paxxi:
@Tolriq could certainly be caused by this, I haven't noticed any issue but then again I don't think I've tested with `-p` so there might be some weirdness there.

Will investigate and see.

Tolriq:
@Paxxi ok no in fact it's the xbmcbin globally that point to KODI_HOME even when forcing a working folder.

Migrating KODI_HOME to the working dir solve that issue but seems to cause a lot of others :(

Issue is a mix between portable mode / working directory in debug settings and KODI_HOME.

Maybe this have changed on another PR and there's just need to have some updates of Wiki to reflect how to handle that ?

Paxxi:
There's been a lot of volatility in this area the last week as one major change led to many things being overlooked and issues popping up all over the place.

Wiki is not updated as no changes are meant to be required by someone opening up the solution and trying to debug. `special://xbmcbin` is now meant to point to the root folder/working directory as that's where all the dlls end up after running the installer.

To get this working with the dlls in the system folder we play some tricks and add system to the path for dll resolution during debugging. This fails for some files currently it seems and I'll get it sorted sometime this week.

As a workaround just copy cpluff.dll into the root folder and it should work.

Tolriq:
Ok thanks, that's what I ended up doing. (All the ones from DllPaths_win32.h as curl is also needed and in my case nfs too).

Just for information since maybe I do things wrong and -p is not the solution, but how are we supposed to manage multiple debugging Kodi profiles for different Kodi versions ? (Like when working on different branches to update previous PR / backport)

Paxxi:
`-p` is perfectly fine for that or script some simple switching of userdata dirs.

Voyager1:
Why don't we simply put all dll's in the xbmc folder instead of xbmc/system, just like it is in the installation target? Having a different structure between development and (install) release is just begging for problems.

MartijnKaijser:
@Voyager1 that likely also means repackaging every dependency again.

PR #10156, opened 2016-07-23
[addons] fix event handling

tamland:
Should hopefully fix all the various updating issues stemming from the lack of a event strategy in the addon system.
This includes a very basic pub/sub utility for sending events; it gets the job done but if anyone know something better, existing, that can be used, free to suggest one.

### DIFF
--- a/xbmc/listproviders/DirectoryProvider.cpp
+++ b/xbmc/listproviders/DirectoryProvider.cpp
@@ -36,6 +36,7 @@
 #include "settings/Settings.h"
 #include "threads/SingleLock.h"
 #include "utils/JobManager.h"
+#include "utils/log.h"
 #include "utils/SortUtils.h"
 #include "utils/URIUtils.h"
 #include "utils/Variant.h"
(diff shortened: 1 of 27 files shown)


tamland:
can someone provide xcode/vs for this?

xbmc/listproviders/DirectoryProvider.cpp
```diff
@@ -253,6 +260,19 @@ void CDirectoryProvider::Fetch(std::vector<CGUIListItemPtr> &items) const
   }
 }
 
+void CDirectoryProvider::OnEvent(const ADDON::AddonEvent& event)
+{
+  CSingleLock lock(m_section);
```
wsnipex: couldn't the lock be moved down inside the if?

tamland: m_currentUrl needs to be protected as well, since it may be written to

ksooo:
@tamland you can cherry-pick the xcode project update for this PR here https://github.com/ksooo/xbmc/commit/1315dc065439d05f623637a5f8f31f1b852caf17 if you like.

tamland:
thanks @ksooo!

tamland:
jenkins build this please

MartijnKaijser:
Did quick test and seems to work fine now with installing Addons and updating vfs

da-anda:
sorry, haven't checked the changes - will this also update the add-on lists in terms of "recently used" on home screen? IIRC this is also not working atm (at least last time I checked).
Also, at least for me PVR channels and file shares in "videos" node not always show up, but that's a different issue.

MartijnKaijser:
@da-anda yep for add-on question :)

MartijnKaijser:
Can we include this for alpha3?

tamland:
yes. just had to squash.

BigNoid:
This PR causes some issues. Easiest to reproduce is starting a video and do a ```ReloadSkin()```. Without this PR it works fine, with this PR it crashes.

PR #10157, opened 2016-07-23
Double the backslash in mysql in case we're in 'LIKE'

phate89:
@DaveTBlake spotted a bug in our current mysqldataset. The bug is more a wrong implementation because of a specific case.
Currently db values are passed through PrepareSQL to get the parameters formatted well (for example strings will have backslash doubled and special chars formatted).
Sqlite has a specific function for it but mysql doesn't so in 2010 (!!!) firnsy (not sure if it was a member) ported the sqlite function to mysql (https://github.com/xbmc/xbmc-antiquated/commit/a2ae89c0e79237d4e86ac42bf4ea08c0cc780be5 https://sourceforge.net/p/xbmc/mailman/message/25665578/ )
The main problem is that mysql has a "unique feature" where in like statements ( for eample `SELECT * FROM movie_view WHERE strPath LIKE 'C:\\Film\\%'`) doubling backslash isn't enough because it parse the string twice (explained in the note here http://dev.mysql.com/doc/refman/5.7/en/string-comparison-functions.html#operator_like ). SQlite doesn't work like this so this isn't handled. for example
this query works in both mysql and sqlite:

```
`SELECT * FROM movie_view WHERE strPath='C:\\Film\\Batman\\'
```

but this didn't work in mysql (work of course in sqlite)

```
`SELECT * FROM movie_view WHERE strPath LIKE 'C:\\Film\\Batman\\%'
```

because you need

```
`SELECT * FROM movie_view WHERE strPath LIKE 'C:\\\\Film\\\\Batman\\\\%'
```

This of course breaks every playlist with windows local paths because they all have backslash..
So we need to double backslash but when we have a like we need to quadruple it.
I tried to insert in this current code a check to detect if we're parsing a like parameter.. I'm not sure it's okay to do it like this but I can't think to another way.. Every suggestion is welcome..
@MilhouseVH not sure if it's worth it to add it to your builds.. I tested it quickly and it seems to work but it might break a lot of other things..

### DIFF
--- a/xbmc/dbwrappers/mysqldataset.cpp
+++ b/xbmc/dbwrappers/mysqldataset.cpp
@@ -809,12 +809,13 @@ void MysqlDatabase::mysqlVXPrintf(
   length = 0;
   bufpt = 0;
   for(; (c=(*fmt))!=0; ++fmt){
+    bool isLike = false;
     if( c!='%' ){
       int amt;
       bufpt = (char *)fmt;
       amt = 1;
       while( (c=(*++fmt))!='%' && c!=0 ) amt++;
-      mysqlStrAccumAppend(pAccum, bufpt, amt);
+      isLike = mysqlStrAccumAppend(pAccum, bufpt, amt);
       if( c==0 ) break;
     }
     if( (c=(*++fmt))==0 ){
@@ -1178,7 +1179,11 @@ void MysqlDatabase::mysqlVXPrintf(
         int needQuote;
         char ch;
         char q = ((xtype==etSQLESCAPE3)?'"':'\'');   /* Quote character */
-        const char *escarg = va_arg(ap, char*);
+        std::string arg = va_arg(ap, char*);
+        if (isLike)
+          StringUtils::Replace(arg, "\\", "\\\\");
+        const char *escarg = arg.c_str();
+
         isnull = escarg==0;
         if( isnull ) escarg = (xtype==etSQLESCAPE2 ? "NULL" : "(NULL)");
         k = precision;
@@ -1241,15 +1246,15 @@ void MysqlDatabase::mysqlVXPrintf(
 /*
 ** Append N bytes of text from z to the StrAccum object.
 */
-void MysqlDatabase::mysqlStrAccumAppend(StrAccum *p, const char *z, int N) {
+bool MysqlDatabase::mysqlStrAccumAppend(StrAccum *p, const char *z, int N) {
   if( p->tooBig | p->mallocFailed ){
-    return;
+    return false;
   }
   if( N<0 ){
     N = strlen(z);
   }
   if( N==0 || z==0 ){
-    return;
+    return false;
   }
   if( p->nChar+N >= p->nAlloc ){
     char *zNew;
@@ -1258,7 +1263,7 @@ void MysqlDatabase::mysqlStrAccumAppend(StrAccum *p, const char *z, int N) {
     if( szNew > p->mxAlloc ){
       mysqlStrAccumReset(p);
       p->tooBig = 1;
-      return;
+      return false;
     }else{
       p->nAlloc = szNew;
     }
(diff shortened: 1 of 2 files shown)


wsnipex:
I didn't look into this issue in detail and I'm probably naive, but doesn't https://dev.mysql.com/doc/refman/5.7/en/mysql-real-escape-string.html or https://dev.mysql.com/doc/refman/5.7/en/string-functions.html#function_quote work?

phate89:
it works and we use it to escape the string (it's used after I doubled the backslash) but in like is a different story. that function doesn't know if we're in a like or not so it just performs a regular escape (Adding backslash before every character needed). Like needs a regular escape PLUS doubled backslash...

wsnipex:
SELECT * FROM movie_view WHERE strPath LIKE QUOTE('C:\\Film\\Batman\\%')?

phate89:
I tried and it doesn't work.. Also if it worked it could create problems with sqlite.. anyway thanks fo trying to find a simpler solution.. this should be our last resort..

DaveTBlake:
Thanks for doing this @phate89 . I know it does not feel like an elegant solution, but I think it is a necessary one. At the moment on MySQL any queries  using LIKE with backslash in the parameter just don't work, which includes any query on path in a Windows system.

+1

Tolriq:
@phate89 not sure how parsing is done in Mysql but I faced a more or like same problem in another project.

Things to tests : 
- What happens if there's "like" string in any parameter / field name like a query with xx = like
- Check if there's case where the parameter is passed through a function like xxx like UPPERCASE(xxx%)

tamland:
What I'm wondering about is why do you have windows paths in you db in the first place.. Shouldn't the vfs handle this?

DaveTBlake:
Not sure how you expect the vfs to handle this @tamland ?
In music database the path table has rows for every folder and sub folder that contains music files, the physical location of the music. This is used by library update to check what files have been changed and need rescanning etc., to deduce the location of NFO files and thumbnails for albums and artists, and as a filter parameter for smart playlists e.g. list only songs in this path. The library needs to know where the music files are.

phate89:
@tamland no if you simply add a local path on Windows you get the windows regular path..

tamland:
Correct me if im wrong (since i dont use window), but afaik the vfs can already translate the platform independent urls to the real windows paths. Not that it in any way helps with the immediate problem but maybe it's time to fix the root cause: leaking platform details into the database. It really doesn't need to know about the path separator. And that would take care of a lot of different issues as well..

phate89:
There's something like that if I remember well..but everything right now is stored with paths (including Linux paths, sources and arts) so changing it it's a huge change and I'm not sure we can change everything safely. Plus it's incorrect it's a path awareness problem.. it's a mysql like problem with \ symbol. It happens because we use like on paths but if we used for something else that required like and could contain a backslash you would have the same problem

DaveTBlake:
Yeap replacing path would be a huge change, and it really is a MySQL problem with backslashes generically that this PR is trying to address. I would also worry about the speed impact on the library update checking too.

phate89:
@DaveTBlake  @MilhouseVH @tamland suggestion on how to go further? Just throw it in? wait for v18?

wsnipex:
anyone tried http://dev.mysql.com/doc/refman/5.6/en/sql-mode.html#sqlmode_no_backslash_escapes or changing the escape char to something different from "/" http://dev.mysql.com/doc/refman/5.6/en/string-comparison-functions.html

phate89:
@wsnipex I tried both.. The first doesn't work because it disables escapes in regular strings but you still need to escape them in like (so you still have to locate if the string is in a like). the second is a specific like option. it needs to be added after like and if there's no like the query fails so you still need to parse the query to determine if it's a like (and sqlite doesn't have it so we can't just use it)..

wsnipex:
Ok, seems there is no better solution then this PR for now. Time for a DB abstraction layer..

DaveTBlake:
I would say merge it soonest. We don't have a better fix, and can't find anything wrong with this, but get it into the nightlies for real world testing.

MartijnKaijser:
jenkins build and merge

MartijnKaijser:
jenkins build this please

MartijnKaijser:
Fails to build
http://jenkins.kodi.tv/job/BuildMulti-All/1557/

phate89:
jenkins build this please

phate89:
jenkins build and merge

PR #10158, opened 2016-07-23
[cmake] Improve generated Visual Studio and Xcode projects

fetzerch:
This PR changes the CMake build system so that usable Visual Studio and Xcode projects are generated.

The initial project files created one static library per subdirectory in xbmc/ (similar to our Autotools project).
While this has some advantages (compile options and include dirs can be set on a directory level, link time is faster), it doesn't play well with the generated Visual Studio and Xcode projects. These treat all subdirectories as separate subprojects and the IDEs project browsers weren't close to the filesystem anymore and rather unusable. Additionally the subprojects caused much longer compile times especially with Visual Studio.

Screenshots can be found in Jira: http://jira.kodi.tv/browse/KODI-16651

Pings for code review: @wsnipex, @notspiff, @hudokkow 
Would be good to get some feedback from Xcode users: @FernetMenta, @Memphiz

Jenkins build: http://jenkins.kodi.tv/view/Helpers/job/BuildMulti-All/1503/
Packages should have been uploaded in case platform devs want to verify that they are still working.


FernetMenta:
@fetzerch any significant changes to the version I tested last week? Are you asking for general feedback or do you want to have this particular version tested?

fetzerch:
@FernetMenta you mentioned some build errors when only building the kodi scheme. That should work now. Other than that, just some minor improvements.

In general I'm interested in anything that prevents you from using it

project/cmake/scripts/common/Macros.cmake
```diff
@@ -176,6 +177,20 @@ function(core_add_shared_library name)
   endif()
 endfunction()
 
+# Sets the compile language for all C source files in a target to CXX.
+# Needs to be called from the CMakeLists.txt that defines the target.
+# This method needs to be called from the CMakeLists.txt that adds the target!
```
hudokkow: A bit redundant? Or shall we all be scared? :smile:

hudokkow:
Quick browse and code looks good but I can't test Xcode stuff.
I can test VS on Monday if I get past the "install'uninstall one more thing because we're so great and it will only take 3h to download, verify, spit an error and tell you this is the wrong install/update exec. Or we won't tell you. We're M$ and we know you have psychic powers!"

project/cmake/scripts/common/Macros.cmake
fetzerch: Hehe, good catch :)

project/cmake/CMakeLists.txt
```diff
@@ -293,24 +301,43 @@ elseif(CORE_SYSTEM_NAME STREQUAL android)
...
 else()
-  add_library(${APP_NAME_LC} SHARED ${CORE_MAIN_SOURCE} $<TARGET_OBJECTS:compileinfo> "${RESOURCES}" ${OTHER_FILES})
+  add_library(${APP_NAME_LC} SHARED ${CORE_MAIN_SOURCE} "${RESOURCES}" ${OTHER_FILES})
   # The libraries depend on the main shared library for wrapping. Because of this we cannot build
   # the libraries as a dependency. This requires to build the all target on android (and not only kodi).
   add_dependencies(${APP_NAME_LC} export-files pack-skins)
 endif()
 whole_archive(_MAIN_LIBRARIES ${core_DEPENDS})
-target_link_libraries(${APP_NAME_LC} ${SYSTEM_LDFLAGS} ${_MAIN_LIBRARIES} ${DEPLIBS} ${CMAKE_DL_LIBS})
+target_link_libraries(${APP_NAME_LC} lib${APP_NAME_LC} ${_MAIN_LIBRARIES} lib${APP_NAME_LC} ${SYSTEM_LDFLAGS} ${DEPLIBS} ${CMAKE_DL_LIBS})
```
wsnipex: double lib${APP_NAME_LC}

fetzerch: thx, fixed

fetzerch:
Currently Kodi won't launch if built on windows with CMake and Visual Studio because there is a mismatch between where the dlls are downloaded/built to and where Kodi tries to load them. This got introduced by: https://github.com/xbmc/xbmc/commit/e3391adb4f74b0a13de1e76a967fd604771ad6e8

fetzerch:
Pushed more commits:
- ccc1ded, 9778be8 and ccebec9 fix the problem of kodi not starting on windows
- ccc1ded together with 40702e8 replace https://github.com/xbmc/xbmc/pull/10162.
  @Memphiz: Do you want me to exclude the commit installing certs.pem on osx/ios here? From what I can see it doesn't cause any harm to have it even if #10162 is not yet included.

Memphiz:
no need to exclude it - once i have finally some time to work on that PR i will do. The packaging of the cert will remain as is - so would be good to have it already in cmake ...

fetzerch:
Build was fine except on linux (caused by PulseAudio incompatibility).
New build triggered after fixing that is green: http://jenkins.kodi.tv/job/LINUX-64/8479/console

The feedback I got so far was positive, so I'll merge it now. Further improvements can be done independently.

FernetMenta:
> The feedback I got so far was positive, so I'll merge it now

only fixes can be merged outside a merge window. please keep in mind next time

fetzerch:
Perfectly clear to me. However windows build was broken without this due to libraries being moved around. Additionally I understood from @MartijnKaijser that cmake things can go in since they make our live easier and have little to no impact on behavior. Otherwise I wouldn't have merged it.

MartijnKaijser:
This code is currently unused by our buildsystem/official builds as such it isn't really affected by merge windows. This will of course change when we start using it.

FernetMenta:
ok, thanks for the update

PR #10159, opened 2016-07-23
[cmake/linux] Move WiiRemote target to ExtraTargets.cmake

fetzerch:
@hudokkow: I forgot to move that one when rebasing my change. There shouldn't be any change needed to the cpack code, right?


hudokkow:
Nope, all good.
Thanks for the refactor/cleanup!

PR #10161, opened 2016-07-24
AESinkAudiotrack: Do not average over 4 seconds but 200 ms

fritsch:
Please someone runtime tests this.

The old 20 worked around the insane flaws of the android delay API by averaging over a very long time. If you send "periodsize" amount of data this was something between 80 ms and 200 ms in PCM mode which made us react very slowly, we needed up to 4 (1.6) seconds in worst case to adjust to audio. After I added the manual blocking (forced usleeps) this should not be needed anymore. 

This could fix Sync Playback to Display's resample ratio.


koying:
Yep, seems ok.
Actually, the previous report might indeed been due to the long time time it took to stabilize...

MartijnKaijser:
Seems to play fine with sync to display enabled

fritsch:
Thanks for your feedback. jenkins build and merge

PR #10162, opened 2016-07-24
[cmake] Allow mirroring of additional system specific files to build tree

fetzerch:
On windows and osx/ios we need to mirror additional files to the build tree. While this is usually done via configuration files in installdata, these files need special treatment (either because they need to go into a different subdir or because they are dynamically generated and their path is not static).

### DIFF
--- a/project/cmake/scripts/osx/Macros.cmake
+++ b/project/cmake/scripts/osx/Macros.cmake
@@ -105,3 +105,15 @@ function(find_soname lib)
   endif()
   set(${lib}_SONAME ${${lib}_SONAME} PARENT_SCOPE)
 endfunction()
+
+# Copies additional files to the root of the buildtree so that kodi can be
+# started from there.
+# On return:
+#   files added to ${install_data}, mirror in build tree
+function(copy_additional_files_to_buildtree)
+  copy_file_to_buildtree(${CORE_SOURCE_DIR}/tools/depends/target/openssl/cacert.pem
+                         ${CORE_SOURCE_DIR}/tools/depends/target/openssl
+                         SUBDIR system/certs)
+
+  set(install_data ${install_data} PARENT_SCOPE)
+endfunction()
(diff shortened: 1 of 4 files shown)


project/cmake/scripts/osx/Macros.cmake
```diff
@@ -105,3 +105,14 @@ function(find_soname lib)
...
   set(${lib}_SONAME ${${lib}_SONAME} PARENT_SCOPE)
 endfunction()
+
+# Copies additional files to the root of the buildtree
+# On return:
+#   files added to ${install_data}, mirror in build tree
+function(copy_additional_files_to_buildtree)
+  copy_file_to_buildtree(${CORE_SOURCE_DIR}/tools/depends/target/openssl/cacert.pem
+                         ${CORE_SOURCE_DIR}/tools/depends/target/openssl
+                         SUBDIR system/certs)
```
wsnipex: mind adding a comment that this is only necessary to run kodi from build dir. Otherwise installing them would be enough

fetzerch: Sure. Added a comment to that function and also mentioned it in the commit message.

fetzerch:
Turned out that this solution would break in-source-builds as copy_file_to_buildtree wouldn't copy anything there. I've been working on a different solution in https://github.com/fetzerch/xbmc/commit/8789261b118f11b481f0b29939e5d36f0691b466.

Instead of these kind of hooks, we use configuration via installdata files. An optional second parameter in the files specifies where to mirror these files to.

```
tools/depends/target/openssl/cacert.pem system/certs
```

We'll need a similar thing for windows, so I think it's the cleaner solution. I'll do a bit more testing on the other platforms and will create a PR then.

fetzerch:
Closing in favor of #10158 (ccc1ded).

project/cmake/scripts/osx/Macros.cmake
Memphiz: just for being complete - this is not necessary only for being able to run kodi from the build tree but it gets also bundled into the final app bundle and is needed for having working ssl support...

PR #10163, opened 2016-07-24
[depends] Install config-binaddons.site and Toolchain_binaddons.cmake

fetzerch:
Installing those two files allows to use xbmc-addons.include to build addons only with the installed depends.

That way multiple versions of depends (android-arm, android-x86, ...) can be built and installed from one kodi source tree. Afterwards the source tree can be cleaned completely.

When building with CMake, this enables to configure with `-DCMAKE_TOOLCHAIN_FILE=/opt/xbmc-depends/<platform>/share/Toolchain.cmake` and compile the binary addons afterwards with `make binary-addons ADDONS="visualization.spectrum"`.

ping: @wsnipex


wsnipex:
+1

PR #10164, opened 2016-07-24
AE: add atempo filter

FernetMenta:
This adds long desired atempo filter to AE. This filter allows audio going at different speed without changing pitch (chipmunk voices)

This is in particular important for live streams. In order to prevent buffers from running dry VideoPlayer slows down clock speed. Currently that does not work very well. A change of 2% already results in noticeable change of pitch and audio may still stall because the change is not big enough.

### DIFF
--- a/project/VS2010Express/XBMC.vcxproj.filters
+++ b/project/VS2010Express/XBMC.vcxproj.filters
@@ -2845,6 +2845,9 @@
     <ClCompile Include="..\..\xbmc\dialogs\GUIDialogPlayerProcessInfo.cpp">
       <Filter>dialogs</Filter>
     </ClCompile>
+    <ClCompile Include="..\..\xbmc\cores\AudioEngine\Engines\ActiveAE\ActiveAEFilter.cpp">
+      <Filter>cores\AudioEngine\Engines\ActiveAE</Filter>
+    </ClCompile>
   </ItemGroup>
   <ItemGroup>
     <ClCompile Include="..\..\xbmc\media\MediaType.cpp">
@@ -6706,6 +6709,9 @@
     </ClInclude>
     <ClInclude Include="..\..\xbmc\dialogs\GUIDialogPlayerProcessInfo.h">
       <Filter>dialogs</Filter>
+    </ClCompile>
+    <ClInclude Include="..\..\xbmc\cores\AudioEngine\Engines\ActiveAE\ActiveAEFilter.h">
+      <Filter>cores\AudioEngine\Engines\ActiveAE</Filter>
     </ClInclude>
   </ItemGroup>
   <ItemGroup>
--- a/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEFilter.cpp
+++ b/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEFilter.cpp
@@ -1,0 +1,332 @@
+/*
+ *      Copyright (C) 2010-2016 Team Kodi
+ *      http://xbmc.org
+ *
+ *  This Program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2, or (at your option)
+ *  any later version.
+ *
+ *  This Program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with XBMC; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "ActiveAEFilter.h"
+#include "utils/log.h"
+#include "utils/StringUtils.h"
+#include <algorithm>
+
+extern "C" {
+#include "libavfilter/avfilter.h"
+#include "libavfilter/buffersink.h"
+#include "libavfilter/buffersrc.h"
+#include "libswresample/swresample.h"
+}
+
+using namespace ActiveAE;
+
+CActiveAEFilter::CActiveAEFilter()
+{
+  m_pFilterGraph = nullptr;
+  m_pFilterCtxIn = nullptr;
+  m_pFilterCtxOut = nullptr;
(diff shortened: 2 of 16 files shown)


xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEStream.cpp
```diff
@@ -671,17 +710,34 @@ bool CActiveAEStreamBuffers::IsDrained()
 
 void CActiveAEStreamBuffers::SetRR(double rr)
 {
-  m_resampleBuffers->SetRR(rr);
+  if (rr == 1.0)
```
popcornmix: Isn't this block redundant? Looks like the next if block will just do the same calls again.

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEStream.cpp
```diff
@@ -671,17 +710,34 @@ bool CActiveAEStreamBuffers::IsDrained()
...
+    m_resampleBuffers->SetRR(1.0);
+    m_atempoBuffers->SetTempo(1.0/rr);
+  }
 }
 
 double CActiveAEStreamBuffers::GetRR()
 {
-  return m_resampleBuffers->GetRR();
+  double tempo = m_resampleBuffers->GetRR();
+  tempo *= m_atempoBuffers->GetTempo();
```
popcornmix: Should this be a ``/=``? Looks like when it is set we use the inverse of ``rr``.

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEStream.cpp
FernetMenta: yes, thanks.

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEStream.cpp
FernetMenta: right, I will correct it.

wsnipex:
cmake fix: http://paste.ubuntu.com/20743795/

FernetMenta:
@wsnipex thanks updated

fritsch:
Very nice work! Thank you very much.

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEFilter.cpp
```diff
@@ -192,7 +192,7 @@ int CActiveAEFilter::ProcessFilter(uint8_t **dst_buffer, int dst_samples, uint8_
     frame->format = m_sampleFormat;
 
     result = avcodec_fill_audio_frame(frame, channels, m_sampleFormat,
-                             src_buffer[0], channels * src_samples * av_get_bytes_per_sample(m_sampleFormat), 0);
+                             src_buffer[0], channels * src_samples * av_get_bytes_per_sample(m_sampleFormat), 1);
```
FernetMenta: @fritsch I don't think this is correct because all of our buffers have a alignment of 16

@popcornmix can you try setting it to 16 on pi?

fritsch: @fernetMenta: Nope they don't have as the m_frames of pulseaudio might be something like: 2205 and then there is a problem. So the m_frames directly influences the src_samples. In our case 2205 * 8 = 17640 which is not dividable by 32. So fill frame fails cause it wants 17664. We cannot assume the m_frames coming from a Sink is mod 32.

FernetMenta: again, all buffers are allocated with alignment of 16. This is a fact and not related to any sink.

fritsch: You compute the requested size as: channels * src_samples*av_get_sample_size.

src_samples: 2205
Channels: 2
Sample size: 8

This is put into avcodec_fill_audio_frame. Your alloced size Plays zero role here. Ffmpeg Computes this and fails, cause src_buffer needs to be at least some Bytes larger. So either the method is called with wrong src_buffer size or alignment needs to be 1 as 32 needs a mod 32 number as requested size number.

fritsch: In short: replacing 1 with 16 will work, too, but only by chance for 4 byte format... If Format is 3 Byte and two channel this method will fail, too.

FernetMenta: you are wrong here. for planar data, size has to be calculated considering line size.

fritsch: You are right. I was wrong and misinterpreted the debugger.

popcornmix: Changing the last parameter of avcodec_fill_audio_frame to 16 doesn't help the segfault on Pi.

fritsch: Can you post a backtrace? Funnily I cannot reproduce mine anymore, one of fernet's commits has most likely fixed it for me.

FernetMenta: > Changing the last parameter of avcodec_fill_audio_frame to 16 doesn't help the segfault on Pi.

this change alone is not enough for planar formats. buf_size has to be calculated taking line_size into account. I will fix this evening

FernetMenta:
jenkins build this please

MartijnKaijser:
Win32 timed out 
Jenkins build this please

project/VS2010Express/XBMC.vcxproj.filters
```diff
@@ -6706,6 +6709,9 @@
     </ClInclude>
     <ClInclude Include="..\..\xbmc\dialogs\GUIDialogPlayerProcessInfo.h">
       <Filter>dialogs</Filter>
+    </ClCompile>
```
ace20022: this should be  ```</ClInclude>```

FernetMenta: sorry, I will fix in a minute

PR #10165, opened 2016-07-24
[droid] split SDK and NDK api level. They are unrelated

MartijnKaijser:


wsnipex:
Please add this to the Droid Readme, It's a change in behavior.

PR #10166, opened 2016-07-24
Readd Player process info and PVR info dialog

ksooo:
This functionality got removed with recent VideoPlayer rewrite. This PR readds it.

Please note that skin changes are required to support the new dialog.

@FernetMenta for review?
@Razzeee could you provide VS project file update?


ksooo:
@BigNoid fyi re Estuary
@HitcherUK maybe you want to adapt Confluence to support the new dialog?

FernetMenta:
nice!
@MartijnKaijser despite the API change, we want this for v17, right?

ksooo:
The API change actually is a non-breaking API addition (a new json-rpc method) but is just sugar and I easily can remove it from the PR (that why I made it a separate commit). So, just tell me what you want.

MartijnKaijser:
Indeed since it's a non-breaking api change i'd say go for it. Up to the PVR guys really

MilhouseVH:
PR looks good to me. JSON is working (and toggling). Button mappings look fine too (and toggle).

One thing that is possible, and probably shouldn't be, is that JSON will open the PlayerProcessInfo dialog even when no video is playing, ie. at the Home screen (or even when playing music) - in both cases it shows details of the last video to have played, if there has been one (and junk if not).

The image below is a clean start of Kodi, no video played, then executing JSON `Input.ShowPlayerProcessInfo`:

Imgur

It would be better to not open the dialog when there is no (video) player active.

Razzeee:
@ksooo 
Pick this https://github.com/Razzeee/xbmc/commit/65230991c5434552cec5fb1e39aee62939f466b9
But it's not test compiled, so make sure to test it with jenkins

ksooo:
@Razzeee thx for the vs project update.
@MilhouseVH i think i found a good check to prevent the dialog popping up when player is not active.

ksooo:
jenkins build this please

ksooo:
jenkins: no idea why it is acting up on WIN32, seems unrelated to this PR, though.

Razzeee:
Just a test failing and it doesn't even know which one.

MilhouseVH:
> @MilhouseVH i think i found a good check to prevent the dialog popping up when player is not active.

Thanks, this is better - I'm not seeing anything when there is no player active and executing `Input.ShowPlayerProcessInfo`.

However there is still an issue when playing a song and using JSON `Input.ShowPlayerProcessInfo`.

This is a freshly started kodi.bin - no video played yet - while playing a music track and executing JSON `Input.ShowPlayerProcessInfo`:

Imgur

If I had played a video the dialog would be showing details of that video.

Is this PR intending to support audio playback? If not then it should only appear when the video player is active.

HitcherUK:
Can't it only be triggered if `System.IsFullscreen`?

ksooo:
The nonsense displayed for music should be fixed by another PR. It's a core issue.

I don't think the dialog should only work when in full screen, because most (all?) information shown by it is also useful if not in fullscreen.

FernetMenta:
> The nonsense displayed for music should be fixed by another PR.

+1

joethefox:
Testing the new window, I have noticed a small issue. Via JSON you can use the command "GUI.ActivateWindow" to activate a specific window. Via JSONRPC.Introspect you can obtain all the windows name/id that can be activated. The new window name "playerprocessinfo" is returned via introspect but when calling 

`{"method":"GUI.ActivateWindow", "params":{"window":"playerprocessinfo"}}`

nothing is showed and logs says unable to load DialogPlayerProcessInfo.xml from the skin folder.

BigNoid:
> nothing is showed and logs says unable to load DialogPlayerProcessInfo.xml from the skin folder.

https://github.com/xbmc/skin.estuary/pull/172 is needed for the skin part.

ksooo:
jenkins build this please

ksooo:
Looking at my end I'm done with this PR and will merge it as soon as the needed Estuary changes are merged to xbmc master. The latter will hopefully be done before release of alpha 3. Ping @BigNoid.

ksooo:
travis and jenkins errors are not related to this PR.

ksooo:
jenkins build this please

xbmc/Application.cpp
```diff
@@ -2356,6 +2356,15 @@ bool CApplication::OnAction(const CAction &action)
       g_windowManager.ActivateWindow(WINDOW_MUSIC_PLAYLIST);
     return true;
   }
+  if (action.GetID() == ACTION_PLAYER_PROCESS_INFO)
+  {
+    if ((m_pPlayer->IsPausedPlayback() || m_pPlayer->IsPlaying()) &&
```
FernetMenta: the first statement can be dropped. m_pPlayer->IsPlaying() gives no info on playback is paused or not

ksooo: Confused: which one can be dropped? m_pPlayer->IsPausedPlayback() or m_pPlayer->IsPlaying() ? How to detect whether player is active but paused?

FernetMenta: if (m_pPlayer->GetCurrentPlayer() == "VideoPlayer")

is correct because this dialog is only valid for VideoPlayer

ksooo: I understood that <code>if (m_pPlayer->GetCurrentPlayer() == "VideoPlayer")</code> return <code>false</code> in case videoplayer is inactive (does not play something, is paused, rewinds, ff, ...). Is this correct?

FernetMenta: yes

ksooo: changed.

ksooo:
jenkins build this please

MartijnKaijser:
@ksooo needs a rebase

ksooo:
rebased

ksooo:
jenkins build this please

ksooo:
rebased

ksooo:
jenkins build this please

MartijnKaijser:
ignore ios fail

ksooo:
jenkins IOS error is not related to this PR

HitcherUK:
I've updated Confluence if anyone is interested in testing.

screenshot079

HitcherUK:
How about making the old one open on long press of 'o'?

FernetMenta:
what do you consider the "old one"?

ksooo:
playerdebug I guess.

HitcherUK:
Yes, just a thought unless it's going to be removed altogether.

FernetMenta:
there are less than 10 people on this planet who can explain the parameters playerdebug shows. I see no need for a standard mapping of this action. further longpress does not work for remotes.

PR #10168, opened 2016-07-25
AESinkPULSE: Align m_frames to 32

fritsch:
Still thinking about a proper solution to directly hit the unterlaying ALSA device.


fritsch:
The server shall handle that for us. That's why we have it.

PR #8290, opened 2015-10-25
[input] add WETEK remote to peripherals

da-anda:
this has been lying around in my repo for some time. It adds the USB ID of the WETEK remote to our peripherals system which allows to ship customized keymaps for the WETEK remote while they only apply when the remote is present (just like for Nyxboard).

I'm happy to add a suiteable default keymap to this PR if somebody points me to one. @NedScott do you consider the keymap in the wiki as good?
I don't have a WETEK Core box, so was not able to add it's USB ID. If somebody can grab it, I'll add it.

PR #10169, opened 2016-07-26
Default keymaps for WeTek USB remotes

NedScott:
Successor to https://github.com/xbmc/xbmc/pull/8290

This is using updated and safe keymaps that work on all three of the current WeTek USB remotes. This has gotten really good feedback from users and from people on the LibreELEC team.

This does not currently impact using these remotes in Android, so this is linux only (LibreELEC/OpenELEC, OSMC, etc).


Jalle19:
Any objections?

PR #10171, opened 2016-07-26
More long-press enhancements

NedScott:
This adds some additional long-press actions that users seemed to find useful from other keymaps I've done. None of these should change any existing default behavior. This also adds equivalent long-press actions for remote.xml and updated keymap header, which I had forgotten to do in the previous v16 PR.

Added:

LP on [Play]/[Play_pause] - Global - will bring up Info (both for library data, add-ons, and on-screen metadata/info).
LP on [Play]/[Play_pause] - Virtual Keyboard - Enter/done. (Similar to how Amazon does it for their remotes on most data-entry screens.)
LP on [Left] - Fullscreen video/audio - Rewind
LP on [Right] - Fullscreen video/audio - FastForward
LP on [Up] - Fullscreen video - Change audio track
LP on [Down] - Fullscreen video - Change subtitle track (which can also turn subtitles off)
LP on [Select]/[Enter]/[Return] - Fullscreen audio - PlayPause (this was already enabled for fullscreen video)

### DIFF
--- a/system/keymaps/keyboard.xml
+++ b/system/keymaps/keyboard.xml
@@ -12,7 +12,7 @@
 <!-- not used in the current window's section.                                            -->
 <!--                                                                                      -->
 <!-- Actions can be built-in functions.                                                   -->
-<!--  eg <B>ActivateWindow(Music)</B>                                                   -->
+<!--  eg <B>ActivateWindow(Music)</B>                                                     -->
 <!-- would automatically go to Music on the press of the B button.                        -->
 <!--                                                                                      -->
 <!--  Long presses                                                                        -->
@@ -123,6 +123,7 @@
       <prev_track>SkipPrevious</prev_track>
       <stop>Stop</stop>
       <play_pause>PlayPause</play_pause>
+      <play_pause mod="longpress">Info</play_pause>
       <fastforward>FastForward</fastforward>
       <rewind>Rewind</rewind>
       <record/>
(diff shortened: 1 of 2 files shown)


MartijnKaijser:
i'd see no problems with this :)

MilhouseVH:
This has broken both my remotes (Ortek VRC-1100 and Harmony One emulating MCE remote) in my latest #0729 build, on both [RPi/RPi2](http://forum.kodi.tv/showthread.php?tid=269814&pid=2383680#pid2383680) and [Generic (X86)](http://forum.kodi.tv/showthread.php?tid=269815&pid=2383681#pid2383681).

I'm getting various reports of buttons not working or not working as expected, including over CEC, and also from an Xbox 360 USB user. In my case it seems that the longpress buttons are being interpreted as the normal press buttons, making the remotes now unusable.

Still trying to work out if this is specific to LibreELEC - feedback from other distributions would help narrow that down.

NedScott:
Now that I think about it, I don't recall there being any real feedback about MCE remotes and long-press. This is the first time we've had them in by default, besides long-press on "menu", which doesn't seem to actually exist on MCE remotes (maybe the DVD menu key?).

Maybe we should revert the remote.xml map for now?

NedScott:
https://github.com/xbmc/xbmc/pull/10186

MilhouseVH:
Yes, it definitely seems to be the case that longpress remote buttons are now being processed as the "normal" button, at least in LibreELEC.

PR #10172, opened 2016-07-27
[droid] - cleanup binary addon handling

ghost:
This commit enables inputstream addons to be loaded on droid devices (trying to avoid #if ANDROID).

The first commit cleans up duplicated / platform dependend code in every addon helper library,
Generation of the binary addon dll name has completely moved to libXBMC_addon.h wich is included in every binary addon. Name genaration is done by 2 makros.

The second commit introduces a new special path for binary addons, because there could (are) differences between the different platforms. For any reasons all binary addons on Android are in the folder where libkodi is also located, this differes from other platforms, where we have subfolders like /addons/library.kody.xy.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -540,7 +540,7 @@ bool CApplication::Create()
       CLog::Log(LOGNOTICE, "WARNING: unsupported ffmpeg version detected");
   }
 #endif
-  
+
   std::string cpuModel(g_cpuInfo.getCPUModel());
   if (!cpuModel.empty())
     CLog::Log(LOGNOTICE, "Host CPU: %s, %d core%s available", cpuModel.c_str(), g_cpuInfo.getCPUCount(), (g_cpuInfo.getCPUCount() == 1) ? "" : "s");
@@ -926,6 +926,11 @@ bool CApplication::InitDirectoriesLinux()
     CSpecialProtocol::SetLogPath(strTempPath);
     CreateUserDirs();
   }
+  CSpecialProtocol::SetXBMCBinAddonPath(appBinPath + "/addons");
+
+#if defined(TARGET_ANDROID)
+  CXBMCApp::InitDirectories();
+#endif
 
   return true;
 #else
@@ -1012,7 +1017,7 @@ bool CApplication::InitDirectoriesOSX()
     CSpecialProtocol::SetTempPath(strTempPath);
     CSpecialProtocol::SetLogPath(strTempPath);
   }
-
+  CSpecialProtocol::SetXBMCBinAddonPath(appPath + "/addons");
   return true;
 #else
   return false;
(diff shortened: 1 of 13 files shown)


FernetMenta:
looks ok to me
@wsnipex what do you think?

MartijnKaijser:
Gave it a quick try and pvr still works

wsnipex:
For Linux and Freebsd special://xbmcbinaddons will only point to the actual libraries, addon's data still resides in another dir.
e.g.
library: /usr/lib/kodi/addons/<addonname>
data: /usr/lib/kodi/addons/<addonname>

Kodi's paths make this difference (XBMCBinPath vs XBMCPath), so maybe we should have the same for addons to be consistent here?

mapfau:
"/addons" is concatted to the xbmcbin path in Applications.cpp.
So xbmcbinaddons is always xbmcbin + "/addons"

can you pls. look in slack?

wsnipex:
I probably was unclear above. I know that behavior is unchanged and the PR looks ok to me as well. 

What I wanted to say is that maybe sometime we want to support installing binaddons in dirs different from kodi. For this case we might need to differ between library and data locations.
Just a point for consideration/discussion, not blocking this PR in any way.

mapfau:
If I don't misunderstand again: This is exactly done here: there is a new special path now (xbmcbinaddons) that is only there for setting the binary addon path separately fronm xbmcbin.

Means, we're now completely free to define an own path only for the binary addons.

FernetMenta:
jenkins build this please

wsnipex:
no, its not.
kodi has appBinPath (binaries) and appPath (arch independent data). Those two are usually different on linux.
You only add the binary addon equivalent of appBinPath

mapfau:
Is there something to do from my side? IMO we can add an alternative path for linux at the time it is necessary, would this be OK?

FernetMenta:
if I understood @wsnipex correctly he said that this is ok for now. but there is room for future improvements.

wsnipex:
yes, thats what I meant, thanks for the translation @FernetMenta :)

MartijnKaijser:
@FernetMenta should this wait til 1 august or merge now?

mapfau:
My question was simply because I didn't got the reason why it is not merged - now I know that you have merge freeze windows. Thanks for all your statements!!

FernetMenta:
@MartijnKaijser for my understanding of our process regarding merge windows it should wait until Aug. 1st.

MartijnKaijser:
The tag confused me a bit so thought i asked :)

wsnipex:
we could argue that it is a fix(as the tad suggests as well), so it could go in anytime

MartijnKaijser:
I'm fine either way. You guys know the possible risk if any better as me :)

FernetMenta:
ok :)

mapfau:
thx

PR #10174, opened 2016-07-27
[depends][win32][fix] new shairplay.dll that actually exports methods

Paxxi:
Added shairplay.dll as a delay loaded lib to VS solution as well.
This does not require cmake changes to match currently


Paxxi:

PR #10175, opened 2016-07-27
[SpecialProtocol] Fix crash caused by envhome on Windows

fetzerch:
The path envhome is only set when TARGET_POSIX is defined. On windows this prevents Kodi from starting. Aborts in GetPath due to missing map entry.

Bug was introduced by: https://github.com/xbmc/xbmc/commit/3e2d04ffb4e52118825a1d774e9a03b4d3d8f237

### DIFF
--- a/xbmc/filesystem/SpecialProtocol.cpp
+++ b/xbmc/filesystem/SpecialProtocol.cpp
@@ -253,7 +253,9 @@ void CSpecialProtocol::LogPaths()
   CLog::Log(LOGNOTICE, "special://xbmc/ is mapped to: %s", GetPath("xbmc").c_str());
   CLog::Log(LOGNOTICE, "special://xbmcbin/ is mapped to: %s", GetPath("xbmcbin").c_str());
   CLog::Log(LOGNOTICE, "special://masterprofile/ is mapped to: %s", GetPath("masterprofile").c_str());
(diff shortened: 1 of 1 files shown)


Paxxi:
To be clear, it's an assert that triggers but very annoying :)

Tolriq:
Should not the envhome defined for all ? How could someone browsing that path know from the doc that it's only available on some OS ? Windows users have home directory too ;)

Paxxi:
@Tolriq this is used specifically for an ios function, it's not something anyone has to know about or other platforms have to care about

Tolriq:
https://github.com/xbmc/xbmc/pull/9781 Memphiz answer to my query about the assert.

Then maybe the path should only be defined for IOS and not all posix ?

Paxxi:
maybe, but that's not relevant to this PR, this is a quick fix to a minor issue.

DanTheMan827:
@Tolriq I did this for POSIX rather than just iOS because envhome literally maps to the $HOME variable

I was originally trying to do ~/ using path substitution but that introduced another set of bugs...

Namely, browsing into a ISO or archive in file manager, going into a folder inside it, and then going back didn't keep the substituted path but rather returned to the absolute path which just doesn't work with sandboxed iOS...

If someone could make it work with ~/ that would probably make the most sense

PR #10178, opened 2016-07-28
[PVR][json-rpc] Fix PVR.AddTimer for timer rules.

ksooo:
This fixes the issue mentioned here: https://github.com/xbmc/xbmc/pull/10074#issuecomment-235228038

Now, if timerrule= true is set for PVR.AddTimer, a EPG-event-based timer rule will be created. channel and epg search string are taken from the epg event, start any time and end any time will be set to true, weekdays will be set to any day.

@joethefox can you test this, please?


joethefox:
Tested and I can confirm that the timer rule is correctly created with all the defaults values filled.

ksooo:
jenkins build this please

PR #10180, opened 2016-07-28
[xbmc][improvement] Fix detection of KODI_HOME

Paxxi:
Currently we require this to be set as an environment variable which complicates generating a VS solution with cmake. This fix changes behaviour and tries to find the correct directory in 3 ways.
1. check `KODI_HOME` environment variable
2. check the directory the executable lives in (this will be the common case for users)
3. check the current working directory (this will be the common case for dev env)

This does not change the behaviour of other platforms but I did refactor the code so testing on Linux and OSX will be required to make sure nothing broke.

@Memphiz can you try it out and see?
@fetzerch or @wsnipex can you try it out on linux?

### DIFF
--- a/xbmc/Util.cpp
+++ b/xbmc/Util.cpp
@@ -239,7 +239,7 @@ std::string GetHomePath(const std::string& strTarget, std::string strPath)
 {
   if (strPath.empty())
   {
-    auto strHomePath = CUtils::ResolveExecutablePath();
+    auto strHomePath = CUtil::ResolveExecutablePath();
     size_t last_sep = strHomePath.find_last_of(PATH_SEPARATOR_CHAR);
     if (last_sep != std::string::npos)
       strPath = strHomePath.substr(0, last_sep);
--- a/xbmc/Util.cpp
+++ b/xbmc/Util.cpp
@@ -107,6 +107,167 @@ using namespace PLAYLIST;
 unsigned int CUtil::s_randomSeed = time(NULL);
 #endif
 
+namespace
+{
+#ifdef TARGET_WINDOWS
+bool IsDirectoryValidRoot(std::wstring path)
+{
+  path += L"\\system\\settings\\settings.xml";
+  auto h = CreateFileW(path.c_str(), GENERIC_READ, 0, nullptr,
+    OPEN_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr);
+  if (h != INVALID_HANDLE_VALUE)
+  {
+    CloseHandle(h);
+    return true;
+  }
+
+  return false;
+}
+
+std::string GetHomePath(const std::string& strTarget, std::string strPath)
+{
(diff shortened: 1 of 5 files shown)


Voyager1:
should the Kodi project file not be updated as well, because currently the KODI_HOME variable is set to ${SolutionDir}..\..  i.e. xbmc\project\VS2010Express\..\.. i.e. the xbmc folder. Since all DLLs are in xbmc\system, kodi crashes on startup in debug mode.

Paxxi:
Ahh yeah, only tested it with cmake, will update that as well but VS solution will be gone soon

fetzerch:
Does it make sense to mention "windows" in the commit message?

BTW: Your points 2 and 3 are the same for all single-configuration generators (GNU/NMake/Ninja Makefiles) where the executable is built directly in the root dir. They are only really different on multi-configuration generators (Visual Studio and Xcode). They allow to specify the configuration (Release/Debug) at build-time and need separate paths where they put the executables.

Fails to compile on Linux:
```
xbmc/Application.cpp: In member function ‘bool CApplication::InitDirectoriesLinux()’:
xbmc/Application.cpp:883:47: error: no matching function for call to ‘CUtil::GetHomePath(std::__cxx11::string&, const char*&)’
   CUtil::GetHomePath(appBinPath, envAppBinHome);
                                               ^
In file included from xbmc/Application.cpp:32:0:
xbmc/Util.h:63:22: note: candidate: static std::__cxx11::string CUtil::GetHomePath(std::__cxx11::string)
   static std::string GetHomePath(std::string strTarget = "KODI_HOME"); // default target is "KODI_HOME"
                      ^
xbmc/Util.h:63:22: note:   candidate expects 1 argument, 2 provided
```

xbmc/Util.cpp
```diff
@@ -107,6 +107,167 @@ using namespace PLAYLIST;
...
+  }
+  return strPath;
+}
+#endif
+#if defined(TARGET_POSIX) && !defined(TARGET_DARWIN)
+std::string GetHomePath(const std::string& strTarget, std::string strPath)
+{
+  if (strPath.empty())
+  {
+    auto strHomePath = CUtils::ResolveExecutablePath();
```
fetzerch: `CUtil::`

fetzerch:
seems to not cause regressions on osx or linux

Memphiz:
@Paxxi well the regression this introduced on osx was non obvious. Users pointed it out. Next time its better to check all source files for the call of the changed function (CUtil::GetHomePath in this case) and alter the occurrence blindly so the behavior stays as before (in this case by moving the first parameter of the call to the left side of the "=").

Paxxi:
Oh, thought I got them all, in retrospect I should have renamed the method so any mistakes would have been build errors and not silently fail

Memphiz:
Oh right - good idea too :)

PR #10182, opened 2016-07-29
[osx] fix media key control for next and previous track

Rechi:
This allows to use the media controls for previous and next track on mac.
Without this PR they would invoke f14 and f15 which have no action set in the keymaps.

### DIFF
--- a/xbmc/platform/darwin/osx/SDLMain.mm
+++ b/xbmc/platform/darwin/osx/SDLMain.mm
@@ -438,12 +438,12 @@ static void setupWindowMenu(void)
 #define VK_VOLUME_MUTE      0xAD
 #define VK_VOLUME_DOWN      0xAE
 #define VK_VOLUME_UP        0xAF
-#define VK_MEDIA_NEXT_TRACK 0xB0
-#define VK_MEDIA_PREV_TRACK 0xB1
+#define VK_MEDIA_NEXT_TRACK 0x9E
+#define VK_MEDIA_PREV_TRACK 0x9D
 #define VK_MEDIA_STOP       0xB2
 #define VK_MEDIA_PLAY_PAUSE 0xB3
-#define VK_REWIND           0x9D
-#define VK_FAST_FWD         0x9E
+#define VK_REWIND           0xB1
+#define VK_FAST_FWD         0xB0
 
 - (void)powerKeyNotification
 {


MartijnKaijser:
@Memphiz ping

Memphiz:
mhhh i wonder why those had wrong keycodes then before @davilla do you have a clue about that? (unlikely i know...) - well lets hope that its not something that only works with newer osx versions then ... jenkins build this please

MrMC:
not a clue, too far back in time.

xbmc/platform/darwin/osx/SDLMain.mm
```diff
@@ -442,8 +442,8 @@ - (void) deviceDidUnMountNotification:(NSNotification *) note
 #define VK_MEDIA_PREV_TRACK 0xB1
 #define VK_MEDIA_STOP       0xB2
 #define VK_MEDIA_PLAY_PAUSE 0xB3
-#define VK_REWIND           0x9D
-#define VK_FAST_FWD         0x9E
+#define VK_REWIND           0xB1
```
ksooo: VK_REWIND now has the same value as VK_MEDIA_PREV_TRACK (0xB1). Looks insane, no?

xbmc/platform/darwin/osx/SDLMain.mm
```diff
@@ -442,8 +442,8 @@ - (void) deviceDidUnMountNotification:(NSNotification *) note
...
+#define VK_FAST_FWD         0xB0
```
ksooo: VK_FAST_FWD now has the same value as VK_MEDIA_NEXT_TRACK (0xB0). Looks insane, no?

xbmc/platform/darwin/osx/SDLMain.mm
Rechi: Should I remove `VK_REWIND` and `VK_FAST_FWD` and replace them with `VK_MEDIA_PREV_TRACK` and `VK_MEDIA_NEXT_TRACK`?

ksooo: No idea, tbh. i'm not an osx/ios expert, only stumbled over the logical clash here.

@Memphiz ?

Memphiz: i'd say swap the hex codes from rewind with prevtrack and fast fwd with nexttrack - so we still have f15 and f16 around in case there is something that needs it?

Memphiz:
I'm ok with this. Please have a look in the osx forum in case there are issues with this reported by users - if you can .

PR #6392, opened 2015-02-10
drop some useless level 4 settings for player and renderer

FernetMenta:
see title

PR #10184, opened 2016-07-29
[cleanup] remove unused leftovers from settings removals

jjd-uk:
This removes leftovers from https://github.com/xbmc/xbmc/pull/6392 by @FernetMenta 

These were copied and pasted into the new layout as when I first started on the settings refactor I wasn't confident to remove this and didn't want to break anything. I now feel I have enough understanding that it would seem to me that these requirements now make no sense, as they are unused with no settings within this `<group>`, since `<requirements>` specified within a `<group>` only apply to that `<group>`.

FernetMenta can you confirm I'm correct and that this is safe to remove.

### DIFF
--- a/system/settings/darwin_ios.xml
+++ b/system/settings/darwin_ios.xml
@@ -7,7 +7,7 @@
           <visible>false</visible>
         </setting>
       </group>
-      <group id="4">
+      <group id="3">
         <setting id="videoplayer.hqscalers">
           <visible>false</visible>
         </setting>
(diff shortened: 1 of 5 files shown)


FernetMenta:
@jjd-uk looks ok to me, thanks

MartijnKaijser:
No changes needed in the platform files?

jjd-uk:
Silly me forgot since I'd changed the `<group>` id's to account for this removed group the platform files might need adjustment to account for the change of id's. Will check and add the necessary commits.

jjd-uk:
Ok that's the platform files done

PR #10185, opened 2016-07-29
[estuary] skin sync 0.8.4

BigNoid:
Adds all the commits from Estuary repo since #9933


MilhouseVH:
No TexturePacker (xbt) support `resources/*`?

MilhouseVH:
One problem with the screenshots is that they're only accessible via the mouse. If you use a keyboard or remote control it's not possible to highlight the screenshots - the input is "locked" to the buttons at the bottom of the screen.

MartijnKaijser:
ping @Memphiz 
Why are the icon.png and fanart.jpg handled different here compared to Estouchy?
https://github.com/xbmc/xbmc/blob/master/tools/darwin/Support/CopyRootFiles-ios.command#L56
https://github.com/xbmc/xbmc/blob/master/tools/darwin/Support/CopyRootFiles-osx.command#L38

They are now moved to /resources/ including some screenshots used in addon browser

BigNoid:
@MartijnKaijser I forgot to add it. Will do a follow up PR two fix these two issues.

MartijnKaijser:
What seemed strange to is the difference between Estuary and Estouchy

Memphiz:
because "KODI_HOME" is different between osx and ios app bundles @MartijnKaijser

Voyager1:
the removal of the first control in VideoFullScreen.xml also causes that you can't see anymore what you're typing when trying to enter a time to seek. I restored the deleted control and it works again, but probably too much stuff is still in there... This is the desired behaviour: e.g. enter 123 to seek to 1:23

Voyager1:
update: fix here. Feel free to cherry pick. https://github.com/Voyager1/xbmc/commit/179c7d7428021eac42d6321bbd642efec0a841d0

BigNoid:
Thx, I've included it in #10253

PR #10186, opened 2016-07-30
Remove all long-press actions from remote.xml

NedScott:
It appears that long-press does not behave well with input devices that use remote.xml (MCE, non-android CEC, etc). This removes those actions for now.

### DIFF
--- a/system/keymaps/remote.xml
+++ b/system/keymaps/remote.xml
@@ -27,7 +27,6 @@
   <global>
     <remote>
       <play>PlayPause</play>
-      <play mod="longpress">Info</play>
       <pause>Pause</pause>
       <stop>Stop</stop>
       <forward>FastForward</forward>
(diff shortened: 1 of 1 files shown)


MilhouseVH:
@Ned: I just ran some tests, and I added the following mappings to a custom `remote.xml` on LibreELEC RPi2 with Ortek VRC-1100 remote:
```
<play>Right</play>
<play mod="longpress">Down</play>
```
Effect: Pressing `Play` (long or short) results in a DOWN action.

And then I replaced the above with:
```
<play mod="longpress">Down</play>
<play>Right</play>
```
Effect: Pressing `Play` (long or short) results in a RIGHT action.

So it simply looks like the last mapping always overrides the first regardless of modifier, when instead there should be two distinct mappings (one for longpress, the other for a normal "shortpress"). It does appear that `remote.xml` (`<remote>`) doesn't support the "longpress" modifier, although I'm sure it can't be that simple as presumably it is working for whatever remote you have.

I can't see any reason why "longpress" shouldn't work with MCE (and possibly other) remotes - the Ortek VRC-1100 MCE remote I have generates two distinct KEY_DOWN/KEY_UP events, and presumably it's the time interval between the KEY_DOWN/KEY_UP events that determines if it's a short or long press.

MilhouseVH:
I could be reading the source code incorrectly, but in ButtonTranslator.cpp the "longpress" modifier seems to be supported only for `<keyboard>` mappings: https://github.com/xbmc/xbmc/blob/master/xbmc/input/ButtonTranslator.cpp#L1361-L1418

There doesn't appear to be any modifier support (longpress or otherwise) for `<remote>` mappings: https://github.com/xbmc/xbmc/blob/master/xbmc/input/ButtonTranslator.cpp#L1250-L1325

PR #10187, opened 2016-07-30
[EPG] Fix and simplify epg entry update.

ksooo:
No idea what i was smoking back then when i implemented this, but as it is with current master it can produce tons of log errors, like "Error: EPG_EVENT_CREATED: uid 12345678 found! Updating existing event." and "Error: EPG_EVENT_UPDATED: uid 87654321 not found. Inserting new event."

### DIFF
--- a/xbmc/epg/Epg.cpp
+++ b/xbmc/epg/Epg.cpp
@@ -309,129 +309,6 @@ void CEpg::AddEntry(const CEpgInfoTag &tag)
   }
 }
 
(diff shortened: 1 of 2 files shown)


ksooo:
jenkins build this please

Jalle19:
I saw those too in a user log but thought they were generated by some addon or something.

ksooo:
jenkins ios error is not related to this pr.

PR #10188, opened 2016-07-30
[cmake] handle optional dependencies as "auto" by default

wsnipex:
this aligns handling of optional dependecies with autotools:
ENABLE_FOO == AUTO:
 if dep is not found it will be disabled with a warning

ENABLE_FOO == ON:
 if dep is not found config will fail as fatal


wsnipex:
@fetzerch this addresses the cmake issue we discussed in https://github.com/xbmc/xbmc/pull/10170

fetzerch:
+1

I think it would be a good idea to refactor the option handling again a bit later.
- Platforms that don't support certain options shouldn't show them.
- Features vs. dependent libraries. For example for ENABLE_OPTICAL the user would get an error that Cdio wasn't found, it's not obvious that they are related. Another example is ENABLE_PLIST that enables Airplay.
- [FeatureSummary](https://cmake.org/cmake/help/latest/module/FeatureSummary.html) could be used to enrich options and libs with user friendly messages.
- Minimum versions should be set at the find_package call, not inside the module. Best would be if the modules would be as independent from kodi as possible (maybe with exceptions of cpluff and ffmpeg).
- Continue to modernize the find modules (good example how they should look like is FindTinyXML). Then instead of using _INCLUDE_DIRS, _LIBRARIES and _DEFINES, we can simply add the imported target (for example TinyXML::TinyXML) to the target_link_library call of kodi.

Paxxi:
@wsnipex This broke cmake on win32
```
C:\code\kodi\build [addfolder ↕ +49 ~0 -0 !]> cmake -G "Visual Studio 14" ..\project\cmake\
-- Warning: Did not find file Compiler/MSVC-ASM
-- Source directory: C:/code/kodi
-- Build directory: C:/code/kodi/build
-- Generator: Multi-configuration (Visual Studio 14 2015)
-- CMake Version: 3.6.1
-- System type: Windows
-- Core system type: windows
-- Platform:
-- CPU: AMD64, ARCH: win32
-- Could NOT find PkgConfig (missing:  PKG_CONFIG_EXECUTABLE)
-- MICROHTTPD_HOME env is not set, setting it to /usr/local
-- Looking for microhttpd in /usr/local
-- Could NOT find Alsa (missing:  ALSA_INCLUDE_DIRS ALSA_LIBRARIES)
-- Could NOT find UDev (missing:  UDEV_INCLUDE_DIRS UDEV_LIBRARIES)
-- Could NOT find Dbus (missing:  DBUS_INCLUDE_DIRS DBUS_LIBRARIES)
CMake Error at C:/Program Files/CMake/share/cmake-3.6/Modules/FindPackageHandleStandardArgs.cmake:148 (message):
  Could NOT find Avahi (missing: AVAHI_INCLUDE_DIRS AVAHI_LIBRARIES)
  C:/Program Files/CMake/share/cmake-3.6/Modules/FindPackageHandleStandardArgs.cmake:388 (_FPHSA_FAILURE_MESSAGE)
  modules/FindAvahi.cmake:23 (find_package_handle_standard_args)
  scripts/common/Macros.cmake:335 (find_package)
  CMakeLists.txt:149 (core_optional_dep)


See also "C:/code/kodi/build/CMakeFiles/CMakeOutput.log".
See also "C:/code/kodi/build/CMakeFiles/CMakeError.log".
```
Reverting this and it works again

PR #10189, opened 2016-07-30
[estuary] skin sync pt II

BigNoid:
Follow up to #10185 

This adds the add-on news item as clickable item in add-on info which was added by phil but got lost in the rebase (and with that fixes the nav to the screenshots). 
Also changes the build script for iOS and macOS to reflect the change to `resources` folder
schermafbeelding 2016-07-30 om 11 32 11

schermafbeelding 2016-07-30 om 11 30 17


MartijnKaijser:
@BigNoid screenshots now can also be reached with keyboard/remote?

BigNoid:
> @BigNoid screenshots now can also be reached with keyboard/remote?

Yep @MartijnKaijser 

> (and with that fixes the nav to the screenshots)

MilhouseVH:
Screenshot nav with keyboard/remote is now working just fine, thanks!

PR #10192, opened 2016-07-30
rework to ACTION_PLAYER_PROCESS_INFO

FernetMenta:
- rename old codecinfo to playerdebug, this fixes showing codec info for pictures
- move handling of ACTION_PLAYER_PROCESS_INFO to player. this is much cleaner than handling it in application

@ksooo /cc


FernetMenta:
jenkins build this please

ksooo:
Is player process info really designed to be a VideoPlayer only feature? I thought it is designed player agnostic and that current state that only VideoPlayer supports this 'interface' is only for limited time until other players implement this.

ksooo:
That was the reason for putting the code handling the new action to CApplication.

MartijnKaijser:
Ping @Montellese as this will break JSON-RPC compatibility from previous Kodi versions

xbmc/input/ButtonTranslator.cpp
```diff
@@ -87,7 +87,7 @@ static const ActionMapping actions[] =
     { "showsubtitles"            , ACTION_SHOW_SUBTITLES },
     { "nextsubtitle"             , ACTION_NEXT_SUBTITLE },
     { "cyclesubtitle"            , ACTION_CYCLE_SUBTITLE },
-    { "codecinfo"                , ACTION_SHOW_CODEC },
+    { "playerdebug"              , ACTION_PLAYER_DEBUG },
```
ksooo: imo codecinfo is currently still used in system/keymaps/(gamepad|joystick|remote).xml You need to change this as well.

xbmc/pictures/GUIWindowSlideShow.cpp
```diff
@@ -737,7 +737,7 @@ bool CGUIWindowSlideShow::OnAction(const CAction &action)
 {
   switch (action.GetID())
   {
-  case ACTION_SHOW_CODEC:
+  case ACTION_PLAYER_PROCESS_INFO:
     {
       CGUIDialogPictureInfo *pictureInfo = (CGUIDialogPictureInfo *)g_windowManager.GetWindow(WINDOW_DIALOG_PICTURE_INFO);
```
ksooo: imo neither ACTION_SHOW_CODEC nor ACTION_PLAYER_PROCESS_INFO is correct here. Seems that slide show simply misuses an existing action with completely unrelated semantics to display a picture info dialog.

ksooo:
> this fixes showing codec info for pictures

 ooops. my intention was to stick with 'codecinfo' assignment in "slideshow' section for all input devices, as there obviously is done something not related to player process info, but obviously i accidentally changed it in keyboard.xml (while the others gamepad, remote,joystick are fine). So, it is 'only' broken for 'keyboard'.

xbmc/input/ButtonTranslator.cpp
FernetMenta: will do, I thought you changed this all to processinfo

xbmc/pictures/GUIWindowSlideShow.cpp
FernetMenta: right but this has ever been wrong. for now it is important that the info will show on "o".

FernetMenta:
> Is player process info really designed to be a VideoPlayer only feature?

if other players want to handle it, they first need to implement it. then they can handle the action.

FernetMenta:
> That was the reason for putting the code handling the new action to CApplication.

we don't handle all window actions in windowManager only because the same action is handled by more than a single window.

FernetMenta:
@MartijnKaijser compatibility got long broken with dropping codecInfo dialog. Exposing some removed function to an API is useless and wrong.

xbmc/input/ButtonTranslator.cpp
ksooo: see my other comment, actually wanted to keep it, but accidentally changed it for one input source (keyboard).

FernetMenta: codecinfo is history, makes no sense to keep the action

ksooo: yeah yeah, all fine with this. go ahead, just wanted to explain my intention and what actually went wrong .

FernetMenta: ah, ok. sorry if you perceived my comments as offensive.

ksooo: While you are at it (changing the cpp and xml): Can't the slide show picture info dialog stuff be changed to react on ACTION_SHOW_INFO which is more appropriate or is this already used differently there?

FernetMenta: good idea. I'll look into it.

ksooo: thx

FernetMenta: this is more work than I thought :) already found 2 bugs and there is at least a third one making my system crash when opening pictures.

FernetMenta:
@ksooo picture info is now triggered by ACTION_SHOW_INFO. This was already mapped but the action was not taken. For v18 we can drop one of those mappings. For now I would keep both because users would expect it to show on "o" key.

@MartijnKaijser I dropped the JSON-RPC changes in this commit. Can be fixed later. Since ACTION_SHOW_CODEC is gone JSON-RPC returns MethodNotFound on ShowCodec.

ksooo:
Looks all good to me now.

Tolriq:
Since I'm posting on the other PR let's do all
Technically this is a fix for JSON as thing is deprecated, so version.txt should have rev updated. '7.20.1'

Any change of behavior as simple as it could be must have version updated.

FernetMenta:
bumped version
jenkins build this please

MartijnKaijser:
ok to merge?

FernetMenta:
you mean merge window opened in the far east already? :)

MartijnKaijser:
yep :)

PR #10194, opened 2016-07-30
[cmake/win32] Don't find system wide installed mysqlclient on Windows

fetzerch:
There might be better ways to do this (on a broader level similar to cross compiling options on other platforms). But mysql is the one dependency that has a high chance to be installed system wide on dev machines, so it might be good to get that as fix in.

ping: @mapfau, @Paxxi

### DIFF
--- a/project/cmake/modules/FindMySqlClient.cmake
+++ b/project/cmake/modules/FindMySqlClient.cmake
@@ -14,11 +14,20 @@
 #
 #   MySqlClient::MySqlClient   - The MySqlClient library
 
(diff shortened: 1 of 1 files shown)


fetzerch:
I've implemented an alternative more generic approach in https://github.com/fetzerch/xbmc/commit/92bb56974499311c6e5486823e24ea9d3ba59891. However with @Paxxi we decided to package a cmake config for mysqlclient and make sure cmake uses that.

fetzerch:
Jenkins build this please

PR #10197, opened 2016-07-30
remove win32 win10sdk dependency

ghost:
This enables VS2015 builds with cmake generated sln if winsdk10 version is > the one specified in this project file


MartijnKaijser:
jenkins build and merge

PR #10198, opened 2016-07-30
[PVR] Timers: Fix changing timer type after creation.

ksooo:
Followup to #10057 
Bug reported in the forum: http://forum.kodi.tv/showthread.php?tid=284616

@metaron-uk fyi
@Jalle19  for review?

### DIFF
--- a/xbmc/pvr/windows/GUIWindowPVRTimersBase.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRTimersBase.cpp
@@ -94,6 +94,8 @@ void CGUIWindowPVRTimersBase::GetContextButtons(int itemNumber, CContextButtons 
 
         if (timerType && !timerType->IsReadOnly() && timer->GetTimerRuleId() == PVR_TIMER_NO_PARENT)
           buttons.Add(CONTEXT_BUTTON_EDIT_TIMER, 21450);  /* Edit */
+        else
+          buttons.Add(CONTEXT_BUTTON_EDIT_TIMER, 19241);  /* View timer information */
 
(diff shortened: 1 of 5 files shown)


ksooo:
jenkins build this please

ksooo:
@Jalle19 I really would like to have this in alpha 3 as it also  fixes a crash. If you won't find time for review today, np. Changes are straight forward, imo very low regression potential.

xbmc/pvr/windows/GUIWindowPVRTimersBase.cpp
```diff
@@ -372,12 +374,7 @@ bool CGUIWindowPVRTimersBase::ActionShowTimer(CFileItem *item)
   }
   else
   {
-    const CPVRTimerInfoTagPtr tag(item->GetPVRTimerInfoTag());
-    if (ShowTimerSettings(tag) && !tag->GetTimerType()->IsReadOnly())
-    {
-      /* Update timer on pvr backend */
-      bReturn = g_PVRTimers->UpdateTimer(tag);
-    }
+    bReturn = EditTimer(item);
```
ksooo: Note to self/reviewer: this fixes a crash after changing timer type, because I forgot to prpare this code for the new feature. Best is to use EditTimer here as it does exactly the right thing and we have no code duplication.

Jalle19:
Had a quick look through, I suppose it's good. +1.

ksooo:
Thanks @Jalle19.

metaron-uk:
Seems to address the issue I reported. Thanks @ksooo  :)

PR #10199, opened 2016-07-31
[SETTINGS] Add Source settings to call source selection windows...

jjd-uk:
.... and rename Library settings section to Media to allow creation of Media -> Library category to contain new Source settings and previous Database settings.

This was inspired by @phil65 **"Media sources"** item in Estuary, so this add similar ability to core in order to be available to everyone and not just Estuary users.

image

image

These changes would allow removal of "Media sources" from Estuary as this function could now be handled by core, however didn't want to be presumptuous about doing that myself, so will leave that decision to the skinners.

I've marked this as [WIP] since I've renamed the Library section to Media, however to be done in a 100% proper manner the underlying window would also need to be renamed, for example:

https://github.com/xbmc/xbmc/blob/master/xbmc/guilib/WindowIDs.h#L46

**WINDOW_SETTINGS_LIBRARY** to **WINDOW_SETTINGS_MEDIA**

However this would break skins already updated to use the `ActivateWindow(LibrarySettings)` call in their skin `Settings.xml`, so would require changing to `ActivateWindow(MediaSettings)` if the window was renamed. 

Thus at the moment I've played it safe and only included the labelling changes and I've **NOT** yet renamed the underlying window name in code to avoid breaking skins. If this is to be merged for Krypton a decision is needed on whether we:
- rename window to **WINDOW_SETTINGS_MEDIA** now and break skins `ActivateWindow(LibrarySettings)` call
- keep **WINDOW_SETTINGS_LIBRARY**  and `ActivateWindow(LibrarySettings)` for Krypton then rename to **WINDOW_SETTINGS_MEDIA** in v18

ping @MartijnKaijser @Paxxi and additional pings for opinions on whether to do the window rename at this point of the release cycle @ronie @BigNoid


jjd-uk:
jenkins build this please

Paxxi:
Too late in the release cycle to change the window name in my opinion, tag it with a todo and schedule it for v18 alpha 1

jjd-uk:
Just occured to me, I could do the Window renaming now in a seperate PR then that PR could either be accepted for Kyprton (assuming this one goes in), or the merger of that PR is delayed until we've branched Kypton and Master is open for v18.

MartijnKaijser:
I like the change.  I'll leave it up to @ronie and skinners as he can tell if skins already use the name. If not I'd see no issue including it.

ronie:
i'm fine with the window name change as well.
we're still in alpha, so no problem with skin breakage.

HitcherUK:
It's no big deal to change `ActivateWindow(LibrarySettings)` to `ActivateWindow(MediaSettings)`

jjd-uk:
PR updated to include the underlying code changes necessary to rename the window from LIBRARY to MEDIA.

jenkins build this please

jjd-uk:
Windows test build if anyone wants to see this change http://mirrors.xbmc.org/test-builds/win32/KodiSetup-20160731-b94a377-media.exe

MartijnKaijser:
Looks good but shouldn't the media sources button be removed?

jjd-uk:
After this is merged the "Media sources" button is indeed not required, but I took that as being a skin decision as to whether @phil65 @ronie @BigNoid want to keep it or not, don't want to intrude too much on other peoples turf ;)

ronie:
i'd say ditch it :)

HitcherUK:
@jjd-uk Might want to add descriptions for each item as well for the textbox ID(6)

jjd-uk:
@HitcherUK do you mean the Help descriptions? if so now added.

MartijnKaijser:
@jjd-uk you can add the removal of the skin button in this PR or make it a separate one. Up to you.

jjd-uk:
I'd prefer not to do it without phils ok, so I could PR the button removal to the Estuary repo

jjd-uk:
Jenkins has gone missing since my last force push, just to be sure

jenkins build this please

PR #10200, opened 2016-07-31
Avoid addons to be detected as system libraries

ghost:
Earlier detection of if a library is system or not was done by detection if it is hooked or not (afaik)
Beginning with NeedHook() binary addons are detected as systemlibraries and not freed after use.

This commit sets the IsYstemLibrary to false if the creation path is from Addons.

### DIFF
--- a/xbmc/cores/DllLoader/DllLoaderContainer.cpp
+++ b/xbmc/cores/DllLoader/DllLoaderContainer.cpp
@@ -251,7 +251,7 @@ LibraryLoader* DllLoaderContainer::LoadDll(const char* sName, bool bLoadSymbols)
 #ifdef TARGET_POSIX
   pLoader = new SoLoader(sName, bLoadSymbols);
 #elif defined(TARGET_WINDOWS)
-  pLoader = new Win32DllLoader(sName);
+  pLoader = new Win32DllLoader(sName, false);
 #else
   pLoader = new DllLoader(sName, m_bTrack, false, bLoadSymbols);
 #endif
--- a/xbmc/cores/DllLoader/Win32DllLoader.cpp
+++ b/xbmc/cores/DllLoader/Win32DllLoader.cpp
@@ -132,10 +132,11 @@ Export win32_exports[] =
   { NULL,                          -1, NULL,                                NULL }
 };
 
-Win32DllLoader::Win32DllLoader(const std::string& dll) : LibraryLoader(dll)
+Win32DllLoader::Win32DllLoader(const std::string& dll, bool isSystemDll)
+  : LibraryLoader(dll)
+  , bIsSystemDll(isSystemDll)
 {
   m_dllHandle = NULL;
-  bIsSystemDll = false;
   DllLoaderContainer::RegisterDll(this);
 }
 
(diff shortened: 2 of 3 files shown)


xbmc/DynamicDll.cpp
```diff
@@ -25,17 +25,19 @@
 
 using namespace XFILE;
 
-DllDynamic::DllDynamic()
+DllDynamic::DllDynamic():
```
Paxxi: this is completely pointless comment but moving the `:` to the next row gives a nice symmetry as everything lines up :)

Paxxi:
I like it but since you're touching this can you maybe reword the log message as well, change the type to DEBUG maybe and something less scary like "Tried to unload system dll, this may indicate a bug but Kodi will function as expected as the issue was prevented"

xbmc/DynamicDll.cpp
```diff
@@ -83,6 +85,16 @@ bool DllDynamic::EnableDelayedUnload(bool bOnOff)
   return true;
 }
 
+bool DllDynamic::SetSystemLibrary(bool isSystem)
```
Paxxi: This can be changed to return void as the return value is never checked and Coverity like to complain about ignoring return values.

xbmc/DynamicDll.cpp
mapfau: Yes I like it also with : on next line. I followed a discussion in the past between @fritsch and @FernetMenta  and kept in mind that they want to have it in the first line. but maybe I mix up things :-) So if I'm not the only one who likes it in line 2. I'll love to change it

xbmc/DynamicDll.cpp
mapfau: I followed the line of this one:

bool DllDynamic::EnableDelayedUnload(bool bOnOff)

Same situation -> should I remove the return value also?
RetVal seems to be never used also.

mapfau:
https://github.com/xbmc/xbmc/blob/master/xbmc/network/AirPlayServer.cpp#L888

I don't think that this would work with the current code. EnableDelayedUnload call afert Load will fail, because m_dll is valid after load.

mapfau:
I have one commit regarding Airplay / Airtunes waiting for this one waiting here.
Is there any reason for holding it back? Should I change something?

MartijnKaijser:
@Paxxi @mapfau ping if this is something for v17?

mapfau:
I would love to see it in v17 because I can cleanup branches :-)

Currently all binary addons are loaded on start but not unloaded, because they are detected as system dlls. Not a very nice situation....

tyl0re:
Would be good. not only Unload isnt Possible also Update an Addon via Repo isnt possible(On WIndows) because deleting the Old Version cant be done because its open. ) Hope this would bring the Solution

Paxxi (commented):
Looking over it again I'm not sure I like the extra boolean, rather have a separate LoadSystemDll to make it more readable. 

There's some minor styling issues with missing space around = and could be updated to use nullptr while we're touching this anyway

peak3d:
moved to https://github.com/xbmc/xbmc/pull/10661

PR #10201, opened 2016-07-31
[settings] restore input devices to system, and move json-rpc app set…

jjd-uk:
…tings back to services & combine with webserver settings.

A bit more minor reshuffling to get as many rough edges for the settings refactor ironed out as possible before release. Following feedbck it seems I unintentionally made life slightly more difficult for those users that use the json-rpc remote control apps.

Idea of Control in Interface section was to group settings for both physical and app remotes, but I forgot you still need to enable the webserver to use json-rpc.

This PR places a Input category to System (effectively old Input Devices)

image

Then moves the json-rpc app settings to a new Control category in Services, and combines these with the webserver settings. Therefore all settings to get the remote apps working are now in the one window. ~~Also since the webserver is required to be enable to use json-rpc app i've introduced a dependency in the settings~~ (removed from pr)

image

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -7326,7 +7326,7 @@ msgid "Security"
 msgstr ""
 
 msgctxt "#14094"
-msgid "Input devices"
+msgid "Devices"
 msgstr ""
 
 #: system/settings/settings.xml
@@ -8136,6 +8136,7 @@ msgstr ""
 
 #. Audio DSP process chain indication
 #: xbmc/settings/dialogs/GUIDialogAudioDSPSettings.cpp
+#: system/settings/settings.xml
 msgctxt "#15087"
 msgid "Input"
 msgstr ""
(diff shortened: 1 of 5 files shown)


Tolriq:
Seems it's way harder to not care about something I cared for 7 years :)

Anyway some details about those settings :

- **There's no relation at all between WebServer and the 2 settings in application control.**

Web server is one way to use JSON.

The other way is JSON over TCP that is enabled with the application control setting. 
Most application use both but they are really independent settings.

But (it's a very very big but) the application control settings, primarily manage **Event Server** that is used also for hardware remotes and many other things.

The allow remote control from app on this systems needs to be on by default even with no http as it's needed by default on Mac see @Memphiz Event Server recent experiments.

Anyway no one really know those subtilities and in future a separation of Event Server and JSON over TCP could avoid some of the recent miss understandings.

And thanks for putting those settings back where users seek them :)

jjd-uk:
Ok that is probably why there was no dependency before, I will remove that addition.

jjd-uk:
Rebased.

jenkins build this please

MilhouseVH:
Was it intentional that this PR would disable "Allow remote control via HTTP" for users that had already enabled this setting? Presumably there is no way to migrate the old setting. Just that it might cause a lot of support grief when users upgrade to a build with this change, and suddenly find their remote control etc. is no longer working because the webserver has been disabled again.

MilhouseVH:
Apologies for the noise: this is a LibreELEC issue as we enable the webserver by default in [appliance.xml](https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/mediacenter/kodi/config/appliance.xml#L53-L60) and this needs a tweak to match the updated settings.

PR #10202, opened 2016-07-31
[xbmc/win32/fix] Fix crash because of heap mismatch

Paxxi:
When running a debug version of Kodi we allocate
plugin_path on the debug heap and then pass it to
cpluff to free. cpluff being a release version does not
use the debug heap and crashes because it tries
to free an invalid address.

Also added some minor safety to the string copy

This is only an issue during development so no need to rush it in for alpha.


Paxxi:
jenkins build and merge

PR #10203, opened 2016-07-31
fix listprovider updating

tamland:
This fixes updating of dynamic lists that are not visible, but have been invalidated by a background thread. This is due to PENDING to not being included in IsUpdating, which  causes `Update` to never getting called. Also fixes some other issues found (see commits).


Razzeee:
jenkins build this please

tamland:
jenkins build this please

PR #10204, opened 2016-07-31
[xbmc/fix] Fix crash on shutdown

Paxxi:
On teardown the circular dependencies of Observer and Observable often lead to a crash.
Removed all state from Observer, it shouldn't care about
what it's Observing.
Switched to use atomic for message tracking to simplify and get rid of a lock.

@FernetMenta what do you think? I've noticed both deadlocks and plenty of crashes when closing Kodi related to this so removed most of it that wasn't really used anyway

### DIFF
--- a/xbmc/pvr/dialogs/GUIDialogPVRChannelsOSD.cpp
+++ b/xbmc/pvr/dialogs/GUIDialogPVRChannelsOSD.cpp
@@ -56,10 +56,8 @@ CGUIDialogPVRChannelsOSD::~CGUIDialogPVRChannelsOSD()
 {
   delete m_vecItems;
 
-  if (IsObserving(g_infoManager))
-    g_infoManager.UnregisterObserver(this);
-  if (IsObserving(g_EpgContainer))
-    g_EpgContainer.UnregisterObserver(this);
+  g_infoManager.UnregisterObserver(this);
+  g_EpgContainer.UnregisterObserver(this);
 }
 
 bool CGUIDialogPVRChannelsOSD::OnMessage(CGUIMessage& message)
@@ -183,14 +181,11 @@ CPVRChannelGroupPtr CGUIDialogPVRChannelsOSD::GetPlayingGroup()
 
 void CGUIDialogPVRChannelsOSD::Update()
 {
+  g_infoManager.RegisterObserver(this);
+  g_EpgContainer.RegisterObserver(this);
+
   // lock our display, as this window is rendered from the player thread
   g_graphicsContext.Lock();
-
-  if (!IsObserving(g_infoManager))
-    g_infoManager.RegisterObserver(this);
-  if (!IsObserving(g_EpgContainer))
-    g_EpgContainer.RegisterObserver(this);
-
   m_viewControl.SetCurrentView(DEFAULT_VIEW_LIST);
 
   // empty the list ready for population
--- a/xbmc/utils/Observer.cpp
+++ b/xbmc/utils/Observer.cpp
@@ -24,72 +24,20 @@
 
 #include <algorithm>
 
-Observer::~Observer(void)
-{
-  StopObserving();
-}
-
-void Observer::StopObserving(void)
-{
-  CSingleLock lock(m_obsCritSection);
-  for (auto& observable : m_observables)
-    observable->UnregisterObserver(this);
-}
-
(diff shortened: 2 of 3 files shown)


xbmc/pvr/dialogs/GUIDialogPVRChannelsOSD.cpp
```diff
@@ -183,14 +181,11 @@ CPVRChannelGroupPtr CGUIDialogPVRChannelsOSD::GetPlayingGroup()
 
 void CGUIDialogPVRChannelsOSD::Update()
 {
+  g_infoManager.RegisterObserver(this);
```
Paxxi: @FernetMenta is this okay, Thought there's no reason to hold the graphicsContext while grabbing locks in g_infoManager and g_EpgContainer.

xbmc/utils/Observer.cpp
```diff
@@ -98,44 +46,40 @@ void Observable::RegisterObserver(Observer *obs)
...
 void Observable::SetChanged(bool SetTo)
 {
-  CSingleLock lock(m_obsCritSection);
   m_bObservableChanged = SetTo;
 }
 
 void Observable::SendMessage(const ObservableMessage message)
 {
   CSingleLock lock(m_obsCritSection);
+  auto obs = m_observers;
```
FernetMenta: you introduce an unused variable here.
I understand what you have in mind but this would only work with shared pointers. If an observer unregisters after leaving the lock, this segfaults.

xbmc/pvr/dialogs/GUIDialogPVRChannelsOSD.cpp
FernetMenta: yes, holding gfx lock here was definitely undesired.

xbmc/utils/Observer.cpp
```diff
@@ -24,72 +24,20 @@
 
 #include <algorithm>
 
-Observer::~Observer(void)
```
FernetMenta: nice! lots of unused code?

FernetMenta:
I like the idea of simplifying this but the lock has to be held while notifying. The problem is that the objects hooked into this do way too much processing on the notify call. To be safe no Notify method must call other functions.

xbmc/utils/Observer.cpp
Paxxi: This was sloppiness on my part. 
Having shared pointers would make it sager but the current usage pattern I think is mostly long lived objects like *manager, will try and investigate a bit more

Paxxi:
@FernetMenta I reverted the change to SendMessage so it still holds the lock while notifying. This is still valuable as it fixes some crashes. Good for merge?

xbmc/utils/Observer.cpp
```diff
@@ -24,72 +24,20 @@
...
-  for (auto& observer : m_observers)
-    observer->UnregisterObservable(this);
-}
-
 bool Observable::IsObserving(const Observer &obs) const
 {
   CSingleLock lock(m_obsCritSection);
-  return find(m_observers.begin(), m_observers.end(), &obs) != m_observers.end();
+  return std::find(m_observers.begin(), m_observers.end(), &obs) != m_observers.end();
 }
```
Jalle19: Doesn't the act that this used to compile at all mean we have a `using namespace std;` in a header somewhere?

Paxxi: yup it would seem so, didn't track it down though

FernetMenta:
@Paxxi go ahead

Paxxi:
jenkins build and merge

Paxxi:
jenkins build and merge

xbmc/utils/Observer.cpp
```diff
@@ -24,72 +24,20 @@
 
 #include <algorithm>
 
-Observer::~Observer(void)
-{
-  StopObserving();
```
ksooo: @paxxi: This is imo a bad idea, because now no longer is guaranteed that on destruction observers are removed from their observable. Thus Observable::NotifyObservers may call observers which are already destructed. This may be the cause of the crash reported here http://forum.kodi.tv/showthread.php?tid=269814&pid=2388702#pid2388702.

... meanwhile I got a similar crash which seems to confirm my assumption: http://pastebin.com/sQsMHCDi

PR #10205, opened 2016-07-31
[settings] change System > Add-ons groupings/labels and add button to access browser

jjd-uk:
Since we already have buttons for manage dependencies and show running add-ons, this adds a button to access the add-on browser. With this the groupings have been slightly changed and re-labelled.

image

Where skins such as Confluence currently provide access to the add-on browser via a button in the list of settings Sections, this could be dropped if user can access the browser from System > Add-ons. For example the following "Add-ons" button could be removed if @HitcherUK wished.

image

Everyone will be relieved to hear this is my last planned settings change that I'll be submitting for consideration to be included in Krypton. Everything from here on in will be cleanup/fixes only where required.

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -7714,7 +7714,17 @@ msgctxt "#14260"
 msgid "Debug"
 msgstr ""
 
-#empty strings from id 14261 to 14269
+#: system/settings/settings.xml
+msgctxt "#14261"
+msgid "Add-on Browser"
+msgstr ""
+
+#: system/settings/settings.xml
+msgctxt "#14262"
+msgid "Update Policy"
+msgstr ""
+
+#empty strings from id 14263 to 14269
 
 #: system/settings/settings.xml
 msgctxt "#14270"
@@ -14551,7 +14561,12 @@ msgctxt "#24147"
 msgid "Failed to scan %s: %s"
 msgstr ""
 
-#empty strings from id 24148 to 24990
+#empty strings from id 24148 to 24989
+
+#: system/settings/settings.xml
+msgctxt "#24990"
+msgid "Browse add-ons..."
+msgstr ""
 
 #. Used as error message in add-on browser when add-on repository data could not be downloaded
 #: xbmc/filesystem/AddonsDirectory.cpp
@@ -14573,7 +14588,7 @@ msgstr ""
 
 #: system/settings/settings.xml
 msgctxt "#24994"
-msgid "Running"
+msgid "Running..."
 msgstr ""
 
 #: xbmc/filesystem/AddonsDirectory.cpp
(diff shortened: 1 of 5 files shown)


ronie:
the option to browse/install addons should be easily accessible and not hidden three levels deep in system settings imo.
not that i mind putting a shortcut there, but i wouldn't remove the main button (in confluence).

i don't like the current implementation in Estuary either tbh.

jjd-uk:
My main reason for adding the additional shortcut is easy access if you're already in there looking at dependencies for example, or changing the unknown sources. However I'm of the opinion the main means of accessing add-ons shouldn't be in Settings, and buttons such as I highlighted above shouldn't be there. My preference would be for skins like Confluence to have a main menu item to showcase the presence of add-ons more.

HitcherUK:
Adding the browser to the main menu is something I did for fTV -

screenshot001

I've considered adding it to Confluence but not sure how well it would go down with the team but I think we should be pushing our own addon as much as possible.

BigNoid:
I don't like the idea of adding a shortcut to addonbrowser in settings menu. That's not what you should be looking for when browsing that section imo.

Let's focus on making it better in Estuary for this point.

jjd-uk:
Then why do we have the other buttons there for dependencies and running add-ons?

BigNoid:
> Then why do we have the other buttons there for dependencies and running add-ons?

If you are in the settings menu, it's to setup something or fix a problem. Those two shortcuts can help diagnose a problem with the addons.

The addonbrowser has nothing to do with settings imo. If we put it back there, it's only because people are used to having it there.

jjd-uk:
But if you're trying to diagnose a problem wouldn't you want quick access to the main browser as well so you can easily switch between looking at dependencies/what's running and the main browser. Not a big deal for me personally, I just don't understand the logic, if I've not convinced people then to avoid spamming everyone I'll close this PR and perhaps it's something we can discuss at devcon.

BigNoid:
The problem is that people have a hard time finding the addonbrowser in Estuary. Imo adding a shortcut here is not the way to fix it. We should look at how to improve that in Estuary. The main problem is that it's an icon without text now and not obvious from first sight that's where the addonbrowser is. 
But indeed, maybe we should discuss this elsewhere.

HitcherUK:
This works nicely -

screenshot081

NedScott:
All add-on settings can be accessed from the add-on browser, making them all settings. Add-ons themselves can be seen as a setting because they are extending the very functions of Kodi. It is essential to have a connection there. It makes no sense to have to back out of one area of Kodi and go into an entirely different area to get complete control.

There's already a button to go from the add-on manager to settings, and all this does is adding a button to go the other way as well. That seems perfectly logical to me.

Finding the add-on manager is not the issue here, IMO. It's about users expecting a path/flow. Having part of the add-on settings in settings, but not the add-on manager, is very confusing. It does not flow.

PR #10206, opened 2016-07-31
[xbmc/fix/filesystem] Add a recursive version of CDirectory::Remove

Paxxi:
Tested on Windows and works fine
Untested on POSIX

@MilhouseVH care to try it out and see? If it builds that is :)


Paxxi:
Jenkins build this please

MilhouseVH:
Thanks Paxxi - seems to be working on Linux! The `archive_cache` folder is now being cleared on startup, thanks.

I did wonder if adding a boolean `recurse` parameter (default false) to the existing `::Remove()` would be a better implementation?

Paxxi:
I hate Boolean parameters as there's no indication what they do, this is much clearer when reading

PR #10149, opened 2016-07-21
VideoPlayer: drop deinterlace mode

FernetMenta:
deinterlacing can't be forced. either the stream has information on interlacing and what field is first or it can't be deinterlaced.
giving users the option to force deinterlacing of a progressive stream makes zero sense.

@popcornmix cold you please update OMX

PR #10207, opened 2016-07-31
[buildfix] cleanup after deinterlace mode drop

stefansaraev:
ref #10149


FernetMenta:
At least the change for IMX looks incomplete. IMX supports deinterlacing, hence it should handle VS_INTERLACEMETHOD_NONE
IMO better leave it broken in an obvious state than pretending it was adapted.

stefansaraev:
lol.

PR #10209, opened 2016-08-01
[droid] show correct labels of plugged USB drives

da-anda:
As the title says, we currently only show some rather odd device ID/hash/string instead of the label of the plugged storage device on Android. This should fix it. It's cherrypicked from SPMC, so thanks to @koying. I only moved the files to the new location (compared to Jarvis) and added them to the cmake list. This is entirely untested so far (waiting for a test build) - all I can say atm is that it's working fine in SPMC on the SHIELD and that it would be good to have in Krypton unless there are any objections.


koying:
Actually, looks like the API went public for N: https://developer.android.com/reference/android/os/storage/StorageVolume.html

da-anda:
didn't know I also had to add the new files to makefile.in as well. Fixed that and jenkins should be happy now

da-anda:
tested on the SHIELD - working as expected. Feel free to hit the button. Test build by jenkins is on mirrors

MartijnKaijser:
jenkins build and merge

PR #10177, opened 2016-07-28
Remove EnableStreamAtPTS

ghost:
see subject

PR #10212, opened 2016-08-02
inputstream: bump version

FernetMenta:
was forgotten with #10177

@tamland what is the story with this versions? It does not seem to be verified when binary addons are loaded and as a result outdated addons make the system crash.
Shouldn't we check the abi version of extension points when addons are loaded?

@mapfau /cc


tamland:
> @tamland what is the story with this versions? It does not seem to be verified when binary addons are loaded and as a result outdated addons make the system crash.
Shouldn't we check the abi version of extension points when addons are loaded?

From long before my time. It's only an install-time check + some repository hacks. On the long list of things that should be fixed..

PR #10214, opened 2016-08-02
BuildSetup: don't copy the contents of \system in \BUILD_WIN32\applic…

wiromare:
…ation

@Paxxi

### DIFF
--- a/project/Win32BuildSetup/BuildSetup.bat
+++ b/project/Win32BuildSetup/BuildSetup.bat
@@ -1,4 +1,4 @@
-@ECHO OFF
+﻿@ECHO OFF
 SETLOCAL ENABLEDELAYEDEXPANSION
 REM setup all paths
 SET cur_dir=%CD%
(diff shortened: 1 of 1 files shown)


Paxxi:
As far as I can tell from the diff this doesn't change the behaviour in any way but it does clean up the exclude_dll.txt file which is good ofc.

Have you seen any specific issue that this would solve?

wiromare:
in my case, i don't know why, xcopy fails to copy some extra files included in my modified version of kodi

after xcopy ..\..\system\*.dll BUILD_WIN32\application /E /Q /I /Y /EXCLUDE:exclude.txt > NUL

maybe it could be them same for you in future

anyway, even if nsis don't catch them I thought it was wrong to copy those folders in application

Paxxi:
Those folders aren't copied into the application folder in any of our nightly builds so either there's been a regression today or some other modification of yours may be at fault.

wiromare:
because nsis don't catch them 

  ;Start copying files
  SetOutPath "$INSTDIR"
  File "${app_root}\application\*.*"
  SetOutPath "$INSTDIR\addons"
  File /r "${app_root}\application\addons\*.*"
  File /nonfatal /r "${app_root}\addons\peripheral.*"
  File /r "${app_root}\addons\skin.*"
  SetOutPath "$INSTDIR\media"
  File /r "${app_root}\application\media\*.*"
  SetOutPath "$INSTDIR\system"
  File /r "${app_root}\application\system\*.*"
  SetOutPath "$INSTDIR\userdata"
  File /r "${app_root}\application\userdata\*.*"

in your setup if you run buildsetup.bat in BUILD_WIN32\application you should see the system's folders also in application

Paxxi:
ahh ok, I'll try it out later today and see

Paxxi:
yup you were entirely correct, thank you!

PR #10217, opened 2016-08-02
[linux] fix datadir (KODI_HOME) detection

wsnipex:
fix "Unable to find path to Kodi data files" after https://github.com/xbmc/xbmc/commit/c027eaf0d4a8096d18c9ff00513a757cf9fbbf6e
reported in http://forum.kodi.tv/showthread.php?tid=285263

PR #10218, opened 2016-08-02
[KODI_HOME] fix after c9bd52302a6451a4b8822562a13880ca0a7a1eae

mk01:
kodi not starting with error "Unable to find path to %s data files!",
failing to identify /XXX/share vs. /XXX/lib.


bkuhls:
https://github.com/xbmc/xbmc/pull/10217 should already contain a fix.

PR #10221, opened 2016-08-03
[PVR][guiinfo] Recordings: Add more VIDEOPLAYER info labels (e.g. VIDEOPLAYER_EPISODENAME).

ksooo:
This makes it possible for skinners to get more data for PVR recordings, for instance VIDEOPLAYER_EPISODENAME now also works for PVR recordings.

@Jalle19 for review?
@phil65 @BigNoid @HitcherUK fyi
@HitcherUK i will come up with a PR for confluence for recording episode name display

### DIFF
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -8674,10 +8674,91 @@ std::string CGUIInfoManager::GetVideoLabel(int item)
   }
   else if (m_currentFile->HasPVRRecordingInfoTag())
   {
+    const CPVRRecordingPtr tag(m_currentFile->GetPVRRecordingInfoTag());
+
     switch (item)
     {
-    case VIDEOPLAYER_PLOT:
-      return m_currentFile->GetPVRRecordingInfoTag()->m_strPlot;
+      case VIDEOPLAYER_TITLE:
+        return tag->m_strTitle;
+
+      case VIDEOPLAYER_GENRE:
+        return StringUtils::Join(tag->m_genre, g_advancedSettings.m_videoItemSeparator);
+
+      case VIDEOPLAYER_PLOT:
+        return tag->m_strPlot;
+
+      case VIDEOPLAYER_PLOT_OUTLINE:
+        return tag->m_strPlotOutline;
+
+      case VIDEOPLAYER_STARTTIME:
+        return tag->RecordingTimeAsLocalTime().GetAsLocalizedTime("", false);
+
+      case VIDEOPLAYER_ENDTIME:
+        return (tag->RecordingTimeAsLocalTime() + tag->m_duration).GetAsLocalizedTime("", false);
+
+      case VIDEOPLAYER_YEAR:
+        if (tag->HasYear())
+          return StringUtils::Format("%i", tag->GetYear());
+        break;
+
+      case VIDEOPLAYER_EPISODE:
+        if (tag->m_iEpisode > 0)
+        {
+          if (tag->m_iEpisode == 0) // prefix episode with 'S'
+            return StringUtils::Format("S%i", tag->m_iEpisode);
+          else
+            return StringUtils::Format("%i", tag->m_iEpisode);
+        }
+        break;
+
+      case VIDEOPLAYER_SEASON:
+        if (tag->m_iSeason > 0)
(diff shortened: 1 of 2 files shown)


xbmc/GUIInfoManager.cpp
```diff
@@ -8674,10 +8674,91 @@ std::string CGUIInfoManager::GetVideoLabel(int item)
...
+      case VIDEOPLAYER_YEAR:
+        if (tag->HasYear())
+          return StringUtils::Format("%i", tag->GetYear());
+        break;
+
+      case VIDEOPLAYER_EPISODE:
+        if (tag->m_iEpisode > 0)
+        {
+          if (tag->m_iEpisode == 0) // prefix episode with 'S'
+            return StringUtils::Format("S%i", tag->m_iEpisode);
```
Jalle19: Should probably be `E`, not `S`?

xbmc/GUIInfoManager.cpp
```diff
@@ -8674,10 +8674,91 @@ std::string CGUIInfoManager::GetVideoLabel(int item)
...
+          else
+            return StringUtils::Format("%i", tag->m_iEpisode);
+        }
+        break;
+
+      case VIDEOPLAYER_SEASON:
+        if (tag->m_iSeason > 0)
+          return StringUtils::Format("%i", tag->m_iSeason);
```
Jalle19: If you prefix the episode, perhaps it would make sense to prefix the season too? Or prefix none and leave it up to skins to do the formatting.

xbmc/GUIInfoManager.cpp
```diff
@@ -8674,10 +8674,91 @@ std::string CGUIInfoManager::GetVideoLabel(int item)
...
+        break;
+
+      case VIDEOPLAYER_EPISODENAME:
+        return tag->EpisodeName();
+
+      case VIDEOPLAYER_CHANNEL_NAME:
+        return tag->m_strChannelName;
```
Jalle19: Are these properties public or is this class a friend class? Not quite related to this PR but perhaps this could use some refactoring.

xbmc/GUIInfoManager.cpp
ksooo: No discussion about this, please. It is exactly like it is done for VideoInfoTags etc. I once had a discussion about this with montellese and he was very clear that this has to be exactly like it is. ;-) Okay?

xbmc/GUIInfoManager.cpp
ksooo: See my other comment, please. ;-)

xbmc/GUIInfoManager.cpp
ksooo: public. Just one of so many examples of not encapsulating data properly.

xbmc/GUIInfoManager.cpp
```diff
@@ -8674,10 +8674,91 @@ std::string CGUIInfoManager::GetVideoLabel(int item)
...
+        const CPVRChannelPtr channel(tag->Channel());
+        if (channel)
+          return channel->FormattedChannelNumber();
+        break;
+      }
+
+      case VIDEOPLAYER_CHANNEL_GROUP:
+      {
+        if (!tag->IsRadio())
+          return g_PVRManager.GetPlayingTVGroupName();
```
ksooo: No, we havn't. I have just copied the gist of the implementation of VIDEOPLAYER_CHANNEL_GROUP for CPVRChannel. I guess this can be improved, but wanted to do this in another PR.

ksooo:
jenkins build this please

xbmc/GUIInfoManager.cpp
Jalle19: This can't be right, no one in their right mind would except "episode 4" to be abbreviated `S04`?

ksooo: I have to admit that this looks really strange. If it is actually a bug, it has to be fixed here and at least two other places in this file. The other places date back until 2013... I would prefer to do the fix in a separate PR. Okay?

Jalle19: Of course

ksooo:
jenkins build this please

ksooo:
jenkins and travis errors are unrelated

PR #10224, opened 2016-08-03
Rename README.ubuntu to README-ubuntu.md

RodrigoHahn:
Improves compatibility with most tools by using a standard markdown file extension.


MartijnKaijser:
in that case all docs should be renamed and not just one

un1versal:
This change is ill conceived, renaming it alone will break formatting and make reading extremely difficult

PR #10226, opened 2016-08-03
[droid] remove google-breakpad

MartijnKaijser:
We don't use it so might as well stop building it


MrMC:
+1, MrMC fork removed it long ago.

MartijnKaijser:
jenkins build and merge

MartijnKaijser:
jenkins build and merge

tamland:
Last time I checked it worked fine and created stack traces, so what's the replacement?

MrMC:
gdb :) any competent android developer knows how to get gdb running under droid to debug issues.

tamland:
So you'll be providing training for users that is asked for a crashlog?

Breakpad is for crash reporting, and for when things have already happen, not for debugging. So no, gdb is no way a replacement for breakpad. But you're competent android developer so you already knew that ofc :)

MrMC:
I don't ask for crashlogs from Android users, never had, never will :) I might ask for a logcat which will typically point towards the suspect area. If they can't figure out how to get a logcat, then the last log file.

More important is their setup and what they did. If I can't reproduce the issue then I don't understand the real problem and any fix is a blind fix. And any competent developer knows that blind fixes are a last resort that is sure to come back and bite back in the rear later.

tamland:
That's.. really off-topic, and I can't say I'm that interested in the story of your workflow, sorry:\

The premise of this pr was that "We don't use it so might as well stop building it", which is false: #10230

MartijnKaijser:
In all these years it has been in i think it was only used once to get a crashlog from a user. Not sure why you refer to that PR as it was just a bit of code i forgot to remove with this PR. Any time i had an issue or crash koying asked for a logcat.

tamland:
Because the way you said it it sounded like it was just a build thing and the functionality was already gone. I have used it successfully a few times, it's useful, and atm I have no interest in starting to debug on android. I'm sure others find it useful as well, so unless there's a reason...

MrMC:
reason is... nice idea but turns out to be useless in development/production practice. Both previous android maintainers have dropped it in their own forks, nuff said.

PR #10228, opened 2016-08-04
inputstream: add API version check

FernetMenta:
adds API version check for inputstream addons

@mapfau /cc

### DIFF
--- a/xbmc/addons/InputStream.cpp
+++ b/xbmc/addons/InputStream.cpp
@@ -69,6 +69,18 @@ CInputStream::CInputStream(const AddonProps& props,
   }
 }
 
+bool CInputStream::CheckAPIVersion()
+{
+  std::string dllVersion = m_pStruct->GetApiVersion();
+  if (dllVersion.compare(INPUTSTREAM_API_VERSION) != 0)
+  {
+    CLog::Log(LOGERROR, "CInputStream::CheckAPIVersion - API version does not match");
+    return true;
+  }
+
+  return true;
+}
+
 void CInputStream::SaveSettings()
 {
   CAddon::SaveSettings();
--- a/xbmc/addons/InputStream.h
+++ b/xbmc/addons/InputStream.h
@@ -50,6 +50,7 @@ namespace ADDON
     virtual ~CInputStream() {}
 
     virtual void SaveSettings() override;
+    virtual bool CheckAPIVersion(void) override;
 
     bool UseParent();
     bool Supports(const CFileItem &fileitem);
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/kodi_inputstream_dll.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/kodi_inputstream_dll.h
@@ -214,6 +214,12 @@ extern "C"
   */
   bool IsRealTimeStream();
 
+  const char* GetApiVersion()
+  {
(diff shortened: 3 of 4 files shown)


mapfau:
Looks good, thx

FernetMenta:
jenkins build this please

FernetMenta:
Android build error unrelated

xbmc/addons/InputStream.cpp
```diff
@@ -69,6 +69,18 @@ CInputStream::CInputStream(const AddonProps& props,
...
+bool CInputStream::CheckAPIVersion()
+{
+  std::string dllVersion = m_pStruct->GetApiVersion();
+  if (dllVersion.compare(INPUTSTREAM_API_VERSION) != 0)
+  {
+    CLog::Log(LOGERROR, "CInputStream::CheckAPIVersion - API version does not match");
+    return true;
+  }
+
+  return true;
```
da-anda: why are you in both cases returning `true`? Doesn't make much sense.

FernetMenta: oops, this is obviously not correct

MilhouseVH:
When a older version of `inputstream.mpd` is installed (eg. v1.2.0, though probably not specific to this particular inputstream add-on) and the user upgrades to a version of Kodi with that includes this API check, the newer version of Kodi will crash on startup (see post [here](https://github.com/LibreELEC/LibreELEC.tv/pull/641#issuecomment-241618163)).

This currently makes it impossible for inputstream users to upgrade Kodi without manual intervention (ie. deleting inputstream add-ons, and then re-installing updated versions of the same add-ons).

Is there a simple fix to prevent a new version of Kodi from crashing when an older add-on is installed, as this is going to be a fairly common scenario.

FernetMenta:
ideally cpluff would care about versions but this is not implemented yet for binary addons. unfortunately I can't think of any simple fix for this. taking into account that inputstream api comes with v17 only alpha users are affected.

MilhouseVH:
It's a little surprising that a previously released add-on will be able to crash Kodi in this way.

Although it's currently affecting only alpha users (and we have quite a few!) it's likely to be an issue even with users upgrading from an alpha build to a full Kodi release build unless there's a fix of some kind added to Kodi. Upgrading the current add-on is a bit of a chicken/egg situation, as I'm not sure if the new add-on (ie. with API check support) will work with the older non-API Kodi builds.

Is it not possible to simply catch the error (and possibly disable the add-on), rather than segfault?

FernetMenta:
Nope, this is an incompatibility of ABI. The only way to prevent from segfault would be to have cpluff check version is advance and not load the shared library. But this is not done by our addon system. C has no capability of catching errors.

Alpha users have to be aware that those type of things can happen with developer versions.

garbear:
@MilhouseVH I've run into a similar problem, Kodi crashing on startup due to an add-on being present. I tracked it down to a circular dependency on initialization. Can you give 230c1d8 a try?

MilhouseVH:
@garbear thanks, I'll give it a try (might take a few hours before I can confirm)

MilhouseVH:
@garbear unfortunately your patch hasn't made any difference, Kodi continues to crash on startup. http://sprunge.us/Vdhj

mapfau:
@FernetMenta shouldn't we check for the kodi.inputstream version always on load time?

It's 1.0.4 with incompatible versions and was bumped to 1.0.5. There is no check at all against this version when loading inputstream.xy addons.

PR #10229, opened 2016-08-04
VideoPlayer: fix state of StreamInfo

FernetMenta:
see title

@ksooo /cc

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -857,8 +857,7 @@ bool CVideoPlayer::OpenInputStream()
 
 bool CVideoPlayer::OpenDemuxStream()
 {
-  if(m_pDemuxer)
-    SAFE_DELETE(m_pDemuxer);
+  CloseDemuxer();
 
   CLog::Log(LOGNOTICE, "Creating Demuxer");
 
@@ -896,6 +895,16 @@ bool CVideoPlayer::OpenDemuxStream()
   m_offset_pts = 0;
 
   return true;
+}
+
+void CVideoPlayer::CloseDemuxer()
+{
+  delete m_pDemuxer;
+  m_pDemuxer = nullptr;
+  m_SelectionStreams.Clear(STREAM_NONE, STREAM_SOURCE_DEMUX);
+
+  CServiceBroker::GetDataCacheCore().SignalAudioInfoChange();
+  CServiceBroker::GetDataCacheCore().SignalVideoInfoChange();
 }
 
 void CVideoPlayer::OpenDefaultStreams(bool reset)
@@ -1469,7 +1478,7 @@ void CVideoPlayer::Process()
       CDVDInputStream::ENextStream next = m_pInputStream->NextStream();
       if(next == CDVDInputStream::NEXTSTREAM_OPEN)
       {
-        SAFE_DELETE(m_pDemuxer);
+        CloseDemuxer();
 
         SetCaching(CACHESTATE_DONE);
         CLog::Log(LOGNOTICE, "VideoPlayer: next stream, wait for old streams to be finished");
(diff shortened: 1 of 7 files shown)


ksooo:
@FernetMenta good to go now. :-)

ksooo:
@FernetMenta just noticed that we still need to adapt CGUIInfoManager::m_audioInfo for PVR radio channels, exatly like we did with CGUIInfoManager::m_videoInfo for PVR TV channels. I suggest to put this into this PR.

xbmc/pvr/PVRManager.cpp
```diff
@@ -1471,6 +1478,16 @@ bool CPVRManager::PerformChannelSwitch(const CPVRChannelPtr &channel, bool bPrev
   return bSwitched;
 }
 
+void CPVRManager::SetChannelPreview(bool preview)
+{
+  m_isChannelPreview = preview;
+}
+
+bool CPVRManager::IsChannelPreview()
```
ksooo: this method should be const

xbmc/pvr/PVRManager.h
```diff
@@ -559,6 +559,10 @@ namespace PVR
      */
     void ConnectionStateChange(int clientId, std::string connectString, PVR_CONNECTION_STATE state, std::string message);
 
+    void SetChannelPreview(bool preview);
+
+    bool IsChannelPreview();
```
ksooo: Would you mind adding doxy for the two new member functions?

xbmc/pvr/PVRManager.cpp
```diff
@@ -1411,6 +1412,12 @@ bool CPVRManager::PerformChannelSwitch(const CPVRChannelPtr &channel, bool bPrev
     {
       delete m_currentFile;
       m_currentFile = new CFileItem(channel);
+
+      if (IsPlayingChannel(m_currentFile->GetPVRChannelInfoTag()))
```
FernetMenta: right, will change

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -4629,6 +4641,13 @@ void CVideoPlayer::GetVideoStreamInfo(int streamId, SPlayerVideoStreamInfo &info
   if (s.name.length() > 0)
     info.name = s.name;
 
+  if (s.width == 0 || s.height == 0)
```
ksooo: According to @fritsch checking width and height against zero is no good idea to check validity. He told me that tvheadend "sometimes" returns zero dimensions for "certain" channels. Maybe to workaround this tvh bug checking against string length of info.videoCodecName is acceptable?

FernetMenta: This is some left over and I am going to drop it because it did not change anything. Some parameters like aspect ratio or audio channels get updated later by stream players.

> checking against string length of info.videoCodecName is acceptable

codec name is always valid and does never change.

xbmc/pvr/PVRManager.h
FernetMenta: will do

xbmc/pvr/PVRManager.cpp
FernetMenta: ok

ksooo:
@FernetMenta works like a charme now for both tv and radio channels. Code looks good. +1 for merge.

FernetMenta:
jenkins build this please

axbmcuser:
http://trac.kodi.tv/ticket/16840
Maybe related - never had this problem ever before. Tried to double check with non self compiled nightly from http://mirrors.kodi.tv/nightlies/win32/ (4a61bf5) to be sure my build is not messed up.
Thanks!

ksooo:
@FernetMenta confirmed. Seems <code>g_PVRManager.SetChannelPreview(false)</code> gets not called when stopping playback. Inserting <code>g_PVRManager.SetChannelPreview(false)</code> in <code>CVideoPlayer::OnExit()</code> fixes the issue. Shall I open a PR or will you?

FernetMenta:
That would be the wrong place. The state belongs to pvrmanager and pvrmanager should reset it. pvrmanager.ispreviewactive should return false if no channel is active

ksooo:
> The state belongs to pvrmanager and pvrmanager should reset it.

 Okay. I will open a PR for this.

PR #10231, opened 2016-08-04
Fix warnings

rbalint:
Fixing warnings found by Debian unstable's gcc

### DIFF
--- a/lib/cpluff/libcpluff/ploader.c
+++ b/lib/cpluff/libcpluff/ploader.c
@@ -159,17 +159,17 @@ static void descriptor_errorf(ploader_context_t *plcontext, int warn,
 	message[127] = '\0';
 	if (warn) {
 		cpi_warnf(plcontext->context,
-			N_("Suspicious plug-in descriptor content in %s, line %d, column %d (%s)."),
+			N_("Suspicious plug-in descriptor content in %s, line %llu, column %llu (%s)."),
 		plcontext->file,
-		XML_GetCurrentLineNumber(plcontext->parser),
-		XML_GetCurrentColumnNumber(plcontext->parser) + 1,
+		(long long unsigned int)XML_GetCurrentLineNumber(plcontext->parser),
+		(long long unsigned int)XML_GetCurrentColumnNumber(plcontext->parser) + 1,
 		message);
 	} else {				
 		cpi_errorf(plcontext->context,
-			N_("Invalid plug-in descriptor content in %s, line %d, column %d (%s)."),
+			N_("Invalid plug-in descriptor content in %s, line %llu, column %llu (%s)."),
 			plcontext->file,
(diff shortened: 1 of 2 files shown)


MartijnKaijser:
Could you also send them upstream as we'll soon be removing to downloaded depends (at already done for windows)
https://github.com/jlehtine/c-pluff

rbalint:
They are already fixed in a slightly different way:
https://github.com/jlehtine/c-pluff/commit/b305106eb9996021bfd094e375d77eadcf22bf11

MartijnKaijser:
jenkins build this please
@wsnipex @Paxxi maybe for a comment?

xbmc/utils/StringUtils.cpp
```diff
@@ -1297,7 +1297,7 @@ std::string StringUtils::FormatFileSize(uint64_t bytes)
 {
   const std::array<std::string, 6> units{"B", "kB", "MB", "GB", "TB", "PB"};
   if (bytes < 1000)
-    return Format("%lluB", bytes);
+    return Format("%"PRIu64"B", bytes);
```
wsnipex: Please wrap in if defined(HAVE_INTTYPES_H) and make a fallback

Paxxi: Do we support a compiler that does not have it? We define PRIu64 ourselves on windows afaik in platformdefs to avoid ugly ifdefs sprinkled around the code

lib/cpluff/libcpluff/ploader.c
```diff
@@ -159,17 +159,17 @@ static void descriptor_errorf(ploader_context_t *plcontext, int warn,
 	message[127] = '\0';
 	if (warn) {
 		cpi_warnf(plcontext->context,
-			N_("Suspicious plug-in descriptor content in %s, line %d, column %d (%s)."),
+			N_("Suspicious plug-in descriptor content in %s, line %llu, column %llu (%s)."),
 		plcontext->file,
-		XML_GetCurrentLineNumber(plcontext->parser),
-		XML_GetCurrentColumnNumber(plcontext->parser) + 1,
+		(long long unsigned int)XML_GetCurrentLineNumber(plcontext->parser),
```
Paxxi: I'd prefer using uint64_t but just cosmetic so no real issue

xbmc/utils/StringUtils.cpp
wsnipex: well, we have this check in other places already and android is always a candidate for toolchain weirdness.

Razzeee:
I think you should probably drop the c-pluff stuff, as it will be superseeded by https://github.com/xbmc/xbmc/pull/10741

Razzeee:
On the other hand
> They are already fixed in a slightly different way:
> jlehtine/c-pluff@b305106

Doesn't seem to be in the 0.1.x branch we're using.

PR #10234, opened 2016-08-05
settings: Add settings option to control atempo resampling

popcornmix:
Atempo rsampling does increase cpu usage so may want to be disabled (e.g. on Pi 1).
The threshold when atempo resampling is used compared to normal resampling could be usefully changed.

This adds a setting for controlling this and defaults to disabled on Pi 1

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -6818,7 +6818,18 @@ msgctxt "#13516"
 msgid "Add art"
 msgstr ""
 
-#empty strings from id 13517 to 13549
+#: system/settings/settings.xml
+msgctxt "#13517"
+msgid "Threshold for pitch correction"
+msgstr ""
+
+#: system/settings/settings.xml
+msgctxt "#13518"
+msgid "When the speed change exceeds this threshold, a pitch-correction filter will be applied. This avoids the \"chipmonk voices\" that normally result from speeding up a video"
+msgstr ""
+
+
+#empty strings from id 13519 to 13549
 
 #: system/settings/settings.xml
 msgctxt "#13550"
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -2370,6 +2370,16 @@
           </constraints>
           <control type="list" format="string" />
         </setting>
+        <setting id="audiooutput.atempothreshold" type="integer" label="13517" help="13518">
+          <level>3</level>
+          <default>2</default> <!-- 2% -->
+          <constraints>
+            <minimum>0</minimum>
+            <step>1</step>
+            <maximum>100</maximum>
+          </constraints>
+          <control type="edit" format="integer" />
+        </setting>
         <setting id="audiooutput.samplerate" type="integer" label="458" help="36523">
           <level>2</level>
           <default>48000</default>
--- a/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEStream.cpp
+++ b/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEStream.cpp
@@ -714,9 +714,9 @@ bool CActiveAEStreamBuffers::IsDrained()
     return false;
 }
 
-void CActiveAEStreamBuffers::SetRR(double rr)
+void CActiveAEStreamBuffers::SetRR(double rr, double atempoThreshold)
 {
-  if (rr < 1.02 && rr > 0.98)
+  if (fabs(rr - 1.0) < atempoThreshold)
   {
     m_resampleBuffers->SetRR(rr);
     m_atempoBuffers->SetTempo(1.0);
--- a/system/settings/rbp.xml
+++ b/system/settings/rbp.xml
@@ -92,6 +92,9 @@
         <setting id="audiooutput.processquality">
           <default>101</default> <!-- AE_QUALITY_GPU -->
         </setting>
+        <setting id="audiooutput.atempothreshold">
+          <default>100</default> <!-- disabled -->
+        </setting>
       </group>
       <group id="3">
         <setting id="audiooutput.ac3transcode" help="37024">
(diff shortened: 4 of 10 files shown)


xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEStream.cpp
```diff
@@ -584,6 +585,7 @@ CActiveAEStreamBuffers::CActiveAEStreamBuffers(AEAudioFormat inputFormat, AEAudi
   m_inputFormat = inputFormat;
   m_resampleBuffers = new CActiveAEBufferPoolResample(inputFormat, outputFormat, quality);
   m_atempoBuffers = new CActiveAEBufferPoolAtempo(outputFormat);
+  m_atempoThreshold = CSettings::GetInstance().GetInt(CSettings::SETTING_AUDIOOUTPUT_ATEMPOTHRESHOLD) / 100.0;
```
FernetMenta: audio settings are managed by AE, loaded by CActiveAE::LoadSettings()
m_atempoThreshold can be set by CActiveAE::CreateStream

system/settings/settings.xml
```diff
@@ -2370,6 +2370,16 @@
           </constraints>
           <control type="list" format="string" />
         </setting>
+        <setting id="audiooutput.atempothreshold" type="integer" label="13517" help="13518">
+          <level>2</level>
```
MartijnKaijser: Shouldn't it be an advanced setting?

FernetMenta: Could be even an expert setting

popcornmix:
@FernetMenta updated. Now expert setting and setting is set in CActiveAE::LoadSettings.

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -6818,7 +6818,18 @@ msgctxt "#13516"
 msgid "Add art"
 msgstr ""
 
-#empty strings from id 13517 to 13549
+#: system/settings/settings.xml
+msgctxt "#13517"
+msgid "Threshold when atempo resampling kicks in"
```
garbear: Can we remove the word atempo? This will confuse users, and googling returns no results

popcornmix:
Suggestions welcome.
"Threshold when pitch preserving resampling is applied"?

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEStream.cpp
```diff
@@ -716,7 +718,7 @@ bool CActiveAEStreamBuffers::IsDrained()
 
 void CActiveAEStreamBuffers::SetRR(double rr)
 {
-  if (rr < 1.02 && rr > 0.98)
+  if (fabs(rr - 1.0) < m_atempoThreshold)
```
FernetMenta: if we passed threshold as parameter of SetRR instead of constructor, users would be able to change threshold on the fly. just add the setting to CActiveAE::OnSettingsChange

FernetMenta:
@popcornmix thanks, almost done. see my comment

addons/resource.language.en_gb/resources/strings.po
FernetMenta: This is wrong in almost every sense :)

> Above this speed

threshold is no speed

> audio is resampled to preserve the pitch

it is the other way round. resampling does not preserve the pitch

> This has a minor performance impact

we don't have any measurements on how atempo filter compares to resampling. apart from pi both is done by CPU.

garbear: Then let my ignorance be an example of one of those confused users I mentioned :) The takeaway is that this needs to be explained much, much better.

FernetMenta: This is not a proposed help text but some background info:

For minor speed adjustments resampling is the preferred method. In regard to a/v sync it is more accurate. I am not completely sure but I think audio quality is also better.
For larger speed adjustments resampling is not ideal because it changes the pitch -> chipmunk voices. With v17 we introduced ffmpeg's atempo filter but we don't have too much experience with it yet. The atempo filter tries to keep the audio pitch.

popcornmix:
@FernetMenta updated

I did a quick profile. ActiveAE thread on Pi3 takes:
gpu resampling: 5%
low resampling: 15%
medium resampling 20%
high resampling 35%
atempo resampling: 15%

So it appears the atempo resampling is actually cheaper than medium resampling, but stopping the use of gpu  resampling is the problem for Pi1.

I wonder if the audio quality of atempo resampling is noticeably lower for the audiophiles?

addons/resource.language.en_gb/resources/strings.po
garbear: Thanks for the explanation. Instead of a "pitch preserving resample algorithm", i would call this feature "pitch correction". So a better title text would be "Threshold for pitch correction".

The help text is still a little confusing to me. Is "resample ratio" just the speed factor, e.g. a 2% resample ratio is a 2% increase/decrease in speed?

popcornmix: Yes, we are referring to a 2% increase/decrease in speed as the threshold.
Technically that is a resample ratio of 98%/102%

garbear: ok, title text: "Threshold for pitch correction"

help text: "When the speed change exceeds this threshold, a pitch-correction filter will be applied. This avoids the "chipmonk voices" that normally result from speeding up a video."

FernetMenta:
> I wonder if the audio quality of atempo resampling is noticeably lower for the audiophiles?

Ask them and they answer yes. Just change an they won't notice :)

Thanks for your good work here!

addons/resource.language.en_gb/resources/strings.po
FernetMenta: Thanks. I am ok with this.

popcornmix:
Updated with @garbear's suggested wording (which looks good to me).

FernetMenta:
jenkins build this please

MartijnKaijser:
Merge?

PR #10235, opened 2016-08-05
Music Library Filtering Simple Fixes

DaveTBlake:
Fixing a few bugs in the way filtering, the option on the side menu,  is applied to the music library.

The most obvious change it that filter albums and songs by year now works. The bug for year was also stopping the lower filter items from displaying, so these missing items now appear. Songs by duration also works now. Using artist name to filter albums was only allowing selection from album artists, but then appying the filter to the song artist table, which can produce inconsistent results.


DaveTBlake:
jenkins build this please

DaveTBlake:
It seems that no one uses music filtering, at least no one has complained about the bugs. But hey, let's fix'em any way.

PR #10236, opened 2016-08-05
use final redirected path for internet streams

kenji123:
This PR makes sure input streams use the final redirected url for internet based streams.

Currently, even in some cases when the stream seems to successfully open, subsequent requests use the base path of the original url rather than the final redirected one. For example load balancer urls for HLS streams.

There may be a better way for handling this, but it's the broadest, yet simplest way that I could find.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDInputStreams/DVDFactoryInputStream.cpp
+++ b/xbmc/cores/VideoPlayer/DVDInputStreams/DVDFactoryInputStream.cpp
@@ -36,6 +36,7 @@
 #include "FileItem.h"
 #include "storage/MediaManager.h"
 #include "URL.h"
+#include "filesystem/CurlFile.h"
 #include "filesystem/File.h"
 #include "utils/URIUtils.h"
 #include "ServiceBroker.h"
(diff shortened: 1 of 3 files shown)


FernetMenta:
if ContentLookup is disabled, this new code must be bypassed

MartijnKaijser:
@kenji123 if you update the PR always tell so as we don't get notified by a push.

FernetMenta:
i almost missed this. could be solution for trac 16732

manijak:
This could definitely be a solution for trac 16732! I've done a few tests on a OSX build and it works. 
More info here: http://trac.kodi.tv/ticket/16732#comment:17

FernetMenta:
do you recommend to merge this change?

manijak:
Yes, for sure!

MilhouseVH:
Had a report of a problem after this merged, is it related?

http://forum.kodi.tv/showthread.php?tid=269814&pid=2406314#pid2406314

Hopefully the requested crashlog may provide more information. A user in the LibreELEC x86_64 thread is also asking about problems with various streaming add-ons but has provided no information.

FernetMenta:
please open trac tickets for bug reports

MilhouseVH:
http://trac.kodi.tv/ticket/16874

MilhouseVH:
This PR is definitely the cause of the Shoutcast-related crashes, details added on tract ticket.

tojo2503:
This one is maybe related to this bug: http://trac.kodi.tv/ticket/17005 ?
Could someone look into it? Can also provide more details / logs...

EusebiuOprinoiu:
Hey, guys!
I have the same problem as @tojo2503. 
Streaming media files from a password protected HTTP(S) directory doesn't work when accessed from the "Movies" or "TV Shows" menus. (They work from the Files Manager. Scraping works as well.)

arnova:
Ah yes, that must be the piece of the puzzle I previously read over: It's password protected. I'll try to reproduce again this weekend using a password protected http server.

MartijnKaijser:
@arnova https://github.com/xbmc/xbmc/pull/10799 ?

PR #10237, opened 2016-08-05
VideoPlayer: fix early stall of audio stream, sync is now in AE

FernetMenta:
see title + cosmetics


FernetMenta:
jenkins build this please

PR #10238, opened 2016-08-06
[lang] Add "On" back to strings.po

BigNoid:
It was removed in https://github.com/xbmc/xbmc/commit/25718f4ccebd227eaf63f0f8b53f0552f89f01e5
but it's still used in https://github.com/xbmc/xbmc/blob/master/xbmc/guilib/GUIRadioButtonControl.cpp#L101


HitcherUK:
@BigNoid 'Off' is still broken even after this commit.

PR #10239, opened 2016-08-06
add missing part of the curl 7.50 patch

BlackIkeEagle:
PR #10160 was not properly merged it seems, the
xbmc/filesystem/CurlFile.h change was missing


stefansaraev:
you should consider 7.50.0 broken and not use it, better upgrade to 7.50.1 sooner than later.

#10160 will work with curl >= 7.50.1 and <= 7.49.1. (I've build and runtime tested both)

BlackIkeEagle:
ok I'll keep it local until there is curl 7.50.1 available in our repo's

PR #10240, opened 2016-08-06
[PVR] Fix crash during shutdown due to wrong usage of std::remove

Paxxi:
@ksooo can you try it out?


ksooo:
works like it should now.

ksooo:
jenkins build this please

PR #10241, opened 2016-08-06
[pvr/cleanup/improvement] Some minor things I found while researching

Paxxi:
Observer crash in pvr

### DIFF
--- a/xbmc/epg/EpgContainer.cpp
+++ b/xbmc/epg/EpgContainer.cpp
@@ -65,7 +65,10 @@ CEpgContainer::CEpgContainer(void) :
 
 CEpgContainer::~CEpgContainer(void)
 {
-  Unload();
+  // Don't call Unload here, it'll try
+  // to notify observers which may be in
+  // different states of teardown
+  Stop();
 }
 
 CEpgContainer &CEpgContainer::GetInstance()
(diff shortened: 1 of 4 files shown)


xbmc/epg/EpgContainer.cpp
```diff
@@ -65,7 +65,10 @@ CEpgContainer::CEpgContainer(void) :
 
 CEpgContainer::~CEpgContainer(void)
 {
-  Unload();
+  // Don't call Unload here, it'll try
+  // to notify observers which may be in
+  // different states of teardown
+  Stop();
```
ksooo: Imo, reaching dtor does not necessarily mean Kodi "shutdown". Thus I'm not sure this is okay.

Unload() does much more than Stop().

ksooo:
Re: "Remove strange circular dependency way to trigger an epg update": good catch!

xbmc/epg/EpgContainer.cpp
Paxxi: Unload() does more stuff but when looking at it, none of it seemed important in a destructor as all objects will clean themselves up. I may be mistaken but it looked perfectly safe to remove the call to it.

xbmc/pvr/PVRManager.cpp
```diff
@@ -365,6 +365,7 @@ void CPVRManager::SetState(ManagerState state)
     CSingleLock lock(m_managerStateMutex);
     m_managerState = state;
     SetChanged();
+    TriggerEpgsCreate();
```
ksooo: Calling this on every (!) pvr manager state change looks horribly wrong to me (although in fact the existing code does eaxtly this). So, your change does not change behavior, but I doubt that the existing behavior is okay. We need to find a good condition for when to call TriggerEpgsCreate...

PR #10242, opened 2016-08-06
[xbmc/cleanup] Remove some win32 hard coded defines that cmake handles

Paxxi:


MartijnKaijser:
jenkins build this please

fetzerch:
+1, but basically all occurrences of BUILDING_WITH_CMAKE could go (it's defined for win only and appears in a few more places).

Paxxi:
They're all on the todo-list to get dropped but it can wait for now

PR #10243, opened 2016-08-06
Resolution: Choose most matching refreshrate

fritsch:
Admiral Schneider on the forum had an issue with his new 120 hz TV. This TV also has a 119.8804 mode which is used for 23.97608 content. Because our metric always ends with a distance measure of 0.0 if (refreshrate / video_fps) / round(refreshrate / video_fps) == 0.0. So basically every division without rest gets this 0.0

As we render with refreshrate 120 hz and this puts immense load for 24.0 fps content onto GPUs, we shall punish modes so that 24.0 gets a lower measure than 120 fps.

Another example: 30 fps
TV has: 30 hz, 60 hz and 120 hz

Before this PR:

w(30) == w(60) == w(120) = 0.0

After this PR:

w(30) < w(60) < w(120)

but still:

w(60) < w(59.94) < w(118.88)

### DIFF
--- a/xbmc/guilib/Resolution.cpp
+++ b/xbmc/guilib/Resolution.cpp
@@ -303,8 +303,21 @@ float CResolutionUtils::RefreshWeight(float refresh, float fps)
   float div   = refresh / fps;
   int   round = MathUtils::round_int(div);
 
+  float weight = 0.0f;
+
   if (round < 1)
-    return (fps - refresh) / fps;
+    weight = (fps - refresh) / fps;
   else
-    return (float)fabs(div / round - 1.0);
+    weight = (float)fabs(div / round - 1.0);
+
+  // punish higher refreshrates and prefer better matching
+  // e.g. 30 fps content at 60 hz is better than
+  // 30 fps at 120 hz - as we sometimes don't know if
+  // the content is interlaced at the start, only
+  // punish when refreshrate > 60 hz to not have to switch
+  // twice for 30i content
+  if (refresh > 60 && round > 1)
+    weight += round / 1000.0;
+
+  return weight;
 }


popcornmix:
I used to do this in Pi builds, but there was an issue with interlaced video.

They tend to vary between 25 and 50 (or 30 and 60) with different deinterlace modes.
You'll often start playback with 25fps and detect interlaced after the first frame is decoded and then the fps is doubled when deinterlace is activated.
This scheme is likely to be switching HDMI mode more often.

Special casing 25 and 30/29.97 to prefer 2x the framerate would be one way of avoiding this issue, but I guess @FernetMenta would be better placed to comment.

fritsch:
@popcornmix In fact I talked with @FernetMenta before that and he told me "we can get interlaced content detected properly", but yes - you are right - we need something ontop of that most likely.

fritsch:
Easy fix however: http://sprunge.us/MQcU but I would consider this a hack :-)

FernetMenta:
> Special casing 25 and 30/29.97 to prefer 2x the framerate would be one way of avoiding this issue

It is not on renderer to decide but player.

fritsch:
Yeah, but we need a player change. If you play 1080i50 player does not know about the 50 hz at the beginning, so it will switch to 25 hz first and after some seconds it will switch to 50hz mode.

FernetMenta:
> Yeah, but we need a player change

a tiny one. if it is interlaced, it can't be 25hz

fritsch:
@FernetMenta two hacks added - I am not sure how to solve it properly.

FernetMenta (commented):
drop this please. the strategy for videoplayer is making to kill all dependencies to pvr. this change does not cover all cases anyway

fritsch:
force pushed

FernetMenta:
the comment was for 2nd commit, don't know if github shows this correctly

xbmc/guilib/Resolution.cpp
```diff
@@ -303,8 +303,21 @@ float CResolutionUtils::RefreshWeight(float refresh, float fps)
...
-    return (float)fabs(div / round - 1.0);
+    weight = (float)fabs(div / round - 1.0);
+
+  // punish higher refreshrates and prefer better matching
+  // e.g. 30 fps content at 60 hz is better than
+  // 30 fps at 120 hz - as we sometimes don't know if
+  // the content is interlaced at the start, only
+  // punish when refreshrate > 60 hz to not have to switch
+  // twice for 30i content
+  if (refresh > 60 && round > 1)
```
FernetMenta: lets just ignore all resolutions > 60

fritsch:
Not possible. Some tvs only have 120 and 119.9x

Did I remove the right one?

Am 01.10.2016 7:47 vorm. schrieb "Rainer Hochecker" <
notifications@github.com>:

FernetMenta:
If tvs have only a single resolution, we can't switch anyway and stay at desktop. So it is possible to ignore all > 60

FernetMenta:
+1 after  discussion with @fritsch on slack

fritsch:
jenkins build this please

PR #10245, opened 2016-08-06
Estuary: Fix Paused label (seeklabel variable)

wiromare:
see here #10129 for more info

p.s.
I found another strange behavior, the new dialog player process info disapear while moving the mouse, since this can be connected on how to show livetv info etc etc... I prefer only to report

@BigNoid


BigNoid:
Thx, but this fix is already in estuary repo and will be included in xbmc master next skin sync.

MartijnKaijser:
see that's what i hate about having two repos. This should stop imo

ksooo:
Absolutely.

PR #10247, opened 2016-08-07
VideoPlayer: passthrough fixes

FernetMenta:
see title

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Audio/DVDAudioCodecPassthrough.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Audio/DVDAudioCodecPassthrough.cpp
@@ -27,6 +27,10 @@
 
 #include "cores/AudioEngine/AEFactory.h"
 
+extern "C" {
+#include "libavcodec/avcodec.h"
+}
+
 #define TRUEHD_BUF_SIZE 61440
 
 CDVDAudioCodecPassthrough::CDVDAudioCodecPassthrough(CProcessInfo &processInfo) :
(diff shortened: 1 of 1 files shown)


FernetMenta:
jenkins build this please

MilhouseVH:
@FernetMenta: This introduces a massive memory leak when seeking during playback of certain files. Reverting the change [fixes the leak](http://forum.kodi.tv/showthread.php?tid=269815&pid=2402265#pid2402265), so it's definitely the cause.

Sample file in this [post](http://forum.kodi.tv/showthread.php?tid=269814&pid=2398013#pid2398013). Enable passthrough, play sample file, seek forwards or backwards -> freeze then crash due to OOM.

No crash with passthrough disabled. Reproducible on both RPi2 (LibreELEC) and x86 (LibreELEC), and - probably - [Lubuntu](http://forum.kodi.tv/showthread.php?tid=288754).

MartijnKaijser:
Same on Android when i skip it quits to Android main screen just after or during the buffering circle. A build with just this PR reverted doesn't show this sympton.

xbmc/cores/VideoPlayer/DVDCodecs/Audio/DVDAudioCodecPassthrough.cpp
```diff
@@ -127,12 +132,25 @@ int CDVDAudioCodecPassthrough::Decode(uint8_t* pData, int iSize, double dts, dou
...
+  if (pData)
+  {
+    AVPacket pkt;
+    av_init_packet(&pkt);
+    pkt.data = pData;
+    pkt.size = iSize;
+    av_packet_split_side_data(&pkt);
+    skip = iSize - pkt.size;
+    pData = pkt.data;
+    iSize = pkt.size;
```
fritsch: I think you miss an av_packet_free_side_data here, as av_packet_split_side_data mallocs a buffer to copy the sidedata?

PR #10248, opened 2016-08-07
[PVR] CDirectoryProvider now supports async PVR startup. Fixes favori…

ksooo:
…te tv and radio channels not appearing on Home screen, sort method and order, item re-selection after update, start and end time when no epg.

Currently, Estuary's favorite tv and radio channels widget gets not filled on initial Kodi startup. One needs to open any other window and to go back to home screen to get the widgets populated. This is due to the related directoty list providers not respecting asynchronous pvr startup.

The problem is similar to and the solution is logically the same as for #10203  

Important: This also needs a skin fix to work properly.

@da-anda fyi
@Jalle19 for review?

### DIFF
--- a/xbmc/listproviders/DirectoryProvider.cpp
+++ b/xbmc/listproviders/DirectoryProvider.cpp
@@ -33,6 +33,7 @@
 #include "music/dialogs/GUIDialogMusicInfo.h"
 #include "music/MusicThumbLoader.h"
 #include "pictures/PictureThumbLoader.h"
+#include "pvr/PVRManager.h"
 #include "settings/Settings.h"
 #include "threads/SingleLock.h"
 #include "utils/JobManager.h"
@@ -48,6 +49,7 @@
 using namespace XFILE;
 using namespace ANNOUNCEMENT;
 using namespace KODI::MESSAGING;
+using namespace PVR;
 
 class CDirectoryJob : public CJob
 {
@@ -263,7 +265,7 @@ void CDirectoryProvider::Fetch(std::vector<CGUIListItemPtr> &items) const
   }
 }
 
-void CDirectoryProvider::OnEvent(const ADDON::AddonEvent& event)
+void CDirectoryProvider::OnAddonEvent(const ADDON::AddonEvent& event)
 {
   CSingleLock lock(m_section);
   if (URIUtils::IsProtocol(m_currentUrl, "addons"))
@@ -272,7 +274,20 @@ void CDirectoryProvider::OnEvent(const ADDON::AddonEvent& event)
         typeid(event) == typeid(ADDON::AddonEvents::Disabled) ||
         typeid(event) == typeid(ADDON::AddonEvents::InstalledChanged) ||
         typeid(event) == typeid(ADDON::AddonEvents::MetadataChanged))
-    m_updateState = INVALIDATED;
+      m_updateState = INVALIDATED;
+  }
+}
+
+void CDirectoryProvider::OnPVRManagerEvent(const PVR::ManagerState& event)
+{
+  CSingleLock lock(m_section);
+  if (URIUtils::IsProtocol(m_currentUrl, "pvr"))
+  {
+    if (event == ManagerStateStarted ||
+        event == ManagerStateStopped ||
+        event == ManagerStateError ||
+        event == ManagerStateInterrupted)
+      m_updateState = INVALIDATED;
   }
 }
 
@@ -300,6 +315,7 @@ void CDirectoryProvider::Reset(bool immediately /* = false */)
       m_isAnnounced = false;
       CAnnouncementManager::GetInstance().RemoveAnnouncer(this);
       ADDON::CAddonMgr::GetInstance().Events().Unsubscribe(this);
+      g_PVRManager.Events().Unsubscribe(this);
     }
   }
 }
@@ -393,7 +409,8 @@ bool CDirectoryProvider::UpdateURL()
   {
     m_isAnnounced = true;
     CAnnouncementManager::GetInstance().AddAnnouncer(this);
-    ADDON::CAddonMgr::GetInstance().Events().Subscribe(this, &CDirectoryProvider::OnEvent);
+    ADDON::CAddonMgr::GetInstance().Events().Subscribe(this, &CDirectoryProvider::OnAddonEvent);
+    g_PVRManager.Events().Subscribe(this, &CDirectoryProvider::OnPVRManagerEvent);
   }
   return true;
 }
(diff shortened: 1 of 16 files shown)


ksooo:
https://github.com/xbmc/skin.estuary/pull/180

ksooo:
jenkins build this please

xbmc/listproviders/DirectoryProvider.cpp
```diff
@@ -272,7 +274,16 @@ void CDirectoryProvider::OnEvent(const ADDON::AddonEvent& event)
...
-    m_updateState = INVALIDATED;
+      m_updateState = INVALIDATED;
+  }
+}
+
+void CDirectoryProvider::Notify(const Observable &obs, const ObservableMessage msg)
+{
+  if (URIUtils::IsProtocol(m_currentUrl, "pvr"))
+  {
+    if (msg == ObservableMessageManagerStateChanged && g_PVRManager.IsStarted())
```
tamland: Don't call g_PVRManager.IsStarted here please. The directory provider should not need to know about things like this; instead pvr should sent appropriate events.

xbmc/listproviders/DirectoryProvider.cpp
```diff
@@ -272,7 +274,16 @@ void CDirectoryProvider::OnEvent(const ADDON::AddonEvent& event)
         typeid(event) == typeid(ADDON::AddonEvents::Disabled) ||
         typeid(event) == typeid(ADDON::AddonEvents::InstalledChanged) ||
         typeid(event) == typeid(ADDON::AddonEvents::MetadataChanged))
-    m_updateState = INVALIDATED;
+      m_updateState = INVALIDATED;
+  }
+}
+
+void CDirectoryProvider::Notify(const Observable &obs, const ObservableMessage msg)
+{
```
tamland: missing lock

xbmc/listproviders/DirectoryProvider.cpp
```diff
@@ -387,13 +403,19 @@ bool CDirectoryProvider::UpdateURL()
...
+
   m_currentUrl = value;
 
   if (!m_isAnnounced)
   {
     m_isAnnounced = true;
     CAnnouncementManager::GetInstance().AddAnnouncer(this);
     ADDON::CAddonMgr::GetInstance().Events().Subscribe(this, &CDirectoryProvider::OnEvent);
+
+    if (URIUtils::IsProtocol(m_currentUrl, "pvr"))
```
tamland: This wont work. m_currentUrl can change, and if it does, it wouldn't register correctly (just do the same as the two other registration systems)

xbmc/listproviders/DirectoryProvider.cpp
ksooo: ooops.

xbmc/listproviders/DirectoryProvider.cpp
ksooo: I don't get your point. What exactly is *wrong* with calling g_PVRManager? Need to know this object and thus pvr details anyway in order to register as observer at the pvr manager singleton. Maybe it is not the nicest api, but this is how pvr manager works today - sending one state changed event and offering several methods to obtain the current state. I would like to keep this code as it is now and put pvr messaging rework on my list for v18.

tamland: I'm just trying to keep this clean from any dependency that isn't strictly registering and not leaking in details that doesn't have anything to do with the vfs from those singletons. Why does the directory provider need to know, or care, about which state pvr manager is in? Addons have similar state change events (InstalledChanged, MetadataChanged) but it doesn't call back into addonmgr.

xbmc/listproviders/DirectoryProvider.cpp
tamland: Right. Though, it would do an unnecessary unregister/register if url changes from pvr to a different pvr path.

ksooo: I'm sure I can optimize this. ;-)

xbmc/listproviders/DirectoryProvider.cpp
ksooo: I understand what you mean and your intention for sure is okay. But changing pvr messaging is high effort and i do not want to do this **now** as we are near beta 1 and this is highly sensitive pvr code (I never touched/read before).  As i wrote i will be happy to (completely) rework this during alpha phase for v18.

ksooo:
@tamland take 2. should be much better now. any objections?

ksooo:
jenkins build this please

da-anda:
only to have this future proof, can this event handling in future be extended to also update PVR channels f.e. on EPG updates, playcount updates of channels etc, or would this require some other man in the middel event handler?

ksooo:
@BigNoid are you fine with the Estuary fix?

BigNoid:
>  are you fine with the Estuary fix?

Yeah looks good. I'll rebase #10253 after this is in then.

xbmc/pvr/PVRManager.h
```diff
@@ -70,6 +71,21 @@ namespace PVR
...
   };
 
+  struct PVRManagerEvent
+  {
+    virtual ~PVRManagerEvent() {};
+  };
+
+  namespace PVRManagerEvents
+  {
+    struct Starting    : PVRManagerEvent {};
```
FernetMenta: I am wondering if polymorphism is the right concept here.

ksooo: @FernetMenta ask @tamland. he wanted me to do it like he does with addons (at least this is what i understood from his review).

ksooo: i think the idea is that if an event is a struct it can carry aditional data. i do not use this this for pvr manager, but some of the addon events do.

ksooo: but yeah, i guess i will simplify this. if we need to carry data along with the events, we easily can change that later.

ksooo:
@FernetMenta addressed your review comment. thx.

FernetMenta:
thanks, imo much better. as you said, in case some future requirements need a change, it still can be changed.

Jalle19:
Looks good to me

ksooo:
jenkins build this please

ksooo:
jenkins build this please

ksooo:
jenkins build this please

ksooo:
rebased

ksooo:
heads-up: if nobody objects i will merge this tomorrow. :-)

ksooo:
jenkins build this please

da-anda:
I've seen @MilhouseVH included it in his nightlies, so maybe wait one or two days if regressions are poping up in these build and then merge?

ksooo:
final (?) rebase

ksooo:
jenkins build this please

ksooo:
No idea why jenkins failed on osx, but reason is not related to this PR, I guess.

ksooo:
last rebase before merge
jenkins build this please

ksooo:
no complaints from Milhouse build users...

PR #10249, opened 2016-08-07
[upnp] startup honors user settings.

MartijnKaijser:
Patch from @Rallymen007
https://github.com/xbmc/xbmc/pull/10072

Ping @Montellese

### DIFF
--- a/xbmc/network/NetworkServices.cpp
+++ b/xbmc/network/NetworkServices.cpp
@@ -827,9 +827,20 @@ bool CNetworkServices::StartUPnP()
   bool ret = false;
 #ifdef HAS_UPNP
   ret |= StartUPnPClient();
-  ret |= StartUPnPServer();
-  ret |= StartUPnPController();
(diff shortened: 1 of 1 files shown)


MartijnKaijser:
jenkins build and merge

Montellese:
Sorry for the (too) late reply. To match our coding style the curly braces should be removed.

And for the upnp controller to work the server has to be running as well i.e. the controller must only be started if the server is enabled as well. That's also how it is in the settings i.e. the controller setting is greyed out if the server isn't enabled.

MartijnKaijser:
oops sorry. Best revert this instead of adjusting?

Rallymen007:
These should have been fixed, the content of the file in commit 852f4d7 seems correct to me

Rallymen007:
Oh no I got it wrong, sorry.

I can do it either way

Montellese:
Follow-up PR is enough.

PR #10251, opened 2016-08-07
VideoPlayer: rework rtmp options for ffmpeg-demuxer

FernetMenta:
rework after https://github.com/xbmc/xbmc/pull/10015

in particular fixes non stripped http options from http urls


FernetMenta:
jenkins build this please

PR #10252, opened 2016-08-07
[settings] make skin settings button more obvious

jjd-uk:
There's been a number of comments over the years that access to the skin specific settings is not obvious enough to a new users, thus they miss out on the options to configure what's shown on main menu for example.

I've tried to do this in as low impact way as possible, so as I see it there are two options:

1) A simple labelling to "Configure skin settings.." so the wording points it's something to be selected, plus the UI convention of have three trailing dots "..." to suggest there's something more.

image

2) Move this setting to it's own Group then similar to above have a "Configure..." label for the button to suggest something to be selected.

image

For easy comparison this PR currently contains both options.

As well as trying to make access to these settings more obvious in "Interface" -> "Skin", I've personally got a dislike for Estuary having skin settings in two locations, and I can see that being a source of user confusion as to why we have 

image

as well as "Interface" -> "Skin". Thus if this goes in using one of the options I've given, then there's the possiblity the "Skin settings" shortcut is no longer need in the main settings window. That decision would of course be up to the skinners.

What are peoples thought? @MartijnKaijser @phil65 @ronie @BigNoid @HitcherUK

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -7771,7 +7771,12 @@ msgctxt "#14260"
 msgid "Debug"
 msgstr ""
 
-#empty strings from id 14261 to 14269
+#: system/settings/settings.xml
+msgctxt "#14261"
+msgid "Configure skin..."
+msgstr ""
+
+#empty strings from id 14262 to 14269
 
 #: system/settings/settings.xml
 msgctxt "#14270"
@@ -14685,6 +14690,7 @@ msgid "Manage dependencies"
 msgstr ""
 
 #: xbmc/filesystem/AddonsDirectory.cpp
+#: system/settings/settings.xml
 msgctxt "#24997"
 msgid "Look and feel"
 msgstr ""
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -3016,7 +3016,7 @@
   </section>
   <section id="interface" label="14206" help="38102">
     <category id="skin" label="166" help="36102">
-      <group id="1" label="16000">
+      <group id="1" label="24997">
         <setting id="lookandfeel.skin" type="addon" label="166" help="36103">
           <level>0</level>
           <default>skin.estuary</default>
@@ -3027,7 +3027,7 @@
             <show more="true" details="true">installed</show>
           </control>
         </setting>
-        <setting id="lookandfeel.skinsettings" type="action" parent="lookandfeel.skin" label="21417" help="36104">
+        <setting id="lookandfeel.skinsettings" type="action" parent="lookandfeel.skin" label="14261" help="36104">
           <level>0</level>
           <dependencies>
             <dependency type="enable" on="property" name="AddonHasSettings" setting="lookandfeel.skin" />
@@ -3046,7 +3046,7 @@
           <control type="list" format="string" />
         </setting>
         <setting id="lookandfeel.skincolors" type="string" parent="lookandfeel.skin" label="14078" help="36106">
-          <level>2</level>
+          <level>1</level>
           <default>SKINDEFAULT</default>
           <constraints>
             <options>skincolors</options>
@@ -3057,7 +3057,7 @@
           <control type="list" format="string" />
         </setting>
         <setting id="lookandfeel.font" type="string" parent="lookandfeel.skin" label="13303" help="36107">
-          <level>2</level>
+          <level>1</level>
           <default>Default</default>
           <constraints>
             <options>skinfonts</options>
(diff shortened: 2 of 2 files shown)


BigNoid:
It makes sense to remove the shortcut from the main settings screen. It's confusing to have it at two places. No strong opinion on renaming or moving it, but having a group called skin settings might be double as we are already in the skin section of the settings.

Btw, theres a third option to only show the shortcut in the main settings screen by removing skinsettings.xml from estuary and switch to a custom dialog. The settings button won't be shown or will be disabled in that case iirc, but I am personally not in favor of that.

jjd-uk:
What I wanted mainly was feedback on was whether either idea was acceptable, and then address any concerns. For example on your point the labels can easily be changed if that was the preferred way to do this, so instead maybe a Group label of "Skin specific" then a "Configure settings" label for the button.

As for your third option, the primary reason for this PR is that all skins would benefit, anything Estuary specific could be decided separately.

jjd-uk:
Anymore thoughts on this from anyone? no big deal waiting for v18 but thought it might be a nice to have low impact improvement for v17.

da-anda:
I'm in favor of removing the custom skin settings button in settings home screen. Mentioned it the first second it showed up, but phil wanted to have and keep it.

As for the two options, I'm for 1). Either label is fine with me. As another label alternative, maybe `Customize skin`.

I also just noticed, that "zoom" is not really skin specific, but UI ingeneral, so the indention doesn't make much sense, does it?

PS; I personally hate the three dots, worst usability ever. These elements should be styled by the skin f.e. as button, but this requires core changes and is a different, general issue.

jjd-uk:
So is this wanted at all? if so do we want it for Krypton allowing Estuary to drop the Skin settings shortcut from the main Settings view? and if wanted should I add the commit here dropping the shortcut from Estuary?

I ask because I should have the time to get this into a mergeable state tomorrow.

ronie:
> There's been a number of comments over the years that access to the skin specific settings is not obvious enough to a new users, thus they miss out on the options to configure what's shown on main menu for example.

that was the case with Confluence probably?

since it's now listed on the main settings screen (in Estuary) i would be hard to miss, even for new users. for this reason, i would keep it there.

jjd-uk:
@ronie to my mind it could be very confusing to users to duplicate access to skin settings in this manner in Estuary as it means we now have 2 areas in the settings titled "Skin".

da-anda:
we removed it from settings home screen for a reason back then (to have all UI/skin related settings in one place IIRC) and I think we should keep it that way -> remove from Estuary settings screen

jjd-uk:
Forget about Estuary for a moment, primary purpose is to make skins settings option more obvious in any skin.

Following feedback this is latest iteration in:

Confluence
image

Aeon Nox
image

Rapier
image

system/settings/settings.xml
```diff
@@ -3022,14 +3022,14 @@
...
-        <setting id="lookandfeel.skinsettings" type="action" parent="lookandfeel.skin" label="21417" help="36104">
+        <setting id="lookandfeel.skinsettings" type="action" parent="lookandfeel.skin" label="14266" help="36104">
           <level>0</level>
           <dependencies>
             <dependency type="enable" on="property" name="AddonHasSettings" setting="lookandfeel.skin" />
           </dependencies>
           <control type="button" format="action" />
         </setting>
-        <setting id="lookandfeel.skintheme" type="string" parent="lookandfeel.skin" label="15111" help="36105">
+        <setting id="lookandfeel.skintheme" type="string" label="15111" help="36105">
```
MartijnKaijser: why are you removing the parents?

jjd-uk: I was looking for a variation on option 1) to make "- Configure skin..." stand out better without resorting to using a separate group as in option 2) above.

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -7752,7 +7752,17 @@ msgctxt "#14260"
 msgid "Debug"
 msgstr ""
 
-#empty strings from id 14261 to 14269
+#empty strings from id 14261 to 14264
+
+msgctxt "#14265"
+msgid "Look & Feel"
```
MartijnKaijser: We already have such a label which is used in addon manager

jjd-uk:
Bumping I still think this is a useful change to make the Skin settings button stand out more, but maybe it's something to be discussed over the weekend at Devcon. I didn't see any particular need to keep the parent relationship for theme, colours, fonts, & zoom, and removing those makes the Skin settings button stand out more in my opinion, but will put those parent relationships back if that gets this in.

Please ignore the skin settings shortcut question for Estuary for the moment, that can be discussed separately, perhaps again over the weekend.

NedScott:
This really is needed, and dates back to when the setting was first moved to a "sub-window" of appearance settings. It doesn't look like anyone objects to the idea, and the only discussion was about the mostly-unrelated Estuary-specific button.

I will paypal someone beer money to push the button on this.

HitcherUK:
Can someone remind me again why skins are the only addon that doesn't use the addon settings dialog?

MartijnKaijser:
Legacy and time :)

jjd-uk:
image

@MartijnKaijser Anything else needed for this?

PR #10254, opened 2016-08-07
[cmake] Update Effects11 and Build it with CMake instead VS

fetzerch:
ping: @Paxxi, @afedchin


xbmc/filesystem/CMakeLists.txt
```diff
@@ -180,7 +180,7 @@ if(BLURAY_FOUND)
                       BlurayFile.h)
 endif()
 
-core_add_library(filesystem upnp)
+core_add_library(filesystem)
```
MartijnKaijser: Seems not related?

afedchin:
did you update fx11 from upstream?

Paxxi:
@afedchin it's updated with the latest upstream release from August 2

afedchin:
oki. that's fine for me. thanks

xbmc/filesystem/CMakeLists.txt
fetzerch: moved to separate commit

fetzerch:
@Paxxi: I think it would be still good to get this in, even if we replace it with a prebuilt version from depends at some point.

With the changes in https://github.com/xbmc/xbmc/pull/10256 appveyor fails on the effects11 solution: https://ci.appveyor.com/project/MartijnKaijser/xbmc/build/1.0.2827.

Jenkins build this please

Paxxi:
agreed, I also wanted this in for the x64 work because it also got stuck on building this

PR #10255, opened 2016-08-07
[depends/win32] Update mini wdk with files from 10.0.14393.0 as

Paxxi:
there's some new defines that break the build


Paxxi:
jenkins build this please

PR #10256, opened 2016-08-07
[cmake] Modernize find modules and show less verbose output

fetzerch:
Modernize Find Modules (according to guideline: https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#find-modules)
- Don't trust pkg-config, use it as hint and ensure that the things it suggests exist.
- Show versions
- Don't cache things that shouldn't be cached

It also shows less verbose output when not building with VERBOSE=1. 
- output before: http://paste.ubuntu.com/22462822/
- output after: http://paste.ubuntu.com/22607511/

In addition it fixes a warning due to breakpad being removed and an issue with depends on linux/rbpi that could lead to host files being picked up. (@wsnipex might want to check 2be1703).

Still todo:
- use ffmpeg from depends and don't build a second time
- fix FindMMAL


wsnipex:
looks good to me

fetzerch:
Thx, will update the remaining changes in separate PRs. This PR now also fixes the Appveyor build.

Jenkins build this please

fetzerch:
One thing I forgot:
@Fneufneu: If you get some errors or libs or dependencies not found when you run cmake on BSD, please ping me.

PR #10258, opened 2016-08-08
VideoPlayer: drop IsCaching, useless and not thread safe

FernetMenta:
see title

@HitcherUK @BigNoid @ronie @phil65 
the gui label will be still there for a while. I suggest that you drop usage of this label from skins.

### DIFF
--- a/xbmc/ApplicationPlayer.cpp
+++ b/xbmc/ApplicationPlayer.cpp
@@ -319,12 +319,6 @@ int64_t CApplicationPlayer::GetTime() const
     return 0;
 }
 
-bool CApplicationPlayer::IsCaching() const
-{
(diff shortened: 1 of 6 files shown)


FernetMenta:
jenkins build this please

MartijnKaijser:
Will the label just be empty? Maybe just drop it simultaneously then if is has no negative impact on skins.

FernetMenta:
actually it is/was a bool that now returns always false.

FernetMenta:
Just to prevent from misunderstandings. This change has no negative impact on skins in terms of empty labels.

FernetMenta:
caching == buffering,  I mixed this up.
I will make it thread safe. Sorry for confurion

PR #10260, opened 2016-08-08
respect start_pts for omx_pts (32-bit) handling in aml

ghost:
AMlogics omx_pts is 32 bit int. Streams with large pts values will fail, because kodi uses internally 64 bit pts. This PR fixes the PTS issue by storing the start_pts value and adding / removing this offset at the interface kodi / aml.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
@@ -401,23 +401,6 @@ void dumpfile_write(am_private_t *para, void* buf, int bufsiz)
     write(para->dumpfile, buf, bufsiz);
 }
 
-static int set_pts_pcrscr(int64_t value)
-{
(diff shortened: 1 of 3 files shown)


stefansaraev:
https://github.com/xbmc/xbmc/commit/59828832455794019c59d46e931b1f96efa312f5#diff-4e99a88003816cb71e489e8a0db42345R1922

cc @codesnake

mapfau:
I have recognized that decoder seems to eat too much frames. I'm not sure if this problem is because of this commit, but I run often out of frames... I have to check if the 2 sec. feed logic is broken now after these changes.....

mapfau:
Now its correct playing on my side.

O.T.:: I pushed today a fix to hardkernel for odroid c1/c1+ wich enables omx_pts in the amogic video driver. Its running well now on my odrois C1. Thx, @codesnake .

codesnake:
Works fine on my side. Thanks @mapfau.

MrMC:
Nice,  after 4+ years, AML finally understands that pts needs to be exposed as 64 bit :)

PR #10261, opened 2016-08-08
Fix joystick input while app is in background

garbear:
This fixes joysticks being able to control Kodi when the app is unfocused.

A second fix is included. When Kodi's main thread freezes for > 0.5s, spurious button press events can appear. This was fixed by factoring the button handler to avoid using a separate thread.


garbear:
jenkins build this please

PR #10264, opened 2016-08-10
[PVR] Reset ischannelpreview flag when pvr stream gets closed. Fixes trac#16840.

ksooo:
Regression introduced by #10229 
Fixes http://trac.kodi.tv/ticket/16840

@FernetMenta for review?
@axbmcuser maybe you can runtime test?

### DIFF
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -10959,12 +10959,21 @@ bool CGUIInfoManager::ConditionsChangedValues(const std::map<INFO::InfoPtr, bool
 
 bool CGUIInfoManager::IsPlayerChannelPreviewActive() const
 {
-  bool streamValid = m_videoInfo.valid;
-  if (m_currentFile->HasPVRChannelInfoTag() && m_currentFile->GetPVRChannelInfoTag()->IsRadio())
-    streamValid = m_audioInfo.valid;
-
-  return m_playerShowInfo &&
-         (m_isPvrChannelPreview || !streamValid);
+  bool bReturn(false);
+  if (m_playerShowInfo)
+  {
+    if (m_isPvrChannelPreview)
+    {
+      bReturn = true;
+    }
+    else
+    {
+      bReturn = !m_videoInfo.valid;
+      if (bReturn && m_currentFile->HasPVRChannelInfoTag() && m_currentFile->GetPVRChannelInfoTag()->IsRadio())
+        bReturn = !m_audioInfo.valid;
+    }
+  }
+  return bReturn;
 }
 
 CEpgInfoTagPtr CGUIInfoManager::GetEpgInfoTag() const
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -9229,6 +9229,15 @@ bool CGUIInfoManager::GetDisplayAfterSeek()
 void CGUIInfoManager::SetShowInfo(bool showinfo)
 {
   m_playerShowInfo = showinfo;
+
+  if (!showinfo)
+    m_isPvrChannelPreview = false;
+}
+
+bool CGUIInfoManager::ToggleShowInfo()
+{
+  SetShowInfo(!m_playerShowInfo);
+  return m_playerShowInfo;
 }
 
 void CGUIInfoManager::Clear()
--- a/xbmc/GUIInfoManager.h
+++ b/xbmc/GUIInfoManager.h
@@ -200,7 +200,7 @@ public:
   void SetShowTime(bool showtime) { m_playerShowTime = showtime; };
   void SetShowInfo(bool showinfo);
   bool GetShowInfo() const { return m_playerShowInfo; }
-  bool ToggleShowInfo() { m_playerShowInfo = !m_playerShowInfo; return m_playerShowInfo; };
+  bool ToggleShowInfo();
   bool IsPlayerChannelPreviewActive() const;
 
   std::string GetSystemHeatInfo(int info);
--- a/xbmc/pvr/PVRManager.cpp
+++ b/xbmc/pvr/PVRManager.cpp
@@ -1178,6 +1178,7 @@ void CPVRManager::CloseStream(void)
     g_application.SaveFileState();
   }
 
+  m_isChannelPreview = false;
   m_addons->CloseStream();
   SAFE_DELETE(m_currentFile);
 }
--- a/xbmc/pvr/PVRManager.cpp
+++ b/xbmc/pvr/PVRManager.cpp
@@ -1227,6 +1227,7 @@ void CPVRManager::UpdateCurrentChannel(void)
     delete m_currentFile;
     m_currentFile = new CFileItem(playingChannel);
     UpdateItem(*m_currentFile);
+    m_isChannelPreview = false;
   }
 }
 


axbmcuser:
Yes. Just did some quick tests.
It looks like the problem is partly fixed - especially the crash seems to be gone.

What remains are problems within LiveTV-playback itself, again, best reproduced with Confluence:

- Start LiveTV playback
- Press "I" for VideoFullScreen INFO (_without_ player buttons!)
- Press "UP" a few times on the arrow keys (causing channelpreviewactive to be true, which is correct at this point, other channels are previews)
- While GUI shows one of those previewed channels (don't confirm the switch!) press "I" to close VideoFullScreen INFO
- Normal LiveTV-playback continues, GUI is now closed
-  Press "I" for VideoFullScreen INFO again (_without_ player buttons!)
- Problem: channelpreviewactive still is true, although the INFO shows the active channel playing and you are not in the process doing channelpreview

Thanks!

ksooo:
Second commit should fix also the other problem reported.

axbmcuser:
Quick tests with the additional second commit lead to strange results for me:
Like INFO dialog closing by itself after a short while and acting strangely. Did not change anything else in the build compared to compiling with only the first commit included. Will try to test some more as soon as i can.

axbmcuser:
Just some quick feedback without being able to do extensive tests regarding the mentioned weird behaviour of the LiveTV-GUI:

(Follow exact steps - maybe _not_ as easily reproducible as before and doesn't fail every time (Still: Confluence)):
- Start LiveTV playback
- Press "I" for VideoFullScreen INFO (_without_ player buttons!)
- Press "UP" a few times on the arrow keys (causing channelpreviewactive to be true, which is correct at this point, other channels are previewed within the GUI)
- While GUI shows one of those previewed channels (don't confirm the switch!) press "I" to close VideoFullScreen INFO
- Normal LiveTV-playback continues, GUI is now closed
-  Press "I" to open VideoFullScreen INFO again (_without_ player buttons!)
- Press "UP" a fast few times on the arrow keys (causing channelpreviewactive to be true, which is correct at this point, other channels are previewed within the GUI)
IMPORTANT: the following step has to follow very very closely to the step before. Means: Pressing the arrow key UP the last time has to be followed instantly by pressing "I" (!)
- While GUI shows one of those previewed channels (don't confirm the switch!) press "I" to close VideoFullScreen INFO
- Normal LiveTV-playback continues, GUI is now closed
- Press "I" for VideoFullScreen INFO (_without_ player buttons!)
- Problem: 
INFO GUI still shows old channel preview data, after a short delay it flickers back to the correct channel which is playing. Sometimes the just opened INFO dialog (without player buttons) even closes by itself.

Sorry that it's so complicated, but maybe it helps a bit.

FernetMenta:
I will think about it in more detail this evening.

ksooo:
The last problem with quickly reopening the info osd after closing it in preview mode is a known problem and not related to the recent channel preview changes or this PR. You should be able to reproduce with alpha 3 or even Jarvis. I know the root cause for this. A fix is imo not trivial.

Please, let's concentrate in this PR on fixing the regression introduced by recent changes. There are old bugs with channel preview, which should not be mixed in here.

axbmcuser:
Did some testing on Jarvis, thanks for the hint. Now i can differ the new from the old bugs. 
Since i barely use LiveTV and had disabled channelpreviews in the past, i never stumbled upon it. 
Looks like the new " channelpreviewactive-bug" is fixed with the two commits discussed. If i notice otherwise, i'll let you know. Thanks.

ksooo:
Many thanks for reporting the issue and for testing. Much appreciated.

xbmc/GUIInfoManager.cpp
```diff
@@ -9281,9 +9281,8 @@ void CGUIInfoManager::UpdateAVInfo()
 
       m_videoInfo = video;
       m_audioInfo = audio;
-
-      m_isPvrChannelPreview = g_PVRManager.IsChannelPreview();
     }
+    m_isPvrChannelPreview = g_PVRManager.IsChannelPreview();
```
FernetMenta: this is not ideal for 2 reasons

1) might bring the flickering back. updating a/v streams and preview need to be in sync.
2) it would call IsChannelPreview on every render cycle, even if the INFO is not active

How about setting IsChannelPreview to false in CGUIInfoManager::SetShowInfo if m_playerShowInfo is set to false?

ksooo: > How about setting IsChannelPreview to false in CGUIInfoManager::SetShowInfo if m_playerShowInfo is set to false?

Yeah that works - if CGUIInfoManager::ToggleShowInfo also resets m_isPvrChannelPreview. I've just learned that ToggleInfo is called when hitting "I". Earlier today I tried what you suggested (yeah had the same idea!), but only in SetShowInfo. That was obviously not sufficient...

ksooo:
@FernetMenta commit 2 updated as you suggested.
@axbmcuser mind retesting?

xbmc/GUIInfoManager.cpp
```diff
@@ -9229,6 +9229,15 @@ bool CGUIInfoManager::GetDisplayAfterSeek()
 void CGUIInfoManager::SetShowInfo(bool showinfo)
 {
   m_playerShowInfo = showinfo;
+
+  if (!showinfo)
+    m_isPvrChannelPreview = false;
+}
+
+bool CGUIInfoManager::ToggleShowInfo()
```
FernetMenta: good idea to move the body here. much clearer now.

axbmcuser:
Just compiled with commit 1 and 2(updated). Just saw commit 3. Recompiling including commit 3 now. Will retest afterwards - or are there more changes to wait for?

ksooo:
No, that should be it.

axbmcuser:
All 3 commits included, i did not notice any new problems being introduced while making non-extensive tests. From user point of view these variants now seem to work equally:
- commit 1 + 2 included
- commit 1 + 2(updated after FernetMenta suggestions) included
- comit 1 + 2(updated) + 3 included

Old bugs ignored this time, of course.

ksooo:
Cool thanks.

ksooo:
jenkins build this please

ksooo:
travis error due to job timeout; not related to this PR.

axbmcuser:
Update side note: Just noticed that there may be some side effects left regarding the ischannelpreview-issues.

Having mouse support always disabled, i noticed that having channelpreview in System -> TV -> Playback enabled causes MusicOSD and "Music Fullscreen Info" seem to act strangely. Since i never use music, i just checked MusicVisualisation.xml and saw that ischannelpreview is also used in there.

(Again, to reproduce i use Confluence)

Quick test 1 (not much time today):

- ENABLE channelpreview in System -> TV -> Playback
- Start playing music, go to fullscreen after playing has started
- Press I for Info Dialog on playing Music
- Hit ENTER (MusicOSD should open - but does not)

Quick test 2:
- ENABLE channelpreview in System -> TV -> Playback
- Start playing music, go to fullscreen after playing has started
- Do _not_ open Fullscreen Info Dialog, just fullscreen music (without visualization in my case)
- Hit ENTER (MusicOSD opens as it should) (in oppsite to test case 1 where it seems to be blocked)

Quick test 3:
- DISABLED channelpreview in System -> TV -> Playback
- retest quick case 1
- Different behaviour, MusicOSD opens as it should

Maybe someone wants to look into it and create a new PR if confirmed. I'm happy to participate in testing later on, if my find seems to be valid.

xbmc/GUIInfoManager.cpp
```diff
@@ -10950,12 +10959,21 @@ bool CGUIInfoManager::ConditionsChangedValues(const std::map<INFO::InfoPtr, bool
...
 bool CGUIInfoManager::IsPlayerChannelPreviewActive() const
 {
-  bool streamValid = m_videoInfo.valid;
-  if (m_currentFile->HasPVRChannelInfoTag() && m_currentFile->GetPVRChannelInfoTag()->IsRadio())
-    streamValid = m_audioInfo.valid;
-
-  return m_playerShowInfo &&
-         (m_isPvrChannelPreview || !streamValid);
+  bool bReturn(false);
+  if (m_playerShowInfo)
```
FernetMenta: @ksooo I think we need to check for m_currentFile->HasPVRChannelInfoTag() at this level too

ksooo: Absolutely. This change will fix all of the issues reported above. Will open a PR in a minute.

PR #10265, opened 2016-08-10
add backward compatibility for AppearanceSettings

ronie:
allow users with a jarvis compatible skin (which is still using AppearanceSettings) to navigate to InterfaceSettings in order to switch skins.

### DIFF
--- a/xbmc/input/ButtonTranslator.cpp
+++ b/xbmc/input/ButtonTranslator.cpp
@@ -332,7 +332,8 @@ static const ActionMapping windows[] =
     { "pvrsettings"              , WINDOW_SETTINGS_MYPVR },
     { "playersettings"           , WINDOW_SETTINGS_PLAYER },
     { "mediasettings"            , WINDOW_SETTINGS_MEDIA },
-    { "interfacesettings"        , WINDOW_SETTINGS_INTERFACE },	
+    { "interfacesettings"        , WINDOW_SETTINGS_INTERFACE },
+    { "appearancesettings"       , WINDOW_SETTINGS_INTERFACE },	// backward compatibility to v16
     { "videoplaylist"            , WINDOW_VIDEO_PLAYLIST },
     { "loginscreen"              , WINDOW_LOGIN_SCREEN },
     { "profiles"                 , WINDOW_SETTINGS_PROFILES },


MartijnKaijser:
Shouldn't those skin be auto marked broken and changed to Estuary as well?

ronie:
that would be best, but i don't think we have such logic in our code...

MartijnKaijser:
iirc it worked with v15 -> v16. iirc @tamland also did a rescan of all deps when addon db was updated to new version to catch these issues.

MartijnKaijser:
Seems indeed the dependency check which happened on addon DB upgrade is missing/non functioning.

xbmc/input/ButtonTranslator.cpp
```diff
@@ -332,7 +332,8 @@ static const ActionMapping windows[] =
     { "pvrsettings"              , WINDOW_SETTINGS_MYPVR },
     { "playersettings"           , WINDOW_SETTINGS_PLAYER },
     { "mediasettings"            , WINDOW_SETTINGS_MEDIA },
-    { "interfacesettings"        , WINDOW_SETTINGS_INTERFACE },	
+    { "interfacesettings"        , WINDOW_SETTINGS_INTERFACE },
+    { "appearancesettings"       , WINDOW_SETTINGS_INTERFACE },	// backward compat
```
MartijnKaijser: could you add the version to it as well?

tamland:
There's never been any db upgrade checks.. What we have is an *invalidation* of the stored repository content when the repository addon version differ. *If* auto updates are on, it will trigger an update, which triggers a dependency check for addons in the repo and show a notification. (That last part I consider a bug). Afaik nothing's changed in this area.

That wont help if there's no update or you cant navigate to change it, so I'm wondering, since we break skin every release anyway, if we should just revert to default skin on upgrade. Otherwise this (and all future breaking changes) will have to stay in for as long as we want support upgrading from..

MartijnKaijser:
Well it's big issue that dependencies are not checked again on upgrade for every installed Addon. I've saying this for years already. This is not only a skin issue but everything else as well. Always default back to default skin is not an option and it doesn't fix the issue if some pvr Addon or whatever should become incompatible.

MartijnKaijser:
Actually i can tell you why it used to work. Back with v16 and before we just created the addons.xml files per addon branch and Kodi itself would gather all these XML files into a big list and from that filter out relevant addons. This method as such also contained skins or whatever from previous Kodi version, If there was no new version available it used the old version and compared it to the dependency on upgrade. That caused it to become "broken". 
Nowdays we prefilter all addons on the server and only provide a full compatible list. As such the Kodi install has no idea the current installed version has become incompatible as it's no longer in the addons.xml list.

tamland:
Yeah, makes sense. As I've said, that thing is broken by design and has never done any proper dependency checking, so no surprises there.

ronie:
added version number as requested.

PR #10216, opened 2016-08-02
player: make playspeed floating point and add tempo controls

FernetMenta:
- make playspeed a floating point variable.
- allow speed with audio enabled between 0.8 and 1.5 of normal speed
- add commands "tempoup" and "tempodown" to PlayerControls
- add gui labels to enable skins to set tempo
  - "tempoenabled": bool to test if active player supports tempo
  - "istempo": bool to test if speed is not 1 but in range of 0.8 to 1.5
  - "playspeed": label that holds playspeed

PlayerControls can be used by skins or mapped to keys. 
Example:

```
<keymap>
  <global>
    <keyboard>
      <f mod="ctrl">xbmc.playercontrol(tempoup)</f>
      <r mod="ctrl">xbmc.playercontrol(tempodown)</r>
    </keyboard>
  </global>
</keymap>
```

PR #10266, opened 2016-08-10
[skin] show playspeed

Rechi:
This shows the playspeed in the skins from #10216

### DIFF
--- a/addons/skin.estouchy/xml/Includes.xml
+++ b/addons/skin.estouchy/xml/Includes.xml
@@ -75,6 +75,7 @@
 		<value condition="Player.Forwarding">$LOCALIZE[31044] $VAR[Speed]</value>
 		<value condition="Player.Rewinding">$LOCALIZE[31045] $VAR[Speed]</value>
 		<value condition="Player.Paused">[UPPERCASE]$LOCALIZE[31043][/UPPERCASE]</value>
+		<value condition="Player.IsTempo">[UPPERCASE]$LOCALIZE[31046]: $INFO[Player.PlaySpeed][/UPPERCASE]</value>
 	</variable>
 	<variable name="Speed">
 		<value condition="Player.Forwarding2x | Player.Rewinding2x">2x</value>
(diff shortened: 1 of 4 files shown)


addons/skin.estouchy/xml/Includes.xml
```diff
@@ -75,6 +75,7 @@
 		<value condition="Player.Forwarding">$LOCALIZE[31044] $VAR[Speed]</value>
 		<value condition="Player.Rewinding">$LOCALIZE[31045] $VAR[Speed]</value>
 		<value condition="Player.Paused">[UPPERCASE]$LOCALIZE[31043][/UPPERCASE]</value>
+		<value condition="Player.IsTempo">$LOCALIZE[31046]: $INFO[Player.PlaySpeed]</value>
```
ronie: please use [UPPERCASE] here

Rechi:
@ronie updated

FernetMenta:
great!

Rechi:
I thought it might be useful to replace the condition `Player.IsTempo` with `Player.TempoEnabled + Player.Playing`.
Then it will also display `Play speed: 1.00`, but only if the player can change speed. This would show the ability of changing speed for the actual stream.

FernetMenta:
Player.Playing is redundant in this context. TempoEnabled returns false if player is not active.

Rechi:
TempoEnabled also returns true if player is fastforwarding or rewinding, but I think `Play speed: *` only should be displayed if playing.
My question was if it also makes sense to display `Play speed: 1.00` or should it display nothing in this case.

FernetMenta:
right, this differs from the internal implementation where IsPlaying is also true for ff/rw.

> My question was if it also makes sense to display Play speed: 1.00 or should it display nothing in this case.

Good question. Now that we have a label for speed, does it still make sense to display "forwarding" or "rewinding"? Why not just show the speed label for all cases?

MartijnKaijser:
@ronie @BigNoid @phil65 ?

ronie:
tested and works ok at my end.

thx!

PR #10267, opened 2016-08-10
[PVR] Guide window: Fix selcting channel via channel number input, if…

ksooo:
… target channel is on first epg page (channel 1 to 8 for Estuary and Confluence).

In Guide window, one can type in numbers to jump to the given channel row in the epg grid. If the target channel row is on the first page of the grid, this did not work. The wrong channel row got selected. This PR fixes this bug.

@Jalle19 for review? Note: The respective methods of the epg grid container control simply had no handling for this special case.


Jalle19:
Looks good

ksooo:
jenkins build this please

ksooo:
travis error is unrelated.

PR #10269, opened 2016-08-10
Split artist and title (shoutcast) and added the playing station

soerendd:
The solved problem is that when listening to internet radio most stations embedd meta information in the mp3 stream. The current implementation takes that information as it is and sets it to tag->title. In my case all of the stations i listen to use the same "format" to give information about the artist and the song title. 
The format is:
Artist - Title

My patch takes the string and tries to use a regex to split this into artist and title. 

Second part of the patch is to add a new field to the tag. Reason is that I would like to get the station name in an kodi addon (for LCDd). So i added also the needed interface definition. Furthermore I changed the guiinfomanager to recognize the field.


MartijnKaijser:
Please use correct indentation (spaces).

soerendd:
I habe committed additional changes to ensure indentation

xbmc/filesystem/ShoutcastFile.cpp
```diff
@@ -163,12 +167,28 @@ bool CShoutcastFile::ExtractTagInfo(const char* buf)
...
-
+  
   if (reTitle.RegFind(strBuffer.c_str()) != -1)
   {
     std::string newtitle(reTitle.GetMatch(1));
-    result = (m_tag.GetTitle() != newtitle);
-    m_tag.SetTitle(newtitle);
+
+    CRegExp reParseTitle(true);
+    reParseTitle.RegComp("(?P<a>.*) - (?P<t>.*)");
```
phil65: Wouldnt this lead to lot of false positives?

soerendd: As I've written most radio stations a have seen so far are using this scheme. Other stations which split their station names with a '-' mostly doing this without spaces. Replace the regex if you think an other would fit better. I use this currently in an custom build of libreelec to display this information in a 4x20 display. The artist and song title get their own lines. If this will not go into Kodi i have to find an other way to extract this information afterwards. Nevertheless the station name should be published by kodi.

Razzeee:
@soerendd 
can you split this into two PRs? So that we can look at station and title/artist seperatly?

Razzeee:
Ping @soerendd

Rechi:
@soerendd this needs a rebase

DaveTBlake:
Not sure why this was built by jenkins, but brought it to my attention.

If this does get continued, I think what rather than adding yet more methods to CMusicInfoTag, and info labels etc., getting stream info to the skins for shoutcasts could be better handled by using FileItem properties not tags.

MartijnKaijser:
@soerendd if you like to continue on this please update this PR or create new one

PR #10272, opened 2016-08-10
Support input streams having their own demuxer for the "OtherStream" …

rbuehlma:
…of CDVDInputStreamPVRManager

This allows the use of inputstream.mpd for PVR addons by simply passing the manifest URL

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDFactoryDemuxer.cpp
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDFactoryDemuxer.cpp
@@ -96,7 +96,7 @@ CDVDDemux* CDVDFactoryDemuxer::CreateDemuxer(CDVDInputStream* pInputStream, bool
         if(demuxer->Open(pOtherStream))
           return demuxer.release();
         else
-           return nullptr;
+          return nullptr;
       }
 
       /* Used for MediaPortal PVR addon (uses PVR otherstream for playback of rtsp streams) */
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDFactoryDemuxer.cpp
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDFactoryDemuxer.cpp
@@ -90,6 +90,15 @@ CDVDDemux* CDVDFactoryDemuxer::CreateDemuxer(CDVDInputStream* pInputStream, bool
 
     if (pOtherStream)
     {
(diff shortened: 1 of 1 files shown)


rbuehlma:
As this is my first PR to Kodi, please let me know if this is the wrong way to get this patch into Kodi and how it should be done correctly.

xbmc/cores/VideoPlayer/DVDDemuxers/DVDFactoryDemuxer.cpp
```diff
@@ -90,6 +90,15 @@ CDVDDemux* CDVDFactoryDemuxer::CreateDemuxer(CDVDInputStream* pInputStream, bool
...
     if (pOtherStream)
     {
+      if (pOtherStream->GetIDemux())
+      {
+        std::unique_ptr<CDVDDemuxClient> demuxer(new CDVDDemuxClient());
+        if(demuxer->Open(pOtherStream))
+          return demuxer.release();
+        else
+           return nullptr;
+      }
```
hudokkow: not commenting on the code but indent is wrong

rbuehlma: Is fixed now

FernetMenta:
absolutely no-go. the otherstream hack will disappear in v18 and the entire pvr streaming interface will be moved to inputstream. please read http://forum.kodi.tv/showthread.php?tid=259830

garbear:
Hi  rbuehlma, thanks for your contribution. We're working with a player that has over a decade of hacky patches and poorly implemented features. FernetMenta has been really firm about not making things worse. As things get better, bugs like this disappear. So while this PR improves things, ultimately the code is getting ripped out.

If you find any other bugs, let us know and FernetMenta will be able to tell if the code is slated for removal or not.

rbuehlma:
Thanks for the explanation. I do like the target solution of separating the PVR part from the stream handling. It was just a bit a throwback that the proposed workaround cannot be used until the new solution is in place.

But actually its not a big deal, I can still use the workaround to continue working on the addon and as soon as the implementation is done, I will switch to that one.

@FernetMenta could you send me a notification as soon as it is ready to be tested?

And one additional question: Will Kodi by itself know, that the current stream (handled by the inputstream addon) is a PVR Stream, and will it display the correct start/end time and description of the current program? Or is this something that needs to be handled in the addon?

PR #10273, opened 2016-08-10
File extension support for input stream

rbuehlma:
I'm not sure about the change in DVDFactoryInputStream.cpp.
It was required as the function is called without a player from VideoPlayerCodec::Init
m_pInputStream = CDVDFactoryInputStream::CreateInputStream(NULL, fileitem);

### DIFF
--- a/xbmc/addons/InputStream.cpp
+++ b/xbmc/addons/InputStream.cpp
@@ -186,6 +186,21 @@ bool CInputStream::Supports(const CFileItem &fileitem)
       return true;
   }
 
+  // check extensions
+  for (auto ext : m_extensionsList)
+  {
+    if (ext.empty())
+      continue;
+
+    if (StringUtils::EndsWith(fileitem.GetPath(), ext)) {
+      return true;
+    }
+
+    if ((protocol == "http" || protocol == "https") && fileitem.GetPath().find(ext + "?") != std::string::npos) {
+      return true;
+    }
+  }
+
   // check paths
   CSingleLock lock(m_parentSection);
   auto it = m_configMap.find(ID());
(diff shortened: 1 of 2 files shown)


FernetMenta:
-1 the only reason for this pr is the otherstream hack

rbuehlma:
This PR is not only for the case of the otherstream hack. This also works if a URL is provided some other way (e.g. .strm file and probably also the URL directly).

FernetMenta:
for .strm files it is better to use pathList. Trying to select an inputstream addon by an extension won't work reliable.

rbuehlma:
I agree that the extension is not always reliable but it is probably the most generic way and should work very well in most cases (e.g. .mpd for DASH-Manifests) in my case.

I do not know a more reliable way as the path differs for each provider. Even the mime-type is probably not always the same. Common on all the stream I observed so far is currently only the extension ".mpd" (somtimes followed by some parameters "?..." in case of http).

Additionally, the field m_extensionsList already exists but without any functionality.

FernetMenta:
> Additionally, the field m_extensionsList already exists but without any functionality.

this is a c/p left-over

ListItems can carry a property for requesting a particular inputstream addon. That is the method to be used by addons.

garbear:
Is there somewhere we can document that this property is named "`inputstreamaddon`"? Game add-ons have a similiar property, "`gameclient`".

MartijnKaijser:
isn't it mentioned in the python modules doxygen?

garbear:
`inputstreamaddon` doesn't show up anywhere else in the codebase

MartijnKaijser:
@garbear 
https://github.com/xbmc/xbmc/blob/efcc4347c8e5e3d85fff5bbff55c8d41cdfbadd2/xbmc/addons/InputStream.cpp#L170
`item.setProperty('inputstreamaddon', 'inputstream.mpd')`

garbear:
anywhere *else*, as in InputStream.cpp is the only instance. exposing this to add-on devs might be useful, I've had a lot of requests for the "gameclient" attribute (but that's because there's multiple game clients for many games)

PR #10275, opened 2016-08-11
[osx] media key: send also SDL_KEYUP

Rechi:
Without this PR the osx media keys are mapped to longpress actions if one exists.
Sine #10171 there exists a longpress action for `play_pause` and the play key doesn't do the expected thing.

### DIFF
--- a/xbmc/platform/darwin/osx/SDLMain.mm
+++ b/xbmc/platform/darwin/osx/SDLMain.mm
@@ -434,6 +434,17 @@ static void setupWindowMenu(void)
   [pool release];
 }
 
+static void keyPress(SDLKey key)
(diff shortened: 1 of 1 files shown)


MartijnKaijser:
@Memphiz ping

Memphiz:
Looks correct - jenkins build this please
@MartijnKaijser will this travis stuff ever show a green icon? I always find it irritating that it shows a red cross...

MartijnKaijser:
@Memphiz from the log sometimes :)
Can't we just drop the autotools builds from travis to start with and keep cmake?

Memphiz:
thx @Rechi

PR #10277, opened 2016-08-11
paplayer: interims fix for speed while paused

FernetMenta:
@rechi @Tolriq this fixes the speed issue for paplayer

### DIFF
--- a/xbmc/cores/paplayer/PAPlayer.cpp
+++ b/xbmc/cores/paplayer/PAPlayer.cpp
@@ -949,6 +949,13 @@ void PAPlayer::SetSpeed(float speed)
 
 float PAPlayer::GetSpeed()
 {
+  //! @todo: remove extra member for pause state
+  //! there was inconsistency throughout the entire application on how speed
+  //! and pause were used. Now speed is defined as current playback speed.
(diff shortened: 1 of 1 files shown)


FernetMenta:
jenkins build this please

xbmc/cores/paplayer/PAPlayer.cpp
```diff
@@ -949,6 +949,11 @@ void PAPlayer::SetSpeed(float speed)
 
 float PAPlayer::GetSpeed()
 {
+  // maximum crap that pause is handled independent from speed
+  // TODO: change this
```
MartijnKaijser: could you use doxygen style todo so it shows up in the todo list?

FernetMenta:
@MartijnKaijser ok this way? I am no doxy expert

MartijnKaijser:
`//! @todo Change this`
Maybe extend it a bit more

FernetMenta:
doxy updated

Rechi:
@FernetMenta since this PR `PlayerControl(Play)` doesn't resume playback for `PAPlayer`, only pausing.
Resuming with the keyboard still works, but not in the GUI and web interface.

PR #10278, opened 2016-08-12
VideoPlayer: clock fixes

FernetMenta:
- GetClockInfo was not thread safe
- wrong method call in RenderManager


FernetMenta:
jenkins build this please

PR #10280, opened 2016-08-12
[PVR] Fix PVR channels not working as Kodi startup window

ksooo:
Fixes http://trac.kodi.tv/ticket/16765

Note: currently only radio and tv channel window will work as startup window. Only these windows can be activated as startup windows in Kodi settings GUI. This has always been the case, as not changed by this PR (only wording in the UI), although it also makes sense to use the guide window as startup window. But this will be some more work and thus will be done in another PR.

Settings: 
screenshot004

Video, to get an impression how it looks like: https://www.dropbox.com/s/5molkfr0b3xicde/IMG_2229.MOV

@Jalle19 @BigNoid for review?

EDIT: Compared to the solution we had before PVRManager Krypton rework (that broke the old solution, btw) the new approach does not need any busy waiting dialogs until PVR has been started which makes it technically "correct" now, imo.

EDIT2: guide windows now also can be used as start windows; was less work than expected. :-)

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -10181,7 +10181,17 @@ msgctxt "#19272"
 msgid "You need a tuner, backend software, and an add-on for the backend to be able to use PVR. Please visit http://kodi.wiki/view/PVR to learn more."
 msgstr ""
 
-#empty strings from id 19273 to 19274
+#. Settings -> Interface -> Other -> Startup window
+#: system/settings/settings.xml
+msgctxt "#19273"
+msgid "TV guide"
+msgstr ""
+
+#. Settings -> Interface -> Other -> Startup window
+#: system/settings/settings.xml
+msgctxt "#19274"
+msgid "Radio guide"
+msgstr ""
 
 #: xbmc/pvr/timers/PVRTimerInfoTag.cpp
 msgctxt "#19275"
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -9742,9 +9742,10 @@ msgctxt "#19179"
 msgid "Deleted"
 msgstr ""
 
+#. Settings -> Interface -> Other -> Startup window
 #: system/settings/settings.xml
 msgctxt "#19180"
-msgid "TV"
+msgid "TV channels"
 msgstr ""
 
 #: system/settings/settings.xml
@@ -9757,9 +9758,10 @@ msgctxt "#19182"
 msgid "Days to display"
 msgstr ""
 
+#. Settings -> Interface -> Other -> Startup window
 #: system/settings/settings.xml
 msgctxt "#19183"
-msgid "Radio"
+msgid "Radio channels"
 msgstr ""
 
 #. Used on pvr recordings as button to show of them
--- a/xbmc/guilib/GUIWindowManager.cpp
+++ b/xbmc/guilib/GUIWindowManager.cpp
@@ -651,9 +651,11 @@ void CGUIWindowManager::PreviousWindow()
   }
   // get the previous window in our stack
   if (m_windowHistory.size() < 2)
-  { // no previous window history yet - check if we should just activate home
+  {
+    // no previous window history yet - check if we should just activate home
     if (GetActiveWindow() != WINDOW_INVALID && GetActiveWindow() != WINDOW_HOME)
     {
+      CloseWindowSync(pCurrentWindow);
       ClearWindowHistory();
       ActivateWindow(WINDOW_HOME);
     }
@@ -667,6 +669,7 @@ void CGUIWindowManager::PreviousWindow()
   if (!pNewWindow)
   {
     CLog::Log(LOGERROR, "Unable to activate the previous window");
+    CloseWindowSync(pCurrentWindow);
     ClearWindowHistory();
     ActivateWindow(WINDOW_HOME);
     return;
@@ -1500,8 +1503,10 @@ void CGUIWindowManager::AddToWindowHistory(int newWindowID)
   { // found window in history
     m_windowHistory = historySave;
   }
-  else
-  { // didn't find window in history - add it to the stack
+  else if (newWindowID != WINDOW_SPLASH)
+  {
+    // didn't find window in history - add it to the stack
+    // but do not add the splash window to history, as we never want to travel back to it
     m_windowHistory.push(newWindowID);
   }
 }
--- a/addons/skin.estuary/1080i/Includes_MediaMenu.xml
+++ b/addons/skin.estuary/1080i/Includes_MediaMenu.xml
@@ -253,13 +253,10 @@
 	<include name="MediaMenuCommon">
 		<include>OpenClose_Left</include>
 		<depth>DepthSideBlade</depth>
-		<animation type="Conditional" condition="$EXP[sidebar_focused]">
+		<left>-484</left>
+		<animation type="Conditional" condition="$EXP[sidebar_focused]" reversible="true">
 			<effect type="fade" start="0" end="100" time="300" tween="sine" easing="out" />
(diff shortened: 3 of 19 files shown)


Jalle19:
Couldn't find anything that looked obviously wrong, +1.

ksooo:
rebased

@Jalle19 added two more commits. 'PVRGuideWindow* cosmetics' and 'guide window as start window'. Any objections?

ksooo:
jenkins build this please

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -9742,9 +9742,10 @@ msgctxt "#19179"
 msgid "Deleted"
 msgstr ""
 
+#. Settings -> Interface -> Other -> Startup window
 #: system/settings/settings.xml
 msgctxt "#19180"
-msgid "TV"
+msgid "TV channels"
```
MartijnKaijser: Are you sure can change this string without negative impact on how it was used before?

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -9757,9 +9758,10 @@ msgctxt "#19182"
 msgid "Days to display"
 msgstr ""
 
+#. Settings -> Interface -> Other -> Startup window
 #: system/settings/settings.xml
 msgctxt "#19183"
-msgid "Radio"
+msgid "Radio channels"
```
MartijnKaijser: Same here

addons/resource.language.en_gb/resources/strings.po
ksooo: Yep. It's exclusively used for Settings -> Interface -> Other -> Startup window value. Checked xbmc core (incl. Estuary and Estouchy) + Confluence.

addons/resource.language.en_gb/resources/strings.po
ksooo: Yep. It's exclusively used for Settings -> Interface -> Other -> Startup window value. Checked xbmc core (incl. Estuary and Estouchy) + Confluence.

ksooo:
If nobody objects i will merge this tomorrow, in time for beta 1

Jalle19:
You're doing great work @ksooo, keep it up!

bkuhls:
Hi, this commit leads to a segfault here during startup: http://trac.kodi.tv/ticket/16848

xbmc/guilib/GUIWindowManager.cpp
```diff
@@ -1500,8 +1503,10 @@ void CGUIWindowManager::AddToWindowHistory(int newWindowID)
   { // found window in history
     m_windowHistory = historySave;
   }
-  else
-  { // didn't find window in history - add it to the stack
+  else if (newWindowID != WINDOW_SPLASH)
+  {
+    // didn't find window in history - add it to the stack
+    // but do not add the splash window to history, as we never want to travel back to it
```
xhaggi: This is wrong, we we need the splash in history to render it properly. Without `GetActiveWindow` never return the `CGUIWindowSplash` and therefore it's not rendered. We remove the splash window right after the GUI is ready https://github.com/xbmc/xbmc/blob/master/xbmc/Application.cpp#L3928

ksooo: I added this for a reason. If I only could remember what is was. But if you simply revert this another Jack will jump out of the box.

ksooo: If I revert the above line I get "Unable to activate the previous window" in the log if I hit "back" in the startup PVR window. "Back" looks into history, finds WINDOW_SPLASH and tries to go back to the (already removed) WINDOW_SPLASH and then, after the log error, defaults to WINDOW_HOME.

<pre>
(lldb) thread backtrace
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x00000001010560f0 kodi.bin`CGUIWindowManager::PreviousWindow(this=0x000000010a1068d0) at GUIWindowManager.cpp:709
    frame #1: 0x000000010104584c kodi.bin`CGUIWindow::OnBack(this=0x000000010b0c2e00, actionID=10) at GUIWindow.cpp:964
    frame #2: 0x0000000102628216 kodi.bin`CGUIMediaWindow::OnBack(this=0x000000010b0c2e00, actionID=10) at GUIMediaWindow.cpp:217
    frame #3: 0x0000000101d85502 kodi.bin`PVR::CGUIWindowPVRBase::OnBack(this=0x000000010b0c2e00, actionID=10) at GUIWindowPVRBase.cpp:129
    frame #4: 0x0000000101041226 kodi.bin`CGUIWindow::OnAction(this=0x000000010b0c2e00, action=0x00007ffeefbfd850) at GUIWindow.cpp:449
    frame #5: 0x00000001026277ef kodi.bin`CGUIMediaWindow::OnAction(this=0x000000010b0c2e00, action=0x00007ffeefbfd850) at GUIMediaWindow.cpp:172
    frame #6: 0x0000000101d84d0c kodi.bin`PVR::CGUIWindowPVRBase::OnAction(this=0x000000010b0c2e00, action=0x00007ffeefbfd850) at GUIWindowPVRBase.cpp:113
    frame #7: 0x0000000101d88378 kodi.bin`PVR::CGUIWindowPVRChannelsBase::OnAction(this=0x000000010b0c2e00, action=0x00007ffeefbfd850) at GUIWindowPVRChannels.cpp:132
    frame #8: 0x000000010105a4c8 kodi.bin`CGUIWindowManager::HandleAction(this=0x000000010a1068d0, action=0x00007ffeefbfd850) const at GUIWindowManager.cpp:1094
    frame #9: 0x000000010105a16b kodi.bin`CGUIWindowManager::OnAction(this=0x000000010a1068d0, action=0x00007ffeefbfd850) const at GUIWindowManager.cpp:1044
    frame #10: 0x00000001001bfff0 kodi.bin`CApplication::OnAction(this=0x000000010b800600, action=0x00007ffeefbfd850) at Application.cpp:1998
    frame #11: 0x0000000101186edc kodi.bin`CInputManager::ExecuteInputAction(this=0x000000010a1e5b10, action=0x00007ffeefbfd850) at InputManager.cpp:729
    frame #12: 0x000000010118b099 kodi.bin`CInputManager::HandleKey(this=0x000000010a1e5b10, key=0x00007ffeefbfdca0) at InputManager.cpp:669
    frame #13: 0x0000000101189626 kodi.bin`CInputManager::OnKeyUp(this=0x000000010a1e5b10, key=0x00007ffeefbfe090) at InputManager.cpp:681
    frame #14: 0x00000001011884ad kodi.bin`CInputManager::OnEvent(this=0x000000010a1e5b10, newEvent=0x00007ffeefbfe660) at InputManager.cpp:390
    frame #15: 0x00000001001bf18d kodi.bin`CApplication::HandleWinEvents(this=0x000000010b800600) at Application.cpp:335
    frame #16: 0x00000001001ee31f kodi.bin`CApplication::FrameMove(this=0x000000010b800600, processEvents=true, processGUI=true) at Application.cpp:2670
    frame #17: 0x0000000100485ab6 kodi.bin`CXBApplicationEx::Run(this=0x000000010b800600, params=0x00007ffeefbfeb88) at XBApplicationEx.cpp:108
    frame #18: 0x0000000101a13a61 kodi.bin`::XBMC_Run(renderGUI=true, params=0x00007ffeefbfeb88) at xbmc.cpp:88
</pre>

ksooo: So, we need to remove WINDOW_SPLASH from the history when it gets destroyed (we should do this with every window we destroy, right?!) ... and make sure we are going to WINDOW_HOME, then.

xhaggi: Okay will adjust the PR tommorow and remove the destoryed window from history.

ksooo: > remove the destroyed window from history

 Every window we destroy should to be removed from history, not only WINDOW_SPLASH.

PR #10282, opened 2016-08-12
[epg] CEpg::UpdateEntry: Remove wrong error log. Not finding an epg e…

ksooo:
…vent can be okay due to epg linger time contraints.


ksooo:
jenkins build this please

PR #10283, opened 2016-08-13
[processInfo] reset audio/video codec info

mk01:
(if particular stream type is missing within the stream, process info
values won't be initialised)

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -928,7 +928,10 @@ void CVideoPlayer::OpenDefaultStreams(bool reset)
     }
   }
   if(!valid)
+  {
     CloseStream(m_CurrentVideo, true);
+    m_processInfo->ResetVideoCodecInfo();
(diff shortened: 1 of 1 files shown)


xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -909,6 +909,9 @@ void CVideoPlayer::CloseDemuxer()
 
 void CVideoPlayer::OpenDefaultStreams(bool reset)
 {
+  m_processInfo->ResetAudioCodecInfo();
+  m_processInfo->ResetVideoCodecInfo();
```
FernetMenta: that won't work. OpenDefaultStream can be called often but won't do anything if parameters did not change.

better explicitly reset if there is no stream of a particular type

mk01:
@FernetMenta 
this variant ?

FernetMenta:
looks good, thanks
ready to go?

mk01:
yes

FernetMenta:
jenkins build this please

PR #10284, opened 2016-08-13
VideoPlayer: fix false positive drop detection for ffmpeg decoder

FernetMenta:
false positive drops results in a/v going out of sync

@fritsch mind testing this with you sample?


fritsch:
Via our discussion on IRC I propose to change CDVDClock::ErrorAdjust

to do an ahead correction when > 0.75 m_frametime and for the other direction by -0.75 * frametime, which will give us a band where nothing happens of 0.5 * m_frametime where nothing happens.

FernetMenta:
DVDDlock::ErrorAdjust is completely disconnected from this change

fritsch:
@FernetMenta you asked me testing the sample from above and I did so.

FernetMenta:
@fritsch yes thanks for testing but this was before I got to know that the issue is related to clock feedback sync mode.

FernetMenta:
jenkins build this please

PR #10285, opened 2016-08-13
[osx/XBMCHelper] - some fixes

Memphiz:
Here are some fixes for the XBMCHelper service which we use on osx to handle IR Remote Controls for example.
1. This updates the 3rd party class HIDRemote which we use to latest stable version
2. Adds some logging
3. Ensures that the service is started even when Kodi is run from the Login Items (autostart on login) - which failed since the needed services might not be in a responding shape at that point.
4. Ensure that the service is restarted in case it gets killed

This is OSX only stuff and i consider it safe for krypton (it doesn't fix all known issues with IR remotes but it makes it more reliable when Kodi is autostarted on login).


Memphiz:
jenkins build and merge

PR #10286, opened 2016-08-13
[PVR] Fix home screen channel widget context menu.

ksooo:
CFileItem::IsVideo must not return true for any PVR item except tv recordings. Fixes wrong context menu entries for homescreen channel widget.

Note: currently the channel widget items have no context menu items. Those are still about to be implemented. Will do later, when i find the time.

@Jalle19 good to go?

### DIFF
--- a/xbmc/FileItem.cpp
+++ b/xbmc/FileItem.cpp
@@ -733,8 +733,13 @@ bool CFileItem::IsVideo() const
   if (HasPictureInfoTag())
     return false;
 
+  // only tv recordings are videos...
   if (IsPVRRecording())
-    return true;
+    return !GetPVRRecordingInfoTag()->IsRadio();
+
+  // ... all other PVR items are not.
+  if (IsPVR())
+    return false;
 
   if (URIUtils::IsDVD(m_strPath))
     return true;


ksooo:
jenkins build this please

xbmc/FileItem.cpp
```diff
@@ -733,8 +733,15 @@ bool CFileItem::IsVideo() const
   if (HasPictureInfoTag())
     return false;
 
-  if (IsPVRRecording())
-    return true;
+  if (IsPVR())
```
Jalle19: Couldn't this still be IsPVRRecording?

ksooo: No, the actual problem is that some lines below the 'video' filename url extensions are checked and .pvr (which matches both channels and recordings) and this leads to a false positive. Thus, pvr must be handled without filename extension check.

ksooo: but yeah maybe this can be coded  a bit more efficient. Will change. Thanks.

ksooo:
@Jalle19 good to go now?

ksooo:
jenkins build this please

Jalle19:
Looks good

ksooo:
jenkins build this please

PR #10287, opened 2016-08-13
[OSX/CoreAudioDevice] - use the alternate method for retrieving the a…

Memphiz:
…udio device name (by using a CFString object). This fixes broken audio device names in certain localisations (f.e. umlauts in the device name).

Mostly a nobrainer. @FernetMenta for a second pair of eyes. (my test was to add an aggregated audio device via audio-midi-setup which had the name "Hauptgerät". In Kodi with the old code only "Hauptgerä" was displayed. With the new code the name is displayed correct in Kodi too).

### DIFF
--- a/xbmc/cores/AudioEngine/Sinks/osx/CoreAudioDevice.cpp
+++ b/xbmc/cores/AudioEngine/Sinks/osx/CoreAudioDevice.cpp
@@ -215,17 +215,14 @@ std::string CCoreAudioDevice::GetName() const
   AudioObjectPropertyAddress  propertyAddress;
   propertyAddress.mScope    = kAudioDevicePropertyScopeOutput;
   propertyAddress.mElement  = 0;
-  propertyAddress.mSelector = kAudioDevicePropertyDeviceName;
-
-  UInt32 propertySize;
-  OSStatus ret = AudioObjectGetPropertyDataSize(m_DeviceId, &propertyAddress, 0, NULL, &propertySize);
-  if (ret != noErr)
-    return "";
+  propertyAddress.mSelector = kAudioDevicePropertyDeviceNameCFString;
 
   std::string name;
-  char *buff = new char[propertySize + 1];
-  buff[propertySize] = 0x00;
-  ret = AudioObjectGetPropertyData(m_DeviceId, &propertyAddress, 0, NULL, &propertySize, buff);
+  CFStringRef deviceName = NULL;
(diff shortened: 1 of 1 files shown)


xbmc/cores/AudioEngine/Sinks/osx/CoreAudioDevice.cpp
```diff
@@ -215,28 +215,24 @@ std::string CCoreAudioDevice::GetName() const
...
-  if (ret != noErr)
-    return "";
+  propertyAddress.mSelector = kAudioDevicePropertyDeviceNameCFString;
 
   std::string name;
-  char *buff = new char[propertySize + 1];
-  buff[propertySize] = 0x00;
-  ret = AudioObjectGetPropertyData(m_DeviceId, &propertyAddress, 0, NULL, &propertySize, buff);
+  CFStringRef deviceName = NULL;
+  UInt32 propertySize = sizeof(deviceName);
```
FernetMenta: that looks strange: 
sizeof(some ref that points to null)

what is the value of propertySize after this call?

Memphiz: 8 (size of a pointer) - thats what i found in examples for this. I can also query for the propertysize which also returns 8.

Memphiz: Those XXXXRef types in apple land ar simple pointers ...

FernetMenta: I am sure this is correct, I just don't get the sense why propertySize is passed as a pointer :)
anyway, go ahead

Memphiz: I guess in case of error it writes the needed size into it - so the caller can resize the passed buffer with the needed size and try again ...

Memphiz:
Kicked a testbuild just for osx ... will merge once its ok

MrMC:
AudioObjectGetPropertyData is a general routine used by many things including returning variable length structures.

    @param          ioDataSize
                        A UInt32 which on entry indicates the size of the buffer pointed to by
                        outData and on exit indicates how much of the buffer was used.

In this case, the returned data is not variable length as it's a CFStringRef type.

Memphiz:
thx for the info :) - build was fine ... merging

PR #10288, opened 2016-08-13
[upnp] Controller only starts when Server is started

Rallymen007:
Follow up to #10249, hopefully I got it right this time.

### DIFF
--- a/xbmc/network/NetworkServices.cpp
+++ b/xbmc/network/NetworkServices.cpp
@@ -829,12 +829,9 @@ bool CNetworkServices::StartUPnP()
   ret |= StartUPnPClient();
   if (CSettings::GetInstance().GetBool(CSettings::SETTING_SERVICES_UPNPSERVER))
   {
-   ret |= StartUPnPServer();
-  }
-
-  if (CSettings::GetInstance().GetBool(CSettings::SETTING_SERVICES_UPNPCONTROLLER))
-  {
-    ret |= StartUPnPController();
+    ret |= StartUPnPServer(); 
+    if (CSettings::GetInstance().GetBool(CSettings::SETTING_SERVICES_UPNPCONTROLLER))
+      ret |= StartUPnPController();
   }
 
   if (CSettings::GetInstance().GetBool(CSettings::SETTING_SERVICES_UPNPRENDERER))


MartijnKaijser:
@Montellese

xbmc/network/NetworkServices.cpp
```diff
@@ -829,12 +829,9 @@ bool CNetworkServices::StartUPnP()
...
-
-  if (CSettings::GetInstance().GetBool(CSettings::SETTING_SERVICES_UPNPCONTROLLER))
-  {
-    ret |= StartUPnPController();
+    ret |= StartUPnPServer(); 
+    if (CSettings::GetInstance().GetBool(CSettings::SETTING_SERVICES_UPNPCONTROLLER))
+      ret |= StartUPnPController();
   }
 
   if (CSettings::GetInstance().GetBool(CSettings::SETTING_SERVICES_UPNPRENDERER))
```
Montellese: Could you add a second commit to remove the curly braces belonging to this `if` statement?

Montellese:
I just looked at the whole code again and wondering what the original problem was. `CNetworkServices::StartUPnP()` used to call all the different upnp-related `StartFoo()` methods. All those methods already contain logic to check if the respective settings are enabled i.e. `StartUPnPServer()` checks if `CSettings::SETTING_SERVICES_UPNPSERVER` is enabled. Why do we need to perform these checks in `CNetworkServices::StartUPnP()` as well?

xbmc/network/NetworkServices.cpp
Rallymen007: Those two ifs are nested as to have the "controller needs server" behavior, and I think those curly braces are needed as the if block isn't one instruction long.

Rallymen007:
Indeed they do some checks. If I recall correctly, my issue was that under no circumstances Kodi left port 1900 alone which is something I needed as the UPnP library I use has issues with port reuse despite all the flags being set properly.

This was the spirit of the original fix, but as you pointed out the only way to have this behavior is to fully recompile Kodi without UPnP as there is no suitable setting to shut down the UPnP client. The original fix helped me because of my wrong understanding of the settings, but after restoring the correct behavior it doesn't do shit.

So yeah, apart from adding a separate setting to shut down the UPnP client (which I could try but I'm sure I'll just butcher everything :p, not to mention it might be wrong) this fix is no longer useful.

MartijnKaijser:
@Rallymen007 @Montellese should anything changed?

Montellese:
@MartijnKaijser actually #10249 was unnecessary as well. We should probably revert it because all the checks that were added are already done in the UPnP logic.

PR #10293, opened 2016-08-13
Controller dialog: Feature groups

garbear:
This PR puts controller features into groups in the controller dialog, similar to how settings groups work.

feature groups


da-anda:
just noticed in the screenshot that the word "configuration" is cut off in the dialogs title bar

garbear:
that's because I resized the window. somewhere some bounding boxes aren't getting updated when the window is resized

MartijnKaijser:
jenkins build this please

MartijnKaijser:
@ronie @BigNoid @phil65 any objections?

ronie:
nope.

HitcherUK:
Why are there 2 labels with the same ID (8)?

garbear:
An error. The labels should be: https://github.com/garbear/xbmc/blob/8824342/xbmc/games/controllers/windows/GUIControllerDefines.h

garbear:
Problem is harmless, the first #8 is left over from a rebase. Fix is: 81b9e7f1d

Should I send the fix here or to the skin.estuary repo?

MartijnKaijser:
@garbear master branch here

PR #10294, opened 2016-08-14
[iOS/XMemUtils] - fix pagesize retrieval once and for all for all ios…

Memphiz:
… devices (32bit and 64bit ARM CPUs). Fixes free memory count being off by factor 4.

tested on ios 5.1.1 32bit arm, ios 6.0.2 32bit arm, ios 9.3.3 64bit arm.


Memphiz:
jenkins build this please

PR #10299, opened 2016-08-16
Bump peripheral API - enable translating between controller profiles

garbear:
This is the change needed for https://github.com/kodi-game/peripheral.joystick/pull/45 (which i accidentally merged and reverted already) to be pulled into master.

This PR allows the user to map any controller profile and have it work in the GUI. For example, you could map just the SNES profile (if you have a SNES controller) and peripheral.joystick will derive how it should be mapped to a 360 controller.

Once this change is in, I'll take care of updating peripheral.joystick.

### DIFF
--- a/addons/kodi.peripheral/addon.xml
+++ b/addons/kodi.peripheral/addon.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<addon id="kodi.peripheral" version="1.0.19" provider-name="Team-Kodi">
+<addon id="kodi.peripheral" version="1.0.20" provider-name="Team-Kodi">
 	<backwards-compatibility abi="1.0.19"/>
 	<requires>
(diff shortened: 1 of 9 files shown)


MartijnKaijser:
@garbear we already compile from peripheral.joystick master by default.

MartijnKaijser:
@garbear might merge it so it's ready for beta2 with some weeks of testing in master?
jenkins build this please

garbear:
I added two commits that are pretty inconsequential, but it'll make my life a lot easier when I merge buttonmap-deriving and udev support into peripheral.joystick. Both commits are backwards compatible - the min version doesn't change.

jenkins build and merge

garbear:
peripheral.joystick has been updated from v1.1.0 (beta1) to v1.2.0 (beta2).

This means that in the 17-beta2 changelog, we can add controller translation and udev support.

PR #10300, opened 2016-08-16
[OSX] - fixed regression (non working ir remote and not executed pref…

Memphiz:
…light script) introduced in behavior change of CUtil:GetHomePath in #10180.

What the title says ... will kick osx testbuilds as this is osx only.


Memphiz:
build was fine

PR #10302, opened 2016-08-16
Settings: make Peripherals a Standard setting, not Advanced

MilhouseVH:
This change will avoid [this situation](http://forum.kodi.tv/showthread.php?tid=269814&pid=2395203#pid2395203) where new users have trouble configuring CEC which is hidden by default, as it's an "Expert" setting.

Ping @jjd-uk.

### DIFF
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -2596,7 +2596,7 @@
     <category id="input" label="14125" help="36374">
       <group id="1" label="35000">
         <setting id="input.peripherals" type="action" label="35000" help="36375">
-          <level>2</level>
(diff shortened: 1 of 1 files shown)


jjd-uk:
I don't see the need for this to be a Level 0 Basic setting, since the default Level on install is Level 1 Standard I would prefer any change to be to Level 1.

MilhouseVH:
I assumed the default level would have been Basic, but since it's Standard I've updated the PR accordingly - thanks! Also changed the title as the current level isn't Expert, it's Advanced. Such a small change, so many mistakes. :)

PR #10305, opened 2016-08-17
TextureManager: We still need to free textures when playing video

popcornmix:
Currently if you background video and scroll through movie library we never release the textures resulting in a huge memory leak.

See: http://forum.kodi.tv/showthread.php?tid=228120


popcornmix:
Ping @FernetMenta

FernetMenta:
+1
no need for jenkins here

PR #10308, opened 2016-08-18
[PVR] Recordings window context menu fixes

ksooo:
4 fixes:

1) 'Mark as watched': must not set play count to one, unconditionally, but increment it.
2) Fix duplicate context menu entries (regression introduced by non-pvr krypton context menu rework) 
3) 'Resume from ...': Fix time format to be aligned with format used for non-pvr videos
4) 'Play / Play from beginning': Fix wording to be aligned with strings used for non-pvr videos

before:
screenshot005

after:
screenshot007

@Jalle19 objections?

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -923,6 +923,7 @@ msgctxt "#207"
 msgid "Plot"
 msgstr ""
 
+#: xbmc/pvr/windows/PVRGUIWindowRecordings.cpp
 #: xbmc/dialogs/GUIDialogPlayEject.cpp
 #: system/settings/settings.xml
 msgctxt "#208"
@@ -5176,6 +5177,8 @@ msgctxt "#12022"
 msgid "Resume from %s"
 msgstr ""
 
+#. Label of video/recordings context menu items
+#: xbmc/pvr/windows/PVRGUIWindowRecordings.cpp
 #: xbmc/video/ContextMenus.cpp
 msgctxt "#12023"
 msgid "Play from beginning"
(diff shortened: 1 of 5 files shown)


ksooo:
jenkins build this please

MartijnKaijser:
I can't image the that mark as watched button should be there if playcount != 0  (could it then would be mark as watched again). As such you always set it to one if it's watched

ksooo:
> I can't image the that mark as watched button should be there if playcount != 0 (could it then would be mark as watched again). As such you always set it to one if it's watched

'Mark as watched' will actually not be visible if playcount > 0. Independently playcount will be increased on every 'successful' completed watch of a recording (automatically, not via context menu). This is exactly how it is implemented for non-pvr videos. I checked that.

Jalle19:
Don't library items show Mark as unwatched if playcount > 0?

ksooo:
So do we for pvr recordings.

ksooo:
It did you actually mean Mark as watched?

ksooo:
BTW, I did not change the visibility logic in this PR. So, if there is something buggy with visibility, it is not introduced by/related to this PR.

ksooo:
@Jalle19 thus us the logic. 

<code>
      if (recording->m_playCount > 0)
        buttons.Add(CONTEXT_BUTTON_MARK_UNWATCHED, 16104); /* Mark as unwatched */
      else
        buttons.Add(CONTEXT_BUTTON_MARK_WATCHED, 16103);   /* Mark as watched */
</ code>

I don't see a problem here.

ksooo:
@Jalle19 good to go? I would like to have this in beta 1.

Jalle19:
Shove it in

PR #10309, opened 2016-08-18
Deadzone settings

garbear:
This PR promotes deadzone configuration to first-class peripheral settings. It adds the following settings to peripherals.xml:

``` xml
<peripheral class="joystick" mapTo="joystick">
    <setting key="left_stick_deadzone" type="float" label="35078" order="1" value="0.2" min="0.0" step="0.05" max="1.0" />
    <setting key="right_stick_deadzone" type="float" label="35079" order="2" value="0.2" min="0.0" step="0.05" max="1.0" />
</peripheral>
```

The result is that all unique joysticks can have their own deadzone settings. (If two joystick have the same USB VID/PID, then they will have to share deadzone settings.) Additionally, there are separate settings for each analog stick.

deadzone setting

### DIFF
--- a/xbmc/peripherals/devices/Peripheral.cpp
+++ b/xbmc/peripherals/devices/Peripheral.cpp
@@ -158,10 +158,11 @@ bool CPeripheral::Initialise(void)
   }
   else
   {
-    m_strSettingsFile = StringUtils::Format("special://profile/peripheral_data/%s_%s_%s.xml",
+    m_strSettingsFile = StringUtils::Format("special://profile/peripheral_data/%s_%s_%s_%s.xml",
                                             PeripheralTypeTranslator::BusTypeToString(m_mappedBusType),
                                             m_strVendorId.c_str(),
-                                            m_strProductId.c_str());
+                                            m_strProductId.c_str(),
+                                            CUtil::MakeLegalFileName(m_strDeviceName, LEGAL_WIN32_COMPAT).c_str());
   }
 
   LoadPersistedSettings();
--- a/xbmc/peripherals/devices/Peripheral.cpp
+++ b/xbmc/peripherals/devices/Peripheral.cpp
@@ -33,6 +33,7 @@
 #include "utils/StringUtils.h"
 #include "utils/XBMCTinyXML.h"
 #include "utils/XMLUtils.h"
+#include "Util.h"
 
 using namespace JOYSTICK;
 using namespace PERIPHERALS;
@@ -149,10 +150,20 @@ bool CPeripheral::Initialise(void)
     return bReturn;
 
   g_peripherals.GetSettingsFromMapping(*this);
-  m_strSettingsFile = StringUtils::Format("special://profile/peripheral_data/%s_%s_%s.xml",
-                                          PeripheralTypeTranslator::BusTypeToString(m_mappedBusType),
-                                          m_strVendorId.c_str(),
-                                          m_strProductId.c_str());
+  if (m_iVendorId == 0x0000 && m_iProductId == 0x0000)
+  {
+    m_strSettingsFile = StringUtils::Format("special://profile/peripheral_data/%s_%s.xml",
+                                            PeripheralTypeTranslator::BusTypeToString(m_mappedBusType),
+                                            CUtil::MakeLegalFileName(m_strDeviceName, LEGAL_WIN32_COMPAT).c_str());
+  }
+  else
+  {
+    m_strSettingsFile = StringUtils::Format("special://profile/peripheral_data/%s_%s_%s.xml",
+                                            PeripheralTypeTranslator::BusTypeToString(m_mappedBusType),
+                                            m_strVendorId.c_str(),
+                                            m_strProductId.c_str());
+  }
+
   LoadPersistedSettings();
 
   for (unsigned int iFeaturePtr = 0; iFeaturePtr < m_features.size(); iFeaturePtr++)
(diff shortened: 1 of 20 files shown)


garbear:
This feature is included in my newest [retroplayer builds](http://forum.kodi.tv/showthread.php?tid=173361) if anyone wants to test

garbear:
Added a commit to improve logging. I went a little verbose in the commit message, but it explains what's going on pretty well.

```
    The button map allows Kodi to translate driver input (anonymous button, hats
    and axes) into the button format of the requested controller type.
    
    When a device is connected, Kodi first checks if the device came from an
    add-on. If so, Kodi checks if the add-on provides button maps for its
    device. Otherwise, Kodi looks for ANY add-on that can map buttons.
```

garbear:
Seeking opinion: I added `CPeripheralUtils::SanitizeFileName()` that takes a peripheral name and replaces (arbitrary) invalid characters with '_'. Does Kodi already have something like this? If not, should this be moved to StringUtils?

garbear:
jenkins build this please

akva2:
https://github.com/xbmc/xbmc/blob/master/xbmc/Util.h#L111

garbear:
Exactly what I needed, thanks.

I noticed that it was easier to identify peripheral's setting files by name than VID/PID, so I added c4b6eb3 which makes the name always appear in the settings' file name.

garbear:
jenkins build this please

xbmc/peripherals/devices/Peripheral.cpp
```diff
@@ -149,10 +150,21 @@ bool CPeripheral::Initialise(void)
...
+                                            PeripheralTypeTranslator::BusTypeToString(m_mappedBusType),
+                                            CUtil::MakeLegalFileName(m_strDeviceName, LEGAL_WIN32_COMPAT).c_str());
+  }
+  else
+  {
+    m_strSettingsFile = StringUtils::Format("special://profile/peripheral_data/%s_%s_%s_%s.xml",
+                                            PeripheralTypeTranslator::BusTypeToString(m_mappedBusType),
+                                            m_strVendorId.c_str(),
+                                            m_strProductId.c_str(),
+                                            CUtil::MakeLegalFileName(m_strDeviceName, LEGAL_WIN32_COMPAT).c_str());
```
MilhouseVH: This changes the name of the RPi CEC settings file from `rpi_2708_1001.xml` to `rpi_2708_1001_CEC Adapter.xml` which is going to be a small support headache when users upgrading to Kodi 17 complain their CEC isn't working (because they're now back to default settings).

garbear: This is a good point. until we get migration in place, I recommend switching back to just vid/pid if they are known

MilhouseVH: Thanks @garbear.

http://sprunge.us/aRSA

What are the plans for migration, is this going to happen in Kodi 17? If not I'll submit a PR, otherwise I'll just temporarily add the patch to my LE builds.

garbear: it'll prolly look like this: check if pid_vid exists, if so use that, if not use pid_vid_name

MilhouseVH: So...

```
  if (m_iVendorId == 0x0000 && m_iProductId == 0x0000)
  {
    m_strSettingsFile = StringUtils::Format("special://profile/peripheral_data/%s_%s_%s_%s.xml",
                                            PeripheralTypeTranslator::BusTypeToString(m_mappedBusType),
                                            m_strVendorId.c_str(),
                                            m_strProductId.c_str(),
                                            CUtil::MakeLegalFileName(m_strDeviceName, LEGAL_WIN32_COMPAT).c_str());
  }
  else
  {
    m_strSettingsFile = StringUtils::Format("special://profile/peripheral_data/%s_%s_%s.xml",
                                            PeripheralTypeTranslator::BusTypeToString(m_mappedBusType),
                                            m_strVendorId.c_str(),
                                            m_strProductId.c_str());
  }
```
?

MilhouseVH: I wonder also if it might be a good idea to replace spaces in `m_strDeviceName` with underscores?

garbear: I was thinking of changing the second block: 

else
 {
// Backwards compatibility 
 m_strSettingsFile = (without name) 
if! CFile::Exists( m_strSettingsFile) 
 m_strSettingsFile = (with name) 
} 

I prefer to replace spaces (and non-url-safe characters) with underscores. that's what I do in peripheral.joystick. see https://github.com/kodi-game/peripheral.joystick/blob/master/src/storage/StorageUtils.cpp#L58

MilhouseVH: http://nmacleod.com/public/oebuild/patches/00_backward_compat_peripheral_settings_name_change.txt

I'll test this now.

Not sure what plan is here, will you PR this or shall I? I can use the patch in my builds until there's something more permanent.

garbear: I like how that looks. I only got a phone until devcon, and prolly little if any internet, so you'll have to pr

MilhouseVH: PR: https://github.com/xbmc/xbmc/pull/10393

Unfortunately `StringUtils::MakeSafeUrl()` doesn't exist in Kodi, so I've just used `StringUtils::Replace(... ' ', '_')` instead.

PR #10311, opened 2016-08-19
remove longpress from left/right keys

FernetMenta:
hardly anybody wants to sacrifice repeating seeks steps for rw/ff

### DIFF
--- a/system/keymaps/keyboard.xml
+++ b/system/keymaps/keyboard.xml
@@ -323,8 +323,6 @@
       <right>StepForward</right>
       <up>ChapterOrBigStepForward</up>
       <down>ChapterOrBigStepBack</down>
-      <left mod="longpress">Rewind</left>
-      <right mod="longpress">FastForward</right>
       <up mod="longpress">AudioNextLanguage</up>
       <down mod="longpress">NextSubtitle</down>
       <a>AudioDelay</a>


NedScott:
Why was this removed? Repeating seek steps don't make any sense because they'll change so fast that you won't know what seek step increments you are on. It's entirely unpractical.

Meanwhile, rw/ff support for low-button remotes has been requested several times.

NedScott:
To add, I don't even feel strongly about this. I'm normally the one to remove changes like this when people ask me, no matter how much I think they make sense. However, you've done it on a whim while I took nearly a year getting positive feedback on that keymap.

I guess you're now the gatekeeper of keymaps, too.

FernetMenta:
there are still a lot of users who have disabled additive seek and for this scenario key repeats do make sense.

FernetMenta:
to add: I did this because users complained about your change. btw: 2 team members have given positive votes on this PR. I rank this higher than a vote of a EX team member

NedScott:
oh wow, I was literally writing an apology when the page updated with your last comment, trying to insult me based on my team member status. Even if you think I don't count, my point was that a significant number of users wanted this change. Any time I propose a keymap change it is with a lot of testing, feedback, and consideration. Often, this means making changes that I don't want, but the users do want.

**Original:** okay, fair point, I didn't consider that. Sorry for the rush to judgement.

MilhouseVH:
Perhaps a non-adaptive variant of the `FastForward`/`Rewind` action might be useful (eg. `NonAdaptiveFastForward`/`NonAdaptiveRewind`), as this could be mapped regardless of whether adaptive seeking is in effect or not, and would make more sense as a longpress mapping.

The problem in this PR (and which I think @NedScott is trying to solve) is that the existing auto-repeat mappings are only useful when adaptive seeking is disabled, and these same auto-repeat mappings are entirely unusable when adaptive seeking is enabled (which also happens to be the default, so the default behaviour is unusable). 

Mapping the longpress behaviour to actions that don't change behaviour according to adaptive seeking might be a better long term solution.

PR #10313, opened 2016-08-19
rename librarysettings to mediasettings

ronie:
fix a few leftovers that were missed during the window renaming.

@jjd-uk


jjd-uk:
+1  Sorry forgot there might be other places using the window name.

PR #10315, opened 2016-08-20
Fix crash in slider control if range is zero

garbear:
If a slider control is created with no start or end values, they default to zero. The control's proportion is a percentage of its range, so a divide by zero occurs.

This change returns a proportion of 0.0 if the range is zero.


garbear:
jenkins build this please

PR #10317, opened 2016-08-21
DVDVideoCodecAmlogic: Do not H/W decode Hi10P

kszaq:
No Amlogic chip can hardware decode 10-bit H.264 properly, we should use s/w decoding.


fritsch:
Looks okay @codesnake please

codesnake:
Looks good to me.

PR #10318, opened 2016-08-21
Added 2 view modes with zoom that depends on AR

nicklauslittle:
I added 2 new video view modes that will attempt to zoom the video to match the screen height. If this zoom causes the width of the video to exceed the width of the screen by some percent (20% for one, 10% for the other), then the zoom is reduced to that amount. (This works well for 2.39:1 on a 16:9 screen or 16:9 videos on a 4:3 screen.) IMHO this provides a good compromise between using most of the screen and minimizing the amount of video cropped, while still maintaining the aspect ratio. (I couldn't find an easy way to do this currently without having to set a custom zoom for each video.)

### DIFF
--- a/xbmc/cores/IPlayer.h
+++ b/xbmc/cores/IPlayer.h
@@ -224,7 +224,9 @@ enum ViewMode {
   ViewModeStretch16x9,
   ViewModeOriginal,
   ViewModeCustom,
-  ViewModeStretch16x9Nonlin
+  ViewModeStretch16x9Nonlin,
+  ViewModeZoom120Width,
+  ViewModeZoom110Width
 };
 
 class IPlayer
--- a/xbmc/cores/VideoPlayer/VideoRenderers/BaseRenderer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/BaseRenderer.cpp
@@ -428,7 +428,7 @@ void CBaseRenderer::ManageRenderArea()
 
 void CBaseRenderer::SetViewMode(int viewMode)
 {
-  if (viewMode < ViewModeNormal || viewMode > ViewModeStretch16x9Nonlin)
+  if (viewMode < ViewModeNormal || viewMode > ViewModeZoom110Width)
     viewMode = ViewModeNormal;
 
   CMediaSettings::GetInstance().GetCurrentVideoSettings().m_ViewMode = viewMode;
@@ -536,6 +536,18 @@ void CBaseRenderer::SetViewMode(int viewMode)
     CDisplaySettings::GetInstance().SetPixelRatio(CMediaSettings::GetInstance().GetCurrentVideoSettings().m_CustomPixelRatio);
     CDisplaySettings::GetInstance().SetNonLinearStretched(CMediaSettings::GetInstance().GetCurrentVideoSettings().m_CustomNonLinStretch);
     CDisplaySettings::GetInstance().SetVerticalShift(CMediaSettings::GetInstance().GetCurrentVideoSettings().m_CustomVerticalShift);
+  }
+  else if (CMediaSettings::GetInstance().GetCurrentVideoSettings().m_ViewMode == ViewModeZoom120Width)
+  {
+    float fitHeightZoom = sourceFrameRatio * screenHeight / (info.fPixelRatio * screenWidth);
+    CDisplaySettings::GetInstance().SetPixelRatio(1.0f);
+    CDisplaySettings::GetInstance().SetZoomAmount(fitHeightZoom < 1.0f ? 1.0f : (fitHeightZoom > 1.2f ? 1.2f : fitHeightZoom));
+  }
+  else if (CMediaSettings::GetInstance().GetCurrentVideoSettings().m_ViewMode == ViewModeZoom110Width)
+  {
+    float fitHeightZoom = sourceFrameRatio * screenHeight / (info.fPixelRatio * screenWidth);
+    CDisplaySettings::GetInstance().SetPixelRatio(1.0f);
+    CDisplaySettings::GetInstance().SetZoomAmount(fitHeightZoom < 1.0f ? 1.0f : (fitHeightZoom > 1.1f ? 1.1f : fitHeightZoom));
   }
   else // if (CMediaSettings::GetInstance().GetCurrentVideoSettings().m_ViewMode == ViewModeNormal)
   {
--- a/xbmc/settings/MediaSettings.cpp
+++ b/xbmc/settings/MediaSettings.cpp
@@ -94,7 +94,7 @@ bool CMediaSettings::Load(const TiXmlNode *settings)
       scalingMethod = (int)VS_SCALINGMETHOD_LINEAR;
     m_defaultVideoSettings.m_ScalingMethod = (ESCALINGMETHOD)scalingMethod;
 
-    XMLUtils::GetInt(pElement, "viewmode", m_defaultVideoSettings.m_ViewMode, ViewModeNormal, ViewModeCustom);
+    XMLUtils::GetInt(pElement, "viewmode", m_defaultVideoSettings.m_ViewMode, ViewModeNormal, ViewModeZoom110Width);
     if (!XMLUtils::GetFloat(pElement, "zoomamount", m_defaultVideoSettings.m_CustomZoomAmount, 0.5f, 2.0f))
       m_defaultVideoSettings.m_CustomZoomAmount = 1.0f;
     if (!XMLUtils::GetFloat(pElement, "pixelratio", m_defaultVideoSettings.m_CustomPixelRatio, 0.5f, 2.0f))
--- a/xbmc/video/ViewModeSettings.cpp
+++ b/xbmc/video/ViewModeSettings.cpp
@@ -1,0 +1,111 @@
+/*
+ *      Copyright (C) 2016 Team Kodi
+ *      http://kodi.tv
+ *
+ *  This Program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2, or (at your option)
+ *  any later version.
+ *
+ *  This Program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with XBMC; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "ViewModeSettings.h"
+
+#include "cores/IPlayer.h"
+#include "guilib/LocalizeStrings.h"
+#include "settings/VideoSettings.h"
+
+struct ViewModeProperties
+{
+  int stringIndex;
+  int viewMode;
+  bool hideFromQuickCycle;
+  bool hideFromList;
+};
+
+#define HIDE_ITEM true
+
+/** The list of all the view modes along with their properties
+ */
+static const ViewModeProperties viewModes[] =
+{
+  { 630,   ViewModeNormal },
+  { 631,   ViewModeZoom },
+  { 39008, ViewModeZoom120Width },
+  { 39009, ViewModeZoom110Width },
+  { 632,   ViewModeStretch4x3 },
+  { 633,   ViewModeWideZoom },
+  { 634,   ViewModeStretch16x9 },
+  { 644,   ViewModeStretch16x9Nonlin, HIDE_ITEM, HIDE_ITEM },
+  { 635,   ViewModeOriginal },
+  { 636,   ViewModeCustom, HIDE_ITEM, HIDE_ITEM }
+};
+
+#define NUMBER_OF_VIEW_MODES (sizeof(viewModes) / sizeof(viewModes[0]))
+
+/** Gets the index of a view mode
+ *
+ * @param viewMode The view mode
+ * @return The index of the view mode in the viewModes array
+ */
+static int GetViewModeIndex(int viewMode)
+{
+  size_t i;
+
+  // Find the current view mode
+  for (i = 0; i < NUMBER_OF_VIEW_MODES; i++)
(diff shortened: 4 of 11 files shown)


nicklauslittle:
Here are some sample screen shots of what it looks like on content that is too wide for the screen. (On content where the width and height fit the screen no zooming is performed, i.e. it looks the same as the "Normal" setting.)

Normal:
normal

Zoom:
zoom

Zoom - 120% Width:
zoom 20

Zoom - 110% Width:
zoom 10

Razzeee:
Hey there and thanks for the submission!

xbmc/cores/VideoPlayer/VideoRenderers/BaseRenderer.cpp
```diff
@@ -537,6 +537,18 @@ void CBaseRenderer::SetViewMode(int viewMode)
     CDisplaySettings::GetInstance().SetNonLinearStretched(CMediaSettings::GetInstance().GetCurrentVideoSettings().m_CustomNonLinStretch);
     CDisplaySettings::GetInstance().SetVerticalShift(CMediaSettings::GetInstance().GetCurrentVideoSettings().m_CustomVerticalShift);
   }
+  else if (CMediaSettings::GetInstance().GetCurrentVideoSettings().m_ViewMode == ViewModeZoom120Width)
+  {
+    float fitHeightZoom = sourceFrameRatio * screenHeight / (info.fPixelRatio * screenWidth);
```
Razzeee: Can you move this part into a method, as it's the exact same one as below?

xbmc/cores/IPlayer.h
```diff
@@ -226,7 +226,9 @@ enum ViewMode {
   ViewModeStretch16x9,
   ViewModeOriginal,
   ViewModeCustom,
-  ViewModeStretch16x9Nonlin
+  ViewModeStretch16x9Nonlin,
+  ViewModeZoom120Width,
```
Razzeee: Just a small suggestion, but I thiink it would be more natural to have the 110 first and then the 120

xbmc/settings/MediaSettings.cpp
```diff
@@ -94,7 +94,7 @@ bool CMediaSettings::Load(const TiXmlNode *settings)
       scalingMethod = (int)VS_SCALINGMETHOD_LINEAR;
     m_defaultVideoSettings.m_ScalingMethod = (ESCALINGMETHOD)scalingMethod;
 
-    XMLUtils::GetInt(pElement, "viewmode", m_defaultVideoSettings.m_ViewMode, ViewModeNormal, ViewModeCustom);
+    XMLUtils::GetInt(pElement, "viewmode", m_defaultVideoSettings.m_ViewMode, ViewModeNormal, ViewModeZoom110Width);
```
Razzeee: I'm not sure if you did that on purpose (or if that was a bug on our side) but this also exposes ViewModeStretch16x9Nonlin on our gui. Which wasn't the case before, if I'm reading it right.

xbmc/cores/VideoPlayer/VideoRenderers/BaseRenderer.cpp
FernetMenta: nope, I consider it bad practice to write methods with two lines of code.

xbmc/cores/IPlayer.h
nicklauslittle: I originally had it that way, but I changed it to 120% first to match the progression of the spinner in the settings (Normal -> Zoom -> Zoom 120% width -> Zoom 110% width. To me this progression made the most sense, so that Normal and Zoom are first and then the 120% width zoom is smaller than full zoom for some content and then the 110% width zoom is smaller than that. Otherwise, it just looks like it flips back and forth when going through the spinner.). I can change it if you really want and no one else objects.

xbmc/cores/VideoPlayer/VideoRenderers/BaseRenderer.cpp
nicklauslittle: Yeah, two lines is kind of short. I usually wait until at least 4 lines (or if it is several times instead of just twice).

xbmc/settings/MediaSettings.cpp
nicklauslittle: I struggled with this particular bit of code for a little while. I didn't realize it affected the GUI, though. I thought it just loaded the video settings. I didn't want to change the numbering of any of the existing view modes as that could mess up existing settings. But I also saw that the last view mode was intentionally excluded from this. I honestly don't know enough about the code to make a good decision on this one. If you or anyone else has any suggestions, let me know.

xbmc/cores/IPlayer.h
Razzeee: No, if the progression is right, it's fine. Just read funny ;)

xbmc/settings/MediaSettings.cpp
nicklauslittle: I examined the code further and think I have figured out what is going on. It looks like the last view mode was accidentally excluded from the view mode spinner. I have added a new commit to try to better abstract the view mode order and text shown in the GUI from the constants that are stored in the settings database. This code can easily be tweaked, if necessary, to hide the non-linear mode.

nicklauslittle:
I think I have resolved most of the issues. Let me know if there is anything else. Thanks.

xbmc/settings/VideoSettings.cpp
```diff
@@ -86,3 +86,49 @@ bool CVideoSettings::operator!=(const CVideoSettings &right) const
   if (m_VideoStream != right.m_VideoStream) return true;
   return false;
 }
+
+int CVideoSettings::GetNextViewMode(int viewMode)
```
FernetMenta: I don't like those methods. Assume someone wants to insert another view mode in the middle. First the enum in iPlayer has to be changed, then this entire switch/case has to be reorganized.
An array that maps the enum to the string would fit better. Example: https://github.com/xbmc/xbmc/blob/master/xbmc/GUIInfoManager.cpp#L206

xbmc/settings/VideoSettings.h
```diff
@@ -61,8 +61,13 @@ class CVideoSettings
   bool m_StereoInvert;
   int m_VideoStream;
 
-  static int GetNextViewMode(int viewMode);
-  static int GetViewModeStringIndex(int viewMode);
+  typedef struct
```
FernetMenta: I see no need for a typedef.

xbmc/video/dialogs/GUIDialogVideoSettings.cpp
```diff
@@ -333,9 +333,9 @@ void CGUIDialogVideoSettings::InitializeSettings()
     entries.clear();
 
     // add all view modes
-    entries.push_back(std::make_pair(CVideoSettings::GetViewModeStringIndex(ViewModeNormal), ViewModeNormal));
-    for (int i = CVideoSettings::GetNextViewMode(ViewModeNormal); i != ViewModeNormal; i = CVideoSettings::GetNextViewMode(i))
-      entries.push_back(std::make_pair(CVideoSettings::GetViewModeStringIndex(i), i));
+	entries.push_back(std::make_pair(CVideoSettings::m_ViewModeProperties[ViewModeNormal].stringIndex, ViewModeNormal));
+	for (int i = CVideoSettings::m_ViewModeProperties[ViewModeNormal].nextViewMode; i != ViewModeNormal; i = CVideoSettings::m_ViewModeProperties[i].nextViewMode)
```
FernetMenta: should be possible to use 
for (auto &mode : CVideoSettings::m_ViewModeProperties)

xbmc/settings/VideoSettings.h
```diff
@@ -61,8 +61,13 @@ class CVideoSettings
...
+  {
+    const int stringIndex;
+    const int nextViewMode;
```
FernetMenta: why "next"? this array should defines pairs of modes and their stringId

FernetMenta:
Looks like we need to change the method how the user sets the mode first. From a user perspective it makes no sense to toggle through a list of 10 modes. If you accidentally press one time too often, you have to press the button for another 9 times.
From a developer perspective it does not make sense either. you started to build the design around this silly UI interaction method.

For mode selection bring up a list with available modes. This is also aligned with other settings we have.

xbmc/settings/VideoSettings.h
nicklauslittle: "next" represents the next view mode to display in the GUI. The enumeration order does not map well to the order in which the modes should be displayed in the GUI, IMO. The enumeration order also cannot be changed since the enumeration value is used when stored in the settings. Changing them would mess up everyone's view mode settings.

nicklauslittle:
The view mode can actually be set 2 different ways. First, it can be set the "normal way" using the video settings menu, with the spinner that is bi-directional, so you can go backwards. The second way, which I had no idea existed until I started working on this, is to cycle through the modes using the "Zoom" adjustment key (typically 'Z'). This only goes one direction, and I assume is there so someone can cycle through them quickly until the video content looks correct. Both of these work the exact same way they worked before I added the 2 new view modes.

I see several different options here: 1) Leave it how it is, 2) Add another field to the structure that specifies a subset of the view modes to cycle through when hitting the Zoom adjustment key, or 3) Add a reverse Zoom adjustment key that cycles through them backwards. If anyone has any other suggestions, let me know.

Razzeee:
Not that you can know, but we're actively trying to get rid of spinners (bi directional lists) for ux reasons. So if you can change it to actually show the select dialog it would be appreciated. But that should not block this from going in (current beta status might).

xbmc/settings/VideoSettings.h
FernetMenta: And exactly this is not good. You make this class dependent on the representation. VIdeoSettings should not care about the method the UI sets the mode

nicklauslittle:
Ah. That makes sense. I will try to look into this sometime soon. I'm guessing this would be something like an additional dialog that would pop up with a list of all the view modes?

As for the Zoom adjustment key, should I just leave as is? I was thinking that at least the Custom mode could be removed when cycling through the modes, since it does really do anything unless you adjust the other properties in the video settings dialog.

FernetMenta:
>  I'm guessing this would be something like an additional dialog that would pop up with a list of all the view modes?

Yes, check out this example: https://github.com/xbmc/xbmc/blob/9b312193ff99da42bdedb5e72c4240b0cd057a87/xbmc/pvr/addons/PVRClients.cpp#L896

> I was thinking that at least the Custom mode could be removed when cycling

cycling through modes can be removed

FernetMenta:
@nicklauslittle you are doing are good job here!

nicklauslittle:
Thanks! I decided not to remove the cycling through modes (I figured there are probably some people who use it and would be upset if I removed it.) But everything else, I tried to incorporate all the advice you guys have given me. Let me know if you find any issues.

xbmc/video/ViewModeSettings.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+/** Gets the next view mode for quick cycling through the modes
+ *
+ * @param viewMode The current view mode
+ * @return The next view mode
+ */
+int CViewModeSettings::GetNextQuickCycleViewMode(int viewMode)
+{
+  // Find the next quick cycle view mode
+  for (size_t i = GetViewModeIndex(viewMode) + 1; i < NUMBER_OF_VIEW_MODES; i++)
```
FernetMenta: please don't omit the curly brackets here. this can easily break later.

xbmc/video/ViewModeSettings.cpp
```diff
@@ -0,0 +1,108 @@
...
+{
+  return viewModes[GetViewModeIndex(viewMode)].stringIndex;
+}
+
+/** Fills the list with all visible view modes
+ */
+void CViewModeSettings::ViewModesFiller(const CSetting *setting, std::vector< std::pair<std::string, int> > &list, int &current, void *data)
+{
+  // Add all appropriate view modes to the list control
+  for (const auto&item : viewModes)
```
FernetMenta: white space after auto would be nice

FernetMenta:
@MartijnKaijser @Razzeee If you agree, I'll add this to beta2

xbmc/video/ViewModeSettings.cpp
```diff
@@ -0,0 +1,108 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with XBMC; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+// ViewModeSettings.cpp: implementation of the CViewModeSettings class.
```
MartijnKaijser: I'm sure we don't need these lines

xbmc/video/ViewModeSettings.cpp
```diff
@@ -0,0 +1,108 @@
+/*
+ *      Copyright (C) 2005-2016 Team XBMC
+ *      http://xbmc.org
+ *
```
MartijnKaijser: Only "2016 Team Kodi"
http://kodi.tv

xbmc/video/ViewModeSettings.h
```diff
@@ -0,0 +1,57 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with XBMC; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+// ViewModeSettings.h: interface for the CViewModeSettings class.
+//
+//////////////////////////////////////////////////////////////////////
+
+#pragma once
```
MartijnKaijser: Same in this file. Pragma is usually first line in the fil but doesn't really matter

MartijnKaijser:
@fernetmenta I'm fine if the comments you made are addressed.

FernetMenta:
@nicklauslittle please squash commits and rebase

FernetMenta:
jenkins build this please

MartijnKaijser:
@nicklauslittle thx

axbmcuser:
@nicklauslittle Great work with the new selection dialog! No toggle-madness anymore :)
Related trac.kodi.tv/ticket/16904

PR #9586, opened 2016-04-09
[pvr] improve value increase calls

AlwinEsch:
This request is from a old rework from me. This change the `++`where possible to the front and use `auto` for loops.

PR #10320, opened 2016-08-21
[PVR] Fix regression in CPVRRecordings::UpdateEpgTags.

ksooo:
Introduced by #9586, => https://github.com/xbmc/xbmc/commit/6ff0b2cf4ac1#diff-821f9fc69b834bb338f060b8fcb6e838R505

Symptoms: EPG tags for events which are still recording/got recorded had no recordings assigned after kodi restart (because due to the regression only deleted recording were ever assigned to epg tags.) Visible for example as missing play icon in epg grid.

@Jalle19 for review, please.


Jalle19:
Sounds sane, +1

ksooo:
jenkins build this please

PR #10322, opened 2016-08-22
[estuary] - show buffering progress in video full screen

aleksrozman:
The current estuary skin just has a rotating icon, with no indication how bad the streaming source is when it comes to filling the buffer. Suggesting to restore the progress indicator (using the volume progress images for now) and add a label so it is clear that the source is either buffering slow or fast. Grey font color may need tweaking, it could blend with the video but at least the progress bar will be enough.

### DIFF
--- a/addons/skin.estuary/1080i/VideoFullScreen.xml
+++ b/addons/skin.estuary/1080i/VideoFullScreen.xml
@@ -23,21 +23,26 @@
 				<height>100</height>
 				<texture colordiffuse="button_focus">dialogs/extendedprogress/loading-back.png</texture>
 			</control>
-<!-- 			<control type="image" id="1">
+			<control type="image" id="1">
 				<left>910</left>
 				<top>490</top>
 				<width>100</width>
 				<height>100</height>
 				<texture>dialogs/volume/progress/p$INFO[Player.CacheLevel].png</texture>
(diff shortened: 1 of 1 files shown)


BigNoid:
The code was commented out because it was unreliable according to @phil65 but I see no problem in adding that back. If there's a problem it should be fixed in core imo.
Text without some shade behind it will make it unreadable in most cases, especially when the text is grey, so not sure that is a good idea.
You did a PR to your own master branch and merged it and than did a PR. You should just create a PR upstream directly from the ```stealth flyer-estuary``` branch. Now there's an unnecessary merge commit in this PR.

stealthflyer:
Agreed, grey text probably isnt the best choice but it is the skin theme. I can test out tonight with the addition of <shadowcolor>text_shadow</shadowcolor>.  I did not have any stability issues with the commented code.

@BigNoid I'll close this pull request and resubmit a new one from my estuary branch unless you know of a better way to change the base branch.

BigNoid:
How about making the circle a bit bigger and have the percentage inside the circle and label ```Buffering...``` underneath?
There's no way to alter the base branch afaik.

stealthflyer:
Using PR https://github.com/xbmc/xbmc/pull/10325 instead

PR #10324, opened 2016-08-22
Use uint64_t to represent times for seeking

rbuehlma:
There are live streams out there (in my case Zattoo) where the PTS is the time in msec since 1970.  Therefore the "current time in msec" of the stream does not fit into a 32-bit integer.
This causes problems with seeking in case all the seeking functions use 32-bit integer instead of 64-bit.

It is probably too late to get such a change into v17 but can it be considered for v18?
If no, is there any alternative solution for such cases?
If yes, I most probably have missed impacted sections/functions. Please let me know what else would be required.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -3677,7 +3677,7 @@ void CApplication::OnPlayBackSpeedChanged(int iSpeed)
   CAnnouncementManager::GetInstance().Announce(Player, "xbmc", "OnSpeedChanged", m_itemCurrentFile, param);
 }
 
-void CApplication::OnPlayBackSeek(int iTime, int seekOffset)
+void CApplication::OnPlayBackSeek(uint64_t iTime, int seekOffset)
 {
 #ifdef HAS_PYTHON
   g_pythonParser.OnPlayBackSeek(iTime, seekOffset);
--- a/xbmc/Application.h
+++ b/xbmc/Application.h
@@ -175,7 +175,7 @@ public:
   virtual void OnPlayBackResumed() override;
   virtual void OnPlayBackStopped() override;
   virtual void OnQueueNextItem() override;
-  virtual void OnPlayBackSeek(int iTime, int seekOffset) override;
+  virtual void OnPlayBackSeek(uint64_t iTime, int seekOffset) override;
   virtual void OnPlayBackSeekChapter(int iChapter) override;
   virtual void OnPlayBackSpeedChanged(int iSpeed) override;
 
--- a/xbmc/ApplicationPlayer.cpp
+++ b/xbmc/ApplicationPlayer.cpp
@@ -280,7 +280,7 @@ bool CApplicationPlayer::SeekScene(bool bPlus)
   return (player && player->SeekScene(bPlus));
 }
 
-void CApplicationPlayer::SeekTime(int64_t iTime)
+void CApplicationPlayer::SeekTime(uint64_t iTime)
 {
   std::shared_ptr<IPlayer> player = GetInternal();
   if (player)
@@ -295,7 +295,7 @@ void CApplicationPlayer::SeekTimeRelative(int64_t iTime)
     // use relative seeking if implemented by player
     if (!player->SeekTimeRelative(iTime))
     {
-      int64_t abstime = player->GetTime() + iTime;
+      uint64_t abstime = player->GetTime() + iTime;
       player->SeekTime(abstime);
     }
   }
--- a/xbmc/ApplicationPlayer.h
+++ b/xbmc/ApplicationPlayer.h
@@ -161,7 +161,7 @@ public:
   int   SeekChapter(int iChapter);
   void  SeekPercentage(float fPercent = 0);
   bool  SeekScene(bool bPlus = true);
-  void  SeekTime(int64_t iTime = 0);
+  void  SeekTime(uint64_t iTime = 0);
   void  SeekTimeRelative(int64_t iTime = 0);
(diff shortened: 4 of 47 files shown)


rbuehlma:
The time seems to be an absolute time in most cases. For these, I did also change to value from signed to unsigned.

FernetMenta:
I don't think those streams are handled by ffmpeg. Please elaborate on the issue.

rbuehlma:
Actually I have the stream working for me, including seeking in the stream (using this patch and the other two PR which are rejected). Therefore, ffmpeg does not seem to be a problem.

After cleaning and re-compiling I now have this error/warning:

`Caught: java.lang.RuntimeException: WARNING: Cannot convert the return value of swig type uint64_t for the call XBMCAddon::xbmc::Player::onPlayBackSeek`

Do you have a hint what could be causing this?

garbear:
@rbuehlma not commenting on the state of the PR, but you could try deleting `xbmc/interfaces/python/generated`

rbuehlma:
Thanks, but unfortunately, this did not help. I did a fresh checkout and already with "./bootstrap" I get this error:

`[xbmc/interfaces/python/generated/AddonModuleXbmc.xml, xbmc/interfaces/python/PythonSwig.cpp.template, xbmc/interfaces/python/generated/AddonModuleXbmc.cpp]
Processing "xbmc/interfaces/python/PythonSwig.cpp.template" using the module specification for module "xbmc/interfaces/python/generated/AddonModuleXbmc.xml"
Caught: java.lang.RuntimeException: WARNING: Cannot convert the return value of swig type uint64_t for the call XBMCAddon::xbmc::Player::onPlayBackSeek
java.lang.RuntimeException: WARNING: Cannot convert the return value of swig type uint64_t for the call XBMCAddon::xbmc::Player::onPlayBackSeek
	at Helper.getOutConversion(Helper.groovy:232)
	at Helper$getOutConversion$16.callStatic(Unknown Source)
	at Helper.getOutConversion(Helper.groovy:230)
	at Helper.getOutConversion(Helper.groovy)
	at Helper$getOutConversion$13.call(Unknown Source)
	at SimpleTemplateScript1$_run_closure11$_closure49$_closure52.doCall(SimpleTemplateScript1.groovy:965)
	at SimpleTemplateScript1$_run_closure11$_closure49.doCall(SimpleTemplateScript1.groovy:960)
	at SimpleTemplateScript1$_run_closure11.doCall(SimpleTemplateScript1.groovy:948)
	at SimpleTemplateScript1.run(SimpleTemplateScript1.groovy:918)
	at Generator.run(Generator.groovy:69)
codegenerator.mk:37: recipe for target 'xbmc/interfaces/python/generated/AddonModuleXbmc.cpp' failed
make: *** [xbmc/interfaces/python/generated/AddonModuleXbmc.cpp] Error 1
`

FernetMenta:
unsigned is wrong anyway because we support negative times. you'd better implement an offset in your demuxer.

rbuehlma:
Changing to signed 64-bit would be fine as well (currently there is a mixup in the code, some functions already use unsigned for seek times).

The demuxer is part of inputstream.mpd. I guess it could be implemented there but wouldn't it be better to fix the problem in Kodi instead of doing a workaround in the Addon?

FernetMenta:
I don't see a problem with current code. Your addon may have a new requirement that we can discuss.
I am not convinced that we should change API for inputstream and pvr.

rbuehlma:
I could fix the problem above by using "long long" instead for the python interface. The "wrapper" does not know of uint64_t.

Would changing the types from unsigned to signed give this PR the chance to get into Kodi?
With this change, I could eliminate quite a few casts to int. This can't be that bad :)

FernetMenta:
The method is called SeekTime, not SeekPts. It seeks within the timespan presented to the user by the gui. Currently pvr bypasses player with its own start and end times but this will go away. Finally you will need an offset in demuxer anyway.

rbuehlma:
Is implementing this offset already possible?
What exactly would be the "presented timespan"?
- The timespan which is shown in the time-bar in the GUI (in my case the start/end of a TV-Program)
- The timespan a user can use to seek (in my case 7 days ago until now)
- a different time?

How are the offsets of the Demuxer and the one of the player aligned? Would the Demuxer need to adjust the PTS/DTS of each frame or is the PTS/DTS not relevant for this time?
Can this "presented timespan" change during playback (e.g. if a new TV-program changes)?

FernetMenta:
Yes, this is possible now. Check out what ffmpeg demuxer does. 
Seeking can be done by clicking in the progress bar shown on the osd. For pvr this is not finished yet: osd should have indicators for the entire timespan and markers for current show. Amswer: 2nd bullet, you can seek 7 days in your case.
Next question: yes the timespan can change. typical timeshift buffer is a ring buffer.

rbuehlma:
I implemented the offset as well, and it seems to be working.
Still in my opinion it would be more consistent to represent the times as 64bit values but it does not seem to be required for my case.
Shall I close this PR or would it be of interest, if I change the type to singed?

FernetMenta:
I agree that we have inconsistencies but changing all integers to 64 bit values is imo not the right solution. We don't want all integers to be 64bit, do we? 32 bit is enought for the timespan to seek

rbuehlma:
No, not all integers, but in my opinion, the ones that might exceed the 32bit size. The time is one of them. The biggest indication is probably when you need to cast a 64bit integer to a 32bit one.

But I agree that it is mostly academically in this case. I will therefore close this PR

PR #10326, opened 2016-08-23
Fix osd  not showing while watching tv

rjsachse:
Was reported on the forum bugs.


rjsachse:
Here the 2 links about this.
http://forum.kodi.tv/showthread.php?tid=267397
http://forum.kodi.tv/showthread.php?tid=262373&pid=2399281#pid2399281

ksooo:
Works like a charme. Thanks @rjsachse

PR #10327, opened 2016-08-23
[PVR] Fix epg tag <-> recording association

ksooo:
Fix epg tag <-> recording association (epg tags had not always the right recordings assigned).

Goto TV Guide, start a recording that will last some more minutes (at least)
=> resp. epg event gets marked with 'recording' icon
Quit Kodi, immediately restart Kodi
Goto TV Guide
=> resp. epg event does not have the 'recording' icon (although it still gets recorded)

Happens only in combination with async epg transfer (currently only supported by pvr.hts).

@Jalle19 I fixed the same problem for timer icons months ago, did forget about recordings at that time. Good to go?

### DIFF
--- a/xbmc/epg/Epg.cpp
+++ b/xbmc/epg/Epg.cpp
@@ -28,6 +28,7 @@
 #include "guilib/LocalizeStrings.h"
 #include "pvr/addons/PVRClients.h"
 #include "pvr/PVRManager.h"
+#include "pvr/recordings/PVRRecordings.h"
 #include "pvr/timers/PVRTimers.h"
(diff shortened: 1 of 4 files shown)


Jalle19:
+1

ksooo:
jenkins build this please

ksooo:
Travis error is not related to this PR.

The following tests FAILED:
	 64 - TestURIUtils.ReplaceExtension (SEGFAULT)
	317 - TestCrc32.Compute_1 (SEGFAULT)
Errors while running CTest

PR #10332, opened 2016-08-23
EGLNativeTypeAmlogic: Enable GUI scaling when framebuffer size is less than output res

kszaq:
For 4K output Kodi renders GUI at 1080p and this results in GUI covering only 1/4 screen.

Example: https://forum.libreelec.tv/thread-1355-post-9983.html#pid9983

This patch enables hardware upscaling if output resolution is greater than rendered GUI resolution.

Another solution would be to render GUI at full resolution but there are 2 disadvantages:
- more reserved memory is needed for framebuffer
- rendering performance is not satisfactory at 4K

### DIFF
--- a/xbmc/windowing/egl/EGLNativeTypeAmlogic.h
+++ b/xbmc/windowing/egl/EGLNativeTypeAmlogic.h
@@ -53,6 +53,8 @@ public:
 protected:
   bool SetDisplayResolution(const char *resolution);
   void SetupVideoScaling(const char *mode);
+  void DealWithScale(const RESOLUTION_INFO &res);
+  void EnableFreeScale(const RESOLUTION_INFO &res);
   void DisableFreeScale();
 
 private:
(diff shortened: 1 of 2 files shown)


xbmc/windowing/egl/EGLNativeTypeAmlogic.h
```diff
@@ -53,6 +53,8 @@ class CEGLNativeTypeAmlogic : public CEGLNativeType
 protected:
   bool SetDisplayResolution(const char *resolution);
   void SetupVideoScaling(const char *mode);
+  void DealWithScale(const RESOLUTION_INFO &res);
```
fritsch: Can be private, right?

fritsch:
@codesnake that looks not clean in my opinion, please review as it is "your" platform :-)

xbmc/windowing/egl/EGLNativeTypeAmlogic.h
kszaq: `EnableFreeScale` and `DisableFreeScale` can be private as well. I made it protected to match `DisableFreeScale`. Should I move all three to private?

MartijnKaijser:
@kszaq please do not refer this PR in a commit message cause that will spam the PR

codesnake:
@kszaq Can you confirm this PR handles correctly all video view modes (I mean Normal, Zoom, Wide Zoom etc)?

PR #10333, opened 2016-08-23
[PVR] Recordings window: Respect setting SETTING_MYVIDEOS_SELECTACTIO…

ksooo:
Recordings window: Respect setting SETTING_MYVIDEOS_SELECTACTION also for recordings. Fixes inconsistent 'select item' behavior wrt 'normal' videos and recordings.

Reported here: http://forum.kodi.tv/showthread.php?tid=249318

@Jalle19: next one for review, if you like.

### DIFF
--- a/xbmc/pvr/windows/GUIWindowPVRRecordings.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRRecordings.cpp
@@ -262,11 +262,34 @@ bool CGUIWindowPVRRecordings::OnMessage(CGUIMessage &message)
           {
             case ACTION_SELECT_ITEM:
             case ACTION_MOUSE_LEFT_CLICK:
+              switch(CSettings::GetInstance().GetInt(CSettings::SETTING_MYVIDEOS_SELECTACTION))
+              {
+                case SELECT_ACTION_CHOOSE:
+                  OnPopupMenu(iItem);
+                  bReturn = true;
+                  break;
+                case SELECT_ACTION_PLAY_OR_RESUME:
+                  PlayFile(m_vecItems->Get(iItem).get(), false, true);
+                  bReturn = true;
+                  break;
+                case SELECT_ACTION_RESUME:
+                {
+                  CFileItemPtr item(m_vecItems->Get(iItem));
+                  const std::string resumeString = GetResumeString(*item);
+                  item->m_lStartOffset = resumeString.empty() ? 0 : STARTOFFSET_RESUME;
+                  PlayFile(item.get(), false, false);
+                  bReturn = true;
+                  break;
+                }
+                case SELECT_ACTION_INFO:
+                  ShowRecordingInfo(m_vecItems->Get(iItem).get());
+                  bReturn = true;
+                  break;
+              }
+              break;
             case ACTION_PLAY:
-            {
               bReturn = PlayFile(m_vecItems->Get(iItem).get());
               break;
-            }
             case ACTION_CONTEXT_MENU:
             case ACTION_MOUSE_RIGHT_CLICK:
               OnPopupMenu(iItem);


xbmc/pvr/windows/GUIWindowPVRRecordings.cpp
```diff
@@ -262,11 +262,33 @@ bool CGUIWindowPVRRecordings::OnMessage(CGUIMessage &message)
...
             case ACTION_SELECT_ITEM:
             case ACTION_MOUSE_LEFT_CLICK:
+              switch(CSettings::GetInstance().GetInt(CSettings::SETTING_MYVIDEOS_SELECTACTION))
+              {
+                case SELECT_ACTION_CHOOSE:
+                  OnPopupMenu(iItem);
+                  bReturn = true;
+                  break;
+                case SELECT_ACTION_PLAY_OR_RESUME:
+                  PlayFile(m_vecItems->Get(iItem).get(), false, true);
```
Jalle19: Perhaps the value of `m_vecItems->Get(iItem).get()` could be stored before the `switch` statement begins since it's used 3-4 times?

ksooo: Good point for <code>case ELECT_ACTION_RESUME</code> below, where it is used **multiple times in one case**. Will change this.

ksooo:
@Jalle19 good to go now?

xbmc/pvr/windows/GUIWindowPVRRecordings.cpp
```diff
@@ -262,11 +262,34 @@ bool CGUIWindowPVRRecordings::OnMessage(CGUIMessage &message)
           {
             case ACTION_SELECT_ITEM:
             case ACTION_MOUSE_LEFT_CLICK:
+              switch(CSettings::GetInstance().GetInt(CSettings::SETTING_MYVIDEOS_SELECTACTION))
```
Jalle19: @ksooo I was thinking of adding it above this line instead.

ksooo: I know but I think this does not make too much sense, because we then would do the call for every ACTION_*, even for those we don't handle.

Jalle19: Okay, it's fine as it is then.

ksooo:
jenkins build this please

MilhouseVH:
According to a [forum comment](http://forum.kodi.tv/showthread.php?tid=269814&pid=2401522#pid2401522), "recording folders don't seem to work under liveTV", which I'm assuming is due to this change (as there's no other relevant change since the last working build).

ksooo:
@MilhouseVH Thanks for the heads up. I will take look later today.

axbmcuser:
With default select action set to "Show information", KODI now opens the Information Dialog of a recording (which is correct).

When i hit "play recording" within the info dialog, the dialog closes and opens up again, becuase Estuary does Dialog.close(PVRRecordingInfo,true) & Action(Select). I hope it hasn't been reported before. Thanks.

ksooo:
Not related to this PR. Open a TRACK ticket if you find a new issue, please.

ksooo:
On second thought this actually is related to this PR, although an Estuary problem. Thanks for reporting.

axbmcuser:
I noticed it before and after this PR got merged - combined with the intention of this PR it made sense to me, that it's most likely related.

Side note: Replacing Action(Select) with Action(Play) in Estuary did not solve the issue in my quick tests. (Nothing happens with  Action(Play))
But that was just a lucky guess - most likely that's the wrong approach and someone got another fix in mind.

ksooo:
Using "Action(Play)" is the right approach; unfortunately currently this reveals another bug. I just tracked it down and will come up with a fix.

axbmcuser:
@ksooo

This time i can't tell if truly related, but something quite bad regarding the recordings-folder i noticed recently is:

Reproduce 1 (almost works every time in my case):
- Play LiveTV (in my case via tvheadend)
- While playback in background, open the discussed "recording" folder
- Background playback of LiveTv freezes (Complete KODI GUI still active, just the LiveTV playback is stopped - sometimes Estuary showes the buffering indicator in this state, but playback never resumes playback)

Reproduce 2 (if example 1 does not reproduce):
- Play LiveTV (in my case via tvheadend)
- While playback in background, open "recording" folder
- Open Information Dialog on any recording
- Background playback of LiveTv freezes (Complete KODI GUI still active, just the LiveTV playback is stopped - sometimes Estuary showes the buffering indicator in this state, but playback never resumes playback)

Debug log:
http://pastebin.com/raw/Gyk59ALa

Tract is down (404), i'm happy to move the report there, once it's back. 

Additional Info: Tested on LibreELEC nightly (debug log), additional testing with Millhouse LibreELEC #0828 + Windows win32 nightly 608220d all with same error/result)

ksooo:
This is definitely not related to this PR. please open a track ticket.

ksooo:
@FernetMenta fyi, i quickly debugged this. When opening the recordings window, a thumbextractor gets created, but not by pvr code! You know the consequences... The playing Live TV channel gets stopped.

<pre>
(lldb) thread backtrace
* thread #35: tid = 0xd6ea2, 0x0000000116f04524 pvr.hts.3.4.6.dylib`::CloseLiveStream() + 4 at client.cpp:322, name = 'JobWorker', stop reason = breakpoint 1.1
  * frame #0: 0x0000000116f04524 pvr.hts.3.4.6.dylib`::CloseLiveStream() + 4 at client.cpp:322
    frame #1: 0x0000000100d6b6ea Kodi`PVR::CPVRClient::CloseStream(this=0x0000000109ba0800) + 58 at PVRClient.cpp:1817
    frame #2: 0x00000001015de583 Kodi`PVR::CPVRClients::CloseStream(this=0x0000000106e3a710) + 99 at PVRClients.cpp:1345
    frame #3: 0x00000001016b0dca Kodi`PVR::CPVRManager::CloseStream(this=0x000000010705f000) + 474 at PVRManager.cpp:1204
    frame #4: 0x00000001017911de Kodi`CDVDInputStreamPVRManager::Close(this=0x0000000108c35bd0) + 110 at DVDInputStreamPVRManager.cpp:233
    frame #5: 0x000000010178f062 Kodi`CDVDInputStreamPVRManager::~CDVDInputStreamPVRManager(this=0x0000000108c35bd0) + 66 at DVDInputStreamPVRManager.cpp:63
    frame #6: 0x000000010178f185 Kodi`CDVDInputStreamPVRManager::~CDVDInputStreamPVRManager(this=0x0000000108c35bd0) + 21 at DVDInputStreamPVRManager.cpp:62
    frame #7: 0x000000010178f1e9 Kodi`CDVDInputStreamPVRManager::~CDVDInputStreamPVRManager(this=0x0000000108c35bd0) + 25 at DVDInputStreamPVRManager.cpp:62
    frame #8: 0x0000000100bec1fc Kodi`CDVDFileInfo::GetFileStreamDetails(pItem=0x000000011b5eb3d0) + 1868 at DVDFileInfo.cpp:351
    frame #9: 0x00000001018f6b4c Kodi`CThumbExtractor::DoWork(this=0x000000011b5eb390) + 5068 at VideoThumbLoader.cpp:140
    frame #10: 0x0000000100cae9df Kodi`CJobWorker::Process(this=0x000000011639cf10) + 111 at JobManager.cpp:69
    frame #11: 0x000000010094b080 Kodi`CThread::Action(this=0x000000011639cf10) + 896 at Thread.cpp:221
    frame #12: 0x0000000100949789 Kodi`CThread::staticThread(data=0x000000011639cf10) + 969 at Thread.cpp:131
    frame #13: 0x0000000106cc1805 libsystem_pthread.dylib`_pthread_body + 131
    frame #14: 0x0000000106cc1782 libsystem_pthread.dylib`_pthread_start + 168
    frame #15: 0x0000000106cbefa1 libsystem_pthread.dylib`thread_start + 13
(lldb) 
</pre>

ksooo:
@FernetMenta Happens only if you select a recording folder. Then, for this folder on the right half of the recordings window a "folder listing" is displayed. I think the directory provider behind this listing tries to extract the thumbnails.

screenshot002

ksooo:
@FernetMenta yep, the code defining the folder listing ui (around https://github.com/xbmc/xbmc/blob/master/addons/skin.estuary/1080i/MyPVRRecordings.xml#L252) is causing the trouble. I removed it for teting pupose and as expected Live TV no longer gets stopped when openeing the recordings window.

Bad thing is that I don't have a good ida for a fix right now. Maybe we should not do the folder listing if pvr is playing live tv or a recording? What do you think?

axbmcuser:
@ksooo 
Another "recordings"-folder issue i noticed (not sure if you get notified by tract since it's assigned to "opdenkamp")
http://trac.kodi.tv/ticket/16866
Maybe - if you got the time - you want to take a look. :) Thanks, as always.

ksooo:
@axbmcuser May I kindly ask you not to misuse random unrelated github PRs to contact me personally. Every github comment generates an email for every Team Kodi member. I check trac on a regular base, btw. Thanks.

axbmcuser:
Understood.

PR #10253, opened 2016-08-07
[estuary] skin sync

BigNoid:
adds the following:
- Show channel number, channel name, episode name in second row of topbar, when playing live tv or per recordings by ksooo
- MyPVRRecordings: Use Listitem.ActualIcon to ensure channel logo is displayed by ksooo
- Fix for seeklabel by ronie
- Remove media source from estuary settings window by jjd-uk
- Adds Kodi repository as widget by HitcherUK
- Fix timeseek label by Voyager1
- Drop Player.SeekOffset

PR #10336, opened 2016-08-24
[estuary] revert addition of repository widget on home screen

tamland:
Reverting because:
-there's no refresh logic implement
-causes 'random' modal popus on home screen 
-it violate the update policy setting user has set

As mentioned in #10253 this path is solely intended for internal use in the addon browser. That's why it behaves this way. A new endpoint retrieving the repository content have to be added and designed for use in directory providers; we cant just throw arbitrary urls in it.

### DIFF
--- a/addons/skin.estuary/1080i/Home.xml
+++ b/addons/skin.estuary/1080i/Home.xml
@@ -450,24 +450,10 @@
 							<param name="sortorder" value="descending"/>
 							<param name="list_id" value="8500"/>
 							<param name="onup_id" value="8400"/>
-							<param name="ondown_id" value="8600"/>
-							<param name="single_label" value="$INFO[ListItem.Label]"/>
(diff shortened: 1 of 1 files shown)


BigNoid:
If it behaves like that, yeah remove it.

ronie:
thanx for the explanation :)

@tamland 
can you change the id on this line to 20001
https://github.com/xbmc/xbmc/pull/10336/commits/4ce0d4d67f5478a1735b073b242ab73574cbf661#diff-6a8a015fcc8e96e1ccbdc9626e1cf61eL453

tamland:
Will do!

If you've seen the various 'could not connect to repository' reports in the forums, this would be it fyi.
If nobody beats me to it I'll look into adding a proper endpoint for this, but right now I don't have a lot of spare time and there are a lot of other issues that are more important to get done first...

PR #10337, opened 2016-08-24
Ticket#16638 - No proper handling of WM_DISPLAYCHANGE and related events 

marianob85:
Should resolve #16638 ticket

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -336,14 +336,16 @@ bool CApplication::OnEvent(XBMC_Event& newEvent)
         CApplicationMessenger::GetInstance().PostMsg(TMSG_QUIT);
       break;
     case XBMC_VIDEORESIZE:
-      if (g_windowManager.Initialized() &&
-          !g_advancedSettings.m_fullScreen)
+      if (g_windowManager.Initialized())
       {
         g_Windowing.SetWindowResolution(newEvent.resize.w, newEvent.resize.h);
-        g_graphicsContext.SetVideoResolution(RES_WINDOW, true);
-        CSettings::GetInstance().SetInt(CSettings::SETTING_WINDOW_WIDTH, newEvent.resize.w);
-        CSettings::GetInstance().SetInt(CSettings::SETTING_WINDOW_HEIGHT, newEvent.resize.h);
-        CSettings::GetInstance().Save();
+        if (!g_advancedSettings.m_fullScreen)
+        {
+          g_graphicsContext.SetVideoResolution(RES_WINDOW, true);
+          CSettings::GetInstance().SetInt(CSettings::SETTING_WINDOW_WIDTH, newEvent.resize.w);
+          CSettings::GetInstance().SetInt(CSettings::SETTING_WINDOW_HEIGHT, newEvent.resize.h);
+          CSettings::GetInstance().Save();
+        }
       }
       break;
     case XBMC_VIDEOMOVE:
@@ -352,8 +354,8 @@ bool CApplication::OnEvent(XBMC_Event& newEvent)
       {
         // when fullscreen, remain fullscreen and resize to the dimensions of the new screen
         RESOLUTION newRes = (RESOLUTION) g_Windowing.DesktopResolution(g_Windowing.GetCurrentScreen());
-        if (newRes != g_graphicsContext.GetVideoResolution())
-          CDisplaySettings::GetInstance().SetCurrentResolution(newRes, true);
+        CDisplaySettings::GetInstance().SetCurrentResolution(newRes, true);
+        g_graphicsContext.SetVideoResolution(g_graphicsContext.GetVideoResolution(), true);
       }
       else
 #endif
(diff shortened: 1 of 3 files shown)


MartijnKaijser:
@afedchin ping 
@marianob85 you seem to have used tabs instead of spaces which we use. Once changes are accepted please squash all commits into one and force push to the same branch so it updates this Pull request. Thx

marianob85:
Ok. I will fix it, but I would like to also test those changes for few days. Compiled version is also available on forum http://forum.kodi.tv/showthread.php?tid=262233&page=3 for user who also have experienced this problem ( maybe we get some feedback from others - in my case seems working correctly now ). 
I've create pull request so early to get feedback from kodi developers: maybe it's a better way to make it work properly. I've downloaded source code few days ago so it's hard to fix something without good knowledge of code :) 
I will commit requested changes in few days after some deeper home tests.

xbmc/windowing/windows/WinSystemWin32.h
```diff
@@ -153,6 +153,7 @@ class CWinSystemWin32 : public CWinSystemBase
   virtual bool Restore();
   virtual bool Hide();
   virtual bool Show(bool raise = true);
+  virtual void Resize();
```
Jalle19: Please add the `override` keyword.

marianob85: Nope. Still does not work for not-windowed full screen mode.

Jalle19: What does that have to do with adding the `override` keyword?

afedchin: XBMC_VIDEORESIZE is not handled in true fullscreen mode see https://github.com/xbmc/xbmc/blob/master/xbmc/Application.cpp#L340

afedchin: you can try this https://github.com/afedchin/xbmc/commit/6c7469abab8ce349c9010c17bed77a6bd8a5c8c0

marianob85:
> you can try this afedchin@ 6c7469a 
Nope. Doesn't work for not windowed full screen ( work for 'use full screen window )

Any clue what else can I check?

marianob85:
Based on afedchin src another prototype. 

Checked for normal full screen and using window mode. Work for both

xbmc/Application.cpp
```diff
@@ -336,24 +336,27 @@ bool CApplication::OnEvent(XBMC_Event& newEvent)
...
-        CSettings::GetInstance().SetInt(CSettings::SETTING_WINDOW_HEIGHT, newEvent.resize.h);
-        CSettings::GetInstance().Save();
-      }
-      break;
+      if (g_windowManager.Initialized())
+        {
+          g_Windowing.SetWindowResolution(newEvent.resize.w, newEvent.resize.h);
+          if (!g_advancedSettings.m_fullScreen)
+          {
+            g_Windowing.SetWindowResolution(newEvent.resize.w, newEvent.resize.h);
```
afedchin: no need to set resolution twice

marianob85:
Ok. Seems to be working.

xbmc/windowing/windows/WinEventsWin32.cpp
```diff
@@ -667,6 +667,25 @@ LRESULT CALLBACK CWinEventsWin32::WndProc(HWND hWnd, UINT uMsg, WPARAM wParam, L
         m_pEventFunc(newEvent);
       }
       return(0);
+    case WM_DISPLAYCHANGE:
```
afedchin: you should also update the list of supported resolutions. because it may be a new display which may not support some or all resolutions from old display.

wiromare:
hi @marianob85 
will this fix also the problem with a dual monitor setup?

I have an yamaha 475 and after power on kodi is moved on pc monitor and then it stays there, i had this issue some time ago, a while before to put my hand on the code for others reason and i completely forget this

currently i'm using an optical cable to leave the hdmi connected direct to the tv,

unfortunately now the cables are all hidden and it's difficult to me to test this PR for a dual monitor setup

marianob85:
Hard to say, theoretically should :)

afedchin:
please squash commits.

afedchin (changes requested):
please fix comments

xbmc/Application.cpp
```diff
@@ -336,24 +336,26 @@ bool CApplication::OnEvent(XBMC_Event& newEvent)
...
-      {
-        g_Windowing.SetWindowResolution(newEvent.resize.w, newEvent.resize.h);
-        g_graphicsContext.SetVideoResolution(RES_WINDOW, true);
-        CSettings::GetInstance().SetInt(CSettings::SETTING_WINDOW_WIDTH, newEvent.resize.w);
-        CSettings::GetInstance().SetInt(CSettings::SETTING_WINDOW_HEIGHT, newEvent.resize.h);
-        CSettings::GetInstance().Save();
-      }
-      break;
+      if (g_windowManager.Initialized())
+        {
```
marianob85: All should be correct now

afedchin (approved):

afedchin:
jenkins build this please

PR #10338, opened 2016-08-24
Revert "VideoPlayer: move supported deinterlacing methods into ProcessInfo

fritsch:
This reverts commit 4a22da3fe41baab0c075b5626a73f89b63cf1398.

Sorry - I don't have the time currently to debug this issue properly. Whenever the first loop checks for deint methods, VAAPI Picture is not yet available, therefore the Supports method is not delegated to the proper class. We end up later with the Auto method only.


fritsch:
Fernet will fix :-)

PR #10339, opened 2016-08-25
VideoPlayer: fix and cleanup deinterlacing methods

FernetMenta:
I wanted to do this for v18 but it is broken now and it makes no sense to implement another work-around for the old design flaw. Defining de-interlacing methods does not belong to renderer. It is defined by video decoder and platform. This change simplifies this:
- SW decoding with ffmpeg provides the basic methods: none, yadif, yadif-half
- platforms can override and add additional like render_bob, blend, etc.
- hw decoders define their own

@Memphiz could you add platform overrides for osx/ios? OSX can add render bob and bland, ios, render_bob
@afedchin could you please do the required changes for Windows
@fritsch vaapi, vdpau
@popcornmix mmal
@mk01 imx

Android is same as iOS or other platform that use GLES renderer.


fritsch:
There is some thing, that might confuse users: Whenever you play progressive only content, you only see Yadif and Yadif half, as those are the fallback methods we have when hw decoder does not overwrite. Is that intended?

xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp
```diff
@@ -150,21 +150,19 @@ enum AVPixelFormat CDVDVideoCodecFFmpeg::GetFormat( struct AVCodecContext * avct
   const char* pixFmtName = av_get_pix_fmt_name(*fmt);
 
   // if frame threading is enabled hw accel is not allowed
-  if(ctx->m_decoderState != STATE_HW_SINGLE)
+  // 2nd condition:
```
Memphiz: This changeset looks unrelated

Memphiz:
@FernetMenta i am unsure what needs to be overiddn ... Only the new method in processinfo ? (You already took care of the vtb hw decoder as i see)...

xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp
FernetMenta: this block had a return statement and would have required to set deint methods before return.

FernetMenta:
@Memphiz yes, this method

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererVTBGLES.h
```diff
@@ -39,7 +39,6 @@ class CRendererVTB : public CLinuxRendererGLES
 
   // Feature support
   virtual bool Supports(EINTERLACEMETHOD method) override;
-  virtual EINTERLACEMETHOD AutoInterlaceMethod() override;
```
Memphiz: the supports method needs to be removed here too i guess ... at least my compiler tells me that supports method is marked as override baut there is nothing in the base class to be overridden...

Memphiz: will pr the fix with the ios / osx overrides ...

FernetMenta: thanks

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

FernetMenta:
I added a method to query the default de-interlacing method. That is used by the settings dialog in order to display something useful if current method is not supported. Codecs are supposed to set their desired default method.

@afedchin is there anything to be done for Windows?

FernetMenta:
I will check Windows tomorrow. Then we can get this in. @afedchin is back after Sept. 10th and can verify.

objections?

FernetMenta:
jenkins build this please

MilhouseVH:
No VAAPI hardware acceleration after this change, and all VAAPI deinterlace modes are missing (and those that are present, don't work - everything is `yadif:1:-1:1` in PlayerProcessInfo). 

This had been working until a couple of days ago, before the cherry pick from https://github.com/FernetMenta/xbmc/pull/404.

Maybe this is important...
```
17:41:31  24.553089 T:140558567728896   ERROR: VAAPI::ConfigVAAPI - failed to init output
```
Debug log playing 720p: http://sprunge.us/CQNY

http://forum.kodi.tv/showthread.php?tid=269815&pid=2408161#pid2408161

afedchin:
@FernetMenta windows changes looks good, but. I didn't understand how it will work with sw decoding and dxva rendering. On some hw dxva deinterlacing are better than ffmpeg.

FernetMenta:
> I didn't understand how it will work with sw decoding and dxva rendering

for that purpose you can override processInfo and provide different or additional methods.

afedchin:
@FernetMenta I still don't see how to solve this. 
Only WinRenderer knows which methods are supported by itself.  It has no access to processInfo. 
1. There is no issue with DXVA decoder and any renderer it works as before.
2. with SW decoder there are three possible renderers: DXVAHD (which support deinterlacing) and PS/SW renderer which does not.

In my mind I see that in case SW decoding and DXVAHD renderer user should be able to select the following deinterlacing methods: None, Auto - which actually is DXVA (as before), DXVA and YADIFF. But I do not see how to access to processInfo from WinRenderer which knows actual rendering method.

FernetMenta:
I will think about it

PR #10340, opened 2016-08-25
[skin.estuary] Fix broken addon message

Hitcher:
Aligns the background texture and message with the addon icon.

screenshot094

@BigNoid @ronie @MartijnKaijser


BigNoid:
Ah missed that when doing the redesign. Thx!

ronie:
thx!

PR #10341, opened 2016-08-25
AESinkPULSE: Try to resolve channel layout - fallback if not possible fixes #16862

fritsch:
Since we try to avoid PA's upmixing strategy we might end up with a non supported output map that matches PA channels, but does not reach the actual layout. This avoids it, with a two step solution:

a) Try to map and resolve to the sound device's map
b) If we cannot open the required number of channels -> fallback to PA's remixing


fritsch:
jenkins build and merge

PR #10342, opened 2016-08-26
[PVR] Recordings window: Fix recording groups not working (regression introduced by #10333)

ksooo:
Fix recording groups not working (regression introduced by #10333).

@MilhoseVH fyi
@Jalle19 for review?

### DIFF
--- a/xbmc/pvr/windows/GUIWindowPVRRecordings.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRRecordings.cpp
@@ -258,6 +258,14 @@ bool CGUIWindowPVRRecordings::OnMessage(CGUIMessage &message)
         int iItem = m_viewControl.GetSelectedItem();
         if (iItem >= 0 && iItem < m_vecItems->Size())
         {
+          const CFileItemPtr item(m_vecItems->Get(iItem));
+          if (item->m_bIsFolder)
+          {
+            // recording folders are handled by base class.
+            bReturn = false;
+            break;
+          }
+
           switch (message.GetParam1())
           {
             case ACTION_SELECT_ITEM:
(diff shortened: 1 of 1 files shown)


MilhouseVH:
Ah missed this for tonight's build, will include it tomorrow assuming @Jalle19 finds no issues. Thanks.

Jalle19:
Too bad we didn't catch this last time. +1 if it fixes the regression.

ksooo:
Yeah I've learned now that PlayFile returns false if it is called with a folder item and this is the condition to call the base class which handles folders correctly. Imo bad design, should be much more clear now.

ksooo:
jenkins build this please

ksooo:
Travis error is not related to this PR

bas-t:
I see you fixed the resume option too.  But wasn't it supposed to ask wether to resume or start from the beginning?

ksooo:
I did not change resume behavior. What line of code in this PR you are referring to?

ksooo:
In case you mean "case SELECT_ACTION_RESUME:", there is no change in behavior, I just added a comment explaining what the two bool parameters are doing.

The abovementioned case is called when a user has the setting "default action when selecting video files" set to "resume". Then, no popup menu shall appear. If resume is not possible, the recording will automatically be played. Makes sense?

bas-t:
Resume was broken, I guessed that lines 280, 287, and 301 in xbmc/pvr/windows/GUIWindowPVRRecordings.cpp restored the function, as it is working again. Then again, I might be wrong.
Only thing: while it is working again, I'd like to have the option, as it was previously, to either resume or start from the beginning. That was a pop up dialog. That dialog does not show right now.

I see you beat me to it, while typing I receive your reply.
The use case you refer to is indeed what I mean

bas-t:
The context menue still offers me the choise I need (play from the beginning or play from the moment I quited), I wasn't aware of that.
While I still think it is a good choise to implement the previous behaviour: show that popup, I can do without it.
Kind regards,
Tycho.

ksooo:
Sorry, but i'm not aware that i have changed the behavior.

Media settings -> Videos -> Default select action = "Play"
+ select a recording with resume info => popup with two items "play from beginning" and "resume from xx:xx:xx" appears
+ select a recording without resume info => it will automatically be played from the beginning
=> this is exactly how select on a recording worked before this PR, no change in behavior

Media settings -> Videos -> Default select action = "Resume"
+ select a recording with resume info => it will be automatically started from the saved resume point
+ select a recording without resume info => it will automatically be played from the beginning

bas-t:
"select a recording with resume info => popup with two items "play from beginning" and "resume from xx:xx:xx" appears"

That does not happen atm, prreviously it did. Bottomline: what do I have to do to make this popup show? In a previous reply you stated that "Then, no popup menu shall appear.".

I  guess we have a communication problem.
Nevertheless, let's try to get on the same page.
Kind regards,
Tyucho.

ksooo:
> That does not happen atm, prreviously it did. Bottomline: what do I have to do to make this popup show?

Okay, this is my last attempt to explain it. There is a setting that controls the behavior. Open Kodi settings, Goto "Media settings" -> "Videos". Set the setting "Default select action" to "Play".

screenshot003

bas-t:
OK thanks, this was previously the behaviour when default action was set to resume. I obviously missed that.

ksooo:
"Previously" this setting was not respected by the recordings window. You could have set it to whatever you wanted, it would not have changed behavior of the recordings window. So, your statement that "this was behavior when set to 'resume' is just plain wrong, because "previously" recordings window and this setting were 100% unrelated. 

It is important to me that you get that, because everything you are writing here implies that now something with 'play' action on recordings is broken. It's not! Okay?

bas-t:
No, not OK. Everything I've written here implies only that I had it working (the resume or start from the beginning popup) if, and only if I set the default action to "resume" previously.
Where "previously" means like in Isengard, Jarvis and as I recall Krypton when it was in early alfa state.
It might be so that it never worked for you, regardless the setting of whatever default action, but that does not mean that it never worked for anyone else.
But let's not fight over this, we can simply conclude that my experience regarding the matter is not matching yours. Okay?

bas-t:
Moreover, I thank you very much for fixing the matter, it is very usefull to me in it's present state.

PR #10345, opened 2016-08-26
check for kodi.inputstream version

ghost:
see title

### DIFF
--- a/xbmc/addons/InputStream.cpp
+++ b/xbmc/addons/InputStream.cpp
@@ -36,8 +36,22 @@ std::unique_ptr<CInputStream> CInputStream::FromExtension(AddonProps props, cons
   std::string extensions = CAddonMgr::GetInstance().GetExtValue(ext->configuration, "@extension");
   std::string protocols = CAddonMgr::GetInstance().GetExtValue(ext->configuration, "@protocols");
   std::string name(ext->plugin->identifier);
+
+  bool matchAPIVersion(true);
+  for (unsigned int i(0); i < ext->plugin->num_imports; ++i)
+    if (strcmp(ext->plugin->imports[i].plugin_id, "kodi.inputstream") == 0 && strcmp(ext->plugin->imports[i].version, INPUTSTREAM_API_VERSION) != 0)
+    {
+      matchAPIVersion = false;
+      CLog::Log(LOGNOTICE, "CInputStream::FromExtension - %s (%s) not loaded due to API mismatch, has %s, wanted %s",
+        props.id.c_str(),
+        props.version.Upstream().c_str(),
+        ext->plugin->imports[i].version,
+        INPUTSTREAM_API_VERSION);
+      break;
+    }
+
   std::unique_ptr<CInputStream> istr(new CInputStream(props, name, listitemprops,
-                                                      extensions, protocols));
+                                                      extensions, protocols, matchAPIVersion));
   istr->CheckConfig();
   return istr;
 }
@@ -46,8 +60,10 @@ CInputStream::CInputStream(const AddonProps& props,
                            const std::string& name,
                            const std::string& listitemprops,
                            const std::string& extensions,
-                           const std::string& protocols)
+                           const std::string& protocols,
+                           bool matchAPIVersion)
 : InputStreamDll(std::move(props))
+, m_matchAPIVersion(matchAPIVersion)
 {
   m_fileItemProps = StringUtils::Tokenize(listitemprops, "|");
   for (auto &key : m_fileItemProps)
@@ -71,6 +87,9 @@ CInputStream::CInputStream(const AddonProps& props,
 
 bool CInputStream::CheckAPIVersion()
 {
+  if (!m_matchAPIVersion)
+    return false;
+
   std::string dllVersion = m_pStruct->GetApiVersion();
   if (dllVersion.compare(INPUTSTREAM_API_VERSION) != 0)
   {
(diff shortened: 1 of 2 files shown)


MilhouseVH:
This does the job here, many thanks!

Logging when there is an API mismatch might be a good idea, but not essential. We will include this as a kodi 17 beta1 backport in LibreELEC.

xbmc/addons/InputStream.cpp
```diff
@@ -36,8 +36,14 @@ std::unique_ptr<CInputStream> CInputStream::FromExtension(AddonProps props, cons
   std::string extensions = CAddonMgr::GetInstance().GetExtValue(ext->configuration, "@extension");
   std::string protocols = CAddonMgr::GetInstance().GetExtValue(ext->configuration, "@protocols");
   std::string name(ext->plugin->identifier);
+
+  bool matchAPIVersion(true);
+  for (unsigned int i(0); i < ext->plugin->num_imports; ++i)
+    if (strcmp(ext->plugin->imports[i].plugin_id,"kodi.inputstream")==0 && strcmp(ext->plugin->imports[i].version, INPUTSTREAM_API_VERSION))
+      matchAPIVersion = false;
```
da-anda: shouldn't we break the loop once we found the desired plugin_id to save a couple CPU cycles?

xbmc/addons/InputStream.cpp
```diff
@@ -36,8 +36,14 @@ std::unique_ptr<CInputStream> CInputStream::FromExtension(AddonProps props, cons
   std::string extensions = CAddonMgr::GetInstance().GetExtValue(ext->configuration, "@extension");
   std::string protocols = CAddonMgr::GetInstance().GetExtValue(ext->configuration, "@protocols");
   std::string name(ext->plugin->identifier);
+
+  bool matchAPIVersion(true);
+  for (unsigned int i(0); i < ext->plugin->num_imports; ++i)
+    if (strcmp(ext->plugin->imports[i].plugin_id,"kodi.inputstream")==0 && strcmp(ext->plugin->imports[i].version, INPUTSTREAM_API_VERSION))
```
fritsch: Please add != 0 to the second statement, as people not knowing what strcmp returns lead to confusions.

FernetMenta:
-1
no cpluff hacks for developer versions. absolutely no need for this.

MilhouseVH:
@FernetMenta I know where you're coming from and I'm sympathetic to that (and I don't even care if this isn't merged as we'll still use it) but please understand you now have thousands of installations that can't be upgraded easily because of this issue.

They may be testers and this may be alpha, but they all give their spare time just like everyone else working on Kodi so I don't think we should dump on them if we can avoid it, otherwise we'll have a lot less testers in future.

Please at least review this as if it were to be merged, even if ultimately it isn't merged as then it's down to distributions to carry this patch temporarily until test users have migrated to a non-crashing combination of Kodi 17 and inputstream.

FernetMenta:
I reviewed it and I already told you that I did think about a solution. But this change is a hack and should not be merged to master. LE can include it and drop it later.
It is appreciated that all those users contribute by testing alpha versions but we can't introduce a fully functional versioning system for developer branches. Every single hack like this will hit you at a later time with even more work to correct it.

mapfau:
I think it could be a good solution to check dependencies always on load time of whatever addons. Not only inputstream. Then the chance that kodi doesnt segfaults on version mismatch is much greater.
Unfortunately i haden't the time to look deeper up the stack trace.

From my personal experience on kodinerds: tons of hours were spend because of exactly this issue.
There are plenty of posts because of kodi crash in the time the api has changed. A nice log entry wich explains this would be better than a stack trace (log entry is missing here, sry).

If you @FernetMenta agree to the idea of a solution for this wich is not hacky and more generic, I culd look deeper. If you don't want such a checkforever reason, I can save my time.... Pls. let me know.

MartijnKaijser:
Doesn't https://github.com/xbmc/xbmc/pull/10335 partially solve it? At least when it's properly set in addon.xml

MilhouseVH:
@martihn my latest master build that includes #10335 continues to crash if the user has an older inputstream addon installed, so no (at least, not in #10335's current state).

tamland:
Consider #10335 a wip, but the proper solution for inputstream is to build on that and make something general that works for the entire addon system; not just adding even more binary addon/inputstream specific handling.

mapfau:
I'll look into the PR today

AlwinEsch:
Related to version check, I has also another way added on my rework (related to a API level system, not to a Version string), see https://github.com/AlwinEsch/kodi/commit/104a4ecca49e325a1b0a71c8f3636c1abeaa38fb

Not sure if my way becomes better, but another way.

mapfau:
@AlwinEsch this means that the api level is already defined in addons. Unfortunately the child has already fallen into the fountain. We try to fix already installed addons.

My knowledge about kodi versioning is not quite good, but I think there is a simple check of all the requirements missing. This should be checked before something is loaded.

My approach would be collecting all "importable" kernel modules in a map, and then let AddonLoader or DLLLoader or CPLuff or whatever makes sense let check all import statements

This one for example.
```
  <requires>
    <import addon="xbmc.core" version="0.1.0"/>
  </requires>
```
would be one wich has to be in the map. Unfortunately I'm not able to find some #define of this xbmc.core version in the kodi sources. Anybody an idea?

MartijnKaijser:
@mapfau is this still needed or obsolete with Addon manager rework of week ago?

mapfau:
closed -> it was just an idea

PR #10347, opened 2016-08-27
Miscellaneous keymap cleanup

NedScott:
These are all the little things I've changed that didn't seem big enough for their own PR:

Updated the header area to finally remove all those lingering XBMC references (some even mentioned the xbox!)

touchscreen.xml - Removed two-finger-swipe-left = seek(-7). It will now always be "back". This was originally added when remote.xml used to do smallstepback, but JMarshall thought back should always be back, and I agree. I just forgot to change touchscreen.xml

Joystick keymap had an unneeded < programs > section that was left over from a previous fix.

Added a couple of mod-combo actions to keyboard.xml for the sake of users who can't use longpress (Flirc, etc).

### DIFF
--- a/system/keymaps/customcontroller.AppleRemote.xml
+++ b/system/keymaps/customcontroller.AppleRemote.xml
@@ -1,33 +1,40 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- This file contains the mapping of keys (gamepad,remote and keyboard) to actions within XBMC -->
-<!-- The <global> section is a fall through - they will only be used if the button is not        -->
-<!-- used in the current window's section.  Note that there is only handling                     -->
-<!-- for a single action per button at this stage.                                               -->
-<!-- For joystick/gamepad configuration under linux/win32, see below as it differs from xbox     -->
-<!-- gamepads.                                                                                   -->
-
-<!-- The format is:                                                                              -->
-<!--    <device>                                                                                 -->
-<!--      <button id=""#>xbmc action</button>                                                    -->
-<!--    </device>                                                                                -->
-
-<!-- To map keys from other remotes using the RCA protocol, you may add <customcontroller name="AppleRemote"> blocks -->
-<!-- In this case, the tags used are <button id=""#> where # is the original button code (OBC) of the key -->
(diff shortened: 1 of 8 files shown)


NedScott:
Added mappings for the following for keyboard.xml

AudioNextLanguage
VolAmpUp
VolAmpDown
CreateBookmark
CreateEpisodeBookmark
PlayerControl(TempoUp)
PlayerControl(TempoDown)
Enter

garbear:
A user on our forum requested this change: https://github.com/garbear/xbmc/commit/0010463. Good idea?

NedScott:
That would make Back act like "escape" does on a keyboard. If you are in a directory of files or in a nested node in the video library, you end up going back to the home screen or whatever the previous "menu" was, rather than the previous directory.

garbear:
Is this better than having two buttons that do the same thing (B+Back)?

NedScott:
I thought it might be a problem when retroplayer is introduced later on, but maybe not? "Back" (the button) would exit retroplayer and "B" would become usable again for normal "back" (the action). I'll toss it in. At the very least, it won't hurt for Kodi v17.

NedScott:
I've also removed some lingering "My-" stuff for windows. Ping @ronie to make sure this is correct.

ronie:
yup, looking good.
some of the old MyFoo mappings likely didn't even work anymore.

perhaps you can add keymappings for the new TVTimerRules / RadioTimerRules windows as well?

NedScott:
Sure thing, done.

NedScott:
Anything else? good to merge? :)

NedScott:
Is there any reason this isn't going in yet? It fixes a few things for v17, and it would be really nice to see it included. Everything changed is very safe and should not have any issues. Especially now that tempo up/down is a feature for v17, people are asking how to add this to their keymaps.

I'll buy you beer or wash your car. Please? :)

Memphiz:
@NedScott can you retry the tempoup/down thing? Make sure sync to display is enabled for this test...

Memphiz:
@MartijnKaijser i think this is good stuff...

NedScott:
@Memphiz Thanks, that was what I didn't have set. Works for me as well now :)

PR #10348, opened 2016-08-27
[cmake] Fix lib/lib64 paths by using GNUInstallDirs module

fetzerch:
An attempt to let cmake decide the right install directories.
Problem reported in http://forum.kodi.tv/showthread.php?tid=288447.

Ping: @wsnipex and @hudokkow, would be great if you could test that.

Docs:
- https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html
- https://github.com/Kitware/CMake/blob/master/Modules/GNUInstallDirs.cmake


wsnipex:
thanks, that was on my list. Didn't test it yet, but it will most likely change our libdir location when prefix=/usr and therefore impact all our binary addons.

hudokkow:
I can only test this later this week. Back from vacation means a backlog of stuff todo.

fetzerch:
superseded by #10359

PR #10351, opened 2016-08-27
Cmake fixes for offline/disconnected builds

wsnipex:
Currently our cmake buildsys assumes internet access.
This PRs allows offline building by passing a local tarball for ffmpeg, libdvd\* and crossguid

These changes allow us to switch our official ubuntu PPA packaging to use cmake.
The last commit is the same workaround for i386 that have in autotools already.

@fetzerch @hudokkow 

PS: I originally wanted a macro to unify the FOO_URL handling, but then remembered that Find modules should be as portable as possible. Opinions?


wsnipex:
jenkins build this please

fetzerch:
was just reviewing it, lgtm!

PR #10353, opened 2016-08-28
Make "Next" visible for chapters in Estuary's OSD

NedScott:
Thanks to Guilouz and the users on the forum for catching this.

"Previous" on the OSD for Estuary is always visible (I assume because it can be used to "restart" a video), but "Next" has a condition to only show when there is a playlist with more than one item. I think this is a simple error, and it was forgotten that next/prev buttons are also used for chapter skipping, rather than an intentional design decision.

EDIT: updated to only show up for when there are chapters or playlists

### DIFF
--- a/addons/skin.estuary/1080i/VideoOSD.xml
+++ b/addons/skin.estuary/1080i/VideoOSD.xml
@@ -102,7 +102,7 @@
 								<param name="texture" value="osd/fullscreen/buttons/next.png"/>
 							</include>
 							<onclick>PlayerControl(Next)</onclick>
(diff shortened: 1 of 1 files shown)


NedScott:
ping @ronie @BigNoid @HitcherUK (sorry if I'm forgetting anyone that should be poked for skin stuff)

BigNoid:
It's a design decision to hide the next button when it has no use (eg clicking on it and nothing happens). So instead of removing that, I would add the following:
```[Player.ChapterCount | Integer.IsGreater(Playlist.Length(video),1)]```

NedScott:
Like this?

BigNoid:
Yep, looks good. Thx!

PR #10362, opened 2016-08-29
Fix query that was missing the actual join condition for unique_id

phate89:


phate89:
jenkins build this please

PR #10363, opened 2016-08-29
[PVR] Add IsRecordable, IsPlayable, GetEpgTagUrl to PVRClient

rbuehlma:
My IPTV provider allows recordings back in the past. As Kodi currently does not allow that, I would like to give the PVR addon the possibility to decide if a EPG entry can be recorded or not. If the addon does not implement the functions (returns PVR_ERROR_NOT_IMPLEMENTED), the old behaviour will be used.

Please let me know if there is a preferred way for this to be implemented. And lets hope I didn't touch a deferred feature again ;)

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -10500,7 +10500,13 @@ msgctxt "#19262"
 msgid "Parental control. Enter PIN:"
 msgstr ""
 
-#empty string with id 19263
+#. Label of a context menu entry to replay a programs
+#: xbmc/pvr/PVRContextMenuItem.cpp
+msgctxt "#19263"
+msgid "Replay program"
+msgstr ""
+
+#empty string with id 19264
 
 #. label for 'incorrect pin' error dialog header
 #: xbmc/pvr/PVRGUIActions.cpp
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -10367,7 +10367,9 @@ msgctxt "#19239"
 msgid "Starting background threads"
 msgstr ""
 
-#empty string with id 19240
+msgctxt "#19240"
+msgid "Replay"
+msgstr ""
 
 #. Label for context menu entry to open settings dialog for a timer (read-only)
 #: xbmc/pvr/PVRContextMenus.cpp
--- a/addons/skin.estouchy/xml/DialogPVRInfo.xml
+++ b/addons/skin.estouchy/xml/DialogPVRInfo.xml
@@ -198,6 +198,13 @@
 				<label>19165</label>
 				<visible>Window.IsActive(PVRGuideInfo)</visible>
 			</control>
+			<control type="button" id="10">
+				<description>Replay Program</description>
+				<width>230</width>
+				<include>ButtonInfoDialogsCommonValues</include>
+				<label>19240</label>
+				<visible>Window.IsActive(PVRGuideInfo)</visible>
+			</control>
 			<control type="button" id="6">
 				<description>Record</description>
 				<width>230</width>
--- a/xbmc/addons/PVRClient.cpp
+++ b/xbmc/addons/PVRClient.cpp
@@ -682,6 +682,28 @@ PVR_ERROR CPVRClient::IsRecordable(const CConstPVREpgInfoTagPtr &tag, bool *isRe
   return retVal;
 }
 
+bool CPVRClient::IsPlayable(const CConstPVREpgInfoTagPtr &tag)
+{
+  if (!m_bReadyToUse)
+    return false;
+
+  EPG_TAG pvrTag;
+  WriteEpgTag(tag, pvrTag);
+  return m_struct.toAddon.IsPlayable(pvrTag);
+}
+
+const std::string CPVRClient::GetEpgTagUrl(const CConstPVREpgInfoTagPtr &tag)
+{
+  if (!m_bReadyToUse)
+    return nullptr;
+
+  EPG_TAG pvrTag;
+  WriteEpgTag(tag, pvrTag);
+  char url[4096];
+  m_struct.toAddon.GetEpgTagUrl(pvrTag, url, sizeof(url));
+  return url;
+}
+
 void CPVRClient::CallMenuHook(const PVR_MENUHOOK &hook, const CFileItem *item)
 {
   if (!m_bReadyToUse)
--- a/xbmc/addons/PVRClient.cpp
+++ b/xbmc/addons/PVRClient.cpp
@@ -692,7 +692,7 @@ bool CPVRClient::IsPlayable(const CConstPVREpgInfoTagPtr &tag)
   return m_struct.toAddon.IsPlayable(pvrTag);
 }
 
-const std::string CPVRClient::GetEpgTagUrl(const CConstPVREpgInfoTagPtr &tag)
+const std::string CPVRClient::GetEpgTagUrl(const CConstPVREpgInfoTagPtr &tag, const CStringPropertyMapPtr &properties)
 {
   if (!m_bReadyToUse)
     return nullptr;
@@ -700,7 +700,7 @@ const std::string CPVRClient::GetEpgTagUrl(const CConstPVREpgInfoTagPtr &tag)
   EPG_TAG pvrTag;
   WriteEpgTag(tag, pvrTag);
   char url[4096];
-  m_struct.toAddon.GetEpgTagUrl(pvrTag, url, sizeof(url));
+  m_struct.toAddon.GetEpgTagUrl(pvrTag, url, sizeof(url), properties);
   return url;
 }
 
--- a/xbmc/addons/PVRClient.cpp
+++ b/xbmc/addons/PVRClient.cpp
@@ -30,6 +30,8 @@
 #include "dialogs/GUIDialogKaiToast.h"
 #include "events/EventLog.h"
 #include "events/NotificationEvent.h"
+#include "pvr/epg/Epg.h"
+#include "pvr/epg/EpgInfoTag.h"
 #include "filesystem/SpecialProtocol.h"
 #include "guilib/LocalizeStrings.h"
 #include "settings/AdvancedSettings.h"
@@ -369,6 +371,37 @@ void CPVRClient::WriteClientChannelInfo(const CPVRChannelPtr &xbmcChannel, PVR_C
   strncpy(addonChannel.strStreamURL, xbmcChannel->StreamURL().c_str(), sizeof(addonChannel.strStreamURL) - 1);
 }
 
+void CPVRClient::WriteEpgTag(const CConstPVREpgInfoTagPtr &tag, EPG_TAG &pvrTag)
+{
+  time_t t;
+  tag->StartAsUTC().GetAsTime(t);
+  pvrTag.startTime = t;
+  tag->EndAsUTC().GetAsTime(t);
+  pvrTag.endTime = t;
+  pvrTag.iParentalRating = tag->ParentalRating();
+  pvrTag.iUniqueBroadcastId = tag->UniqueBroadcastID();
+  pvrTag.bNotify = tag->Notify();
+  tag->FirstAiredAsUTC().GetAsTime(t);
+  pvrTag.firstAired = t;
+  pvrTag.iSeriesNumber = tag->SeriesNumber();
+  pvrTag.iEpisodeNumber = tag->EpisodeNumber();
+  pvrTag.iEpisodePartNumber = tag->EpisodePart();
+  pvrTag.iStarRating = tag->StarRating();
+  pvrTag.iYear = tag->Year();
+  pvrTag.iFlags = tag->Flags();
+  pvrTag.strTitle = tag->Title(true).c_str();
+  pvrTag.strPlotOutline = tag->PlotOutline().c_str();
+  pvrTag.strPlot = tag->Plot().c_str();
+  pvrTag.strOriginalTitle = tag->OriginalTitle(true).c_str();
+  pvrTag.strCast = tag->Cast().c_str();
+  pvrTag.strDirector = tag->Director().c_str();
+  pvrTag.strWriter = tag->Writer().c_str();
+  pvrTag.strIMDBNumber = tag->IMDBNumber().c_str();
+  pvrTag.strEpisodeName = tag->EpisodeName().c_str();
+  pvrTag.strIconPath = tag->Icon().c_str();
+  pvrTag.iChannelNumber = tag->ChannelTag()->ClientChannelNumber();
+}
+
 bool CPVRClient::GetAddonProperties(void)
 {
   std::string strBackendName, strConnectionString, strFriendlyName, strBackendVersion, strBackendHostname;
@@ -628,6 +661,24 @@ PVR_ERROR CPVRClient::RenameChannel(const CPVRChannelPtr &channel)
 
   PVR_ERROR retVal = m_struct.toAddon.RenameChannel(addonChannel);
   LogError(retVal, __FUNCTION__);
+  return retVal;
+}
+
+PVR_ERROR CPVRClient::IsRecordable(const CConstPVREpgInfoTagPtr &tag, bool *isRecordable)
+{
+  if (!m_bReadyToUse)
+    return PVR_ERROR_SERVER_ERROR;
+
+  if (!m_clientCapabilities.SupportsRecordings())
+    return PVR_ERROR_NOT_IMPLEMENTED;
+
+  PVR_ERROR retVal(PVR_ERROR_UNKNOWN);
+
+  EPG_TAG pvrTag;
+  WriteEpgTag(tag, pvrTag);
+  retVal = m_struct.toAddon.IsRecordable(pvrTag, isRecordable);
+  LogError(retVal, __FUNCTION__);
+
   return retVal;
 }
 
--- a/xbmc/addons/PVRClient.h
+++ b/xbmc/addons/PVRClient.h
@@ -362,6 +362,14 @@ namespace PVR
      */
     PVR_ERROR RenameChannel(const CPVRChannelPtr &channel);
 
+    /*
+     * @brief Check if an epg tag can be recorded
+     * @param tag The epg tag
+     * @param isRecordable Set to true if the tag can be recorded
+     * @return PVR_ERROR_NO_ERROR if isRecordable was set by the addon
+     */
+    PVR_ERROR IsRecordable(const CConstPVREpgInfoTagPtr &tag, bool *isRecordable);
+
     /*!
      * @return True if this add-on has menu hooks, false otherwise.
      */
@@ -806,6 +814,13 @@ namespace PVR
      * @param addonChannel The channel on the addon's side.
      */
     static void WriteClientChannelInfo(const CPVRChannelPtr &xbmcChannel, PVR_CHANNEL &addonChannel);
+
+    /*!
+     * @brief Copy over epg info from CPVREpgInfoTag to EPG_TAG.
+     * @param tag The epg tag on Kodi's side.
+     * @param pvrTag The epg tag on the addon's side.
+     */
+    static void WriteEpgTag(const CConstPVREpgInfoTagPtr &tag, EPG_TAG &pvrTag);
 
     /*!
      * @brief Whether a channel can be played by this add-on
--- a/xbmc/addons/PVRClient.h
+++ b/xbmc/addons/PVRClient.h
@@ -370,6 +370,24 @@ namespace PVR
      */
     PVR_ERROR IsRecordable(const CConstPVREpgInfoTagPtr &tag, bool *isRecordable);
 
+    /*
+     * Check if the given EPG tag can be played.
+     * @param tag The EPG tag
+     * @return True if the EPG tag can be played
+     * @remarks Required, always return false if not supported by the addon
+     *
+     */
+    bool IsPlayable(const CConstPVREpgInfoTagPtr &tag);
+
+    /*
+     * Get the URL to play a given EPG tag
+     * @param tag The EPG tag
+     * @return The url to play for the EPG tag, empty string if it cannot be played
+     * @remarks Required, always return empty string if not supported by the addon
+     *
+     */
+    const std::string GetEpgTagUrl(const CConstPVREpgInfoTagPtr &tag);
+
     /*!
      * @return True if this add-on has menu hooks, false otherwise.
      */
--- a/xbmc/addons/PVRClient.h
+++ b/xbmc/addons/PVRClient.h
@@ -382,11 +382,12 @@ namespace PVR
     /*
      * Get the URL to play a given EPG tag
      * @param tag The EPG tag
+     * @param properties The properties to be provided to the file item
      * @return The url to play for the EPG tag, empty string if it cannot be played
      * @remarks Required, always return empty string if not supported by the addon
      *
      */
-    const std::string GetEpgTagUrl(const CConstPVREpgInfoTagPtr &tag);
+    const std::string GetEpgTagUrl(const CConstPVREpgInfoTagPtr &tag, const CStringPropertyMapPtr &properties);
 
     /*!
      * @return True if this add-on has menu hooks, false otherwise.
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
@@ -568,6 +568,26 @@ extern "C"
    */
   PVR_ERROR IsRecordable(const EPG_TAG &tag, bool *isRecordable);
 
+  /*
+   * Check if the given EPG tag can be played.
+   * @param tag The EPG tag
+   * @return True if the EPG tag can be played
+   * @remarks Required, always return false if not supported by the addon
+   *
+   */
+  bool IsPlayable(const EPG_TAG &tag);
+
+  /*
+   * Get the URL to play a given EPG tag
+   * @param tag The EPG tag
+   * @param url The url to be returnd
+   * @param urlLen The length of the url buffer
+   * @return The length of the url. -1 if the tag is not playable
+   * @remarks Required, always return -1 if not supported by the addon
+   *
+   */
+  int GetEpgTagUrl(const EPG_TAG &tag, char *url, int urlLen);
+
   /*!
    * @brief Notify the pvr addon that XBMC (un)paused the currently playing stream
    */
@@ -723,6 +743,9 @@ extern "C"
     pClient->toAddon.PositionRecordedStream         = PositionRecordedStream;
     pClient->toAddon.LengthRecordedStream           = LengthRecordedStream;
     pClient->toAddon.IsRecordable                   = IsRecordable;
+    pClient->toAddon.IsPlayable                     = IsPlayable;
+    pClient->toAddon.GetEpgTagUrl                   = GetEpgTagUrl;
+
 
     pClient->toAddon.DemuxReset                     = DemuxReset;
     pClient->toAddon.DemuxAbort                     = DemuxAbort;
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
@@ -558,6 +558,16 @@ extern "C"
    */
   bool CanSeekStream();
 
+  /*
+   * Check if the given EPG tag can be recorded.
+   * @param tag the epg tag
+   * @param isRecordable Set to true if the tag can be recorded
+   * @return PVR_ERROR_NO_ERROR if isRecordable was set by the addon
+   * @remarks Optional, return PVR_ERROR_NOT_IMPLEMENTED to let kodi decide
+   *
+   */
+  PVR_ERROR IsRecordable(const EPG_TAG &tag, bool *isRecordable);
+
   /*!
    * @brief Notify the pvr addon that XBMC (un)paused the currently playing stream
    */
@@ -712,6 +722,7 @@ extern "C"
     pClient->toAddon.SeekRecordedStream             = SeekRecordedStream;
     pClient->toAddon.PositionRecordedStream         = PositionRecordedStream;
     pClient->toAddon.LengthRecordedStream           = LengthRecordedStream;
+    pClient->toAddon.IsRecordable                   = IsRecordable;
 
     pClient->toAddon.DemuxReset                     = DemuxReset;
     pClient->toAddon.DemuxAbort                     = DemuxAbort;
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
@@ -582,11 +582,12 @@ extern "C"
    * @param tag The EPG tag
    * @param url The url to be returnd
    * @param urlLen The length of the url buffer
+   * @param properties The properties to be provided to the file item
    * @return The length of the url. -1 if the tag is not playable
    * @remarks Required, always return -1 if not supported by the addon
    *
    */
-  int GetEpgTagUrl(const EPG_TAG &tag, char *url, int urlLen);
+  int GetEpgTagUrl(const EPG_TAG &tag, char *url, int urlLen, const CStringPropertyMapPtr&);
 
   /*!
    * @brief Notify the pvr addon that XBMC (un)paused the currently playing stream
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h
@@ -663,6 +663,8 @@ extern "C" {
     long long (__cdecl* PositionRecordedStream)(void);
     long long (__cdecl* LengthRecordedStream)(void);
     PVR_ERROR (__cdecl* IsRecordable)(const EPG_TAG&, bool*);
+    bool (__cdecl* IsPlayable)(const EPG_TAG&);
+    int (__cdecl* GetEpgTagUrl)(const EPG_TAG&, char*, int);
     void (__cdecl* DemuxReset)(void);
     void (__cdecl* DemuxAbort)(void);
     void (__cdecl* DemuxFlush)(void);
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h
@@ -662,6 +662,7 @@ extern "C" {
     long long (__cdecl* SeekRecordedStream)(long long, int);
     long long (__cdecl* PositionRecordedStream)(void);
     long long (__cdecl* LengthRecordedStream)(void);
+    PVR_ERROR (__cdecl* IsRecordable)(const EPG_TAG&, bool*);
     void (__cdecl* DemuxReset)(void);
     void (__cdecl* DemuxAbort)(void);
     void (__cdecl* DemuxFlush)(void);
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h
@@ -31,6 +31,8 @@
 #include <string.h>
 #include <stdint.h>
 #include <stdio.h>
+#include <map>
+#include <memory>
 
 #include "xbmc_addon_types.h"
 #include "xbmc_epg_types.h"
@@ -102,6 +104,10 @@ extern "C" {
     xbmc_codec_type_t codec_type;
     xbmc_codec_id_t   codec_id;
   } xbmc_codec_t;
+
+  typedef std::map<std::string, std::string> CStringPropertyMap;
+  typedef std::shared_ptr<CStringPropertyMap> CStringPropertyMapPtr;
+
 
   /*!
    * @brief numeric PVR timer type definitions (PVR_TIMER.iTimerType values)
@@ -664,7 +670,7 @@ extern "C" {
     long long (__cdecl* LengthRecordedStream)(void);
     PVR_ERROR (__cdecl* IsRecordable)(const EPG_TAG&, bool*);
     bool (__cdecl* IsPlayable)(const EPG_TAG&);
-    int (__cdecl* GetEpgTagUrl)(const EPG_TAG&, char*, int);
+    int (__cdecl* GetEpgTagUrl)(const EPG_TAG&, char*, int, const CStringPropertyMapPtr&);
     void (__cdecl* DemuxReset)(void);
     void (__cdecl* DemuxAbort)(void);
     void (__cdecl* DemuxFlush)(void);
--- a/xbmc/pvr/PVRGUIActions.cpp
+++ b/xbmc/pvr/PVRGUIActions.cpp
@@ -1092,13 +1092,17 @@ namespace PVR
       if (!epgTag)
         return false;
 
-      std::string stream = epgTag->GetStreamUrl();
+      CStringPropertyMapPtr properties(new CStringPropertyMap());
+      std::string stream = epgTag->GetStreamUrl(properties);
       if (stream.empty())
       {
         return false;
       }
 
       item->SetPath(stream);
+      for (auto const &entry : *properties.get()) {
+        item->SetProperty(entry.first, entry.second);
+      }
 
       if (!bCheckResume || CheckResumeRecording(item))
         CApplicationMessenger::GetInstance().PostMsg(TMSG_MEDIA_PLAY, 0, 0, static_cast<void*>(new CFileItem(*item)));
--- a/xbmc/pvr/PVRGUIActions.cpp
+++ b/xbmc/pvr/PVRGUIActions.cpp
@@ -1086,6 +1086,26 @@ namespace PVR
     return true;
   }
 
+  bool CPVRGUIActions::PlayEpgTag(const CFileItemPtr &item, bool bPlayMinimized, bool bCheckResume) const
+  {
+      const CPVREpgInfoTagPtr epgTag(CPVRItem(item).GetEpgInfoTag());
+      if (!epgTag)
+        return false;
+
+      std::string stream = epgTag->GetStreamUrl();
+      if (stream.empty())
+      {
+        return false;
+      }
+
+      item->SetPath(stream);
+
+      if (!bCheckResume || CheckResumeRecording(item))
+        CApplicationMessenger::GetInstance().PostMsg(TMSG_MEDIA_PLAY, 0, 0, static_cast<void*>(new CFileItem(*item)));
+
+      return true;
+    }
+
   bool CPVRGUIActions::SwitchToChannel(const CFileItemPtr &item, bool bCheckResume) const
   {
     return SwitchToChannel(item, bCheckResume, m_settings.GetBoolValue(CSettings::SETTING_PVRPLAYBACK_SWITCHTOFULLSCREEN));
--- a/xbmc/pvr/PVRGUIActions.h
+++ b/xbmc/pvr/PVRGUIActions.h
@@ -245,6 +245,15 @@ namespace PVR
     bool PlayRecording(const CFileItemPtr &item, bool bCheckResume) const;
 
     /*!
+     * @brief Play epg tag.
+     * @param item containing a an epg tag.
+     * @param bPlayMinimized controls whether the recording should be played fullscreen or in a window.
+     * @param bCheckResume controls resume check.
+     * @return true on success, false otherwise.
+     */
+    bool PlayEpgTag(const CFileItemPtr &item, bool bPlayMinimized, bool bCheckResume) const;
+
+    /*!
      * @brief Switch channel.
      * @param item containing a channel or an epg tag.
      * @param bCheckResume controls resume check in case a recording for the current epg event is present.
--- a/xbmc/pvr/addons/PVRClients.cpp
+++ b/xbmc/pvr/addons/PVRClients.cpp
@@ -28,6 +28,7 @@
 #include "ServiceBroker.h"
 #include "cores/IPlayer.h"
 #include "guilib/LocalizeStrings.h"
+#include "pvr/epg/EpgInfoTag.h"
 #include "messaging/ApplicationMessenger.h"
 #include "pvr/channels/PVRChannelGroupInternal.h"
 #include "pvr/channels/PVRChannelGroups.h"
@@ -1213,6 +1214,21 @@ time_t CPVRClients::GetPlayingTime() const
   }
 
   return time;
+}
+
+bool CPVRClients::IsRecordable(const CConstPVREpgInfoTagPtr &tag) const
+{
+  PVR_CLIENT client;
+  bool isRecordable = false;
+
+  if (GetClient(tag->ChannelTag()->ClientID(), client))
+  {
+     if(client->IsRecordable(tag, &isRecordable) != PVR_ERROR_NO_ERROR) {
+       isRecordable = tag->EndAsLocalTime() > CDateTime::GetCurrentDateTime();
+     }
+  }
+
+  return isRecordable;
 }
 
 bool CPVRClients::IsTimeshifting(void) const
--- a/xbmc/pvr/addons/PVRClients.cpp
+++ b/xbmc/pvr/addons/PVRClients.cpp
@@ -1231,6 +1231,30 @@ bool CPVRClients::IsRecordable(const CConstPVREpgInfoTagPtr &tag) const
   return isRecordable;
 }
 
+bool CPVRClients::IsPlayable(const CConstPVREpgInfoTagPtr &tag)
+{
+  PVR_CLIENT client;
+
+  if (GetClient(tag->ChannelTag()->ClientID(), client))
+  {
+    return client->IsPlayable(tag);
+  }
+
+  return NULL;
+}
+
+const std::string CPVRClients::GetEpgTagUrl(const CConstPVREpgInfoTagPtr &tag)
+{
+  PVR_CLIENT client;
+
+  if (GetClient(tag->ChannelTag()->ClientID(), client))
+  {
+    return client->GetEpgTagUrl(tag);
+  }
+
+  return "";
+}
+
 bool CPVRClients::IsTimeshifting(void) const
 {
   PVR_CLIENT client;
--- a/xbmc/pvr/addons/PVRClients.cpp
+++ b/xbmc/pvr/addons/PVRClients.cpp
@@ -1243,13 +1243,13 @@ bool CPVRClients::IsPlayable(const CConstPVREpgInfoTagPtr &tag)
   return NULL;
 }
 
-const std::string CPVRClients::GetEpgTagUrl(const CConstPVREpgInfoTagPtr &tag)
+const std::string CPVRClients::GetEpgTagUrl(const CConstPVREpgInfoTagPtr &tag, const CStringPropertyMapPtr &properties)
 {
   PVR_CLIENT client;
 
   if (GetClient(tag->ChannelTag()->ClientID(), client))
   {
-    return client->GetEpgTagUrl(tag);
+    return client->GetEpgTagUrl(tag, properties);
   }
 
   return "";
--- a/xbmc/pvr/addons/PVRClients.h
+++ b/xbmc/pvr/addons/PVRClients.h
@@ -626,6 +626,8 @@ namespace PVR
 
     int GetClientId(const std::string& strId) const;
 
+    bool IsRecordable(const CConstPVREpgInfoTagPtr &tag) const;
+
     bool IsRealTimeStream() const;
 
     void ConnectionStateChange(CPVRClient *client, std::string &strConnectionString, PVR_CONNECTION_STATE newState,
--- a/xbmc/pvr/addons/PVRClients.h
+++ b/xbmc/pvr/addons/PVRClients.h
@@ -630,7 +630,7 @@ namespace PVR
 
     bool IsPlayable(const CConstPVREpgInfoTagPtr &tag);
 
-    const std::string GetEpgTagUrl(const CConstPVREpgInfoTagPtr &tag);
+    const std::string GetEpgTagUrl(const CConstPVREpgInfoTagPtr &tag, const CStringPropertyMapPtr &properties);
 
     bool IsRealTimeStream() const;
 
--- a/xbmc/pvr/addons/PVRClients.h
+++ b/xbmc/pvr/addons/PVRClients.h
@@ -628,6 +628,10 @@ namespace PVR
 
     bool IsRecordable(const CConstPVREpgInfoTagPtr &tag) const;
 
+    bool IsPlayable(const CConstPVREpgInfoTagPtr &tag);
+
+    const std::string GetEpgTagUrl(const CConstPVREpgInfoTagPtr &tag);
+
     bool IsRealTimeStream() const;
 
     void ConnectionStateChange(CPVRClient *client, std::string &strConnectionString, PVR_CONNECTION_STATE newState,
--- a/xbmc/pvr/dialogs/GUIDialogPVRGuideInfo.cpp
+++ b/xbmc/pvr/dialogs/GUIDialogPVRGuideInfo.cpp
@@ -242,7 +242,7 @@ void CGUIDialogPVRGuideInfo::OnInitWindow()
       bHideRecord = false;
     }
   }
-  else if (CServiceBroker::GetPVRManager().Clients()->SupportsTimers() && m_progItem->EndAsLocalTime() > CDateTime::GetCurrentDateTime())
+  else if (CServiceBroker::GetPVRManager().Clients()->SupportsTimers() && m_progItem->IsRecordable())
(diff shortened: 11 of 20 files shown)


xbmc/addons/PVRClient.cpp
```diff
@@ -730,6 +730,35 @@ PVR_ERROR CPVRClient::RenameChannel(const CPVRChannelPtr &channel)
   return retVal;
 }
 
+PVR_ERROR CPVRClient::IsRecordable(const EPG::CEpgInfoTag *tag, bool *isRecordable) {
```
AlwinEsch: Can you change this to:
```cpp
PVR_ERROR CPVRClient::IsRecordable(const EPG::CEpgInfoTag *tag, bool *isRecordable)
{
```

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
```diff
@@ -576,6 +576,16 @@ extern "C"
    */
   bool CanSeekStream();
 
+  /*
+   * Check if the given EPG tag can be recorded.
+   * @param tag the eog tag
+   * @param the returned value, true if the tag can be recorded
+   * @return true if the tag can be recored.
```
AlwinEsch: `true` about a `PVR_ERROR` does not match exactly.

xbmc/pvr/addons/PVRClients.cpp
```diff
@@ -1454,6 +1454,21 @@ time_t CPVRClients::GetPlayingTime() const
...
 
+bool CPVRClients::IsRecordable(const EPG::CEpgInfoTag *tag) const
+{
+  PVR_CLIENT client;
+  bool isRecordable = false;
+
+  if (GetClient(tag->ChannelTag()->ClientID(), client))
+  {
+     if(client->IsRecordable(tag, &isRecordable) != PVR_ERROR_NO_ERROR) {
+       isRecordable = tag->EndAsLocalTime() > CDateTime::GetCurrentDateTime();
```
AlwinEsch: If it returns `PVR_ERROR_NO_ERROR` is `isRecordable` always false.

AlwinEsch:
The idea to have something for this is where nice and the basic way looks good for me.

Maybe we can add a new Error like `PVR_ERROR_NOT_RECORDABLE`

```cpp
     if(client->IsRecordable(tag, &isRecordable) == PVR_ERROR_NOT_RECORDABLE)
       return false;
     isRecordable = tag->EndAsLocalTime() > CDateTime::GetCurrentDateTime();
```

AlwinEsch:
Also we can use maybe direct a `true/false` of the return on `IsRecordable`. So if a normal add-on support recordings it always returns true as default otherwise false and always the same as on `PVR_ADDON_CAPABILITIES`.

ksooo:
Thanks for the contribution, looking good in general.

Some comments from first quick review:
- I agree to @alwinus on the signature for the new api function, returning a bool is fine here, no error code needed
- A pvr add-on api bump is needed, but missing in the PR
- have you checked that the feature also works if you are for instance in the channels window or pvr guide window? Just asking because you only modified the guide dialog, but there are other places where a decision whether a broadcast can be recorded is done.
- your change can be earliest merged for Kodi v18, because for Krypton we're in besta now and now api changes are allowed at this stage

ksooo:
Let's start with getting the api defined before commenting on code  that might be obsoleted once we have a decision on that. ;-)

My take (after thinking about this some more minutes):

- The signature as it currently is, is fine
- Return ERROR_NONE: Kodi takes the bool value from the api function to decide on recordability
- Return any other error: Kodi decides on recordabiliy according to start and end time of the epg tag, like it does today
- no need for an additional error code, especially as "cannot be recorded" is not an error!
- if we would change return value to bool, Kodi no longer can decide whether to use start / end time to decide, it had to take the value as returned by the add-on. The latter implies that every (!) add-on must properly support the new api function. This is problematic as not all add-ons have maintainers any more and the proper implementation  for most if the add-ons is more complicated than just "return true/false;"

ksooo:
@AlwinEsch ^^^ is my proposal to let the new api function like it is designed now, okay?

rbuehlma:
Wow, thanks for the fast feedback. Its perfectly fine to have this PR only in v18.
I will work on feedback the next days.

Do you prefer incremental commits or shall I amend the current one?

> have you checked that the feature also works if you are for instance in the channels window or pvr guide window? Just asking because you only modified the guide dialog, but there are other places where a decision whether a broadcast can be recorded is done.

You are right, my first thought was that its not possible to select past events elsewhere, but this call should also be used in the other cases. I will pick this up.

xbmc/pvr/addons/PVRClients.cpp
rbuehlma: isRecordable is passed as "return"-parameter to IsRecordable and set there in case of PVR_ERROR_NO_ERROR (the PVR-addon does the decision). In all other cases, the previous behaviour of Kodi is used as this is probably what most PVR addons require.

rbuehlma:
BTW: I did write down some ideas of what I would like to implement for IPTV in the board:
http://forum.kodi.tv/showthread.php?tid=287864
This PR is one of them.
My next attempt would be to also allow "Switch" to an EPG entry (e.g. select a past (or even future) show and play it instead of recording it).
May be you could comment on these ideas in the thread if you have some input.

FernetMenta:
Why do you want to overcomplicate things? Just drop the entire check and have the timer creation fail if recording is not supported on a show.

rbuehlma:
> Why do you want to overcomplicate things? Just drop the entire check and have the timer creation fail if recording is not supported on a show.

The check is also used to decide whether the "Record" button is shown. In my opinion, we could drop the check in PVRTimerInfoTag.cpp, but the remaining of the code would still be required for the button.

ksooo:
@FernetMenta I don't want to handle all the bug reports saying "kodi recording does not work. There is a 'record' menu item and sometimes when I use it I get an error message. Please make that there is no 'record' item if the event can't be recorded." ;-)

xbmc/pvr/addons/PVRClients.cpp
ksooo: > isRecordable is passed as "return"-parameter 
Yep. The code is fine as it is.

xbmc/addons/PVRClient.cpp
ksooo: Yep. Please change. Curly braces start on next line is Kodi coding convention.

xbmc/addons/PVRClient.cpp
```diff
@@ -730,6 +730,35 @@ PVR_ERROR CPVRClient::RenameChannel(const CPVRChannelPtr &channel)
...
+  PVR_ERROR retVal(PVR_ERROR_UNKNOWN);
+  try
+  {
+    EPG_TAG pvrTag;
+    time_t t;
+    tag->StartAsUTC().GetAsTime(t);
+    pvrTag.startTime = t;
+    tag->EndAsUTC().GetAsTime(t);
+    pvrTag.endTime = t;
+    pvrTag.iUniqueBroadcastId = tag->UniqueBroadcastID();
```
ksooo: Submitting a partially filled EPG_TAG "feels" not good (although working for your use case). I can imagine that addons could make their "isrecordable" decision based on other criteria contained in the EPG_TAG. Short, I suggest to fill the EPG_TAG completely here, like done here for recordings https://github.com/xbmc/xbmc/blob/master/xbmc/addons/PVRClient.cpp#L275

xbmc/addons/PVRClient.h
```diff
@@ -219,6 +219,13 @@ namespace PVR
      */
     PVR_ERROR RenameChannel(const CPVRChannelPtr &channel);
 
+    /*
+     * @brief Check if an epg tag can be recorded
+     * @param tag The epg tag
+     * @return true if the tag can be recorded
```
ksooo: doxy for param isrecordable is missing, please add

return doxy does not match actual return type. Please fix.

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
```diff
@@ -576,6 +576,16 @@ extern "C"
    */
   bool CanSeekStream();
 
+  /*
+   * Check if the given EPG tag can be recorded.
+   * @param tag the eog tag
```
ksooo: typo: eog vs. epg

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
```diff
@@ -576,6 +576,16 @@ extern "C"
...
+   * @param the returned value, true if the tag can be recorded
+   * @return true if the tag can be recored.
+   * @remarks Optional, return PVR_ERROR_NOT_IMPLEMENTED to let kodi decide base on the current time
```
ksooo: Please remove "base on the current time", because from the api perspective of view oit does not matter what kodi makes out of this.

xbmc/pvr/addons/PVRClients.h
```diff
@@ -28,6 +28,8 @@
 
 #include "addons/PVRClient.h"
 
+#include "epg/EpgInfoTag.h"
```
ksooo: no need for an include here. use forward declaration

xbmc/pvr/addons/PVRClients.h
```diff
@@ -657,6 +659,8 @@ namespace PVR
 
     int GetClientId(const std::string& strId) const;
 
+    bool IsRecordable(const EPG::CEpgInfoTag *tag) const;
```
ksooo: Avoid raw pointers, use shared pointer instead => <code>const EPG::CEpgInfoTagPtr &tag</code>, yes this can be forward declared as well. still no need for the include above.

xbmc/pvr/addons/PVRClients.h
rbuehlma: If I do a forward declaration, I get this error:
`PVRClients.cpp: In member function ‘bool PVR::CPVRClients::IsRecordable(const EPG::CEpgInfoTag&) const’:
PVRClients.cpp:1462:20: error: invalid use of incomplete type ‘const class EPG::CEpgInfoTag’
   if (GetClient(tag.ChannelTag()->ClientID(), client))
                    ^
In file included from PVRClients.h:26:0,
                 from PVRClients.cpp:21:
PVRChannel.h:39:9: note: forward declaration of ‘class EPG::CEpgInfoTag’
   class CEpgInfoTag;
         ^
`
Am I doing something wrong here?

ksooo: Of course you need the include in the cpp file.

xbmc/addons/PVRClient.h
```diff
@@ -707,6 +708,13 @@ namespace PVR
     static void WriteClientChannelInfo(const CPVRChannelPtr &xbmcChannel, PVR_CHANNEL &addonChannel);
 
     /*!
+     * @brief Copy over epg info from CEpgInfoTag to EPG_TAG.
+     * @param tag The epg tag on XBMC's side.
```
ksooo: XBMC -> Kodi, please.

rbuehlma:
If I did not miss one, I have implemented all your feedback.

I hope I did also find all the other checks which should be replaced with this new function.

xbmc/addons/PVRClient.h
```diff
@@ -707,6 +708,13 @@ namespace PVR
...
+     * @param addonChannel The epg tag on the addon's side.
```
ksooo: There is no param addonChannel. please fix.

And please, don't mis(use) me as your doxy / typo checker. That's no fun. Do this by yourself, please, before committing. Thanks.

xbmc/epg/EpgInfoTag.cpp
```diff
@@ -240,7 +240,7 @@ bool CEpgInfoTag::IsActive(void) const
 
 bool CEpgInfoTag::IsRecordable(void) const
 {
-  return g_PVRClients->IsRecordable(this);
+  return g_PVRClients->IsRecordable(shared_from_this());
```
ksooo: Ha, I was curios whether you would do this right. Check. Nice.

xbmc/addons/PVRClient.cpp
```diff
@@ -364,6 +364,36 @@ void CPVRClient::WriteClientChannelInfo(const CPVRChannelPtr &xbmcChannel, PVR_C
...
+  time_t t;
+  tag->StartAsUTC().GetAsTime(t);
+  pvrTag.startTime = t;
+  tag->EndAsUTC().GetAsTime(t);
+  pvrTag.endTime = t;
+  pvrTag.iParentalRating = tag->ParentalRating();
+  pvrTag.iUniqueBroadcastId = tag->UniqueBroadcastID();
+  pvrTag.bNotify = tag->Notify();
+  tag->FirstAiredAsUTC().GetAsTime(t);
+  pvrTag.firstAired = t;
```
ksooo: this looks wrong. you have to take the value from tag.

xbmc/addons/PVRClient.h
rbuehlma: I'm sorry, will be more carefully in future commits

xbmc/pvr/addons/PVRClients.h
```diff
@@ -28,8 +28,6 @@
 
 #include "addons/PVRClient.h"
 
-#include "epg/EpgInfoTag.h"
```
ksooo: good catch ;-)

xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
```diff
@@ -759,8 +759,11 @@ void CGUIDialogPVRTimerSettings::InitializeTypesList()
     if (type->ForbidsEpgTagOnCreate() && m_timerInfoTag->GetEpgInfoTag())
       continue;
 
-    // Drop TimerTypes that aren't rules if end time is in the past
-    if (!type->IsTimerRule() && m_timerInfoTag->EndAsLocalTime() < CDateTime::GetCurrentDateTime())
+    EPG::CEpgInfoTagPtr epgTag = m_timerInfoTag->GetEpgInfoTag(false);
+    //If we have an epg tag, ask the pvr addon if recording is possible, else, use the current time
+    bool canRecord = epgTag ? epgTag->IsRecordable() : m_timerInfoTag->EndAsLocalTime() >= CDateTime::GetCurrentDateTime();
+    // Drop TimerTypes that aren't rules or cannot be recorded
+    if (!type->IsTimerRule() && !canRecord)
```
ksooo: For efficiency reasons I would first check IsTimerRule, which is a cheap (local) call and only if it's true also check IsRecordable, which is expensive, because a (potential) remote call.

xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
```diff
@@ -759,8 +759,11 @@ void CGUIDialogPVRTimerSettings::InitializeTypesList()
     if (type->ForbidsEpgTagOnCreate() && m_timerInfoTag->GetEpgInfoTag())
       continue;
 
-    // Drop TimerTypes that aren't rules if end time is in the past
-    if (!type->IsTimerRule() && m_timerInfoTag->EndAsLocalTime() < CDateTime::GetCurrentDateTime())
+    EPG::CEpgInfoTagPtr epgTag = m_timerInfoTag->GetEpgInfoTag(false);
```
ksooo: const EPG::CEpgInfoTagPtr epgTag(m_timerInfoTag->GetEpgInfoTag(false)); if you don't mind

xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
```diff
@@ -759,12 +759,15 @@ void CGUIDialogPVRTimerSettings::InitializeTypesList()
...
-    //If we have an epg tag, ask the pvr addon if recording is possible, else, use the current time
-    bool canRecord = epgTag ? epgTag->IsRecordable() : m_timerInfoTag->EndAsLocalTime() >= CDateTime::GetCurrentDateTime();
-    // Drop TimerTypes that aren't rules or cannot be recorded
-    if (!type->IsTimerRule() && !canRecord)
-      continue;
+    // Drop TimerTypes that aren't rules and cannot be recorded
+    if (!type->IsTimerRule())
+    {
+      EPG::CEpgInfoTagPtr epgTag = m_timerInfoTag->GetEpgInfoTag(false);
+      //If we have an epg tag, ask the pvr addon if recording is possible, else, use the current time
```
ksooo: Now we're getting to the nitpicking part: always put a space between // and the actual comment. Kodi coding convention. Thanks.

ksooo:
@rbuehlma thanks for changing everything in almost lightning speed. Could you please squash everything into a single commit now that basic review is done (at least from my side, you never know who else suddenly will jump onto the train).

rbuehlma:
I am the one to thank you all for the fast review. Its now squashed. Whats the process now?

ksooo:
> Whats the process now?

 Well, we have to wait until master gets opened for v18 development, which is several months away from tody - we are in early beta for v17 right now. Only fixes currently go into master.

 Maybe we need to discuss changes intended by this PR further if another team/community member comments here.

 The only thing that is currently missing in this PR is actually bumping the PVR addon API in Kodi. You can refer to previous PRs with the flag "PVR API Change" to learn how this is to be done. And of course all pvr addons in the official pvr-addons repo must be adapted with a (at least default) implementation of the new function, changelog and the repective api bump.  As you are the initiator of this api change I see you in charge for doing the addon adjustments, too. ;-) But i will be happy to assist you on this if something is unclear.

I suggest not yet to do the API bumps, as we are always trying to combine multiple changes into a single bump.

rbuehlma:
> And of course all pvr addons in the official pvr-addons repo must be adapted with a (at least default) implementation of the new function, changelog and the repective api bump.

Well, I will wait with this until my second API change "IsPlayable" is done. There I probably will re-use a function introduced in this change (WriteEpgTag). How should I do that? I personally would create a branch from this one to prevent a merge conflict, but then the new PR will contain both changes as this branch does not exists in your repo. Is there a better way?

ksooo:
> Is there a better way?

 I would not open another PR, but extend the existing PR, keeping "IsRecordable" and "IsPlayable" changes in separate commits.

Jalle19:
I was about to suggest tacking on IsPlayable too. It would be a first step in being able to grey out channels that can't be played at the moment. We should think about the API design though since it would make most sense for channels, not just individual EPG tags. Preferably there would also be a bulk method, e.g. GetPlayableChannels, based on which Kodi can grey out channels. Otherwise there would be a lot of back and forth communication with the backend, especially with hundreds of channels.

rbuehlma:
@Jalle19 Shall we take this discussion to the board? I expect that it does not belong in a PR but aligning is definitely a good idea.
We could use this thread: http://forum.kodi.tv/showthread.php?tid=287864

rbuehlma:
I have added the IsPlayable flag currently only for EPG tags. If this returns true, an additional button "RePlay" will be show, which will cause the stream url to be requested from the addon (using GetEpgTagUrl).

With this PR, replay of old EPG tags does work. To get the correct EPG shown, I had to do some hacks which will not be pushed as the refactoring of the PVR interface will probably make them obsolete.

@Jalle19 from my point of view, is your usecase of IsPlayable a different one and also requires a different functions. If you suggest the interface, I can try to implement that as well.

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -10056,7 +10056,9 @@ msgctxt "#19239"
 msgid "Starting background threads"
 msgstr ""
 
-#empty string with id 19240
+msgctxt "#19240"
+msgid "RePlay"
```
ksooo: I don't think the capital P is okay here. Even this might be a proper nound for Zattoo, Kodi UI only contains 'generic' strings.

ksooo: When adding strings, plase always provide subscriptive text (#.) and info in which file(s) the new string is used (#:)

addons/skin.estouchy/xml/DialogPVRInfo.xml
```diff
@@ -198,6 +198,13 @@
 				<label>19165</label>
 				<visible>Window.IsActive(PVRGuideInfo)</visible>
 			</control>
+			<control type="button" id="9">
+				<description>RePlay Program</description>
```
ksooo: Again, no captial P in Replay. 
Additionally, not "Program", but "program" (Kodi UI design guideline).
I'm not sure whether 'program' is the right word here. We use "show" elsewhere.

ksooo: yeah yeah I know it's just a comment, but still...

xbmc/FileItem.cpp
```diff
@@ -124,6 +124,10 @@ CFileItem::CFileItem(const CEpgInfoTagPtr& tag)
   Initialize();
 
   m_bIsFolder = false;
+  if (tag->HasPVRChannel())
+  {
+    m_pvrChannelInfoTag = tag->ChannelTag();
```
ksooo: I'm afraid that this may cause serious strange side affects, because usually a file item contains at most one of the pvr types! Using code assumes that there are no "multitype" file items.

xbmc/addons/PVRClient.cpp
```diff
@@ -393,6 +393,7 @@ void CPVRClient::WriteEpgTag(const EPG::CConstEpgInfoTagPtr &tag, EPG_TAG &pvrTa
   pvrTag.strIMDBNumber = tag->IMDBNumber().c_str();
   pvrTag.strEpisodeName = tag->EpisodeName().c_str();
   pvrTag.strIconPath = tag->Icon().c_str();
+  pvrTag.iChannelNumber = tag->PVRChannelNumber();
```
ksooo: seems unrelated to this PR. why have you added this?

xbmc/addons/PVRClient.cpp
```diff
@@ -785,6 +786,46 @@ PVR_ERROR CPVRClient::IsRecordable(const EPG::CConstEpgInfoTagPtr &tag, bool *is
...
+  }
+  catch (std::exception &e)
+  {
+    LogException(e, __FUNCTION__);
+  }
+
+  return false;
+}
+
+const std::string CPVRClient::GetEpgTagUrl(const EPG::CConstEpgInfoTagPtr &tag)
```
ksooo: Usually we use all uppercase letters for abbaviations. => GetEPGTagURL

xbmc/addons/PVRClient.cpp
```diff
@@ -785,6 +786,46 @@ PVR_ERROR CPVRClient::IsRecordable(const EPG::CConstEpgInfoTagPtr &tag, bool *is
...
+{
+  if (!m_bReadyToUse)
+    return NULL;
```
ksooo: We're doing c++, thus "return nullptr;" No need to change all the old code still using NULL, but for new code we should use "modern" c++ features.

xbmc/addons/PVRClient.cpp
```diff
@@ -785,6 +786,46 @@ PVR_ERROR CPVRClient::IsRecordable(const EPG::CConstEpgInfoTagPtr &tag, bool *is
...
+
+  try
+  {
+    EPG_TAG pvrTag;
+    WriteEpgTag(tag, pvrTag);
+    char url[4096];
```
ksooo: where does this magic number come from???

xbmc/pvr/dialogs/GUIDialogPVRGuideInfo.cpp
```diff
@@ -127,14 +128,18 @@ bool CGUIDialogPVRGuideInfo::OnClickButtonPlay(CGUIMessage &message)
...
   {
     Close();
 
     if (m_progItem)
     {
       if (message.GetSenderId() == CONTROL_BTN_PLAY_RECORDING && m_progItem->HasRecording())
         g_application.PlayFile(CFileItem(m_progItem->Recording()), "videoplayer");
+      if (message.GetSenderId() == CONTROL_BTN_PLAY_TAG && m_progItem->IsPlayable()) {
+        m_progItem->SetPath(m_progItem->GetStreamUrl());
+        g_application.PlayFile(CFileItem(m_progItem), "videoplayer");
```
ksooo: I'm not an expert on this, but your concept to rely on a stream URL that can be played is imo not good. Not sure, though. @FernetMenta might comment on this?

ksooo: Why is this functionality only available from the guide dialog, not from the guide window?

xbmc/FileItem.cpp
rbuehlma: I wasn't sure about this. Who could guide here how to solve this correctly? If the channel is not set, lots of checks need to be changed to ensure that the channel information is still printed (we are actually playing a channel here, just not live).

xbmc/addons/PVRClient.cpp
rbuehlma: Ii is related to his PR but the functions was introduced in the first commit. iChannelNumber was missing in the mapping.

xbmc/addons/PVRClient.cpp
rbuehlma: Its the maximum URL size we might expect. I don't think Kodi handles longer URLs, neither do most browsers. Shall I extract it to a constant or is there already one?

xbmc/pvr/dialogs/GUIDialogPVRGuideInfo.cpp
rbuehlma: Its currently only available in the dialog as I expect that this part needs to be adapter as soon as the new PVR/Stream API is inplace.

xbmc/FileItem.cpp
ksooo: > If the channel is not set, lots of checks need to be changed

 can you point me to this checks, please?

> we are actually playing a channel here

then, a fileitem containing only a pvrchannel and no epg tag should be used (whatever used means, thats why I'm asking for the code places checks would need to be adapted)?

Maybe, instead of <code>g_application.PlayFile(CFileItem(m_progItem), "videoplayer");</code>, <code>g_application.PlayFile(CFileItem(m_progItem->ChannelTag()), "videoplayer");</code> is correct?

rbuehlma: I first tried to use a file item containing an epg tag, but there most of the code to properly display the replay is missing.

Thinking about it, I agree that the fileitem containg a pvrchannel should be used, but then we would need to tell the player the offset of the show such that it knows which EPG tag to display (instead of now), just like timeshift. I'm not sure if there already is a way to do this?
There comes an additional challenge: I'm not using the PVR addon as input stream but an inputstream addon (mpd) and as war as I can see, there is no way to let the player know about this offset?

rbuehlma:
I did a rebase to the current master.

rbuehlma:
Instead of an EPG-Tag FileItem, I now create a Channel FileItem but override the URL with the one of the EPG-Tag. If the PVR addon returns the correct GetPlayingTime, this works well.

There arise some problems, if the URL is handled by an inputstream addon instead of the PVR addon, but that case does not seem to be supported currently anyway.

Who can tell me, if this PR will be accepted once the cycle of v18 has started?
And beside of bumping the api version and the depending addons, what else will be required from my side?

metaron-uk:
@rbuehlma this 'IsPlayable' feature interests me. 
Is the plan that when selecting 'RePlay' Kodi queries the Addon for a URL and then plays that URL directly, or that it then plays the stream via the PVR backend?
If kodi plays the URL directly this might allow some sort of tie-in between a PVR backend provided EPG and for example iplayer-www.

metaron-uk:
Reading the code in more detail I see this is exactly what it does :+1: 

Would video player also follow a URL in the form pvr://recordings/..... ?
If yes, it might allow the mechanism for playing recorded files via the guide to be improved (the current mechanism for associating a recording to just one EPG entry is a little inflexible given the number of repeats we get on TV these days)
This looks like it might allow a backend just the right amount of control...

rbuehlma:
I'm not sure if the current implementation is the correct way of passing the URL. 
Also I think "pvr://" URLs will be dropped in favour of input stream add-ons. If your stream/recordings are not supported by any of the input streams available, you would then need to implement your own input stream.

@FernetMenta can probably give you a more accurate answer here.

rbuehlma:
Rebased to current master.

FernetMenta:
@rbuehlma right. VIdeoPlayer will get rid of all dependencies to pvr. All information about start/end times etc will come from VP, not any other component like EPG. A pvr addon can set properties on a file item to select a particular imputstream handler.

rbuehlma:
How will it be possible to handle cases where this information start/end time & content information changes during the stream (e.g. when playing a live stream and the show changes)?

FernetMenta:
The source (inputstream/backend) is always aware of this information. No need to pull this info from EPG that originally came from the backend anyway.

rbuehlma:
Yes, the pvr backend is aware of this, but not the inputstream (inputstream.adaptive) in my and probably others cases. But as far as I understood, there is no interaction any more with the pvr addon as soon as the stream is playing.

Am I missing something?

FernetMenta:
How it is currently handled:
Information about the currently playing item bypasses VP and hits the UI directly. This does not make much sense and also has flaws that would prevent it from working in a headless scenario. btw, pvr is not the only case, Shoutcast input stream for example sends messages to the UI to trigger updates on current file item.

Future:
This info won't bypass VP. There may be cases where the fileitem requires some means to get updated. This has to be designed properly.

metaron-uk:
@FernetMenta I'm confused. You seem to be suggesting that the PVR API will handle only content management including EPG. In order to perform playback, PVR will need to hand off to a separate inputsteam addon.
This makes sense from a video player perspective (no need to depend on the PVR API) but PVR client maintainers will then need to implement and maintain a separate inputstream addon in order convert whatever protocol the backend speaks for video playback into something VP understands.
I can see this approach has merit, but it will involve quite a lot of work for PVR maintainers to re-structure their code. Is this on your roadmap for L***?
I agree wholeheartedly with you that the UI shouldn't be handling any of this by the way.
Probably this ought to be socialized with @ksooo etc... on the forum (if it hasn't been already that is!)

FernetMenta:
@metaron-uk partly correct. @AlwinEsch is currently working on a solution that binary addons can implement more than a single extension point. That means that a pvr addon is also the required inputstream addon. There is just a split of those two APIs.
Yes, addon maintainers need to adapt their addons but work stays in a reasonable amount of effort.

metaron-uk:
@FernetMenta - sounds like a plan :)

FernetMenta:
@metaron-uk btw, with that split and having the possibility to instantiate several addons at the same time, we can enable thumbnail extraction for pvr recordings. Further picture-in-picture is possible too.
Still some way to go ...

gena01:
I am interested in this as well. My use case might be a bit different. IPTV provider has most of the content recorded (with time difference, which can be changed in options). Ideally I would love to be able to pick any program (even if the time has been passed) which has been recorded and watch it. I would also love to be able to watch the program from the beginning and/or "live". (Lets say that it started 10 minutes ago.. i could watch from start of the program stream or at current position). The way the IPTV API works is via SOAP and i can fetch the stream URL right before playback based on user credentials (and possibly other things like video stream type/etc..)

rbuehlma:
@gena01, from my point of view, your use case should already be covered with this PR.

Though one part is missing: Some properties need to be set on the file item. At least the ones required by the inputstream.addon. So beside of fetching the stream-url from the addon, we also need to receive the properties.

As the file item might change its information during the playback (e.g. if the current show changes) I suggest to create a function which regularly polls the PVR addon for the current information and updates the file item accordingly. This function would then probably require the current playing time as input value such that the PVR addon can determine the currently playing program.

@FernetMenta would such a polling against the PVR addon be the right way and should information which is displayed in the UI (program name, start/end time, plot and so on) also be set as a property in the file item?

FernetMenta:
@rbuehlma videoplayer already "polls" the source, because it reads the input stream. this layer can be used to transport additional information from the source to the player. 
the fileitem itself should not be changed. some old pvr hacks do so and that resulted in big issues. videoplayer can expose this information in a separate object for the UI to access.

rbuehlma:
Thanks @FernetMenta, in this case, I will only implement the possibility to set the properties of the file item through the pvr addon in this PR.

I plan to replace GetEpgTagUrl with GetEpgTagFileItem which directly returns the complete FileItem containing the URL and the properties. The FileItem is created in PVRClient with the information (url and properties) received from the addon. Does that sound ok to you?

FernetMenta:
> The FileItem is created in PVRClient

I won't do this. PVRClient is the low level interface to the addon and should not access classes of upper layers. Assume all addons did this, then a change to FIleItem could break all those addons.

rbuehlma:
Ok, would it be better to add a second call GetEpgTagProperties which is called after GetEpgTagUrl to get the required properties?
Another option would be to add a return-parameter to GetEpgTagUrl which contains the properties after calling it.

I do not really like the idea to call the addon twice for a set of information belonging together. But I also try to prevent such return-parameter.

I guess the return parameter is the one I would prefer here.

metaron-uk:
@rbuehlma, sorry, but I don't see why a PVR client needs to update any properties once it has passed the 'playback' URL to PVR core which I guess would pass it on to videoplayer (which then passes it on to an inputstream addon ?) saying 'play this' and playback commences.

From this point onwards inputstream/videoplayer should manage whatever they need regarding the in-progress file using their own APIs.
I see this feature as allowing a PVR client provided EPG entry to have a URL which can be played back directly by videoplayer, thereby bypassing the PVR backend:
Possible example URLs: 
* nfs://server/area/myfile.mpg
* cifs:\\\\windowsbox\myVideos\MyRecording.qt
* pvr://recordings/active/default/myrecording.mpg
OR if that is dropped in future something like:
* inputstream://mythtv-istream/11928_20150103190000.mpg
* iptv://my-longurl.xxx
* upnp://mymediabox/videos/myfile.mp4
* etc...

Surely if the metadata associated with the in progress show needs to change that's a matter for intputstream API, not the PVR client API. This seems to be what @FernetMenta is suggesting.

If the playback URL 'finishes' - because the connection is lost or the file at that location ends, then inpustream / videoplayer stop and pass control back to Kodi core. This then presents the EPG entry where you presumably selected 'play this' previously.

Your currently proposed code seems to support this just fine already. I don't think you need to change it to add any 'Properties'.

FernetMenta:
> I see this feature as allowing a PVR client provided EPG entry to have a URL which can be played back directly by videoplayer, thereby bypassing the PVR backend:
Possible example URLs:

If you refer to streamURLs you should read: http://forum.kodi.tv/showthread.php?tid=259830&pid=2245238#pid2245238

rbuehlma:
@metaron-uk 

There are properties on the file item which need to be set, such that e.g. inputstream.adaptive can play the stream (manifest_type). I would say that this property needs to come from the pvr addon.

The other "properties" I was talking about, you called them meta-data which I pref (e.g. name of show, EPG info...) should in my opinion not be part of this PR but should still come from the PVR addon as we do not want to implement DASH stream for every PVR provider to get its meta information.

rbuehlma:
I have added a commit to allow the pvr addon providing properties for the file item.
Is this the right way to do it?

With this patch, URLs provided by the addon through this new API-functions can be played by eg. inputstream.adaptive.

If the PR is ok, which steps are required from my side to get it merged?

ksooo:
> I have added a commit to allow the pvr addon providing properties for the file item.
Is this the right way to do it?

 A general problem with your API change is that PVR addon API is a C API, not C++. I know that we are already using references and other c++ language constructs in the API, but I would not introduce more C++ in new APIs. You are using references, std::map etc pp... which I consider not okay.

> If the PR is ok, which steps are required from my side to get it merged?

 TBH, I have no plan. Other people currently doing larger changes to the binary addon component of Kodi. So I don't know when and how API changes for Leia are good to go.

rhocheck:
1)
streamURL is deprecated
http://forum.kodi.tv/showthread.php?tid=259830

2)
changing path of fileitem is not allowed. This has led to huge problems in caching those items.

FernetMenta:
streamURL is deprecated
http://forum.kodi.tv/showthread.php?tid=259830

changing path of fileitem is not allowed. This has led to huge problems in caching those items.

xbmc/addons/PVRClient.h
```diff
@@ -681,6 +708,13 @@ namespace PVR
     static void WriteClientChannelInfo(const CPVRChannelPtr &xbmcChannel, PVR_CHANNEL &addonChannel);
 
     /*!
+     * @brief Copy over epg info from CEpgInfoTag to EPG_TAG.
+     * @param tag The epg tag on Kodi's side.
+     * @param pvrTag The epg tag on the addon's side.
+     */
+    static void WriteEpgTag(const EPG::CConstEpgInfoTagPtr &tag, EPG_TAG &pvrTag);
```
ksooo: if you use EPG_TAG.iUniqueBroadcastId in the API you do not need this (expensive) method.

xbmc/addons/PVRClient.h
```diff
@@ -217,6 +217,33 @@ namespace PVR
      */
     PVR_ERROR RenameChannel(const CPVRChannelPtr &channel);
 
+    /*
+     * @brief Check if an epg tag can be recorded
+     * @param tag The epg tag
+     * @param isRecordable Set to true if the tag can be recorded
+     * @return PVR_ERROR_NO_ERROR if isRecordable was set by the addon
+     */
+    PVR_ERROR IsRecordable(const EPG::CConstEpgInfoTagPtr &tag, bool *isRecordable);
```
ksooo: Ooops, just realized that we are in CPVRClient here. Sooo it's quite opposite. Use bool& instead of bool*, const ref tag is okay.

xbmc/addons/PVRClient.h
```diff
@@ -217,6 +217,33 @@ namespace PVR
...
+
+    /*
+     * Check if the given EPG tag can be played.
+     * @param tag The EPG tag
+     * @return True if the EPG tag can be played
+     * @remarks Required, always return false if not supported by the addon
+     *
+     */
+    bool IsPlayable(const EPG::CConstEpgInfoTagPtr &tag);
```
ksooo: Ooops, just realized that we are in CPVRClient here. const ref tag is okay, then. My other comments stay valid.

xbmc/addons/PVRClient.h
```diff
@@ -217,6 +217,33 @@ namespace PVR
...
+
+    /*
+     * Get the URL to play a given EPG tag
+     * @param tag The EPG tag
+     * @param properties The properties to be provided to the file item
+     * @return The url to play for the EPG tag, empty string if it cannot be played
+     * @remarks Required, always return empty string if not supported by the addon
+     *
+     */
+    const std::string GetEpgTagUrl(const EPG::CConstEpgInfoTagPtr &tag, const CStringPropertyMapPtr &properties);
```
ksooo: Ooops, just realized that we are in CPVRClient here. const refs are okay, then. My other comments stay valid.

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
```diff
@@ -576,6 +576,37 @@ extern "C"
...
+  PVR_ERROR IsRecordable(const EPG_TAG &tag, bool *isRecordable);
+
+  /*
+   * Check if the given EPG tag can be played.
+   * @param tag The EPG tag
+   * @return True if the EPG tag can be played
+   * @remarks Required, always return false if not supported by the addon
+   *
+   */
+  bool IsPlayable(const EPG_TAG &tag);
```
ksooo: Please see the comments I accidentally made for CPVRClient your changes and apply them here. ;-)

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
```diff
@@ -576,6 +576,37 @@ extern "C"
...
 
+  /*
+   * Check if the given EPG tag can be recorded.
+   * @param tag the epg tag
+   * @param isRecordable Set to true if the tag can be recorded
+   * @return PVR_ERROR_NO_ERROR if isRecordable was set by the addon
+   * @remarks Optional, return PVR_ERROR_NOT_IMPLEMENTED to let kodi decide
+   *
+   */
+  PVR_ERROR IsRecordable(const EPG_TAG &tag, bool *isRecordable);
```
ksooo: Okay.

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
```diff
@@ -576,6 +576,37 @@ extern "C"
...
+   * Get the URL to play a given EPG tag
+   * @param tag The EPG tag
+   * @param url The url to be returnd
+   * @param urlLen The length of the url buffer
+   * @param properties The properties to be provided to the file item
+   * @return The length of the url. -1 if the tag is not playable
+   * @remarks Required, always return -1 if not supported by the addon
+   *
+   */
+  int GetEpgTagUrl(const EPG_TAG &tag, char *url, int urlLen, const CStringPropertyMapPtr&);
```
ksooo: No C++ here for the map, use C.

Return PVR_ERROR

Purpose of this function unclear. Docs not sufficient, see my comment for CPVRClient. fileitem? wtf?

xbmc/epg/EpgInfoTag.h
```diff
@@ -85,6 +85,22 @@ namespace EPG
...
+    /* @brief Check if this event can be played
+     * @return True if it can be played.
+     */
+    bool IsPlayable(void) const;
+
+    /* @brief Get the stream url for this event
+     * @param properties The properties to be provided to the file item
+     * @return The stream url, empty string if playing is not possible
+     */
+    const std::string GetStreamUrl(CStringPropertyMapPtr &properties) const;
```
ksooo: const CStringPropertyMapPtr &?

ksooo: okay, obviously the properties are filled by this method. Please make the documentation more clear.

xbmc/pvr/PVRGUIActions.cpp
```diff
@@ -926,6 +926,30 @@ namespace PVR
...
+        return false;
+      }
+
+      item->SetPath(stream);
+      for (auto const &entry : *properties.get()) {
+        item->SetProperty(entry.first, entry.second);
+      }
+
+      if (!bCheckResume || CheckResumeRecording(item))
+        CApplicationMessenger::GetInstance().PostMsg(TMSG_MEDIA_PLAY, 0, 0, static_cast<void*>(new CFileItem(*item)));
```
ksooo: I guess this has to be tweeked a little, but sorry no time atm, will take a closer look later. Basic idea is to have exactly one method in CPVRGUIActions calling this. I just merged a larger refactoring PR to achieve this...

Quick hint: you are missing for instance parental check here, but adding it here is wrong, we have one core "check and play" method here.

xbmc/pvr/PVRGUIActions.h
```diff
@@ -233,6 +233,15 @@ namespace PVR
     bool PlayRecording(const CFileItemPtr &item, bool bCheckResume) const;
 
     /*!
+     * @brief Play epg tag.
+     * @param item containing a an epg tag.
+     * @param bPlayMinimized controls whether the recording should be played fullscreen or in a window.
+     * @param bCheckResume controls resume check.
+     * @return true on success, false otherwise.
+     */
+    bool PlayEpgTag(const CFileItemPtr &item, bool bPlayMinimized, bool bCheckResume) const;
```
ksooo: No need for playminimized param. If we do this right, this will be handled internally by CPVRGUIActions.

ksooo (changes requested):

xbmc/pvr/addons/PVRClients.h
```diff
@@ -660,6 +660,12 @@ namespace PVR
 
     int GetClientId(const std::string& strId) const;
 
+    bool IsRecordable(const EPG::CConstEpgInfoTagPtr &tag) const;
+
+    bool IsPlayable(const EPG::CConstEpgInfoTagPtr &tag);
+
+    const std::string GetEpgTagUrl(const EPG::CConstEpgInfoTagPtr &tag, const CStringPropertyMapPtr &properties);
+
```
ksooo: Please add documentation!

rbuehlma:
@FernetMenta 
You are right, changing the URL is not a good idea.
I could create a FileItem constructor for this case but I'm not sure if that would be the solution you are expecting.

I'm currently using the FileItem constructor taking a PVRChannel as that provides the best information on the OSD screens (Basically, its like watching the channel in time-shift).

How will such a file item be created in the future?

Is it better to put this PR on hold until all the changes to the PVR api and FileItem handling are done?

xbmc/addons/PVRClient.cpp
```diff
@@ -657,6 +689,46 @@ PVR_ERROR CPVRClient::RenameChannel(const CPVRChannelPtr &channel)
...
+  if (!m_bReadyToUse)
+    return PVR_ERROR_SERVER_ERROR;
+
+  if (!m_addonCapabilities.bSupportsRecordings)
+    return PVR_ERROR_NOT_IMPLEMENTED;
+
+  PVR_ERROR retVal(PVR_ERROR_UNKNOWN);
+
+  EPG_TAG pvrTag;
+  WriteEpgTag(tag, pvrTag);
```
rbuehlma: As far as I understand, Kodi does cache the EPG. I expect that most add-ons do not cache this information. Also, I expect that It is possible that this function is called with an EPG entry which is not in the cache of the add-on (e.g. if Kodi used the information form its cache on did not request a refresh from the add-on).
In my case (Zattoo) it would be possible to request the EPG information from their website, but that would be far more expensive. Also I'm not sure if that will be possible for all other providers.
This is why I decided to pass the full entry back to the addon. Do you still suggest to only pass the id? If yes, I will change the API.

FernetMenta:
> Is it better to put this PR on hold until all the changes to the PVR api and FileItem handling are done?

I can't answer this because I don't know when this will be finished. 

What I am saying is that we should not continue with that old pvr hack streamURL. This is announced in the forum and devs have had plenty of time to adapt to this. pvr needs to follow kodi's design. FileItems can be prefixed by a protocol like pvr://. If a path is prefixed with "pvr", the pvr addon must handle the stream.

rbuehlma:
>  If a path is prefixed with "pvr", the pvr addon must handle the stream.
Doesn't that contradict the separation of PVR-API and InputStream? As far as I unterstood the PVR addon should only be responsible to provide the URL which is then handled by one of the inputstream addons. Is this wrong?

FernetMenta:
Nope, this is not wrong but it has to apply to the rules already in place. A resource prefixed with "http://" won't get picked up by some ftp handler and a resource prefixed with protocol "pvr" won't get picked up by an mpeg-dash addon. We must not abuse what has been designed to be a protocol. pvr is actually a protocol implemented by addons like vnsi or tvh. The problem is that it got abused and we need to take corrective action.

One of the first things that need be corrected is prefixing channel items with "pvr://".

rbuehlma:
Ok, if I understand this correctly, an EPG item (and also other PVR related items) should always produce a FileItem containing a "pvr://".
This URL will then be handled by the corresponding PVR-addon.
Is that correct so far?
If yes, how should the PVR-addon then pass the actual URL (e.g DASH or what ever) to the inputstream addon?

FernetMenta:
> Ok, if I understand this correctly, an EPG item (and also other PVR related items) should always produce a FileItem containing a "pvr://".

This is how it's currently done. This is a big issue. Again: "pvr://" is a protocol that is not applicable to all pvr addons. You know what a protocol is, don't you?

rbuehlma:
A protocol defines how something is transferred but not what it transfers. If there is no further specification of the pvr://, it can be perfectly fine to use it to return the actual stream URL to the player (similar to http 302).

Is there some specification of that protocol available?

Its still completely unclear to me on how the desired solution will look like. What happens after the user has selected a show (epg tag), channel or recording to de played? Maybe its easier for you to describe the approach than for me trying to guess it.

FernetMenta:
> Maybe its easier for you to describe the approach than for me trying to guess it.

pvr architecture is exhausted and I know many weaknesses but this does not mean that I have a solution for everything. 

> pvr://, it can be perfectly fine to use it to return the actual stream URL to the player (similar to http 302)

Sure but this mechanism is not implemented in Kodi. FileItem is a generic element and you would need to touch many other components than only pvr. Every component that works with FileItems would be required to resolve this kind of redirect url before knowing how to deal with the item.

rbuehlma:
Rebase to current master.

rbuehlma:
I have added the new functionality to the context menu of the guide. Also "Smart select" does now replay a program if possible.

ksooo:
As I explained in the forum, I think that using recordings to achieve replay functionality is the right approach. Almost everything we need for "replay" is already in place and possible today with recordings. Recordings conceptional fit really nice here, because a replayable item for me feels exactly like a recording - with the only difference that somebody else did the recording, not me. Why should we introduce new API functions for something we already have?

The only unsolved problem with recordings as they work today is memory usage, as for your use case there is large amount of recordings involved. Currently the list of all available recordings is fetched from the client during Kodi pvr startup. For your use case this could get a killer memory-wise. I suggest to concentrate on solving this general conceptional problem instead of introducing more and more of functionality for something we basically already have.

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
```diff
@@ -558,6 +558,37 @@ extern "C"
...
 
+  /*
+   * Check if the given EPG tag can be recorded.
+   * @param tag the epg tag
+   * @param isRecordable Set to true if the tag can be recorded
+   * @return PVR_ERROR_NO_ERROR if isRecordable was set by the addon
+   * @remarks Optional, return PVR_ERROR_NOT_IMPLEMENTED to let kodi decide
+   *
+   */
+  PVR_ERROR IsRecordable(const EPG_TAG &tag, bool *isRecordable);
```
ksooo: This is a C API. Using C++ language constructs (here: referenecs) is a nogo.

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
```diff
@@ -558,6 +558,37 @@ extern "C"
...
+
+  /*
+   * Check if the given EPG tag can be played.
+   * @param tag The EPG tag
+   * @return True if the EPG tag can be played
+   * @remarks Required, always return false if not supported by the addon
+   *
+   */
+  bool IsPlayable(const EPG_TAG &tag);
```
ksooo: This is a C API. Using C++ language constructs (here: referenecs) is a nogo.
For consistency reasons, return value should be PVR_ERROR and the bool should be an out param (exactly like done for IsRecordable)

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
```diff
@@ -558,6 +558,37 @@ extern "C"
...
+   * Get the URL to play a given EPG tag
+   * @param tag The EPG tag
+   * @param url The url to be returnd
+   * @param urlLen The length of the url buffer
+   * @param properties The properties to be provided to the file item
+   * @return The length of the url. -1 if the tag is not playable
+   * @remarks Required, always return -1 if not supported by the addon
+   *
+   */
+  int GetEpgTagUrl(const EPG_TAG &tag, char *url, int urlLen, const CStringPropertyMapPtr&);
```
ksooo: This is a C API. Using C++ language constructs (here: referenecs and std containers) is a nogo.
Return value should be PVR_ERROR and urlLen should be an inout param (in buffer size, out url length)

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h
```diff
@@ -103,6 +105,10 @@ extern "C" {
     xbmc_codec_id_t   codec_id;
   } xbmc_codec_t;
 
+  typedef std::map<std::string, std::string> CStringPropertyMap;
+  typedef std::shared_ptr<CStringPropertyMap> CStringPropertyMapPtr;
```
ksooo: This is a C API. Using C++ here is a nogo.

ksooo (changes requested):

rbuehlma:
I agree that you could probably implement replay as recordings but there are some recording feature you do not want to have in case of replay. Some I am thinking of:
- Should not need to load half of the guide (the past) again as recordings
- Should not show up in the recordings list
- Should not generate a popups if a recording is planned/completed
- Should not be possible to edit/delete in any way
- Should not show the "recorded" icon in the guide
and probably more...

I'm therefore not sure if the better way is to "duplicate" some of the feature or implement exception for quite some feature used for recordings. What do you think?

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
rbuehlma: I will look into these c++ constructs as soon as it is decided which way the feature should be implemented. I'm not really used to C nor C++ and am therefore having some trouble to distinguish each feature set, but that is something I can learn :)

FernetMenta:
there are indeed considerable differences. treating those cases differently, learn, then generalize and factor out common code is imo much better than do generalization up front. the latter always results in a mess.

ksooo:
>    Should not need to load half of the guide (the past) again as recordings

 That's what I said. we should concentrate on solving this.

>    Should not show up in the recordings list

 Why? What would be the disadvantage of having them there?

 >   Should not generate a popups if a recording is planned/completed

 Will not happen as the popups are for timers, not recordings.

 >   Should not be possible to edit/delete in any way

 Could easily added to the API (just one read-only flag).

 >   Should not show the "recorded" icon in the guide

 There is no such icon. There is a "play" icon if a recording is available for an epg event. This fits perfect, imo.

>    and probably more...

 Name it, then we can discuss it.

rbuehlma:
>> Should not show up in the recordings list

> Why? What would be the disadvantage of having them there?

I as a user would be confused if I find multiple 10k of recordings which I have never recorded. I would need some sort of filter to find my own recordings. Forcing the user to enable a filter to be able to use the recordings is probably not the way to go.

>> Should not generate a popups if a recording is planned/completed

> Will not happen as the popups are for timers, not recordings.

If I remember it correctly, I get a popup if a new recording is inserted after the initial load, but not sure about that.

>> Should not show the "recorded" icon in the guide

> There is no such icon. There is a "play" icon if a recording is available for an epg event. This fits perfect, imo.

I agree that it would technically fit, but do you want to have this icon on every past show in the guide? I expect that it would look a bit ugly, and you would not find out which show has actually been recorded by you (and will persist after the replay period).

Considering that the transfer of the Replay-recordings needs to somehow be prevented (for example by delivering the "recorded" information together with the EPG item), I do only see the "Play"-action in the guide as a feature I would want to reuse from the recordings.
Did I miss some there?

To minimize the API changes from my approach, we could add "isPlayable" and "isRecordable" to the EPG item and removing these two functions from the API. But retrieving the URL still needs to somehow be done after the user has selected the program to replay.

ksooo:
Sorry, I'm not with you. You need to convince somebody else from the Team if you want to get this PR merged some day or we should agree on the recordings approach. For the latter you have my full backing, including me providing the necessary core and API changes. I don't want to discuss this to death, tbh.

ksooo:
>  I would need some sort of filter to find my own recordings.

 The addon can define virtual folders to organize recordings. For instance, there could be two toplevel folders, one labeled "my recordings", the other "TV program replay". You can even add subfolders for the different channels or whatever you like.

> If I remember it correctly, I get a popup if a new recording is inserted after the initial load, but not sure about that.

 You don't remember this correctly. Should be no problem at all.

> but do you want to have this icon on every past show in the guide?

 Yes, I want a "play" icon for every epg item which can be played, no matter who "recorded" this.

> I do only see the "Play"-action in the guide as a feature I would want to reuse from the recordings.

 What other recordings functionality is in the guide window you don't want to use?

> Considering that the transfer of the Replay-recordings needs to somehow be prevented (for example by delivering the "recorded" information together with the EPG item)

> To minimize the API changes from my approach, we could add "isPlayable" and "isRecordable" to ...

Please don't get me wrong, but I have the bigger picture. Even adding small things like isplayable and isrecordable in whatever way (function, struct field - does not matter) means, that every (!) pvr addon implementation needs to be adapted to support this properly or ugly chunks of compatibility code need to be injected into the Kodi pvr code ("if functionality supported to it the new way if not do it like in the past" - no go for this!). Given that some of the addons don't have active maintainers anymore changes like you suggest will lead to the immediate death of those addons - which I want to prevent if possible.

rbuehlma:
Well, I'm fine with also implementing your proposed approach. I'm not strictly against it, I'm just not sure which one will cause less trouble :)

Probably the best is, to try both since I don't expect a lot implementation effort.

For me, the most unclear point is this:

>> Should not need to load half of the guide (the past) again as recordings

> That's what I said. we should concentrate on solving this.

Do you have something in mind how we could solve that? First try to simply load them as recordings an see how it behaves?

ksooo:
> First try to simply load them as recordings an see how it behaves?

 This would be a good first step, yes. If it turns out then that there will be no problems, we would have saved much time for a concept and implementation nobody needs. ;-)

 Can you do e a favor and close this PR and open another one for the new approach. This one has so many comments and stuff that it is really hard to keep track what's still of interest...

rbuehlma:
Ok, I'll implement the second approach in the next days and will create a new PR for that one. Until we have one of the solutions accepted, I will keep this one open. Is that ok?

ksooo:
Ok, I will just stop watching this PR.

rbuehlma:
Superseeded by https://github.com/xbmc/xbmc/pull/12689

PR #10364, opened 2016-08-30
Peripheral API v1.1.0: Fix mapping D-pad on RPi for some controllers

garbear:
Some controllers send both a button press and an axis change when the D-pad is pressed. To compensate for this, I added a "cooldown" to Kodi so that input arriving within 50ms would be ignored.

A design decision I made in #8807 was to save the button map to the disk after every button press (it simplified the API). When the button map was saved, it would trigger a refresh, and Kodi would hit the disk another few times to calculate the updated button map. I suspect these disk accesses were adding up to >50ms, which is why Kodi wasn't ignoring the second driver input.

This PR adds a SaveButtonMap() function to the peripheral API and only saves the button map when the button-mapping wizard comes to an end.

Also included is a pot-pourri of improvements, some of which the v1.1.0 bump relies on.

Corresponds to https://github.com/kodi-game/peripheral.joystick/pull/50


garbear:
Unrelated improvements have been separated out into #10371

jenkins build this please

PR #10366, opened 2016-08-30
allow passing an alternative dyload library path

wsnipex:
it's only implemented for linux atm, but I'll happily add other platforms, if the general idea is accepted and also wanted on other platforms.

If the env var KODI_BINADDON_PATH is non empty, we first try to find shared libs in this path.

### DIFF
--- a/xbmc/addons/AddonDll.h
+++ b/xbmc/addons/AddonDll.h
@@ -133,6 +133,7 @@ bool CAddonDll<TheDll, TheStruct, TheProps>::LoadDll()
     return true;
 
   std::string strFileName;
+  std::string strAltFileName;
   if (!m_bIsChild)
   {
     strFileName = LibPath();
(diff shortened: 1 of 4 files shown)


xbmc/addons/AddonDll.h
```diff
@@ -176,6 +177,25 @@ bool CAddonDll<TheDll, TheStruct, TheProps>::LoadDll()
 #endif
   if (!XFILE::CFile::Exists(strFileName))
   {
+    std::string altbin = CSpecialProtocol::TranslatePath("special://xbmcaltbinaddons/");
+    if (!altbin.empty())
+    {
+      strAltFileName = altbin + m_props.libname;
```
akva2: one huge dump dir for all binary add-ons?

wsnipex: yeah, that was the intention. But the case /some/path/<addon_dir>/<addon_lib> is also checked afterward.

akva2: Right. Just making sure you didnt intend dir/id/binary (in case LibPath () helper is what was wanted.

wsnipex:
jenkins build this please

stefansaraev:
this PR is responsible for at least pvr.hts not loading http://sprunge.us/UJSN

does it fix something or it is a feature?

if not a fix - please revert.

FernetMenta:
I suggest that we revert this until problems got solved

wsnipex:
see https://github.com/xbmc/xbmc/pull/10419

PR #10367, opened 2016-08-30
DVDAudioCodecPassthrough: Fix memory leak after 

fritsch:
14af37aac18b0f471694ac7c65a96d39afaf327

av_packet_split_side_data(&pkt); allocates own buffers which are not freed properly

This only fixes the memleak, it does not care for the unsmooth seeking with that PR users reported.


fritsch:
jenkins build this please

FernetMenta:
thanks

PR #10368, opened 2016-08-30
FIX: [droid] ignore wifi-direct net intf (p2p)

fritsch:
For Android. The class is full of ifdeffery so it does not really matter.

### DIFF
--- a/xbmc/network/linux/NetworkLinux.cpp
+++ b/xbmc/network/linux/NetworkLinux.cpp
@@ -116,6 +116,12 @@ bool CNetworkInterfaceLinux::IsEnabled()
 
 bool CNetworkInterfaceLinux::IsConnected()
 {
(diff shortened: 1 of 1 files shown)


fritsch:
See: http://forum.kodi.tv/showthread.php?tid=288755

Jalle19:
Is the ifdef really necessary? Theoretically any Linux box could have Wifi Direct interfaces if I'm not mistaken.

fritsch:
No idea. Find out and send me a PR, will Squash it in :-)

Am 30.08.2016 10:20 nachm. schrieb "Sam Stenvall" <notifications@github.com

Jalle19:
From what I could fine the interface name is arbitrary when configured manually on Linux.

fritsch:
Standing here for too long, seems no one needs it.

PR #10369, opened 2016-08-30
Add a numeric field type that doesn't need casting

phate89:
Currently in smart playlists in integer fields we do `CAST(bla as DECIMAL(5,1))` because of the videodatabase fields that were all strings.
Now there are legitimate numeric db fields so this creates problems with votes where we have integers bigger than 5 characters that gets changed by casting messing the result.
So with this pr I'll add a field for true numeric fields that doesn't need casting.
This fixes http://forum.kodi.tv/showthread.php?tid=220012&pid=2404701#pid2404701
@DaveTBlake I added also tracknumber to these fields. can you check if it's ok?


phate89:
jenkins build this please

DaveTBlake:
Thanks Phate I need to look more at this, will do over the weekend. 
I suspect that track need completely different handling.  The iTrack field of the song table holds both tack number and disc number in upper and lower bits, so CAST(songview.iTrack as DECIMAL(5,1)) = 3 is never going to find track number 3.

phate89:
@DaveTBlake did you have time to look into it? Do I merge it excluding tracknumber to solve the rest and you'll fix it afterwards?

DaveTBlake:
@phate89 I had a quick look. As I feared track number rule doesn't work, and still won't after your change. So yes leave it out, and I will add to my list of things to fix with music smart playlists allowing for both the new numeric type and decoding the disc/track combination.

But good for fixing the rest.

phate89:
jenkins build this please

phate89:
jenkins build this please

PR #10371, opened 2016-08-31
Break out joystick improvements from #10364

garbear:
This PR breaks out the changes that the fix in #10364 doesn't depend on.

One cool change is that the add-on icon is shown in the peripherals dialog. This won't have any effect until X-Arcade Tankstick and Steam Controller support is finished, but in the future we can look forward to this:

peripheral icons


garbear:
jenkins build this please

garbear:
added a commit from may that I forgot about and have been carrying in my retroplayer branch.

jenkins build and merge

PR #10372, opened 2016-08-31
[cmake] fix sse detection

wsnipex:
Fixes false SSE detection in some environments, e.g. buildroot for arm

@fetzerch @Memphiz @fritsch for sanity checks and testing


wsnipex:
jenkins build this please

fritsch:
@wsnipex from what I see - and I don't see much in cmake - it seems you are doing the right thing, checking sse4.1 and so on. If the others are fine, we can test it in master. The none SSE4 cpus get less and less - barely not many ION's available anymore :-)

Only @MilhouseVH is still using one in a productive environment.

wsnipex:
jenkins build this please

MilhouseVH:
I believe Hitcher is an ION lover too.

xbmc/linux/sse4/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
-set(SOURCES CopyFrame.cpp)
+if(HAVE_SSE4_1)
```
fetzerch: @mapfau: Might be interesting for you as it's one line you can then drop from https://github.com/mapfau/xbmc/commit/b2357611a8c2f81418d61a5911d85db915a8c698

fetzerch:
Looks good to me, thx!

Memphiz:
Currently unable to check this out :( are you sure with those -mblah flags for clang?

wsnipex:
didn't change the flags, I just added avx. Those I'm not sure about, but since sse* seems to be the same as on gcc, I guess they are the same for avx as well.

wsnipex:
updated. We now enable all SSE versions <= 4.1 by default(if available).
@fritsch @FernetMenta :  I've replaced all the __SSE__ macros in AE, so its now possible to disable SSE use.
Disabling SSE altogether brings up an old issue we only observed on i386 before, so it seems we don't compile without it.
see http://pastebin.com/cq4aK60E

wsnipex:
jenkins build this please

wsnipex:
jenkins build this please

wsnipex:
jenkins build this please

Memphiz:
@wsnipex looking at the jenkins output it looks ok'ish (for ios no extensions are found as expected). For OSX the output is as follows:

http://pastebin.com/kR10RFdv

Not sure if this is correct. I wonder how this works. It finds the extensions that are available on the build host - right? How to ensure that the target where this runs has the same abilities in the end?

Beside that i wonder why it says "-- Cross-Compiling: FALSE" on osx - maybe @fetzerch has an idea? (you know osx is a cross compile as it uses the sdk buildchain, not the host one...)

Memphiz:
Beside that i i've given the kodi-20160901-906b6a8-cmake_sse-x86_64.dmg a shot  (from sept. 1st) and it seems to work ok. Won't be able to test anything else for this PR - so OK from me.

wsnipex:
yes, it checks the build host, same as before my change. The difference is that now you can disable sse extensions, which wasn't possible before. Currently only SSE(1) seems to be absolutely necessary to build(at least on linux), and we only compile with up to SSE2 in AudioEngine. I don't think there are any intel based Macs that don't support this.
The SSE4.1 based libary is linux only and is runtime checked

fetzerch:
> Beside that i wonder why it says "-- Cross-Compiling: FALSE" on osx - maybe @fetzerch has an idea?
> (you know osx is a cross compile as it uses the sdk buildchain, not the host one...)

`CMAKE_CROSSCOMPILE` is just a variable that is set if the toolchain has `CMAKE_SYSTEM_NAME` set (as for IOS: https://github.com/xbmc/xbmc/blob/master/tools/depends/target/Toolchain.cmake.in#L18). 
We could set it for OSX too, it just needs a few adaptions then. For example `CMAKE_CROSSCOMPILE` currently defines if unit testing should be enabled.

FernetMenta:
this broke cmake build with XCode target

```
CMake Error at /Users/DE213022/progs/src/xbmc/xbmc/rendering/CMakeLists.txt:8 (target_compile_options):
  Cannot specify compile options for target "rendering" which is not built by
  this project.
```

Memphiz:
in addition - its fine when generating gnu make files - strange ...

PR #10373, opened 2016-08-31
[PVR] Recordings: Prevent thumbnail extraction (as it cannot work pro…

ksooo:
…perly without major pvr addon api changes).

Fixes the ugly problem reported here: https://github.com/xbmc/xbmc/pull/10333#issuecomment-243210489

@Jalle19, @FernetMenta good to go? I think I now found a good place for the "pvr recording check"...

### DIFF
--- a/xbmc/video/VideoThumbLoader.cpp
+++ b/xbmc/video/VideoThumbLoader.cpp
@@ -86,6 +86,9 @@ bool CThumbExtractor::operator==(const CJob* job) const
 bool CThumbExtractor::DoWork()
 {
   if (m_item.IsLiveTV()
+  // Due to a pvr addon api design flaw (no support for multiple concurrent streams
+  // per addon instance), pvr recording thumbnail extraction does not work (reliably).
+  ||  m_item.IsPVRRecording()
   ||  URIUtils::IsUPnP(m_item.GetPath())
   ||  URIUtils::IsBluray(m_item.GetPath())
   ||  m_item.IsBDFile()
--- a/xbmc/cores/VideoPlayer/DVDFileInfo.cpp
+++ b/xbmc/cores/VideoPlayer/DVDFileInfo.cpp
@@ -103,10 +103,6 @@ bool CDVDFileInfo::ExtractThumb(const std::string &strPath,
   unsigned int nTime = XbmcThreads::SystemClockMillis();
   CFileItem item(strPath, false);
 
-  if (item.IsDiscImage() ||
-      item.IsPVR())
-    return false;
-
   item.SetMimeTypeForInternetFile();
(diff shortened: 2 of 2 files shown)


ksooo:
@axbmcuser fyi

xbmc/video/VideoThumbLoader.cpp
```diff
@@ -86,6 +86,9 @@ bool CThumbExtractor::operator==(const CJob* job) const
 bool CThumbExtractor::DoWork()
 {
   if (m_item.IsLiveTV()
```
ksooo: > Doesn't this include recordings?

No. Definitely not. I checked the implementation of IsLiveTV() and ofc did runtime tests.

Jalle19: Just checking, +1 in that case.

ksooo:
jenkins build this please

FernetMenta:
Could you wait until Saturday? Why got this broken again? iirc I have fixed this some time ago.

ksooo:
> iirc I have fixed this some time ago.

 @FernetMenta You mean this fix, right? https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDFileInfo.cpp#L107

 Fine, but there are more cases you didn't catch. So, it's still broken, not again, imo. ;-)

<code>CVideoThumbLoader</code> instanciates and starts <code>CThumbExtractor</code>s => https://github.com/xbmc/xbmc/blob/master/xbmc/video/VideoThumbLoader.cpp#L417

<code>CThumbExtractor::DoWork</code> not only calls <code>CDVDFileInfo::ExtractThumb</code> (the case you fixed), but also <code>CDVDFileInfo::GetFileStreamDetails</code> =>  https://github.com/xbmc/xbmc/blob/master/xbmc/video/VideoThumbLoader.cpp#L140

My approach is more restrictive than yours because it generally prevents CThumbExtractor to access pvr recordings, regardless from where it's called. BTW, I did not invent this "type filter" approach at the beginning of <code>CThumbExtractor::DoWork</code>, I just added one case.

FernetMenta:
Thanks for explanation. Go ahead

> In case we agree to merge this PR, your abovementioned fix could even be removed

could you do this? maybe move the other exceptions like IsDiscImage to the other place too?

ksooo:
@FernetMenta yep, will do.

ksooo:
>> In case we agree to merge this PR, your abovementioned fix could even be removed

> could you do this? maybe move the other exceptions like IsDiscImage to the other place too?

@FernetMenta done. IsDiscImage check was at "the other place" already, btw.

ksooo:
jenkins build this please

ksooo:
Travis error is unrelated

PR #9336, opened 2016-03-12
[PVR] prevent crash on .. when recordings list empty

metaron-uk:
Currently when the pvr.mythtv client starts up it returns an empty recordings list until backend sync is complete. If you enter the recordings list during this time, you end up with only .. present.
If you select .., Kodi crashes.
screenshot043

With commit 1 you get taken back to the home screen instead, which is what I expected would happen.

Commit 2 contains some 'robustness' additions I made while debugging.
It seems kodi tries to exit when it can't follow the .. path. These changes let it carry on trying to exit for a little longer, which makes it easier to work out what's happening from the log.
Happy to drop this commit if it isn't seen as useful.

pings @ksooo @janbar

PR #10374, opened 2016-08-31
[PVR] Recordings window: fix crash when selecting '..' (fixes trac#16866)

ksooo:
http://trac.kodi.tv/ticket/16866

Before:
screenshot006

After:
screenshot005

@Jalle19 any objections?
@BigNoid for the (small) Estuary change?
@axbmcuser fyi

### DIFF
--- a/addons/skin.estuary/1080i/MyPVRRecordings.xml
+++ b/addons/skin.estuary/1080i/MyPVRRecordings.xml
@@ -243,12 +243,12 @@
 						<height>262</height>
 						<label>$LOCALIZE[19076] ($INFO[Container(5000).NumItems,[B],[/B] $LOCALIZE[31036]])</label>
 						<font>font16</font>
-						<visible>ListItem.IsFolder</visible>
+						<visible>ListItem.IsFolder + !ListItem.IsParentFolder</visible>
 					</control>
 					<control type="group">
 						<left>-10</left>
 						<top>60</top>
-						<visible>ListItem.IsFolder</visible>
+						<visible>ListItem.IsFolder + !ListItem.IsParentFolder</visible>
 						<include content="InfoList">
 							<param name="path" value="$INFO[ListItem.Path]/$INFO[ListItem.Label]" />
 							<param name="height" value="362" />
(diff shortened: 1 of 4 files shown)


Jalle19:
Wouldn't it be better to just not show the go to parent item if you're already in the root directory?

addons/skin.estuary/1080i/MyPVRRecordings.xml
```diff
@@ -243,12 +243,12 @@
 						<height>262</height>
 						<label>$LOCALIZE[19076] ($INFO[Container(5000).NumItems,[B],[/B] $LOCALIZE[31036]])</label>
 						<font>font16</font>
-						<visible>ListItem.IsFolder</visible>
+						<visible>ListItem.IsFolder + !String.IsEmpty(ListItem.Date)</visible>
```
ronie: perhaps use this instead?
```ListItem.IsFolder + !ListItem.IsParentFolder```

ksooo:
> Wouldn't it be better to just not show the go to parent item if you're already in the root directory?

 yes, this would be the best solution, imo. Unfortunately GUIMediaWindow cannot handle empty lists (i was told). There is even really ugly code that adds some ".." (for pvr not really working => root fs listing!) item to prevent problems in case of empty list. => https://github.com/xbmc/xbmc/blob/master/xbmc/windows/GUIMediaWindow.cpp#L1816 ff

addons/skin.estuary/1080i/MyPVRRecordings.xml
ksooo: Yeah. Much better. Will change.

Jalle19:
Okay, in that case I suppose it's good. Would be nice to be able to fix the root cause somehow though, this path hackery is a bit hard to follow.

ksooo:
> this path hackery is a bit hard to follow.

@Jalle19 reworked. Looks much cleaner to me now. Agreed?

ksooo:
jenkins build this please

metaron-uk:
I suspect this may fix the problem  I was  having
https://github.com/xbmc/xbmc/pull/9336

PR #10376, opened 2016-09-01
log failed texture loading

tamland:
Currently the job manager is flooded with texture caching jobs trying to load an empty path. This should prevent that and actually log the error where it occurs. 

This doesn't fix the cause of the issue though (it still polls the filesystem) as I don't know what it is, but it seem to come from estuary attempting to lookup a texture named  '-'. <- @phil65 @BigNoid any ideas?

### DIFF
--- a/xbmc/GUILargeTextureManager.cpp
+++ b/xbmc/GUILargeTextureManager.cpp
@@ -49,6 +49,9 @@ bool CImageLoader::DoWork()
   std::string loadPath;
 
   std::string texturePath = g_TextureManager.GetTexturePath(m_path);
+  if (texturePath.empty())
+    return false;
+
   if (m_use_cache)
     loadPath = CTextureCache::GetInstance().CheckCachedImage(texturePath, needsChecking);
   else
@@ -209,6 +212,9 @@ void CGUILargeTextureManager::ReleaseImage(const std::string &path, bool immedia
 // queue the image, and start the background loader if necessary
 void CGUILargeTextureManager::QueueImage(const std::string &path, bool useCache)
 {
+  if (path.empty())
+    return;
+
   CSingleLock lock(m_listSection);
   for (queueIterator it = m_queued.begin(); it != m_queued.end(); ++it)
   {
--- a/xbmc/TextureCache.cpp
+++ b/xbmc/TextureCache.cpp
@@ -114,18 +114,28 @@ std::string CTextureCache::CheckCachedImage(const std::string &url, bool &needsR
 
 void CTextureCache::BackgroundCacheImage(const std::string &url)
 {
+  if (url.empty())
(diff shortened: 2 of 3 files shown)


BigNoid:
Problem is using ```-``` as fallback for textures. Afaik it can just be left blank to make sure it's an empty path (and it won't try to cache it than with this pr). You can pick https://github.com/BigNoid/xbmc/commit/1bbc5bed80b7d9507b8d9af0a431b56a486b2b5a

tamland:
Perfect, thanks!

jenkins build this please

ronie:
@BigNoid 
does this only apply to ```<texture fallback="-">``` or ```<texture>-</texture>``` as well?

and what about ```<texturenofocus>-</texturenofocus>``` ?

since it's common practice to use "-" to specify an empty value (not just images, but fonts and labels as well), i'm wondering what the recommended way is...

tamland:
Yeah, there's a lot more it seems..
Why is using "-" instead of empty value a common practice??

ronie:
common practice as in 'everybody has been doing it for as long as i can remember'
afaik no one ever mentioned it shouldn't be done like that.

i'm not sure if just using an empty value will work as well, but i'll test it in a minute.

BigNoid:
From my tests the error only appears when ```-``` is used in variables.

ronie:
> From my tests the error only appears when - is used in variables.

hmm... that's kinda odd isn't it?

either way, specifying an empty value seems to works just fine

tamland:
I see it's mentioned a few times in core code that '-' means 'no texture' but clearly this broke at some point as there are many things that doesn't handle it. Will open a new pr. Merging the rest as those changes should be correct regardless.

MilhouseVH:
Just an observation when using stock Estuary (latest Kodi master), but since this change when viewing videos and pulling up the Info panel, I see the following errors being logged:
```
18:21:06 51041.457031 T:1962061824   ERROR: CGUITextureManager::GetTexturePath: could not find texture 'flags/video/1080.png'
18:21:06 51041.457031 T:1962061824   ERROR: CGUITextureManager::GetTexturePath: could not find texture 'flags/audio/truehd.png'
18:21:06 51041.457031 T:1962061824   ERROR: CGUITextureManager::GetTexturePath: could not find texture 'flags/audio/2.png'
```
1080.png is in `flags/videoresolution/1080.png`
truehd.png is in `flags/audiocodec/truehd.png`
2.png is in `flags/audiochannel/2.png`

Is Estuary looking in the wrong place, or is this some sort of alternative fallback location that is now logged as an error? Ping @phil65 @ronie etc. Thanks.

ronie:
thx for the report@
https://github.com/xbmc/xbmc/pull/10415

PR #10377, opened 2016-09-01
unify language select with other addon select dialogs

tamland:
Why does language select have to have its own special dialog..? I've always disliked that this dialog shows a bunch of languages I don't care about and that it hides the fact that selecting one involves downloading and installing an addon.

### DIFF
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -3111,7 +3111,7 @@
             <addontype>kodi.resource.language</addontype>
           </constraints>
           <control type="button" format="addon">
-            <show more="false" details="false">all</show>
+            <show more="true" details="true">installed</show>
           </control>
         </setting>
         <setting id="locale.charset" type="string" label="14091" help="36116">


Montellese:
This was specifically requested when I PRed the addonification of languages. I also think it makes sense because all systems I know show you all available languages independent of whether they are available offline or not. It also makes it easier to change the language for people who don't understand English. But since I never change this setting I don't really care.

system/settings/settings.xml
```diff
@@ -3111,7 +3111,7 @@
             <addontype>kodi.resource.language</addontype>
           </constraints>
           <control type="button" format="addon">
-            <show more="false" details="false">all</show>
+            <show more="true" details="true">installed</show>
```
Montellese: I'd keep "details" to "false" because we don't have proper addon icons for language addons.

MartijnKaijser:
I'd agree with Montellese that for languages it's best to show all available. 
Partially related it might be a good idea to have some confirmation that one the language is switched we ask the user if he wants to keep it as easy to click the wrong one. 

All language addons should have their own icon now days (they are just all the same)

scott967:
Changing languages to addons was a design decision.  It's hard enough for users that the default font for the default skin can't display characters for various languages without making them search for language support.

PR #10378, opened 2016-09-01
[Estuary] missing icons and a few updated ones

ronie:
-add a few missing default icons

-updated some that still smelled like confluence

@phil65


phil65:
Thx!

PR #10380, opened 2016-09-02
[cmake] Fixes: in-tree addon builds and CMAKE_INSTALL_PREFIX inconsistency

hudokkow:
Fix in-tree addon build env after 4a65001. See [comment](https://github.com/xbmc/xbmc/commit/4a65001cafb54f8c925498d6c90d28a5793a068e#commitcomment-18689259)

Fix CMAKE_INSTALL_PREFIX not being honored by CPack generated DEBs if script is called without -DCMAKE_INSTALL_PREFIX=chihuahua, leading to CMake assuming default value (/usr/local for linux).

Second one costed me 2h until I realized what the ***\* was going on and went digging in docs to find CPACK_SET_DESTDIR.

### DIFF
--- a/project/cmake/CMakeLists.txt
+++ b/project/cmake/CMakeLists.txt
@@ -408,10 +408,13 @@ if(CMAKE_GENERATOR STREQUAL "Unix Makefiles")
 endif()
 
 # Prepare add-on build env
-foreach(binding ${addon_bindings})
-  get_filename_component(file ${binding} NAME)
-  set(BINDING_TARGET ${CORE_BUILD_DIR}/include/${APP_NAME_LC}/${file})
-  configure_file(${binding} ${BINDING_TARGET} COPYONLY)
+core_file_read_filtered(bindings ${CORE_SOURCE_DIR}/xbmc/addons/addon-bindings.mk)
+foreach(binding ${bindings})
+  string(REPLACE " =" ";" binding "${binding}")
+  string(REPLACE "+=" ";" binding "${binding}")
+  list(GET binding 1 header)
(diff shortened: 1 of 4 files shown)


hudokkow:
> if you don't set CMAKE_INSTALL_PREFIX, the package still builds fine, but you can't run kodi because the script in /usr/bin/kodi expects the things in /usr/local/ (but the package installs to /usr)

@fetzerch, second commit fixes ^^

git commit not being updated on a rerun is less important and I'll fix it when I find time.

jenkins build this please

project/cmake/CMakeLists.txt
```diff
@@ -408,7 +408,13 @@ if(CMAKE_GENERATOR STREQUAL "Unix Makefiles")
 endif()
 
 # Prepare add-on build env
-foreach(binding ${addon_bindings})
+file(STRINGS ${CORE_SOURCE_DIR}/xbmc/addons/addon-bindings.mk bindings)
+string(REPLACE "\n" ";" bindings "${bindings}")
```
wsnipex: you might want to use our macros for this, core_file_read_filtered

hudokkow: No comment lines so no need for it. But you gave me an idea. Will update later today

fetzerch:
Thanks much, works :) But it's probably a good idea to use the macro as wsnipex mentioned. Then it'll still work even if someone adds a comment to that file.

hudokkow:
Updated.

hudokkow:
Usual suspects failed. Rest is [fine](http://jenkins.kodi.tv/job/BuildMultiWithAddons-PR/25/).

PR #10381, opened 2016-09-02
[Estuary] fix / unify PVR window breadcrumbs

ksooo:
@phil65 @ronie @BigNoid good to go?

@da-anda this fixes the problem you lately mentioned on slack

### DIFF
--- a/addons/skin.estuary/1080i/Includes.xml
+++ b/addons/skin.estuary/1080i/Includes.xml
@@ -897,7 +897,7 @@
 					<top>0</top>
 					<width>120</width>
 					<height>100</height>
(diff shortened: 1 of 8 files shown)


ksooo:
No experienced skinner around to give feedback?

ronie:
breadcrumb changes look fine to me.

i'm not really familiar with the Container.Update() action though
that one needs to be checked by @phil65 / @BigNoid

ksooo:
passing a string containing just a blank, how it is done before this PR, is definitely not the intention. I looked at the implementation of container.update and found that (of course) the path of the container to update is to pass to that function.

PR #10383, opened 2016-09-02
VideoPlayer: fix pts for passthrough audio

FernetMenta:
see title

@MilhouseVH does this fix the seeking issue with passthrough enabled?

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Audio/DVDAudioCodecPassthrough.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Audio/DVDAudioCodecPassthrough.cpp
@@ -121,12 +121,6 @@ int CDVDAudioCodecPassthrough::Decode(uint8_t* pData, int iSize, double dts, dou
   int skip = 0;
   if (m_backlogSize)
   {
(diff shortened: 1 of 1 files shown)


MilhouseVH:
No, unfortunately seeking remains the same with this PR.

I'm using this sample file: https://www.dropbox.com/s/uptq22l2nv95h7r/blechnarrisch.ts?dl=0 (streamed over smb:// or from local playback - same behaviour with both).

I am testing a build that includes #10247 and #10367 (and not LE master which is based on Kodi 17 beta1 with #10247 reverted), plus this PR.

On an Intel NUC, with passthrough enabled, seeking (step forward/step backward) will result in the seek position becoming very "jittery" (the position indicator visibly moving backwards and forwards, basically wobbling), with playback hanging and the GUI showing "100% buffering".

I see the exact same behaviour on RPi2 when using MMAL, but not when using OMX (the latter is fine, no problems).

FernetMenta:
should work now

FernetMenta:
jenkins build this please

MilhouseVH:
Yes, perfect now! Many thanks.

PR #10385, opened 2016-09-03
Update uniqueids when we change imdbnumber via json

phate89:
Fixes http://trac.kodi.tv/ticket/16873


phate89:
jenkins build and merge

PR #10386, opened 2016-09-03
add scan to library context menu

tamland:
A "bug" in the old show info context menu that got removed during refactoring. But since it turned out to be a useful feature, adding it back;)


tamland:
jenkins build and merge

NedScott:
Semi-related to this, if a person has the info button mapped on their remote and presses it on an item not scanned in, will it still prompt to scan?

MartijnKaijser:
fixed strings in https://github.com/xbmc/xbmc/commit/e51e752fd22f8adc86f0e67a7b22f4537943a680

PR #10387, opened 2016-09-03
VideoPlayer: catch silly CRedirectException, fixes crash

FernetMenta:
fix http://trac.kodi.tv/ticket/16874

We really should not deal with C++ exceptions as long catch blocks can't be forced.

@MilhouseVH /cc

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDInputStreams/DVDFactoryInputStream.cpp
+++ b/xbmc/cores/VideoPlayer/DVDInputStreams/DVDFactoryInputStream.cpp
@@ -147,11 +147,22 @@ CDVDInputStream* CDVDFactoryInputStream::CreateInputStream(IVideoPlayer* pPlayer
     {
       XFILE::CCurlFile url;
       // try opening the url to resolve all redirects if any
-      if (url.Open(finalFileitem.GetURL()))
+      try
       {
-        finalFileitem.SetPath(url.GetURL());
+        if (url.Open(finalFileitem.GetURL()))
+        {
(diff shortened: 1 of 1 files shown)


MilhouseVH:
Tested on RPi2, problem appears to be fixed - thanks.

arnova:
@FernetMenta : IMHO the proper fix for this should not use CCurlFile open/close to het the final URL but just have a separate function in CCurlFile (without the exception part) giving us the final URL. Seems much cleaner ?

FernetMenta:
@arnova not sure what exactly you are proposing. Do you have time to do this?

arnova:
@FernetMenta : Working on it. I think in half an hour or so.

arnova:
@FernetMenta : What I had in mind doesn't seem that straightforward after all. But it still feels like we shouldn't be bothered with handling such exceptions at this level?

FernetMenta:
@arnova not sure if I got your question. What level do you mean?

arnova:
VideoPlayer level. The redirect was implemented as a hack for IFile, therefor it's handled there. Anything else not going through IFile, requires special treatment which sucks imho.

FernetMenta:
yes, VideoPlayer only need to catch the error.
Are you ok with this PR now?

arnova:
I think I can live with it ;-), although as *pretty* as I would like (note: can't think of anything else atm)...

FernetMenta:
@arnova thanks. still enough work to keep you busy in the future :)

PR #10388, opened 2016-09-03
skin: remove all use of hyphen as none value

tamland:
Follow-up on #10376. Appears to work fine..


BigNoid:
Maybe nitpicking, but would be better to format like ```<texturenofocus />``` imo

tamland:
Nothing against that, but too much work;) This is just a find and replace on `>-<`.

ronie:
let's get it in and see if any issues pop up...

ksooo:
I have noticed that now all parent folder list items have an "unread" overlay. Before, they had now overlay.

screenshot005

PR #10389, opened 2016-09-03
[guilib] drop support for using hyphen as none value

tamland:
There's currently many code paths that doesn't handle the "-" magic value. It's not clear to me why this value was used instead of empty values in the first place; which is commonly always handled...

We should probably schedule this v18 as it breaks all skins relying on "-" being interpreted as an empty value.

### DIFF
--- a/xbmc/TextureCache.cpp
+++ b/xbmc/TextureCache.cpp
@@ -62,7 +62,9 @@ void CTextureCache::Deinitialize()
 
 bool CTextureCache::IsCachedImage(const std::string &url) const
 {
-  if (url != "-" && !CURL::IsFullPath(url))
+  if (url.empty())
+    return false;
+  if (!CURL::IsFullPath(url))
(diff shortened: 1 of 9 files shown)


MartijnKaijser:
@ronie perhaps you can already include this in skin changes thread? If no skin uses it anymore we can perhaps add it.

HitcherUK:
What should we use instead? If I use -
`<texture></texture>`
Notepad++ will change to this when I use Pretty print -
`<texture/>`

ronie:
both are fine

rjsachse:
Would this work with the progress control as well? At the moment I have to create a blank texture for the background. eg for the cache progress.

ronie:
good to go?

tamland:
if jenkins still approves, so jenkins build this please

ronie:
windows failure is unrelated

PR #10390, opened 2016-09-03
[Estuary] Fix channel icons / thumbnails.

ksooo:
Made usage of channel icons / thumbnails consistent across all (hopefully) all pvr windows...

Notes:

Listitem.Icon for PVR items will return thumbnail, and as fallback the channel icon. Use this if you prefer thumbnail over channel icon.

Listitem.ActualIcon for PVR items returns the channel icon

@BigNoid @ronie any objections?


BigNoid:
No objections here

PR #10392, opened 2016-09-04
[estuary] Avoid empty variables

BigNoid:
Empty variables are skipped, so instead of empty use a window property `Window(Home).Property(empty.value)`

@ronie would this be acceptable or should core be changed to not skip an empty variable?

@ksooo this fixes the overlay issue on the parent item folder.

### DIFF
--- a/addons/skin.estuary/1080i/Variables.xml
+++ b/addons/skin.estuary/1080i/Variables.xml
@@ -19,16 +19,14 @@
 		<value>$INFO[Player.Time]$INFO[Player.Title, - ]</value>
 	</variable>
 	<variable name="AutoCompletionContentVar">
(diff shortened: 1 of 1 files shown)


ronie:
parhaps @tamland has an idea why empty values in variables are ignored?

@BigNoid what about this approach:

tamland:
like #10395? certainly doesn't look right to just skip it..

BigNoid:
Yep better @ronie as that's not a workaround like I had

ronie:
this is not really needed anymore after https://github.com/xbmc/xbmc/pull/10395

@phil65 think it's worth to merge it anyway? 
it's a bit of a cleanup/simplification of variable values.

phil65:
Some of these "weird" value orders were needed when I wrote it, but that might have changed with latest @tamland changes... Might be worth it to give it a try and watch out for regressions.

PR #10394, opened 2016-09-04
disable plugin caching

tamland:
Supersedes #9898

### DIFF
--- a/xbmc/filesystem/PluginDirectory.cpp
+++ b/xbmc/filesystem/PluginDirectory.cpp
@@ -227,7 +227,7 @@ void CPluginDirectory::EndOfDirectory(int handle, bool success, bool replaceList
     return;
 
   // set cache to disc
-  dir->m_listItems->SetCacheToDisc(cacheToDisc ? CFileItemList::CACHE_IF_SLOW : CFileItemList::CACHE_NEVER);
+  dir->m_listItems->SetCacheToDisc(CFileItemList::CACHE_NEVER);
 
   dir->m_success = success;
   dir->m_listItems->SetReplaceListing(replaceListing);
--- a/xbmc/interfaces/legacy/ModuleXbmcplugin.h
+++ b/xbmc/interfaces/legacy/ModuleXbmcplugin.h
@@ -131,8 +131,8 @@ namespace XBMCAddon
     /// @param updateListing        [opt] bool - True=this folder should
     ///                             update the current listing/False=Folder
     ///                             is a subfolder(Default).
-    /// @param cacheToDisc          [opt] bool - True=Folder will cache if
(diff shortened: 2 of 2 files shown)


L2501:
this causes some bad behaviour like having to enter search queries every time you go up back into a search result like in the twitch addon.

phil65:
might be that this should be solved inside the plugin now? @tamland ?

tamland:
Well yes obviously, disabling plugin caching will cause plugins to not cache.. If we want it in core it needs to be re-implement (in a way that it doesn't get stuck indefinitely). But preferably plugins shouldn't be so slow they need caching. If that's an answer(?):)

popcornmix:
Got a [report](http://forum.kodi.tv/showthread.php?tid=269814&pid=2465506#pid2465506) of youtube add-on being slow after this PR went in.

I know that is not unexpected and there are reasons for disabling caching, but thought I'd mention it in case anyone has ideas for improving this. I suspect there may be many similar reports.

zag2me:
Generally you would want to cache online plugins from my experience using them daily. I value speed over updates. Just my opinion...

dentaku65:
Please restore the cache for plugins; the perception is lack of performance using Kodi, at least on RPI3. Or place an option to enable the cache for plugins, if possible.

arnova:
@tamland : I agree with @dentaku65 I think this change has too many drawbacks.

DaveTBlake:
This seems to be the cause of  http://trac.kodi.tv/ticket/17211 and multiple complaints on the forum.

@tamland this change is causing problems, but I'm afraid I don't know enough to be able to solve the issues

koying:
Maybe implementing a TTL for transient cache would be an alternative solution?

arnova:
I think for v17 this should just be reverted. As that seems the safest solution for now.

tamland:
The fix is to remove the permanent/global caching in GUIMediaWindow, and instead cache until you go back or leave the window. That should solve the navigation sluggishness. After that we should really think about whether we should provide caching for plugins, because what we have now at least is completely broken.

@DaveTBlake if so, that would be really weir (and unrelated) as this pr just flips the cache policy for plugins.

koying:
@tamland From the ticket, "If the playlist is populated by web query" + "Install "radio" music addon", so that seems addon related indeed.

dentaku65:
Installed Krypton RC3 today via LibreElec... still having navigation slowness inside the add-ons. Using this on every day basis is a nightmare.

MartijnKaijser:
@dentaku65 this is not in RC3. Also use the forum

dentaku65:
@MartijnKaijser ah ok, sorry

PR #10395, opened 2016-09-04
fix skin variable parsing: dont skip if element text is empty

tamland:


BigNoid:
+1 I expect it to stop if the condition is met, no matter if the value is empty. Tested and works as intended.

ronie:
great, thx!

tamland:
jenkins build and merge

MartijnKaijser:
jenkins build this please

phil65:
Late thx, this was the reason I think why I put "-" in there instead.

PR #10396, opened 2016-09-04
[cmake/win32] Bump minimum CMake version to 3.6

fetzerch:
There seem to be some issues with older CMake versions on windows (< 3.4).
Therefore let's bump the minimum required version to 3.6 which introduced the VS_STARTUP_PROJECT var. We use that to allow users to directly build and run Kodi with F5. (older versions would have the ALL_BUILD target preselected which builds everything including tests and doesn't start kodi).

On windows this bump should be non-intrusive, because devs have to download and install CMake anyways manually, so let's save us some troubles.

ping: @Paxxi

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,7 +19,6 @@ init:
 install:
   - SET PATH=C:\Program Files (x86)\CMake\bin;C:\msys64\bin;C:\msys64\usr\bin;%PATH%
   - bash -lc "pacman --needed --noconfirm -Sy"
-  - bash -lc "pacman --needed --noconfirm -S curl"
   - mklink /j %APPVEYOR_BUILD_FOLDER%\project\BuildDependencies\msys64 C:\msys64
 
 build_script:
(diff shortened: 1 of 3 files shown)


Paxxi:
Fully agree, there's no real reason to have an older version intalled on windows and it's a fairly important feature for usability

fetzerch:
AppVeyor still has 3.5.1 installed. Seems like there's no way to manually update that via appveyor.yml, so it's have to wait until they have updated. Ticket: https://github.com/appveyor/ci/issues/1001

MartijnKaijser:
Ping @fetzerch

fetzerch:
Build fails because the mingw libs (ffmpeg, libdvd) can't be found. @Paxxi any idea?
I think the error is unrelated to this PR, other appveyor builds fail as well. Maybe we should get this one in anyways as it fixes appveyor log output.

MartijnKaijser:
jenkins build this please

Paxxi:
Appveyor is a bit special as we use pre-built ffmpeg and libdvd afaik to save time, not sure if they end up in the correct place

fetzerch:
@Paxxi: I've temporarily added a few "dir" calls yesterday. I think it fails even to download ffmpeg/libdvdnav - even though the url works locally. Does it makes sense to try adding the files to our mirror?

fetzerch:
Appveyor build fixed, merging.

PR #10397, opened 2016-09-04
[cmake] Optimizations for ARM

fetzerch:
A couple of (mostly) ARM related fixes. Kudos to @mapfau (most of the things popped up in his odroid work).

See commit messages for details.

Ping: @wsnipex, @hudokkow 

@mapfau: mind testing this again?


project/cmake/scripts/linux/ArchSetup.cmake
```diff
@@ -11,6 +11,11 @@ else()
   elseif(CPU MATCHES "i.86")
     set(ARCH i486-linux)
     add_options(CXX ALL_BUILDS "-msse")
+  elseif(CPU STREQUAL arm)
```
wsnipex: MATCHES, cpu will hardly ever be arm exactly.

project/cmake/scripts/linux/ArchSetup.cmake
```diff
@@ -11,6 +11,11 @@ else()
...
+    set(NEON True)
+    set(ARCH arm)
+  elseif(CPU STREQUAL aarch64)
```
wsnipex: please also check for "arm64" here

project/cmake/scripts/rbpi/ArchSetup.cmake
```diff
@@ -18,6 +18,7 @@ else()
     set(ARCH arm-linux-gnueabihf)
```
wsnipex: NEON must always be off for rpi1

project/cmake/scripts/rbpi/ArchSetup.cmake
```diff
@@ -18,6 +18,7 @@ else()
...
   elseif(CPU MATCHES "cortex-a7")
     set(ARCH arm-linux-gnueabihf)
+    set(NEON True)
   else()
```
wsnipex: can you also add RPI3? should be cortex-a53, correct @popcornmix?

wsnipex:
Nice.
while you are at it, can you remove the dependency on a Toolchain file for Rbpi? There are users out there that build directly on the rpi.

project/cmake/scripts/rbpi/ArchSetup.cmake
fetzerch: NEON is not set to true, so the option will default to off. do you want me to add an additional check in the common ArchSetup.cmake file? Not sure if that's necessary, cause if someone turns that open on, he likely knows what he's doing.

fetzerch:
thx, done

project/cmake/scripts/rbpi/ArchSetup.cmake
wsnipex: Never underestimate some users ability to turn every possible option on ;)
I'd simply force it off here, to be on the save side

fetzerch:
jenkins build this please

fetzerch:
- Disabled neon also for x86 and x86_64 (linux and osx)
- AML only for android and linux (was incorrectly set to optional dep on osx, or ios)
- Fixed a missing include for atoi on android, not sure if that is caused by NEON flags

jenkins build this please

project/cmake/scripts/linux/ArchSetup.cmake
```diff
@@ -8,9 +8,17 @@ if(WITH_ARCH)
...
   elseif(CPU MATCHES "i.86")
     set(ARCH i486-linux)
+    set(NEON False)
     add_options(CXX ALL_BUILDS "-msse")
+  elseif(CPU MATCHES arm)
+    set(ARCH arm)
+    set(NEON True)
+  elseif(CPU MATCHES aarch64 OR CPU MATCHES arm64)
+    set(ARCH aarch64)
+    set(NEON False)
```
wsnipex: should be True for aarch64

stefansaraev: nope. neon is always present on arch64 and -mfpu=neon is not valid gcc opt.

fetzerch:
- Option overrides in platform ArchSetup.cmake files must be cached vars since they are now set before the option is defined. This is due to the moved includes in the first commit.
- iOS/AppleClang doesn't support neon flags

jenkins build this please

fetzerch:
@wsnipex good to go?

PR #10400, opened 2016-09-04
VideoPlayer: fix lateframes if fps does not equal refresh rate

FernetMenta:
fix for http://trac.kodi.tv/ticket/16872

@HitcherUK could you test this please?

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp
@@ -1303,7 +1303,7 @@ void CRenderManager::PrepareNextRender()
       m_QueueSkip++;
     }
 
-    int lateframes = (renderPts - m_Queue[idx].pts) / frametime;
+    int lateframes = (renderPts - m_Queue[idx].pts) * m_fps / DVD_TIME_BASE;
     if (lateframes)
       m_lateframes += lateframes;
     else


HitcherUK:
@MilhouseVH If you have the time could you include this in a LE  x86_64 build please?

Thanks both.

MilhouseVH:
Yep

axbmcuser:
I included the possible fix https://github.com/xbmc/xbmc/pull/10400/commits/5bf2202cf4a680cf86f7a47b119300826c84bed0 in my build - 25fps stutter issues seem to remain for me. More info/debug log on http://trac.kodi.tv/ticket/16872

FernetMenta:
could you provide a sample file?

axbmcuser:
Since it looks like can't even nearly reproduce it as easily as HitcherUK (although i very likely got my hands on the same 25fps file he used), i unfortunately can't right now. HitcherUK apparently used the TV Show Suspects s05e05 (720p - should be only one possible file/rls with this specs).

Findings from my end so far:

Without fix https://github.com/xbmc/xbmc/pull/10400/commits/5bf2202cf4a680cf86f7a47b119300826c84bed0:
- 25fps (and only this fps) content from Amazon Prime VOD stream sometimes stutters after long time of viewing. Often at least 1 episode in a playlist-row plays without any issues. If the problem occurs, the drop-frames-stutter is persisent until you try to "fix" it by skipping/seeking. Just to be safe: Servers and internet connection are fine. So are every other fps files from this online source. It happens only with such rare 25fps content (which is typically played at 50fps refresh rate).
- I can't really reproduce the issue with my even more rare local/smb 25fps content, it seems. I don't have the problem after a period of 4 minutes like HitcherUK reported.

Including fix https://github.com/xbmc/xbmc/pull/10400/commits/5bf2202cf4a680cf86f7a47b119300826c84bed0:
- Local content (like HitcherUK s file) + Amazon Prime VOD 25fps stream content stutter completely from the beginning of both test sources.

Bottom line: Looks like the possible fix https://github.com/xbmc/xbmc/pull/10400/commits/5bf2202cf4a680cf86f7a47b119300826c84bed0 makes things worse in my case. I barely had the issue before and as could fix it easily for for some time with a simple seek. Not nearly as bad as HitcherUK having the issue after every first 4 minutes of a file. (?)

Happy to test more as soon as i have more time - maybe HitcherUK can give some additional feedback in the meantime.

I'll provide an additional log without the fix https://github.com/xbmc/xbmc/pull/10400/commits/5bf2202cf4a680cf86f7a47b119300826c84bed0 since my log on tract already was with the fix included.

xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp
```diff
@@ -1303,7 +1303,7 @@ void CRenderManager::PrepareNextRender()
       m_QueueSkip++;
     }
 
-    int lateframes = (renderPts - m_Queue[idx].pts) / frametime;
+    int lateframes = (renderPts - m_Queue[idx].pts) * (m_fps * DVD_TIME_BASE);
```
fritsch: In fact as we discussed on IRC:

A frame is: (1 / m_fps) * DVD_TIME_BASE in length (pts)
So here we need: (r - i) / ((1 / m_fps) * DVD_TIME_BASE) == (r -i) / (DVD_TIME_BASE / m_fps) == (r - i) * m_fps / DVD_TIME_BASE

FernetMenta:
updated. @axbmcuser mind testing again?

axbmcuser:
Had the chance to make a first quick+small test:
Started same 25fps tv show episode on Amazon VOD as i always use for testing - instantly skip 10 minutes forward after playback begins - let the video play from there without any visible GUI. (This was the best way for me to reproduce the (in my case) "medium-rare" issue on purpose.

**Result without fix:** Test case almost always causes stutter until you fix it with another easy skip/seek
**Result including the just updated fix:** No stutter present with test case. Dropped frame near 0 (only the typical 2-4 from the actual +10min skip/seek event, which should be completely normal)

I did not have the time to make any longer real life testing - i'll will continue to use the build and report further asap.

HitcherUK:
Played the same video for 25mins with no drops whatsoever.

Thanks.

PR #10403, opened 2016-09-05
fix SSE usage

wsnipex:
this is needed on platforms where gcc disables SSE by default, mainly i386


wsnipex:
jenkins build this please

PR #10404, opened 2016-09-05
[PVR] Add timer/timer rule: display error box in case pvr addon does …

ksooo:
…not support a suitable timer type.

Problem reported here: http://forum.kodi.tv/showthread.php?tid=261847&pid=2407997#pid2407997 => "when I open the EPG and select "add timer" from the context menu I don't get any response at all."

The root cause for this particular problem is a small bug in pvr.vdr.vnsi (https://github.com/kodi-pvr/pvr.vdr.vnsi/pull/89) , but not every addon must support all kind of timer types, thus the "error" may occure even if pvr addon is not buggy, thus user should be informed.

@Jalle19 no-brainer. Any objections?

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -9361,7 +9361,19 @@ msgctxt "#19093"
 msgid "Instant recording: %s"
 msgstr ""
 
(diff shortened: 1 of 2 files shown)


Jalle19:
+1

jenkins build this please

PR #10406, opened 2016-09-05
[gui] use CVariant on SetVisualisation(...)

AlwinEsch:
This is a part what was needed on my add-on system rework to prevent compile error, is independent and make sense for me to have.

### DIFF
--- a/xbmc/music/dialogs/GUIDialogVisualisationPresetList.cpp
+++ b/xbmc/music/dialogs/GUIDialogVisualisationPresetList.cpp
@@ -26,6 +26,7 @@
 #include "input/Key.h"
 #include "guilib/LocalizeStrings.h"
 #include "utils/StringUtils.h"
+#include "utils/Variant.h"
 
 using ADDON::CVisualisation;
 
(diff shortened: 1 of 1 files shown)


xbmc/music/dialogs/GUIDialogVisualisationPresetList.cpp
```diff
@@ -61,7 +62,8 @@ void CGUIDialogVisualisationPresetList::SetVisualisation(CVisualisation* vis)
   {
     SetUseDetails(false);
     SetMultiSelection(false);
-    SetHeading(StringUtils::Format(g_localizeStrings.Get(13407).c_str(), m_viz->Name().c_str()));
+    std::string heading = StringUtils::Format(g_localizeStrings.Get(13407).c_str(), m_viz->Name().c_str());
+    SetHeading(CVariant{heading});
```
Paxxi: Shouldn't be a need for this temporary

Paxxi:
:thumbs_up: one less missing include is always a good thing

xbmc/music/dialogs/GUIDialogVisualisationPresetList.cpp
AlwinEsch: Was a c&p from the other places where it is used, thought is wanted this way ;-)

AlwinEsch:
jenkins build this please

xbmc/music/dialogs/GUIDialogVisualisationPresetList.cpp
Paxxi: Both are fine really, when doing the original I mainly chose for readability so there's probably a mix.

AlwinEsch:
Ok to merge in?

PR #10407, opened 2016-09-05
[addons] add read of api level from addon.xml

AlwinEsch:
This is the part from my addon system rework where the used API level becomes read on add-ons XML file.

Is currently not direct required, but good for me to do a review and may be to bring in to have all binary add-ons already with the value set for future usage.

``` xml
<addon
  id="audiodecoder.dumb"
  version="2.0.0"
  name="Modplug Audio Decoder"
  provider-name="spiff">
  <requires>
    <import addon="kodi.audiodecoder" version="1.0.0"/>
  </requires>
  <extension
    point="kodi.audiodecoder"
    apilevel="1" <!-- See here the set of used level -->
    name="dumb"
    extension=".it|.xm|.s3m"
    library_@PLATFORM@="@LIBRARY_FILENAME@"/>
  <extension point="xbmc.addon.metadata">
    <summary lang="en">DUMB Audio Decoder</summary>
    <description lang="en">DUMB Audio Decoder</description>
    <platform>@PLATFORM@</platform>
  </extension>
</addon>
```

### DIFF
--- a/addons/xbmc.addon/metadata.xsd
+++ b/addons/xbmc.addon/metadata.xsd
@@ -20,6 +20,7 @@
       <xs:attribute name="point" type="xs:string" use="required"/>
       <xs:attribute name="id" type="simpleIdentifier"/>
       <xs:attribute name="name" type="xs:string"/>
+      <xs:attribute name="apilevel" type="xs:string"/>
     </xs:complexType>
   </xs:element>
   <xs:simpleType name="simpleIdentifier">
--- a/xbmc/addons/Addon.h
+++ b/xbmc/addons/Addon.h
@@ -56,11 +56,12 @@ void OnPostUnInstall(const AddonPtr& addon);
 class AddonProps
 {
 public:
-  AddonProps() : type(ADDON_UNKNOWN), packageSize(0) {};
-  AddonProps(std::string id, TYPE type) : id(std::move(id)), type(type), packageSize(0) {}
+  AddonProps() : type(ADDON_UNKNOWN), apiLevel(-1), packageSize(0) {};
+  AddonProps(std::string id, TYPE type) : id(std::move(id)), type(type), apiLevel(-1), packageSize(0) {}
 
   std::string id;
   TYPE type;
+  int apiLevel;
   AddonVersion version{"0.0.0"};
   AddonVersion minversion{"0.0.0"};
   std::string name;
@@ -99,6 +100,7 @@ public:
   std::string ID() const override{ return m_props.id; }
   std::string Name() const override { return m_props.name; }
   bool IsInUse() const override{ return false; };
+  int APILevel() const override { return m_props.apiLevel; }
   AddonVersion Version() const override { return m_props.version; }
   AddonVersion MinVersion() const override { return m_props.minversion; }
   std::string Summary() const override { return m_props.summary; }
--- a/xbmc/addons/AddonBuilder.h
+++ b/xbmc/addons/AddonBuilder.h
@@ -45,6 +45,7 @@ public:
   void SetBroken(std::string broken) { m_props.broken = std::move(broken); }
   void SetPath(std::string path) { m_props.path = std::move(path); }
   void SetLibName(std::string libname) { m_props.libname = std::move(libname); }
+  void SetAPILevel(int apiLevel) { m_props.apiLevel = apiLevel; }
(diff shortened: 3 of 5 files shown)


tamland:
We already have dependencies with min/max version. Why do you need *another* versioning system?
Either way, binary addon specifics should be added to binary addon interfaces, not IAddon/AddonProps.

AlwinEsch:
See here the coming structure: https://github.com/AlwinEsch/kodi/tree/binary-addon-rework-complete-2/xbmc/addons/kodi-addon-dev-kit/include/kodi/api2

Is to define later the used include for add-on headers and to know the `<kodi/api2/...>`, this becomes on end read from `cmake` there and macro on add-on define the right include.

The Macro way already visible here: https://github.com/AlwinEsch/kodi/blob/binary-addon-rework-complete-2/xbmc/addons/kodi-addon-dev-kit/src/api2/kodi-addon-sharedlibrary/AddonLib.cpp#L22

The `min/max` is on end no more needed and can be removed.

MartijnKaijser:
@AlwinEsch the min/max is the default for every add-on we have so why are you inventing something else?

AlwinEsch:
You can leave a include and handling of a e.g. `#include <kodi/api2/...>` and have also a `#include <kodi/api3/...>` present and with them on xml becomes the used folder and level defined.

To use a complete version string like a `2.1.3` for a folder name make no sense for me and brings to much on changes.

My planned way (who I also want to show a bit on DevCon) is to have a stable API Level e.g. `2` and during development time a unstable e.g. `3`. All add-ons with `2` are still usable without changes and on next release becomes this `3` stable and all add-ons can be swapped to them.

After a release level `4` comes, goes the `2`complete away. So you have always for add-on updates a minimum time for updates of one release to another.

Also is this level used for everything of add-on type, so that the version of e.g. PVR is no more direct usable this way.

MartijnKaijser:
Sorry but that makes absolutely no sense at all. Because you change something in PVR or whatever code and then bump the API every other addon like audio decoder/encoder must also be updated or whatever. Why do you want to force changes on addons each time that haven't changed in past year(s).  Currently if you break PVR add-on API only those are broken and the rest stays as is. With you new "system" you break everything.

AlwinEsch:
Problematic becomes the very big increase of add-on functionality on `libXBMC_addon.h` (who is used on much add-ons) and `libKODI_guilib.h`. 

If you have a change on `libXBMC_addon.h` then need all add-ons a update. That's also why I want to standardize everything to `C++` so that on changes on add-on itself nearly nothing need to do.

Currently must on a e.g. PVR Version change all PVR add-ons asap be updated, to have support of them after changes, I want to have a way to have a bit more time of them.

The interface supports on end the same as the way used with Python add-ons.

Also becomes the add-on load way changed, currently is a load of binary add-on always fixed to something like screensaver, PVR, ADSP... but never alone as independent add-on possible. So that it becomes possible to create a `C++` add-on to do something like currently with Python add-ons.

AlwinEsch:
Currently is the way on https://github.com/xbmc/xbmc/tree/master/xbmc/addons/kodi-addon-dev-kit/include/kodi not the way I would imagine as an "include" add-ons for strangers.

This here is a example (not final) how I thing it is more easy for add-on development: https://raw.githubusercontent.com/AlwinEsch/kodi/binary-addon-rework-complete/xbmc/addons/kodi-addon-dev-kit/include/kodi/api2/pvr/Addon.hpp

There you have one parent class for your add-on and you need only the parts to add in your child who needed and bring something like this:
https://github.com/AlwinEsch/pvr.demo/tree/change-to-api3/src

But to have them for external developers more easy must we find a way to support different Versions and we hold now 13 different types of them (**and there comes a lot more in future**).
- add-on
- gui
- pvr
- adsp
- screensaver
- visualization
- audioencoder
- audiodecoder
- inputstream
- peripheral
- audioengine
- codec

Few of them can be removed (e.g. codec) or passed (e.g. adsp to audioengine)  to others.
To allow there a backward compatibility with older version becomes it a big mess.

Paxxi:
Easier way would be using an API level define before the addon includes anything from us, depending on its version we drag in things from v1 , v2 etc into the kodiaddon namespace

AlwinEsch:
I leave it on the old without the API, but I want todo is a rework to have to have the interface more clear

PR #10408, opened 2016-09-05
[pvr] bye, bye libXBMC_pvr.cpp

AlwinEsch:
Hi all,

my big add-on rework takes to much time and that's why I bring this smaller one. This remove the usage of shared pvr library on add-on's and the communication is done direct on header with the function table from Kodi (What was before on the cpp of lib). 

If this is OK and works can also this way be done for the other libraries.


AlwinEsch:
jenkins build this please

FernetMenta:
great! what are your plans for integrations? You don't plan this for v17, do you?

AlwinEsch:
If acceptable for you and the others we can do it for v17, was not sure if OK on the Beta.

Normally are the old who use the library (if lib present) are also working, nothing changed on the function table itself.

AlwinEsch:
If all are swapped to the new way can be the `..._RegisterMe` and `..._UnRegisterMe` removed, but then all add-ons need a Version change to prevent compatibility faults.

MartijnKaijser:
Then that's a no.

ksooo:
@AlwinEsch sorry, API changes for v17 are a no go as we are already in beta.

AlwinEsch:
@ksooo no problem thought the same, and sorry that it come so late.

ksooo:
@AlwinEsch no problem, you are doing a great job. Keep it up.

AlwinEsch:
OK to have on end all library parts in one request or better to have every part in a own pull request?

FernetMenta:
I think separate PRs are easier to review

xbmc/addons/kodi-addon-dev-kit/include/kodi/libXBMC_pvr.h
```diff
@@ -27,29 +27,78 @@
...
-
 #define DVD_TIME_BASE 1000000
 
 //! @todo original definition is in DVDClock.h
 #define DVD_NOPTS_VALUE 0xFFF0000000000000
 
+namespace V1
+{
+namespace KodiAPI
+{
```
Paxxi: I think KodiAPI should be the outer namespace as it won't change and have the version as the second level

Paxxi:
Nice work, seems like a sensible way to handle it

AlwinEsch:
About the:
`
If all are swapped to the new way can be the ..._RegisterMe and ..._UnRegisterMe removed, but then all add-ons need a Version change to prevent compatibility faults.
`
above. Is not needed.

All add-ons need only recompile and nothing else :)

Acceptable to bring this and the changes for other libraries in?

AlwinEsch:
jenkins build addons please

AlwinEsch:
jenkins build this please

fritsch:
jenkins build this please

AlwinEsch:
jenkins build this please

AlwinEsch:
OK to bring in?

jenkins build this please

AlwinEsch:

PR #10409, opened 2016-09-06
[addons] move ./xbmc/addons/binary/interfaces one down and remove binary folder

AlwinEsch:
Remove not needed binary folder


AlwinEsch:
jenkins build this please

stefansaraev:
not sure how jenkins build passed, but seems you forgot to update makefiles: http://sprunge.us/GSdB

popcornmix:
This breaks building for me. The files:
```
xbmc/addons/interfaces/api1/AudioEngine/Makefile
xbmc/addons/interfaces/api1/AudioDSP/Makefile
xbmc/addons/interfaces/api1/Addon/Makefile
xbmc/addons/interfaces/api1/Codec/Makefile
xbmc/addons/interfaces/api1/PVR/Makefile
xbmc/addons/interfaces/api1/InputStream/Makefile
xbmc/addons/interfaces/api1/GUI/Makefile
xbmc/addons/interfaces/api1/Peripheral/Makefile
```
all include ``include ../../../../../../Makefile.include`` which needs to have one ``../`` removed after this PR.

ksooo:
@AlwinEsch please, only fixes at this stage. As we can see here, even a non-functional change can cause trouble. IMHO, it is important that we're stabilizing or code base in this phase where we nearing a relaese. Best approach still is not to commit stuff that is not really needed.

MartijnKaijser:
That's because we only use cmake with jenkins

FernetMenta:
why don't we trash autotools now?

hudokkow:
Started work this morning just to see the extent of work needed: https://github.com/xbmc/xbmc/compare/master...hudokkow:goodbye_old_friend?expand=1

Not that I don't trust CMake but it's L*** stuff IMO.

FernetMenta:
if all builds on Jenkins use cmake, we have to force developers to use cmake too.

stefansaraev:
it is not yet ready to be forced on distributions. here my thoughts:

1. it is not yet possible to cross compile kodi using cmake, without using internal depends. distros like openelec/libreelec/buildroot and more need that. they have own cross-toolchain with all target dependencies already compiled. they also want to use own cmake toolchain files and not add kodi specific stuff (NATIVEPREFIX, DEPENDS_PATH and so on) 
2. it should be possible to disable ANY optional dependency, as it is with autoconf now (yes, smbclient on non-win too!)
3. there should be a separate (native) build step / projcet for native tools such as texturepacker and jsonschemabuilder.
4. cmake-3.1 requirment is too high imo, at least debian jessie (ships with cmake 3.0.2) should be suppoted. ubuntu 14.04 lts (EOL april 2019!) even ships with 2.8.x :(

in short, our cmake buildsystem is not (yet) friendly to distro packagers, as it shoule be.

sure, once master is open for v18 development, those issues can be solved and autoconf completely removed. but autoconf should stay for v17

do you want me to PR the build fix or you will push it (there is a paste above, you can pick it as-is) ?

samnazarko:
>cmake-3.1 requirment is too high imo,

+1.  I reverted this change for OSMC Krypton and haven't noticed any issues so far using cmake for binary add-ons. The original commit that bumps the version seems to only do so to fixup Windows build issues. With cmake on Jessie, all seems fine. 

Please do not remove autotools during the beta cycle. It's too late, and OSMC is just preparing for Krypton.

wsnipex:
there is no way we remove autotools for v17.

@stefansaraev points 1 to 3 are already possible/true. And I don't agree with 4, cmake 3.1 is pretty old.

AlwinEsch:
Ah, sorry to all, has seen it not as complicated and with all test builds OK thought to bring it in.

PR #10411, opened 2016-09-06
[GUIFont] - make BufferHandle explicit GLuint, use initializer consistently

guidosarducci:
Recent commits fixed potential pointer/integer conversion problems using GLES, but didn't use GL types to ensure portability and avoid future integer/integer issues. This also expands use of defined initializers.

### DIFF
--- a/xbmc/guilib/GUIFontCache.h
+++ b/xbmc/guilib/GUIFontCache.h
@@ -39,6 +39,10 @@
 #include "TransformMatrix.h"
 #include "system.h"
 
+#if defined(HAS_GL) || defined(HAS_GLES)
+#include "system_gl.h"
+#endif
+
 #define FONT_CACHE_TIME_LIMIT (1000)
 #define FONT_CACHE_DIST_LIMIT (0.01f)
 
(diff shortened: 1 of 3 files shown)


xbmc/guilib/GUIFontCache.h
```diff
@@ -39,6 +39,10 @@
 #include "TransformMatrix.h"
 #include "system.h"
 
+#if defined(HAS_GL) || defined(HAS_GLES)
+#include "system_gl.h"
```
FernetMenta: including this header here is not a good design

FernetMenta:
does this fix anything?

xbmc/guilib/GUIFontCache.h
guidosarducci: It's arguably a bigger design issue to take the implementation-independent GUIFontCache.h and introduce GLES and DX specific preprocessing, but that's exactly what was done in the previous commits: https://github.com/xbmc/xbmc/commit/fa6f5d761b600b0a8b52db18168e540b03f79ceb and https://github.com/xbmc/xbmc/commit/e9dd97bac1dc97f9fbdba2b81402aed2c9b92713.

The potential issue here is in assuming the `bufferHandle` is always `unsigned int` instead of correctly using the GL opaque types, and to fix that requires pulling in the right header.  Before these commits, when `bufferHandle` was a `void*` with casting left and right, I spent hours tracking down a GL crash that could have been easily avoided by directly using the GL type as intended.

So yes, I also had qualms about including the header and breaking implementation independence, but that ship sailed two commits ago, and this the safer approach.

What are your suggestions?

Also pinging @Memphiz who's made previous changes to have a look.

FernetMenta: GLuint is defined as unsigned int. I don't see this this PR changes anything in behaviour.

Memphiz:
no opinion on that - my changes where a backport from @MrMC

MrMC:
if you are going to use it as a GLuint, define it as a GLuint and skip the casting madness. Most problematic on arm or other embedded.

MartijnKaijser:
@guidosarducci ping?

Razzeee:
@guidosarducci 
I'm closing this as the PR seems to be stale, if you want to revisit it, just let us know and we will reopen it.
Thanks!

PR #10412, opened 2016-09-06
VideoPlayer: avoid null render if renderer has no frame or is not con…

FernetMenta:
…figured


FernetMenta:
jenkins build this please

PR #10415, opened 2016-09-06
[Estuary] fix texture paths

ronie:
fixes the 'could not find texture' errors reported by @MilhouseVH


MilhouseVH:
No more errors - thanks for the quick fix!

MilhouseVH:
Do you know where the following is coming from?
```
20:59:21 104.878525 T:1961242624   ERROR: CGUITextureManager::GetTexturePath: could not find texture 'dialogs/volume/progress/p0.png'
```

stefansaraev:
@MilhouseVH https://github.com/xbmc/xbmc/blob/master/addons/skin.estuary/1080i/DialogVolumeBar.xml#L36

MilhouseVH:
@stefansaraev it's not actually the volume indicator - if I reduce the volume to zero I get the muted image, and there is no error logged. The error being logged is actually from the "Buffering" overlay that appears when watching a video - at the beginning this is at 0% buffered and it obviously tries to load a non-existent 0% image.

stefansaraev:
sure, same thing. :)
https://github.com/xbmc/xbmc/blob/master/addons/skin.estuary/1080i/VideoFullScreen.xml#L32

MilhouseVH:
Snap. :)

I guess a blank p0.png image would be one way to solve it...

ronie:
yup, added.

PR #10416, opened 2016-09-06
[cmake] Fix globbing of single files on export

fetzerch:
GLOB_RECURSE when called with a file name will list all files with the given name in all subdirectories. This is unintuitive for files specified in installdata. Here if a full file name is specified only that single file should be mirrored/installed. Since we mirror them to the same directory, files are overridden and the last file wins.

Fixes installation of certificates as mentioned in https://github.com/xbmc/xbmc/pull/10108#issuecomment-244591110
Ping: @Memphiz, @wsnipex, @hudokkow

### DIFF
--- a/project/cmake/scripts/common/Macros.cmake
+++ b/project/cmake/scripts/common/Macros.cmake
@@ -259,7 +259,15 @@ function(copy_files_from_filelist_to_buildtree pattern)
         else()
           list(GET dir -1 dest)
         endif()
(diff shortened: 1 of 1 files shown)


fetzerch:
jenkins build this please

Memphiz:
Looks good to me - thx for the quick fix :)

wsnipex:
looks good

PR #10417, opened 2016-09-06
[droid] Drop RK specific code and fix cmake AML building

MartijnKaijser:
We only support official Android API so RK specific can be removed
Second commits only uses AMLutils and EGLNativeTypeAmlAndroid when it's actually enabled.


MartijnKaijser:
Jenkins build this please

MartijnKaijser:
@fritsch @fernetmenta any comments? 
@seo still good on AML linux? 
jenkins build this please

fritsch:
Cleanup is always good, but keep in mind: Android's API be it Audio or Video is broken and not working well on < v23 firmware, so everything that forces older boxes to standards might kill them.

That being said, currently no device besides the shield does implement even the audio API correctly, while also having a broken firmware since 3.2 release.

No strong feelings here. The define-hell in the EGLWrapper should go completely though.

xbmc/windowing/egl/EGLWrapper.cpp
```diff
@@ -88,16 +91,15 @@ bool CEGLWrapper::Initialize(const std::string &implementation)
...
+#if defined(TARGET_ANDROID) && defined(HAS_LIBAMCODEC)
+      (nativeGuess = CreateEGLNativeType<CEGLNativeTypeAmlAndroid>(implementation))
+#elif defined(TARGET_ANDROID)
+      (nativeGuess = CreateEGLNativeType<CEGLNativeTypeAndroid>(implementation))
+#elif defined(TARGET_RASPBERRY_PI)
       (nativeGuess = CreateEGLNativeType<CEGLNativeTypeRaspberryPI>(implementation))
 #elif defined(HAS_IMXVPU)
       (nativeGuess = CreateEGLNativeType<CEGLNativeTypeIMX>(implementation))
-#else
+#elif defined(HAS_LIBAMCODEC)
```
wsnipex: maybe defined(TARGET_LINUX) && defined(HAS_LIBAMCODEC)

FernetMenta:
Ok because it is a fix. Next time SystemInfo is touched it should be moved to platform and use platform overrides instead of ifdeffery.

MartijnKaijser:
PR updated 
Jenkins build this please

MartijnKaijser:
Confirmed working on android and amlogic linux

PR #10418, opened 2016-09-06
[cmake/android] Fix startup without AML

fetzerch:
Attempt to fix startup due to missing AML related symbols on android.
Does what autotools does, might be a good idea to check if these aml related files are really needed if aml is disabled.

@MartijnKaijser My local build failed to install (complained that it was too big). Let's try the jenkins one.

--> jenkins.kodi.tv/job/Android-ARM/9125

### DIFF
--- a/xbmc/utils/CMakeLists.txt
+++ b/xbmc/utils/CMakeLists.txt
@@ -177,11 +177,14 @@ if(NOT CORE_SYSTEM_NAME STREQUAL windows)
                       GLUtils.h)
 endif()
 
+if(OPENGLES_FOUND)
+  list(APPEND SOURCES AMLUtils.cpp)
+  list(APPEND HEADERS AMLUtils.h)
+endif()
+
 if(AML_FOUND)
-  list(APPEND SOURCES AMLUtils.cpp
-                      ScreenshotAML.cpp)
-  list(APPEND HEADERS AMLUtils.h
-                      ScreenshotAML.h)
+  list(APPEND SOURCES ScreenshotAML.cpp)
+  list(APPEND HEADERS ScreenshotAML.h)
 endif()
 
 core_add_library(utils)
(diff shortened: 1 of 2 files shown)


MartijnKaijser:
It starts however you shouldn't make it look like autotools but but fix autotools.

MartijnKaijser:
Looks like amutils is used inconsistent with opengles or has_amlcodec in different places

stefansaraev:
amlutils is used unconditionaly on android 

https://github.com/xbmc/xbmc/blob/master/xbmc/utils/SystemInfo.cpp#L57-L60
https://github.com/xbmc/xbmc/blob/master/xbmc/utils/SystemInfo.cpp#L870-L873
https://github.com/xbmc/xbmc/blob/master/xbmc/rendering/gles/RenderSystemGLES.cpp#L660

no idea if this special aml case is still needed nowadays..

xbmc/utils/CMakeLists.txt
```diff
@@ -177,11 +177,14 @@ if(NOT CORE_SYSTEM_NAME STREQUAL windows)
                       GLUtils.h)
 endif()
 
+if(OPENGLES_FOUND)
+  list(APPEND SOURCES AMLUtils.cpp)
+  list(APPEND HEADERS AMLUtils.h)
+endif()
+
 if(AML_FOUND)
```
wsnipex: ENABLE_AML

wsnipex:
as pointed out above, AML is a mess. Looks like its currently not possible to properly disable it.

xbmc/utils/CMakeLists.txt
fetzerch: Don't get it. We only look for AML if ENABLE_AML is set. But if you check with ENABLE_AML here we compile the files even if AML was enabled but not found. Is that what we want?

wsnipex: you're right, was a brain fart. Sorry

MartijnKaijser:
Alternative approach 
https://github.com/xbmc/xbmc/pull/10417/commits/71a73cc04ede13f6fdbc7be5d8f6c3631b1fb6cf

fetzerch:
I'm all for fixing it properly, just thought you wanted a quick fix yesterday evening to get master working again. Closing this in favor of https://github.com/xbmc/xbmc/pull/10417.

PR #10421, opened 2016-09-07
[cmake] linux: install cmake files in arch independent dir

wsnipex:
fixes KodiConfig.cmake not being found when building on a different cpu architecture
e.g. on the PPA, when building binary addons on i386, while kodi-addon-dev was created on amd64

@fetzerch @hudokkow ping


wsnipex:

hudokkow:
Looks good!

akva2:
No arch paths in that file sounds like something is wrong.

wsnipex:
@akva2 Sorry, I don't follow. Can you elaborate please?

akva2:
Typically /usr/lib/<arch> end up in such config files. But it was a knee jerk, thinking a bit further i guess it might very well be fine here since there are no binaries involves.

wsnipex:
yes, we had /usr/lib there, which became /usr/lib/x86_64-linux-gnu/kodi after multi-arch went in, because on the PPA kodi-addon-dev is built by amd64.
If you now build addons on i386 against those headers, which include KodiConfig.cmake, cmake doesn't find it, because it looks in /usr/lib/i686-linux-gnu/kodi.
Furthermore, cmake files are arch independent and therefore should go to /usr/share out of principle.

akva2:
Cmake files in general yes, but config files are comparabke to .pc files. You write explicit paths to linked libs and such and that makes them arch dependent. But there are no libs on KodiCondig so this pr is fine.

PR #10423, opened 2016-09-07
[cmake] xslt is optional

stefansaraev:
I just switched my OE build to cmake and realized I forgot to add this in #9863


hudokkow:
Looks good, thanks!
jenkins build this please

PR #10424, opened 2016-09-07
[cpack/deb] Force GIT_HASH update on DEB package generation

hudokkow:
@fetzerch, as per our private slack chat.

I was wondering if we could get into trouble calling core_find_git_rev() twice at different times, thus possibly generating two different APP_SCMID stamps. Instead of duplicating code into two functions, I decided to separate APP_SCMID and GIT_HASH generation based on function call sig. Old behavior is preserved if called without parameters.

Also cleaned up/modernized a bit.

### DIFF
--- a/project/cmake/scripts/common/Macros.cmake
+++ b/project/cmake/scripts/common/Macros.cmake
@@ -460,21 +460,6 @@ macro(core_add_optional_subdirs_from_filelist pattern)
   endforeach()
 endmacro()
 
-macro(today RESULT)
-  if(WIN32)
-    execute_process(COMMAND "cmd" " /C date /T" OUTPUT_VARIABLE ${RESULT})
-    string(REGEX REPLACE "(..)/(..)/..(..).*" "\\1/\\2/\\3" ${RESULT} ${${RESULT}})
-  elseif(UNIX)
-    execute_process(COMMAND date -u +%F
-                    OUTPUT_VARIABLE ${RESULT})
-    string(REGEX REPLACE "(..)/(..)/..(..).*" "\\1/\\2/\\3" ${RESULT} ${${RESULT}})
-  else()
-    message(SEND_ERROR "date not implemented")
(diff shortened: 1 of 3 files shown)


hudokkow:
jenkins build this please

project/cmake/scripts/common/Macros.cmake
```diff
@@ -509,13 +494,14 @@ endfunction()
...
-# The following variables are set:
+# The following variables may be set:
 #   APP_SCMID - git HEAD commit in the form of 'YYYYMMDD-hash'
 #               if git tree is dirty, value is set in the form of 'YYYYMMDD-hash-dirty'
 #               if no git tree is found, value is set in the form of 'YYYYMMDD-nogitfound'
 #   GIT_HASH  - git HEAD commit in the form of 'hash'. if git tree is dirty,
 #               value is set in the form of 'hash-dirty'
-function(core_find_git_rev)
+#               if no git tree is found, value is set in the form of 'nogitfound'
+function(core_find_git_rev stamp)
```
fetzerch: Thx! Just wondering if we should aligning this a bit with cmake convention? the function could take the output variable name as first parameter (as it does already, but don't restrict the names to those 2) and have an [option](https://cmake.org/cmake/help/latest/command/cmake_parse_arguments.html#command:cmake_parse_arguments) that defines if the date is prefixed?
If we ever want to store the version in a different variable for some reason, with the current function it wouldn't work.

hudokkow: Good idea! Thanks!

hudokkow:
Updated.
@stefansaraev, if you set -DGIT_VERSION=foobar it should do what you want now.
Best viewed with ?w=1 to ignore white space changes only -> https://github.com/xbmc/xbmc/pull/10424/files?w=1

hudokkow:
jenkins build and merge

PR #10434, opened 2016-09-08
fixed: remove accidentially added Makefile.in from header list

notspiff:
I assume this is a woopsie. If not, it should be in others, not HEADERS. ref http://jenkins.kodi.tv/job/Android-X86/8241/console


hudokkow:
@fetzerch, git blame points to you but it could also come from spiff's original work. Without history to confirm it's only speculation.
Either way, it looks like it slipped in so jenkins build this please.

fetzerch:
Yeah, spiff is correct. That's an whoopsi ;)

Am 08.09.2016 16:58 schrieb "Cristiano A. Silva" <notifications@github.com>:

PR #10435, opened 2016-09-08
Revert "AE: change usage of swr_set_compensation"

FernetMenta:
This reverts commit 7f4f89ba3a36454b01c80103068e338a7d65dbfa.

fix for http://trac.kodi.tv/ticket/16887

### DIFF
--- a/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEResampleFFMPEG.cpp
+++ b/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEResampleFFMPEG.cpp
@@ -58,7 +58,6 @@ bool CActiveAEResampleFFMPEG::Init(uint64_t dst_chan_layout, int dst_channels, i
   m_src_fmt = src_fmt;
   m_src_bits = src_bits;
   m_src_dither_bits = src_dither;
-  m_resampleRatio = 1.0;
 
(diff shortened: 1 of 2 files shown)


fritsch:
Please let's wait until: http://trac.kodi.tv/ticket/16887#comment:22 has verified. Perhaps that only covers another issue :-)

FernetMenta:
I have confirmed it working. You can do this too if you player the provided sample. I will merge this evening.

FernetMenta:
got confirmed this very moment, what a surprise :)

jenkins build this please

fritsch:
This breaks DTS in flac. Which is not really a breakge as I consider the dts wav encoded to flac as something totally non standard. But it seems we somehow apply some compensation even when just doing audio playback, meaning: the very moment we call sws_whatever_compensation we alter audio so that it is not "bit perfect" anymore and this results in DTS RAW issues (e.g. noise).

fritsch:
Some comments by our ffmpeg friends:

```
20:50 <@nevcairiel> fritsch: tell them to diaf rather quickly =p
21:13 <@nevcairiel> for some reason thats a recurring idiocy
21:13 <+wm4> fritsch: use the kodi funds or whatever you have to hire a hitman?
21:22 < rcombs> fritsch: I'd recommend telling the user to fuck off
(asked before I added this cites here)
```

PR #10437, opened 2016-09-08
cmake cosmetics

stefansaraev:
no need to build those static libs and link them into the target binary.


stefansaraev:
hum. OPTICAL was only unix thing (see last commit) so appveyor failed. now it complains about HAS_DVD_DRIVE redefinition (because it's already defined in xbmc/system.h for TARGET_WINDOWS)

not sure how to proceed from there..

wsnipex:
I'd fix the CMakeFile.txt for both libs instead. Just wrap stuff in if(ENABLE_OPTICAL) in there.

fetzerch:
The windows build has been switched over to CMake completely which now allows us to turn options on and off for that platform as well. So I would just remove the define from system.h.

@Paxxi: what do you think?

stefansaraev:
define removed.

fetzerch:
jenkins build this please

wsnipex:
looks good

PR #10442, opened 2016-09-09
[Estuary] bugfixes

ronie:
10 fixes for issues reported here:
http://forum.kodi.tv/showthread.php?tid=262373

and there (team only):
https://github.com/xbmc/skin.estuary/issues

will squash after review


BigNoid:
Looks good to me. We can probably get rid of the old rating flags completely with https://github.com/BigNoid/xbmc/commit/2d7c3e1cddcc91b3d5a1e3d08aa8c184a409154e

ronie:
@BigNoid but subtitles use a 0-5 rating scale...i don't think that would work
additionally, the new images are twice as wide as the old ones.

BigNoid:
> but subtitles use a 0-5 rating scale...i don't think that would work
> additionally, the new images are twice as wide as the old ones.
Yes of course, I mistakenly thought everything was already converted to 10 ratings. Sry for noise.

MartijnKaijser:
/offtopic
Maybe we should make the subtitle ratings also 0-10 for consistency?

PR #10445, opened 2016-09-09
require click action to activate slider controls

ronie:
require slider controls to be activated before you can move the slider(s).

this fixes the problem of being unable to navigate to other controls (left/right) when a slider control is focused.
for example: http://forum.kodi.tv/showthread.php?tid=262373&pid=2410300#pid2410300

ping @Montellese @HitcherUK @BigNoid @phil65

### DIFF
--- a/xbmc/dialogs/GUIDialogSlider.cpp
+++ b/xbmc/dialogs/GUIDialogSlider.cpp
@@ -81,6 +81,8 @@ void CGUIDialogSlider::SetSlider(const std::string &label, float value, float mi
   m_callbackData = callbackData;
   if (slider)
   {
+    slider->SetActive();
+    slider->KeepActive();
     slider->SetType(SLIDER_CONTROL_TYPE_FLOAT);
     slider->SetFloatRange(min, max);
     slider->SetFloatInterval(delta);
--- a/xbmc/guilib/GUISliderControl.cpp
+++ b/xbmc/guilib/GUISliderControl.cpp
@@ -59,6 +59,8 @@ CGUISliderControl::CGUISliderControl(int parentID, int controlID, float posX, fl
   m_iInfoCode = 0;
   m_dragging = false;
   m_action = NULL;
+  m_active = false;
+  m_keepactive = false;
 }
 
 CGUISliderControl::~CGUISliderControl(void)
@@ -85,7 +87,7 @@ void CGUISliderControl::Process(unsigned int currentTime, CDirtyRegionList &dirt
   dirty |= m_guiBackground.SetWidth(m_width);
   dirty |= m_guiBackground.Process(currentTime);
 
-  CGUITexture &nibLower = (m_bHasFocus && !IsDisabled() && m_currentSelector == RangeSelectorLower) ? m_guiSelectorLowerFocus : m_guiSelectorLower;
+  CGUITexture &nibLower = (m_active && m_bHasFocus && !IsDisabled() && m_currentSelector == RangeSelectorLower) ? m_guiSelectorLowerFocus : m_guiSelectorLower;
   float fScale;
   if (m_orientation == HORIZONTAL)
     fScale = m_height == 0 ? 1.0f : m_height / m_guiBackground.GetTextureHeight();
@@ -95,7 +97,7 @@ void CGUISliderControl::Process(unsigned int currentTime, CDirtyRegionList &dirt
   dirty |= ProcessSelector(nibLower, currentTime, fScale, RangeSelectorLower);
   if (m_rangeSelection)
   {
-    CGUITexture &nibUpper = (m_bHasFocus && !IsDisabled() && m_currentSelector == RangeSelectorUpper) ? m_guiSelectorUpperFocus : m_guiSelectorUpper;
+    CGUITexture &nibUpper = (m_active && m_bHasFocus && !IsDisabled() && m_currentSelector == RangeSelectorUpper) ? m_guiSelectorUpperFocus : m_guiSelectorUpper;
     if (m_orientation == HORIZONTAL)
       fScale = m_height == 0 ? 1.0f : m_height / m_guiBackground.GetTextureHeight();
     else
@@ -157,11 +159,11 @@ bool CGUISliderControl::ProcessSelector(CGUITexture &nib, unsigned int currentTi
 void CGUISliderControl::Render()
 {
   m_guiBackground.Render();
-  CGUITexture &nibLower = (m_bHasFocus && !IsDisabled() && m_currentSelector == RangeSelectorLower) ? m_guiSelectorLowerFocus : m_guiSelectorLower;
+  CGUITexture &nibLower = (m_active && m_bHasFocus && !IsDisabled() && m_currentSelector == RangeSelectorLower) ? m_guiSelectorLowerFocus : m_guiSelectorLower;
   nibLower.Render();
   if (m_rangeSelection)
   {
-    CGUITexture &nibUpper = (m_bHasFocus && !IsDisabled() && m_currentSelector == RangeSelectorUpper) ? m_guiSelectorUpperFocus : m_guiSelectorUpper;
+    CGUITexture &nibUpper = (m_active && m_bHasFocus && !IsDisabled() && m_currentSelector == RangeSelectorUpper) ? m_guiSelectorUpperFocus : m_guiSelectorUpper;
     nibUpper.Render();
   }
   CGUIControl::Render();
@@ -196,7 +198,7 @@ bool CGUISliderControl::OnAction(const CAction &action)
   switch ( action.GetID() )
   {
   case ACTION_MOVE_LEFT:
-    if (m_orientation == HORIZONTAL)
+    if (m_active && m_orientation == HORIZONTAL)
     {
       Move(-1);
       return true;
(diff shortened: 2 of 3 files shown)


HitcherUK:
Nice, thanks ronie. Will test over the weekend.

BigNoid:
Massive improvement in usability! I've only seen one minor nitpick when testing this and that's if you mapped for instance ```SubtitleDelayPlus```. Before this PR the focus nib is shown in DialogSlider and with this PR it shows the unfocused nib.

tamland:
Why feature and L***? It's a fix!

MartijnKaijser:
I'd agree. If there are no major concerns please add it to v17

phil65:
Yes, v17 please. From UX perspective this a definiteily a fix.
Thx @ronie !

ronie:
it was never broken, so can't be a fix ;-)
it' a functionality change and given the fact we're in beta, i'm not going to mess with our gui engine now.

you all should know 9 out of 10 gui engine changes we push are causing some sort of regression.

ronie:
guess we need to decide whether to apply this to the sliderex control only or to the slider control as well.

i think @AchimTuran uses a lot of slider controls in some of his adsp addons, that might be a reason to include those as well..

it would probably be possible to make an exception for the one slider in DialogSlider though.

i'm not sure if this pr is likely to have a negative affect other slider controls, one used in the VideoOSD or SeekBar, for example?

@BigNoid how can i reproduce the issue with seeking? i'm not sure what to look for?

MartijnKaijser:
Let's ask @milhouse to include it for while as that will give it a broader test base

BigNoid:
> how can i reproduce the issue with seeking? i'm not sure what to look for?

Testing again and I was actually seeing that slider ```401``` in DialogSeekbar never uses the focus texture, which has nothing to do with this PR.

MilhouseVH:
> Let's ask @milhouse to include it for while as that will give it a broader test base

Been in since #0909 (last 2 builds).

phil65:
> it was never broken, so can't be a fix ;-)

Technically not, but in terms of UX I would consider it a fix because those buttons were broken design.
Vertical lists filled by core / addons (=possibly very large amount of buttons) in combination with buttons which do not allow going left and right (like sliders) simply dont work and make it impossible to come up with a sane navigation layout for remotes / keyboards because you cant leave the focused list properly with button presses.
Still hoping for a v17 inclusion. :)

ronie:
updated. the slider in dialogslider should now behave as before.

BigNoid:
> the slider in dialogslider should now behave as before.

That non-focused issue in slider dialog is gone now and much better to not have to click in that dialog to activate it first. Thx!

AchimTuran:
@ronie 
> i think @AchimTuran uses a lot of slider controls in some of his adsp addons, that might be a reason to include those as well..
Right every audiodsp add-on has a minimum of one slider. e.g. the EQ dialog has only sliders.

I use both sliderex and slider. Consequently I need the activation on enter for both. Is that possible with your implementation?
In the last days I tested my add-ons with your commit and they worked out of the box.

ronie:
@AchimTuran yes, that will still work.

ronie:
jenkins build this please

MartijnKaijser:
No side affects so far using it :)

axbmcuser:
I like the idea of this change, but one thing feels strange:

Example 1:
- Play a movie, open VideoOSD
- Button "Settings" -> "Audio and subtitle settings"
- Slider "Volume" is selected, but not clicked yet so not changeable (which is correct and by intention)
- Click on slider so it becomes changeable
- Learned intuition tells you you may leave the "activated" slider state with a usual back button press
- Hitting the back button leads to closing the dialog, not leaving the activated slider state

If you have an addon which has a slider as the joystick addon from the initial post or the dim-screensaver for example, same thing happens and feels imho wrong.

Any chance to improve the situation? Making the "Back" action possibly check if a slider is in activated and before processing the usual Back-action-stuff close the activated slider instead?

ronie:
> Learned intuition tells you you may leave the "activated" slider state with a usual back button press

why do you think that's intuitive? i don't think we have other controls that can be deactivated with back?

axbmcuser:
Are there even any comparable controls which have such an "activated" state? I don't think so? 

Hard to describe, let's see:
In my opinion - regarding intuition, "activating" the slider into the "changeable-sub-state" has the same sub-level usability characteristics as opening a small on-top-context-menu which would supply the possibility to quickly change the values.
Of course i know it's in fact not a dialog on top of the slider, which still does not hurt my statement imho.

I mainly can tell that doing ux stuff my whole life i - as usual - just tried out the changes and caught myself (and also a friend i supplied the build) countless times closing the complete dialog by "accident" when by intuition it should not have been closed but first leave the new "activated-changeable-slider-sub-state" instead.

Feel free to disagree of course - i only mean well :)

phil65:
I agree that ideally we should open a dialog (--> DialogSlider.xml might do the job)
This still is an improvement though compared to the previous way, so for now i think this is fine.

phil65:
> Move cursor back up to the slider and see it's still in the activated state, which also feels wrong, "changeable-slider-state" should be set to "false" once you leave the slider
@axbmcuser this should be fixed now. Thx for testing.

MilhouseVH:
Is this related: http://forum.kodi.tv/showthread.php?tid=269814&pid=2418282#pid2418282

I can reproduce with my latest builds - you can use the sample AVI+SRT from [this post](http://forum.kodi.tv/showthread.php?tid=269814&pid=2416492#pid2416492)

ronie:
@MilhouseVH are you testing with the current version of this PR?
it's been updated a few times since you initially added it to your builds.

MilhouseVH:
Yes, I always build with the latest version of this PR that is current at the time of the build.

The subtitle offset breakage has occurred only in the last few days - build #0916 is OK, but build #0917 is the first build with the problem, so it's in a change added to this PR between the times below (all times UK BST/CET+1):
```
-rw-rw-r-- 1 neil neil 149207040 Sep 16 23:59 LibreELEC.tv/target/pi2/LibreELEC-RPi2.arm-8.0-devel-20160916235324-#0916-gb094ddc.tar
-rw-rw-r-- 1 neil neil 149207040 Sep 17 21:27 LibreELEC.tv/target/pi2/LibreELEC-RPi2.arm-8.0-devel-20160917212349-#0917-g082a8ea.tar
```

The only candidate is https://github.com/xbmc/xbmc/pull/10445/commits/ea253be7e08507264f3a3006c681d0a1f0e91dcd

ronie:
ping @phil65 must be the unfocus stuff we've added...

ronie:
@MilhouseVH updated. 
https://github.com/xbmc/xbmc/pull/10445/commits/ae66de78778a4364c4273048a15765d19593b00a should fix the issue.

MilhouseVH:
Looks good now, many thanks. I've tested the sliders in audio/video and they're also still working.

axbmcuser:
I coincidentally noticed that the latest PS GUI has a very similar slider activation mechanism like it's intended by this PR. They wanted to avoid the same left/right navigation problems.

Regarding back-button intuition issues: On PS, while being on an activated slider, the back-button also does not close the dialog, but only ends the activated slider state at first. Just for statistics.

Some screens - if interested:

http://i.imgur.com/Kr2LRJA.png
http://i.imgur.com/VFjZb5U.png
http://i.imgur.com/3xVUeJW.png
http://i.imgur.com/nSkVVW3.png

phil65:
Hmm I think a nicer fix would be to set m_active of the slider control to TRUE in CGUIDialogSlider::ShowAndGetInput instead of checking g_windowManager.IsWindowActive(WINDOW_DIALOG_SLIDER)). Perhaps @tamland could take a look?

ronie:
cool, thx! looks like an interesting option. will check it out.

ronie:
i've checked but, ShowAndGetInput() makes a call to SetSlider() which already calls SetActive()
so i'm afraid that's not it.

tamland:
Not sure I understood what the problem was.. but definitely do not want controls to depend on global window state.

ronie:
reverted the unfocus function.
the slider will now stay activated (as it was before) when you navigate to another control.

phil65:
@tamland:
the problem happens on following scenario:

Go to Addon Browser, select Dim Screensaver addon, open the addon settings which contain a slider "dim level"
Activate slider by pressing it, change it as you like, but dont confirm with "select" action
Move cursor down to the "OK" button, don't press it yet
Move cursor back up to the slider and see it's still in the activated state, which also feels wrong, "changeable-slider-state" should be set to "false" once you leave the slider

MilhouseVH:
Need to add `OnUnFocus()` to ` xbmc/guilib/GUISliderControl.h`?

```
CPP     xbmc/guilib/Texture.o
GUISliderControl.cpp:252:35: error: no 'void CGUISliderControl::OnUnFocus()' member function declared in class 'CGUISliderControl'
 void CGUISliderControl::OnUnFocus()
                                   ^
/home/neil/projects/LibreELEC.tv/build.LibreELEC-RPi2.arm-8.0-devel/kodi-17.0-beta2-6e9d6fb/Makefile.include:97: recipe for target 'GUISliderControl.o' failed
make[2]: *** [GUISliderControl.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```

ronie:
yup, didn't have the time to test it yet.
updated and squashed now.

ronie:
jenkins build this please

MartijnKaijser:
jenkins build this please

PR #10446, opened 2016-09-09
[cmake] Move to root

hudokkow:
Just to get the ball rolling.
cc/ @fetzerch, @wsnipex, @notspiff and anyone else interested in testing/debugging.

Build-tested under linux (core/addons).

buildsteps/depends stuff separated on purpose. Unified deps and jenkins scripts are voodoo to me.


fetzerch:
this btw would also allow later also to get rid of the CORE_SOURCE_DIR variable (and we can use CMAKE_SOURCE_DIR) instead

hudokkow:
Yeap, that's the plan. BTW, I need a bool in jenkins to test this. Something like "Use CMakeLists.txt is in root folder."

Paxxi:
Can we get rid of the treedata with this? It feels wrong but maybe there's a good reason for it that I am missing

fetzerch:
The treedata is sth different. That tells cmake which CMakeLists from which subfolders to include.
Alternative is to use add_subdirectories in the main CMakeLists.txt and all that contain folders.
I think @notspiff's main intention with that was to be able to conditionally include folders on a per platform level. I don't have a very strong opinion about it, the add_subdirectories approach is probably easier to understand for people that have used cmake in other projects.

Paxxi:
That's my thinking, add_directory is easier to follow and translates better between projects. I know it's not really related but thought maybe we can sneak it in here

notspiff:
Separation of code and data is the reason.

7. okt. 2016 19:36 skrev "Pär Björklund" <notifications@github.com>:

afedchin:
+1 to get rid of treedata. I think add_directory for each first level subdirectory it'a a better way.
something like this https://github.com/afedchin/xbmc/commit/84aa2cb0daf270ef7349f2a895704dc7daa1398b?diff=unified#diff-3de6f69aaaeae7bbc934887287370c56R69

hudokkow:
This is now (almost) ready to go in.

A review would be nice, specially on the buildsteps/depends commit. Right now it [successfully builds](http://jenkins.kodi.tv/view/Helpers/job/BuildMulti-All/1813/) core and add-ons but, for some reason, jenkins forgets to report the addons sucess/failure list. After battling 3 days trying to find out why, I can only assume it's something in front of my eyes that I can't see or jenkins build scripts need adapting to work with the new location of stuff.

As for the rest, I'm having trouble with s/CORE_SOURCE_DIR/CMAKE_SOURCE_DIR/ so I'll leave it to another PR. treedata stuff is also for another PR, if we decide that's the way to go.

Note to self: drop travis-ci temp commit.

hudokkow:
CORE_SOURCE_DIR is gone, except for add-ons project.

Furthermore, $(ADDON_PROJECT_DIR)/.failure and $(ADDON_PROJECT_DIR)/.success are being correctly populated in jenkins builds.
Why jenkins doesn't pick them up after building and creates the neat green lights is beyond me.

Note to self, again: drop travis-ci temp commit.

wsnipex (approved):
looks good.
Are you coordinating the merge with notspiff's autotools removal?

hudokkow:
This is a tricky one because it needs someone with jenkins admin access.
I suggest we merge this and adapt jenkins jobs to look for the CMake project in root folder at the same time.
After we're sure everything is working, we can test and merge autotools removal.

@notspiff, your thoughts?

akva2:
i think opposite. remove autotols first. then we move cmake. i really do not recommend having cmake coexist with autotools on top of the build tree, it causes a shitload of headache on build servers and not. i have tried that before it's no fun. and sure, we could probably eat those issues temporarily, but in case it takes more time than anticipated...

we can however not remove autotools before i / somebody find the time to fix up the toolchain mess.

hudokkow:
Works for me too. Can you rebase your PR so we can get it in?
I don't want to keep rebasing this more than needed. ;)

akva2:
again, we cannot remove autotools before cmake build is fixed. it will break downstreams e.g. gentoo, as they cannot package using their build scripts because they use toolchains. once that is in place i can rebase. but it's too messy to do too often as you have noticed so i will hold to it's ready.

hudokkow:
Ah! You're talking about https://github.com/xbmc/xbmc/pull/11088
So order should be: fix toolchain (wrong) assumptions -> remove autotools - > move cmake.

Okido!

akva2:
yup

fetzerch (approved):
I've adapted jenkins to look in the right place for the addon marker files. (depending on CMAKE_ROOT - which we have to default=1 once this goes in).

Triggered a new testbuild: jenkins.kodi.tv/view/Helpers/job/BuildMulti-All/1845

Otherwise any objections?
Just a fair warning: this invalidates all generated project files and devs have to run cmake again and do a full rebuild.

Memphiz:
@fetzerch i think the osx and ios docs need another adaption for this too ;)

ksooo:
@fetzerch after merging this some documentation needs to be adapted to the folder strucure changes, e.g. https://github.com/xbmc/xbmc/blob/master/docs/README.osx.md where "projects/cmake" is mentioned. Could you pleaes do the respective updates? Thanks.

fetzerch:
Fixed both in: https://github.com/xbmc/xbmc/pull/11264

ksooo:
thx

PR #10447, opened 2016-09-10
CGUIDialogContentSettings: fix crash when choosing a scraper and no scraper is set yet

Montellese:
This fixes a crash I came across after deleting my MyVideosXXX.db file and setting up my sources from scratch. When I tried to choose a scraper for my music videos Kodi crashed on me because it tries to access `m_scraper->ID()` even though `m_scraper` is an empty `std::shared_ptr`. I checked all the other places in the file and they all already have the proper checks in place.

### DIFF
--- a/xbmc/settings/dialogs/GUIDialogContentSettings.cpp
+++ b/xbmc/settings/dialogs/GUIDialogContentSettings.cpp
@@ -249,10 +249,13 @@ void CGUIDialogContentSettings::OnSettingAction(const CSetting *setting)
   else if (settingId == SETTING_SCRAPER_LIST)
   {
     ADDON::TYPE type = ADDON::ScraperTypeFromContent(m_content);
-    std::string selectedAddonId = m_scraper->ID();
+    std::string currentScraperId;
+    if (m_scraper != nullptr)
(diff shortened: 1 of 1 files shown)


Montellese:
jenkins build this please

Montellese:
@ronie: You touched this code last in b89eab75f3991a2a7f8454dbaf0953a26f0d12a7. Any objections?

Tolriq:
@Montellese sorry to hijack this, but since I'm no more involved and can't use forum to PM and see you are around :) 

Anyway if you recall JSON version was said to be updated to 8 for beta / release, don't know if you still agree with that decision and could manage it.

ronie:
@Montellese not at all. thanx for the fix!

zag2me:
Thanks, I believe this is the music vid problem I cam across recently, Thinking about it I probably deleted the database manually.

Montellese:
I haven't been around for a while. Should AppVeyor and Travis CI builds be re-triggered to be merge ready or can this go in?

MartijnKaijser:
@Montellese if jenkins is fine you can ignore the others.

PR #10450, opened 2016-09-11
[Estuary][PVR] Guide window: add vertical scrollbar to channel list.

ksooo:
Requested many times in the forum, because mouse users currently have no (!) way to scroll through the channels displayed in the epg grid.

So, I consider this an important bug fix.

screenshot002

@BigNoid are the skin changes okay?
@Jalle19 mind taking a look at the code changes?

### DIFF
--- a/addons/skin.estuary/1080i/MyPVRGuide.xml
+++ b/addons/skin.estuary/1080i/MyPVRGuide.xml
@@ -27,11 +27,12 @@
 					<top>0</top>
 					<width>1700</width>
 					<height>510</height>
-					<pagecontrol>10</pagecontrol>
+					<pagecontrol>60</pagecontrol>
 					<scrolltime tween="quadratic" easing="out">200</scrolltime>
 					<timeblocks>36</timeblocks>
 					<rulerunit>6</rulerunit>
 					<onleft>9000</onleft>
+					<onright>60</onright>
 					<onup>10</onup>
 					<ondown>10</ondown>
 					<viewtype label="19032">list</viewtype>
@@ -153,6 +154,16 @@
 							<texture>$VAR[PVRTimerIcon]</texture>
 						</control>
 					</focusedlayout>
+				</control>
+				<control type="scrollbar" id="60">
+					<right>85</right>
+					<top>45</top>
+					<width>13</width>
+					<height>465</height>
+					<onleft>10</onleft>
+					<onright>10</onright>
+					<orientation>vertical</orientation>
+					<texturesliderbackground colordiffuse="22FFFFFF">colors/white.png</texturesliderbackground>
 				</control>
 				<control type="group">
 					<top>540</top>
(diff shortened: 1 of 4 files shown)


ksooo:
@HitcherUK you might be interested in adding this to Confluence

BigNoid:
The scrollbar stops in the middle of the bar when the list end is reached, while it should stop at the end of the bar. it's not the skin code afaict.

schermafbeelding 2016-09-11 om 13 00 28

Maybe we should hide the scrollbar when there's no mouse enabled as it can't be (reasonably) navigated too with the keyboard. (```system.getbool(input.enablemouse)```)

ksooo:
@BigNoid 

> The scrollbar stops in the middle of the bar when the list end is reached, while it should stop at the end of the bar. it's not the skin code afaict.

 code issue. will look into it...
> Maybe we should hide the scrollbar when there's no mouse enabled as it can't be (reasonably) navigated too with the keyboard. (system.getbool(input.enablemouse))

there is not a single reference to "enablemouse" in Estuary xml. maybe this is a general issue that should not be fixed in this PR, then.

BigNoid:
> there is not a single reference to "enablemouse" in Estuary xml. maybe this is a general issue that should not be fixed in this PR, then.

True, there are no controls that are only useful for mouse input up to now. You will find plenty of those in Confluence for example.
I don't mind not hiding it, but than we should add ```<onright>60</onright>``` to the epggrid control. It can only be reached by scrolling completely to the right first so it's not very useful, but at least it is reachable by keyboard.

ksooo:
@BigNoid Inserted system.getbool(input.enablemouse), as you suggested. Scrollbar range issues should be solved now. Would you mind retesting?

MartijnKaijser:
Mouse is always enabled by default so not sure how much it would really avoid confusion for keyboard users :)

BigNoid:
> Mouse is always enabled by default so not sure how much it would really avoid confusion for keyboard users :)

True and it might also work as just a visual indicator of how far you are in the epg on second thought so might be beneficial for non-mouse users too.

Tested and works good now @ksooo

ksooo:
@BigNoid okay, will remove the enablemouse visibility condition and insert onright 60, no?

BigNoid:
> okay, will remove the enablemouse visibility condition and insert onright 60, no?

yes

ksooo:
This fixes http://trac.kodi.tv/ticket/15886 , btw.

ksooo:
jenkins build this please

HitcherUK:
> @HitcherUK you might be interested in adding this to Confluence

Done -

https://github.com/xbmc/skin.confluence/commit/fa8d8c210c374f55f8d2e163818dadaa01b2af4d

ksooo:
Cool, thanks @HitcherUK.

PR #10453, opened 2016-09-11
win32: fix eac3 for wasapi, 2nd round

FernetMenta:
2nd round


FernetMenta:
confirmed fixed: http://forum.kodi.tv/showthread.php?tid=289215&pid=2413068#pid2413068
no need for other jenkins builds.

PR #10454, opened 2016-09-11
[cmake] Fix skin packaging for windows

fetzerch:
This reverts commit fc70dc26dd87977129f990fd20e371fbc208b549 as it might have broken skin packaging on windows.

If that fixes it, we need a proper fix for the double installation on linux.

@MartijnKaijser: Triggerd jenkins build: jenkins.kodi.tv/job/WIN-32/9924


MartijnKaijser:
confirmed this fixes the broken Estuary skin

PR #10455, opened 2016-09-12
[win32] Fix a possible leak and a micro-optimisation.

ace20022:


xbmc/Util.cpp
```diff
@@ -172,16 +172,18 @@ std::string GetHomePath(const std::string& strTarget, std::string strPath)
   auto bufSize = GetCurrentDirectoryW(0, nullptr);
   if (bufSize > 0)
   {
-    auto buf = new wchar_t[bufSize];
-    if (0 != GetCurrentDirectoryW(bufSize, buf))
+    std::unique_ptr<wchar_t> buf(new wchar_t[bufSize]);
```
neo1973: Shouldn't this be `std::unique_ptr<wchar_t[]>` for arrays? (It doesn't make difference for PODs I believe, but it's nicer)

ace20022: thanks!

Paxxi:
Was there a bad return path that could leak? Using unique_ptr is obviously correct so +1

ace20022:
> Was there a bad return path that could leak? Using unique_ptr is obviously correct so +1

The next lines are:
```
 if (IsDirectoryValidRoot(currentDirectoryW))
         return currentDirectory;
     }
   }
 
   // If we ended up here we're most likely screwed
   // we will crash in a few seconds
   return strPath;
}
```

So according to the comment, if the leak would happen, the program would exit anyway ;)

MartijnKaijser:
Jenkins build this please

Paxxi:
Ahh right, any failure in that method will cause our initialization to fail and we crash or exit

MartijnKaijser:
ok to merge?

Paxxi:
Yup

PR #10460, opened 2016-09-13
[RDS] enable internal announcements by default

ronie:
remove the setting to enable internal RDS announcement messages.
i don't think we should bother users with low-level stuff like this. just enable it by default.

afaik none of the other messages send to the AnnouncementManager are user-configurable.

http://forum.kodi.tv/showthread.php?tid=290588

ping @AlwinEsch


AlwinEsch:
Is OK to remove

ronie:
thx!

jenkins build this please

PR #10461, opened 2016-09-13
 [pvr] show notification if no pvr addon is enabled while entering a pvr window

xhaggi:
This PR will show a notification dialog (DialogOK) if you enter a PVR window with no active/enabled PVR addon.

@ksooo @Jalle19 @ronie what do you think?

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -10354,7 +10354,13 @@ msgctxt "#19295"
 msgid "Delete timer rule"
 msgstr ""
 
-#empty strings from id 19296 to 19498
+#. Label if no pvr addons are enabled
+#: xbmc/pvr/windows/GUIWindowPVRBase.cpp
+msgctxt "#19296"
+msgid "No PVR add-on enabled"
+msgstr ""
+
+#empty strings from id 19297 to 19498
 
 #: xbmc/epg/Epg.cpp
 msgctxt "#19499"
--- a/xbmc/pvr/windows/GUIWindowPVRBase.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRBase.cpp
@@ -22,6 +22,7 @@
 #include "GUIWindowPVRRecordings.h"
 
 #include "Application.h"
+#include "addons/AddonManager.h"
 #include "cores/AudioEngine/Engines/ActiveAE/AudioDSPAddons/ActiveAEDSP.h"
 #include "dialogs/GUIDialogKaiToast.h"
 #include "dialogs/GUIDialogNumeric.h"
@@ -339,8 +340,14 @@ void CGUIWindowPVRBase::SetInvalid()
 
 bool CGUIWindowPVRBase::CanBeActivated() const
 {
-  // No activation if PVR is not (yet) available.
-  return g_PVRManager.IsStarted();
+  // check if there is at least one enabled PVR add-on
+  if (!ADDON::CAddonMgr::GetInstance().HasAddons(ADDON::ADDON_PVRDLL))
+  {
+    CGUIDialogOK::ShowAndGetInput(CVariant{19296}, CVariant{19272}); // No PVR add-on enabled, You need a tuner, backend software...
+    return false;
+  }
+
+  return true;
 }
 
 bool CGUIWindowPVRBase::OpenChannelGroupSelectionDialog(void)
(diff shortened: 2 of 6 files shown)


xhaggi:
BTW we should change the message which is currently used in the case we don't have the PVR stuff ready. Something like "No PVR client available. Waiting for one to be started."

ronie:
i think the root issue is skins don't have an infobool anymore to check whether or not to hide the pvr related items on the home menu.

see: http://forum.kodi.tv/showthread.php?tid=265728

xhaggi:
@ronie not really. For Xperience1080 I don't want to hide the Section "TV" anymore because it's part of our core. Therefor it's possible to hide the section manually within the skin settings. The infobool `HasTVChannels` and `HasRadioChannels` are kicked in after the PVR client is fully initialized which results in a really odd behavior within the skin, because the section is visible 5 to 10 sec after the home screen is visible.

ronie:
i don't mind if you want to offer those menu items by default :-)

but most skinners prefer to auto enable them based on whether pvr is enabled on the users system.
given the fact the vast majority of the kodi users don't use pvr, i don't think that's unreasonable.

xhaggi:
Ok I have refactored the whole PR after a nice talk with @ksooo. Now we show a notification to the user in case there is no enabled pvr addon.

FernetMenta:
I am not in favour of this. The strategy goes in opposite direction. We have dynamic connection handling, means that clients can come and go.
Having PVR windows as startup windows is a unmaintainable mess if windows depend on clients connected.

xhaggi:
@FernetMenta I've changed my first approach and now the goal of the PR is to show a notification dialog if a user does not have an enabled PVR addon while entering a PVR window. It does not depends on client connection or on the PVR manager like it was before.

FernetMenta:
@xhaggi following scenario:

- you have a connected client
- you have opened a pvr window
- go to home window
- for some reason client disconnects (this can happen)
- you can't enter pvr window again

in addition, what should happen if user is in a pvr window and client disconnects?

we don't lock video or music windows either, if network drives are not available. (and those are the only configured sources)

I don't see any disadvantage in opening an empty window. Maybe display some notification but don't lock the window.

xhaggi:
We only check if the user has an enabled PVR client addon via the addon manager nothing else atm. In case he didn't have an enabled PVR client addon we display the notification and do not activate the PVR window. This has no effect on your described scenario.

FernetMenta:
Sorry, you are right. For some reason I though you were checking for "connected clients". Disregard my comments and sorry for the noise.

ksooo:
@xhaggi runtime tested. works as designed. +1

xhaggi:
@FernetMenta np
@ksooo thanks for runtime testing it
@MartijnKaijser good to go in?

MartijnKaijser:
i guess
jenkins build this please

xhaggi:
build errors in jenkins are unrelated. @MartijnKaijser your button.

xhaggi:
I've updated the PR again, because there is no need to check `g_PVRManager.IsStarted()` in `CanBeActivated()` because all windows are fail-safe (can be activated without a started PVR manager) and reload its content if the connection established.

@FernetMenta that's what you suggested.

>* for some reason client disconnects (this can happen)
* you can't enter pvr window again

FYI: ATM if a PVR client disconnects after the PVR manager was started the state of the pvr manager didn't change anymore. Entering a PVR window is possible. If you start/restart Kodi and the PVR backend is not reachable the PVR manager would not start. In that case you can't enter the PVR manager.
That's the reason why it's safe to remove the `g_PVRManager.IsStarted()` check, because a connection lost while running Kodi allow you to enter the PVR windows so why should this be prevented if you start Kodi.

xhaggi:
jenkins build this please

xhaggi:
@ronie @phil65 ping you too because it make sense to show the TV section without checking the infobools `PVR.HasTVChannels` and `PVR.HasRadioChannels` and better let the user decide in the skin settings if he want to hide the TV section.

xhaggi:
any objections .. if not I'll shove it in

ksooo:
+1

NedScott:
Semi-related to this, but it is essential that there is an easy way to get to "PVR add-ons" from within the PVR settings. This is needed not only for enabling/disabling the PVR-add-ons, but for accessing settings in the add-on window, changing the PVR add-on, etc.

A shortcut to the PVR add-on grouping would be enough to improve this significantly for users. Even when you know where to go, it makes the workflow easier and faster to have a connection between PVR add-ons and PVR settings.

The notification for "no PVR add-on" is only a partial solution to that bigger problem.

PR #10462, opened 2016-09-13
switched <enter> from FullScreen back to Select as it breaks remotes …

antigenx:
…using lirc/devinput

### DIFF
--- a/system/keymaps/remote.xml
+++ b/system/keymaps/remote.xml
@@ -32,7 +32,7 @@
       <up>Up</up>
       <down>Down</down>
       <select>Select</select>
-      <enter>FullScreen</enter>
+      <enter>Select</enter>
       <pageplus>PageUp</pageplus>
       <pageminus>PageDown</pageminus>
       <back>Back</back>


FernetMenta:
there is nothing to switch back because enter has always been mapped to FullScreen

antigenx:
Dunno what to tell ya, upgrading to Ubuntu 16.04 broke my remote's behaviour and this was the fix. Maybe they switched a key definition from KEY_SELECT to KEY_ENTER. FullScreen still has KEY_ZOOM which makes more logical sense to me. Should KEY_SELECT and KEY_ENTER not be considered generally equivalent, rationally speaking?

FernetMenta:
> Should KEY_SELECT and KEY_ENTER not be considered generally equivalent

if they were supposed to be the same, it wouldn't be two different keys. we don't want to break behavior for all other users.
you are digging in the wrong place. find our why behavior changed on your system. you may want to change you lircmap.xml to adapt to that change.

antigenx:
I agree that we don't want to break behaviour for other folks, I don't know why this changed suddenly. I don't (well, didn't until now) have any customisations in lirc, the lircmap or the remote files, and yet the behaviour changed on upgrade to 16.04.1 -- that suggests to me that the behaviour probably changed for other folks too.

Respectfully, I disagree, there are plenty of examples where there are different button codes for the same expected result such as KEY_OK and KEY_SELECT and KEY_ENTER. It's not like there's a hard standard in the IR world. Logically, these buttons should have the same effect. Do you toggle FullScreen when you hit ENTER on your keyboard? No. 

End of the day, I solved my problem with a simple one-line patch. PR is there so if you feel you need it in the future, it's just a click away.

Cheers,
d

NedScott:
I guess the question is: what does KEY_ENTER physically look like on your remote? It might be that for some past remote it was labeled in such a way that fullscreen made sense, but maybe things are different now for more modern remotes? (I honestly don't know)

antigenx:
For me, it's the button in the centre of the dpad. I'll keep digging and see if I can find out more. I find it curious that it hasn't been talked about in the forum... maybe it does only affect me. 

My reasoning is that key codes should have an expected outcome based on their implied purpose. However, FullScreen is one of those things where there's not going to be a dedicated button for it on any remote. KEY_ZOOM strikes me as a logical choice for it, but maybe not all remotes have that, so its understandable that KEY_ENTER was chosen as a backup. A bit of a kludge, but "yous do what you gotta"

I'll keep digging on this one. Clearly it's not as simple as one would hope :)

NedScott:
You've peaked my curiosity, and I have an obsession with remotes and keymaps. If you make a thread for this on the Kodi forum and tell me about it, we can resume the discussion there. I'm sure there's something we can fix. What you describe should be "select" on most remotes, but I haven't used Ubuntu in years.

phunkyfish:
If the author would like it kept open please just request so on the PR. If you feel there is still merit in merging this PR you can request the label be changed to PR Cleanup: Backburner.

PR #10464, opened 2016-09-14
[pvr][epg] fix division by zero in CGUIEPGGridContainer::GetLabel

xhaggi:
This PR fixes the crash while entering the Guide window while PVR manager is starting up.

@ksooo mind taking a look.


ksooo:
I have no clue how this can happen, but the fix looks sane.

ksooo:
jenkins build and merge

PR #10465, opened 2016-09-14
[WIP] [skin.estuary] create custom system window + tidy up settings window

jjd-uk:
This something I wanted to discuss at Devcon, decided to put this up now so people not attending can comment and also to get early feedback from those that are.

The settings window has historically been a dumping ground for windows such the Add-on browser and System Info that don't fit else where, this seek to resolve this by creating a new custom System window containing:

Add-on browser - easy access to this is still needed in the event users turn off the Add-ons home menu item

File manager - I don't believe a place on the Home screen is merited for this.

Event Log - several have bemoaned that this is so hidden within the settings, this brings it an an easier to find place.

System Info - should not be in Settings window as there is nothing changeable within this window. 

custom_system_window

This window is accessed by replacing the File Manager button on the Home window

home_window

In addition removing System Info from Settings window allows tidy up there by dropping the addition settings wording after each Section name, so from

image

to

settings_window

Note - I did look see if I could code a new window in core so this is something all skins could have, but quickly got lost so that for the moment that is well above my skill level.

ping @HitcherUK @phil65 @ronie @BigNoid @MartijnKaijser @da-anda

### DIFF
--- a/addons/skin.estuary/1080i/Custom_1110_System.xml
+++ b/addons/skin.estuary/1080i/Custom_1110_System.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<window>
+<window id="1110">
 	<defaultcontrol always="true">9000</defaultcontrol>
 	<controls>
 		<include>DefaultBackground</include>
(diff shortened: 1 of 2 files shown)


NedScott:
The only problem I have with this is it becomes confusing about "system settings" and this "system" window. Also, add-ons really need to be accessible from settings, as they often are settings from the perspective of the user. This feels more like a workaround for that problem, but it doesn't get closer to fixing it, IMO.

If anything, I would welcome this if the button was accessed from settings (or both home screen and settings) and was called something other than System, simply to avoid confusion. However, I don't have a better suggestion.

To be clear, I don't oppose this idea. I see it as no better or worse than we currently have.

jjd-uk:
One is System activities the other System settings to my mind that was a clear enough distinction, however I'm willing to consider any other ideas for any appropriate name for the new window if the idea is otherwise acceptable.

BigNoid:
> Add-on browser - easy access to this is still needed in the event users turn off the Add-ons home menu item

We may need a shortcut to the "my addons" list for this case in the add ons setting section as installed add ons should be accessible so they can be uninstalled. 
But associating installing new add ons with settings or system is what was tried to correct with Estuary from Confluence.

jjd-uk:
I'll happily resurrect https://github.com/xbmc/xbmc/pull/10205 with whatever you want to the shortcut button to do, but ronie didn't add-on access being buried so far down the structure, adding to this new window resolves that.

Is there a case for doing both? so adding the shortcut button for installed addons as you suggest but also what I've here adding the add-on browser in this custom window. 

I can demo this over the weekend if you want to see it in action.

PR #10466, opened 2016-09-14
AESinkWASAPI: Move deprecated layout before first 7.1 layout (#16903)

fritsch:
Fixes http://trac.kodi.tv/ticket/16903


FernetMenta:
thanks!
already tested by jenkins. inject?

fritsch:
I sent the user out to find the following:

a) AMD users non legacy, testing 5.1 dts decode 5.1 flac 7.1 whatever
b) Intel user with SNB or older
c) Intel user with HSW or newer
d) Nvidia user with ION-2 or older
e) Nvidia user with GT720 or newer
f) For both Windows 10 and Windows 7

We need to take care for regressions. This file is like that for many years :-)

FernetMenta:
the fix is correct. if there is breakage, then we hit another issue. no need to delay this fix.

fritsch:
jenkins build and merge

PR #10467, opened 2016-09-16
[PVR] add support for streams provided by inputstreamaddons

frankr612:
There are several plugins available providing iptv channels through inputstreamaddons (smoothstream, mpd ...) which can't be accessed by the pvr addons yet. With this simple solution it's possible to use a plugin url given as strStreamUrl which will then call the plugin and have the path and properties set to support an inputstream addon. So basically this means that "setResolvedUrl" called by a plugin can be used even for PVR addons/content. The main use case for this right now would be the Simple IPTV Client addon where one would be able to add "plugin://" paths to the m3u playlist.

I tested this with the IPTV Simple Client and the SkyGo plugin (german version) and other plugins providing content through inputstreamaddons (free and drm content) and it works flawlessly.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDFactoryDemuxer.cpp
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDFactoryDemuxer.cpp
@@ -99,6 +99,15 @@ CDVDDemux* CDVDFactoryDemuxer::CreateDemuxer(CDVDInputStream* pInputStream, bool
         else
           return nullptr;
       }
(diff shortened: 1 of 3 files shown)


FernetMenta:
the otherStream hack is going to die soon. see dev section of the forum.

frankr612:
Sorry, wasn't aware of this. Now found the thread you are reffering to I guess...seems it was all done and rejected before. ;)

Jalle19:
Should we close this one?

MartijnKaijser:
@frankr612 thx for trying to contribute though it won't be merged

PR #10468, opened 2016-09-16
[win32][fix] Add app manifest to the build to mark us as dpi aware

Paxxi:
This solves the issue where we display outside the screen if the dpi scaling is set to anything higher than 100%. Tested with 125/150/175 and looks good


afedchin:
Can't cmake set DPI aware via project property? I don't like an idea to support the manifest file.

Paxxi:
The other way is to call the manifest tool to generate it on the fly but I read somewhere that method is discouraged since cmake 3.4 and the app.manifest file only really needs changing when windows 11 comes out probably so it's low maintenance

afedchin:
oki. that's fine for me.

PR #10470, opened 2016-09-16
[Estuary] Misc stuff

phil65:
- UX improvement: Added some fallback labels in case no poster is available
  bildschirmfoto von 2016-09-16 13-41-03
- small cleanup


AlwinEsch (approved):
Looks good and also a test for me with latest github changes.

PR #10471, opened 2016-09-16
AE: fix mem leak in tempo filter

FernetMenta:
see title


fritsch:
jenkins build and merge

PR #10472, opened 2016-09-16
[Estuary][PVR] Handle visibility of PVR home screen entries like the other home screen entries.

ksooo:
… for consistency reasons this more than useful, imo.

screenshot002
screenshot003

@phil65 @BigNoid @ronie what do you guys thin about this?
@xhaggi fyi


MartijnKaijser:
@ksooo the text "enter add-ons browser" needs to be adjusted.

addons/skin.estuary/1080i/Home.xml
```diff
@@ -565,6 +565,20 @@
...
+						</include>
+						<include content="ImageWidgetHeader">
+							<param name="widget_header" value="$LOCALIZE[19020]" />
+							<param name="visible" value="!System.HasPVRAddon"/>
+						</include>
+						<include content="ImageWidget">
+							<param name="image_path" value="special://skin/extras/home-images/tv.jpg"/>
+							<param name="text_label" value="$LOCALIZE[31143]" />
+							<param name="button_label" value="$LOCALIZE[31144]" />
+							<param name="button_onclick" value="ActivateWindow(addonbrowser,addons://,return)"/>
```
xhaggi: are we able to go further down to pvr addons?

addons/skin.estuary/1080i/Home.xml
```diff
@@ -659,6 +675,20 @@
...
+						</include>
+						<include content="ImageWidgetHeader">
+							<param name="widget_header" value="$LOCALIZE[19021]" />
+							<param name="visible" value="!System.HasPVRAddon"/>
+						</include>
+						<include content="ImageWidget">
+							<param name="image_path" value="special://skin/extras/home-images/tv.jpg"/>
+							<param name="text_label" value="$LOCALIZE[31143]" />
+							<param name="button_label" value="$LOCALIZE[31144]" />
+							<param name="button_onclick" value="ActivateWindow(addonbrowser,addons://,return)"/>
```
xhaggi: same here

xbmc/GUIInfoManager.cpp
```diff
@@ -1190,7 +1195,8 @@ const infomap system_labels[] =  {{ "hasnetwork",       SYSTEM_ETHERNET_LINK_ACT
                                   { "startupwindow",    SYSTEM_STARTUP_WINDOW },
                                   { "stereoscopicmode", SYSTEM_STEREOSCOPIC_MODE },
                                   { "hasadsp",          SYSTEM_HAS_ADSP },
-                                  { "hascms",           SYSTEM_HAS_CMS }};
+                                  { "hascms",           SYSTEM_HAS_CMS },
+                                  { "haspvraddon",      SYSTEM_HAS_PVR_ADDON }};
```
xhaggi: it would make sense if you split those into a separate commit `[gui] new infobool HasPVRAddon`

MilhouseVH:
"Browse add-ons" would be better.

ksooo:
> "Browse add-ons" would be better.

 i've chosen "enter" because we use this wording on the other home screen setup pages as well, for instance "Enter file section" for videos setup.

addons/skin.estuary/1080i/Home.xml
ksooo: i would like to, but i think this currently not possible. ping @tamland

xhaggi: we are .. use `ActivateWindow(addonbrowser,addons://user/xbmc.pvrclient,return)`

ksooo: what we would need is a listing of all available pvr client addons from all active repos. i tried with "addons://search/xbmc.pvrclient", (xbmc.pvrclient is the official internal type string for pvr client addons) but this did not work

xbmc/GUIInfoManager.cpp
ksooo: no problem can do.

addons/skin.estuary/1080i/Home.xml
xhaggi: as we deliver the PVR addons on each plattform we can do it that way. better as pointing the user to the root only.

xhaggi: if the user didn't find the right here, he is able to search for it manually etc.

xbmc/GUIInfoManager.cpp
xhaggi: thanks

xhaggi:
beside my comments, great and thx @ksooo  👍

ksooo:
@xhaggi split into two commits now

addons/skin.estuary/1080i/Home.xml
ksooo: > we can do it that way.

@xhaggi  sorry, what do you mean by "that way"?

xhaggi: i mean pointing the user directly to the installed pvr addons

ksooo: addons://user/xbmc.pvrclient i guess...

> as we deliver the PVR add-ons on each platform 

 do we? what about linux?

> we can do it that way. better as pointing the user to the root only.

 you mean to use "addons://user/xbmc.pvrclient", right? yep, this is actually better. will change

ksooo:
@xhaggi changed addon browse url like you suggested.

addons/skin.estuary/1080i/Home.xml
xhaggi: on linux you need to install the add-ons via ppa or build them by yourself, but all pvr addons are listed under "My add-ons > PVR clients" then.

xhaggi:
@ksooo thanks now it becomes really handy 👍

ksooo:
jenkins build this please

MilhouseVH:
Any ideas, could this be related?

http://forum.kodi.tv/showthread.php?tid=269814&pid=2418005#pid2418005

User is seeing freezes at startup, that seem to have started since this PR was added - he's generated a backtrace using `kill -SIGSEGV $(kodi.bin)` which references `ADDON::CAddonMgr::HasAddons()`. If a debug-enabled backtrace well help let me know and I'll try to get you one.

ksooo:
A stack trace with debug symbols would be great, yes.

ksooo:
@FernetMenta this is regression caused by the pvr mananger rewrite you did for for Krypton. http://sprunge.us/FIQN => thread 1 vs thread 122 => CPVRClients calling AddonMananger and vica versa... Any advise how to fix this properly?

ksooo:
@FernetMenta nvm, I just found a proper fix

ksooo:
> the root cause is ADDON::AddonPtr CPVRClient::GetRunningInstance() const
lower layers must not call into upper layers.

 in this case it's up to @tamland to fix addon mgr?

FernetMenta:
only PVR has this bad implementation of GetRunningInstance

xbmc/GUIInfoManager.cpp
```diff
@@ -6954,6 +6960,8 @@ bool CGUIInfoManager::GetBool(int condition1, int contextWindow, const CGUIListI
     bReturn = CProfilesManager::GetInstance().GetMasterProfile().getLockMode() != LOCK_MODE_EVERYONE;
   else if (condition == SYSTEM_HAS_PVR)
     bReturn = true;
+  else if (condition == SYSTEM_HAS_PVR_ADDON)
+    bReturn = CAddonMgr::GetInstance().HasAddons(ADDON::ADDON_PVRDLL);
```
FernetMenta: this should not be done here. it triggers creating instances of tons of addons on every render cycle.

FernetMenta: Use CBinaryAddonCache::GetAddons and the issue with the deadlock is solved too

MilhouseVH:
Here's a debug-enabled crashlog if you still need it: http://sprunge.us/YagS

ksooo:
@FernetMenta thanks for the advice. Well fix it that way.

PR #10473, opened 2016-09-16
[osx] - fix sse flags when using autoconf/xcode

Memphiz:
@FernetMenta ok? (its the similar approach in configure.ac like linux does)


FernetMenta:
Thanks
+1

Memphiz:
Jenkins build and merge

PR #10475, opened 2016-09-17
[contextmenu] remove incorrect log message

tamland:


xbmc/ContextMenuManager.cpp
```diff
@@ -78,11 +78,7 @@ void CContextMenuManager::Init()
 void CContextMenuManager::ReloadAddonItems()
 {
   VECADDONS addons;
-  if (!m_addonMgr.GetAddons(addons, ADDON_CONTEXT_ITEM))
-  {
-    CLog::Log(LOGERROR, "ContextMenuManager: failed to load addons.");
-    return;
-  }
```
fritsch: You don't only remove the logging - that you are aware, right?

tamland: Yeah, it's correct. I originally though GetAddons returned error code, but it doesn't. So it should continue. Sorry for the confusing commit message..

PR #10476, opened 2016-09-17
maybe fix android crash after #10335

tamland:
untested..

### DIFF
--- a/xbmc/addons/AddonSystemSettings.cpp
+++ b/xbmc/addons/AddonSystemSettings.cpp
@@ -31,6 +31,23 @@
 namespace ADDON
 {
 
+CAddonSystemSettings::CAddonSystemSettings() :
+  m_activeSettings{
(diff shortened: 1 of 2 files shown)


bkuhls:
Thanks! I can confirm that this patch applied on top of e36734747739f52df6c9efe012253484ff163de0 fixes the segfault I reported in https://github.com/xbmc/xbmc/pull/10335#issuecomment-247637674

MartijnKaijser:
will test on android
jenkins build this please

koying:
@tamland You could keep it static and load on first ctor only if empty...
That'll avoid the member var.

tamland:
But I have no need to avoid it.. Having it as member is fine; and I get constness. I was just being lazy with the original code

MartijnKaijser:
@tamland tested on Android and works

mazeyx:
awesome thanks guys. :)

MartijnKaijser:
objections?

PR #10479, opened 2016-09-17
cmake: make external, shared ffmpeg default

wsnipex:
Since our own ffmpeg is - as of v3.1 - essentially vanilla (-1 unimportant build related commit), lets use external, shared ffmpeg libs by default.
This should make the life of all linux distro kodi maintainers easier.

I have left out autotools on purpose as incentive for packagers to switch to cmake.
supersedes https://github.com/xbmc/xbmc/pull/10432

ping @fetzerch @notspiff @hudokkow @rbalint


wsnipex:
jenkins build this please

FernetMenta:
I would keep it as e have it now. Currently we don't carry any changes of ffmpeg in our repo but this can change tomorrow.

wsnipex:
If really necessary the default can be changed, but we should really try to stay away from custom patches after all those years trying to get rid of them.

jenkins build this please

project/cmake/modules/FindFFMPEG.cmake
```diff
@@ -1,18 +1,140 @@
...
+  set(ENABLE_INTERNAL_FFMPEG OFF)
+endif()
 
-if(ENABLE_INTERNAL_FFMPEG)
+# external FFMPEG
+if(NOT ENABLE_INTERNAL_FFMPEG)
   if(FFMPEG_PATH)
-    message(WARNING "Internal FFmpeg enabled, but FFMPEG_PATH given, ignoring")
+    set(ENV{PKG_CONFIG_PATH} "${FFMPEG_PATH}/lib/pkgconfig")
+    list(APPEND ${CMAKE_PREFIX_PATH} ${FFMPEG_PATH})
```
fetzerch: CMAKE_PREFIX_PATH doesn't need ${} here

project/cmake/modules/FindFFMPEG.cmake
```diff
@@ -1,18 +1,140 @@
...
+
+  if(PC_FFMPEG_FOUND
+     AND PC_FFMPEG_libavcodec_VERSION
+     AND PC_FFMPEG_libavfilter_VERSION
+     AND PC_FFMPEG_libavformat_VERSION
+     AND PC_FFMPEG_libavutil_VERSION
+     AND PC_FFMPEG_libswscale_VERSION
+     AND PC_FFMPEG_libswresample_VERSION
+     AND PC_FFMPEG_libpostproc_VERSION)
+    set(FFMPEG_VERSION 3.1)
```
fetzerch: I would set a variable at the top where you define FFMPEG_PKGS, that way if we bump the version, the version variables are in once place.

project/cmake/modules/FindFFMPEG.cmake
```diff
@@ -1,18 +1,140 @@
...
+                                    REQUIRED_VARS FFMPEG_INCLUDE_DIRS
+                                                  FFMPEG_LIBAVCODEC
+                                                  FFMPEG_LIBAVFILTER
+                                                  FFMPEG_LIBAVFORMAT
+                                                  FFMPEG_LIBAVUTIL
+                                                  FFMPEG_LIBSWSCALE
+                                                  FFMPEG_LIBSWRESAMPLE
+                                                  FFMPEG_LIBPOSTPROC
+                                                  FFMPEG_VERSION
+                                    FAIL_MESSAGE "FFmpeg 3.1 not found, please consider using -DENABLE_INTERNAL_FFMPEG=ON")
```
fetzerch: you can use the same var here as well

project/cmake/modules/FindFFMPEG.cmake
```diff
@@ -1,18 +1,140 @@
...
+
+  if(FFMPEG_FOUND)
+    set(FFMPEG_LIBRARIES ${FFMPEG_LIBAVCODEC} ${FFMPEG_LIBAVFILTER}
+                         ${FFMPEG_LIBAVFORMAT} ${FFMPEG_LIBAVUTIL}
+                         ${FFMPEG_LIBSWSCALE} ${FFMPEG_LIBSWRESAMPLE}
+                         ${FFMPEG_LIBPOSTPROC})
+    set(FFMPEG_LDFLAGS ${PC_FFMPEG_LDFLAGS})
+    list(APPEND FFMPEG_DEFINITIONS -DFFMPEG_VER_SHA=\"${FFMPEG_VERSION}\")
```
fetzerch: are you sure that quotes need to be escaped here? I've done a similar change for lirc recently and there it was not needed

fetzerch (commented):
I've mostly looked at the cmake module, others know better if we want internal or external ffmpeg by default.

One additional thing I would like to get rid of is the ffmpeg link wrapper (https://github.com/xbmc/xbmc/blob/master/project/cmake/CMakeLists.txt#L391). Don't yet have a good idea how, but if you have one, let me know.

FernetMenta:
the default should be our fork. don't change this

project/cmake/modules/FindFFMPEG.cmake
wsnipex: I'm pretty sure I put it like this for a reason, but don't remember why exactly. This var is only use internally for some loglines and seems to have worked as is for quite some time now.

wsnipex:
changed to fallback to internal if required version is not found.

@fetzerch to get rid of the link wrapper we'd have to build ffmpeg outside of the kodi project or introduce some magic to build it during config stage.

wsnipex:
@fetzerch  https://github.com/wsnipex/xbmc/commit/e4d36ab2518646c46599a244a32dd1fabb97dbb1 is a POC on building ffmpeg during config stage and getting rid of the link wrapper. Not sure if I like it better.

project/cmake/modules/FindFFMPEG.cmake
```diff
@@ -1,17 +1,164 @@
...
+if(NOT ENABLE_INTERNAL_FFMPEG)
   if(FFMPEG_PATH)
-    message(WARNING "Internal FFmpeg enabled, but FFMPEG_PATH given, ignoring")
+    set(ENV{PKG_CONFIG_PATH} "${FFMPEG_PATH}/lib/pkgconfig")
+    list(APPEND CMAKE_PREFIX_PATH ${FFMPEG_PATH})
+  endif()
+
+  # FFMPEG v3.1
+  set(REQUIRED_FFMPEG_VERSION 3.1)
+  set(FFMPEG_PKGS libavcodec>=57.48.101
```
FernetMenta: imo those checks are a bit overenginered. note that developers need to be able to build with any version of ffmpeg. if external ffmpeg is used we can expect from that persons that they know what they are doing. hence no version check required.

wsnipex: are you suggesting that we should just accept any libs thrown at us? If yes, we can scrap a whole lot of our build system and deal with the fallout every day.
A dev that insists on building with a lower version can patch this.
The same version checks exists in autotools for years and have been in cmake from the start as well.

wsnipex: oh and you call it over engineered, I call it architecture and good practice

rbalint:
Thank you for working on making our (distro packagers) life easier.
The smaller the patches are I have to carry the more time I can spend on other quality-related issues.

project/cmake/modules/FindFFMPEG.cmake
FernetMenta: > are you suggesting that we should just accept any libs thrown at us?

that is not the case here because we use our fork as default.

> A dev that insists on building with a lower version can patch this.

I disagree. Development is a main use case that has to be considered. In general developers don't be experts on build system. hence this has to be kept as simple as possible.

fritsch:
My take on this from a technical level (without politics):
As a developer I easily want to test newer / custom ffmpeg as in a lot of times, we work with upstream  to either fix issues or develop new features or doing API transitions. Until now I just used --with-ffmpeg=/opt/ffmpeg and all was working for me. I'd like to also have an easy solution in the future in a documented way, so that I don't have to dig into whatever, but just can keep my development use case as I did before without too much hazzle.

As a Kodi-Supporter especially on the bugtracker and the forum I don't want to fix bugs, that I solved with upstream ffmpeg again and again. That means: Though we work on getting fixes included with upstream the very same day - it won't be always possible and we will be "shipping" a custom kodi ffmpeg version, which then still is mandatory for distributions as everything else means known (!) (cause we obviously fixed bugs) breakage. Same accounts for cherry-picks we do from upstream, e.g.: hevc-10 bit fixes for dxva, or VAAPI / VP9 enhancements, which might break compatibility with distributions as certain fields are not defined (good case as it breaks in the compiler) or certain code paths (e.g. format queries return NULL) -> runtime segfault not happening for us and so on.

There are also other things besides the "versioning", which is: missing options. If you use an ffmpeg without mjpeg, png encoder support, our Image handling won't work. If you don't enable ffmpeg's dts-hd decoder,  we won't display metadata correctly anymore. So if you see the same version does not mean we run on this build, that's perhaps the biggest technical difference not yet named: versioning is not enough to make sure that everything is fine.

The message should not be: kodi is now "supporting" your distribution's specific ffmpeg version, cause we can't do that out of stated reasons. 

Btw. ffmpeg has deprecated codec_info and when fernet and I are done with v17 we will transition ffmpeg to codecpar - this will break backwards compatibility to every ffmpeg version < 3.1. Distributions need to be aware of this.

wsnipex:
jenkins build this please

wsnipex:
once more, jenkins build this please

wsnipex:
jenkins build this please ....

wsnipex:
jenkins build this please

wsnipex:
github issues, lets try again:
jenkins build this please

wsnipex:
jenkins build this please

wsnipex:
jenkins build this please

wsnipex:
@fetzerch this is ready from my side, can you please check again?

the following changed:
- new option for devs: -DWITH_FFMPEG=/path/to/prefix which _disables version checks_
nothing else needed to build with whatever (old) version you want
- existing option FFMPEG_PATH retains version checks

jenkins build this please

project/cmake/modules/FindFFMPEG.cmake
```diff
@@ -1,17 +1,199 @@
...
+# WARNING: this option is for developers as it will _disable ffmpeg version checks_!
+# Consider using FFMPEG_PATH instead, which _does_ check library versions
+#
+# Allows building with external ffmpeg not found in system paths,
+# without library version checks
+#
+# usage: -DWITH_FFMPEG=/path/to/ffmpeg_install_prefix
+if(WITH_FFMPEG)
+  set(FFMPEG_PATH ${WITH_FFMPEG})
+  message(STATUS "Warning: FFmpeg version checking disabled")
```
fetzerch: message(WARING)?

project/cmake/modules/FindFFMPEG.cmake
```diff
@@ -1,17 +1,199 @@
...
+
+# external FFMPEG
+if(NOT ENABLE_INTERNAL_FFMPEG OR CMAKE_CROSSCOMPILING)
   if(FFMPEG_PATH)
-    message(WARNING "Internal FFmpeg enabled, but FFMPEG_PATH given, ignoring")
+    set(ENV{PKG_CONFIG_PATH} "${FFMPEG_PATH}/lib/pkgconfig")
+    list(APPEND CMAKE_PREFIX_PATH ${FFMPEG_PATH})
+  endif()
+
+  # FFMPEG v3.1
```
fetzerch: remove the comment as we have the versions at the top?

project/cmake/CMakeLists.txt
```diff
@@ -46,16 +46,16 @@ option(ENABLE_AIRTUNES    "Enable AirTunes support?" ON)
...
-    option(ENABLE_INTERNAL_FFMPEG "Enable internal ffmpeg?" ON)
-  endif()
   option(FFMPEG_PATH        "Path to external ffmpeg?" "")
   option(ENABLE_INTERNAL_CROSSGUID "Enable internal crossguid?" ON)
   option(ENABLE_OPENSSL     "Enable OpenSSL?" ON)
   option(ENABLE_SDL         "Enable SDL?" OFF)
   if(CORE_SYSTEM_NAME STREQUAL linux OR CORE_SYSTEM_NAME STREQUAL freebsd)
     option(ENABLE_X11         "Enable X11 support?" ON)
     option(ENABLE_AML         "Enable AML?" OFF)
+    option(ENABLE_INTERNAL_FFMPEG "Enable internal ffmpeg?" OFF)
```
fetzerch: The module is quite difficult to read, maybe it's a good idea to have a comment in the module that explains the default behavior. Assuming I build on linux without depends installed, it would now try to find ffmpeg 3.1, and because I don't have that installed, it would build the kodi ffmpeg version for me, correct?

fetzerch (commented):
cmake wise it looks good to me, not yet tested

project/cmake/modules/FindFFMPEG.cmake
wsnipex: I had that first, but don't really like the cmake Warning. It looks too much like an error.

project/cmake/CMakeLists.txt
wsnipex: correct. Will add a comment

wsnipex:
addressed comments
one last time: jenkins build this please

wsnipex:
heads up: I'll merge this in 24h.

PR #10480, opened 2016-09-17
[PVR] Fix PVR not working after using kodi login screen (trac#16813)

ksooo:
Problem reported in trac http://trac.kodi.tv/ticket/16813 and the forum http://forum.kodi.tv/showthread.php?tid=282907&pid=2378582#pid2378582

Cause: Regression introduced by PVRManager refactoring for Krypton.
- first commit is basically preparation for the actual bug fix, to be able to do it in clean way
- second commit is the fix
- third commit fixes a race/deadlock that showed up after the login screen bug was fixed

@Jalle19 mind taking a look.

### DIFF
--- a/xbmc/pvr/windows/GUIWindowPVRBase.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRBase.cpp
@@ -115,6 +115,9 @@ void CGUIWindowPVRBase::UnregisterObservers(void)
 
 void CGUIWindowPVRBase::Notify(const Observable &obs, const ObservableMessage msg)
 {
+  if (msg == ObservableMessageManagerStopped)
+    ClearData();
+
   if (IsActive())
   {
     CGUIMessage m(GUI_MSG_REFRESH_LIST, GetID(), 0, msg);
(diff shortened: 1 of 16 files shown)


Jalle19 (approved):
Trying this new review feature. +1

ksooo:
jenkins build this please

xbmc/pvr/windows/GUIWindowPVRBase.cpp
```diff
@@ -1047,3 +1053,29 @@ bool CGUIWindowPVRBase::ConfirmStopRecording(const CPVRTimerInfoTagPtr &timer)
...
+{
+  if (!m_progressHandle)
+  {
+    CGUIDialogExtendedProgressBar *loadingProgressDialog = dynamic_cast<CGUIDialogExtendedProgressBar *>(g_windowManager.GetWindow(WINDOW_DIALOG_EXT_PROGRESS));
+    if (!loadingProgressDialog)
+    {
+      CLog::Log(LOGERROR, "CGUIWindowPVRBase - %s - unable to get WINDOW_DIALOG_EXT_PROGRESS!", __FUNCTION__);
+      return;
+    }
+    m_progressHandle = loadingProgressDialog->GetHandle(g_localizeStrings.Get(19235)); // PVR manager is starting up
```
Jalle19: Is `GetHandle()` guaranteed to not return null?

ksooo: yep, it is

PR #10481, opened 2016-09-17
Fix "Custom" view mode not showing in view mode list

nicklauslittle:
This fixes at least part of http://trac.kodi.tv/ticket/16904. I'm not sure it's feasible to show the "Custom" label without putting it in the list. If somebody knows otherwise, let me know.


FernetMenta:
thanks!

PR #10482, opened 2016-09-17
[cmake] Fix LIRC_DEVICE option

fetzerch:
Fix handling of quotes in LIRC_DEVICE option. Currently if a user wants to set LIRC_DEVICE quotes have to be escaped.

Code is used for linux, rbpi and bsd, so keep it in the main CMakeLists.txt similar to other options.

Fixes: http://trac.kodi.tv/ticket/16860.


akva2:
this doesn't fix anything, in fact it will break the build.

fetzerch:
This used to be necessary in old cmake versions (https://cmake.org/cmake/help/v3.0/policy/CMP0005.html). I've tested it here and it works for me. What version are you running at the moment? Is there any warning about this policy?

i (usually) don't PR without at least compile testing it locally ;)

akva2:
after first trying to track down the why and then grumping over the stupid behavior added to the build system, forcing a build of friggan ffmpeg to check the lirc code,

```
OPENSSL=1 -DUSE_UPNP=1 -DHAVE_XBMC_NONFREE=1 -DHAS_DVD_DRIVE -DLIRC_DEVICE=\"\"/dev/lircd\"\" -DHAVE_LIRC=1 -DHAVE_SDL=1 -DHAVE_SDL_VERSION=2 -DHAVE_LIBGL=1 -DHAVE_X11=1 -DHAVE_LIBXRANDR=1 -DHAVE_LIBEGL=1 -DBIN_INSTALL_PATH=\"/usr/local/lib/kodi\" -DINSTALL_PATH=\"/usr/local/share/kodi\" -std=c++11 -o CMakeFiles/input_linux.dir/LIRC.cpp.o -c /home/akva/kode/kodi-cmake/xbmc/input/linux/LIRC.cpp
/home/akva/kode/kodi-cmake/xbmc/input/linux/LIRC.cpp: In constructor ‘CRemoteControl::CRemoteControl()’:
/home/akva/kode/kodi-cmake/xbmc/input/linux/LIRC.cpp:54:21: error: ‘dev’ was not declared in this scope
   , m_deviceName(LIRC_DEVICE)
                     ^
/home/akva/kode/kodi-cmake/xbmc/input/linux/LIRC.cpp:54:25: error: ‘lircd’ was not declared in this scope
   , m_deviceName(LIRC_DEVICE)
                         ^
/home/akva/kode/kodi-cmake/xbmc/input/linux/LIRC.cpp: In member function ‘void CRemoteControl::SetDeviceName(const string&)’:
/home/akva/kode/kodi-cmake/xbmc/input/linux/LIRC.cpp:115:21: error: ‘dev’ was not declared in this scope
     m_deviceName=LIRC_DEVICE;
                     ^
/home/akva/kode/kodi-cmake/xbmc/input/linux/LIRC.cpp:115:25: error: ‘lircd’ was not declared in this scope
     m_deviceName=LIRC_DEVICE;
                         ^
```

this is not all that wierd, considering  the rather funky value; ```-DLIRC_DEVICE=\"\"/dev/lircd\"\"````
i figured this is could be due to reconfiguring an old build directory, so i cleaned and indeed then it worked. so yes, this works due to a cmake change, but it can confuse peeps in the process.

i get no policy warnings, only a warning about my 3.5.1. i guess there is a policy statement somewhere.

fetzerch:
Since the behaviour is fixed long time ago (and covered by our minimum cmake version which is 3.1) I still think it's a good idea to get rid of those hacks. Overall even such small improvements make the whole system (being it just a bit) more readable.

Can we live with the fact that devs have to clean the build dir from time to time?
Any strong objections? Also poking @hudokkow and @wsnipex again.

akva2:
No. It confused me was my main point. I tested, it failed and i jumped a bit quick on the solution expecting it to break in the 'old' way.

wsnipex:
LGTM

hudokkow:
No objections!

fetzerch:
Alright, thanks for the confirmation.

jenkins build this please

fetzerch:
test failure in osx unrelated

PR #10484, opened 2016-09-18
Picture improvements

garbear:
As discussed at devcon, we really want to get as many people as possible involved in improving the quality and architecture of our code. New features are awesome, but the legacy code from our early days is making it hard to properly implement these features. The pictures code is one of the best examples of an area with lots of low-hanging fruit, so I figured I'd start there with some general cleanup.

### DIFF
--- a/xbmc/pictures/GUIViewStatePictures.cpp
+++ b/xbmc/pictures/GUIViewStatePictures.cpp
@@ -82,8 +82,20 @@ std::string CGUIViewStateWindowPictures::GetExtensions()
 VECSOURCES& CGUIViewStateWindowPictures::GetSources()
 {
   VECSOURCES *pictureSources = CMediaSourceSettings::GetInstance().GetSources("pictures");
+
+  // Guard against source type not existing
+  if (pictureSources == nullptr)
+  {
+    static VECSOURCES empty;
+    return empty;
+  }
+
+  // Picture add-ons
   AddAddonsSource("image", g_localizeStrings.Get(1039), "DefaultAddonPicture.png");
+
+  // Global sources
   AddOrReplace(*pictureSources, CGUIViewState::GetSources());
+
   return *pictureSources;
 }
 
--- a/xbmc/pictures/GUIWindowPictures.cpp
+++ b/xbmc/pictures/GUIWindowPictures.cpp
@@ -582,7 +582,7 @@ void CGUIWindowPictures::LoadPlayList(const std::string& strPlayList)
 
 void CGUIWindowPictures::OnItemInfo(int itemNumber)
 {
-  CFileItemPtr item = (itemNumber >= 0 && itemNumber < m_vecItems->Size()) ? m_vecItems->Get(itemNumber) : CFileItemPtr();
+  CFileItemPtr item = m_vecItems->Get(itemNumber);
   if (!item)
     return;
   if (!m_vecItems->IsPlugin() && (item->IsPlugin() || item->IsScript()))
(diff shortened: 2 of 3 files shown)


xbmc/pictures/GUIWindowPictures.cpp
```diff
@@ -602,9 +602,11 @@ void CGUIWindowPictures::OnItemInfo(int itemNumber)
 
 std::string CGUIWindowPictures::GetStartFolder(const std::string &dir)
 {
-  std::string lower(dir); StringUtils::ToLower(lower);
-  if (lower == "plugins" || lower == "addons")
+  if (StringUtils::EqualsNoCase(dir, "plugins") ||
```
xhaggi: you've changed the implementation from equals to equals no case and not only the readability ;)

xhaggi: not sure if we really want it to be changed

garbear: Doesn't the `StringUtils::ToLower(lower)` have the effect of a case-insensitive comparison? So this isn't a change in functionality, is it?

xhaggi: ohh sorry for the noise, only saw the conditions changed 😀

xbmc/pictures/GUIWindowPictures.cpp
```diff
@@ -607,9 +602,11 @@ void CGUIWindowPictures::OnItemInfo(int itemNumber)
...
+      StringUtils::EqualsNoCase(dir, "addons"))
+  {
```
Montellese: Why the curly braces? That doesn't fit our coding style.

xbmc/windows/GUIMediaWindow.cpp
```diff
@@ -1990,9 +1990,12 @@ bool CGUIMediaWindow::Filter(bool advanced /* = true */)
 
 std::string CGUIMediaWindow::GetStartFolder(const std::string &dir)
 {
-  std::string lower(dir); StringUtils::ToLower(lower);
-  if (lower == "$root" || lower == "root")
+  if (StringUtils::EqualsNoCase(dir, "$root") ||
+      StringUtils::EqualsNoCase(dir, "root"))
+  {
```
Montellese: Same as above with curly braces.

Montellese (changes requested):
Looks good apart from the cosmetics.

xbmc/pictures/GUIViewStatePictures.cpp
```diff
@@ -82,8 +82,20 @@ std::string CGUIViewStateWindowPictures::GetExtensions()
 VECSOURCES& CGUIViewStateWindowPictures::GetSources()
 {
   VECSOURCES *pictureSources = CMediaSourceSettings::GetInstance().GetSources("pictures");
+
+  // Guard against source type not existing
+  if (pictureSources == nullptr)
```
Paxxi: You do not need the explicit nullptr check, `if(pictureSources) ` is fine

Paxxi (changes requested):
Add the braces montellese asks for and it should be good

xbmc/pictures/GUIViewStatePictures.cpp
garbear: i'm an explicit programmer

Paxxi: Yup, both are fine, just a matter of taste

garbear:
braces removed. jenkins build this please

PR #10485, opened 2016-09-18
Fix Music Roles Node passing of options string

DaveTBlake:
A simple fix to the Roles node, other options were not being correctly passed through in the way they are for the genres node etc. Not that anyone is using it yet to notice!

Addition of GetRoleById allows the select role to be displayed e.g. "Library - Roles - Composer" when drilling down to a list of aritsts with that role from the roles node (previously is just said "Artists")

### DIFF
--- a/xbmc/music/MusicDatabase.cpp
+++ b/xbmc/music/MusicDatabase.cpp
@@ -3615,9 +3615,11 @@ bool CMusicDatabase::GetRolesNav(const std::string& strBaseDir, CFileItemList& i
       CFileItemPtr pItem(new CFileItem(labelValue));
       pItem->GetMusicInfoTag()->SetTitle(labelValue);
       pItem->GetMusicInfoTag()->SetDatabaseId(m_pDS->fv("role.idRole").get_asInt(), "role");
-
-      std::string artistrolepath = StringUtils::Format("musicdb://artists/?roleid=%i", m_pDS->fv("role.idRole").get_asInt());
-      pItem->SetPath(artistrolepath);
+      CMusicDbUrl itemUrl = musicUrl;
+      std::string strDir = StringUtils::Format("%i/", m_pDS->fv("role.idRole").get_asInt());
+      itemUrl.AppendPath(strDir);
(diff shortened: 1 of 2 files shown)


Paxxi (approved):

phate89 (approved):
The code is ok.. Not sure if it fixes the problem since I don't know well music part well

DaveTBlake:
jenkins build this please

xbmc/music/MusicDatabase.cpp
```diff
@@ -3615,9 +3615,11 @@ bool CMusicDatabase::GetRolesNav(const std::string& strBaseDir, CFileItemList& i
       CFileItemPtr pItem(new CFileItem(labelValue));
       pItem->GetMusicInfoTag()->SetTitle(labelValue);
       pItem->GetMusicInfoTag()->SetDatabaseId(m_pDS->fv("role.idRole").get_asInt(), "role");
-
-      std::string artistrolepath = StringUtils::Format("musicdb://artists/?roleid=%i", m_pDS->fv("role.idRole").get_asInt());
-      pItem->SetPath(artistrolepath);
+      CMusicDbUrl itemUrl;
```
Montellese: Shouldn't the resulting path be based on `strBaseDir`? At least that's how it works for videodb stuff. But I don't know what the base directory is.

DaveTBlake: In this case no.  The path to the roles node is `musicdb://roles/` that is what will be in strBaseDir.  But the required role ID, when navigating from a role, is passed as an option value rather than extend the existing 4 parameter (idGenre, IdArtist, idAlbum, IdSong) system. That way both genre and role can be used together.

The items of the roles node are given a path of musicdb://artists/?roleid=n plus any other options from strBaseDir

DaveTBlake: As follow up to this I found how to use path based on strBaseDir to make it more consistent with other node handling.  But still adding roleid as option immediately rather than extend the parameter system.

PR #10486, opened 2016-09-18
Recently Played Albums exclude singles

DaveTBlake:
Fix GetRecentlyPlayedAlbums to only return albums, not the fake entries (releasetype = 'single') made for singles. Looking back we have never done this, although it seems odd to allow singles to come up as "played albums"

@HitcherUK would you like to try this out? You mentioned it on Slack.

It will also impact what is returned by JSON API CAudioLibrary::GetRecentlyPlayedAlbums so I am wondering about an API minor version bump?

Worth noting that lastplayed needs adding as a filter/playlist rule and JSON API filter so that users can create/adjust what recently played things they see as custom nodes etc. But that is part of a bigger change.


HitcherUK:
Will test today, thanks.

DaveTBlake:
Went for a minor bump to API - well that is what the version number is for, showing that there has been a behind the scenes correction to what data is being returned.

HitcherUK:
All good -

Played some singles - not added to Recently Played Albums
Played some albums - added to Recently Played Albums

+1

DaveTBlake:
jenkins build this please

DaveTBlake:
Thanks @HitcherUK 
Have checked the modified SQL on MySQL too, all fine.

PR #10487, opened 2016-09-18
[estuary] avoid hyphens in front of temperatures

ronie:
some people may interpret them as a minus sign.

@phil65  @chewitt


phil65:
Thx!

PR #10489, opened 2016-09-18
[cmake/addons] Fix packaging of binary addons on android 

fetzerch:
@MartijnKaijser: Found the commit that broke the binary addon packaging on android. https://github.com/xbmc/xbmc/commit/441c73f714386fe22ba8937b29590b64475e646a 
CMAKE_SYSTEM_NAME is a cmake variable that is set to Linux for android as well. To better distinguish, we've introduced CORE_SYSTEM_NAME.

@wsnipex: I'm not sure if the fix is logically correct. It would probably also work to just set OVERRIDE_PATHS.

build triggered: http://jenkins.kodi.tv/job/Android-ARM/9231/


wsnipex:
thanks, fine for me

mrdominuzq:
Pretty New to cmake but what about STREQUAL "(linux|rbpi|freebsd)") 

Doesnt it so The same :)?

wsnipex:
STREQUAL looks for the literal string, MATCHES or REGEX MATCH would work though.
The way it is now is probably more explicit to read though.

mrdominuzq:
Make sense :)

PR #10492, opened 2016-09-19
[CEC] fixes and inconsistencies [WIP]

Glenn-1990:
Current problems with cec implementation:
1) built in function "CECStandby" not working when the option to power down devices on startup isn't set. 
2) built in function "CECToggleState" not working
3) many low power devices can't suspend although we try to do that when the TV powers off
4) resuming from standby ignores the activate source setting and 
5) "bPowerOnScreensaver" is behaving different than "bPowerOffScreensaver"  as "bPowerOnScreensaver" is only working when the activate source option is enabled
6) "bPowerOffDevicesOnStandby" is not sent to libcec
7) According to libcec, "wakedevices" should not power on devices when starting. Wakedevices defines a list of devices allowed to power op when calling "PowerOnDevices()".
8) Strings in settings dialog are sometimes unclear and not ordered properly.
example: the "WakeDevices" setting is listed as "Devices to power off during shutdown" in Kodi while it actually means "List of devices to power off when calling StandbyDevices() without any parameter" according to libcec. So basically this is a list of devices which may be powered down.
9) COM port setting in the settings dialog isn't used in Kodi.
10) when the screensaver and stop playback on inactive source settings are enabled (power on/off devices disabled), the following looping occurs: screensaver activates and sending inactive source -> tv sending source deactivated -> kodi send stop command -> stop command wakes screenserver -> source becomes active again -> screensaver becomes active again -> ...
11) The activate source setting will also wake up TV's regardless of the "wakedevices" setting in libcec. in the other hand, the inactive source setting does not power off the TV -> inconsistency 
12) sometimes a Kodi restart is needed in order to pickup changed settings

What is changed with this PR:
1) renamed and changed "PowerOffDevices" setting like it should be according to libcec, it should define what devices we may power down. 
2) base the toggle action on the power state of the TV when this one is set in "poweroffdevices". When power off isn't allowed, toggle will activate/deactivate the source.
3) When suspend isn't possible, the existing pause/stop media playback will be used.
4) add check for this setting
5) Removed activate source check
6) "bPowerOffDevicesOnStandby" will be sent to libcec now
7) renamed and changed "wakedevices" setting like it should be according to libcec, it should define what devices we may power up. Currently Kodi was misusing this setting as a "power up these devices on Kodi startup" functionality. The activate source setting will be used for that.
8) 9) reworked strings and settings order
10) fixed
11) Not fixed, not sure what to do here...
Libcec says "Make libCEC the active source when starting the client application", but currently kodi will not only activate but also power up devices. We can check for the power status or such.
12) not fixed

@opdenkamp 
Several questions for you:
a) PowerOnDevices(CECDEVICE_BROADCAST) will power on devices listed in "wakedevices" but what will ActivateSource() do, only wake up the TV or multiple devices? And should it power up devices when "wakedevices" is empty.
b) What do you think about issue 11?
c) Do you agree that "poweroffdevices" and "wakedevices" should define the devices which can be powered up/down by Kodi. So when these are set, the built in commands CECToggleState and such should work properly independent from other settings. Also setting these should not lead to powering up/down devices automatically when all other settings (activate/deactivate source, screensaver, standby,..) are unset? 

PR was tested on raspbian with a LG, Sony and Philips TV (without AVR). PR not ready to merge as some issues are not resolved yet.

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -16264,12 +16264,12 @@ msgstr ""
 
 #: system/peripherals.xml
 msgctxt "#36007"
-msgid "Devices to power on during startup"
+msgid "Allowed devices to power on"
 msgstr ""
 
 #: system/peripherals.xml
 msgctxt "#36008"
-msgid "Devices to power off during shutdown"
+msgid "Allowed devices to power off"
 msgstr ""
 
 #: system/peripherals.xml
@@ -16321,7 +16321,7 @@ msgstr ""
 
 #: system/peripherals.xml
 msgctxt "#36020"
-msgid "Switch source to this device on startup"
+msgid "Switch source to this device on startup/media playback"
 msgstr ""
 
 #: system/peripherals.xml
@@ -16351,7 +16351,7 @@ msgstr ""
 
 #: system/peripherals.xml
 msgctxt "#36026"
-msgid "Devices to also put in standby mode"
+msgid "Put devices in standby mode when this system enters standby mode"
 msgstr ""
 
 #: xbmc/peripherals/devices/PeripheralCecAdapter.cpp
--- a/system/peripherals.xml
+++ b/system/peripherals.xml
@@ -10,28 +10,31 @@
 
   <peripheral bus="cec" name="CEC Adapter" mapTo="cec">
     <setting key="enabled" type="bool" value="1" label="305" order="1" />
-    <setting key="activate_source" type="bool" value="1" label="36020" order="2" />
-    <setting key="wake_devices" type="enum" value="36037" label="36007" lvalues="36037|36038|36039|231" order="3" />
-    <setting key="standby_devices" type="enum" value="36037" label="36008" lvalues="36037|36038|36039|231" order="4" />
-    <setting key="send_inactive_source" type="bool" value="1" label="36025" order="5" />
-    <setting key="cec_standby_screensaver" type="bool" value="0" label="36009" order="6" />
-    <setting key="cec_wake_screensaver" type="bool" value="1" label="36010" order="7" />
-    <setting key="standby_pc_on_tv_standby" type="enum" value="13011" label="36029" order="8" lvalues="36028|13005|13011" />
-    <setting key="standby_tv_on_pc_standby" type="bool" value="1" label="36026" order="9" />
-    <setting key="use_tv_menu_language" type="bool" value="1" label="36018" order="10" />
-    <setting key="pause_playback_on_deactivate" type="bool" value="1" label="36033" configurable="0" />
-    <setting key="pause_or_stop_playback_on_deactivate" type="enum" value="231" label="36033" order="11" lvalues="231|36044|36045" />
-    <setting key="connected_device" type="enum" label="36019" value="36037" lvalues="36037|36038" order="12" />
-    <setting key="cec_hdmi_port" type="int" value="1" min="1" max="15" label="36015" order="13" />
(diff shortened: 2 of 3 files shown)


opdenkamp:
> Several questions for you:
> a) PowerOnDevices(CECDEVICE_BROADCAST) will power on devices listed in "wakedevices" but what will ActivateSource() do, only wake up the TV or multiple devices? And should it power up devices when "wakedevices" is empty.

There's a little misunderstanding here how it's supposed to work with CEC.

Kodi is a source device, and the only things that a standard source device does in CEC for power control are <activate source>, <device power status> and <inactive source>.
<activate source> tells the other CEC devices that we'd like to become the displayed source in the chain. This will power on the TV, switches between the source and TV, and the TV will power on your AVR if you've configured it to do this.
<device power status> tells the TV that the source powered on or off, and the TV may decide what to do with that information.
<inactive source> tells the TV that we no longer want to be the displayed source, and the TV may decide what to do with that again.

This generally doesn't work well for anything except <activate source>, as this results in different behaviour between vendors and often models from the same vendor too.
Because Kodi is the only media player connected the TV for lots of our users, Kodi is configured to power off all HDMI devices when the PC powers off or goes to standby.

The "wakedevices" setting just tells libCEC which devices to power on explicitly, if they're not woken up by <activate source>.

> b) What do you think about issue 11?

This is not an issue, see previous.

> c) Do you agree that "poweroffdevices" and "wakedevices" should define the devices which can be powered up/down by Kodi. So when these are set, the built in commands CECToggleState and such should work properly independent from other settings. Also setting these should not lead to powering up/down devices automatically when all other settings (activate/deactivate source, screensaver, standby,..) are unset?

This can't work, because <activate source> will override this. <activate source> must be used, or switches and sinks won't switch to the correct inputs and the remote control won't work otherwise.

> What is changed with this PR:
> 1) renamed and changed "PowerOffDevices" setting like it should be according to libcec, it should define what devices we may power down.
> 2) base the toggle action on the power state of the TV when this one is set in "poweroffdevices". When power off isn't allowed, toggle will activate/deactivate the source.

NAK, active source status is primary

> 3) When suspend isn't possible, the existing pause/stop media playback will be used.

Please change the settings instead if it's not supported, same as is done with Kodi's shutdown menu.

> 4) add check for this setting

What setting?

> 5) Removed activate source check

See 2, active source is primary in CEC

> 6) "bPowerOffDevicesOnStandby" will be sent to libcec now
> 7) renamed and changed "wakedevices" setting like it should be according to libcec, it should define what devices we may power up. Currently Kodi was misusing this setting as a "power up these devices on Kodi startup" functionality. The activate source setting will be used for that.

The current behaviour in Kodi (at least when I last touched it) is correct, it's meant to explicitly wake those devices that are not covered by <activate source>

Glenn-1990:
@opdenkamp thank for explanation, is there a way to not automatically power up/down devices while maintaining working builtin CEC commands? User case: "CECToggleState" can be mapped to the power button of the remote or the user can have an add on to "CECStandby" or "CECActivateSource" on some Kodi actions. 

> The current behaviour in Kodi (at least when I last touched it) is correct, it's meant to explicitly wake those devices that are not covered by <activate source>

Yes, on Kodi start, Kodi will call  \<activate source\>. When there are multiple power on devices defined, this will be followed by  \<power on devices (broadcast)\>. On Kodi shutdown \<standby devices (broadcast)\> will be called, this seems all correct to me. But, when deactivating the screensaver or waking from sleep, only \<activate source\> gets called, while activating the screensaver or sleep will shutdown all devices configured. \<power on devices (broadcast)\> is missing here for all devices other than TV?

Glenn-1990:
replaced by: https://github.com/xbmc/xbmc/pull/10512

PR #10495, opened 2016-09-19
[cmake/osx] Fix generating Xcode projects

fetzerch:
Fixes broken Xcode project generation after https://github.com/xbmc/xbmc/pull/10372.

`core_add_library` does a different thing depending on the generator. For makefiles, it creates a new static library target (everytime it gets called). This leads to hunderts of small libraries (one per each directory). For Xcode or VS there is one libkodi static library and `core_add_library` only appends to that one. This was done to get a decent project structure in the IDEs. Using the same thing for makefiles leads to >60s linking time (every time you touch only a single line).

So in CMakeLists.txt files inside of the xbmc directory, make sure to always use the `CORE_LIBRARY` variable when specifying compile options or dependencies. 

Unfortunately this is currently not very well documented, and I expect errors like this to appear from now and then. Don't yet have a good idea how to prevent them. One thing I could imagine is to have a script that runs some checks on the project files in CI.


fetzerch:
@hudokkow, @Paxxi: Ping just to make sure that you're aware of this black magic

Paxxi:
Thanks, that's some non obvious behavior :)

PR #10496, opened 2016-09-19
Fix spelling error in german skin.estuary strings.po

seahawk1986:
A small fix for a spelling error in german translation file for the estuary skin.


phil65:
Thx for trying to contribute, but we do translations via transifex nowadays, not via direct pull requests.
(--> https://www.transifex.com/teamxbmc/kodi-main/ )

PR #10497, opened 2016-09-19
[mingwlibs] a set of small fixes for ffmpeg build proc.

afedchin:
This fixes several issues with ffmpeg build for windows:
- [x] Fixed build ffmpeg via MSVC toolchain
- [x] Fixed build hanging if ffmpeg build failed.
- [x] Added liblzma to dependencies which adds lzma support to ffmpeg.
- [x] Bump gnutls version which speed up build (require upload builded lib to our mirror)

@Paxxi @FernetMenta take a look please.


afedchin:
jenkins build this please

tools/buildsteps/win32/fmpeg_options.txt
```diff
@@ -18,3 +18,4 @@
 --enable-encoder=mjpeg
 --enable-protocol=http
 --enable-runtime-cpudetect
+--disable-dxva2
```
FernetMenta: @afedchin why do you disable dxva2 here?

fritsch: We use: d3d11va afaik --dxva2 is the old dx9 method, right?

afedchin: actually it's not needed anymore since we switched to d3d11va

afedchin: @fritsch you are right. dxva2 is old dx9 api

Paxxi:
Got cut of, was doing review on the plane.  Looks good, will try to test it out tomorrow

akva2:
probably for later, but wouldn't moving it from buildsteps to main project make more sense now that it builds with msvc? it would maximize the code shared with the other platforms and such.

Paxxi:
@akva2 yes that is what I think as well, we still need msys2 for building it but can link with our openssl and zlib and such minimizing depends and downloads and speeding up jenkins builds. Other upside is easier debugging with proper pdb  files. 
Should be a goal for v18

akva2:
all game add-ons build in a mingw environment orchestrated by cmake. i believe @montellese make it flexible enough to be reused anywhere.

Paxxi (changes requested):
Seems to have built fine now. Thank you for fixing this.

Paxxi:
@akva2 right, I didn't mean to get rid of msys or mingw but that we should build ffmpeg with msvc by default. Would probably require splitting make-mingwlibs into two, one for ffmpeg and one for libdvd as they need different compilers and paths set up.

afedchin:
jenkins build this please

Voyager1:
can someone take care of the missing gnutls dependency on the mirror?

afedchin:
@MartijnKaijser please upload this package https://www.dropbox.com/s/tzizdldyvp0dn7p/i686-gnutls-3.4.14-static.tar.gz?dl=0 to the http://mirror.kodi.tv/build-deps/win32/msys2/locals/

MartijnKaijser:
@Voyager1 @afedchin done (takes about 30 minutes from now)

Voyager1:
thanks!

PR #10498, opened 2016-09-19
Mention-Bot: Let's use a more sensible default

fritsch:
More sensible defaults for Mention-Bot:
- blacklist fernetmenta, he does not want to receive messages
  ~~-only send messages to xbmc team members~~ <- let's try it first to also get new devs from outside that contributed before

### DIFF
--- a/.mention-bot
+++ b/.mention-bot
@@ -1,0 +1,18 @@
+{
+  "maxReviewers": 3, // Maximum  number of people to ping in the PR message, default is 3
+  "numFilesToCheck": 5, // Number of files to check against, default is 5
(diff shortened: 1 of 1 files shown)


kib (approved):
Looks very reasonable.

There are a few other settings that might be experimented with later.

piejanssens:
Didn't you forget setting 'requiredOrgs'?

fritsch:
Nope. I wanted, but it was internally decided that we now ping everyone and add those that don't want to be mentioned, e.g. not continue contributing to our blacklist.

This is just trial and error to see if this bot is useful for us. We will evaluate it for some days / weeks and decide afterwards if it is useful.

PR #10500, opened 2016-09-19
RenderManager: Bypass renderers need updating even when renderer has …

popcornmix:
Otherwise subs and overlays don't get updated properly.
See: http://forum.kodi.tv/showthread.php?tid=269814&pid=2416492#pid2416492


popcornmix:
Ping @FernetMenta

FernetMenta:
that does not seem to be correct. presentstep states are valid for bypass renderers too.

popcornmix:
Closing in favour of #10501

PR #10501, opened 2016-09-19
VideoPlayer: drop obsolete player method HasFrame, fix subs for bypas…

FernetMenta:
follow-up from https://github.com/xbmc/xbmc/pull/10500

@popcornmix I think this is a cleaner fix


popcornmix:
+1 fixes the omxplayer subs issue

FernetMenta:
jenkins build this please

PR #10502, opened 2016-09-19
VideoPlayer: make processInfo a smart pointer, fixes mem leak

FernetMenta:
fix mem leak

credits to @popcornmix for bringing this up


FernetMenta:
jenkins build this please

PR #10504, opened 2016-09-19
Mention-Bot: Config File should be proper json

fritsch:
mention-bot configuration needs to be proper json. The should write that more "fat" on their Readme


kib (approved):
Bye bye comments in the config :)
Valid JSON.... eugh

PR #10506, opened 2016-09-19
[PVR][guiinfo] Optimize implementation of System.HasPVRAddon. Also fixes possible deadlock on PVR startup.

ksooo:
Optimize implementation of System.HasPVRAddon. Also fixes possible deadlock on PVR startup.

Fixes the deadlock reported here: https://github.com/xbmc/xbmc/pull/10472#issuecomment-247881160

@FernetMenta does the code change look okay to you?
@MilhouseVH maybe you can put this patch into your build and report back?

### DIFF
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -90,6 +90,7 @@
 #include "pvr/recordings/PVRRecording.h"
 
 #include "addons/AddonManager.h"
+#include "addons/BinaryAddonCache.h"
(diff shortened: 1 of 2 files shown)


MilhouseVH:
Will do, thanks.

FernetMenta:
looks ok to me

MilhouseVH:
Positive reply: http://forum.kodi.tv/showthread.php?tid=269814&pid=2418860#pid2418860

ksooo:
jenkins build this please

PR #8953, opened 2016-01-24
[PVR] handle tv<->radio channel change in database, fixes track #16031

Glenn-1990:
There are lots of reports of not starting PVR manager and I was also facing this from time time time.
After a bit of debugging, I figured out where the problem lies and how to reproduce.

1) the problem
- Assume the backend has a tv channel with an unique id of "1234".
- So the channel is stored in the Kodi database as tv channel with an unique id of "1234".
- Assume kodi is not running and the user is adding and deleting channels on the backend.
  Let's say we delete the tv channel with id "1234", after this we create a new radio channel, it's possible that the backend assigns id "1234" to that radio channel (as it's free to use now).
- Now we fire up kodi, kodi will first load radio channels (by design)

-> radio channels are loaded from db
-> non existing radio channels (on backend) are deleted from the db
-> new radio channels are added to the db, so radio channel "1234" gets added, but there is still a tv channel in the db with id "1234" -> clash -> PVR manager stops here. 
-> Assume is we still continue (in our mind).
-> tv channels are loaded from db
-> non existing tv channels (on backend) are deleted from the db, so tv channel "1234" gets deleted.
-> new tv channels are added.
- Following this loading order, it's obvious that this only happens when a channel changes from tv to radio and not the other way around.

Some users are reporting that they need to clear the tv database every few hours because of this. 
https://github.com/kodi-pvr/pvr.hts/issues/172

This can be explained that for example some channels are not broadcasting video all the time and are handled as radio, when a video pid comes available, they are television. Also bouquet updates can easily mess things up like this. I mainly see reports when using tvheadend or vdr. 

Some other reports that that are caused by this issue (notice that bIsRadio flag is always "1"):
http://trac.kodi.tv/ticket/16031
http://forum.kodi.tv/showthread.php?tid=189534&page=5
http://forum.kodi.tv/showthread.php?tid=233588&pid=2067509#pid2067509
http://forum.kodi.tv/showthread.php?tid=235246

2) The solution:
- One way to handle this is to check all channels in one time instead (radio+tv), 
  but it's hard to this this with the current way kodi handles channel groups.
- Another way is to let radio and tv channels with the same unique id coexist in the database, this is what I've don in this PR. I'm not sure if this is the best way to handle this.
- If someone has a better alternative?

@ksooo @Jalle19 @xhaggi @opdenkamp

PR #10507, opened 2016-09-19
Fix potential SQL constraint integrity violation in the PVR database

### LINKED ISSUE #172: CPVRChannelGroup - Persist: SQL constraint violation
Loading the hts plugin will frequently on reboot and the screen gets stuck in 'Starting PVR manager'.
I'm using TVHeadend with two satellite networks (Astra + Hotbird) so there are plenty of services and channels mapped. It seems the issue occurs when a video PID is missing and a channel that was classified 'tv' is changed to 'radio' or vice versa. This happens every few hours so I have to clear the TV DB on every reboot.

In this such a case the logs would give me an error like this:
```
22:13:46 T:139793007634176   ERROR: SQL: Abort due to constraint violation
        Query: INSERT INTO channels (iUniqueId, bIsRadio, bIsHidden, bIsUserSetIcon,
        bIsUserSetName, bIsLocked, sIconPath, sChannelName, bIsVirtual, bEPGEnabled, 
        sEPGScraper, iLastWatched, iClientId, idEpg) VALUES (1800902271, 1, 0, 0, 0, 0,
        'http://192.168.10.2:9981/imagecache/1577', 'Bahai Radio', 0, 1, 'client', 0, 1600, -1)
22:13:46 T:139793007634176   ERROR: CommitInsertQueries - failed to execute queries
```

At the same time TV29.db already contains the following entry in table channels:
```
sqlite> select * from channels where iUniqueId=1800902271;
1711|1800902271|0|0|0|0|0|http://192.168.10.2:9981/imagecache/1577|Bahai Radio|0|1|client|0|1600|1711
```

The records only differ in bIsRadio (insert statement wants to set it 1, exists as 0) and idEpg (insert: -1, exists: 1711). The error gets thrown because the records would share the same values for iUniqueId and iClientId (unique index).

Environment:
ArchLinux 4.3.3-2 x86_64
Kodi 15.2
kodi-addon-pvr-hts 2.1.18-1 (built from Isengard branch)

Note: I'm not sure if it's a bug in the addon or rather in the Kodi PVR manager but I understand that existing channels should have the m_iChannelId field populated in order to have the PVR manager call 'REPLACE INTO' rather than 'INSERT INTO'.

Jalle19:
Supersedes https://github.com/xbmc/xbmc/pull/8953
Fixes trac #16031
Closes kodi-pvr/pvr.hts#172

This fixes the problem without bumping the database version so it should be more possibly to get it merged for V17.

Basically the code that persists channels has been made smarter. Instead of just blindly looking at the primary key stored in the object we query the database to check if a channel with the unique ID and client ID already exists. If it does, we update that channel, if not we create a new channel. There is more background information in the referenced PRs.

This fix also has the added side-effect that it increases database performance. While a new SELECT statement has been added, we now do an UPDATE instead of a REPLACE INTO which prevents indexes from having to be re-calculated (REPLACE INTO basically does a DELETE followed by an INSERT). This also means we don't need to worry about the channel ID changing after the channel has been updated (which was a side effect of using REPLACE INTO).

@ksooo now's a good time to test that new review feature, @Glenn-1990 can you runtime test this? I tested with your reproducible test case (change a channel's service from TV to radio while Kodi is off) and it seems to fix the issue.

### DIFF
--- a/xbmc/pvr/PVRDatabase.cpp
+++ b/xbmc/pvr/PVRDatabase.cpp
@@ -24,6 +24,7 @@
 
 #include "dbwrappers/dataset.h"
 #include "addons/PVRClient.h"
+#include "pvr/channels/PVRChannel.h"
 #include "pvr/channels/PVRChannelGroupInternal.h"
 #include "pvr/channels/PVRChannelGroupsContainer.h"
 #include "pvr/PVRManager.h"
@@ -604,17 +605,15 @@ bool CPVRDatabase::Persist(CPVRChannelGroup &group)
 
 bool CPVRDatabase::Persist(CPVRChannel &channel)
 {
-  bool bReturn(false);
-
   /* invalid channel */
   if (channel.UniqueID() <= 0)
   {
     CLog::Log(LOGERROR, "PVR - %s - invalid channel uid: %d", __FUNCTION__, channel.UniqueID());
-    return bReturn;
+    return false;
   }
 
   std::string strQuery;
-  if (channel.ChannelID() <= 0)
+  if (!ChannelExists(channel.m_iClientId, channel.m_iUniqueId))
   {
     /* new channel */
     strQuery = PrepareSQL("INSERT INTO channels ("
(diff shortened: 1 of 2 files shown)


xbmc/pvr/PVRDatabase.cpp
```diff
@@ -625,31 +624,25 @@ bool CPVRDatabase::Persist(CPVRChannel &channel)
...
+        "iUniqueId = %i, bIsRadio = %i, bIsHidden = %i, bIsUserSetIcon = %i, bIsUserSetName = %i, bIsLocked = %i, "
+        "sIconPath= %s, sChannelName = %s, bIsVirtual = %i, bEPGEnabled = %i, sEPGScraper = %s, iLastWatched = %u, "
+        "iClientId = %i, idChannel = %i, idEpg = %i "
+        "WHERE iUniqueId = %i AND iClientId = %i",
         channel.UniqueID(), (channel.IsRadio() ? 1 :0), (channel.IsHidden() ? 1 : 0), (channel.IsUserSetIcon() ? 1 : 0), (channel.IsUserSetName() ? 1 : 0), (channel.IsLocked() ? 1 : 0),
         channel.IconPath().c_str(), channel.ChannelName().c_str(), 0, (channel.EPGEnabled() ? 1 : 0), channel.EPGScraper().c_str(), channel.LastWatched(), channel.ClientID(),
-        channel.ChannelID(),
-        channel.EpgID());
-  }
+        channel.ChannelID(), channel.EpgID(), channel.UniqueID(), channel.ClientID());
```
xhaggi: no need to update iUniqueId and iClientId as they are part of the where condition.

xhaggi (changes requested):

bkuhls:
@Jalle19: You referenced pull request 8963 but I think you mean PR 8953, right?

Jalle19:
@bkuhls you're right, I meant https://github.com/xbmc/xbmc/pull/8953

xbmc/pvr/PVRDatabase.cpp
Jalle19: You're right, that's a left over from when it used REPLACE INTO

Glenn-1990:
@Jalle19 thanks for this, now I can remove the "delete pvr database" script from my clients. I will test this tomorrow. :-)

xbmc/pvr/PVRDatabase.cpp
Jalle19: Fixed in the latest fixup commit

xhaggi (approved):
thx

Jalle19:
jenkins build this please

Memphiz:
@Jalle19 i am currently working on new jenkins slave and need the server for some more time. I have canceled your build request and will mail around once i am done.

xbmc/pvr/PVRDatabase.h
```diff
@@ -185,6 +185,14 @@ namespace PVR
...
 
     bool RemoveChannelsFromGroup(const CPVRChannelGroup &group);
+
+    /*!
+     * Checks whether the specified unique combination for a channel exists in the database
+     * @param iClientId
+     * @param iUniqueId
+     * @return
+     */
+    bool ChannelExists(int iClientId, int iUniqueId);
```
ksooo: shouldn't this member function be 'const'?

xbmc/pvr/PVRDatabase.cpp
```diff
@@ -667,3 +659,14 @@ bool CPVRDatabase::UpdateLastWatched(const CPVRChannelGroup &group)
 
   return ExecuteQuery(strQuery);
 }
+
+bool CPVRDatabase::ChannelExists(int iClientId, int iUniqueId)
+{
+  std::string strQuery = PrepareSQL("SELECT idChannel FROM channels WHERE iClientId = %d AND iUniqueId = %d",
```
ksooo: const std::string strQuery(Prepare... ?

ksooo (approved):
despite from const nitpicking this is looking good to me, though i have not runtime tested it.

xbmc/pvr/PVRDatabase.h
Jalle19: I would like that too, but all database access methods are non-const only.

xbmc/pvr/PVRDatabase.cpp
Jalle19: Sure

Jalle19:
@Memphiz okay, no problem.

Glenn-1990:
@Jalle19 run time tested: reproduced the issue and restarted kodi several times, PVR manager was not starting up, then applied your commits and it came back to life ;-)

Glenn-1990:
Btw, do we need the same for channel groups?

Jalle19:
I need to check, I've never thought about that scenario.

xbmc/pvr/PVRDatabase.h
ksooo: yikes. So, leave it like it is, i'd say.

Jalle19:
jenkins build this please

Jalle19:
jenkins build this please

Jalle19:
@ksooo good to go?

ksooo:
> Btw, do we need the same for channel groups?

 what about this? supposed to be handled in another PR?

Jalle19:
Yeah I'd say so, I'm not even sure if there's an issue in that regard yet.

Jalle19:
@ksooo the code for persisting channel groups is very similar to what we had here. I would still prefer to get this merged so I have time to think of a proper solution for the groups part, if that's okay with you?

Jalle19:
Actually the persisting of groups cannot have this exact same issue because there's no `UNIQUE INDEX` that could trigger an integrity violation. It still uses `REPLACE INTO` though which is not nice, but the performance impact is much lesser since one can assume that `num(channels) > num(groups)`.

ksooo:
> Actually the persisting of groups cannot have this exact same issue

 thanks for the update on this.

PR #10508, opened 2016-09-19
Musicroles Menu Additional default nodes 

DaveTBlake:
I would like to add some more default nodes beneath the music roles submenu item in the music library menu. This will make some of the new music library features more accessible without cluttering the menu layout for those users that don't want any of this (or have not tagged their music with composers etc., musician credits or people invloved).

This submenu currently shows nodes for the standard roles - composer, conductor, orchestra, arranger, lyricist, DJmixer,  remixer. This change adds nodes:
a) all roles - standard roles, plus musician credits e.g. "piano",  and people involved e.g. engineer
b) all contributors - every artist that Kodi knows about regardless of role they contribute to the song
c) all artists  - song and album artists 
d) album artists only 

c ) and d)  are effectively the usual artists node  but independent of the "Show song and album artists" system setting.

@jjd-uk I thought this was a gentle step towards depricating the "Show song and album artists" system setting as we discussed - the ability to switch between seeing the song artists or not without going off to the settings menu.

@ronie or @phil65 care to check this for me.

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -19464,7 +19464,31 @@ msgctxt "#38042"
 msgid "[Missing]"
 msgstr ""
 
-#empty strings from id 38043 to 38099
+#. Title album artists node
+#: system/library/music/musicroles/albumartists.xml
+msgctxt "#38043"
+msgid "Album artists"
+msgstr ""
+
+#. Title all artists node
+#: system/library/music/musicroles/allartists.xml
+msgctxt "#38044"
+msgid "Song & album artists"
+msgstr ""
+
+#. Title all contributors node (any role)
+#: system/library/music/musicroles/allcontributors.xml
+msgctxt "#38045"
+msgid "All contributors"
+msgstr ""
+
+#. Title all roles node
+#: system/library/music/musicroles/allroles.xml
+msgctxt "#38046"
+msgid "All roles"
+msgstr ""
+
+#empty strings from id 38047 to 38099
 
 #. Description of section #14200 "Player""
 #: system/settings/settings.xml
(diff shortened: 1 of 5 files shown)


ronie:
i don't have any music with additional role tags, so likely not a good candidate for testing this

DaveTBlake:
Fair enough @ronie :)
The nodes work OK, this is just making some example custom nodes available by default. So I wanted approveal that doing that was OK, and that they syntax of the xml was to standard.

Not sure who in the team has fully tagged music files to see this in action.... @zag2me or @HitcherUK ?

zag2me:
imo any new feature such as roles or whatever should be accompanied by the appropriate node folder so +1 from me. The xml looks fine.

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -19464,7 +19464,31 @@ msgctxt "#38042"
...
+#. Title all artists node
+#: system/library/music/musicroles/allartists.xml
+msgctxt "#38044"
+msgid "Song & Album Artists"
+msgstr ""
+
+#. Title all contributors node (any role)
+#: system/library/music/musicroles/allcontributors.xml
+msgctxt "#38045"
+msgid "All Contributors"
```
phil65: Didnt we decide to use lowercase except for first word?

DaveTBlake: Just my natural habit to capitalise titles, no idea of the Kodi design policy.  Was that a question for others to confirm or rhetorical (meaning change to lower case).?

DaveTBlake: Since other nodes are only first word capitals e.g. "Recently played albums", I have adjusted strings.po to match. Thanks for picking this up.

(oddly new GitHub keeps showing the now out of date line above and expanded comments)

phil65: cool, thx.

PR #10509, opened 2016-09-19
[cmake/ios] Enable Xcode generator for IOS

fetzerch:
Not yet tested, as there's still something wrong with my local ios depends.
cmake run succeeds at least. 

@Memphiz I'll let you know once I could verify the build.


Memphiz:
@fetzerch it creates the xcode project now. Next error is this:

http://pastebin.com/uYk903jQ

It somehow tries to compile an universal binary (which is also called fat binary and can be used to have multiple architectures into one dylib or executable at the same time). It fails because it tries to add the armv7 architecture into the universal binary 2 times (only one architecture can be in a fat binary once). Basically we don't want to and can't use universal binaries at all (because depends only builds for one single architecture).
Thats really strange. Wasn't cmake supposed to get this stuff right? I mean why does it work ok with makefiles but becoming such a bitch for xcode (which should only be a different generator and not change behavior at all)? ;)

akva2:
because Makefiles are sane and xcode is a huge pile of hack. /runs

Memphiz:
@akva2 yep - maybe - but cmake is supposed to abstract the huge pile of hack ;)

akva2:
touchè

MrMC:
HAHAHAHAHAHA

MrMC:
Just an FYI, you can built a universal binary. Depends supports it, Xcode supports it, just have to package it right. We did an experimental built for MrMC on iOS to support armv7 and arm64 under one binary using embedded bitcode. Was eventually abandoned as hitting armv7 devices were just not worth it.

fetzerch:
@Memphiz: Got it working now. Main issue was that for ios, xcode can only build bundles. This required to change the way I used to control the packaging in Install.cmake. Then there were a few minor issues (see commit message). I've successfully built and run the app on the ipad from xcode. In addition building the deb works now with make and xcodebuild and after signing I can install and run both.

Could you give it another try? After opening the xcode project, you still need to select the signing properties (bundle identifier, team). And since the toolchain file is touched, it needs a reconfigure of depends.

jenkins build this please with addons

Memphiz:
Nice - will give it a shot asap. Jenkins seems unhappy? Or just a fluke?

fetzerch:
Right, the changes to the toolchain must be only done when building with xcode.
The other errors seem unrelated, but let's try again.

jenkins build this please with addons

fetzerch:
Jenkins build is fine now as well (http://jenkins.kodi.tv/job/IOS/8546)

Memphiz:
Great. Will give it a try the next days and adapt the osx/ios readmes to reflect the new workflow

project/cmake/scripts/ios/ArchSetup.cmake
```diff
@@ -13,6 +13,7 @@ if(WITH_ARCH)
   set(ARCH ${WITH_ARCH})
 else()
   if(CPU STREQUAL armv7 OR CPU STREQUAL arm64)
+    set(CMAKE_OSX_ARCHITECTURES armv7)
```
Memphiz: set CMAKE_OSX_ARCHITECTURES CPU? - we want to be able to build arm64 aswell ... also where does the CPU come from? i thought cmake variables must be references by ${CPU} then?

Memphiz:
nevermind my comments - hadn't reset my branch to you remote ... works fine for me on xcode7.3 so far.

fetzerch:
arm64 build works as well now.
jenkins build this please with addons

Memphiz:
Nice :)

fetzerch:
come on jenkins, you can do it ;)

fetzerch:

fetzerch:

fetzerch:

PR #10510, opened 2016-09-19
Fix some low-hanging fruit (CLion warnings)

Jalle19:
@ksooo @xhaggi the commits that are less than completely trivial have more specific commit messages, you may want to look at those.

Briefly runtime tested, but it wouldn't hurt if someone else did it too.

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayerRadioRDS.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayerRadioRDS.cpp
@@ -819,7 +819,6 @@ void CDVDRadioRDSData::ProcessUECP(const unsigned char *data, unsigned int len)
               case UECP_RDS_PS:                 ret = DecodePS(msg);                  break;  //!< Program Service name (PS)
               case UECP_RDS_DI:                 ret = DecodeDI(msg);                  break;  //!< Decoder Identification and dynamic PTY indicator
               case UECP_RDS_TA_TP:              ret = DecodeTA_TP(msg);               break;  //!< Traffic Announcement and Traffic Programme bits.
-              case UECP_RDS_MS:                 ret = DecodeMS(msg);                  break;  //!< Music/Speech switch
               case UECP_RDS_PTY:                ret = DecodePTY(msg);                 break;  //!< Program Type
               case UECP_RDS_PTYN:               ret = DecodePTYN(msg);                break;  //!< Program Type Name
               case UECP_RDS_RT:                 ret = DecodeRT(msg, len);             break;  //!< RadioText
@@ -964,19 +963,6 @@ unsigned int CDVDRadioRDSData::DecodeTA_TP(uint8_t *msgElement)
     CVariant data(CVariant::VariantTypeObject);
     data["on"] = false;
     ANNOUNCEMENT::CAnnouncementManager::GetInstance().Announce(ANNOUNCEMENT::PVR, "xbmc", "RDSRadioTA", data);
-  }
-
-  return 4;
-}
-
-unsigned int CDVDRadioRDSData::DecodeMS(uint8_t *msgElement)
-{
-  bool speechActive = msgElement[3] == 0;
-  if (m_MS_SpeechActive != speechActive)
-  {
-    m_currentInfoTag->SetSpeechActive(m_MS_SpeechActive);
-    CLog::Log(LOGDEBUG, "Radio UECP (RDS) Processor - %s - Stream changed over to %s", __FUNCTION__, speechActive ? "Speech" : "Music");
-    m_MS_SpeechActive = speechActive;
   }
 
   return 4;
--- a/xbmc/pvr/PVREvent.h
+++ b/xbmc/pvr/PVREvent.h
@@ -27,7 +27,6 @@ namespace PVR
     ManagerError = 0,
     ManagerStopped,
     ManagerStarting,
-    ManagerStopping,
     ManagerInterrupted,
     ManagerStarted,
 
--- a/xbmc/pvr/PVRManager.h
+++ b/xbmc/pvr/PVRManager.h
@@ -72,8 +72,7 @@ namespace PVR
   enum ContinueLastChannelOnStartup
   {
     CONTINUE_LAST_CHANNEL_OFF  = 0,
-    CONTINUE_LAST_CHANNEL_IN_BACKGROUND,
-    CONTINUE_LAST_CHANNEL_IN_FOREGROUND
+    CONTINUE_LAST_CHANNEL_IN_BACKGROUND
   };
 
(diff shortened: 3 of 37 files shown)


Jalle19:
@AlwinEsch I dropped `DecodeMS` from the RDS parser since the member it populates is never used. Let me know if you still want to keep it. We have some other parts of the code where there is technically unused code but you can see that it's there for a reason (maybe for easier additions in the future) so I didn't bother with everything.

xbmc/pvr/dialogs/GUIDialogPVRGroupManager.cpp
```diff
@@ -144,7 +144,7 @@ bool CGUIDialogPVRGroupManager::ActionButtonDeleteGroup(CGUIMessage &message)
 
     if (pDialog->IsConfirmed())
     {
-      if (((CPVRChannelGroups *) g_PVRChannelGroups->Get(m_bIsRadio))->DeleteGroup(*m_selectedGroup))
+      if ((g_PVRChannelGroups->Get(m_bIsRadio))->DeleteGroup(*m_selectedGroup))
```
xhaggi: do we need the extra brackets?

xbmc/pvr/PVRManager.h
```diff
@@ -72,8 +72,7 @@ namespace PVR
   enum ContinueLastChannelOnStartup
   {
     CONTINUE_LAST_CHANNEL_OFF  = 0,
-    CONTINUE_LAST_CHANNEL_IN_BACKGROUND,
-    CONTINUE_LAST_CHANNEL_IN_FOREGROUND
+    CONTINUE_LAST_CHANNEL_IN_BACKGROUND
```
xhaggi: we use this in settings IIRC.

xbmc/pvr/PVREvent.h
```diff
@@ -27,7 +27,6 @@ namespace PVR
     ManagerError = 0,
     ManagerStopped,
     ManagerStarting,
-    ManagerStopping,
```
xhaggi: not sure if we should remove it.

xbmc/cores/VideoPlayer/VideoPlayerRadioRDS.cpp
```diff
@@ -969,19 +968,6 @@ unsigned int CDVDRadioRDSData::DecodeTA_TP(uint8_t *msgElement)
   return 4;
 }
 
-unsigned int CDVDRadioRDSData::DecodeMS(uint8_t *msgElement)
```
AlwinEsch: I understand not complete why it becomes removed? Can you describe it.

If it becomes removed are also other parts not needed anymore. The value `m_MS_SpeechActive` and also the parts from `CPVRRadioRDSInfoTag` need then removed.

To leave the parts on `CPVRRadioRDSInfoTag` and `CGUIInfoManager` make remove of them no sense.

AlwinEsch:
Hi @Jalle19, for some unknown reason has not seen you comment above ;), but has send a comment :)

With RDS is always a bit hard, all radio channels with RDS are always a bit different. But the remove of them is OK for me, also not seen direct one who use it.

Hope after I'm or someone else has created a DAB+ radio add-on can I recheck them, on DAB+ they have changed some way's for RDS not sure how it comes then.

xbmc/pvr/dialogs/GUIDialogPVRGroupManager.cpp
Jalle19: Probably not, I'll remove them

xbmc/pvr/PVRManager.h
Jalle19: Perhaps, but the value isn't used anywhere in the actual code. It's basically "if not foreground", not explicitly "is background". Not sure if we should leave this here or not.

xbmc/pvr/PVREvent.h
Jalle19: @ksooo what do you think?

xbmc/cores/VideoPlayer/VideoPlayerRadioRDS.cpp
Jalle19: I'll try to find the rest of the references to it.

xbmc/cores/VideoPlayer/VideoPlayerRadioRDS.cpp
```diff
@@ -819,7 +819,6 @@ void CDVDRadioRDSData::ProcessUECP(const unsigned char *data, unsigned int len)
               case UECP_RDS_PS:                 ret = DecodePS(msg);                  break;  //!< Program Service name (PS)
               case UECP_RDS_DI:                 ret = DecodeDI(msg);                  break;  //!< Decoder Identification and dynamic PTY indicator
               case UECP_RDS_TA_TP:              ret = DecodeTA_TP(msg);               break;  //!< Traffic Announcement and Traffic Programme bits.
-              case UECP_RDS_MS:                 ret = DecodeMS(msg);                  break;  //!< Music/Speech switch
```
AlwinEsch: If unused change it to:
´´´cpp
               case UECP_LINKAGE_INFO:           ret = 4;                              break;   //!< Music/Speech switch
´´´
To prevent stream flow pointer failure!

ksooo:
Hmmm, didn't we have an agreement that only fixes should go into Krypton from beta2 on?

Jalle19:
@ksooo probably, though I think I've seen other PRs marked both Cleanup and Krypton. This is not really important and not worth rebasing (it's easy to do the code inspection again) so feel free to close if you don't want it for Krypton.

ksooo:
> so feel free to close if you don't want it for Krypton.

 is it okay for you if we leave it open and label it "v18"?

Jalle19:
Yeah I guess

ksooo:
Thanks @Jalle19

MartijnKaijser:
@ksooo well I'd say if we're pretty sure it works I wouldn't hold it up. However it was more to put features on hold and still let some cleaning happen till RC

ksooo:
In this case, no objections. @Jalle19 shove it in, man.

MartijnKaijser:
jenkins build and merge

MartijnKaijser:
this new github review stuff might not make it really clear if all comments are really addressed. Could you check @Jalle19 and then jenkins build it?

AlwinEsch:
If it becomes merged from jenkins I create a RDS fix request. Otherwise think about my comment with `UECP_RDS_MS` above. Sorry.

MartijnKaijser:
Needs a rebase

ksooo:
@Jalle19 what to do with this one?

Jalle19:
Let's close it for now

PR #10511, opened 2016-09-19
[python] fix dialog documentation

ronie:
see: http://forum.kodi.tv/showthread.php?tid=291175&pid=2417848

@tamland


tamland:
:+1:

PR #10512, opened 2016-09-19
[CEC] fix CECToggleState, screensaver looping etc. 

Glenn-1990:
This one superseeds https://github.com/xbmc/xbmc/pull/10492 and contains only the most important code changes. Setting that don't apply for unsuspendable devices (RPI, IMX, etc.) are not shown to avoid user confusion.

Issues that should be solved now:
http://forum.kodi.tv/showthread.php?tid=225779
https://discourse.osmc.tv/t/control-tv-on-standby-toggle-using-cec-kodi-built-in-functions/3638/2

@opdenkamp

### DIFF
--- a/system/peripherals.xml
+++ b/system/peripherals.xml
@@ -17,6 +17,7 @@
     <setting key="cec_standby_screensaver" type="bool" value="0" label="36009" order="6" />
     <setting key="cec_wake_screensaver" type="bool" value="1" label="36010" order="7" />
     <setting key="standby_pc_on_tv_standby" type="enum" value="13011" label="36029" order="8" lvalues="36028|13005|13011" />
+    <setting key="stop_on_tv_standby" type="enum" value="36044" label="36029" order="8" lvalues="36028|13005|36044" />
     <setting key="standby_tv_on_pc_standby" type="bool" value="1" label="36026" order="9" />
     <setting key="use_tv_menu_language" type="bool" value="1" label="36018" order="10" />
     <setting key="pause_playback_on_deactivate" type="bool" value="1" label="36033" configurable="0" />
--- a/system/peripherals.xml
+++ b/system/peripherals.xml
@@ -25,7 +25,6 @@
     <setting key="connected_device" type="enum" label="36019" value="36037" lvalues="36037|36038" order="12" />
     <setting key="cec_hdmi_port" type="int" value="1" min="1" max="15" label="36015" order="13" />
     <setting key="physical_address" type="string" label="36021" value="0" order="14" />
-    <setting key="port" type="string" value="" label="36022" order="15" />
 
     <setting key="tv_vendor" type="int" value="0" configurable="0" />
     <setting key="device_name" type="string" value="Kodi" configurable="0" />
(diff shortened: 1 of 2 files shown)


kib:
ping @opdenkamp

xbmc/peripherals/devices/PeripheralCecAdapter.cpp
```diff
@@ -1185,12 +1185,12 @@ void CPeripheralCecAdapter::CecSourceActivated(void *cbParam, const CEC::cec_log
...
     }
-    else if (bPlayingAndDeactivated
-      && adapter->GetSettingInt("pause_or_stop_playback_on_deactivate") == LOCALISED_ID_STOP)
+    else if (activated == 0 && adapter->GetSettingInt("pause_or_stop_playback_on_deactivate") == LOCALISED_ID_STOP)
     {
       if (pSlideShow)
         pSlideShow->OnAction(CAction(ACTION_STOP));
-      else
+      else if (g_application.m_pPlayer->IsPlaying())
+        // only when playing as TMSG_MEDIA_STOP will wake the screensaver, this can make the source active again
```
Jalle19: Perhaps a better solution would be to not wake the screen saver when stop is pressed? Most people just press any button I think, I doubt anyone would even notice if stop doesn't work.

system/peripherals.xml
```diff
@@ -25,7 +25,6 @@
     <setting key="connected_device" type="enum" label="36019" value="36037" lvalues="36037|36038" order="12" />
     <setting key="cec_hdmi_port" type="int" value="1" min="1" max="15" label="36015" order="13" />
     <setting key="physical_address" type="string" label="36021" value="0" order="14" />
-    <setting key="port" type="string" value="" label="36022" order="15" />
```
Jalle19: Is the string (36022) used anywhere else? If not, drop that one as well.

xbmc/peripherals/devices/PeripheralCecAdapter.cpp
Glenn-1990: I'm not sure, I do not want break existing logic. In any case, we sending "stop" when we are not playing isn't needed.

system/peripherals.xml
Glenn-1990: Indeed, that seems to be the case ;-)

tudstudent:
@Glenn-1990 , would this resolve my issue which I posted here:
http://trac.kodi.tv/ticket/16842

If it is I can test this PR tomorrow/thursday. Would be very welcome if it solves the standby issue on my set.

Glenn-1990:
@tudstudent Probably not, can you post a logfile with debugging enabled in Kodi (when sending Kodi to sleep and CEC devices stay on)?

tudstudent:
@Glenn-1990 I have uploaded a log including CEC peripheral logging:
http://paste.ubuntu.com/23250235/
Hope you can help with this problem

xbmc/peripherals/devices/PeripheralCecAdapter.cpp
opdenkamp: I don't mind, either works for me. Not using the stop button to wake up the screensaver myself ;-)

MartijnKaijser:
is this PR obsolete or does it need updating ?

opdenkamp:
It'll need updating as people are still having issues. Just got an email from @olympia about this problem earlier week. Looking at CEC related issues atm.

olympia:
Indeed personally I am also having the same issues what @tudstudent is describing in the above trac ticket backed up with logs. If I remember correctly there were different opinions in the forums whether this is a libcec or Kodi issue, but was definitely working well in Kodi 16

Rechi:
@Glenn-1990 this needs a rebase

Glenn-1990:
@opdenkamp You are talking about "this problem" while this PR covers more issues, are you talking about the "CECToggleState" issue? 
My CEC enabled development environment is gone and I'm limited in time. Can you pick the commits you need?

PR #10513, opened 2016-09-19
[estouchy] exit button should execute the selected shutdown function

ronie:
quit may not be applicable on some platforms, use the selected shutdown function instead.

http://forum.kodi.tv/showthread.php?tid=262145&pid=2418740#pid2418740

### DIFF
--- a/addons/skin.estouchy/xml/Home.xml
+++ b/addons/skin.estouchy/xml/Home.xml
@@ -211,7 +211,7 @@
 				<content>
 					<item>
 						<label>13012</label>
-						<onclick>Quit</onclick>
+						<onclick>Shutdown</onclick>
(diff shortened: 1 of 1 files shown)


ksooo:
Didn't we have an agreement only to include fixes for Krypton after beta2?

ronie:
perhaps i should've tagged it fix, i dunno..
it fixes the problem, but it's also a behaviour change.

will tag it for v18, np.

ksooo:
Reading what it does, it sounds like a fix to me. We should actually label it "fix" and shove into Krypton i guess. ;-)

ronie:
you mean we should actually read stuff before we comment? ;-)

ksooo:
I mean we should use correct PR labels in the first place. :-)

PR #10514, opened 2016-09-20
drop some unneeded ifdefs related to ffmpeg

FernetMenta:
drop some ugly ifdeffery
logging ffmpeg version info is good enough for determining the source.


FernetMenta:
jenkins build this please

ksooo:
Didn't we have an agreement only to include fixes for Krypton after beta2?

FernetMenta:
I consider this as a fix :)

FernetMenta:
@ksooo it makes some commit on our ffmpeg fork obsolete. We need to bump our ffmpeg fork again for final.

ksooo:
> I consider this as a fix :)
 
 In this case we should put the "fix" label on it.

PR #10516, opened 2016-09-20
VideoPlayer: handle exceptional case where distance between keyframes…

FernetMenta:
The user who created http://trac.kodi.tv/ticket/16911 has issues with self encoded videos. The distance between keyframes is chosen much to large, 240 frames. For 24fps that results in 10 seconds while demux queue is only 8 seconds. As a result audio queue may be full while video has still not found a key frame required to start.

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -2060,6 +2060,18 @@ void CVideoPlayer::HandlePlaySpeed()
 
       m_syncTimer.Set(3000);
     }
+    else
+    {
+      // exceptions for which stream players won't start properly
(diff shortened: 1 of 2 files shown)


Jalle19:
Does this affect situations where you pull the antenna cable while watching live TV, then insert it again after let's say 30 seconds?

FernetMenta:
Not sure what exactly you mean by "affect". The scenario you described has always been working for me using vnsi. I just tried and it is not broken with this change here.

Jalle19:
@FernetMenta that's all I wanted to know, thanks

FernetMenta:
jenkins build this please

FernetMenta:
Android build error unrelated
Fix confirmed by http://trac.kodi.tv/ticket/16911#comment:17

PR #10517, opened 2016-09-20
kodi.guilib: fix version in generated addon.xml

FernetMenta:
fix regex that extracts guilib version on OSX
required after dependency checking https://github.com/xbmc/xbmc/pull/10335


Memphiz:
no need to wait on a comment from me here - if it works - it works :)

PR #10518, opened 2016-09-20
[PVR] timersettings: show unknown when the current setting is unknown

Glenn-1990:
When the duplicate setting for a timer rule is unknown to Kodi, an empty box is present. Other fields in timersettings already have a fall back, but the duplicate detection not. 
I know that the pvr addon should push all known types, but backends may change constantly while pvr addon updates are not so frequently. This PR just let the user known that the setting is unknown. This principle is the same as what is done with other settings as lifetime, priority, ... 

before PR:
empty

after PR:
not_empty

@ksooo @Jalle19

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -5944,6 +5944,7 @@ msgstr ""
 #: xbmc/addons/GUIDialogAddonInfo.cpp
 #: xbmc/peripherals/devices/Peripheral.cpp
 #: xbmc/peripherals/bus/PeripheralBus.cpp
+#: xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
 msgctxt "#13205"
 msgid "Unknown"
 msgstr ""
--- a/xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
+++ b/xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
@@ -905,6 +905,21 @@ void CGUIDialogPVRTimerSettings::DupEpisodesFiller(
     list.clear();
     pThis->m_timerType->GetPreventDuplicateEpisodesValues(list);
     current = pThis->m_iPreventDupEpisodes;
+
+    auto it = list.begin();
+    while (it != list.end())
+    {
+      if (it->second == current)
+        break; // value already in list
+
+      ++it;
+    }
+
+    if (it == list.end())
+    {
+      // PVR backend supplied value is not in the list of predefined values. Insert it.
+      list.insert(it, std::make_pair(g_localizeStrings.Get(13205), current)); // "Unknown"
+    }
   }
   else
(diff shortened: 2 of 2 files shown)


Jalle19 (approved):
I would probably use `std::find_if` instead of manual iterator looping, but it's up to you.

ksooo:
@Glenn-1990: I do not very like this fix, because nobody can actually understand what "Unknown" means in this context. So my take is that is does not improve the status quo, maybe it is even ore confusing to the user. Sorry, -1 from me for this change.

For the other fallbacks we are able to "generate" some at least somehow meaningful strings for unknown values (e.g. unknown margin time value => "%i min").

> backends may change constantly

 That's the root cause with tvheadend. It changes HTSP API without bumping API version. This needs to be fixed. I'm not willing to taint PVR core with workarounds for tvheadend flaws.

Jalle19:
If this only affects users of certain pre-stable tvheadend builds then -1 for me too. If it's a generic issue, maybe we can change the string to something less ambiguous?

ksooo:
> If it's a generic issue

 it's not. It only an issue with software changing API without bumping. ;-)

> maybe we can change the string to something less ambiguous

 in difference to the other timer attributes i have no idea for a 'less ambiguous' string.

ksooo:
@Glenn-1990 for this particular case i really would appreciate if we change pvr.hts to support the new dup detect values. For the future i would hope that tvheadend would bump htsp api whenever an api change occurs. It turned out not to be a good idea to bump only on tvheadend releases (as the release cycle of tvh is waaaaaaaaaay to long for this).

Glenn-1990:
@ksooo I understand, I was just testing tvh master and saw this issue. I know that pvr addons should handle this. But in my opinion Kodi should display something instead of nothing. When the box is empty, people will post bug reports to Kodi instead of the PVR addon.  
Btw, actually the htsp code was untouched: https://github.com/tvheadend/tvheadend/commit/66ad72055d07183b199d84ae934ad1b598e29acf

ksooo:
> @ksooo I understand, I was just testing tvh master and saw this issue. I know that pvr addons should handle this. But in my opinion Kodi should display something instead of nothing. 

 sorry, my point here still is that 'unknown' is not more meaningful than an ampty string. So, if we are not able to find something really meaningful i would like to stick with 'nothing'.

> Btw, actually the htsp code was untouched: tvheadend/tvheadend@66ad720

 not sure waht you mean. This is exactly the problem I mentioned. API was changed (new attributes added) without bumping HTSP version. tvheadend must not deliver the new attributes if the connecting client's supported htsp version is to low. Strictly speaking, that's what went wrong here.

Glenn-1990:
@ksooo something like "not supported by pvr addon", "pvr addon error", "check your pvr addon" etc?Otherwise I will close this PR :-)

ksooo:
@Glenn-1990 i would appreciate if you fix this issue in pvr.hts and close this PR.

Glenn-1990:
@ksooo , will do htsp api bump ;-)

ksooo:
> will do htsp api bump ;-)

@Glenn-1990 many thanks. much appreciated.

PR #10521, opened 2016-09-21
Ticket #16906 - Fix iEndOffset property when parsing cue files

iscy:
## Description

When CueDocument is parsing the cue file and generating the data, it blindly sets "iEndOffset" by assuming that the last track ends when the current one starts. This assumption is only right if the 'FILE' pointed by this cue sheet does not change. In the current case, each 'TRACK' has its own flac file, so this assumption is not right and PAPlayer will act on this data. Therefore, the files won't play since they will end shortly (immediately) after being started.
## Motivation and Context

http://trac.kodi.tv/ticket/16906
## How Has This Been Tested?

Once patched and the library re-imported, the album used for debugging could be played without problems.
## Types of change
## Checklist:


FernetMenta:
thanks you very much
jenkins build this please

PR #10522, opened 2016-09-21
Some improvements

garbear:
Some more low-hanging fruit that I found when rewriting RetroPlayer 4 months ago

### DIFF
--- a/xbmc/addons/AddonManager.cpp
+++ b/xbmc/addons/AddonManager.cpp
@@ -20,6 +20,7 @@
 
 #include "AddonManager.h"
 
+#include <algorithm>
 #include <iterator>
 #include <memory>
 #include <utility>
@@ -514,22 +515,25 @@ bool CAddonMgr::GetInstallableAddons(VECADDONS& addons, const TYPE &type)
   CSingleLock lock(m_critSection);
 
   // get all addons
-  VECADDONS installableAddons;
-  if (!m_database.GetRepositoryContent(installableAddons))
+  if (!m_database.GetRepositoryContent(addons))
     return false;
 
   // go through all addons and remove all that are already installed
-  for (const auto& addon : installableAddons)
-  {
-    // check if the addon matches the provided addon type
-    if (type != ADDON::ADDON_UNKNOWN && addon->Type() != type && !addon->IsType(type))
-      continue;
 
-    if (!CanAddonBeInstalled(addon))
-      continue;
+  addons.erase(std::remove_if(addons.begin(), addons.end(),
+    [this, type](const AddonPtr& addon)
+    {
+      bool bErase = false;
 
-    addons.push_back(addon);
-  }
+      // check if the addon matches the provided addon type
+      if (type != ADDON::ADDON_UNKNOWN && addon->Type() != type && !addon->IsType(type))
+        bErase = true;
+
+      if (!this->CanAddonBeInstalled(addon))
+        bErase = true;
+
+      return bErase;
+    }), addons.end());
 
   return true;
 }
--- a/xbmc/dbwrappers/Database.h
+++ b/xbmc/dbwrappers/Database.h
@@ -139,12 +139,6 @@ public:
   bool CommitMultipleExecute();
 
   /*!
-   * @brief Open a new dataset.
-   * @return True if the dataset was created successfully, false otherwise.
-   */
-  bool OpenDS();
-
-  /*!
    * @brief Put an INSERT or REPLACE query in the queue.
    * @param strQuery The query to queue.
    * @return True if the query was added successfully, false otherwise.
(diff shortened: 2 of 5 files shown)


xbmc/addons/AddonManager.cpp
```diff
@@ -514,22 +515,25 @@ bool CAddonMgr::GetInstallableAddons(VECADDONS& addons, const TYPE &type)
...
-  for (const auto& addon : installableAddons)
-  {
-    // check if the addon matches the provided addon type
-    if (type != ADDON::ADDON_UNKNOWN && addon->Type() != type && !addon->IsType(type))
-      continue;
 
-    if (!CanAddonBeInstalled(addon))
-      continue;
+  addons.erase(std::remove(addons.begin(), addons.end(),
+    [&](const AddonPtr& addon)
```
Paxxi: Not sure it matters but always specify the variable to capture as it can lead to performance issues capturing everything. Probably more of a good habit than real issue

Paxxi (approved):
Looks good overall, the comments are up to you if you want to change as it's nothing critical.

xbmc/addons/AddonManager.cpp
garbear: It looks like I captured everything to call `CanAddonBeInstalled()`. Should I capture this and call `this->CanAddonBeInstalled()` ?

MartijnKaijser:
@tamland ping for some review

xbmc/addons/AddonManager.cpp
Paxxi: I think that would be the correct way

tamland: Not sure I agree with explicit capture being a good habit *everywhere*. When the lambda outlives the scope, for sure, because then horrible things can happen. But for things like std::remove it's perfectly safe.

xbmc/addons/AddonManager.cpp
```diff
@@ -514,22 +515,25 @@ bool CAddonMgr::GetInstallableAddons(VECADDONS& addons, const TYPE &type)
...
   // go through all addons and remove all that are already installed
-  for (const auto& addon : installableAddons)
-  {
-    // check if the addon matches the provided addon type
-    if (type != ADDON::ADDON_UNKNOWN && addon->Type() != type && !addon->IsType(type))
-      continue;
 
-    if (!CanAddonBeInstalled(addon))
-      continue;
+  addons.erase(std::remove(addons.begin(), addons.end(),
```
tamland: remove_if

xbmc/addons/AddonManager.cpp
```diff
@@ -514,22 +515,25 @@ bool CAddonMgr::GetInstallableAddons(VECADDONS& addons, const TYPE &type)
...
-    addons.push_back(addon);
-  }
+      // check if the addon matches the provided addon type
+      if (type != ADDON::ADDON_UNKNOWN && addon->Type() != type && !addon->IsType(type))
+        bErase = true;
+
+      if (!CanAddonBeInstalled(addon))
+        bErase = true;
+
+      return bErase;
```
tamland: Maybe it's just a copy-paste but does this improve readability? I don't think so.. Just returning the removal condition in one line is much clearer imo. The variables and assignments is just extra noise..

xbmc/addons/AddonManager.cpp
Paxxi: Whether you capture explicitly or implicitly does not matter when it out lives the scope, bad things will happen then unless you capture by value which excludes the this pointer

xbmc/addons/AddonManager.cpp
garbear: good catch

xbmc/addons/AddonManager.cpp
garbear: so like
```
addons.erase(std::remove(addons.begin(), addons.end(),
   [this](const AddonPtr& addon)
```
?

Paxxi: Yep should be fine

xbmc/addons/AddonManager.cpp
Paxxi: Early return works fine as well in this case, still easy to read and add more checks, but it's mostly about preference I guess

garbear:
@Paxxi @tamland your fixes for the CAddonMgr commit. looks good?

xbmc/addons/AddonManager.cpp
tamland: Fine, but at least follow the surrounding style. Hungarian notation and single return abomination is not used here;)

Paxxi:
Ok for me

garbear:
jenkins build this please

garbear:
jenkins build this please

garbear:
OSX failure looks unrelated

PR #10524, opened 2016-09-21
[PVR] Fix epg data unload on application exit / log off.

ksooo:
This fixes a problem with epg data not properly re-initialized after switching user profile. Reported here http://forum.kodi.tv/showthread.php?tid=282907&pid=2418766#pid2418766

Fix is not only to stop the epg updater thread when logging off, but also to wipe all epg data.

@Jalle19 for review?


ksooo:
jenkins build this please

Jalle19 (approved):
Sounds reasonable

PR #10525, opened 2016-09-21
[RFC] [cmake] RPi: to be consistent with AMLogic add support for -DENABLE_RPI=<value>

sraue:
- rename MMAL_FOUND to RPI_FOUND
- remove CORE_SYSTEM_NAME=rbpi and use RPI_FOUND
- use RPI_FOUND for OMXLIB_FOUND and OMX_FOUND
- use linux platform support for 'rbpi'

ping @popcornmix


sraue:
will take a look regarding the builderror

project/cmake/CMakeLists.txt
```diff
@@ -211,8 +212,11 @@ endif()
 
 if(CORE_SYSTEM_NAME STREQUAL android OR CORE_SYSTEM_NAME STREQUAL linux)
   core_optional_dep(AML)
-elseif(CORE_SYSTEM_NAME STREQUAL rbpi)
```
wsnipex: this breaks unified deps build.

wsnipex (commented):
sorry, this changes way too much without need. We are looking for libraries(MMAL) in find modules, not SOCs. There is absolutely no need to rename anything to add a simply option that turns on RPI specifics.

Simply set CORE_SYSTEM_NAME to rbpi if ENABLE_RPI=1 before Archsetup is called.

sraue:
my intention is to have it consistent... not only for the config options, also internally... if we start something new why dont do this in a consistent way if possible... internally we use for AML "FOUND_AML" for RPI: FOUND_MMAL, CORE_SYSTEM_NAME STREQUAL rbpi and some others... for example for IMX i dont found the logic yet... (if its already fully implemented)
why use rpi as a platform if its a linux system and dont use the linux platform for rpi too (and share the same code)... for others like AML it checks for needed librarys instead setup a new platform.

wsnipex:
No we use FOUND_MMAL if mmal is available and FOUND_AML if amlcodec is available.
You can build for RPI without mmal libs and for amlogic socs without amlcodec.
Having any of those libs doesn't necessarily mean that its a certain platform.

sraue:
then using FOUND_MMAL instead FOUND_RPI and removing "rbpi" from platforms?
but FindMMAL also checks for VC libs, which is needed on RPI to use Kodi... and should then be seperated to FindVC or so...
Regarding @popcornmix there is no other known system which uses MMAL because its RPi specific, like VC libs and the OMX implementation
quote from @popcornmix 

> Personally I think the "rbpi" / "mmal" / "omxplayer" distinction is unnecessary. I'm pretty you you can't build with omxplayer but not mmal, or with mmal but not omxplayer (it's never tested) so combining them all into a "rbpi" option would be fine by me.

wsnipex:
Once again, MMAL is a libary, which iirc depends on VC libs. Thats why we check for VC libs in the FindMMAL module. Even if there is not a single other soc/platform/whatever that needs MMAL, we still need that find module.
And there is no FOUND_RPI, there is only CORE_SYSTEM_NAME/platform, which we use to distinguish between different platforms that use the same OS. Removing rpbi from there will break all kinds of things, all over the place.

popcornmix:
I'll start this by saying I don't know the build system well.

I do know that within the code there is enough overlap between the defines ``TARGET_RASPBERRY_PI``/``HAS_MMAL``/``HAS_OMXPLAYER`` that I would be surprised if it were possible to build with anything other than all enabled or all disabled (no non-raspberry-pi platform uses ``HAS_MMAL`` or ``HAS_OMXPLAYER``).

It is possible that this could be improved, but for now the distinction between these "features" is not very useful.

A standard linux build can be run on Raspberry Pi using the experimental arm side graphics driver (see [here](http://geektillithertz.com/wordpress/index.php/2016/05/09/kodi-on-raspberry-pi-open-source-gpu-driver-mesa-11/)) obviously without video or other gpu acceleration. But it does show the majority of the code runs on pi when built for linux (at least when built locally - I don't think this has been attempted as a cross compile). Getting that to build with acceleration options may be of interest in the future as the experimental arm side graphics driver gets more stable.

So it may make sense to treat Pi as a linux system with extra acceleration options, but I'll defer to people who know the build system better to decide what the better approach is.

wsnipex:
If you want to treat Rpi just like normal linux and therefore get rid of all TARGET_RASPBERRY_PI ifdeffery, then - by all means - lets do it.
They way to go in this case is: 
- remove TARGET_RASPBERRY_PI ifdefs
- remove CORE_SYSTEM_NAME rbpi
- fix linux/ArchSetup.cmake to properly detect all RPI variants(current PR only detects v1)
- fixup the whole depends Toolchain stuff incl binary addons toolchain
- fix binary addon building
- fix binary addon breakage in (probably a lot of) addons
- keep MMAL and OMX detection and ifdefs as is

anthonyryan1:
I definitely think that it would be a step in the correct direction to merge the raspberry pi target into the  linux target.

* While OMX is supported in some Mesa Gallium3D drivers, the the current OMX code within Kodi would have no rendering path on such devices. Because of this it would probably be bad idea to expose any sort of `-DENABLE_OMX` without correcting that first.
* MMAL is still raspberrypi specific, so it could probably be considered synonymous with the raspberry pi.

I think all the raspberrypi ifdefs should be replaced with mmal ifdefs because the use of MMAL is the only thing that makes the raspberry pi special when compared to a generic Linux build on arm. I don't think we should bother trying to detect all the raspberry pi variants. We should just support generic arm targets and detect or provide a `-DENABLE_NEON` cmake argument for differentiating the only arm feature that I see being used that might not be available on all arm targets.

Right now the only purpose served by detecting all the variants is to automatically optimize the CFLAGS when cross compiling. That sort of thing belongs in a wiki, rather than in our build system. All distros and cross compile environments targeting the Pis would be better off setting the same arguments system wide for their building if they're concerned about focusing that intently on tweaking their CFLAGS. It's a slippery slope sort of problem, if we optimize CFLAGS for one SoC, why shouldn't we optimize for all of them.

Finally, I'm torn on OMX. If the OMX stuff could be generalized such that it works on any platform providing OMX, it should be on it's own and toggled with a `-DENABLE_OMX` independent from the MMAL stuff. While it's raspberry pi specific I think it should be behind a `-DENABLE_MMAL` with the rest of the raspberry pi and mmal stuff.

I look forward to everyone's thoughts.

popcornmix:
omxplayer and stuff inside ``HAVE_OMXLIB`` are unlikely to be usable on non-pi platforms.
The openmax spec is not sufficient to do all we need, so there are vendor specific parameters (e.g. ``OMX_PARAM_BRCM*``) which won't work on other platforms.

Also note that ``HAVE_LIBOPENMAX`` is a different openmax target (Tegra?) and is not Pi related.

As such I don't mind the Pi specific openmax stuff being included inside the ``ENABLE_MMAL`` define.
As ``TARGET_RASPBERRY_PI``, ``HAS_MMAL`` and ``HAVE_OMXLIB`` are all defined on Pi builds and none are defined on non-pi builds, I'd be surprised if you can successfully build with just a subset of those defines (may be worth a try to confirm). If they don't work independently then merging them seems reasonable.

MartijnKaijser:
Still wanted?

PR #8922, opened 2016-01-19
[guiinfo] extend ranges for listitem properties

phil65:
Since the amount of properties skins are using grew quite significantly over the last years, the ranges we set back then for the listitem properties are not large enough anymore. This should fix that by approx. doubling the ranges.
As a testcase you can try running a simple script

```
import xbmc
for i in range(1, 90):
    InfoLabel = xbmc.getInfoLabel("Container(9000).ListItem.Property(%i)" % (i))
```

or
https://www.dropbox.com/s/z4xhfy7caeipfkd/script.bug.zip?dl=0

and see that this breaks confluence without this commit (no fanart visible anymore in library for example)

@mkortstiege @ronie

PR #10527, opened 2016-09-21
increase property and variable ranges for skins

ronie:
two commits:

**1 - increase range for listitem properties by @phil65** 
this is **not** a fix, but a workaround for a bug that would really need fixing!!!1!
for discussion, please see: https://github.com/xbmc/xbmc/pull/8922

**2 - increase range for variables**
we currently 'only' reserve 500 places for skin variables (this used to be 1000 when the feature was introduced) and some skins are hitting (or exceeding) this limit.
as a result these skins pretty much get crippled after a few minutes of use.
this commit restores the reserved range for variables back to the original value of 1000.
see: http://forum.kodi.tv/showthread.php?tid=288287

### DIFF
--- a/xbmc/guiinfo/GUIInfoLabels.h
+++ b/xbmc/guiinfo/GUIInfoLabels.h
@@ -804,18 +804,19 @@
 #define LISTITEM_ADDON_NEWS         (LISTITEM_START + 178)
 #define LISTITEM_ADDON_SIZE         (LISTITEM_START + 179)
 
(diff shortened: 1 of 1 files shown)


ronie:
squashed.
jenkins build this please.

MartijnKaijser:
Perhaps add a doxygen todo comment in this location so someone looking at the code sees there's an issue to be fixed?

ronie:
i don't master the art of doxygen, mind checking if this is correct?

Paxxi:
Changes look OK, can't say if this will cause any performance issues though

MartijnKaijser:
Jep looks good. Squash and merge i would say

PR #10528, opened 2016-09-22
[cmake] Fix FindCCACHE module

fetzerch:
@wsnipex, @fritsch followup on yesterdays slack discussion

Clean build with empty ccache:
2127.38user 227.17system 5:10.50elapsed 758%CPU (0avgtext+0avgdata 745860maxresident)k
10040inputs+8928728outputs (381major+39179319minor)pagefaults 0swaps

Clean build with ccache enabled:
751.87user 59.98system 1:56.25elapsed 698%CPU (0avgtext+0avgdata 745836maxresident)k
80inputs+4412512outputs (348major+19134578minor)pagefaults 0swaps

### DIFF
--- a/project/cmake/modules/FindCCache.cmake
+++ b/project/cmake/modules/FindCCache.cmake
@@ -7,7 +7,11 @@
 # See: https://crascit.com/2016/04/09/using-ccache-with-cmake/
 
 find_program(CCACHE_PROGRAM ccache)
-if(CCACHE_PROGRAM)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(CCACHE REQUIRED_VARS CCACHE_PROGRAM)
+
+if(CCACHE_FOUND)
   # Supports Unix Makefiles and Ninja
   set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
   set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK "${CCACHE_PROGRAM}")


akva2:
just curious, which distros do not setup ccache wrappers for default compilers ?

fetzerch:
Ubuntu if i'm not mistaken.

I'll plan to extend the module also for osx (https://crascit.com/2016/04/09/using-ccache-with-cmake/) when i find time.

akva2:
Then you are mistaken. 'type cc' -> /usr/lib/ccache/gcc. I have had ccache for years without doing anything so i am surprised if it is needed on ubuntu.

akva2:
Note not speaking against the work. other platforms, distros et al may very well need it.

fetzerch:
Hm, I get this on xenial:
```
$ ll /usr/bin/cc
lrwxrwxrwx 1 root root 20 Mar 26 09:32 /usr/bin/cc -> /etc/alternatives/cc
$ ll /etc/alternatives/cc
lrwxrwxrwx 1 root root 12 Mar 26 09:32 /etc/alternatives/cc -> /usr/bin/gcc*
$ ll /usr/bin/gcc
lrwxrwxrwx 1 root root 5 Mar 26 09:32 /usr/bin/gcc -> gcc-5
```
The find module has the other advantage that it should work also out of the box with clang if you export CC and CXX (but then you might as well just export it to a clang ccache wrapper).

On arch according to the wiki you'd also have to export it to use it. https://wiki.archlinux.org/index.php/ccache

akva2:
Confusing. Clang has the same treatment on my xenial, all compilers get a wrapper and /usr/lib/ccache is first in PATH. I have probably installed some package. Anyways.. was just curious.

fetzerch:
My installation is relatively new, and i just figured out yesterday that i wasn't using it.
Might be that this comes with some other package.

PR #10533, opened 2016-09-22
[cmake] FindSmbClient: find libraries needed by libsmbclient so we link against them

sraue:
esp. with libsambaclient static build we must link against this libs. this libs are samba default dependencies so search this libs and link against this


project/cmake/modules/FindSmbClient.cmake
```diff
@@ -23,6 +23,12 @@ find_path(SMBCLIENT_INCLUDE_DIR NAMES libsmbclient.h
 find_library(SMBCLIENT_LIBRARY NAMES smbclient
                                PATHS ${PC_SMBCLIENT_LIBDIR})
 
+find_library(TALLOC_LIBRARY talloc)
+find_library(TDB_LIBRARY tdb)
+find_library(TEVENT_LIBRARY tevent)
+find_library(WBCLIENT_LIBRARY wbclient)
+find_library(RESOLV_LIBRARY resolv)
+
```
fetzerch: Are this libs required on all platforms or only on some?
It should be appended to the FPHSA call's REQUIRED_VARS as well so that cmake fails if one of the libs is not available.

sraue:
@fetzerch can you check if on the system which runs the checks here this libraries are installed so this checks fails here? samba depends on this libs, but ships this libs byself of not compiled against this libs as external libs (what many distros are doing). so at least for linux i think samba will not work without this libs.

project/cmake/modules/FindSmbClient.cmake
sraue: appended to the FPHSA call

wsnipex:
those libs are only needed when smbclient is static.
Can you check if "pkg-config --libs --static smbclient" outputs the correct libs?
If yes, you can use PC_SMBCLIENT_STATIC_LIBARIES

sraue:
@wsnipex yes they are shown... how to use PC_SMBCLIENT_STATIC_LIBARIES do we have already a sample with a similiar package?

sraue:
event without this patch its already set... need to find the way using this for linking...:
`PC_SMBCLIENT_STATIC_LIBRARIES:INTERNAL=smbclient;talloc;tdb;tevent;wbclient;smbclient;resolv;nsl;dl`

wsnipex:
you can do something like this: https://github.com/wsnipex/xbmc/blob/32b9af0dd62feb69fc53d29ff9923ca6892936cf/project/cmake/modules/FindFFMPEG.cmake#L128-L136

If needed, you can also use find_library to check each one

lrusak:
@wsnipex what about something like this? https://github.com/lrusak/xbmc/commit/e24bfeee3d4d114de6e3c0fcf9ccae62d466630a

stefansaraev:
@lrusak here is how it's done for python: https://github.com/xbmc/xbmc/pull/11183/commits/1fd33407e72c58da4db53eb6fb4f3ab0cb9a39fa

lrusak:
@stefansaraev something like this works, however, I'm not sure if that is the correct way or not,
https://github.com/lrusak/xbmc/commit/ff90a7551a080ab9555f526ab497b5278d8144a3

wsnipex:
like https://github.com/xbmc/xbmc/blob/master/cmake/modules/FindFFMPEG.cmake#L183-L191 please

lrusak:
@wsnipex I really don't know what else to do. This is all I can come up with, https://github.com/lrusak/xbmc/commit/3f8992e2522c377fcc98439bb1bf2321adbba09b

Rechi:
@sraue this needs a rebase

Geremia:
Is this related to [this build error I'm getting on 17.6](https://forum.kodi.tv/showthread.php?tid=325100&pid=2675974#pid2675974)?:

>[100%] Linking CXX executable kodi.bin
>…
>/usr/lib/liblibsmb-samba4.so: undefined reference to `_tevent_req_error@TEVENT_0.9.9'
>/usr/lib/libtevent-util.so.0: undefined reference to `tevent_req_is_error@TEVENT_-1.9.9'
>…

wsnipex (changes requested):
as already said, this should only be used when we actually link a static lib.

phunkyfish:

PR #10491, opened 2016-09-18
Drop AsyncFileCopy and some random seemingly unused stuff

Jalle19:
## Description

Just some minor things I've found while trying to pick up coding on Kodi again.
## Motivation and Context

Gotta start somewhere
## How Has This Been Tested?

It seems to build

PR #10534, opened 2016-09-22
[macos][ios] fix xcode project after #10491.

ksooo:
Fix xcode project after #10491 (removal of AsyncFileCopy.(cpp|h)

@Jalle19 fyi, if you remove a file, xcode project must be adapted.


ksooo:
jenkins build and merge

L-S-D:
I have a similar problem on Linux, 
xbmc/utils/utils.a needs AsyncFileCopy.o

bkuhls:
Same here
```make[2]: *** No rule to make target 'AsyncFileCopy.o', needed by 'utils.a'.  Stop.```

This line needs to be removed:
https://github.com/xbmc/xbmc/blob/master/xbmc/utils/Makefile.in#L4

ksooo:
xbnc/utils/Makefile must be adapted as well. will do. :-/

Jalle19:
Thanks guys!

PR #10535, opened 2016-09-22
Fix autoconf-based build after #10491.

bkuhls:
Fix Makefile after #10491 (removal of AsyncFileCopy.(cpp|h)

Reported by https://github.com/xbmc/xbmc/pull/10491#issuecomment-248974387 & https://github.com/xbmc/xbmc/pull/10534#issuecomment-248986438


Paxxi:
Thank you

PR #10538, opened 2016-09-22
[cmake] Improve documentation

fetzerch:
Small improvements for the CMake user documentation: [Preview](https://github.com/fetzerch/xbmc/blob/cmake_docs/project/cmake/README.md)

### DIFF
--- a/project/cmake/README.md
+++ b/project/cmake/README.md
@@ -3,13 +3,12 @@
 This files describes Kodi's CMake based buildsystem. CMake is a cross-platform
 tool for generating makefiles as well as project files used by IDEs.
 
-The current version of the buildsystem is capable of building the main Kodi
-executable (but no packaging or dependency management yet) for the following
-platforms:
+The current version of the buildsystem is capable of building and packaging
+Kodi for the following platforms:
 
 - Linux (GNU Makefiles, Ninja)
 - Windows (NMake Makefiles, Visual Studio 14 (2015), Ninja)
-- OSX and IOS (GNU Makefiles, Xcode, Ninja)
+- macOS and iOS (GNU Makefiles, Xcode, Ninja)
 - Android (GNU Makefiles)
 - FreeBSD (GNU Makefiles)
 
@@ -45,12 +44,12 @@ are downloaded using `DownloadBuildDeps.bat` and `DownloadMingwBuildEnv.bat`
 and that the mingw libs (ffmpeg, libdvd and others) are built using
 `make-mingwlibs.bat`.
 
-### OSX
+### macOS
 
-For OSX the required dependencies can be found in
+For macOS the required dependencies can be found in
 [docs/README.osx](https://github.com/xbmc/xbmc/tree/master/docs/README.osx).
 
-On OSX it is necessary to build the dependencies in `tools/depends` using
+On macOS it is necessary to build the dependencies in `tools/depends` using
 `./bootstrap && ./configure --host=<PLATFORM> && make`. The other steps such
 as `make -C tools/depends/target/xbmc` and `make xcode_depends` are not needed
 as these steps are covered already by the CMake project.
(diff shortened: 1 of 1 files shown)


wsnipex:
nice! Can you also add a section on building binary addons?

project/cmake/README.md
```diff
@@ -142,7 +146,23 @@ cmake --build . -- VERBOSE=1 -j$(sysctl -n hw.ncpu)  # or: make VERBOSE=1 -j$(sy
...
+
+#### OSX installer generation
+
+Afterwards an installable DMG for OSX can be built with the following command:
+
+```
+cmake --build . --config "Release" --target "dmg"  # or: make dmg
+```
+
+#### IOS package generation
```
Jalle19: It's `iOS`, not `IOS` which is a Cisco brand.

fritsch:
Tank you very much!

fetzerch:
@wsnipex: Added section for binary addons. I'm sure there are at least 5 more ways to build addons ;)
@Jalle19: Nitpicker ;P Changed also OSX to macOS.

project/cmake/README.md
```diff
@@ -168,13 +221,36 @@ scanning them would take up an unreasonable amount of configure time.
...
 - `make check` builds and executes the test suite.
+- `make check-valgrind` builds and executes the test suite with valgrind memcheck.
 
 Code coverage (with Gcov, LCOV and Gcovr) can be built on Linux:
 
 - CMake has to be executed with `-DCMAKE_BUILD_TYPE=Coverage`
 - `make coverage` generates an HTML code coverage report.
 - `make coverage_xml` generates an XML code coverage report.
 
+## Building binary addons
```
FernetMenta: what audience is this targeted to? developers, in particular developers who build new addons, need to build this differently.
this doc should stress the pitfall: if you built addon 1, you can't build addon 2 unless deleting the bootsstrap dir first.

fetzerch: My main usecase when adding the `make binary-addons` was testing packaging for osx/ios/android. I thought it would be useful also for users that build kodi, but don't develop addons.
I'll make this more clear in the documentation.

Since I didn't yet have played much with the binary-addon build system, maybe you, @wsnipex, @hudokkow, @garbear or @AchimTuran can help documenting their workflow. That would also make it more clear to me what issues you currently face when working on the addons.

FernetMenta: see my post on the forum: http://forum.kodi.tv/showthread.php?tid=219166&pid=1934922#pid1934922
developing addons is cumbersome. if the addon does not exist in our binary-addon-repo you have to create a folder manually in project/cmake/addons/addons or the build systems fails silently

fetzerch: Let me know if it's acceptable now. We can always incrementally improve it as we go on.

project/cmake/README.md
```diff
@@ -168,13 +221,43 @@ scanning them would take up an unreasonable amount of configure time.
...
+```
+make binary-addons ADDONS="visualization.spectrum pvr.demo"
+```
+
+Addon developers can build single addons into the Kodi build directory
+so that the addon can be tested with self-compiled specific versions of Kodi.
+
+```
+mkdir pvr.demo-build && cd pvr.demo-build
+cmake -DCMAKE_PREFIX_PATH=<KODI_BUILD_DIR>/build -DCORE_BUILD_DIR=<KODI_BUILD_DIR> <pvr.demo-SRC>
```
FernetMenta: - in 99.99% of all cases devs want to build debug
- there was confusion about whether the path to addon src must be relative or absolute. would be nice if this was clarified here.
- the bootstrap issue should be either fixed or documented

fetzerch: Thanks, updated.

For the bootstrap issue, I just mentioned that it only works for existing addons. If you try to build a non existing addon, make binary-addon fails with "No addons available to be built".
There's certainly more work needed for both: fixing and documenting.

garbear:
Here's the instructions I wrote a long time ago (they are now out of date): https://github.com/kodi-game/peripheral.joystick

Notice separate sections for building vs developing

AchimTuran:
Here you can see my current scripts for developing my addons.
https://github.com/AchimTuran/xbmc/tree/AudioDSP/V2/usability_core_fixes/tools/windows/buildtools
Furthermore here you can see some of my notes to build binary addons locally.
https://github.com/kodi-adsp/adsp.biquad.filters/blob/master/README.md

Currently I miss source groups for binary Addon workspaces. Here is a first workaround, it could be improved if we can use the macros from the core build system. But I don't know how to include on cmake file into another.
https://github.com/AchimTuran/xbmc/commit/8295db89387dcd643994798e7815b576215ceeee

fetzerch:
Merging this as it's already an improvement, will improve the addon docs further on in a separate iteration.

PR #10540, opened 2016-09-23
[ios] - added latest ios devices for proper screen scale

Memphiz:
I got a bit lazy latly. Here is the bunch of latest ios devices added for proper screen scaling.

This adds:
iPodTouch6G
iPadAir (TD LTE Version)
iPhone6s
iPhone6s Plus
iPhoneSE
iPhone7
iPhone7 Plus
iPadMini4 (both versions)
iPadPro 9,7 inch (both versions)
iPadPro 12,9 inch (both versions)


Memphiz:
no need for jenkins i have build tested this (ands only an enum + some if conditions)

PR #10543, opened 2016-09-23
[platform] - add unique hardware identifier as info label

Memphiz:
As discussed at DevCon we would benefit from having a possibility to better identifer unique kodi instances.

This is the core work for this. ATM it simply uses the mac address of the first connected network interface and builds the MD5 of it. This value is:
1. visible in the SystemInfo -> Hardware Section
2. Logged to the header of the kodi.log
3. Available as info label with the name "hardwareuuid"

For osx and ios i already implemented overrides that use proper os methods for getting a unique hardware id (which might be better as a mac address).

@FernetMenta @fritsch @Paxxi @popcornmix please have a look and in case there is something better on your platforms - implement an override class in 
https://github.com/xbmc/xbmc/tree/master/xbmc/platform/overrides/*yourplatform*/Platform*Name*.cpp with overridden GetUniqueHardwareIdentifier method.

The better we get this right in the first shot - the more accurate the statistics will be!

@MartijnKaijser can you do anything good with this? (from an addon side i mean)

Beside that - we agreed that we will document the purpose of that identifier. We should do this before anything gets into production.

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -5525,7 +5525,13 @@ msgctxt "#12395"
 msgid "Battery level"
 msgstr ""
 
-#empty strings from id 12396 to 12599
+#. Please don't translate this but simply use the english version!
+#: xbmc/windows/GUIWindowSystemInfo.cpp
+msgctxt "#12396"
+msgid "Hardware UUID"
+msgstr ""
+
+#empty strings from id 12397 to 12599
 
 #: xbmc/guilib/WindowIDs.h
 msgctxt "#12600"
--- a/xbmc/platform/Platform.cpp
+++ b/xbmc/platform/Platform.cpp
@@ -21,7 +21,7 @@
 #include "Platform.h"
 
 // Override for platform ports
-#if !defined(PLATFORM_OVERRIDE)
+#if !defined(PLATFORM_OVERRIDE_CLASSPLATFORM)
 
 CPlatform* CPlatform::CreateInstance()
 {
--- a/xbmc/platform/Platform.cpp
+++ b/xbmc/platform/Platform.cpp
@@ -18,7 +18,14 @@
  *
  */
 
+#include "system.h"
 #include "Platform.h"
+#include "Application.h"
+#include "network/Network.h"
+#include "utils/log.h"
+#include "utils/md5.h"
+
+const std::string CPlatform::NoValidUUID = "NOUUID";
 
 // Override for platform ports
 #if !defined(PLATFORM_OVERRIDE_CLASSPLATFORM)
@@ -34,16 +41,30 @@ CPlatform* CPlatform::CreateInstance()
 
 CPlatform::CPlatform()
 {
-  
-}
-
-CPlatform::~CPlatform()
-{
-  
+  m_uuid = NoValidUUID;
 }
 
 void CPlatform::Init()
 {
-  // nothing for now
+  InitUniqueHardwareIdentifier();
+}
+
+void CPlatform::InitUniqueHardwareIdentifier()
+{
+  CNetworkInterface* iface = g_application.getNetwork().GetFirstConnectedInterface();
+  if (iface)
+  {
+    m_uuid = iface->GetMacAddress();
+#if defined(_DEBUG)
+    CLog::Log(LOGDEBUG, "HardwareUUID (nomd5): %s", m_uuid.c_str());
+#endif
+    m_uuid = XBMC::XBMC_MD5::GetMD5(m_uuid);
+    CLog::Log(LOGNOTICE, "HardwareUUID: %s", m_uuid.c_str());
+  }
+}
+
+std::string CPlatform::GetUniqueHardwareIdentifier()
+{
+  return m_uuid;
 }
 
--- a/xbmc/platform/Platform.cpp
+++ b/xbmc/platform/Platform.cpp
@@ -55,11 +55,7 @@ void CPlatform::InitUniqueHardwareIdentifier()
   if (iface)
   {
     m_uuid = iface->GetMacAddress();
-#if defined(_DEBUG)
-    CLog::Log(LOGDEBUG, "HardwareUUID (nomd5): %s", m_uuid.c_str());
-#endif
     m_uuid = XBMC::XBMC_MD5::GetMD5(m_uuid);
-    CLog::Log(LOGNOTICE, "HardwareUUID: %s", m_uuid.c_str());
   }
 }
 
--- a/xbmc/platform/overrides/linux/PlatformLinux.cpp
+++ b/xbmc/platform/overrides/linux/PlatformLinux.cpp
@@ -47,9 +47,6 @@ void CPlatformLinux::InitUniqueHardwareIdentifier()
       if (fgets(line, sizeof line, f) != nullptr)
       {
           m_uuid = line;
-#if defined(_DEBUG)
-          CLog::Log(LOGDEBUG, "HardwareUUID (nomd5): %s", m_uuid.c_str());
-#endif
       }
       fclose(f);
   }
@@ -69,9 +66,6 @@ void CPlatformLinux::InitUniqueHardwareIdentifier()
           if (colon)
           {
             m_uuid = colon + 2;
-#if defined(_DEBUG)
-              CLog::Log(LOGDEBUG, "HardwareUUID (nomd5): %s", m_uuid.c_str());
-#endif
           }
         }
       }
@@ -87,6 +81,5 @@ void CPlatformLinux::InitUniqueHardwareIdentifier()
   else
   {
     m_uuid = XBMC::XBMC_MD5::GetMD5(m_uuid);
-    CLog::Log(LOGNOTICE, "HardwareUUID: %s", m_uuid.c_str());
   }
 }
--- a/xbmc/platform/overrides/linux/PlatformLinux.cpp
+++ b/xbmc/platform/overrides/linux/PlatformLinux.cpp
@@ -1,0 +1,92 @@
+/*
+ *      Copyright (C) 2016 Team Kodi
+ *      http://kodi.tv
+ *
+ *  This Program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2, or (at your option)
+ *  any later version.
+ *
+ *  This Program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with Kodi; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "PlatformLinux.h"
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include "utils/log.h"
+#include "utils/md5.h"
+
+CPlatform* CPlatform::CreateInstance()
+{
+    return new CPlatformLinux();
+}
+
+void CPlatformLinux::Init()
+{
+  // call base init
+  CPlatform::Init();
+  // add platform specific init stuff here
+}
+
+void CPlatformLinux::InitUniqueHardwareIdentifier()
+{
+  // try to read the machine-id from etc first
+  FILE *f = fopen("/etc/machine-id", "r");
+  if (f)
+  {
+      char line[33]; // 32 + zero byte
+      if (fgets(line, sizeof line, f) != nullptr)
+      {
+          m_uuid = line;
+#if defined(_DEBUG)
+          CLog::Log(LOGDEBUG, "HardwareUUID (nomd5): %s", m_uuid.c_str());
+#endif
+      }
+      fclose(f);
+  }
+    
+  // if we didn't get an uuid yet - try to get a serialnumber
+  if (m_uuid == NoValidUUID)
+  {
+    f = fopen("/proc/cpuinfo", "r");
+    if (f)
+    {
+      char line[256];
+      while (fgets(line, sizeof line, f))
+      {
+        if (strncmp(line, "Serial", 6) == 0)
+        {
+          char *colon = strchr(line, ':');
+          if (colon)
+          {
+            m_uuid = colon + 2;
+#if defined(_DEBUG)
+              CLog::Log(LOGDEBUG, "HardwareUUID (nomd5): %s", m_uuid.c_str());
+#endif
+          }
+        }
+      }
+      fclose(f);
+    }
+  }
+
+  // fallback to base implementation if needed
+  if (m_uuid == NoValidUUID)
+  {
+    CPlatform::InitUniqueHardwareIdentifier();
+  }
+  else
+  {
+    m_uuid = XBMC::XBMC_MD5::GetMD5(m_uuid);
+    CLog::Log(LOGNOTICE, "HardwareUUID: %s", m_uuid.c_str());
+  }
+}
--- a/xbmc/platform/overrides/rbpi/PlatformRbpi.h
+++ b/xbmc/platform/overrides/rbpi/PlatformRbpi.h
@@ -1,0 +1,37 @@
+#pragma once
+
+/*
+ *      Copyright (C) 2016 Team Kodi
+ *      http://kodi.tv
+ *
+ *  This Program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2, or (at your option)
+ *  any later version.
+ *
+ *  This Program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with Kodi; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "platform/Platform.h"
+
+class CPlatformRbpi : public CPlatform
+{
+  public:
+    /**\brief C'tor */
+    CPlatformRbpi() = default;
+  
+    /**\brief D'tor */
+    virtual ~CPlatformRbpi() = default;
+  
+    void Init() override;
+  
+    void InitUniqueHardwareIdentifier() override;
+};
--- a/Kodi.xcodeproj/project.pbxproj
+++ b/Kodi.xcodeproj/project.pbxproj
@@ -820,7 +820,7 @@
 		DF033D381946612400BFC82E /* AEDeviceEnumerationOSX.cpp in Sources */ = {isa = PBXBuildFile; fileRef = DF033D361946612400BFC82E /* AEDeviceEnumerationOSX.cpp */; };
 		DF0ABB73183A94A30018445D /* Utf8Utils.cpp in Sources */ = {isa = PBXBuildFile; fileRef = DF0ABB71183A94A30018445D /* Utf8Utils.cpp */; };
 		DF0ABB74183A94A30018445D /* Utf8Utils.cpp in Sources */ = {isa = PBXBuildFile; fileRef = DF0ABB71183A94A30018445D /* Utf8Utils.cpp */; };
-		DF0D9F4D1D63931F006A7DBB /* Platform.cpp in Sources */ = {isa = PBXBuildFile; fileRef = DF0D9F4B1D63931F006A7DBB /* Platform.cpp */; settings = {COMPILER_FLAGS = "-DPLATFORM_OVERRIDE"; }; };
+		DF0D9F4D1D63931F006A7DBB /* Platform.cpp in Sources */ = {isa = PBXBuildFile; fileRef = DF0D9F4B1D63931F006A7DBB /* Platform.cpp */; settings = {COMPILER_FLAGS = "-DPLATFORM_OVERRIDE_CLASSPLATFORM"; }; };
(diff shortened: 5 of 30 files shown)


Memphiz:
@chewitt fyi ping

xbmc/platform/Platform.cpp
```diff
@@ -47,3 +53,22 @@ void CPlatform::Init()
   // nothing for now
 }
 
+std::string CPlatform::GetUniqueHardwareIdentifier()
+{
+  static std::string uuid = NoValidUUID;
+
+#if defined(HAS_LINUX_NETWORK) || defined(HAS_WIN32_NETWORK)
```
FernetMenta: I would expect this in the overrides for Linux and Windows

xbmc/platform/Platform.cpp
```diff
@@ -47,3 +53,22 @@ void CPlatform::Init()
   // nothing for now
 }
 
+std::string CPlatform::GetUniqueHardwareIdentifier()
+{
+  static std::string uuid = NoValidUUID;
```
FernetMenta: please make uuid a member

xbmc/platform/Platform.cpp
Memphiz: but the check is for LINUX_NETWORK ... it is true for other platforms too ... its not a check for TARGET_LINUX and TARGET_WIN32

popcornmix:
@Memphiz I've added support for reading hardware serial number from ``/proc/cpuinfo``.
This works on Pi and many linux platforms (e.g. Odroid). The serial is a 64-bit hex number (although Pi only uses lower 32-bits). My x64 Ubuntu machine doesn't support this.
Currently it is ``#ifdef TARGET_RASPBERRY_PI``, but I think it may make sense to always try this before mac address check on linux platforms.
See: https://github.com/popcornmix/xbmc/commit/829254256666061766fe0ed82e4dd51f6a8cc549

Memphiz:
@popcornmix thanks - sounds good in general - @FernetMenta how do we fit this in the overrides design. We can add an override rbpi which only works for rbpi - there is no way to have something for all linux systems with our principle - @popcornmix we are evaluating a new design pattern which gets rid of platform ifdefs - once we are clear about it i will take your code and add it in the new way and ping you so that you get an idea how it is meant to be done :)

Memphiz:
@popcornmix see https://github.com/xbmc/xbmc/pull/10543/commits/41a6486f59db2749bcbbdec4e4084b405c4687a4 - notice it only adds a cpp and a h file - the buildsystem (cmake and makefiles) will pick this up automagically because it is in the overrides folder and has a match on CORE_SYSTEM_NAME

Though this concept doesn't allow yet to have something common for linux for example (so its no 1:1 replacement for TARGET_LINUX for example)

You get the idea ...

Memphiz:
@popcornmix sorry - had to force push - https://github.com/xbmc/xbmc/pull/10543/commits/eaa955252486bf21ded41645a02c9071729d6afa is the commit for rbpi

Hedda:
Tip; You can also use Time and Date to make an even more unique identifier.

That is, in case you have several Kodi installations on the same hardware.

https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_1_.28date-time_and_MAC_address.29

This is what a universally unique identifier (UUID) does in Version 1.

Memphiz:
@Hedda that uuid would change on each start of the software - we want a unique identifier which is as stable as possible - similar to an IMEI on phones - we want to gather some accurate usage counts of unique devices.

Hedda:
@Memphiz Ah, ok I see. Sorry I misunerstod your intended purpose.

chewitt:
LGTM 👍

FernetMenta:
@Memphiz I would do it the same why as for ProcessInfo:
https://github.com/xbmc/xbmc/tree/master/xbmc/cores/VideoPlayer/Process/overrides/rbpi

Memphiz:
@FernetMenta yes i di that in my last commit - but that means its only used on rbpi and not general on linux ... i take it - we have to duplicate that code for linux then (and fall back to the base implementation there)?

Also why is in the processinfo a check for target_raspberry_pi? Is this really needed?

Memphiz:
this is a tradeoff ... instead of having something like if defined(TARGET_LINUX) || defined(TARGET_FOO) we have to duplicate the code into overrides/linux and overrides/foo to achieve the same. Do we want that sort of code duplication just for the sake of removing ifdefs?

chewitt:
FWIW.. on Linux nothing apart from Pi has a useable serial identifier and the only devices we've seen where MAC address is an issue is a statistically irrelevant minority of older Amlogic devices where u-boot had a hardcoded MAC address. In our own UUID script we try MAC and if for some reason it's not available we generate our own random number and store it so it survives a reboot (but not reinstall).

FernetMenta:
> Also why is in the processinfo a check for target_raspberry_pi? Is this really needed?

It should not be required.

> Do we want that sort of code duplication just for the sake of removing ifdefs?

I don't see this kind of code duplication as an issue. In particular not in this stage. Later, should we notice that bigger portions of code are duplicated, we can share this amongst the platforms.

Memphiz:
@chewitt well as said - odroid has and maybe other embedded platforms - i'll add the serial check to linux aswell and fallback to macaddr if not found - better save then sorry :)

Memphiz:
@FernetMenta ok - i'll add overrides for win32, linux and android then ...

xbmc/platform/overrides/rbpi/PlatformRbpi.h
```diff
@@ -0,0 +1,37 @@
...
+ *
+ */
+
+#include "platform/Platform.h"
+
+class CPlatformRbpi : public CPlatform
+{
+  public:
+    /**\brief C'tor */
+    CPlatformRbpi();
```
Paxxi: Use `= default;` instead of having empty definitions

Paxxi (commented):
Will have a look at win32 and see what we can use

xbmc/platform/Platform.cpp
Memphiz: i will simply remove those ifdefs - if g_application.getNetwork().GetFirstConnectedInterface(); the platform is broken - there is no platform which doesn't implement that anyway ... the ifdefs where a c&p from somewhere else - i don't see why those are required at all.

xbmc/platform/overrides/rbpi/PlatformRbpi.h
Memphiz: @Paxxi thx for the =default trick - changed ... (and learned something new)

stefansaraev:
> In our own UUID script we try MAC and if for some reason it's not available we generate our own random number and store it so it survives a reboot (but not reinstall)

@chewitt perhaps a bit irrelevant. but you should not track users by mac address. you should use systemd's machineid instead. 

@Memphiz 

on linux, if /etc/machine-id exists, use that.

on android: https://developer.android.com/reference/android/provider/Settings.Secure.html#ANDROID_ID

on windiws: every windows machine has unique id that can be fetched via wmi, not sure if that's easy to do from kodi.

popcornmix:
@Memphiz there is a compile error in platformRBbpi.cpp - changeing ``<stdio.h>`` to ``<string.h>`` fixes it. However the code doesn't get called.

The MAC address scheme should work on Pi, so using ``proc/cpuinfo`` isn't critical, but MAC address is limited to 24-bits on Pi (the OUI is always the same), so MAC addresses and not guaranteed unique. The serial number whilst also not guaranteed unique has 32 bits so collisional are less likely.

I can confirm that ``/etc/machine-id`` does exist on raspbian and libreelec on Pi. I am assuming this doesn't persists over reinstalls so may be less accurate that a MAC address or hardware serial number which never changes for counting unique users.

Memphiz:
@FernetMenta added linux override with machine-id - thx @stefansaraev - so now only android and windows is missing ... @popcornmix i have force pushed again (sorry for that - i like to have my PRs clean) and changed the include. I have no idea why the method isn't called on rbpi - it should as long as you build with either makefiles or cmake (might need to run cmake again so it picks up the rbpi implementation though).

Memphiz:
@stefansaraev that android thingy is for api level 24 - we are at 17 if i am right ...

Memphiz:
@stefansaraev nevermind - api 24 was just selected - the id is available since api level 3 ...

chewitt:
LE `/etc/machine-id` persists over reboots AND reinstallation in all but a tiny handful of cases which is why we use MAC and not the systemd approach. If your userbase is 75% pi devices that boot from SD cards the reinstall rate is somewhat higher than normal and persistence negates that skewing the stats. It's probably not the issue it once was, but it works, so no plan to fix it.

stefansaraev:
HKEY_LOCAL_MACHINE\Software\Microsoft\Cryptography\MachineGuid should be easier for windows.

xbmc/platform/overrides/linux/PlatformLinux.cpp
```diff
@@ -0,0 +1,81 @@
...
+  // add platform specific init stuff here
+}
+
+void CPlatformLinux::InitUniqueHardwareIdentifier()
+{
+  // try to read the machine-id from etc first
+  FILE *f = fopen("/etc/machine-id", "r");
+  if (f)
+  {
+      char line[256];
```
wsnipex: fyi: machine-id is defined as 32 chars(hex)

Memphiz: thx- will change on sunday

Memphiz:
I will look into the android part on sunday ...

MilhouseVH:
FYI this doesn't build on Ubuntu:
```
CPP     xbmc/addons/PVRClient.o
CPP     xbmc/platform/XbmcContext.o
CPP     xbmc/platform/overrides/linux/PlatformLinux.o
overrides/linux/PlatformLinux.cpp: In member function ‘virtual void CPlatformLinux::InitUniqueHardwareIdentifier()’:
overrides/linux/PlatformLinux.cpp:62:38: error: ‘strncmp’ was not declared in this scope
         if (strncmp(line, "Serial", 6) == 0)
                                      ^
overrides/linux/PlatformLinux.cpp:64:41: error: ‘strchr’ was not declared in this scope
           char *colon = strchr(line, ':');
                                         ^
/home/neil/projects/kodi.git/Makefile.include:97: recipe for target 'overrides/linux/PlatformLinux.o' failed
make[1]: *** [overrides/linux/PlatformLinux.o] Error 1
Makefile:428: recipe for target 'xbmc/platform/platform.a' failed
make: *** [xbmc/platform/platform.a] Error 2
make: *** Waiting for unfinished jobs....
```

fritsch:
@MilhouseVH mind including: <string.h>

xbmc/platform/overrides/linux/PlatformLinux.cpp
```diff
@@ -0,0 +1,81 @@
...
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with Kodi; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "PlatformLinux.h"
+#include <stdlib.h>
+#include <stdio.h>
```
MilhouseVH: Thanks @fritsch. Adding: `#include <string.h>` here fixes my build issue.

Memphiz:
rebased to master and added the android part (compile tested only atm - testbuild cooking on jenkins currently). I also added the missing include string.h on linux and changed the array size for machine-id to 33 (thx @fritsch for the hint) as suggested by wsnipex. 

@popcornmix if you could try this current state of the PR - i don't know why it wouldn't be called on rbpi. The expected call stack is lined out here:

http://pastebin.com/WbmaEkhN

Memphiz:
android arm testbuild here:
http://mirrors.kodi.tv/test-builds/android/arm/kodi-20160924-e4cbc81-hw_identifier-armeabi-v7a.apk

@MartijnKaijser you know someone who can test this? (if all is right it should show a md5 hash in the systeminfo->hardware dialog) - if all is wrong it either shows "NOUUID" or it crashs ...

Memphiz:
will try todo the windows version tomorrow if noone beats me to it ...

MilhouseVH:
Looks like Linux has a problem - latest PR, on Ubuntu with wired eth0: http://sprunge.us/iPgE

Memphiz:
@martijnkaijser is this the uuid you get?
d41d8cd98f00b204e9800998ecf8427e
Thats the MD5 of an empty String and would mean our jni Context is not yet initialised when we init the uuid ...

MartijnKaijser:
@Memphiz nope
screenshot_20160924-133819

This is  my NVIDIA Tablet and my Nexus Player has the same  number

Memphiz:
I need to add logging to see the plain ID - tomorrow

MilhouseVH:
> Looks like Linux has a problem - latest PR, on Ubuntu with wired eth0: http://sprunge.us/iPgE

What's odd is that the UUIID appears correctly in the GUI (same as screenshot from @MartijnKaijser,  above) but `HardwareUUID: NOUUID` is logged.

So on Linux there appears to be a logging issue (logging before UUID is available?)

This is repeatable on Ubuntu (wired, eth0) and LibreELEC (NUC, wifi, wlan0).

FernetMenta:
the issue is that logging is before ServiceManager has initialised CPlatform

Memphiz:
Ha! Good catch :)

MilhouseVH:
Looks good now on Linux, thanks: http://sprunge.us/KKVI

```
22:02:08   6.435283 T:139732003137600  NOTICE: special://masterprofile/playercorefactory.xml does not exist. Skipping.
22:02:08   6.452078 T:139732003137600  NOTICE: Running database version Addons26
22:02:08   6.452197 T:139732003137600  NOTICE: HardwareUUID: CEB455B58D74416F39D65402BAADAAD7
22:02:08   6.506041 T:139732003137600  NOTICE: ADDONS: Using repository repository.kodi.game
```

Memphiz:
wow @MilhouseVH have you some sort of a trigger whenever i push? :D

MilhouseVH:
I was in the process of building something else so it was no bother. :)

Memphiz:
@MartijnKaijser another testbuild for android which has the md5 sum switched off - can you give this a shot? i want to see what the playn androi_id value is:

http://mirrors.kodi.tv/test-builds/android/arm/kodi-20160924-8ab1bfe-hw_identifier-armeabi-v7a.apk

slack seems to be down for me ... (dns can't be resolved for the team slack url)

Memphiz:
@MartijnKaijser also for the json-rpc request

{"jsonrpc":"2.0","method":"XBMC.GetInfoLabels","params":{"labels": ["system.hardwareuuid"] },"id":1}

works for me - would this be enough?

Memphiz:
added windows version with the reg key. Built and tested via jenkins slave. We still have a problem with the overrides principle. ATM cmake defines PLATFORM_OVERRIDE because it found a PlatformWindow.cpp - sadly it does so for the whole kodi project and not only the Platform.cpp file. That means that now our ProcessInfo.cpp (which uses the same concept) thinks its overridden and so the CProcessInfo::CreateInstance isn't compiled in anymore. So atm this PR is compile broken on windows until we find a way to only set a compiler define on a single source file.

MartijnKaijser:
screenshot_20160925-082922
Here's Android

Memphiz:
@fritsch could you have a look at the jni commit? Any idea how it could hold the string "android_id" in lowercase instead of the id?

koying:
@Memphiz Looks like the default value:
http://androidxref.com/5.1.1_r6/xref/frameworks/base/core/java/android/provider/Settings.java#3504

@MartijnKaijser On what device was that?

Memphiz:
@fetzerch we seem to have a cmake problem here
https://github.com/xbmc/xbmc/blob/master/xbmc/platform/CMakeLists.txt#L18 - this define is set on windows because there was an override file found. For some reason its also defined here then:
https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/Process/ProcessInfo.cpp#L27

We need a solution where those defines are only hitting the files in the same folder like the cmakelists.txt or a possibility to make those defines per source file.

You can test this by simply compiling this pr on windows it will fail because of. The missing CProcessInfo CreateInstance because the PLATFORM_OVERRIDE is defined globally.

koying:
@Memphiz nvm ;)
ANDROID_ID is just the string identifying the setting.
You have to do an actual get for the value https://developer.android.com/reference/android/provider/Settings.Secure.html#getLong(android.content.ContentResolver,%20java.lang.String)

Memphiz:
Args ok - thx @koying

koying:
@Memphiz In case you wonder, you get context resolver there:
https://github.com/xbmc/xbmc/blob/master/xbmc/platform/android/jni/Context.cpp#L201

wsnipex:
@Memphiz quick fix/workaround for the define: make the 2 different ones

Memphiz:
@MartijnKaijser would you be so kind again:

http://mirrors.kodi.tv/test-builds/android/arm/kodi-20160925-4129ddc-hw_identifier-armeabi-v7a.apk

MartijnKaijser:
UUID phone: edb0a391192e33e9
UUID shield: 1e1df1a4980fddb8
UUID Nexus PLayer: 2b6ef6384dba72ad

Memphiz:
its because i removed the md5 hash for a test - that looks as expected - thx for testing - squashing time :)

xbmc/Application.cpp
```diff
@@ -574,6 +574,7 @@ bool CApplication::Create()
   else
     CLog::Log(LOGNOTICE, "ARM Features: Neon disabled");
 #endif
+
```
MartijnKaijser: unneeded

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -5525,7 +5525,12 @@ msgctxt "#12395"
 msgid "Battery level"
 msgstr ""
 
-#empty strings from id 12396 to 12599
+#: xbmc/windows/GUIWindowSystemInfo.cpp
```
MartijnKaijser: could you add a description text using  #.   ?

MartijnKaijser:
Need testing in windows as well?

Memphiz:
basically this needs testing on all platforms but osx/ios/android and we need some sort of sign off of each ...

Memphiz:
@MartijnKaijser added a description to the language change (i don't want this to be translated at all - to prevent stupid ideas of how to translate this ;) ). My time is over for now next up is my move into the flag - could you somehow ensure that platform devs test this / sign it off  and then take care of the merge in case we go forward to next beta?

MartijnKaijser:
@Paxxi @wsnipex @FernetMenta ping

MilhouseVH:
On RPi, I see the HardwareUUID being logged correctly (http://sprunge.us/aYGO) but it's not appearing in the System Info > Hardware section:

s1

I think there's not enough labels. If I apply [this patch](http://sprunge.us/UZdD) on top of this PR, then I see:

s2

Hedda:
Tip; UUID is in most cases displayed in a format with inserted hyphen (dash) characters:

e.g. 600e2d98-c0bf-4d6b-be53-88de15a01346 or {600E2D98-C0BF-4D6B-BE53-88DE15A01346}

https://en.wikipedia.org/wiki/Universally_unique_identifier

The uses of hyphens (dashes) makes it a little more human-readable when displayed.

Also, using all lowercase as standard on all platforms is also more common than using all uppercase

https://henbo.wordpress.com/2007/12/02/the-mystery-of-upper-case-and-lower-case-guid-values/

Memphiz:
@Hedda i know you mean good. But i would like to avoid a change on all platform files again for the sake of what is imo cosmetics. I only show the uuid in the GUI for transparency reasons. It is not ment to be human readable. If you want to change this you can do a PR against my repo no problem.

wsnipex:
quick test on ubuntu, shows the UUID in system info and log.

MartijnKaijser:
Windows shows
17:11:26 T:12744  NOTICE: HardwareUUID: E8F4044081A6C078F55F926F0F2CD05E

jenkins build this please

Paxxi (approved):
Windows bits looks fine

popcornmix:
If the md5sum of the mac address is used as a way of anonymising the data I wonder if it would be better moved up one layer so it also applies to platform specific implementations too.
Technically the Pi's MAC can be determined from the serial number, so I wonder if that should really be md5summed.
Or the md5sum sum could be added to the Pi specific implementation.

xbmc/platform/overrides/linux/PlatformLinux.cpp
```diff
@@ -0,0 +1,92 @@
...
+{
+  // try to read the machine-id from etc first
+  FILE *f = fopen("/etc/machine-id", "r");
+  if (f)
+  {
+      char line[33]; // 32 + zero byte
+      if (fgets(line, sizeof line, f) != nullptr)
+      {
+          m_uuid = line;
+#if defined(_DEBUG)
```
wsnipex: don't we have log levels for a reason?

wsnipex (commented):
I know its annoying, but if we want to merge this for v17, we really also need autotools fixed up

Memphiz:
@popcornmix could you clarify your comment please? I don't understand it :) - if it is about using the mac address instead of the serialnumber - i wanted to use mac only in worst case scenarios where everything else failed. Because i don't trust the network interfaces being up when needed (well as of murphy we will init the uuid in exactly that moment where the if is down ...) - thats why i did the specific implementations in the first place - to have something that works from the first second and can't be easily changed (which is at least true for ios/osx/android and maybe windows ...)

Memphiz:
logging of uuid removed for privacy reasons (logs from forum users would allow to bind usernames to uuids ...)

Memphiz:
Jenkins build this please

ronie:
may i suggest to start a public discussion on the forum before merging this?

kib:
For forum readers visiting github: 
It is not silent on this subject, but we are having an internal discussion about it currently before involving the rest of the community. This code is not merged and likely will not be merged before a team consensus on scope and size and a public involvement on the forum has taken place. 

Team members can find my lengthy post on intended usage discussed during Devcon [here](http://forum.kodi.tv/showthread.php?tid=292162&pid=2442361#pid2442361) in the team discussion thread.

_Further comments on the PR should be technical in dealing with how to properly do a UUID on different platforms._

deejayetech (approved):

Rechi:
@Memphiz this needs a rebase

Memphiz:
Not Interested in working on this any further - sorry ...

PR #10546, opened 2016-09-23
upgrade autotools for aarch64 depends build

ghost:
upgrade autotools for aarch64 depends build
## Description

upgrade autotools for aarch64 depends build
config.sub/guess are the latest ones from autotools side
## Motivation and Context

upgrade autotools for aarch64 depends build
## How Has This Been Tested?

cross compiled on ubuntu x86-64 14.04 lts
## Types of change


tools/depends/configure.ac
```diff
@@ -169,6 +169,18 @@ case $host in
       platform_os="linux"
     fi
   ;;
+  aarch64*-*-linux-gnu*|aarch*-*-linux-be*)
+    if test "$use_platform" = "auto"; then
+      if test "x$use_cpu" = "xauto"; then
+        use_cpu=$host_cpu
+      fi
+      use_toolchain="${use_toolchain:-/usr}"
+      platform_cflags="-Os -fPIC -DPIC"
```
wsnipex: are we sure we want -Os here? @stefansaraev any opinion?

wsnipex (commented):
thanks!
a bit hard to review due to the config.sub/guess update, but looks good to me.

wsnipex:
jenkins build this please

Memphiz:
mhh depends was already able to build with aarch64 / arm64 for ios/tvos. I have to check if this is still working with this PR - but can't do it until sunday.

tools/depends/configure.ac
stefansaraev: it's a c/p from arm*-*-linux-gnu few lines above and it does not hurt.

Memphiz:
tvos and ios arm64 depends still building fine with this ... good from me.

mapfau:
still not 100% finished.... final commit will arrive soon

wsnipex:
jenkins build this please

Memphiz:
That last commit sounds like it would need a change for Jenkins too here no?:

https://github.com/xbmc/xbmc/blob/master/tools/buildsteps/defaultenv#L33

IIRC API 21 is installed on the slaves already

hudokkow:
jenkins build this please

mapfau:
1.) stripping still fails from time to time (make apk)

2.) Some addons fail with the droid::32bit build:
Jenkins: http://jenkins.kodi.tv/job/Android-ARM/9361/

@wsnipex @garbear here is one example:

"/home/jenkins/workspace/android-arm/tools/depends/target/binary-addons/arm-linux-androideabi-android-21/build/pcre/src/pcre/pcrecpp.cc:855:50: error: 'strtouq' was not declared in this scope
  unsigned long long r = strtouq(str, &end, radix);

Any idea (strouq????, whats this?)

wsnipex:
@Memphiz yes, this needs ndk 12. I tried with ndk 10, but it fails on aarch64.

@mapfau some addons always fail, we won't be able to fix this in this PR. As long as not _all_ fail, I wouldn't care atm.

mapfau:
most of the addons are running well.
some fail because of header files (sys/timeb.h), some because of reasons as above.

mapfau:
will fix soon :-)

wsnipex:
as said on slack, you have to rebase project/cmake/modules/FindLibDvd.cmake

xbmc/cores/DllLoader/exports/CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ elseif(NOT CORE_SYSTEM_NAME STREQUAL windows)
   add_custom_target(wrapper.def ALL nm ${CMAKE_CURRENT_BINARY_DIR}/CMakeFiles/wrapper.dir/wrapper.c.o | grep __wrap | awk '{ printf(\"%s \", \$\$3) }' | sed \"s/___wrap_/__wrap_/g\" | sed \"s/__wrap_/-Wl,-wrap,/g\" > wrapper.def)
 
   if(CORE_SYSTEM_NAME STREQUAL android)
-    add_custom_command(TARGET wrapper.def COMMAND echo \"-l$<TARGET_FILE:${APP_NAME_LC}>\" >> wrapper.def)
+    add_custom_command(TARGET wrapper.def COMMAND echo \"-L${kodi_BINARY_DIR} -l:libkodi.so\" >> wrapper.def)
     add_dependencies(wrapper.def ${APP_NAME_LC})
```
fetzerch: You could use the TARGET_FILE_DIR and TARGET_FILE_NAME generator expression. (https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html)
They have the advantage that cmake already complains if the library target does for some reason not exist. So it makes things more robust. `kodi` should always be APP_NAME_LC, otherwise it breaks builds where someone tries to rebrand.  Also I'm wondering where the kodi_BINARY_DIR comes from.

xbmc/platform/android/jni/AudioFormat.h
```diff
@@ -19,6 +19,8 @@
  *
  */
 
+#define HAVE_INTTYPES_H 1
```
fetzerch: are these needed when you set the define from cmake/autotools?

xbmc/cores/DllLoader/exports/CMakeLists.txt
wsnipex: this is wrong anyway, it is -l[name] without lib prefix and .so suffix.
 i.e. -lkodi
Anyway, autotools links against libdummy-kodi, so see https://github.com/wsnipex/xbmc/commit/07477c1dbd7a6f953636a8e2ba7089cbca3c6134

mapfau: its fom an android documentation with ":", seems to work well.
But, why use dummy? The problem seems more that the full path is given in -l

mapfau: Pls let me know what way you prefer. I'll make the final shot then

wsnipex: I prefer it the same as in autotools unless someone can explain why libdummy-kodi is not needed.
You can simply reset on my branch, I also fixed an rebase error there.

xbmc/platform/android/jni/AudioFormat.h
mapfau: @wsnipex I agree, they shouldn't be in source files but in cmake project

wsnipex: see top 2 commits: https://github.com/wsnipex/xbmc/commits/aarch64

warning: this is _not_ build tested

peak3d:
moved to https://github.com/xbmc/xbmc/pull/10659

PR #10550, opened 2016-09-24
Revert "[osx/readme] - remove KODI_HOME bits from readme (not needed …

Memphiz:
No codechange - just readme. Default XCode setup seems to build the binaries out of tree where we have no chance to finde our KODI_HOME programmatically.
This reverts only the readme bit.


Memphiz:
@ksooo done

PR #10552, opened 2016-09-24
VideoPlayer: fix wav playback on win32

FernetMenta:
see title

@Paxxi @fetzerch @fritsch my guess is that the VS compiler does optimize away buffer_size because it is only used in an conditional expression for the function call. I hate this coding style anyway because you can't inspect the value of buffer_size before the function is called.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
@@ -299,8 +299,9 @@ bool CDVDDemuxFFmpeg::Open(CDVDInputStream* pInput, bool streaminfo, bool filein
 
         // av_probe_input_buffer might have changed the buffer_size beyond our allocated amount
         int buffer_size = std::min((int) FFMPEG_FILE_BUFFER_SIZE, m_ioContext->buffer_size);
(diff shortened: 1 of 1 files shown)


FernetMenta:
I want to thank mention-bob in advance :)

fritsch:
I hessitated to do a similar fix, though we knew it was working. I fear that we might cover an issue somewhere else, but as we did not find anything after really trying to do so. Go for that.

Paxxi:
Let's leave it open for now, if we can't find it and it's holding up the release we can merge it

FernetMenta:
@Paxxi I want to get this is because plugin devs depend on it. If you want to do further experiment, you can revert it on your system.

Paxxi:
Oh, if it's causing issues for them currently then sure, let's merge it so they can continue to work.

FernetMenta:
The author of the spotify addon brought this to my attention first.

FernetMenta:
jenkins build this please

fetzerch:
Just for reference, drmemory claims an uninitialized register access and that also looks like an optimization issue to me.

I've executed gflags and drmemory with both debug and release builds. On debug builds there are no warnings. Also not when moving the probe_buffer from stack to heap.
Additionally I've checked with valgrind on linux and looked at the coverity findings. There's absolutely no indication for an error.

PR #10553, opened 2016-09-24
Setting album user rating using album ID not GetAlbumIdByPath

DaveTBlake:
Fixes a bug when setting album user rating when the user has songs from more than one album in the same folder.

There is nothing to say that all albums must have a unique folder that does not contain songs from any other album, even if this is the common way to organise your music files.  Path may not be sufficient to identify an album uniquely. Hence `GetAlbumIdByPath` needs to allow for more than one album being found for a path, just as `GetAlbumByName` and `GetArtistByName` do, rather than return the ID of the first in the results set.

Setting album user rating was using path to attempt to identify the album when the unique album ID is readily available. Switch to use ID not path.

Also `SetAlbumVotes` was removed as it was unused. The idea seems to be that "rating" and "votes" are values loaded from scraping NFO or online sources, and are never set via the GUI, hence this routine was unnecessary as  votes are never set in isolation ( and there is no routine to set rating).  Rating and votes are loaded by `CAlbum::Load`.

@Razzeee perhaps you would like to comment as you did the original work on this in  #8405
@zag2me most likely to have data to test this out.

### DIFF
--- a/xbmc/music/MusicDatabase.cpp
+++ b/xbmc/music/MusicDatabase.cpp
@@ -2167,12 +2167,18 @@ int CMusicDatabase::GetAlbumIdByPath(const std::string& strPath)
 {
   try
   {
-    std::string strSQL=PrepareSQL("select distinct idAlbum from song join path on song.idPath = path.idPath where path.strPath='%s'", strPath.c_str());
-    m_pDS->query(strSQL);
-    if (m_pDS->eof())
-      return -1;
+    if (NULL == m_pDB.get()) return false;
+    if (NULL == m_pDS.get()) return false;
 
-    int idAlbum = m_pDS->fv(0).get_asInt();
+    std::string strSQL = PrepareSQL("SELECT DISTINCT idAlbum FROM song JOIN path ON song.idPath = path.idPath WHERE path.strPath='%s'", strPath.c_str());
+    // run query
+    if (!m_pDS->query(strSQL)) return false;
+    int iRowsFound = m_pDS->num_rows();
+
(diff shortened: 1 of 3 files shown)


DaveTBlake:
To test this fix put music files from more than one album all in the same folder. Add folder as new music source and scan into music library. Edit the user rating for the different albums from the Album Info dialog, and see them being correctly applied when you look at the albums node (sorted by "my rating").

Without this fix when you change the user rating for an album once you leave the Album Info dialog then actual edit is applied to the first album.

Razzeee:
LGTM 

Have you tried if votes can still be set via json?

DaveTBlake:
Votes, for songs nor albums, have never been added as something that can be set from JSON

Razzeee:
Sorry, I always forget how scrapping is done. So I didn't ment json, but
reading from nfo.

DaveTBlake:
Yes, reading of votes, rating and userrating from NFO is fine. All done within CAlbum::load, SetAlbumVotes is not called anywhere.  But, as I mentioned on Slack, album rating is limited to 5. I will fix that

zag2me:
Probably a Feature Request... but setting the album and song rating via JSON would be very useful to me to cloud backup my user ratings.

DaveTBlake:
PR #10570  to fix the album rating range from being trucated at 5 when loaded.

@zag2me you can set rating and user rating for both songs and albums via JSON, just not votes. But if votes were to be added by another PR it still would not need the SetAlbumVotes routine I removed.

DaveTBlake:
jenkins build this please

DaveTBlake:
jenkins build this please

PR #10554, opened 2016-09-24
Cache controller layouts

garbear:
AddonMgr used to create a new uninitialised controller add-on every time GetAddon() was called. This means that CController::LoadLayout() was called a whole bunch. In some extreme cases, like every time a button is mapped in the controller dialog, the layout would be reloaded from disk three times.

This change caches controller add-ons so that the layout is only loaded once (and reloaded if the add-on version is updated).

The strategy for caching controllers is identical to binary add-ons, so I reused the binary add-on mgr, even though controllers are just XML files.

This change exposes a pre-existing bug with cyclical dependencies in the service manager. Kodi crashes on startup if a system add-on in addonmanifest.xml has a type that is cached by the binary add-on manager. I fixed this by creating the binary add-on cache, then initializing the add-on manager (which uses the empty binary add-on cache), then initializing the binary add-on cache. Not sure if there's a better approach.

### DIFF
--- a/xbmc/ServiceManager.cpp
+++ b/xbmc/ServiceManager.cpp
@@ -45,7 +45,9 @@ bool CServiceManager::Init1()
 bool CServiceManager::Init2()
 {
   m_Platform->Init();
-  
+
+  m_binaryAddonCache.reset(new ADDON::CBinaryAddonCache());
+
   m_addonMgr.reset(new ADDON::CAddonMgr());
   if (!m_addonMgr->Init())
   {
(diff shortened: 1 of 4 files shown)


tamland:
This globals stuff is a no-go. GetRunningInstance and OnPostInstall are both mistakes that will be removed and should not be used for anything new.
You should this move somewhere more appropriate as it's not the job of the addon system to handle lazy loading of controller layouts.

garbear:
I'll think of another approach

PR #10556, opened 2016-09-24
[estuary] check if fanart pack is selected

ronie:
don't try to display fanart if no fanart pack is selected.
also changed the settings label, as the fanart pack is not only used on the home window.

@phil65


phil65:
👍

PR #10557, opened 2016-09-24
[paplayer] fix codec deinit for audio decoder add-on (was never called)

AlwinEsch:
## Description

Fix of never called decoder deinit
## Motivation and Context

Has found during my add-on system rework that on audio decoder add-on the `DeInit` was never called and a mem leak with this come.

This add the use of `ICodec::DeInit` call to the `CAudioDecoder::Destroy()` function.

On the other class (`VideoPlayerCodec`) who use also the `ICodec` is it done from his destructor.
## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/AudioDecoder.cpp
+++ b/xbmc/addons/AudioDecoder.cpp
@@ -46,6 +46,11 @@ CAudioDecoder::CAudioDecoder(AddonProps props, std::string extension, std::strin
   m_strExt = std::move(strExt);
 }
 
+CAudioDecoder::~CAudioDecoder()
+{
+  DeInit();
+}
+
 bool CAudioDecoder::Init(const CFileItem& file, unsigned int filecache)
 {
   if (!Initialized())
--- a/xbmc/addons/AudioDecoder.h
+++ b/xbmc/addons/AudioDecoder.h
@@ -56,7 +56,7 @@ namespace ADDON
     CAudioDecoder(AddonProps props, std::string extension, std::string mimetype, bool tags,
         bool tracks, std::string codecName, std::string strExt);
 
-    virtual ~CAudioDecoder() {}
+    virtual ~CAudioDecoder();
 
     // Things that MUST be supplied by the child classes
     bool Init(const CFileItem& file, unsigned int filecache) override;
--- a/xbmc/cores/paplayer/ICodec.h
+++ b/xbmc/cores/paplayer/ICodec.h
@@ -56,11 +56,6 @@ public:
   // 3.  Fill in the m_TotalTime, m_SampleRate, m_BitsPerSample and m_Channels parameters.
   virtual bool Init(const CFileItem &file, unsigned int filecache)=0;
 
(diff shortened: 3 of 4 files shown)


xbmc/cores/paplayer/AudioDecoder.cpp
```diff
@@ -55,6 +55,8 @@ void CAudioDecoder::Destroy()
   CSingleLock lock(m_critSection);
   m_status = STATUS_NO_FILE;
 
+  if (m_codec)
+    m_codec->DeInit();
   m_pcmBuffer.Destroy();
 
   if ( m_codec )
```
AlwinEsch: Maybe another way can be to add it on `virtual ~ICodec() {};`???

FernetMenta:
thanks bob for pinging the right people :)

xbmc/cores/paplayer/AudioDecoder.cpp
```diff
@@ -58,7 +58,10 @@ void CAudioDecoder::Destroy()
...
+  {
+    m_codec->DeInit();
```
FernetMenta: I don't see what is here. Now you call Deinit before the destructor.
This change violates the principle of encapsulation. Why do you want to call DeInit from the outside?

FernetMenta: What does this entire PR change in behaviour?

AlwinEsch: An alternative way can be to remove the:
```cpp
  // DeInit()
  // Should just cleanup anything as necessary.  No need to free buffers here if they
  // are allocated and destroyed in the destructor.
  virtual void DeInit()=0;
```
complete and to do this on destructors from `ADDON::CAudioDecoder` and like before on `VideoPlayerCodec`. What do you think about them?

FernetMenta: please answer my question first

AlwinEsch: It change in behaviour nearly nothing, the only new thing is that a function who was not called on audio decoder add-on's before is called now.

Not sure if it was never called or it was lost on a update of PAPlayer.

AlwinEsch: Since the `DeInit` is set in` ICodec` I am assuming that it will be called from the outside and not in the child class itself uses the `ICodec`.

FernetMenta: ah ok, we are talking about an addon issue here. I would drop DeInit from the interface ICodec and make the addon CAudioDecoder behave like VideoPlayerCodec. Do deinit in the destructor: https://github.com/xbmc/xbmc/blob/master/xbmc/addons/AudioDecoder.h#L59

AlwinEsch:
Is updated and during test the DeInit on add-on itself was called good.

jenkins build this please

FernetMenta:
+1

PR #10558, opened 2016-09-24
[epg] fix mouse focus

ronie:
mouse focus on the epggrid was a bit off.

before:
before

after:
after

@ksooo @xhaggi

### DIFF
--- a/xbmc/epg/GUIEPGGridContainer.cpp
+++ b/xbmc/epg/GUIEPGGridContainer.cpp
@@ -838,8 +838,8 @@ bool CGUIEPGGridContainer::SelectItemFromPoint(const CPoint &point, bool justGri
   if (!m_focusedProgrammeLayout || !m_programmeLayout || (justGrid && point.x < 0))
     return false;
 
-  int channel = MathUtils::round_int(point.y / m_channelHeight);
-  int block   = MathUtils::round_int(point.x / m_blockSize);
+  int channel = point.y / m_channelHeight;
+  int block   = point.x / m_blockSize;
 
   if (channel > m_channelsPerPage)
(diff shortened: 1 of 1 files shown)


ksooo (approved):
+1

ksooo:
As I'm not an expert in this mouse stuff it might be good to get approval from @xhaggi as well..

ksooo:
I think the right fix has to look a little different. Instead of the changes done in this PR, do:

<pre>
diff --git a/xbmc/epg/GUIEPGGridContainer.cpp b/xbmc/epg/GUIEPGGridContainer.cpp
index 8cfa281..5ac7735 100644
--- a/xbmc/epg/GUIEPGGridContainer.cpp
+++ b/xbmc/epg/GUIEPGGridContainer.cpp
@@ -838,8 +838,8 @@ bool CGUIEPGGridContainer::SelectItemFromPoint(const CPoint &point, bool justGri
   if (!m_focusedProgrammeLayout || !m_programmeLayout || (justGrid && point.x < 0))
     return false;
 
-  int channel = MathUtils::round_int(point.y / m_channelHeight);
-  int block   = MathUtils::round_int(point.x / m_blockSize);
+  int channel = point.y / m_channelHeight;
+  int block   = point.x / m_blockSize;
 
   if (channel > m_channelsPerPage)
     channel = m_channelsPerPage - 1;
</pre>

@ronie feel free to change the PR this way, if you are with me.

ronie:
yup, that's it. it also fixes the small offset on the x-axis i noticed.

xhaggi:
looks good too

ksooo:
jenkins build this please

ksooo:
jenkins errors are unrelated

PR #10562, opened 2016-09-25
[depends] sqlite 3140200

MartijnKaijser:
## Motivation and Context

keep it updated
## How Has This Been Tested?

runtime


MartijnKaijser:
We seem keep shipping two patches. Does it make sense that some one tries to send these upstream?

Paxxi:
Are the patches even needed?

MartijnKaijser:
@koying added one of them. The other is from  @bavison

MartijnKaijser:
jenkins build this please

PR #10375, opened 2016-09-01
[IMX] compilation error fix & others

mk01:

PR #10563, opened 2016-09-25
[cmake] [IMX] add cmake option -DENABLE_IMXVPU to add and complete su…

sraue:
…pport for IMX based devices. Rename IMX_FOUND to IMXVPU_FOUND so it matches with the needed libraries.

build and runtime tested, building needs https://github.com/xbmc/xbmc/pull/10375

ping @wsnipex


project/cmake/modules/FindIMXVPU.cmake
```diff
@@ -0,0 +1,27 @@
+# - Try to find IMXVPU
+# Once done this will define
+#
+# IMXVPU_FOUND - system has IMXVPU
+# IMXVPU_INCLUDE_DIRS - the IMXVPU include directory
+# IMXVPU_LIBRARIES - The IMXVPU libraries
+
```
fetzerch: Please adapt as in https://github.com/xbmc/xbmc/commit/d8877a284c63350dd0cea25c0275a258f6cc4a4d.

XX_INCLUDE_DIRS and XX_LIBRARIES should not be cached vars (so they can't be used directly by find_path/library. Use XX_INCLUDE_DIR and XX_LIBRARY. The paths from package config should can hint find_path/library by setting PATHS. Docs: https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#find-modules

lrusak:
@fetzerch how about something like this? https://github.com/lrusak/xbmc/commit/6d60b5046861ab844c055b25b2ba2da2b8b323e9

lrusak:
@sraue do you want to update this PR with the changes here https://github.com/lrusak/xbmc/commit/6d60b5046861ab844c055b25b2ba2da2b8b323e9

PR #10564, opened 2016-09-26
[python] change signature for ListItem.setRating

phil65:
At the moment the signature of ListItem.setRating contains a keyword argument called "def". Since this is a builtin keyword for python we should use something else for this, otherwise it cant be used explicitely. This PR adjusts it to use "defaultt", as we do in other parts of the API, too.
Also fixes some stuff in the docs.

@tamland

### DIFF
--- a/xbmc/interfaces/legacy/ListItem.cpp
+++ b/xbmc/interfaces/legacy/ListItem.cpp
@@ -157,12 +157,12 @@ namespace XBMCAddon
         vtag.SetUniqueID(it->second, it->first);
     }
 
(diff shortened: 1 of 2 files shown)


InuSasha:
defaultt with two t's?

MartijnKaijser:
@InuSasha default is a keyword as well hence we use defaultt

InuSasha:
@MartijnKaijser ah good to know, sorry for the noise

tamland (approved):

phate89 (approved):

phil65:
I think this qualifies as a fix and can get merged?

PR #10567, opened 2016-09-26
[infomanager] add player.dbid infolabel

BigNoid:
Add player.dbid as infolabel
## Motivation and Context

We have listitem.dbid so it makes sense to have player.dbid. It may be useful for starting scripts from the video osd.
## How Has This Been Tested?

Runtime tested. Infolabel returns the same info as listitem.dbid
## Types of change

## Checklist:


xbmc/GUIInfoManager.cpp
```diff
@@ -506,9 +506,10 @@ const infomap player_labels[] =  {{ "hasmedia",         PLAYER_HAS_MEDIA },
                                   { "pauseenabled",     PLAYER_CAN_PAUSE },
                                   { "seekenabled",      PLAYER_CAN_SEEK },
                                   { "channelpreviewactive", PLAYER_IS_CHANNEL_PREVIEW_ACTIVE},
-                                  { "tempoenabled", PLAYER_SUPPORTS_TEMPO},
-                                  { "istempo", PLAYER_IS_TEMPO},
-                                  { "playspeed", PLAYER_PLAYSPEED}};
+                                  { "tempoenabled",     PLAYER_SUPPORTS_TEMPO},
```
FernetMenta: please keep as it was. we agreed on no stupid vertical alignment anymore

MartijnKaijser:
jenkins build this please

xbmc/GUIInfoManager.cpp
```diff
@@ -508,6 +508,7 @@ const infomap player_labels[] =  {{ "hasmedia",         PLAYER_HAS_MEDIA },
                                   { "channelpreviewactive", PLAYER_IS_CHANNEL_PREVIEW_ACTIVE},
                                   { "tempoenabled", PLAYER_SUPPORTS_TEMPO},
                                   { "istempo", PLAYER_IS_TEMPO},
+                                  { "dbid", PLAYER_DBID },
```
FernetMenta: this is wrong section. dbid is no player state

BigNoid:
@FernetMenta moved it to underneath title now

FernetMenta:
I don't see any change. dbid is not a player state. dbid an attribute of the videotag. Label needs to be something like XXX_DBID, but not PLAYER_DBID

FernetMenta:
I would expect this in MUSICPLAYER_xx or VIDEOPLAYER_xx
There is also a section PLAYER_ITEM_XX, maybe this one fits but not sure why this was introduced.

BigNoid:
> dbid an attribute of the videotag. Label needs to be something like XXX_DBID, but not PLAYER_DBID

This is true from a dev point of view, but for the ones who this is exposed too it is an attribute of the player as it is for both video and music.
I have no problems changing it to ```musicplayer.dbid``` and ```videoplayer.dbid``` as it returns empty for everything but database items

FernetMenta:
> but for the ones who this is exposed too

Those are also devs who need to understand what they are want to show on the UI. Otherwise we get something like the DisplayAfterSeek messed up states we had a couple of weeks ago.

BigNoid:
> Those are also devs who need to understand what they are want to show on the UI. Otherwise we get something like the DisplayAfterSeek messed up states we had a couple of weeks ago.

I see your point. For me player label was always if it was generic for both audio and video, like ```player.title``` and ```player.art(xx)```

FernetMenta:
compare the code of PLAYER_TITLE with the code of VIDEO/MUSICPLAYER_TITLE
the first is a mess compared to the latter where the tag is already accessed any new attributes can be easily added.

BigNoid:
@FernetMenta updated now to ```music player.dbid``` and ```video player.dbid```. Thx for the review, this looks much cleaner.

ronie:
iirc the long term goal was to drop the separate musicplayer.xxx / videoplayer.xxx labels completely and only use player.xxx ones instead.
http://forum.kodi.tv/showthread.php?tid=106772&pid=853324#pid853324

no strong opinion either way, just as a fyi :-)

FernetMenta:
@BigNoid looks good now

@ronie this is a rather old post. Considering how the app has evolved, I don't think this makes sense anymore. Player itself exposes more and more info that separated views on player and items are more appropriate.

BigNoid:
jenkins build this please

PR #10569, opened 2016-09-27
Compilation albums - all songs marked means album is compilation (fixes Trac tickets)

DaveTBlake:
There is an omission in the current logic for determining if an album is a compilation from the music file tags. Fundamentally if all the songs from an album are tagged as being part of a compilation (via the TCMP or COMPILATION tag) then, regardless of album artist or the lack variation of song artists across the album, the album should be flagged as a compilation in the music libary. But this does not currently happen consistently, only for those albums that also have musicbrainz id tags.

For music files without Musicbrainz id tags the current processing logic recognises an album is a compliation if 
    1. album name is non-empty AND
    2a. no tracks overlap OR
    2b. all tracks are marked as part of compilation AND
    3a. a unique primary artist is specified as "various" or "various artists"  OR
    3b. we have at least two primary artists and no album artist specified.

It means that compilation albums are picked up even if the music files are not tagged as such, but just from the "various" album artist, or the lack of album artist but a diversity of song artists. All this is correct and remains unchanged.

However users may want to tag anthologies - a collection of recordings from various old sources combined together such as a "best of", retrospective or rarities type release - as a compilation. Or they may want to use album artist to differentiate between different kinds of song artist varieties. Tagging using Picard can certainly produce both of these examples, and because of the presence of mbids the albums are flagged in the music library as compilations. The same album without the mbids is not.

So this fix makes the treatment of anthologies consistent, by ensuring when all the songs from an album are tagged as being part of a compilation that the album is flagged as a compilation. 

A minor addition is also  made to accept the local language setting of "various artists" in the album artist tag as indicating a compilation.

Note we do not need to use album artist name as an indicator of compilation, we have a separate data field to do that, and it is far better design to use that flag rather than do string comparisons on artist name. Kodi will continue to apply the local language VA string (at the time the music is scanned) as the album artist for those compilation albums that satisfy 3a or 3b above. But many of the language changing issues associated with VA can be mitigated by using the compilation flag in filters and rules.

I believe this fixes trac tickets 15773 and 15812
Discussed on forum here http://forum.kodi.tv/showthread.php?tid=290477
But who would like to approve?  @jjd-uk @zag2me @Razzeee

### DIFF
--- a/xbmc/music/infoscanner/MusicInfoScanner.cpp
+++ b/xbmc/music/infoscanner/MusicInfoScanner.cpp
@@ -674,35 +674,53 @@ void CMusicInfoScanner::FileItemsToAlbums(CFileItemList& items, VECALBUMS& album
     }
 
     /*
-     We have a compilation if
-     1. album name is non-empty AND
-     2a. no tracks overlap OR
-     2b. all tracks are marked as part of compilation AND
-     3a. a unique primary artist is specified as "various" or "various artists" OR
-     3b. we have at least two primary artists and no album artist specified.
-     */
+    We have a Various Artists compilation if
+    1. album name is non-empty AND
+    2a. no tracks overlap OR
+    2b. all tracks are marked as part of compilation AND
+    3a. a unique primary artist is specified as "various", "various artists" or the localized value
+    OR
+    3b. we have at least two primary artists and no album artist specified.
+    */
+    std::string various = g_localizeStrings.Get(340); // Various Artists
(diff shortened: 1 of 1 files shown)


jjd-uk:
I'll try to fit in some testing, probably won't have a lot of time this week though but I'll see what I can do.

DaveTBlake:
There is a test build here http://mirrors.kodi.tv/test-builds/win32/KodiSetup-20160930-db2aaf8-musictest01.exe that includes this PR and the Album Record Label and Type from music files tags fix (10573)

DaveTBlake:
Since this fixes 2 tickets I would like to get it into beta please. Anyone object?

PR #10570, opened 2016-09-27
Fix album rating to 0 to 10 range when loaded

DaveTBlake:
Left out by #8405 when album rating was changed from 0 to 5 range.

@Razzeee @zag2me @MartijnKaijser


Razzeee:
Jenkins build and merge

PR #10572, opened 2016-09-27
[CryptThreading] replace hardcoded value for CRYPTO_LOCK by macro

Rechi:
## Description

Replace hard coded value for CRYPTO_LOCK by macro.
## Motivation and Context

Cleanup

### DIFF
--- a/xbmc/utils/CryptThreading.cpp
+++ b/xbmc/utils/CryptThreading.cpp
@@ -50,6 +50,7 @@ GCRY_THREAD_OPTION_PTHREAD_IMPL;
 
 /* ========================================================================= */
 /* openssl locking implementation for curl */
+#if defined(HAVE_OPENSSL) && OPENSSL_VERSION_NUMBER < 0x10100000L
 static CCriticalSection* getlock(int index)
 {
   return g_cryptThreadingInitializer.get_lock(index);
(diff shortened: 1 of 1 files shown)


MartijnKaijser:
@wsnipex ping

wsnipex:
it doesn't look wrong, but does it fix anything? @jimfcarroll ping, since you added this

MartijnKaijser:
@wsnipex related? http://trac.kodi.tv/ticket/16760

wsnipex:
only in so far that we can scrap the whole locking function once openssl 1.1.x is used.
So it would be great if @Rechi could ifdef this for openssl versions <= 1.1.x

Rechi:
@wsnipex do you really mean **<= 1.1.x**?

wsnipex:
not literally. crypto locking can/should only be used on versions < 1.1.0

Rechi:
I mean if I should check `version < 1.2` (what your first comment means) or `version < 1.1`? (obviously no string checks)

wsnipex:
< 1.1.0

Rechi:
@wsnipex updated

MartijnKaijser:
jenkins build this please

wsnipex:
thanks! Can someone try this with openssl 1.1.0? maybe @MilhouseVH ?

MilhouseVH:
I can include this in tonight's LibreELEC build, however we use LibreSSL 2.3.5 instead of OpenSSL (LibreSSL has `#define OPENSSL_VERSION_NUMBER 0x20000000L`). This builds OK, not entirely sure how to test...

MilhouseVH:
This has been in my builds for 2 weeks, no reported issues.

jimfcarroll:
This was originally added due to inconsistencies in curl libraries. I don't recall all of the details but I'd love to see it go if there's a better way to handle it now.

PR #10573, opened 2016-09-27
Album Record Label and Type from music files tags 

DaveTBlake:
Albums have label and type fields that can currently only be filled by scraping from NFO files and maybe from online sources (?). Yet music files can be tagged with record label and release type information that could be used to populate these fields.

This adds processing of the LABEL (TPUB)  and RELEASETYPE (MUSICBRAINZ ALBUM TYPE) tags and uses them to poplate the album library data. Both label and type fields are avaliable in smart playlist rules, so this extends the way that users can explore their music  collection.

Note the album table also has a "releasetype" field, this is used internally to flag if the entry in the album table is a real "album" or a "single" e.g. a fake album created to cover all the singles by that artist. We must take care not to confuse the two.

### DIFF
--- a/xbmc/music/Album.cpp
+++ b/xbmc/music/Album.cpp
@@ -174,6 +174,7 @@ CAlbum::CAlbum(const CFileItem& item)
 
   iYear = stTime.wYear;
   strLabel = tag.GetRecordLabel();
+  strType = tag.GetMusicBrainzReleaseType();
   bCompilation = tag.GetCompilation();
   iTimesPlayed = 0;
   dateAdded.Reset();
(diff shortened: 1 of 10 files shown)


MartijnKaijser:
No db bump?

FernetMenta:
If this is no fix, and I don't see the fix label, I suggest to remove the Krypton label. DB changes are always risky and should be avoided on short final

MartijnKaijser:
i don't see short final happening soon with current issues popping up

FernetMenta:
we can focus on current issues and try to resolve them. if some of us try and other use this time to throw new risky code in, we'll never get somewhere. 
we did agree on fixes only on devcon, didn't we?

DaveTBlake:
No db change needed guys, the data is already in the structure we just aren't filling it. Could call that a broken thing and this a fix  rather than an improvement. 

No risky code changes either, I know better than that for beta.

PR #10575, opened 2016-09-27
[WIP] Estuary settings window changes

jjd-uk:
## Description

As discuss at Devcon, the layout of the Settings window is being adjusted to accomodate File Manager, Add-on Browser, Event Log, and System Info (renamed here to "About system")
## Motivation and Context

Keep Settings and system utilities one one place
## How Has This Been Tested?

As only xml file changes, this was tested by replacing the relevant files from nightly build 27/09/2016

estuary1

estuary2
## Types of change

## Checklist:

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -7521,67 +7521,45 @@ msgctxt "#14200"
 msgid "Player"
 msgstr ""
 
-#: system/settings/settings.xml
-msgctxt "#14201"
-msgid "Player settings"
-msgstr ""
+#empty string id 14201
 
 #: system/settings/settings.xml
 msgctxt "#14202"
 msgid "Library"
 msgstr ""
 
-#: system/settings/settings.xml
-msgctxt "#14203"
-msgid "Library settings"
-msgstr ""
+#empty string id 14203
 
 #: system/settings/settings.xml
 msgctxt "#14204"
 msgid "TV"
 msgstr ""
 
-#: system/settings/settings.xml
-msgctxt "#14205"
-msgid "TV settings"
-msgstr ""
+#empty string id 14205
 
 #: system/settings/settings.xml
 msgctxt "#14206"
 msgid "Interface"
 msgstr ""
 
-#: system/settings/settings.xml
-msgctxt "#14207"
-msgid "Interface settings"
-msgstr ""
-
-#: system/settings/settings.xml
-msgctxt "#14208"
-msgid "Service settings"
-msgstr ""
-
-#: system/settings/settings.xml
-msgctxt "#14209"
-msgid "System settings"
-msgstr ""
-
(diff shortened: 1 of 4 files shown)


jjd-uk:
Some internal discussion of what the Add-ons button should do exactly at http://forum.kodi.tv/showthread.php?tid=292274

Additional pings for @MartijnKaijser @ronie @HitcherUK @da-anda 

Feedback also needed for Home.xml and Includes_Buttons.xml changes to get the 3 button bottom menu bar, did this by eye so the size changes may not be 100% right.

ronie:
> did this by eye so the size changes may not be 100% right.

they're spot on ;-)

HitcherUK:
'About' sounds better to me than 'About system' as 'System' is also an option there.

jjd-uk:
@HitcherUK that was the reason I didn't want to keep "System info" to avoid confusion with "System" settings, so tried to find a less confusing wording while keeping system in the name, since the window it goes to is still called "System info". This "About system" change is basically a quick & dirty hack to have something low impact for Krypton. However I've no issue with changing it to just "About" if others feel the same. 

If we do go for using "About" then something for consideration for v18 would be to do it properly, and completely rename the "System info" window to "About" in core. Can't do that now as it's obviously a skin breaking change

da-anda:
just an idea - on home screen we could show the avatar of the currently logged in user right next to the "power" button now that we have the space. Clicking the avatar toggles "switch profile" and it should only show up if more than one profile exists. What do you guys think?

marcelerkel:
If you want to get rid of the word system, then why not simply call it Info. There's summary information, storage information, network information, video information, hardware information, and PVR information behind that button.

If I see a button called About, then I expect to find the application's version number, some license info and copyright; that's what any other application shows when you open the About menu.

jjd-uk:
@da-anda As it happens I did consider whether we could have a swich profiles button along the power button and keep 4 buttons, but I wasn't sure if that would be a popular change so it was left as only a thought.

@marcelerkel plenty of devices such as iOS and Android devices use "About" where things like capacity, battery, and other statuses reside, so it's not an aplication version only thing, that's why I had the idea to use it. I'm not 100% sure on using "Info" but it's a possibility I hadn't considered.

mazeyx:
if you reduce it to 3 buttons, the watched list will scroll and be annoying, try and find something for that 4th button on home screen :) personally id like the about there instead :)

phil65:
Just lookin at code, I´m quite sure that this breaks other places in the skin since "BottomMainMenuItem" include is used for some more stuff.

jjd-uk:
@phil65 Sorry I missed it's also used in "Includes_MediaMenu.xml". I'll look at alternatives, since I based the dimension changes on the 3 button gouplist in "Custom_1100_AddonLauncher.xml" I could look to see if it would make sense for a new include for both 3 button layouts.

PR #10576, opened 2016-09-28
[gui] fix some fallback icons in Art selection dialog

phil65:
we currently use a wrong fallback icon for the "no art" entry in art selection dialog.
DefaultVideo.png is normally used for Video files and similar, DefaultPicture.png is more appropriate in this case.

@ronie @BigNoid 

Perhaps I should also add a defaulticon to the thumb://Current entry?

EDIT: added fallback icons for all entries which had none.

### DIFF
--- a/xbmc/video/dialogs/GUIDialogVideoInfo.cpp
+++ b/xbmc/video/dialogs/GUIDialogVideoInfo.cpp
@@ -730,7 +730,7 @@ void CGUIDialogVideoInfo::OnGetArt()
       // which is probably the IMDb thumb.  These could be wrong, so allow the user
       // to delete the incorrect thumb
       CFileItemPtr item(new CFileItem("thumb://None", false));
-      item->SetIconImage("DefaultVideo.png");
+      item->SetIconImage("DefaultPicture.png");
       item->SetLabel(g_localizeStrings.Get(13515));
       items.Add(item);
     }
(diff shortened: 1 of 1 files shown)


BigNoid:
Yeah makes sense to have default picture icon there. Not sure if it's worth the trouble for the thumb://current entry as that is hidden when there's no art and in cases where it's empty it's because the url image can't be loaded, but does has a path to the image. In which case the icon never kicks in.

phil65:
yeah but skinners can also use .ActualIcon and thus explicitely use the icon image.

BigNoid:
That's kind of an edge case for that dialog I think, but sure no objections if you want to add it.

ronie:
no objections from me either

phil65:
jenkins build this please

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -794,6 +797,7 @@ void CGUIDialogVideoInfo::OnGetFanart()
   {
     CFileItemPtr itemCurrent(new CFileItem("fanart://Current",false));
     itemCurrent->SetArt("thumb", m_movieItem->GetArt("fanart"));
+    itemNone->SetIconImage("DefaultPicture.png");
```
Rechi: maybe `itemCurrent` instead of `itemNone`

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -819,6 +823,7 @@ void CGUIDialogVideoInfo::OnGetFanart()
   {
     CFileItemPtr itemLocal(new CFileItem("fanart://Local",false));
     itemLocal->SetArt("thumb", strLocal);
+    itemNone->SetIconImage("DefaultPicture.png");
```
Rechi: maybe `itemLocal` instead of `itemNone`

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
phil65: oops sorry, c&p ftw :/

phil65:
jenkins build this please

PR #10577, opened 2016-09-28
[PVR] Fix: pvr client addons must be reloaded on profile switch to re-read addon config etc.

ksooo:
When switching user profiles, pvr client addons must be reloaded to re-read addon config etc.

This should fix a problem reported in the forum http://forum.kodi.tv/showthread.php?tid=282907&pid=2422637#pid2422637

@Jalle19 for review?

### DIFF
--- a/xbmc/pvr/addons/PVRClients.cpp
+++ b/xbmc/pvr/addons/PVRClients.cpp
@@ -1112,8 +1112,6 @@ bool CPVRClients::IsKnownClient(const AddonPtr client) const
 void CPVRClients::UpdateAddons(void)
 {
   VECADDONS addons;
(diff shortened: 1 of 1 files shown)


Jalle19 (approved):

ksooo:
jenkins build this please

da-anda:
I guess this should apply to other binary add-ons as well. Did anybody check if it's the case for these already?

ksooo:
> I guess this should apply to other binary add-ons as well. 

 yes

> Did anybody check if it's the case for these already?

 i don't think so.

Jalle19:
I was thinking the same when looking at the code. @garbear and @AlwinEsch might be interested.

PR #10578, opened 2016-09-28
[gui] fix header on sort dialog

BigNoid:
This fixes the empty header on the select dialog when selecting a sort method.
This was added in #9647 but I can't seem to find label 32104 anywhere in history. That number should also be reserved for scripts.

Before:
schermafbeelding 2016-09-28 om 10 47 23

After:
schermafbeelding 2016-09-28 om 10 48 01

@phil65 since you added the string id in #9647

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -19625,3 +19625,8 @@ msgstr ""
 msgctxt "#39009"
 msgid "Zoom - 110% width"
 msgstr ""
+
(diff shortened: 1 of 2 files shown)


MartijnKaijser:
@bignoid could you use a way higher number for the string?

BigNoid:
> could you use a way higher number for the string?

Sure, it's the highest now :)

MartijnKaijser (approved):

phil65 (approved):

PR #10583, opened 2016-09-28
[PVR] Guide window: fix ruler date might taking to much space (result ing in ugly overlap).

ksooo:
Before:

screenshot002

After:

screenshot000

@phil65 fyi
@Jalle19 mind taking a look at the code changes

### DIFF
--- a/xbmc/XBDateTime.cpp
+++ b/xbmc/XBDateTime.cpp
@@ -1327,12 +1327,12 @@ std::string CDateTime::GetAsLocalizedTime(const std::string &format, bool withSe
   return strOut;
 }
 
-std::string CDateTime::GetAsLocalizedDate(bool longDate/*=false*/) const
+std::string CDateTime::GetAsLocalizedDate(bool longDate/*=false*/, bool withYear /* = true */) const
 {
-  return GetAsLocalizedDate(g_langInfo.GetDateFormat(longDate));
+  return GetAsLocalizedDate(g_langInfo.GetDateFormat(longDate), withYear);
 }
 
-std::string CDateTime::GetAsLocalizedDate(const std::string &strFormat) const
+std::string CDateTime::GetAsLocalizedDate(const std::string &strFormat, bool withYear /* = true */) const
 {
   std::string strOut;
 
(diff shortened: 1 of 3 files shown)


Jalle19 (approved):
Looks good to me

ksooo:
Jenkins build this please

xbmc/XBDateTime.cpp
```diff
@@ -1327,12 +1327,12 @@ std::string CDateTime::GetAsLocalizedTime(const std::string &format, bool withSe
...
 
-std::string CDateTime::GetAsLocalizedDate(bool longDate/*=false*/) const
+std::string CDateTime::GetAsLocalizedDate(bool longDate/*=false*/, bool withYear /* = true */) const
 {
-  return GetAsLocalizedDate(g_langInfo.GetDateFormat(longDate));
+  return GetAsLocalizedDate(g_langInfo.GetDateFormat(longDate), withYear);
 }
 
-std::string CDateTime::GetAsLocalizedDate(const std::string &strFormat) const
+std::string CDateTime::GetAsLocalizedDate(const std::string &strFormat, bool withYear /* = true */) const
```
xhaggi: why do we need this here? you should be able to leave the year out of the format pattern.

xhaggi:
I'm not very familiar with our date time formatting class, but why not use a pattern for the date we need in the guide ruler for example like java does with its date time pattern `E, d M`.

xhaggi:
after a second look it should be possible right now `DDD,  D MMM` or `ddd, d mmm`. Both do the same because it's case-insensitve.

xhaggi:
sorry for the noise again but IMO this should be handled by the skin because other skins don't have an issue with limited space for that label e.g. Xperience1080.

-1 from my side

ksooo:
Regarding format string: you cannot hardcode one because it's depended on the locale. Thus, my approach of obtaining the correct locale depended format string and than to remove the year is correct and definitely better than hardcoding something locale independent like you suggested. It is fine the same way for GetLocalizedTime(..., bool withoutSeconds), btw

Regarding coding it in the skin: @phil65 told me that this is currently not possible.

xbmc/XBDateTime.cpp
ksooo: true, but i did it exactly like it was done for <code>GetAsLocalizedTime</code>. i wanted to change as less code as possible at this stage in the release cycle.

ksooo: See line line 1332. There it is called.

Paxxi: I have a real dislike for bool parameters, I'd rather see a new method like GetAsLocalizedDateWithoutYear as it makes it very clear what it does without looking at the declaration

ksooo: > GetAsLocalizedDateWithoutYear

to get rid of the two bools we would end with GetAsLongLocalizedDateWithoutYear for the guide window ruler case. fine by me.

ksooo:
On second thought, I'm with you @xhaggi. This should be solved in the skin. If this is currently not possible this should be tackled by extending guiinfo, not by hardcoding the (year-less) date format in the core.

ksooo:
@phil65 fyi, closing this PR.

xhaggi:
@ksooo You are right, if the predefined format pattern is locale depended. Adding more localized pattern seems reasonable for me as doing it like this.

Anyway, I don't want to miss the year in the guide window only because our default skin is not able to handle it right. At this stage we should fix it there. Modifying the core will effects all skins.

xhaggi:
wow .. simultaneously 😄

PR #10584, opened 2016-09-29
minor keymap change for tempo up/down

NedScott:
As requested here: https://github.com/xbmc/xbmc/commit/9719da3ea8046d363f3ebf650368ce7ef0838ed3#commitcomment-19212541

change control + arrow to alt + arrow, to avoid conflict on Mac OS X.


FernetMenta:
thanks.
does anybody object object to this change

Rechi:
would be nice if also works on Info and OSD screen.

PR #10586, opened 2016-09-29
[python] - allow detailedmode for xbmcgui.Dialog().select() / multiselect()

phil65:
Allows to use the detailedmode for our Dialog().select() and Dialog().multiselect() methods
1) by adding a bool parameter called "useDetails"
2) by being able to pass xbmcgui listitems instead of strings so that these items can carry a label2 and/or an icon.

@ronie @tamland

### DIFF
--- a/xbmc/interfaces/legacy/Dialog.cpp
+++ b/xbmc/interfaces/legacy/Dialog.cpp
@@ -101,7 +101,7 @@ namespace XBMCAddon
     }
 
 
-    int Dialog::select(const String& heading, const std::vector<String>& list, int autoclose, int preselect)
+    int Dialog::select(const String& heading, const std::vector<Alternative<String, const ListItem* > > & list, int autoclose, int preselect, bool useDetails)
     {
       DelayedCallGuard dcguard(languageHook);
       CGUIDialogSelect* pDialog= (CGUIDialogSelect*)g_windowManager.GetWindow(WINDOW_DIALOG_SELECT);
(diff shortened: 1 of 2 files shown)


ronie:
tested and works fine on my end. thx!

ganesh1515:
Thankyou

Sent from my iPhone

On Sep 29, 2016, at 11:36 PM, ronie <notifications@github.com<mailto:notifications@github.com>> wrote:


tested and works fine on my end. thx!

xbmc/interfaces/legacy/Dialog.cpp
```diff
@@ -101,7 +101,7 @@ namespace XBMCAddon
     }
 
 
-    int Dialog::select(const String& heading, const std::vector<String>& list, int autoclose, int preselect)
+    int Dialog::select(const String& heading, const std::vector<Alternative<String, const ListItem* > > & list, int autoclose, int preselect, bool useDetails)
```
tamland: Would it make sense to make it `Alternative<vector, vector>` instead? As it is now, you're allowed to pass a lists with both string and ListItems, contrary to what the docs say.

xbmc/interfaces/legacy/Dialog.cpp
```diff
@@ -111,26 +111,25 @@ namespace XBMCAddon
       pDialog->Reset();
       if (!heading.empty())
         pDialog->SetHeading(CVariant{heading});
-
-      String listLine;
-      for(unsigned int i = 0; i < list.size(); i++)
+      for (const auto item : list)
```
tamland: `auto&`

phil65: oki

phil65:
@tamland updated, including the docs.

xbmc/interfaces/legacy/Dialog.cpp
tamland: i guess..

tamland:
looks good

phil65:
Updated to include @python_v17 remarks.

MartijnKaijser:
Jenkins build and merge

PR #10592, opened 2016-09-29
[gui] change rule type selection from spinner to selectdialog	

phil65:
We sometimes have approx 30 different rule types. Doing that via spinners is quite a PITA, so lets use dialogselect instead.


phil65:
superseded by https://github.com/xbmc/xbmc/pull/10594

PR #10593, opened 2016-09-29
[gui] change group-by selection from spinner to selectdialog

phil65:
Next one, this time it´s the group-by button in smartplaylisteditor dialog.

PR #10594, opened 2016-09-29
[gui] smartplaylisteditor: convert spinners etc to regular buttons

phil65:
Next one...this time the order-by button in Smartplaylisteditor.
this PR adds the same header file to GUIDialogSmartPlaylistEditor.cpp as https://github.com/xbmc/xbmc/pull/10593, so remind me to update the PR which gets pulled last. :)
Could have put all commits into one big PR, so sorry for the additional pinging. 

EDIT: all PRs merged into this one.

EDIT: FInished!
screenshot031

### DIFF
--- a/xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
+++ b/xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
@@ -26,6 +26,7 @@
 #include "filesystem/File.h"
 #include "GUIDialogContextMenu.h"
 #include "GUIDialogSmartPlaylistRule.h"
+#include "GUIDialogSelect.h"
 #include "guilib/GUIKeyboardFactory.h"
 #include "guilib/GUIWindowManager.h"
 #include "guilib/LocalizeStrings.h"
@@ -294,9 +295,18 @@ void CGUIDialogSmartPlaylistEditor::OnType()
 
 void CGUIDialogSmartPlaylistEditor::OnOrder()
 {
-  CGUIMessage msg(GUI_MSG_ITEM_SELECTED, GetID(), CONTROL_ORDER_FIELD);
-  OnMessage(msg);
-  m_playlist.m_orderField = (SortBy)msg.GetParam1();
+  std::vector<SortBy> orders = CSmartPlaylistRule::GetOrders(m_playlist.GetType());
+  CGUIDialogSelect* dialog = (CGUIDialogSelect*)g_windowManager.GetWindow(WINDOW_DIALOG_SELECT);
+  dialog->Reset();
+  for (auto order: orders)
+    dialog->Add(g_localizeStrings.Get(SortUtils::GetSortLabel(order)));
+  dialog->SetHeading(CVariant{ 21429 });
+  dialog->SetSelected(g_localizeStrings.Get(SortUtils::GetSortLabel(m_playlist.m_orderField)));
+  dialog->Open();
+  int newSelected = dialog->GetSelectedItem();
+  if (!dialog->IsConfirmed() || newSelected < 0 || orders[newSelected] == m_playlist.m_orderField)
+    return;
+  m_playlist.m_orderField = orders[newSelected];
   UpdateButtons();
 }
 
@@ -374,15 +384,10 @@ void CGUIDialogSmartPlaylistEditor::UpdateButtons()
     CONTROL_DESELECT(CONTROL_ORDER_DIRECTION);
   }
 
-  // sort out the order fields
-  std::vector< std::pair<std::string, int> > labels;
-  std::vector<SortBy> orders = CSmartPlaylistRule::GetOrders(m_playlist.GetType());
-  for (unsigned int i = 0; i < orders.size(); i++)
-    labels.push_back(make_pair(g_localizeStrings.Get(SortUtils::GetSortLabel(orders[i])), orders[i]));
-  SET_CONTROL_LABELS(CONTROL_ORDER_FIELD, m_playlist.m_orderField, &labels);
+  SET_CONTROL_LABEL2(CONTROL_ORDER_FIELD, g_localizeStrings.Get(SortUtils::GetSortLabel(m_playlist.m_orderField)));
 
   // setup groups
-  labels.clear();
+  std::vector< std::pair<std::string, int> > labels;
   std::vector<Field> groups = CSmartPlaylistRule::GetGroups(m_playlist.GetType());
   Field currentGroup = CSmartPlaylistRule::TranslateGroup(m_playlist.GetGroup().c_str());
   for (unsigned int i = 0; i < groups.size(); i++)
--- a/xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
+++ b/xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
@@ -189,9 +189,7 @@ bool CGUIDialogSmartPlaylistEditor::OnMessage(CGUIMessage& message)
 
 void CGUIDialogSmartPlaylistEditor::OnPopupMenu(int item)
 {
-  if (item < 0 || item >= m_ruleLabels->Size())
-    return;
-  if (m_playlist.m_ruleCombination.m_rules.size() == 1 && m_playlist.m_ruleCombination.m_rules[0]->m_field == FieldNone)
+  if (item < 0 || static_cast<size_t>(item) >= m_playlist.m_ruleCombination.m_rules.size())
     return;
   // highlight the item
   m_ruleLabels->Get(item)->Select(true);
@@ -209,14 +207,17 @@ void CGUIDialogSmartPlaylistEditor::OnPopupMenu(int item)
 }
 
 void CGUIDialogSmartPlaylistEditor::OnRuleList(int item)
-{
-  if (item < 0 || item >= (int)m_playlist.m_ruleCombination.m_rules.size()) return;
-
-  CSmartPlaylistRule rule = *std::static_pointer_cast<CSmartPlaylistRule>(m_playlist.m_ruleCombination.m_rules[item]);
-
-  if (CGUIDialogSmartPlaylistRule::EditRule(rule,m_playlist.GetType()))
-    *m_playlist.m_ruleCombination.m_rules[item] = rule;
-
+{ 
+  if (item == static_cast<int>(m_playlist.m_ruleCombination.m_rules.size()))
+    OnRuleAdd();
+  else if (item < 0 || item > static_cast<int>(m_playlist.m_ruleCombination.m_rules.size()))
+    return;
+  else
+  {
+    CSmartPlaylistRule rule = *std::static_pointer_cast<CSmartPlaylistRule>(m_playlist.m_ruleCombination.m_rules[item]);
+    if (CGUIDialogSmartPlaylistRule::EditRule(rule,m_playlist.GetType()))
+      *m_playlist.m_ruleCombination.m_rules[item] = rule;
+  }
   UpdateButtons();
 }
 
@@ -343,11 +344,6 @@ void CGUIDialogSmartPlaylistEditor::UpdateButtons()
 {
   CONTROL_ENABLE(CONTROL_OK); // always enabled since we can have no rules -> match everything (as we do with default partymode playlists)
   
-  // if there's no rule available, add a dummy one the user can edit
-  if (m_playlist.m_ruleCombination.m_rules.size() <= 0)
-    m_playlist.m_ruleCombination.AddRule(CSmartPlaylistRule());
-
-  // name
   if (m_mode == "partyvideo" || m_mode == "partymusic")
   {
     SET_CONTROL_LABEL2(CONTROL_NAME, g_localizeStrings.Get(16035));
@@ -369,15 +365,15 @@ void CGUIDialogSmartPlaylistEditor::UpdateButtons()
   CGUIMessage msgReset(GUI_MSG_LABEL_RESET, GetID(), CONTROL_RULE_LIST);
   OnMessage(msgReset);
   m_ruleLabels->Clear();
-  for (unsigned int i = 0; i < m_playlist.m_ruleCombination.m_rules.size(); i++)
+  for (const auto& rule: m_playlist.m_ruleCombination.m_rules)
   {
     CFileItemPtr item(new CFileItem("", false));
-    if (m_playlist.m_ruleCombination.m_rules[i]->m_field == FieldNone)
-      item->SetLabel(g_localizeStrings.Get(21423));
-    else
-      item->SetLabel(std::static_pointer_cast<CSmartPlaylistRule>(m_playlist.m_ruleCombination.m_rules[i])->GetLocalizedRule());
+    item->SetLabel(std::static_pointer_cast<CSmartPlaylistRule>(rule)->GetLocalizedRule());
     m_ruleLabels->Add(item);
   }
+  CFileItemPtr item(new CFileItem("", false));
+  item->SetLabel(g_localizeStrings.Get(21423));
+  m_ruleLabels->Add(item);
   CGUIMessage msg(GUI_MSG_LABEL_BIND, GetID(), CONTROL_RULE_LIST, 0, 0, m_ruleLabels);
   OnMessage(msg);
   SendMessage(GUI_MSG_ITEM_SELECT, GetID(), CONTROL_RULE_LIST, currentItem);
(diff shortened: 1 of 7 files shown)


phil65:
Added another commit which converts two more buttons.
If it helps testing or whatever I could also merge the other PR´s into this one.

MilhouseVH:
Merge them and I'll include them in test builds - keeping them separate makes that a bit trickier (as they're both patching the same code).

phil65:
@MilhouseVH done.

MilhouseVH:
Thanks, will include in the next build.

phil65:
...aaand another commit.

xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
```diff
@@ -208,14 +207,17 @@ void CGUIDialogSmartPlaylistEditor::OnPopupMenu(int item)
...
-
-  CSmartPlaylistRule rule = *std::static_pointer_cast<CSmartPlaylistRule>(m_playlist.m_ruleCombination.m_rules[item]);
-
-  if (CGUIDialogSmartPlaylistRule::EditRule(rule,m_playlist.GetType()))
-    *m_playlist.m_ruleCombination.m_rules[item] = rule;
-
+{ 
+  if (item == static_cast<int>(m_playlist.m_ruleCombination.m_rules.size()))
+    OnRuleAdd();
+  else if (item < 0 || item > static_cast<int>(m_playlist.m_ruleCombination.m_rules.size()))
```
Montellese: Please move this condition to the top. IMO early returns should always be at the top if there are multiple conditions. Furthermore after an early return you don't need an `else`.

xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
```diff
@@ -208,14 +207,17 @@ void CGUIDialogSmartPlaylistEditor::OnPopupMenu(int item)
...
+    return;
+  else
+  {
+    CSmartPlaylistRule rule = *std::static_pointer_cast<CSmartPlaylistRule>(m_playlist.m_ruleCombination.m_rules[item]);
+    if (CGUIDialogSmartPlaylistRule::EditRule(rule,m_playlist.GetType()))
```
Montellese: Space after `rule,`

xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
```diff
@@ -270,33 +272,80 @@ void CGUIDialogSmartPlaylistEditor::OnCancel()
...
+    UpdateButtons();
+  }
+}
+
 void CGUIDialogSmartPlaylistEditor::OnLimit()
 {
-  CGUIMessage msg(GUI_MSG_ITEM_SELECTED, GetID(), CONTROL_LIMIT);
-  OnMessage(msg);
-  m_playlist.m_limit = msg.GetParam1();
+  const int limits[] = {0, 10, 25, 50, 100, 250, 500, 1000, -1 };
```
Montellese: You can use an `std::vector` to get rid of the awkward `-1` special item and then you can use a ranged `for` loop.

xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
```diff
@@ -270,33 +272,80 @@ void CGUIDialogSmartPlaylistEditor::OnCancel()
...
+    m_playlist.m_ruleCombination.SetType(CSmartPlaylistRuleCombination::CombinationAnd);
+  else
+    m_playlist.m_ruleCombination.SetType(CSmartPlaylistRuleCombination::CombinationOr);
   UpdateButtons();
 }
 
+void CGUIDialogSmartPlaylistEditor::OnName()
+{
+  std::string name = m_playlist.m_playlistName;
+  if (CGUIKeyboardFactory::ShowAndGetInput(name, CVariant{g_localizeStrings.Get(16012)}, false))
```
Montellese: You can use `CVariant{ 16012 }` directly. The `CGUIKeyboardFactory::ShowAndGetInput()` knows how to deal with that.

xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
```diff
@@ -546,6 +548,37 @@ void CGUIDialogSmartPlaylistEditor::HighlightItem(int item)
   OnMessage(msg);
 }
 
+std::vector<CGUIDialogSmartPlaylistEditor::PLAYLIST_TYPE> CGUIDialogSmartPlaylistEditor::GetAllowedTypes(std::string mode)
```
Montellese: `const std::string& mode`

xbmc/dialogs/GUIDialogSmartPlaylistRule.cpp
```diff
@@ -371,54 +371,15 @@ void CGUIDialogSmartPlaylistRule::OnBrowse()
...
-  UpdateButtons();
-}
-
 std::pair<std::string, int> OperatorLabel(CDatabaseQueryRule::SEARCH_OPERATOR op)
 {
   return std::make_pair(CSmartPlaylistRule::GetLocalizedOperator(op), op);
 }
 
-void CGUIDialogSmartPlaylistRule::UpdateButtons()
+std::vector<std::pair<std::string, int>> CGUIDialogSmartPlaylistRule::GetValidOperators(CSmartPlaylistRule rule)
```
Montellese: `const CSmartPlaylistRule& rule`

xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
```diff
@@ -270,33 +272,80 @@ void CGUIDialogSmartPlaylistEditor::OnCancel()
...
   UpdateButtons();
 }
 
 void CGUIDialogSmartPlaylistEditor::OnType()
 {
-  CGUIMessage msg(GUI_MSG_ITEM_SELECTED, GetID(), CONTROL_TYPE);
-  OnMessage(msg);
-  m_playlist.SetType(ConvertType((PLAYLIST_TYPE)msg.GetParam1()));
+  std::vector<PLAYLIST_TYPE> allowedTypes = GetAllowedTypes(m_mode);
+  CGUIDialogSelect* dialog = (CGUIDialogSelect*)g_windowManager.GetWindow(WINDOW_DIALOG_SELECT);
```
Montellese: Please use `static_cast`.

xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
```diff
@@ -270,33 +272,80 @@ void CGUIDialogSmartPlaylistEditor::OnCancel()
...
   UpdateButtons();
 }
 
 void CGUIDialogSmartPlaylistEditor::OnOrder()
 {
-  CGUIMessage msg(GUI_MSG_ITEM_SELECTED, GetID(), CONTROL_ORDER_FIELD);
-  OnMessage(msg);
-  m_playlist.m_orderField = (SortBy)msg.GetParam1();
+  std::vector<SortBy> orders = CSmartPlaylistRule::GetOrders(m_playlist.GetType());
+  CGUIDialogSelect* dialog = (CGUIDialogSelect*)g_windowManager.GetWindow(WINDOW_DIALOG_SELECT);
```
Montellese: `static_cast`

xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
```diff
@@ -311,11 +360,22 @@ void CGUIDialogSmartPlaylistEditor::OnOrderDirection()
 
 void CGUIDialogSmartPlaylistEditor::OnGroupBy()
 {
-  CGUIMessage msg(GUI_MSG_ITEM_SELECTED, GetID(), CONTROL_GROUP_BY);
-  OnMessage(msg);
-  m_playlist.SetGroup(CSmartPlaylistRule::TranslateGroup((Field)msg.GetParam1()));
+  std::vector<Field> groups = CSmartPlaylistRule::GetGroups(m_playlist.GetType());
+  Field currentGroup = CSmartPlaylistRule::TranslateGroup(m_playlist.GetGroup().c_str());
+  CGUIDialogSelect* dialog = (CGUIDialogSelect*)g_windowManager.GetWindow(WINDOW_DIALOG_SELECT);
```
Montellese: `static_cast`

xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
```diff
@@ -345,50 +400,50 @@ void CGUIDialogSmartPlaylistEditor::UpdateButtons()
     SET_CONTROL_LABEL2(CONTROL_NAME, m_playlist.m_playlistName);
   
   UpdateRuleControlButtons();
+
```
Montellese: Unneeded newline

xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
```diff
@@ -345,50 +400,50 @@ void CGUIDialogSmartPlaylistEditor::UpdateButtons()
     SET_CONTROL_LABEL2(CONTROL_NAME, m_playlist.m_playlistName);
   
   UpdateRuleControlButtons();
+  
 
+  if (m_playlist.m_ruleCombination.GetType() == CSmartPlaylistRuleCombination::CombinationOr)
+    SET_CONTROL_LABEL2(CONTROL_MATCH, g_localizeStrings.Get(21426));
```
Montellese: Could you add comments to these two `SET_CONTROL_LABEL2` lines to know what the new labels are?

xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
```diff
@@ -420,73 +475,20 @@ void CGUIDialogSmartPlaylistEditor::UpdateRuleControlButtons()
...
-  for (unsigned int i = 0; i < allowedTypes.size(); i++)
-    labels.push_back(make_pair(GetLocalizedType(allowedTypes[i]), allowedTypes[i]));
-  // check our playlist type is allowed
+  std::vector<PLAYLIST_TYPE> allowedTypes = GetAllowedTypes(m_mode);
+  // check if our playlist type is allowed
   PLAYLIST_TYPE type = ConvertType(m_playlist.GetType());
   bool allowed = false;
-  for (unsigned int i = 0; i < allowedTypes.size(); i++)
-    if (type == allowedTypes[i])
+  for (auto allowedType: allowedTypes)
```
Montellese: Please use curly braces (even though the original code was missing them as well).

xbmc/dialogs/GUIDialogSmartPlaylistRule.cpp
```diff
@@ -465,15 +426,83 @@ void CGUIDialogSmartPlaylistRule::UpdateButtons()
...
-  SET_CONTROL_LABELS(CONTROL_OPERATOR, m_rule.m_operator, &labels);
+void CGUIDialogSmartPlaylistRule::OnCancel()
+{
+  m_cancelled = true;
+  Close();
+}
+
+void CGUIDialogSmartPlaylistRule::OnField()
+{
+
```
Montellese: Unneeded newline

xbmc/dialogs/GUIDialogSmartPlaylistRule.cpp
```diff
@@ -465,15 +426,83 @@ void CGUIDialogSmartPlaylistRule::UpdateButtons()
...
+  std::vector<Field> fields = CSmartPlaylistRule::GetFields(m_type);
```
Montellese: You could use `const auto` here.

xbmc/dialogs/GUIDialogSmartPlaylistRule.cpp
```diff
@@ -465,15 +426,83 @@ void CGUIDialogSmartPlaylistRule::UpdateButtons()
...
+  CGUIDialogSelect* dialog = (CGUIDialogSelect*)g_windowManager.GetWindow(WINDOW_DIALOG_SELECT);
```
Montellese: `static_cast`

xbmc/dialogs/GUIDialogSmartPlaylistRule.cpp
```diff
@@ -465,15 +426,83 @@ void CGUIDialogSmartPlaylistRule::UpdateButtons()
...
+  dialog->Reset();
+  dialog->SetHeading(CVariant{20427});
+  int selected = -1;
+  for (unsigned int i = 0; i < fields.size(); i++)
```
Montellese: ranged `for` loop.

xbmc/dialogs/GUIDialogSmartPlaylistRule.cpp
```diff
@@ -465,15 +426,83 @@ void CGUIDialogSmartPlaylistRule::UpdateButtons()
...
+  // check if operator is still valid. if not, reset to first valid one
+  std::vector< std::pair<std::string, int> > validOperators = GetValidOperators(m_rule);
+  bool isValid = false;
+  for (auto op : validOperators)
+    if (std::get<0>(op) == std::get<0>(OperatorLabel(m_rule.m_operator)))
+      isValid = true;
+  if (!isValid)
+    m_rule.m_operator = (CDatabaseQueryRule::SEARCH_OPERATOR)std::get<1>(validOperators[0]);
+  m_rule.SetParameter("");
+  UpdateButtons();
```
Montellese: Please throw in some empty lines in this code block (e.g. before an `if` statement and after the early return). It's hard to read if it's so dense.

xbmc/dialogs/GUIDialogSmartPlaylistRule.cpp
```diff
@@ -465,15 +426,83 @@ void CGUIDialogSmartPlaylistRule::UpdateButtons()
...
+}
+
+void CGUIDialogSmartPlaylistRule::OnOperator()
+{
+
```
Montellese: Unneeded newline

xbmc/dialogs/GUIDialogSmartPlaylistRule.cpp
```diff
@@ -465,15 +426,83 @@ void CGUIDialogSmartPlaylistRule::UpdateButtons()
...
+  std::vector< std::pair<std::string, int> > labels = GetValidOperators(m_rule);
+  CGUIDialogSelect* dialog = (CGUIDialogSelect*)g_windowManager.GetWindow(WINDOW_DIALOG_SELECT);
```
Montellese: `static_cast`

xbmc/dialogs/GUIDialogSmartPlaylistRule.cpp
```diff
@@ -465,15 +426,83 @@ void CGUIDialogSmartPlaylistRule::UpdateButtons()
...
+  if (!isValid)
+    m_rule.m_operator = (CDatabaseQueryRule::SEARCH_OPERATOR)std::get<1>(validOperators[0]);
+  m_rule.SetParameter("");
+  UpdateButtons();
+}
+
+void CGUIDialogSmartPlaylistRule::OnOperator()
+{
+
+  std::vector< std::pair<std::string, int> > labels = GetValidOperators(m_rule);
```
Montellese: `const auto`

Montellese (changes requested):
Looks good apart from the minors. This will require all skins to adjust the dialog right?

xbmc/dialogs/GUIDialogSmartPlaylistRule.cpp
phil65: i need `i`... and I would need sth like `for count, element in enumerate(fields)` How do I do that in Cpp? :)

phil65:
@Montellese 
addressed all of your comments in a separate commit, except for the loops, see question above.

Yes, this needs some trivial changes for all skins.

xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
```diff
@@ -546,6 +549,37 @@ void CGUIDialogSmartPlaylistEditor::HighlightItem(int item)
   OnMessage(msg);
 }
 
+std::vector<CGUIDialogSmartPlaylistEditor::PLAYLIST_TYPE> CGUIDialogSmartPlaylistEditor::GetAllowedTypes(const std::string& mode)
+{
```
phil65: Should I move this method somewhere else? The method is static and perhaps better belongs to CSmartPlaylist?

xbmc/dialogs/GUIDialogSmartPlaylistRule.cpp
Montellese: Right it doesn't work with ranged `for` loop. But if you use iterators there's a method called `std::distance()` and you use it like `size_t pos = std::distance(fields.begin(), field);` where `field` is the iterator.

phil65: does that really make the code easier in this case? :)

xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
Paxxi: Ranged for loop works for plain array as well, the requirements are that std:begin and end exists and those are implemented by default

xbmc/dialogs/GUIDialogSmartPlaylistRule.cpp
Montellese: It makes it more generic in case something around it changes.

xbmc/dialogs/GUIDialogSmartPlaylistEditor.cpp
Montellese: @Paxxi: correct but the `-1` would have to go in that case as well which is my main point/concern.

phil65:
@Montellese switched to using iterators. Anything else?
Thx for reviewing.

phil65:
jenkins build this please

phil65:
@Montellese 	general skinner´s sentiment was to merge this, even though we´re already in a later beta, since needed skin changes are very trivial and our skinners are quite good in keeping up with these changes.	Are you fine with this?

Montellese:
Fine with me but don't forget to squash the cleanup commit(s).

phil65:
All follow-up stuff squashed into one commit.
jenkins build this please.

jingai:
Why was the Order button (id=19 in SmartPlaylistEditor.xml) changed from togglebutton to button?  It only toggles between two values.

PR #10596, opened 2016-10-01
[osx] - Fix build on case-sensitive filesystems.

jingai:
Fix build on OSX when underlying filesystem is case-sensitive.  XBMCHelper.cpp includes "url.h" when it should be "URL.h".


fritsch:
jenkins build this please

@FernetMenta @Memphiz this is a no brainer, right?

FernetMenta:
yes

Memphiz:
Thx @jingai :)

PR #10597, opened 2016-10-01
Amlogic: fixes to hardware decoding blacklist

kszaq:
These two commits extend a list of formats when Amlogic hardware decoder should not be used.
## Description

Adds 2 more cases to fall back to software decoding:
- High 4:2:2 and 4:4:4 profiles for all Amlogic devices, as these are not supported by H/W decoder
- HEVC 10bit decoding for SoCs that don't support it (S8xx)
## Motivation and Context

Fix black screen/artifacts when hardware decoder is used for unsupported formats.
## How Has This Been Tested?

Compiled and tested for LibreELEC, Kodi 17.0b2, S805 and S905 devices.
## Types of change
## Checklist:

### DIFF
--- a/xbmc/utils/AMLUtils.cpp
+++ b/xbmc/utils/AMLUtils.cpp
@@ -191,6 +191,23 @@ bool aml_support_hevc_4k2k()
   return (has_hevc_4k2k == 1);
 }
 
+bool aml_support_hevc_10bit()
+{
+  static int has_hevc_10bit = -1;
+
+  if (has_hevc_10bit == -1)
+  {
+    CRegExp regexp;
+    regexp.RegComp("hevc:.*10bit");
+    std::string valstr;
+    if (SysfsUtils::GetString("/sys/class/amstream/vcodec_profile", valstr) != 0)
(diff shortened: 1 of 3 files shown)


xbmc/utils/AMLUtils.cpp
```diff
@@ -191,6 +191,23 @@ bool aml_support_hevc_4k2k()
   return (has_hevc_4k2k == 1);
 }
 
+bool aml_support_hevc_10bit()
+{
+  static int has_hevc_10bit = -1;
+
+  if (has_hevc_10bit == -1)
```
stefansaraev: you dont need this check

fritsch: it's a static. He wants to avoid doing the regex more than once.

kszaq: Yes, this is almost 1:1 copy from `aml_support_hevc_4k2k()` above.

Stane1983: Looks fine to me

codesnake:
Looks good to me.

PR #7843, opened 2015-08-18
CApplication: improve skin reloading in case theme/color/font is not default

Montellese:
This is an attempt at improving the logic in `CApplication::OnSettingChanged()` that is responsible for handling all the different skin related settings. The main issue is that changes to 4 different settings should result in reloading the skin:
- `lookandfeel.skin`
- `lookandfeel.font`
- `lookandfeel.skintheme`
- `lookandfeel.skincolors`

Where it gets complicated is with the dependencies:
- changing `lookandfeel.skin` might change `lookandfeel.font`, `lookandfeel.skintheme` and `lookandfeel.skincolors`
- changing `lookandfeel.skintheme` might change `lookandfeel.skincolors`

The fact that chaging a setting's value results in a call to the `OnSettingChanged()` implementation (even if it's called from an existing `OnSettingChanged()` implementation) makes it very difficult to keep track of the chain of changes invoked by changing a single setting due to dependencies. Ideally this would be handled by the settings system but that will be a major change.

In the case of skin settings this leads to reloading the skin multiple times unnecessarily. Therefore I'm falling back on a member variable in `CApplication` where I store the ID of a setting which should not be handled in the next `OnSettingChanged()` because I know which setting I'm changing and that it's callback handler will (most likely) be called.

So as an example when changing `lookandfeel.skin` what happens is
1. `CApplication::OnSettingChanged()` is called for `lookandfeel.skin`
2. `lookandfeel.skincolors` is reset to its default value
3. `CApplication::OnSettingChanged()` is called for `lookandfeel.skincolors` but is ignored
4. `lookandfeel.skintheme` is reset to its default value
5. `CApplication::OnSettingChanged()` is called for `lookandfeel.skintheme` but is ignored
6. `lookandfeel.font` is reset to its default value but nothing happens because it already is set to its default value
7. The skin is reloaded asynchronously

Furthermore as an improvement I've changed the code to make use of the setting's `Reset()` implementation to reset the setting's value to its default value instead of manually setting it's value to the default value. That way in case the default value changes in `settings.xml` it will automatically be handled correctly.

It also merges the handling of `lookandfeel.skintheme` with the handling of the other 3 skin related settings (especially their skin reloading logic).

PR #10602, opened 2016-10-01
fix font setting reset on skin change 

tamland:
Typo from #7843. No idea how this worked..

Also simplified it a bit by making m_skinReloadSettingIgnore a bool. As everything here should be synchronous there is no need to track which setting is being changed.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -262,6 +262,7 @@ CApplication::CApplication(void)
   , m_fallbackLanguageLoaded(false)
   , m_WaitingExternalCalls(0)
   , m_ProcessedExternalCalls(0)
+  , m_ignoreSkinSettingChanges(false)
 {
   m_network = NULL;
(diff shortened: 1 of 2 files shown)


MartijnKaijser:
@Montellese ping
Jenkins build this please

scott967:
Is this why when I have Estuary font set to non-default (Arial based) and switch skins, the new skin doesn't load, instead Estuary reloads with font set to default.  Then I can switch skins.  I though maybe it was some bug in my skin?

Montellese:
Thanks for the fix. It will probably work with the boolean. IIRC my initial approach was a vector of all setting IDs to be ignored but then realized it's not necessary and I only used the affected setting ID.

Have you tried the cases where changing a setting (e.g. the skin itself) will change the value of several other settings that would all cause a skin reload?

tamland:
^yeah. no issues with that.

Montellese (approved):

Paxxi (approved):

PR #10603, opened 2016-10-01
[xbmc][fix] Fix issue where yes/no dialog would show blank buttons

Paxxi:
Default values given to ShowYesNoDialog was wrong and caused the null check to think they contained valid values.
## Description

Add a default constructor to CVariant to allow instantiating a null CVariant
## Motivation and Context

Solves a bug seen when downloading Metropolis skin(likely any skin) and being asked to install the Skin Helper Addon
## How Has This Been Tested?

Run time tested
## Types of change
## Checklist:
  Setting default values to empty strings is not very smart
  when checking for a null variant. It's not null and we
  end up displaying empty strings instead of yes/no


Paxxi:
jenkins build this please

ronie:
thx. came across this issue a few times as well.

PR #10606, opened 2016-10-01
[guiinfo] System.HasAddon now also checks installed addons by type

xhaggi:
This will extend the info bool `System.HasAddon` to support checking for at least one installed addon by type. Currently it's only possible to check for a specific addon.

**How it works**
If the given parameter could be translated to one of our internal addon types it will lookup for that type, if not it will lookup for a specific addon.

**Why**
It's quite useful if you want to redirect a user to our addon browser -> repos if he needs at least one addon of the given type e.g for the weather widget.

### DIFF
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -7633,8 +7633,32 @@ bool CGUIInfoManager::GetMultiInfoBool(const GUIInfo &info, int contextWindow, c
         break;
       case SYSTEM_HAS_ADDON:
       {
-        AddonPtr addon;
-        bReturn = CAddonMgr::GetInstance().GetAddon(m_stringParameters[info.GetData1()],addon) && addon;
+        std::string param = m_stringParameters[info.GetData1()];
+        // check if parameter is a specific addon or addon type
+        ADDON::TYPE type = ADDON::TranslateType(param);
+        switch (type)
+        {
+          // type is unknown, lookup if the specific addon is installed
+          case ADDON::ADDON_UNKNOWN:
+            AddonPtr addon;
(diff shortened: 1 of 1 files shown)


xhaggi:
@ronie @phil65 @BigNoid forgot to ping you and our @mention-bot don't know that you are right guys 😄

ronie:
no objections from my end

phil65:
yup, nice. thx.

xhaggi:
jenkins build this please

xhaggi:
@MartijnKaijser something for Krypton. It will only extend the existing info bool without breaking the current functionality.

ksooo:
ping @tamland as I'm not sure whether it is okay to expose our internal add-on types to skins.

ksooo:
ping @FernetMenta as I'm not sure whether this causes performance issues. I takeover you requested to use binary add-on cache for the guiinfo pvr client add-on installation check I did recently.

tamland:
It will. It's a no-go performance wise. HasAddon is already a big resource hog as it involves reparsing/-constructing the addon, every update. Now multiply that by the number of addons installed. In addition, all multi-addon getter methods in AddonMgr involves database calls.

xhaggi:
@tamland okay, then i'll add new info bool `System.HasAddons(type)`

xhaggi:
> Now multiply that by the number of addons installed. In addition, all multi-addon getter methods in AddonMgr involves database calls.

If you look at the implementation I never call the lookup by type if the given parameter is an unknow type. Only if the type is valid we'll do the lookups.

xhaggi:
@tamland after a deeper look at the addon manager code I see what you mean. We definitely need a cache to solve this issue, but that is out of the scope of this PR. I'll close it for now.

PR #10609, opened 2016-10-02
Added option for disabling grouping movies in sets when browsing a tag

ThreePinkApples:
## Description

Added a option under Mediasettings -> Video to choose to group movies when browsing a tag. 
This options behaves completely separately from the existing general option of grouping movies in sets. 

I'm quite sure some of the things I've done in the code is quite bad, and I will add my own comments with questions. 
## Motivation and Context

The innability to disabled grouping of movies when viewing a tag has annoyed me from the first time I created a tag. I just don't feel it always makes sense. My best example is the MCU. I have a MCU tag, and it doesn't make sense to then group the movies, I would rather be able to sort them by year which would be the intented viewing order. 
Optimally, it should be possible to manually sort movies in each tag, separately, but that's probably much more work than this
## How Has This Been Tested?

I've manually tested this by running a build on my PC where I've have both set grouping settings on, both off, one on and the other off, and then checked that it behaved as it should. 
## Types of change

## Checklist:


addons/resource.language.en_gb/resources/strings.po
```diff
@@ -19630,3 +19630,14 @@ msgstr ""
 msgctxt "#39010"
 msgid "Select sort method"
 msgstr ""
+
+#: system/settings/settings.xml
+msgctxt "#40000"
+msgid "Show movie sets in tags"
+msgstr ""
+
```
aerandir92: I guess all of this is very wrong?
I don't know how you handle new strings, so I just put them here and gave them a high ID number for testing

xbmc/video/windows/GUIWindowVideoBase.cpp
```diff
@@ -1377,7 +1377,8 @@ void CGUIWindowVideoBase::GetGroupedItems(CFileItemList &items)
     VIDEODATABASEDIRECTORY::NODE_TYPE nodeType = CVideoDatabaseDirectory::GetDirectoryChildType(m_strFilterPath);
     if (items.GetContent() == "movies" && params.GetSetId() <= 0 &&
         nodeType == NODE_TYPE_TITLE_MOVIES &&
-       (CSettings::GetInstance().GetBool(CSettings::SETTING_VIDEOLIBRARY_GROUPMOVIESETS) || (StringUtils::EqualsNoCase(group, "sets") && mixed)))
+       (CSettings::GetInstance().GetBool(CSettings::SETTING_VIDEOLIBRARY_GROUPMOVIESETS) && !StringUtils::StartsWithNoCase(m_strFilterPath, "videodb://movies/tags/") || (StringUtils::EqualsNoCase(group, "sets") && mixed)) ||
+	   (CSettings::GetInstance().GetBool(CSettings::SETTING_VIDEOLIBRARY_GROUPMOVIESETSINTAGS) && StringUtils::StartsWithNoCase(m_strFilterPath, "videodb://movies/tags/") || (StringUtils::EqualsNoCase(group, "tags") && mixed)))
```
aerandir92: It feels wrong to check the filterpath with a typedout string, is there a constant somwhere I can use?

Also, the last part about group && mixed, I don't know if that is needed, I just added it to replecate the line abbove.

phil65:
Sorry, but -1 for custom options for specific nodes. Either we change default behaviour or not.

aerandir92:
@phil65 can this be done in an addon in a good way then? This is kinda part 1 of my quest to make tags actually usable. Next step is to be able to list tags together with movies (so that they work mostly as custom collections). Having tags hidden away in small submenu is just bad.

phil65:
I would start by adding a "add tag..." context menu item to movies / tvshows..That would improve usability quite a lot. I am not sure if a mixed movie-tag listing is what we want.

aerandir92:
That's a great start, actually, didn't think about that. It just annoyes me that I have to navigate back if I say I decide "I want to watch a Disney movie", instead of scrolling to D to find the Disney tag, I have to go back to home and to tags, then Disney

MartijnKaijser:
Can we close this one?

PR #10611, opened 2016-10-02
drop cookies.dat, not required, not thread safe

FernetMenta:
fix http://trac.kodi.tv/ticket/16935

A systemwide storage for cookies makes no sense and is not thread safe. We may have multiple CURL sessions open at once.

### DIFF
--- a/xbmc/filesystem/CurlFile.cpp
+++ b/xbmc/filesystem/CurlFile.cpp
@@ -506,13 +506,8 @@ void CCurlFile::SetCommonOptions(CReadState* state)
   g_curlInterface.easy_setopt(h, CURLOPT_FOLLOWLOCATION, TRUE);
   g_curlInterface.easy_setopt(h, CURLOPT_MAXREDIRS, 5);
 
-  // Enable cookie engine for current handle to re-use them in future requests
-  std::string strCookieFile;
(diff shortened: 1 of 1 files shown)


kib:
I don't see the benefit in this either. The cookies.dat shared storage doesn't select the correct cookie anyway. Dropping it makes more sense - the cookie should only exist in the session content anyway.

kib (approved):

arnova:
This may also fix http://trac.kodi.tv/ticket/16849

arnova:
Btw. expect breakage for some stuff, this was once implemented for a reason (not by me) I guess.

FernetMenta:
Since having a global cookie store is wrong by design, I don't expect any well designed behaviour to break.

> this was once implemented for a reason

maybe, maybe not. the code was very old. maybe the author was not aware that cookie engine could be enabled without having a file, or this is some later feature of libcurl.

fritsch:
+1

Long Island Icetea 15 euro

Am 05.10.2016 12:37 nachm. schrieb "Rainer Hochecker" <
notifications@github.com>:

arnova:
@FernetMenta : Please note that I'm fine with this change nevertheless since this obviously can never work for multithread.

PR #10614, opened 2016-10-03
[cmake] Fix always outdated targets

fetzerch:
CMake detects some targets (pack-skin, addon libraries) as always outdated when autotools leftovers are present in the source directory.

The issue is caused by xbt and so files that are present in the kodi source tree from an previous autotools build. The export-files target would override correctly generated libraries or skin packs by the old ones in the source tree. I've blacklisted the problematic file endings, because we never want to mirror them, as they are built by the project.

ping: let's see if mention-bot figures it out ;) --> also reported by @Memphiz for osx


fetzerch:
jenkins build this please

PR #10615, opened 2016-10-04
VTB on OSX: fix flickering video caused by buffer overwrite

FernetMenta:
see title. This applies the same concept we already have for vdpau and vaapi.

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoRenderers/BaseRenderer.h
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/BaseRenderer.h
@@ -90,7 +90,7 @@ public:
   virtual void Flush() {};
   virtual void SetBufferSize(int numBuffers) { }
   virtual void ReleaseBuffer(int idx) { }
-  virtual bool NeedBufferForRef(int idx) { return false; }
+  virtual bool NeedBuffer(int idx) { return false; }
   virtual bool IsGuiLayer() { return true; }
   // Render info, can be called before configure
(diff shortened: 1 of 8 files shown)


FernetMenta:
@Memphiz I think iOS has the same issue. There you could use the standard functions glFenceSync that are available for GLES.

FernetMenta:
jenkins build this please

MartijnKaijser:
jenkins build this please  (github died for awhile)

MartijnKaijser:
ignore the win32 fail

Memphiz:
Nice - great work - thx - so thats the fix for the stuttering issue?

FernetMenta:
@Memphiz "stuttering" is not the right term if you mean this case: http://forum.kodi.tv/showthread.php?tid=281517

"choppy" is better suited here. But since there a always multiple definitions out there, I requested a video for better diagnosis :)

jenkins build this please

MrMC:
might want to check the ref count on m_vtbbuf, vda/vtb will not re-use it if greater than one.

popcornmix:
glFenceSync is GLES3 only, not GLES2:
https://www.khronos.org/opengles/sdk/docs/man3/html/glFenceSync.xhtml

FernetMenta:
> might want to check the ref count on m_vtbbuf, vda/vtb will not re-use it if greater than one.

that is not the problem here

MrMC:
ok mr.expert :)

PR #10616, opened 2016-10-04
[python] add xbmcgui.Dialog().videoinfo() and .musicinfo()

phil65:
Allows to open musicinfo and videoinfo dialog via python, which will allow several script add-ons to drop their custom infowindow implementations.
@tamland @xhaggi @ronie

### DIFF
--- a/xbmc/interfaces/legacy/Dialog.cpp
+++ b/xbmc/interfaces/legacy/Dialog.cpp
@@ -27,6 +27,8 @@
 #include "guilib/GUIWindowManager.h"
 #include "dialogs/GUIDialogFileBrowser.h"
 #include "dialogs/GUIDialogNumeric.h"
+#include "video/dialogs/GUIDialogVideoInfo.h"
+#include "music/dialogs/GUIDialogMusicInfo.h"
 #include "settings/MediaSourceSettings.h"
 #include "dialogs/GUIDialogKaiToast.h"
 #include "ModuleXbmcgui.h"
@@ -36,6 +38,7 @@
 #include "WindowException.h"
 #include "messaging/ApplicationMessenger.h"
 #include "Dialog.h"
+#include "ListItem.h"
 #ifdef TARGET_POSIX
 #include "linux/XTimeUtils.h"
 #endif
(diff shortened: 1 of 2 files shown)


xhaggi:
looks good +1

xbmc/interfaces/legacy/Dialog.cpp
```diff
@@ -85,6 +88,27 @@ namespace XBMCAddon
       return pDialog->IsConfirmed();
     }
 
+    bool Dialog::videoinfo(const ListItem* item)
+    {
+      const AddonClass::Ref<xbmcgui::ListItem> listitem(item);
+      auto fileItem = new CFileItem(*listitem->item);
```
tamland: Leaks. thou shalt not use new in c++ you know;) Something like this should suffice:

```
bool Dialog::videoinfo(const ListItem& item)
{
  if (!item.item->HasVideoInfoTag())
    return false;
  CGUIDialogVideoInfo::ShowFor(*item.item);
  return true;
}
```

tamland:
Wouldn't it make more sense to combine these two to one 'info dialog' api? Or, why would you ever need to have api for both for the same file item?

xbmc/interfaces/legacy/Dialog.cpp
phil65: Thx! My C++ skills are still WIP, this was just taken from another part of our API code.

phil65:
so you mean open a or b based on the mediatype of the listitem, right?
Would also be fine for me. @xhaggi , what do you think?

xhaggi:
sure makes sense

phil65:
@ping tamland, should I choose based on mediatype, or based on HasVideoInfoTag() / HasMusicInfoTag() ?

ronie:
can you pick something that's easily extendable?
i intent to add support for the pictureinfo / epginfo dialogs as well.

tamland:
I'd say HasVideoInfoTag (etc) as that's what context menus use. This is an implementation details though and can be changed later.

phil65:
passing by reference somehow didnt work, so I decided to pass by pointer instead, and got rid of new keyword. Is it fine this way?

phil65:
let´s give it a try: jenkins build this please.

phil65:
Seems to build fine.
ping @tamland @xhaggi for another review. :)

tamland:
also all good:)

MartijnKaijser:
Jenkins build this please

xbmc/interfaces/legacy/Dialog.h
```diff
@@ -108,7 +108,37 @@ namespace XBMCAddon
...
+      ///
+      info(...);
+#else
+      bool info(const ListItem* item);
+#endif
+
+#ifdef DOXYGEN_SHOULD_USE_THIS
+      ///
+      /// \ingroup python_Dialog
+      /// \python_func{ xbmcgui.Dialog().select(heading, list[, autoclose,preselect]) }
```
ronie: ^ houston, we've lost useDetails

MartijnKaijser: fix https://github.com/xbmc/xbmc/pull/10833

phil65: oops, sorry, rebase fuckery. Thx for fixing.

PR #10617, opened 2016-10-04
 [eventlog] implement high resolution datetime sort method (CDateTime granularity of 1 sec is not sufficient).

ksooo:
[eventlog] implement special datetime sort method (CDateTime granularity of 1 sec is not sufficient).

Log entries get sorted by date using their internal CDateTime value, which is set to log entry instance creation date and time. So far so good, but there can be created more than one log entry per second. If this happens, the log entries time stamps are no longer unique. Thus they may get sorted logically wrong. Example: In the chronologically sorted list of log entries "PVR connecting to backend" appears after "PVR backend connection established".

The solution is pretty simply. On creation of a log entry, additionally to the "now" CDateTime, a high resolution timestamp (std::chrono functionality) will be assigned to the new instance and when comparing log entries by date, this high res timestamp will be used instead of the CDateTime value.

@Montellese what do you think about this approach (and the code change)?

### DIFF
--- a/xbmc/FileItem.h
+++ b/xbmc/FileItem.h
@@ -73,6 +73,9 @@ class CFileItemList;
 class CCueDocument;
 typedef std::shared_ptr<CCueDocument> CCueDocumentPtr;
 
+class IEvent;
+typedef std::shared_ptr<const IEvent> EventPtr;
+
 /* special startoffset used to indicate that we wish to resume */
 #define STARTOFFSET_RESUME (-1)
 
@@ -120,6 +123,8 @@ public:
   CFileItem(const PVR::CPVRTimerInfoTagPtr& timer);
   CFileItem(const CMediaSource& share);
   CFileItem(std::shared_ptr<const ADDON::IAddon> addonInfo);
+  CFileItem(const EventPtr& eventLogEntry);
+
   virtual ~CFileItem(void);
   virtual CGUIListItem *Clone() const { return new CFileItem(*this); };
 
@@ -542,6 +547,7 @@ private:
   PVR::CPVRRadioRDSInfoTagPtr m_pvrRadioRDSInfoTag;
   CPictureInfoTag* m_pictureInfoTag;
   std::shared_ptr<const ADDON::IAddon> m_addonInfo;
+  EventPtr m_eventLogEntry;
   bool m_bIsAlbum;
 
   CCueDocumentPtr m_cueDocument;
--- a/xbmc/events/BaseEvent.cpp
+++ b/xbmc/events/BaseEvent.cpp
@@ -18,8 +18,11 @@
  *
  */
 
+#include <chrono>
+
 #include "BaseEvent.h"
 #include "guilib/LocalizeStrings.h"
+#include "utils/StringUtils.h"
 
 CBaseEvent::CBaseEvent(const std::string& identifier, const CVariant& label, const CVariant& description, EventLevel level /* = EventLevel::Information */)
   : m_level(level),
@@ -29,6 +32,7 @@ CBaseEvent::CBaseEvent(const std::string& identifier, const CVariant& label, con
     m_description(description),
     m_details(),
     m_executionLabel(),
+    m_timestamp(GetInternalTimestamp()),
     m_dateTime(CDateTime::GetCurrentDateTime())
 { }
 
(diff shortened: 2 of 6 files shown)


xbmc/events/BaseEvent.cpp
```diff
@@ -20,6 +20,9 @@
 
 #include "BaseEvent.h"
 #include "guilib/LocalizeStrings.h"
+#include "utils/StringUtils.h"
+
+unsigned int CBaseEvent::s_nextIndex = 0;
```
Paxxi: Is this something you forgot to remove? Can't see it used and can't see a reason for it either, increasing the granularity of the timestamp should be enough

ksooo: Not a leftover. Look for the constructor changes below.

Sure, increasing timestamp granularity would be sufficient. But do we have a high resolution timer in Kodi? I'm not aware of any. That's why I took this shortish.

Paxxi: We have the millisecond one in CThread or you can use `std::chrono::steady_clock` which does nanosecond granularity on windows and Linux at least. I don't really like adding a static and state just to solve sorting. 

In missed that you use it in the format call.

xbmc/events/BaseEvent.cpp
```diff
@@ -97,3 +104,13 @@ std::string CBaseEvent::VariantToLocalizedString(const CVariant& variant)
...
   return "";
 }
+
+void CBaseEvent::ToSortable(SortItem& sortable, Field field) const
+{
+  if (field == FieldDate)
+  {
+    const CDateTime dt(GetDateTime());
+    const std::string strDt(dt.IsValid() ? dt.GetAsDBDateTime() : "");
+    sortable[FieldDate] = StringUtils::Format("%s %08u", strDt.c_str(), m_index);
```
Montellese: Why `8` digits`? 2^32 - 1 = 4294967295 and that's 10 digits. Obviously we're not gonna have that many events in one session but we should do things properly anyway.

xbmc/FileItem.h
```diff
@@ -542,6 +547,7 @@ class CFileItem :
   PVR::CPVRRadioRDSInfoTagPtr m_pvrRadioRDSInfoTag;
   CPictureInfoTag* m_pictureInfoTag;
   std::shared_ptr<const ADDON::IAddon> m_addonInfo;
+  EventPtr m_eventLogEntry;
```
Montellese: I'm not a huge fan of this but since `CFileItem` is hopelessly overloaded anyway I'm not gonna object.

Montellese (changes requested):
Looks ok apart from the minor.

xbmc/events/BaseEvent.cpp
ksooo: Cool, not loving my own solution very much, I will give std::chrono::steady_clock a try. Hopefully it is available and working on all Kodi platforms. Thanks for the input, @Paxxi

ksooo:
@Paxxi what do you think of my new std:.chrono approach?

xbmc/events/BaseEvent.cpp
ksooo: fixed with the new std::chrono based approach

ksooo:
lets see whether this build on all platforms...

jenkins build this please

ksooo:
jenkins build this please

xbmc/events/BaseEvent.cpp
```diff
@@ -29,6 +32,7 @@ CBaseEvent::CBaseEvent(const std::string& identifier, const CVariant& label, con
     m_description(description),
     m_details(),
     m_executionLabel(),
+    m_timestamp(std::chrono::time_point_cast<std::chrono::nanoseconds>(std::chrono::steady_clock::now()).time_since_epoch().count()),
```
ksooo: First, it seems that you looked  not at the most recent commit? Meanwhile I moved the chrono stuff into a static member function, ctors call this function no, so no more code duplication here.

Second, I wanted to avoid chrono data types in the header file. That's why I decided to have a primitive type as member.

Okay?

Paxxi: ahh yeah that's perfectly fine, no objections.

PR #10618, opened 2016-10-04
Album is compilation and last played rules and fix missing sort orders in smart playlists

DaveTBlake:
Fix missing rules and sort orders.
The compilation flag had been omitted as a rule for albums playlists, adding it means that users can now create playlists and filter type custom nodes for compilations (or exclude them).

Similarly with Last Played date, including it means that users can create their own variations of the recently played albums node without core code changes.

While doing that found that some of the sort orders were missing too, causing the display to ignore the order defined in the playlist. 

A good test would be to create a playlist of recently played compliations.

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -900,7 +900,10 @@ msgctxt "#203"
 msgid "Plot outline"
 msgstr ""
 
-#empty string with id 204
+#: xbmc/playlists/SmartPlaylist.cpp
+msgctxt "#204"
+msgid "Compilation"
+msgstr ""
 
 #: xbmc/playlists/SmartPlaylist.cpp
 msgctxt "#205"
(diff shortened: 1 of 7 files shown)


DaveTBlake:
Recently played compilation albums
`<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
<smartplaylist type="albums">
    <name>Last Played Compilations</name>
    <match>all</match>
    <rule field="compilation" operator="true" />
    <limit>25</limit>
    <order direction="descending">lastplayed</order>
</smartplaylist>`

@zag2me would you like a build to test?

DaveTBlake:
jenkins build this please

DaveTBlake:
Test build available here http://mirrors.kodi.tv/test-builds/win32/KodiSetup-20161008-ef6f476-MusicTest02.exe (It also includes fix to songs node for MariaDB PR10657)

jjd-uk:
The example playlist you've given doesn't seen to work, I assuming that the most recently played album should appear at the top? if so then that isn't happening.

DaveTBlake:
Thanks for looking at this @jjd-uk, it is always good to get someone else to try out changes. 

That example playlist will find the most recent 25 compilation albums (assuming that you have that many) and show the most recently played at the top. But the order you see them displayed can also depend on the sort you apply to that view - you can change it from the side blade. And if you have not played the compilation albums, then the date is null and they revert to sorting by title (and no date is shown). I guess that is what you are seeing, but do let me know either way (with screen shot if not).

jjd-uk:
Now I know how Last Played is supposed to behave all now appears to be fine to me, so a +1 from me to get this pulled in. I must say the default "Last Played" behaviour of having play all the way through every track on the album is not very intuitive to me.

PR #10619, opened 2016-10-04
[python] add xbmcgui.ListItem.setCast()

phil65:
A separate setter for cast.
With current means it is not possible to set actor thumbnails since setInfo() only takes a name-role tuple for cast.
This PR adds a separate new setter to keep bw compatibility. I adjusted the dict keys in a way that it can directly take a cast dict received via JSON getMovie(s) method.

@tamland, as usual. :)

Should I mark the setinfo-cast as deprecated?

EDIT: also recently requested in http://forum.kodi.tv/showthread.php?tid=293913

### DIFF
--- a/xbmc/interfaces/legacy/ListItem.cpp
+++ b/xbmc/interfaces/legacy/ListItem.cpp
@@ -588,6 +588,30 @@ namespace XBMCAddon
       }
     } // end ListItem::setInfo
 
+    void ListItem::setCast(const std::vector<Properties>& actors)
+    {
+      LOCKGUI;
+      item->GetVideoInfoTag()->m_cast.clear();
+      for (const auto& dictionary: actors)
+      {
+        SActorInfo info;
+        for (auto it = dictionary.begin(); it != dictionary.end(); ++it)
+        {
+          const String& key = it->first;
+          const String& value = it->second;
+          if (key == "name")
+            info.strName = value;
+          else if (key == "role")
+            info.strRole = value;
+          else if (key == "thumbnail")
+            info.thumbUrl = value;
+          else if (key == "order")
+            info.order = strtol(value.c_str(), NULL, 10);
+        }
+        item->GetVideoInfoTag()->m_cast.push_back(std::move(info));
+      }
+    }
+
     void ListItem::addStreamInfo(const char* cType, const Properties& dictionary)
     {
       LOCKGUI;
--- a/xbmc/interfaces/legacy/ListItem.h
+++ b/xbmc/interfaces/legacy/ListItem.h
@@ -629,6 +629,40 @@ namespace XBMCAddon
 #ifdef DOXYGEN_SHOULD_USE_THIS
       ///
       /// \ingroup python_xbmcgui_listitem
+      /// @brief \python_func{ setCast(actors) }
+      ///-----------------------------------------------------------------------
+      /// @brief Set cast including thumbnails
+      ///
+      /// @param actors            list of dictionaries (see below for relevant keys)
+      ///
+      /// - Keys:
+      /// | Label         | Description                                     |
+      /// |--------------:|:------------------------------------------------|
+      /// | name          | string (Michael C. Hall)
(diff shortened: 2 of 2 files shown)


xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -588,6 +588,30 @@ namespace XBMCAddon
       }
     } // end ListItem::setInfo
 
+    void ListItem::setCast(const std::vector<Properties>& actors)
+    {
+      LOCKGUI;
+      item->GetVideoInfoTag()->m_cast.clear();
+      for (const auto dictionary: actors)
```
tamland: `const auto&`

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -588,6 +588,30 @@ namespace XBMCAddon
...
+      {
+        SActorInfo info;
+        for (auto it = dictionary.begin(); it != dictionary.end(); ++it)
+        {
+          const String& key = it->first;
+          const String value(it->second.c_str());
```
tamland: unnecessary copy. bind to `&` like above instead

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -588,6 +588,30 @@ namespace XBMCAddon
...
+          if (key == "name")
+            info.strName = value;
+          else if (key == "role")
+            info.strRole = value;
+          else if (key == "thumbnail")
+            info.thumbUrl = value;
+          else if (key == "order")
+            info.order = strtol(value.c_str(), NULL, 10);
+        }
+        item->GetVideoInfoTag()->m_cast.push_back(info);
```
tamland: info can be moved. i.e `.push_back(std::move(info));`

phil65:
updated.

MartijnKaijser:
Jenkins build this please

phil65:
jenkins build this please.

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -588,6 +588,30 @@ namespace XBMCAddon
...
+          const String& key = it->first;
+          const String& value = it->second.c_str();
+          if (key == "name")
+            info.strName = value;
+          else if (key == "role")
+            info.strRole = value;
+          else if (key == "thumbnail")
+            info.thumbUrl = value;
+          else if (key == "order")
+            info.order = strtol(value, NULL, 10);
```
Rechi: `strtol(value.c_str(), NULL, 10)`

phil65: just to learn: why do i need .c_str() twice for this? It´s already applied in line 601.

phil65:
jenkins build this please.

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -588,6 +588,30 @@ namespace XBMCAddon
...
+    {
+      LOCKGUI;
+      item->GetVideoInfoTag()->m_cast.clear();
+      for (const auto& dictionary: actors)
+      {
+        SActorInfo info;
+        for (auto it = dictionary.begin(); it != dictionary.end(); ++it)
+        {
+          const String& key = it->first;
+          const String& value = it->second.c_str();
```
phil65: I used this because ListItem.addStreamInfo() uses the same. What would be a better alternative here in your opinion?

phil65: that´s how i did it in the beginning, but changed it based on @tamland comment (see above)
So should I change it back?

phil65: ok, thx!

MartijnKaijser:
@tamland ping for review if you have a moment

tamland:
looks good (sorry for slow reply)

MartijnKaijser:
Jenkins build this please

PR #10622, opened 2016-10-05
VAAPI: restore accidentally lost call to sync of vaapi render picture

FernetMenta:
see title

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererVAAPI.cpp
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererVAAPI.cpp
@@ -202,5 +202,13 @@ bool CRendererVAAPI::UploadTexture(int index)
   return true;
 }
 
+void CRendererVAAPI::AfterRenderHook(int idx)
+{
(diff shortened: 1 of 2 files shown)


FernetMenta:
jenkins build this please

fritsch:
To play devil's advocate here: does that fix a known issue?
Background: we run that since several months without any bugreport now by accident. I currently don't understand why it works currently without the fences and wanted to understand that.

Any idea?

FernetMenta:
> Any idea?

The sync is required for a deterministic behaviour. Without it you can get buffer overwrites or decoder pipeline can block waiting for gl operations to be finished.
Without the sync it may work by chance depending on the drivers implementation and other components you can't influence directly. Not seeing any obvious issues does not mean that the code was correct. See vda on OSX. It has always been broken but the symptoms started showing on some systems as we changed the format of the video surface.

If you don't do this right, you'll end up in very hard to track down errors like the AMD asic hang you should remember: https://github.com/xbmc/xbmc/pull/1649

fritsch:
Thx. So in essence: it worked by chance and not cause we have some serialized code somewhere, which should be investigated.

FernetMenta:
I don't claim this but I have no indicators that we do.

PR #10623, opened 2016-10-05
paplayer: fix gapless playback

FernetMenta:
see title

http://forum.kodi.tv/showthread.php?tid=250744

### DIFF
--- a/xbmc/cores/paplayer/AudioDecoder.cpp
+++ b/xbmc/cores/paplayer/AudioDecoder.cpp
@@ -170,7 +170,7 @@ int64_t CAudioDecoder::TotalTime()
   return 0;
 }
 
-unsigned int CAudioDecoder::GetDataSize()
+unsigned int CAudioDecoder::GetDataSize(bool checkPktSize)
 {
   if (m_status == STATUS_QUEUING || m_status == STATUS_NO_FILE)
     return 0;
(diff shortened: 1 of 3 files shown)


fritsch:
That thingy gets more ugly over time with every commit. The special casing in all its places makes it kind of unmaintainable. 

Rewrite for v18? :-)

FernetMenta:
I won't hold you back if you like to do so.

FernetMenta:
I won't hold you back from looking into this either :)

arnova:
I've been using videoplayer (previously dvdplayer) for music playback since the Xbox age. For me it's the only way to get guaranteed stutter-free playback (and I don't care for gapless/crossfade). So +1 from me.

fritsch:
The stutter should be fixed, btw since 1.5 years now. Mind retrying?

Am 05.10.2016 12:35 nachm. schrieb "arnova" <notifications@github.com>:

FernetMenta:
jenkins build this please

fritsch:
Background: the stutter was a combined softae misconception and in thread next song queing issue by paplayer.

popcornmix:
Stutter is very common with audio when using paplayer with a slow network, like wifi.
This is because paplayer has no buffering unless you set buffermode=1.
VideoPlayer avoids this as it has substantial internal buffering.

arnova:
@fritsch : I'll give it another try. @popcornmix : I'm aware of that, but in this case it concerned local (hdd) playback. And like I said it dates back to the Xbox age, so I'm not 100% it's a SoftAE issue but rather a design flaw.

PR #10625, opened 2016-10-05
Add custom request support to CCurl

ghost:
Add custom request support to CCurl
## Description

This PR adds support to override the standard GET / POST requests
## Motivation and Context

Some content provider provide necessary information in HEAD / OPTIONS call
## How Has This Been Tested?

Runtime test
## Types of change
- [ x] New feature (non-breaking change which adds functionality)
## Checklist:
- [x ] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 

### DIFF
--- a/xbmc/filesystem/File.cpp
+++ b/xbmc/filesystem/File.cpp
@@ -235,6 +235,11 @@ bool CFile::CURLAddOption(XFILE::CURLOPTIONTYPE type, const char* name, const ch
     m_curl.SetOption(name, value);
     break;
   }
+  case XFILE::CURL_OPTION_REQUEST:
+  {
+    m_curl.SetCustomRequest(value);
+    break;
+  }
   default:
     return false;
   }
--- a/xbmc/utils/URIUtils.cpp
+++ b/xbmc/utils/URIUtils.cpp
@@ -466,7 +466,9 @@ std::string URIUtils::ChangeBasePath(const std::string &fromPath, const std::str
 CURL URIUtils::SubstitutePath(const CURL& url, bool reverse /* = false */)
 {
   const std::string pathToUrl = url.Get();
-  return CURL(SubstitutePath(pathToUrl, reverse));
+  CURL url_tmp(CURL(SubstitutePath(pathToUrl, reverse)));
+  url_tmp.SetCustomRequest(url.GetCustomRequest());
+  return url_tmp;
 }
 
 std::string URIUtils::SubstitutePath(const std::string& strPath, bool reverse /* = false */)
--- a/xbmc/URL.cpp
+++ b/xbmc/URL.cpp
@@ -809,3 +809,13 @@ void CURL::RemoveProtocolOption(const std::string &key)
   m_protocolOptions.RemoveOption(key);
   m_strProtocolOptions = m_protocolOptions.GetOptionsString(false);
 }
+
+void CURL::SetCustomRequest(const std::string &request)
+{
+  m_strCustomRequest = request;
+}
+
+const std::string CURL::GetCustomRequest() const
+{
+  return m_strCustomRequest;
+}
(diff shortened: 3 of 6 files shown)


xbmc/filesystem/File.cpp
```diff
@@ -283,7 +288,10 @@ bool CFile::Open(const CURL& file, const unsigned int flags)
         return m_pFile->Open(url);
       }
     }
-    m_pFile = CFileFactory::CreateLoader(url);
+    if (m_pFile)
+      m_pFile->Close();
+    else
+      m_pFile = CFileFactory::CreateLoader(url);
```
arnova: What's the idea behind this?

xbmc/utils/URIUtils.cpp
```diff
@@ -466,7 +466,9 @@ std::string URIUtils::ChangeBasePath(const std::string &fromPath, const std::str
 CURL URIUtils::SubstitutePath(const CURL& url, bool reverse /* = false */)
 {
   const std::string pathToUrl = url.Get();
-  return CURL(SubstitutePath(pathToUrl, reverse));
+  CURL url_tmp(CURL(SubstitutePath(pathToUrl, reverse)));
+  url_tmp.SetCustomRequest(url.GetCustomRequest());
+  return url_tmp;
```
arnova: Why here? Not really the place IMO.

mapfau: The existing concept at this place is soooooo ugly:

- build an URL String from all members (get()), this is [Protocol]/URL/....|Header|options
- Then build a copy of this (wich means exploding all the things wich was concatted a line above) and substituting path.

Never ever in my life seen such a waste of resources.

But: I don't see any other place. 

Cold you pls. so nice and let me know wich place is better?
Concatting to the string (get) is imo WRONG!

xbmc/filesystem/File.cpp
mapfau: keep-alive (reusing of conection) works only on the same curl_easy handle.
You can call multiple times open() now, was not possible before.

(yes it was possible, but lead to memory leaks :-) So its more an bugfix as an feature).

arnova: This only concerns CurlFile, right? IMO it should be in that class, not here.

And if, for some reason, it really needs to be here please add a proper comment. In one month no one will know why it is the way it is.

mapfau: It's for every file. If you call open multiple times without closing close() before, the m_pFile pointer gets lost as memoryleak.

xbmc/utils/URIUtils.cpp
mapfau: In CCurlfile this member is already gone away. No chance without reengeneering all this mess here

xbmc/filesystem/File.cpp
arnova: Nevertheless: Please add a proper comment

xbmc/utils/URIUtils.cpp
arnova: My problem here is that the function is called "SubstitutePath" and not "SubstitutePath and Other Do Some Other Stuff Too", so that's why it doesn't belong there. By accident SubstitudePath() is called in places where you need to set the custom header right, but I wouldn't rely on the fact that this is always in the same cases as SubstitudePath. Perhaps we just need to add a new member variable to IFile ? This seems like a much cleaner approach.

xbmc/filesystem/File.cpp
arnova: Ideally yes, but still I doubt everyone will understand why this is, the way it is now. That's why I'd like a comment.

mapfau:
No need from my side - will close this issue. Sorry for the time.

PR #10626, opened 2016-10-05
[estuary] country flags - use RGB color space

ronie:
fixes: http://forum.kodi.tv/showthread.php?tid=262373&pid=2430406#pid2430406

@phil65


kib (approved):
A fix where you can't even see the change 👍

ronie:
let's get it in....even if it's for the time being.

suggestions have been made to outsource these flags to a resource addon
or perhaps just drop them all together (as it's not really possible to map a language to a country)

PR #10627, opened 2016-10-05
Export max value in ratings by default

phate89:
We support a max value to calc the true rating but we never exported it or documented it. This could have solved some problem we have now with switched music ratings from 5 to 10..


phate89:
@DaveTBlake as discussed in slack

DaveTBlake (approved):
Looks fine to me

phate89:
jenkins build this please

PR #10628, opened 2016-10-05
close and reuse a file if open is called multiple times

ghost:
Reuse CFile in case of multiple open() calles without close()
## Description

see title
## Motivation and Context

Calling multiple times open on a filesystem file without calling Close() after each call lead to memory leaks (m_pFile is not closed / freed) -> Avoid memory leaks.

Calling multiple times Open() makes sense if you have to deal with keep-alive connections.
Conection reuse in curl is only supporten on one curl_easy handle.

This PR simply Closes() the underlying file but holds the object alive.
Keepalive connections are now possible.
## How Has This Been Tested?

Send OPTIONS with following GET over the Addon CURLxxx Addon API
## Types of change
## Checklist:

### DIFF
--- a/xbmc/filesystem/File.cpp
+++ b/xbmc/filesystem/File.cpp
@@ -283,7 +283,11 @@ bool CFile::Open(const CURL& file, const unsigned int flags)
         return m_pFile->Open(url);
       }
     }
-    m_pFile = CFileFactory::CreateLoader(url);
+
+    if (m_pFile)
+      m_pFile->Close();
+    else
+      m_pFile = CFileFactory::CreateLoader(url);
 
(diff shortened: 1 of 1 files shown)


FernetMenta:
imo this is a bit hacky and tailored to a very specific use case. open() should not implicitly close/open the underlying file. Nobody ever worked with an abstraction of a file based API would expect this.
why don't you create a new method that implements this specific behaviour?

mapfau:
IMO its normal usecase, and how its now its an open bug wich can produce memory leaks.
2 opions to close this bug: delete m_pfile, or close it.

I'm good in closing this PR. It was just an Idea, I don't need it.

PR #10629, opened 2016-10-05
Dont open CFile on Open again if its already opened

ghost:
Dont open CFile on Open again if its already opened
## Description

see title
## Motivation and Context

Calling multiple times open on a filesystem file without calling Close() after each call lead to memory leaks (m_pFile is not closed / freed).

CCurlfile is not able to handle multiple Open calls -> Close it on start of Open
## How Has This Been Tested?

Send OPTIONS with following GET over the Addon CURLxxx Addon API
## Types of change
## Checklist:

### DIFF
--- a/xbmc/filesystem/File.cpp
+++ b/xbmc/filesystem/File.cpp
@@ -254,6 +254,19 @@ bool CFile::Open(const std::string& strFileName, const unsigned int flags)
 
 bool CFile::Open(const CURL& file, const unsigned int flags)
 {
+  if (m_pFile)
+  {
+    if ((flags & READ_REOPEN) == 0)
+    {
+      CLog::Log(LOGERROR, "File::Open - already open: %s", file.GetRedacted().c_str());
+      return false;      
+    }
+    else
+    {
+      return m_pFile->ReOpen(URIUtils::SubstitutePath(file));
+    }
+  }
+
   m_flags = flags;
   try
   {
@@ -280,6 +293,10 @@ bool CFile::Open(const CURL& file, const unsigned int flags)
       {
         // for internet stream, if it contains multiple stream, file cache need handle it specially.
         m_pFile = new CFileCache(m_flags);
+
+        if (!m_pFile)
+          return false;
+
         return m_pFile->Open(url);
       }
     }
--- a/xbmc/filesystem/File.h
+++ b/xbmc/filesystem/File.h
@@ -64,12 +64,25 @@ public:
   bool CURLAddOption(XFILE::CURLOPTIONTYPE type, const char* name, const char * value);
   bool CURLOpen(unsigned int flags);
 
+  /**
+  * Attempt to open an IFile instance.
+  * @param file reference to CCurl file description
+  * @param flags see IFileTypes.h
+  * @return true on success, false otherwise
+  *
+  * Remarks: Open can only be called once. Calling
+  * Open() on an already opened file will fail
+  * exept flag READ_REOPEN is set and the underlying
+  * file has an implementation of ReOpen().
+  */
   bool Open(const CURL& file, const unsigned int flags = 0);
+  bool Open(const std::string& strFileName, const unsigned int flags = 0);
+
   bool OpenForWrite(const CURL& file, bool bOverWrite = false);
+  bool OpenForWrite(const std::string& strFileName, bool bOverWrite = false);
+
   ssize_t LoadFile(const CURL &file, auto_buffer& outputBuffer);
 
-  bool Open(const std::string& strFileName, const unsigned int flags = 0);
-  bool OpenForWrite(const std::string& strFileName, bool bOverWrite = false);
   /**
    * Attempt to read bufSize bytes from currently opened file into buffer bufPtr.
    * @param bufPtr  pointer to buffer
--- a/xbmc/filesystem/CurlFile.cpp
+++ b/xbmc/filesystem/CurlFile.cpp
@@ -1189,6 +1189,12 @@ bool CCurlFile::CReadState::ReadString(char *szLine, int iLineLength)
   return (bool)((pLine - szLine) > 0);
 }
 
+bool CCurlFile::ReOpen(const CURL& url)
+{
+  Close();
+  return Open(url);
+}
+
 bool CCurlFile::Exists(const CURL& url)
 {
   // if file is already running, get info from it
--- a/xbmc/filesystem/CurlFile.h
+++ b/xbmc/filesystem/CurlFile.h
@@ -51,6 +51,7 @@ namespace XFILE
       virtual ~CCurlFile();
       virtual bool Open(const CURL& url);
       virtual bool OpenForWrite(const CURL& url, bool bOverWrite = false);
+      virtual bool ReOpen(const CURL& url);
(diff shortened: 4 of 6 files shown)


mapfau:
better?

Paxxi:
Seems sane

FernetMenta:
open() is not supposed to be called on an open file. I suggest you create an extra method for this special case.

mapfau:
Its not suggested, but possible.
No place in the code disallows it.

Why should I implement a special method to fix a leak?
I don't understand.

Its a bug imo.

xbmc/filesystem/File.cpp
```diff
@@ -283,7 +283,9 @@ bool CFile::Open(const CURL& file, const unsigned int flags)
         return m_pFile->Open(url);
```
ace20022: Is ```m_file``` guaranteed to be ```null``` here (too)

FernetMenta:
It is not really a bug. This method was not designed to be called if already open. The correct fix is to make open fail if there is already an implementation open.
I don't think you have checked all implementations of IFile if they are safe to be opened multitple times, did you?

xbmc/filesystem/File.cpp
mapfau: I wouldn't be sure here.... if cache is cleared between the check and the call it could be null.

mapfau:
I'm quite sure that there are only 2 Opens, but the first calls the second one.

xbmc/filesystem/File.cpp
mapfau: ah, missed the line - no must NOT be != nullptr , this has to be checked

FernetMenta:
As mentioned this is not a bug. The API is designed in a way that open is supposed to be called only once. Hence there is no mem leak. This could be enforced by a check if open, then return false.

Allowing open to be called on already open instances would be a major API change that would require every single implementation of IFile, including open PR for a VFS extension point, to be reworked that multiple opens are supported.

FernetMenta:
@mapfau looking at you last answer I am not sure if you are aware that CFile serves as container for all our implementations of IFile, not only CCurlFile. If Open is redefined that is complies to the API that it can be called multiple times, you would have one of the next issues here in FTPFile. https://github.com/xbmc/xbmc/blob/master/xbmc/filesystem/SFTPFile.cpp#L604
Calling FTPFile::Open an an already open sessions creates a mem leak too.

FernetMenta:
If open had some documentation it would say that you only can call it once, not on instances that were already opened. All implementations of IFile comply to this. A method that is not supposed to be called multiple times does not need protection for mem leaks, it should log an error instead.

garbear:
Right, the problem is that there's no documentation. A description should be added [here](https://github.com/xbmc/xbmc/blob/c21bf03/xbmc/filesystem/IFile.h#L62) to make it mandatory that Open() fail if already open.

mapfau:
If this is from concept / naming OK I would continue with the addon interface

xbmc/filesystem/File.h
```diff
@@ -64,12 +64,33 @@ class CFile
...
   bool CURLOpen(unsigned int flags);
 
+  /**
+  * Attempt to open an IFile instance.
+  * @param file reference to CCurl file description
+  * @param flags see IFileTypes.h
+  * @return true on success, false otherwise
+  *
+  * Remarks: Open can only be called once. Calling
+  * Open() on an already opend file will fail.
```
garbear: should be `opened`

xbmc/filesystem/File.cpp
```diff
@@ -387,6 +397,21 @@ bool CFile::OpenForWrite(const CURL& file, bool bOverWrite)
   return false;
 }
 
+bool CFile::ReOpen(const CURL& file)
+{
+  if (!m_pFile)
```
FernetMenta: how about?
if (!m_pFile || !m_pFile->SupportsReOpen())

virtual bool IFile::SupportsReOpen() { return false;}

implementation of IFile can override this

FernetMenta:
concept is ok in general

xbmc/filesystem/File.h
mapfau: thx

xbmc/filesystem/File.cpp
mapfau: good idea

MrMC:
you might take a real close look at the life cycle of m_state before calling this done :)

xbmc/filesystem/File.cpp
arnova: An alternative could be:
bool CFile::Open(const CURL& file, const bool &bReOpen = false)
....
I think this doesn't break API as well, right?

mapfau:
@MrMC The tests I did have been successful (connection was reused).
If you have seen something I have overseen, pls. let me know. Thx!

xbmc/filesystem/File.cpp
mapfau: Open already has "flags" parameter wich could also be used (without any API change),
@FernetMenta pls let me know if this is an option, then I can save the time putting ReOpen at 100 places in the api

peak3d:
Implementation uses flags field now

FernetMenta:
looks good

xbmc/filesystem/File.cpp
```diff
@@ -254,6 +254,19 @@ bool CFile::Open(const std::string& strFileName, const unsigned int flags)
 
 bool CFile::Open(const CURL& file, const unsigned int flags)
 {
+  if (m_pFile)
+  {
+    if ((flags & READ_REOPEN) == 0)
+    {
+      CLog::Log(LOGERROR, "File::Open - already open: %s", file.GetRedacted().c_str());
+      throw;
```
garbear: can you add a message so that a handler (or at least a debugger) can show what's wrong?

peak3d: sure, just add a string behind the throw?

FernetMenta: throw will likely kill the app because there is no catch. please drop throw. a bid disadvantage of current c++ standard is that there is no way to enforce a catch statement.

peak3d: https://github.com/xbmc/xbmc/pull/10629/files#diff-7a28d3fe5a6629e2db83f3e96cc1b10bR262

FernetMenta:
@peak3d this pr is now submitted by a deleted user. please close an re-submit with reference to this one.

xbmc/filesystem/File.cpp
Paxxi: @FernetMenta I suggested it for that reason, as this code path is a developer error and not runtime error it helps catch mistakes early. I prefer an uncaught throw over terminate as throw will break into debugger or trigger a crash dump.

peak3d:
https://github.com/xbmc/xbmc/pull/10658

xbmc/filesystem/File.cpp
FernetMenta: for that purpose you could place an assert(0). as long as catch can not be enforced, we should not use throw at all. such kind of things can spread like cancer.

recently I tracked down an issue caused by one of our silly throw statements. this was not trivial to spot when all you get is a log file.

PR #10632, opened 2016-10-05
[PVR] Fix: PVR windows never must list removable media.

ksooo:
PVR windows are derived from CGUIMediaWindow. Default behavior of this class is to include removable media (if present) in directory listing in case the listing is empty otherwise.

=> https://github.com/xbmc/xbmc/blob/master/xbmc/filesystem/VirtualDirectory.cpp#L168

Example:

screenshot032-phil65

This makes no sense for PVR windows. Thus, this PR disables this nonsense base class functionality for all PVR windows. 

@xhaggi for code review?
@phil65 fyi

### DIFF
--- a/xbmc/pvr/windows/GUIWindowPVRBase.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRBase.cpp
@@ -69,6 +69,9 @@ CGUIWindowPVRBase::CGUIWindowPVRBase(bool bRadio, int id, const std::string &xml
   m_bRadio(bRadio),
   m_progressHandle(nullptr)
 {
+  // prevent removable drives to appear in directory listing (base class default behavior).
+  m_rootDir.AllowNonLocalSources(false);
+
   m_selectedItemPaths[false] = "";
   m_selectedItemPaths[true] = "";
 


ronie:
Will there be a '..' item in the list instead?

xhaggi:
looks sane +1, never had this issue.

xhaggi:
there should be an item `No item found` IIRC, becauee `..` makes no sense in the pvr search window

ronie:
yes that would work as well. as long as there's something listed, it's ok.

xhaggi:
BTW with the current beta I have the same issue. My internal mounted HDD is shown here.

xhaggi:
@ronie sure because of the empty list container selection issue ;)

ksooo:
> Will there be a '..' item in the list instead?

search window: "Search..."
timers / timer rules window: "Add timer..."
channels: ".." (leading you to home screen - kinda hacky, but not introduced by this PR)

other windows are not affected.

ksooo:
> BTW with the current beta I have the same issue. My internal mounted HDD is shown here.

 it's an ancient bug, even present Jarvis, maybe even before.

xhaggi:
@ksooo don't know the right label, but if this PR fix it, it'sgood to go in from my side.

ksooo:
jenkins build this please

xhaggi:
never had this issue in Jarvis, but don't you this window as much.

ksooo:
> never had this issue in Jarvis

 it IS in Jarvis. ;-)

phil65:
Thanks for the quick fix!

ksooo:
Jenkins WIN-32 error is unrelated.

PR #10634, opened 2016-10-06
guilib: fix initial focus when opening GUiMediaWindow with plugin con…

FernetMenta:
fix for http://trac.kodi.tv/ticket/16786


phil65:
Thx, this bugged me for a while. :)

mazeyx:
this also happens to me when playing stuff from tvshows/movie library, but not video. not sure if you guys experience this also might be android only bug.

PR #10640, opened 2016-10-06
CUtil::MakeShortenPath() should return true if the given path is already short enough (fixes #16783)

Montellese:
This fixes the behaviour of `CUtil::MakeShortenPath()` which in turn fixes the fact that right now any setting pointing at a path which is 30 characters or shorter does not show up in the GUI at all. If the path is longer than 30 characters it is shortened and properly shown. This has been reported in http://trac.kodi.tv/ticket/16783.


Montellese:
jenkins build this please

MartijnKaijser:
Thx.  I thought it was fixed in mean time as I saw text in screenshot location hence I closed my ticket

Montellese:
Yeah because the problem only exists for paths shorter than 31 characters.

MartijnKaijser:
feel free to merge when ready.

PR #10645, opened 2016-10-07
[gui] field and operator buttons from label2 to label

BigNoid:
Change the button labels from right to left.

Before:
schermafbeelding 2016-10-07 om 15 32 07

After:
schermafbeelding 2016-10-07 om 15 29 55


phil65:
yup, fine for me, thx.

MartijnKaijser (approved):

ronie:
no-brainer. 
thx @BigNoid

HitcherUK:
@BigNoid @ronie @phil65 Capitalise 'contains' now? Or in case it's used elsewhere add a new string?

ronie:
it isn't used elsewhere, but if you change one, you need to change all the other operators as well

MartijnKaijser:
i don't think "contains" should be capitalised as it's a mid sentence operator.

> Title FOO contains BAR

HitcherUK:
Good point MartijnKaijser, I wasn't thinking about it in that way.

PR #10646, opened 2016-10-07
[estuary] another sync, misc fixes / additions

phil65:
@ronie

### DIFF
--- a/addons/skin.estuary/1080i/DialogContextMenu.xml
+++ b/addons/skin.estuary/1080i/DialogContextMenu.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <window>
-	<defaultcontrol>1000</defaultcontrol>
+	<defaultcontrol>996</defaultcontrol>
 	<include>Animation_DialogPopupOpenClose</include>
 	<coordinates>
 		<left>0</left>
@@ -16,6 +16,15 @@
 			<animation effect="fade" start="0" end="100" time="300">WindowOpen</animation>
 			<animation effect="fade" start="100" end="0" time="200">WindowClose</animation>
 			<texture colordiffuse="B0FFFFFF">colors/black.png</texture>
+		</control>
+		<control type="button" id="22001">
+			<left>-1920</left>
+			<top>-1080</top>
+			<width>5760</width>
+			<height>3240</height>
+			<texturefocus></texturefocus>
+			<texturenofocus></texturenofocus>
+			<onclick>Action(close)</onclick>
 		</control>
 		<control type="image" id="999">
 			<description>background image</description>
--- a/addons/skin.estuary/1080i/DialogAddonSettings.xml
+++ b/addons/skin.estuary/1080i/DialogAddonSettings.xml
@@ -47,7 +47,7 @@
 			<width>1060</width>
 			<height>680</height>
 			<onup>2</onup>
-			<ondown>9001</ondown>
+			<ondown>2</ondown>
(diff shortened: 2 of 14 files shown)


addons/skin.estuary/1080i/DialogContextMenu.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <window>
-	<defaultcontrol>1000</defaultcontrol>
+	<defaultcontrol>996</defaultcontrol>
```
phil65: no real hard reason, but I think it makes more sense to put focus on the grouplist instead of a button template?

phil65: The "real" buttons start with 1001 (see https://github.com/xbmc/xbmc/blob/8d4a5bba55638dfd0bdc5e7de34f3e5293f99933/xbmc/dialogs/GUIDialogContextMenu.cpp#L53 ), so i think 1000 doesnt make too much sense here.

phil65: in fact 1000 gets disabled: https://github.com/xbmc/xbmc/blob/8d4a5bba55638dfd0bdc5e7de34f3e5293f99933/xbmc/dialogs/GUIDialogContextMenu.cpp#L123

ronie: thx!

ronie:
looking good

MilhouseVH:
Are the changes to Wall and Infowall views intentional? I (and [others](http://forum.kodi.tv/showthread.php?tid=269814&pid=2432859#pid2432859) preferred the original more compact layouts...

Here's Wall:

Before:
s1

After:
s2

And InfoWall:

Before:
s3

After:
s4

phil65:
yup, change was made because now it supports thumbs with non-poster format.
not happy with it? i might tweak it a bit more.

MilhouseVH:
Personally I preferred the more compact look as all of my thumbs are posters. Now, it just looks a lot less neat with so much wasted space.

However as it's intentional I guess the discussion should continue in the Estuary forum - I'll make a post and then you can gauge the feedback.

phil65:
ok, cool.

phil65:
I changed wall now to the layout used by infowall:
screenshot035
...and i might remove the year label since it looks a bit messy i think.

This is the corresponding confluence layout which gets "replaced".
screenshot033

Tbh I would also prefer a just-poster view, but i think this here works better when takin plugins, alternate scrapers etc into account.

MilhouseVH:
OK, if you've changed it already there's no point me posting the current (new) look on the forum (I was in the process of uploading the old/new look to an imgur gallery, and failing!), but I do think this may need wider discussion before committing to these changes - I don't really see the point of adding the extra text below the thumbs in these Wall/InfoWall views, as to me a "Wall" view should just be about the artwork, with as much artwork visible as possible. Having to support non-poster thumbs now means that the majority of artwork (typically posters) is now surrounded by blank space. If there's a need to support non-poster artwork then maybe it should be done with a new/extra view?

The revised Wall view is better, but that text... hmmm. :)

rottengibbs:
Can we get the old Wall view back? For me the new look is terrible.. No need for that Background and text. Make it optional ;)

zag2me:
Just adding my support to removing the year from title. It is unnecessary on that view.

redglory:
+1 for previous wall and infowall versions

Cocotus:
+1 for previous poster only option please!

PR #10647, opened 2016-10-07
VideoPlayer: aml: Disable pullup correction when using Amlogic decoder

codesnake:
Disable pullup correction in video player when using Amlogic decoder.
## Description

When using Amlogic decoder to play a video, playback may freeze at some point. Disabling pullup correction fixes that issue.
## Motivation and Context

Before switching to VideoPlayer, Amlogic decoder used RENDER_FMT_BYPASS picture format and pullup correction was disabled for it, but after the switch the decoder uses RENDER_FMT_AML picture format and pullup correction became enabled for it.
## How Has This Been Tested?

This was tested on WeTek Play, WeTek Play 2 and WeTek Hub with LibreELEC installed, using different video formats.
## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayerVideo.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayerVideo.cpp
@@ -800,7 +800,7 @@ int CVideoPlayerVideo::OutputPicture(const DVDVideoPicture* src, double pts)
   int result = 0;
 
   //correct any pattern in the timestamps
-  if (picture.format != RENDER_FMT_BYPASS)
+  if (picture.format != RENDER_FMT_BYPASS && picture.format != RENDER_FMT_AML)
   {
     m_pullupCorrection.Add(pts);
     pts += m_pullupCorrection.GetCorrection();


FernetMenta:
-1
bypassing this function is wrong, for AML and for  all renderer.

FernetMenta:
maybe m_pullupCorrection.GetCorrection has an issue in general, maybe not. until we don't know the exactly what's going wrong, I don't agree to alternative code paths in VideoPlayer

codesnake:
But Kodi doesn't do any rendering for AML at all.

FernetMenta:
it controls the timing and this need to be correct. otherwise audio and subs can't be synced.

xbmc/cores/VideoPlayer/VideoPlayerVideo.cpp
```diff
@@ -800,7 +800,7 @@ int CVideoPlayerVideo::OutputPicture(const DVDVideoPicture* src, double pts)
   int result = 0;
 
   //correct any pattern in the timestamps
-  if (picture.format != RENDER_FMT_BYPASS)
+  if (picture.format != RENDER_FMT_BYPASS && picture.format != RENDER_FMT_AML)
   {
```
FernetMenta: a hack that is already in place does not justify another hack

PR #10649, opened 2016-10-07
[cmake/jenkins] Build with CMake

fetzerch:
Removes the possibility to build with legacy build systems on jenkins.
Jenkins is already building with cmake on all platforms for some time.


Memphiz:
fine by me - maybe leave the var in the jenkins job for a couple of weeks in case devs without this PR want to build ... (it wouldn't set CMAKE_BUILD at all if it was removed from jenkins, and then it might try to build with native buildsys ...)

PR #10650, opened 2016-10-07
[PVR] Do not restart pvr manager on async client reconnect.

ksooo:
Although PVR manager must be started on first async client connect, do not restart pvr manager on subsequent async client reconnect. This is not necessary and a waste of machine time. 

@FernetMenta your opinion?


FernetMenta:
not sure. if PVRManager is not restarted, it may not be in sync with backend. vnsi triggers updates of epg or channels for example. if not connected kodi's pvr records won't get updated. a restart synchronises pvr with backend. hence I think it is required.
I don't see an disadvantages of a restart.

ksooo:
I guess you are right. if we are not sure whether this causes problems we should not change this. At least not now.

PR #10398, opened 2016-09-04
[cmake] Add uninstall target for linux

fetzerch:
Requested by @FernetMenta. Unfortunately a bit slow.

ping @hudokkow for fc70dc2

PR #10656, opened 2016-10-08
[cmake] Fix double install of skins

fetzerch:
Skins were installed twice on linux (to the same place) so that make uninstall reported errors.

https://github.com/xbmc/xbmc/pull/10398 contained a commit that ensured that skins are not installed twice on linux. The PR was reverted in https://github.com/xbmc/xbmc/pull/10454 because it broke skin packaging on windows.

After investigation, the second "install" on windows covered the issue, that the windows build script created empty xbt files. Those were overridden by cmake again so that it wasn't noticed.

Ping: @Paxxi


fetzerch:
jenkins build this please

hudokkow:
win32 error: Error in script "genNsisInstaller.nsi" on line 179 -- aborting creation process
https://github.com/xbmc/xbmc/blob/master/project/Win32BuildSetup/genNsisInstaller.nsi#L179

Paxxi:
GenNsisInstaller expects the skin files in win32buildsetup/BUILD_WIN32 so it likely needs to be modified to find the ones built by cmake

fetzerch:
Next round. http://jenkins.kodi.tv/job/WIN-32/10218/

I'm not sure if the change to genNsisIncludes.bat is correct, because also the current installer doesn't show skin related options.

MartijnKaijser:
@fetzerch nope it's not correct. You need to fix this line https://github.com/xbmc/xbmc/pull/10656/files#diff-8baa35259a2ba40e143d79c0a0328babL179  not remove it

fetzerch:
jenkins build this please

MartijnKaijser:
APK size looks normal

fetzerch:
good to go? quite surprisingly it didn't break android skin or binary addon packaging ;)

PR #10657, opened 2016-10-08
Add explicit ordering on song join queries - fix bug on MariaDB

DaveTBlake:
Fix bug that appears when using Kodi with a **MariaDB** server that causes tracks to be listed multiple times when music files are tagged with other artist contributions e.g. composer, lyricist, remixer etc.

Results dataset processing depends upon records being in song id order, but this was not explicit in the SQL. SQLlite and MySQL happen to order the results that way, MariaDB does not, and better to be explicit.

Testing needs songs with multiple artists and other contributors tagged. Look at songs filtered by genre, artist or album or full songs node. 
Also use `AudioLibrary.GetRecentlyAddedSongs`, and `GetRecentlyPlayedSongs` via JSON API. I suspect that these routines are rarely used, let alone on a MariaDB server but have been updated because they were similarly impacted.

I have tested on SQLite and MySQL that nothing is broken by this change, but would appreciate someone with MariaDB showing this does fix the issue.


BigNoid:
Tested with MariaDB and no more dupe tracks if there's composer information.

DaveTBlake:
Fix also confirmed on forum by MariaDB user that reported the issue.

DaveTBlake:
Bump JSON API minor version since has fixed  `AudioLibrary.GetSongs`, `GetRecentlyAddedSongs`, and `GetRecentlyPlayedSongs`  from returning incorrect results on a MariaDB server (even though no one has actually reported this).

Is that correct @Montellese ?

Montellese:
@DaveTBlake you bumped the fix version (not the minor version) but that's correct.

DaveTBlake:
So major.minor.fix - got it. Thanks @Montellese

PR #10658, opened 2016-10-08
support reusing of CFile::m_pFile

peak3d:
Dont open CFile on Open again if its already opened
## Description

see title
## Motivation and Context

Calling multiple times open on a filesystem file without calling Close() after each call lead to memory leaks (m_pFile is not closed / freed).

CCurlfile is not able to handle multiple Open calls -> Close it on start of Open
## How Has This Been Tested?

CURLCreate -> CURLOpen -> CURLOpen -> CURLClose
## Types of change
## Checklist:

### DIFF
--- a/xbmc/filesystem/CurlFile.cpp
+++ b/xbmc/filesystem/CurlFile.cpp
@@ -1189,6 +1189,12 @@ bool CCurlFile::CReadState::ReadString(char *szLine, int iLineLength)
   return (bool)((pLine - szLine) > 0);
 }
 
+bool CCurlFile::ReOpen(const CURL& url)
+{
+  Close();
+  return Open(url);
+}
+
 bool CCurlFile::Exists(const CURL& url)
 {
   // if file is already running, get info from it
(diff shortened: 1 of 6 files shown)


FernetMenta:
@peak3d this is required for addons, right?
@MartijnKaijser ok for v17?

MartijnKaijser:
@FernetMenta if you think it's OK no problem.

MartijnKaijser:
jenkins build this please

peak3d:
It's quite safe, and yes I'll use it for inputstream addons if it's merged.
Could also be used for e.g. scrapers as curl can better reuse once opened connections,

arnova:
I'm confused: I thought the idea was that when a file is already open and we perform a re-open we would simply use the old m_pFile. But now it seems we first close the file and then open it again. Isn't the latter less efficient? Or am I missing something?

peak3d:
Only CCurlFile implements ReOpen(). CCurlFile Close() lets m_state intact but resets all the other things wich needs to be resetted to allow a new Open() call. Open() then opens and writes using m_state with the still valid underlying easy_handle. This leads to reuse of the connection (if url matches the previous call)

IMO its the correct behaviour for a method called ReOpen.

arnova:
Ah yes, now I see what you're trying to do. Looks ok to me.

PR #10661, opened 2016-10-08
[Win32] DynamicDlls are not system-dll's

peak3d:
Dont mark DynamicDlls as SystemDll on Win32
## Description

see title
## Motivation and Context

Currently all dynamic loaded dll's wich are not hooked are marked as systemdlls.
This leads to the fact, that binary addons cannot be updated anymore, because they reside loaded.
Analogous to the other platforms isSystemDll is set to false if its loaded through DllLoaderContainer::LoadDll()
## How Has This Been Tested?

Start kodi, try to update inputstream.mpd
## Types of change
## Checklist:

### DIFF
--- a/xbmc/cores/DllLoader/DllLoaderContainer.cpp
+++ b/xbmc/cores/DllLoader/DllLoaderContainer.cpp
@@ -251,7 +251,7 @@ LibraryLoader* DllLoaderContainer::LoadDll(const char* sName, bool bLoadSymbols)
 #ifdef TARGET_POSIX
   pLoader = new SoLoader(sName, bLoadSymbols);
 #elif defined(TARGET_WINDOWS)
-  pLoader = new Win32DllLoader(sName);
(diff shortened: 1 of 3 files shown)


peak3d:
@paxxi no bool anymore :-)

Paxxi:
Imo system dlls should not be delayed but that's for another version, no objections to this

peak3d:
Just a side note from debugging: libcpluff.dll is also now marked as "Not a system dll" and can therefore be theoretically freed. I made some tests but there was no approach freing cpluff during runtime so it should be good to go.
Pls. let me know if the remark above has to be investigated, but I was pretty sure that's OK as cpluff is not marked as system dll on lall other builds.

Paxxi:
Let's wait a bit, want to take a look at it more thoroughly

Paxxi:
yeah we should be fine merging this, need to redesign it for v18

PR #8178, opened 2015-10-03
[GUILIB] add override onclick to dynamic content lists

InuSasha:
<content list="videodb://recentlyaddedmovies/">
&nbsp;&nbsp;<onclick>PlayMedia("$INFO[ListItem.FileNameAndPath]")</onclick>
</content>

if no onclick is defined, the default behavior is used.
the <content>videodb://..</content> is still working, too

additional: ExecuteAction methods gets an optional item as parameter.
This is used for ListItem infolabel while parse action string.

PR #10663, opened 2016-10-09
[guilib] make onclick override conditional. Fallback to default actio…

BigNoid:
…n if no condition is met.

In #8178 override on clicks were introduced for dynamic content lists. Problem is that once you add an onclick tag to a container it will always try to execute that onclick even if you put a condition to it. This fixes that problem.
Thx to @phil65 for the hint.

Imo this is a fix as previous behaviour was wrong but no problem to lift it to v18 too if anyone objects.


ronie:
yup, krypton material imo

phil65:
jenkins build this please.

PR #10664, opened 2016-10-09
X11, osx: yield to main loop after input action has been handled

FernetMenta:
see title

I noticed on OSX where holding down a key was blocking all other processing including rendering video.


FernetMenta:
jenkins build this please

ronie:
this introduces odd mouse behaviour on linux...
the cursor is lagging and keeps moving for a while after i stopped moving my mouse.

FernetMenta:
yes, we need to eat all mouse event until the last one when processing messages. I will think about this

FernetMenta:
in the meantime I revert this

PR #10665, opened 2016-10-09
VideoPlayer: fix subsequent seeks

FernetMenta:
skip seeks if processing of last seek is still in process

Actual change is 64c6ce2 and 8134a7e
The rest is cosmetics.

There is still a bit work left for getting smooth video when holding down key for StepForward/Backward

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -2496,6 +2496,20 @@ void CVideoPlayer::HandleMessages()
         continue;
       }
 
+      // skip seeks if player has not finished the last seek
+      if (m_CurrentVideo.id >= 0 &&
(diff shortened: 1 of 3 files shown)


popcornmix:
A quick test of this PR didn't show any problems with MMAL or OMXPlayer.

When disabling all skip steps except -30s,30s and holding down left or right I don't see picture updating (video continues to play normally), but seek bar updates and video resumes when the key is released.
I don't really mind that behaviour compared to the previous rather sporadic seeking.

FernetMenta:
@popcornmix yes, this is the remaining issue I mentioned in the comment. I want to fix this in the next round.

popcornmix:
No objections here.

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

MartijnKaijser:
jenkins build this please

PR #10666, opened 2016-10-09
AE: Explicitely check for DTS for making settings invisible

fritsch:
There are passthrough configurations where only AC3 is available. So - explicitely check for that setting.

### DIFF
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -2516,7 +2516,12 @@
           <level>2</level>
           <default>false</default>
           <dependencies>
-            <dependency type="visible" on="property" name="aesettingvisible" setting="audiooutput.config">audiooutput.passthrough</dependency>
+            <dependency type="visible">
+              <and>
+                <condition on="property" name="aesettingvisible" setting="audiooutput.config">audiooutput.dtspassthrough</condition>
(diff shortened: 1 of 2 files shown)


fritsch:
jenkins build this please (no computer for building here).

koying (approved):
Not tested, but +1 on the concept

fritsch:
@FernetMenta I have the feeling that I'm missing something here as in: it does not work :-)

fritsch:
I got it working (did not specify the samplerate correctly), though I am not happy with the AC3Transcode setting change, as it depends on the visibility now. Also this will most likely hide already enabled settings.

Open for ideas.

FernetMenta:
-1 for v17
There may be cases where DTS works although not explicitly set by the sink for what reason ever. When using IEC there is no difference between AC3 and DTS regarding the transport layer.

We should not risk breakage just for a displayed setting that may not work on a small number of systems.

fritsch:
Oki. I keep it open. Also I changed it - as AC3 works everywhere, so it's minimal now.

FernetMenta:
@fritsch yes, I missed this
+1

fritsch:
jenkins build this please

MartijnKaijser:
FireTV: AC3/EAC3options
SATV: Shows all options

PR #10671, opened 2016-10-09
Revert "X11, osx: yield to main loop after input action has been handled"

FernetMenta:
Reverts xbmc/xbmc#10664


FernetMenta:
causes issues with mouse events. I revert until I have a solution.

PR #10672, opened 2016-10-09
[estuary] smartplaylist rules container did not work with a mouse

ronie:
@phil65


phil65:
interesting... smells like a bug in core though, too?

phil65 (approved):

ronie:
you tell me :-)

but yeah, this was an odd one to track down.

phil65:
got it. The issue was that the layouts didnt have a width defined. So nothing really which has to get fixed in core i would say.

ronie:
feel free to close if you're going to fix it elsewhere

phil65:
nope, this one makes more sense. :)

PR #10674, opened 2016-10-10
VideoPlayer: fix relative seeks

FernetMenta:
Holding down the key generates many relative seek commands. If the calling thread is blocked, there is not enough time for rendering.
Now VideoPlayer processes those messages asynchronously.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDMessage.h
+++ b/xbmc/cores/VideoPlayer/DVDMessage.h
@@ -219,6 +219,7 @@ public:
   struct CMode
   {
     int time = 0;
+    bool relative = false;
     bool backward = false;
     bool flush = true;
     bool accurate = true;
@@ -231,6 +232,7 @@ public:
     m_mode(mode)
   {}
   int GetTime() { return m_mode.time; }
+  bool GetRelative() { return m_mode.relative; }
   bool GetBackward() { return m_mode.backward; }
   bool GetFlush() { return m_mode.flush; }
   bool GetAccurate() { return m_mode.accurate; }
--- a/xbmc/cores/VideoPlayer/DVDMessage.h
+++ b/xbmc/cores/VideoPlayer/DVDMessage.h
@@ -216,31 +216,30 @@ private:
 class CDVDMsgPlayerSeek : public CDVDMsg
 {
 public:
-  CDVDMsgPlayerSeek(int time, bool backward, bool flush = true, bool accurate = true, bool restore = true, bool trickplay = false, bool sync = true)
-    : CDVDMsg(PLAYER_SEEK)
-    , m_time(time)
-    , m_backward(backward)
-    , m_flush(flush)
-    , m_accurate(accurate)
-    , m_restore(restore)
-    , m_trickplay(trickplay)
-    , m_sync(sync)
+  struct CMode
+  {
+    int time = 0;
+    bool backward = false;
+    bool flush = true;
+    bool accurate = true;
+    bool sync = true;
+    bool restore = true;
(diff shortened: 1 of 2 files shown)


FernetMenta:
jenkins build this please

popcornmix:
You'll need a fix like: https://github.com/popcornmix/xbmc/commit/459aebfcb6dd7833590917e5908a87d0b971b400
Should I see any change with this PR? Behaviour when holding down right hasn't changed (video plays at normal speed until released).

FernetMenta:
I will pick the omx fix, thanks

> video plays at normal speed until released

if this is the case, something is wrong with mmal or input. a seek flushes demux queue and decoder. does player get seek commands?

popcornmix:
I've added a log message to SeekTimeRelative, but I only see that when I release the key.
The seek bar does move while I'm holding the key down, and it does seek to the expected place when I release.
http://paste.ubuntu.com/23305122/

FernetMenta:
Where did you add the log? Maybe in issue in SeekHandler. Could you add some logging to this chain of functions please. Something must go wrong before it comes to VideoPlayer.

da-anda:
might you guys be talking about different things? Multiple skips steps vs only one with instant seek (aka old behavior before skip steps)?

FernetMenta:
@da-anda use case here is having additive seek disabled by setting delay to zero

popcornmix:
Log was in SeekTimeRelative.
I've found the issue. I had set a single seek step forwards and backwards but not zeroed the skip delay. I had imagined that with only a single seek step then the delay would be ignored (what does it mean in this case?)
I now get ~10 log message per second from SeekTimeRelative. Player shows 0% buffering constantly while holding the seek key down, but you do get one or two video frames displayed about once per second.

Seeking is much slower when delay is set to 0.
With delay=750ms and step=30s I can seek from start to end of 42 minute video in 7 seconds.
With delay=0ms and step=30s I can seek from start to end of 42 minute video in 114 seconds.

I assume that in the first case we get ~10 seeks of 30s per second.
In the second I assume we only update the current seek position when video frame emerge, so ~1 seek of 30s per second.

Log file with delay set to zero.
http://paste.ubuntu.com/23307782/

FernetMenta:
> Seeking is much slower when delay is set to 0.
With delay=750ms and step=30s I can seek from start to end of 42 minute video in 7 seconds.
With delay=0ms and step=30s I can seek from start to end of 42 minute video in 114 seconds.

That's expected behaviour, isn't it? Before being able to do the next jump you have to land first.

> but you do get one or two video frames displayed about once per second.

That's a remaining issue. Every relative seek should result in a displayed video frame.

popcornmix:
> That's expected behaviour, isn't it? Before being able to do the next jump you have to land first.

It's one possible expected behaviour...
Compare with 8x fast forward. The clock moves at 8x real time regardless of how long it is taking the seek/demux/decode/render paths. On a slower platform/network/disk the clock moves at a fixed rate but you get fewer decoder video frames during the process.
The repeated skip could behave the same way. The number of 30s seeks per second depends on the key repeat rate. Skipping from one end of a one hour video could always take the same time, with speed of platform determining the time between video frames actually appearing.

But that is probably a harder system to implement, and isn't a blocker for this PR.

FernetMenta:
if all you want to achieve is getting fast from one end to the other, you should use additive seek or big step forward. not producing a picture after a relative seek makes no sense bacause you don't know where you are.

popcornmix:
No problem - as long as the behaviour is expected I'm happy.

FernetMenta:
jenkins build this please

PR #10675, opened 2016-10-10
Dont load inputstream addons on startup if they are disabled

peak3d:
Dont load inputstream addons on startup if they are disabled
## Description

see title
## Motivation and Context

Faster startup if one or more inputstream addons are dsabled
## How Has This Been Tested?

runtime test, start with disabled inputstream.mpd / enable it / read log
## Types of change

- [?] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/xbmc/addons/InputStream.cpp
+++ b/xbmc/addons/InputStream.cpp
@@ -38,7 +38,8 @@ std::unique_ptr<CInputStream> CInputStream::FromExtension(AddonProps props, cons
   std::string name(ext->plugin->identifier);
   std::unique_ptr<CInputStream> istr(new CInputStream(props, name, listitemprops,
                                                       extensions, protocols));
-  istr->CheckConfig();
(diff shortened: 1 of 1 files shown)


FernetMenta:
I am ok with that but it also tells us that the current addon handling is not appropriate for binary addons. It makes no sense to create an instance of an addon that is disabled.

peak3d:
@FernetMenta correct place?

peak3d:
I can try to check earlier for the state (not inputstream dependent), but I don't have enough background to overview side effects. Let me try one other...

FernetMenta:
When you compare python addons with binary addons you find more differences than things in common. Trying to develop binary addons within the constraints of the current addon system leads to nowhere. We need to ged rid of these constraints and do what's best for binary addons in the current stage.

peak3d:
OK, nothing what can be done in some days - still OK with the PR?

FernetMenta:
+1

peak3d:
jenkins build this please

PR #10676, opened 2016-10-10
[PVR] Recording home screen widget: Fix info dialog (show recording info, not generic video info).

ksooo:
Problem reported internally by @da-anda: "I use "show info" as default select action, and from home screen there is a different "info" view than from PVR -> recordings window. From the default "info" view, there is no "resume point" handling when clicking the play button while form the PVR info view it's working."

Cause of the bug is that directory provider did display the info dialog for videos instead of the special one for pvr recordings. This PR fixes that misbehavior.

@xhaggi okay to go in?


xhaggi (approved):
looks good

ksooo:
jenkins build this please

ksooo:
Travis error is unrelated.

da-anda:
thanks

PR #10679, opened 2016-10-11
[gui] fix slider focus handling

phil65:
PR https://github.com/xbmc/xbmc/pull/10445 intended to change the handling of sliderex controls ("SettingsSlider"), but adjusted the base slidercontrol class instead and therefore also the behaviour of "normal" slider controls. That way we also had to introduce unneeded complexity by adding a keepActive member bool to CGUISliderControl, which basically was a workaround to not break normal slider controls.

This PR changes that so that original behaviour for normal slider controls is restored and only behaviour for sliderex controls gets changed. That way we can also drop the keepActive bool again.
The base slider control is now always "active" (like it always was), while the derived settingsslider control becomes active after havin clicked on it.

This needs proper testing, some parts of guilib can be a minefield :)
@Montellese perhaps for review? or @xhaggi ?
Some testing from @ronie would also be cool.
Should fix an issue reported by @romanvm.

### DIFF
--- a/xbmc/dialogs/GUIDialogSlider.cpp
+++ b/xbmc/dialogs/GUIDialogSlider.cpp
@@ -81,8 +81,6 @@ void CGUIDialogSlider::SetSlider(const std::string &label, float value, float mi
   m_callbackData = callbackData;
   if (slider)
   {
-    slider->SetActive();
-    slider->KeepActive();
     slider->SetType(SLIDER_CONTROL_TYPE_FLOAT);
     slider->SetFloatRange(min, max);
     slider->SetFloatInterval(delta);
--- a/xbmc/guilib/GUISettingsSliderControl.cpp
+++ b/xbmc/guilib/GUISettingsSliderControl.cpp
@@ -19,6 +19,7 @@
  */
 
 #include "GUISettingsSliderControl.h"
+#include "input/Key.h"
 
 CGUISettingsSliderControl::CGUISettingsSliderControl(int parentID, int controlID, float posX, float posY, float width, float height, float sliderWidth, float sliderHeight, const CTextureInfo &textureFocus, const CTextureInfo &textureNoFocus, const CTextureInfo& backGroundTexture, const CTextureInfo& nibTexture, const CTextureInfo& nibTextureFocus, const CLabelInfo &labelInfo, int iType)
     : CGUISliderControl(parentID, controlID, posX, posY, sliderWidth, sliderHeight, backGroundTexture, nibTexture,nibTextureFocus, iType, HORIZONTAL)
@@ -27,6 +28,7 @@ CGUISettingsSliderControl::CGUISettingsSliderControl(int parentID, int controlID
 {
   m_label.SetAlign((labelInfo.align & XBFONT_CENTER_Y) | XBFONT_RIGHT);  
   ControlType = GUICONTROL_SETTINGS_SLIDER;
+  m_active = false;
 }
 
 CGUISettingsSliderControl::~CGUISettingsSliderControl(void)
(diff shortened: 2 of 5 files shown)


mgonzales71:
thank you for fixing the focus issue with the new slider(x) changes!  
will test once Jenkins builds a test or it rolls up into a nightly once committed.

ronie:
> PR #10445 intended to change the handling of sliderex controls ("SettingsSlider")

nope, my intention was to change **both** :-)
https://github.com/xbmc/xbmc/pull/10445#issuecomment-246280378

phil65:
Sorry  but as discussed on slack,  that doesn't make much sense ;-)
A normal slider does not have enough visual states to be either activated and/or focused,  it can only be both at the same time.  @wisler would have to expose sliderex buttons to the addon interfaces if he wants that behavior

ronie:
> @wisler would have to expose sliderex buttons to the addon interfaces if he wants that behavior

makes me wonder, if they can easily be exchanged, do we still have a need for slider controls at all?
are they any use-cases where you can't use a sliderex control?

phil65:
DialogSlider.xml and python add-ons mainly. I also make use of regular slider controls in estuary since they can act as a nib for a progressbar, but in that specific case they dont get focused. (see VideoOSD.xml)

ronie:
i know, but can you use by a sliderex control in those examples instead?

just trying to find out if we can perhaps deprecate the slider control...

phil65:
I dont think so.
- DialogSlider.xml would need a hack to directly activate sliderex
- Perhaps some python devs have valid use cases for this (perhaps @romanvm knows whether someone make use of them via his framework?)
We would also have to expose sliderex to python first, otherwise there is no way to use sliders at all anymore via python
- About that stuff in VideoOSD... didnt try yet if it could be exchanged, but would definitely feel a bit weird since in that use case a label / focustexture arent needed.

romanvm:
Sorry guys, but I have almost zero knowledge of Python skinning so I don't know the difference between slider and sliderx. From Python perspective it really does not matter what is under the hood, if it walks like a slider and quacks like a slider. 😄 

I don't track the usage of my plugins, but from the questions in PyXBMCt forum topic I see that some addon developers use it to create dashboards for various automation addons to control things from Kodi. So I guess some of them use Slider too. But I don't have any solid data.

ronie:
this pr will re-introduce the navigation issues (for slider controls) we were trying to fix,
but the focus issue is more annoying/urgent, so no objection from my end to merge this.
but let's wait a bit for @AchimTuran to respond.

deprecation is perhaps something we can re-evaluate for v18.
another possibility would be to add texturefocus/unfocus textures to slider controls.

phil65:
the navigation issues were almost exclusive to sliderex controls because they get used by our core to fill certain settings list (--> DialogSettingsBase stuff). that was the issue we intended to fix in the beginning.

phil65:
jenkins build this please.

MartijnKaijser:
was this ok to merge?

ronie:
i guess there's no way around it :-)

final ping to @AchimTuran

AchimTuran:
Sorry guys for the late response, but I was very busy during the last days with real life stuff.

In my opinion both sliders should work in the same way. But I'm not a show stopper here because I didn't follow the discussion on slack. Currently adsp.bI quad.filters uses normal sliders which could be replaced by sliderex. But we will have again the navigation issue. 

Is it not possible to use both behaviours and turn one option on with a bool value or XML tag?
Didn't looked into the code bit it should be possible.

ronie:
if you can replace all sliders with sliderex controls, there shouldn't be any navigation issues?

only slider controls will have navigation issues once this pr is merged.

AchimTuran:
After discussing stuff with @ronie I also think it is the best what we could do.
@phil65 feel free to merge it.

MartijnKaijser:
Not sure if related to this change but Coverity created following report just now
http://paste.ubuntu.com/23347252/

PR #10680, opened 2016-10-11
[AE/OsxSink] fix 16bit ne passthrough when there is no dedicated encoded stream

Memphiz:
This restores the pcm passthrough mode for the osx sink when only 32bit float virtual formats are available.

We have multiple ways of passing through on osx
1. Use dedicated passthrough streams - this is the prefered way but not all audio devices have those streams
2. If we have a 16 bit integer stream with 48000 Hz sample rate we try to use this.
   a. We open the physical stream and try to open a virtual stream with the same format so that no conversion is taking place and we get a raw passthrough situation
   b. If we fail to open a virtual stream with 16bit integer 48000 Hz (which is the case if there simply is no suche virtual stream like on my mac mini for example), we use the bitstream option we had before (developed by jmarshall) which reverses the conversion from virtual to physical format which is done in the device/CoreAudio layer to achieve raw passthrough.

This PR adds the 2.b. use case again which was removed in 

https://github.com/xbmc/xbmc/commit/ddc3d29e50105180f4cfeddf462dfa6f934f923f#diff-3dd678432bba9d51a39816409d376717

@FernetMenta you can try this with this patch http://pastebin.com/6MM375AX which basically disables the dedicated encoded streams and leaves you with only integer streams. On my macmini i get the same behavior as the user in the forum - it fails to find a virtual stream which matches the physical format and opens the nullsink ... (because my macmini also only has 32bit float virtual formats).

Jenkins bakes a testbuild i want to provide to the user too.

### DIFF
--- a/xbmc/cores/AudioEngine/Sinks/AESinkDARWINOSX.cpp
+++ b/xbmc/cores/AudioEngine/Sinks/AESinkDARWINOSX.cpp
@@ -251,8 +251,9 @@ bool CAESinkDARWINOSX::Initialize(AEAudioFormat &format, std::string &device)
     }
     else
     {
-      CLog::Log(LOGERROR, "%s, Unable to find suitable stream", __FUNCTION__);
-      return false;
+      CLog::Log(LOGERROR, "%s, Unable to find suitable virtual stream", __FUNCTION__);
+      //return false;
+      numOutputChannelsVirt = 0;
     }
   }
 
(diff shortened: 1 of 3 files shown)


Memphiz:
User confirmed that this fixes the regression here http://forum.kodi.tv/showthread.php?tid=293500&pid=2434785#pid2434785

FernetMenta:
wow, what a hack :)
but since it was already in, it sould be ok

fritsch:
We need something similar for certain intel rigs that only speak 24 bit :-)

Am 12.10.2016 10:14 vorm. schrieb "Rainer Hochecker" <
notifications@github.com>:

Memphiz:
Feel free to steal it for your needs @fritsch ;)

fritsch:
On ALSA ist would be a bit different. We needed to enumerate a pseudo format 16 bit and just align it into the 24bit output. Need to think if I want to do that :-) or directly doing it in kernel level.

Btw. checking your code in detail again, there might be an issue, as you use a float multiplication with an inverse. There is no guarantee that this holds for every input float.

Memphiz:
don't wanna ping him but you would need to discuss this with jmarshall :)

PR #10683, opened 2016-10-11
[android] don't strip binaries in debug builds

wsnipex:
allows us to package debug builds for testing


wsnipex:
jenkins build this please

MartijnKaijser:
@wsnipex builds are now indeed unstripped and log is at debug level

PR #10684, opened 2016-10-11
[video] Support Rockchip's VAAPI hardware decoding

wzyy2:
## Description

Rockchip Linux Team have implement a vaapi backend driver.See wiki below.
http://rockchip.wikidot.com/rockchip-va-driver
http://linux-rockchip.info/mw/index.php?title=VideoAcceleration

This PR intended to make kodi support it. (most changes actually is made for gles-libmali)

I might add too much ifdef in vaapi.c since i'm not very familiar with kodi and not dare to change much....... :-)
## Test

If you have a rockchip sbc(such as firefly, miqi....) , you can follow this guide to install vaapi-ready kenrel and rootfs(http://rockchip.wikidot.com/linux-user-guide) then build kodi from my branch.
I played some demo h264 video and it seems work well.


wzyy2:
uh....The color is not right...

FernetMenta:
you should adapt gles rendering to support vaapi instead of adapting to a specific hardware.

wzyy2:
> you should adapt gles rendering to support vaapi instead of adapting to a specific hardware.
Ok, i will adapt gles rendering and also test it in intel pc with gles.

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererRKVA.cpp
```diff
@@ -0,0 +1,214 @@
+/*
```
FernetMenta: when adapting to gles this file should be renamed to RenderVAAPI-GLES or something

MartijnKaijser:
Closing this as it needs to be implemented in a different way

PR #10685, opened 2016-10-11
[android] update readme for compiling with NDK 12b and SDK21

MartijnKaijser:
Fixes
http://trac.kodi.tv/ticket/16974

### DIFF
--- a/docs/README.android
+++ b/docs/README.android
@@ -13,7 +13,7 @@ TOC
 1. Introduction
 -----------------------------------------------------------------------------
 
-We currently recommend Ubuntu "Trusty Tahr" (14.04) 64Bit. This is what our continuous
+We currently recommend Ubuntu "Xenial Xerus" (16.04) 64Bit. This is what our continuous
 integration system "jenkins" is using.
 Additionally, building from OSX Snow Leopard or Mavericks (others on your own risk ;) ) is working (see 3.1).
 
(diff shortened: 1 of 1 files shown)


phillipao:
Looks like 12b still requires ndk-dir but doesn't recognize --system. So for x86:

./make-standalone-toolchain.sh --ndk-dir=../.. --install-dir=<android-toolchain-x86>/android-21 --platform=android-21 --toolchain=x86-4.9 --arch=x86

docs/README.android
```diff
@@ -133,25 +133,34 @@ common scenario.
...
+      --install-dir=<android-toolchain-arm>/android-21 --platform=android-21 \
+      --toolchain=arm-linux-androideabi-4.9
+
+Building for aarch64 architecture:
+
+    $ cd <android-ndk>
+    $ ls platforms
+    $ cd build/tools
+    $ ./make-standalone-toolchain.sh --ndk-dir=../../ \
+      --install-dir=<android-toolchain-arm>/android-21 --platform=android-21 \
```
fetzerch: should the install-dir for aarch64 be `android-linux-aarch64-4.9`?

fetzerch:
https://github.com/MartijnKaijser/xbmc/blob/9dd3894cfc649ed712edf234dd31e89544053d51/docs/README.android#L91 should be updated as well.

MartijnKaijser:
updated

PR #10686, opened 2016-10-12
[PVR] Fix: activate/deactivate read-only timer

Glenn-1990:
Activate/deactivate buttons are not working for read-only timers.

Reproducible with rule based timers and tvheadend 4.2.

@ksooo


ksooo:
Good catch.

jenkins build this please.

ksooo:
@Glenn-1990 thanks for the fix.

PR #10630, opened 2016-10-05
Controller fixes

garbear:
This PR manages the remaining bugs in the input system from #8807. I'll break out fixes in this PR to another PR as they're ready.

The remaining bugs are:
### Input
- [x] Joysticks not being detected on android startup (#10689)
- [x] Restart needed to enable/disable peripheral add-ons (#10690, #10734)
- [x] Missing lock in CAddonButtonMap (#10691)
- [x] Broken Shield controller if add-on isn't present (#10695)
- [x] Problems with accelerometers ([report and linux solution](http://forum.kodi.tv/showthread.php?tid=211138&pid=2207458#pid2207458)) (#10725, #10910)
- [x] PS4 controllers skip because button presses are sent when trigger is partially pressed (#10910)
- [x] Deadlock when updating peripheral add-ons (#10726)
- [x] Button map doesn't refresh when the user hits reset (#10735)
- [x] Possible crash when controllers are unplugged (#10740)
- [x] Incompatible with Unified Remote server ([report](http://forum.kodi.tv/showthread.php?tid=164725&pid=2119476#pid2119476), [support thread](http://forum.kodi.tv/showthread.php?tid=295961))
- [x] Constant joystick discovery on Windows ([report](http://forum.kodi.tv/showthread.php?tid=295548&pid=2449281#pid2449281), [peripheral.joystick#62](https://github.com/xbmc/peripheral.joystick/pull/62))
- [x] 100% cpu when scanning for joystick events after 6 days (#10921)
- [x] User should be able to open controller dialog with controllers unplugged (#10923)
- [x] Use 300ms rumble on weak motor for notifications (#10929)
- [ ] Incompatible with xpadder ([report](http://forum.kodi.tv/showthread.php?tid=173361&pid=2393609#pid2393609))
- [ ] All controllers vibrate whenever a new one is plugged in ([report](http://forum.kodi.tv/showthread.php?tid=173361&pid=2423447#pid2423447))
- [ ] Launching a game by holding A causes all hell to break loose ([report](http://forum.kodi.tv/showthread.php?tid=173361&pid=2413498#pid2413498))
- [ ] Missing a `mod="longpress"`  parameter for [joystick.xml](https://github.com/xbmc/xbmc/blob/master/system/keymaps/joystick.xml) keymap (should fix above bug)
- [ ] If Kodi crashes, the controller could vibrate without stopping ([report](http://forum.kodi.tv/showthread.php?tid=173361&pid=2430664#pid2430664))
[...]

PR #10689, opened 2016-10-12
Fix joysticks not being detected on android startup

garbear:
Proper fix for #9823

This fixes an error I had when testing on android. The error in the log was

```
WARNING: CPeripheralBusAndroid: ignoring input event for unknown input device with ID 1
```

This occurred because even though the joystick was discovered, Kodi wasn't doing an initial scan on the android bus (an initial scan only occurs if the bus requires Kodi to poll it). This fixes the bug by doing a scan when the bus is initialized.

Broken out from #10630

### DIFF
--- a/xbmc/peripherals/bus/android/PeripheralBusAndroid.cpp
+++ b/xbmc/peripherals/bus/android/PeripheralBusAndroid.cpp
@@ -108,6 +108,12 @@ bool CPeripheralBusAndroid::InitializeProperties(CPeripheral* peripheral)
   return true;
 }
 
+void CPeripheralBusAndroid::Initialise(void)
+{
+  CPeripheralBus::Initialise();
+  TriggerDeviceScan();
+}
+
 void CPeripheralBusAndroid::ProcessEvents()
 {
   std::vector<ADDON::PeripheralEvent> events;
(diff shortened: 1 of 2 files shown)


xbmc/peripherals/bus/android/PeripheralBusAndroid.cpp
```diff
@@ -108,6 +108,12 @@ bool CPeripheralBusAndroid::InitializeProperties(CPeripheral* peripheral)
   return true;
 }
 
+void CPeripheralBusAndroid::Initialise(void)
```
peak3d: don't want to be picky, but should this better Initiali**z**e ??

garbear: not if it was originally written by a dutch person

garbear: it's `Initialise()` in the base class so it needs to be like this in the overriding class too. You can PR a separate change to rename the function if you'd like.

peak3d: Ah, oh, have overseen the override, sry :-)

garbear: was referring to the author of the base class, @opdenkamp

MartijnKaijser:
Jenkins build this please

PR #10690, opened 2016-10-12
[addons] Fix restart needed to enable/disable peripheral add-ons

garbear:
Previously, if the joystick add-on was disabled, it would continue to run in the background. This fixes that.

Should we warn the user about disabling this add-on if controllers are connected, and then reenable the add-on if they decline? We do this with unknown sources, but the settings system makes this easy. The addon manager doesn't have a clean way to prevent a disable operation from happening, so this might require some heavier modifications.

Broken out from #10630

### DIFF
--- a/xbmc/addons/Addon.cpp
+++ b/xbmc/addons/Addon.cpp
@@ -345,7 +345,8 @@ void OnEnabled(const std::string& id)
   // If the addon is a special, call enabled handler
   AddonPtr addon;
   if (CAddonMgr::GetInstance().GetAddon(id, addon, ADDON_PVRDLL) ||
-      CAddonMgr::GetInstance().GetAddon(id, addon, ADDON_ADSPDLL))
+      CAddonMgr::GetInstance().GetAddon(id, addon, ADDON_ADSPDLL) ||
+      CAddonMgr::GetInstance().GetAddon(id, addon, ADDON_PERIPHERALDLL))
     return addon->OnEnabled();
 
   if (CAddonMgr::GetInstance().ServicesHasStarted())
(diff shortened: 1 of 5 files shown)


garbear:
@MartijnKaijser can this go in, even if it means that users with only joysticks can disable their ability to control Kodi? This is the current behavior, it just doesn't take effect immediately due to this bug

MartijnKaijser:
jenkins build and merge

tamland:
@garbear I have told you before not to use these addon hooks. They're broken and I have almost removed everything using so don't start adding new things to clean up now.

tamland:
Also, `IsInUse` can be used to prevent disabling (as already done by several other addon types)

MartijnKaijser:
@tamland if you thing it's better to revert and done correctly go ahead.

garbear:
@tamland - sorry, should have pinged you for this. Are these hooks replaced by anything? I think PVR uses a similar strategy, so when that's cleaned up I can adapt peripheral add-ons accordingly

garbear:
And we don't want to prevent disabling, only warn the user and prevent disabling if they recline. This can't be done currently, so maybe we need a pre-disable hook that can return false to block the disabling?

tamland:
The preferred way is to reload async via CAddonMgr::Events() so the entire AddonMgr (and everything that use it) isn't blocked while addons are being installed/enabled.

As for disabling, IsInUse can probably be refactored to do that as most of the currently implementations seems to just check a setting. Which is the same as 'active' available here: https://github.com/xbmc/xbmc/blob/eee0fd7f25abc5577a92cf0e9ac80230db997515/xbmc/addons/AddonSystemSettings.h#L43

PR #10691, opened 2016-10-12
Fix missing lock in CAddonButtonMap

garbear:
This fixes a race condition I hit while pausing/resuming during debugging. peripheral.joystick was doing an extra refresh of the button map, which was fixed in https://github.com/kodi-game/peripheral.joystick/commit/f902039. This is the corresponding change needed in core to prevent other add-ons from crashing kodi.

Broken out from #10630


garbear:
jenkins build this please

PR #10695, opened 2016-10-12
Fix controller input on the Shield if peripheral.joystick is not present

garbear:
The shield controller sends keypresses when it hasn't been scanned by Kodi, so if no add-on is present to translate joystick input, skip the controller so that it continues sending somewhat-functional keypresses.

Broken out from #10630

### DIFF
--- a/xbmc/peripherals/bus/PeripheralBus.cpp
+++ b/xbmc/peripherals/bus/PeripheralBus.cpp
@@ -43,6 +43,24 @@ CPeripheralBus::CPeripheralBus(const std::string &threadname, CPeripherals *mana
 {
 }
 
+bool CPeripheralBus::InitializeProperties(CPeripheral* peripheral)
+{
+  if (peripheral == nullptr)
+    return false;
+
+  if (peripheral->Type() == PERIPHERAL_JOYSTICK)
+  {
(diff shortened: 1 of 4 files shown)


MartijnKaijser:
jenkins build and merge

xbmc/peripherals/bus/PeripheralBus.cpp
```diff
@@ -43,6 +43,24 @@ CPeripheralBus::CPeripheralBus(const std::string &threadname, CPeripherals *mana
...
+  if (peripheral == nullptr)
+    return false;
+
+  if (peripheral->Type() == PERIPHERAL_JOYSTICK)
+  {
+    // Ensure an add-on is present to translate input
+    if (!g_peripherals.GetInstance().GetAddonWithButtonMap(peripheral))
+    {
+      CLog::Log(LOGWARNING, "Button mapping add-on not present for %s (%s), skipping", peripheral->Location().c_str(), peripheral->DeviceName().c_str(), peripheral->VendorIdAsString(), peripheral->ProductIdAsString());
+      return false;
```
popcornmix: You have two ``%s`` format strings, but four arguments...

PR #10696, opened 2016-10-12
VideoPlayer: make sure first frame of a new stream is displayed witho…

FernetMenta:
make sure first frame of a new stream is displayed without delay
signal started when frame is picked for render

this makes continuous seek steps look smooth

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayerVideo.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayerVideo.cpp
@@ -910,7 +910,7 @@ int CVideoPlayerVideo::OutputPicture(const DVDVideoPicture* src, double pts)
     return EOS_DROPPED;
   }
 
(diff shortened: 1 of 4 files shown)


FernetMenta:
@popcornmix anything to adapt for omxplayer?

popcornmix:
Simple compile fix for omxplayer:
https://github.com/popcornmix/xbmc/commit/d1c1e64f47c75529d182128427fa2ba42b803af1
Not sure we can do much better with a bypass renderer.

FernetMenta:
thanks @popcornmix 

jenkins build this please

stefansaraev:
fyi, this also fixed my gles render crash/deadlock we discussed few months ago in slack. thanks!

FernetMenta:
thanks for feedback. actually I had hoped to fix this earlier by some other changes. good to hear that it is fixed finally.

PR #10697, opened 2016-10-12
VideoPlayer: fix CDVDMessageQueue::WaitUntilEmpty

FernetMenta:
see title, this has been broken since ever but did not show because the entire sync did not work


FernetMenta:
jenkins build this please

PR #10698, opened 2016-10-13
[gui] add listitem.appearances for actors / directors

phil65:
see title.


BigNoid:
Nice one!
schermafbeelding 2016-10-13 om 07 50 53

One thing I've noticed is that after a skin reload appearances returns empty.

MartijnKaijser:
Jenkins build this please

braz96:
Now if only we can sort by number of appearances. ;)

phil65:
that already works.

PR #10699, opened 2016-10-13
[python] Add xbmcgui.DialogBusy()

phil65:
## Description

Adds a new dialog class to the python interface.
## Motivation and Context

Lot of python scripters are making use of our busy dialog although there is no proper interface to do that. So their solution is to use executebuiltin("ActivateWindow(busydialog)") and executebuiltin("Dialog.Close(busydialog)") in order to open / close it, which feels very workaround-ish.
Because of that, I added busydialog handling to our xbmcgui class, with a similar interface as our progress dialog.
When doin this I also found out that our busydialog core code supports a progress bar. That one can be used via python as well now, and I also added one to estuary in a separate commit.
## How Has This Been Tested?

With some simple test scripts

screenshot041
## Types of change
## Checklist:

### DIFF
--- a/xbmc/interfaces/legacy/Dialog.cpp
+++ b/xbmc/interfaces/legacy/Dialog.cpp
@@ -512,6 +512,61 @@ namespace XBMCAddon
       return dlg->IsCanceled();
     }
 
+    DialogBusy::~DialogBusy() { XBMC_TRACE; deallocating(); }
+
+    void DialogBusy::deallocating()
+    {
+      XBMC_TRACE;
+
+      if (dlg && open)
+      {
+        DelayedCallGuard dg;
+        dlg->Close();
+      }
+    }
+
+    void DialogBusy::create()
+    {
+      DelayedCallGuard dcguard(languageHook);
+      CGUIDialogBusy* pDialog = static_cast<CGUIDialogBusy*>(g_windowManager.GetWindow(WINDOW_DIALOG_BUSY));
+
+      if (pDialog == nullptr)
+        throw WindowException("Error: Window is NULL, this is not possible :-)");
+
+      dlg = pDialog;
+      open = true;
+
+      pDialog->Open();
+    }
+
+    void DialogBusy::update(int percent) const
+    {
+      DelayedCallGuard dcguard(languageHook);
+
+      if (dlg == nullptr)
+        throw WindowException("Dialog not created.");
+
+      if (percent >= -1 && percent <= 100)
+        dlg->SetProgress(percent);
+
+    }
+
+    void DialogBusy::close()
+    {
+      DelayedCallGuard dcguard(languageHook);
+      if (dlg == nullptr)
+        throw WindowException("Dialog not created.");
+      dlg->Close();
+      open = false;
+    }
+
+    bool DialogBusy::iscanceled() const
+    {
+      if (dlg == nullptr)
+        throw WindowException("Dialog not created.");
+      return dlg->IsCanceled();
+    }
+
     DialogProgressBG::~DialogProgressBG() { XBMC_TRACE; deallocating(); }
 
     void DialogProgressBG::deallocating()
--- a/xbmc/interfaces/legacy/Dialog.h
+++ b/xbmc/interfaces/legacy/Dialog.h
@@ -29,6 +29,7 @@
 #include "ListItem.h"
 #include "dialogs/GUIDialogProgress.h"
 #include "dialogs/GUIDialogExtendedProgressBar.h"
+#include "dialogs/GUIDialogBusy.h"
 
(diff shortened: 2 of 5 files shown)


MartijnKaijser:
jenkins build this please

phil65:
@tamland for review, please.

xbmc/interfaces/legacy/Dialog.cpp
```diff
@@ -491,6 +491,62 @@ namespace XBMCAddon
...
+    void DialogBusy::create()
+    {
+      DelayedCallGuard dcguard(languageHook);
+      CGUIDialogBusy* pDialog = static_cast<CGUIDialogBusy*>(g_windowManager.GetWindow(WINDOW_DIALOG_BUSY));
+
+      if (pDialog == nullptr)
+        throw WindowException("Error: Window is NULL, this is not possible :-)");
+
+      dlg = pDialog;
+      open = true;
```
tamland: this open flag doesn't seem to do anything..

xbmc/interfaces/legacy/Dialog.cpp
```diff
@@ -491,6 +491,62 @@ namespace XBMCAddon
...
+
+      pDialog->Open();
+    }
+
+    void DialogBusy::update(int percent) const
+    {
+      DelayedCallGuard dcguard(languageHook);
+      auto pDialog = dlg;
```
tamland: does nothing

xbmc/interfaces/legacy/Dialog.h
```diff
@@ -679,6 +680,131 @@ namespace XBMCAddon
...
+      ///
+      /// \ingroup python_DialogBusy
+      /// \python_func{ xbmcgui.DialogBusy().update(percent) }
+      ///------------------------------------------------------------------------
+      ///
+      /// Updates the busy dialog.
+      ///
+      /// @param percent        integer - percent complete. (0:100)
+      ///
+      /// @note If percent == 0, the progressbar will be hidden.
```
tamland: Sounds like a bug. It should be possible to have a progress bar at 0%

xbmc/interfaces/legacy/Dialog.h
```diff
@@ -679,6 +680,131 @@ namespace XBMCAddon
...
+      ///
+      ///
+      ///------------------------------------------------------------------------
+      ///
+      /// **Example:**
+      /// ~~~~~~~~~~~~~{.py}
+      /// ..
+      /// dialog.close()
+      /// ..
+      /// ~~~~~~~~~~~~~
```
tamland: I think you can drop all of these "examples". They are pretty useless!:)

xbmc/interfaces/legacy/Dialog.h
phil65: Yup, I just did it in same way as progress dialog. I agree that it makes sense to change this behaviour for both.

xbmc/interfaces/legacy/Dialog.h
phil65: Ok, will only keep the one for initialization then. ( .create() )

phil65:
@MartijnKaijser Should I add the @python_v17 stuff just to the class description or to every single method of the class?

xbmc/interfaces/legacy/Dialog.h
```diff
@@ -744,14 +746,7 @@ namespace XBMCAddon
...
       ///
-      /// **Example:**
-      /// ~~~~~~~~~~~~~{.py}
-      /// ..
-      /// dialog.update(25)
-      /// ..
-      /// ~~~~~~~~~~~~~
-      ///
-      update(...);
+       update(...);
```
garbear: whitespace oops

MartijnKaijser:
Each one please as that's clearest

MartijnKaijser:
@tamland ping for review if you have a moment

tamland:
Need to update the documentation on the percent parameter, otherwise it looks good.

phil65:
updated.

MartijnKaijser:
Jenkins build this please

FernetMenta:
@phil65 calling busy dialog from python is nonsense. the busy dialog is made to keep the GUI going while the calling thread is blocked.

calling dialogBusy from python leads to having multiple busy dialogs open at the same time:
https://trac.kodi.tv/ticket/17896

further: opening a modal dialog via a message is wrong anyway. leads to having the busy dialog open but obscured by some other dialog.

FernetMenta:
I will deprecte this and log a message when this is called.

FernetMenta:
> Please also make sure to block Dialog.Close(busydialog) and ActivateWindow(busydialog) then.

already had ActivateWindow(busydialog), added closedialog now

PR #10700, opened 2016-10-13
smart playlists: fix drill down from artists lists that have role based rules

DaveTBlake:
If an "artists" type smart playlist has role rules e.g. "artist is composer", then clicking on a composer from such an artist list can result in nothing being displayed despite the fact thay they do compose some songs. This is because the role rule is not passed down to the subsequent list of albums or songs.

This is a minimal change to ensure that such incorrect blank lists are avoided.

An attempt to solve the more general problems caused by the rules from smart playlists not being passed on to the next level when an item in a list is clicked is made in #10098 . But it is probably now too late in the release cycle for such a design change, and I think there is more to be done. Hence this fix is just a stop gap to provide at least minimal functionality.

### DIFF
--- a/xbmc/music/MusicDatabase.cpp
+++ b/xbmc/music/MusicDatabase.cpp
@@ -6284,9 +6284,39 @@ bool CMusicDatabase::GetFilter(CDbUrl &musicUrl, Filter &filter, SortDescription
   std::string type = musicUrl.GetType();
   const CUrlOptions::UrlOptions& options = musicUrl.GetOptions();
 
+  // Check for playlist rules first, they may contain role criteria
+  bool hasRoleRules = false;
+  auto option = options.find("xsp");
+  if (option != options.end())
+  {
+    CSmartPlaylist xsp;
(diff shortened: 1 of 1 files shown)


DaveTBlake:
This just moves the xsp processing to the start of CMusicDatabase::GetFilter from the end so that artist role rules can be taken into account.

Any comment @Montellese as smart playlist involved?

DaveTBlake:
This just needs a sanity check, anyone?

The kind of smart playlist that would list artists, but then nothing beneath when you click on one is like this for composers. You need some music files in your library that have composer tagged, and the composer not an album or song artist.
```
<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
<smartplaylist type="artists">
<name>Role Composer</name>   
<match>all</match>
<rule field="role" operator="is">   
<value>Composer</value>
</rule>
</smartplaylist>
```

DaveTBlake:
Any objections? I would like to get ths merged and composer etc. smart playlists working, rather than come up blank.

DaveTBlake:
jenkins build this please

PR #10704, opened 2016-10-14
[Texture][FIX] Prevent an access violation in CFFmpegImage::DecodeFra…

ace20022:
…me triggered by a failed alloc due to insufficient free memory.

This fixes a crash reported at http://forum.kodi.tv/showthread.php?tid=293557
Kodi is still quite unresponsive when opening folder with many huge pictures.

@fritsch for review.


fritsch:
Change is fine, thx much.

What do you think: Do we want to get rid of the _aligned_malloc also? We use av_malloc in other place, too?

fritsch:
jenkins build this please

ace20022:
If av_malloc is at least as good, then I can make the change tomorrow

fritsch:
We can do that separately. This fix here is good to go on it own. We can also discuss the pros and cons on slack before.

fritsch:
jenkins build and merge

PR #10705, opened 2016-10-14
VideoPlayer: fix seek beyond eof for mp4

FernetMenta:
- disable ffmpeg's generic seek for remote files because this is very very slow
- terminate playback if seek fails because it was beyond eof


FernetMenta:
jenkins build this please

peak3d:
Just made a quick test - looks good! Thx!

PR #10706, opened 2016-10-14
VideoPlayer: drop unmaintained code in favour of fixes

FernetMenta:
this code is obsolete, we have additive seek

we don't need seek functions all over the place.

### DIFF
--- a/xbmc/video/windows/GUIWindowFullScreen.cpp
+++ b/xbmc/video/windows/GUIWindowFullScreen.cpp
@@ -76,10 +76,6 @@ static CLinuxResourceCounter m_resourceCounter;
 CGUIWindowFullScreen::CGUIWindowFullScreen(void)
     : CGUIWindow(WINDOW_FULLSCREEN_VIDEO, "VideoFullScreen.xml")
 {
-  m_timeCodeStamp[0] = 0;
-  m_timeCodePosition = 0;
-  m_timeCodeShow = false;
-  m_timeCodeTimeout = 0;
   m_bShowViewModeInfo = false;
   m_dwShowViewModeTimeout = 0;
   m_bShowCurrentTime = false;
@@ -108,17 +104,6 @@ CGUIWindowFullScreen::~CGUIWindowFullScreen(void)
 
 bool CGUIWindowFullScreen::OnAction(const CAction &action)
 {
-  if (m_timeCodePosition > 0 && action.GetButtonCode())
-  { // check whether we have a mapping in our virtual videotimeseek "window" and have a select action
-    CKey key(action.GetButtonCode());
-    CAction timeSeek = CButtonTranslator::GetInstance().GetAction(WINDOW_VIDEO_TIME_SEEK, key, false);
-    if (timeSeek.GetID() == ACTION_SELECT_ITEM)
-    {
-      SeekToTimeCodeStamp(SEEK_ABSOLUTE);
-      return true;
-    }
-  }
-
   if (CSettings::GetInstance().GetBool(CSettings::SETTING_PVRPLAYBACK_CONFIRMCHANNELSWITCH) &&
       g_infoManager.IsPlayerChannelPreviewActive() &&
       (action.GetID() == ACTION_SELECT_ITEM || CButtonTranslator::GetInstance().GetGlobalAction(action.GetButtonCode()).GetID() == ACTION_SELECT_ITEM))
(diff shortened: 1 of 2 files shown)


FernetMenta:
I consider this a fix because this is pre-requisite for fixing a bug related to seek.

FernetMenta:
jenkins build this please

FernetMenta:
any objections?

MartijnKaijser:
nope

afedchin:
Please correct me if I'm wrong, now there is no a way to seek to specific time with digits on remote/keyboard?
hehe.. already merged while I'm writing comment

FernetMenta:
If this was the function of this code, yes. If this is still a desired feature, it has to be implemented in CSeekHandler. Otherwise the code is not maintainable.

jjd-uk:
After answering http://forum.kodi.tv/showthread.php?tid=294731 it's just been pointed out to me that this PR has caused http://kodi.wiki/view/skip_steps#Time_skipping to no longer work, seems there were users still using it as it allowed to jump to a precise moment in time.

FernetMenta:
I didn't even know what this code did when I created this pr. I was fixing a bug related to seeking and noticed that we had seek handler and this very old code, obviously not used by many and not migrated to new seek handler.
Sometimes we have to leave things behind that are not maintained.

MilhouseVH:
Why "obviously not used by many"? I miss this feature, it's very useful while testing. I know other users on the forum have noticed it is now missing, and of course they're just the users testing nightlies. While it may have been old code, it still worked and shouldn't have been removed without plans to migrate it before the release of Kodi 17 (IMHO).

jjd-uk:
Perhaps it wad old untouched code because it simply worked and since this in not in our Beta!s yet then yes this removal would have gone unnoticed. I know there's certainly one than that one user who will miss this.

jjd-uk:
Btw your PR title was not at all helpful as I completely missed this removal of seek functionality.

FernetMenta:
It was unmaintained code is it did harm fixing bugs in maintained code. Feel free to reimplement or suggest users to stay with v16 if they think this was the most important feature ever. 
You can't expect to me to migrate unmaintained and decayed code that was left forgotten when seekhandler and additive seek was implemented, do you? Do you want to to revert all my recent fixed and get this old code back?

ronie:
not wanting to be a part of this discussion, but i'm seeing a trend lately where pr's are being submitted and are merged within a few hours without any code review by a secondary member.

perhaps we can improve this a bit?
no idea who would qualify to review video related code... and i wouldn't be surprised if there is no one due to lack of team devs.
but we can at least keep pr's open for a few days, so everyone has a change to leave feedback i think?

jjd-uk:
@xhaggi was the direct seeking functionality deliberately left out of seekhandler? or was it simple missed?

FernetMenta:
please move further discussion to internal forum.

xhaggi:
We should revert this PR because it removes a feature in a stage where we shouldn't. I left this code untouched while working on the adaptive seeking stuff because it was unrelated to it. I agree with @FernetMenta that we need to move it over to our seek handler, but we should not remove it without an alternative implementation IMO.

xhaggi:
BTW if we decide to drop support for it we should do it for v18 instead.

FernetMenta:
I can't think of much contributions to videoplayer recently apart from myself. You can either support my tansformation of videoplayer or try to tell my what I should do or not. Decide for the latter and you are on your own on this journey.

xhaggi:
Don't get me wrong I don't want to tell you anything. You are right. You are the only one currently maintaining this code part but we should decide together or better respect our policies while in beta stage, or don't we? If you decide for us it's also fine by me because I don't want to open an endless discussion about a feature I don't use very often by myself.

MilhouseVH:
Cut whatever features you like, no point discussing it really, is there?

fritsch:
This (in general) was discussed and agreed at DevCon. Sorry guys. Everyone is free to reimplement a feature. Remember the architecture presentation fernetmenta gave at devcon, the broken house image. Sometimes the only way forward is to tear down a wall. Oh, it hit your feature this time?

So then add it back in a maintainable way but blaming the cleaning up guy for throwing away things laying on the floor after a party.

Again: after the seek fixes, an improvement, someone needs to fix the old crap.

xhaggi:
>So then add it back in a maintainable way but blaming the cleaning up guy for throwing away things laying on the floor after a party is just unfair.

We only want to know why he dropped a feature while fixing an issue, nothing else. No one is blaming him. For my part, I thought we should not drop a feature while in beta stage. I just wanted to clarify that and got that. 

>You can either support my tansformation of videoplayer or try to tell my what I should do or not. Decide for the latter and you are on your own on this journey.

Take it or leave it 😯

PR #10707, opened 2016-10-15
SeekHandler: fix state InProgress

FernetMenta:
Fixes jumping playback position when using additive seek


FernetMenta:
jenkins build this please

PR #10708, opened 2016-10-15
VideoPlayer: ffmpeg - use codec aspect ratio for stereo modes

FernetMenta:
fix http://trac.kodi.tv/ticket/16876


fritsch:
jenkins build this please

Something I don't like about the ffmpeg av_dict api is, that one always has to look into the sourcecode if this ptr needs to be freed by one self at the end (which is not the case).

PR #10711, opened 2016-10-15
[addons] improve safety of installing/uninstalling addon files

tamland:
This change should ensure that addons are either completely installed to the file system, or not installed if something fails, not leaving any partially installed/removed files.

Unlike the current procedure which just copies files one-by-one over the existing folder _praying_ everything goes ok.

Also includes some fixes to RemoveRecursive.

cc @Paxxi

### DIFF
--- a/Kodi.xcodeproj/project.pbxproj
+++ b/Kodi.xcodeproj/project.pbxproj
@@ -840,6 +840,8 @@
 		DF1D2DF11B6E85EE002BB9DB /* XbtFile.cpp in Sources */ = {isa = PBXBuildFile; fileRef = DF1D2DE91B6E85EE002BB9DB /* XbtFile.cpp */; };
 		DF1D2DF31B6E85EE002BB9DB /* XbtManager.cpp in Sources */ = {isa = PBXBuildFile; fileRef = DF1D2DEB1B6E85EE002BB9DB /* XbtManager.cpp */; };
 		DF1D2DF41B6E85EE002BB9DB /* XbtManager.cpp in Sources */ = {isa = PBXBuildFile; fileRef = DF1D2DEB1B6E85EE002BB9DB /* XbtManager.cpp */; };
+		DF209C121DB2A0C300515D1F /* FilesystemInstaller.cpp in Sources */ = {isa = PBXBuildFile; fileRef = DF209C101DB2A0C200515D1F /* FilesystemInstaller.cpp */; };
+		DF209C131DB2A0D400515D1F /* FilesystemInstaller.cpp in Sources */ = {isa = PBXBuildFile; fileRef = DF209C101DB2A0C200515D1F /* FilesystemInstaller.cpp */; };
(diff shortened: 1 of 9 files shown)


MartijnKaijser:
@Memphiz ping as well since you might want xcode still added right?

Memphiz:
thx @MartijnKaijser 

can't do a PR to tamlands repo - so @tamland please pick https://github.com/Memphiz/xbmc/commit/4eae05e7cace139e57ff18e6998606c0bd2063af

wsnipex:
the new file needs adding to xbmc/addons/CMakeLists.txt

tamland:
already are..

tamland:
jenkins build this please

MartijnKaijser:
@tamland it fails on Windows

tamland:
jenkins build this please

Paxxi (approved):
Sorry for the delay, changes look solid.

tamland:
did anyone test this on windows..?

MartijnKaijser:
@tamland nope not yet. Last i wanted to try it didn't build yet. Will build one now and let you know

tamland:
Can any windows devs help with this? It seems like unpacking the archive suddenly fails (and silently, without returning error) with this weird message:
>XFILE::CDirectory::Create - Error creating C:\
>ERROR: Previous line repeats 264 times.

and `XFILE::CFile::Rename` also fails for a directory that should have been checked to exist. Log: http://paste.ubuntu.com/23359807/

Paxxi:
@tamland I'll look at it now

Paxxi:
The reason for the 260 error messages is this bit
https://github.com/xbmc/xbmc/blob/master/xbmc/filesystem/File.cpp#L116
For every file in the addon it will try to create the directory C:, c:\users, c:\users\per, c:\users\per\appdata, c:\users\per\appdata\local, c:\users\per\appdata\local\temp, c:\users\per\appdata\local\temp\guid

This should at a minimum be wrapped in an `if (CDirectory::Exists())` but it really should be moved into the CDirectory::Create implementations and probably reversed, try to create the full path, if that fails chop of a bit and try again, it's most likely that we're only missing one bit and not the whole path.

In my case installing from repo works but will spam the log.

Regarding the rename error I assume it's this bit

tamland:
> That's a directory and it's expected that CFile::Rename fails on it.

Sure about that? That's not how it's implemented on other platforms. And there is no CDirectory::Rename. So I think it's a windows bug. It calls the [correct function](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365240(v=vs.85).aspx) for renaming directories but fail for some reason

Paxxi:
You're right about the CFile::Rename, it seems I spoke too soon. It should support moving a directory on windows as well, as long as it's not moving it across drive/filesystem boundries.

I'm taking another look at it

Paxxi:
I don't seem to be able to reproduce the issue with the rename failing.

One theory could be that something has an open file handle to a file in it, that will prevent a move/rename.

tamland:
Ok, seem like there was just a misunderstand on my part and that the change works as indented on windows. Thanks for the help anyway @Paxxi :)
The only issue is with updating the skin that is in-use, which is not related to this pr. Same with the 'creating C:\': this looks like an old issue. So if no objections I'll go ahead with this.

Paxxi:
Wait until tonight, @martijn mentioned a way to repro yesterday so I want to have a look at if we can fix updating the active skin as well

Paxxi:
Okay I tested this and the file causing the issue is Textures.xbt that we keep open. I'm not familiar enough with our texture loading to solve this and poking at it may introduce some unpleasant regressions so lets go with this as is for now.

stefansaraev:
this is quite intrusive change.

http://forum.kodi.tv/showthread.php?tid=269815&pid=2443873#pid2443873

it breaks installing/upgrading addons at least on LibreELEC.

popcornmix:
Also fails on Pi LibreELEC
http://forum.kodi.tv/showthread.php?tid=269814&pid=2445125#pid2445125

tamland:
#10794

PR #10713, opened 2016-10-16
configure: fix aarch64 builds

koenkooi:
## Description

Just like x86\* aarch64 has platforms that can be full openGL and/or full openGLES. In the case of mesa/freedreno on Qualcomm SoCs, building against GLES will fail, but full GL works.

So stop hardcoding GL/GLES/Wayland and let honour the --enable-gl flags passed to configure.
## Motivation and Context

After running into Keith and @MartijnKaijser last week I wanted to switch from Jarvis to Krypton on my Dragonboard-410c to test the ffmpeg-v4l2 patches. It didn't build, so I tracked the failure down to the point where I replaced the aarch64 patch I wrote with the upstream one.
## How Has This Been Tested?

Configure now does the right thing now:

```
------------------------
  Kodi Configuration:
------------------------
  Kodi Version: 17.0-BETA4
  git Rev.: 20161015-2f143a8
  Debugging:    Yes
  Profiling:    No
  Optimization: Yes
  SWIG Available:   Yes
  JRE Available:    Yes
  Doxygen Available:    No
  Crosscomp.:   Yes
  target ARCH:  aarch64
  target CPU:   
  OpenGL:   Yes
```

Compilation succeeds, and it runs, see screenshot below.

img_7394
## Types of change

I'm not 100% sure this is non-breaking change, but Libreelec always sets the appropriate --enable-opengl/gles, so they won't see breakage.
## Checklist:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -626,9 +626,13 @@ case $host in
      ARCH="aarch64"
      use_arch="aarch64"
      use_neon=yes
-     use_gles=yes
-     use_gl=no
-     use_wayland=no
+     # Check whether --enable-gl was given.
+     if test "${enable_gl+set}" = set; then :
+       enableval=$enable_gl; use_gl=$enableval
+     else
+       use_gl=no
+       use_gles=yes
+     fi
      USE_STATIC_FFMPEG=1
      ;;
   arm*-*linux-android*)


configure.ac
```diff
@@ -626,9 +626,6 @@ case $host in
      ARCH="aarch64"
      use_arch="aarch64"
      use_neon=yes
-     use_gles=yes
```
wsnipex: can you please add a test that keeps this as default, if the user didn't specify --enable-gles or --enable-gl?

wsnipex:
thanks! Please note that we are moving to a cmake based build system and will deprecate autotools soon, so would be great if you could test https://github.com/xbmc/xbmc/tree/master/project/cmake as well.

wsnipex (changes requested):

koenkooi:
Updated the branch, a check has been added to see if --enable-gl has been set and if it isn't, default to GLES.

configure.ac
```diff
@@ -626,9 +626,12 @@ case $host in
...
-     use_gl=no
-     use_wayland=no
+     # Check whether --enable-gl was given.
+     if test "${enable_gl+set}" = set; then :
+       enableval=$enable_gl; use_gl=$enableval
+     else
+       use_gles=yes
```
wsnipex: we also need use_gl=no here, because it defaults to "yes"

koenkooi:
New version, now with use_gl=no

wsnipex:
jenkins build this please

wsnipex:
thank you

PR #10714, opened 2016-10-16
AESinkAUDIOTrack: Don't false alarm with m_offset

fritsch:
Fixes some false alarm. Tracking of m_offset is not optimal, but it's currently used when filling up the silence buffer with RAW PT.

After v17, this complete sink will be renamed to Legacy and a new one is written with the Android N functionality only. So all these hacks will die in fire then :-)


MartijnKaijser:
Jenkins build this please

MartijnKaijser:
tested and all seems fine

PR #10718, opened 2016-10-16
Fix c/p error in stereoscopic inversion for block format

garbear:
Just a quick c/p error I came across


da-anda:
good catch, thanks

PR #10719, opened 2016-10-16
[estuary] grey text on focused item == unreadable

ronie:
grey text against a blue background is hardly readable, use white instead.

@phil65


phil65:
Thx!

PR #10721, opened 2016-10-17
[android] strip binaries in debug builds 

MartijnKaijser:
[strip binaries in debug builds as well as they do not bring any extra information in logcat

@wsnipex @koying


wsnipex:
why not build release instead?

koying:
Well, release means no debug info at all (no -g), which is another story.
You then get no symbol at all in logcat (IIRC), only lib name + displacement.

MartijnKaijser:
jenkins build and merge

PR #10723, opened 2016-10-17
[PVR] Fix search window update

ksooo:
The problem:
- open pvr search window, click "Search...", enter some search string, search
  => search dialog closes, progress dialog appears, progress dialog closes, search window fills with search results
- now, plug in a usb stick (yeah, no kidding)
  => Bug: all search results disappear from search window

Background/cause: plugging in a usb stick triggers an update of the pvr search window. where first update works fine, any subsequent update erases search results due to wrong reset of <code>m_bSearchConfirmed</code> during search window content update in <code>OnPrepareFileItems</code>.

The second commit introduces to ignore the window message sent upon plugging/unplugging removable media as it makes no sense to update pvr search window content in this case. Nevertheless, the first commit is still needed to ensure proper updates (not erasing the search results) triggered by other window messages/actions.

@xhaggi for review?

### DIFF
--- a/xbmc/pvr/windows/GUIWindowPVRBase.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRBase.cpp
@@ -33,6 +33,7 @@
 #include "epg/Epg.h"
 #include "epg/GUIEPGGridContainer.h"
 #include "filesystem/StackDirectory.h"
+#include "GUIUserMessages.h"
 #include "guilib/GUIMessage.h"
 #include "guilib/GUIWindowManager.h"
 #include "input/Key.h"
@@ -227,6 +228,21 @@ bool CGUIWindowPVRBase::OnMessage(CGUIMessage& message)
         UpdateSelectedItemPath();
       }
       bReturn = true;
+    }
+    break;
+
+    case GUI_MSG_NOTIFY_ALL:
+    {
+      switch (message.GetParam1())
+      {
+        case GUI_MSG_UPDATE_SOURCES:
+        {
+          // removable drive connected/disconnected. base class triggers a window
+          // content refresh, which makes no sense for pvr windows.
+          bReturn = true;
+          break;
+        }
+      }
     }
     break;
   }
--- a/xbmc/pvr/windows/GUIWindowPVRSearch.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRSearch.cpp
@@ -169,8 +169,6 @@ void CGUIWindowPVRSearch::OnPrepareFileItems(CFileItemList &items)
 
   if (m_bSearchConfirmed)
   {
-    m_bSearchConfirmed = false;
-
     bAddSpecialSearchItem = true;
 
     items.Clear();


xbmc/pvr/windows/GUIWindowPVRSearch.cpp
```diff
@@ -22,6 +22,7 @@
 #include "dialogs/GUIDialogOK.h"
 #include "dialogs/GUIDialogProgress.h"
 #include "epg/EpgContainer.h"
+#include "GUIUserMessages.h" // GUI_MSG_UPDATE_SOURCES
```
xhaggi: should we really add this comment?

xbmc/pvr/windows/GUIWindowPVRSearch.cpp
```diff
@@ -241,6 +240,14 @@ bool CGUIWindowPVRSearch::OnMessage(CGUIMessage &message)
...
     }
   }
+  else if (message.GetMessage() == GUI_MSG_NOTIFY_ALL)
+  {
+    if (message.GetParam1() == GUI_MSG_UPDATE_SOURCES)
+    {
+      // removable drive connected/disconnected. base class triggers a window content refresh, which makes no sense for pvr search window.
+      return true;
+    }
+  }
```
xhaggi: should we move this code to `CGUIWindowPVRBase::OnMessage` instead, because it make no sense for all pvr windows.

xhaggi (dismissed):

xbmc/pvr/windows/GUIWindowPVRSearch.cpp
phil65: or, even more generic, move it to WindowBase and check m_allowNonLocalSources of the actual virtual directory?

xbmc/pvr/windows/GUIWindowPVRSearch.cpp
ksooo: i thought this might be useful information, but if you don't like it i can remove it, np

xbmc/pvr/windows/GUIWindowPVRSearch.cpp
ksooo: good point, will do

ksooo: done

ksooo: @phil65 might be a good idea for v18. at this point for v17 i only want to make changes i'm pretty sure not to break something.

ksooo: @phil65 for now, i moved this code to CGUIWindowPVRBase like @xhaggi suggested

phil65: sure, makes sense.

xbmc/pvr/windows/GUIWindowPVRBase.cpp
```diff
@@ -229,6 +230,15 @@ bool CGUIWindowPVRBase::OnMessage(CGUIMessage& message)
       bReturn = true;
     }
     break;
+
+    case GUI_MSG_NOTIFY_ALL:
+    {
+      if (message.GetParam1() == GUI_MSG_UPDATE_SOURCES)
+      {
+        // removable drive connected/disconnected. base class triggers a window content refresh, which makes no sense for pvr windows.
+        return true;
```
ksooo: are you serious? why?

ksooo: okay. not in the mood to discuss things not worth discussing. changed.

xhaggi:
thanks for fixing it .. looks good so far.

ksooo:
jenkins build this please

ksooo:
jenkins build this please

xhaggi (approved):

ksooo:
jenkins build this please

ksooo:
Travis error is unrelated.

PR #10724, opened 2016-10-18
ActiveAESink: Allow sending of plain zeros

fritsch:
This gives the advanced possibility to stream zeros instead of noise. That won't keep AVRs alive anymore, but on Analog / Heaphones there is no need to hammer random noise on the sink.

Background: The noise was just made fool AVRs to not go off while we are idleing.

### DIFF
--- a/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp
+++ b/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp
@@ -533,12 +533,15 @@ void CActiveAE::StateMachine(int signal, Protocol *port, Message *msg)
         switch (signal)
         {
         case CActiveAEControlProtocol::RECONFIGURE:
+        {
           if (m_streams.empty())
           {
             streaming = false;
             m_sink.m_controlPort.SendOutMessage(CSinkControlProtocol::STREAMING, &streaming, sizeof(bool));
           }
           LoadSettings();
+          m_sink.m_controlPort.SendOutMessage(CSinkControlProtocol::SETNOISETYPE, &m_settings.streamNoise, sizeof(bool));
+          m_sink.m_controlPort.SendOutMessage(CSinkControlProtocol::SETSILENCETIMEOUT, &m_settings.silenceTimeout, sizeof(int));
           ChangeResamplers();
           if (!NeedReconfigureBuffers() && !NeedReconfigureSink())
             return;
@@ -547,6 +550,7 @@ void CActiveAE::StateMachine(int signal, Protocol *port, Message *msg)
           // don't accept any data until we are reconfigured
           m_extDeferData = true;
           return;
+        }
         case CActiveAEControlProtocol::SUSPEND:
           UnconfigureSink();
           m_stats.SetSuspended(true);
@@ -1724,6 +1728,9 @@ bool CActiveAE::InitSink()
                                                                      &m_settings.device;
 
   // send message to sink
+  m_sink.m_controlPort.SendOutMessage(CSinkControlProtocol::SETNOISETYPE, &m_settings.streamNoise, sizeof(bool));
+  m_sink.m_controlPort.SendOutMessage(CSinkControlProtocol::SETSILENCETIMEOUT, &m_settings.silenceTimeout, sizeof(int));
+
   Message *reply;
   if (m_sink.m_controlPort.SendOutMessageSync(CSinkControlProtocol::CONFIGURE,
                                                  &reply,
@@ -2560,6 +2567,8 @@ void CActiveAE::LoadSettings()
 
   m_settings.resampleQuality = static_cast<AEQuality>(CSettings::GetInstance().GetInt(CSettings::SETTING_AUDIOOUTPUT_PROCESSQUALITY));
   m_settings.atempoThreshold = CSettings::GetInstance().GetInt(CSettings::SETTING_AUDIOOUTPUT_ATEMPOTHRESHOLD) / 100.0;
+  m_settings.streamNoise = CSettings::GetInstance().GetBool(CSettings::SETTING_AUDIOOUTPUT_STREAMNOISE);
+  m_settings.silenceTimeout = CSettings::GetInstance().GetInt(CSettings::SETTING_AUDIOOUTPUT_STREAMSILENCE) * 60000;
 }
 
 bool CActiveAE::Initialize()
@@ -2622,7 +2631,8 @@ void CActiveAE::OnSettingsChange(const std::string& setting)
       setting == CSettings::SETTING_AUDIOOUTPUT_PASSTHROUGH            ||
       setting == CSettings::SETTING_AUDIOOUTPUT_SAMPLERATE             ||
       setting == CSettings::SETTING_AUDIOOUTPUT_MAINTAINORIGINALVOLUME ||
-      setting == CSettings::SETTING_AUDIOOUTPUT_GUISOUNDMODE)
+      setting == CSettings::SETTING_AUDIOOUTPUT_GUISOUNDMODE           ||
+      setting == CSettings::SETTING_AUDIOOUTPUT_STREAMNOISE)
   {
     m_controlPort.SendOutMessage(CActiveAEControlProtocol::RECONFIGURE);
   }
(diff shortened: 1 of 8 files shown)


fritsch:
Needs more work - no settings callback in ActiveAESink and setting is ignored in ctor.

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAESink.cpp
```diff
@@ -870,6 +870,7 @@ void CActiveAESink::OpenSink()
   delete m_sampleOfSilence.pkt;
   m_sampleOfSilence.pkt = new CSoundPacket(config, m_sinkFormat.m_frames);
   m_sampleOfSilence.pkt->nb_samples = m_sampleOfSilence.pkt->max_nb_samples;
+  m_streamNoise = CSettings::GetInstance().GetBool(CSettings::SETTING_AUDIOOUTPUT_STREAMNOISE);
```
FernetMenta: all audio settings are supposed to be managed by ActiveAE and centralised in CActiveAE::m_settings;

fritsch:
You said: do it simple, so i did the same as you did with the silence timer.

Am 19.10.2016 7:27 vorm. schrieb "Rainer Hochecker" <
notifications@github.com>:

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAESink.cpp
fritsch: You are obviously right. That's why the commit says: discuss. Will fix.

FernetMenta: please move setting for silence timer too

fritsch:
On my way.

Am 19.10.2016 6:31 nachm. schrieb "Rainer Hochecker" <
notifications@github.com>:

fritsch:
In the words of Knuth: I proved it correct, build it - but had no time for runtime testing. Tomorrow next timeslot.

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAESink.h
```diff
@@ -40,6 +40,8 @@ struct SinkConfig
   AEAudioFormat format;
   CEngineStats *stats;
   const std::string *device;
+  int extSilenceTimeOut;
+  bool streamNoise;
```
FernetMenta: this does not fit well here, better send it with CSinkControlProtocol::STREAMING

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.h
```diff
@@ -72,6 +72,8 @@ struct AudioSettings
   unsigned int samplerate;
   AEQuality resampleQuality;
   double atempoThreshold;
+  bool streamNoise;
```
FernetMenta: those don't fit here, send them is a separate struct

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAESink.cpp
```diff
@@ -322,6 +321,14 @@ void CActiveAESink::StateMachine(int signal, Protocol *port, Message *msg)
           m_extStreaming = *(bool*)msg->data;
           return;
 
+        case CSinkControlProtocol::SILENCECOUNTERUPDATED:
+          m_silenceTimeOut = *(int*)msg->data;
+        return;
```
FernetMenta: wrong indentation

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAESink.h
```diff
@@ -63,6 +63,8 @@ class CSinkControlProtocol : public Protocol
     VOLUME,
     FLUSH,
     TIMEOUT,
+    SILENCECOUNTERUPDATED,
+    NOISECHANGED,
```
FernetMenta: here "changed", the other "updated". the the goal to have as many different names as possible? :)
I would call the signal SetXXX, that also fits initial setting of those parameters

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.h
fritsch: Read again. Those are settings members.

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAESink.h
```diff
@@ -63,6 +63,8 @@ class CSinkControlProtocol : public Protocol
     VOLUME,
     FLUSH,
     TIMEOUT,
+    SILENCECOUNTERUPDATED,
```
fritsch: As you use it as a tristate in the sink. Once for using an infty value, then to set it to zero. This is just used to transmit the setting into the sink.

fritsch: Concerning the second comment: putting it into the streaming message makes no sense, as it is not related.

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAESink.cpp
fritsch: will fix

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp
```diff
@@ -1724,6 +1731,9 @@ bool CActiveAE::InitSink()
                                                                      &m_settings.device;
 
   // send message to sink
+  m_sink.m_controlPort.SendOutMessage(CSinkControlProtocol::NOISECHANGED, &m_settings.streamNoise, sizeof(bool));
+  m_sink.m_controlPort.SendOutMessage(CSinkControlProtocol::SILENCECOUNTERCHANGED, &m_settings.silenceTimeout, sizeof(int));
```
FernetMenta: again, I won't accept "counter"

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp
```diff
@@ -533,12 +533,18 @@ void CActiveAE::StateMachine(int signal, Protocol *port, Message *msg)
...
         case CActiveAEControlProtocol::RECONFIGURE:
+        {
           if (m_streams.empty())
           {
             streaming = false;
             m_sink.m_controlPort.SendOutMessage(CSinkControlProtocol::STREAMING, &streaming, sizeof(bool));
           }
+          AudioSettings settings = m_settings;
           LoadSettings();
+          if (settings.streamNoise != m_settings.streamNoise)
```
FernetMenta: what the the magic here? first you assign m_settings to settings, than you compare those two. what do you expect? what behaviour do you want to achieve? if you followed my advice with naming of the signals, you would have avoided this case. when "setting" a value the old value is not of any interest. you can always send the message and no comparison is needed.

fritsch: My idea: Only if the setting has changed, we need to send the signal. Intention: Avoid GenerateNoise() if not needed.

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAESink.cpp
```diff
@@ -394,6 +401,17 @@ void CActiveAESink::StateMachine(int signal, Protocol *port, Message *msg)
           m_volume = *(float*)msg->data;
           m_sink->SetVolume(m_volume);
           return;
+
+        case CSinkControlProtocol::NOISETYPECHANGED:
+        {
+          bool streamNoise = *(bool*)msg->data;
+          if (streamNoise != m_streamNoise)
```
fritsch: Yes. You are right here. It basically is a "triggered" signal. With a bool it will always change - so this is most likely not needed here, right? But think of e.g. streamsilence parameter, where user could pick up the setting, scroll through it and select the same value as before.

FernetMenta:
looks good now, thanks!

MartijnKaijser:
Please do not use string ID lower than 10000

fritsch:
@MartijnKaijser for the record: why? (PR is already changed)

MartijnKaijser:
@fritsch because try to find a strings with a low id gives you a gazillion unrelated results in the source code when trying to track down where these are used. Anything about 10000 is almost a perfect hit. Since we are trying to clean up unused strings removing the lower ones and not adding them again makes this easier.

MartijnKaijser:
jenkins build this please

fritsch:
No problem, can change. What's the reason for that?

Am 22.10.2016 12:44 nachm. schrieb "Martijn Kaijser" <
notifications@github.com>:

PR #10725, opened 2016-10-18
Fix mapping controllers with accelerometers

garbear:
If a controller has an accelerometer, when the user goes to map a button, the configuration wizard will immediately skip to the next button. This is because the Z axis is "activated", so Kodi thinks the user is trying to map this axis to the button.

The fix is to wait 50ms when the user starts mapping buttons. If an axis is activated, then it is ignored until deactivated again. This means that as long as the controller stays upright, the accelerometer shouldn't get in the way anymore.

Only tested with a faked accelerometer so far, I'm in the process of getting a controller that has a physical one.

Broken out from #10630


garbear:
Superseded by #10910

PR #10726, opened 2016-10-19
Fix deadlock introduced by #10690

garbear:
This fixes a possible deadlock by querying the add-on database before acquiring the lock.


garbear:
jenkins build this please

PR #10727, opened 2016-10-19
VideoPlayer: ffmpeg, fix calculating aspect ratio for 3d modes

FernetMenta:
finally fix http://trac.kodi.tv/ticket/16876

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
@@ -1209,19 +1209,21 @@ int CDVDDemuxFFmpeg::GetNrOfStreams() const
 
 double CDVDDemuxFFmpeg::SelectAspect(AVStream* st, bool& forced)
 {
-  // for stereo modes, use codec aspect ratio
-  AVDictionaryEntry *entry = av_dict_get(st->metadata, "stereo_mode", NULL, 0);
-  if (entry)
-  {
-    forced = false;
-    return av_q2d(st->codecpar->sample_aspect_ratio);
-  }
-
   // trust matroshka container
   if (m_bMatroska && st->sample_aspect_ratio.num != 0)
   {
     forced = true;
-    return av_q2d(st->sample_aspect_ratio);
+    double dar = av_q2d(st->sample_aspect_ratio);
+    // for stereo modes, use codec aspect ratio
+    AVDictionaryEntry *entry = av_dict_get(st->metadata, "stereo_mode", NULL, 0);
+    if (entry)
+    {
+      if (strcmp(entry->value, "left_right") == 0 || strcmp(entry->value, "right_left") == 0)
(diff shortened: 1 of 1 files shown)


afedchin:
thanks. will test this tonight.

xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp
```diff
@@ -833,7 +833,13 @@ bool CDVDVideoCodecFFmpeg::GetPictureCommon(DVDVideoPicture* pDvdVideoPicture)
     aspect_ratio = av_q2d(pixel_aspect) * pDvdVideoPicture->iWidth / pDvdVideoPicture->iHeight;
 
   if (aspect_ratio <= 0.0)
+  {
```
afedchin: If aspect is calculated already then all code below will not work. so the correction doesn't work also.

afedchin:
I can't test with all my samples, but seems with this an anamorph video is displayed wrong. 
I'll try to explain:
1. for anamorph videos (1920x1080) we should use codec aspect without changes.
2. for non anamorph videos (1920x2160 or 3840x1080) we should change aspect to *2 or /2

afedchin:
here is a sample https://www.dropbox.com/s/q5y3q4jee30y4hx/3d.sbs.disney_3d_global_sizzle_2012-DWEU.m2ts?dl=0 which doesn't work with this

xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp
FernetMenta: this block is only executed if codec does not provide AR

afedchin:
with this https://github.com/afedchin/xbmc/commit/39898904ee90a6866b20d122cf025d81423dfaf5 I have correct aspect for anamorph and full stereo. as I said above I will test with more samples tonight.

FernetMenta:
your sample does not have stereo_mode set by ffmpeg. stereo_mode is set by CStereoscopicsManager::DetectStereoModeByString. 
this is all rather hacky and a can of worms ...
we have to keep things clear and separated.

afedchin:
but we support both stereo flags from a container and file naming convention.

FernetMenta:
but the one form file naming must not be used to determine aspect ratio. PR is updated.

afedchin:
this https://www.dropbox.com/s/an8v07r1t3k0e9h/3d.Kodi.test.mkv?dl=0 sample (anamorph stereo) still not work as expected.

FernetMenta:
@afedchin I looked into ffmpeg code and noticed that AR already observes stereo modes. All samples I got work with this change.

afedchin:
after quick test seems it returns pre ffmpeg 3.x behavior

FernetMenta:
great, thanks

jenkins build this please

PR #10730, opened 2016-10-19
Allow multiple content elements in a single container

b-pass:
## Description

This change adds a new IListProvider, called CMultiProvider, which aggregates several independent &lt;content&gt; blocks (either DirectoryProviders or StaticProviders or both).
## Motivation and Context

DirectoryProviders (dynamically filled &lt;content&gt; blocks) are a powerful feature.  But currently you can only fill a list with a single directory at a time, and you also cannot add other static elements to a dynamic list.  For example, you might like to add some navigational list entries at the top of a list, and have it fill with dynamic content below those static entries.  Or maybe you would like to combine two different directories in one list for some reason.  This was hinted as a potential new feature in 2014, but it seems it never got added: http://forum.kodi.tv/showthread.php?tid=176864
## How Has This Been Tested?

I tested this by trying a few different combinations of multiple &lt;content&gt; blocks.
## Types of change

## Checklist:

### DIFF
--- a/xbmc/listproviders/MultiProvider.cpp
+++ b/xbmc/listproviders/MultiProvider.cpp
@@ -1,0 +1,118 @@
+/*
+ *      Copyright (C) 2013-2017 Team Kodi
+ *      http://kodi.tv
+ *
+ *  This Program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2, or (at your option)
+ *  any later version.
+ *
+ *  This Program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with XBMC; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "MultiProvider.h"
+#include "utils/XBMCTinyXML.h"
+#include "threads/SingleLock.h"
+
+CMultiProvider::CMultiProvider(const TiXmlNode *first, int parentID)
+ : IListProvider(parentID)
+{
+  for (const TiXmlNode *content = first; content; content = content->NextSiblingElement("content"))
+  {
+    IListProviderPtr sub(IListProvider::CreateSingle(content, parentID));
+    if (sub)
+      m_providers.push_back(sub);
+  }
+}
+
+bool CMultiProvider::Update(bool forceRefresh)
+{
+  bool result = false;
+  for (auto& provider : m_providers)
+    result |= provider->Update(forceRefresh);
+  return result;
+}
+
+void CMultiProvider::Fetch(std::vector<CGUIListItemPtr> &items)
+{
+  CSingleLock lock(m_section);
+  std::vector<CGUIListItemPtr> subItems;
+  items.clear();
+  m_itemMap.clear();
+  for (auto const& provider : m_providers)
+  {
+    provider->Fetch(subItems);
+    for (auto& item : subItems)
+    {
+      auto key = GetItemKey(item);
+      m_itemMap[key] = provider;
+      items.push_back(item);
+    }
+    subItems.clear();
+  }
+}
+
+bool CMultiProvider::IsUpdating() const
+{
+  bool result = false;
+  for (auto const& provider : m_providers)
+    result |= provider->IsUpdating();
+  return result;
+}
+
+void CMultiProvider::Reset()
+{
+  {
+    CSingleLock lock(m_section);
+    m_itemMap.clear();
+  }
+
+  for (auto const& provider : m_providers)
+    provider->Reset();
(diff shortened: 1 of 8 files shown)


HitcherUK:
Nice, thanks.

BigNoid:
Nice, tested and works great. Thx for this addition!
You can pick https://github.com/BigNoid/xbmc/commit/241de15944ef145150a3a07f190cd5d731d5def4 for Xcode sync

b-pass:
Picked and Squashed, thanks!

xbmc/listproviders/MultiProvider.h
```diff
@@ -0,0 +1,51 @@
...
+public:
+  CMultiProvider(const TiXmlNode *first, int parentID);
+  
+  virtual bool Update(bool forceRefresh);
+  virtual void Fetch(std::vector<CGUIListItemPtr> &items) const;
+  virtual bool IsUpdating() const;
+  virtual void Reset(bool immediately = false);
+  virtual bool OnClick(const CGUIListItemPtr &item);
+  virtual bool OnInfo(const CGUIListItemPtr &item);
+  virtual bool OnContextMenu(const CGUIListItemPtr &item);
```
garbear: can you add the override keyword on these functions?

xhaggi:
looks good +1

da-anda:
this would be great for some home screen widgets on Estuary in v17. Are there any major concerns to hold it back for v17 (besides of feature freeze)?

xbmc/listproviders/MultiProvider.h
```diff
@@ -0,0 +1,51 @@
...
+  virtual void Fetch(std::vector<CGUIListItemPtr> &items) const override;
+  virtual bool IsUpdating() const override;
+  virtual void Reset(bool immediately = false) override;
+  virtual bool OnClick(const CGUIListItemPtr &item) override;
+  virtual bool OnInfo(const CGUIListItemPtr &item) override;
+  virtual bool OnContextMenu(const CGUIListItemPtr &item) override;
+  
+protected:
+  std::vector<IListProviderPtr> m_providers;
+  mutable std::map<CGUIListItem*, IListProviderPtr> m_itemMap;
```
b-pass: The idea is that the MultiProvider needs to know which sub-provider will be sent Click/Info/Context messages so it needs to keep track of which content element belong to which sub-provider.

The only way I see to maintain that mapping is to build it during Fetch().  Unfortunately, Fetch is const in the base class.

I could maybe change that without too much impact on other code.  Would that be preferable to a mutable here?

b-pass: I went ahead and made that change an squashed it in

b-pass:
Fixed conflicts and squashed

b-pass:
Is there anything else I need to get this merged?  I don't want it to get forgotten and miss the boat for v19....

ronie:
jenkins build this please

xbmc/listproviders/MultiProvider.cpp
```diff
@@ -0,0 +1,109 @@
+/*
+ *      Copyright (C) 2013 Team XBMC
+ *      http://xbmc.org
```
xhaggi: Please update the copyright in all files you've touched.

```
*      Copyright (C) 201X-2017 Team Kodi
*      http://kodi.tv
```

xhaggi (changes requested):

xbmc/listproviders/MultiProvider.cpp
```diff
@@ -0,0 +1,109 @@
...
+  CSingleLock lock(m_section);
+  std::vector<CGUIListItemPtr> subItems;
+  items.clear();
+  m_itemMap.clear();
+  for (auto const& provider : m_providers)
+  {
+    provider->Fetch(subItems);
+    for (auto& item : subItems)
+    {
+      m_itemMap[item.get()] = provider;
```
tamland: This doesn't look very safe. There is no guarantee other providers aren't going to kill these items at any time, so you should hold a shared pointer.

b-pass: I never access the items from the pointers held in this map.  When events come in they have a shared pointer, the map just directs which provider should receive the call.

xbmc/listproviders/MultiProvider.cpp
b-pass: @xhaggi Done.

xbmc/listproviders/MultiProvider.cpp
b-pass: @tamland I changed the map definition to use a void* for this key, which will help discourage someone from accidentally using the key as a pointer.

tamland: Well, it's still storing possibly dangling pointer which is being accessed by `.find`. Are you sure this is legal and safe?

Paxxi: I'm with @tamland on this one. It's a neat trick but it's going to blow up sooner or later, it's also quite non-intuitive to use a pointer address as the key value. There should be a better way to keep track of the items.

b-pass: @tamland @Paxxi Alright, I've changed it to use a size_t for keys.

tamland:
jenkins build this please

PR #10731, opened 2016-10-19
Change splash db upgrade logic

phate89:
Change the splash screen to show the message of upgrading database only when we're actually upgrading a database.
## Motivation and Context

The old method wassn't aware if there was some upgrade and what db was doing the upgrade
## How Has This Been Tested?

Tested upgrading my old mysql databases
## Types of change

## Checklist:

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -14654,13 +14654,13 @@ msgstr ""
 #. Progress text on splash screen
 #: xbmc/Application.cpp
 msgctxt "#24150"
-msgid "Database migration in progress - please wait..."
+msgid "Database migration in progress - please wait"
 msgstr ""
 
 #. Progress text on splash screen
 #: xbmc/Application.cpp
 msgctxt "#24151"
-msgid "Add-on migration in progress - please wait..."
+msgid "Add-on migration in progress - please wait"
 msgstr ""
 
 #empty strings from id 24152 to 24990
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -25,6 +25,7 @@
 #include "events/EventLog.h"
 #include "events/NotificationEvent.h"
 #include "interfaces/builtins/Builtins.h"
+#include "utils/JobManager.h"
 #include "utils/Variant.h"
 #include "utils/Splash.h"
 #include "LangInfo.h"
@@ -1123,8 +1124,22 @@ bool CApplication::Initialize()
   g_curlInterface.Unload();
 
   // initialize (and update as needed) our databases
-  CSplash::GetInstance().Show(g_localizeStrings.Get(24150));
-  CDatabaseManager::GetInstance().Initialize();
+  CEvent event(true);
+  CJobManager::GetInstance().Submit([&event]() {
+    CDatabaseManager::GetInstance().Initialize();
+    event.Set();
+  });
+  std::string localizedStr = g_localizeStrings.Get(24150);
+  int iDots = 1;
+  while (!event.WaitMSec(1000))
+  {
+    if (CDatabaseManager::GetInstance().m_bIsUpgrading)
+      CSplash::GetInstance().Show(std::string(iDots, ' ') + localizedStr + std::string(iDots, '.'));
+    if (iDots == 3)
+      iDots = 1;
+    else
+      ++iDots;
+  }
   CSplash::GetInstance().Show();
 
   StartServices();
@@ -1139,11 +1154,30 @@ bool CApplication::Initialize()
     g_windowManager.CreateWindows();
 
     m_confirmSkinChange = false;
-    m_incompatibleAddons = CAddonSystemSettings::GetInstance().MigrateAddons([](){
-      CSplash::GetInstance().Show(g_localizeStrings.Get(24151));
+
+    std::vector<std::string> incompatibleAddons;
+    event.Reset();
+    std::atomic<bool> isMigratingAddons(false);
+    CJobManager::GetInstance().Submit([&event, &incompatibleAddons, &isMigratingAddons]() {
+      incompatibleAddons = CAddonSystemSettings::GetInstance().MigrateAddons([&isMigratingAddons]() {
+        isMigratingAddons = true;
+      });
+      event.Set();
     });
-    m_confirmSkinChange = true;
+    localizedStr = g_localizeStrings.Get(24151);
+    iDots = 1;
+    while (!event.WaitMSec(1000))
+    {
+      if (isMigratingAddons)
+        CSplash::GetInstance().Show(std::string(iDots, ' ') + localizedStr + std::string(iDots, '.'));
+      if (iDots == 3)
+        iDots = 1;
+      else
+        ++iDots;
+    }
     CSplash::GetInstance().Show();
+    m_incompatibleAddons = incompatibleAddons;
+    m_confirmSkinChange = true;
 
     std::string defaultSkin = ((const CSettingString*)CSettings::GetInstance().GetSetting(CSettings::SETTING_LOOKANDFEEL_SKIN))->GetDefault();
     if (!LoadSkin(CSettings::GetInstance().GetString(CSettings::SETTING_LOOKANDFEEL_SKIN)))
--- a/xbmc/utils/Splash.cpp
+++ b/xbmc/utils/Splash.cpp
@@ -39,14 +39,9 @@ CSplash& CSplash::GetInstance()
   return instance;
 }
 
-void CSplash::Show()
+void CSplash::Show(const std::string& message /* = "" */)
 {
-  Show("");
-}
-
-void CSplash::Show(const std::string& message)
-{
-  if (!g_advancedSettings.m_splashImage)
+  if (!g_advancedSettings.m_splashImage && !(m_image || !message.empty()))
     return;
 
   if (!m_image)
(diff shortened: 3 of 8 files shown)


phate89:
@Paxxi @DaveTBlake for review/test

DaveTBlake:
A welcome follow on from #10483 that only shows when migration tasks happening rather than every start up. @MilhouseVH, @tamland and @fritsch may be interested in this too.

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -14663,7 +14663,55 @@ msgctxt "#24151"
 msgid "Add-on migration in progress - please wait..."
 msgstr ""
 
-#empty strings from id 24152 to 24990
+#. Localized addon database name for splash screen
+#: xbmc/addons/AddonDatabase.h
+msgctxt "#24152"
+msgid "Addon"
```
phil65: nitpick: Normally we use "Add-on" instead of "Addon".

MilhouseVH:
Thanks @Phaeodaria - it works well, I've tested upgrading MyMusic50 to MyMusic60 and MyVideos90 to MyVideos107, the latter is particularly time consuming. I'll include this PR in future builds.

There does however remain one issue: when `<splash>false</splash>` is added to advancedsettings.xml, there is no splash displayed, and also no migration messages (even when Music and Videos databases are being migrated). All the user sees is a totally blank (black) screen, with no messages.

IMHO the splash should be forced to appear along with the migration messages, regardless of the `<splash>` setting, whenever a migration is taking place.

One other nitpick would be to position the text a little higher, just in case of any overscanning or weird resolution issues - it seems a little unnecessary to position the text quite so low down (it appears to be close to the bottom edge of my 1920x1080 1:1 pixel screen).

DaveTBlake:
+1 
Tested with a large MyMusic48 to MyMusic60.

I too would put the text a little higher, and probably shouldn't capatalise "Database", e.g.
"Music database migration in progress - please wait..."
But minor stuff, and I would happily live with it.

addons/resource.language.en_gb/resources/strings.po
MartijnKaijser: You might as well repeat the complete sentence each time instead of stitching the strings together.

DaveTBlake: The other view is that storing the same text over and over is wasteful.
But is it a translation issue e.g. in some languages the object needs to go else where in the phrase?

phate89:
I updated the pr.. Now we update the text adding dot every second..
If someone is familiar with our CThread should check my code...

tamland:
Sorry but not liking how this is currently implemented. The database should not have a dependency on the global application messenger. Instead it should be passed in as a callback/another agnostic structure, or the entire migration refactored and uncoupled from the database. This is why I didn't bothered with it in the first place.

xbmc/utils/Splash.cpp
```diff
@@ -39,14 +45,46 @@ CSplash& CSplash::GetInstance()
...
+  while (!m_bStop)
+  {
+    for (int i = 1; i <= 3 && !m_bStop; ++i)
+    {
+      Show(std::string(i, ' ') + text + std::string(i, '.'));
+      if (!m_bStop)
+        Sleep(1000);
+    }
+  }
+  Show();
```
FernetMenta: Show must only be called by the render thread.

phate89: @FernetMenta  I started with something simpler and I moved to something out of my area.. Thanks for checking the pr.. Can you point me to an example how I can I call it every second from render thread? I really know nothing about this part..

Paxxi:
@Tamland trust that passing in a callback could be cleaner but more work. This is a step in the right direction and app messenger is purposely designed for this with very few dependencies itself. 
Any work that removes some lower level dependency of a higher level is progress and I encourage it. Will review the implementation later when I'm at a computer

phate89:
@tamland I added a new commit that moves the update part (the same update procedure) to databasemanager. So only databasemanager has a dependency on applicationmessenger..Is it better?

xbmc/utils/Splash.cpp
Paxxi: You can use app messenger and post a guide message, those are handled on the correct thread. Fernet might have a better idea though

FernetMenta: This is dead end and it should become clearer now that the approach of showing upgrade info with CSplash was an abuse of this class. The splash image is shown on init before the main loop gets active. This is also before app messenger can be used. 
I recommend to revert the upgrade messages and start from scratch.

phate89: As I said it's out of my area and I have n idea where to start so starting from scratch is probably too hard for me.. I'll close the pr so if someone wants to try it can..

FernetMenta:
you could offload the migrations tasks to an extra thread. then the main thread can wait in init for this task to be completed and display splash and messages accordingly.

phate89:
@fernetmenta thanks for the suggestions. Any help is appreciated. 
The main problem wasn't to actually update from main thread but update the text every second to keep something moving.
You are suggesting move the loop in capplication? Something like 
'''
CDatabaseManager::GetInstance().StartInitialize();
while (CDatabaseManager::GetInstance().IsRunning())
{
  CSplash::GetInstance().Show(updatedMsg);
  if (CDatabaseManager::GetInstance().IsRunning())
    sleep (1000);
}
'''

FernetMenta:
@phate89 yes, this will hold processing of the main thread until upgrade has finished. This is intended behaviour.

phate89:
@FernetMenta I updated the pr.. I moved the update logic to the db manager and moved the db upgrade and the addon migration to a separate thread (If I understood well our cthread)

tamland:
Not exactly thread safe.. Try creating a task and send it off with CJobManager::Submit instead, or spawn a new thread; don't inherit from Cthread. Things like CAddonSystemSettings shouldn't *be* a thread.

FernetMenta:
@phate89 using CThread may work but is not the best choice. What you really want to do is spawning an upgrade thread if required. Look into CJob and CJobManager. Those classes do what you need here.

phate89:
Thanks from the tip.. it'' quite nice ttt see we already have so many classes ready for several tasks.. I''l look into CJobManager and report back

phate89:
@tamland  @FernetMenta I updated the pr using CJobManager::Submit. I hope I got it right this time :|

xbmc/Application.cpp
```diff
@@ -1123,8 +1124,22 @@ bool CApplication::Initialize()
...
+    notFinished = false;
+  });
+  std::string localizedStr = g_localizeStrings.Get(24150);
+  while (notFinished)
+  {
+    for (int i = 1; i <= 3 && notFinished; ++i)
+    {
+      if (CDatabaseManager::GetInstance().m_bIsUpgrading)
+        CSplash::GetInstance().Show(std::string(i, ' ') + localizedStr + std::string(i, '.'));
+      if (notFinished)
```
FernetMenta: is is already a condition of the for loop. why do you test again?

xbmc/Application.cpp
```diff
@@ -1123,8 +1124,22 @@ bool CApplication::Initialize()
...
+  CJobManager::GetInstance().Submit([&notFinished]() {
+    CDatabaseManager::GetInstance().Initialize();
+    notFinished = false;
+  });
+  std::string localizedStr = g_localizeStrings.Get(24150);
+  while (notFinished)
+  {
+    for (int i = 1; i <= 3 && notFinished; ++i)
+    {
+      if (CDatabaseManager::GetInstance().m_bIsUpgrading)
```
FernetMenta: what is the difference between m_bIsUpgrading and notFinished?

phate89: notFinished is about the whole procedure. Not only the actual upgrade but also checking if there iis an old db and copying if we don't find one.  Isupgrading is when the actual migration from an old one to a new one happens. That's the slow part we want to display.
There are 2 different because there's no simple check ahead of time if we need to do some upgrade. For every database we get back version by version to find if there is a ddatabase.  If the upgrade fails we keep going back..

xbmc/Application.cpp
phate89: I added it because the work might end while drawing the message and might avoid to wait 1 sec in that case.. not sure if it's worth it

FernetMenta:
Looks ok from my pov. I can't say much about the upgrade itself

phate89:
jenkins build this please

tamland:
Getting closer:) but you need to use std::future or events, and wait. With the sleeps the migration is up to 1 second slower than it need to be.

phate89:
@tamland
I know waiting is better but we also want to update the text every second to show some progress. It's something that can take several minutes so moving dots helps to see we're not stuck.

tamland:
That's what it does. If you pass 1 second to the wait method it returns `ready` or `timeout` after 1 second. If it's ready before that it will be interrupted.

phate89:
@tamland  I'm dumb of course you're right.. i updated it using our CEvent class

phate89:
now it should be ok.. thanks @FernetMenta @tamland for all the help..
jenkins build this please

MartijnKaijser:
win32 upgrade from v14 and all is fine. thx
jenkins build this please.

MartijnKaijser:
@FernetMenta @tamland if all good please merge :)

FernetMenta:
fine by me, @tamland your button

PR #10732, opened 2016-10-19
[CMAKE] - fixups for ios/binaddons

Memphiz:
Those are some fixes which i found during building with cmake 3.6 for tvos on our jenkins slave which triggered a bug due to sdk/osx host version constellation. I filed a bug report on cmake and it will be fixed in cmake 3.7 - those commits from this PR are needed in addition.

@fetzerch for review


fetzerch (commented):
LGTM :)

Memphiz:
Jenkins build and merge

PR #10734, opened 2016-10-19
Use new add-on hooks to enable/disable peripheral add-ons

garbear:
This improves on #10690 by avoiding the old hooks and using the new ones that @tamland added


garbear:
@tamland how's this look?

tamland:
good. it was already implemented (..?) It's also possible to get addon id from those event btw; if you want avoid running the whole 'UpdateAddons'

garbear:
k, thanks for the feedback. gotta love when you go to fix something and it's already implemented :) the problem was that we weren't listening for enable/disable, just install, so that's why enable/disable had no immediate effect in the gui. Adding enable/disable events fixed this.

jenkins build this please

PR #10735, opened 2016-10-19
Fix button maps not refreshing after a reset

garbear:
If the user resets the controller in the controller dialog, the effect doesn't take effect immediately.

controller reset

This fix causes the button map to be refreshed when the user hits reset.

Broken out from #10630


garbear:
jenkins build this please

garbear:
@MartijnKaijser can you tag as a fix for Krypton 17.0-beta4?

PR #10736, opened 2016-10-20
[win32] A set of fixes for dx resources.

afedchin:
see title.


afedchin:
jenkins build this please

PR #10737, opened 2016-10-20
[filesystem/nfs] - ensure that the last access time of a context is u…

Memphiz:
…pdated during keep alive and that clearmembers doesn't clear the list of keep alive handlers during context change - fixes #16576

@MilhouseVH please verify


MilhouseVH:
That looks good, no more crashes!

Is there likely to be anything similar that might affect `smb://`? I don't use `smb://` for very much so wouldn't really see such an issue.

Memphiz:
No i don't think so. Its special to nfs...

Memphiz:
Jenkins build and merge

PR #10738, opened 2016-10-20
AESinkAudioTrack: Don't use IEC if we would loose DTSHD (API 7/7.1 sh…

fritsch:
Sadly with Android N 7 and 7.1 there won't be DTS-HD / TrueHD passthrough via the superior IEC API. Don't switch to IEC API if user also has DTSHD API available.


fritsch:
Good that the age of RAW PT is nearly gone ...

PR #10739, opened 2016-10-20
Fixed OnPlayBackResumed not getting fired anymore.

razzeee:
## Description

Make sure OnPlayBackResumed get's fired when we resume.
## Motivation and Context

http://trac.kodi.tv/ticket/16976#ticket
## How Has This Been Tested?

Tested on windows 10 in the vs debugger
## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -3104,12 +3104,10 @@ void CVideoPlayer::Pause()
   if (GetSpeed() == 0)
   {
     SetSpeed(1);
-    m_callback.OnPlayBackResumed();
   }
   else
   {
     SetSpeed(0);
-    m_callback.OnPlayBackPaused();
   }
 }
 
@@ -3646,6 +3644,13 @@ void CVideoPlayer::SetSpeed(float speed)
     return;
   
   m_newPlaySpeed = speed * DVD_PLAYSPEED_NORMAL;
+  if (m_newPlaySpeed != m_playSpeed)
+  {
+    if (m_newPlaySpeed == DVD_PLAYSPEED_NORMAL)
+      m_callback.OnPlayBackResumed();
+    else if (m_newPlaySpeed == DVD_PLAYSPEED_PAUSE)
+      m_callback.OnPlayBackPaused();
+  }
   SetPlaySpeed(speed * DVD_PLAYSPEED_NORMAL);
 }
 


xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3104,7 +3104,6 @@ void CVideoPlayer::Pause()
   if (GetSpeed() == 0)
   {
     SetSpeed(1);
```
Razzeee: @FernetMenta 
I'm a little bit concerned about this code, as it seems like it might be dead code right now. Should I remove it?

FernetMenta: what exactly do you consider dead code?

please move both callbacks to SetSpeed

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3647,6 +3646,8 @@ void CVideoPlayer::SetSpeed(float speed)
   
   m_newPlaySpeed = speed * DVD_PLAYSPEED_NORMAL;
   SetPlaySpeed(speed * DVD_PLAYSPEED_NORMAL);
+  if (speed == 1)
```
FernetMenta: you need to compare to old speed or you get resume calls even if nothing changed.

Razzeee:
Updated

garbear:
I don't think that this will work, because if you call SetSpeed(1) then SetSpeed(2) it will call `m_callback.OnPlayBackPaused();`

Razzeee:
Right forgot about fast forward

garbear:
add the new speed to the comparison.

if (old == 0 && new != 0) OnResume();
if (old != 0 && new == 0) OnPause();

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3644,9 +3642,17 @@ void CVideoPlayer::SetSpeed(float speed)
 
   if (!CanSeek())
     return;
-  
+  float m_oldPlayerSpeed = m_playSpeed;
   m_newPlaySpeed = speed * DVD_PLAYSPEED_NORMAL;
   SetPlaySpeed(speed * DVD_PLAYSPEED_NORMAL);
+  if (m_oldPlayerSpeed == 0)
```
FernetMenta: Remember what I said on the other PR: player commands are asynchronous. Means it does not matter if you call the callback before or after SetPlaySpeed, speed won't have changed anyway at this time. No need for m_oldPlayerSpeed, compare m_newPlaySpeed with m_PlaySpeed.
The behaviour is as follows:
resume: this is whenever new speed is normal and old speed was not normal speed
pause: new speed is DVD_PLAYSPEED_PAUSE and old speed was not DVD_PLAYSPEED_PAUSE

Razzeee:
Updated, I hope I understood you this time :)

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3647,6 +3645,14 @@ void CVideoPlayer::SetSpeed(float speed)
   
   m_newPlaySpeed = speed * DVD_PLAYSPEED_NORMAL;
   SetPlaySpeed(speed * DVD_PLAYSPEED_NORMAL);
+  if (m_newPlaySpeed == DVD_PLAYSPEED_NORMAL && m_playSpeed != DVD_PLAYSPEED_NORMAL)
```
FernetMenta: this block needs to be before SetSplaySpeed

with an outer condition and simpler inner conditions the code would be more readable
what do you think?

```
if (m_newPlaySpeed != m_playSpeed)
{
  if (m_newPlaySpeed == DVD_PLAYSPEED_NORMAL)
    m_callback.OnPlayBackResumed();
  else if (m_newPlaySpeed == DVD_PLAYSPEED_PAUSED)
    m_callback.OnPlayBackPaused();
}
SetPlaySpeed(speed * DVD_PLAYSPEED_NORMAL);
```

Razzeee: Why does it need to be before SetPlaySpeed? I thought that was what you ment, when you said the code is async and it doesn't matter in which order it get's called? Just trying to understand things here.

I like what you did to the code format, will test and push the changes.

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3646,6 +3644,13 @@ void CVideoPlayer::SetSpeed(float speed)
...
   
   m_newPlaySpeed = speed * DVD_PLAYSPEED_NORMAL;
+  if (m_newPlaySpeed != m_playSpeed)
+  {
+      if (m_newPlaySpeed == DVD_PLAYSPEED_NORMAL)
+          m_callback.OnPlayBackResumed();
+      else if (m_newPlaySpeed == DVD_PLAYSPEED_PAUSE)
+          m_callback.OnPlayBackPaused();
+  }
   SetPlaySpeed(speed * DVD_PLAYSPEED_NORMAL);
```
garbear: SetPlaySpeed() is public, so should this check occur within SetPlaySpeed()?

Side comment: i sure hope so, because razzeee hasn't updated this pr nearly enough 😉

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3646,6 +3644,13 @@ void CVideoPlayer::SetSpeed(float speed)
     return;
   
   m_newPlaySpeed = speed * DVD_PLAYSPEED_NORMAL;
+  if (m_newPlaySpeed != m_playSpeed)
+  {
+      if (m_newPlaySpeed == DVD_PLAYSPEED_NORMAL)
```
FernetMenta: wrong indentation, apart from this it is ok now

xbmc/cores/VideoPlayer/VideoPlayer.cpp
FernetMenta: I could not find my initial comment to @Razzeee anymore, so again: This is an intermediate fix and it does not fix the real issue that is the blocking callback interface. Because it is blocking VideoPlayer's internal thread can't call it. @garbear SetPlaySpeed is public but must not block because it is also called by the internal thread.

the current state: it works somehow and continues to have issues as it has always had. An external call to Pause() or SetSpeed just triggers an asynchronous command to VideoPlayer. The real action happens later. As a result the callback is just a guess what player might do later on. It can be wrong.

Razzeee:
Updated

Razzeee:
jenkins build this please

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3646,6 +3644,13 @@ void CVideoPlayer::SetSpeed(float speed)
     return;
   
   m_newPlaySpeed = speed * DVD_PLAYSPEED_NORMAL;
+  if (m_newPlaySpeed != m_playSpeed)
+  {
+    if (m_newPlaySpeed == DVD_PLAYSPEED_NORMAL)
+        m_callback.OnPlayBackResumed();
+    else if (m_newPlaySpeed == DVD_PLAYSPEED_PAUSE)
+        m_callback.OnPlayBackPaused();
```
Razzeee: Hrm, seems like I've reset my visual studio somewhen, grml

Razzeee:
jenkins build this please

MartijnKaijser:
@FernetMenta all good now?
jenkins build this please

FernetMenta:
looks good

PR #10740, opened 2016-10-21
Move peripherals to shared_ptr

garbear:
This stores peripherals as shared_ptrs instead of raw pointers.

The danger of raw pointers is that peripherals are exposed to outside classes. For example, whenever a key is pressed, InputManager will ask g_peripherals for all HID peripherals, and then ask each HID peripheral for a translated key. This is dangerous if a CEC adapter is unplugged while the key is being translated, because the peripheral will be destroyed before InputManager asks it for the translated key.

As a second example, the settings system enables/disables the rumble setting by asking for all joysticks, and then checking each joystick for rumble. If the joystick is unplugged at the exact instance of these checks, then things blow up.

That's all the examples in the current code. Everything else I've added in the past, including the controller dialog, is safe because the reasoning occurs inside the peripherals subsystem. Still, better safe then sorry, especially when RetroPlayer is merged and we allocate ports to controllers in the games code instead of the peripherals code.

The third commit removes the ability to have sub devices. I'm not sure what the intended purpose of this was, but it's not used currently, and the ownership of sub devices should be reevaluated after the change in this PR.

Broken out from #10630

### DIFF
--- a/xbmc/input/KeyboardStat.cpp
+++ b/xbmc/input/KeyboardStat.cpp
@@ -60,13 +60,13 @@ void CKeyboardStat::Initialize()
 
 bool CKeyboardStat::LookupSymAndUnicodePeripherals(XBMC_keysym &keysym, uint8_t *key, char *unicode)
 {
-  std::vector<CPeripheral *> hidDevices;
+  PeripheralVector hidDevices;
   if (g_peripherals.GetPeripheralsWithFeature(hidDevices, FEATURE_HID))
   {
-    for (unsigned int iDevicePtr = 0; iDevicePtr < hidDevices.size(); iDevicePtr++)
+    for (auto& peripheral : hidDevices)
     {
-      CPeripheralHID *hidDevice = (CPeripheralHID *) hidDevices.at(iDevicePtr);
-      if (hidDevice && hidDevice->LookupSymAndUnicode(keysym, key, unicode))
+      std::shared_ptr<CPeripheralHID> hidDevice = std::static_pointer_cast<CPeripheralHID>(peripheral);
+      if (hidDevice->LookupSymAndUnicode(keysym, key, unicode))
         return true;
     }
   }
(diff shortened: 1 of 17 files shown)


garbear:
jenkins build this please

garbear:
I don't blame you, I took on the input refactoring challenge for controllers and I'm still fixing issues almost two years later!

Look at the removed code, subdevices are accessed a few times but there's no way to actually add subdevices. The only time m_subDevices is modified is when it's cleared. Also subdevices aren't referenced anywhere outside of CPeripheral. We should either connect the orphaned code to make it reachable, or remove it.

opdenkamp:
I was talking about the code that polls lirc, keyboard, peripherals, etc. for input, which is called from the main loop, which I didn't find very nice, which I didn't want to address at the time when I wrote the peripherals stuff ;) That's why there's a subdevice, which I didn't use.

garbear:
Yea, I started down that path and it was like chewing class so I added a dedicated thread for joystick event handling: https://github.com/xbmc/xbmc/tree/master/xbmc/peripherals (EventScanner.cpp/h).

It defaults to polling at 60fps, but another system can request a handle for a different poll rate that the event thread will use as long as the handle is held. If this handle strategy sounds familiar, it's cause i loosely ripped it from some code you wrote for a project a while back, so thanks for that.

So what should we do with subdevices? I don't like unconnected code hanging about. What if we merge this PR, then I create a branch (and maybe even open a WIP PR so it's more discoverable) containing the removed code? That way, you or someone with time will have a launching point (and a fancy new event thread) for implementing this feature in the future.

garbear:
@opdenkamp what do you want to do about subdevices? Remove and save the code for later, or keep the code?

opdenkamp:
I mixed up what I intended to use these for earlier in https://github.com/xbmc/xbmc/pull/10740#issuecomment-255438310
Implemented the device features for that kind of stuff.

Subdevices were meant for composite USB devices. I've used a hack now to make the current code work for the USB-CEC adapter, which uses a composite USB device: https://github.com/xbmc/xbmc/blob/master/xbmc/peripherals/bus/linux/PeripheralBusUSBLibUdev.cpp#L148

To get this type of device supported without about hacks or abusing "features", you're going to need subdevices.

garbear:
ok, i'll leave subdevices in. jenkins build this please

PR #10750, opened 2016-10-23
[AirTunes] - fix the "IsRunning" method.

Memphiz:
The thread only runs during playback so we can't use it as the indicator for the service itself.

This fixes the problem that the airtunes server doesn't stop when you deactivate it in the Settings (because IsRunning always returned false).


Memphiz:
jenkins build this please

PR #10751, opened 2016-10-23
fix skin settings loading on skin reaload

tamland:
Due to 'new' skin being loaded before the old is unloaded (which saves the settings), the new settings isn't being properly loaded until the next reload.

cc @BigNoid


BigNoid:
Working as intended, thx!

PR #10752, opened 2016-10-23
[WIP][addons] Fix broken libKODI_audioengine

AchimTuran:
Some header files for this library where missing. 

Currently I can't get information from the sink, which I will try to fix in the next days.


hudokkow:
Also needs an update in https://github.com/xbmc/xbmc/blob/master/project/cmake/scripts/linux/Install.cmake#L177-L187

AchimTuran:
Thx, I will also add the header files there.

AlwinEsch:
Have added the cmake related header install.

PR #10753, opened 2016-10-23
[WIP] Win32 build tools

AchimTuran:
This PR will include wrapper scripts to simplify add-on development with Visual Studio. Furthermore it adds [jom](https://wiki.qt.io/Jom) to for faster binary add-on builds. It also adds [ninja](https://ninja-build.org/), which we can use to speed up our Windows build with jenkins.

During development I also noticed that multiple calls to `"%VS140COMNTOOLS%..\..\VC\bin\vcvars32.bat"` breaks an open command line window. To workaround I inserted an if statement.


Montellese:
Please don't check in exe's into git. They should be downloaded before or during build like all other build tools.

afedchin:
I think binaries in repo it's a bad idea.

AchimTuran:
I fully agree with you @Montellese and @afedchin.

Could you mention me a place where this can be done properly?

tools/buildsteps/win32/bootstrap-addons.bat
```diff
@@ -32,6 +35,7 @@ IF "%WORKDIR%" == "" (
 )
 
 rem setup some paths that we need later
+SET BUILD_ON_CORES=8
```
Paxxi: There is an environment variable for it in Windows but don't remember the name now

Paxxi:
On my phone currently so won't comment about the code. 
Why include both the zip and exe? Also can we drop the changelogs?

tools/buildsteps/win32/bootstrap-addons.bat
AchimTuran: Ah didn't know that. You mean %NUMBER_OF_PROCESSORS%?
Which is discussed [here](http://stackoverflow.com/questions/22919076/find-number-of-cpus-and-cores-per-cpu-using-command-prompt).

Paxxi: Yep that's the one

AchimTuran:
As I'm not very familiar with our current build tools, I committed them all to show from where I get them.
Sure we can drop the changeling if it isn't needed. That's why I added WIP to the title.

AchimTuran:
@afedchin @Montellese 
Would it be ok for you to download and extract jom and ninja during the run of [DownloadBuildDeps.bat](https://github.com/xbmc/xbmc/blob/master/project/BuildDependencies/DownloadBuildDeps.bat)?

Montellese:
@AchimTuran yes that's how it should be done. And the downloaded files should be extracted to the subdirectories in `project/BuildDependencies/` like we already do for `wget.exe`, `doxygen.exe` et al.

AchimTuran:
@Montellese 
> @AchimTuran yes that's how it should be done. And the downloaded files should be extracted to the  subdirectories in project/BuildDependencies/ like we already do for wget.exe, doxygen.exe et al.

Then we should add ninja and jom to http://mirrors.kodi.tv right? Do know who could do this?

Montellese:
Unfortunately I don't know the new credentials for our FTP server. But I'm sure @MartijnKaijser does.

Rechi:
@AchimTuran is this still something we want?

AchimTuran:
I don't know if it is something what we will need. These scripts helped me a lot to properly build binary add-ons for windows. You can see them more like a workaround. Our build system for binary add-ons needs a refactoring.

MartijnKaijser:

PR #10757, opened 2016-10-23
AMC(surface) is not a BYPASS

koying:
Cut the crap about amcs being a bypass.
Also cut the crap about having to wait for vsync
## Motivation and Context

Cut the crap ;)
## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.cpp
@@ -209,6 +209,11 @@ void CDVDMediaCodecInfo::Validate(bool state)
   m_valid = state;
 }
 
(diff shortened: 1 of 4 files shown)


MartijnKaijser:
Already build and tested on Android (also only affects android specific code) and no negative affects

mazeyx:
seems to use alot of cpu again with this upto 60% from 30%. also when you backout when video is playing you dont see the video in the background anymore.

MartijnKaijser:
I can't reproduce the CPU usage at all. Both the Nexus Player and the NVIDIA SHIELD use the same before and after. Are you sure this PR is the cause as in try nightly before this one?
Only the playing in background is correct as i never use that at all so didn't check.

mazeyx:
@MartijnKaijser just tried the test build which fixed the background image, seems the cpu load is down to 40% from 60% now much more acceptable, yeah the nightly before yesterday was giving me 30% cpu, i guess with the logic change my bravia cpu is less powerful then both those boxes so i notice it. 40% is ok if its a bug fix, 60% was abit too high and reminded me of the old days lol thanks

PR #10759, opened 2016-10-23
Forced rescan of music file tags for v17

DaveTBlake:
Add a forced rescan of tags to capture the new tag data that is processed when upgrading to v17 from earlier versions.

Users are not going to many of see the improvements to the music library when they upgrade to V17 from earlier versions unless they rescan the additional tags that v17 processes from their music files. However there is no easy way for users to initiate a rescan of exsisting files without either a) dropping that source and losing their playcount etc. data or b) "touching" the music files so that Kodi thinks they have changed. A library update started by the user will only check the hash table, it then only scans the tags of those that have a different name or timestamp. If users retag for the new features then this is fine, many will have the tags already in their music files but fail to benefit when using Kodi. 

The user will be prompted to rescan the tags from all their music files the first time they enter the music library after upgrading from an older version of Kodi. They can opt not to do so at that time, but they will be asked again the next time they enter the music library. This prompting will repeat until scanning is started.  This scan could be interrupted if the user does not want to completely scan all their files, and Kodi will not prompt again.

Rather than automatically fetch additional data for artists and albums (from online sources or NFO files), if the user has "fetch additional information during update" enabled then they will also be asked if they want to fetch ithis information as well. 

This can take some time, and it is something that the user can do at a later date using "Query Information for all" from the context menu of the artist or album nodes. It also may be unnecessary - most of the additional information previously scraped will still be in the libaray, only the album genre(s), year, compilation, label and type will revert to what is determined from the music  file tags. 

To test this install Kodi as an upgrade from an earlier version (the existence of an earlier MyMyusicXX.db than 60), or modiy the guisettings.xml to trick it into rescanning by

```
    <mymusic>
        <needsupdate>60</needsupdate>
    </mymusic>
```

Plenty of scope to discuss the best words to use. An alternative  3 button dialog "scan now", "scan later", "never scan" is also an option. But we really need something in place before release.

@MartijnKaijser and @jjd-uk after our discussion. @phate89 could be combined with testing of  #10731

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -900,11 +900,13 @@ msgctxt "#203"
 msgid "Plot outline"
 msgstr ""
 
+#. label of a smartplaylist filter option to filter for albums on compilations
 #: xbmc/playlists/SmartPlaylist.cpp
 msgctxt "#204"
 msgid "Compilation"
 msgstr ""
 
+#. label of a smartplaylist filter option to filter for songs or albums on votes for rating
 #: xbmc/playlists/SmartPlaylist.cpp
 msgctxt "#205"
 msgid "Votes"
@@ -3477,12 +3479,7 @@ msgctxt "#799"
 msgid "Library update"
 msgstr ""
 
-#: xbmc/music/windows/GUIWindowMusicBase.cpp
-msgctxt "#800"
-msgid "Music library needs to rescan tags from files. Would you like to scan now?"
-msgstr ""
-
-#empty string with id 801
+#empty strings from id 800 to 801
 
 #. Label of progess bar which shows the available/total space of pvr backend
 #: xbmc/pvr/PVRGUIInfo.cpp
@@ -14820,6 +14817,7 @@ msgid "Style"
 msgstr ""
 
 #. Concert composer of the music (if present)
+#: Label of ListItem.Property(Role.Composer)
 #: addons/skin.estuary/1080i/DialogPVRRadioRDSInfo.xml
 msgctxt "#29903"
 msgid "Composer"
@@ -14832,6 +14830,7 @@ msgid "Artist"
 msgstr ""
 
 #. Conductor of the classic music (if present)
+#: Label of ListItem.Property(Role.Conductor)
 #: addons/skin.estuary/1080i/DialogPVRRadioRDSInfo.xml
 msgctxt "#29905"
 msgid "Conductor"
@@ -15274,49 +15273,49 @@ msgctxt "#29985"
 msgid "If traffic advisory is send from RDS, volume is increased"
 msgstr ""
 
-#. Music role category
+#. Node title for music artist role, plural of #38036
 #: system/library/music/musicroles/Remixers.xml
 msgctxt "#29987"
 msgid "Remixers"
 msgstr ""
 
-#. Music role category
+#. Node title for music artist role, plural of #38037
 #: system/library/music/musicroles/Arrangers.xml
 msgctxt "#29988"
 msgid "Arrangers"
 msgstr ""
 
-#. Music role category
+#. Node title for music artist role, plural of #29903
 #: system/library/music/musicroles/Composers.xml
 msgctxt "#29989"
 msgid "Composers"
 msgstr ""
 
-#. Music role category
+#. Node title for music artist role, plural of #29905
 #: system/library/music/musicroles/Conductors.xml
 msgctxt "#29990"
 msgid "Conductors"
 msgstr ""
 
-#. Music role category
+#. Node title for music artist role, plural of #38040
 #: system/library/music/musicroles/DJ.xml
 msgctxt "#29991"
 msgid "DJ Mixers"
 msgstr ""
 
-#. Music role category
+#. Node title for music artist role, plural of #38035
 #: system/library/music/musicroles/Lyricists.xml
 msgctxt "#29992"
 msgid "Lyricists"
 msgstr ""
 
-#. Music role category
+#. Node title for music artist role, plural of #38034
 #: system/library/music/musicroles/Orchestras.xml
 msgctxt "#29993"
 msgid "Orchestras"
 msgstr ""
 
-#. Roles from music contributor
+#. Title for music artist roles node
 #: system/library/music/musicroles/index.xml
 msgctxt "#29994"
 msgid "Roles"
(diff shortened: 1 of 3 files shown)


addons/resource.language.en_gb/resources/strings.po
```diff
@@ -3482,7 +3482,10 @@ msgctxt "#800"
 msgid "Music library needs to rescan tags from files. Would you like to scan now?"
 msgstr ""
 
-#empty string with id 801
+#: xbmc/music/windows/GUIWindowMusicBase.cpp
+msgctxt "#801"
+msgid "Fetch additional information for albums and artists?[CR]This could take some time so you may prefer to do this later"
```
MartijnKaijser: Please don't use line breaks as translations might need more space anyway.

MartijnKaijser: Also do not use strings below 10000. If possible also move the other one to a higher number cause we need to get rid of the lower ID

DaveTBlake: OK, any band allocated to anything specific?  Or just pick the next number at the end?
I just put the new message next to the related old one because there was a gap.  Dotting text about seems really confusing to me

DaveTBlake:
Have moved message strings to higher ids as requested. 

@da-anda  while editing strings.po I have added some better contextual help for translators on a few music related strings that have been added recently.

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -900,11 +900,13 @@ msgctxt "#203"
 msgid "Plot outline"
 msgstr ""
 
+#. label of a smartplaylist filter option to filter for albums on compilations
 #: xbmc/playlists/SmartPlaylist.cpp
 msgctxt "#204"
 msgid "Compilation"
 msgstr ""
 
+#. label of a smartplaylist filter option to filter for songs or albums on votes for rating
```
da-anda: votes in terms of the value of the user rating?

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -3477,10 +3479,7 @@ msgctxt "#799"
 msgid "Library update"
 msgstr ""
 
-#: xbmc/music/windows/GUIWindowMusicBase.cpp
-msgctxt "#800"
-msgid "Music library needs to rescan tags from files. Would you like to scan now?"
-msgstr ""
+#empty string with id 800
 
 #empty string with id 801
```
MartijnKaijser: #empty strings from id 800 to 801

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -19439,46 +19440,54 @@ msgid "Role"
...
+#: Label of ListItem.Property(Role.Mixer)
 msgctxt "#38041"
 msgid "Mixer"
 msgstr ""
 
-#. Missing artist name
+#. Name used when artist name is missing from music file tags
 #: music/MusicDatabase.cpp
 msgctxt "#38042"
 msgid "[Missing]"
```
da-anda: I know this label didn't change, but wouldn't "unknown" be more "user friendly"? I mean, it's technically missing from the tags, but I'm not a friend of using "technical" terminology in the UI, especially in non-technical contexts. But maybe I understood the usecase of this label wrong

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -19496,7 +19505,7 @@ msgctxt "#38044"
 msgid "Song & album artists"
 msgstr ""
 
-#. Title all contributors node (any role)
+#. Title all contributors node (any role e.g. composer, piano, producer, egnineer)
```
da-anda: typo in `egnineer`

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -19508,7 +19517,21 @@ msgctxt "#38046"
...
+#. Question when forced rescan of music tags needed after upgrade from previous version of Kodi
+#: xbmc/music/windows/GUIWindowMusicBase.cpp
+msgctxt "#38060"
+msgid "Music library needs to rescan tags from files. Would you like to scan now?"
+msgstr ""
+
+#. Extra scraping question on upgrade from previous version 
+#: xbmc/music/windows/GUIWindowMusicBase.cpp
+msgctxt "#3861"
+msgid "Fetch additional information for albums and artists?This could take some time so you may prefer to do this later"
```
da-anda: whitespace missing after `?`

da-anda:
Thanks for adding contextual comments - much appreciated

addons/resource.language.en_gb/resources/strings.po
DaveTBlake: When importing an album rating from an online source you can also get a count of how many people rated the album, that is "votes". I think, as I didn't add and don't use, this bit of music functionality.

addons/resource.language.en_gb/resources/strings.po
DaveTBlake: The idea is to stick out, as it indicates an error in tagging etc. It could happen when upgrading library from early versions, but mostly should not occur. It means that the artist name tag was blank/missing. In comparison an artist could be tagged as "unknown" or "anonymous" etc. that is not the same as a missing tag.  There was much talk of both forum and the PR that implemented this that chose "[missing]", here I'm just making the context clearer (or trying to!).

da-anda: thanks for the background info

addons/resource.language.en_gb/resources/strings.po
da-anda: ah ok, so really votes from the metadata sites. Just wasn't sure as I was not aware of any voting capability or info in Kodi, but haven't used music section lately. Sorry for the noise.

MartijnKaijser:
If i hit NO it will keep asking me to rescan which is quite irritating :) There's no way to give it some memory to not ask again 5 seconds later?

DaveTBlake:
It asks when you reenter the music library having left it, but you should be able to browse around the library or leave and stay away without any reminder.  Do you go in and out the library that quickly? Or is something else happening?

MartijnKaijser:
Enter library, pick an album, play it, get out of library, later go back to library to queue another....

DaveTBlake:
>...later go back to library to queue another.... 

The key is "later". Pop in and out the music library and yes the current implementation will pester you. I'm just not sure that the pestering is bad?

Montellese:
IMO if the user selects NO we shouldn't ask him again. He can still manually update/refresh his library if he changes his mind or not?

MartijnKaijser:
@Montellese nope he can't. Update library just skips the file rescan as the hash didn't change. 
I also suggest to just do it on next library update if the user didn't answer YES.
Adding an this to the settings menu seems bit overdoing it as it's just a one time rescan in the rest of v17 usage.

DaveTBlake:
@Montellese as @MartijnKaijser says, this scan ignores the hash values so that tags Kodi previously didn't process can be scanned from unchanged music files. I did suggest adding this kind of scan as a user option, but I can see that isn't necessary.

Since this is an only on upgrade from earlier version event I was hoping to keep it as simple as possible without taking away the user choice completely. Sure I can complicate it with a delay check of some kind, or with 3 options (yes, no, or never) but I want the team to be sure they want that.

da-anda:
> I also suggest to just do it on next library update if the user didn't answer YES.

sounds like the most elegant solution IMO

DaveTBlake:
>    I also suggest to just do it on next library update if the user didn't answer YES.
> sounds like the most elegant solution IMO

Could be really nasty though, especially if you have a large library :(
You have a new album on your drive, so you hit library update to add it thinking that won't take long only to find Kodi scanning for hours from scratch. Most likely result is an abort, then the user never sees the added facilities.

IMO pestering is OK, but just doing it without user consent is not.

MartijnKaijser:
if you abort the scan it will pick up next time where it left off

DaveTBlake:
>if you abort the scan it will pick up next time where it left off

That works for updated files, but the forced scan has no way to know if it completed. Stop a forced scan and you never get another one.

DaveTBlake:
Can we go with this, or do we have a clear alternative for me to implement?

MartijnKaijser:
We'll go with this and see what the users think.
Jenkins build and merge

MartijnKaijser (approved):

MilhouseVH:
So I use MySQL for MyMusic, but every device has now forced a rescan (and they're all on identical versions of Kodi - RPi1, RPi2 and Generic LibreELEC builds). Three devices, one MySQL database, three forced rescans. Is this expected?

A forced rescan on the NUC wasn't so bad. Then I tried to access Music on the RPi2 - same MySQL database as the NUC - and it too wanted a forced rescan even though I'd already done this on the NUC. The RPi3 was also fairly tolerable. But now I'm trying to access Music on the RPi1 and it also wants a forced rescan, surely this isn't right? I really don't fancy a forced rescan on an RPi1...

DaveTBlake:
>Three devices, one MySQL database, three forced rescans. Is this expected?

That is accident not design, I simply didn't consider the client server setup.  Immediate workround is to let scanning on subsequent devices start and then stop it (from the side blade).

The prompt to rescan is using a vaue in GUIsettings.xml, of course each device has its own hence the behaviour is to be expected, but obviously isn't desirable.

Montellese:
This is simply one of the drawbacks that come with the hack of using MySQL as a central database for multiple Kodi installations.

PR #10761, opened 2016-10-23
Fade the current controller while mapping its buttons

garbear:
This keeps the controller button focused (albeit faded) so that the user can know which controller they're mapping at all times.

screen shot 2016-10-23 at 2 34 18 pm

Broken out from #10630


garbear:
jenkins build this please

PR #10763, opened 2016-10-24
instance id

tamland:
Alternative to #10543 which uses uuid4. This ensures that the id is not reproducible and trackable by others.

Also, this makes it easier for user to reset the ID by simply deleting it, in contrast to a hardware IDs which are much harder to change.

This should be more than good enough for our purposes.


FernetMenta:
+1

Rechi:
@tamland this needs a rebase

tamland:
On second though: if we ever decide we want this, we can just do it in the add-on. There's not really any need for it to be in core

PR #10765, opened 2016-10-25
Fix 1 error and 5 compiler warnings

garbear:
This PR silences some compiler warnings, and also fixes an error in CVideoInfoScanner where an int parameter was used as a string.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -252,60 +252,55 @@ using KODI::MESSAGING::HELPERS::DialogResponse;
 //extern IDirectSoundRenderer* m_pAudioDecoder;
 CApplication::CApplication(void)
   : m_pPlayer(new CApplicationPlayer)
+  , m_iScreenSaveLock(0)
+  , m_bPlaybackStarting(false)
+  , m_ePlayState(PLAY_STATE_NONE)
+  , m_confirmSkinChange(true)
+  , m_ignoreSkinSettingChanges(false)
   , m_saveSkinOnUnloading(true)
   , m_autoExecScriptExecuted(false)
+  , m_bScreenSave(false)
+  , m_bInhibitIdleShutdown(false)
+  , m_dpms(nullptr)
+  , m_dpmsIsActive(false)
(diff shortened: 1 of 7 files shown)


xbmc/Application.cpp
```diff
@@ -252,60 +252,55 @@ using KODI::MESSAGING::HELPERS::DialogResponse;
 //extern IDirectSoundRenderer* m_pAudioDecoder;
 CApplication::CApplication(void)
   : m_pPlayer(new CApplicationPlayer)
+  , m_iScreenSaveLock(0)
```
akva2: i'd consider using class initialization instead of this nightmare init list. it's just asking for never-ending order issues later imo.

garbear: The list is only a nightmare because the class is poorly designed. I find initialization lists more maintainable because the order is enforced, and as a result it's easier to catch variables that go uninitialized.

akva2: just so we are clear, i mean initialization in the header, ```bool member = false;``` style. order is just as guaranteed (by declaration), but if i add something, i do not have to spend time finding where in the init list it fits. but to each his own, was just a suggestion.

akva2: that is how i do it at least.

xbmc/addons/GUIDialogAddonInfo.cpp
```diff
@@ -144,7 +144,7 @@ bool CGUIDialogAddonInfo::OnMessage(CGUIMessage& message)
           CGUIMessage msg(GUI_MSG_ITEM_SELECTED, GetID(), iControl);
           OnMessage(msg);
           int start = msg.GetParam1();
-          if (start >= 0 && start < m_item->GetAddonInfo()->Screenshots().size())
+          if (start >= 0 && start < static_cast<int>(m_item->GetAddonInfo()->Screenshots().size()))
```
FernetMenta: why not making start size_t?

garbear:
jenkins build and merge

PR #10769, opened 2016-10-25
SeekHandler: handle the seeking with numeric input.

afedchin:
this returns the seeking with numpad after #10706 

@FernetMenta ping

### DIFF
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -620,6 +620,18 @@ const infomap player_param[] =   {{ "art",              PLAYER_ITEM_ART }};
 ///     used (xx) will return AM/PM. Also supported: (hh:mm)\, (mm:ss)\,
 ///     (hh:mm:ss)\, (hh:mm:ss).
 ///   }
+///   \table_row3{   <b>`Player.StartTime`</b>,
+///                  \anchor Player_StartTime
+///                  _string_,
+///     Time playing media began
+///   }
+///   \table_row3{   <b>`Player.StartTime(format)`</b>,
+///                  \anchor Player_StartTime_format
+///                  _string_,
+///     Shows hours (hh)\, minutes (mm) or seconds (ss). When 12 hour clock is
+///     used (xx) will return AM/PM. Also supported: (hh:mm)\, (mm:ss)\,
+///     (hh:mm:ss)\, (hh:mm:ss).
+///   }
 ///   \table_row3{   <b>`Player.SeekNumeric`</b>,
 ///                  \anchor Player_SeekNumeric
 ///                  _string_,
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -620,6 +620,18 @@ const infomap player_param[] =   {{ "art",              PLAYER_ITEM_ART }};
 ///     used (xx) will return AM/PM. Also supported: (hh:mm)\, (mm:ss)\,
 ///     (hh:mm:ss)\, (hh:mm:ss).
 ///   }
+///   \table_row3{   <b>`Player.SeekNumeric`</b>,
+///                  \anchor Player_SeekNumeric
+///                  _string_,
+///     Time to which the user is seeking via numeric keys. 
+///   }
+///   \table_row3{   <b>`Player.SeekNumeric(format)`</b>,
+///                  \anchor Player_SeekNumeric_format
+///                  _string_,
+///     Shows hours (hh)\, minutes (mm) or seconds (ss). When 12 hour clock is used
+///     (xx) will return AM/PM. Also supported: (hh:mm)\, (mm:ss)\, (hh:mm:ss)\,
+///     (hh:mm:ss).
+///   }
 /// \table_end
 ///
 /// -----------------------------------------------------------------------------
@@ -632,7 +644,8 @@ const infomap player_times[] =   {{ "seektime",         PLAYER_SEEKTIME },
                                   { "time",             PLAYER_TIME },
                                   { "duration",         PLAYER_DURATION },
                                   { "finishtime",       PLAYER_FINISH_TIME },
-                                  { "starttime",        PLAYER_START_TIME}};
+                                  { "starttime",        PLAYER_START_TIME },
+                                  { "seeknumeric",      PLAYER_SEEKNUMERIC } };
 
 /// \page modules__General__List_of_gui_access
 /// \section modules__General__List_of_gui_access_Weather Weather
@@ -8000,6 +8013,16 @@ std::string CGUIInfoManager::GetMultiInfoLabel(const GUIInfo &info, int contextW
       return "-" + strSeekSize;
     if (seekSize > 0)
       return "+" + strSeekSize;
+  }
+  else if (info.m_info == PLAYER_SEEKNUMERIC)
+  {
+    if (!CSeekHandler::GetInstance().HasTimeCode())
+      return "";
+    int seekTimeCode = CSeekHandler::GetInstance().GetTimeCodeSeconds();
+    TIME_FORMAT format = (TIME_FORMAT)info.GetData1();
+    if (format == TIME_FORMAT_GUESS && seekTimeCode >= 3600)
+      format = TIME_FORMAT_HH_MM_SS;
+    return StringUtils::SecondsToTimeString(seekTimeCode, format);
   }
   else if (info.m_info == PLAYER_ITEM_ART)
   {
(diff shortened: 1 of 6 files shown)


xbmc/utils/SeekHandler.cpp
```diff
@@ -298,9 +307,123 @@ bool CSeekHandler::OnAction(const CAction &action)
...
+    {
+      // rotate around
+      for (int i = 0; i < 5; i++)
+        m_timeCodeStamp[i] = m_timeCodeStamp[i + 1];
+      m_timeCodeStamp[5] = remote - REMOTE_0;
+    }
+   }
+ }
+
+int CSeekHandler::GetTimeCodeAmount() const
```
xhaggi: IMO we should rename this method to `GetTimeCodeSeconds()` to be clear what it does return.

xbmc/utils/SeekHandler.cpp
```diff
@@ -298,9 +307,123 @@ bool CSeekHandler::OnAction(const CAction &action)
...
+    int s = tot % 100; tot /= 100;
+    int m = tot % 100; tot /= 100;
+    int h = tot % 100;
+
+    return h * 3600 + m * 60 + s;
+  }
+  return 0;
+}
+
+bool CSeekHandler::GetTimeCodeStamp(std::string *strDispTime /*= nullptr*/) const
```
xhaggi: and instead of doing all formatting magic here, should we use `CDateTime` instead?

```
CDateTime time = CDateTime(static_cast<time_t>(GetTimeCodeSeconds()));
return time->GetAsLocalizedTime("HH:MM:SS");
```

xhaggi:
@afedchin thank you for jumping on the train.

afedchin:
@ronie ping. new info label added.

xbmc/video/windows/GUIWindowFullScreen.cpp
```diff
@@ -104,6 +105,18 @@ CGUIWindowFullScreen::~CGUIWindowFullScreen(void)
 
 bool CGUIWindowFullScreen::OnAction(const CAction &action)
 {
+  if (CSeekHandler::GetInstance().HasTimeCode() && action.GetButtonCode())
```
afedchin: still don't find a solution for this hack.

xbmc/utils/SeekHandler.cpp
afedchin: i'm not sure, because as an user I'm expecting to see what I'm typing.

xbmc/utils/SeekHandler.cpp
```diff
@@ -298,9 +307,120 @@ bool CSeekHandler::OnAction(const CAction &action)
...
+    int s = tot % 100; tot /= 100;
+    int m = tot % 100; tot /= 100;
+    int h = tot % 100;
+
+    return h * 3600 + m * 60 + s;
+  }
+  return 0;
+}
+
+std::string CSeekHandler::GetTimeCodeStamp() const
```
xhaggi: sorry for being too picky but what do you think about renaming it to `GetTimeCodeAsString()` like we do in CDateTime.

xbmc/GUIInfoManager.cpp
```diff
@@ -8001,6 +8002,10 @@ std::string CGUIInfoManager::GetMultiInfoLabel(const GUIInfo &info, int contextW
     if (seekSize > 0)
       return "+" + strSeekSize;
   }
+  else if (info.m_info == PLAYER_SEEKNUMERIC)
+  {
+    return CSeekHandler::GetInstance().GetTimeCodeStamp();
```
xhaggi: it should be possible for the skinner to specify a different time format.

```
int seekTimeCode = CSeekHandler::GetInstance().GetTimeCodeSeconds();
if (format == TIME_FORMAT_GUESS && seekTimeCode >= 3600)
    format = TIME_FORMAT_HH_MM_SS;
StringUtils::SecondsToTimeString(seekTimeCode, (TIME_FORMAT)info.GetData1())
```

xbmc/video/windows/GUIWindowFullScreen.cpp
xhaggi: don't care about that IMO. The reason for it is to have a different button mapped to select. IMO we should load the default instead of that virtual window.

xbmc/GUIInfoManager.cpp
```diff
@@ -632,7 +632,8 @@ const infomap player_times[] =   {{ "seektime",         PLAYER_SEEKTIME },
                                   { "time",             PLAYER_TIME },
                                   { "duration",         PLAYER_DURATION },
                                   { "finishtime",       PLAYER_FINISH_TIME },
-                                  { "starttime",        PLAYER_START_TIME}};
+                                  { "starttime",        PLAYER_START_TIME },
+                                  { "seeknumeric",      PLAYER_SEEKNUMERIC } };
```
xhaggi: should we use `seektimecode` instead.

xbmc/utils/SeekHandler.cpp
xhaggi: if we drop that method ignore my comment 😄

xbmc/video/windows/GUIWindowFullScreen.cpp
FernetMenta: we do care about this because this code has to leave CGUIWindowFullScreen

xbmc/utils/SeekHandler.cpp
afedchin: it's dropped

xbmc/GUIInfoManager.cpp
afedchin: I think `seeknumeric` is saying for itself

xhaggi: Okay, the rest looks good. thanks 👍

xbmc/guilib/GUIWindowManager.cpp
```diff
@@ -1397,13 +1397,19 @@ int CGUIWindowManager::GetActiveWindowID()
     // check for LiveTV and switch to it's virtual window
     else if (g_PVRManager.IsStarted() && g_application.CurrentFileItem().HasPVRChannelInfoTag())
       iWin = WINDOW_FULLSCREEN_LIVETV;
+    // special casting for numeric seek
```
garbear: casting or casing?

afedchin: casing ofc.

afedchin:
@ronie @phil65  @HitcherUK , guys, please someone adapt estuary for these changes please.

xbmc/utils/SeekHandler.cpp
```diff
@@ -298,9 +307,102 @@ bool CSeekHandler::OnAction(const CAction &action)
...
+    case REMOTE_2:
+    case REMOTE_3:
+    case REMOTE_4:
+    case REMOTE_5:
+    case REMOTE_6:
+    case REMOTE_7:
+    case REMOTE_8:
+    case REMOTE_9:
+    {
+      if (!g_application.CurrentFileItem().IsLiveTV())
```
da-anda: I know it's been like that before, but this IMO looks wrong in two ways:
a) when having a timeshift buffer, seeking to a time code within the buffer might still make sense
b) g_application.m_pPlayer->CanSeek() (which is checked in "onAction" method) should already take care of this

da-anda: ah wait, it's conflicting with channel number input, so it might make sense to keep it, unless channel# input isn't processed upfront

ksooo: ... i'm getting scared. Do we actually have a clue what we're doing here? ;-)

da-anda: well, you pvr guys should know as likely one of you added this special case handling back then ;) I'm not at all into the codebase anymore, just spotted this and thought this is looking weird

afedchin: If I'm not wrong a window handle this upfront, so we can remove this to do seeking for recordings. I can't check this right now, but will do asap.

afedchin: just tested. it works as it should.

ronie:
@afedchin skin part in https://github.com/xbmc/xbmc/pull/10786

afedchin:
@ronie thank you!

if there are no objections, I will merge this tomorrow morning.

jenkins build this please

xhaggi (approved):
+1

xhaggi:
@afedchin I've done some tests while adding the new info label to Xperience1080. The info label works great but it won't jump to the given time. Could someone else confirm if it works?

xhaggi:
@afedchin sorry for the noise i missed that I've to press Enter/OK/Select 😄

PR #10770, opened 2016-10-25
ListItem.Year - prefer to use the year value from the firstaired date

ronie:
ListItem.Year currently returns the year taken from the Premiered date.

for episodes it would be more accurate to use the year from the FirstAired date.

http://forum.kodi.tv/showthread.php?tid=294755

before:
before

after:
after

### DIFF
--- a/xbmc/video/VideoInfoTag.cpp
+++ b/xbmc/video/VideoInfoTag.cpp
@@ -718,11 +718,13 @@ const std::string& CVideoInfoTag::GetDefaultRating() const
 
 const bool CVideoInfoTag::HasYear() const
 {
-  return m_premiered.IsValid();
+  return m_firstAired.IsValid() || m_premiered.IsValid();
 }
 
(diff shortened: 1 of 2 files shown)


Montellese:
Recently I've been wondering why we have premiered and first aired. The latter is only used for episodes and the former isn't used for episodes at all. So in CVideoInfoTag it could be merged into one property.

phil65:
Shouldnt we better drop year entirely from that confluence infoscreen for episodes? A user might wonder what the difference should be between year and premiered...

phate89 (changes requested):
You should also add firstaired to HasYear() because it's used to check if we have a valid year...

ronie:
@phate89 good catch, thx!
should be fixed now.

phate89 (approved):

MartijnKaijser:
Jenkins build this please

PR #10771, opened 2016-10-25
Singledecoder

peak3d:
Support streamchanges on systems with only 1 h/w decoder
## Description

If there is a stream change on systems wich only allow 1 h/w decoder, every stream change leads to toggle between ffmpeg fallback and h/w. This is because a new decoder on streamchange is prepared before it is used. For a short time 2 decoders are up regarding the current implementation.

This PR introduces a hint in DVDStreamInfo::flags wich can tell the decoder, that the Open() call leads from a streamchange and one more decoder is running. If codec information and capabilities are checked, but the decoder fails, we can be quite sure that this is because of resources.

At the time the stream change is applied, it is checked, if the decoder needs to be reopened. 
## Motivation and Context
- Android encrypted media has to play on with mediaCodec, a fallback to ffmpeg is no alternative.
- Adaptive streamchange (inputstream.mpd) would lead to a mix of s/w and h/w decoding.
## How Has This Been Tested?
- create a .strm file with this content:
  #KODIPROP:inputstreamaddon=inputstream.mpd
  https://storage.googleapis.com/wvmedia/clear/h264/tears/tears.mpd
- enable inputstream.mpd
- The first packet from inputstream generates a DMX_GENERAL_STREAMCHANGE.
- Without this PR on machines with limited resources (Odroid-C2 here) the video is played with ffmpeg. Using this PR, its played with mediaCodec.
## Types of change
## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemux.cpp
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemux.cpp
@@ -20,6 +20,29 @@
 
 #include "DVDDemux.h"
 
+unsigned int CDemuxStream::AvDispositionToDemuxFlags(unsigned int disposition)
+{
+  static unsigned int translate[][2]={
+    {AV_DISPOSITION_DEFAULT, FLAG_DEFAULT},
+    {AV_DISPOSITION_DUB, FLAG_DUB},
+    {AV_DISPOSITION_ORIGINAL, FLAG_ORIGINAL},
+    {AV_DISPOSITION_COMMENT, FLAG_COMMENT},
+    {AV_DISPOSITION_LYRICS, FLAG_LYRICS},
+    {AV_DISPOSITION_KARAOKE, FLAG_KARAOKE},
+    {AV_DISPOSITION_FORCED, FLAG_FORCED},
+    {AV_DISPOSITION_HEARING_IMPAIRED, FLAG_HEARING_IMPAIRED},
+    {AV_DISPOSITION_VISUAL_IMPAIRED, FLAG_VISUAL_IMPAIRED},
+    {AV_DISPOSITION_ATTACHED_PIC, FLAG_DISP_ATTACHED_PIC}
+  };
+
+  unsigned int ret(0);
+  for (auto avt : translate)
+    if (disposition & avt[0])
+      ret |= ret[1]);
+
+  return ret;
+}
+
 std::string CDemuxStreamAudio::GetStreamType()
 {
   char sInfo[64] = {0};
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemux.h
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemux.h
@@ -127,18 +127,24 @@ public:
 
   int  changes; // increment on change which player may need to know about
 
-  enum EFlags
-  { FLAG_NONE             = 0x0000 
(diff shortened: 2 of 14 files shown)


FernetMenta:
There is more work needed to get this fully right. Something for v18

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemux.h
```diff
@@ -127,18 +127,20 @@ class CDemuxStream
 
   int  changes; // increment on change which player may need to know about
 
-  enum EFlags
```
peak3d: Using enum for flag values might work, but its IMO not the best choice

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
```diff
@@ -1484,7 +1484,9 @@ CDemuxStream* CDVDDemuxFFmpeg::AddStream(int streamIdx)
 
     stream->source = STREAM_SOURCE_DEMUX;
     stream->pPrivate = pStream;
-    stream->flags = (CDemuxStream::EFlags)pStream->disposition;
```
ace20022: Doesn't you loose all other flags now? Additionally FLAG_DISP_ATTACHED_PIC = 0x0200 collides with AV_DISPOSITION_CLEAN_EFFECTS   0x0200

peak3d: (Coming back to your question: I have searched for all occurrences of the flags field. Only the AV_DISPOSITION_CLEAN_EFFECTS  was explecitely checked on various places.)

ace20022: flags are used all over the place, they are forwarded directly here: https://github.com/peak3d/xbmc/blob/1800fcf24736faf870da4e23ce28c3964019d2a5/xbmc/cores/VideoPlayer/VideoPlayer.cpp#L523

If we want to be safe, we would have to translate all used av flags to our flags

peak3d: just looked at the definition of the AV_DISPOSITION flags, right, didn't knew that the demuxstream EFlags values are already a translation of AV_DISPOSITION, so you're right, the |= is false.

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemux.cpp
```diff
@@ -20,6 +20,29 @@
...
+    {AV_DISPOSITION_FORCED, FLAG_FORCED},
+    {AV_DISPOSITION_HEARING_IMPAIRED, FLAG_HEARING_IMPAIRED},
+    {AV_DISPOSITION_VISUAL_IMPAIRED, FLAG_VISUAL_IMPAIRED},
+    {AV_DISPOSITION_ATTACHED_PIC, FLAG_DISP_ATTACHED_PIC}
+  };
+
+  unsigned int ret(0);
+  for (auto avt : translate)
+    if (disposition & avt[0])
+      ret |= ret[1]);
```
peak3d: @ace20022 : you are really an eagleeye :-) Its fixed... thx!

peak3d:
There are other ways floating around....

PR #10772, opened 2016-10-25
[estuary] fallback to icon in poster view

ronie:
fixes trac #16977

@phil65


phil65:
should be fine, thx. (although probably any non-poster stuff will get stretched / scaled in a not-so-nice way?)

ronie:
it fixes the case were an addon sets content to movies...
if they don't provide a poster sized thumb, it's not our problem :-)

they are scaled.. doesn't look all that bad ;-)

phil65:
ok, green button is yours.

PR #10774, opened 2016-10-25
Parse the id tag only if the uniqueid tag is missing as fallback

phate89:
This avoids to add the default id twice with "unknown" and the right name in case of scrapers that supports backwards compatibility


phate89:
jenkins build this please

PR #10781, opened 2016-10-26
Hide controller profiles in the add-on browser

garbear:
## Motivation and Context

The original reason I used the add-on system for controller profiles was so that installing an emulator would automatically install the required controllers.

However, controller profiles are really internal to the system and, without games, don't need to be shown outside of the controller dialog. We especially don't need to show them as a single node inside another node called "Game add-ons". Talk about confusing.

This PR hides the "Game add-ons" node for the v17 release. As a result, if the user deletes their addon DB, all controllers disappear from the GUI and the user has no way to bring them back. There are two possible solutions:
- Don't disable them in the first place
- Show disabled controller profiles in the controller dialog

I submitted a PR for the first option (#9700) and it was rejected. So this PR includes the second solution.

### DIFF
--- a/xbmc/addons/binary/interfaces/api1/Peripheral/AddonCallbacksPeripheral.cpp
+++ b/xbmc/addons/binary/interfaces/api1/Peripheral/AddonCallbacksPeripheral.cpp
@@ -87,7 +87,7 @@ unsigned int CAddonCallbacksPeripheral::FeatureCount(void* addonData, const char
   unsigned int count = 0;
 
   AddonPtr addon;
-  if (CAddonMgr::GetInstance().GetAddon(controllerId, addon, ADDON_GAME_CONTROLLER))
+  if (CAddonMgr::GetInstance().GetAddon(controllerId, addon, ADDON_GAME_CONTROLLER, false))
   {
     ControllerPtr controller = std::static_pointer_cast<CController>(addon);
     if (controller->LoadLayout())
--- a/xbmc/filesystem/AddonsDirectory.cpp
+++ b/xbmc/filesystem/AddonsDirectory.cpp
@@ -50,7 +50,6 @@ CAddonsDirectory::~CAddonsDirectory(void) {}
 
 const auto CATEGORY_INFO_PROVIDERS = "category.infoproviders";
 const auto CATEGORY_LOOK_AND_FEEL = "category.lookandfeel";
-const auto CATEGORY_GAME_ADDONS = "category.gameaddons";
 
 const std::set<TYPE> dependencyTypes = {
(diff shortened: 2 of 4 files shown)


MartijnKaijser:
@tamland for review

stefansaraev:
this, as a side effect of auto-disabling non-system addons on addons.db deletion, was expected.

one should rather rethink not to auto disable those addons..

tamland:
>This PR hides the "Game add-ons" node for the v17 release. As a result, if the user deletes their addon DB, all controllers disappear from the GUI and the user has no way to bring them back. 

As I've said before I'm disagreeing with the whole "user deletes their addon DB" usecase. It's like complaining about "I deleted my video db and now all my videos are gone" and having to setup everything over again. IMO every special case that is added is going hurt usability and should therefor be avoided as much as possible. So *why should* controller profiles be special here and break out of the whole enable/disable functionality?

There is a third option I've been thinking about doing to help making the re-enabling process bit easier, that works for *all* addons: to show them in the 'get more' dialog together with installable addons.

tamland:
@stefansaraev If you have nothing to contribute, please stop commenting. github is for developers.

stefansaraev:
@tamland 

I did have anything to to contribute, and I did contribute. 

I did not blame you, I did not point a finger at you. please dont act like Its me wanting to intentionaly offend you.

deleting addons db is very real use case, and is very different than deleting music or video db (as scaning for installed addons happens automaticaly and the user has no control over that process), of course you are free to disagree.

about github being for developers, you should look at kodi (and few other projects that resolve around kodi) commit logs / stats.
 
this PR is just another workaround for a wrong design mistake. we'll see more in future. now I have nothing else to contribute and am out.

MartijnKaijser:
Deleting the addons.db is not a use case. It's intentionally breaking something.

stefansaraev:
unless it's already broken. but we discussed this many times, no point to discuss it again if you can't understand why one would be forced to delete addons db.

Montellese:
IMO if there would be a valid reason to manually delete addons db something else went wrong earlier which needs to be fixed. Users should not have to touch any Kodi database at all.

stefansaraev:
my last words here: there may be non kodi-related reasons for corrupted / removed addons db. like e2fsck splitting in few parts and moving them to /lost+found. and more. good luck.

tamland:
At some point we have to trust that the filesystem can store the files it is asked to store. Exactly the same can happen to guisettings and other databases. But we're not going make design decision based on the fact that any file can be corrupted at any time by external sources.

garbear:
No longer an issue after #11571

PR #10782, opened 2016-10-26
libdvd/Makefile: cross-compile, allow to specify the ar command

tgurr:
This commit fixes a build issue on cross environments where the ar command may be prefixed with e.g. the host triplet, like x86_64-pc-linux-gnu-ar and results in a build failure not finding the ar command.
## How Has This Been Tested?

Building kodi-17.0_beta4 locally with the changes applied.
## Types of change
- [ x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/lib/libdvd/Makefile.in
+++ b/lib/libdvd/Makefile.in
@@ -1,4 +1,5 @@
 
+AR=@AR@
 ARCH=@ARCH@
 CC=@CC@
 CXX=@CXX@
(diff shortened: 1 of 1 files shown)


wsnipex:
Thanks, the fix is ok, but we're in the process of deprecating autotools, so you should probably check if a similar fix is needed for the cmake build system in project/cmake.

tgurr:
I came across the error when packaging Kodi 17.0 beta4 (autotools based build) to see if there are any issues which would be nice to have fixed before the final is released. I'm afraid I can't promise that I'll have time to look into the CMake based build any time soon.

wsnipex:
ok
jenkins build this please

koenkooi:
Cmake tends to be less susceptible to issues like this if you're using a toolchain file.

PR #10784, opened 2016-10-26
[deps][fix][win32] Fix embedded fonts in libass

Paxxi:
Fix embedded font loading in libass on Windows
## Description

Implemented embedded font loading for libass when building with DirectWrite
Code can be found at https://github.com/Paxxi/libass it's quite ugly still and needs some cleanup but it's tested and solves the issue
## How Has This Been Tested?

Runtime tested by user that reported the issue
## Types of change

## Checklist:

Closes trac 16864

### DIFF
--- a/project/BuildDependencies/scripts/0_package.list
+++ b/project/BuildDependencies/scripts/0_package.list
@@ -16,7 +16,7 @@ expat-2.2.0-win32-vc140.7z
 freetype-2.6.3-win32-vc140.7z
 giflib-5.1.4-win32-vc140.7z
 jsonschemabuilder-1.0.0-win32-3.7z
-libass-660b2d-win32-vc140.7z
+libass-542975a-win32-vc140.7z
 libbluray-0.9.3-win32-vc140.7z
 libcdio-0.9.3-win32-vc140.7z
 libcec-3.1.0-win32-vc140.7z
--- a/project/BuildDependencies/scripts/0_package.list
+++ b/project/BuildDependencies/scripts/0_package.list
@@ -16,7 +16,7 @@ expat-2.2.0-win32-vc140.7z
 freetype-2.6.3-win32-vc140.7z
 giflib-5.1.4-win32-vc140.7z
 jsonschemabuilder-1.0.0-win32-3.7z
-libass-791d94-win32-vc140.7z
+libass-660b2d-win32-vc140.7z
 libbluray-0.9.3-win32-vc140.7z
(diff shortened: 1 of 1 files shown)


taxigps:
@Paxxi have tested your new test build for ass. it's not crash now. but still have some problem.
1. when have more than one embedded fonts in the ass sub and one font of them installed as window system font, other embedded fonts will using default font arial.ttf. 
2. how to add additional fonts path like `media/Fonts` as we supported before?
by the way, we should get rid of path `players/VideoPlayer/etc/fonts` in our source of kodi if don't need fontconfig any more.

taxigps:
and ass sub will disppear or clipped without this patch (https://github.com/libass/libass/pull/120) when set `subtitle position on screen` to `above video`

Paxxi:
Thanks
1. Is probably because of how the fallback handling is implemented, will try and solve this one. 
2. We still use fontconfig on all other platforms so it should stay. 

I'll add that patch to the next build

Paxxi:
@taxigps new test build here http://mirrors.kodi.tv/test-builds/win32/KodiSetup-20161029-aa07769-libass.exe that includes your patch and hopefully solves the fallback issue as well

taxigps:
test new build. when play, rewind and play again, sometimes can select right font, sometimes can't.
and an other problem, can't load font when embedded font filename use chinese.
how to send my sample mkv to you, an one minute movie clip with embedded chinese ass sub and fonts.

taxigps:
@Paxxi sample file upload to ftp.xbmc.org/upload/Oblivion.2013.ASS_TEST.mkv. 
four font embedded:

Font name | TTF filename | Note
------------ | -------------- | -------------
方正超粗黑简体 | FZCCHJW.TTF | Sometimes be selected right
方正黑体_GBK | FZHTK.TTF | Can't be selected, maybe because the font name?
方正稚艺简体 | FZZHYJW.TTF | Sometimes be selected right
方正粗圆简体 | 方正粗圆简体.TTF | [ass] Error opening font

Paxxi:
thanks @taxigps , I'll look into it and see if I can get it working

Paxxi:
Comparing that sample against v16.1 I think it looks the same. The last font fails to load because Freetype doesn't understand utf-8 on Windows and passes it along straight to fopen trying to open a bogus path.

I will get that bit fixed but until then I'll merge this as it fixes several issues.

Paxxi:
Created track ticket http://trac.kodi.tv/ticket/17057#comment:1 to keep track of it.

PR #10785, opened 2016-10-26
add privacy policy

ronie:
as reported by @MartijnKaijser on trac (http://trac.kodi.tv/ticket/17016) it's mandatory to provide a way for users to read our privacy policy within kodi.

this pr adds this functionality to the systeminfo window.

please check if i got the install path on each platform correct, the file should be available in special://xbmc/
(assuming we want to save it there...)

privacypolicy

### DIFF
--- a/privacy-policy.txt
+++ b/privacy-policy.txt
@@ -1,0 +1,55 @@
+[B]Kodi Privacy Policy[/B]
+[I]2016 Oct 23[/I]
+
+Your privacy is an important factor that the XBMC Foundation and Team Kodi considers in the development of all of our software and services. We are committed to being transparent and open. This Privacy Policy explains generally how we receive information about you, and what we do with that information once we have it.
+
+
+[B]What do we mean by “personal information?”[/B]
+
+For us, “personal information” means information which identifies you, like your name or email address.
+
+Any information that falls outside of this is “non-personal information.”
+
+If we store your personal information with information that is non-personal, we will consider the combination as personal information. If we remove all personal information from a set of data then the remaining is non-personal information.
+
+
+[B]How do we learn information about you?[/B]
+
+We learn information about you when:
+
+    • you give it to us directly (e.g., when you choose to send us logs in the forums);
+    • we collect it automatically through our software and services (e.g., when your Kodi connects with our servers to update add-ons, or in download statistics provided by Google Play);
+    • when we try and understand more about you based on information you’ve given to us (e.g., when we use your platform information provided to the add-on server to figure out which platforms use Kodi the most, to better understand where we should focus our efforts).
+
+
+[B]What do we do with your information once we have it?[/B]
+
+Generally, we use your information to help us provide and improve our software and services for you (e.g., we use a log you send us to figure out why Kodi isn’t playing a video right or why it might have crashed, or we determine how many active users are using each platform in order to determine how to allocate resources per platform).
+
+
+[B]When do we share your information with others?[/B]
+
+    • When we have asked and received your permission to share it.
+    • When we are fulfilling our educational purpose. We sometimes publicly release information to make our software better and foster an open web, but when we do so, we will remove your personal information and try to disclose it in a way that minimizes the risk of you being re-identified. For example, in a blog post we might attempt to analyze from available data how many active users of Kodi existed at the time and share that info with the community.
+    • When the law requires it. To date, this has never happened, and we do not anticipate it happening in the future, as the majority of the information we collect is already almost entirely anonymous and never shows unique user behavior, except in voluntarily submitted logs on our forum.
+    • If our organizational structure or status changes (if we undergo a restructuring, are acquired, or go bankrupt) we may pass your information to a successor or affiliate.
+
+
+[B]How do we store and protect your personal information?[/B]
+
+We are committed to protecting your personal information once we have it. We implement physical, business and technical security measures. Despite our efforts, if we learn of a security breach, we’ll notify you so that you can take appropriate protective steps.
+
+We also don’t want your personal information for any longer than we need it, so we only keep it long enough to do what we collected it for. Once we don’t need it, we take steps to destroy it unless we are required by law to keep it longer. In the case of logs provided to us in the forums, our servers do not store those logs, and we rely on the users submitting those logs to remove the links when they deem appropriate.
+
+
+[B]What else should you know?[/B]
+
+We’re a global organization and our computers are in several different places around the world. We also use service providers whose computers may also be in various countries. This means that your information might end up on one of those computers in another country, and that country may have a different level of data protection regulation than yours. By giving us information, you consent to this kind of transfer of your information. No matter what country your information is in, we comply with applicable law and will also abide by the commitments we make in this privacy policy.
+
+If you are under 13, we don’t want your personal information, and you must not provide it to us. If you are a parent and believe that your child who is under 13 has provided us with personal information, please contact us to have your child’s information removed.
+
+
+[B]What if we change this privacy policy or any of our privacy notices?[/B]
+
+We may need to change this policy and our notices. The updates will be posted online. If the changes are substantive, we will announce the update through Kodi’s usual channels for such announcements such as blog posts and forums. Your continued use of the product or service after the effective date of such changes constitutes your acceptance of such changes. To make your review more convenient, we will post an effective date at the top of the page.
+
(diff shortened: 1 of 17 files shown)


ronie:
todo: read the file only once, on window open (instead of each time the button is focused)

privacy-policy.txt
```diff
@@ -0,0 +1,55 @@
...
+We learn information about you when:
+
+    • you give it to us directly (e.g., when you choose to send us logs in the forums);
+    • we collect it automatically through our software and services (e.g., when your Kodi connects with our servers to update add-ons, or in download statistics provided by Google Play);
+    • when we try and understand more about you based on information you’ve given to us (e.g., when we use your platform information provided to the add-on server to figure out which platforms use Kodi the most, to better understand where we should focus our efforts).
+
+
+[B]What do we do with your information once we have it?[/B]
+
+Generally, we use your information to help us provide and improve our software and services for you (e.g., we use a log you send us to figure out why Kodi isn’t playing a video right or why it might have crashed, or we determine how many active users are using each platform in order to determine how to allocate resources per platform).
```
da-anda: shouldn't it be `sent` instead of `send` in "e.g., we use a log you sent us to figure..."

ronie: i didn't proof-read it ;-)
it's a copy of our policy taken from here: https://kodi.tv/kodi-privacy-policy/

hudokkow: ping @jjd-uk to proof-read it

project/cmake/scripts/linux/Install.cmake
```diff
@@ -132,6 +132,10 @@ install(FILES ${CORE_SOURCE_DIR}/copying.txt
         DESTINATION ${datarootdir}/doc/${APP_NAME_LC}
         COMPONENT kodi)
 
+install(FILES ${CORE_SOURCE_DIR}/privacy-policy.txt
+        DESTINATION ${datarootdir}${APP_NAME_LC}
+        COMPONENT kodi)
+
```
hudokkow: Any reason you didn't used the code block above and install it to ${datarootdir}/doc/${APP_NAME_LC}, same as the other files?

ronie: special://xbmc/ points to ${datarootdir}${APP_NAME_LC}
(i need a special path that's the same across platforms)

privacy-policy.txt
garbear: The document uses simple present tense, so send is correct

privacy-policy.txt
```diff
@@ -0,0 +1,55 @@
...
+
+
+[B]When do we share your information with others?[/B]
+
+    • When we have asked and received your permission to share it.
+    • When we are fulfilling our educational purpose. We sometimes publicly release information to make our software better and foster an open web, but when we do so, we will remove your personal information and try to disclose it in a way that minimizes the risk of you being re-identified. For example, here is a blog post where we attempt to analyze from available data how many active users of XBMC existed at the time and share that info with the community.
```
garbear: I suggest we change:

`For example, here is a blog post where we attempt to analyze...`

to:

`For example, in a blog post we might attempt to analyze...`

Also, XBMC -> Kodi

garbear:
Overall, I think we have a good privacy policy. Good job guys.

privacy-policy.txt
natethomas: Good change. The original language is based on having a link to a blog post, which we obviously can't have here.

ronie: both fixed now

Hedda:
Should this privacy policy be translated via Transifex when shown in the GUI?

http://www.transifex.com/teamxbmc/kodi-main/

If not using English inside Kodi then could expect this to be translated too?

MartijnKaijser:
@Hedda sorry but that would be a really bad idea cause as we noticed in general the translations are quite horribly done. As this is a "legal"  document you cannot trust regular users to translate it.

xbmc/windows/GUIWindowSystemInfo.cpp
```diff
@@ -199,3 +219,16 @@ void CGUIWindowSystemInfo::SetControlLabel(int id, const char *format, int label
...
+{
+  CFile file;
+  XFILE::auto_buffer buf;
+  if (file.LoadFile("special://xbmc/privacy-policy.txt", buf) > 0)
+  {
+    std::string strBuf(buf.get(), buf.length());
+    return strBuf;
+  }
+  else
+    return "";
```
tamland: Please *always* log errors;)

tamland:
I think this should rather be added as an info label (similar to all the other system info). Loading a file in CGUIWindowSystemInfo is going to cause it to lag, and the privacy policy will be coupled to and only viewable in this specific window.

fetzerch:
For android this has to be added to https://github.com/xbmc/xbmc/blob/master/tools/android/packaging/Makefile.in#L111 i think. But I'm not sure which directory it needs to go inside the apk. maybe @koying can help here?

ronie:
@tamland moved the functionality to CSysInfo and made it available as an infolabel.
please check if this is what you've had in mind.

MartijnKaijser:
Now also works on Android. 
Any chance of making it possible to scroll through the text manually?

MartijnKaijser:
@Memphiz @fetzerch could you check on iOS?

Memphiz:
@MartijnKaijser how am i supposed to scroll down in the policy? (just trying it out on osx - its shown - but only the upper lines which fit into the sysinfo rect ...)

ronie:
currently it autoscrolls, i'll look into adding a scrollbar.

Memphiz:
works on ios too

ronie:
cool, all platforms confirmed working then.

MartijnKaijser:
Tested and looks fine to me. Jenkins build this please. 
@tamland was the implementation ok?

xbmc/windows/GUIWindowSystemInfo.cpp
```diff
@@ -81,6 +82,23 @@ bool CGUIWindowSystemInfo::OnMessage(CGUIMessage& message)
...
+      else if (m_section == CONTROL_BT_HARDWARE)
+        SET_CONTROL_HIDDEN(13);
+      else if (m_section == CONTROL_BT_PVR)
+        SET_CONTROL_HIDDEN(13);
+      else if (m_section == CONTROL_BT_POLICY)
+      {
+        SET_CONTROL_LABEL(13, g_infoManager.GetLabel(SYSTEM_PRIVACY_POLICY));
+        SET_CONTROL_VISIBLE(13);
+      }
       return true;
```
phil65: just
```
if (m_section > CONTROL_START && m_section < CONTROL_END)
{
 SET_CONTROL_HIDDEN(13);
}

```
should be fine.
For bonus points, use a #define for the textbox. :)

tamland:
yeah, looks good

xbmc/windows/GUIWindowSystemInfo.cpp
phil65: oops, I meant
`if (m_section >= CONTROL_START && m_section <= CONTROL_BT_PVR)`

ronie:
updated as @phil65 suggested.

Jenkins build this please.

xbmc/windows/GUIWindowSystemInfo.cpp
```diff
@@ -182,6 +191,11 @@ void CGUIWindowSystemInfo::FrameMove()
     SetControlLabel(i++, "%s: %s", 19168, PVR_BACKEND_DELETED_RECORDINGS);  // Deleted and recoverable recordings
     SetControlLabel(i++, "%s: %s", 19025, PVR_BACKEND_TIMERS);
   }
+
+  else if (m_section == CONTROL_BT_POLICY)
+  {
+    SET_CONTROL_LABEL(40, g_localizeStrings.Get(12389));
```
phil65: the 40 should be CONTROL_TB_POLICY  :)

phil65: oops, sorry. false alarm.

ronie:
> works on ios too

jenkins says no

Memphiz:
i guess because Jenkins builds with cmake - i didn't ...

Memphiz:
@ronie i think you Need to add privacy-policy.txt to this file here:

https://github.com/xbmc/xbmc/blob/master/project/cmake/installdata/ios/packaging.txt

and here

https://github.com/xbmc/xbmc/blob/master/project/cmake/installdata/osx/packaging.txt

because on osx it failed "silently" too: 

rsync: link_stat "/Users/Shared/jenkins/workspace/OSX-64/build/privacy-policy.txt" failed: No such file or directory (2)

@fetzerch any idea why osx Fails silently and ios correclty Fails the Job?

ronie:
ok added, thx! 

Jenkins build this please.

PR #10786, opened 2016-10-26
[skins] add seek label to video fullscreen

ronie:
can go in together with https://github.com/xbmc/xbmc/pull/10769

### DIFF
--- a/addons/skin.estouchy/xml/VideoFullScreen.xml
+++ b/addons/skin.estouchy/xml/VideoFullScreen.xml
@@ -51,5 +51,16 @@
 				<label></label>
 			</control>
 		</control>
+		<control type="label" id="1">
+			<description>seek label</description>
+			<posx>50</posx>
+			<posy>30</posy>
+			<width>1180</width>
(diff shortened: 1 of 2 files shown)


afedchin (approved):

addons/skin.estouchy/xml/VideoFullScreen.xml
```diff
@@ -51,5 +51,16 @@
...
+		<control type="label" id="1">
+			<description>seek label</description>
+			<posx>50</posx>
+			<posy>30</posy>
+			<width>1180</width>
+			<height>30</height>
+			<align>left</align>
+			<aligny>center</aligny>
+			<font>font16</font>
+			<label>$INFO[Player.SeekNumeric]</label>
```
xhaggi: @ronie adding the format pattern (hh:mm:ss) to the info label will be look a bit better here.
Without it will show you the time as `1:36`, with it will zeros the rest`00:01:36`.

xhaggi:
No problem :)

PR #10787, opened 2016-10-26
[skin] Add "skin.select" builtin.

BigNoid:
## Description

This adds a new builtin `Skin.Select(<header>, <skin.string>, label1|value1, label2|value2, etc|etc)`
## Motivation and Context

Currently if you want to let a user choose between a list of values for a skin setting, a custom dialog or a lot of conditional onclicks are required. Furthermore you need a variable to display the chosen skin setting.
This makes it possible to present a list of predefined values to use for a setting. It writes the label from the list as `<skin.string>.Label` and the value as `<skin.string>.Value`
This way the chosen option can be presented without using variables.
If  `-` is set as value the `<skin.string>.Value` will be reset.
## How Has This Been Tested?

Runtime tested
## Types of change

## Checklist:

### DIFF
--- a/xbmc/interfaces/builtins/SkinBuiltins.cpp
+++ b/xbmc/interfaces/builtins/SkinBuiltins.cpp
@@ -92,6 +92,50 @@ static int SetAddon(const std::vector<std::string>& params)
   return 0;
 }
 
+/*! \brief Select and choose a value for a skin setting.
+ *  \param params The parameters.
+ *  \details params[0] = Header of the dialog.
+ *  \details params[1] = Names of skin settings.
+ */
+static int SkinSelect(const std::vector<std::string>& params)
+{
+  std::vector<std::pair<std::string, std::string>> settings;
+
+  auto pDlgSelect = static_cast<CGUIDialogSelect*>(CServiceBroker::GetGUI()->GetWindowManager().GetWindow(WINDOW_DIALOG_SELECT));
+  pDlgSelect->Reset();
+  pDlgSelect->SetHeading(CVariant{atoi(params[0].c_str())});
+
+  for (auto param = params.begin() + 2; param < params.end(); ++param)
+  {
+    if (param->find('|') != std::string::npos)
+    {
+      std::vector<std::string> values = StringUtils::Split(*param, '|', 2);
+      std::string label = g_localizeStrings.Get(atoi(values[0].c_str()));
+      settings.push_back(std::make_pair(label, values[1]));
+      pDlgSelect->Add(label);
+    }
+  }
+
+  pDlgSelect->Open();
+  if (pDlgSelect->IsConfirmed())
+  {
+    unsigned int item = pDlgSelect->GetSelectedItem();
+    int settingLabel = CSkinSettings::GetInstance().TranslateString(params[1] + ".Label");
+    int settingValue = CSkinSettings::GetInstance().TranslateString(params[1] + ".Value");
+    std::string value1 = settings[item].first;
+    std::string value2 = settings[item].second;
+
+    CSkinSettings::GetInstance().SetString(settingLabel, value1);
+    if (value2 == "-")
+      CSkinSettings::GetInstance().Reset(params[1] + ".Value");
+    else
+      CSkinSettings::GetInstance().SetString(settingValue, value2);
+    CServiceBroker::GetSettings()->Save();
+
+  }
+  return 0;
+}
+
 /*! \brief Select and set a skin bool setting.
  *  \param params The parameters.
  *  \details params[0] = Names of skin settings.
@@ -546,6 +590,7 @@ CBuiltins::CommandMap CSkinBuiltins::GetOperations() const
            {"skin.reset",         {"Resets a skin setting to default", 1, SkinReset}},
            {"skin.resetsettings", {"Resets all skin settings", 0, SkinResetAll}},
            {"skin.setaddon",      {"Prompts and set an addon", 2, SetAddon}},
+           {"skin.select",        {"Prompts and choose a value for a skin setting", 3, SkinSelect}},
            {"skin.selectbool",    {"Prompts and set a skin setting", 2, SelectBool}},
            {"skin.setbool",       {"Sets a skin setting on", 1, SetBool}},
            {"skin.setfile",       {"Prompts and sets a file", 1, SetFile}},


xbmc/interfaces/builtins/SkinBuiltins.cpp
```diff
@@ -102,6 +102,50 @@ static int SetAddon(const std::vector<std::string>& params)
...
+
+/*! \brief Select and choose a value for a skin setting.
+ *  \param params The parameters.
+ *  \details params[0] = Names of skin settings.
+ */
+static int SkinSelect(const std::vector<std::string>& params)
+{
+  std::vector<std::pair<std::string, std::string>> settings;
+  
+  CGUIDialogSelect* pDlgSelect = (CGUIDialogSelect*)g_windowManager.GetWindow(WINDOW_DIALOG_SELECT);
```
phil65: static cast + auto keyword.

xbmc/interfaces/builtins/SkinBuiltins.cpp
```diff
@@ -102,6 +102,50 @@ static int SetAddon(const std::vector<std::string>& params)
...
+  pDlgSelect->Reset();
+  pDlgSelect->SetHeading(CVariant{g_localizeStrings.Get(atoi(params[0].c_str()))});
```
phil65: not 100% sure, but I think this also works without g_localizeStrings.Get().

xbmc/interfaces/builtins/SkinBuiltins.cpp
```diff
@@ -102,6 +102,50 @@ static int SetAddon(const std::vector<std::string>& params)
...
+  
+  for (unsigned int i = 1 ; i < params.size() ; i++)
```
phil65: range-based for loop

BigNoid:
thx for the pointers @phil65 should be good now.

xbmc/interfaces/builtins/SkinBuiltins.cpp
```diff
@@ -102,6 +102,50 @@ static int SetAddon(const std::vector<std::string>& params)
...
+  auto pDlgSelect = static_cast<CGUIDialogSelect*>(g_windowManager.GetWindow(WINDOW_DIALOG_SELECT));
+  pDlgSelect->Reset();
+  pDlgSelect->SetHeading(CVariant{atoi(params[0].c_str())});
+  
+  for (auto &param : params)
+  {
+    if (param.find('|') != std::string::npos)
+    {
+      std::vector<std::string> values = StringUtils::Split(param, '|');
+      std::string label = g_localizeStrings.Get(atoi(values[0].c_str()));
```
xhaggi: It would be better you make sure you got only 2 elements in the values array before you access the second index.

xbmc/interfaces/builtins/SkinBuiltins.cpp
```diff
@@ -102,6 +102,50 @@ static int SetAddon(const std::vector<std::string>& params)
...
+      settings.push_back(std::make_pair(label, values[1].c_str()));
+      pDlgSelect->Add(label);
+    }
+  }
+  
+  pDlgSelect->Open();
```
xhaggi: If no label/value pairs are found, do we need to show the dialog?

xbmc/interfaces/builtins/SkinBuiltins.cpp
```diff
@@ -102,6 +102,50 @@ static int SetAddon(const std::vector<std::string>& params)
...
+  if(pDlgSelect->IsConfirmed())
+  {
+    unsigned int item = pDlgSelect->GetSelectedItem();
+    int setting1 = CSkinSettings::GetInstance().TranslateString(params[1] + ".Label");
+    int setting2 = CSkinSettings::GetInstance().TranslateString(params[1] + ".Value");
```
xhaggi: `settingLabel` and `settingValue` would fit better here.

xbmc/interfaces/builtins/SkinBuiltins.cpp
```diff
@@ -102,6 +102,50 @@ static int SetAddon(const std::vector<std::string>& params)
...
+    std::string value1 = settings[item].first;
```
xhaggi: `label`

xbmc/interfaces/builtins/SkinBuiltins.cpp
```diff
@@ -102,6 +102,50 @@ static int SetAddon(const std::vector<std::string>& params)
...
+    std::string value2 = settings[item].second;
```
xhaggi: `value`

xhaggi (changes requested):

Rechi:
@BigNoid please update the PR with above comments

MartijnKaijser:
@BigNoid nudge :)

xbmc/interfaces/builtins/SkinBuiltins.cpp
```diff
@@ -104,6 +104,49 @@ static int SetAddon(const std::vector<std::string>& params)
   return 0;
 }
 
+/*! \brief Select and choose a value for a skin setting.
+ *  \param params The parameters.
+ *  \details params[0] = Names of skin settings.
```
xhaggi: `param[0]` is the header if I'm right.

xbmc/interfaces/builtins/SkinBuiltins.cpp
```diff
@@ -104,6 +104,49 @@ static int SetAddon(const std::vector<std::string>& params)
...
+{
+  std::vector<std::pair<std::string, std::string>> settings;
+
+  auto pDlgSelect = static_cast<CGUIDialogSelect*>(CServiceBroker::GetGUI()->GetWindowManager().GetWindow(WINDOW_DIALOG_SELECT));
+  pDlgSelect->Reset();
+  pDlgSelect->SetHeading(CVariant{atoi(params[0].c_str())});
+
+  for (auto &param : params)
+  {
+    if (param.find('|') != std::string::npos)
```
xhaggi: We should skip the first two parameters before we make the loop over the parameters.  The current implementation is not very nice, because you could add a pipe to header parameter 0 and it is interpreted as a setting value.

xhaggi (changes requested):

phunkyfish:

PR #10788, opened 2016-10-27
[gui] - remove fallbacks for pvr next* infolabels

phil65:
At the moment the PVR infolabels nextstarttime/date fallback to current time in case the info is not available. Makes much more sense to fallback to an empty label instead, skinners can do fallbacks via $VARs.

@ronie  @xhaggi


xhaggi:
Makes totally sense IMO. Never noticed that there is a fallback. @ksooo mind taking a look at it too.

ksooo:
yeah, this is fine. I did exactly this change (no fallback to current date/time) for LISTITEM_STARTTIME, LISTITEM_ENDTIME, LISTITEM_STARTDATE and LISTITEM_ENDDATE with a previous PR: https://github.com/xbmc/xbmc/pull/10248/commits/fc0b7aa69bed48a8b45b5fd594207d48c8e19644

ksooo:
jenkins build this please

PR #10789, opened 2016-10-27
[AirTunes/DACP] - fix remote control

Memphiz:
This fixes the DACP remote control feature which allows us to control the client when streaming audio/music via airtunes with the Kodi playback controls.

@FernetMenta any idea why i only get an OnSpeedChanged announcement when playback is resumed? Is this expected?

Also who the fuck stripped off the arbitrary request header support in CurlFile. Imo those changes should be communicated. Or at the least have a grep at the code on who is using "SetProtocolOption" before restricting request headers and simply throw away anything else.

Its always the apple features that get broken _grrr_

### DIFF
--- a/xbmc/filesystem/CurlFile.cpp
+++ b/xbmc/filesystem/CurlFile.cpp
@@ -825,6 +825,10 @@ void CCurlFile::ParseAndCorrectUrl(CURL &url2)
           m_postdata = Base64::Decode(value);
           m_postdataset = true;
         }
+        else if (name == "active-remote")// needed for DACP!
+        {
+          SetRequestHeader(it->first, value);
+        }
         // other standard headers (see https://en.wikipedia.org/wiki/List_of_HTTP_header_fields)
         else if (name == "accept" || name == "accept-language" || name == "accept-datetime" ||
                  name == "authorization" || name == "cache-control" || name == "connection" ||


Memphiz:
jenkins build this please

FernetMenta:
@Memphiz there is an open pr that fixes the OnSpeedChanged issue

Memphiz:
Mhh which one is it? I might need to change this very PR then maybe...

MartijnKaijser:
https://github.com/xbmc/xbmc/pull/10739

akva2:
@Memphiz let us summarize some history.

1) the codec info was simply removed from the application. bam, users hated it and it has to be brought back. Offender: FernetMenta
2) Jump to time was removed. Users hated it. offender: FernetMenta. still not resolved and kodi 17 will be released without a feature a lot of users used.
3) All audio decoders was almost removed, I managed to stop that. almost-offender: FernetMenta
4) HTTP header support was removed. offender: FernetMenta

FernetMenta:
@akva2 not sure what you are trying to say. Look at github stats you you see that I do by far the most changes. As a result things break every now and then. 
btw: it is unfair to just list some negatives. I am sure that the positives outweigh those few issues.

akva2:
oh i knew i should have been more verbose, but was on the phone. the positives outweigh the negatives, they are not even on the same page, and this was NOT an attempt at picking a fight. it was just to point out a trend. my point is that you (and @tamland) in particular, have a tendency to shoot first, ask later when it comes to removal of code (and features) and that brings trouble at times. there's lots of dodo in the kodi code base and no doubt a lot of the code deserves to be kicked out. but there is also a reason for the code being there in 95% of the cases. all parties would benefit if you put some more effort into mapping the why before the boot is activated.

Memphiz:
@FernetMenta so what is this PR that fixes something - do i need to adapt this PR for it?

Memphiz:
@FernetMenta sorry - just saw the comment from martijn ...

Memphiz:
@FernetMenta i don't think the PR mentioned does anything for announcements. On resuming Playback i only get the OnSpeedChanged announcement with param 1 - so basically my commit for this is still valid. Should there be any other announcement on resume?

FernetMenta:
isn't OnSpeedChanged good enough for this purpose?

> so basically my commit for this is still valid

what commit are you referring to?

Memphiz:
https://github.com/xbmc/xbmc/pull/10789/commits/6eb46b22fe667c2da790820dd7a99158326c1892

I had to add this - looks like at some point on resume the OnPlay announcememt was sent - or i simply forgot this scenario in the first place...

FernetMenta:
you should still get OnPlay announcement on resume

FernetMenta:
@Memphiz https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/VideoPlayer.cpp#L3650

Memphiz:
@FernetMenta this is paplayer ...

https://github.com/xbmc/xbmc/blob/master/xbmc/cores/paplayer/PAPlayer.cpp#L946

there is no onresumed callback fired in the case of unpause.

https://github.com/xbmc/xbmc/blob/master/xbmc/cores/paplayer/PAPlayer.cpp#L924 <- this code is supposed to do it - but the method is not called.

For clarification. I play an mp3 file for example ... hit pause on the osd ... and then hit play on the osd again ...

FernetMenta:
@Memphiz papalyer, right. that needs the same change as VP, see liked pr. do you have time to do this?

Memphiz:
Yes - maybe tomorrow

Memphiz:
Removed the setspeed workaround in this PR in favour of #10886

Memphiz:
jenkins build and merge

PR #10791, opened 2016-10-27
ActiveAEFilter: Add missing include for avcodec_fill_audio_frame

fritsch:
This worked by accident. Found it when I had a brief preview of ffmpeg 3.2 in combination with kodi.


fritsch:
@FernetMenta ping

PR #10760, opened 2016-10-23
[xbmc][win]Build addons with debug info on Windows

Paxxi:
## Description

Change buildsetup to bundle all our PDB files into an
archive, should hopefully make it easier to look at addon
crashes
## Motivation and Context

Get better crash logs and make investigation of crashes easier.
## Types of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
## Checklist:

- [ ] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [ ] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #10793, opened 2016-10-27
Revert "[xbmc][win]Build addons with debug info on Windows"

MartijnKaijser:
Reverts xbmc/xbmc#10760

Causes binary add-on to not load. Will investigate later why and do proper fix.


MartijnKaijser:
jenkins build and merge

PR #10794, opened 2016-10-27
[addons] make sure temp dir exists before running CFileOperationJob

tamland:
As a workaround as the whole CFileOperationJob breaks down if it happens to find a directory before a file in the archive..

### DIFF
--- a/xbmc/addons/FilesystemInstaller.cpp
+++ b/xbmc/addons/FilesystemInstaller.cpp
@@ -40,6 +40,9 @@ bool CFilesystemInstaller::InstallToFilesystem(const std::string& archive, const
   auto newAddonData = URIUtils::AddFileToFolder(m_tempFolder, StringUtils::CreateUUID());
   auto oldAddonData = URIUtils::AddFileToFolder(m_tempFolder, StringUtils::CreateUUID());
 
+  if (!CDirectory::Create(newAddonData))
+    return false;
+
   if (!UnpackArchive(archive, newAddonData))
   {
     CLog::Log(LOGERROR, "Failed to unpack archive '%s' to '%s'", archive.c_str(), newAddonData.c_str());


stefansaraev:
works for me. thanks much

MartijnKaijser:
jenkins build this please

MartijnKaijser:
@tamland current install issue on windows is it can't create the temp folders.
http://paste.ubuntu.com/23393516/  (fresh install)

If i manually create the  /temp/ folder, wipe the userdata again (except the temp folder) it does update the add-ons. Still one add-on still errors after that
http://paste.ubuntu.com/23393551/

Memphiz:
we see the same on atv4. @tamland and who is supposed to create the directory special://home/temp ??? - are you sure you don't wanna simply use special://temp which already exists for sure?

Either use special://temp or ensure that special://home/temp is created at some point.

maybe here:

https://github.com/xbmc/xbmc/blob/master/xbmc/Application.cpp#L1062

But i see no point in having another temp directory tbh.

MartijnKaijser:
@Memphiz have you tested with https://github.com/xbmc/xbmc/pull/10811  ?
I think one /temp/ should suffice as well but there might be an additional reason?

tamland:
>we see the same on atv4. @tamland and who is supposed to create the directory special://home/temp ??? - are you sure you don't wanna simply use special://temp which already exists for sure?

>Either use special://temp or ensure that special://home/temp is created at some point.

They are different thing? All mapped to ~/.kodi/temp here; and used for many other things..

Memphiz:
the only spot i even found special://home/temp was in the filesysteminstaller. Well basically you can't assume that there is a temp folder inside of special://home on all platforms... might be the case on some - but nothing ensures that .

Memphiz:
so if you want to have an ensured temp folder - special://temp would be it ... because this is a virtual path ... special://home/temp is not a virtual path - its a concatination of special://home (virtual) and temp (a real directory that must be created first)

Memphiz:
@MartijnKaijser #10811 fixes it too - but i think the special://home/temp has to go - because it creates another temp directory on some platforms ...

tamland:
On which platform is it not mapped to home? FilesystemInstaller can be changed to use any temporary directory *as long as it's on the same filesystem as home/addons*, otherwise it won't work.

Memphiz:
@tamland osx for example:

20:47:11 T:4410646528  NOTICE: special://home/ is mapped to: /Users/memphis/Library/Application Support/Kodi
20:47:11 T:4410646528  NOTICE: special://temp/ is mapped to: /Users/memphis/.kodi/temp

app data that should be retained has to go into Application Support ... os policy.

PR #10795, opened 2016-10-27
[PVR][EPG] Feature: gridcontainer orientation

ksooo:
A feature that has been requested several times: PVR guide window with grid layout with channels as columns and timeline as rows.

screenshot004

screenshot003

Please note that the skin changes are not intended to be merged with this PR (unless people say they want it in estuary). It's just a test/reference implementation. The respective commit will be removed from this PR before merge. 

@xhaggi for code review
@ronie heads-up for skinning engine extension

### DIFF
--- a/xbmc/epg/GUIEPGGridContainer.cpp
+++ b/xbmc/epg/GUIEPGGridContainer.cpp
@@ -44,15 +44,17 @@ using namespace EPG;
 static const int BLOCK_SCROLL_OFFSET = 60 / CGUIEPGGridContainerModel::MINSPERBLOCK; // how many blocks are jumped if we are at left/right edge of grid
 static const int PAGE_NOW_OFFSET = CGUIEPGGridContainerModel::GRID_START_PADDING / CGUIEPGGridContainerModel::MINSPERBLOCK; // this is the 'now' block relative to page start  
 
-CGUIEPGGridContainer::CGUIEPGGridContainer(int parentID, int controlID, float posX, float posY, float width,
-                                           float height, int scrollTime, int preloadItems, int timeBlocks, int rulerUnit,
+CGUIEPGGridContainer::CGUIEPGGridContainer(int parentID, int controlID, float posX, float posY, float width, float height,
+                                           ORIENTATION orientation, int scrollTime, int preloadItems, int timeBlocks, int rulerUnit,
                                            const CTextureInfo& progressIndicatorTexture)
 : IGUIContainer(parentID, controlID, posX, posY, width, height),
+  m_orientation(orientation),
   m_channelLayout(nullptr),
   m_focusedChannelLayout(nullptr),
   m_programmeLayout(nullptr),
   m_focusedProgrammeLayout(nullptr),
   m_rulerLayout(nullptr),
+  m_rulerDateLayout(nullptr),
   m_pageControl(0),
   m_rulerUnit(rulerUnit),
   m_channelsPerPage(0),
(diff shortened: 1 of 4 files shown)


ronie:
looks great! ;-)

xhaggi:
@ksooo we dropped this feature a year ago or so 🤔 
I wonder why it comes back now. Don't get me wrong I'm sure you re-added it in a more suitable way.
I'll take a look at it later.

ksooo:
As I wrote there are plenty of people requesting this feature. Are you sure we ever had this feature? Quick look through the commit history does not reveal anything like this.

xhaggi:
https://github.com/ksooo/xbmc/commit/a1fb9dcf4c2c94880422a1742667ebab84698790

phil65:
Since we are in the process of removing all other viewtypes (now, next, channel) from the pvrguide window, perhaps we could give that window proper viewtype support just for the epg grid? That way skinners would have a sane way to optionally allow an additional layout.

ksooo:
Okay. Found it, was removed in mid of 2014. But I wonder why this happened. People consider this a neat feature and I agree (otherwise I never would have implemented it).

Maybe back then it was the right decision, because shit simply did not work? Now it does. ;-) I changed fundamentally things in this control during the last year. Maybe with this I "accidentally" fixed stuff that prevented the horizontal layout from working properly before.

ksooo:
@phil65 sorry, I have no idea what "proper view type support means".

ksooo:
... but would be happy to implement it.

phil65:
Have a look at CGUIViewControl, that´s what would be needed for that window.

ksooo:
> Have a look at CGUIViewControl, that´s what would be needed for that window. That´s how mediawindows (and some other special cases) deal with viewtypes.

 The CGUIWindoPVRGuide is derived from CGUIMediaWindow. Does this mean it already supports "proper viewtypes"?

phil65:
Possible. @xhaggi can probably tell us how feasible it is to implement that.

xhaggi:
@ksooo don't get me wrong again, if this is requested by several users and if it's done in a proper way I'm fine with it. Thanks for doing it 😄 

Regarding the view type support: The window supports it since I did the rewrite of the PVR views to be normal windows. You are able to specify a new views like you can do on every other media window e.g. "Vertical", "Horizontal".

xhaggi:
Ahh sorry but I'm not sure if it works currently because of the static view types "Now", "Next", "Channel" we want to drop for v18. After that it definitely should.

ksooo:
> don't get me wrong again, if this is requested by several users and if it's done in a proper way I'm fine with it. 

 yeah, this is how i got it in the first place. all good.

ksooo:
> Ahh sorry but I'm not sure if it works currently because of the static view types "Now", "Next", "Channel" we want to drop for v18. After that it definitely should.

 good to know. thanks for clarifying. thus, "proper viewtypes" (TM)  should be no big deal for v18, right.

phil65:
> Ahh sorry but I'm not sure if it works currently because of the static view types "Now", "Next", "Channel" we want to drop for v18. After that it definitely should.

Yeah that´s why I asked, viewtype handling was always bit special for this window. If everything works like a normal MediaWindow after removal of Next, Now etc, then all perfect. 👍

xhaggi:
no, I'll create a PR which drops all other static view types except the timeline view and after that we're able to use the view type feature like on every other media window.

ksooo:
Rebased and removed skin reference  implementation which now can be found here: https://github.com/ksooo/xbmc/commit/860acc82599724b53a06fc6a2753a215c3f8a32e

jenkins build this please

PR #10797, opened 2016-10-28
deprecate autotools

wsnipex:
Give a fair warning that autotools will die soon

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2408,3 +2408,8 @@ AC_OUTPUT
 
 final_message="$final_message\n  prefix:\t$prefix\n$dashes"
 echo -e "$final_message\n"
+echo "#------------------------------------ WARNING ----------------------------------------#"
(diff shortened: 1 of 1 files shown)


MartijnKaijser (approved):

configure.ac
```diff
@@ -2408,3 +2408,8 @@ AC_OUTPUT
 
 final_message="$final_message\n  prefix:\t$prefix\n$dashes"
 echo -e "$final_message\n"
+echo "#------------------------------------ WARNING ----------------------------------------#"
+echo "The autotools based build system is deprecated and will be removed in the next version!"
```
wsnipex: No strong opinion, but I prefer it as is. That way we don't have to change it, if we need to postpone removal for some reason...

MartijnKaijser: Fine by me :)
Merge away

PR #10799, opened 2016-10-28
VideoPlayer: add user name and password back to url after testing for…

FernetMenta:
fix http://trac.kodi.tv/ticket/17005

This is not the only issue. Scraped item lose the property "IsHTTPDirectory" and are treated as internet streams.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDInputStreams/DVDFactoryInputStream.cpp
+++ b/xbmc/cores/VideoPlayer/DVDInputStreams/DVDFactoryInputStream.cpp
@@ -154,6 +154,8 @@ CDVDInputStream* CDVDFactoryInputStream::CreateInputStream(IVideoPlayer* pPlayer
         {
           CURL finalUrl(curlFile.GetURL());
           finalUrl.SetProtocolOptions(origUrl.GetProtocolOptions());
+          finalUrl.SetUserName(origUrl.GetUserName());
+          finalUrl.SetPassword(origUrl.GetPassWord());
           finalFileitem.SetPath(finalUrl.Get());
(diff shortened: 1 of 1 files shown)


MartijnKaijser:
jenkins build this please

arnova:
@FernetMenta : Shouldn't we fix this inside CCurlFile? As in: shouldn't it just return the proper path including eg. username/password?

FernetMenta:
@arnova we pass the redirected path, that doesn't have the original parameter anymore

arnova:
@FernetMenta : Yes, I know but shouldn't CCurlFile::GetURL() handle  putting back username, password and protocol options?

And one other thing that puzzles me (note: haven't looked at the code yet): why does the problem here cause video files to be played from the library to break, while it still works when played from the files section?

FernetMenta:
@arnova 

> why does the problem here cause video files to be played from the library to break, while it still works when played from the files section?

see my comment on top of this pr

> Yes, I know but shouldn't CCurlFile::GetURL() handle putting back username, password and protocol options?

not sure, maybe this would harm on some other place.

arnova:
@FernetMenta : Note that I'm fine with this PR as is. Just wanted to point out that GetURL() was introduced, and only introduced, to be used here so it shouldn't matter whether you add the change here or in CurlFile itself.

FernetMenta:
@arnova I did not know this. A quick search on the the code showed that GetURL has been used on a few other places meanwhile.
I agree that all this could be made less error prone but would like to postpone to v18. ok?

arnova:
@FernetMenta : Fine by me.

PR #10766, opened 2016-10-25
VideoPlayer: fixes for hls

FernetMenta:
fix for http://trac.kodi.tv/ticket/16670

fix for http://trac.kodi.tv/ticket/16999

PR #10801, opened 2016-10-29
VideoPlayer: hls, select best resolution closest to desired bandwidth

FernetMenta:
see title

@popcornmix that fixes the issue for the url you gave me


popcornmix:
Yes, fixes the URL. However now playing that URL takes between 6 and 12 seconds to start stream.
With #10766 reverted the stream starts in 1 or 2 seconds.

Any reason why it's much slower?

FernetMenta:
It is much faster here on my Mac. Stream also starts within 1-2 secs. The old function did parse the m3m8 playlist file to select a particular url. But it did not properly and as a result on some playlists audio was missing. Now we let ffmpeg do this.

Maybe the self-selected stream was identified as mpegts and demuxer used the "fast-channel-switch" path by reducing time for analysis in find_stream_info.

jenkins build this please

MartijnKaijser:
good for merge?

PR #10802, opened 2016-10-29
[strings] Clarify it's system memory/cpu usage

MartijnKaijser:
## Description

see title
## Motivation and Context

Users complaining Kodi is using a lot of memory/cpu while it can be anything else running on the system that is hogging it. Stop blaming Kodi.
## How Has This Been Tested?

not

### DIFF
--- a/addons/skin.estouchy/language/resource.language.en_gb/strings.po
+++ b/addons/skin.estouchy/language/resource.language.en_gb/strings.po
@@ -179,7 +179,7 @@ msgstr ""
 #empty strings from id 31304 to 31308
 
 msgctxt "#31309"
-msgid "Memory Used:"
+msgid "System Memory Used:"
 msgstr ""
 
 #empty strings from id 31310 to 31319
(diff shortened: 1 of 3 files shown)


addons/skin.estouchy/language/resource.language.en_gb/strings.po
```diff
@@ -179,7 +179,7 @@ msgstr ""
 #empty strings from id 31304 to 31308
 
 msgctxt "#31309"
-msgid "Memory Used:"
+msgid "System Memory Used:"
```
hudokkow: Since your touching it, mind choosing used or usage? For consistency sake...
Also, please don't capitalize the second word.

MartijnKaijser: Sure and it was already there (and every other word in the language file also has it)

jjd-uk: Looks ok to me

PR #10805, opened 2016-10-29
Remove XMemUtils' alloc when ffmpeg intrinsics might be used

fritsch:
This replaces our homebaked _aligned_malloc with just ffmpeg's av_malloc as we continue to run into alignment related intrinsic segfaults with ffmpeg, let ffmpeg decide about it's alignment.

The only areas the old code is left is:
VAAPI.cpp where we need a 64 byte aligned memory area (av_malloc provides)
GUIShaderDX.h which uses some "special" alignment, no idea here

### DIFF
--- a/xbmc/cores/AudioEngine/Utils/AEBuffer.cpp
+++ b/xbmc/cores/AudioEngine/Utils/AEBuffer.cpp
@@ -40,7 +40,7 @@ CAEBuffer::~CAEBuffer()
 void CAEBuffer::Alloc(const size_t size)
 {
   DeAlloc();
-  m_buffer     = (uint8_t*)_aligned_malloc(size, 16);
+  m_buffer     = (uint8_t*)_aligned_malloc(size, 32);
   m_bufferSize = size;
   m_bufferPos  = 0;
 }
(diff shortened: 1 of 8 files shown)


FernetMenta:
this is no-go and I told you already. I won't make everything dependent on libav.

further I see zero advantage in doing so. ffmpeg aligns by 16 or 32
https://github.com/FFmpeg/FFmpeg/blob/415f907ce8dcca87c9e7cfdc954b92df399d3d80/libavutil/mem.c#L64
we can do as well

fritsch:
@fernetmenta: as you see from the code you linked if we keep 16 - no AVX will be used. Do we want that?

FernetMenta:
point me to the code where AVX would be applied to buffers allocated by us

xbmc/cores/AudioEngine/Utils/AEBuffer.cpp
```diff
@@ -20,9 +20,9 @@
 
 #include "AEBuffer.h"
 #include <algorithm>
-#ifdef TARGET_POSIX
```
FernetMenta: why don't you just remove entire CAEBuffer? it is not used anymore (at least I could not find any usage)

fritsch:
@FernetMenta https://github.com/xbmc/xbmc/blob/master/xbmc/guilib/FFmpegImage.cpp#L464

fritsch:
I will close that and test with an 32 byte alignment

fritsch:
Here, also: https://github.com/xbmc/xbmc/blob/master/xbmc/pictures/Picture.cpp#L179

xbmc/pictures/Picture.cpp
```diff
@@ -176,7 +180,7 @@ bool CPicture::ResizeTexture(const std::string &image, uint8_t *pixels, uint32_t
   // create a buffer large enough for the resulting image
   GetScale(width, height, dest_width, dest_height);
 
-  uint8_t *buffer = new uint8_t[dest_width * dest_height * sizeof(uint32_t)];
+  uint8_t *buffer = (uint8_t*) _aligned_malloc(dest_width * dest_height * sizeof(uint32_t), 32);
```
FernetMenta: this file already makes use of ffmpeg (sws_scale). use of av_malloc is welcome here

PR #10807, opened 2016-10-29
workaround for addon installer deadlocks

tamland:
This fixes deadlocks caused by the addon installer giving jobs to the job manager which depends on waiting on other jobs to finish (see https://github.com/xbmc/xbmc/blob/49cf9e6a5cb2e393cd1769c0b2ae34e2689e4931/xbmc/addons/AddonInstaller.cpp#L729). If the job manager is out of workers, it deadlocks.

Since this is the addon installer being broken at its core,  I see no other way (other than a rewrite) than to give it as many threads as required.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -1159,11 +1159,11 @@ bool CApplication::Initialize()
     event.Reset();
     std::atomic<bool> isMigratingAddons(false);
     CJobManager::GetInstance().Submit([&event, &incompatibleAddons, &isMigratingAddons]() {
-      incompatibleAddons = CAddonSystemSettings::GetInstance().MigrateAddons([&isMigratingAddons]() {
-        isMigratingAddons = true;
-      });
-      event.Set();
-    });
+        incompatibleAddons = CAddonSystemSettings::GetInstance().MigrateAddons([&isMigratingAddons]() {
+          isMigratingAddons = true;
+        });
+        event.Set();
+      }, CJob::PRIORITY_DEDICATED);
     localizedStr = g_localizeStrings.Get(24151);
     iDots = 1;
     while (!event.WaitMSec(1000))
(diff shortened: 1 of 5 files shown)


MartijnKaijser:
@Paxxi @FernetMenta any remarks or suggestions that fit with the v17 scope?

FernetMenta:
JobManager can be used by every component. Allowing an undefined number of threads is high risk. That would just create another weak point instead of fixing one.

notspiff:
wouldn't some (variant of) https://github.com/notspiff/xbmc/tree/draft_installer_fix do the job ? not tested.

tamland:
>wouldn't some (variant of) https://github.com/notspiff/xbmc/tree/draft_installer_fix do the job ? not tested.

Not tested either but doesn't this end in a loop of constant rescheduling the same job?

>JobManager can be used by every component. Allowing an undefined number of threads is high risk. That would just create another weak point instead of fixing one.

To be fair, it's set to 10000, which is not even that high. Not saying this is 'good' in any way but is it *really* better to have a job manager that can deadlock, and where the 'high' priority isn't 'high' at all, but involves blocking until there's a free thread?

Either way, the proper fix is redesign the whole installer and make it non-blocking. But we need a quick fix for v17.

FernetMenta:
@tamland how about adding an extra prio for that work-around?

akva2:
The idea was to put the job at the end of the queue but i didnt check if queue is lifo or fifo

akva2:
on a pc so can be more verbose. the situation used to be this. consider some addon, with dependency dep1, which again depends on dep2. the installer is a job queue with two workers. the installation jobs were scheduled as 'dep2 dep1 addon'. eventually you ended up in the situation where 'dep1' and 'addon' were executing in parallel. the 'sleep' code was there to make the 'addon' job wait for the 'dep1' job to finish.

if this now deadlocks, it somehow must end up with execution of 'dep1' and 'addon' before 'dep2'. so 'dep1' will wait for 'dep2', 'addon' waits for 'dep1' and it deadlocks. iow, the scheduling code must have changed. the cure then is to give up the worker slot to the 'dep2' job, and reschedule 'dep1' to execute afterwards. which is what i tried to implement in my patch.

tamland:
Added a new priority. Can you please have a look @FernetMenta?

FernetMenta:
my concerns are addressed, thanks

tamland:
ok, 
jenkins build this please

PR #10809, opened 2016-10-29
AudioTrack: Workarounds

fritsch:
Yet another bunch of workarounds for broken firmware. This time AFTV2 via: http://forum.kodi.tv/showthread.php?tid=295014&pid=2446406#pid2446406

### DIFF
--- a/xbmc/cores/AudioEngine/Sinks/AESinkAUDIOTRACK.cpp
+++ b/xbmc/cores/AudioEngine/Sinks/AESinkAUDIOTRACK.cpp
@@ -162,8 +162,10 @@ static int AEChannelMapToAUDIOTRACKChannelMask(CAEChannelInfo info)
          | CJNIAudioFormat::CHANNEL_OUT_SIDE_RIGHT;
   else if (info.Count() > 2)
     return CJNIAudioFormat::CHANNEL_OUT_5POINT1;
-  else
+  else if (info.Count() == 2)
     return CJNIAudioFormat::CHANNEL_OUT_STEREO;
+  else
+    return CJNIAudioFormat::CHANNEL_OUT_MONO;
 #endif
 
   info.ResolveChannels(KnownChannels);
@@ -323,6 +325,7 @@ bool CAESinkAUDIOTRACK::Initialize(AEAudioFormat &format, std::string &device)
 
   while (!m_at_jni)
   {
+    CLog::Log(LOGNOTICE, "Trying to open: %u samplerate %d channelMask %d encoding", m_sink_sampleRate, atChannelMask, m_encoding);
     int min_buffer = CJNIAudioTrack::getMinBufferSize(m_sink_sampleRate,
                                                          atChannelMask,
                                                          m_encoding);
--- a/xbmc/cores/AudioEngine/Sinks/AESinkAUDIOTRACK.cpp
+++ b/xbmc/cores/AudioEngine/Sinks/AESinkAUDIOTRACK.cpp
@@ -245,12 +245,19 @@ bool CAESinkAUDIOTRACK::Initialize(AEAudioFormat &format, std::string &device)
   int stream = CJNIAudioManager::STREAM_MUSIC;
   m_encoding = CJNIAudioFormat::ENCODING_PCM_16BIT;
 
-  // Get equal or lower supported sample rate
-  std::set<unsigned int>::iterator s = m_sink_sampleRates.upper_bound(m_format.m_sampleRate);
-  if (--s != m_sink_sampleRates.begin())
-    m_sink_sampleRate = *s;
-  else
-    m_sink_sampleRate = CJNIAudioTrack::getNativeOutputSampleRate(CJNIAudioManager::STREAM_MUSIC);
+  uint32_t distance = 192000; // max upper distance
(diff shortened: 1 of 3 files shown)


fritsch:
jenkins build this please

MrMC:
how does a requested sample rate of 22050 get past the check at line 247 when the min supported sink sample rate is 32000 ?

AESinkAUDIOTRACK - 32000 supported

fritsch:
That's a very good question - let's find that out.

fritsch:
Build pending here: http://jenkins.kodi.tv/job/Android-ARM/9720/
Let's use the opportunity to correctly support MONO and see if it works: https://github.com/fritsch/xbmc/commit/0f873e8c659dcbaea3b28bdbdd731a2398ea74eb

Here is a testfile: https://dl.dropboxusercontent.com/u/55728161/audio.mp3

Thanks for the feedback

fritsch:
The problem was with the way the samplerates were chosen. The iterator comparison was wrongly used. If changed the metric for matching samplerates. In order to resample more correctly if needed.

Let's wait until that aac 1.0 user from the forum has tested it. My samples are opened as stereo, so that needs to be found out.

fritsch:
For the record: 

The bug is here: https://github.com/xbmc/xbmc/blob/master/xbmc/cores/AudioEngine/Sinks/AESinkAUDIOTRACK.cpp#L248

If you query with 22050 hz the element the upper selects is 32000. That very moment the iterator points to begin() as 32000 is the first element in the set, therefore --s will not (!) point to begin() but to somewhere undefined. And this value we assign -> butz :-)

fritsch:
jenkins build and merge

fritsch:
jenkins build and merge after being interrupted by non jenkins merge.

MartijnKaijser:
Jenkins build this please

xbmc/cores/AudioEngine/Sinks/AESinkAUDIOTRACK.cpp
```diff
@@ -245,12 +245,19 @@ bool CAESinkAUDIOTRACK::Initialize(AEAudioFormat &format, std::string &device)
...
-    m_sink_sampleRate = *s;
-  else
-    m_sink_sampleRate = CJNIAudioTrack::getNativeOutputSampleRate(CJNIAudioManager::STREAM_MUSIC);
+  uint32_t distance = 192000; // max upper distance
+  for (auto& s : m_sink_sampleRates)
+  {
+     // prefer best match or alternatively something that divides nicely and
+     // is not too far away
+     uint32_t d = std::abs((int)m_format.m_sampleRate - (int)s) + 8 * (s > m_format.m_sampleRate ? (s % m_format.m_sampleRate) : (m_format.m_sampleRate % s));
+     if (d < distance)
```
MrMC: I think if you worked a bit more, you could put more on that line :)

xbmc/platform/android/jni/AudioFormat.cpp
```diff
@@ -34,6 +34,7 @@ int CJNIAudioFormat::ENCODING_DTS_HD                   = 0x00000008;
 int CJNIAudioFormat::ENCODING_DOLBY_TRUEHD             = 0x00000009;
 int CJNIAudioFormat::ENCODING_IEC61937                 = 0x0000000d;
 
+int CJNIAudioFormat::CHANNEL_OUT_MONO                  = 0x00000004;
 int CJNIAudioFormat::CHANNEL_OUT_STEREO                = 0x0000000c;
```
MrMC: not enough 👍 

need below like how CHANNEL_OUT_STEREO is handled

      CJNIAudioFormat::CHANNEL_OUT_MONO  = get_static_field<int>(c, "CHANNEL_OUT_MONO ");

fritsch: I want to change that in the future, this get_static_field helps us to support that on different platforms, but it also makes us incompatible with "real upstream API".

Think of AML's ENCODING_AC3 with the static_field getter we overwrite this field but don't know anymore if we now need to talk RAW or IEC.

fritsch: for the time being I added it. Will remove all of that after v17 :-)

fritsch: Oki, added it. Thanks for the feedback.

xbmc/platform/android/jni/AudioFormat.cpp
```diff
@@ -76,6 +77,7 @@ void CJNIAudioFormat::PopulateStaticFields()
     if (sdk >= 5)
     {
       CJNIAudioFormat::CHANNEL_OUT_STEREO = get_static_field<int>(c, "CHANNEL_OUT_STEREO");
+      CJNIAudioFormat::CHANNEL_OUT_MONO = get_static_field<int>(c, "CHANNEL_OUT_MONO");
```
MrMC: above stereo so it looks pretty and matches the order define order. Yea, yea OCD is a bitch :)

fritsch: Now looks sooo pretty :-) fixed it up.

fritsch:
jenkins build this please

PR #10814, opened 2016-10-30
[PVR] Guide window: Fix responsiveness on first open. 

ksooo:
Under special circumstances -- for example if guide window is set as kodi startup window or if opened very(!) quickly after kodi startup -- the window can be experienced as kinda "unresponsive", because due to the async update of the window which is per default every 5 seconds (to avoid unnecessary flickering and cpu load) , it can happen that first epg data are displayed after 5 seconds.

This PR fixes this glitch by introducing an initial boost for the window update.

Second commit just adds some more fail safeness (i once experienced a crash there).

note: pvr.demo + guide window as kodi startup window is nice test setup to test the change.

@xhaggi for review?

### DIFF
--- a/xbmc/pvr/windows/GUIWindowPVRGuide.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRGuide.cpp
@@ -515,6 +515,9 @@ bool CGUIWindowPVRGuide::RefreshTimelineItems()
     if (epgGridContainer)
     {
       const CPVRChannelGroupPtr group(GetChannelGroup());
+      if (!group)
+        return false;
+
       std::unique_ptr<CFileItemList> timeline(new CFileItemList);
 
       // can be very expensive. never call with lock acquired.
--- a/xbmc/pvr/windows/GUIWindowPVRGuide.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRGuide.cpp
@@ -695,6 +695,11 @@ CPVRRefreshTimelineItemsThread::CPVRRefreshTimelineItemsThread(CGUIWindowPVRGuid
 
 void CPVRRefreshTimelineItemsThread::Process()
 {
+  static const int BOOSTED_SLEEPS_THRESHOLD = 4;
+
+  int iLastEpgItemsCount = 0;
+  int iUpdatesWithoutChange = 0;
+
   while (!m_bStop)
   {
     if (m_pGuideWindow->RefreshTimelineItems() && !m_bStop)
@@ -702,6 +707,25 @@ void CPVRRefreshTimelineItemsThread::Process()
       CGUIMessage m(GUI_MSG_REFRESH_LIST, m_pGuideWindow->GetID(), 0, ObservableMessageEpg);
       KODI::MESSAGING::CApplicationMessenger::GetInstance().SendGUIMessage(m);
     }
-    Sleep(5000);
+
+    // in order to fill the guide window asap, use a short update interval until we the
+    // same amount of epg events for BOOSTED_SLEEPS_THRESHOLD + 1 times in a row .
+    if (iUpdatesWithoutChange < BOOSTED_SLEEPS_THRESHOLD)
+    {
+      int iCurrentEpgItemsCount = m_pGuideWindow->CurrentDirectory().Size();
+
+      if (iCurrentEpgItemsCount == iLastEpgItemsCount)
+        iUpdatesWithoutChange++;
+      else
+        iUpdatesWithoutChange = 0; // reset
+
+      iLastEpgItemsCount = iCurrentEpgItemsCount;
+
+      Sleep(1000); // boosted update cycle
+    }
+    else
+    {
+      Sleep(5000); // normal update cycle
+    }
   }
 }


Jalle19:
Isn't there some way to determine when we can go back to 5 second intervals without having to specify an arbitrary boost parameter?

ksooo:
> Isn't there some way to determine when we can go back to 5 second intervals without having to specify an arbitrary boost parameter?

 atm, i have no better idea. for instance, if you use async epg transfer (pvr.hts supports this), there you never know when an update starts/ends. there is no fixed interval, progress, ... updates just come in at any time. but ofc i'm open for suggestions.

ksooo:
@Jalle19 what do you think about this more advanced approach?

Jalle19 (approved):
+1 if it works

ksooo:
jenkins build this please

xbmc/pvr/windows/GUIWindowPVRGuide.cpp
```diff
@@ -695,13 +698,37 @@ CPVRRefreshTimelineItemsThread::CPVRRefreshTimelineItemsThread(CGUIWindowPVRGuid
 
 void CPVRRefreshTimelineItemsThread::Process()
 {
+  // we want this number of refresh calls without new data in a row, before going to normal refresh cycle.
+  static const int BOOSTED_CYCLES_THRESHOLD = 5;
```
xhaggi: Should we move this constant to the top out of this method.

xbmc/pvr/windows/GUIWindowPVRGuide.cpp
```diff
@@ -695,13 +698,37 @@ CPVRRefreshTimelineItemsThread::CPVRRefreshTimelineItemsThread(CGUIWindowPVRGuid
...
+
+      if (iCurrentEpgItemsCount == iLastEpgItemsCount)
+        iUpdatesWithoutChange--;
+      else
+        iUpdatesWithoutChange = BOOSTED_CYCLES_THRESHOLD; // reset; we want BOOSTED_CYCLES_THRESHOLD in a row
+
+      iLastEpgItemsCount = iCurrentEpgItemsCount;
+    }
+
+    if (iUpdatesWithoutChange)
```
xhaggi: Please compare it to a number instead of use the bool comparator.

xbmc/pvr/windows/GUIWindowPVRGuide.cpp
```diff
@@ -695,13 +698,37 @@ CPVRRefreshTimelineItemsThread::CPVRRefreshTimelineItemsThread(CGUIWindowPVRGuid
...
     if (m_pGuideWindow->RefreshTimelineItems() && !m_bStop)
     {
       CGUIMessage m(GUI_MSG_REFRESH_LIST, m_pGuideWindow->GetID(), 0, ObservableMessageEpg);
       KODI::MESSAGING::CApplicationMessenger::GetInstance().SendGUIMessage(m);
     }
-    Sleep(5000);
+
+    // in order to fill the guide window asap, use a short update interval until
+    // BOOSTED_CYCLES_THRESHOLD times in a row we get no more epg events between two cycles.
+    if (iUpdatesWithoutChange)
```
xhaggi: Same here

xhaggi (changes requested):

xbmc/pvr/windows/GUIWindowPVRGuide.cpp
ksooo: What would be the benefit? It is not used outside this member function?

xbmc/pvr/windows/GUIWindowPVRGuide.cpp
ksooo: Will do.

xbmc/pvr/windows/GUIWindowPVRGuide.cpp
ksooo: Will do.

xbmc/pvr/windows/GUIWindowPVRGuide.cpp
xhaggi: could we save the second check if we change the code to:

```
static const int BOOSTED_CYCLES_THRESHOLD = 5;

int iLastEpgItemsCount = 0;
int iUpdatesWithoutChange = 0;

while (!m_bStop)
{
  if (m_pGuideWindow->RefreshTimelineItems() && !m_bStop)
  {
    CGUIMessage m(GUI_MSG_REFRESH_LIST, m_pGuideWindow->GetID(), 0, ObservableMessageEpg);
    KODI::MESSAGING::CApplicationMessenger::GetInstance().SendGUIMessage(m);
  }

  // in order to fill the guide window asap, use a short update interval until
  // BOOSTED_CYCLES_THRESHOLD times in a row we get no more epg events between two cycles.
  if (iUpdatesWithoutChange < BOOSTED_CYCLES_THRESHOLD)
  {
    int iCurrentEpgItemsCount = m_pGuideWindow->CurrentDirectory().Size();

    if (iCurrentEpgItemsCount == iLastEpgItemsCount)
      iUpdatesWithoutChange++;
    else
      iUpdatesWithoutChange = 0; // reset; we want BOOSTED_CYCLES_THRESHOLD in a row

    iLastEpgItemsCount = iCurrentEpgItemsCount;

    Sleep(1000); // boosted update cycle
  }
  else 
  {
    Sleep(5000); // normal update cycle
  }    
}
```

ksooo: @xhaggi just one small change... i wanted 5 calls to refresh without new data in a row, which means 4 times sleep(1000). With your suggested change we would do 5 Sleep(1000). So, the value of the constant needs be changed to 4 now. will change accordingly.

ksooo:
@xhaggi requested changes are in. good to go now?

xbmc/pvr/windows/GUIWindowPVRGuide.cpp
xhaggi: Thanks looks better now.

ksooo:
jenkind build this please

ksooo:
jenkins build this please

PR #10820, opened 2016-10-30
Fixed: RTMP swfvfy also accepts 1 instead of just true

basrieter:
## Description

It turns out that SWFVFY=1 will break RTMP playback on Kodi 17.x. The documentation at http://rtmpdump.mplayerhq.hu/librtmp.3.html does show it as a valid option. 
## Motivation and Context

Currently it breaks backward compatibility and if a user goes to the RTMP documentation, they will find a value for swfvfy that will not work with Kodi.
## How Has This Been Tested?

I compiled and tested with my add-on (Retrospect http://www.rieter.net/content/xot/downloads/) by playing a stream from the "Nickelodeon (Sverige)" channel (from the 'Big Time Rush' show).
## Types of change

Just single new or statement added.
## Checklist:


FernetMenta:
thanks!

MartijnKaijser:
jenkins build this please

PR #10822, opened 2016-10-30
[AMLUtils] Update aml_permissions() to check for RW access to more paths

samnazarko:
- /sys/class/vfm/map:

This is normally set in initramfs on OSMC and LibreELEC, but after @codesnake's 59828832455794019c59d46e931b1f96efa312f5, Kodi now needs to be able to write to this file during when setting up playback.
- /sys/class/tsync/enable

RW has been needed for this file for a while, but it was never in aml_permissions() check.


fritsch:
Thx. Jenkins does not build that code path anyways, so direct merge.

PR #10824, opened 2016-10-30
fix streaminfo set by python

ronie:
we need to call DetermineBestStreams() after adding streamdetails.

this fixes retrieving ListItem.VideoCodec (and the likes...) in skins, for listitems created by python.


MartijnKaijser:
Jenkins build this please

PR #10826, opened 2016-10-31
Revert "Picture: Use av_malloc when using ffmpeg scalers"

fritsch:
Currently in a hurry, that broke Pictures. I need to have a look after being back.

Remember: It's not safe to put custom buffers into ffmpeg methods, cause they don't have proper alignment. With the above original code - it happened the other way round :-)

### DIFF
--- a/xbmc/pictures/Picture.cpp
+++ b/xbmc/pictures/Picture.cpp
@@ -42,7 +42,6 @@
 
 extern "C" {
 #include "libswscale/swscale.h"
-#include "libavutil/mem.h"
 }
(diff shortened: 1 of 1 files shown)


fritsch:
Thanks @akva2 for pinging me.

fritsch:
jenkinds build and merge

fritsch:
@akva2 can you please revert the other commit with av's malloc and free or remove the & before the argument of av_free?

fritsch:
Other fix on the way.

PR #10829, opened 2016-10-31
[cmake] add our own FindLibXml2.cmake, needed for static libs

wsnipex:
fixes zlib dependency issues on some platforms when libxml2 is static

@peak3d @fetzerch  ping


wsnipex:
jenkins build this please

PR #10830, opened 2016-10-31
FIX: [rar] deadlock with cbr + net vfs (credit notspiff)

koying:
## Description

from notspiff:

specifically we are called off-thread with the main thread waiting on us while holding the gui lock. so it will deadlock on obtaining dialog. the bt was funkily confusing though, as it will actually show a "downstream" deadlock on the rar manager, as there's two concurrent calls to it (to preload/load images). the deadlock that was triggered in the debugger was the second call to the rar manager spinning on the rar manager lock, which is held by the first call to rar manager, which again is spinning on the gui lock.

oh, and using a "exotic" vfs was only part of the problem to the extent that it's slower so the timeout on the progress dialog kicked in.
## Types of change

## Checklist:

### DIFF
--- a/xbmc/filesystem/RarManager.cpp
+++ b/xbmc/filesystem/RarManager.cpp
@@ -19,6 +19,7 @@
  */
 
 #include "system.h"
+#include "Application.h"
 #include "RarManager.h"
 #include "Util.h"
 #include "utils/CharsetConverter.h"
@@ -74,7 +75,7 @@ public:
   }
   ~progress_info()
   {
-    if (shown)
+    if (shown && g_application.IsCurrentThread())
     {
       // close progress dialog
       CGUIDialogProgress* dlg = (CGUIDialogProgress*)g_windowManager.GetWindow(WINDOW_DIALOG_PROGRESS);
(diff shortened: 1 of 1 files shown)


koying:
jenkins build this please

xbmc/filesystem/RarManager.cpp
```diff
@@ -88,7 +89,7 @@ class progress_info
   bool progress(int progress, const char *text)
   {
     bool cont(true);
-    if (shown || showTime.IsTimePast())
+    if (shown || showTime.IsTimePast() && g_application.IsCurrentThread())
```
popcornmix: This adds a warnng:
```
RarManager.cpp: In member function 'bool progress_info::progress(int, const char*)':
RarManager.cpp:92:40: warning: suggest parentheses around '&&' within '||' [-Wparentheses]
```

koying:
@popcornmix Bad c/p ;)

notspiff:
.. and i see my request wasn't honored and yet another mail got sent.. :p (i assume you haven't read your mail @koying)

koying:
jenkins build this please

MartijnKaijser:
jenkins build this please

PR #10834, opened 2016-10-31
[python] add getTrailer() function to the xbmc.InfoTagVideo() class

ronie:
allow addons to retrieve the trailer from the selected listitem.

### DIFF
--- a/xbmc/interfaces/legacy/InfoTagVideo.h
+++ b/xbmc/interfaces/legacy/InfoTagVideo.h
@@ -546,6 +546,24 @@ namespace XBMCAddon
 #else
       String getFirstAired();
 #endif
+
+#ifdef DOXYGEN_SHOULD_USE_THIS
+      ///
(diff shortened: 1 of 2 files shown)


xbmc/interfaces/legacy/InfoTagVideo.h
```diff
@@ -546,6 +546,24 @@ namespace XBMCAddon
...
+      /// \ingroup python_InfoTagVideo
+      /// @brief \python_func{ getTrailer() }
+      ///-----------------------------------------------------------------------
+      /// To get the path where the trailer is stored.
+      ///
+      /// @return [string] Trailer path
+      ///
+      ///
+      ///-----------------------------------------------------------------------
+      ///
```
phil65: `@python_v17` version info is missing

ronie: thx, added.

MartijnKaijser (approved):

PR #10835, opened 2016-11-01
slideshow: only stop video player, keep background music playing

FernetMenta:
see title

@DaveTBlake this restores background music during slideshow which is more a feature by accident than anything well designed. Now we wait for the bug report that music does not continue after having played a video :)


DaveTBlake:
Thanks for restoring the "happy accident" of background music during pictures.

MartijnKaijser:
jenkins build and merge

PR #10840, opened 2016-11-01
[depends] alsa-lib: fix typo in configure.in

XECDesign:
## Description

Change `${host-os}` to `${host_os}` in `tools/depends/target/alsa-lib`'s `configure.in`.

## Motivation and Context

Currently, if alsa-lib's configure finds `${host_cpu}-${host_os}-gcc`, it will set `$CC` to `${host_cpu}-${host-os}-gcc`. Since the variables are not the same (on the raspberry pi `${host-os}` is set to `arm-unknown`, resulting in `$CC` being a non-existent `arm-arm-unknown-..`), `configure` fails when testing the compiler.

## How Has This Been Tested?

On a raspberry pi:
```
cd tools/depends
./configure --with-platform=raspberry-pi --prefix=${HOME}/opt/kodi
make
```
Without the patch, configure claims the C compiler is not able to produce binaries. With the patch, the build completes.

## Types of change

## Checklist:

### DIFF
--- a/tools/depends/target/alsa-lib/Makefile
+++ b/tools/depends/target/alsa-lib/Makefile
@@ -39,6 +39,7 @@ ifeq ($(OS),android)
 	cd $(PLATFORM); patch -p0 < ../removeshm-3.patch
 	cd $(PLATFORM); patch -p0 < ../timeval.patch
 endif
+	cd $(PLATFORM); patch -p0 < ../host-os.patch
 	cd $(PLATFORM); $(CONFIGURE)
 
 $(LIBDYLIB): $(PLATFORM)
--- a/tools/depends/target/alsa-lib/host-os.patch
+++ b/tools/depends/target/alsa-lib/host-os.patch
@@ -1,0 +1,11 @@
+--- configure.in	2010-04-16 12:11:05.000000000 +0100
++++ configure.in	2016-11-01 15:15:21.645908396 +0000
+@@ -31,7 +31,7 @@
+ 
+   which ${program_prefix}gcc >/dev/null 2>&1 && CC=${program_prefix}gcc
+   which ${host_cpu}-${host_os}-gcc >/dev/null 2>&1 \
+-  && CC=${host_cpu}-${host-os}-gcc
++  && CC=${host_cpu}-${host_os}-gcc
+   which ${host_cpu}-${host_vendor}-${host_os}-gcc >/dev/null 2>&1 \
+   && CC=${host_cpu}-${host_vendor}-${host_os}-gcc
+ 


MartijnKaijser:
@fritsch any idea why we use this ancient also lib?

fritsch:
We don't use newer features :-) @wsnipex you'd prefer a bump instead?

XECDesign:
Is it worthwhile opening a PR of changes required to make builds work like it says on the tin natively on ARM (on a pi3 or a chromebook running Raspbian, in my case)?

There is another change needed to get libssh to build, but it's a little less trivial. The issue is that cmake fails with:
> Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR)

I worked around the problem by modifying the Makefile to pass the necessary `-D` flags to cmake, but that's not portable. I haven't looked at exactly why cmake doesn't find and set those itself. It would be nice to have a place to put such changes and get a bit of review or suggestions from kodi devs.

Or should I just open a trac issue for that?

wsnipex:
Can you please explain how you build?
Usually building natively (not cross compiling) doesn't use depends at all. And cross compiling with depends shouldn't hit that bug at all, since we explicitly set compilers and flags.

XECDesign:
For now, I have only built depends.

The plan is to build kodi and set the appropriate `PATH` and `LD_PRELOAD_PATH` in the wrapper. I have been advised that Debian's packaging is frowned upon and not supported by kodi, that it's best to use the bundled dependencies where possible. and that the official kodi PPA is the way to go. Porting packages from the PPA back to something that works on Debian is proving to be challenging and still heavily relies on system libraries without kodi's patches.

I have run into issues with replacing the distro's own packages in order to meet the required dependencies. For example, libbluray jumps from 0.6.2 to 0.9.2. Many raspbian/debian packages are linked against it and I don't know whether it's entirely backwards compatible. libgif goes from 4.1.6 to 5.0.5. dcadec comes from an entirely different source.

That means that with each release, I would need to go over all of the dependencies changes, figure out which are safe to upgrade and what to do with those that aren't. I have no way of knowing if all the libraries are backwards compatible or whether the SOVERSION of each library is set correctly. I worry that this will only become more and more time consuming over time as new dependencies and patches are added and versions are bumped.

The short version is that I would like a kodi package for raspbian that has all the patches, most compatibility, but doesn't replace system packages and is more maintainable in the long term.

wsnipex:
ok, makes sense, specially for debian stable based distros. Thanks for the explanation.
The PR here is ok and pending a successful build test from jenkins, will be merged.
For the libssh issue: can you either post a log showing the issue on the forum, or hop on IRC (freenode, #kodi-dev) and ping me?

jenkins build this please

PR #10841, opened 2016-11-01
[PVR] Fix epg update deadlock.

ksooo:
I recently experienced deadlocks when opening PVR Guide window, especially if this window was set as Kodi startup window.

Misc stack traces revealed that the deadlock was always caused by CEpg <-> CPVRTimers

Problem was (once more) that it is not really working to call a PVR global (like g_PVRTimers aor g_PVRRecordings) while having own mutex locked. Fix is to just not to do this. ;-)

@Jalle19 for review?


Jalle19 (approved):

ksooo:
jenkins build this please

PR #10842, opened 2016-11-02
[Fix] access violation in CFFmpegImage::Initialize

Rechi:
## Description

Check number of AVStreams is at least one before accessing the first AVStream.

## Motivation and Context
Kodi was crashing if I opened a specific directory.

## How Has This Been Tested?
Runtime tested

## Types of change

## Checklist:

### DIFF
--- a/xbmc/guilib/FFmpegImage.cpp
+++ b/xbmc/guilib/FFmpegImage.cpp
@@ -249,6 +249,12 @@ bool CFFmpegImage::Initialize(unsigned char* buffer, unsigned int bufSize)
     return false;
   }
 
+  if (m_fctx->nb_streams <= 0)
+  {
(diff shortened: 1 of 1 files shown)


FernetMenta:
thanks,
jenkins build and merge

fritsch:
Could you upload that specific sample? Might be sth for upstream, too.

Rechi:
@fritsch I think it's not a direct ffmpeg problem. `CBaseTexture::LoadFromFileInternal` (https://github.com/xbmc/xbmc/blob/master/xbmc/guilib/Texture.cpp#L230) gets called with a directory as texturePath for the faulty directory. Normally it is called with file paths.

fritsch:
Then we should also fix the root cause but this code here is never the less correct. Thx again.

PR #10845, opened 2016-11-03
Fix episode selection when entering a show from the homescreen

razzeee:
## Description

So now, that we can access shows from the homescreen, the call order for the item selection got doesn't work for that order.

## Motivation and Context

http://trac.kodi.tv/ticket/16793#comment:5

## How Has This Been Tested?

Only tested on a few shows on my dev pc. Accessing them from homescreen and library.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/video/windows/GUIWindowVideoNav.cpp
+++ b/xbmc/video/windows/GUIWindowVideoNav.cpp
@@ -130,6 +130,10 @@ bool CGUIWindowVideoNav::OnMessage(CGUIMessage& message)
       if (!CGUIWindowVideoBase::OnMessage(message))
         return false;
 
+      // This needs to be done again, because the initialization of CGUIWindow overwrites it with default values
+      // Mostly affects cases where GUIWindowViedeoNav is constructed and we're already in a show, e.g. entering from the homesceen
+      SelectFirstUnwatched();
+
       return true;
     }
     break;
@@ -284,6 +288,12 @@ bool CGUIWindowVideoNav::Update(const std::string &strDirectory, bool updateFilt
   if (!CGUIWindowVideoBase::Update(strDirectory, updateFilterPath))
     return false;
 
+  SelectFirstUnwatched();
+
+  return true;
+}
+
+void CGUIWindowVideoNav::SelectFirstUnwatched() {
   // Check if we should select the first unwatched item
   SelectFirstUnwatchedItem selectFirstUnwatched = GetSettingSelectFirstUnwatchedItem();
   if (selectFirstUnwatched != SelectFirstUnwatchedItem::NEVER)
@@ -302,8 +312,6 @@ bool CGUIWindowVideoNav::Update(const std::string &strDirectory, bool updateFilt
       m_viewControl.SetSelectedItem(iIndex);
     }
   }
-
-  return true;
 }
 
 bool CGUIWindowVideoNav::GetDirectory(const std::string &strDirectory, CFileItemList &items)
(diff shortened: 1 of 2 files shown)


Montellese:
What is the difference in order with this change? `CGUIMediaWindow` calls `Update()` through `Refresh()` in `OnInitWindow()` which is called from `OnMessage(GUI_MSG_WINDOW_INIT)` so the result should be the same or not?

But +1 on moving that logic into a separate method :-)

FernetMenta:
CGUIWindow::OnInitWindow sets selected item to default. Custom selection needs to be done after this.

FernetMenta:
Specializations are best done at a place where it is mot obvious. CGUIWindowVideoNav overrides OnMessage and handles GUI_MSG_WINDOW_INIT. There is already extra code at this place.
When you look at CGUIWindowVideoNav there is no indicator that OnInitWindow does even exist.

Selecting an particular item is not related to init of a window. The logical sequence is to first init a window, then set the cursor on an item.

Montellese:
You do realize that `OnInitWindow()` was initially added to replace having to manually handle `GUI_MSG_WINDOW_INIT` right? That's why I try to do everything in `OnInitWindow()` because it's much easier to jump to method implementations that to a specific part of a huge `switch` statement.

Your argument that selecting an item is not related to init of a window is no argument against `OnInitWindow()` because then it should also not be part of `GUI_MSG_WINDOW_INIT` which does exactly the same.

FernetMenta:
IMO specializations should be done first at the most outer level. Handling GUI_MSG_WINDOW_INIT is before OnInitWindow.

> You do realize that OnInitWindow() was initially added to replace having to manually handle GUI_MSG_WINDOW_INIT right?

Nope, how should I? I suggest to remove one or the other method to avoid further confision.

Razzeee:
So anything I should fix or can this go in like this?
@Montellese @FernetMenta

FernetMenta:
@Razzeee I am ok with this, but not sure about @Montellese

Montellese:
At least add a comment explaining why it's necessary because it's not obvious at all.

Razzeee:
Will do, good comment :)

Razzeee:
@Montellese 
I've added a comment please check if that's clear enough

Razzeee:
jenkins build this please

Razzeee:
Whoops, let's try that again.

jenkins build this please

Montellese:
Is Github messing with me? I don't see any comment in the code...

Razzeee:
No, it's not. 
I had to fix every platform other than Windows yesterday, seems like my branch wasn't up to date. Probably force pushed my comment into nirvana. :(
Will update later, Thx.

Razzeee:
@Montellese 
Updated, sorry :)

Montellese (approved):
Thanks

Razzeee:
jenkins build and merge

hudokkow:
a bunch of builds failed to start
jenkins build and merge

Razzeee:
Seems to be back in v18

Razzeee:
Bisect nailed it down to four possibles:

```
There are only 'skip'ped commits left to test.
The first bad commit could be any of:
01dbdf956f47e1df3c96261d154848b673fe996a
11bb44266299d3f7ca784f0c69d827297ccb94a9
96602b959f20a42c84924b7bebe51d53b8df9c9d
111af227aec2e530ccea359443023f1e95fd912f
We cannot bisect more!
```

PR #10857, opened 2016-11-04
Align play button in context menu to behavior in library

razzeee:
## Description

Use the global way to play files not the one from GUIWindowVideoNav, which creates a playlist for some reason. Before this PR, if you go into the library and open the info for a movie, then choose play from there, you will end up with a playlist running. After this PR you won't have anything in your playlist when you do that. Same as when you start playback from anywhere else.

## Motivation and Context

http://trac.kodi.tv/ticket/16982

## How Has This Been Tested?

Tested on win10 from various parts of the places where the info window is available from.

## Types of change

## Checklist:

Closes #16982

### DIFF
--- a/xbmc/video/ContextMenus.cpp
+++ b/xbmc/video/ContextMenus.cpp
@@ -21,6 +21,7 @@
 #include "ContextMenus.h"
 #include "Application.h"
 #include "Autorun.h"
+#include "playlists/PlayList.h"
 #include "Util.h"
 #include "video/dialogs/GUIDialogVideoInfo.h"
 #include "video/windows/GUIWindowVideoBase.h"
@@ -119,7 +120,17 @@ static void SetPathAndPlay(CFileItem& item)
     item.SetPath(item.GetVideoInfoTag()->m_strFileNameAndPath);
   }
   item.SetProperty("check_resume", false);
-  g_application.PlayMedia(item, "", PLAYLIST_NONE);
+
+  CFileItemPtr movieItem((std::make_shared<CFileItem>(item)));
+
+  g_playlistPlayer.Reset();
+  g_playlistPlayer.SetCurrentPlaylist(PLAYLIST_VIDEO);
+  PLAYLIST::CPlayList& playlist = g_playlistPlayer.GetPlaylist(PLAYLIST_VIDEO);
+  playlist.Clear();
+  playlist.Add(movieItem);
+
+  // play movie...
+  g_playlistPlayer.Play(0, "");
 }
 
 bool CResume::Execute(const CFileItemPtr& itemIn) const


Montellese:
This might break a few JSON-RPC clients (including Kore) which might be relying on the fact that there's a playlist when something is playing. IMO that should be the correct case i.e. whenever we start playing something it must first be added to a playlist.

Razzeee:
If that's the case we do now have 98% broken JSON-RPC clients. We only add stuff to a playlist, when you play it from the video info screen.
And yes, I would have also thought that the correct case would be that everything goes to a playlist.

da-anda:
I know it's not directly related to this PR, but as I've seen it in the changed codebase: Wouldn't it make sense to also support resume point handling everywhere and not just within the VideoInfo dialog? So the "playFile" should probably handle resume points? Also, why do we create a copy of the original fileitem in first place? But that's a whole different story.

Razzeee:
We do support it everywhere the video info is just different as we would
have a potential endless loop o is we would use the default handling. As
that would check your setting for the play action and if it's too too video
info, you'll be trapped.

Razzeee:
Btw the mentioned play and add to playlist case is still around, even after this PR. When you select and start a file via the "Play only this" context menu entry, I don't think I've ever seen it, but it's in the code.

Razzeee:
So can this be merged?

axbmcuser:
Am i understanding correctly, that after this PR the handling which imho feels right, namely "always create a 1-item-playlist when starting playback of a single video" is not used anymore? 

If so:
I understand that the removal of the playlist creation from video info dialog looks like it's good in terms of uniformity, but in terms of end user experience and skinning possibilities (making already great use of the created playlist item) it would only be an improvement if there would be more cases where starting playback of a single video creates the mentioned 1-item-playlist - not less cases. (at best, it would always be the case, of course)

So, for what it's worth, until changes would go in that direction, i'd consider this change +1 in theoretical code uniformity, but a -2 for losing skinnig possibilities and therefore end user disadvantage. Would it be a possibility to leave as is and reconsider improving for v18?

Thanks for looking into it. :) Just my 2 cents.

Razzeee:
I don't get how this impacts skinners in any important way and even the end users shouldn't be affected. Why would I care if I get a 1 item playlist when I start a video from the info screen (I mostly start my videos by clicking on it, so I would not know)
And yes, uniformity is what this brings, probably/hopefully to make it easier to get everything moved to the preferred/same type of handling, whatever that might be. But well, small changes for now.

phil65:
The so-called "skinning possibilities" refer to a hack which may not be possible anymore after this. Not something we need to care about imo.

axbmcuser:
Yes, i know that it's only used from the video info dialog right now. Since it's a very common way for users to start playing a movie, this case is used not that rarely according to that. 
Skins/users use the playlist for being able to open the video info dialog directly while playing a video on VideoFullscren/VideoOSD (In short: Jumping to the "now playing"-playlist and execute "Info").

Looks like there may be new ways in v18 to open the video info dialog from VideoFullscren/VideoOSD directly, but until then, this way seems the only/most reliable way to bypass this missing feature/possibility and there are users happily using it. Why not time this PR simultaneously with the most likely better ways of v18 which are about to come?

But well. As said. Just my 2 cents.

BigNoid:
Not sure what the fuss is about. You can still do that with this PR, the same as you can do that now when starting an item from the listings.
In any case you can use [service.library.data.provider](https://github.com/BigNoid/service.library.data.provider#single-item) for that functionality in Krypton too when it's on the repo, that will save you from opening a window first.

axbmcuser:
Great. Thanks for this as it looks for most ppl unknown solution. Some quick tests with an invisible list using "service.library.data.provider" content look promising.
Hoping that besides the install of the required additional addon there are no other yet unknown downsides, i will add an info about this to the forums, since all threads i saw discussing this issue ended up with no answers or people thinking that there just i no good or even clean way to achieve this. I'll be happily sending the old workaround to nirvana. :*) Thanks again.

DaveTBlake:
Can I have a sanity check here, or maybe I misunderstand. The correct behaviour when playing something, even a single item, is that it is added to a current playlist. At the moment laying a video from info does this, even when other places don't. So with this PR we are removing this correct behaviour to make it consistent with the other places that behave wrongly? 

That just does not sound like a fix to me.

Razzeee:
Well, that would be the ideal case yes.
But an inconsistent handeln is worse than an incorrect one (this is not
affecting any database)

So:
1. Correct behavior
2. Consistent behavior
3. Inconsistent behavior
4. Wrong behavior

That's what I think priority wise. And it should hopefully be much easier
to change to the correct one when every piece of code is channeled through
the same code.

Razzeee:
Had a look at this again and changed the direction.

@tamland might be interested and maybe even @ksooo as this seems to also affect PVR (which I can't test)

Razzeee:
jenkins build this

Razzeee:
jenkins build this please

xbmc/video/ContextMenus.cpp
```diff
@@ -119,7 +120,17 @@ static void SetPathAndPlay(CFileItem& item)
     item.SetPath(item.GetVideoInfoTag()->m_strFileNameAndPath);
   }
   item.SetProperty("check_resume", false);
-  g_application.PlayMedia(item, "", PLAYLIST_NONE);
+
+  CFileItemPtr movieItem((std::make_shared<CFileItem>(item)));
```
ksooo: I would suggest to write <code>const CFileItemPtr movieItem(new CFileItem>(item));</code> here. "const as const can" and less letters to type for the object creation.

xbmc/video/ContextMenus.cpp
```diff
@@ -119,7 +120,17 @@ static void SetPathAndPlay(CFileItem& item)
...
+
+  g_playlistPlayer.Reset();
+  g_playlistPlayer.SetCurrentPlaylist(PLAYLIST_VIDEO);
+  PLAYLIST::CPlayList& playlist = g_playlistPlayer.GetPlaylist(PLAYLIST_VIDEO);
+  playlist.Clear();
+  playlist.Add(movieItem);
+
+  // play movie...
```
ksooo: Isn't it obvious that the next line will start playing the movie? I would only write comments for things not obvious.

xbmc/video/ContextMenus.cpp
Montellese: IMO `std::make_shared<>()` should always be preferred if possible. It's more efficient and exception-safe.

xbmc/video/ContextMenus.cpp
Razzeee: It's basically a copy from a different file, so I tried to keep it similiar. But will remove.

xbmc/video/ContextMenus.cpp
ksooo: Exception safe? 

=> http://en.cppreference.com/w/cpp/memory/shared_ptr/make_shared => "May throw std::bad_alloc or any exception thrown by the constructor of T. If an exception is thrown, this function has no effect."

And why is it more efficient? Don't get this.

Montellese: It is more efficient because `std::make_shared` only requires a single heap allocation whereas `CFileItemPtr(new CFileItem())` requires two heap allocations (one for the reference counter and one for the item itself).

And yes it can throw exceptions itself but in that case the `std::shared_ptr` instance hasn't been created yet so there's no memory leak. If you however do something like
```
SomeMethod(CFileItemPtr(new CFileItem("foo")), CFileItemPtr(new CFileItem("bar")))
```
the compiler can decide in what order to execute the statements. He could choose to do
1. `new CFileItem("foo")`
2. `new CFileItem("bar")`
3. `CFileItemPtr(foo)`
4. `CFileItemPtr(bar)`

and for some reason you get an exception in step 2 there will be a memory leak because step 1 has already been executed and you can't get to that memory anymore. This can't happen with
```
SomeMethod(std::make_shared<CFileItem>("foo"), std::make_shared<CFileItem>("bar"))
```

Razzeee:
I would like to merge this in the next days, so please speak up if somebody sees something wrong.

Razzeee:
jenkins build and merge

dgam1:
For anyone who's interested, my PR #11747 fixes two other cases where playback starts without playlists.

PR #10858, opened 2016-11-04
[python] add dbid to musicinfotag

ronie:
allow python addons to get/set the dbid of a (music) listitem.

added logging for invalid videoinfo tags (we already do this for musicinfo).

### DIFF
--- a/xbmc/interfaces/legacy/InfoTagMusic.h
+++ b/xbmc/interfaces/legacy/InfoTagMusic.h
@@ -67,6 +67,24 @@ namespace XBMCAddon
 #ifdef DOXYGEN_SHOULD_USE_THIS
       ///
       /// \ingroup python_InfoTagMusic
+      /// @brief \python_func{ getDbId() }
+      ///-----------------------------------------------------------------------
+      /// Get identification number of tag in database.
+      ///
+      /// @return [integer] database id.
+      ///
+      ///
+      ///-----------------------------------------------------------------------
+      /// @python_v17 New function added.
+      ///
+      getDbId();
+#else
+      int getDbId();
+#endif
+
+#ifdef DOXYGEN_SHOULD_USE_THIS
+      ///
+      /// \ingroup python_InfoTagMusic
       /// @brief \python_func{ getURL() }
       ///-----------------------------------------------------------------------
       /// Returns url of source as string from music info tag.
--- a/xbmc/interfaces/legacy/ListItem.cpp
+++ b/xbmc/interfaces/legacy/ListItem.cpp
@@ -479,6 +479,8 @@ namespace XBMCAddon
             else
               CLog::Log(LOGWARNING, "Invalid media type \"%s\"", value.c_str());
           }
+          else
+            CLog::Log(LOGERROR,"NEWADDON Unknown Video Info Key \"%s\"", key.c_str());
         }
       }
       else if (strcmpi(type, "music") == 0)
--- a/xbmc/interfaces/legacy/ListItem.cpp
+++ b/xbmc/interfaces/legacy/ListItem.cpp
@@ -483,6 +483,25 @@ namespace XBMCAddon
       }
       else if (strcmpi(type, "music") == 0)
       {
+        std::string type;
+        for (auto it = infoLabels.begin(); it != infoLabels.end(); ++it)
+        {
+          String key = it->first;
+          StringUtils::ToLower(key);
+          const InfoLabelValue& alt = it->second;
+          const String value(alt.which() == first ? alt.former() : emptyString);
+
+          if (key == "mediatype")
+          {
+            if (CMediaTypes::IsValidMediaType(value))
+            {
+              type = value;
+              item->GetMusicInfoTag()->SetType(value);
+            }
+            else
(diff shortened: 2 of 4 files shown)


xbmc/interfaces/legacy/InfoTagMusic.h
```diff
@@ -67,6 +67,24 @@ namespace XBMCAddon
...
       /// \ingroup python_InfoTagMusic
+      /// @brief \python_func{ getDbId() }
+      ///-----------------------------------------------------------------------
+      /// Get identification number of tag in database.
+      ///
+      /// @return [integer] database id
+      ///
+      ///
+      ///-----------------------------------------------------------------------
+      ///
```
phil65: version note is missing.

phil65: I think this returns -1 in case there is no dbid? might be worth adding a comment for that.

ronie: wish that was all :-)
still working on it....

phil65: ok. 👍

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -479,6 +479,20 @@ namespace XBMCAddon
       }
       else if (strcmpi(type, "music") == 0)
       {
+        std::string type;
+        for (InfoLabelDict::const_iterator it = infoLabels.begin(); it != infoLabels.end(); ++it)
```
phil65: auto + range based for loop

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -479,6 +479,20 @@ namespace XBMCAddon
...
+        {
+          String key = it->first;
+          StringUtils::ToLower(key);
+          const InfoLabelValue& alt = it->second;
+          const String value(alt.which() == first ? alt.former() : emptyString);
+
+          if (key == "mediatype")
```
phil65: you can combine these two checks with AND.

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -479,6 +479,20 @@ namespace XBMCAddon
...
+          {
+            if (CMediaTypes::IsValidMediaType(value))
+              type = value;
```
phil65: You can also already set the mediatype here (setType), that way you can remove it from below.

ronie: good thinking, thx!

ronie:
updated

MartijnKaijser (approved):

tamland:
I'm realizing it was a big mistake exposing dbid as int in the api as this will most likely break with the future database refactor. To be future proof it should be returned as a string.

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -488,7 +509,9 @@ namespace XBMCAddon
           const String value(alt.which() == first ? alt.former() : emptyString);
 
           //! @todo add the rest of the infolabels
-          if (key == "tracknumber")
+          if (key == "dbid" and !type.empty())
```
ronie: lol, fixed :-)

MartijnKaijser:
So if you want to switch it to be a string we'd better do it now before we hit RC (same for video?)

ronie:
no idea.. json returns dbid's as an int as well, is that also a problem?

MartijnKaijser:
Only if the new db is going to return something else as an int somewhere in the future

ronie:
i'd say let's address all that when we get there.
'till then we should keep things unified between interfaces imo.

tamland:
For this PR, at least remove the whole '-1' documentation as api users have no business knowing about/relying on this implementation detail.

Depending on how much is already use this, it might be worth fixing now rather than later when more addons are using it. The main issue here is that the id type is a such a low level internal thing of the database. A *slight* api misuse and you have coupled an addon to the current database.

ronie:
ok, i'll drop the '-1' comment.

as for converting it to a string, i'm afraid i don't understand the problem.
in my simple logic dbid is an int, same like year is an int, so we return it as an int.

if we return it as a string, i can convert it back to an int in my python code, then what have we gained?

ping @phil65 since your addons are the ones that are using it currently.

phil65:
Will adapt in case we change behaviour.

tamland:
>as for converting it to a string, i'm afraid i don't understand the problem.
in my simple logic dbid is an int, same like year is an int, so we return it as an int.

>if we return it as a string, i can convert it back to an int in my python code, then what have we gained?

The point is: dbids are opaque identifiers, not ints. they just happens to currently be represented by ints. Using a string is just *one way* to make it very explicit that it is opaque. If you don't treat it like that and do things with it, that's your own doing. But just go ahead with this. I can make a pr to 'clarify' this api.

ronie:
thanx for the feedback.
i'm in no rush to get this in.

will get over it again when the time comes to merge this.

ronie:
jenkins build this please

PR #10859, opened 2016-11-04
Code cleanup for GUIDialogVideoInfo

razzeee:
## Description

Just some cleanup with the help of resharper

## Motivation and Context

It's just cleanup, nothing in that regard.

## How Has This Been Tested?

Tested the basic functions of the window.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/video/dialogs/GUIDialogVideoInfo.cpp
+++ b/xbmc/video/dialogs/GUIDialogVideoInfo.cpp
@@ -50,8 +50,6 @@
 #include "GUIUserMessages.h"
 #include "TextureCache.h"
 #include "music/MusicDatabase.h"
-#include "URL.h"
-#include "video/VideoLibraryQueue.h"
 #include "video/VideoThumbLoader.h"
 #include "filesystem/Directory.h"
 #include "filesystem/VideoDatabaseDirectory.h"
@@ -93,6 +91,7 @@ CGUIDialogVideoInfo::CGUIDialogVideoInfo(void)
   m_bRefresh = false;
   m_hasUpdatedThumb = false;
   m_hasUpdatedUserrating = false;
+  m_bViewReview = false;
   m_castList = new CFileItemList;
   m_loadType = KEEP_IN_MEMORY;
   m_startUserrating = -1;
@@ -251,7 +250,7 @@ void CGUIDialogVideoInfo::OnInitWindow()
   // Disable video user rating button for plugins as they don't have tables to save this
   CONTROL_ENABLE_ON_CONDITION(CONTROL_BTN_USERRATING, !m_movieItem->IsPlugin());
 
-  VIDEODB_CONTENT_TYPE type = (VIDEODB_CONTENT_TYPE)m_movieItem->GetVideoContentType();
+  VIDEODB_CONTENT_TYPE type = static_cast<VIDEODB_CONTENT_TYPE>(m_movieItem->GetVideoContentType());
   if (type == VIDEODB_CONTENT_TVSHOWS || type == VIDEODB_CONTENT_MOVIES)
     CONTROL_ENABLE_ON_CONDITION(CONTROL_BTN_GET_FANART, (CProfilesManager::GetInstance().GetCurrentProfile().canWriteDatabases() || g_passwordManager.bMasterUser) && !StringUtils::StartsWithNoCase(m_movieItem->GetVideoInfoTag()->GetUniqueID().c_str() + 2, "plugin"));
   else
@@ -283,7 +282,7 @@ bool CGUIDialogVideoInfo::OnAction(const CAction &action)
   return CGUIDialog::OnAction(action);
 }
 
-void CGUIDialogVideoInfo::SetUserrating(int userrating)
+void CGUIDialogVideoInfo::SetUserrating(int userrating) const
 {
   userrating = std::max(userrating, 0);
   userrating = std::min(userrating, 10);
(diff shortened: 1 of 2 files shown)


xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -471,12 +469,12 @@ void CGUIDialogVideoInfo::OnSearch(std::string& strSearch)
...
   if (items.Size())
   {
-    CGUIDialogSelect* pDlgSelect = (CGUIDialogSelect*)g_windowManager.GetWindow(WINDOW_DIALOG_SELECT);
+    CGUIDialogSelect* pDlgSelect = static_cast<CGUIDialogSelect*>(g_windowManager.GetWindow(WINDOW_DIALOG_SELECT));
     pDlgSelect->Reset();
     pDlgSelect->SetHeading(CVariant{283});
 
     CVideoThumbLoader loader;
-    for (int i = 0; i < (int)items.Size(); i++)
+    for (int i = 0; i < int(items.Size()); i++)
```
ace20022: Is an int really required?

xbmc/video/dialogs/GUIDialogVideoInfo.h
```diff
@@ -31,18 +31,18 @@ class CGUIDialogVideoInfo :
 public:
   CGUIDialogVideoInfo(void);
   virtual ~CGUIDialogVideoInfo(void);
-  virtual bool OnMessage(CGUIMessage& message);
-  virtual bool OnAction(const CAction &action);
+  bool OnMessage(CGUIMessage& message) override;
+  bool OnAction(const CAction &action) override;
```
garbear: i prefer to leave the `virtual` keywords because it's easier to spot a keyword at the beginning of a line as opposed to the end, but to each his own

Montellese: In the beginning I did that as well but then I changed to the opinion that it should only be there if there's also someone overriding it.

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -283,7 +282,7 @@ bool CGUIDialogVideoInfo::OnAction(const CAction &action)
   return CGUIDialog::OnAction(action);
 }
 
-void CGUIDialogVideoInfo::SetUserrating(int userrating)
+void CGUIDialogVideoInfo::SetUserrating(int userrating) const
```
Montellese: Are you sure about this? Line 291 modifies `m_movieItem` so the call to `CFileITem::GetVidoeInfoTag()` requires a non-cost class member.

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -505,10 +503,8 @@ void CGUIDialogVideoInfo::OnSearch(std::string& strSearch)
   }
 }
 
-/// \brief Make the actual search for the OnSearch function.
-/// \param strSearch The search string
-/// \param items Items Found
-void CGUIDialogVideoInfo::DoSearch(std::string& strSearch, CFileItemList& items)
+
```
Montellese: Unneeded empty line.

Montellese (changes requested):
Looks good apart from the two open points.

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
Paxxi: And if it's required then it should be a static_cast

xbmc/video/dialogs/GUIDialogVideoInfo.h
Paxxi: I understand both sides, when not using an ide it's easier to spot the virtual. When using ReSharper or Clion it's distracting seeing it marked as unnecessary code, I lean towards getting rid of it though

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
Razzeee: Yep, removing it.

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
Razzeee: Thanks, missed that one.

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
Razzeee: Well I actually wondered about that too yesterday, so I did try to set an userrating and it seemed to did work without problems. I will try again just in case I missed it going wrong.

xbmc/video/dialogs/GUIDialogVideoInfo.h
Paxxi: Formatting framework? As in a rule file for clang format?

Razzeee: Yes

Paxxi: I like it!

tamland: [core guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final) says no

Paxxi: Seems sane following those and with the core guidelines checkers it should be easy to stay consistent

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
Razzeee: Just checked again, ratings can still be set and persist, even after a restart.

Razzeee:
This should be good to go then.

PR #10860, opened 2016-11-05
Extract digits also from the new ratings' vote tag

phate89:
Some scraper website might have formatted vote tag so we need to parse them (for example imdb uses 700,000)


phate89:
jenkins build this please

PR #10863, opened 2016-11-05
[cmake] Unify usage of CMAKE_CROSSCOMPILE throughout the project

fetzerch:
## Description
With @wsnipex we decided to set CMAKE_CROSSCOMPILING always when we use a Toolchainfile (and build with depends). This was so far not the case on OSX. In order to allow this, we introduce another variable CORE_HOST_IS_TARGET that determines if we can run build artefacts such as unit tests on the build host.

+includes some cleanups of the toolchain files.

It's also worth mentioning, that with this change, we can build cover coverage of the unit tests on linux on jenkins.

## Motivation and Context
Consistency, and fixes a bug that currently on jenkins `make test` will not work because because CMAKE_CROSSCOMPILING is set and hence testing related targets are not created.

## How Has This Been Tested?
Will need jenkins for testing.

## Types of change

## Checklist:


wsnipex:
looks good

fetzerch:

fetzerch:

project/cmake/modules/FindPython.cmake
```diff
@@ -5,7 +5,7 @@
 # PYTHON_INCLUDE_DIRS - the python include directory
 # PYTHON_LIBRARIES - The python libraries
 
-if(PKG_CONFIG_FOUND AND NOT CMAKE_CROSSCOMPILING)
```
fetzerch: @wsnipex: Are you familiar with the constrains of this module? (going back in the history i could see that you added the NOT CMAKE_CROSSCOMPILING here). according to the build log, i'd also have to exclude linux here now. which basically means we only use pkgconfig on osx.

fetzerch:

fetzerch:
d359195 restores the behavior we had before, the FindPython module needs more love, but that can happen independently.

fetzerch:
@wsnipex: good to go?

project/cmake/modules/FindPython.cmake
```diff
@@ -5,7 +5,7 @@
...
+if(PKG_CONFIG_FOUND AND (NOT CMAKE_CROSSCOMPILING OR APPLE))
```
wsnipex: what I was trying to say above is that _all platforms_ but windows should work with pkgconfig.

wsnipex:
can we get this in? We can fix the findpython module later.

fetzerch:
yes, fine by me

PR #10870, opened 2016-11-05
[PVR] context menu rewrite & major gui actions refactoring

ksooo:
This all started with the wish to support context menus for the new Estuary PVR home screen widgets and quickly turned big, because i wanted to avoid duplicate code. At the end i was in a rage to eliminate as many duplicate pvr gui actions code as possible.

This is not completely finished, but all runtime tested and working well.

@xhaggi @Jalle19 for review and general comments?
@tamland any comments regarding the new CContextMenuManager related code pieces?


tamland:
wow, all at once. great work!

xbmc/ContextMenuManager.cpp
```diff
@@ -72,7 +73,12 @@ void CContextMenuManager::Init()
       std::make_shared<CONTEXTMENU::CMarkWatched>(),
       std::make_shared<CONTEXTMENU::CMarkUnWatched>(),
   };
+
   ReloadAddonItems();
+
+  std::vector<std::shared_ptr<IContextMenuItem>> pvrItems = m_pvrMgr.GetContextMenuItems();
```
tamland: Not sure what you had in mind with the CPVRContextMenuManager as it doesn't seem to do anything but hold the static items. I think it's fine to just include all the classes directly like above (for now, this whole part is a candidate for refactoring anyway). In a way they are 'easier' dependencies to deal with than the huge pvr manager.

xbmc/pvr/PVRContextMenus.cpp
```diff
@@ -0,0 +1,785 @@
...
+      item->m_lStartOffset = 0; // must always be set if PlayRecording is called with bCheckResume == false
+      return CPVRGUIActions::GetInstance().PlayRecording(item, false /* bPlayMinimized */, false /* bCheckResume */);
+    }
+
+    ///////////////////////////////////////////////////////////////////////////////
+    // Mark watched
+
+    std::string MarkWatched::GetLabel(const CFileItem &item) const
+    {
+      return g_localizeStrings.Get(16103); /* Mark as watched */
```
tamland: There's a `CStaticContextMenuAction` class, which can be used to reduce the boilerplate a bit for cases like these where you have a static label.

xbmc/ContextMenuManager.cpp
ksooo: I want to stick with with <code>CPVRContextMenuManager </code> for decoupling purposes. I'm trying to keep all pvr code in xbmc/pvr. But you are perfectly right. It makes perfect sense to get rid of the dependency to the "heavy" <code>CPVRMananager</code>. I will change the PR accordingly. Thanks for the hint.

xbmc/pvr/PVRContextMenus.cpp
ksooo: Cool. I will change the code to use <code>CStaticContextMenuAction</code> where possible.

ksooo:
@tamland thanks for the review. The changes you suggested are now in.

xbmc/pvr/PVRContextMenus.cpp
```diff
@@ -0,0 +1,730 @@
...
+
+    ///////////////////////////////////////////////////////////////////////////////
+    // Play recording
+
+    std::string PlayRecording::GetLabel(const CFileItem &item) const
+    {
+      const CPVRRecordingPtr recording(item.GetPVRRecordingInfoTag());
+      if (recording)
+      {
+        if (!recording->IsDeleted())
```
Jalle19: This and the previous `if` could be combined

xbmc/pvr/PVRContextMenus.cpp
```diff
@@ -0,0 +1,730 @@
...
+      CPVRRecordingPtr recording;
+
+      const CEpgInfoTagPtr epg(item.GetEPGInfoTag());
+      if (epg)
+        recording = epg->Recording();
+
+      if (!recording)
+        recording = item.GetPVRRecordingInfoTag();
+
+      if (recording)
```
Jalle19: Could be `else` for the previous `if` clause instead

xbmc/pvr/PVRContextMenus.cpp
```diff
@@ -0,0 +1,730 @@
...
+    bool EditTimerRule::IsVisible(const CFileItem &item) const
+    {
+      CPVRTimerInfoTagPtr timer;
+
+      const CEpgInfoTagPtr epg(item.GetEPGInfoTag());
+      if (epg)
+        timer = epg->Timer();
+
+      if (!timer)
+        timer = item.GetPVRTimerInfoTag();
```
Jalle19: This logic seems to be duplicated in quite a few places now, now would probably be a good time to factor it out to a separate method somewhere.

xbmc/pvr/PVRContextMenus.h
```diff
@@ -0,0 +1,45 @@
...
+  public:
+    static CPVRContextMenuManager& GetInstance();
+
+    std::vector<std::shared_ptr<IContextMenuItem>> GetMenuItems() const { return m_items; }
+
+  private:
+    CPVRContextMenuManager();
+    CPVRContextMenuManager(const CPVRContextMenuManager&) = delete;
+    CPVRContextMenuManager const& operator=(CPVRContextMenuManager const&) = delete;
+    virtual ~CPVRContextMenuManager() {}
```
Jalle19: `= default`

xbmc/pvr/windows/GUIWindowPVRChannels.cpp
```diff
@@ -150,7 +120,6 @@ bool CGUIWindowPVRChannels::OnAction(const CAction &action)
 {
   switch (action.GetID())
   {
-    case REMOTE_0:
```
Jalle19: Is this correct?

da-anda:
I know I'm annoying, but I noticed that you kindly added some file reference comments to the strings.po file and now should also know their context. So I wanted to ask if you're also in the mood to add a small contextual help comment to those strings? At least an indication if a certain string is a settings option or a settings label. Thanks.

ksooo:
@da-anda will do, remember, this is still wip.

ksooo:
jenkins build this please

xbmc/pvr/PVRContextMenus.cpp
ksooo: Very good point.

xbmc/pvr/PVRContextMenus.h
ksooo: Yep.

xbmc/pvr/windows/GUIWindowPVRChannels.cpp
ksooo: What do you expect if you press '0' when in channels window? '1-9' is fine to init numeric input dialog imo, but zero?

xbmc/pvr/PVRContextMenus.cpp
ksooo: okay

xbmc/pvr/PVRContextMenus.cpp
ksooo: nope. look closer. ;-)

ksooo:
@Jalle19 thanks for the review. The changes you suggested are now in.

xbmc/pvr/PVRContextMenus.cpp
ksooo: Stupid github diff. The function above is not what you initially commented on, but the result of my refactoring...

ksooo:
@ronie mind taking a look at the strings.po cleanup commit? I inserted several "#. @@@ unused?" comments. maybe you can tell for some of those whether they are actually unused?

ronie:
these are used by several skins:
13113
13204
19005
19080
19081
19144
19145
19203
19207
19216
19284

ksooo:
> these are used by several skins:

thanks @ronie. will change remarks accordingly.

ksooo:
jenkins build this please

ksooo:
jenkins build this please

ksooo:
jenkins build this please

ksooo:
As I do not want to rebase this "forever" and i consider the code stable enough and well runtime-tested for months I will merge this tomorrow, if nobody objects.

ksooo:
jenkins build this please

ksooo:
jenkins build this please

xbmc/pvr/PVRContextMenus.cpp
```diff
@@ -0,0 +1,694 @@
...
+#include "pvr/channels/PVRChannel.h"
+#include "pvr/PVRGUIActions.h"
+#include "pvr/PVRManager.h"
+#include "pvr/recordings/PVRRecording.h"
+#include "pvr/recordings/PVRRecordingsPath.h"
+#include "settings/Settings.h"
+#include "pvr/timers/PVRTimers.h"
+#include "utils/URIUtils.h"
+
+#include "PVRContextMenus.h"
```
Montellese: IMO the header file belonging to the source file should always be the very first include in the source file. That way it is easier to spot missing includes in the header file. The way it is now you already (recursively) include a lot of other include files which might accidentally be missing from this specific header file.

ksooo: @Montellese Good point. I guess I will come up with a PR reordering all PVR/EPG headers to follow this scheme:

1.    h file corresponding to this cpp file (if applicable)
2.    headers from the same component
3.    headers from other components
4.    system headers

Montellese: Do we have something on this topic in our guidelines? At work we always follow the Google Coding Guidelines, see https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes. There the order is
1. header file corresponding to this cpp file
2. C system files
3. C++ system files
4. Other libraries' header files
5. Own header files

ksooo: AFAIK we have nothing for this in our coding guidelines and currently you can find any possible combination in our code bare.

I'm fine with the Google guideline btw. 

Despite from including the counterpart  header of a cpp file first (for the reasons you have above) there is no real difference wrt advantages/disadvantages, it's just a definition) coding guideline) for the sake consistency.

ksooo: Ha, we have something already... => https://codedocs.xyz/xbmc/xbmc/code_guidelines.html#code_guidelines_1_3

<pre>
Included header files *.h, are to sort alphabetical to prevent double used file definition and allow better overview

    On cpp define used header first to confirm needed headers are present on them itself
    As next use the global system headers e.g. <cassert> or "system.h"
    Then insert all the headers needed for the file itself
</pre>

Reads like the Google guideline (though not formulated clear enough).

MartijnKaijser: https://github.com/xbmc/xbmc/pull/11043 that ok?

PR #10873, opened 2016-11-06
DVDOverlaySSA: Fix Leak with SSA subtitles

popcornmix:
## Description
CDVDSubtitlesLibass is reference counted and only releases libass
reasources (typically hundreds of MBs) when count gets to zero.

But when replacing a subtitle it is acquiring it a second time
and this is never released.

## Motivation and Context
Fixes: http://trac.kodi.tv/ticket/15820

## How Has This Been Tested?
Tested with subtitle file from trac ticket.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDSubtitles/DVDSubtitleParserSSA.cpp
+++ b/xbmc/cores/VideoPlayer/DVDSubtitles/DVDSubtitleParserSSA.cpp
@@ -54,7 +54,6 @@ bool CDVDSubtitleParserSSA::Open(CDVDStreamInfo &hints)
     if (curEvent)
     {
       CDVDOverlaySSA* overlay = new CDVDOverlaySSA(m_libass);
-      overlay->Acquire(); // increase ref count with one so that we can hold a handle to this overlay
 
       overlay->iPTSStartTime = (double)curEvent->Start * (DVD_TIME_BASE / 1000);
       overlay->iPTSStopTime  = (double)(curEvent->Start + curEvent->Duration) * (DVD_TIME_BASE / 1000);


popcornmix:
Ping @FernetMenta

xbmc/cores/VideoPlayer/DVDCodecs/Overlay/DVDOverlaySSA.h
```diff
@@ -34,7 +34,6 @@ class CDVDOverlaySSA : public CDVDOverlay
   {
     replace = true;
     m_libass = libass;
-    libass->Acquire();
```
FernetMenta: where is the acquire done?

popcornmix: We are passing in an already acquired libass pointer.

popcornmix: Removing this results in the refcount reaching zero at shutdown.
Either we need to avoid the Acquire here, or add another Release to match.

FernetMenta: we release in the destructor. means with every instance of CDVDOverlaySSA there must be also an acquire. where is this in the code when not here?

popcornmix: @FernetMenta the object created here: https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDSubtitles/DVDSubtitleParserSSA.cpp#L56

never gets deleted. This results in the refcount of libass not reaching zero, and hence the libass memory not being freed.

popcornmix: I'm suspicious of:
https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDSubtitles/DVDSubtitleParser.h#L53
which seems to create a new CDVDOverlay and lose track of the original.
Returning ``o`` rather than ``o->Clone()`` also seems to fix the leak.
Do you know why we are cloning the object?

FernetMenta: I have not written this code so I can only guess. The subtitle parser owns the collection of line items. For rendering particular line items get cloned (for what reason ever). Looks kind of strange but I don't see an error there. In regard to libass cloning or increasing ref count directly is the same thing.

> never gets deleted.

this seems wrong. Items should be deleted with disposal of the collection.

FernetMenta: @popcornmix 
https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDSubtitles/DVDSubtitleParserSSA.cpp#L57

this looks wrong. the instance is already create with a ref count of 1

popcornmix: Agree that Acquire does look wrong and removing it also fixes the libass not being freed.

popcornmix:
Updated. jenkins build this please

popcornmix:
I've confirmed that with this PR libass releases all allocated memory.
It was included in last night's Milhouse build.
Okay to merge?

FernetMenta:
+1

PR #10875, opened 2016-11-06
[cmake] Win-Binary-Addon fixes + Handle "select" as setting type for binary addons

fetzerch:
A couple of changes from the retroplayer branch.

## Description
- Two changes related to binary addons. Ping @hudokkow, @wsnipex or @Paxxi.
- One change for an addon callback. Not sure whom to ping here.

## Motivation and Context
Upstreaming them, to reduce rebase work for @garbear.

## How Has This Been Tested?
Locally and on jenkins.

## Types of change

## Checklist:

---
Jenkins build (retroplayer branch): http://jenkins.kodi.tv/job/WIN-32/10573/

### DIFF
--- a/project/cmake/scripts/common/HandleDepends.cmake
+++ b/project/cmake/scripts/common/HandleDepends.cmake
@@ -161,6 +161,10 @@ function(add_addon_depends addon searchpath)
                                   PATCH_COMMAND ${PATCH_COMMAND}
                                   "${INSTALL_COMMAND}")
 
+        if(CMAKE_VERSION VERSION_GREATER 3.5)
+          list(APPEND EXTERNALPROJECT_SETUP GIT_SHALLOW 1)
+        endif()
+
         # if there's an url defined we need to pass that to externalproject_add()
         if(DEFINED url AND NOT "${url}" STREQUAL "")
           # check if there's a third parameter in the file
--- a/project/cmake/scripts/common/HandleDepends.cmake
+++ b/project/cmake/scripts/common/HandleDepends.cmake
@@ -172,6 +172,20 @@ function(add_addon_depends addon searchpath)
                                 GIT_REPOSITORY ${url}
                                 GIT_TAG ${revision}
                                 "${EXTERNALPROJECT_SETUP}")
+
+            # For patchfiles to work, disable (users globally set) autocrlf=true
+            if(CMAKE_MINIMUM_REQUIRED_VERSION VERSION_GREATER 3.7)
+              message(AUTHOR_WARNING "Make use of GIT_CONFIG")
+            endif()
+            if(WIN32 AND patches)
+              externalproject_add_step(${id} gitconfig
+                                       COMMAND git config core.autocrlf false
+                                       COMMAND git rm -rf --cached .
+                                       COMMAND git reset --hard HEAD
+                                       COMMENT "Performing gitconfig step: Disabling autocrlf to enable patching for '${id}'"
+                                       DEPENDERS patch
+                                       WORKING_DIRECTORY <SOURCE_DIR>)
+            endif()
           else()
             set(CONFIGURE_COMMAND "")
             if(NOT WIN32)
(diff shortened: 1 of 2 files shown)


Paxxi:
Will this change the setting for the Kodi repo or just some addon?

MartijnKaijser:
it's a setting option that we already have for python addons but it seems binary didn't had it added yet?

fetzerch:
cdac36e is only for binary addon dependencies. Those are downloaded as external projects and usually you don't ever change anything in there. that's also why we need only the last change.

fetzerch:
@MartijnKaijser: Yes exactly. It's mentioned in the wiki http://kodi.wiki/view/Add-on_settings#type.3D.22select.22 but didn't work for binary addons.

hudokkow:
cmake changes look good.

btw, 5+ hours to build win32 with retroplayer addons? ouch!

fetzerch:
Yep because it uses only one core. https://github.com/xbmc/xbmc/pull/10753 should speed it up.

Paxxi:
No, just no, doing ugly tricks in our build system to work around a broken application (patch) to try and patch something from a git repo is just begging for trouble.

Fork the repo, patch it, upstream it or whatever but if we start down this road it will only end in tears and gothic synth pop while eating ice cream.

fetzerch:
What we currently have is just broken and only works accidentally if you don't have autocrlf=true in your global gitconfig. So if we want to keep the possibility of applying patches, this is necessary.

And I don't think anyone wants to maintain local forks of repos for small patches like this: https://github.com/notspiff/inputstream.rtmp/tree/master/depends/common/librtmp
Because you'll have to bump the repo manually.

Paxxi:
We still have to bump the repo manually for new versions so it's no more work.

garbear:
i've forked a few cores and it's more work that adding a .patch file because you have to manually rebase often, whereas the patch is always applied to master

wsnipex:
the option to apply patches is an integral part of the addons build system, we need this to work.
Note that upstream sources are often _not repositories_ but simple tarballs. We definitely not going to put those into repos, just to apply minimal patches that are only needed for build system.

hudokkow:
Right now my github account holds 180+ repos. 180 are Kodi related.
Keeping local forks of every repo out there in need of a patch is crazy.

akva2:
look to peeps who have been doing this for a long time. look at any /random linux distro. they have been doing this for years on end. and they all use orig tarballs and patchery in the packaging.  look no further than your own repo. depends is all tarballs and patchery (not counting the windows stuff which is just meh all over). do you really think it's a good idea to go for a model different from what every single linux distro out there does?

you can use whatever tool you want locally to generate them patches. such as git. i for one often do; import tarball1 in a git, apply patches, import tarball2 (the new tarball), rebase patch branch on that and voila patches. but that should not affect deployment.

fetzerch:
I'll try to upstream a feature for the autocrlf change as well: https://gitlab.kitware.com/cmake/cmake/merge_requests/240

@MartijnKaijser: Want me to add all missing addon config options then?

fetzerch:

fetzerch:
android build error unrelated

ronie:
@fetzerch 
building of binary addons now fails on ubuntu 16.10, which uses cmake 3.5.2
```/bin/sh: 1: GIT_SHALLOW: not found```

afaik GIT_SHALLOW was introduced in cmake 3.6
if so, this check needs to be adjusted i think?:
https://github.com/xbmc/xbmc/pull/10875/commits/f095fe3f6df192552695c22ab0b94516d9d06714#diff-792731e8f7b3b5e722d1697dd3998296R164

File: project/cmake/scripts/common/HandleDepends.cmake [hunk 1]
```diff
@@ -161,6 +161,10 @@ function(add_addon_depends addon searchpath)
                                   PATCH_COMMAND ${PATCH_COMMAND}
                                   "${INSTALL_COMMAND}")
 
+        if(CMAKE_VERSION VERSION_GREATER 3.5)
+          list(APPEND EXTERNALPROJECT_SETUP GIT_SHALLOW 1)
```
Rechi: @fetzerch this conflicts with https://github.com/xbmc/xbmc/blob/a18701e/cmake/addons/depends/README#L17-L18
`<git-revision>` can't be a commit which is not the last commit for a branch

PR #10877, opened 2016-11-06
New default webinterface Chorus2

MartijnKaijser:
Initially it was the plan to add Chorus1 but since @jez500 wanted to do a rewrite that was delayed a bit. 
Current webinterface is hopeless outdated (broken) and we really need an upgrade.

Since we don't support simply pulling a .zip from a server and package it inside Kodi i'll just be swapping them out this way. The addon ID are the same so there is no problems doing this.


ronie:
was about time ;-)

i do smell licensing issues though with the fanart images included in this addon.
^ @jez500

jez500:
Hey @ronie 
The fanart came from a wallpaper site, from memory it was wallbase, none of them had author information otherwise, I would have included attributions.
Maybe I should get the ol' DLSR out and take some original snaps.

ronie:
yeah, it would be preferred to change them if we don't have any licensing info.

addons/webinterface.default/addon.xml
```diff
@@ -1,145 +1,19 @@
...
-    <description lang="uk_UA">Типовий веб-інтерфейс  для Kodi; Розроблений для пристроїв з будь-якою роздільною здатністю</description>
-    <description lang="uz_UZ">Kodi uchun andoza veb interfeysi; barcha ekran o'lchamlari bilan uskunalar uchun yaratilgan</description>
-    <description lang="vi_VN">Giao diện Web mặc định cho Kodi; Thiết kế cho các thiết bị ở tất cả độ phân giải khác nhau</description>
-    <description lang="zh_CN">Kodi 的默认 web 界面；为支持所有分辨率的设备而设计</description>
-    <description lang="zh_TW">Kodi的預設網頁介面；針對所有解析度的設備而設計</description>
+    <summary lang="en">Kodi web interface - Chorus2</summary>
+    <description lang="en">Default Kodi web interface. Browse and interact with your Music, Movies, TV Shows and more via a web browser.</description>
     <platform>all</platform>
+    <language>en</language>
+    <source>https://github.com/jez500/chorus2</source>
```
da-anda: shouldn't it be moved to our github account? Also not sure about having "jez500" as provider on a default add-on instead of Team Kodi. Maybe we should differ between "provider" and "author/developer"? Don't get me wrong, not trying to take credit away from Jeremy, it just feels odd.

MartijnKaijser: It's already on our github account. It's simply not changed yet

jez500:
I just pushed up 2.3.2 with some tweeks

@ronie - Fanart replaced with CC0 images
@da-anda - Updated the source in the addon.xml to point to the xbmc repo

https://github.com/xbmc/chorus2/commit/67ea92f9d742e37100760f707e077c4a7eaf4f18

ronie:
> Also not sure about having "jez500" as provider on a default add-on instead of Team Kodi.

perhaps list both, like we did with Confluence?
https://github.com/xbmc/xbmc/blob/Jarvis/addons/skin.confluence/addon.xml#L3-L6

MartijnKaijser:
We should also add some screenshots like defined in
http://forum.kodi.tv/showthread.php?tid=250936&pid=2422343#pid2422343

jez500:
@ronie & @da-anda - Added Team Kodi as provider
@MartijnKaijser - Screenshots added

https://github.com/xbmc/chorus2/commit/2266953d70bfb6017d9f37210e81aecc9f7fc587

MartijnKaijser:
Changed my mind

PR #10878, opened 2016-11-06
[gui] change icons for playlist section

phil65:
Changes this:
screenshot011
to this:
screenshot010

@ronie @BigNoid 	@HitcherUK


ronie:
looks good to me

BigNoid:
I'd like to replace ```DefaultPlaylist.png``` with ```DefaultPlaylistMusic.png``` and add ```DefaultPlaylistVideo.png```. 
Up till now ```DefaultPlaylist.png``` was only used in the music library and skins have reflected that by putting a music symbol on that icon.
```DefaultMusicPlaylists.png``` and ```DefaultVideoPlaylists.png``` might also be used, but since those are for nodes, skins may have put folder icons on those.

phil65:
In general I think our direction should be to merge instead of to separate, because we have way too many super-specific defaulticons now already.
Example: We have 3 different icons for "Search" I think, just havin one called "DefaultSearch" makes much more sense, because it´s always clear from context whether it it is add-on related, pvr-related or whatever. By using a more generic naming, add-on authors can also make more use of these icons.

Another reason could be mixed playlists, that would need another icon in case we start to separate them.

HitcherUK:
+1

BigNoid:
I don't think there can be too many icons, although that specific one about search makes sense.
But I realise that may be personal preference so won't mind keeping the one for both audio and video.

phil65:
we´re well above 100 icons already.... My plan would be to get rid of a whole bunch for v18.

da-anda:
how about default icons combined with type specific overlays for skins that want to do that? So the playlist icon would be the same but skins could add an overlay to f.e. the top right corner indicating a specific media type. These overlay icons could again also be used in other places for consistancy and would reduce the number of icons that have to be created by skinners while still giving visual distinction.

phil65:
> edit: being a visual guy, I'd f.e. prefer the "party mode" playlist to have a 🎉 like overlay to spot it more easily in the view of your screenshot above. You find a certain icon more easily than reading each and every label.

I agree that it would be nicer, intention of this pr though was to do a quick improvement without any breakage.

HitcherUK:
Can't we just make sure the fallback works correctly and then we can have as many or as few icons as we want?

eg `DefaultPlaylistMusic.png`, `DefaultPlaylistVideo.png`, `DefaultPlaylistMixed.png` and if none are included then fallback to `DefaultPlaylist.png`

phil65:
Sure, technically that´s possible, but then I am the wrong person to do it ;) The only thing I would tackle is to simplify things, not the opposite.

phil65:
jenkins build this please.

phil65:
updated and also added a separate icon for partymode now.
Credits for the icon go to @IchabodFletchman

phil65:
jenkins build this please.

PR #10881, opened 2016-11-07
Move favorites to the new contextmenu handling

razzeee:
## Description

This will enable having "Add to favourites" and "Remove from favourites" on the homescreen. It also changes the handling inside the library. But it should be the same as before.

One "problem" could be the place I've choosen. As this part is library agnostic (not really video) it might make sense to have a new contextmenu class. But I've choosen the video one for now.

## How Has This Been Tested?

Needs more testing, I have tested adding/deleting. Which works fine, we need to check all corners, if there are places, that have multiple items for "Add to favourites"

## Types of change

## Checklist:

### DIFF
--- a/xbmc/video/ContextMenus.cpp
+++ b/xbmc/video/ContextMenus.cpp
@@ -21,9 +21,14 @@
 #include "ContextMenus.h"
 #include "Application.h"
 #include "Autorun.h"
+#include "ServiceBroker.h"
 #include "Util.h"
 #include "video/dialogs/GUIDialogVideoInfo.h"
 #include "video/windows/GUIWindowVideoBase.h"
+#include "favourites/FavouritesService.h"
+#include "filesystem/FavouritesDirectory.h"
+#include "utils/StringUtils.h"
+#include "utils/URIUtils.h"
 
 
 namespace CONTEXTMENU
@@ -190,5 +195,27 @@ bool CPlay::Execute(const CFileItemPtr& itemIn) const
   SetPathAndPlay(item);
   return true;
 };
+
+
+std::string CFavorite::GetLabel(const CFileItem& item) const
+{
+  return g_localizeStrings.Get(CServiceBroker::GetFavouritesService().IsFavourited(item,
+    g_windowManager.GetActiveWindowOrDialog()) ? 14077 : 14076); // Add/Remove Favourite
+}
+
+bool CFavorite::IsVisible(const CFileItem& item) const
+{
+  //! @todo FAVOURITES Conditions on masterlock and localisation
+  return !item.IsParentFolder() && !item.IsPath("add") && !item.IsPath("newplaylist://") &&
+    !URIUtils::IsProtocol(item.GetPath(), "newsmartplaylist") && !URIUtils::IsProtocol(item.GetPath(), "newtag") &&
+    !URIUtils::IsProtocol(item.GetPath(), "musicsearch") && !URIUtils::IsProtocol(item.GetPath(), "favourites") &&
+    !StringUtils::StartsWith(item.GetPath(), "pvr://guide/") && !StringUtils::StartsWith(item.GetPath(), "pvr://timers/");
+}
+
+bool CFavorite::Execute(const CFileItemPtr& item) const
+{
+  return CServiceBroker::GetFavouritesService().AddOrRemove(*item.get(), g_windowManager.GetActiveWindowOrDialog());
+};
+
 
 }
--- a/xbmc/ContextMenuManager.cpp
+++ b/xbmc/ContextMenuManager.cpp
@@ -83,6 +83,7 @@ void CContextMenuManager::Init()
       std::make_shared<CONTEXTMENU::CMarkWatched>(),
       std::make_shared<CONTEXTMENU::CMarkUnWatched>(),
       std::make_shared<CONTEXTMENU::CRemoveResumePoint>(),
+      std::make_shared<CONTEXTMENU::CFavorite>(),
       std::make_shared<CONTEXTMENU::CEjectDisk>(),
       std::make_shared<CONTEXTMENU::CEjectDrive>(),
       std::make_shared<CONTEXTMENU::CRemoveFavourite>(),
--- a/xbmc/video/ContextMenus.h
+++ b/xbmc/video/ContextMenus.h
@@ -90,4 +90,11 @@ struct CPlay : IContextMenuItem
   bool IsVisible(const CFileItem& item) const override;
   bool Execute(const CFileItemPtr& _item) const override;
 };
+
+struct CFavorite : IContextMenuItem
+{
+  std::string GetLabel(const CFileItem& item) const override;
+  bool IsVisible(const CFileItem& item) const override;
+  bool Execute(const CFileItemPtr& _item) const override;
+};
 }
--- a/xbmc/windows/GUIMediaWindow.cpp
+++ b/xbmc/windows/GUIMediaWindow.cpp
@@ -1720,18 +1720,6 @@ void CGUIMediaWindow::GetContextButtons(int itemNumber, CContextButtons &buttons
   if (!item || item->IsParentFolder())
     return;
 
-  //! @todo FAVOURITES Conditions on masterlock and localisation
-  if (!item->IsParentFolder() && !item->IsPath("add") && !item->IsPath("newplaylist://") &&
-      !URIUtils::IsProtocol(item->GetPath(), "newsmartplaylist") && !URIUtils::IsProtocol(item->GetPath(), "newtag") &&
-      !URIUtils::IsProtocol(item->GetPath(), "musicsearch") &&
-      !StringUtils::StartsWith(item->GetPath(), "pvr://guide/") && !StringUtils::StartsWith(item->GetPath(), "pvr://timers/"))
-  {
-    if (CServiceBroker::GetFavouritesService().IsFavourited(*item.get(), GetID()))
-      buttons.Add(CONTEXT_BUTTON_ADD_FAVOURITE, 14077);     // Remove Favourite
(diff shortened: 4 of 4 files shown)


MartijnKaijser:
jenkins build this please

Razzeee:
jenkins build this please

Razzeee:
jenkins build this please

da-anda:
time to merge?

Razzeee:
jenkins build and merge

Razzeee:
jenkins build this please

phil65:
needs squashing first.

Razzeee:
Yes, just waiting for jenkins and squashing with github then.

Razzeee:
jenkins build this please

Razzeee:
jenkins build this please

Razzeee:
no clue why droid fails

Rechi:
@Razzeee Jenkis has build an old version of your PR. Maybe you have triggered the build too quickly after pushing.
jenkins build this please

xbmc/video/ContextMenus.cpp
```diff
@@ -168,4 +171,26 @@ bool CPlay::Execute(const CFileItemPtr& itemIn) const
   return true;
 };
 
+
+std::string CFavorite::GetLabel(const CFileItem& item) const
+{
+  //The 0 we're passing for contextWindow should only matter to folders
+  return g_localizeStrings.Get(XFILE::CFavouritesDirectory::IsFavourite(std::make_shared<CFileItem>(item).get(), 0) ? 14077 : 14076); // Add/Remove Favourite
```
tamland: Hold a second. Won't this re-read the favourites from disk? If so, not a good idea as it will block gui every time the context menu is opened.

Razzeee: Possible, it should be similar to the previous implementation. But this function should only be called, whenever Visible returns true. Not sure how else to do that.

Razzeee: @tamland so anything I can do about it or can we merge it as is, as it basically should be the same behaviour as before

tamland: Well it's extended to director providers too now, and not just media windows. We really should fix favourites to keep things in memory instead of deserialize/serialize on every change.

xbmc/video/ContextMenus.cpp
```diff
@@ -168,4 +171,26 @@ bool CPlay::Execute(const CFileItemPtr& itemIn) const
...
+  //! @todo FAVOURITES Conditions on masterlock and localisation
+  return !item.IsParentFolder() && !item.IsPath("add") && !item.IsPath("newplaylist://") &&
+    !URIUtils::IsProtocol(item.GetPath(), "newsmartplaylist") && !URIUtils::IsProtocol(item.GetPath(), "newtag") &&
+    !URIUtils::IsProtocol(item.GetPath(), "musicsearch") &&
+    !StringUtils::StartsWith(item.GetPath(), "pvr://guide/") && !StringUtils::StartsWith(item.GetPath(), "pvr://timers/");
+}
+
+bool CFavorite::Execute(const CFileItemPtr& item) const
+{
+  return XFILE::CFavouritesDirectory::AddOrRemove(item.get(), 0);
```
tamland: Btw, does this work? Usually a window is passed as seconds argument (for some reason). If not, the whole contextWindow thing can just be removed.

xbmc/video/ContextMenus.cpp
Razzeee: Can I do that in a followup to this?

tamland: I guess.. I started a refactor here: https://github.com/tamland/xbmc/commits/favourites_refactor btw, but haven't had  time to complete it

ksooo: Intstead of 0, I would pass </code>g_WindowManager.GetActiveWindowID()</code> here.

xbmc/video/ContextMenus.cpp
ksooo: Intstead of 0, I would pass g_WindowManager.GetActiveWindowID() here.

xbmc/video/ContextMenus.cpp
ksooo: XFILE::CFavouritesDirectory::IsFavourite also returns true for a favourite iself. After #11929 got merged, this will lead now to an item "Remove from favourites" and an additional item "Remove" for every favourite. 

This can be corrected by changing the condition here to 

<code>!URIUtils.IsProtocol(item.GetPath(), "favourites") && XFILE::CFavouritesDirectory::IsFavourite(std::make_shared<CFileItem>(item).get()</code>

ksooo: Or even better, you could add <code>!URIUtils.IsProtocol(item.GetPath(), "favourites")</code> to <code>CFavorite::IsVisible</code> instead

xbmc/video/ContextMenus.cpp
Razzeee: Doesn't seem to be in scope and I fail to get it into scope. Somehow.

ksooo: > Doesn't seem to be in scope and I fail to get it into scope. Somehow.

-- return XFILE::CFavouritesDirectory::AddOrRemove(item.get(), 0);
++ return XFILE::CFavouritesDirectory::AddOrRemove(item.get(), g_windowManager.GetActiveWindowID());

Razzeee: fooled by the upercase `W` I feel stupid

Razzeee:
jenkins build this please

xbmc/video/ContextMenus.cpp
tamland: Hold please, we should definitely *not* do this. This can be executed from anywhere, and you can't assume one of those windows are present. On homescreen this will not give you the video window.

Actually, the entire point of the context menus redesign in the first place, was to get rid of exactly this kind of global states and window dependencies that caused them all to break when used from elsewhere.

Razzeee:
This should be ready then, I'm going to merge this tomorrow.

ksooo:
+1

tamland:
I guess you missed https://github.com/xbmc/xbmc/pull/10881#discussion_r109842173? I don't know why ksooo asked for this...

Razzeee:
@tamland 
Nope, should be here https://github.com/xbmc/xbmc/pull/10881/files#diff-ac52b9bdb9f873addec552b2df3ac874R186

ksooo:
>> Instead of 0, I would pass g_WindowManager.GetActiveWindowID() here.

> Hold please, we should definitely not do this. This can be executed from anywhere, and you can't assume one of those windows are present. On homescreen this will not give you the video window.

@tamland I missed your comment, sorry. Looking at the implementation of CFavouritesDirectory::GetExecutePath, where this id is used, passing g_WindowManager.GetActiveWindowID() actually does not work. Instead, in case the favourite points to a folder, the id of the window to activate (to display the folder contents)  is needed here. Means, passing zero will not work either.

tamland:
It *should* be possible to get this info from the file item. E.g if click on a tv show, it will have a media type set, and the target window would be 'videos'. No need to get the actual window.

I'm not entirely sure what the failing cases are, but if there are any, throwing it out (don't allow favourite arbitrary folders) or attach something (sensible) to the items, would be two possible solutions...

ksooo:
> I'm not entirely sure what the failing cases are, but if there are any, throwing it out (don't allow favourite arbitrary folders) or attach something (sensible) to the items, would be two possible solutions...

 I'd go for option two, because I think being able to add folders as favourites (e.g. a folder containg a complete eason of a tv show) is a neat and valuable feature (which I'm using personally, btw :-).

Razzeee:
I'm not sure what you want me to do.

ksooo:
> I'm not sure what you want me to do.

 We need to sort out how to deal properly with the second parameter of XFILE::CFavouritesDirectory::AddOrRemove

Razzeee:
That part I did understand :D

ksooo:
So, what's unclear? I think the message is: Don't merge this PR until we found a solution for the id problem.

Razzeee:
I've got that, just finding the solution is the thing (as always)

tamland:
@Razzeee I would start by replacing this else block: https://github.com/xbmc/xbmc/blob/d3fc85626de56f097c03fc431a2316e5a4bdd148/xbmc/filesystem/FavouritesDirectory.cpp#L210-L215

Instead of inserting `contextWindow` check item properties such as `item.HasVideoInfoTag()` (maybe also `m_type`?), `HasMusicInfoTag` etc.  Video things should have execute action `ActivateWindow(Videos, ...`, music things `Music`. This should at least take of those, and addon folders (which will have `addons://` path). Rest I don't know on top of my head.

ksooo:
>> Instead of inserting contextWindow check item properties such as item.HasVideoInfoTag() (maybe also m_type?), HasMusicInfoTag etc.

Do folder items contain videoinfotags, musicinfotags and the like?

tamland:
Probably not plain folders, but *library folders* like tv shows, seasons, albums etc. should.

da-anda:
beware that for things like the videoInfoDialog the fileItems will be altered and won't contain the library path anymore (stumbled accross this when I tried to refactor this part and have one "manage" button trigger a context menu instead of duplicating context menu buttons in the dialog, which is still a goal I have). While I agree that cFileItems should be selfaware and be able to tell the required context, the required changes might have a bigger scope than anticipated (as always).

Rechi:
@Razzeee this needs a rebase

white-gecko:
Will anything happen to this PR? Might this fix the issue as pointed out here: https://forum.kodi.tv/showthread.php?tid=304000 ?

PR #10882, opened 2016-11-07
Ask user to configure new controllers (once)

garbear:
If a button is pressed and the controller isn't mapped, this shows a dialog asking if the user wants to go to the controller dialog.

If the user selects yes, the controller dialog is opened. If the user selects no, nothing happens and the prompt is never shown again.

## Motivation and Context
Better UX

new controller detected

## Types of change

Broken out from #10630

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -16016,7 +16016,19 @@ msgctxt "#35010"
 msgid "Peripheral libraries"
 msgstr ""
 
-#empty strings from id 35011 to 35048
+#. Title of dialog to ask users if they would like to configure a new controller
+#: xbmc/input/joysticks/dialogs/GUIDialogNewJoystick.cpp
+msgctxt "#35011"
+msgid "New controller detected"
+msgstr ""
+
+#. Text of dialog to ask users if they would like to configure a new controller
+#: xbmc/input/joysticks/dialogs/GUIDialogNewJoystick.cpp
+msgctxt "#35012"
+msgid "A new controller has been detected. Configuration can be done at any time in \"Settings -> System Settings -> Input\". Would you like to configure it now?"
+msgstr ""
+
+#empty strings from id 35013 to 35048
 
 #. Name of game add-ons category
 #: xbmc/filesystem/AddonsDirectory.cpp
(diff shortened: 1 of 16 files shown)


addons/resource.language.en_gb/resources/strings.po
```diff
@@ -16016,7 +16016,19 @@ msgctxt "#35010"
...
+#. Title of dialog to ask users if they would like to configure a new controller
+#: xbmc/input/joysticks/dialogs/GUIDialogNewJoystick.cpp
+msgctxt "#35011"
+msgid "New controller detected"
+msgstr ""
+
+#. Text of dialog to ask users if they would like to configure a new controller
+#: xbmc/input/joysticks/dialogs/GUIDialogNewJoystick.cpp
+msgctxt "#35012"
+msgid "Would you like to configure your controller? This can be done later in System Settings."
```
da-anda: I'm not a native speaker, but for me the following wording would be more clear. Also, IMO the question the user has to base the decision on should ideally be the last part of the message in a yes/no dialog so that the button labels are the immediate answer to the question.

`A new controller has been detected and can be configured. Configuration can be done at any time by navigating to "Settings -> System Settings -> Input". Would you like to configure it now?` [YES] [NO]

Likely needs improvements as I'm still half asleep, but you get the idea.

garbear: yea, good point about the question being at the end. I'd like to try for a little more brevity, so how about

`A new controller has been detected. Configuration can be done at any time in "Settings -> System Settings -> Input". Would you like to configure it now?`

garbear:
@FernetMenta or @Paxxi for an architecture review. I managed to avoid a global variable, is a static const one ok here?

xbmc/input/joysticks/dialogs/GUIDialogNewJoystick.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  if (bShow)
+    Create();
+}
+
+void CGUIDialogNewJoystick::Process()
+{
+  // "New controller detected"
+  // "Would you like to configure your controller? This can also be done in System Settings."
+  if (CGUIDialogYesNo::ShowAndGetInput(CVariant{ 35011 }, CVariant{ 35012 }))
```
da-anda: mind using the new suggested approach for the YesNo dialog instead? See https://github.com/xbmc/team-kanban-board/issues/5

garbear:
jenkins build this please

PR #10884, opened 2016-11-07
Display milliseconds in log file

peak3d:
Display milliseconds in log file

## Description
Display milliseconds in log file

## Motivation and Context
Analysing PTS values and timing is nearly impossible without milliseconds

## How Has This Been Tested?
Runtime test

## Types of change

## Checklist:

### DIFF
--- a/xbmc/utils/log.cpp
+++ b/xbmc/utils/log.cpp
@@ -93,7 +93,7 @@ void CLog::LogString(int logLevel, const std::string& logString)
       WriteLogString(s_globals.m_repeatLogLevel, strData2);
       s_globals.m_repeatCount = 0;
     }
-    
+
     s_globals.m_repeatLine = strData;
     s_globals.m_repeatLogLevel = logLevel;
 
@@ -198,19 +198,21 @@ void CLog::PrintDebugString(const std::string& line)
 
 bool CLog::WriteLogString(int logLevel, const std::string& logString)
 {
-  static const char* prefixFormat = "%02.2d:%02.2d:%02.2d T:%" PRIu64" %7s: ";
+  static const char* prefixFormat = "%02d:%02d:%02d.%03d T:%" PRIu64" %7s: ";
 
   std::string strData(logString);
   /* fixup newline alignment, number of spaces should equal prefix length */
   StringUtils::Replace(strData, "\n", "\n                                            ");
 
   int hour, minute, second;
-  s_globals.m_platform.GetCurrentLocalTime(hour, minute, second);
-  
+  double millisecond;
+  s_globals.m_platform.GetCurrentLocalTime(hour, minute, second, millisecond);
+
   strData = StringUtils::Format(prefixFormat,
                                   hour,
                                   minute,
                                   second,
+                                  static_cast<int>(millisecond),
                                   (uint64_t)CThread::GetCurrentThreadId(),
                                   levelNames[logLevel]) + strData;
 
--- a/xbmc/utils/posix/PosixInterfaceForCLog.cpp
+++ b/xbmc/utils/posix/PosixInterfaceForCLog.cpp
@@ -21,6 +21,7 @@
 #include "PosixInterfaceForCLog.h"
 #include <stdio.h>
 #include <time.h>
+#include <sys/time.h>
 
 #if defined(TARGET_DARWIN)
 #include "platform/darwin/DarwinUtils.h"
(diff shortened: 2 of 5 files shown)


koying:
+1 on principle. Droid logcat is in milli as well

popcornmix:
Personally I'd prefer microseconds. I've been using this in Milhouse builds for years:
https://github.com/popcornmix/xbmc/commit/49013505dd0c71a8f0d318fb13402ccfde5f59f6

fritsch:
As you use the msec value anyways, use it directly, saves us div

peak3d:
@popcornmix can you pls post an example line?

popcornmix:
```20:23:06  46.670376 T:1995576096  NOTICE: Starting Kodi (17.0-BETA6 Git:2016-11-06-a287ef9-dirty). Platform: Linux ARM 32-bit```

The second number is seconds/microseconds elapsed. I don't mind it being merged into the HH:MM:SS part. I agree more time precision is useful in some cases. Sometimes even more than millisecond resolution is useful for me (for approximate profiling when log messages are frequent, e.g. with video component specific logging on mmal).

**Comment by peak3d** [comment]
ok, thx @popcornmix I would format it to %03.6 then to have it in one column, ok?

20:23:06 046.670376

**Comment by popcornmix** [comment]
What you've written looks just like mine with a leading zero.
Do you mean ``20:23:06.670376`` ?

One issue with my logging is it uses a wall clock for first column and an elapsed time for second colums, so you have (different) seconds reported twice. It might be better to avoid that.

**Comment by peak3d** [comment]
Last commit is: 20:23:06 046.670376, but 046 are milliseconds, not elapsed.

So it should it be better: 20:23:06:046.670376 inside the HH:MM:SS, correct?

I'll wait now for the vote: either elapsed in seconds like @popcornmix' approach, or part of HH::MM::SS::MS.mS

**Comment by peak3d** [comment]
approach a) Milliseconds.Micorseconds as part of HH:MM:SS

**Comment by peak3d** [comment]
approach b) elapsed seconds with fractional part since kodi start.

**Comment by popcornmix** [comment]
I'm happy with (a).

**Comment by koying** [comment]
(a) Indeed, let's not mix pears and apples

not sure about Milliseconds.Micorseconds, though
Why not HH:MM:SS.Microseconds? That'd be iso...

**Comment by peak3d** [comment]
Its HH:MM:SS:MS now where MS is Milliseconds (0-999).
To log.h milliseconds are still passed as double, so if anyone needs microseconds he can simply change the formatting.

**Comment by peak3d** [comment]
jenkins build this please

**Comment by koying** [comment]
Please stick to standards, if possible ;)
It's HH:MM:SS[**.**sss...]

Both rfc3339 and iso 8601 have it that way. It's a fraction of a second, with limitless precision.

**Comment by peak3d** [comment]
ok

**Comment by peak3d** [comment]
HH:MM:SS.SSS now

**Comment by ksooo** [comment]
Is there today really the need to implement this platfom dependend? Wouldn't std::chrono fit nicely here? => http://www.cplusplus.com/reference/chrono/

File: xbmc/utils/log.cpp [hunk 1]
```diff
@@ -198,19 +198,21 @@ void CLog::PrintDebugString(const std::string& line)
...
-  s_globals.m_platform.GetCurrentLocalTime(hour, minute, second);
-  
+  double millisecond;
+  s_globals.m_platform.GetCurrentLocalTime(hour, minute, second, millisecond);
+
   strData = StringUtils::Format(prefixFormat,
                                   hour,
                                   minute,
                                   second,
+                                  (int)millisecond,
```
ksooo: please use c++ style casts

xbmc/utils/posix/PosixInterfaceForCLog.cpp
```diff
@@ -94,16 +95,18 @@ void CPosixInterfaceForCLog::PrintDebugString(const std::string &debugString)
...
   struct tm localTime;
-  if (time(&curTime) != -1 && localtime_r(&curTime, &localTime) != NULL)
+  struct timeval tv;
+
+  if (gettimeofday(&tv, nullptr) != -1 && localtime_r(&tv.tv_sec, &localTime) != NULL)
   {
     hour   = localTime.tm_hour;
     minute = localTime.tm_min;
     second = localTime.tm_sec;
+    milliseconds = (double)tv.tv_usec / 1000;
```
ksooo: c++ style cast, please (and maybe at all other places you changed). thx

peak3d:
c++ style OK, std::chrono postponed and goes into the kanban board because of too many other things to do, @ksooo

ksooo:
> std::chrono postponed and goes into the kanban board because of too many other things to do

fine by me.

peak3d:
jenkins build this please

peak3d:
jenkins build

peak3d:
jenkins build this

peak3d:
jenkins build this please

peak3d:
jenkins build this please

ksooo:
being devil's advocate again: do we really need this kind of stuff in Krypton? Every code change can introduce regressions and tbh, this one is no more than "sugar", right?

koying:
Trust us, it's not sugar when you try to make sense of a debug log where a/v is involved ;)

ksooo:
> Trust us, it's not sugar when you try to make sense of a debug log where a/v is involved ;)

k

PR #10885, opened 2016-11-07
[subtitles] decode html escape characters (reverted from commit 56ff2…

MartijnKaijser:
Partly revert PR10813 as each word got reverted for RTL languages like Hebrew


MartijnKaijser:
Jenkins build this please

PR #10889, opened 2016-11-07
More code modernization for dialogVideoInfo

razzeee:
As suggested by @ksooo 
Big thanks!

### DIFF
--- a/xbmc/video/dialogs/GUIDialogVideoInfo.cpp
+++ b/xbmc/video/dialogs/GUIDialogVideoInfo.cpp
@@ -199,7 +199,7 @@ bool CGUIDialogVideoInfo::OnMessage(CGUIMessage& message)
               pDlgSelect->Add(director);
             pDlgSelect->Open();
 
-          int iItem = pDlgSelect->GetSelectedItem();
+            int iItem = pDlgSelect->GetSelectedItem();
             if (iItem < 0)
               return true;
             OnSearch(directors[iItem]);
@@ -497,11 +497,7 @@ void CGUIDialogVideoInfo::OnSearch(std::string& strSearch)
       if (iItem < 0)
         return;
 
-      CFileItem* pSelItem = new CFileItem(*items[iItem]);
-
-      OnSearchItemFound(pSelItem);
-
-      delete pSelItem;
+      OnSearchItemFound(items[iItem].get());
     }
   }
   else
--- a/xbmc/video/dialogs/GUIDialogVideoInfo.cpp
+++ b/xbmc/video/dialogs/GUIDialogVideoInfo.cpp
@@ -191,16 +191,19 @@ bool CGUIDialogVideoInfo::OnMessage(CGUIMessage& message)
         else
         {
           auto pDlgSelect = g_windowManager.GetWindow<CGUIDialogSelect>();
-          pDlgSelect->Reset();
-          pDlgSelect->SetHeading(CVariant{22080});
-          for (const auto &director: directors)
-            pDlgSelect->Add(director);
-          pDlgSelect->Open();
+          if (pDlgSelect)
+          {
+            pDlgSelect->Reset();
+            pDlgSelect->SetHeading(CVariant{22080});
+            for (const auto &director: directors)
+              pDlgSelect->Add(director);
+            pDlgSelect->Open();
 
           int iItem = pDlgSelect->GetSelectedItem();
-          if (iItem < 0)
-            return true;
-          OnSearch(directors[iItem]);
+            if (iItem < 0)
+              return true;
+            OnSearch(directors[iItem]);
+          }
         }
       }
       else if (iControl == CONTROL_LIST)
@@ -358,7 +361,7 @@ void CGUIDialogVideoInfo::SetMovie(const CFileItem *item)
       {
         m_movieItem->GetVideoInfoTag()->m_strTrailer = localTrailer;
         CVideoDatabase database;
-        if(database.Open())
+        if (database.Open())
         {
           database.SetSingleValue(VIDEODB_CONTENT_MOVIES, VIDEODB_ID_TRAILER,
                                   m_movieItem->GetVideoInfoTag()->m_iDbId,
@@ -471,32 +474,35 @@ void CGUIDialogVideoInfo::OnSearch(std::string& strSearch)
   if (items.Size())
   {
     CGUIDialogSelect* pDlgSelect = g_windowManager.GetWindow<CGUIDialogSelect>();
-    pDlgSelect->Reset();
-    pDlgSelect->SetHeading(CVariant{283});
-
-    CVideoThumbLoader loader;
-    for (int i = 0; i < items.Size(); i++)
+    if (pDlgSelect)
     {
-      if (items[i]->HasVideoInfoTag() &&
+      pDlgSelect->Reset();
+      pDlgSelect->SetHeading(CVariant{283});
+
+      CVideoThumbLoader loader;
+      for (int i = 0; i < items.Size(); i++)
+      {
+        if (items[i]->HasVideoInfoTag() &&
           items[i]->GetVideoInfoTag()->GetPlayCount() > 0)
-        items[i]->SetLabel2(g_localizeStrings.Get(16102));
+          items[i]->SetLabel2(g_localizeStrings.Get(16102));
 
-      loader.LoadItem(items[i].get());
-      pDlgSelect->Add(*items[i]);
+        loader.LoadItem(items[i].get());
+        pDlgSelect->Add(*items[i]);
+      }
+
+      pDlgSelect->SetUseDetails(true);
+      pDlgSelect->Open();
+
+      int iItem = pDlgSelect->GetSelectedItem();
+      if (iItem < 0)
+        return;
+
+      CFileItem* pSelItem = new CFileItem(*items[iItem]);
+
+      OnSearchItemFound(pSelItem);
+
+      delete pSelItem;
     }
-
-    pDlgSelect->SetUseDetails(true);
-    pDlgSelect->Open();
-
-    int iItem = pDlgSelect->GetSelectedItem();
-    if (iItem < 0)
-      return;
-
-    CFileItem* pSelItem = new CFileItem(*items[iItem]);
-
-    OnSearchItemFound(pSelItem);
-
-    delete pSelItem;
   }
   else
   {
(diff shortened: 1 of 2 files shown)


xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -189,17 +189,20 @@ bool CGUIDialogVideoInfo::OnMessage(CGUIMessage& message)
...
-          for (const auto &director: directors)
-            pDlgSelect->Add(director);
-          pDlgSelect->Open();
-
-          int iItem = pDlgSelect->GetSelectedItem();
-          if (iItem < 0)
-            return true;
-          OnSearch(directors[iItem]);
+          auto pDlgSelect = dynamic_cast<CGUIDialogSelect*>(g_windowManager.GetWindow(WINDOW_DIALOG_SELECT));
+          if(pDlgSelect)
```
ksooo: please put a space between if and (

EDIT: and at all other places where you copied this line. ;-)

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -189,17 +189,20 @@ bool CGUIDialogVideoInfo::OnMessage(CGUIMessage& message)
...
+            pDlgSelect->SetHeading(CVariant{22080});
+            for (const auto &director: directors)
+              pDlgSelect->Add(director);
+            pDlgSelect->Open();
+
+            auto iItem = pDlgSelect->GetSelectedItem();
+            if (iItem < 0)
+              return true;
+            OnSearch(directors[iItem]);
+          }
```
ksooo: i would insert a log error in case dynamic_cast fails (here and in all other places where dynamic_cast fails).

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -469,33 +472,36 @@ void CGUIDialogVideoInfo::OnSearch(std::string& strSearch)
...
+      if (iItem < 0)
+        return;
 
-    OnSearchItemFound(pSelItem);
+      CFileItem* pSelItem = new CFileItem(*items[iItem]);
 
-    delete pSelItem;
+      OnSearchItemFound(pSelItem);
+
+      delete pSelItem;
```
ksooo: Why not using a smart pointer here (CFileItemPtr). Saves you the manual delete. maybe this can be improved at other places in this file (not checked)?

ksooo (changes requested):

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
Razzeee: Any conventions for such a log entry?

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
Razzeee: Still struggling with casting CFileItemPtr to CFileItem. Which I would need for OnSearchItemFound.

ksooo: if you need a raw CFileItem* use smartitem.get() 

If you need a CFileItem use *smartitem

ksooo: I guess you need the first.

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -496,11 +496,9 @@ void CGUIDialogVideoInfo::OnSearch(std::string& strSearch)
       if (iItem < 0)
         return;
 
-      CFileItem* pSelItem = new CFileItem(*items[iItem]);
+      CFileItemPtr pSelItem(new CFileItem(*items[iItem]));
 
-      OnSearchItemFound(pSelItem);
-
-      delete pSelItem;
+      OnSearchItemFound(pSelItem.get());
```
garbear: Here, you create a new CFileItem on the heap with `new`. You then pass it to OnSearchItemFound() and destroy the CFileItem after (when the smart pointer falls out of scope).

However, even though OnSearchItemFound() takes a pointer, it doesn't care if the CFileItem is on the heap or the stack. Creating the CFileItem on the stack is faster because LIFO is O(1). You you could do this:

```c++
CFileItem selItem(*items[iItem]);
OnSearchItemFound(&selItem);
```

ksooo: +1

Razzeee: Doesn't this miss the smart pointer and needs to be cleaned up after?

ksooo: smart pointer no longer needed, no heap alloc involved anymore here

ksooo: To drive it to the max, <code>OnSearchItemFound(&(items[iItem]));</code> also should work.

ksooo: And now the final solution: <code>OnSearchItemFound(items[iItem].get());</code>. I just realized that items is a list of CFileItemPtr...

Razzeee:
Like the latest commit @akva2 @garbear ?

xbmc/video/dialogs/GUIDialogVideoInfo.h
```diff
@@ -107,11 +107,11 @@ class CGUIDialogVideoInfo :
...
-  bool m_hasUpdatedThumb;
-  bool m_hasUpdatedUserrating;
-  int m_startUserrating;
+  CFileItemList *m_castList = new CFileItemList;
+  bool m_bViewReview = false;
+  bool m_bRefresh = false;
+  bool m_bRefreshAll = true;
+  bool m_hasUpdatedThumb = false;
+  bool m_hasUpdatedUserrating = false;
+  int m_startUserrating = -1;
```
ksooo: If there is only one constructor for the class I see no benefit in initializing the members in the header file. The code gets imo not more readable or more maintainable - there is no difference compared to init in the cpp file.

But there is one drawback: If you change an init value, all cpp files which include that header file must be recompiled if the init code is in the header file. If you have the init code in the cpp file, only the cpp file must be recompiled if you change an init value.

ksooo: Moreover, like it is now, init is spread over cpp and header which is imo not as good as before where all was in one file. I have to look into two files now to understand whole init setup.

da-anda:
ping :) Let's get our open PR list down

xbmc/video/dialogs/GUIDialogVideoInfo.h
Razzeee: So should I remove https://github.com/xbmc/xbmc/pull/10889/commits/9da6022ff80c144992ea95433e90059d6e795091 again?

garbear: Also an argument against the `Don't compare with nullptr` change is that it drops a compile-time type safety check. I usually prefer the added verbosity when it extends type understanding to other parts of the code

Razzeee: @ksooo

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -189,17 +189,20 @@ bool CGUIDialogVideoInfo::OnMessage(CGUIMessage& message)
...
+          auto pDlgSelect = dynamic_cast<CGUIDialogSelect*>(g_windowManager.GetWindow(WINDOW_DIALOG_SELECT));
+          if (pDlgSelect)
+          {
+            pDlgSelect->Reset();
+            pDlgSelect->SetHeading(CVariant{22080});
+            for (const auto &director: directors)
+              pDlgSelect->Add(director);
+            pDlgSelect->Open();
+
+            auto iItem = pDlgSelect->GetSelectedItem();
```
xhaggi: I totally agree to use auto only for complex types not for primitive types.

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -189,17 +189,20 @@ bool CGUIDialogVideoInfo::OnMessage(CGUIMessage& message)
...
-          pDlgSelect->SetHeading(CVariant{22080});
-          for (const auto &director: directors)
-            pDlgSelect->Add(director);
-          pDlgSelect->Open();
-
-          int iItem = pDlgSelect->GetSelectedItem();
-          if (iItem < 0)
-            return true;
-          OnSearch(directors[iItem]);
+          auto pDlgSelect = dynamic_cast<CGUIDialogSelect*>(g_windowManager.GetWindow(WINDOW_DIALOG_SELECT));
```
xhaggi: I've introduced a new generic method `g_windowManager.GetWindow<T>()` which do all the cast magic for us. In addition I've changed all the places where we currently use the old way. You'll need to rebase your PR.

For more info see https://github.com/xbmc/xbmc/pull/11869

ksooo:
@Razzeee at the end there is no right and wrong here. The code definitely works like it is now. If you ask me I would drop https://github.com/xbmc/xbmc/pull/10889/commits/ba202b471139d09e6720cc949077d6d547227cdc for the reasons @garbear gave and I also would drop https://github.com/xbmc/xbmc/pull/10889/commits/9da6022ff80c144992ea95433e90059d6e795091, for the reasons I gave - only one ctor atm and recompile of all c++ files that include the header directly or indirectly on every init value change, but yeah this might be a weak argument.

But as I said, it's mainly a matter of taste. So, it's up to you to decide.

ksooo (approved):

Razzeee:
@ksooo 
thx so much for your feedback! I've dropped both :)

@xhaggi 
That I've edited that commit to this https://github.com/xbmc/xbmc/pull/10889/commits/ff1508840b052a54a8d9e402ad0f42d11a84b15a?w=1 which now feels pretty much useless but, should not hurt.

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -1952,13 +1958,16 @@ bool CGUIDialogVideoInfo::LinkMovieToTvShow(const CFileItemPtr &item, bool bRemo
   int iSelectedLabel = 0;
   if (list.Size() > 1 || (!bRemove && !list.IsEmpty()))
   {
-    list.Sort(SortByLabel, SortOrderAscending, CServiceBroker::GetSettings().GetBool(CSettings::SETTING_FILELISTS_IGNORETHEWHENSORTING) ? SortAttributeIgnoreArticle : SortAttributeNone);
+    list.Sort(SortByLabel, SortOrderAscending, CSettings::GetInstance().GetBool(CSettings::SETTING_FILELISTS_IGNORETHEWHENSORTING) ? SortAttributeIgnoreArticle : SortAttributeNone);
```
ksooo: I'm pretty sure that this will not compile. CSettings::GetInstance() was replaced by CServiceBroker::GetSettings() for Leia.

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -1952,13 +1954,16 @@ bool CGUIDialogVideoInfo::LinkMovieToTvShow(const CFileItemPtr &item, bool bRemo
...
```
ksooo: I'm pretty sure that this will not compile. CSettings::GetInstance() was replaced by CServiceBroker::GetSettings() for Leia.

Razzeee: Whoops, though I just dropped a Namespace that was unneeded. Will revert.

Razzeee: Reverted! Thx!

Razzeee:
jenkins build this please

PR #10890, opened 2016-11-07
Move eject drive and eject disk to new contextmenu system

razzeee:
## Description

Marvel your eyes, as your now able to eject your disks and drives right from your homescreen 🚀 

## Motivation and Context

No ejection from the homescreen ☮️ 

## How Has This Been Tested?

Only tested eject disk, as I have no drive to eject in my reach.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/ContextMenus.cpp
+++ b/xbmc/ContextMenus.cpp
@@ -1,0 +1,53 @@
+/*
+ *      Copyright (C) 2016 Team Kodi
+ *      http://kodi.tv
+ *
+ *  This Program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2, or (at your option)
+ *  any later version.
+ *
+ *  This Program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with XBMC; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "ContextMenus.h"
+#include "guilib/GUIWindowManager.h"
+#include "storage/MediaManager.h"
+
+
+namespace CONTEXTMENU
+{
+
+  bool CEjectDisk::IsVisible(const CFileItem& item) const
+  {
+    return item.IsRemovable() && (item.IsDVD() || item.IsCDDA());
+  }
+
+  bool CEjectDisk::Execute(const CFileItemPtr& item) const
+  {
+#ifdef HAS_DVD_DRIVE
+    g_mediaManager.ToggleTray(g_mediaManager.TranslateDevicePath(item->GetPath())[0]);
+#endif
+    return true;
+  }
+
+  bool CEjectDrive::IsVisible(const CFileItem& item) const
+  {
+    // Must be HDD
+    return item.IsRemovable() && !item.IsDVD() && !item.IsCDDA();
+  }
+
+  bool CEjectDrive::Execute(const CFileItemPtr& item) const
+  {
+    return g_mediaManager.Eject(item->GetPath());
+  }
+
+}
--- a/xbmc/dialogs/GUIDialogContextMenu.cpp
+++ b/xbmc/dialogs/GUIDialogContextMenu.cpp
@@ -24,7 +24,6 @@
 #include "guilib/GUIControlGroupList.h"
 #include "GUIDialogFileBrowser.h"
 #include "GUIUserMessages.h"
-#include "Autorun.h"
 #include "GUIPassword.h"
 #include "Util.h"
 #include "utils/URIUtils.h"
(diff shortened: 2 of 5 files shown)


ksooo:
Not sure whether it is a good idea to introduce the new cpp and header file in the xbmc source root folder.

ksooo:
Otherwise, this is a nice one. 

But who nowadays is still using physical disks?. /me runs ;-)

Razzeee:
I'm just moving it because it's there, I'm not using it either. 
And yes, if you have a better idea then a class in root, let me know. We will have more cases where context menus are not addon/video/music specific

ksooo:
>  if you have a better idea then a class in root, let me know.

 sorry, no idea. maybe it is even okay to have it in the root directory for "global" context menus. @tamland ?

Razzeee:
jenkins build this please

Razzeee:
jenkins build this please

Razzeee:
jenkins build this please

ksooo:
@Razzeee hmm, why the hurry? Imo, this one needs to be reworked. Comments follow...

xbmc/ContextMenus.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "guilib/GUIWindowManager.h"
+#include "storage/MediaManager.h"
+
+
+namespace CONTEXTMENU
+{
+
+  bool CEjectDisk::IsVisible(const CFileItem& item) const
+  {
+    return item.IsRemovable() && (item.IsDVD() || item.IsCDDA());
```
ksooo: Imo here a 
<pre>
#ifdef HAS_DVD_DRIVE
  return item.IsRemovable() && (item.IsDVD() || item.IsCDDA());
#else
  return false;
#endif
</pre>

would be a good choice to 100% prevent this item to appear on builds compiled without dvd drive support.

xbmc/ContextMenus.cpp
```diff
@@ -0,0 +1,53 @@
...
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with XBMC; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "ContextMenus.h"
+#include "guilib/GUIWindowManager.h"
```
ksooo: Is this include really needed?

But I'm missing <code>#include "FileItem.h"</code>

xbmc/dialogs/GUIDialogContextMenu.cpp
```diff
@@ -224,20 +222,6 @@ bool CGUIDialogContextMenu::SourcesMenu(const std::string &strType, const CFileI
 
 void CGUIDialogContextMenu::GetContextButtons(const std::string &type, const CFileItemPtr& item, CContextButtons &buttons)
 {
-  // Add buttons to the ContextMenu that should be visible for both sources and autosourced items
-  if (item && item->IsRemovable())
-  {
-    if (item->IsDVD() || item->IsCDDA())
-    {
-      buttons.Add(CONTEXT_BUTTON_EJECT_DISC, 13391);  // Eject/Load CD/DVD!
```
ksooo: CONTEXT_BUTTON_EJECT_DISC is no longer used now and schould be removed from dialogs/GUIDialogContextMenu.h

xbmc/dialogs/GUIDialogContextMenu.cpp
```diff
@@ -224,20 +222,6 @@ bool CGUIDialogContextMenu::SourcesMenu(const std::string &strType, const CFileI
...
-    }
-    else // Must be HDD
-    {
-      buttons.Add(CONTEXT_BUTTON_EJECT_DRIVE, 13420);  // Eject Removable HDD!
```
ksooo: CONTEXT_BUTTON_EJECT_DRIVE is no longer used now and schould be removed from dialogs/GUIDialogContextMenu.h

ksooo (commented):
Could you please address my comments and open a follow-up PR? Thnaks.

Razzeee:
Nobody had anything to say for 1 and a half month. I don't exactly call this hurry.

Master is open for pre alpha again and I feel like having PRs to be perfect and not giving people the doubt to deliver follow up PRs is also hurting us more then helping. Just my 2 cent :)

xbmc/dialogs/GUIDialogContextMenu.cpp
Razzeee: Good catch

xbmc/dialogs/GUIDialogContextMenu.cpp
Razzeee: Good catch

MartijnKaijser:
@Razzeee http://trac.kodi.tv/ticket/17143

ksooo:
@Razzeee just discovered that this PR introduces another regression: "Eject" is no longer available in the context menu in Kodi's file manager. Before this, PR, for example, "Eject" appeared in the context menu of a mounted USB stick. Reason is that file manager does not use the new context menu system, but CGUIDialogContextMenu, where this PR removes the items in question from. :-/

Razzeee:
So it seems that if "fails" when asking for the `m_listProvider` as it's null for that case.
Here: https://github.com/xbmc/xbmc/blob/c37e3c276434b5e8fb93c240cbc26e9e16502352/xbmc/guilib/GUIBaseContainer.cpp#L794

@tamland might have the insight on how to correctly tackle this

PR #10897, opened 2016-11-08
[cmake] Fix issue with PLATFORM_OVERRIDE (on linux)

fetzerch:
## Description
An issue reported in the forum revieled a subtle issue in the platform overrides. 

The way our CMake build system works, we cannot specify defines on a per file level. Mainly because of the ENABLE_STATIC_LIBS option that is necessary to create nice Xcode/VS projects.
Therefore the PLATFORM_OVERRIDE define in `cores/VideoPlayer/ProcessInfo` overrides the one in `platform`. While this was not causing harm on OSX/iOS/Android where we have platform overrides for both, it causes an issue on Linux where `platform` uses the default implementation.

As agreed with @wsnipex, we remove the option to change ENABLE_STATIC_LIBS for users.

Reported in: http://forum.kodi.tv/showthread.php?tid=295948

## Motivation and Context
Fix.

## How Has This Been Tested?
Compile tested on linux.

## Types of change

## Checklist:


wsnipex:
thanks! looks good

wsnipex:
jenkins build this please

fetzerch:
windows build error seems unrelated

jenkins build this please

Memphiz:
So do we still needto ensure that platform override defines are unique for each use case?

fetzerch:
yes. unless you know that a platform override in one directory goes together with one in another directory. then they can use the same define.

pgnd:
As of today's latest head, I'm still having to 
```
-	-DBUILD_SHARED_LIBS=1 \
+	-DBUILD_SHARED_LIBS=1 -DENABLE_STATIC_LIBS=1 \
```
to get builds to work.
Is that the intention?  My understanding was that that flags was to be rendered ineffective/unneeded.  No?

I *see* the commits above -- but one's failing tests ... perhaps that's a prob?

PR #10898, opened 2016-11-08
Mir windowing system for Kodi

BrandonSchaefer:
Mir windowing system for Kodi.

## Description
Implements a mir windowing system. Supports both OpenGL/ES for rendering. Keyboard/Pointer events but I need to test out touch events. Supports both cmake/autotools. I also need to test out arm/64.

cmake -DENABLE_MIR=ON or for autotools --enable-mir

## Motivation and Context
The motivation for this change is to support a mir windowing system. Allow a Kodi media center to run on mir.

## How Has This Been Tested?
Tested on ubuntu 16.10 using mir 0.24 and 0.25. There is some mir 0.25 code that is commented out but can be re-enabled once 0.25 has been released.

To test you'll need a mir server + kodi built with mir enabled

On ubuntu 16.10:

Getting a mir server running:
apt-get install miral-examples (gives you miral-shell the mir server)
Running on X11 you can simply run: ./miral-shell
You'll see a window pop up saying Mir-on-X which is the mir server running on the X11 platform.

Getting kodi compiled/working.
apt-get install libmirclient-dev
From this branch you can either use autotools (--enable-mir) or cmake (-DENABLE_MIR=ON)
Once configured make annnd wait.

Once you have kodi.bin and the miral-shell (server) running:
./kodi.bin

This will start kodi.bin on the mir server.

From there tested video/pictures/add-ons/skins/music etc. I could have missed somethings to manually test that I was unsure about.

This code change is very low impact, as its turned off by default and the code it self is stuffed into a directory.

I also think I need to write a README.mir 

https://www.youtube.com/watch?v=26A6qK7vNGo

## Types of change

## Checklist:


FernetMenta:
Good work. Looks really nice after a fist glance.

xbmc/windowing/WindowingFactory.h
```diff
@@ -29,6 +29,12 @@
 #elif defined(TARGET_WINDOWS) && defined(HAS_DX)
 #include "windows/WinSystemWin32DX.h"
 
+#elif defined(TARGET_LINUX)   && defined(HAVE_MIR)   && defined(HAS_GL)
+#include "mir/WinSystemMirGLContext.h"
```
hudokkow: Not a code review but please use only one space around operators. Yes, I know... Existing code doesn't follow this request. :wink:

hudokkow: heheh, that was fast! Thx!

BrandonSchaefer: No problem was an easy fix!

BrandonSchaefer:
Also note I do plan on maintaining this port. I Plan to maintain with the latest Mir release (currently 0.24.1 but 0.25 is coming soon). That is until I need to support multiple Mir versions.

project/cmake/CMakeLists.txt
```diff
@@ -118,6 +118,12 @@ if(CORE_SYSTEM_NAME STREQUAL android)
   list(APPEND required_deps Zip)
 endif()
 
+# Missing mir support in these libraries
+if (ENABLE_MIR)
```
fetzerch: Please move this to project/cmake/scripts/linux/ArchSetup.cmake

project/cmake/CMakeLists.txt
```diff
@@ -199,14 +205,22 @@ endif()
 if(NOT WIN32)
   core_optional_dep(OpenGl)
   if(OPENGL_FOUND)
-    core_optional_dep(X ENABLE_X11)
-    core_optional_dep(LibDRM ENABLE_X11)
-    core_optional_dep(XRandR ENABLE_X11)
+    if (ENABLE_MIR)
```
fetzerch: Good reminder that I need to clean up this spaghetti code. Ok for now, but make it `if(` without the space everywhere please.

fetzerch (commented):
Nice to see this feature wise :)
I can only comment on the cmake part and found a few minors.

project/cmake/CMakeLists.txt
```diff
@@ -118,6 +118,12 @@ if(CORE_SYSTEM_NAME STREQUAL android)
   list(APPEND required_deps Zip)
 endif()
 
+# Missing mir support in these libraries
+if (ENABLE_MIR)
+  set(ENABLE_VAAPI OFF CACHE BOOL "Disabling VAAPI since no mir support" FORCE)
+  set(ENABLE_VDPAU OFF CACHE BOOL "Disabling VDPAU since no mir support" FORCE)
```
wsnipex: please also add a message telling the user that HW acceleration is disabled because of mir

project/cmake/modules/FindMir.cmake
```diff
@@ -0,0 +1,37 @@
...
+
+if(PKG_CONFIG_FOUND)
+  pkg_check_modules (PC_MIR mirclient QUIET)
+endif()
+
+find_path(MIR_INCLUDE_DIR NAMES mir_toolkit/mir_client_library.h
+                          PATHS ${PC_MIR_INCLUDE_DIRS})
+
+find_library(MIR_LIBRARY NAMES mirclient
+                         PATHS ${PC_MIR_LIBRARIES} ${PC_LIBRARY_DIRS})
```
wsnipex: PC_MIR_LIBRARY_DIRS?

project/cmake/CMakeLists.txt
```diff
@@ -199,14 +205,22 @@ endif()
 if(NOT WIN32)
   core_optional_dep(OpenGl)
   if(OPENGL_FOUND)
-    core_optional_dep(X ENABLE_X11)
-    core_optional_dep(LibDRM ENABLE_X11)
-    core_optional_dep(XRandR ENABLE_X11)
+    if (ENABLE_MIR)
+      core_optional_dep(Mir ENABLE_MIR)
```
wsnipex: core_require_dep please. The dependency is required once the user enables mir.

And please add an option on line 58 for ENABLE_MIR

BrandonSchaefer: For the option on line 58, how would we define a system is mir and not x11? As any system thats linux will most likely be for an x11 system (that possibly supports mir)

project/cmake/CMakeLists.txt
BrandonSchaefer: O didnt even know about that cmake file, excellent!

project/cmake/modules/FindMir.cmake
```diff
@@ -0,0 +1,37 @@
...
+# This will will define the following variables::
+#
+# MIR_FOUND        - the system has Mir
+# MIR_INCLUDE_DIRS - the Mir include directory
+# MIR_LIBRARIES    - the Mir libraries
+# MIR_DEFINITIONS  - the Mir definitions
+#
+# and the following imported targets::
+#
+#   Mir::Mir   - The Mir library
```
BrandonSchaefer: Removed!

project/cmake/CMakeLists.txt
BrandonSchaefer: Im now assuming you ment an option for OFF :)

project/cmake/CMakeLists.txt
BrandonSchaefer: Hmm its still using opengl/es meaning its accelerated. Whats not implemented is hardware encoding/decoding for mir in these libraries.

BrandonSchaefer: Different acceleration you were talking about. Fixed!

wsnipex:
@FernetMenta is this something we could consider for v17, because it's very contained and doesn't touch existing code?

fritsch:
What I wonder is: Can we get VAAPI working? We don't use any X11 dependencies in the code. What is currently missing in that regard?

fritsch:
For my part v17: I does not harm, we have a maintainer with it. It is not build by default via our ppa - so users using it will most likely be only self compilers.

If ubuntu considers it for their next release, the hw decoding, especially with vaapi should be considered investigating.

xbmc/windowing/mir/WinEventsMir.h
```diff
@@ -41,6 +42,7 @@ class CWinEventsMir : public IWinEvents
 
 private:
   std::queue<XBMC_Event> events;
+  std::mutex mutex;
```
fritsch: for members we use m_ so in your case:
```
m_events;
m_mutex;
```

xbmc/windowing/mir/WinEventsMir.cpp
```diff
@@ -329,10 +331,12 @@ bool CWinEventsMir::MessagePump()
 
 size_t CWinEventsMir::GetQueueSize()
 {
+  std::lock_guard<decltype(mutex)> event_lock(mutex);
```
fritsch: that's a nice c++11 feature which could replace most of our own locking patterns, that were implement this "by hand" when there was no std::mutex and no std::lock_guard available.

MartijnKaijser:
no objection

BrandonSchaefer:
For VAAPI support (I dug into a little bit) it needs a mir backend pretty much. Something Im interested in as well. No promises of when though!

https://cgit.freedesktop.org/libva/tree/va/va_backend.h

xbmc/windowing/mir/WinEventsMir.cpp
BrandonSchaefer: c++11 (and 14) have added quite a bit of quality of life :)

xbmc/windowing/mir/WinEventsMir.h
BrandonSchaefer: Opps missed that! Fixed

fritsch:
I thought that would somehow work via EGL / DRM.

BrandonSchaefer:
That... should work, ill have to test that out!

BrandonSchaefer:
Im seeing some issues with the egl windowing system battling against mir when using opengles. I would prefer not to add an EGL Mir type soo im going to mess with that a bit.

fritsch:
If you have comments on how we can improve that without you having to introduce pseudo types that duplicate given definitions, please do not hessitate to point them out.

FernetMenta:
Mir won't work with vaapi unless someone implements something like vaGetDisplayMIR

fritsch:
If they can use DRM directly: https://cgit.freedesktop.org/vaapi/libva/tree/va/drm/va_drm.c#n90

BrandonSchaefer:
I made some changes in https://github.com/xbmc/xbmc/pull/10898/commits/21b89b2376e347a7e3c7effd79c0e5e062f500f1

Soo I dont need to create a EGL Mir type. Pretty much ignore the windowing egl system if mir is enabled!

BrandonSchaefer:
Also looking at cores/VideoPlayer/DVDCodecs/Video/VAAPI.h seems to depend on X11.

FernetMenta:
I know about the DRM backend but I doubt that it would work when the final goal is to display the surface.

BrandonSchaefer:
Soo messing around with the VAAPI, im *pretty* sure its working using DRM/EGL render nodes. (Was running into issues with our kms backend since it opens up the same dri device which work buuut render nodes seems to work)
http://paste.ubuntu.com/23464991/

A very hacked proof of concept:
http://paste.ubuntu.com/23465030/

From there its just manually checking the /dev/dri/renderD* for the correct number depending on the machine. (as long as you've a kernel >= 3.15)

Which brings me to wondering if we can replace the VAAPI DVDCodec bits with drm only? Or does the x11 specific bring better support for x11? I dont really understand everything thats going on there :).

FernetMenta:
> Which brings me to wondering if we can replace the VAAPI DVDCodec bits with drm only?

needs to be tested with wayland too. there is nothing that keeps us sticking to x11 because x11 will die anyway.

did you get a picture with drm?

BrandonSchaefer:
Yeah testing for sure. I did get picture and tested with print statements in the vaapi renderer + break point on vaRenderPicture which were all going off (which makes me think its working!). Let me screen shot though!

BrandonSchaefer:
http://i.imgur.com/2NMQMyX.jpg

fritsch:
Great! Just great!

fritsch:
For the record, still works fine with X11 Display (what was to be expected). Btw. if we are at it - there are some additional options for "transfering" * the surfaces via DRM insteand of using EGL, but still being able to render them afterwards.

drm-vaapi

* sharing in zero copy approach

fritsch:
https://github.com/fritsch/xbmc/commit/372a3ff88360ded6fb09692621cbfe88d34aa9f0 should do :-)

BrandonSchaefer:
O sweet, yeah i double checked with X11 as well which was working but didnt know about that player process info. That helps secure the fact its vaapi. Also note it doesnt work for *.divx (which im not very familiar with codecs in general I just assume vaapi doesnt support that codec.)

Just needs cleaning up on an RAII class to hold the FD as well as selecting the render node (since I assume mir/wayland will hold onto the device so you cannot drmOpen on it)

fritsch:
For DIVX: Player Settings (Expert Hierarchy) -> Enable Mpeg4

BrandonSchaefer:
Comparison on this pull request with mir using VAAPI vs X11 and vs mir with no hardware accel (ie. this branch currently)

xbmc/windowing/mir/WinSystemMirGLESContext.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#ifndef WINDOW_SYSTEM_MIR_GLES_CONTEXT_H_
```
BrandonSchaefer: Im happy to remove them though

FernetMenta: we require it in many places. a compiler that does not support pragma won't ever be able to build kodi

BrandonSchaefer: And perfectly reasonable, removing!

xbmc/windowing/mir/WinSystemMir.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
```
BrandonSchaefer: Opps, missed two files. Fixing

PR #10899, opened 2016-11-08
[guiinfo][pvr] Fix LISTITEM_RESUMABLE to prefer recording info tag over video info tag, fire recordings updated event on playcount change"

ksooo:
Some skins extend the recent recordings pvr widget by displaying the watched state of the recordings. This works quite well, but marking recordings watched / unwatched using the context menu of recordings in te recordings window will only be reflected by the widget after a restart of kodi, or in case of a partly watched recordings, never. This PR fixes this misbehavior.

The code change has been tested on macOS and Linux on latest Krypton master.

@Jalle19 for review?


ksooo:
jenkins build this please

Jalle19 (approved):

PR #10901, opened 2016-11-09
Videoplayer: If subtitles are set to "on" and the preferred language is "original stream's language" at least default subtitles should be shown, regardless of there language.

ace20022:
…default should be shown, regardless of there language.


ace20022:
jenkins build this please

ace20022:
@MartijnKaijser please don't merge as is, I need to change the commit message...

Leatherface75:
Would be nice to be able to select a 2nd language as fallback.
For example Swedish first and if not exists then select English instead.

ace20022:
jenkins build and merge

ace20022:
jenkins build and merge

PR #10903, opened 2016-11-09
[cmake] Fun with flags

hudokkow:
## Description
Change default -O3 flag to -O2 on Linux, OSX, BSD and RBPi and to -Os on droids and iOS for release builds.

## Motivation and Context
Convo in slack channel.

## How Has This Been Tested?
Build/run tested on linux only.

## Types of change

## Checklist:

Original was linux only. I added the other platforms so we could discuss if it's needed/wanted or not.

https://github.com/xbmc/xbmc/commit/7a35049d09a17bc842d37067aa4603d18d495c55 is cosmetics only. It makes reading build info easier on huge logs.

ping @stefansaraev, @wsnipex and @fetzerch


stefansaraev:
+1. I dont think -O3 is safe for project with big codebase (and lot of depends) like kodi.

btw, my real issue was (and still is) that cmake does not honor my $CXXFLAGS from environment (set by my build scripts). I personaly do prefer -Os on linux, I guess I can work around by appending ${CMAKE_CXX_FLAGS} to CMAKE_CXX_FLAGS_RELEASE :)

hudokkow:
Can you upload your script? If it's top-secret, I'll be happy with only cmake calls.

fetzerch:
Can you move it to common/ArchSetup.cmake please? I don't like that we have this code duplication.

stefansaraev:
@hudokkow 

> Can you upload your script? If it's top-secret, I'll be happy with only cmake calls.

nothing fancy. it just exports CXXFALAGS to environment. on my env it is something like

```
-march=armv8-a -mcpu=cortex-a53 -mabi=lp64 -Wno-psabi -w -pipe -Os
```
OpenELEC and LibreELEC (and maybe other distros?) are doing the same. 

when compiling kodi, -O3 (now -O2 with this PR) was automagicaly appended at the end. rendering my -Os useless. with gcc, whatever -Ox comes last - it is used.

project/cmake/CMakeLists.txt
```diff
@@ -460,31 +460,39 @@ configure_file(${PROJECT_SOURCE_DIR}/scripts/common/AddonHelpers.cmake
...
-  message(STATUS "#---------------------------------------------#")
-  message(STATUS "GIT_REV: ${APP_SCMID}")
-  message(STATUS "#---------------------------------------------#")
-  message(STATUS "CPACK_GENERATOR       : ${CPACK_GENERATOR}")
-  message(STATUS "CPACK_SOURCE_GENERATOR: ${CPACK_SOURCE_GENERATOR}")
-  message(STATUS "#---------------------------------------------#")
+message(STATUS "\n-- DLOPEN/DLCLOSE LIB ----------------------------------------------------------")
+message(STATUS "  ${CMAKE_DL_LIBS}")
+
+message(STATUS "\n-- FUN WITH FLAGS --------------------------------------------------------------")
```
wsnipex: COMPILER FLAGS please

project/cmake/CMakeLists.txt
```diff
@@ -460,31 +460,39 @@ configure_file(${PROJECT_SOURCE_DIR}/scripts/common/AddonHelpers.cmake
...
+message(STATUS "  CMAKE_C_FLAGS               : ${CMAKE_C_FLAGS}")
+message(STATUS "  CMAKE_CXX_FLAGS             : ${CMAKE_CXX_FLAGS}")
+message(STATUS "  CMAKE_ASM_FLAGS             : ${CMAKE_ASM_FLAGS}")
+message(STATUS "  CMAKE_EXE_LINKER_FLAGS      : ${CMAKE_EXE_LINKER_FLAGS}\n")
+message(STATUS "  SYSTEM_LDFLAGS              : ${SYSTEM_LDFLAGS}\n")
+
+if(NOT WIN32)
+  string(TOUPPER ${CMAKE_BUILD_TYPE} BUILD_TYPE)
```
wsnipex: this var is not really needed

project/cmake/CMakeLists.txt
```diff
@@ -460,31 +460,39 @@ configure_file(${PROJECT_SOURCE_DIR}/scripts/common/AddonHelpers.cmake
...
+  message(STATUS "  CMAKE_C_FLAGS_${BUILD_TYPE}         : ${CMAKE_C_FLAGS_${BUILD_TYPE}}")
+  message(STATUS "  CMAKE_CXX_FLAGS_${BUILD_TYPE}       : ${CMAKE_CXX_FLAGS_${BUILD_TYPE}}")
+  message(STATUS "  CMAKE_ASM_FLAGS_${BUILD_TYPE}       : ${CMAKE_ASM_FLAGS_${BUILD_TYPE}}")
+  message(STATUS "  CMAKE_EXE_LINKER_FLAGS_${BUILD_TYPE}: ${CMAKE_ASM_FLAGS_${BUILD_TYPE}}")
+
+  message(STATUS "\n-- BUILD TYPE ------------------------------------------------------------------")
+  message(STATUS "  ${BUILD_TYPE}")
```
wsnipex: please move this to the top of final message, but on one line

wsnipex (changes requested):
I'm ok with setting default cflags to -O2, although we have been building with O3 for a long time without issues that boiled down to cflags.

What's not ok is to overwrite user set flags without reason, if I set my cflags I expect them to be used.

hudokkow:
Current (master) flags we're using: http://pastebin.com/8MbAVdkM

project/cmake/CMakeLists.txt
```diff
@@ -460,31 +460,39 @@ configure_file(${PROJECT_SOURCE_DIR}/scripts/common/AddonHelpers.cmake
...
+message(STATUS "  CMAKE_ASM_FLAGS             : ${CMAKE_ASM_FLAGS}")
+message(STATUS "  CMAKE_EXE_LINKER_FLAGS      : ${CMAKE_EXE_LINKER_FLAGS}\n")
+message(STATUS "  SYSTEM_LDFLAGS              : ${SYSTEM_LDFLAGS}\n")
+
+if(NOT WIN32)
+  string(TOUPPER ${CMAKE_BUILD_TYPE} BUILD_TYPE)
+  message(STATUS "  CMAKE_C_FLAGS_${BUILD_TYPE}         : ${CMAKE_C_FLAGS_${BUILD_TYPE}}")
+  message(STATUS "  CMAKE_CXX_FLAGS_${BUILD_TYPE}       : ${CMAKE_CXX_FLAGS_${BUILD_TYPE}}")
+  message(STATUS "  CMAKE_ASM_FLAGS_${BUILD_TYPE}       : ${CMAKE_ASM_FLAGS_${BUILD_TYPE}}")
+  message(STATUS "  CMAKE_EXE_LINKER_FLAGS_${BUILD_TYPE}: ${CMAKE_ASM_FLAGS_${BUILD_TYPE}}")
```
hudokkow: Ah? Not it doesn't. See pastebin below.

project/cmake/CMakeLists.txt
hudokkow: No sense of humor... TBBT? :wink:

project/cmake/CMakeLists.txt
hudokkow: Don't follow.
It's CMAKE_EXE_LINKER_FLAGS_DEBUG, not CMAKE_EXE_LINKER_FLAGS_Debug.

Last time I checked it's case sensitive.

project/cmake/CMakeLists.txt
hudokkow: Okido.

hudokkow:
@fetzerch, I'll change it to only use common/

project/cmake/CMakeLists.txt
wsnipex: you're right, sorry :)

project/cmake/CMakeLists.txt
wsnipex: afaik cmake should automatically use the correct ones for the build type merged with external passed flags. If it doesn't then we have to merge them ourselves.

MartijnKaijser:
Still needed?

hudokkow:
Nope.

PR #10904, opened 2016-11-09
[FilesystemInstaller] - use special://temp which is correct on all platforms

Memphiz:
special://home/temp/ might not exist on some platforms - the assumption is wrong.

## Description
Use the temp dir we already provide via special://temp

## Motivation and Context
Can't install any addons anymore on osx because the temp dir is not created and the installer fails to unzip the files.

## How Has This Been Tested?
Installed an addon with the patch - it worked ...

## Types of change

## Checklist:

@tamland your button

### DIFF
--- a/xbmc/addons/FilesystemInstaller.cpp
+++ b/xbmc/addons/FilesystemInstaller.cpp
@@ -31,7 +31,7 @@ using namespace XFILE;
 CFilesystemInstaller::CFilesystemInstaller()
 {
   m_addonFolder = CSpecialProtocol::TranslatePath("special://home/addons/");
-  m_tempFolder = CSpecialProtocol::TranslatePath("special://home/temp/");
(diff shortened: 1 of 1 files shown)


tamland:
If special://temp/ is mapped to someplace else (and in turn can be placed on a different filesystem) on some systems, it *cannot* be used for this. The installer relies on moves, which doesn't work across filesystem. So either we need to make special://home/temp/ a thing, or special://home/addons/unpack or something..

Memphiz:
@tamland was this restriction there before filesysteminstaller too? And if so - how was it guaranteed back then that zip and temp dir were on the same filesystem?

tamland:
Before it didn't do any of this, no. It just unpacked directly into the addon folder over the existing files. So when something failed halfway, it left the addon in a broken state.

Memphiz:
got it ... then keep special://home/temp ... as the change from paxxi ensure that directories are created recursively now ...

PR #10906, opened 2016-11-09
Picture: Partially revert f5f9b8d and aadfc21d (keeping fixes)

fritsch:
Picture and Texture is a great mess.

I wanted to "fix" something in a fast way, by making sure that ffmpeg methods like scale and others only run on av_malloced buffers. The problem we have is, that two things are not good in current design:

a) memory is alloced "somehow" and just given into our Image
b) memory is not deleted at a central point, but all over the places

I don't want to delay v17 any further, but this needs to change for the future by providing an encapsulated method of creating and deleting memory that we use for picture processing.

### DIFF
--- a/xbmc/pictures/Picture.cpp
+++ b/xbmc/pictures/Picture.cpp
@@ -42,7 +42,6 @@
 
 extern "C" {
 #include "libswscale/swscale.h"
-#include "libavutil/mem.h"
 }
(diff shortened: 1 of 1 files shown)


fritsch:
This build fixes the issues for the windows folks.

jenkins build this please

PR #10908, opened 2016-11-09
[PVR] Guide info dialog: fix record/add timer button visibility.

ksooo:
Before this PR, in the PVR guide info dialog, the timer-related buttons like "Record" and "Add timer" were visible even if not PVR backend with timer support was available. This PR fixes this misbehavior that the buttons in question are only visible if at least one pvr backend with timer support is available.

This fix was tested on Linux on latest Kodi master with pvr.octonet (which does not support timers).

@Jalle19 for review?


ksooo:
jenkins build this please

Jalle19 (approved):

PR #10909, opened 2016-11-10
Reduce bitrate of AC3 Encoder for Android only

fritsch:
1.) Create Factory Method
2.) Add cmake ifdeffery arround factory method
3.) Set defines and additions in CMakeList.txt
4.) Replace hard defined bitrate with method that can be overwritten  
5.) Create override directory, add android override, empty virtual ctor / dtor
6.) Change Baseclass' attributes to be protected to keep it usable after inheritance
7.) Adjust original mainline code to use factory's CreateInstance()


fritsch:
@hudokkow @fetzerch as partyl discussed on IRC. If there is something in cmake that can get us nearer to the goal: separate platforms from mainline core code, please tell. I find that override way totally overcomplicated, it only works for exactly one design pattern (factory with inheritance).

besides that I mostly have done something wrong. Will lett jenkins run over night.

jenkins build this please

MrMC:
or, add a hidden value in settings
SETTING_AUDIOOUTPUT_AC3TRANSCODEBITRATE, then use android.xml to override it.

Less code, much less complications, no extra files and design pattern concerns.

FernetMenta:
> or, add a hidden value in settings

certainly not

fritsch:
It built, but it did not pickup the overrides. Need a bit help here from the cmake guys. I thought the new generic approach here would be: https://github.com/xbmc/xbmc/pull/10909/commits/d99e58ae195e4112582031d4acad9749fe74c218 in combination with: https://github.com/xbmc/xbmc/pull/10909/commits/db5533b15eec3d77a227c37375b2ede9409387ce#diff-942151f62568419073e4e7dee3bf440bR31

but it seems PLATFORM_OVERRIDE does not get set.

Any ideas?

xbmc/cores/AudioEngine/Encoders/overrides/android/AEEncoderFFmpegAndroid.h
```diff
@@ -0,0 +1,37 @@
...
+
+class CAEEncoderFFmpegAndroid: public CAEEncoderFFmpeg
+{
+public:
+  CAEEncoderFFmpeg* CreateInstance()
+  {
+    return new CAEEncoderFFmpegAndroid();
+  }
+  virtual CAEEncoderFFmpegAndroid() {};
+  virtual ~CAEEncoderFFmpegAndroid() {};
```
fetzerch: should be possible to omit both ctor/dtor. virtual destructor is inherited from the base. if you want to keep the dtor, i'd mark it override.

xbmc/cores/AudioEngine/CMakeLists.txt
```diff
@@ -131,6 +134,9 @@ endif()
 
 core_add_library(audioengine)
 target_include_directories(${CORE_LIBRARY} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+if(OVERRIDES_CPP)
+  target_compile_definitions(${CORE_LIBRARY} PRIVATE -DPLATFORM_OVERRIDE)
```
fetzerch: The define PLATFORM_OVERRIDE has to be unique, maybe PLATFORM_OVERRIDE_AE_ENCODERS. (see https://github.com/xbmc/xbmc/pull/10897). Use the same below in the ifdef.

Also OVERRIDES_CPP/OVERRIDES_H are not added to SOURCES/HEADERS and are therefore not compiled. If the overrides are not dependent on any option, this should do:

    if(OVERRIDES_CPP)
      list(APPEND SOURCES ${OVERRIDES_CPP})
      list(APPEND HEADERS ${OVERRIDES_H})
      target_compile_...

For debugging if your code is picked up, add a `message(STATUS "Use AE Encoder override")` or compile with `make VERBOSE=1` to check if the define is passed to the compiler.

fetzerch (commented):
As discussed on IRC, i'm not sure if that's really the right conceptual way to do it. You now create one OVERRIDES_CPP for Encoders, what if you need to do this also in other places of AE? 

Another thing I just realize now is the GLOB. We should in general avoid this. If you now add new files to the overrides directory, they are not picked up by cmake automatically. You have to manually call cmake again after doing that. (Not only you in particular, also the other devs that pull and get new files deployed in that dir).

xbmc/cores/AudioEngine/CMakeLists.txt
fritsch: Thx for the feedback I added some stuff. Btw. aren't the headers missing here: https://github.com/fetzerch/xbmc/blob/8108034415de1882e127ec64d0de9918cf236d2e/xbmc/cores/VideoPlayer/Process/Makefile.in#L9

koying:
-1 
Agree with @davilla 

As said and ignored on Slack, it needs to be a advanced setting or even a plain ifdef.
This is a dogmatic, sub-optimal and over-engineered, imho.

FernetMenta:
there won't be any settings abuse in code I maintain

koying:
I don't see any abuse in replacing a hardcoded value by a setting, rather than having an hardcoded value per platform in code...

fritsch:
@fetzerch told me on IRC that he has a solution in mind that works more easy without any ifdefs. I will close that PR here. For the record: I've been here, I implemented it. I understand how it works, how I don't like it at all and why can be read in slack.

MrMC:
There are plenty of hidden settings values right now. They work just fine, as does the existing platform override method for settings. I don't see this as an abuse of settings at all, settings values can be hidden for all, hidden for specific platforms, or hidden/shown depending on runtime setup.

PR #10910, opened 2016-11-10
Fix PS4 triggers and accelerometers from skipping in the controller dialog

garbear:
This fixes rapid button skips in the controller dialog when mapping certain controllers. These include controllers with accelerometers, and PS4 triggers.

Accelerometers cause skipping because we can't tell the difference between accelerometers and analog sticks/triggers: both use anonymous axes. When motion (or gravity) occurs, the configuration utility thinks that an analog stick is being moved.

In the case of PS4 triggers, the controller sends two values - a button and an axis. The button is sent when the axis crosses 0.5. However, the button mapper only maps when the axis exceeds 0.7, so we need to ignore the button press.

## Description
This fix adds a new button to the controller dialog named "Fix skipping". This opens a dialog that begins capturing all controller input. Anything captured by the dialog is ignored during the mapping process.

## How Has This Been Tested?
Button map generated on windows: [Wireless_Controller_v054C_p09CC_14b_1h_6a.xml](https://github.com/kodi-game/peripheral.joystick/blob/54cfc4c/peripheral.joystick/resources/buttonmaps/xml/directinput/Wireless_Controller_v054C_p09CC_14b_1h_6a.xml)
Button map generated on OSX: [Wireless_Controller_v054C_p09CC_14b_6a.xml](https://github.com/kodi-game/peripheral.joystick/blob/54cfc4c/peripheral.joystick/resources/buttonmaps/xml/cocoa/Wireless_Controller_v054C_p09CC_14b_6a.xml)

On OSX, the D-pad on the PS4 controller is broken. I think this is because we need to extend our cocoa API to accept hat input.

fix skipping

## Types of change

This bumps the peripheral API to v1.2.0. Requires https://github.com/xbmc/peripheral.joystick/pull/64

Broken out from #10630


garbear:
jenkins build this please

garbear:
jenkins build and merge

PR #10911, opened 2016-11-11
[PVR] Settings: change label and description to reflect that 'TV' is actually 'PVR & Live TV.'

ksooo:
We discussed this on slack some time ago, but somehow it got (partly) lost.

screenshot002
screenshot003

@jjd-uk is this okay?

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -7541,12 +7541,12 @@ msgstr ""
 
 #: system/settings/settings.xml
 msgctxt "#14204"
-msgid "TV"
+msgid "PVR & Live TV"
 msgstr ""
 
 #: system/settings/settings.xml
 msgctxt "#14205"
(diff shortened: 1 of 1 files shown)


jjd-uk:
While I agree it change be changed from "TV" I tried to ensure the category labels weren't too long so not 100% on this change. I would be tempted to just call it "PVR" as in the UK at least PVR's have been around 15 years, and is widely understood by nearly everyone here. Is PVR not so widely understood in other countries? or is it more to do with the case that some add-on's don't facilitate recording?

Note this is not an objection so feel free to merge as is, since it's better than what we currently have.

ksooo:
> or is it more to do with the case that some add-on's don't facilitate recording?

 yes indeed.

jjd-uk:
Then ok from my end

ksooo:
> Then ok from my end

 thanks. if nobody objects i will merge this tomorrow.

PR #10912, opened 2016-11-11
AEEncoderFFmpeg: Make bitrate platform replacable

fritsch:
Use cmake / autotools to generate the define for us depending on which platform we run.

384000 on Android, 64000 on the rest


fritsch:
jenkins build this please

FernetMenta:
behavior must not be hidden in the build system. it has to be clear what parts of the app are overridden for platforms. you could define the variable in an separate file in a subfolder

fritsch:
Thanks for that feedback. Looking forward when your remove the build system behaviour of your alternative overrides approach.

wsnipex:
>behavior must not be hidden in the build system. it has to be clear what parts of the app are overridden for platforms. you could define the variable in an separate file in a subfolder

you are aware that tons of defines come from the build system? also the override crap depends on the build system...

MrMC:
hahahahha :)

FernetMenta:
@fetzerch and I are working on a solution

PR #10916, opened 2016-11-12
[jsonrpc] "added" data for new item *Library.OnUpdate notification

rmrector:
## Description

This PR adds `"added": true` to JSON-RPC "OnUpdate" notification "data" if it is announced for a newly added* media item.

__*__ "Refreshing" an item generally removes and then re-adds the item, so this change also affects the OnUpdate notification for refreshed items.

This is a thin addition to the existing API (no code should break with a new key: value pair here), but it might be preferable to instead break compatibility and add a new notification "OnAdd"? I could rework it if so.

## Motivation and Context

Consuming add-ons can easily tell if the notification is for a new* item or contains just a change to some of its info if they need to treat them differently. For instance, add-ons could keep additional information in sync with Kodi's database, but it may be able to skip possibly expensive operations if only play count or the title is updated.

## How Has This Been Tested?

Poked through several situations that generate "OnUpdate" notifications, and ensured data includes the "added" key when and only when a new* item has been added.

## Types of change

## Checklist:


da-anda:
ping @Montellese, mind having a look? thanks.

rmrector:
Small history rewrite because this is not Java: use default value for new parameter of AnnounceUpdate.

MartijnKaijser:
@Montellese new ping :)

rmrector:
Popped off the json-rpc version bump for simpler picking, I can add it back if it gets merged.

And maybe a 🤘 🐰 poke @Montellese

Paxxi:
Sorry about the silence here.
Thank you for your time and patience. 

I believe adding a parameter should also bump the API version. Normally a json deserializer should ignore it or the layer above but there's always the chance that it could break implementations that expects exactly what the schema says and nothing more

xbmc/interfaces/json-rpc/schema/notifications.json
```diff
@@ -119,6 +119,7 @@
         "properties": {
           "id": { "$ref": "Library.Id", "required": true },
           "type": { "type": "string", "id": "Notifications.Library.Audio.Type", "enum": [ "song" ], "required": true },
+          "added": { "$ref": "Optional.Boolean", "description": "True if the update is for a newly added item." },
```
Montellese: Please add the new property at the end.

xbmc/interfaces/json-rpc/schema/notifications.json
```diff
@@ -200,6 +201,7 @@
           "id": { "$ref": "Library.Id", "required": true },
           "type": { "type": "string", "id": "Notifications.Library.Video.Type", "enum": [ "movie", "tvshow", "episode", "musicvideo" ], "required": true },
           "playcount": { "type": "integer", "minimum": 0, "default": -1 },
+          "added": { "$ref": "Optional.Boolean", "description": "True if the update is for a newly added item." },
```
Montellese: Same here

xbmc/music/MusicDatabase.cpp
```diff
@@ -94,13 +94,15 @@ static void AnnounceRemove(const std::string& content, int id)
   ANNOUNCEMENT::CAnnouncementManager::GetInstance().Announce(ANNOUNCEMENT::AudioLibrary, "xbmc", "OnRemove", data);
 }
 
-static void AnnounceUpdate(const std::string& content, int id)
+static void AnnounceUpdate(const std::string& content, int id, bool added=false)
```
Montellese: Spaces around the `=`

Montellese (changes requested):
Are you sure that you covered all the cases where new items are added to the database? I'd have to check myself but I'd have expected that there are places in `CVideoDatabase` that need to be adjusted as well but I might be wrong.

And as @Paxxi has pointed out please bump the minor version of the API.

rmrector:
All updated. I just popped the version off temporarily in case the PR was going to sit here for a while longer, it was out of date anyway.

@Montellese with the myriad ways of playing a media item I had the same thought, "this can't be everything", but Kodi only has the file scanner and import to add new library items, which both call VideoInfoScanner.AddVideo. I also looked through every "OnUpdate" notification generated and the rest were all just updates to data like playcount, title, and fanart.

Montellese (approved):
Thanks for the changes. Nothing else from me. jenkins build and merge

Rechi:
@rmrector this needs a rebase

rmrector:
@Rechi updated. Please don't leave this unmerged until the version goes out of date again.

Rechi:
@rmrector this PR now contains a merge commit. We don't want them inside a PR to have a clean history.
Please remove it and if necessary rebase on our master branch.
If you need assistance for that task, just ask us.

rmrector:
@Rechi Ohh, that's what I get for using Github's helpful little editor. I may have missed a switch for that, though. Anyway, fixed.

Rechi:
@Montellese is this still fine to be merged?

Montellese:
Still fine with me.

rmrector:
@Rechi don't forget about this one.

rmrector:
Thanks machine man Martijn!

PR #10918, opened 2016-11-12
[cmake] Platform specific overrides

fetzerch:
This is a follow-up on the platform override discussions we had on slack and in the 2 predecessor PRs https://github.com/xbmc/xbmc/pull/10912 and https://github.com/xbmc/xbmc/pull/10909.

I've worked on a couple of minimal examples for different options in a local branch: https://github.com/xbmc/xbmc/compare/master...fetzerch:platform_settings, then discussed with @FernetMenta and implemented to most promising one (option5) for the concrete AE usecase.

Nothing set in stone yet, comments on both, this - or any of the 4 other options in my branch are still highly appreciated.

The last commit is just for easier testing on linux, you can revert the changes to locally switch between override and no override.

My notes might be useful for easier understanding what happens:
unbenannt

Some more notes and the requirements for this are accessible at: https://drive.google.com/open?id=1E1P2SWELCtUpJxHFeDf0TGi_H6xge-Co2rhCdJN-c4M

### DIFF
--- a/cmake/scripts/common/Macros.cmake
+++ b/cmake/scripts/common/Macros.cmake
@@ -56,6 +56,53 @@ function(source_group_by_folder target)
   endif()
 endfunction()
 
+# Marks header file as being overridden on a certain list of platforms.
+#
+# Explicitly marking a file as overridden on specific platforms avoids issues with globbing where
+# CMake would have to be called manually when overriding for a new platform.
+#
+# Usage: add_platform_override(${PROJECT_NAME} settings.h PLATFORMS android linux osx)
+function(add_platform_override target filename)
+  cmake_parse_arguments(ARG "" "" "PLATFORMS" ${ARGN})
+  if(NOT ARG_PLATFORMS)
+    message(FATAL_ERROR "Missing parameter PLATFORMS")
+  endif()
+
+  # Generate an _override.h header that is either empty (platform doesn't define overrides)
+  # or includes the corresponding platform override header.
+  # This _override.h has to be included by the generic header.
+
+  # Determine filename of override header.
+  string(REPLACE ".h" "_override.h" override_file ${filename})
+
+  # Check if we have an override defined for this platform.
+  # TODO: Replace by if(IN_LIST) once we bump to CMake 3.3
+  if(";${ARG_PLATFORMS};" MATCHES ";${CORE_SYSTEM_NAME};")
+    message(STATUS "Override active for ${filename} on ${CORE_SYSTEM_NAME}")
+    file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/${override_file} "#include \"overrides/${CORE_SYSTEM_NAME}/${filename}\"")
+
+    # Add platform specific header to target sources (for IDEs)
+    target_sources(${target} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/overrides/${CORE_SYSTEM_NAME}/${filename})
+  else()
+    # Issue an error if a file exists but it's not listed in add_platform_override.
+    if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/overrides/${CORE_SYSTEM_NAME}/${filename})
+      message(FATAL_ERROR "Disabled platform override file detected, add it to the 'add_platform_override' call.")
+    endif()
+
+    message(STATUS "Override disabled for ${filename}, using generic implementation")
+    string(CONCAT COMMENT "// No platform override defined for ${CORE_SYSTEM_NAME}. To add overrides:\n"
+                          "// Create '${CMAKE_CURRENT_SOURCE_DIR}/overrides/${CORE_SYSTEM_NAME}/${filename}' and redefine symbols from '${filename}'.\n"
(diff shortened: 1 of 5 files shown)


FernetMenta:
@fetzerch thanks very much.

project/cmake/scripts/common/Macros.cmake
```diff
@@ -51,6 +51,53 @@ function(source_group_by_folder target)
   endif()
 endfunction()
 
+# Marks header file as being overridden on a certain list of platforms.
+#
+# This avoids globbing as overriding for a new platform has to me marked explicitly,
```
PeterPablo: `me`->`be`

wsnipex:
why is this any better then simply adding the override header from CMakeLists directly?
This still is pretty opaque and doesn't seem to provide any benefits.

fetzerch:
How would that look like? can you show an example?

wsnipex:
like this https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/VideoRenderers/CMakeLists.txt#L22

I have the feeling I'm missing something really obvious here...

fetzerch:
`if(CORE_SYSTEM_NAME STREQUAL android)` works fine if you need to compile in a different or additional cpp file. `list(APPEND HEADERS xyz.h)` alone doesn't work. Adding header files to targets doesn't actually do anything with the header (it doesn't pass `-include` to gcc or anything like that), it's just a hint to the generators to make the files visible in IDEs.

The use case in the PR is that you have an existing cpp file (AEEncoderFFmpeg.cpp) and you only have a couple of defines that are different on some platforms. We don't want to introduce the overhead of maintaining a different version of the cpp file to make it work.

fritsch:
@fetzerch what's the status here? The solution looks fine. It basically allows us to separate includes per platform. Do you still plan on integrating this?

I am still in need for this exact use case as AC3 encoding does not work on Sony, Philips and many more Android TVs which have a max of 384 kbit/s

fetzerch:
I received only very little feedback until now - compared to the rage there was on the old/existing way of doing things in the slack discussion.

For me this still is the best compromise considering the requirements I got from @FernetMenta and you (fritsch). Since no one came up with a better idea in the meantime, we can just proceed with it. I'll rebase it, please let me know if you want me to keep the AE change or if I should keep only the cmake macro.

@wsnipex are your concerns still valid? Or did my comment help?

fritsch:
@fetzerch  Thanks for coming back on that issue. I can easily live with this realization, it is much better and easier to do as the factory approach, which forced everything into one design. It seperates platform includes into separate headers and concentrates the build system changes in one point. The AE change is exactly what I need for Android. Thanks much.

fetzerch:
Thanks @fritsch. Will merge tomorrow if there are now objections.
Guess we need a backport for Krypton?

jenkins build this please

fritsch:
Yes, please. Thanks much. I need the limited bitrate for the broken Android fws around there (sadly some thousands :-() of named brands.

FernetMenta:
+1 thanks!

fetzerch:
jenkins build this please

cmake/scripts/common/Macros.cmake
```diff
@@ -56,6 +56,53 @@ function(source_group_by_folder target)
   endif()
 endfunction()
 
+# Marks header file as being overridden on a certain list of platforms.
+#
+# Explicitly marking a file as overridden on specific platforms avoids issues with globbing where
+# CMake would have to be called manually when overriding for a new platform.
+#
+# Usage: add_platform_override(${PROJECT_NAME} settings.h PLATFORMS android linux osx)
```
FernetMenta: @fetzerch I notices that this does not really fit our requirements. our platforms are defines by the various windowing systems. currently those are:
X11, Android, Windows, OSX, IOS, PI, IMX, AML

could you adapt this?

FernetMenta:
@fetzerch @lrusak can we use this as an example on how to use the new CORE_PLATFORM_NAME?

lrusak:
@FernetMenta totally. I'll try and finish it up soon.

phunkyfish:
@lrusak @fetzerch 

Is this PR still relevant, should it be closed or go on backburner?

phunkyfish:

fetzerch:
To be honest, I have no idea if it's still needed and/or wanted. Maybe @fritsch?

Back then I've been asked to think about a generic solution for this kind of problem and worked out a few different ideas. If I remember correctly this solution got quite some negative feedback, because it moves `#ifdefs` from code to CMake.

phunkyfish:
Let’s see if @lrusak or @fritsch comment on this.

lrusak:
Take my opinion with a grain of salt as I don't really understand the problem that this is trying to solve but by the looks of it this will make the build more opaque as it changes behaviour in a non standard way (ie. via cmake including different headers rather than typical header inclusion in c/c++)

phunkyfish:
I think we need less opacity so maybe we close this ;)

phunkyfish:
Going to close this, can be reopened at a later date if required.

PR #10921, opened 2016-11-12
Fix 100% cpu usage after a while while scanning for peripheral events

garbear:
When float is large enough, small increments no longer increase it. nextScanMs becomes stuck at a large value and scanIntervalMs will not change it.

After a little over 6 days, nextScanMs becomes  536870912.0f ms, after which adding the scan interval (16.66666f ms) no longer has effect.

Broken out from #10630

### DIFF
--- a/xbmc/peripherals/EventScanner.cpp
+++ b/xbmc/peripherals/EventScanner.cpp
@@ -49,16 +49,16 @@ void CEventScanner::Stop(void)
   StopThread(true);
 }
 
-EventRateHandle CEventScanner::SetRate(float rateHz)
+EventRateHandle CEventScanner::SetRate(double rateHz)
 {
   CSingleLock lock(m_mutex);
 
-  const float oldRate = GetRateHz();
+  const double oldRate = GetRateHz();
 
   EventRateHandle handle = EventRateHandle(new CEventRateHandle(rateHz, this));
   m_handles.push_back(handle);
 
-  const float newRate = GetRateHz();
+  const double newRate = GetRateHz();
 
   CLog::Log(LOGDEBUG, "PERIPHERALS: Event sampling rate set from %.2f to %.2f", oldRate, newRate);
 
@@ -69,7 +69,7 @@ void CEventScanner::Release(CEventRateHandle* handle)
 {
   CSingleLock lock(m_mutex);
 
-  const float oldRate = GetRateHz();
+  const double oldRate = GetRateHz();
 
   m_handles.erase(std::remove_if(m_handles.begin(), m_handles.end(),
     [handle](const EventRateHandle& myHandle)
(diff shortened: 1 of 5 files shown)


xbmc/peripherals/EventScanner.cpp
```diff
@@ -69,31 +69,31 @@ void CEventScanner::Release(CEventRateHandle* handle)
...
   while (!m_bStop)
   {
     m_scanEvent.Reset();
 
     m_callback->ProcessEvents();
 
-    const float nowMs = static_cast<float>(SystemClockMillis());
-    const float scanIntervalMs = GetScanIntervalMs();
+    const double nowMs = static_cast<double>(SystemClockMillis());
+    const double scanIntervalMs = GetScanIntervalMs();
```
akva2: consider using xbmcThreads::EndTime. it handles wrap-arounds for you which is a notorious problem on windows (it loops after 40h or so).

garbear: XbmcThreads::EndTime brings the advantage of wrap-around protected, but it could cause drift by relying on too many calls to SystemClockMillis(). One design goal for this class was close to zero drift so that joystick polling could stay in sync with RetroPlayer. EndTime would have to be modified to use a previous start time plus increment instead of a new call to SystemClockMillis(). We could drop the zero-drift requirement, but I'd rather just add a 2-line fix to account for drift in the event scanner.

garbear:
jenkins build this please

xbmc/peripherals/EventScanner.cpp
```diff
@@ -95,6 +95,10 @@ void CEventScanner::Process(void)
     const double nowMs = static_cast<double>(SystemClockMillis());
     const double scanIntervalMs = GetScanIntervalMs();
 
+    // Handle wrap-around
+    if (nowMs < nextScanMs)
+      nextScanMs = nowMs;
```
popcornmix: Has this been tested? I don't believe floating point quantities will wrap back to zero.
They tend to stop incrementing when the exponent gets too high or reach Inf. Never zero.

Personally I think dealing with times is better done with integers.

garbear: The floating point quantity has an integer source, `SystemClockMillis()`, which uses unsigned int and can wrap around, so this check is needed.

I used doubles not because we're working with times, but because we're working with rates. This means that after reciprocating to get time, we're dealing with rational numbers and would only lose precision by casting earlier than the call to Sleep()

popcornmix: You are right - I thought nowMs was being incremented. Looking closer the code looks okay.

garbear:
jenkins build this please

PR #10922, opened 2016-11-12
Use DRM as a fallback if not on X11 for VAAPI

BrandonSchaefer:
Use DRM in VAAPI as a fallback if X11 is not present.

## Description
If we are not on the X11 windowing system lets fall back to using DRM for VAAPI. Which will use vaGetDisplayDRM using render nodes. This requires opening any device in /dev/dri/renderD<num>.

## Motivation and Context
This will allow hardware acceleration in other display servers that use DRM/EGL such as mir (and most likely wayland, untested).

## How Has This Been Tested?
This has been tested on X11 and my Mir branch here: https://github.com/xbmc/xbmc/pull/10898 on ubuntu 16.10 (yakkety)

This change does require at lease a kernel >= 3.15.

http://i.imgur.com/OGtB2W7.jpg (x11)

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
@@ -32,7 +32,7 @@
 #include "guilib/GraphicContext.h"
 #include "settings/MediaSettings.h"
 #include "settings/AdvancedSettings.h"
-#include <va/va_x11.h>
+#include <va/va_drm.h>
 #include <va/va_drmcommon.h>
 #include <drm_fourcc.h>
 #include "linux/XTimeUtils.h"
@@ -45,7 +45,12 @@ extern "C" {
 #include "libavfilter/buffersrc.h"
 }
 
+#ifdef HAVE_X11
+#include <va/va_x11.h>
+#endif
+
 #include <va/va_vpp.h>
+#include <xf86drm.h>
 
 using namespace VAAPI;
 #define NUM_RENDER_PICS 7
@@ -86,6 +91,11 @@ void CVAAPIContext::Release(CDecoder *decoder)
 void CVAAPIContext::Close()
 {
   CLog::Log(LOGNOTICE, "VAAPI::Close - closing decoder context");
+  if (m_renderNodeFD >= 0)
+  {
+    close(m_renderNodeFD);
+  }
+
   DestroyContext();
 }
 
@@ -123,14 +133,67 @@ bool CVAAPIContext::EnsureContext(CVAAPIContext **ctx, CDecoder *decoder)
   return true;
 }
 
-bool CVAAPIContext::CreateContext()
+void CVAAPIContext::SetValidDRMVaDisplayFromRenderNode()
 {
-  { CSingleLock lock(g_graphicsContext);
-    if (!m_X11dpy)
-      m_X11dpy = XOpenDisplay(NULL);
+  int const buf_size{128};
+  char name[buf_size];
+  int fd{-1};
+
+  // 128 is the start of the NUM in renderD<NUM>
+  for (int i = 128; i < (128 + 16); i++)
+  {
+    snprintf(name, buf_size, "/dev/dri/renderD%u", i);
+
+    fd = open(name, O_RDWR);
+
+    if (fd < 0)
+    {
+      continue;
+    }
+
+    auto display = vaGetDisplayDRM(fd);
+
+    if (display != nullptr)
+    {
+      m_renderNodeFD = fd;
+      m_display = display;
+      return;
+    }
   }
 
-  m_display = vaGetDisplay(m_X11dpy);
+  CLog::Log(LOGERROR, "Failed to find any open render nodes in /dev/dri/renderD<num>");
+}
+
+void CVAAPIContext::SetVaDisplayForSystem()
+{
+  auto win_system = g_Windowing.GetWinSystem();
+  if (win_system == WINDOW_SYSTEM_X11)
+  {
+#if HAVE_X11
+    { CSingleLock lock(g_graphicsContext);
+      if (m_X11dpy == nullptr)
+        m_X11dpy = XOpenDisplay(nullptr);
+    }
+
+    m_display = vaGetDisplay(m_X11dpy);
+#endif
+  }
+  else
+  {
+    // Render nodes depends on kernel >= 3.15
+    SetValidDRMVaDisplayFromRenderNode();
+  }
+}
+
+bool CVAAPIContext::CreateContext()
+{
+  SetVaDisplayForSystem();
+
+  if (m_display == nullptr)
+  {
+    CLog::Log(LOGERROR, "Failed to find any VaDisplays for this system");
+    return false;
+  }
 
   int major_version, minor_version;
   if (!CheckSuccess(vaInitialize(m_display, &major_version, &minor_version)))
@@ -248,11 +311,6 @@ bool CVAAPIContext::CheckSuccess(VAStatus status)
 VADisplay CVAAPIContext::GetDisplay()
 {
   return m_display;
-}
-
-Display *CVAAPIContext::GetX11Display()
-{
-  return m_X11dpy;
 }
 
 bool CVAAPIContext::IsValidDecoder(CDecoder *decoder)
@@ -1011,7 +1069,6 @@ bool CDecoder::ConfigVAAPI()
   memset(&m_hwContext, 0, sizeof(vaapi_context));
 
   m_vaapiConfig.dpy = m_vaapiConfig.context->GetDisplay();
-  m_vaapiConfig.x11dsp = m_vaapiConfig.context->GetX11Display();
   m_vaapiConfig.attrib = m_vaapiConfig.context->GetAttrib(m_vaapiConfig.profile);
   if ((m_vaapiConfig.attrib.value & VA_RT_FORMAT_YUV420) == 0)
   {
@@ -2397,7 +2454,6 @@ bool COutput::CheckSuccess(VAStatus status)
 
 bool COutput::CreateEGLContext()
 {
-  m_Display = g_Windowing.GetDisplay();
   EGLDisplay eglDisplay = g_Windowing.GetEGLDisplay();
   EGLContext eglMainContext = g_Windowing.GetEGLContext();
   EGLConfig eglMainConfig = g_Windowing.GetEGLConfig();
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.h
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.h
@@ -125,7 +125,6 @@ struct CVaapiConfig
   VADisplay dpy;
   VAProfile profile;
   VAConfigAttrib attrib;
-  Display *x11dsp;
   CProcessInfo *processInfo;
 };
 
@@ -378,7 +377,6 @@ public:
   static bool EnsureContext(CVAAPIContext **ctx, CDecoder *decoder);
   void Release(CDecoder *decoder);
   VADisplay GetDisplay();
-  Display* GetX11Display();
   bool SupportsProfile(VAProfile profile);
   VAConfigAttrib GetAttrib(VAProfile profile);
   VAConfigID CreateConfig(VAProfile profile, VAConfigAttrib attrib);
@@ -386,14 +384,15 @@ public:
 private:
   CVAAPIContext();
   void Close();
+  void SetVaDisplayForSystem();
   bool CreateContext();
   void DestroyContext();
   void QueryCaps();
   bool CheckSuccess(VAStatus status);
   bool IsValidDecoder(CDecoder *decoder);
+  void SetValidDRMVaDisplayFromRenderNode();
   static CVAAPIContext *m_context;
   static CCriticalSection m_section;
-  static Display *m_X11dpy;
   VADisplay m_display;
   int m_refCount;
   int m_attributeCount;
@@ -401,6 +400,10 @@ private:
   int m_profileCount;
   VAProfile *m_profiles;
   std::vector<CDecoder*> m_decoders;
+  int m_renderNodeFD{-1};
+#ifdef HAVE_X11
+  static Display *m_X11dpy;
+#endif
 };
 
 /**


BrandonSchaefer:
*im* pretty sure this would be consider a breaking change since (marked as such) ... ideally *nothing* changes but now the VAAPI will be using DRM soo there could be some unforeseen edge cases ive yet to hit in test.

BrandonSchaefer:
I bring comparisons using 3840x2160p60 video:

So for mir on its KMS platform fullscreen:
http://imgur.com/a/NhZX3

Mir using no hardware accel:
http://imgur.com/a/drkIg (thats bad)

X11 same DRM VAAPI code:
http://imgur.com/a/YXJPl

Strangely quite a huge different in skip and drop == FPS (not sure if thats expected?).

BrandonSchaefer:
Seems to be hard to get *good* drop/skip numbers when you switch VTs or move workspaces. Though they seem very close in numbers besides X11 have a 0 drop when never moving a workspace (soo very good!) Mir I can seem to get ~20 drop but I have to switch VT to send the debug message not sure if thats causing extra drop doing that.

BrandonSchaefer:
Ok, so turning on debug by default for Mir finally solves the dropping (since i needed to switch a vt to send the PlayerDebug event)

http://i.imgur.com/HkIpbg2.jpg

Drop: 0 Skip: 0... yay!

BrandonSchaefer:
Some more testing running Big Buck Bunny 3840x2160p60 on both Mir and X11 with the same code from start 0:00 until 8:18 (about when the credits start). Kodi is compiled with player debug enabled so no extra WM stuff getting in the way.

Mir: drop 0 skip 3
X11: drop 0 skip 101

If theres a better way to test this sort of thing let me know!

fritsch:
Hi Brandon, I would just map the PlayerDebug to ctrl-o in a custom keyboard.xml, therefore create: ```~/.kodi/userdata/keymaps```

with the following content

```
<keymap>
  <fullscreenvideo>
    <keyboard>
      <o mod="ctrl">PlayerDebug</o>
    </keyboard>
  </fullscreenvideo>
</keymap>
```

Then just press ctrl-o from time to time. Keeping it on screen produces further load, so just check every some seconds / minutes.

fritsch:
The code is fine. I don't see a problem when comparing also our render part with: https://dvdhrm.wordpress.com/tag/drm/ . We succesfully use: EGL_LINUX_DMA_BUF_EXT to display it afterwards. 

On thing that could be tested on MIR though is: Go to setting and disable Prefer VAAPI-Output - then a cpu intensive SSE4 copy is used that is based on vaDeriveImage and vaMapBuffer. It should just continue to work. If not - that's on the list to be dropped anyways.

xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
```diff
@@ -123,14 +128,44 @@ bool CVAAPIContext::EnsureContext(CVAAPIContext **ctx, CDecoder *decoder)
   return true;
 }
 
+namespace
+{
+int find_open_render_node_fd()
```
FernetMenta: I think this is fine as a fallback when running headless (future case), but with UI we should not bypass vaapi api for getting the display handle. Consider the case with multiple cards and monitors connected.

https://cgit.freedesktop.org/libva/tree/va
^^ there should be code to get vaDisplay for MIR. Would you know if anybody intents to write and submit this code?

btw: we prefer camel case for function names.

FernetMenta: > I don't see the point depending a decoder on a display server when it is not needed.

makes no sense using the gpu one one gfx card when you have the display on the other one. vaapi provides the API for getting the display on the various systems and this has to be used.

fritsch: Which is no reason to depend the decoder on a specific display API (like X11, etc.). As you see from the patches in the other PR, it will solely depend on EGL as a standard. If your X11 supports it, your MIR, your wayland whatever - you easily get exactly that egl display, which is totally DisplayServer unaware and just speaks the EGL API.

FernetMenta: You seem not to understand. Are you saying the folks who do vaapi are wrong in providing backend code? You have to find the correct drm fd for the display in use. Pulling this code from vaapi, where it belongs to, to the application level is wrong.

fritsch: I am saying that with the implementation of vaapi drm code, it was exactly (!) designed and implemented to add a drm node into the code. 

The only reason for depending on a server _was_ the rendering part. I emphasize: "was" - we currently depend on X11 without a single reason as this only provides us vaPutSurface which we don't use - we don't use any X11 specifics in our VAAPI. Depending on it is useless.

Then DRM and DMA sharing infrastructure was implemented afterwards - after wayland, long after X11 - and now vaapi can be used without any dependency to any display server, while rendering can be done with a sane API.

if you check the wayland backend, you will see that only vaPutSurface again depends on this display server dependency.

fritsch: You can see that in e.g. Android: https://cgit.freedesktop.org/vaapi/libva/tree/va/android/va_android.cpp

The only method actually using this dependency is: vaPutSurface - nothing else has a dependency on the display server.

FernetMenta: If you finally want to display a decoded frame on a monitor, you have a dependency to that system, whatever it is. Consider a multi-gfx-card, multi-monitor system. That means that windowing has to be involved in selecting the right render node. It is NOT sufficient to select any render node for the give use case.

fritsch: It really seems I don't see it. As in my opinion only the EGLDisplay matters. As this one decides at the end ...

But to shorten this discussion that leads to nowhere. @BrandonSchaefer if you want please add a mir backend - it only needs to do what the android backend does. For the vaPutSurface you can just return UNIMPLEMENT - vaapi has such an error return value.

Then this can be sent to the VAAPI guys for review. If they say: that's totally not needed, just use drm prime - fine. if they take it, also fine.

We - in kodi need that drm mode anyways the very moment we speak transcoding, so it will come back in a different form anyways. Especially for the use case when we use a display and additionally use the kodi instance for transcoding - we will directly open the DRM no matter what display server is running for us.

fritsch: I see. As arguments and technical references don't count, I am really wrong in that discussion anyways. Sorry for disturbing you on your "I am right, no one else is. Else no one has a say besides me" - planet.

BrandonSchaefer: If it works out better I can just ifndef some functions for X11 vs Mir as well. Ill try to take some time to looking at doing an implementation for Mir in vaapi.

Seems like currently its hardcoded for X11 + EGL, I dont have a multiple GPU on this machine but reading about render nodes they are independent of cards ie. any *should* work but I cant test that.

I would prefer to at lease has this DRM as backup support since I would assume most cases have a single GPU, which is better then software rendering.

FernetMenta: > but reading about render nodes they are independent of cards ie. any should work but I cant test that

Assume you have a system with multiple cards. You would have a render node for every device. Does not matter as long you do not intend to bring the vaSurface to display. But in case you do want to display it, you need to find the render node for the device the app has its display on.

BrandonSchaefer: Reading this:
https://dvdhrm.wordpress.com/tag/drm/

It states a few times:
"It’s also important to know that render-nodes are not bound to a specific card."
"Instead, if user-space requires hardware-acceleration, it should open any node and use it."
"Really! Questions like “how do I find the render-node for a given card?” don’t make any sense. Yes, driver-specific user-space can figure out whether and which render-node was created by which driver, but driver-unspecific user-space should never do that!"

From that, and since kodi should be driver unspecific this should be a perfect use case of the any render node?

BrandonSchaefer: Also showing two kodis using DRM VAAPI with different EGL contexts in Mir
http://i.imgur.com/0cEq880.jpg

FernetMenta: You agree that vaapi processing is done by gpu, do you? Assume we have 2 cards: gpu1 and gpu2. We would also have two render nodes. Still agree? Now vaapi processing can either be on gpu1 or gpu2, right? Assume you select render node for device gpu1 but our display in on gpu2.
So vaapi decoding is done on gpu1 but we want to display the pic on display connected to gpu2. Does this sound as a reasonable scenario for you?

FernetMenta: @BrandonSchaefer why don't you implement something like this

```
if (winsystem == wayland)
  vaDisplay = vaGetDisplayWL();
else if (winsystem == X11)
  vaDisplay = vaGetDisplayX11();
else if (winsystem == Mir)
{
  // todo
  // vaDisplay = vaGetDisplayMir();
  // intermediate, better than nothing
  vaDisplay = vaGetDisplayDRM();
}
else
  vaDisplay = vaGetDisplayDRM();
```

BrandonSchaefer: That sounds good, though I think ill stick with just X11/DRM atm and once the Mir branch lands I can purpose a change for Mir (same with wayland since I dont think its around)

BrandonSchaefer: Issue is its going to have to include va_x11 which pulls in Xlib.h. Soo ill have to #ifdef around it.

BrandonSchaefer: We could also ... g_Windowing.GetVaDisplay(); Which if unsupported would just return nullptr

FernetMenta: I'd say we start with ifdef until we have a better idea.
Are you aware about Kodi's release process? Currently we are in beta, close to RC stages. Means that new features are targeted to v18. So we are not in a hurry with this. Once v17 has branched, master should be open for business again.

BrandonSchaefer: I was not! Though this does make this a more or less fall back now (wont be a breaker of things). I just tend to like to get things in a finished state if I can :)

RAOF: So, there are two parts to VAAPI usage:
1) Hardware video decoding
2) Presentation, using vaPutSurface.

(1) doesn't require that the GPU being used for decoding is the same that will be used for presentation; the underlying dma-buf mechanism makes this transparent¹. Indeed, if the non-display GPU happens to have better decoding hardware - for example, it supports H265 whereas the display GPU doesn't - it would be preferable to use the non-display GPU rather than fall back to software decoding.

This could be a sensible extension to the current code - open *each* DRM rendernode and use the most featureful one. A second extension would be to open each rendernode and potentially use both when decoding multiple streams - for picture-in-picture decoding, for example.

¹: It's slightly preferable to decode on the GPU you'll be displaying on, as that saves a GPU→GPU memory copy. They're reasonably fast, though; certainly they won't blow a 60fps budget.

RAOF: So, since presentation is (mostly) unimportant, the most sensible approach would be to initialise VA on all available rendernodes and then pick the one with the best format support?

FernetMenta: Who said that presentation would be unimportant? Kodi maps OpenGL textures to vaSurfaces (zero-copy). Please elaborate how this works when having decoding on one gpu and the GL context on the other.

RAOF: As far as I can tell from VAAPI.cpp, Kodi uses eglCreateImageKHR with the EGL_EXT_image_dma_buf_import extension to import the buffers identified by the dma-buf fds as textures. Those dma-buf fds are cross-GPU handles.

The dma-buf fds you get from the vaSurfaces will be references to memory on the decoder-GPU; if the GL context is not on the same GPU, then eglCreateImageKHR will implicitly perform a GPU→GPU migration.

This means that decoding on one GPU and presenting on a different GPU is not zero-copy, but the copy is hidden in the driver and will generally be performed by a hardware DMA engine. As this is only happening at most 3 times per frame, it shouldn't be a particularly noticeable performance impact.

FernetMenta: I wrote this code in order to get zero-copy and I certainly won't break this strategy for something I don't see any benefit in. Introducing a hidden any highly brittle copy operation done by the driver does not make any sense to me. 

> As this is only happening at most 3 times per frame, it shouldn't be a particularly noticeable performance impact

I seriously doubt. Why don't you try with a 4k @ 60 fps movie and post the numbers.

BrandonSchaefer:
Either way it goes, I think its still a good idea to get a fallback for render nodes/DRM here. Mainly if the windowing system (ie. Mir in this case) doesnt support a direct GetVaDisplay. Mir requires some work to allow direct access to its buffers. So DRM will be far better then software rendering.

This will still pick X11 VaDisplay vs render nodes when on X11.

Let me know when you want me to squash the commits.

xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
```diff
@@ -123,14 +132,64 @@ bool CVAAPIContext::EnsureContext(CVAAPIContext **ctx, CDecoder *decoder)
   return true;
 }
 
-bool CVAAPIContext::CreateContext()
+namespace
+{
+VADisplay FindValidDRMVaDisplayFromRenderNode()
```
FernetMenta: no globals please. make this a method of context

xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.h
```diff
@@ -378,29 +377,29 @@ class CVAAPIContext
...
   static CCriticalSection m_section;
-  static Display *m_X11dpy;
   VADisplay m_display;
   int m_refCount;
   int m_attributeCount;
   VADisplayAttribute *m_attributes;
   int m_profileCount;
   VAProfile *m_profiles;
   std::vector<CDecoder*> m_decoders;
+  int m_fd{-1};
```
FernetMenta: something more meaningful then fd would be great.

xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
```diff
@@ -123,14 +132,64 @@ bool CVAAPIContext::EnsureContext(CVAAPIContext **ctx, CDecoder *decoder)
...
 {
-  { CSingleLock lock(g_graphicsContext);
-    if (!m_X11dpy)
-      m_X11dpy = XOpenDisplay(NULL);
```
FernetMenta: keep the separate display connection

xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
BrandonSchaefer: Its in an unnamed namespace, so it'll have internal linkage. It also doesnt depend on anything in the class it self (soo I tend to like pure functions :). But fine with me, changing!

xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.h
BrandonSchaefer: O yeah thats a bad name right now.

xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
BrandonSchaefer: I removed this, because it wasnt referenced anywhere, but I can put it back (not sure if there were plans for it later). Ill add back!

BrandonSchaefer: O sorry, was thinking of something else, you're right!

FernetMenta: It is referenced in current code:

```
  { CSingleLock lock(g_graphicsContext);
    if (!m_X11dpy)
      m_X11dpy = XOpenDisplay(NULL);
  }

  m_display = vaGetDisplay(m_X11dpy);
```

xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
```diff
@@ -123,14 +132,67 @@ bool CVAAPIContext::EnsureContext(CVAAPIContext **ctx, CDecoder *decoder)
...
+  auto win_system = g_Windowing.GetWinSystem();
+  if (win_system == WINDOW_SYSTEM_X11)
+  {
+#if HAVE_X11
+    Display *x11_display{nullptr};
+    { CSingleLock lock(g_graphicsContext);
+      x11_display = XOpenDisplay(nullptr);
```
FernetMenta: this opens a display connection every time we start some video. this is not what we want here. we want open it once for the lifetime of the application.

xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
```diff
@@ -123,14 +132,67 @@ bool CVAAPIContext::EnsureContext(CVAAPIContext **ctx, CDecoder *decoder)
...
+
+    if (display != nullptr)
+    {
+      m_render_node_fd = fd;
+      m_display = display;
+      return;
+    }
+  }
+
+  CLog::Log(LOGERROR, "FAILED To find any open render nodes in /dev/dri/renderD<num>\n");
```
FernetMenta: there is no need for new line character when using Kodi's log

xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.h
```diff
@@ -378,29 +377,30 @@ class CVAAPIContext
...
   static CCriticalSection m_section;
-  static Display *m_X11dpy;
   VADisplay m_display;
   int m_refCount;
   int m_attributeCount;
   VADisplayAttribute *m_attributes;
   int m_profileCount;
   VAProfile *m_profiles;
   std::vector<CDecoder*> m_decoders;
+  int m_render_node_fd{-1};
```
FernetMenta: chamelCasePlease :)

xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
BrandonSchaefer: Good to know, fixing

BrandonSchaefer:
Opps was a static fixing...

FernetMenta:
looks good now, thanks very much!

xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
```diff
@@ -123,14 +133,67 @@ bool CVAAPIContext::EnsureContext(CVAAPIContext **ctx, CDecoder *decoder)
...
+
+    if (display != nullptr)
+    {
+      m_renderNodeFD = fd;
+      m_display = display;
+      return;
+    }
+  }
+
+  CLog::Log(LOGERROR, "FAILED To find any open render nodes in /dev/dri/renderD<num>");
```
hudokkow: Failed to find any...

BrandonSchaefer: Yeah that does look better

xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
```diff
@@ -123,14 +133,67 @@ bool CVAAPIContext::EnsureContext(CVAAPIContext **ctx, CDecoder *decoder)
...
+}
 
-  m_display = vaGetDisplay(m_X11dpy);
+bool CVAAPIContext::CreateContext()
+{
+  SetVaDisplayForSystem();
+
+  if (m_display == nullptr)
+  {
+    CLog::Log(LOGERROR, "FAILED To find any a VaDisplay for this system\n");
```
BrandonSchaefer: Opps and forgot about the \n ... annd fixing

hudokkow: Sorry for nitpicking. It's either "any VaDisplay" or "a VaDisplay". But English is not my native language...

BrandonSchaefer: Yeah you're right, english is my native language and im not super good at it :)

FernetMenta:
jenkins build this please

candrews:
With this change, I'm getting a build failure:
```
make[1]: Leaving directory '/var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDSubtitles'
x86_64-pc-linux-gnu-g++ -DNDEBUG=1 -O2 -march=native -pipe -fomit-frame-pointer -fno-lto -fPIC -DPIC -D_REENTRANT -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -DNDEBUG=1 -O2 -march=native -pipe -fomit-frame-pointer -fno-lto -fPIC -DPIC -D_REENTRANT -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -Wl,-O1 -Wl,--as-needed -flto -fno-lto -Wl,-O1 -Wl,--as-needed -flto -fno-lto -o kodi.bin xbmc/platform/posix/posix.a -Wl,--start-group xbmc/platform/posix/posix.a xbmc/linux/linux.a xbmc/network/network.a xbmc/video/windows/videowindows.a xbmc/utils/utils.a xbmc/cores/DllLoader/exports/util/exports_utils.a xbmc/cores/DllLoader/exports/exports.a xbmc/settings/settings.a xbmc/video/video.a xbmc/pvr/addons/pvraddons.a xbmc/pvr/windows/pvrwindows.a xbmc/guilib/guilib.a  xbmc/cores/VideoPlayer/VideoPlayer.a xbmc/cores/VideoPlayer/DVDCodecs/DVDCodecs.a xbmc/cores/VideoPlayer/DVDCodecs/Audio/Audio.a xbmc/cores/VideoPlayer/DVDCodecs/Overlay/Overlay.a xbmc/cores/VideoPlayer/DVDCodecs/Video/Video.a xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxers.a xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreams.a xbmc/cores/VideoPlayer/DVDSubtitles/DVDSubtitles.a xbmc/cores/VideoPlayer/Process/Process.a xbmc/addons/addons.a xbmc/addons/binary/interfaces/addon-interfaces.a xbmc/addons/binary/interfaces/api1/Addon/addon-callbacks-addon.a xbmc/addons/binary/interfaces/api1/AudioDSP/addon-callbacks-audiodsp.a xbmc/addons/binary/interfaces/api1/AudioEngine/addon-callbacks-audioengine.a xbmc/addons/binary/interfaces/api1/Codec/addon-callbacks-codec.a xbmc/addons/binary/interfaces/api1/GUI/addon-callbacks-gui.a xbmc/addons/binary/interfaces/api1/InputStream/addon-callbacks-inputstream.a xbmc/addons/binary/interfaces/api1/Peripheral/addon-callbacks-peripheral.a xbmc/addons/binary/interfaces/api1/PVR/addon-callbacks-pvr.a xbmc/contrib/kissfft/kissfft.a xbmc/cores/AudioEngine/audioengine.a xbmc/cores/DllLoader/dllloader.a xbmc/cores/ExternalPlayer/ExternalPlayer.a xbmc/cores/VideoPlayer/VideoRenderers/VideoRenderer.a xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoShaders.a xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/HwDecRender.a xbmc/cores/cores.a xbmc/cores/paplayer/paplayer.a xbmc/cores/playercorefactory/playercorefactory.a xbmc/dbwrappers/dbwrappers.a xbmc/dialogs/dialogs.a xbmc/epg/epg.a xbmc/events/events.a xbmc/filesystem/MusicDatabaseDirectory/musicdatabasedirectory.a xbmc/filesystem/VideoDatabaseDirectory/videodatabasedirectory.a xbmc/filesystem/filesystem.a xbmc/games/controllers/controllers.a xbmc/games/controllers/dialogs/controllerdialogs.a xbmc/games/controllers/guicontrols/controllerguicontrols.a xbmc/games/controllers/windows/controllerwindows.a xbmc/input/input.a xbmc/interfaces/builtins/builtins.a xbmc/interfaces/generic/interfaces-generic.a xbmc/interfaces/info/info.a xbmc/interfaces/interfaces.a xbmc/interfaces/json-rpc/json-rpc.a xbmc/interfaces/legacy/legacy.a xbmc/interfaces/python/python_binding.a xbmc/listproviders/listproviders.a xbmc/platform/platform.a xbmc/media/media.a xbmc/messaging/messaging.a xbmc/messaging/helpers/messagingHelpers.a xbmc/music/dialogs/musicdialogs.a xbmc/music/infoscanner/musicscanner.a xbmc/music/music.a xbmc/music/tags/musictags.a xbmc/music/windows/musicwindows.a xbmc/network/dacp/dacp.a xbmc/network/websocket/websocket.a xbmc/peripherals/addons/peripheral-addons.a xbmc/peripherals/bus/peripheral-bus.a xbmc/peripherals/devices/peripheral-devices.a xbmc/peripherals/dialogs/peripheral-dialogs.a xbmc/peripherals/peripherals.a xbmc/pictures/pictures.a xbmc/playlists/playlists.a xbmc/powermanagement/powermanagement.a xbmc/profiles/profiles.a xbmc/profiles/dialogs/profiles_dialogs.a xbmc/profiles/windows/profiles_windows.a xbmc/programs/programs.a xbmc/pvr/channels/pvrchannels.a xbmc/pvr/dialogs/pvrdialogs.a xbmc/pvr/pvr.a xbmc/pvr/recordings/pvrrecordings.a xbmc/pvr/timers/pvrtimers.a xbmc/rendering/rendering.a xbmc/settings/dialogs/settings_dialogs.a xbmc/settings/lib/settings_lib.a xbmc/settings/windows/settings_windows.a xbmc/storage/storage.a xbmc/video/dialogs/videodialogs.a xbmc/video/jobs/video-jobs.a xbmc/video/videosync/videosync.a xbmc/view/view.a xbmc/windowing/windowing.a xbmc/windows/windows.a xbmc/xbmc.a xbmc/cdrip/cdrip.a xbmc/interfaces/legacy/wsgi/legacy-wsgi.a xbmc/network/httprequesthandler/httprequesthandlers.a xbmc/network/httprequesthandler/python/httprequesthandlers-python.a xbmc/rendering/gl/rendering_gl.a lib/libUPnP/libupnp.a xbmc/network/upnp/upnp.a xbmc/input/joysticks/input_joysticks.a xbmc/input/joysticks/dialogs/input_joystick_dialogs.a xbmc/input/joysticks/generic/input_joysticks_generic.a xbmc/input/linux/input_linux.a xbmc/input/touch/input_touch.a xbmc/input/touch/generic/input_touch_generic.a xbmc/network/linux/network_linux.a xbmc/powermanagement/linux/powermanagement_linux.a xbmc/storage/linux/storage_linux.a xbmc/windowing/X11/windowing_X11.a lib/UnrarXLib/UnrarXLib.a -Wl,--end-group xbmc/threads/threads.a xbmc/commons/commons.a -lavahi-client -lavahi-common -lmicrohttpd -lresolv -ltinyxml -lssl -lcrypto -lz -llzo2 -lpthread -lbz2 -lgcrypt -lgpg-error -lGLU -lGL  -L/usr/lib64 -lpython2.7 -lssl -lcrypto -Wl,-O1 -Wl,--as-needed -flto -fno-lto -lcrossguid -luuid -lyajl -lxml2 -lxslt -lxml2 -lz -llzma -licui18n -licuuc -licudata -lm -ldl -lm -lxml2 -lfribidi -lglib-2.0 -lsqlite3 -lpcrecpp -lpcre -lfreetype -ltag -lcdio -lm -lasound -ldbus-1 -lasound -lpulse -lX11 -lXext -lXrandr -ldrm -lEGL -lssh -lsmbclient -ludev -llcms2 -lbluetooth -lcap -lavcodec -lavfilter -lavformat -lavutil -lpostproc -lswscale -lswresample -lva-x11 -lva -rdynamic
/usr/lib/gcc/x86_64-pc-linux-gnu/6.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: xbmc/cores/VideoPlayer/DVDCodecs/Video/Video.a(VAAPI.o): undefined reference to symbol 'vaGetDisplayDRM'
/usr/lib64/libva-drm.so.1: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make: *** [Makefile:449: kodi.bin] Error 1
```

It looks like "-lva-drm" is missing.

FernetMenta:
yes, seems jenkins does not build vaapi then

FernetMenta:
@BrandonSchaefer  could you submit a fix for this?

BrandonSchaefer:
Opps yeah I didnt add finding libva-drm to the FindVaapi.cmake. Ill get something up tomorrow (a bit late for me)

BrandonSchaefer:
Though whats strange, it builds fine for me as well as here:
https://api.travis-ci.org/jobs/177342760/log.txt?deansi=true

How did you compile this? I should be adding libva-drm to the FindVAAPI.cmake but just curious how I cant get it to fail nor does it fail when ran through travis.

candrews:
I'm using auto tools/make, not cmake, not sure if that matters.

BrandonSchaefer:
Autotools are planned to be removed soon, annnd I didnt test it out there :). You can try cmake in kodi/project/cmake which looks to be working.

https://github.com/xbmc/xbmc/pull/10429

FernetMenta:
> I'm using auto tools/make, not cmake, not sure if that matters.

if you are using autotools, you are on your own.

if it builds with cmake, all fine.

PR #10923, opened 2016-11-13
Fix unable to open controller dialog if all controllers are unplugged

garbear:
While testing, I've become aware of something: there are endless quirks in real life hardware, which means that inevitably the button mapper will get messed up.

Sometimes, things can get so bad that simply connecting the controller causes erratic behavior. In this case, the user's only hope is to unplug the controller, open the controller dialog, start the mapping (or ignoring) process, and then plug in the controller to capture the erratic input. However, if all controllers are disconnected, the user can't open the configuration dialog.

This change removes the constraint that controllers must be connected to open the controller dialog. Also, as there's no check done for add-ons anymore, we ensure that add-ons are enabled so that button mapping is possible when the dialog opens.

## Motivation and Context
Better UX

unable to configure controllers

## Types of change

Broken out from #10630

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -16056,7 +16056,19 @@ msgctxt "#35016"
 msgid "Axis %d"
 msgstr ""
 
-#empty strings from id 35017 to 35048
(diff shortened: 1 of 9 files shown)


garbear:
jenkins build this please

PR #10924, opened 2016-11-13
paplayer: fix replay gain

FernetMenta:
@DaveTBlake according to discussion on slack


DaveTBlake:
Thanks @FernetMenta. Just for the record here, rather than burried in Slack, the album or track replaygain given in music file tags was being ingored unless there was also a peak tag value as well. By definition if no peak value is specifed then a default of 1.0 should be used.

Peak clipping was also being incorrectly applied.  ActiveAE has clipping protection when stream amplification is applied, so we should apply replayGain as replayGain if m_peak equals 1.0 and as volume amplification for the other cases.

This makes the "aviod clipping" `replayGainSettings.bAvoidClipping` setting useless. Should it therefore be removed as part of this PR or a separate PR?

FernetMenta:
> Should it therefore be removed as part of this PR or a separate PR?

right, ideally it should go with this PR. would you have time to remove and PR to this PR?

razzeee (approved):

DaveTBlake:
> right, ideally it should go with this PR. would you have time to remove and PR to this PR?

Happy to do the settings change, but not sure how to PR a PR? Merge this and I'll raise a separate PR to remove the unused setting.

FernetMenta:
@DaveTBlake just PR to my repo branch rgain

DaveTBlake:
Well @FernetMenta  I seem to have PR to your repo, but brought more commits from master with me and it is not shown here. Probably a cock up on my part.

FernetMenta:
@DaveTBlake thanks, I cherry picked your change

jenkins build this please

FernetMenta:
@DaveTBlake @Razzeee merge when you think it's ready

PR #10925, opened 2016-11-13
Correction issue on Android TV when intent usage

deblockt:
Correct issue when two intents are run one after the other

## Description
The function onNewIntent don't set the path of fileitem. So, When the second intent is run, kodi try to play video but it have not the path.

## Motivation and Context
I'm writting an android Tv plugin to provide recommendation. 
To run kodi, i use intent. 
For the first played video, it's work fine. But the second video just open kodi.

## How Has This Been Tested?
My change is test on android TV (shield TV). 
Now, when I play my second video, kodi start fine.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/platform/android/activity/XBMCApp.cpp
+++ b/xbmc/platform/android/activity/XBMCApp.cpp
@@ -858,10 +858,8 @@ void CXBMCApp::onNewIntent(CJNIIntent intent)
   std::string action = intent.getAction();
   if (action == "android.intent.action.VIEW")
   {
-    CFileItem* fileitem = new CFileItem(GetFilenameFromIntent(intent));
-    fileitem->SetPath(GetFilenameFromIntent(intent));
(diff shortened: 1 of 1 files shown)


koying:
Good catch.
```
CApplicationMessenger::GetInstance().SendMsg(TMSG_MEDIA_PLAY, 1, 0, static_cast<void*>( new CFileItem(GetFilenameFromIntent(intent), false)));
```
should work also (uses the CURL(std::string) ctor)) and be more efficient

Razzeee:
I'm not sure why the first try works, but fix seems to be valid.

deblockt:
The first call work because OnNewIntent is not used. 
The first call use CXBMCApp::run and run XBMC like with command line.

I have change my correction using  koying's code.

Razzeee:
jenkins build this please

PR #10926, opened 2016-11-14
[guilib] modify container.refresh to reset any container

BigNoid:
## Description
```Container.Refresh```  and ```Container.Update``` are basically the same. They send the same message, only ```Container.Refresh``` resets the history by default. This merges those two message, so that if you send ```Container.Update``` without parameters it does the same as previously ```Container.Refresh```. Add the ```replace``` parameter to get the same results as previously ```Container.Refresh(path)```.
That only works in media windows and for the main container.
This PR adds a new message that resets the dynamic content container so it will be refreshed.
```Container.Refresh(ID)```
## Motivation and Context
EDIT:
See #10931 for solution for stalled dynamic content lists.

## Types of change

- [] New feature (non-breaking change which adds functionality)

## Checklist:

Code is based on https://github.com/Black09/xbmc/commit/3e31795addaaa3bc37e184ba68db887f7db6c013

@phil65 @ronie @HitcherUK

### DIFF
--- a/xbmc/guilib/GUIBaseContainer.cpp
+++ b/xbmc/guilib/GUIBaseContainer.cpp
@@ -473,6 +473,12 @@ bool CGUIBaseContainer::OnMessage(CGUIMessage& message)
       }
       return true;
     }
+    else if (message.GetMessage() == GUI_MSG_REFRESH_CONTAINER)
+    {
+      Reset();
+      m_listProvider->Reset(true);
+      UpdateListProvider(true);
+    }
   }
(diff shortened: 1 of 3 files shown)


tamland:
We should not allow any api to force update containers like this. Dynamic list are automatic, and therefore never needs to be reset.

BigNoid:
I am fine with closing this, it was born out of a need to reset the stalled containers. But since that is now properly fixed I can't think of a use case anymore.

PR #10370, opened 2016-08-30
Add settings to control notification rumble duration and strength

garbear:
I've added some fun settings that let you control the notification rumble. A forum user requested a "classier" rumble feel, so this lets people experiment with different rumble behaviours.

It might not be desirable to include these extra settings in the final build. However, I would like feedback from several people before hard-coding these values. Maybe @MilhouseVH can put out a couple builds with this patch?

rumble settings

PR #10929, opened 2016-11-14
Set notification rumble to weak motor for 300ms

garbear:
The general consensus seems to be that a weak motor rumble for 300ms is best for notification because this gives a more "classy" feel than full-strength rumble of all motors for 1000ms.

Supersedes #10370

Broken out from #10630


garbear:
jenkins build this please

PR #10438, opened 2016-09-08
[Estuary][PVR] PVR home screen widgets: Misc fixes and improvements

ksooo:
- Add "recent recordings" widget
  screenshot006
- Fix TV "current/upcoming recordings" widgets to display only TV recordings
- Add Radio "current/upcoming recordings" widgets
- Fix "recently played channels" widget to not display channels that were never played
- Fix "resume" not working when activated using "recent recordings" widget
- Fix URL for "recently played channels" widgets (do not use localized URLs in the skin!) 

@da-anda fyi
@Jalle19 mind taking a look at the code
@BigNoid @ronie are the skin changes okay?

PR #10930, opened 2016-11-14
[PVR] Fix recordings url encoding.

ksooo:
Regression introduced by #10438 => 5ba26b9c21c665310d3ce0424958069f29f26b99

If a recording folder name contained "?", the children of this folder were never displayed.

This PR fixes this bug.

Changes were runtime tested on macOS, latest kodi master.

screenshot000
screenshot001

@Jalle19 for code review?
@fetzerch fyi, maybe you can runtime test.

### DIFF
--- a/xbmc/pvr/recordings/PVRRecordingsPath.cpp
+++ b/xbmc/pvr/recordings/PVRRecordingsPath.cpp
@@ -22,6 +22,7 @@
 #include "utils/RegExp.h"
 #include "utils/StringUtils.h"
 #include "utils/URIUtils.h"
+#include "URL.h"
 
 #include "PVRRecordingsPath.h"
 
@@ -98,7 +99,7 @@ CPVRRecordingsPath::CPVRRecordingsPath(bool bDeleted, bool bRadio,
     strDirectoryN = StringUtils::Format("%s/", strDirectoryN.c_str());
 
   std::string strTitleN(strTitle);
-  StringUtils::Replace(strTitleN, '/', ' ');
+  strTitleN = CURL::Encode(strTitleN);
 
   std::string strSeasonEpisodeN;
   if ((iSeason > -1 && iEpisode > -1 && (iSeason > 0 || iEpisode > 0)))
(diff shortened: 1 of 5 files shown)


xbmc/pvr/recordings/PVRRecordingsPath.cpp
```diff
@@ -230,3 +239,17 @@ size_t CPVRRecordingsPath::GetDirectoryPathPosition() const
   }
   // unreachable
 }
+
+void CPVRRecordingsPath::EscapeSegment(std::string &strSegment)
```
xhaggi: Any reason why you don't use `CURL::Decode` and `CURL::Encode`?

fetzerch:
Thanks for the quick fix. Runtime tested here with mythtv, all good :)

xbmc/pvr/recordings/PVRRecordingsPath.cpp
ksooo: well, i was wrong, i was misinterpreting what it does when i was first reading the code. will change, thanks for the hint.

ksooo: done

ksooo:
second commit fixes estuary recordings window sidebar for folders with special characters in it's name.

before:

screenshot000

after:

screenshot002

@phil65 fyi

ksooo:
jenkins build this please

xbmc/pvr/recordings/PVRRecordingsPath.h
```diff
@@ -52,8 +52,8 @@ namespace PVR
     bool IsDeleted() const { return !IsActive(); }
     bool IsRadio() const { return m_bRadio; }
     bool IsTV() const { return !IsRadio(); }
-    std::string GetDirectoryPath() const { return m_directoryPath; }
-    std::string GetSubDirectoryPath(const std::string &strPath) const;
+    std::string GetUnescapedDirectoryPath() const;
+    std::string GetUnescapedSubDirectoryPath(const std::string &strPath) const;
```
ksooo: to make it pretty clear that the returned values are unescaped strings.

PR #10931, opened 2016-11-14
[listprovider] Make sure content is always loaded. Fixes #16635

### LINKED ISSUE #16635: [Feature request] Add autoclose and oklabel parameters to xbmcgui.Dialog().ok
## Roadmap or todo item
**Detailed Description**

Can you please add the autoclose parameter to the xbmcgui.Dialog().ok dialog, in a similar fashion to xbmcgui.Dialog().yesno() Also, can a parameter oklabel also added to the xbmcgui.Dialog().ok dialog? It will be very nice to be able to change the label of the OK dialog.

Is there any chance this can be backported for Leia 18.5?

### Context

In [Advanced Emulator Launcher](https://forum.kodi.tv/showthread.php?tid=287826) some operations take a long time to complete, for example scraping a launcher with many items may take hours to complete. In order to realise unattended scraping (users left the scraper running overnight), if an error happens (network down, etc.) AEL displays the error with a yesno() dialog which autocloses in one minute and then the scraping continues. However, the proper way to display the error must be an OK dialog, just with one button.

I'm sure other plugin/addon authors will find this change useful for their addons.

### Possible Implementation

No idea, sorry. I am not an expert of the C++ core.

### Additional context, screenshots or links

No screenshots needed, description is clear.

*Note: Team Kodi is considering this item, but we are not making any promises if and when it will be included.*

BigNoid:
## Description
Set begin state to ```INVALIDATED``` so that dynamic content always gets fetched when the list is loaded, even if content attributes haven't changed. 

## Motivation and Context
See http://trac.kodi.tv/ticket/16635 for full details.
One side effect of previous behaviour was that a list that had not changed it seemed to load very fast on window load. But in fact that was old content, so could potentially be out of date. Behaviour in this PR is the correct way imo.
## How Has This Been Tested?
Runtime tested

## Types of change

## Checklist:

@phil65 @ronie @HitcherUK

### DIFF
--- a/xbmc/listproviders/DirectoryProvider.cpp
+++ b/xbmc/listproviders/DirectoryProvider.cpp
@@ -322,6 +322,8 @@ void CDirectoryProvider::Reset(bool immediately /* = false */)
   if (m_jobID)
     CJobManager::GetInstance().CancelJob(m_jobID);
   m_jobID = 0;
+  // Make sure job is always fired on start
+  m_updateState = INVALIDATED;
   // reset only if this is going to be destructed
   if (immediately)
   {


BigNoid:
jenkins build and merge please

hudokkow:
jenkins is really temperamental lately. Didn't even started jobs, except for win that failed, again.
jenkins build and merge

BigNoid:
jenkins failure is unrelated

tamland:
Can you please explain why this is correct? Setting it to invalidated is the *opposite* of what is currently done in both initializer and in 'immediate' reset.

BigNoid:
The job needs to be fired when the list is loaded, regardless if sort, limit and url have changed. ```CDirectoryProvider::Reset()``` is called when the list becomes visible, so setting it invalid there made the most sense to me.

mazeyx:
this is causing a strange issue in estuary, if you have videos show up last on list and you go settings then exit out and scroll up it shows "Playlists" under Video > media sources (usually blank under the media source icons) for a second or two then vanishes.

BigNoid:
Yeah that is expected. Before that only happened on first window load (and after some specific calls like updated favourites/playback stopped/database update) and the list would never be loaded again. With this change the list will be updated each time the list becomes visible the first time a window is opened. Skins need to check if the list is updating or that is has more than 0 items. So it's visible for a short time while its updating.

mazeyx:
@BigNoid ok thanks ill get used to it. sounds like this might also fix when you do a first install and start scanning to library the home page has no files on it until you restart kodi then they show up.

tamland:
@BigNoid Ok, I've had a look at the directory provider source again and I'm not sure if this is the right approach. Currently, the whole thing is  base on being initialised to an empty state; and then, only when ::Update is called changes are detected. The 'invalidated' state was only intended to be used when *external* changed to the content occurs; that the directory provider cannot know about.

So instead of just throwing in an invalidation, we should find out an understand why this window issue  is happening

BigNoid:
@tamland 
Imo the problem is that no matter if ```OnJobComplete``` is called the job is considered done and unless sort, limit or url changes ::Update is never called again.
I tried implementing ```OnJobProgress``` but I couldn't get that to work.

BigNoid:
@tamland 
I've did a few debug sessions and ```Reset()``` gets called on window exit. To fix the problem described on trac, but keep functionality as it was before I've now did this: https://github.com/BigNoid/xbmc/commit/fa5f6a03309716255048612af1a7d005b510e071
Would that be the right approach? So it only gets invalidated if ```OnJobComplete``` wan't called.

PR #10932, opened 2016-11-15
Controller dialog fixes and code improvements

garbear:
This PR fixes two bugs in the controller dialog, one is a bug that was introduced by #10910, and one is a preexisting bug that was uncovered by #10910.

Also included are logging and code style improvements.

## Types of change

Broken out from #10630


garbear:
Updated with an additional fix for a bug in #10630: The right analog stick would lowers the volume immediately after mapping is finished in the controller dialog.

a1rwulf:
I definitely don't know a lot about this code, but it looks good to me.

garbear:
Added fix for Coverty warning.

jenkins build this please

PR #10937, opened 2016-11-15
[estuary] fix getting stuck on rss control

ronie:
fixes: http://forum.kodi.tv/showthread.php?tid=295897

the rss control is a focusable control, define up/down/left/right navigation.
(cosmetics: remove unused pagecontrol from home menu)

@phil65

### DIFF
--- a/addons/skin.estuary/1080i/Home.xml
+++ b/addons/skin.estuary/1080i/Home.xml
@@ -993,7 +993,6 @@
 						<top>1</top>
 						<width>382</width>
 						<height>671</height>
-						<pagecontrol>70</pagecontrol>
 						<onfocus>ClearProperty(listposition,home)</onfocus>
 						<onright condition="String.IsEqual(Container(9000).ListItem.Property(id),movies) + Control.IsVisible(5000)">5000</onright>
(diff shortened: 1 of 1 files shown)


phil65:
perhaps better (or in addition) add a hitrect tag with width and height = 0?

ronie:
but then you can't focus the rss control anymore ;-)
the purpose of focusing the rss control is that the rss text will stop scrolling.

phil65:
Do you think that this "purpose" actually makes sense? ;) 
The RSS control has no visual idication that it is focused, so it shouldnt get focused at all imo.

ronie:
i have no opinion if it makes sense or not :-)
obviously the dev who implemented it did think it has purpose, nor did anyone object:
https://github.com/xbmc/xbmc/pull/1709

the 'visual indication' is that the text stops scrolling.

phil65:
I´d prefer a hitrect to make it non-focusable.
the bug this tries to fix is that the startup focus position isnt borked when havin mouse activated. Using a hitrect, it should be solved properly. Using onright/onleft etc only workarounds that by makin sure that the focus can get shifted away again.

> the 'visual indication' is that the text stops scrolling.

sometimes (for example at startup) the text is not visible yet. In those cases there is visual indication at all.

ronie:
closing... issue addressed in estuary 2nd edition

PR #10938, opened 2016-11-15
[estuary] fix missing texture log error

ronie:
use info,prefix,postfix formatting for images to get rid of missing texture errors in the log when info returns empty.

```11:21:02.549 T:139697911798080   ERROR: CGUITextureManager::GetTexturePath: could not find texture 'flags/starrating/.png'```

@phil65


phil65:
+1

PR #10939, opened 2016-11-15
[estuary] fix lost focus in mediasource dialog

ronie:
fixes: http://forum.kodi.tv/showthread.php?tid=262373&pid=2442043#pid2442043

without this kodi remembers the last focused position in this dialog, which would usually be the 'ok button'
but this button is disabled (and can't be focused) if you open the dialog again to add another source.

@phil65


phil65:
makes sense, +1

PR #10945, opened 2016-11-16
MMALCodec: Take more care with the output lock mutex

popcornmix:
## Description
There were reports of "GetPicture called but m_output_ready is empty"
error in log.

It's been hard to reproduce but the only way that seems possible is for
m_output_ready.size to produce an inconsistent result between calls
to CMMALVideo::Decode and CMMALVideo::GetPicture.

Looking closely we don't always have the m_output_mutex lock when
testing the size, so do it more carefully.

## Motivation and Context
Hopefully fix "GetPicture called but m_output_ready is empty" error in log and video stall.

## How Has This Been Tested?
Been in Milhouse builds for last new nights.

## Types of change

## Checklist:


popcornmix:
jenkins build this please

hudokkow:
jenkins build this please

PR #10946, opened 2016-11-16
[cmake/win32] fixes for building with abnormal configuration

Kwiboo:
## Description

Building fails (on windows) when version tag is empty, app name is changed or when optical/upnp/airtunes is disabled

## Motivation and Context

Fixes building of abnormal configuration on windows

## How Has This Been Tested?

Compile tested with cmake 3.6.3 and visual studio 2015 on windows 10

## Types of change

## Checklist:


hudokkow:
Thanks for looking into this. CMake changes look sane. As for the rest, we need an win dev.

ping @Paxxi 
jenkins build this please

Razzeee:
jenkins build failed last time with this PR.
Let's do it again and see why.
jenkins build this please

Kwiboo:
If it helps any review of this PR:
`HAS_AIRTUNES` and `HAS_UPNP` removed from `xbmc/system.h` is defined earlier in the same file depending on cmake config and should not cause any changes in defines with normal cmake config.
`HAS_AUDIO` was removed because I could not find it used anywhere inside Kodi: https://github.com/xbmc/xbmc/search?q=HAS_AUDIO

hudokkow:
Anyone has a reason not to merge this? If so, speak in the next 24h or I'll push it in.
@afedchin, do you have time for a quick look?

afedchin:
what should I take a look on? I know nothing about cmake.

hudokkow:
The ifdeferry? It's not a CMake invention... ;)

xbmc/storage/windows/Win32StorageProvider.cpp
```diff
@@ -96,6 +98,7 @@ CDetectDisc::CDetectDisc(const std::string &strPath, const bool bautorun)
 bool CDetectDisc::DoWork()
 {
   CLog::Log(LOGDEBUG, "%s: Optical media found in drive %s", __FUNCTION__, m_strPath.c_str());
+#ifdef HAS_DVD_DRIVE
```
afedchin: this ifdef isn't needed because the ifdef above avoid running this code. if you want to keep this for safety, I think the log record above shouldn't be occurs also.

Kwiboo: I'll update to ifdef away the complete `CDetectDisc` class.

`g_mediaManager.GetDiskLabel` and `g_mediaManager.GetDiskUniqueId` are the methods missing when `HAS_DVD_DRIVE` is undefined.

afedchin: got it, but please move this up to avoid logging also.

Kwiboo:
Rebased and a moved `ifdef` in `CDetectDisc::DoWork()`

Completely removing `CDetectDisc` was a rather obtrusive change as it also required ifdef in https://github.com/xbmc/xbmc/blob/master/xbmc/windowing/windows/WinEventsWin32.cpp#L790 so I just moved the ifdef in `CDetectDisc::DoWork()`

MartijnKaijser:
jenkins build this please
any further comments/objections?

Paxxi:
Looks good to me

hudokkow:
jenkins build and merge

PR #10947, opened 2016-11-16
[curl] cleanup multi handle in proper order

Kwiboo:
## Description

Changes cleaning up order of curl multi/easy handles

## Motivation and Context

From https://curl.haxx.se/libcurl/c/curl_multi_cleanup.html
1 - curl_multi_remove_handle before any easy handles are cleaned up
2 - curl_easy_cleanup can now be called independently since the easy handle is no longer connected to the multi handle
3 - curl_multi_cleanup should be called when all easy handles are removed

## Types of change

## Checklist:


hudokkow:
Thanks for your contribution.
Not really sure who to ping. @Paxxi, @Montellese, @notspiff

wsnipex:
it looks correct, but since this a pretty critical part of the code, this is something for v18

Paxxi:
Thank you for spotting this. This is a fix and I think it should go in right away since we currently do it wrong and probably in undefined behavior territory

arnova:
Yup, but better shove it in ASAP then so we can spot any regressions way before we go final.

PR #10948, opened 2016-11-16
[cmake] Crossguid offline url override

BrandonSchaefer:
Crossguid offline url override

## Description
When attempting to build offline crossguid will still attempt to download even when you override the URL. It also does not generate the full path for you. Updated it so it will override the URL the same way as ffmpeg and libdvd.

I could also be assuming something here but trying to use it like ffmpeg/libdvd. If so just reject this PR!

## Motivation and Context
Would like to override the URL for offline building.

## How Has This Been Tested?
Test by -DCROSSGUID_URL=foo.tar.bz2 then make crossguid and see that it still downloads the tar ball. With the change, it wont download anymore! (Though it'll fail unless you have a foo.tar.bz2 as the real crossguild tar ball!)

## Types of change

## Checklist:

### DIFF
--- a/project/cmake/modules/FindCrossGUID.cmake
+++ b/project/cmake/modules/FindCrossGUID.cmake
@@ -7,8 +7,13 @@ if(ENABLE_INTERNAL_CROSSGUID)
 
   # allow user to override the download URL with a local tarball
   # needed for offline build envs
-  if(NOT EXISTS ${CROSSGUID_URL})
+  if(CROSSGUID_URL)
+    get_filename_component(CROSSGUID_URL "${CROSSGUID_URL}" ABSOLUTE)
+  else()
     set(CROSSGUID_URL http://mirrors.kodi.tv/build-deps/sources/crossguid-${CGUID_VER}.tar.gz)
+  endif()
+  if(VERBOSE)
+    message(STATUS "CROSSGUID_URL: ${CROSSGUID_URL}")
   endif()
 
   set(CROSSGUID_LIBRARY ${CMAKE_BINARY_DIR}/${CORE_BUILD_DIR}/lib/libcrossguid.a)


BrandonSchaefer:
More or less the issue is, if you provide a non full path it'll fail and download. Since ffmpeg and libdvd generate the fullpath for you it would be more consistent to do it how the others are doing it.

hudokkow:
Sorry, it's almost AM here and I'm not sure I understand the problem. What do you mean by a non full path? Are you only passing the filename?

> Though it'll fail unless you have a foo.tar.bz2 as the real crossguild tar ball!

Again, not fully understanding but you need to check if files exists:

> if(CROSSGUID_URL AND EXISTS CROSSGUID_URL)
[...]

Thanks for taking the time to look into this. CMake is very much a work in progress and all input is welcome. Ideally we should cache downloaded files and look in there first. No there yet. :wink:

BrandonSchaefer:
Non-full path meaning a non absolute path (relative path). So where ever your root CMakeLists.txt file is

For example:
-DCROSSGUILD_URL=downloads/crossguid-tar-ball.tar.bz2 -DFFMPEG_URL=downloads/ffmpeg-tar-ball.tar.bz2

Crossguild url fails, and ffmpeg url works.

Could be some cached download which for the absolute path would be:
/path/to/kodi/project/cmake/downloads/crossguid-tar-ball.tar.bz2

This issue is:
if (NOT EXISTS <path>) requires the absolute path, so giving it:
-DCROSSGUILD_URL=downloads/crossguid-tar-ball.tar.bz2

Will fail since that file does not exists, which means ill need to know the full absolute path from the command line. So it would be better if we did it like ffmpeg and libdvd (in FindFFMPEG.cmake) you give a relative path, which then cmake builds the absolute path with:
    get_filename_component(FFMPEG_URL "${FFMPEG_URL}" ABSOLUTE)

Which is what Ive done here. So now you can use a relative path based on the root CMakeLists.txt vs *knowing* an absolute path from the command line

BrandonSchaefer:
Also why do we need to check if the file exists to fall back? The override is usually manually done, soo if they give an incorrect path, error out. That seems fair to me (otherwise it might be confusing if they are not paying attention and then use the downloaded version vs the one they are attempting to override with)

hudokkow:
Now I understand! Like I said, almost (very) AM. Sorry for wasting your time.

jenkins build this please

BrandonSchaefer:
O no worries i didnt explain it very well to begin with :)

wsnipex:
this is ok, thanks.

project/cmake/modules/FindCrossGUID.cmake
```diff
@@ -7,7 +7,9 @@ if(ENABLE_INTERNAL_CROSSGUID)
 
   # allow user to override the download URL with a local tarball
   # needed for offline build envs
-  if(NOT EXISTS ${CROSSGUID_URL})
+  if(CROSSGUID_URL)
+    get_filename_component(CROSSGUID_URL "${CROSSGUID_URL}" ABSOLUTE)
```
wsnipex: please add:
      if(VERBOSE)
        message(STATUS "CROSSGUID_URL: ${CROSSGUID_URL}")
      endif()

BrandonSchaefer:
Done and squashed

PR #10949, opened 2016-11-16
[listprovider] whitelist mediatypes which may get opened in VideoInfo…

phil65:
only allow the videoinfodialog for specific mediatypes (via whitelisting)
Fixes issues when using genres /tags VFS path etc on home screen, esp. when "Default select action" is set to "Show Information".
@Montellese @tamland

### DIFF
--- a/xbmc/listproviders/DirectoryProvider.cpp
+++ b/xbmc/listproviders/DirectoryProvider.cpp
@@ -402,8 +402,16 @@ bool CDirectoryProvider::OnInfo(const CGUIListItemPtr& item)
   }
   else if (fileItem->HasVideoInfoTag())
   {
-    CGUIDialogVideoInfo::ShowFor(*fileItem.get());
-    return true;
+    auto mediaType = fileItem->GetVideoInfoTag()->m_type;
+    if (mediaType == MediaTypeMovie ||
+        mediaType == MediaTypeTvShow ||
+        mediaType == MediaTypeEpisode ||
+        mediaType == MediaTypeVideo ||
+        mediaType == MediaTypeMusicVideo)
+    {
+      CGUIDialogVideoInfo::ShowFor(*fileItem.get());
+      return true;
+    }
   }
   else if (fileItem->HasMusicInfoTag())
   {


Montellese:
This approach does work for cases like "genre" etc. but it still does not work for a lot of other cases like plugin items or UPnP items which also might have a media type set to movies, episodes or whatever and we must not show the info dialog for these items either. For those cases the only solution is to check if the item is part of the local library (by checking if it has a scraper).

phil65:
So I guess this approach is not fine enough then? Could you point me to the logic needed here? Could give it a try then. Thx!

tamland:
Why would non-video items have video info tag, and why shouldn't videos from upnp have info dialog? (this makes no sense to me..)

tamland:
plugin items definitely should have info dialog though; as it always have..

phil65:
@tamland I agree that it doesnt really make sense that they have a VideoInfoTag. They get one assigned in CVideoDatabase::GetNavCommon() (m_iDbId and m_type are set there)
I think to fix that properly m_iDbId and m_type would have to be members of the listitem itself, not of the VideoInfoTag  / MusicInfoTag? Those changes are out of scope for me unfortunaley though, too many other tasks to work on :)

Montellese:
I'll look up the location where the logic for opening the video info dialog is located when I'm home.

@tamland: Concerning `CVideoInfoTag` @phil65 has already explained this. Concerning showing the video info dialog for e.g. UPnP items the problem is that the info dialog allows you to refresh the item or change the artwork etc. of these items which does not make sense for UPnP or plugin items. A while ago you even got a crash when opening the info dialog for an item without a scraper. I noticed that when I tried the same for imported items in my media import work.

The other solution would be to make `CGUIDialogVideoInfo` much smarter about what items it can handle so that it doesn't provide options like "Choose art" for non-library items but can still show details like the plot etc.

ksooo:
> They get one assigned in CVideoDatabase::GetNavCommon()

 I really do not like the concept of non-const CFileItem::GetVideoInfoTag whcih creates a tag on demand, leading for instance to file items originally created for a pvr recording tag and later getting an additional video info tag assigned. For those I fixed numerous bugs of the kind
<pre>
if (item->HasVideoInfoTag())
 return item->GetVideoInfoTag()->GetSomething();
else if (item->HasPVRRecordingInfoTag())
 return item->GetPVRRecodingInfoTag()->GetSomething();
</pre>

be changing the order of the 'poor man's type check'. Just for the record. ;-)

da-anda:
@Montellese IMO all the extra buttons should go and get replaced by one "manage" button that simply triggers the context menu for that item. I tried to implement this ~2 years ago but failed. I managed to get the button show the context menu etc, but due to the way the infodialog is opened (in a while loop or so) I didn't manage to fix the "update" (rescrape) feature while the dialog is open. Somebody with more insight would have to do it. I also had to add a new method "getPlayableUrl" (or alike) to the cFileItem items, because we currently create a temporary cFileItem, replace it's library VFS path with a translated one and pass it to the info dialog, but the context menu handler required the library VFS path. We did this IMO hack on several places IIRC.

tamland:
@Montellese imo your "other" solution is the one we should try to do. Instead of disabling it, we should work on making CGUIDialogVideoInfo just a dialog that shows info that works for anything video-like, and untangle all the weird scraper/database logic that doesn't belong in here at all.

phil65:
So what do you guys think? From what I understand this PR is a start to get things into right direction. I could try to improve the VideoInfo dialog later on in case i find some time.

tamland:
Just go for it. These other discussions are off-topic;)

phil65:
jenkins build this please.

PR #10950, opened 2016-11-17
Build fixes and musl configuration

kraj:


bkuhls:
We already have https://github.com/xbmc/xbmc/pull/9672 to add musl support, according to travis it needs some work to also make cmake happy.

Rechi:
@kraj this needs a rebase

kraj:
@Rechi true, although now I see autoconf build system is gone in master, so I have to redo some patches in Cmake way.

kraj:
I think this pull is obsolete since autotools are dropped and the fix for RssReader is already available upstream now.

PR #10951, opened 2016-11-17
Song Info Dialog contributors in control list

DaveTBlake:
Enable a control list to show the various contributors to a song e.g. composer, conductor, musicians and other people involved in the recording. Clicking on the artist displays the artist info dialog for that artist.  

This is particularly nice when you have artists covering songs composed by other artists in your library, or doing background vocals, or playing instruments etc. It also lets the user look at conductor or orchestra details for classical music, including the discography of other albums by that orchstra or conductor. Of course this needs the music to be fully tagged, and artist info to be scraped.

This also gives the skin the option to format how the artist name and role are displayed.

@phil65 if you could look at what tweek is needed for Estuary that would be great.
@HitcherUK I got this working myself in Confluence, but probably needs checking.

@tamland not sure I am opeing the artist dialog correctly. I could not see how to set the artist and use `CGUIDialogMusicInfo.showfor`

In Confluence it could look like this
image

### DIFF
--- a/xbmc/music/dialogs/GUIDialogSongInfo.cpp
+++ b/xbmc/music/dialogs/GUIDialogSongInfo.cpp
@@ -80,9 +80,13 @@ bool CGUIDialogSongInfo::OnMessage(CGUIMessage& message)
           db.Close();
         }
       }
+      CGUIMessage msg(GUI_MSG_LABEL_RESET, GetID(), CONTROL_LIST);
+      OnMessage(msg);
       break;
     }
   case GUI_MSG_WINDOW_INIT:
+    CGUIDialog::OnMessage(message);    
+    Update();
     m_cancelled = false;
     break;
 
@@ -110,6 +114,32 @@ bool CGUIDialogSongInfo::OnMessage(CGUIMessage& message)
       {
         OnGetThumb();
         return true;
+      }
+      else if (iControl == CONTROL_LIST)
+      {
+        int iAction = message.GetParam1();
+        if ((ACTION_SELECT_ITEM == iAction || ACTION_MOUSE_LEFT_CLICK == iAction))
+        {
+          CGUIMessage msg(GUI_MSG_ITEM_SELECTED, GetID(), iControl);
+          g_windowManager.SendMessage(msg);
+          int iItem = msg.GetParam1();
+          if (iItem < 0 || iItem >= static_cast<int>(m_song->GetMusicInfoTag()->GetContributors().size()))
+            break;
+          int idArtist = m_song->GetMusicInfoTag()->GetContributors()[iItem].GetArtistId();
+          if (idArtist > 0)
+          {
+              CGUIWindowMusicBase *window = (CGUIWindowMusicBase *)g_windowManager.GetWindow(WINDOW_MUSIC_NAV);
+              if (window)
+              {
+                CFileItem item(*m_song);
+                std::string path = StringUtils::Format("musicdb://artists/%li", idArtist);
+                item.SetPath(path);
+                item.m_bIsFolder = true;
+                window->OnItemInfo(&item, true);
+              }
+          }
+          return true;
+        }
       }
     }
     break;
@@ -147,20 +177,25 @@ bool CGUIDialogSongInfo::OnBack(int actionID)
 
 void CGUIDialogSongInfo::OnInitWindow()
 {
-  CMusicDatabase db;
-  db.Open();
+  // Normally have album id from song
+  m_albumId = m_song->GetMusicInfoTag()->GetAlbumId();
+  if (m_albumId < 0)
+  {
+    CMusicDatabase db;
+    db.Open();
 
-  // no known db info - check if parent dir is an album
-  if (m_song->GetMusicInfoTag()->GetDatabaseId() == -1)
-  {
-    std::string path = URIUtils::GetDirectory(m_song->GetPath());
-    m_albumId = db.GetAlbumIdByPath(path);
-  }
-  else
-  {
-    CAlbum album;
-    db.GetAlbumFromSong(m_song->GetMusicInfoTag()->GetDatabaseId(),album);
-    m_albumId = album.idAlbum;
+    // no known db info - check if parent dir is an album
+    if (m_song->GetMusicInfoTag()->GetDatabaseId() == -1)
+    {
+      std::string path = URIUtils::GetDirectory(m_song->GetPath());
+      m_albumId = db.GetAlbumIdByPath(path);
+    }
+    else
+    {
+      CAlbum album;
+      db.GetAlbumFromSong(m_song->GetMusicInfoTag()->GetDatabaseId(), album);
+      m_albumId = album.idAlbum;
+    }
   }
   CONTROL_ENABLE_ON_CONDITION(CONTROL_ALBUMINFO, m_albumId > -1);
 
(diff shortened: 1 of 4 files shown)


DaveTBlake:
While , without any xml changes, in Estuary this appears

Song Info Estuary

Obviously I need to change something to get artist name in the control list items (showing here as 0).

In the lower panel, where the ContributorsAndRole infolabel currently is, the missing song information could usefully be shown - song genre, mood, comments, rating and votes, userrating, and playcount.

zag2me:
This will be great for testing once we can see those fields in the song info.

Thanks!

DaveTBlake:
@ronie unfortunately adding contributors as a control list to Estouchy is a bit beyond my skinning skills.

addons/skin.estuary/1080i/DialogMusicInfo.xml
```diff
@@ -149,6 +149,42 @@
 				<ondown>130</ondown>
 				<onup>9000</onup>
 				<orientation>horizontal</orientation>
+        <itemlayout height="72.5" width="595" condition="Container.Content(songs)">
```
phil65: indentation

addons/skin.estuary/1080i/DialogMusicInfo.xml
```diff
@@ -149,6 +149,42 @@
...
+            <top>10</top>
+            <width>546</width>
+            <height>60</height>
+            <font>font12</font>
+            <aligny>center</aligny>
+            <label>$INFO[ListItem.Label,[COLOR button_focus], - [/COLOR]]$INFO[ListItem.Label2]</label>
+          </control>
+        </itemlayout>
+        <focusedlayout height="72.5" width="595" condition="Container.Content(songs)">
+          <control type="group">
```
phil65: no need to put these two controls into a group.

DaveTBlake:
Can this go into b6 please?

xbmc/music/dialogs/GUIDialogSongInfo.cpp
```diff
@@ -80,9 +83,14 @@ bool CGUIDialogSongInfo::OnMessage(CGUIMessage& message)
           db.Close();
         }
       }
+      CGUIMessage msg(GUI_MSG_LABEL_RESET, GetID(), CONTROL_LIST);
+      OnMessage(msg);
+      m_songContributors->Clear();
       break;
     }
   case GUI_MSG_WINDOW_INIT:
+    CGUIDialog::OnMessage(message);
```
Razzeee: Just a dumb question, why do we need to do this?

razzeee (approved):

DaveTBlake:
jenkins build this please

xbmc/music/dialogs/GUIDialogSongInfo.cpp
```diff
@@ -58,10 +58,13 @@ CGUIDialogSongInfo::CGUIDialogSongInfo(void)
   m_needsUpdate = false;
   m_startUserrating = -1;
   m_loadType = KEEP_IN_MEMORY;
+  m_songContributors = new CFileItemList;
```
tamland: Don't use new/delete (ever please). this should be a unique_ptr

tamland:
I hope this api change is fully backwards compatible?

And I sorry to be that guy again, but I'm not liking the direction this is going. I know direct db access is already here, but we should work on *removing* this kind of db code from gui, not add more, digging the grave even deeper.

DaveTBlake:
>I hope this api change is fully backwards compatible?

Yes. Skins can continue not displaying this information, although that would be a shame. I hope that Estuary, as the default skin, will kead the way showing all the other skins what is possible.

>I know direct db access is already here, but we should work on removing this kind of db code from gui, not add more, digging the grave even deeper.

I am at a loss to understand what you want. How is information meant to get out of the database and onto the screen? And I don't see any grave.

xbmc/music/dialogs/GUIDialogSongInfo.cpp
Montellese: Because you don't have to care about calling delete, it's taken care of by the smart pointer.

tamland:
Not sure how to explain this; this is the basic architecture of separating code into parts that have different concerns. What this is doing is coupling the gui to the database. CGUIDialogSongInfo is a class that should be responsible for dialoging and displaying info. It's now doing low level database things such as connecting, passing database id, and GetArtistPath something. 

We are removing things like this elsewhere such that it's possible to actually reuse the gui without requiring a local database with this rigid structure to exist. The info dialog should show the info that is passed to it, that's it.

>And I don't see any grave.

For example: add support for plugins to provide this info. You have to trash it all and start over.

xbmc/music/dialogs/GUIDialogSongInfo.cpp
tamland: ^that, and it's explicit who owns it. more reasoning here: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rr-newdelete

tamland:
>It is not as if the GUI is manipulating SQL or something horrible. It is using methods of a separate object to get the artist and path data to populate the dialog. I am always pleased to have input, but just saying "not liking direction" isn't very useful.

For example you can attach contributors to the song file item that the dialog is showing, from where it was first created. This way, the dialog can just read the info from the song directly it without being coupled to the database. In addition, the contributor info will the be available from anywhere in the gui, not just this one dialog.

That's the ideal solution. If that is too much work now (which it probably is), even just moving all data access code to the ShowFor method (which basically acts as a controller) would be great improvement as it keeps all non-gui code separate and compartmentalized; making it much easier to refactor later as opposed to when it's entangled in other dialog code. 

The problem is that this "separate object" (the db) is a leaky abstraction and you depend on a specific one. If it was a clean interface, and was passed *in*, it would be an entirely different story.

zag2me:
This is great, but does it need improvements for the latest v2 version of Estuary?

DaveTBlake:
>This is great, but does it need improvements for the latest v2 version of Estuary?

AFAIK no, it should be fine with v2, assuming that addons/skin.estuary/1080i/Variables.xml is not different in v2.

DaveTBlake:
@tamland / @Razzeee what I have done here is follow the approach taken in the artist/album music info dialog as a simple means to make the new data that Kodi process visible to the user in a nice (for user) way.  What you would prefer is a fundamental redesign of *all* these dialogs to meet current archtectural aspirations.

I am willing to work under your direction to achieve the latter, I fully understand architectural concepts, decoupling and encapsulation etc., and you clearly have a vision for what you want. The dialogs seem a convoluted mess to me too. BUT I would like to see the UI we release for v17 support the new music lib facilities. These were added early in alpha, and a great shame that I didn't have the know-how to do more work on the UI at the time.

So it's like this: you guys can support me to get this released and give the users a taste of more good things to come in v18. And I will feel commited and engaged to improve the whole dialog situation. Or you can just kick the ball into the long grass, and it's game over.

Razzeee:
It's all about being aware and in communication where we want to be in the future. I don't think it should block this, as it's one small field (more) that would need to be moved, once we go for the new structure.

DaveTBlake:
Heartened by your comments @Razzeee I have rebased for Estuary v2.

I have not been able to get the change to unique_ptr to work, seems to clash with CGUIMessage and all the other places it is used have the control list items defined as CFileItemList* with new/delete. Makes me wonder if keeping consistency with them is the best idea here.

DaveTBlake:
@tamland I have decoupled the control list action from the music db, using `CGUIWindowMusicBase::OnItemInfo`, I believe that is what you were getting at?  Better (even if I can't see to use unique_pointer)?

tamland:
Yes, a lot. If GetContributors in the future returned the complete artist (instead of the db id) it could be passed directly to the next info dialog without the need for any 'database id' in the dialog code; then it can in principle work for any music source such as plugins.
But I mean, this will be up to you in the end. I don't an as clear vision as you might think;) It's like Razzeee said.. I'm not trying to block anything here; just getting some small incremental improvement to the current info dialog situation (and other places) and to see things going a right/different direction, and I'm happy:)

DaveTBlake:
Glad you are happier @tamland, and thanks for the improvements with binding the items of the control list. I have updated and squished.

Using the control list in testing makes me even more enthusiastic to have this in v17 please.

jjd-uk:
Code has been reviewed and it seems Dave has addressed the comments as best he can for v17, any objections to this now being merged?

FernetMenta:
DB operation should not be done on the main thread. I think the goal should be to get this right when implementing new features. 

>  any objections to this now being merged?

the labels show api change but no fix

DaveTBlake:
I consider this a fix, in that it completes the addition of a feature added at the beginning of v17 by making it available to the user it a useable way. Label was an omission, so I have added it. But of course others may disagree.

DaveTBlake:
>DB operation should not be done on the main thread. 

Really not sure what to make of that comment @FernetMenta 
It may well be that all the music and video dialogs need reworking collectively to suit a new architecture.  If someone can explain more what is planned then perhaps I could contribute for v18.  Meanwhile I hope that displaying contributors in a control list, just as the artist info dialog displays the discography, or the album info dialog displays the tracks, is reasonable for inclusion in v17.

FernetMenta:
@DaveTBlake I am just saying that DB operations on the main and renderthread are a major design flaw. OnSomething() methods are not supposed to by hijacked by heavy call like db queries.
I know this was the case before but this would have been an opportunity to makes things a little bit better.

DaveTBlake:
> DB operations on the main and renderthread are a major design flaw. OnSomething() methods are not supposed to by hijacked by heavy call like db queries.

@FernetMenta I would like to discuss this further on the forum http://forum.kodi.tv/showthread.php?tid=298153 and flesh out how that is to be achieved.

Razzeee:
jenkins build this please

PR #10954, opened 2016-11-18
[PVR][Estuary] Recordings window: show status for recordings in progress

ksooo:
PVR recordings window list items are missing an indicator for "in progress" recordings. This was reported in the forum.

This PR adds the missing functionality.

screenshot003

The change was tested with latest Krypton master on macOS.

@Jalle19 for review?
@phil65 we discussed this internally

### DIFF
--- a/xbmc/pvr/recordings/PVRRecording.cpp
+++ b/xbmc/pvr/recordings/PVRRecording.cpp
@@ -28,8 +28,9 @@
 
 #include "pvr/PVRManager.h"
 #include "pvr/addons/PVRClients.h"
-#include "pvr/recordings/PVRRecordingsPath.h"
 #include "pvr/channels/PVRChannelGroupsContainer.h"
+#include "pvr/recordings/PVRRecordingsPath.h"
+#include "pvr/timers/PVRTimers.h"
 
 #include "PVRRecording.h"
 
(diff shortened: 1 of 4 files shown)


phil65:
Skin changes approved. ;)

xbmc/pvr/recordings/PVRRecording.cpp
```diff
@@ -481,3 +482,24 @@ int CPVRRecording::ClientID(void) const
 {
   return m_iClientId;
 }
+
+bool CPVRRecording::IsInProgress() const
+{
+  const CDateTime now(CDateTime::GetUTCDateTime());
+  if (m_recordingTime <= now && m_recordingTime + m_duration >= now)
```
Jalle19: Isn't it enough to compare this? Why check the timer too?

xbmc/pvr/recordings/PVRRecording.h
```diff
@@ -226,17 +226,23 @@ namespace PVR
...
+     * @brief Get the identifier of the client that serves this recording
      * @return the client identifier
      */
     int ClientID(void) const;
 
     /*!
-     * @brief Retrieve the recording Episode Name
-     * @note Returns an empty string if no Episode Name was provided by the PVR client
+     * @brief Retrieve the recording episode name
+     * @return the episode name or an empty string if no episode name was provided by the PVR client
```
Jalle19: These cosmetics could go in a separate commit

ksooo:
@Jalle19 Added a comment explaining why comparing recording's start time and duration against 'now' is not sufficient. okay now?

xbmc/pvr/recordings/PVRRecording.cpp
```diff
@@ -481,3 +482,28 @@ int CPVRRecording::ClientID(void) const
...
+  {
+    const std::vector<CFileItemPtr> timers(g_PVRTimers->GetActiveRecordings());
+    for (const auto& timerItem : timers)
+    {
+      const CPVRTimerInfoTagPtr timer(timerItem->GetPVRTimerInfoTag());
+      if (timer->m_iClientId == m_iClientId &&
+          timer->ChannelTag()->UniqueID() == m_iChannelUid &&
+          timer->StartAsUTC() <= m_recordingTime &&
+          timer->EndAsUTC() >= m_recordingTime + m_duration)
```
Glenn-1990: Is there some kind of ID we can check?

It's possible that 1 channel has 2 or more recordings running at the same time.
One of them is actually recording and the other one not (disabled for example, lower prio, no tuner, ...), in this case both recordings are marked as "recording" and only one should.

ksooo: That's why we are checking the time in addition to client Id and channel id.

xbmc/pvr/recordings/PVRRecording.cpp
Jalle19: Okay

Jalle19 (approved):
No further questions your honor

ksooo:
jenkins build this please

xbmc/pvr/recordings/PVRRecording.cpp
ksooo: > Is there some kind of ID we can check?

 No. timers do not know the respective recording ids and vice versa.

> It's possible that 1 channel has 2 or more recordings running at the same time. One of them is actually recording and the other one not (disabled for example

 in this case the backend must only send one PVR_RECORDING to kodi, namely for the recording that is actually in progress. For the disabled timer there must be no PVR_RECORDING transferred to kodi.

PR #10955, opened 2016-11-18
[PVR][Estuary] Folder overlay fixes

Glenn-1990:
Problem: When grouping PVR recordings in folders the overlay shows a "HD" symbol. (screenshot 1)

This is caused by a bug in "PVRRecordings.cpp" which will set the overlay to "ICON_OVERLAY_HD".
Fixing this will solve the issue partially. The folders are handled as normal files now...
Two other fixes in estuary are needed to handle folders just the same as with normal videos. 
Screenshot 2 shows the result with this these three fixes.

@ksooo I will rebase once your other PR is in.

@phil65 I noticed that the "OverlayHD.png" image in estuary might be wrong? OverlayHD in Kodi means "Hard disk" not High Definition".

Before:
schermafdruk van 2016-11-18 21 58 32
After:
schermafdruk van 2016-11-18 20 44 10

### DIFF
--- a/addons/skin.estuary/1080i/Variables.xml
+++ b/addons/skin.estuary/1080i/Variables.xml
@@ -317,9 +317,10 @@
 		<value condition="ListItem.IsRecording">windows/pvr/record.png</value>
 		<value condition="ListItem.IsPlaying">overlays/watched/OverlayPlaying-List.png</value>
 		<value condition="ListItem.IsResumable">overlays/watched/resume.png</value>
+		<value condition="ListItem.IsParentFolder">DefaultFolderBack.png</value>
 		<value condition="ListItem.IsFolder">overlays/folder.png</value>
 		<value condition="!String.IsEmpty(ListItem.Overlay)">$INFO[ListItem.Overlay]</value>
-		<value condition="!ListItem.IsParentFolder">OverlayUnwatched.png</value>
+		<value>OverlayUnwatched.png</value>
 	</variable>
 	<variable name="SettingsSectionIcon">
 		<value condition="Window.IsActive(playersettings)">icons/settings/video.png</value>
--- a/addons/skin.estuary/1080i/Variables.xml
+++ b/addons/skin.estuary/1080i/Variables.xml
@@ -317,8 +317,8 @@
 		<value condition="ListItem.IsRecording">windows/pvr/record.png</value>
 		<value condition="ListItem.IsPlaying">overlays/watched/OverlayPlaying-List.png</value>
 		<value condition="ListItem.IsResumable">overlays/watched/resume.png</value>
+		<value condition="ListItem.IsFolder">overlays/folder.png</value>
 		<value condition="!String.IsEmpty(ListItem.Overlay)">$INFO[ListItem.Overlay]</value>
-		<value condition="ListItem.IsFolder + Container.Content(files)">overlays/folder.png</value>
 		<value condition="!ListItem.IsParentFolder">OverlayUnwatched.png</value>
 	</variable>
 	<variable name="SettingsSectionIcon">
--- a/xbmc/pvr/recordings/PVRRecordings.cpp
+++ b/xbmc/pvr/recordings/PVRRecordings.cpp
@@ -139,9 +139,9 @@ void CPVRRecordings::GetSubDirectories(const CPVRRecordingsPath &recParentPath, 
       unwatchedFolders.insert(pFileItem);
   }
 
-  // Remove the watched overlay from folders containing unwatched entries
+  // Change the watched overlay to unwatched for folders containing unwatched entries
   for (auto item : unwatchedFolders)
-    item->SetOverlayImage(CGUIListItem::ICON_OVERLAY_WATCHED, true);
+    item->SetOverlayImage(CGUIListItem::ICON_OVERLAY_UNWATCHED, false);
 }
 
 int CPVRRecordings::Load(void)


phil65:
I am aware, yes.
This here might be an easier "fix"? https://github.com/phil65/skin.estuary/commit/b16aca6ad5a30b732867a97f01b1ddcc878b81e6

Glenn-1990:
@phil65 Why not replacing the icon with a real "hard disk" icon  like this one?https://github.com/phil65/skin.estuary/blob/master/media/DefaultHardDisk.png
Otherwise other Kodi parts needing the hard disk icon will be broken...

phil65:
>Otherwise other Kodi parts needing the hard disk icon will be broken...

Could you point me to those other parts? 

If it helps to have a hard disk icon then sure, I could change it. For me it seemed that the small default square for files would look more elegant.
@ksooo 	, what´s your opinion?

ksooo:
I'm fine with the square for files. I. Like it more than a hard disk icon, because a file (recording in this case) is (!) not a hard disk, right?

ksooo:
...but for the recording folders, I definitely want a folder icon, neither hd nor square. 

Not sure I understood the question in the first place, sorry. :-)

Glenn-1990:
I'll try to explain the issues more in detail.

1) Kodi sets an incorrect overlay for unwatched recording folders.
Check the argument "false" in combo with ICON_OVERLAY_WATCHED here:
https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/recordings/PVRRecordings.cpp#L144
And the enum "+1" here:
https://github.com/xbmc/xbmc/blob/master/xbmc/guilib/GUIListItem.cpp#L187
This results in an ICON_OVERLAY_HD instead of ICON_OVERLAY_UNWATCHED.
@ksooo Can you confirm this?
2) Not directly related to this PR (as this issue will be hided after fixing the unwatched overlay), but ICON_OVERLAY_HD should be "hard disk" and not "high definition". @phil65  I'm fine that you use the square, as long as it's clear that it's not High def.
3) Recording folders should use the folder overlay and not a square or high def one. Also the parent folder should not be handled as a "file". 

To make a long story short:
- This PR fixes recording folders overlay. The same overlays as the regular file viewer are used now.
- @phil65 's commit fixes the high definition issue for possible other Kodi parts, so this commit is still a must, but not for the PVR part. https://github.com/phil65/skin.estuary/commit/b16aca6ad5a30b732867a97f01b1ddcc878b81e6

phil65:
thx for explanation. Sounds fine then.

xbmc/pvr/recordings/PVRRecordings.cpp
```diff
@@ -125,7 +125,7 @@ void CPVRRecordings::GetSubDirectories(const CPVRRecordingsPath &recParentPath,
       pFileItem->m_dateTime = current->RecordingTimeAsLocalTime();
 
       // Assume all folders are watched, we'll change the overlay later
-      pFileItem->SetOverlayImage(CGUIListItem::ICON_OVERLAY_WATCHED, false);
+      pFileItem->SetOverlayImage(CGUIListItem::ICON_OVERLAY_WATCHED);
```
ksooo: @Glenn-1990 ping. Why this change (and the other in this commit)? What does it fix or the other way around, what does not work without this change?

ksooo: This change is a noop, as the default for parameter bOnOff of method SetOverlayImage is false. So, no change in behavior, but decrease in readability. Please revert.

xbmc/pvr/recordings/PVRRecordings.cpp
```diff
@@ -141,7 +141,7 @@ void CPVRRecordings::GetSubDirectories(const CPVRRecordingsPath &recParentPath,
 
   // Remove the watched overlay from folders containing unwatched entries
   for (auto item : unwatchedFolders)
-    item->SetOverlayImage(CGUIListItem::ICON_OVERLAY_WATCHED, true);
+    item->SetOverlayImage(CGUIListItem::ICON_OVERLAY_UNWATCHED);
```
ksooo: Okay, this makes sense. May I suggest to change the code comment to make clear that we're setting unwatched here. Makes things more clear. 

And could you please add the parameter "false" to SetOverlayImage. Makes the code more clear, imo.

ksooo (changes requested):

xbmc/pvr/recordings/PVRRecordings.cpp
Glenn-1990: I'll update the comment and the parameter

xbmc/pvr/recordings/PVRRecordings.cpp
ksooo: > Indeed this is a noop, but to set an overlay, SetOverlayImage() should be called without the 2nd parameter... 

 This API is crap, if you ask me. But if you insist in removing the parameter, well, so shall it be. ;-)

xbmc/pvr/recordings/PVRRecordings.cpp
ksooo: thanks

xbmc/pvr/recordings/PVRRecordings.cpp
Glenn-1990: It's not that I insist, I just tried to be more consistent with other code.
Ideally there should be two overlay functions:
SetOverlayImage(ICON_OVERLAY) and SetOverlayImageFromPlaycount(int playcount) so that the 2nd parameter isn't needed. But this is outside of this PR ofc.

Glenn-1990:
@ksooo Updated and re added the 2nd parameter ;-)

ksooo:
jenkins build this please

ksooo (approved):

ksooo:
@Glenn-1990 thanks for your contribution.

addons/skin.estuary/1080i/Variables.xml
```diff
@@ -317,9 +317,10 @@
 		<value condition="ListItem.IsRecording">windows/pvr/record.png</value>
 		<value condition="ListItem.IsPlaying">overlays/watched/OverlayPlaying-List.png</value>
 		<value condition="ListItem.IsResumable">overlays/watched/resume.png</value>
+		<value condition="ListItem.IsParentFolder">DefaultFolderBack.png</value>
+		<value condition="ListItem.IsFolder">overlays/folder.png</value>
```
phil65: this would break watchedicon for tvshows / seasons I think.

ksooo: > thx for explanation. Sounds fine then.

ooops. you gave your okay. that's why i merged it. ?!

phil65: hehe, sorry. "sounds fine" wasnt really meant as a code ACK, but as a concept ACK. Will deal with it. 👍

Glenn-1990:
@phil65 Folder overlay (including parent directory) seems to be broken again (see screenshot), maybe accidentally with the massive skin update?

schermafdruk van 2016-12-08 23 15 10

phil65:
I wonder what youre using since it looks quite broken. :)

ksooo:
He has not deleted the 1080i folder.

ksooo:
Otherwise he's right, the folder icon change did not make it into Estuary v2. 

As far as I remember this was be intention as the changes were buggy wrt watched/unwatched markers for recording folders not shown anymore? @phil65

Glenn-1990:
Deleted all skin folders and refetched from git, still the same problem.

@phil65 what's looking broken except the overlay issue? 
I'm running Krypton from git with pvr.hts.

Seems that the fix was overwritten by "[estuary] 2nd edition", huge diff, so I made a screenshot ;-)

https://github.com/xbmc/xbmc/commit/314f286b4356e02ddc59cc2d3d418f382438775c

schermafdruk van 2016-12-09 19 57 18

ksooo:
If you have recording folders containing only watched recordings, a checkmark should be displayed, not a folder icon. 

Phil was referring to the broken overall appearance of your screen shot, not something broken wrt folder icons.

Glenn-1990:
@ksooo Folder watched icon is working ok.

The selected item on the screenshot is a fully watched recording folder, the others not.
schermafdruk van 2016-12-09 21 24 33

ksooo:
> Folder watched icon is working ok.

 This is current master, without your PR, right? Otherwise the folders would have folder icons, right? The question is how this does this look like with your PR.

 Assumption is that on current master there are no folder icons, but watched icons for folders are working okay and we suspect exactly the opposite with your PR. This is what I understand from @phil65

Glenn-1990:
@ksooo the screenshot is from krypton latest, without any patches. When a folder is fully watched, the icon will change from "unwatched" to "watched". With my PR there will be folder icons instead of "watched" and "unwatched" icons, they are meant for files only not folders if I'm right?

@phil65 I just saw that de folder overlays are broken in the file manager too. All folders have the same overlay than regular files. Maybe this is by design for estuary v2?

ksooo:
> With my PR there will be folder icons instead of "watched" and "unwatched" icons, they are meant for files only not folders if I'm right?

 Yes and no. ;-) If a folder contains only watched recordings it shall have not a folder icon, but "watched" icon. If a folder contains at least one unwatched recording, it shall have the folder icon.

> All folders have the same overlay than regular files. Maybe this is by design for estuary v2?

 I consider this a bug.

PR #10956, opened 2016-11-19
guilib: fix segfault caused by not protected method CGUITextureManage…

FernetMenta:
HasTexture was not thread safe. My system crashed reliably with skin confluence.


FernetMenta:
jenkins build this please

PR #10957, opened 2016-11-19
[listprovider] Refresh list when job was not finished

BigNoid:
## Description
Follow up of #10931. Instead of always invalidating, only invalidate if ```OnJobComplete``` wasn't reached with success. 
## Motivation and Context
The job manager isn't aware of jobs that are busy but get interrupted because the window is closed.
At this point ```Reset()``` is called. By marking the job there as invalidated if ```OnJobComplete``` wasn't called, the job will fire again the next time the list is visible and otherwise it will only fire again through external events.
@tamland I don't think I can fix this in any other way. The way the jobmanager is setup, is to get results from ```OnJobComplete```. It doesn't handle the cases where ```OnJobComplete``` isn't reached for whatever reason. There's ```OnJobProgress``` but afaict that is only useful for showing the progress. 
## How Has This Been Tested?
Runtime tested

## Types of change

## Checklist:

### DIFF
--- a/xbmc/guilib/GUIBaseContainer.cpp
+++ b/xbmc/guilib/GUIBaseContainer.cpp
@@ -862,9 +862,10 @@ void CGUIBaseContainer::FreeResources(bool immediately)
   if (m_listProvider)
   {
     if (immediately)
+    {
       Reset();
-
-    m_listProvider->Reset(immediately);
+      m_listProvider->Reset();
+    }
   }
   m_scroller.Stop();
 }
--- a/xbmc/listproviders/DirectoryProvider.cpp
+++ b/xbmc/listproviders/DirectoryProvider.cpp
@@ -187,7 +187,7 @@ CDirectoryProvider::CDirectoryProvider(const TiXmlElement *element, int parentID
 
 CDirectoryProvider::~CDirectoryProvider()
 {
-  Reset(true);
+  Reset();
 }
 
 bool CDirectoryProvider::Update(bool forceRefresh)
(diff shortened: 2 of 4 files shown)


BigNoid:
jenkins build this please

tamland:
Please just hold on a moment. This is getting way too complex now. By adding a bool you are doubling the number of states for the directory provider which is definitely not needed.

Can you explain how to you reproduce this issue? (the original ticket just said 'make a skin' basically)

BigNoid:
You can reproduce on Estuary (v1 and v2) by focusing movies, do a ```Skin.Reload()``` and immediately enter the movies section (before all the widgets are loaded). Go back to home screen and the widget lists that weren't finished loading yet will remain empty until a skin reload or one of the external event triggers happens.

tamland:
The issue here is with the "Do not reset and refresh on window close/open all the time" behaviour that was added in https://github.com/xbmc/xbmc/commit/1da515fb78c0c203afa6f45969719aa53e98b487. #10931 effectively undid this behaviour when state is being forced to invalidated. 

So the fix is to undo #10931 and either always reset on window change (remove the "immediately" thing), or remove the job cancelling (I don't understand why this have special behaviour and wasn't included in 'immediately' clause tbh..)

BigNoid:
Thx for looking into this. Always resetting/reloading on window load was not the right way to fix it I feel now as in a lot of cases this means unnecessary refreshing of the lists. So I've tried your second option and moved the job cancelling to the ```immediately``` section. That works, but I get a lot of ```EXC_BAD_ACCESS``` crashes in xcode on this line: https://github.com/xbmc/xbmc/blob/master/xbmc/utils/JobManager.cpp#L411.
Another drawback of this way is that the bool ```Container.IsUpdating``` isn't true anymore for the list that was busy loading if you reopen the window. (Skinners can show a busy icon while that condition is true to give visual feedback to the user).

tamland:
Not sure how you did it but https://github.com/tamland/xbmc/commits/fix_directory_provider works as far as I can tell.. There should be no changes to IsUpdating here. It's considered 'updating' as long as there's a job running.

BigNoid:
Yeah that works with no side effects afaict 👍 
Dropped my commit and picked 09cb7e00221ec8fa2cb0192fff12fab8ef936b1a.

jenkins build this please

PR #10959, opened 2016-11-19
[PVR] Recordings window: Fix show status for recordings in progress

ksooo:
This is a followup to #10954 that fixes the first approach. 

The quick check comparing recording's start time and duration against 'now' before doing a check against the currently recording timers is wrong. The start time and duration is kinda static and does not (necessarily) reflect any adjustments in the events schedule (for example it might last some minutes longer than originally scheduled). So, this check must not be done at all. Otherwise the in progress indicator will disappear too early.

This fix was tested on latest Krypton master on macOS and Linux.

@Jalle19 the correct approach is actually the opposite of your first thought on #10954 ;-)


ksooo:
jenkins build this please

ksooo:
jenkins build this please

ksooo:
jenkins build this please

PR #10960, opened 2016-11-19
[cmake] Properly find correct libcec version

fetzerch:
Fix compile error if libcec 3.1 is installed.

## Description
Just setting a pkgconfig minimum version doesn't work and just causes pkgconfig to not find an older package. But since our CMake modules don't trust package config and validate that the include directory
really exists, it would still just find whatever version is installed.

The proper way is to either call `find_package` with `VERSION` or set
`<Module Name>_FIND_VERSION` inside the module.

Additionally: If we don't have a version from package config, the code now tries to find the version from the header file.

## Motivation and Context
Fix.

## How Has This Been Tested?
Compile tested on xenial (without build-deb ppa). Btw @wsnipex, will there gonna be a build-dep ppa for xenial?

## Types of change

## Checklist:

### DIFF
--- a/project/cmake/modules/FindCEC.cmake
+++ b/project/cmake/modules/FindCEC.cmake
@@ -15,13 +15,23 @@
 #   CEC::CEC   - The libCEC library
 
 if(PKG_CONFIG_FOUND)
-  pkg_check_modules(PC_CEC libcec>=4.0.0 QUIET)
+  pkg_check_modules(PC_CEC libcec QUIET)
 endif()
 
-find_path(CEC_INCLUDE_DIR libcec/cec.h libCEC/CEC.h
+find_path(CEC_INCLUDE_DIR NAMES libcec/cec.h libCEC/CEC.h
                           PATHS ${PC_CEC_INCLUDEDIR})
 
-set(CEC_VERSION ${PC_CEC_VERSION})
+if(PC_CEC_VERSION)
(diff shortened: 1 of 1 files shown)


fetzerch:
jenkins build this please

project/cmake/modules/FindCEC.cmake
```diff
@@ -15,13 +15,23 @@
...
-  pkg_check_modules(PC_CEC libcec>=4.0.0 QUIET)
+  pkg_check_modules(PC_CEC libcec QUIET)
 endif()
 
-find_path(CEC_INCLUDE_DIR libcec/cec.h libCEC/CEC.h
+find_path(CEC_INCLUDE_DIR NAMES libcec/cec.h libCEC/CEC.h
                           PATHS ${PC_CEC_INCLUDEDIR})
 
-set(CEC_VERSION ${PC_CEC_VERSION})
+if(NOT PC_CEC_VERSION)
```
wsnipex: -NOT

fetzerch: ups, good catch, set that just for testing

fetzerch:
@wsnipex good to go then?

akva2:
this won't be the only such error if you didn't consider this aspect of the do-not-trust-pkgconfig change. just fyi, i had used the pkgconfig version check in more find modules originally.

PR #10670, opened 2016-10-09
AESinkOSS: Fix compilation after STREAM_TYPE transition

fritsch:
This completes the transition from the AE_FMT_Passthrough formats to StreamInfo.

PR #10961, opened 2016-11-19
[audioengine] use OSS only on FreeBSD

AlwinEsch:
This use OSS audio now only on FreeBSD

## Description
This prevent also the compile fault related to https://github.com/xbmc/xbmc/pull/10670 

## Motivation and Context
Want to have compile working to continue work ;-)

@FernetMenta is this way OK?

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/AudioEngine/AESinkFactory.cpp
+++ b/xbmc/cores/AudioEngine/AESinkFactory.cpp
@@ -36,10 +36,12 @@
   #if defined(HAS_ALSA)
     #include "Sinks/AESinkALSA.h"
   #endif
+  #if defined(TARGET_FREEBSD)
+    #include "Sinks/AESinkOSS.h"
+  #endif
   #if defined(HAS_PULSEAUDIO)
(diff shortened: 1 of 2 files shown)


FernetMenta:
+1
I don't think oss fallback was ever used in an case.

xbmc/cores/AudioEngine/AESinkFactory.cpp
```diff
@@ -35,11 +35,12 @@
 #elif defined(TARGET_LINUX) || defined(TARGET_FREEBSD)
   #if defined(HAS_ALSA)
     #include "Sinks/AESinkALSA.h"
+  #else
```
AlwinEsch: OK to change the `else` to `#elif defined(TARGET_FREEBSD)` here and on other positions (to have OSS only on FreeBSD)?

If it becomes separated to a own `#elif defined(TARGET_FREEBSD)` must be the PulseAudio part inserted two times (also available on FreeBSD).

AlwinEsch:
jenkins build this please

PR #10963, opened 2016-11-20
[PVR] Recordings window: fix recordings sub folders after #10930.

ksooo:
Recordings sub-subfolders did not work anymore after #10930. That PR revealed an ancient bug with recordings url/path parsing which is fixed by the PR.

Bug was reported here: http://forum.kodi.tv/showthread.php?tid=269814&pid=2459718#pid2459718

This fix was tested with latest krypton master on macOS.

@MilhouseVH fyi
@Jalle19 for a quick check? Want to have this in beta 6, since it is a regression.


ksooo:
jenkins build this please

ksooo:
jenkins build this please

ksooo:
WIN-32 jenkins error is unrelated.

PR #7133, opened 2015-05-13
[win32] binary addons: add support for MSYS/MinGW

Montellese:
These commits add basic support for downloading and installing the MSYS/MinGW environment (which we already use as part of our normal build system) to the binary addons buildsystem. The result is a toolchain file which can be used as part of the win32 binary addon buildsystem to build dependencies with the MinGW tools.
I have tested this with the `audiodecoder.timidity` addon which can currently not be built for win32 due to the fact that `timidity` needs to be built with MinGW.

This work is largly based on the work of @notspiff so most of the credit goes to him.

In an additional step we may have to look into supporting building binary addon dependencies in the MSYS/MinGW environment (not with a toolchain file) which would support using autotools buildsystems.

PR #10964, opened 2016-11-20
[binary-addons/win32] Add support for MSYS/MinGW

fetzerch:
This is a follow-up on https://github.com/xbmc/xbmc/pull/7133 but makes use of kodi's already existing msys2/mingw environment as suggested by @Paxxi.

Pros:
- Less maintenance effort, only one dependency to keep up2date
- Newer version of mingw/msys allows to compile all existing game.libretro cores.
- Less resources & faster builds (especially on jenkins), no need to download/unpackage/setup a second environment
- Less patching of libretro cores because the newer version is closer to their build system.

Con:
- Closer coupled to kodi. If wanted/needed, the old approach could still be used later. Addons don't have to be adapted.

Pings: @Montellese, @notspiff, @hudokkow, @garbear 

@afedchin: How much effort would it be to update the gcc package to 5.3.0-**2** (https://github.com/xbmc/xbmc/compare/master...fetzerch:binaddons_internalmingw?expand=1#diff-0d5d3df941fb50de5af166d16ef8b5bdR20)


afedchin:
the easiest way to update gcc is update whole mssy2+packages.
https://github.com/xbmc/xbmc/blob/master/project/BuildDependencies/DownloadMingwBuildEnv.bat#L25 this line defines version of msys2. if you update msys2 it will install all new versions of used packages automatically. but need to update msys2 + packages on our mirror also (it's required because life time of packages is limited and there is a situation then a package can go away from msys repo)

afedchin:
my two cents: we can update whole msys2+packages for v18. because current version is nearly an year old.

fetzerch:
Updating for v18 sounds good to me. For the moment we just copy gcc.exe, which is exactly what the updated gcc package would do.

fetzerch:
Since I don't want to block garbear's retroplayer PR: Anyone interested in reviewing this?
It should be non-intrusive for existing binary addons. @Paxxi Is this solution conceptually acceptable now?

Note to myself: remove the tmp commit before merging ;)

project/cmake/addons/depends/windows/cmake/mingw/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+
+# configure the MinGW toolchain file
+configure_file(${PROJECT_SOURCE_DIR}/Toolchain_mingw32.cmake.in ${CMAKE_INSTALL_PREFIX}/Toolchain_mingw32.cmake @ONLY)
+
+# configure MinGWConfig.cmake
+configure_file(${PROJECT_SOURCE_DIR}/MinGWConfig.cmake.in ${CMAKE_INSTALL_PREFIX}/MinGWConfig.cmake)
+
+# MinGW GCC 5.3.0-1 comes without cc.exe, Remove this once package is bumped to 5.3.0-p2
+# See https://github.com/Alexpux/MINGW-packages/pull/1034
+if(NOT EXISTS ${MINGW_PATH}/bin/cc.exe)
```
hudokkow: Maybe prefix comment with TODO:
Probably easier to find later...

project/cmake/scripts/windows/tools/patch.cmake
```diff
@@ -0,0 +1,37 @@
...
+  set(PATCH_ARCHIVE_NAME "patch-2.5.9-7-bin-1")
+  set(PATCH_ARCHIVE "${PATCH_ARCHIVE_NAME}.zip")
+  set(PATCH_URL "${KODI_MIRROR}/build-deps/win32/${PATCH_ARCHIVE}")
+  set(PATCH_DOWNLOAD ${BUILD_DIR}/download/${PATCH_ARCHIVE})
+
+  # download the archive containing patch.exe
+  message(STATUS "Downloading patch utility from ${PATCH_URL}...")
+  file(DOWNLOAD "${PATCH_URL}" "${PATCH_DOWNLOAD}" STATUS PATCH_DL_STATUS LOG PATCH_LOG SHOW_PROGRESS)
+  list(GET PATCH_DL_STATUS 0 PATCH_RETCODE)
+  if(NOT ${PATCH_RETCODE} EQUAL 0)
```
hudokkow: if(PATCH_RETCODE)

hudokkow:
Feel free to disregard my nitpicks. Just a matter of code consistency but this is far from being the only code in need of that kind of attention.
I say shove it in. More rebase fun for me with https://github.com/xbmc/xbmc/pull/10446 ;)

fetzerch:
Thanks @hudokkow: Changed it to `if(NOT PATCH_RETCODE EQUAL 0)`. Find that a bit more readable than just `if(NOT PATCH_RETCODE EQUAL)`

garbear:
thanks guys

PR #10966, opened 2016-11-20
Controller dialog: Fix Kodi actions sent after mapping final analog stick

garbear:
The final step in mapping the Kodi controller profile is "Move right analog stick left...". However, when the stick is moved left and the mapping is successful, the mapping wizard ends and the fully-left stick sends a volume down command.

The solution in this PR is to keep track of which axes have successfully mapped a button, and continue absorbing input from then until the axis is re-centered. The button mapping wizard has been modified to wait until all successfully-mapped axes are centered before unregistering itself.

## Motivation and Context
Better UX

## Types of change

Broken out from #10630


garbear:
jenkins build this please

PR #10967, opened 2016-11-20
Fixed missing binary addons

Memphiz:
897094feeef8357eaf7ce2a4328a062b1af9d151 <- changed the way platforms.txt files are parsed. It assumes that all platforms are in a space seperated list in that file. In reality those platforms are in multiple lines for some addons.

@hudokkow @fetzerch  for review.

Fixes regression where osx is missing some visualisation addons (because the platform is in the second line of platforms.txt)

### DIFF
--- a/project/cmake/scripts/common/CheckTargetPlatform.cmake
+++ b/project/cmake/scripts/common/CheckTargetPlatform.cmake
@@ -9,7 +9,11 @@ function(check_target_platform dir target_platform build)
   if(EXISTS ${dir} AND EXISTS ${dir}/platforms.txt)
     # get all the specified platforms
     file(STRINGS ${dir}/platforms.txt platforms)
-    string(REPLACE " " ";" platforms ${platforms})
+    
+    list( LENGTH platforms listlen )
+    if(${listlen} EQUAL 1)    
+        string(REPLACE " " ";" platforms ${platforms})
+    endif()
 
     # check if the addon/dependency should be built for the current platform
     foreach(platform ${platforms})


hudokkow:
Fix looks good. However, separate_arguments() can't properly handle empty lines and causes [this](https://github.com/xbmc/repo-binary-addons/pull/34), even if one uses `string(STRIP ... ...)` before separate_arguments().

I guess a warning is prefered to silently failing.

In the long term, solution might be to not use separate_arguments() and keep platforms on the same line or REGEX check and substitute `\r\n` or `\n`. Preferably both for good measure.
Stuff for another PR...

Memphiz:
I can give the regex a shot - something like s/[\r\n]*/\s/g - not sure if carriage return and linefeed can be used in Regex though

fetzerch:
Multiline regexes should work according to: https://cmake.org/pipermail/cmake/2011-January/041859.html

Memphiz:
@hudokkow please describe what your initial intention was here. ${platforms} already was a list of strings ... in the case where platform was on one line it woudl make "linux;osx" out of "linux osx" in your case ... after that it it is traversed via foreach ... now i wonder if foreach automatically splits at ";" or what? or are lists always the same as a single string just with some sort of internal seperator?

Memphiz:
i force pushed ... how about that approach? (tested with "all", "linux osx" and linux/osx on seperate lines)

wsnipex:
maybe https://github.com/xbmc/xbmc/blob/master/project/cmake/scripts/common/Macros.cmake#L386-L407 can be used here

hudokkow:
Intention was to deal with the above mentioned empty lines problem. And yes, AFAIK foreach() splits (only) at ";"

Fix looks good.
Alternatives:
* revert to use separate_arguments() and leave a TODO comment. Something like "revisit/refactor needed. separate_arguments() can't handle line LF/CR and empty lines."
* use core_file_read_filtered() like wsnipex suggested
* uniformize platforms.txt - https://github.com/xbmc/repo-binary-addons/pull/38

Memphiz:
@hudokkow looks good from the jenkins output (visualisations for osx were built ...) - i don't think you can combine UPLOAD_RESULT and RUN_TESTS - thats why there wasn't uploaded anything ...

hudokkow:
hmmm, interesting. We either upload or run tests...
Anyway, do we go with the uniformity approach and close this one? We can always improve the code next time it bites us in the ass.

Thx for your patience with my cock up. ;)

Memphiz:
I am for merging this as is. Your button :) - we don't upload when running tests because we don't have anything to upload because we don't build kodi but only the kodi-testsuite...

wsnipex:
I'm for merging this.

PR #10968, opened 2016-11-20
Fix digital buttons with analog actions only sending one action

garbear:
Sometime in the past, a bug was introduced where a digital button mapped to an analog action only sent 1 action, when it used to send 1 action every frame (as a true analog button/trigger/stick would). This fixes the bug by sending an analog action each frame for as long as the button is pressed.

Furthermore, a ramp-up function has been applied to the analog value. This is because with analog buttons/triggers/sticks, you can control the amount of the action, so scrolling slowly is possible. However, with digital buttons, the amount is always 1.0, which can be unpleasant when you're trying to scroll slowly through a long list.

Now, when the button is pressed, it has an initial analog value of 0.3, which ramps up to 1.0 after 1.5 seconds.

## Types of change

Broken out from #10630

### DIFF
--- a/xbmc/input/joysticks/generic/FeatureHandling.cpp
+++ b/xbmc/input/joysticks/generic/FeatureHandling.cpp
@@ -30,6 +30,8 @@
 using namespace JOYSTICK;
 
 #define ANALOG_DIGITAL_THRESHOLD  0.5f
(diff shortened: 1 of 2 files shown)


garbear:
Updated to fix issue where this failed miserably if multiple controllers were attached

garbear:
jenkins build this please

PR #10972, opened 2016-11-22
[addon] remove not needed header comment in CScreenSaver

AlwinEsch:
The comment says 'Things that MUST be supplied by the child classes', but there
are no childs, is everything inside him.

### DIFF
--- a/xbmc/addons/ScreenSaver.h
+++ b/xbmc/addons/ScreenSaver.h
@@ -36,7 +36,6 @@ public:
   virtual ~CScreenSaver() {}
   virtual bool IsInUse() const;
 
-  // Things that MUST be supplied by the child classes
   bool CreateScreenSaver();
   void Start();
   void Render();


AlwinEsch:
jenkins build this please and merge

akva2:
childs refer to instances in the context. so it's not wrong as such, just imprecise.

AlwinEsch:
For me the text was not understandable, to use as calls for screensaver is it clear that the rest of kodi must call them.
And for calls on add-on is it only:
```cpp
  void Start();
  void Render();
  void GetInfo(SCR_INFO* pInfo);
```

The GetInfo becomes also removed in future, was never used in Kodi and add-ons.

akva2:
Not disagreeing or objecting to anything, just providing the information.

AlwinEsch:
jenkins build this please

PR #10973, opened 2016-11-22
Cmake fixes

wsnipex:
assortment of fixes for the cmake build system

## Description
- fix FindPython module to always use pkg-config
~~- disable gold linker on debug builds per default, because it breaks wrap libs~~
 fixed linking with gold in debug builds
- don't rebuild ffmpeg and crossguid with kodi if we already have it in depends

## Motivation and Context
fixbuild breakeage

## How Has This Been Tested?
build tested on linux only

## Types of change

### DIFF
--- a/project/cmake/CMakeLists.txt
+++ b/project/cmake/CMakeLists.txt
@@ -331,7 +331,7 @@ else()
 endif()
 add_dependencies(${APP_NAME_LC} ${APP_NAME_LC}-libraries export-files pack-skins)
 whole_archive(_MAIN_LIBRARIES ${core_DEPENDS})
-target_link_libraries(${APP_NAME_LC} ${_MAIN_LIBRARIES} lib${APP_NAME_LC})
+target_link_libraries(${APP_NAME_LC} ${_MAIN_LIBRARIES} lib${APP_NAME_LC} ${DEPLIBS})
 unset(_MAIN_LIBRARIES)
 
 if(WIN32)
--- a/project/cmake/modules/FindPython.cmake
+++ b/project/cmake/modules/FindPython.cmake
@@ -5,39 +5,34 @@
 # PYTHON_INCLUDE_DIRS - the python include directory
 # PYTHON_LIBRARIES - The python libraries
 
-if(PKG_CONFIG_FOUND AND (NOT CMAKE_CROSSCOMPILING OR APPLE))
-  pkg_check_modules(PYTHON python QUIET)
+if(PKG_CONFIG_FOUND)
+  pkg_check_modules(PC_PYTHON python >= 2.7 QUIET)
 endif()
 
-if(NOT PYTHON_FOUND)
-  if(CMAKE_CROSSCOMPILING)
-    find_program(PYTHON_EXECUTABLE python ONLY_CMAKE_FIND_ROOT_PATH)
-    find_library(PYTHON_LIBRARY NAMES python2.6 python2.7)
-    find_path(PYTHON_INCLUDE_DIRS NAMES Python.h PATHS ${DEPENDS_PATH}/include/python2.6 ${DEPENDS_PATH}/include/python2.7)
-    set(PYTHON_INCLUDE_DIR ${PYTHON_INCLUDE_DIRS} CACHE PATH "python include dir" FORCE)
+find_program(PYTHON_EXECUTABLE python ONLY_CMAKE_FIND_ROOT_PATH)
+find_library(PYTHON_LIBRARY NAMES python2.7 PATHS ${PC_PYTHON_LIBDIR})
+find_path(PYTHON_INCLUDE_DIR NAMES Python.h PATHS ${PC_PYTHON_INCLUDE_DIRS} ${DEPENDS_PATH}/include/python2.7)
 
-    find_library(FFI_LIBRARY ffi)
-    find_library(EXPAT_LIBRARY expat)
-    find_library(INTL_LIBRARY intl)
+if(CMAKE_CROSSCOMPILING)
+  find_library(FFI_LIBRARY ffi REQUIRED)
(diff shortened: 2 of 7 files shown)


wsnipex:
jenkins build this please

project/cmake/modules/LDGOLD.cmake
```diff
@@ -1,4 +1,8 @@
-option(ENABLE_LDGOLD "Use GNU gold linker" ON)
+if(CMAKE_BUILD_TYPE STREQUAL Release)
+  option(ENABLE_LDGOLD "Use GNU gold linker" ON)
+else()
+  option(ENABLE_LDGOLD "Use GNU gold linker" OFF)
+endif()
```
fetzerch: Well, then you can just disable it. It mostly does make sense on debug builds. Because it can split the debug info into extern files for better link performance. Can you please add a note to the commit message why this is needed, otherwise such changes have a very high chance of being accidentally removed again.

project/cmake/CMakeLists.txt
```diff
@@ -331,7 +331,7 @@ else()
 endif()
 add_dependencies(${APP_NAME_LC} ${APP_NAME_LC}-libraries export-files pack-skins)
 whole_archive(_MAIN_LIBRARIES ${core_DEPENDS})
-target_link_libraries(${APP_NAME_LC} ${_MAIN_LIBRARIES} lib${APP_NAME_LC})
+target_link_libraries(${APP_NAME_LC} ${_MAIN_LIBRARIES} lib${APP_NAME_LC} ${DEPLIBS})
```
fetzerch: Sry, i don't understand it from reading the commit message. 3eb4687 was fixing the issue that libs were specified twice in the commit message. How should we then deal with it?

project/cmake/modules/FindPython.cmake
```diff
@@ -5,39 +5,35 @@
...
 
-if(CMAKE_SYSTEM_NAME STREQUAL Darwin)
-  find_library(FFI_LIBRARY ffi REQUIRED)
-  find_library(INTL_LIBRARY intl)
-  list(APPEND PYTHON_LIBRARIES ${FFI_LIBRARY} ${INTL_LIBRARY})
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(PYTHON REQUIRED_VARS PYTHON_INCLUDE_DIR PYTHON_LIBRARY PYTHON_LIBRARIES)
+if(PYTHON_FOUND)
+  set(PYTHON_INCLUDE_DIRS ${PYTHON_INCLUDE_DIR})
+  set(PYTHON_LIBRARIES ${PYTHON_LIBRARIES} CACHE STRING "python libraries" FORCE)
```
fetzerch: Why do you need to cache PYTHON_LIBRARIES?

project/cmake/modules/FindPython.cmake
```diff
@@ -5,39 +5,35 @@
...
 endif()
 
 mark_as_advanced(PYTHON_INCLUDE_DIRS PYTHON_LIBRARIES PYTHON_LDFLAGS)
```
fetzerch: mark_as_advanced should be called for all find_program/path/library calls made in the module. Imho a good idea if cmake did that on it's own.

project/cmake/modules/LDGOLD.cmake
wsnipex: well, this could be a bug in gold, but LE already had the same issue and it crashed on me as well with default ubuntu 16.04 toolchain. We simply cannot enable something that crashes.

wsnipex:
updated. Gold is fixed now, so we keep it enabled by default.

jenkins build this please

wsnipex:
jenkins build this please

wsnipex:
jenkins build this please

wsnipex:
jenkins build this please

fetzerch:
looks good to me now, thanks much

phate89:
the commit 8558d672e98b62f3ea0126ba491376add8ed71cb create a problem in windows env. Now it tries to use my local installed python that is 64 bit and fails to compile..

project/cmake/modules/FindPython.cmake
```diff
@@ -5,39 +5,34 @@
...
-    find_package(PythonLibs 2.7 REQUIRED)
+find_program(PYTHON_EXECUTABLE python ONLY_CMAKE_FIND_ROOT_PATH)
+find_library(PYTHON_LIBRARY NAMES python2.7 PATHS ${PC_PYTHON_LIBDIR})
+find_path(PYTHON_INCLUDE_DIR NAMES Python.h PATHS ${PC_PYTHON_INCLUDE_DIRS} ${DEPENDS_PATH}/include/python2.7)
+
+if(CMAKE_CROSSCOMPILING)
+  find_library(FFI_LIBRARY ffi REQUIRED)
+  find_library(EXPAT_LIBRARY expat REQUIRED)
+  find_library(INTL_LIBRARY intl REQUIRED)
+  find_library(GMP_LIBRARY gmp REQUIRED)
```
bkuhls: Is there a reason why we need GMP support for a cross-compiled python env? In buildroot and Openelec the check is reverted:

https://github.com/bkuhls/buildroot/blob/392aebf38605cc849b4f5295615fd44f1cf56507/package/kodi/0003-CMake-Remove-dependency-on-gmp-and-libintl-from-Find.patch

https://github.com/OpenELEC/OpenELEC.tv/blob/master/packages/mediacenter/kodi/patches/kodi-998.02-cmake-python.patch

bkuhls: ping, this question is still valid for Leia.

wsnipex: This is only used when building with our own depends system and should be totally irrelevant for buildroot. OE is outdated and LE certainly does not set KODI_DEPENDSBUILD

PR #10974, opened 2016-11-22
Fix GetFilter SQL for music artists, albums and songs nodes

DaveTBlake:
Fix a number of glitches and inconsistencies in the SQL created by `CMusicDatabase::GetFilter` for artists, albums and songs nodes. Some of them were introduced by my initial implementation of artist roles, and I would like to get it right before RC. Others are long standing but only show depending on the music data and the combination of multiple criteria. Also the better identification of individual artists, when there is more than one on a song, revealed some.

Unfortunately it is hard to see in what way the SQL has changed by just looking at the code, moreover without familiarity with the music db schema it is difficult to evaluate those changes.  Because of that I have delayed this PR for several months while I test it throughly (on both SQLite and MySQL).  I am confident that it is safe to release even at this stage of the cycle. 

Please don't be put off by the code diff! The SQL is constructed in a maintainable and modular way,  using string manipulation just as before, but the end result looks like a rewrite of this routine. In fact it may be better for me to provide a SQL diff instead if anyone is interested.

Some of the odd behaviour now fixed: 
- Inconsistent results between filtering albums by genre or filtering artists by genre and then navigating to albums, especially when album genre has been modified via NFO.
- Filtering by both genre _and_ artist showing albums with songs with that genre and songs with that artist, but not ensuring they were the same song.
- Displaying only album artists, but when navigate to albums for a selected artist also seeing those albums where they feature on some songs. It sometimes looks like random albums, and you get compilation albums which users complain about.
- Filtering artists for those that are album artists _and_ composers returning those that are an album artist but a composer of a song on some other album where they are not an artist. Similarly filtering albums for such an artist.
- Navigating from composers and seeing not only those albums with songs that they composed but also any album where they were album artist (and composed nothing).

### DIFF
--- a/xbmc/dbwrappers/Database.cpp
+++ b/xbmc/dbwrappers/Database.cpp
@@ -101,6 +101,54 @@ void CDatabase::Filter::AppendGroup(const std::string &strGroup)
     group += ", " + strGroup;
 }
 
+void CDatabase::ExistsSubQuery::AppendJoin(const std::string &strJoin)
+{
+  if (strJoin.empty())
+    return;
+
+  if (join.empty())
+    join = strJoin;
+  else
+    join += " " + strJoin;
+}
+
+void CDatabase::ExistsSubQuery::AppendWhere(const std::string &strWhere, bool combineWithAnd /* = true */)
+{
+  if (strWhere.empty())
+    return;
+
+  if (where.empty())
+    where = strWhere;
+  else
+  {
+    where += combineWithAnd ? " AND " : " OR ";
+    where += strWhere;
+  }
+}
+
+bool CDatabase::ExistsSubQuery::BuildSQL(std::string & strSQL)
+{
+  if (tablename.empty())
+    return false;
+  strSQL = "EXISTS (SELECT 1 FROM " + tablename;
+  if (!join.empty())
+    strSQL += " " + join;
+  std::string strWhere;
+  if (!param.empty())
+    strWhere = param;
+  if (!where.empty())
+  {
+    if (!strWhere.empty())
+      strWhere += " AND ";
+    strWhere += where;
+  }
+  if (!strWhere.empty())      
+    strSQL += " WHERE " + strWhere;
+
+  strSQL += ")";
+  return true;
+}
+
 CDatabase::CDatabase(void)
 {
   m_openCount = 0;
--- a/xbmc/music/MusicDatabase.cpp
+++ b/xbmc/music/MusicDatabase.cpp
@@ -6320,13 +6320,13 @@ bool CMusicDatabase::GetFilter(CDbUrl &musicUrl, Filter &filter, SortDescription
   int idRole = 1; // Default restrict song_artist to "artists" only, no other roles.
   option = options.find("roleid");
   if (option != options.end())
-    idRole = (int)option->second.asInteger();
+    idRole = static_cast<int>(option->second.asInteger());
   else
   {
     option = options.find("role");
     if (option != options.end())
     {
-      if (option->second.asString() == "all")
+      if (option->second.asString() == "all" || option->second.asString() == "%")
         idRole = -1000; //All roles
       else
         idRole = GetRoleByName(option->second.asString());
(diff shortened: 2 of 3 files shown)


zag2me:
Tested and no issues that I can see with my music library, roles working well and totally awesome.

I can't comment too much on the individual cases, but no errors that I can see.

xbmc/dbwrappers/Database.cpp
```diff
@@ -101,6 +101,55 @@ void CDatabase::Filter::AppendGroup(const std::string &strGroup)
...
+    join += " " + strJoin;
+}
+
+void CDatabase::ExistsSubQuery::AppendWhere(const std::string &strWhere, bool combineWithAnd /* = true */)
+{
+  if (strWhere.empty())
+    return;
+
+  if (where.empty())
+  where = strWhere;
```
notspiff: indent

xbmc/dbwrappers/Database.cpp
```diff
@@ -101,6 +101,55 @@ void CDatabase::Filter::AppendGroup(const std::string &strGroup)
...
+  {
+    where += combineWithAnd ? " AND " : " OR ";
+    where += strWhere;
+  }
+}
+
+bool CDatabase::ExistsSubQuery::BuildSQL(std::string & strSQL)
+{
+  std::string strWhere;
+  if (tablename.empty())
```
notspiff: move decl after if

xbmc/dbwrappers/Database.cpp
```diff
@@ -101,6 +101,55 @@ void CDatabase::Filter::AppendGroup(const std::string &strGroup)
...
+    strWhere += where;
+  }
+  if (!strWhere.empty())      
+    strSQL += " WHERE " + strWhere;
+
+  strSQL += ")";
+  return true;
+}
+
+
```
notspiff: one line extra.

xbmc/music/MusicDatabase.cpp
```diff
@@ -6342,104 +6342,130 @@ bool CMusicDatabase::GetFilter(CDbUrl &musicUrl, Filter &filter, SortDescription
...
+        ExistsSubQuery albumArtistSub("album_artist", "album_artist.idArtist = artistview.idArtist");
+        ExistsSubQuery songArtistSub("song_artist", "song_artist.idArtist = artistview.idArtist");
+        if (idRole > 0)
+          songArtistSub.AppendWhere(PrepareSQL("song_artist.idRole = %i", idRole));
+        if (idGenre > 0)
+        {
+          songArtistSub.AppendJoin("JOIN song_genre ON song_genre.idSong = song_artist.idSong");
+          songArtistSub.AppendWhere(PrepareSQL("song_genre.idGenre = %i", idGenre));
+        }
+        if (idRole <= 1 & idGenre > 0)
```
notspiff: &&

xbmc/music/MusicDatabase.cpp
```diff
@@ -6454,51 +6480,94 @@ bool CMusicDatabase::GetFilter(CDbUrl &musicUrl, Filter &filter, SortDescription
...
-        filter.AppendWhere(PrepareSQL(
-          "albumview.idAlbum IN (SELECT song.idAlbum FROM song JOIN song_artist ON song.idSong = song_artist.idSong "
-          "JOIN artist ON artist.idArtist = song_artist.idArtist "
-          "WHERE artist.strArtist like '%s' %s) OR " // All albums linked to this artist via songs
-          "albumview.idAlbum IN (SELECT album_artist.idAlbum FROM album_artist "
-          "JOIN artist ON artist.idArtist = album_artist.idArtist "
-          "WHERE artist.strArtist like '%s')", // All albums where album artists fit
-          option->second.asString().c_str(), strRoleSQL.c_str(), option->second.asString().c_str()));
-      // no artist given, so exclude any single albums (aka empty tagged albums)
+      if (idRole <= 1 & idGenre > 0)
```
notspiff: &&

notspiff (commented):
sorry for all the nits, just caught my attention.

DaveTBlake:
>sorry for all the nits, just caught my attention.

Not at all, thanks for looking.

DaveTBlake:
jenkins build this please

DaveTBlake:
Any objections to this fix being merged?

In addition to my own extensive testing I have had positive feedback from users.  A test build, with this and the other as yet unmerged music PRs, is here http://mirrors.kodi.tv/test-builds/win32/KodiSetup-20161126-f6cce77-Music1126.exe

PR #10976, opened 2016-11-23
FileFactory: handle library protocol

phil65:
when using a path like library://video/movies/ on home screen, the log gets polluted with some error messages:

WARNING: XFILE::CFileFactory::CreateLoader - unsupported protocol(library) in library://video/movies/f.xml/folder.jpg
...

This should fix it. 
@Montellese


Montellese:
Yup, that should be ok.

phil65:
jenkins build and merge.

PR #10979, opened 2016-11-23
fixed: implement the expected method in ActiveAE

notspiff:
Either the interface of CEngineStats and IAE have been confused, or just one of them were updated at some point.

### DIFF
--- a/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp
+++ b/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp
@@ -2856,9 +2856,10 @@ bool CActiveAE::HasDSP()
   return m_stats.HasDSP();
 };
 
-AEAudioFormat CActiveAE::GetCurrentSinkFormat()
+bool CActiveAE::GetCurrentSinkFormat(AEAudioFormat& format)
 {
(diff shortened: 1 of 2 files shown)


FernetMenta:
only used by ADSP. current work on this in also on this branch: https://github.com/FernetMenta/kodi-agile

akva2:
i think this should go to mainline as it fixes a (potential) existing bug though?

FernetMenta:
ADSP will be disabled for v17. @AchimTuran is working on it on the brach mentioned. @AchimTuran would this fix anything for v17?

AchimTuran:
@notspiff @akva2 
Thanks for fixing this bug! Nice to read something from you.

But it is already fixed in [kodi-agile](https://github.com/FernetMenta/kodi-agile/blob/master/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp#L2885-L2889). 

@FernetMenta 
> ADSP will be disabled for v17. @AchimTuran is working on it on the brach mentioned. @AchimTuran would this fix anything for v17?

Right for the final v17 adsp will be disabled, but I will immediately enable it again for v18. So user can't blame me.
This fix will only fix the broken LibKodiAudioEngine, which is only used by binary add-ons. But is is currently not used by any published add-on. :smiley:

PR #10982, opened 2016-11-24
[PVR] fix recordings window subfolder listings

ksooo:
This PR fixes problems with recording folders/subfolders with names containing special characters, like blanks.

The problem was reported in the forum: http://forum.kodi.tv/showthread.php?tid=269814&pid=2462329#pid2462329

The second commit fixes recordings listings containing two ".." items. GUI media window has been fixed some time ago, so this hack causes trouble now.

screenshot000

@MilhouseVH could you please include this PR in your next build and ask the reporter (grisia) to test and report back?

### DIFF
--- a/xbmc/pvr/recordings/PVRRecordings.cpp
+++ b/xbmc/pvr/recordings/PVRRecordings.cpp
@@ -340,14 +340,6 @@ bool CPVRRecordings::GetDirectory(const std::string& strPath, CFileItemList &ite
     }
   }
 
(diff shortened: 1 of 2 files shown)


Jalle19:
The path management for recordings seems to be  quite problematic. Maybe in the future we could refactor it to be more robust.

+1 either way, if it works.

ksooo:
jenkins build this please

ksooo:
bug fix confirmed by the reporter.

PR #10983, opened 2016-11-25
[cmake] split arch / cpu

peak3d:
CMake / depends: introduce a clear split between ARCH and CPU

## Description
This PR allows the definition of ARCH and CPU in the kodi depends and cmake step to be able building optimized versions for a specific machine.
Currently all is done using a single environment value "CPU" wich tries to make magic things during the build steps (sometimes hardcoded CPU's and architectures).

## Motivation and Context
Motivation: Create an optimized build for my armv7-a arch and cortex-a5 cpu and be also able to create generic builds if no special architecture / cpu is targetted.
Context: Allow options to define these values at the configure step of depends and transport make them available to the cmake step

## Notes
- the complete build is done with a generic architecture [X86_64 / arm / aarch64 ...], the parameters --with-arch and --with-cpu are only used for tuning.
- The buildstep folder arm64-v8a was renamed to arm64 because v8a is not a general architecture.
- I didn't modify the hardcoded CPU settings in configure.ac for rpi, because I'm not familiar with them.

## How Has This Been Tested?
- Specific builds for odroids armv7 / aarch64 / linux / android
- Jenkins build for all platforms

## Types of change

## Checklist:


project/cmake/scripts/android/ArchSetup.cmake
```diff
@@ -8,25 +8,29 @@ set(ARCH_DEFINES -DTARGET_POSIX -DTARGET_LINUX -D_LINUX -DTARGET_ANDROID)
...
+endif()
+
+set(FFMPEG_OPTS --enable-cross-compile --arch=${ARCH} --target-os=linux --disable-vdpau 
+                --cc=${CMAKE_C_COMPILER} --host-cc=${CMAKE_C_COMPILER} --strip=${CMAKE_STRIP})
+
+if(NEON)
+  set(FFMPEG_OPTS "${FFMPEG_OPTS} --enable-neon")
+endif()
+
+if(DEFINED CPU AND NOT CPU STREQUAL "")
```
wsnipex: if(CPU)

peak3d: for the complete line?

wsnipex (commented):
looks good, thanks for this.
only a few minors

project/cmake/scripts/android/ArchSetup.cmake
wsnipex: yes

koenkooi:
This is a big improvement, but it has some assumption that aren't always true:

1. Not every 32bit arm chip supports NEON, the most obvious example is armv6 (which is handled by the special rpi configs), but e.g. atmel sells cortex-a5 SoCs without NEON. 
2. -mfpu=neon isn't always the fastest option, depending on the implementation NEON instructions will stall the CPU pipeline. The vfpv3-d16 and vfpv4  settings tend work better with cheap cores like a5/a7/a53.

Item 1 is academic, since I haven't encountered an armv7 SoC yet *with* GPU that lacks NEON and the performance impact of 2. might only be noticed in the audio and thumbnail code.

Unless a platform/SoC gets the same amount of manpower as rpi I'd say leave things like fpu up to the higher level buildsystem (libreelec, openembedded, buildroot, etc), those tend to have better info about the target.

Anyway, this PR as-is is a massive improvement, so feel free to ignore the above :)

tools/depends/configure.ac
```diff
@@ -259,23 +244,25 @@ case $host in
     esac
     case $host in
       *86*-apple-darwin)
-        MC_CHECK_NOT_CPU([$use_cpu], "arm")
+        MC_CHECK_NOT_CPU([$use_arch], "arm")
```
peak3d: cpu should be never be arm[xxx], if you want to have a more specific architecture for arm (e.g. armv7-a) it has to be passed using --with-arch=armv7-a. CPU is something like cortex-A53 or xeon

project/cmake/scripts/linux/ArchSetup.cmake
```diff
@@ -3,25 +3,20 @@ set(SYSTEM_DEFINES -D__STDC_CONSTANT_MACROS -D_FILE_DEFINED
                    -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64)
 set(PLATFORM_DIR linux)
 set(CMAKE_SYSTEM_NAME Linux)
-if(WITH_ARCH)
-  set(ARCH ${WITH_ARCH})
+
+if(ARCH STREQUAL x86_64)
```
peak3d: https://github.com/xbmc/xbmc/pull/10983/commits/9a63443782685b822e128b101a4703f157206bed#diff-6d62c2eff58a3ca4980971f5fdfed800R63

project/cmake/scripts/common/ArchSetup.cmake
```diff
@@ -57,10 +57,14 @@ if(NOT CORE_SYSTEM_NAME)
   string(TOLOWER ${CMAKE_SYSTEM_NAME} CORE_SYSTEM_NAME)
 endif()
 
+if(WITH_ARCH)
+  set(ARCH ${WITH_ARCH})
+elseif(NOT CMAKE_TOOLCHAIN_FILE)
```
peak3d: This is crosscompile yes / no ^^ should be OK (??)

project/cmake/scripts/freebsd/ArchSetup.cmake
```diff
@@ -3,14 +3,11 @@ set(SYSTEM_DEFINES -D__STDC_CONSTANT_MACROS -D_FILE_DEFINED
                    -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64)
 set(PLATFORM_DIR linux)
 set(SYSTEM_LDFLAGS -L/usr/local/lib)
-if(WITH_ARCH)
-  set(ARCH ${WITH_ARCH})
+
+if(ARCH STREQUAL x86_64)
```
wsnipex: thanks, overlooked that :)

peak3d:
1.) Thanks for reviewing, @wsnipex.
2.) @koenkooi great input, learned again things!
Will be considered in the next steps. From idea I could think about optimization settings in ArchSetup.cmake wich will only be used if a user has defined a specific arch and / or cpu.
Builds without specific definition of one of them two  will simply not be tuned.

wsnipex:
@koenkooi 2.) is very interesting. Do you have any specific SOC example where neon is slower?

koenkooi:
@wsnipex anything with only a5/a7 cores. Apart from the GPU the NEON block is the biggest power hog, so a5/a7 designs tend to have a half sized NEON block with worse pipeline behaviour.

The biggest effect will be on handcrafted NEON, but I don't think there are a lot of people using CPU media decode using kodi on ARM. And even if they did, the handcrafted NEON will still be faster than C.

I think, but haven't tested, is that any improvement will only be noticable in the audio code since that's where most of the compiler generated floating point code will live.

tools/depends/configure.ac
```diff
@@ -139,88 +148,64 @@ esac
...
     cross_compiling="yes"
   ;;
   arm*-*-linux-gnu*|arm*-*-linux-uclibc*)
-    if test "$use_platform" = "auto"; then
-      if test "x$use_cpu" = "xauto"; then
-        use_cpu=$host_cpu
-      fi
-      use_toolchain="${use_toolchain:-/usr}"
-      platform_cflags="-Os -fPIC -DPIC"
-      platform_ldflags="-Wl,-rpath-link=$prefix/$deps_dir/lib"
```
wsnipex: too lazy to google? ;)
it adds the dir to the _runtime_ linker path, so shared libs in this dir will be found by the linker even if they are not in LD_LIBRARY_PATH. This is important for libs that come from firmware, e.g. mmal and omx on RBPI, which usually lives in /opt/vc

peak3d:
@koenkooi NEON is now a configure option, by default its value is auto and will only be overridden if arch is armv7*

peak3d:
Jenkins is building all platforms now except arm64-v8 because of the not matching jenkins settings.

tools/depends/configure.ac
```diff
@@ -218,7 +218,7 @@ case $host in
     platform_cflags="-fPIC -DPIC"
     platform_cxxflags="$platform_cflags"
     platform_os="linux"
-    platform_arch="x86_64"
+    platform_arch="x86-64"
```
peak3d: was a mistake :-) This happens if one do things too late at night

project/cmake/scripts/ios/ArchSetup.cmake
```diff
@@ -9,15 +9,12 @@ set(SYSTEM_DEFINES -D_REENTRANT -D_FILE_DEFINED -D_FILE_OFFSET_BITS=64 -D_LARGEF
                    -D__STDC_CONSTANT_MACROS)
 set(PLATFORM_DIR linux)
 set(CMAKE_SYSTEM_NAME Darwin)
-if(WITH_ARCH)
-  set(ARCH ${WITH_ARCH})
+
+if(ARCH STREQUAL arm OR ARCH STREQUAL aarch64)
```
fetzerch: Shouldn't it still be armv7 and arm64 for iOS?

There's one thing I think you didn't take into account in this PR: `ARCH` is included in the ending of binary addons (see https://github.com/xbmc/xbmc/blob/master/xbmc/addons/kodi-addon-dev-kit/include/kodi/libXBMC_addon.h#L64 and https://github.com/xbmc/xbmc/blob/master/project/cmake/scripts/common/Macros.cmake#L151 and https://github.com/xbmc/xbmc/blob/master/configure.ac#L581). If you change `ARCH` in CMake, everything will compile fine and even packaging might work, but then addons will just fail to load.

The cleanest solution would be to specify the binary addon ending in cmake (and autootools for the time being) as an additional variable and pass it to both `libXBMC_addon.h` and to the macro that builds the addon libs. It might be also a good idea to revise if the lib endings could be changed to the new ARCH var - but since these endings are used in a few more scripts, and some platforms may need special names, I'm not sure how feasible this is.
@wsnipex: do you have a better idea? btw: should this be v18?

peak3d:
@Memphiz could you pls. so nice and make an ios build with binary addons?

peak3d:
@FernetMenta can you pls include this into your OSX development and let me know if problems occur?
Thanx!

Memphiz:
@peak3d i tried an ios build on jenkins - full log:

http://jenkins.kodi.tv/job/IOS/lastFailedBuild/consoleFull

breaks on all depends - for example:

http://pastebin.com/BtSKwUwT

arch armv4t? also -mtune=cortex-a8 gets overridden with -mtune=armv7 - and as is see an -march=armv7 is added too ...

tools/depends/configure.ac
```diff
@@ -317,11 +313,6 @@ case $host in
 
         # setup which sdk to use
         if test "$use_platform" = "tvos"; then
-          # setup which cpu to use
-          if test "x$use_cpu" = "xauto"; then
-            use_cpu=arm64
```
Memphiz: tvos needs arm64 - there is no other hardware out there for tvos ...

tools/depends/configure.ac
```diff
@@ -330,18 +321,22 @@ case $host in
...
           fi
-
+        else
           target_platform=iphoneos
 
           found_sdk_version=[`$use_xcodebuild -showsdks | grep $target_platform | sort | tail -n 1 | awk '{ print $2}'`]
           use_sdk="${use_sdk:-$found_sdk_version}"
           sdk_name=$target_platform$use_sdk
           platform_min_version="$target_platform-version-min=5.1"
+          platform_ldflags+=" -Wl,-segalign,4000"
```
Memphiz: change of behavior - those ldflags should only be present when compiling for arm64!

project/cmake/scripts/common/ArchSetup.cmake
```diff
@@ -58,10 +58,14 @@ if(NOT CORE_SYSTEM_NAME)
...
 
+if(WITH_ARCH)
+  set(ARCH ${WITH_ARCH})
+elseif(NOT ARCH)
+  set(ARCH ${CMAKE_SYSTEM_PROCESSOR})
+endif()
+
 if(WITH_CPU)
   set(CPU ${WITH_CPU})
-elseif(NOT CMAKE_TOOLCHAIN_FILE)
```
Memphiz: change of behavior?

Memphiz:
after discussing with peak3d i set the label from v17 to v18 - at this state i feel not comfortable to have the target at v17  because alot of things that are changed here are not well understood yet.

Memphiz:
have to stop the review here. Its just to hard to get read through it. I see that there is a lot of changed behavior. Where ios had a default mcpu of cortex-a8 before the default cpu for ios is no empty. That means that for example in the ffmpeg buildsys the configure line for ios will change and we will effectivly change behavior where we shouldn't. There are some more things i think will change behavior - tried to comment on them as good as i can.

peak3d:
thanx for reviewing, @Memphiz , will look over your comments asap

peak3d:
Will think about other concepts.....

PR #10985, opened 2016-11-25
VideoPlayer: change time parameter of SeekTime from int to double

FernetMenta:
fixes i.e. seeking for PVR for cases where pts is larger than INT_MAX * 1000

pvr and inputstream addons need to be adapted

### DIFF
--- a/addons/kodi.inputstream/addon.xml
+++ b/addons/kodi.inputstream/addon.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<addon id="kodi.inputstream" version="1.0.5" provider-name="Team Kodi">
-  <backwards-compatibility abi="1.0.5"/>
+<addon id="kodi.inputstream" version="1.0.6" provider-name="Team Kodi">
+  <backwards-compatibility abi="1.0.6"/>
   <requires>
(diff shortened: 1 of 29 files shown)


FernetMenta:
@ksooo @peak3d @notspiff ping

peak3d:
ack. Will change as soon as it its merged

ksooo:
v17? Really?

ksooo:
Who is going to adapt the pvr addons?

ksooo:
You need to update https://github.com/xbmc/xbmc/blob/master/addons/xbmc.pvr/addon.xml for the pvr addon api bump.

FernetMenta:
@ksooo pvr addons will break for a short while, I will do the changes online after this was merged. objections?

jenkins build this please

ksooo:
> @ksooo pvr addons will break for a short while, I will do the changes online after this was merged. 

 thanks.

> objections?

 no, not from my side.

notspiff:
@FernetMenta i am travelling. gave you push rights to inputstream.rtmp

FernetMenta:
@notspiff ok, I will update rtmp too.

ksooo:
@FernetMenta please do not forget to announce this api change in the forum.

@dennishamester ping, pvr.octonet needs to be adjusted

PR #10988, opened 2016-11-25
[cmake] allow using an existing Texturepacker binary

wsnipex:
allows passing the PATH to an external Texturepacker binary. Restores the same functionality as in autotools.
If the path is invalid, we still search in default cmake paths.

usage: -DWITH_TEXTUREPACKER=/path/to


lrusak:
what about https://github.com/xbmc/xbmc/blob/master/project/cmake/scripts/linux/Install.cmake#L139-L142

wsnipex:
fixed the install.
jenkins build this please

PR #10990, opened 2016-11-25
RFC: Workaround the skin unload bug - force close TextureBundle

fritsch:
This picks ace's fix and fixes the compile error - so that we come one step forward, as this seems to be the bug keeping us in beta.

### DIFF
--- a/xbmc/addons/Skin.cpp
+++ b/xbmc/addons/Skin.cpp
@@ -369,6 +369,8 @@ const INFO::CSkinVariableString* CSkinInfo::CreateSkinVariable(const std::string
 
 void CSkinInfo::OnPreInstall()
 {
+  if (IsInUse())
+    CApplicationMessenger::GetInstance().SendMsg(TMSG_EXECUTE_BUILT_IN, -1, -1, nullptr, "UnloadSkin");
 }
 
 void CSkinInfo::OnPostInstall(bool update, bool modal)
--- a/xbmc/guilib/TextureBundle.cpp
+++ b/xbmc/guilib/TextureBundle.cpp
@@ -78,6 +78,12 @@ int CTextureBundle::LoadAnim(const std::string& Filename, CBaseTexture*** ppText
   return 0;
 }
 
+void CTextureBundle::Close()
+{
+  if (m_useXBT)
+    m_tbXBT.CloseBundle();
+}
+
 void CTextureBundle::SetThemeBundle(bool themeBundle)
 {
   m_tbXBT.SetThemeBundle(themeBundle);
--- a/xbmc/guilib/TextureBundle.h
+++ b/xbmc/guilib/TextureBundle.h
@@ -39,7 +39,7 @@ public:
   bool LoadTexture(const std::string& Filename, CBaseTexture** ppTexture, int &width, int &height);
 
   int LoadAnim(const std::string& Filename, CBaseTexture*** ppTextures, int &width, int &height, int& nLoops, int** ppDelays);
-
+  void Close();
 private:
   CTextureBundleXBT m_tbXBT;
 
--- a/xbmc/guilib/TextureBundleXBT.cpp
+++ b/xbmc/guilib/TextureBundleXBT.cpp
@@ -55,6 +55,11 @@ CTextureBundleXBT::CTextureBundleXBT(bool themeBundle)
 
 CTextureBundleXBT::~CTextureBundleXBT(void)
 {
+  CloseBundle();
+}
+
+void CTextureBundleXBT::CloseBundle()
+{
   if (m_XBTFReader != nullptr && m_XBTFReader->IsOpen())
   {
     XFILE::CXbtManager::GetInstance().Release(CURL(m_path));
--- a/xbmc/guilib/TextureBundleXBT.h
+++ b/xbmc/guilib/TextureBundleXBT.h
@@ -48,6 +48,8 @@ public:
                 int &width, int &height, int& nLoops, int** ppDelays);
 
   static uint8_t* UnpackFrame(const CXBTFReader& reader, const CXBTFFrame& frame);
+  
+  void CloseBundle();
 
 private:
   bool OpenBundle();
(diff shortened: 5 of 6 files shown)


fritsch:
@ace20022 the only issue in there was a void CloseBundle() vs a bool CloseBundle() ... 

I don't judge about if I like the idea of that approach or not - we need to get v17 out of the door.

ace20022:
Thx! I don't like it myself, since I'm not sure if there're side-effects

MartijnKaijser:
You get stuck when you upgrade from Helix to Krypton when it tries to handle the skin and this is the last what's written in the log although you let it run for much longer than that.
http://paste.ubuntu.com/23536557/

fritsch:
@tamland @Paxxi @ace20022 any ideas on possible other "workarounds"?

MartijnKaijser:
What does now (or still) works is:
* updating currently in use skin (which was the current problem we had to fix)
* installing new skin still works
http://paste.ubuntu.com/23536605/
* upgrading from previous Kodi version where you had Confluence as skin you now end up with Estuary (as it should)

So only the upgrade from previous Kodi version with a non-default skin you end up in some endless loop where nothing is written in the log anymore.

fritsch:
In the provided log it seems the code is not run at all, see: https://github.com/fritsch/xbmc/blob/6ac2f52d9d6135edff9c27caa3c76d45e3457a9c/xbmc/guilib/TextureBundleXBT.cpp#L66

No idea :-)

MartijnKaijser:
Successful skin upgrade when running v17 
http://paste.ubuntu.com/23536705/

Paxxi:
Can't say for sure but reading the log it seems to be an issue with the db migration not completing before updating the addons causing errors in cpluff. Not found, unknown addon and so on and probably leaving the addon db in a weird state.

xbmc/addons/Skin.cpp
```diff
@@ -369,6 +369,8 @@ const INFO::CSkinVariableString* CSkinInfo::CreateSkinVariable(const std::string
 
 void CSkinInfo::OnPreInstall()
 {
+  if (IsInUse())
+    CApplicationMessenger::GetInstance().SendMsg(TMSG_EXECUTE_BUILT_IN, -1, -1, nullptr, "UnloadSkin");
```
fritsch: The problem is a design issue. One cannot assume that you can just "move" a file that has open handles - we need a mechanism to ensure that everything is properly closed when updating a skin under user's ass :-)

tamland: Yes, *on windows*:) This is a windows design issue because there is no way for us to possibly ensure that all processes close the file before we move it.

fritsch: I would highly, highly welcome it - if we could encapsulate that into windows specific code. Any ideas?

tamland:
> Can't say for sure but reading the log it seems to be an issue with the db migration not completing before updating the addons causing errors in cpluff. Not found, unknown addon and so on and probably leaving the addon db in a weird state.

Not sure where you're reading this from.. http://paste.ubuntu.com/23536557/ ? This is just cpluff 'info' and all normal. Where it hangs is at downloading skin.aeon.nox.5. And the only way to check why is to get a stack trace..

Paxxi:
From these that were sprinkled here and there 
```
13560   DEBUG: Unpacking zip://special%3A%2F%2Fhome%2Faddons%2Fpackages%2Fscript.image.resource.select-2.0.1.zip/script.image.resource.select/ to D:\Kodi\Krypton\portable_data\temp\07905361-3d93-4b7b-9b56-dc38dcd8e661
09:59:23.643 T:13560   DEBUG: ADDON: cpluff: 'Unknown plug-in script.image.resource.select could not be uninstalled.'
09:59:23.685 T:13560    INFO: ADDON: cpluff: 'Could not read plug-in directory D:\Kodi\Krypton\portable_data\addons: No such file or directory'
09:59:23.746 T:13560    INFO: ADDON: cpluff: 'Could not read plug-in directory D:\Kodi\Krypton\addons: No such file or directory'
09:59:23.748 T:13560   DEBUG: ADDON: cpluff: 'Plug-in script.image.resource.select has been installed.'
09:59:23.754 T:13560   DEBUG: ADDON: cpluff: 'Not all directories were successfully scanned.'
09:59:23.757 T:13560   DEBUG: CAddonDatabase: script.image.resource.select has been installed.
09:59:23.963 T:13560   DEBUG: CAddonMgr: enabled script.image.resource.select
09:59:24.053 T:13560   DEBUG: Unpacking zip://special%3A%2F%2Fhome%2Faddons%2Fpackages%2Fskin.transparency-10.2.7.zip/skin.transparency/ to D:\Kodi\Krypton\portable_data\temp\71d0c653-13ce-44a8-9c3e-c8d7917d6303
09:59:24.528 T:17284    INFO: CVideoDatabase
```

MartijnKaijser:
@tamland it only hangs downloading Nox with this PR (tried three times and always the same). Taking a nightly and it downloads it just fine so this PR affects something in that.

xbmc/addons/Skin.cpp
afedchin: why we are using `fopen` (`_wfopen` for windows) in `CXBTFReader::Open` instead of using our file API?

Paxxi: I thought about this one but never got around to test it, if it works then it's by far the simplest way

fritsch:
@afedchin: feel free to supersede this PR here. Thx much.

Am 26.11.2016 5:23 nachm. schrieb "Pär Björklund" <notifications@github.com

afedchin:
@fritsch seems I was wrong (partially) a file itself can be moved or/and deleted but a directory which contains the file doesn't

MartijnKaijser:
any other suggestions to tackle this?

afedchin:
I have no suggestions.

MartijnKaijser:
Replaced by 
https://github.com/xbmc/xbmc/pull/11090

PR #10991, opened 2016-11-25
[Estouchy] update to match the new look of the default skin

ronie:
i've updated estouchy to match the new look of estuary.

@phil65 @HitcherUK @BigNoid  for review


BigNoid:
Testing it now on a touch device and it works great so far, just like previous version. One of the new design directions Estuary has taken is to use list overflow to show the user how the panels navigate. I feel this could benefit Estouchy too in the home widgets panels and the panel views in library. Since there are shadows now instead of hardline bars at top/bottom it looks like the rest of the items come out of nowhere, which would also be gone if using list overflow.

ronie:
correct me if i'm wrong, but wouldn't that break touch (and mouse) support?

the side effect of the code needed to get overflow working in panels,
is that the focused position will be off when using touch/mouse input.

estuary-mouse

BigNoid:
Right, good point. We need ```hitrect``` for the focused item I think as this offset is already the case in Estuary which is said to have mouse support.
I tried with the ```offsety``` tag, but the focusedposition is still off with that (so that tag is pretty much useless :) )

PR #10995, opened 2016-11-26
[PVR] Fullscreen video: On up/down and no channel info is currently s…

ksooo:
Fullscreen video: On up/down and no channel info is currently shown, display channel info for playing channel, not info for next/prev channel.

This is, how other PVRs behave and was requested in the forum: http://forum.kodi.tv/showthread.php?tid=297623

This change was tested on latest krypton master on macOS.

@Jalle19  objections?

### DIFF
--- a/xbmc/pvr/PVRManager.cpp
+++ b/xbmc/pvr/PVRManager.cpp
@@ -1438,9 +1438,15 @@ bool CPVRManager::PerformChannelSwitch(const CPVRChannelPtr &channel, bool bPrev
       return false;
     }
 
-    // no need to do anything except switching m_currentFile
     if (bPreview)
     {
+      if (!g_infoManager.GetShowInfo())
+      {
+        // no need to do anything
+        return true;
+      }
+
+      // no need to do anything except switching m_currentFile
       delete m_currentFile;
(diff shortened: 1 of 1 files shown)


Glenn-1990:
@ksooo Does this mean that channel up does not switch a channel on the first press?

ksooo:
> @ksooo Does this mean that channel up does not switch a channel on the first press?

 No. This has only effect if "confirm channel switch with OK is enabled". Then, and only then, first up/down opens the info osd for the playing channel. Without this PR, info osd for next channel is opened. This means, if you want to see see osd for the playing channel, without this PR, you have to do first up, then down.

ksooo:
> If this only happens when "confirm channel switches by pressing OK" is enabled then I guess it's fine. If people want to switch channels by pressing up/down immediately we shouldn't change the behavior for them.

 all good. behaves exactly like you like it.

ksooo:
jenkins build this please

ksooo:
@Glenn-1990 btw, channel up/down always switches channels, regardless of "confirm channel switch with ok" value. The latter only influences behavior of (cursor) up/down.

Glenn-1990:
@ksooo seems a valid addition then ;-)

PR #10996, opened 2016-11-27
Add advanced setting to control party mode mix length

ghost:
## Description
This adds &lt;partymixlength&gt;int&lt;/partymixlength&gt; to advancedsettings.xml and allows the user to control the length of the party mode playlist instead of using a hardcoded length of ten songs.

## Motivation and Context
The party mode list is hardcoded to only ten songs, which is often fine, but my personal preference is to have a longer list to allow for more advanced planning. Since it is likely not a setting most users will want to tune, but I suspect some users wish they could, I added it as an advanced setting.

## How Has This Been Tested?
I have been running this on my personal kodi installation for ~18 months (tracking master every few months). The change is trivial -- it simply replaces the QUEUE_DEPTH #define with g_advancedSettings.m_partyMixLength. It has no effects on any part of kodi other than the party manager and advanced settings loader.

n/a

## Types of change
- [ x] New feature (non-breaking change which adds functionality)

## Checklist:
- [ x] My change requires a change to the documentation, either Doxygen or wiki
- [x ] All new and existing tests passed

I do not think that this change requires a new test to be added. The only change to documentation needed is adding a description to http://kodi.wiki/view/advancedsettings.xml, which I am happy to write if this will be considered for inclusion.

Keep at least &lt;partymixlength&gt; tracks on the playlist when in party
mode.

### DIFF
--- a/xbmc/PartyModeManager.cpp
+++ b/xbmc/PartyModeManager.cpp
@@ -40,10 +40,9 @@
 #include "utils/StringUtils.h"
 #include "utils/Variant.h"
 #include "video/VideoDatabase.h"
+#include "settings/AdvancedSettings.h"
 
 using namespace PLAYLIST;
-
-#define QUEUE_DEPTH       10
 
 CPartyModeManager::CPartyModeManager(void)
 {
@@ -312,7 +311,7 @@ bool CPartyModeManager::AddRandomSongs(int iSongs /* = 0 */)
     iPlaylist = PLAYLIST_VIDEO;
 
   CPlayList& playlist = g_playlistPlayer.GetPlaylist(iPlaylist);
-  int iMissingSongs = QUEUE_DEPTH - playlist.size();
+  int iMissingSongs = g_advancedSettings.m_partyMixLength - playlist.size();
   if (iSongs <= 0)
     iSongs = iMissingSongs;
   // distribute between types if mixed
@@ -601,7 +600,7 @@ bool CPartyModeManager::AddInitialSongs(std::vector< std::pair<int,int > > &song
   int iPlaylist = m_bIsVideo ? PLAYLIST_VIDEO : PLAYLIST_MUSIC;
 
   CPlayList& playlist = g_playlistPlayer.GetPlaylist(iPlaylist);
-  int iMissingSongs = QUEUE_DEPTH - playlist.size();
+  int iMissingSongs = g_advancedSettings.m_partyMixLength - playlist.size();
   if (iMissingSongs > 0)
   {
     // generate iMissingSongs random ids from songIDs
--- a/xbmc/settings/AdvancedSettings.cpp
+++ b/xbmc/settings/AdvancedSettings.cpp
@@ -182,6 +182,8 @@ void CAdvancedSettings::Initialize()
 
   m_cddbAddress = "freedb.freedb.org";
 
+  m_partyMixLength = 10;
+
   m_handleMounting = g_application.IsStandAlone();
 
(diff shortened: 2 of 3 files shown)


jjd-uk:
@DaveTBlake maybe something for you to review

razzeee (approved):

Razzeee:
Code looks fine. Not sure about having more advanced settings, but that's a thing to discuss/solve some other time.

DaveTBlake:
All good, and I think this is a good use of advanced settings.
Happy to see this go into v17.

DaveTBlake (approved):

MartijnKaijser:
i am against having this as advanced setting so it should en up in the correct place in gui settings.

unknownlamer:
Is there any documentation on adding a GUI settings? I dug through the code a bit trying to figure out where to add it, and unless I am missing some code generator, it looks like a fairly complicated task. I am willing to do so since if adding as an advancedsetting is a no-go (I think this is a worthwhile change and I'd like to stop having to build my own packages eventually ;) ), but I'm hitting a wall on how to go about doing so.

Montellese:
Adding a GUI setting should actually be pretty easy in the code. Just look at `system/settings/settings.xml` for adding the definition of the setting and then into `xbmc/settings/Settings.h` and `xbmc/settings/Settings.cpp` on how to access it from the code.

jjd-uk:
If you're going to do a GUI setting then in my view it belongs in:

Settings -> Media -> Music -> Library

Maybe between https://github.com/xbmc/xbmc/blob/master/system/settings/settings.xml#L1126-L1130 and https://github.com/xbmc/xbmc/blob/master/system/settings/settings.xml#L1131-L1135 in the settings.xml with a setting id of something like:

**setting id="musiclibrary.partymixlength"**

Then you simply need to add it to Settings.h and Settings.cpp as Montellese advised, so something like:

**SETTING_MUSICLIBRARY_PARTYMIXLENGTH**

In both of these I used **PARTYMIXLENGTH** as that's what your advanced setting is called, but this can be anything you like.

For an example of how to add settings to GUI see https://github.com/xbmc/xbmc/pull/10199/commits

jjd-uk:
To be clearer:

In https://github.com/xbmc/xbmc/blob/master/xbmc/settings/Settings.h#L221-L231

`static const std::string SETTING_MUSICLIBRARY_PARTYMIXLENGTH;`

Then in https://github.com/xbmc/xbmc/blob/master/xbmc/settings/Settings.cpp#L264-L274 something like:

`const std::string CSettings::SETTING_MUSICLIBRARY_PARTYMIXLENGTH = "musiclibrary.partymixlength";`

and in https://github.com/xbmc/xbmc/blob/master/xbmc/settings/Settings.cpp#L1031-L1033 something like:

`settingSet.insert(CSettings::SETTING_SETTING_MUSICLIBRARY_PARTYMIXLENGTH);`

Then in settings.xml use

`setting id="musiclibrary.partymixlength"`

jjd-uk:
@MartijnKaijser Before unknownlamer changes anything could confirm you'd be ok with what I've directed him/her to do, and would there be a deadline for getting this into v17 since it's only a minor addition.

MartijnKaijser:
Make sure it gets into master first and we'll see.

unknownlamer:
Due to lack of time, if this can't be merged as an advanced setting it'll be at least a month before I can resume work (adding the setting is *dozens* of times the effort, will have to allocate a day to it and have none free for at least as long).

Might be a good idea for someone else later on... I can just close this and keep on building my own copy as it takes less of my time.

PR #11003, opened 2016-11-28
added: make CFileItemList iterable

notspiff:
This makes CFileItemList iterable, i.e.
```
for (auto& it : CFileItemListInstance)
```

### DIFF
--- a/xbmc/FileItem.h
+++ b/xbmc/FileItem.h
@@ -752,6 +752,11 @@ public:
   const std::string &GetContent() const { return m_content; };
 
   void ClearSortState();
+
+  VECFILEITEMS::const_iterator begin() { return m_items.cbegin(); }
+  VECFILEITEMS::const_iterator end() { return m_items.cend(); }
+  VECFILEITEMS::const_iterator cbegin() const { return m_items.begin(); }
+  VECFILEITEMS::const_iterator cend() const { return m_items.end(); }
 private:
   void Sort(FILEITEMLISTCOMPARISONFUNC func);
   void FillSortFields(FILEITEMFILLFUNC func);


xbmc/FileItem.h
```diff
@@ -723,6 +723,9 @@ class CFileItemList : public CFileItem
   const std::string &GetContent() const { return m_content; };
 
   void ClearSortState();
+
+  VECFILEITEMS::const_iterator begin() { return m_items.begin(); }
+  VECFILEITEMS::const_iterator end() { return m_items.end(); }
```
Montellese: Something doesn't fit together. You return `const_iterator` but the methods are called `begin()` and `end()` (as opposed to `cbegin()` and `cend()`) and you call `m_items.begin()` and `m_items.end()` which return `iterator`.
So either we want
* `const_iterator` in which case the methods should be `cbegin()` and `cend()` and `const`
* or `iterator` in which case the return type should be fixed

or we want both :-)

Montellese (commented):
Thanks a lot for this. I was always too lazy to do it :-/

akva2:
On purpose. We do not want a mutable iterator, CFileItemList is based on Add () paradigm. The compiler is smart enough but if you want explicit you shall have explicit give me 15mins to get home .

akva2:
meh. you are of course completely correct, my left hand had forgotten to commit the const versions which i realized was needed after testing. i thought you were (only) referring to begin() and end() returning const iterators wanting cbegin cend non-const overloads when i wrote ^^ (and obviously i "read" your comment too fast..). i did learn something new though - the compiler will not consider non-const overloads of cbegin|cend at all.

i have kept the code that was there and added the missing cbegin() cend() implementations. if you want mutable iteration, just say so. i gave my reasoning above, it just seemed to break surrounding design to me.

Montellese:
No I don't want mutable iterators because it will break stuff that works when using `Add()` and `Remove()`. I was just confused about the naming. But I didn't consider that when I wrote my comment so it wasn't 100% correct either.

hudokkow:
jenkins build this please

MartijnKaijser:
Should there be any updates/changes or can this be merged?

Paxxi:

Montellese:
Let's try again. jenkins build and merge this please

Montellese:
Just gave it a try and it doesn't work in all cases. E.g. when I have a `const CFileItemList` I cannot use the for-range because there's no `begin() const` and `end() const` (returning a `const_iterator`). I also checked the implementation of `std::vector` and there you have
* `begin()` and `end()`
* `begin() const` and `end() const`
* `cbegin() const` and `cend() const`

so we are missing the `begin() const` and `end() const`. Mind providing a follow up PR?

PR #10971, opened 2016-11-21
GUIWindowSlideShow: Protect against empty slides access

popcornmix:
## Description
Fix crash in CGUIWindowSlideShow

## Motivation and Context
Without this when using an iPhone uPnP app kodi crashes trying to dereference the empty vector

## How Has This Been Tested?
Avoids crash locally. Will be in tonight's Milhouse build.

## Types of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [X] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #11006, opened 2016-11-28
rbp: Don't close the display here it should be closed by the opener

popcornmix:
## Description
Don't close the display here it should be closed by the opener

## Motivation and Context
During a normal shutdown, either way is fine as the EGL context has already been shut down and the display will already be closed.
However while debugging the exception in https://github.com/xbmc/xbmc/pull/10971 I found that the cleanup code causes the display to be closed with the EGL context still active.
The firmware doesn't like that and can result in garbage on the display.

## How Has This Been Tested?
In last week of Milhouse builds.

## Types of change

## Checklist:


popcornmix:
jenkins build this please

PR #11012, opened 2016-11-29
[python] allow setting the videoinfotag path

ronie:
ListItem.Path (and FolderName, FolderPath, FileNameAndPath) return empty in the videoinfo dialog for python listings (widgets and the likes).

addons creating such listings need to set the videoinfotag path.

http://forum.kodi.tv/showthread.php?tid=298028

### DIFF
--- a/xbmc/interfaces/legacy/ListItem.cpp
+++ b/xbmc/interfaces/legacy/ListItem.cpp
@@ -454,6 +454,8 @@ namespace XBMCAddon
             item->GetVideoInfoTag()->SetVotes(StringUtils::ReturnDigits(value));
           else if (key == "trailer")
             item->GetVideoInfoTag()->m_strTrailer = value;
+          else if (key == "path")
+            item->GetVideoInfoTag()->m_strPath = value;
           else if (key == "date")
           {
             if (value.length() == 10)
--- a/xbmc/interfaces/legacy/ListItem.h
+++ b/xbmc/interfaces/legacy/ListItem.h
@@ -592,6 +592,7 @@ namespace XBMCAddon
       /// | album         | string (The Joshua Tree)
       /// | artist        | list (['U2'])
       /// | votes         | string (12345 votes)
+      /// | path          | string (/home/user/movie.avi)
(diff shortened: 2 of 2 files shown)


tamland:
and the difference from setPath is?

ronie:
setPath() sets the fileitem path:
https://github.com/xbmc/xbmc/blob/master/xbmc/FileItem.h#L135

setInfo() sets the path in the videoinfotag:
https://github.com/xbmc/xbmc/blob/master/xbmc/video/VideoInfoTag.cpp#L1407-L1410

if i knew the difference, i'd tell you ;-)

tamland:
> if i knew the difference, i'd tell you ;-)

Then it's settled ;) We should not leak these stupid things to the api and addon developers; they have already provided this info once. instead the info dialog should be fixed.

ronie:
ok, i spent some time exploring our code. since it's still a wild-wild west to me, my understanding of things might not be accurate (at all)...

if kodi creates a list of library items, it will *first* create the videoinfotag (including the path) and pass this tag to the creation of the fileitem:
https://github.com/xbmc/xbmc/blob/master/xbmc/video/VideoDatabase.cpp#L6670-L6675

it calls this method:
https://github.com/xbmc/xbmc/blob/master/xbmc/FileItem.cpp#L115-L119

which will copy the path from the videoinfo tag to the fileitem path:
https://github.com/xbmc/xbmc/blob/master/xbmc/FileItem.cpp#L1480-L1492


there's no equivalent to those fileitem creation methods in our python api.
we can only create a simple (non library) file item and optionally add metadata to the videoinfo tag afterwards.

ronie:
^ @tamland ?

tamland:
I think your links are out of date because the highlighted lines make no sense.. (remember to use permalinks!)

tamland:
But anyway, when I think about it, it sort of does make sense in a weird way to have another path on videoinfotag; assuming this other path *strictly* used for display purposes and those info labels.

phil65:
jenkins build this please.

ronie:
jenkins build this please

PR #11017, opened 2016-11-30
Space out order value of library nodes

ghost:
## Description
Adjusts the 'order' value of all library nodes so that the first node has a value of '10', and each subsequent node is 10 higher.

## Motivation and Context
When editing custom nodes, if the user wants to add a node as the first item or between two existing items this requires renumbering a number of pre-existing nodes to make space. This change gives the space to place custom nodes as required without needing to adjust the order of existing nodes.

## How Has This Been Tested?
Runtime testing to ensure all nodes show in the correct order with the updated order values.

## Types of change

## Checklist:

### DIFF
--- a/system/library/music/addons.xml
+++ b/system/library/music/addons.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
-<node order="22" type="folder">
+<node order="150" type="folder">
 	<label>1038</label>
 	<icon>DefaultAddonMusic.png</icon>
 	<path>addons://sources/audio/</path>
--- a/system/library/music/albums.xml
+++ b/system/library/music/albums.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
-<node order="3" type="filter" visible="Library.HasContent(Music)">
+<node order="30" type="filter" visible="Library.HasContent(Music)">
 	<label>132</label>
 	<icon>DefaultMusicAlbums.png</icon>
 	<content>albums</content>
--- a/system/library/music/artists.xml
+++ b/system/library/music/artists.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
-<node order="2" type="filter" visible="Library.HasContent(Music)">
+<node order="20" type="filter" visible="Library.HasContent(Music)">
 	<label>133</label>
 	<icon>DefaultMusicArtists.png</icon>
 	<content>artists</content>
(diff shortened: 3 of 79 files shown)


zag2me:
This is actually quite a good idea, I've thought similar in the past.

hudokkow:
Don't actually know who to ping... @mkortstiege, @DaveTBlake, @phil65 you were the ones last touching these files.

DaveTBlake:
Clearly you didn't test those nodes where you replaced "&" with "&amp". This replacement does not work, the "&" need to be left as they are.

Ignoble61:
Runtime tested it on a Mac - the nodes work fine for me with the &amp;. I can revert to the & if you prefer, but that isn't valid in xml.

DaveTBlake:
I can't test Mac, but fails on both Windows and Pi (LibreElec). Yes revert please, otherwise this PR will not be acceptable IMO.

Ignoble61:
`&`'s restored.

ronie:
'&' characters need to be escaped in xml documents.
if kodi fails to handle that correctly, something else must be wrong.

phil65:
@Ignoble61 Dont you have to rewrite basically all xmls anyways in case someone inserts a new item between  item 1 and item 2 for example? Imo this should be handled on editor side.

Ignoble61:
The editor doesn't have any concept of inserting between items or moving nodes. A new node is added without any order value, and the editor allows the user to specify a specific order value. If preferred, I'll add it to my to-look-at list to see how easy it is to add a 'move' option to the node editor within its current plugin architecture.

Ignoble61:
Note that I've now concluded the basic-coding of a way for the library node editor to 'move' a node (which will adjust all existing nodes to ensure the order is concurrent and unique for the moved node by abusing the select dialog from Python). If this is sufficient - and it will be PR'd against the library node editor towards the end of the week when I've had time to fully test and clean the code - then this PR can now be closed. It's worth noting that it doesn't affect the UX of inserting an item/placing an item at the top of the list when users are manually editing nodes - rather than using the node editor - which remains an issue if this PR is rejected.

phil65:
Thx for adding that functionality to the plug-in, it´s good to have no matter if this PR gets merged or not.
I dont have a real strong opinion on inclusion from this PR, perhaps @Montellese is able to tell us whether we should add some empty spaces to the default library nodes or not.

Ignoble61:
Having adjusted the node editor - which I use myself - I have no strong feelings on whether this is merged or not. All I ask is the team bear in mind that this remains an issue for those who are not using the node editor to customise their nodes, and consider how large an issue this is for those users.

Rechi:
@ Ignoble61 this needs a rebase

DaveTBlake:
The idea of spacing out order id is a good one, but now 18 months later we just need to check that the item order has been maintained by this change, and that no nodes have been ommitted. I have not had time to do that yet

DaveTBlake (approved):
Menu item order checks out, still good to go

PR #11018, opened 2016-11-30
use video disptime only in case video is running

peak3d:
See title

## Description
If Inputstream provides only audio data, the not initialized disptime from video is taken to calculate time offsets wich are displayed later in the UI. This PR fixes the use of this not initialzed value.

## How Has This Been Tested?
Win10 before / after test

## Types of change


peak3d:
@MartijnKaijser its a bug. agreed with @fernetmenta to merge.

PR #11019, opened 2016-11-30
allow absolute item focus

ronie:
allow skins to focus an absolute item number instead of a relative one.
(currently the first visible item is used as an offset, the 'absolute' attribute overrides that)

```SetFocus(9000,0,absolute)```

@phil65 @HitcherUK @BigNoid

### DIFF
--- a/xbmc/interfaces/builtins/GUIControlBuiltins.cpp
+++ b/xbmc/interfaces/builtins/GUIControlBuiltins.cpp
@@ -22,6 +22,7 @@
 
 #include "guilib/GUIWindowManager.h"
 #include "input/ButtonTranslator.h"
+#include "utils/StringUtils.h"
 
 /*! \brief Send a move event to a GUI control.
  *  \param params The parameters.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: xbmc/interfaces/builtins/GUIControlBuiltins.cpp [hunk 1]
```diff
@@ -88,12 +89,14 @@ static int SendMessage(const std::vector<std::string>& params)
...
  *  \details params[0] = ID of control.
  *           params[1] = ID of subitem of control (optional).
+ *           params[2] = "absolute" to focus the absolute position instead of the relative one (optional).
  */
 static int SetFocus(const std::vector<std::string>& params)
 {
   int controlID = atol(params[0].c_str());
   int subItem = (params.size() > 1) ? atol(params[1].c_str())+1 : 0;
-  CGUIMessage msg(GUI_MSG_SETFOCUS, g_windowManager.GetFocusedWindow(), controlID, subItem);
+  int absID = (params.size() > 2) ? StringUtils::EqualsNoCase(params[2].c_str(), "absolute") : 0;
```
garbear: can you make StringUtils::EqualsNoCase() explicitly resolve to 0 or 1, instead of implicitly relying on true == 1?

BigNoid:
+1, however this breaks SetFocus() without extra parameters.

ronie:
> this breaks SetFocus() without extra parameters.

thanx! will fix.

xbmc/interfaces/builtins/GUIControlBuiltins.cpp
ronie: that's way above my paygrade ;-)

you mean (int)StringUtils::EqualsNoCase() ?

HitcherUK:
Does this also fix the problem with fixedlists where the initial focus is always the focus position and not the first item?

xbmc/interfaces/builtins/GUIControlBuiltins.cpp
ronie: ^ @garbear ?

garbear:
```c++
int absID = 0;
if (params.size() > 2 && StringUtils::EqualsNoCase(params[2].c_str(), "absolute"))
  absID = 1;
```

ronie: well.. that's one way to do it ;-)

ronie:
updated.

jenkins build this please

PR #11025, opened 2016-12-01
(Potentially) fix generic touch swipe detector

notspiff:
use fabs(), even if abs() (ie implicit cast to int) probably did the job. But this has the potential to change behavior so carefully consider whether to backport.

this is not an attack: can the windows developers explain what is making them consistently add compiler warnings like the format errors? i assume they drown in other sorts of warnings? can we do something about that so you actually get to see the things that are of interest?

### DIFF
--- a/xbmc/input/touch/generic/GenericTouchSwipeDetector.cpp
+++ b/xbmc/input/touch/generic/GenericTouchSwipeDetector.cpp
@@ -117,8 +117,8 @@ bool CGenericTouchSwipeDetector::OnTouchMove(unsigned int index, const Pointer &
     return false;
   }
 
-  float deltaXabs = abs(pointer.current.x - pointer.down.x);
-  float deltaYabs = abs(pointer.current.y - pointer.down.y);
+  float deltaXabs = fabs(pointer.current.x - pointer.down.x);
+  float deltaYabs = fabs(pointer.current.y - pointer.down.y);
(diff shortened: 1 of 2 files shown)


hudokkow:
ping @Montellese

notspiff:
note: ::abs() is overloaded on windows, so it would be correct there. only clang/gcc would be affected.

Montellese:
@notspiff looks like you answered your own question ;-)

Anyway I just did a clean build on win32 in VS and I got a total of 679 warnings of which 253 are from Kodi code.  Most of them are warnings about type conversions with possible loss of data but yes it makes it harder to spot warnings in newly written code. Here's the list of the 253 warnings: http://pastebin.com/Ljhenqte

MartijnKaijser:
Should there be any updates/changes or can this be merged?

akva2:
maybe something should be done about the comment from garbear (issue is still in master). i can just drop it, but then you better fix the variable type. every time that log statement is executed, memory stomping is performed.

Paxxi:
@notspiff the simple answer is that msvc doesn't complain about format specifiers at our current warning level. With ReSharper it will mark it as an error in the ide but no compile time warning. 

Just for fun I tried upping the warning level and got something like 64k warnings so not really a good idea :) 

All these issues are one of the reasons that I started looking at the fmt and I would guess that it's also why @montellese picked up where I left off

PR #11026, opened 2016-12-01
fixed: do not use move into member variable here

notspiff:
clang points out that it actually hinders copy elusion, not aids it.


hudokkow:
ping @garbear

garbear:
I thought it wouldn't affect copy elision, but if it hinders it then this is preferred.

hudokkow:
jenkins build and merge

PR #11027, opened 2016-12-01
fixed: add appropriate using statements where virtual methods and overloading is mixed

notspiff:
quells a shitload of clang warnings.

for a couple of these methods were marked virtual, but they were not overriding anything. that made the diff a bit larger as i have to select the correct overload explicitly (i have to remove the default param or it gets ambigious with parent method).

### DIFF
--- a/xbmc/addons/Addon.cpp
+++ b/xbmc/addons/Addon.cpp
@@ -167,7 +167,7 @@ bool CAddon::MeetsVersion(const AddonVersion &version) const
  */
 bool CAddon::HasSettings()
 {
-  return LoadSettings();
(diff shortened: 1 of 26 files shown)


hudokkow:
ping @tamland and @Montellese 
jenkins build this please

MartijnKaijser:
@notspiff guess i found the PR i asked somewhere else i think. Could you rebase?

notspiff:
this touch videoplayer, and there is no way not to touch dvdplayer or addon shit cause that's where all the warnings come from. so i'll just close this and the other one to get them out of your eye.

PR #11033, opened 2016-12-02
Fix game add-ons not loading

garbear:
This is the fix I've been using in my RP branch for over a year to fix game add-ons not loading. The problem is that Kodi is trying to load the shared library from `/usr/share/kodi/addons/` instead of `/usr/lib/kodi/addons/`.

This successfully fixes game add-ons not loading, but I have reports from users that it breaks PVR add-ons loading. I figured that I'd post this fix and we could go from there.

## Types of change

### DIFF
--- a/xbmc/addons/Addon.cpp
+++ b/xbmc/addons/Addon.cpp
@@ -30,6 +30,7 @@
 #include "ContextMenuManager.h"
 #include "filesystem/Directory.h"
 #include "filesystem/File.h"
+#include "filesystem/SpecialProtocol.h"
 #include "RepositoryUpdater.h"
 #include "settings/Settings.h"
 #include "ServiceBroker.h"
@@ -331,7 +332,19 @@ std::string CAddon::LibPath() const
 {
   if (m_props.libname.empty())
     return "";
-  return URIUtils::AddFileToFolder(m_props.path, m_props.libname);
+
+  std::string strLibPath = URIUtils::AddFileToFolder(m_props.path, m_props.libname);
+
+  // Check if add-on library has been installed to the binaries path instead
+  std::string strSharePath = CSpecialProtocol::TranslatePath("special://xbmc/");
+  const bool bIsInSharePath = StringUtils::StartsWith(strLibPath, strSharePath);
+  if (bIsInSharePath && !CFile::Exists(strLibPath))
+  {
+    std::string strBinPath = CSpecialProtocol::TranslatePath("special://xbmcbin/");
+    strLibPath.replace(0, strSharePath.length(), strBinPath);
+  }
+
+  return strLibPath;
 }
 
 AddonVersion CAddon::GetDependencyVersion(const std::string &dependencyID) const


wsnipex:
this shouldn't be necessary, we already check /usr/lib/
Also note that on linux it's /usr/lib/x86_64-linux-gnu/kodi/ since a few months.

garbear:
I've added a sample game add-on to our binary add-ons repo: https://github.com/xbmc/repo-binary-addons/pull/42

we'll be able to test when that's merged

xbmc/addons/Addon.cpp
```diff
@@ -331,7 +332,19 @@ std::string CAddon::LibPath() const
...
   if (m_props.libname.empty())
     return "";
-  return URIUtils::AddFileToFolder(m_props.path, m_props.libname);
+
+  std::string strLibPath = URIUtils::AddFileToFolder(m_props.path, m_props.libname);
+
+  // Check if add-on library has been installed to the binaries path instead
+  std::string strSharePath = CSpecialProtocol::TranslatePath("special://xbmc/");
+  const bool bIsInSharePath = StringUtils::StartsWith(strLibPath, strSharePath);
+  if (bIsInSharePath && !CFile::Exists(strLibPath))
```
tamland: calling Exists here is an absolute no. fix this by not setting the wrong path in the first place

garbear: yea, that's why this patch wasn't included in the RP pr. If you have an idea of how to not set the wrong path in the first place, i'd love to hear it.

wsnipex: doesn't https://github.com/xbmc/xbmc/blob/master/xbmc/addons/AddonDll.h#L144-L154 already do the same thing?

tamland: Frankly I'm not sure I understand this whole binary addon issue. Why would the library not be located inside the addon folder?

But if just having `libname` isn't sufficient for binary addons then clearly 'libname' is the wrong solution and can just be removed entirely. Because its only purpose is *here* for resolving LibPath(). More appropriate place to handle this resolving would then be either the [factory](https://github.com/xbmc/xbmc/blob/0bea8ab957a11d1352307efead0bc8c7963275b8/xbmc/addons/AddonManager.cpp#L102) or that^ `LoadDll()`

popcornmix:
I've noticed kodi doesn't seem to be loading game libs with the correct names compared to how they are built. I have to run:
```
cp /opt/xbmc-cmake/arm-linux-gnueabihf/lib/kodi/addons/library.xbmc.addon/libXBMC_addon-arm-linux-gnueabihf.so /opt/xbmc-cmake/arm-linux-gnueabihf/lib/kodi/addons/library.xbmc.addon/libXBMC_addon-arm.so
cp /opt/xbmc-cmake/arm-linux-gnueabihf/lib/kodi/addons/library.kodi.game/libKODI_game-arm-linux-gnueabihf.so /opt/xbmc-cmake/arm-linux-gnueabihf/lib/kodi/addons/library.kodi.game/libKODI_game-arm.so
```
to get them to load. Is ``libKODI_game-arm-linux-gnueabihf.so`` or ``libKODI_game-arm.so`` the correct name? Make system and kodi don't seem to be in sync.

garbear:
Closing, as @wsnipex noted this fix is already present elsewhere

wsnipex:
well, it doesn't work though. Seems there is a difference between game addons and all other binary addons.

garbear:
@MilhouseVH can you leave this out of your builds so we can see if game add-ons still load? assuming you've added the games repo

MilhouseVH:
Sure. I also haven't been including #11034 for the past few days.

garbear:
I saw that, I started replacing the DB with xml files but haven't completed the ability for it to scan for new files. I'll try to merge it within the week, for sure by the end of the year

garbear:
Obsolete. This issue will be addressed in Alwin's binary add-on work eventually.

AlwinEsch:
Not sure if fixed from me ;-), but here at home it works currently :-D.

garbear:
Not on android ;) The current method is hacky, when we load a libretro core we swap the DLL path to game.libretro, and pass game.libretro the original libretro core path. This doesn't work on android because all the .dlls are in a special folder.

AlwinEsch:
Has seen the parts on CAddonDll related to Android, yes there it is a bit hacky. I must go more detailed in this Android to improve them :-)

PR #11036, opened 2016-12-02
Revert "This makes the SubtitleDialog remember the last service used,…

BigNoid:
Reverts #9660 
#9660 introduced remembering last service used, effectively rendering the default tvshow and movie service settings useless.
Also noticed issues where the status of the subtitle dialog was on downloading when opened and focus was lost.

@phil65 @arnova

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -144,7 +144,6 @@
 
 // Dialog includes
 #include "video/dialogs/GUIDialogVideoBookmarks.h"
-#include "video/dialogs/GUIDialogSubtitles.h"
 #include "dialogs/GUIDialogOK.h"
 #include "dialogs/GUIDialogKaiToast.h"
 #include "dialogs/GUIDialogSubMenu.h"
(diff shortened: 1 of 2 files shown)


BigNoid:
jenkins build and merge

arnova:
Well unfortunately you didn't allow me to comment but I think this doesn't overwrite the defaults per se. The defaults are the default after each power cycle, unless the user selects something else that will be the default until the next power cycle. Honestly I'm not sure what the "best" behavior is.

MartijnKaijser:
Well as it was was certainly not correct. Tvshow and movies should be handled separately and not lumped together else you might as well remove the split settings

BigNoid:
I've also encountered unexpected behavior when searching subtitles with
this PR. It is not something we want in final release imo. Reverted it in
master also cause it needs to be done differently when this feature is
wanted.

Op vr 2 dec. 2016 om 17:56 schreef Martijn Kaijser <notifications@github.com

arnova:
Ah yes, didn't realise that. We should indeed remember it separately for tvshow/movies/etc. then

PR #8094, opened 2015-09-19
[xbmc] Change formatting to cppformat instead of sprintf/snprintf style form…

Paxxi:
…atting

Groundwork, still passes most formatting through the old method as the strings aren't updated yet
Updated language addon to use the positional parameters
To avoid packaging troubles and since the library is only 5 files I decided to stick it in contrib as kissfft

PR #11039, opened 2016-12-02
Use {fmt} instead of custom sprintf

Montellese:
This is a rebase of #8094 to use {fmt} (see https://github.com/fmtlib/fmt) instead of custom `sprintf`. It uses python formatting instead of C printf formatting so that's gonna require some getting used to but it has a lot of advantages as well:
* full type safety
* support for `std::string` (no more `.c_str()`)
* positional arguments (especially useful for translations with multiple parameters)

@Paxxi has adjusted all translatable strings but all logging strings etc. still use the old format and are passed through the type safe `fmt::sprintf()`.

As opposed to #8094 I have included {fmt} as an external dependency. I've only done the necessary work for win32 so this needs adjustments for depends. I'm also pretty sure that I didn't do the whole cmake dependency handling right so would be nice if someone could take a look at that.

### DIFF
--- a/tools/depends/target/libfmt/Makefile
+++ b/tools/depends/target/libfmt/Makefile
@@ -1,0 +1,65 @@
+-include ../../Makefile.include
+DEPS = Makefile
+
+# lib name, version
+LIBNAME=fmt
+VERSION=3.0.1
+BASE_URL=https://github.com/fmtlib/fmt/archive
+SOURCE=$(LIBNAME)-$(VERSION)
+ARCHIVE=$(SOURCE).tar.gz
+
+RETRIEVE_TOOL_FLAGS=-Ls --create-dirs -f
+
+CMAKE_OPTIONS=-DFMT_DOC=OFF -DFMT_INSTALL=ON -DFMT_TEST=OFF -DFMT_USE_CPP11=ON
+
+ifeq ($(CROSS_COMPILING), yes)
+  DEPS += ../../Makefile.include
+else
+  CXXFLAGS += -std=c++11
+  ROOT_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+
+  ifeq ($(PLATFORM),)
+    PLATFORM = native
+    TARBALLS_LOCATION = $(ROOT_DIR)
+    RETRIEVE_TOOL := curl
+    ARCHIVE_TOOL := tar --strip-components=1 -xf
+    CMAKE := cmake
+  endif
+endif
+
+LIBDYLIB=$(PLATFORM)/build/$(LIBNAME)/libfmt.a
+
+.PHONY: .installed-$(PLATFORM)
+
+all: .installed-$(PLATFORM)
+$(PREFIX)/lib/lib$(LIBNAME).a:
+	@make .installed-$(PLATFORM)
+
+$(TARBALLS_LOCATION)/$(ARCHIVE):
+	cd $(TARBALLS_LOCATION); $(RETRIEVE_TOOL) $(RETRIEVE_TOOL_FLAGS) -o $(ARCHIVE) $(BASE_URL)/$(VERSION).tar.gz
+
+$(PLATFORM): $(TARBALLS_LOCATION)/$(ARCHIVE) $(DEPS)
+ifeq ($(PREFIX),)
+	@echo
+	@echo "ERROR: please set PREFIX to the kodi install path e.g. make PREFIX=/usr/local"
+	@exit 1
+endif
+	rm -rf $(PLATFORM); mkdir -p $(PLATFORM)
+	cd $(PLATFORM); $(ARCHIVE_TOOL) $(ARCHIVE_TOOL_FLAGS) $(TARBALLS_LOCATION)/$(ARCHIVE)
+	cd $(PLATFORM); rm -rf build; mkdir -p build
+	cd $(PLATFORM)/build; $(CMAKE) $(CMAKE_OPTIONS) ..
+
+$(LIBDYLIB): $(PLATFORM)
+	$(MAKE) -C $(PLATFORM)/build
+
+.installed-$(PLATFORM): $(LIBDYLIB)
+	$(MAKE) -C $(PLATFORM)/build install
+	touch $@
+
+clean:
+	$(MAKE) -C $(PLATFORM)/build clean
+	rm -f .installed-$(PLATFORM)
+
+distclean::
+	rm -rf $(PLATFORM) .installed-$(PLATFORM)
+
--- a/xbmc/rendering/dx/RenderSystemDX.cpp
+++ b/xbmc/rendering/dx/RenderSystemDX.cpp
@@ -32,8 +32,10 @@
 #include "messaging/ApplicationMessenger.h"
 #include "settings/AdvancedSettings.h"
 #include "threads/SingleLock.h"
+#include "utils/CharsetConverter.h"
 #include "utils/MathUtils.h"
 #include "utils/log.h"
+#include "platform/win32/CharsetConverter.h"
 #include "platform/win32/dxerr.h"
 #include "utils/SystemInfo.h"
 #pragma warning(disable: 4091)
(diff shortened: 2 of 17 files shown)


Paxxi:
Thanks for picking up this one. To be clear I started with changing only the translation because it's of most use and to keep work to a minimum for a wip/rfc, we should totally use this for logging as well once it's done

Montellese:
@Paxxi I know that. I actually have more work ready to use spdlog for logging which uses {fmt} internally as well. But I want to do it step by step.

Paxxi:
Perfect, I've tested it out a bit as well and it would certainly be nice to have. Curious to see if the async logging would have any impact on performance critical code.

Montellese:
@Paxxi: see https://github.com/xbmc/xbmc/compare/master...Montellese:spdlog. I haven't looked into async logging yet though.

akva2:
why use this for logging? there's is no security benefit there. it seems like a shitload of work for little gain to me.

Montellese:
What "shitload of work"?
I was looking for a very small and compact logging library and came across `spdlog`. I then tested it inside Kodi and while handling the dependencies I realized that I depends on {fmt}.
You can still use the old log format but you gain type safety which in my book is a huge plus. On win32 you don't get a compiler warning when you try to pass an `std::string` to `CLog::Log()` and the thread running the code with the bad log message will throw an exception and then usually dies silently. That causes a lot of head ache because you're wondering why the rest of your code isn't running.

akva2:
changing all the log statements to use the new syntax rather than the old syntax seems tedious to me. sure, allowing *new code* to use the new syntax makes sense, but rewriting all the old seems to be pain without gain.

Montellese:
Finally got around to looking into the depends integration. I've tested this with an android ARM build but since this is not really my wheelhouse I'd appreciate it if someone (@wsnipex ?) could take a look. I'm also not sure about linux since it probably requires `libfmt3-dev`. Let's see what jenkins says: jenkins build this please

Montellese:
Could jenkins built fine on all platforms. I've taken a look concerning linux system libs. Ubuntu has the package `libfmt3-dev` at http://packages.ubuntu.com/zesty/libfmt3-dev but only for Zesty. @wsnipex can we integrate this into our PPA?

Montellese:
I've added fadcd73ced2ace3739f55cf4c0ff9cf94b682121 which modifies the `libfmt` depends `Makefile` similar to `crossguid` so that it can be built natively on linux.

Paxxi:
@Montellese anything left to do before merging this? The first attempt ended up forgotten and abandoned and I'd hate to see it happen again.

Montellese:
I need to resolve the conflicts and there's no solution for PPA yet but @wsnipex didn't respond to my ping.

Paxxi:
lets hope he sees it this time then :)

wsnipex:
was on holiday when you pinged me, so must have overlooked it in the mess of notifications.
Took some tinkering, but finally got it to build on all our supported ubuntu releases, although <=15.04 lack docs.

Montellese:
@wsnipex: No worries I was on vacation as well in February. So all is good from PPA side? I'll rebase this later today and then we can let jenkins run once more.

Montellese:
jenkins build this please

Montellese:
@Paxxi: had to make another adjustment. Could you please check if https://github.com/xbmc/xbmc/pull/11039/commits/48f6c5310d0d78823662ec0d37662466989ec658#diff-504bcfd2abeae7af84f0ab527d81e08fR355 is correct? There was an issue with formatting an `std::wstring` into an `std::string`.

jenkins build this please

xbmc/rendering/dx/RenderSystemDX.cpp
```diff
@@ -659,7 +660,8 @@ bool CRenderSystemDX::CreateDevice()
     CLog::Log(LOGDEBUG, "%s - on adapter %S (VendorId: %#x DeviceId: %#x) with feature level %#x.", __FUNCTION__, 
                         m_adapterDesc.Description, m_adapterDesc.VendorId, m_adapterDesc.DeviceId, m_featureLevel);
 
-    m_RenderRenderer = StringUtils::Format("%S", m_adapterDesc.Description);
+    std::wstring renderer = StringUtils::Format(L"%s", m_adapterDesc.Description);
+    g_charsetConverter.wToUTF8(renderer, m_RenderRenderer);
```
Paxxi: I'd prefer the use of the FromW/ToW methods in `platform/win32/CharsetConverter.h` as it makes the code easier and simpler with less temporary variables. Not a big deal though.

xbmc/utils/StringUtils.h
```diff
@@ -54,9 +56,26 @@ class StringUtils
   \param ... variable number of value type arguments
   \return Formatted string
   */
-  static std::string Format(PRINTF_FORMAT_STRING const char *fmt, ...) PARAM1_PRINTF_FORMAT;
+  template<typename... Args>
+  static std::string Format(const std::string& fmt, const Args&... args)
```
Paxxi: I think the proper way to handle this is with `Args&&... args` and then calling `fmt::format(fmt, std::forward<Args>(args)...);` to just pass along whatever pointer/const/ref/rvalue specifier that was passed in.

example https://github.com/fmtlib/fmt/blob/master/fmt/format.h#L3950

Paxxi:
The changes for wstring looks sane to me btw, forgot to actually comment on that :)

xbmc/rendering/dx/RenderSystemDX.cpp
Montellese: Will change it. Those methods didn't exist yet when I/you did this ;-)

Paxxi: heh yeah I don't know when I added them but they're simpler to use so I stick to them where it's win32 specific conversions :)

xbmc/utils/StringUtils.h
Montellese: Yeah you're right. That's what you get from years of const-ref'ing everything :-/ This is exactly what universal references are for.

wsnipex:
@Montellese the packages are on the build-deps PPA, but I didn't test your branch with it.

Montellese:
@wsnipex: Thanks. @Paxxi: made the change to use `FromW()` (but there are probably other places as well) and changed the `StringUtils::Format()` wrappers to use universal references. jenkins build this please

Montellese:
Android ARM failure is unrelated and all other builds are fine. Good to go?

phate89:
Do you think it's worth it to start replacing all the different methods we use to format to int with fmt::FormatInt? I was thinking at least here: https://github.com/xbmc/xbmc/blob/master/xbmc/utils/Variant.cpp

Paxxi:
Merge it

Paxxi:
@phate89 yes but not directly with fmt I think, use a wrapper in string utils in case we want to change it someday

phate89:
@Paxxi good point.. also using fmt::Writer should be faster than using format or using stringstreams for single element conversions like we do everywhere..
Maybe a StringUtils:ToString method to overload as we need

xbmc/utils/StringUtils.h
```diff
@@ -54,9 +56,26 @@ class StringUtils
...
+  static std::string Format(const std::string& fmt, Args&&... args)
+  {
+    auto result = fmt::format(fmt, std::forward<Args>(args)...);
+    if (result == fmt)
+      result = fmt::sprintf(fmt, std::forward<Args>(args)...);
+
+    return result;
+  }
+  template<typename... Args>
+  static std::wstring Format(const std::wstring& fmt, Args&&... args)
```
garbear: it looks like this returns a std::string instead of a std::wstring?

Paxxi: No?

Paxxi:
@phate89 I doubt that Perf is worth the extra effort, you're welcome to benchmark it but unless it shows a significant difference the default should be to keep it simple

tools/depends/target/libfmt/Makefile
```diff
@@ -0,0 +1,65 @@
...
+$(PLATFORM): $(TARBALLS_LOCATION)/$(ARCHIVE) $(DEPS)
+ifeq ($(PREFIX),)
+	@echo
+	@echo "ERROR: please set PREFIX to the kodi install path e.g. make PREFIX=/usr/local"
+	@exit 1
+endif
+	rm -rf $(PLATFORM); mkdir -p $(PLATFORM)
+	cd $(PLATFORM); $(ARCHIVE_TOOL) $(ARCHIVE_TOOL_FLAGS) $(TARBALLS_LOCATION)/$(ARCHIVE)
+	cd $(PLATFORM); rm -rf build; mkdir -p build
+	cd $(PLATFORM)/build; $(CMAKE) $(CMAKE_OPTIONS) ..
```
garbear: this line should include the option `-DCMAKE_INSTALL_PREFIX=$(PREFIX)`

wsnipex: not needed, it's already included in ${CMAKE}

wsnipex: ah, but only in the cross compile case, it's missing for native builds

phate89:
@Paxxi excluding for cvariant (if t's worth it) I don't have intention to change any existing stuff but it might nice to have for the future a unified method to do so since we use several different methods to do it but we can't use std::to_string because of android..

xbmc/utils/StringUtils.h
garbear: nvm, didn't notice that fmt::format() was overloaded for wstring

PR #11041, opened 2016-12-03
[gui] return "" instead of "0" for playlist.position in case nothing is playing

phil65:
see title.

### DIFF
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -8422,7 +8422,9 @@ std::string CGUIInfoManager::GetPlaylistLabel(int item, int playlistid /* = PLAY
     }
   case PLAYLIST_POSITION:
     {
-      return StringUtils::Format("%i", g_playlistPlayer.GetCurrentSong() + 1);
+      int currentSong = g_playlistPlayer.GetCurrentSong();
+      if (currentSong > -1)
+        return StringUtils::Format("%i", currentSong + 1);
     }
   case PLAYLIST_RANDOM:
     {


xhaggi:
Looks good, something for Krypton?

phil65:
jenkins build this please.
Should I backport? @ronie @BigNoid

ronie:
seems harmless

BigNoid:
I don't think any skins rely on that value being 0, so backporting wouldn't hurt.

xbmc/GUIInfoManager.cpp
```diff
@@ -8422,7 +8422,9 @@ std::string CGUIInfoManager::GetPlaylistLabel(int item, int playlistid /* = PLAY
     }
   case PLAYLIST_POSITION:
     {
-      return StringUtils::Format("%i", g_playlistPlayer.GetCurrentSong() + 1);
+      int currentSong = g_playlistPlayer.GetCurrentSong();
+      if (currentSong > -1)
+        return StringUtils::Format("%i", currentSong + 1);
```
ksooo: @phil65 <code>break;</code> is missing here. if currentSong is <= -1 execution will fall through PLAYLIST_RANDOM case, which is imo not intended.

Should read:
<pre>
case PLAYLIST_POSITION:
   {
      int currentSong = g_playlistPlayer.GetCurrentSong();
      if (currentSong > -1)
        return StringUtils::Format("%i", currentSong + 1);
      break;
   }
</pre>

ksooo: Causes coverity issue, btw.

PR #11044, opened 2016-12-03
AML: Make DTS-HD, TrueHD work on IEC61937 backport work

fritsch:
This implements IEC mode also for 192 / 8 channels which is for example implemented by AMLogic upcoming firmware. It was succesfully tested by @MartijnKaijser .

It properly checks for 8/192 and if available it enables DTS-HD / TrueHD.

There is a minor EAC3 issue in the firmware only happening with some files, which we reported to them.

Most of the code added is to keep "old" aml stuff working.

This is safe for Krypton.


MartijnKaijser:
Jenkins build and merge

MartijnKaijser:
Jenkins build this please

PR #11045, opened 2016-12-03
Fix upnp iso resume exporting and resume from upnp

phate89:
## Description
Add exporting of playerstate to fix importing of resume point in isos and adds playerstate to upnp to fix resuming of iso via upnp.
@Montellese for review

## Motivation and Context
Right now if we export/import a iso resume point it doesn't work anymore. Same for resume point of a iso via upnp

## How Has This Been Tested?
Tested with 2 local upnp libraries and exporting/importing them once.

## Types of change

## Checklist:


xbmc/video/VideoInfoTag.cpp
```diff
@@ -1187,6 +1197,13 @@ void CVideoInfoTag::ParseNative(const TiXmlElement* movie, bool prioritise)
   {
     XMLUtils::GetDouble(resume, "position", m_resumePoint.timeInSeconds);
     XMLUtils::GetDouble(resume, "total", m_resumePoint.totalTimeInSeconds);
+    const TiXmlElement *playerstate = epbookmark->FirstChildElement("playerstate");
```
Montellese: Something looks wrong here. Shouldn't it be `resume` instead of `epbookmark`?

xbmc/video/VideoInfoTag.cpp
```diff
@@ -1187,6 +1197,13 @@ void CVideoInfoTag::ParseNative(const TiXmlElement* movie, bool prioritise)
...
+    if (playerstate)
+    {
+      const TiXmlElement *value = playerstate->FirstChildElement();
+      if (value)
+        m_EpBookmark.playerState << *value;
```
Montellese: Shouldn't this be `m_resumePoint` instead of `m_EpBookmark` to match the `Save()` implementation?

xbmc/network/upnp/UPnP.cpp
```diff
@@ -249,7 +249,10 @@ class CMediaBrowser : public PLT_SyncMediaBrowser,
             if (time < 0) time = 0;
             curr_value.Append(NPT_String::Format("<upnp:lastPlaybackPosition>%ld</upnp:lastPlaybackPosition>",
                                                  (long)item.GetVideoInfoTag()->m_resumePoint.timeInSeconds));
+            curr_value.Append(NPT_String::Format("<upnp:lastPlayerState>%s</upnp:lastPlayerState>",
```
Montellese: Is `upnp:lastPlayerState` defined in the UPnP specs? If not you need to use another XML namespace like `xbmc:` (which we use in otherplaces).

lib/libUPnP/Platinum/Source/Devices/MediaServer/PltDidl.h
```diff
@@ -123,6 +124,7 @@
 #define PLT_FILTER_FIELD_SERIESTITLE                "upnp:seriesTitle"
 #define PLT_FILTER_FIELD_EPISODE                    "upnp:episodeNumber"
 #define PLT_FILTER_FIELD_LASTPOSITION               "upnp:lastPlaybackPosition"
+#define PLT_FILTER_FIELD_LASTPLAYERSTATE            "upnp:lastPlayerState"
```
Montellese: See my other comment on the XML namespace.

lib/libUPnP/Platinum/Source/Devices/MediaServer/PltMediaItem.h
```diff
@@ -131,6 +131,7 @@ typedef struct {
     NPT_String toc;
     NPT_String user_annotation; //TODO: can be multiple
     NPT_UInt32 last_position;
+    NPT_String last_playerstate;
```
Montellese: Please put all Kodi-specific properties into the `PLT_XbmcInfo` structure because the other ones in `PLT_MiscInfo` are defined in the UPnP spec.

Montellese (changes requested):
What's the format of `playerstate`? Is it a pure string or is it XML or JSON stored in a string?

You certainly need to change the XML namespace of the newly introduced XML tags from `upnp` to `xbmc`.

phate89:
It's an xml stored in a string.. I missed there was the upnp/xbmc distinction.. I'll move them to xbmc

phate89:
@Montellese I updated the pr.. thanks for the review.
 looked at the code and it seems it's always xml the playerstate but I'm not 100% sure..
@FernetMenta maybe you know something more? is playerstate always xml or empty in our videoplayer?

Montellese:
@phate89 If it's XML I have to consult the UPnP specs again. IIRC any XML payload inside the DIDL-Lite elements needs to be encoded for transport but I'm not 100% sure.

lib/libUPnP/Platinum/Source/Devices/MediaServer/PltMediaItem.cpp
```diff
@@ -256,6 +256,7 @@ PLT_MediaObject::Reset()
 
     m_Resources.Clear();
 
+    m_XbmcInfo.last_playerstate      = "";
```
Montellese: Unnecessary spaces.

lib/libUPnP/Platinum/Source/Devices/MediaServer/PltMediaItem.cpp
```diff
@@ -756,7 +764,7 @@ PLT_MediaObject::FromDidl(NPT_XmlElementNode* entry)
     PLT_XmlHelper::GetChildText(entry, "lastPlaybackPosition", str, didl_namespace_upnp);
     if (NPT_FAILED(str.ToInteger(value))) value = 0;
     m_MiscInfo.last_position = value;
-
+
```
Montellese: Unnecessary change.

lib/libUPnP/Platinum/Source/Devices/MediaServer/PltMediaServer.cpp
```diff
@@ -400,7 +400,8 @@ PLT_MediaServer::ParseTagList(const NPT_String& updates, NPT_Map<NPT_String,NPT_
     for (NPT_List<NPT_XmlNode*>::Iterator children = didl_partial->GetChildren().GetFirstItem(); children; children++) {
         NPT_XmlElementNode* child = (*children)->AsElementNode();
         if (!child) continue;
-        tags[child->GetTag()] = *child->GetText();
+        auto txt = child->GetText();
```
phate89: Yes.. gettext returns a pointer that might be null. If we pass to update the playerstate an empty tag (`<xbmc:lastplayerstate></xbmc:lastplayerstate>`) we get a null from gettext. So With `*child->GetText()` we try to retrieve an object from a null pointer

phate89:
@Montellese
If I'm not wrong we already encode the string with `PLT_Didl::AppendXmlEscape`

phate89:
fixed latest comments

lib/libUPnP/Platinum/Source/Devices/MediaServer/PltMediaServer.cpp
```diff
@@ -400,7 +400,8 @@ PLT_MediaServer::ParseTagList(const NPT_String& updates, NPT_Map<NPT_String,NPT_
     for (NPT_List<NPT_XmlNode*>::Iterator children = didl_partial->GetChildren().GetFirstItem(); children; children++) {
         NPT_XmlElementNode* child = (*children)->AsElementNode();
         if (!child) continue;
-        tags[child->GetTag()] = *child->GetText();
```
Montellese: Would you mind splitting this into a separate commit as it's a fundamental fix independent of any Kodi logic.

Montellese:
Moving the generic fix into a separate commit would make upstreaming easier. The rest is fine.

Montellese (approved):

phate89:
@Montellese no problem. I have to split only the commit or also the patch and commit patch?

Montellese:
If you split the commit you also have to redo the patch(es).

phate89:
@Montellese udpated with 2 separate patches
jenkins build this please

MartijnKaijser:
@Montellese that bumps the database.is that something we should still do in beta?

Montellese:
Personally I probably wouldn't backport a database schema change this close to RC.

phate89:
I added the backport needed tag because it's a bug fix and I didn't consider the bump.
The db changes are really small since it only adds a new field in the views..
Otoh it's a bug that's there since forever. I noticed the bug4 months ago looking at the code (not because I actually use this) and I think it's never been reported at all so it's fine by me either way..

MartijnKaijser:
Thanks for fixing it but lets keep db version as is in Krypton :)

PR #11052, opened 2016-12-04
[PVR] CApplication::PlayMedia: Use new PVRGUIActions functionality for playback of PVR channels and recordings.

ksooo:
Mainly refactoring to get rid of duplicated code in PVR, but fixes also PVR channel home screen widget only to switch to fullscreen video instead of re-tuning the channel in case an already playing channel is selected from the widget.

This change has been runtime tested on latest master on LInux and macOS.

@Jalle19 for review?

### DIFF
--- a/xbmc/pvr/PVRManager.cpp
+++ b/xbmc/pvr/PVRManager.cpp
@@ -1228,41 +1228,6 @@ void CPVRManager::CloseStream(void)
   SAFE_DELETE(m_currentFile);
 }
 
-bool CPVRManager::PlayMedia(const CFileItem& item)
-{
-  if (!g_PVRManager.IsStarted())
-  {
-    CLog::Log(LOGERROR, "CApplication - %s PVR manager not started to play file '%s'", __FUNCTION__, item.GetPath().c_str());
-    return false;
-  }
-
-  CFileItem pvrItem(item);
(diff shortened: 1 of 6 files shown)


xbmc/pvr/PVRManager.cpp
```diff
@@ -1228,41 +1228,6 @@ void CPVRManager::CloseStream(void)
...
-
-  if (!pvrItem.HasPVRChannelInfoTag() && !pvrItem.HasPVRRecordingInfoTag())
-    return false;
-
-  // check parental lock if we want to play a channel
-  if (pvrItem.IsPVRChannel() && !g_PVRManager.CheckParentalLock(pvrItem.GetPVRChannelInfoTag()))
-    return false;
-
-  /* copy over resume info from original item */
-  pvrItem.m_lStartOffset = item.m_lStartOffset;
```
Jalle19: What about this logic, has it been been moved somewhere?

ksooo: yes, PVRGUIActions takes care

Jalle19: Thanks

Jalle19 (approved):

ksooo:
jenkins build this please

PR #11053, opened 2016-12-04
AESinkAudioTrack: Use more alsa like settings

fritsch:
In fact most android devices speak ALSA under the hood. the m_frames are directly comparable to periodSize in ALSA wording. Having the periods half the buffer produces huge packages to be added and also the sink's delay can only be estimated as good as half of buffer size. Let's use 1/4 of it.

This is the same what ALSA aims at.

Needs some testing with
a) PT Codecs (DTS, DTS-HD, TrueHD, AC3) (IEC only)
b) FLAC 192/24 and multi channel flac with high samplerate


fritsch:
Cosmetics - no one complained. PA Sink seems stable since a long time (TM)

PR #11054, opened 2016-12-04
[PVR] Fix PVR channel OSD display conditions.

ksooo:
Regression introduced with #10995. Did not work if "channel switch with ok" is deactivated an "channel switch delay" > 0.

nobrainer. :-/


ksooo:
jenkins build this please

ksooo:
Jenkins build this please

PR #11058, opened 2016-12-05
remove deprecated infobools

ronie:
remove deprecated stringcompare infobools.

### DIFF
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -5254,49 +5254,6 @@ int CGUIInfoManager::TranslateSingleString(const std::string &strCondition, bool
       return SYSTEM_ALWAYS_FALSE;
     else if (cat.name == "true" || cat.name == "yes")
       return SYSTEM_ALWAYS_TRUE;
-
-    // deprecated begin
-    // should be removed before L*** v18
-    if (cat.name == "isempty" && cat.num_params() == 1)
-      return AddMultiInfo(GUIInfo(STRING_IS_EMPTY, TranslateSingleString(cat.param(), listItemDependent)));
-    else if (cat.name == "stringcompare" && cat.num_params() == 2)
-    {
-      int info = TranslateSingleString(cat.param(0), listItemDependent);
-      // pipe our original string through the localize parsing then make it lowercase (picks up $LBRACKET etc.)
-      std::string label = CGUIInfoLabel::GetLabel(cat.param(1));
-      StringUtils::ToLower(label);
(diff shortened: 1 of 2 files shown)


ronie:
jenkins build this please

xhaggi (approved):
Looks good

ksooo:
This PR seems to cause serious trouble: http://forum.kodi.tv/showthread.php?tid=298461&pid=2477340#pid2477340

BigNoid:
Estuary still has some deprecated infobools: https://github.com/xbmc/xbmc/blob/master/addons/skin.estuary/xml/DialogSeekBar.xml#L3
https://github.com/xbmc/xbmc/blob/master/addons/skin.estuary/xml/Custom_1109_TopBarOverlay.xml#L29
https://github.com/xbmc/xbmc/blob/master/addons/skin.estuary/xml/Variables.xml#L190

Will do a PR in a minute

AlwinEsch:
This breaks also the skins on pvr.vdr.vnsi.

xbmc/GUIInfoManager.cpp
```diff
@@ -5254,49 +5254,6 @@ int CGUIInfoManager::TranslateSingleString(const std::string &strCondition, bool
       return SYSTEM_ALWAYS_FALSE;
     else if (cat.name == "true" || cat.name == "yes")
       return SYSTEM_ALWAYS_TRUE;
-
-    // deprecated begin
-    // should be removed before L*** v18
-    if (cat.name == "isempty" && cat.num_params() == 1)
-      return AddMultiInfo(GUIInfo(STRING_IS_EMPTY, TranslateSingleString(cat.param(), listItemDependent)));
-    else if (cat.name == "stringcompare" && cat.num_params() == 2)
```
AlwinEsch: This is used on pvr.vdr.vnsi, is there a alternative for a `<visible>StringCompare(Window.Property(menu),osd)</visible>
`

ronie:
then those skins need to be updated as well ;-)

you need to use String.IsEqual() in that case

AlwinEsch:
Thanks a lot! :D

enen92:
@ronie found this comment while going through the infolabels docs:
https://github.com/xbmc/xbmc/blob/master/xbmc/GUIInfoManager.cpp#L6716

Since `IntegerGreaterThan` has been removed, is this still usefull?

ronie:
IntegerGreaterThan() was replaced by Integer.IsGreater(), so i guess the code is still useful..

enen92:
True. Sorry for the noise. I'll update the comment to refer to the correct infobool and avoid future confusion

PR #11061, opened 2016-12-05
fixed: tag updates on shoutcast streams with low meta data intervals

notspiff:
a new tag came in and updated the position before the previous
tag had been processed, meaning the tag never updated after
stream start. fixes some trac ticket.

### DIFF
--- a/xbmc/filesystem/ShoutcastFile.h
+++ b/xbmc/filesystem/ShoutcastFile.h
@@ -64,6 +64,7 @@ protected:
 
   CFileCache* m_cacheReader;
   CEvent m_tagChange;
+  CCriticalSection m_tagSection;
   int64_t m_tagPos;
 };
 }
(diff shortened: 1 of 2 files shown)


FernetMenta:
thanks!
jenkins build this please

notspiff:
i just realized; we probably need a lock on the tag now as we can have update and process happen at the same time and those strings make it dangerous. earlier that couldn't happen (since we assumed we'd set the tag before a new metaint interval hit us), but it can now.

notspiff:
updated.

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

xbmc/filesystem/ShoutcastFile.h
```diff
@@ -64,6 +64,7 @@ class CShoutcastFile : public IFile, public CThread
 
   CFileCache* m_cacheReader;
   CEvent m_tagChange;
+  CCrticalSection m_tagSection;
```
FernetMenta: error: 'CCrticalSection' does not name a type
missing 'i' :)

notspiff:
darn. why can't i ever manage to hide it when i skip steps ;)

FernetMenta:
hehe, let's see if you cheated again :)
jenkins build this please

FernetMenta:
no what idea why jenkins is so lazy. the other 2 are happy, good enough

PR #11062, opened 2016-12-05
[osx/sdl] - fix wrong size in windowed mode

Memphiz:
This fixes a bug we have since osx 10.x i think. When going from Fullscreen to windowed mode the window was all over the place and way bigger then the screen. Distorted rendering was one of the results of this.

## Description
1. Ensure that resize events from SDL which are bigger then screensize are ignored (those might come from some osx HiDIP feature that our SDL doesn't support
2. Reset the window origin after all the resizing is done so that the window not only has the proper size, but also the correct position on the screen.

## How Has This Been Tested?
Tested by me on osx 10.11.6 and 12.1. Also tested be the user who reported the bug here:
http://forum.kodi.tv/showthread.php?tid=298657&pid=2469793#pid2469793

## Types of change

## Checklist:

### DIFF
--- a/xbmc/windowing/WinEventsSDL.cpp
+++ b/xbmc/windowing/WinEventsSDL.cpp
@@ -355,6 +355,15 @@ bool CWinEventsSDL::MessagePump()
       }
       case SDL_VIDEORESIZE:
       {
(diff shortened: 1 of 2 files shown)


Memphiz:
Jenkins build and merge

PR #11065, opened 2016-12-05
Peripherals: Add setting to hide notifications

garbear:
A user has [reported](http://forum.kodi.tv/showthread.php?tid=298891) that a Sony TV running Marshmallow 6.0 continuously connects/disconnects the remote (probably due to the bluetooth bus powering down). As I can't think of any way to detect this, I've added a setting to manually disable the notifications.

notify peripherals

## Types of change

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -16499,7 +16499,19 @@ msgctxt "#35018"
 msgid "Controller configuration depends on a disabled add-on. Would you like to enable it?"
 msgstr ""
 
-#empty strings from id 35019 to 35048
+#. Name of setting for enabling notifications when a peripheral is plugged in
+#: system/settings/settings.xml
+msgctxt "#35019"
+msgid "Notify when device is connected"
+msgstr ""
+
+#. Description of setting with label #35019 "Notify when device is connected"
+#: system/settings/settings.xml
+msgctxt "#35020"
+msgid "Show notification when a device is connected or removed."
+msgstr ""
+
+#empty strings from id 35021 to 35048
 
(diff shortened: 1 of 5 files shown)


MartijnKaijser:
@koying any idea on how to do this without adding this option?

koying:
Would discarding all pseudo-joystick without an analog axis be a solution?

garbear:
No, that would break nes, SNES and genesis controllers. we discard joysticks with more than 100 or 200 buttons. I could ask for a debug log and see if there's any quirks we could look for

garbear:
Superseded by #11162

garbear:
I'm reopening this for consideration. In the last week, enough cases have come forward to warrant a new setting. In addition to the reconnecting of the sony remote, a Rapoo keyboard has been showing frequent notifications, and some CEC adapters take a while to connect, leading to the notification possibly being shown during video playback.

garbear:
Any objections? I plan to change the text to "Show device events" or "Show device notifications"

MartijnKaijser:
Some things:
1. I also find it quite annoying that each time i reconnect my SATV controller i see the popup as i use it as remote but it auto sleeps
2. We shouldn't show notifications on top of playing video i guess?
I'd be fine with default off.

garbear:
if we're going to default off, why don't we just omit the setting and kill connect/disconnect notifications all together?

We could also leave just the sound by default, with an option to turn that off

MartijnKaijser:
Well it's a though call cause it differs what use case you have. 
When doing normal Kodi usage like video the popups could become bit much. When doing gaming you would actually want to know if new controller gets hooked up and what player it is?

garbear:
Superseded by #11279 and #11329

PR #11022, opened 2016-11-30
Inconspicuous fix

garbear:
I mean, this bug has been around for a really long time. I blame the developers who ported XBMC away from the Xbox. Seriously, who uses Windows, Linux, OSX, Android, iOS, Raspberry Pi, FreeBSD, Fire Stick, or any Chinese box? The original linux developers should have known that the lack of games would be Kodi's longest standing bug, lasting over 9 1/2 years, and just given up. All it took to add games back was a minor fix of tens of thousands of lines of code. Some people are so lazy...

## Description
A small fix, really.

## Motivation and Context
I've been motivated by all the people who've contributed code:

* Game info tags: fetzerch for fixing a PVR-related crash
* Game add-ons: Themaister for rewind functionality, fetzerch for mouse support, file length check and cmake modifications, topfs2 for fixing a crash when loading game clients, eibma for fixing linux compilation errors, a1rwulf for catching a missing callback symbol and fixing some rebase errors, notspiff for helping with the rebrand and cmake, hudokkow for updating libretro cores, and lrusak for updating libretro cores and LibreELEC builds.
* RP: elpendor for RGB565 support, poisson for RAII improvements, ChrisMyhre for catching a compile error, notspiff for CMake fixes, acmiyaguchi for video and audio codec support, and popcornmix for Raspberry Pi support.

## How Has This Been Tested?
Lots of drunken Mario Kart.

## Types of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #11066, opened 2016-12-05
Fix coverty defects

garbear:
This fixes 3 coverty defects detected in the RP pr (#11022). The third fix also updates CGameInfoTag to fix missing fields and the setting of the loaded state.

CGameInfoTag was the first code I wrote for RP back in 2012, and at the time it was capable of parsing info directly from the ROM data for NES, SNES and GBA. The parsing has since been removed from kodi and added to [PyRomInfo](https://github.com/garbear/pyrominfo), which I plan to use when we implement python-based game scrapers. As such, CGameInfoTag may need to be changed again when game info tags are actually used. Some GameInfoTag fields are exposed to Python in the commit 844912f from #11022.

## Types of change


garbear:
jenkins build this please

PR #11072, opened 2016-12-06
[cmake] Add support for LCMS2

hudokkow:
## Description
Add missing icc support to color management.

## Motivation and Context
Fix http://trac.kodi.tv/ticket/17106

## How Has This Been Tested?
Built and briefly runtime tested with and without liblcms2 present in system.

## Types of change

## Checklist:

ping @fetzerch and @wsnipex

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,7 @@ install:
       sudo apt-get install -qq automake autopoint build-essential cmake curl default-jre gawk gdb gdc
       gettext git-core gperf libasound2-dev libass-dev libbz2-dev libcap-dev libcdio-dev libcec4-dev libcrossguid-dev libcurl3
       libcurl4-openssl-dev libdbus-1-dev libfmt3-dev libfontconfig-dev libegl1-mesa-dev libfreetype6-dev libfribidi-dev libgif-dev
-      libiso9660-dev libjpeg-dev libltdl-dev liblzo2-dev libmicrohttpd-dev libmysqlclient-dev libnfs-dev
+      libiso9660-dev libjpeg-dev liblcms2-dev libltdl-dev liblzo2-dev libmicrohttpd-dev libmysqlclient-dev libnfs-dev
       libpcre3-dev libplist-dev libpng-dev libpulse-dev libsdl2-dev libsmbclient-dev libsqlite3-dev libssh-dev
       libssl-dev libtag1-dev libtinyxml-dev libtool libudev-dev libusb-dev libva-dev libvdpau-dev
       libxml2-dev libxmu-dev libxrandr-dev libxrender-dev libxslt1-dev libxt-dev libyajl-dev mesa-utils
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,7 +115,7 @@ if(CORE_SYSTEM_NAME STREQUAL android)
 endif()
 
 # Optional dependencies
-set(optional_deps MicroHttpd MySqlClient SSH XSLT
+set(optional_deps LCMS2 MicroHttpd MySqlClient SSH XSLT
                   Alsa UDEV DBus Avahi MDNS SmbClient CCache
                   PulseAudio VDPAU VAAPI Bluetooth CAP Python)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,10 +101,21 @@ endif()
 find_package(Threads REQUIRED QUIET)
 list(APPEND DEPLIBS ${CMAKE_THREAD_LIBS_INIT})
 
-# Required dependencies
-set(required_deps Sqlite3 FreeType PCRE Cpluff LibDvd
-                  TinyXML Yajl Cdio Fmt
-                  Lzo2 Fribidi TagLib FFMPEG CrossGUID)
+# Required dependencies. Keep in alphabetical order please
+set(required_deps Cdio
+                  Cpluff
+                  CrossGUID
+                  FFMPEG
+                  Fmt
+                  FreeType
+                  Fribidi
+                  LibDvd
+                  Lzo2
+                  PCRE
+                  Sqlite3
+                  TagLib
+                  TinyXML
+                  Yajl)
 if(NOT WIN32)
   list(APPEND required_deps ZLIB)
 else()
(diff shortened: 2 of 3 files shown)


project/cmake/modules/FindLCMS2.cmake
```diff
@@ -0,0 +1,47 @@
...
+if(PKG_CONFIG_FOUND)
+  pkg_check_modules(PC_LCMS2 lcms2 QUIET)
+endif()
+
+find_path(LCMS2_INCLUDE_DIR NAMES lcms2.h
+                            PATHS ${PC_LCMS2_INCLUDEDIR})
+find_library(LCMS2_LIBRARY NAMES lcms2 liblcms2
+                           PATHS ${PC_LCMS2_LIBDIR})
+
+set(LCMS2_VERSION ${PC_LCMS2_VERSION})
```
wsnipex: will be emtpy if pkg_config is not used

wsnipex:
do we need lcms in depends as well?

project/cmake/modules/FindLCMS2.cmake
hudokkow: humm, than we have problems with other modules. I used https://github.com/xbmc/xbmc/blob/master/project/cmake/modules/FindFribidi.cmake as base.

wsnipex: right, I've skimmed over it too fast. LCMS2_VERSION isn't in REQUIRED_VARS, so it will just be empty.
But we generally have a few find modules that still need work.

hudokkow:
Don't know. Do we want to enforce it?

wsnipex:
depends if all platforms can use color correction or just linux.

project/cmake/modules/FindLCMS2.cmake
hudokkow: Yeap.

hudokkow:
That's above my pay-grade. Who's the main dev? fernet?

FernetMenta:
Currently this is implemented for OpenGL. OSX can use it too once it's profile has been upgraded.

wsnipex:
Do we want it on OSX? If yes, we need to provide the lib in depends

project/cmake/modules/FindLCMS2.cmake
fetzerch: As long as we don't request specific versions, it doesn't cause any harm. The advantage is that fphsa below will show the version number if it's set. (and just omits it if it wasn't found through pkgconfig)

FernetMenta:
> Do we want it on OSX?

yes we do. I hope migration to shader based pipeline on OSX will be completed soon. After that we can bump GL version on OSX.

hudokkow:
Apparently we have some core problems with ICC profiles: https://github.com/LibreELEC/LibreELEC.tv/pull/1038

arucard21:
I use the ICC Profile with Colour Management on my Linux installation without any problems. I noticed that the problem in LibreELEC occurs with the lcms2 library, and I'm currently using the 2.7 version whereas the crash log shows that they're using the 2.8 version. 

Just looking at the code of these 2 versions, there do seem to be some substantial changes in the code, especially in the underlying methods used (the xform method). It seems to use something called "strides" now (no idea what that is though). Here is a full comparison of the changes in the source file between 2.7 and 2.8:
https://github.com/mm2/Little-CMS/compare/lcms2-2.7...lcms2.8#diff-c9c81367c364178743f2533cc6dfd3a3

I believe this is the relevant commit where most of these changes seem to be made:
https://github.com/mm2/Little-CMS/commit/a7fc6e39ceeeac8bdbd24f08ebb08330d15c5d33

I don't know what these changes in the library actually do, but I think it may have changed the behaviour of the cmsDoTransform method in way that could cause this problem or maybe even have introduced a bug there. I hope this information will help solve the problem.

hudokkow:
Thx for taking the time to investigate.
@lrusak ^^

arucard21:
I just tested whether the new version really does break the color correction, but it doesn't seem like it does. On Debian Unstable, I upgraded liblcms2-dev from 2.7-1 to 2.8-2 and Kodi still plays everything correctly with color correction. I tried turning color management on and off while playing a video (as described in the LibreELEC PR), but that works correctly too. I looked at the patches applied to liblcms2 on Debian (https://sources.debian.net/patches/lcms2/2.8-2/) and they don't seem relevant to this problem. 

I think a more consistent reproduction recipe would be needed here.

arucard21:
I think I managed to reproduce the problem. It seems to occur when you use the ICC mode but don't configure an actual ICC profile file. 

Reproduction recipe: 
* Enable Colour Management
* Choose to use ICC instead of 3D LUT
* Don't select an ICC Profile file
* Play a video 
* Kodi crashes (Segmentation fault) on thread 1 in cmsDoTransform()

The obvious solution is to actually select an ICC profile when you configure that mode in Colour Management. Once I selected my ICC profile, it all worked correctly. Perhaps this can be enforced so you always have to provide an ICC profile when you use that mode in Colour Management.

FernetMenta:
@laurimyllari ping ^^

laurimyllari:
I'll take a look. Any suggestions on how the UI should work are welcome.

Maybe the mode and filename should be required before CMS can be turned on?

arucard21:
I looked at some other places in the Settings where similar behaviour is implemented and I think there are 2 ways for this to work:

1. You could check the settings and disable Colour Management if it's not configured correctly
    * This should require few changes to the implementation but may cause people to think Colour Management is working (because it's enabled) when it's actually not (because it's mis-configured). This could be covered by providing a warning in the OSD menu for Colour Management or just showing it as disabled in the OSD when it's mis-configured (allowing the used to fix the configuration right there). 
2. You could add the filename setting to the popup dialog for choosing a Colour Management mode. 
    * I think that this would technically make it an entirely different kind of popup (not a list choice, but an actual settings dialog). But it would force the user to choose a filename before the "OK" button is enabled. 
    * A benefit of this is that the dialog used here can actually be the same as the dialog used in the OSD (if technically possible). You can just put all settings in that dialog so it gets configured in one go. This also ensures that the OSD settings page is always the same as the System settings page (which currently it isn't, as the filename setting is missing in the OSD and the table size seems to use a different list of values)

I would personally prefer the second option as it feels the most user-friendly (but I'm not an expert in these things). Either way though, I think the discrepancies between the OSD and System settings I mentioned in the second option,  should also be resolved.

MartijnKaijser:
any update?

hudokkow:
Nope?

arucard21:
Shouldn't the change in how the UI looks and behaves for Colour Management be complete separate from whether or not it should be included in cmake?

I think this could be merged and the UI and behaviour fixes for Colour Management can be done independently. Currently, Colour Management works just fine, the problems in the LibreELEC issue were due to improper configuration.

Of course, the question of which platforms this should apply to is still open.

hudokkow:
jenkins build this please

wsnipex:
lets get this in. Adding the lib to depends can be done once needed

PR #11079, opened 2016-12-06
FIX: [rar] deadlock with cbr + net vfs (credit notspiff)

MartijnKaijser:
## Description

from notspiff:

specifically we are called off-thread with the main thread waiting on us while holding the gui lock. so it will deadlock on obtaining dialog. the bt was funkily confusing though, as it will actually show a "downstream" deadlock on the rar manager, as there's two concurrent calls to it (to preload/load images). the deadlock that was triggered in the debugger was the second call to the rar manager spinning on the rar manager lock, which is held by the first call to rar manager, which again is spinning on the gui lock.

oh, and using a "exotic" vfs was only part of the problem to the extent that it's slower so the timeout on the progress dialog kicked in.

### DIFF
--- a/xbmc/filesystem/RarManager.cpp
+++ b/xbmc/filesystem/RarManager.cpp
@@ -19,6 +19,7 @@
  */
 
 #include "system.h"
+#include "Application.h"
 #include "RarManager.h"
 #include "Util.h"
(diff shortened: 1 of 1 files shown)


MartijnKaijser:
jenkins build this please

PR #11083, opened 2016-12-06
Explicitly check/find libva-drm in CMake

BrandonSchaefer:
Explicitly check/find libva-drm

## Description
As of right now we implicitly get -lva-drm added to the building of xbmc. We should explicitly find it in case others are building their own toolchains.

## Motivation and Context
This fixes issues of where -lva-drm is not being found when building through cmake.

## How Has This Been Tested?
Everything still compile for me and Neil MacLeod tested this patch against his build which was failing before with out it.

## Types of change

## Checklist:

### DIFF
--- a/project/cmake/modules/FindVAAPI.cmake
+++ b/project/cmake/modules/FindVAAPI.cmake
@@ -24,6 +24,8 @@ find_library(VAAPI_libva_LIBRARY NAMES va
                                  PATHS ${PC_VAAPI_libva_LIBDIR})
 find_library(VAAPI_libva-x11_LIBRARY NAMES va-x11
                                      PATHS ${PC_VAAPI_libva_LIBDIR})
+find_library(VAAPI_libva-drm_LIBRARY NAMES va-drm
+                                     PATHS ${PC_VAAPI_libva_LIBDIR})
 
 if(PC_VAAPI_libva_VERSION)
   set(VAAPI_VERSION_STRING ${PC_VAAPI_libva_VERSION})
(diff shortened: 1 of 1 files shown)


project/cmake/modules/FindVAAPI.cmake
```diff
@@ -39,27 +41,33 @@ endif()
...
+  endif()
   if(NOT TARGET VAAPI::VAAPI)
     add_library(VAAPI::VAAPI UNKNOWN IMPORTED)
     set_target_properties(VAAPI::VAAPI PROPERTIES
                                        IMPORTED_LOCATION "${VAAPI_libva_LIBRARY}"
                                        INTERFACE_INCLUDE_DIRECTORIES "${VAAPI_INCLUDE_DIR}"
                                        INTERFACE_COMPILE_DEFINITIONS HAVE_LIBVA=1
-                                       INTERFACE_LINK_LIBRARIES VAAPI::VAAPI_X11)
+                                       INTERFACE_LINK_LIBRARIES VAAPI::VAAPI_X11
+                                       INTERFACE_LINK_LIBRARIES VAAPI::VAAPI_DRM)
```
wsnipex: Docs state:
>This property contains the list of transitive link dependencies.

so this should work:
INTERFACE_LINK_LIBRARIES "VAAPI::VAAPI_X11 VAAPI::VAAPI_DRM"

BrandonSchaefer: That makes much more sense, thanks!

wsnipex:
jenkins build this please

PR #11085, opened 2016-12-07
App Name for Windows OS is not reflected at APPDATA directory.

pjmavadiya:
While creating a build with different app name for Windows OS, new app name is not reflecting at APPDATA directory. In order to change new app name we need to change name at this location manually .


hudokkow:
Thx.
jenkins build this please

PR #11086, opened 2016-12-07
support for image decoder add-ons

notspiff:
This is my work on support for image add-ons. The main motivation for this is bringing back libraw support that was lost when cximage was kicked out.

i have confirmed that decode works. you find addons in my github account (raw only for now). if you want to try the raw stuff https://www.rawsamples.ch/index.php/en/canon is convenient.


notspiff:
@AlwinEsch hope this doesn't create too much problems for your refactoring effort! it should be straight forward to move to a new API like other stuff. of course, it means more work, but that's the cost of progress ;)

hudokkow:
Probably should have a new component? kodi-imageencoder-dev? Something for later.

xbmc/addons/kodi-addon-dev-kit/include/kodi/kodi_imageenc_types.h
```diff
@@ -0,0 +1,82 @@
...
+  {
+    int dummy;
+  };
+
+  struct IMAGEENC_PROPS
+  {
+    const char* mimetype;
+  };
+
+  struct ImageEncoder
```
AlwinEsch: Can you change `struct ImageEncoder` to `typedef struct KodiToAddonFuncTable_ImageEncoder`.
Is to have a standardized start name of function tables on interface ([example](https://github.com/FernetMenta/kodi-agile/blob/master/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h#L533). 

Related to the add-on system rework.

notspiff: i certainly can. but i still need feedback on what to do with thumbs before i do more work on this PR, as it potentially affects naming of everything (if thumbs are only done by core, it makes the add-ons image decoders only).

notspiff:
due to the underwhelming feedback i have taken a few executive decisions:
- no support for thumbnail creation in add-ons. decode / encode cycle suffices for these (was only ever relevant for jpeg)
- the factor isn't hot enough to be worth complicating. i listed a dir with 1000 raw's and the time to instance decoders was utterly irrelevant.
- kodi-image-dev is a separate package

i also rebased for the new add-on stuff (struct table and such).

as far as i am concerned this is ready for merge.

hudokkow:
CMake changes look good to me. Can't comment on the rest.
You might want to update the wording in kodi-image-dev.txt.in to match https://github.com/xbmc/xbmc/pull/11370/files

Happy New Year!

xbmc/windowing/X11/WinSystemX11.cpp
```diff
@@ -911,7 +911,8 @@ bool CWinSystemX11::CreateIconPixmap()
   gRatio = vis->green_mask / 255.0;
   bRatio = vis->blue_mask / 255.0;
 
-  CBaseTexture *iconTexture = CBaseTexture::LoadFromFile("special://xbmc/media/icon256x256.png");
+  CBaseTexture *iconTexture = CBaseTexture::LoadFromFile("special://xbmc/media/icon256x256.png",
+                                                         0, 0, "image/png");
```
fritsch: Why is this needed? (That's my only question for that PR)

fritsch:
@AlwinEsch as you are currently reworking binary addon interfaces, please also have a look.

Thanks very much for doing that work. Very nice and non intrusive.

xbmc/windowing/X11/WinSystemX11.cpp
notspiff: it might not be required any longer, i'll check. i suspect it's a leftover from my ffmpeg image stuff that you adopted.

notspiff: indeed. icon shows fine without. removed.

notspiff:
packaging comment updated.

Paxxi:
If there's no objections, rebase and merge?

hudokkow:
No objections from me.

hudokkow:
@AlwinEsch, any comments or can this be rebased and merged?

MartijnKaijser:
can we just move forward or do we need to wait?

xbmc/addons/AddonBuilder.cpp
```diff
@@ -136,6 +138,8 @@ std::shared_ptr<IAddon> CAddonBuilder::Build()
       return CAudioEncoder::FromExtension(std::move(m_props), m_extPoint);
     case ADDON_AUDIODECODER:
       return CAudioDecoder::FromExtension(std::move(m_props), m_extPoint);
+    case ADDON_IMAGEDECODER:
+      return CImageDecoder::FromExtension(std::move(m_props), m_extPoint);
```
AlwinEsch: There is maybe also the add to `FromProps¸ needed. Normally is the FromProps used for Repository contents and I'm think `CAddon` is enough, but completely not sure.

xbmc/guilib/TextureFormats.h
```diff
@@ -0,0 +1,34 @@
...
+#define XB_FMT_UNKNOWN     0
+#define XB_FMT_DXT1        1
+#define XB_FMT_DXT3        2
+#define XB_FMT_DXT5        4
+#define XB_FMT_DXT5_YCoCg  8
+#define XB_FMT_A8R8G8B8   16 // texture.xbt byte order (matches BGRA8)
+#define XB_FMT_A8         32
+#define XB_FMT_RGBA8      64
+#define XB_FMT_RGB8      128
+#define XB_FMT_OPAQUE  65536
```
AlwinEsch: Also not the best to have a Kodi include on add-on headers, but to prevent compatibility faults and have the ___dev-kit___ includes complete is for me the better place there.

Another way (but a bit Balla-Balla) is to add a Translator and use a own header for add-on system. 
```cpp
int TextureFormatAddonToKodi(int a)
{
  int b = XB_FMT_UNKNOWN;
  if (a & ADDON_XB_FMT_UNKNOWN)
    b |= XB_FMT_UNKNOWN;
  ...
}
```
And the other way.

Background from me is to hold on end add-on's so much independent from Kodi as possible.

AlwinEsch:
On the whole it is very good and nice.

xbmc/guilib/TextureFormats.h
AlwinEsch: You can leave it currently like them on you request. In my rework steps, where still a lot is to do I can add something like `TextureFormatAddonToKodi`. Still want to cleanup the whole `addons` source folder of Kodi ;-) but is very, very big work.

notspiff: I can do the translation layer but making guilib depend on addon bindings is worse than the other way around so flipping the situation is unacceptable imo.

xbmc/addons/AddonBuilder.cpp
notspiff: There is no such method, at least there was none when i updated this last time. I had to *remove* FromProps in that update.

xbmc/guilib/TextureFormats.h
AlwinEsch: Yea, is a bit hard, with a translation layer is for me the best for things like this.

Another background is that when a header is changed in Kodi, all add-ons that use it are affected. Unfortunately, this can be overlooked and compiled add-ons do not work properly anymore.

However, is my point of view and do not know what the other thinking.

notspiff:
well, my reasoning is simply that guilib can be used without add-on support
right now, the other way not so much. you will have your layer as soon as i
find some time (perhaps as early as tomorrow morning if the scraper fix in
my backlog doesn't eat the whole train ride).

xbmc/addons/ImageDecoder.cpp
```diff
@@ -0,0 +1,81 @@
...
+  return m_image != nullptr;
+}
+
+bool CImageDecoder::Decode(unsigned char* const pixels, unsigned int width,
+                           unsigned int height, unsigned int pitch,
+                           unsigned int format)
+{
+  if (!Initialized())
+    return false;
+
```
AlwinEsch: A call like `unsigned int TextureFormatKodiToAddon(unsigned int kodiFormat)` here, does not direct affect the GUI lib.

notspiff:
no, but sticking the header in add-on bindings, then having to include it
in guilib instead would!

notspiff:
branch rebased, translation layer added.

notspiff:
everybody seems happy, can we please merge so i don't have to rebase this again.

Paxxi:
If anything breaks, I never touched the button! 😀

AlwinEsch:
Have tested it and works nice.

Found by chance and test a good picture :D :
bildschirmfoto vom 2017-02-08 16-54-35

fritsch:
Thank you very much!

Am 08.02.2017 5:09 nachm. schrieb "Alwin Esch" <notifications@github.com>:

Paxxi:
I completely forgot about it, threw together a quick build for libraw
https://github.com/Paxxi/LibRaw builds fine but untested

notspiff:
pushed a dependency definition in the add-on. untested but should be able kick an add-on build on jenkins to check thinngs.

MartijnKaijser:
running now 
http://jenkins.kodi.tv/job/BuildMulti-All/1908/

notspiff:
damn close but somebody else keeps the sigar. mismatch in config file name. it expects RAW-config.cmake, not libraw-config.cmake. needs to be adjusted in either add-on or your repo (idc which).

Paxxi:
I can fix it tomorrow, you want the whole project/target to be just raw or config enough?

notspiff:
only care about config file, it needs to match the find rule in the add-on which is currently FindRAW.cmake

Paxxi:
I renamed everything to just raw, would likely just cause issues in the future having different names

PR #11090, opened 2016-12-07
Workaround the skin unload bug - force close TextureBundle

ace20022:
The migration step and the unloadSkin meassage/action yield a deadlock.

### DIFF
--- a/xbmc/addons/Skin.cpp
+++ b/xbmc/addons/Skin.cpp
@@ -369,6 +369,9 @@ const INFO::CSkinVariableString* CSkinInfo::CreateSkinVariable(const std::string
 
 void CSkinInfo::OnPreInstall()
 {
+  bool skinLoaded = g_SkinInfo != nullptr;
+  if (IsInUse() && skinLoaded)
+    CApplicationMessenger::GetInstance().SendMsg(TMSG_EXECUTE_BUILT_IN, -1, -1, nullptr, "UnloadSkin");
 }
 
 void CSkinInfo::OnPostInstall(bool update, bool modal)
--- a/xbmc/guilib/TextureBundle.cpp
+++ b/xbmc/guilib/TextureBundle.cpp
@@ -78,6 +78,12 @@ int CTextureBundle::LoadAnim(const std::string& Filename, CBaseTexture*** ppText
   return 0;
 }
 
+void CTextureBundle::Close()
+{
+  if (m_useXBT)
+    m_tbXBT.CloseBundle();
+}
+
 void CTextureBundle::SetThemeBundle(bool themeBundle)
 {
   m_tbXBT.SetThemeBundle(themeBundle);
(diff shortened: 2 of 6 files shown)


xbmc/addons/Skin.cpp
```diff
@@ -359,7 +359,8 @@ int CSkinInfo::GetFirstWindow() const
 bool CSkinInfo::IsInUse() const
 {
   // Could extend this to prompt for reverting to the standard skin perhaps
-  return CSettings::GetInstance().GetString(CSettings::SETTING_LOOKANDFEEL_SKIN) == ID();
+  bool active = GetRunningInstance() != nullptr;
```
tamland: GetRunningInstance is a binary addon hack. For all other addons it's always null

tamland:
maybe i missed something, but isn't this identical to #10990?

MartijnKaijser:
* Upgrade towards v17 works
* updating a skin in use in v17 doesn't work.
http://paste.ubuntu.com/23594578/

ace20022:
damn, forgot to check the original case. Is there a way to check if a skin is loaded/active?

MartijnKaijser:
Tests performed:
* Upgrade to v17: OK
* Update skin in use: OK
* Update skin using a non default theme: OK

MartijnKaijser:
Jenkins build this please 
If others could test this change as well if they see negative side effects that would be great.

MartijnKaijser:
@tamland @Paxxi @fritsch any objections?

xbmc/addons/Skin.cpp
```diff
@@ -369,6 +369,9 @@ const INFO::CSkinVariableString* CSkinInfo::CreateSkinVariable(const std::string
 
 void CSkinInfo::OnPreInstall()
 {
+  bool skinLoaded = g_SkinInfo != nullptr;
```
tamland: Not sure what this supposed to do (?) g_SkinInfo will never be null except during startup, before the first skin is loaded

tamland: And as mentioned in #10990 can we please add a comment here that says what the UnloadSkin is for? Because it makes no sense unless you already know it's a workaround for closing TextureBundle on windows.

tamland:
Other than a comment, no objections. But just to be clear: as far as I can tell it's still exactly the same as the previous patches. So if it didn't work, but now works, I think something is up with your tests martijn...

MartijnKaijser:
exact test before

xbmc/addons/Skin.cpp
MartijnKaijser: Well previous patch had endless loop when upgrading the skin from v14 to v17 so isnt that the case here?

MartijnKaijser:
@ace20022 thx.
Can you also PR it to master and include an inline comment about the reload skin?

ace20022:
But imo it's really a workaround and should be done properly for master. Maybe as a interim solution?

MartijnKaijser:
imo yes as interim solution with big fat warning unless some one already knows how to solve it properly.

PR #11094, opened 2016-12-08
Alternative: Workaround the skin unload bug - force close TextureBundle

ace20022:
Alternative for https://github.com/xbmc/xbmc/pull/11090


MartijnKaijser:
Tests performed:
* Upgrade to v17: OK
* Update skin in use: OK
* Update skin using a non default theme: FAIL

PR #10883, opened 2016-11-07
[gui] allow implicit container referencing for non-media windows

phil65:
At the moment implicit referencing of containers only works in MediaWindows. This PR also allows implicit referencing for non-media windows by choosing the actual focused container. That way common constructs like 
```
<value condition="!String.IsEmpty(Container(5400).ListItem.Art(fanart)) + Control.HasFocus(5400)">$INFO[Container(5400).ListItem.Art(fanart)]</value>
<value condition="!String.IsEmpty(Container(6100).ListItem.Art(fanart)) + Control.HasFocus(6100)">$INFO[Container(6100).ListItem.Art(fanart)]</value>
<value condition="!String.IsEmpty(Container(6200).ListItem.Art(fanart)) + Control.HasFocus(6200)">$INFO[Container(6200).ListItem.Art(fanart)]</value>
...
```
can get reduced to
```
<value condition="!String.IsEmpty(Container.ListItem.Art(fanart))">$INFO[Container.ListItem.Art(fanart)]</value>
```

@ronie @BigNoid @HitcherUK for opinions.
@xhaggi for review

PR #11096, opened 2016-12-09
properly check if a python window is a media window

ronie:
currently implicit container referencing (https://github.com/xbmc/xbmc/pull/10883) fails for python windows.
(and likely other items depending on IsMediaWindow() don't work either)

while a python window can be a media window, in many cases it is not.

this adds a check by evaluating the number of views defined in the xml file.

@tamland @phil65

### DIFF
--- a/xbmc/interfaces/legacy/WindowXML.cpp
+++ b/xbmc/interfaces/legacy/WindowXML.cpp
@@ -278,6 +278,14 @@ namespace XBMCAddon
       return A(m_viewControl.GetCurrentControl());
     }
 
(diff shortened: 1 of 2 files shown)


phil65:
I think the most elegant solution would be to allow python authors to choose between mediawindow and non-mediawindow when instanciating the window object (either by adding a param to constructor or by adding / inheriting another class which overrides IsMediaWindow()).
I would be fine with this solution too for now though. We could still improve it later on.

ronie:
jenkins build this please

ronie:
jenkins build this please

PR #11097, opened 2016-12-09
ListItems: Rename "emulator" property to "gameclient"

garbear:
From commit description:

This change apparently got dropped in a RetroPlayer rebase. The reason is because gameclient is more generic than emulator, as the game might be a launchable asset instead of a ROM. Also, I would expect "emulator" to be a human-readable string, e.g. "PCSX Re-armed", whereas "gameclient" is more obviously an add-on ID.

I wanted to get this in ASAP before add-ons start using this value.

### DIFF
--- a/xbmc/interfaces/legacy/ListItem.h
+++ b/xbmc/interfaces/legacy/ListItem.h
@@ -640,7 +640,7 @@ namespace XBMCAddon
       /// | developer     | string (Square)
       /// | overview      | string (Long Description)
       /// | year          | integer (1980)
-      /// | emulator      | string (game.libretro.fceumm)
+      /// | gameclient    | string (game.libretro.fceumm)
       ///
       ///
       ///-----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


xbmc/interfaces/legacy/ListItem.h
```diff
@@ -640,7 +640,7 @@ namespace XBMCAddon
       /// | developer     | string (Square)
       /// | overview      | string (Long Description)
       /// | year          | integer (1980)
-      /// | emulator      | string (game.libretro.fceumm)
+      /// | gameclient    | string (game.libretro.fceumm)
```
MartijnKaijser: Please also add
`@python_v18 Added new  types **game** and associated infolabels.`
or something few lines lower

garbear: done

garbear:
jenkins build and merge

Hedda:
@garbear can maybe be seen as just nitpicking but for consistency in Kodi should this not be called "game client" instead of "gameclient"?

This is as all other clients in Kodi's common grammar for the English languge are written with separation between words for the function noun and "client". IMHO this seperation of the word "client" should also make translation to other languges just a little bit easier.

Examples where the word "client" is already written seperated in Kodi: "PVR client", "VFS client", "UPnP client", "InputStream client".

MartijnKaijser:
You cannot split a label that is used in coding.

PR #11107, opened 2016-12-10
Move game extensions to advanced settings

garbear:
This imports the 97 game extensions used by our 58 libretro cores to advanced settings.

Previously, the set of game extensions was generated dynamically from the enabled game add-ons, plus all remote game add-ons. Using remote add-ons allows for just-in-time installation, because Kodi is able to show the game in MyGames even if the add-ons isn't installed yet.

However, this caused two problems. First, UX was impacted because even though a file is a game, if no local or remote add-ons could play it, then it wouldn't show up in MyGames. This was especially confusing if no game add-ons or game add-on repos were installed, because then MyGames would be completely empty. The second problem was a performance impact due to hitting the add-on DB for remote game extensions whenever all game extensions were needed.

### DIFF
--- a/xbmc/games/GameUtils.cpp
+++ b/xbmc/games/GameUtils.cpp
@@ -29,6 +29,7 @@
 #include "filesystem/SpecialProtocol.h"
 #include "utils/StringUtils.h"
 #include "utils/URIUtils.h"
+#include "settings/AdvancedSettings.h"
 #include "FileItem.h"
 #include "ServiceBroker.h"
 #include "URL.h"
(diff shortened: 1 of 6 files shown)


garbear:
Currently, games with `.bin`, `.cue`, `.iso`, `.img`, and `.zip` extensions can't be played because these conflict with music/video extensions. Any ideas on how to fix this?

AlwinEsch:
Good question, have also not direct a good idea.

About the `dvdimage`, maybe the way like here https://github.com/xbmc/xbmc/blob/master/xbmc/dialogs/GUIDialogSimpleMenu.cpp#L64-L77

To add functions like this:
```cpp
bool CFileItem::IsVideoDiscImage() const
{
  if (item.IsDiscImage())
  {
    CURL url2("udf://");
    url2.SetHostName(item.GetPath());

    url2.SetFileName("VIDEO_TS/VIDEO_TS.IFO");
    if (XFILE::CFile::Exists(url2.Get()))
      return true;

    url2.SetFileName("BDMV/index.bdmv");
    if (XFILE::CFile::Exists(url2.Get()))
      return true;
  }
  return false
}
```

da-anda:
IMO the old approach was the better one. What is it good for to see games listed when I can't play them due to non-existent cores/repos? Don't see how this improves UX.
Next issue I see is that some emulators/games will not run on all platforms, so you'd be listing games that can never be played on a certain platform (.exe, .bat?). Imagine a network share full with games and you browse them once from your full blown Windows gaming HTPC and once from your PI. Doesn't make much sense to show "blobbyvolley.exe" on the PI

The overhead in accessing the add-ons DB could be fixed by caching the aggregated result in a hidden setting f.e. and update this cache every time the add-on repos are rescanned for updates (have add-on manager fire a signal and hook on to it). That's probably how I would have done it. But you have more insight and can judge better what's a suiteable way.

garbear:
these are good points. Caching the extensions would fix the performance issue. And a error dialog upon MyGames open shown if the local install and remote repos have no game add-ons would fix the UX issue. Will pursue, thanks

Still need to figure out how to handle ambiguous extensions: .bin, .cue, .iso, .img, and .zip. I guess I'll just have to parse each format.

Montellese:
I don't know how the extension mapping works right now but couldn't we extend it with an optional media type? If your in the games library it is obvious that you want to play a game and not watch in ISO DVD. Kodi already has this information for library items. For file items (e.g. in the file manager) we might still have to guess blindly though.

LipkeGu:
@garbear remove the bin extension please psx emulators has problems with plain bin files without cuesheets (.cue) for example FF7

garbear:
I think a better fix would be to stack .bin and .cue files like we do for multi-cd files. Anyways, the real solution is to create a parser that can detect the type of media the file is. When the parser is ready I'll open a PR and close this one.

Montellese:
Do we really need a generic parser to detect the media type. Obviously it would be great but in most cases we already know the media type or at least the media context (games, video, music) from the GUI window which is used to activate an item. We just don't have a way to pass that information along with the action to "play" the file yet.

garbear:
What about in the file manager? It doesn't have a media type, it only detects the extension.

da-anda:
I know it's far fetched and we don't support these emulators yet, but an ISO could be f.e. PS1 or Wii

Montellese:
@garbear: TBH I've never used the file manager because I have everything in my libraries ;-)

LipkeGu:
@garbear : Is it able to bring kodi to use the info provided by the cue sheet (.cue)? the cue is a text file which contains the track and Disc infos.

```
FILE "Metal Gear Solid (1).bin" BINARY
TRACK 1 MODE2/2352
INDEX 1 00:00:00
```

garbear:
Closing. Ultimately extension handling is insufficient and we need a parser.

PR #11115, opened 2016-12-11
[PVR][video] Consolidate implementation of "Play", "Mark watched", "Mark unwatched", "Resume from" context menu items

ksooo:
Both video files and pvr recordings have similar context menu items, like "Play", "Mark watched", "Mark unwatched" and "Resume from". For historical reasons, there are different implementations for these menu items, although lots of code could be shared.

After #10870 got merged it is now possible to consolidate these implementations.  This PR does exactly this. :-) A bunch of code refactoring needed as prerequisite for this is also included here.

A nice visual side effect of these changes is that now "Resume" and "Play" for PVR recordings appear at the topmost positions in the context menu. This has been requested by or users for several times.

The code was runtime-tested on Linux an macOS, latest master.

@Jalle19 for review?

### DIFF
--- a/xbmc/FileItem.cpp
+++ b/xbmc/FileItem.cpp
@@ -1522,10 +1522,24 @@ void CFileItem::UpdateInfo(const CFileItem &item, bool replaceLabels /*=true*/)
   if (item.HasVideoInfoTag())
   { // copy info across
     //! @todo premiered info is normally stored in m_dateTime by the db
-    *GetVideoInfoTag() = *item.GetVideoInfoTag();
-    // preferably use some information from PVR info tag if available
-    if (m_pvrRecordingInfoTag)
-      m_pvrRecordingInfoTag->CopyClientInfo(GetVideoInfoTag());
+
+    if (item.m_videoInfoTag)
+    {
+      if (m_videoInfoTag)
+        *m_videoInfoTag = *item.m_videoInfoTag;
+      else
+        m_videoInfoTag = new CVideoInfoTag(*item.m_videoInfoTag);
+    }
+    else
+    {
+      if (m_videoInfoTag)
+        delete m_videoInfoTag;
+
+      m_videoInfoTag = new CVideoInfoTag;
+    }
+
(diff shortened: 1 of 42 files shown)


ksooo:
jenkins build this please

ksooo:
@Jalle19 any comments? I know it's pain to read the large diff, but that's why i split it up into multiple commits. ;-) Stuff is well runtime tested for some time now, btw. Like always I use my family members as beta testers in the living room. No complaints so far. ;-)

Jalle19:
Sorry, missed this one.

ksooo:
rebased, good to go?

ksooo:
jenkins build this please

xbmc/FileItem.cpp
```diff
@@ -3533,17 +3534,42 @@ int CFileItem::GetVideoContentType() const
...
 
+CFileItem CFileItem::GetItemToPlay() const
+{
+  if (HasEPGInfoTag())
+  {
+    const CPVRChannelPtr channel(GetEPGInfoTag()->ChannelTag());
+    if (channel)
+      return CFileItem(channel);
+  }
+  return CFileItem(*this);
```
Jalle19: Wouldn't `return *this;` have the same effect as this but save one copy?

ksooo: Hehe, good catch.

Jalle19: Even better would be to potentially use `CFileItemPtr` instead, but that probably requires more careful thought.

ksooo: This would require to derive CFileItem from std::enable_shared_from_this which I wanted to prevent.

ksooo:
jenkins build this please

ksooo:
Jenkins macOS error is not related to this PR.

<pre>
make: *** [/Users/Shared/jenkins/workspace/OSX-64/tools/depends/pre-depends/autoconf-pre-depends/../../pre-build-deps/autoconf-2.68.tar.gz] Error 7
autoconf was not found and could not be built. Aborting.
</pre>

PR #11118, opened 2016-12-11
VideoPlayer: fix cc after https://github.com/xbmc/xbmc/commit/03aa244…

FernetMenta:
see title
fix http://trac.kodi.tv/ticket/17102


FernetMenta:
no need to bother jenkins with this

PR #11120, opened 2016-12-11
[addons] move installer temp dir to addons/temp/

tamland:
This is mostly cosmetic change to hopefully prevent more issues such as http://trac.kodi.tv/ticket/17049 and #10904. In short, this temp directory must be on the same filesystem special://home/addons/ for the safety to be achieved (otherwise there is no way to move and it will fallback to copy+delete). So safest thing is to just keep it in inside the addons folder.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -1059,6 +1059,7 @@ void CApplication::CreateUserDirs() const
   CDirectory::Create("special://home/");
   CDirectory::Create("special://home/addons");
   CDirectory::Create("special://home/addons/packages");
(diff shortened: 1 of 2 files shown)


candrews:
Would it not be better to fix the file copying methods so they work across file systems as suggested at http://trac.kodi.tv/ticket/17049 ?

tamland:
That would defeat the purpose of this installer. Both unpacking, copying and deleting are unsafe (especially on windows), and if something fails you end up with either a partially installed or partially installed addon. The easiest way to handle this safely is to instead use atomic move.

MartijnKaijser (approved):

PR #11123, opened 2016-12-11
[language] Let's call it what it is: English (United Kingdom)

MartijnKaijser:
@Montellese you know of any downsides this might have?

### DIFF
--- a/addons/resource.language.en_gb/addon.xml
+++ b/addons/resource.language.en_gb/addon.xml
@@ -2,7 +2,7 @@
 <addon
   id="resource.language.en_gb"
   version="2.0.1"
-  name="English"
+  name="English (United Kingdom)"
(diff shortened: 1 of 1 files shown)


MartijnKaijser:
@Montellese this indeed causes issues for add-on that have not yet moved to the new language folder layout where we use the language codes as folder names.
Is there any fallback we can add for that?

Montellese:
@MartijnKaijser I don't follow. Is there a problem with the current status quo or with this PR?

MartijnKaijser:
@Montellese With this PR. 
Addons that still use "English" as language folder name won't get get the language file loaded anymore

Montellese:
TBH I already disliked having to add the existing fallback so not a fan of adding another fallback.

PR #11124, opened 2016-12-11
[addons] add enable/disable context menu

tamland:
adds enable and disable to the context menu, just to make it a bit more accessible


tamland:
jenkins build and merge

PR #11125, opened 2016-12-11
[gui] DialogSeekbar should use g_application.GetPercentage() for progress info

phil65:
Forgot to PR this some time ago...
@FernetMenta

### DIFF
--- a/xbmc/dialogs/GUIDialogSeekBar.cpp
+++ b/xbmc/dialogs/GUIDialogSeekBar.cpp
@@ -18,6 +18,8 @@
  *
  */
 
+#include <math.h>
+
 #include "GUIDialogSeekBar.h"
 #include "Application.h"
 #include "GUIInfoManager.h"
@@ -68,7 +70,7 @@ void CGUIDialogSeekBar::FrameMove()
   // update controls
   if (!CSeekHandler::GetInstance().InProgress() && g_infoManager.GetTotalPlayTime())
   { // position the bar at our current time
-    CONTROL_SELECT_ITEM(POPUP_SEEK_PROGRESS, (unsigned int)(static_cast<float>(g_infoManager.GetPlayTime()) / g_infoManager.GetTotalPlayTime() * 0.1f));
+    CONTROL_SELECT_ITEM(POPUP_SEEK_PROGRESS, lrintf(g_application.GetPercentage()));
     SET_CONTROL_LABEL(POPUP_SEEK_LABEL, g_infoManager.GetCurrentPlayTime());
   }
   else
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -64,7 +64,6 @@
 #include "utils/CPUInfo.h"
 #include "utils/SortUtils.h"
 #include "utils/StringUtils.h"
-#include "utils/MathUtils.h"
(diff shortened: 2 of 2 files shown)


FernetMenta:
thanks!
+1

phil65:
jenkins build this please.

phil65:
changed it to use MathUtils::round_int() instead of a static cast.
that way it aligns with https://github.com/xbmc/xbmc/blob/master/xbmc/GUIInfoManager.cpp#L6741 and fixes a visual glitch for seekbar (different values for player.progress vs value of dialogseekbar slider, control id 401)
Your choice if this is worth backporting, @FernetMenta 
Jenkins build this please.

garbear:
I think we prefer std::roundf() as opposed to MathUtils because the latter is intended for pixel rounding in GUI code

phil65:
@FernetMenta , @garbear 	
updated, and appended a 2nd commit to drop MathUtils in InfoManager.
Fine that way?
And do you want a backport?

FernetMenta:
thanks. I would keep backports to a minimum of critical fixes but I am fine if you want to backport it.

jenkins build this please

phil65:
jenkins build this please.

phil65:
now using roundf(), round() and lround() + proper casting where appropriate to avoid warnings.
jenkins build this please.

xbmc/dialogs/GUIDialogSeekBar.cpp
```diff
@@ -18,6 +18,8 @@
  *
  */
 
+#include <math.h>
```
phil65: Thx! Will fix.

phil65:
jenkins build this please.

garbear:
sorry, this should be `std::lrintf()` instead of `static_cast<int>(std::roundf())`

FernetMenta:
lrintf still requires a cast to int

garbear:
Documentation says `returns it as a value of type long int`, why would it still require a cast?

FernetMenta:
> Documentation says returns it as a value of type long int, why would it still require a cast?

it gets assigned to parameter "value" which is of type int, not long int.

garbear:
then you mean it requires an implicit cast instead of an explicit one, right?

FernetMenta:
if you assign long to int without explicit cast you get a warning, don't you?

garbear:
> if you assign long to int without explicit cast you get a warning, don't you?

On gcc you only get a warning if -Wconversion is specified, and I don't think it is for our project

FernetMenta:
ok, then it's fine

phil65:
updated. @garbear mind takin another look again?

garbear:
looks good, thanks

phil65:
jenkins build this please.

phil65:
android std library doesnt seem to contain lrint(), is that correct? What should I use instead?

phil65:
switched from <cmath> to <math.h>, that´s what we use everywhere else in code, too.

phil65:
jenkins build this please.

phil65:
jenkins build this please.

phil65:
jenkins build this please.

FernetMenta:
@phil65 afaik there is no lrintf on Android, just use lrint

FernetMenta:
@phil65 great :) then shove it in

PR #11143, opened 2016-12-11
[PVR] trac17040: fix crashes on pvr manager deinit/reinit.

ksooo:
This fixes several crashes on pvr manager reinit and deinit, like trac#17040 (http://trac.kodi.tv/ticket/17040)

PVR manager gets reinited at runtime under certain circumstances, like disabling or uninstalling pvr client addons. Objects managed by PVR manager (timers, recordings, channels, ...) are accessed from various kodi threads. During reinit the above mentioned objects will be destroyed and recreated.

For unknown reasons  neither the destruct/recreate sequence of the pvr objects nor the access to these objects were protected by pvr manager's lock.

Moreover, the objects exposed by PVR manager were raw pointers.

This PR makes PVR manager reinit/deinit thread safe and raw pointers are replaced by shared pointers to ensure pvr objects stay alive once they have been obtained.

Also necessary for fixing trac17040 is to stop playback before stopping a pvr client (for instance upon update/disable/deinstall). This fix is also part of this PR.

The change has been tested on macOS, latest Kodi master.

@ace20022 would be great if you could runtime test this.
@MilhouseVH could you please include this patch in your build to get broader feedback on possible regressions?

### DIFF
--- a/xbmc/pvr/PVRManager.cpp
+++ b/xbmc/pvr/PVRManager.cpp
@@ -77,6 +77,7 @@ using KODI::MESSAGING::HELPERS::DialogResponse;
 
 CPVRManager::CPVRManager(void) :
     CThread("PVRManager"),
+    m_addons(new CPVRClients),
     m_triggerEvent(true),
     m_currentFile(NULL),
     m_database(NULL),
@@ -88,15 +89,12 @@ CPVRManager::CPVRManager(void) :
     m_isChannelPreview(false)
 {
   CAnnouncementManager::GetInstance().AddAnnouncer(this);
-  m_addons.reset(new CPVRClients);
 }
 
 CPVRManager::~CPVRManager(void)
 {
   CServiceBroker::GetSettings().UnregisterCallback(this);
   CAnnouncementManager::GetInstance().RemoveAnnouncer(this);
-  Stop();
-  m_addons.reset();
   CLog::Log(LOGDEBUG,"PVRManager - destroyed");
 }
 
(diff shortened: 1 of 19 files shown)


xbmc/pvr/PVRManager.cpp
```diff
@@ -138,6 +137,30 @@ CPVRManager &CPVRManager::GetInstance()
...
+
+CPVRTimersPtr CPVRManager::Timers(void) const
+{
+  CSingleLock lock(m_critSection);
+  return m_timers;
+}
+
+CPVRClientsPtr CPVRManager::Clients(void) const
+{
+  // note: m_addons is only set/reset in ctor/dtor, thus logically const. no need for a lock here.
```
Jalle19: Maybe it's better to be safe than sorry in case someone changes something in the future?

ksooo: np, can do.

ksooo: done

Jalle19 (approved):

mcaptur:
Would be great if this is backported to Krypton

ksooo:
> Would be great if this is backported to Krypton

 Yeah, that's the plan, once it has been proven to be stable enough. There is risk for deadlocks.

ksooo:
The new commit fixes a crash on pvr mannager restart if the guide window is open at that time. Guide window gets automatically reinited and the updater thread worker might not exit fast enough. Thus the thread object should be recreated instead of reused (to prevent CThread::Create implementation calling exit()).

ksooo:
Regardingd deadlock risk: third commit is one of those:

ksooo:
@MilhouseVH could you please update this patch in your next build? Thanks.

ksooo:
@Jalle19 had to remove the lock from CPVRManager::Clients() again as I encountered a deadlock where three threads were involved and CPVRManager::Clients() was part of that game. Boy, this is complex shit.

ksooo:
@MilhouseVH fyi, I force pushed

ksooo:
@Jalle19 I added some more commits to this PR since you last reviewd. Could you pleaes take another look? Thanks.

I use this patches for some time now on LInux and mcOS, no crashes or deadlocks so far and we have this PR in @MilhouseVH builds, no negative feedback from the testers so far. Thus, I would like to shove this in soon, given you approve the code changes ofc.

ksooo:
jenkins build this please

xbmc/pvr/PVRManager.cpp
```diff
@@ -138,6 +136,30 @@ CPVRManager &CPVRManager::GetInstance()
...
+
+CPVRTimersPtr CPVRManager::Timers(void) const
+{
+  CSingleLock lock(m_critSection);
+  return m_timers;
+}
+
+CPVRClientsPtr CPVRManager::Clients(void) const
+{
+  // note: m_addons is const (only set/reset in ctor/dtor). no need for a lock here.
```
Jalle19: It looks like the lock we discussed earlier got removed? Or is this intentional?

Jalle19: Nevermind, just remembered your last comment...

ksooo:
@Jalle19 Good to go?

mcaptur:
added backport https://github.com/xbmc/xbmc/pull/11197

AlwinEsch:
Thanks a lot!

PR #11145, opened 2016-12-12
fix compile without microhttpd

notspiff:
These directories need to be optional. http://trac.kodi.tv/ticket/17131


wsnipex:
thanks, jenkins build this please.
needs backport as well

PR #11156, opened 2016-12-13
[estuary] rm deprecated infobool: isempty->string.isempty

BigNoid:
see http://forum.kodi.tv/showthread.php?tid=298461&pid=2477340#pid2477340

@phil65


phil65:
Thx!
Have to say that I liked the previous less-verbose infobools more compared to these new ones...

BigNoid:
```String.IsEmpty``` would probably be better if it was left ```IsEmpty``` indeed.

PR #11162, opened 2016-12-13
Blacklist problematic Sony TV remote

garbear:
See discussion: http://forum.kodi.tv/showthread.php?tid=298891

Alternative to #11065 to avoid introducing a setting.

Before fix:
Device removed
New device configured

## Types of change

### DIFF
--- a/xbmc/peripherals/bus/android/AndroidJoystickState.cpp
+++ b/xbmc/peripherals/bus/android/AndroidJoystickState.cpp
@@ -88,6 +88,13 @@ bool CAndroidJoystickState::Initialize(const CJNIViewInputDevice& inputDevice)
 
   const std::string deviceName = inputDevice.getName();
 
+  // Skip problematic devices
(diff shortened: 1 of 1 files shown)


garbear:
untested - @Montellese does this look right?

MartijnKaijser:
maybe some other way to extend the blacklist in case there will be more like these in the future?

NedScott:
There will be an endless and nearly impossible to maintain list when you consider all of the possible input devices this could include. I can already reproduce the issue with a generic bluetooth mouse. IMO, if you have to choose between maintaining a a blacklist and adding a setting, go with the setting.

CiNcH83:
I support what @NedScott just wrote. I also have a Rapoo wireless keyboard which I connect to may Android TV from time to time for better debugging, and it also often gives me the connect/disconnect messages.

MartijnKaijser:
I guess we close this one as specifying single remotes won't work

PR #11163, opened 2016-12-13
Fix silent failure when game add-on can't be installed

garbear:
On the latest Milhouse builds, game add-ons can't be installed because game.libretro (a dependency) isn't in the libreelec repo. Instead of failing silently, this adds an error dialog.

## Screenshots
error dialog

## Types of change

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -1179,6 +1179,7 @@ msgid "Fetching CD information"
 msgstr ""
 
 #: xbmc/dialogs/GUIDialogFileBrowser.cpp
+#: xbmc/games/dialogs/GUIDialogSelectGameClient.cpp
 #: xbmc/pvr/PVRGUIActions.cpp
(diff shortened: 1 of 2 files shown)


MartijnKaijser:
ping @tamland 
This seems like something we should already have from normal add-ons and i don't like duplication

tamland:
we already have that. and im assuming the commented out code above should be commented in, removing this custom game 'dialog' entirely.

garbear:
what commented-out code?

tamland:
the one in GUIDialogSelectGameClient.cpp that says 'todo Switch to add-on browser when more emulators have icons'

MartijnKaijser:
Still needed?

garbear:
Yea, this is still an issue. The PR stalled because `CGUIDialogSelectGameClient` needs to be removed, but we can't do this until we can get the game client from a game database.

However, seeing as we don't have a game database, I recommend we merge this to fix the silent error until a game database is ready.

PR #11169, opened 2016-12-13
Fix queueing .strm and .pls music files

Memesa:
.strm and .pls music (playlist) files cannot be queued but are treated instead as directories when in the Music Navigation screen, leading to errors. Other music playlist files (e.g. .m3u) can be queued.

This is a Krypton backport of https://github.com/xbmc/xbmc/pull/11175.
## Description
Queueing was not possible for music playlist items because the playlist would be processed as a directory resulting in the error:
`DEBUG: CGUIMediaWindow::GetDirectory (<path>/test.strm)`
`XFILE::CDirectory::GetDirectory - Error getting <path>/test.strm`

To my knowledge and tests, the former if statements are incorrect. There is no possibility for .strm and .pls files (playlist files) to be listed as a directory, and this part can therefore be removed. This fixes the queuing these files. Other music files such as .m3u can be listed as directories, but are caught by the if statement as virtual folder.

## How Has This Been Tested?
By queuing an .strm file containing an http link to a music file; didn't work before the fix, did work after the fix (with a reboot and user setting reset to get rid of the memorized path).

## Types of change
- [X]  Bug fix (non-breaking change which fixes an issue)
- [ ]  New feature (non-breaking change which adds functionality)
- [ ]  Breaking change (fix or feature that would cause existing functionality to change)

PR #11175, opened 2016-12-14
Fix queueing .strm and .pls music files

Memesa:
.strm and .pls music (playlist) files cannot be queued but are treated instead as directories when in the Music Navigation screen, leading to errors. Other music playlist files (e.g. .m3u) can be queued.

PR correctly referenced to Master now, in contrary to https://github.com/xbmc/xbmc/pull/11169. This is a very minor fix and I suggest to consider it for inclusion in Krypton.

## Description

Queuing was not possible for .strm and .pls music items because the playlist would be processed as a directory resulting in the error:
DEBUG: CGUIMediaWindow::GetDirectory (<path>/test.strm)
XFILE::CDirectory::GetDirectory - Error getting <path>/test.strm

To my knowledge and tests, the former if statements are incorrect. There is no possibility for .strm and .pls files (playlist files) to be listed as a directory, and this part can therefore be removed. This fixes the queuing these files. Other music files such as .m3u can be listed as directories, but are caught by the if statement as  virtual folder.

## How Has This Been Tested?

By queuing an .strm file containing an http link to a music file; didn't work before the fix, did work after the fix (with a reboot and user setting reset to get rid of the memorized path).

## Types of change

### DIFF
--- a/xbmc/music/windows/GUIWindowMusicBase.cpp
+++ b/xbmc/music/windows/GUIWindowMusicBase.cpp
@@ -669,7 +669,7 @@ void CGUIWindowMusicBase::AddItemToPlayList(const CFileItemPtr &pItem, CFileItem
       return;
     }
   }
-  if (pItem->m_bIsFolder || (g_windowManager.GetActiveWindow() == WINDOW_MUSIC_NAV && pItem->IsPlayList()))
+  if (pItem->m_bIsFolder)
   {
     // Check if we add a locked share
     if ( pItem->m_bIsShareOrDrive )
@@ -996,7 +996,7 @@ void CGUIWindowMusicBase::PlayItem(int iItem)
 #endif
 
   // if its a folder, build a playlist
-  if ((pItem->m_bIsFolder && !pItem->IsPlugin()) || (g_windowManager.GetActiveWindow() == WINDOW_MUSIC_NAV && pItem->IsPlayList()))
+  if (pItem->m_bIsFolder && !pItem->IsPlugin())
   {
     // make a copy so that we can alter the queue state
     CFileItemPtr item(new CFileItem(*m_vecItems->Get(iItem)));


jjd-uk:
Thanks.

Let's see if @DaveTBlake has the time to take a look.
Perhaps @Razzeee could also take a look.

DaveTBlake:
Sorry I'm too busy with other things to check this properly at the moment, but hope to get to it.
The line of code that's changed has been that way for a long time, so I would want to be sure that there is no regression. I'm sure it fixes queue of *.strm files, but have all the other routes to playing stuff been checked for breakage? Unfortunately there are many diverse ways the player gets called (from UI, JSON API, Python etc.), and historically easy to cause unexpected issues.

Memesa:
I can add to that, that this 'if line' checks for the combination of being a playlist and if kodi is currently in the music browser. This means that it should only impact UI; when a playlist is called from every other situation it is redirected to the 'if playlist' part of the code. This becomes more clear if you queue media from the remote app, because then the items do get queued as it fails the 'in music browser' statement.

Also, in the same dialog for video, this code does not check for the playlist+browser either;
in [CGUIWindowMusicBase::AddItemToPlayList](https://github.com/xbmc/xbmc/blob/master/xbmc/video/windows/GUIWindowVideoBase.cpp#L461)
`line 498: if (items[i]->m_bIsFolder)`

in [CGUIWindowVideoBase::PlayItem](https://github.com/xbmc/xbmc/blob/master/xbmc/video/windows/GUIWindowVideoBase.cpp#L1255)
`line 1263: if (pItem->m_bIsFolder && !pItem->IsPlugin())`

Basically, removing these statements in GUIWindowMusicBase.cpp brings it in sync with its video counterpart.

croneter:
Anything we can do to help include this for the final version 17.0?

jjd-uk:
It won't make v17 now as only vital fixes are being allowed in now as we try and stabilise for releasing v17.

MartijnKaijser:
Maybe if @DaveTBlake looks into it we could maybe get it for v17.1 but it should of course first go to master

DaveTBlake:
OK @Memesa, @croneter  since @MartijnKaijser is offering hope of merge in 17.1, what I need to test this out is some .strm files, or at least some remote audio to stream. It may seem crazy to you, but I keep all my music locally.

Do  standard playlists (.PLS or .M3U) or local music have the same issue?

croneter:
@DaveTBlake, sorry, no idea. I'm only trying to facilitate here. 

@Memesa, can you chime in here to help out Dave?

Memesa:
I've seen the request, will post some .strm files ASAP and will test the .pls and .m3u too (pretty sure I already tested those to be non-working before this PR and working after).

Memesa:
You can download an .strm file at https://kodistrmtest.azurewebsites.net/kodistrmtest.strm.html (remove the .html extension after downloading) which will link to a music file at https://kodistrmtest.azurewebsites.net/Test.mp3.

### What does not work before applying the PR and works after:
right click/context menu on the file -> queue item

### What does work:
left click/enter on the file

in regard to your question about standard playlists; .M3U files behave exactly the same, .PLS files do not work at all when the .STRM is renamed to .PLS (maybe a different content is expected?)

DaveTBlake:
Thanks @Memesa for the test stream. It is also useful for some other testing I am doing, remote files is not my ususal use mode at all, so is it possible to leave that file there longer term?

However I am getting some odd behaviour.  Even before applying the PR I find that 
1) right click/context menu on the file -> queue item
does work.

I guess this could be related to my window history? However I would be more confident in the solution to the problem if I could consistently repeat the problem.

DaveTBlake:
> in regard to your question about standard playlists; .M3U files behave exactly the same

Navigate to a .M3U file in file view, right click/context menu - > queue item, the songs listed in M3U get added to the current playlist. @Memesa you seem to say that they don't, am I missing something?

Memesa:
I just checked and I can get your behavior after left click/enter/play where the file info gets cached. (Kodi 17.0 git 20170201 and Kodi 16.1, both Windows 10)

However it does not work for me if not 'played' first (not cached), so I generally restart Kodi before trying.

Also if you try to queue the file when cached, it does play the song but when trying to queue twice it does not add it and there will only be a single song in the playlist.

Can you see the `ERROR: XFILE::CDirectory::GetDirectory - Error getting` errors in the kodi.log?

DaveTBlake:
This change checks out as far as I can tell, but it only seems to effect .strm and .pls files. The only way I can see to test these files is from file view, is that correct, or is there some other approach?

Standard .M3U playlists have `m_bIsFolder = true`, either when invoked from file view or from the playlists node, so are not effected by this. Actually  `m_bIsFolder` is also true for both folders and files, virtual and real.   

Possibly worth editing both the PR title and initial desc to clarify this is for .strm and .pls files, "playlist" alone can mean so many different things in kodi.

But I do wonder what the `|| (g_windowManager.GetActiveWindow() == WINDOW_MUSIC_NAV && pItem->IsPlayList()` test was about originally.  Nested playlists? Maybe at sometime M3U files were not flagged with `m_bIsFolder = true` when played from within the library? I can only guess.

Memesa:
I'm also testing them from the file view, that is correct. The music scraper does not work with any non-music files.

The .M3U files indeed work to be queued but also allow to be browsed into using Kodi, an option that does show up for .STRM and .PLS but does not have any function. Good to know that those are caught by `m_bIsFolder = true`, I missed that one during testing. I also tested the .M3U functionality for the build after applying this PR, and both browse into and queuing works as expected.

I'll edit the title and description to clarify that the effect is on .strm and .pls files, as other playlists formats (e.g. .M3U) are caught with the virtual `m_bIsFolder = true`.

DaveTBlake:
Happy with this now, thanks @Memesa 
jenkins build this please

croneter:
No chance to get this into v17.1?

DaveTBlake:
>No chance to get this into v17.1?

Yes a backport is feasible, #11169 just needs editing etc.

Memesa:
@croneter The current PR has been referenced to master, as this is the way to implement new fixes and features. You can read about backports in my other/backport PR at https://github.com/xbmc/xbmc/pull/11169 through jjd-uk's comment https://github.com/xbmc/xbmc/pull/11169#issuecomment-267055045

PR #11177, opened 2016-12-14
[xcode] Add some recently introduced files to xcode to fix build.

ksooo:
xbmc/ContextMenus.(cpp|h) for macOS and iOS
xbmc/pvr/PVRGUIActions.(cpp|h) for iOS
xbmc/pvr/PVRContextMenus(cpp|h) for iOS
xbmc/pvr/PVRItem(cpp|h) for iOS


ksooo:
jenkins build not needed, as only xcode project file modified, which is not used by jenkins

PR #11179, opened 2016-12-14
[addons] remove use of try, catch

AlwinEsch:
This remove the not good usable use of try and catch on binary add-on handling.

## Description
The try catch was never really used and also is in handling not the best, with them becomes it removed from binary add-on handling

## Motivation and Context
Continued cleanup to bring complete rework step by step in.

## How Has This Been Tested?
With binary add-on's.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/InputStream.cpp
+++ b/xbmc/addons/InputStream.cpp
@@ -110,14 +110,7 @@ void CInputStream::UpdateConfig()
 
   if (status != ADDON_STATUS_PERMANENT_FAILURE)
   {
-    try
-    {
-      pathList = m_pStruct->GetPathList();
-    }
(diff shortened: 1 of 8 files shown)


garbear:
I approve of the peripheral changes, because we have control over the code, but the gameclient try catches should stay. notice the catch(...) - I had to change this from std::exception because some emulators were throwing ints. emulator territory is truly scary and we should guard against exceptions in their code. 

BTW, OOP execution can't some soon enough, there's still some emulators that consistently crash or hang

AlwinEsch:
Thanks @garbear! Have removed the commit to them.

AlwinEsch:
jenkins build this please

PR #11182, opened 2016-12-14
[addons] remove the use of template's from add-on interface

AlwinEsch:
Remove the use of template in Kodi's add-on classes.

## Description
This is the first major part to becomes in future instances of add-on's multiple times together usable.

With them becomes the `template<class TheDll, typename TheStruct, typename TheProps> class CAddonDll` changed to `class CAddonDll` only.

This is required to use the `CAddonDll` in future complete alone and to bring a way to create e.g. `CPVRClient` several times.

Further has the `template` never make sense for me. It is OK if nearly every function need to know the used type, but for `CAddonDll` was the template value only on `Create` used and with one function is not much.

For this reasons becomes now the add-on function table added to the place where it is used (e.g. `struct PVRClient` now in `CPVRClient` as `PVRClient m_struct;`).

The only bit bad thing (in the moment) is the use of values as `void*`, but required to have `CAddonDll` independent from structure names:

In future becomes the now `ADDON_STATUS Create(void* funcTable, void* info);` replaced back to `ADDON_STATUS Create();` and his values called direct to a special function for them with `CreateInstance`.

The function `ADDON_Create` becomes in future used like a `int main()`, that's also a reason why this `CreateInstance` comes, a main only called one time and the new as much as needed.

## Motivation and Context
Bring in a complete reworked binary add-on interface what looks like them: https://github.com/AlwinEsch/kodi/tree/add-addon-instance-handling/xbmc/addons/kodi-addon-dev-kit/include/kodi (this part there not complete finished)

## How Has This Been Tested?
Build of kodi and test with add-on's

## Types of change

## Checklist:

>*Note:* currently contains changes of request #11179


akva2:
the templating there was there because the original code intended to use a traits based construct (TheProps) as well as avoid evil void* usage (TheStruct). it was never used to any large extent so removing it should have no consequence other than the knee-jerk void* usage (which i suspect will go away eventually, if i read you correctly).

what's the problem with the exceptions? afaik even mingw exceptions are translated to win32 so i don't see why it would cause issues ? (note, i'm just curious, i'm sure you have good reasons!)

AlwinEsch:
Hi @akva2, yes the `void*` goes complete away from CAddonDll. 

The only places where void* becomes still used after final rework are:

1. On all Kodi to add-on calls a `void* addonInstance` value, to have a pointer to class who handle it on add-on
2. On all add-on to Kodi calls a `void* kodiInstance` value, to have a pointer to class who handle it in Kodi
3. On the [CreateInstance](https://github.com/AlwinEsch/kodi/blob/add-addon-instance-handling/xbmc/addons/kodi-addon-dev-kit/include/kodi/AddonBase.h#L252) call, where the add-on gives his instance address.

All this `void*`'s are defined in future by a `typedef void* KODI_HANDLE;` and **not visible or used direct on add-on's**, they are only be handled from the headers.

About exceptions, was also requested from @FernetMenta, from my side it can be make more sense to do this direct inside the coming add-on headers and leave away from Kodi itself.

Currently was on nearly all try, catch's only a write of a log entry, but from my side is this alone not good. If it stays in must be more done after a `throw`.

AlwinEsch:
Ah forgot, about the `void* addonInstance` in point 2. there seems possible to pass the parent add-on class direct and prevent the use of them.

Here the coming parent class header for PVR add-ons: https://github.com/AlwinEsch/kodi/blob/add-addon-instance-handling/xbmc/addons/kodi-addon-dev-kit/include/kodi/pvr/PVRClient.h#L652-L2360

AlwinEsch:
jenkins build this please

AlwinEsch:
jenkins build this please

AlwinEsch:
jenkins build this please

FernetMenta:
jenkins build this please

Paxxi:
I'd rather see an interface used instead of a void* but if it's temporary(fixed before v18) then a void* can be allowed.

AlwinEsch:
Yes after final changes becomes the void* removed from CAddonDll if everything goes right.

Like on DevCon described are this changes a major part to allow:
- Use e.g. CPVRClient multiple times
- Use CAddonDll alone like a Python add-on
- Use different add-on types e.g. PVR together with Inputstream and audio decoder who all on one add-on 
- Without the template is the code more clear
- Remove of some not needed code possible, e.g. the various calls on https://github.com/xbmc/xbmc/blob/master/xbmc/addons/AddonBuilder.cpp are no more needed, like `return AddonPtr(new PVR::CPVRClient(std::move(addonProps)));`, for them and any other binary can it be replaced to `return AddonPtr(new CAddonDll(...));`

That is why I will make this comparatively fast on begin, this change now can prevent a lot of update faults if something becomes changed.

Basically are on them not direct changes of add-on process.

AlwinEsch:
OK to merge?

Paxxi:
No objections

AlwinEsch:
Thanks a lot!

PR #8418, opened 2015-11-21
Added: set the correct proxy for FFMPEG to pick up if HttpProxy was p…

basrieter:
This PR fixes the issue I reported at http://trac.kodi.tv/ticket/16336 and http://forum.kodi.tv/showthread.php?tid=248440.

It sets the proxy value (passed in via the HttpProxy url parameter) to the environment variables HTTP_PROXY and HTTPS_PROXY. If a proxy was already configured, the original value is stored and restored again during the Dispose of the DVDDemuxFFmpeg.

PR #11186, opened 2016-12-15
remove header filter

peak3d:
Remove filter of HEADER fields in CCurlFile 

## Description
Current curl implementation allows only a set of HEADER fields somewhen grepped from a wiki page.
This PR removes this filter and allows all HEADER's 

## Motivation and Context
During implementation of DASH content we had many issues by not being able to provide all HEADERS wich are necessary to get the required information. Some provider expect headers like Magine-Token: XYZ wich are currently ignored.

### DIFF
--- a/xbmc/filesystem/CurlFile.cpp
+++ b/xbmc/filesystem/CurlFile.cpp
@@ -826,25 +826,8 @@ void CCurlFile::ParseAndCorrectUrl(CURL &url2)
           m_postdata = Base64::Decode(value);
           m_postdataset = true;
         }
-        else if (name == "active-remote")// needed for DACP!
-        {
-          SetRequestHeader(it->first, value);
-        }
         // other standard headers (see https://en.wikipedia.org/wiki/List_of_HTTP_header_fields)
-        else if (name == "accept" || name == "accept-language" || name == "accept-datetime" ||
-                 name == "authorization" || name == "cache-control" || name == "connection" ||
-                 name == "content-md5" || name == "content-type" || name == "date" ||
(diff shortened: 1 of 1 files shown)


wsnipex:
this reverts the decision to not blindly pass on everything addons set and opens up the security issues again.
There must be another way to do what you need

peak3d:
Could you pls. explain the reason fo this filter? I had some discussions a while ago with someone from kodi team (don't remember who it was) and we both have not seen a use case for this.

wsnipex:
e.g. see discussion on https://github.com/xbmc/xbmc/pull/8418

FernetMenta:
-10000

peak3d:
@FernetMenta I opened this PR for discussion and to let everybody know whats the reasons for it.
-10000 is not helpful :-)

Every addon can open an own HTTP server / use its own CURL instance for making things wich we try to avoid with this filter. So, the filter does IMO no sense at all because there are so many ways to do what peaple want to do. But: It means more coding for everyone.

You may know that I started inputstream.mpd with an own CURL handle, I switched to kodis implementation to benefit form the development there. So, adding this filter does not enhance any security .

Why not make a blacklist instead a whitelist? Would this be a solution? Look over the ickets with issues and blacklist the headers wich are security issues?

FernetMenta:
A blacklist does not make sense because there is no control over this.
You could bypass the filter in case the addon is the caller but not in general.

peak3d:
ok

peak3d:
If necessary overruling will appear triggered from inputstream.

PR #11199, opened 2016-12-17
[addons] rename add-on function structures

AlwinEsch:
Rename to structures used on Kodi to Addon interface.

## Description
This rename them that all starts with `KodiToAddonFuncTable_*`, this is one time to prevent conflicting names like "PVRClient" if someone want to use on add-on and the other is to standardize all add-on function tables with the same name on begin.

## Motivation and Context
Binary add-on system rework.

## How Has This Been Tested?
With binary add-on's.

## Types of change

## Checklist:


AlwinEsch:

PR #11200, opened 2016-12-17
[addons] change binary add-on version check way

AlwinEsch:
## Description
This add a new global add-on function with `GetInstanceVersion` who can called from Kodi and from add-on.

Everything about them is inserted to https://github.com/AlwinEsch/kodi/blob/change-addon-version-way2/xbmc/addons/kodi-addon-dev-kit/include/kodi/versions.h and more clear on reviews about changes.

To call add-on is the new function `const char* __declspec(dllexport) ADDON_GetInstanceVersion(int instanceType)` used.

## Motivation and Context
Binary add-on system rework.

## How Has This Been Tested?
With binary add-on's.

bildschirmfoto vom 2016-12-17 15-13-47

bildschirmfoto vom 2016-12-19 19-47-42

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/AddonDll.cpp
+++ b/xbmc/addons/AddonDll.cpp
@@ -169,7 +169,7 @@ bool CAddonDll::LoadDll()
   return true;
 }
 
-ADDON_STATUS CAddonDll::Create(void* funcTable, void* info)
+ADDON_STATUS CAddonDll::Create(int instanceType, void* funcTable, void* info)
 {
   /* ensure that a previous instance is destroyed */
   Destroy();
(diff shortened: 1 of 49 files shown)


AlwinEsch:
jenkins build this please

Only to test builds.

AlwinEsch:
Hi, @FernetMenta, @notspiff, @garbear, @opdenkamp 

This request is primary to ask. Is a way like this OK. Want to use them on the reworked system.

Somewhere in future I want to remove also the version from xml and binaries only be checked by a standardized function call added here on request.

Idea is to define in addon.xml only the used parts, by this way:
```xml
  ...
  <requires>
    <use instance="main"/> <!-- use main function e.g. for file open -->
    <use instance="pvr"/> <!-- use a pvr client interface -->
    <use instance="gui"/> <!-- use gui interface, to show e.g. dialogs -->
    <use instance="inputstream"/> <!-- use input stream interface, to handle the stream -->
  </requires>
  ...
```
If it leaves undefined and not present on xml becomes the add-on checked for all versions, if set by them only the selected ones.

With this need a add-on normally only a recompile and no code changes (so long no bigger changes comes).

xbmc/addons/AddonDll.cpp
```diff
@@ -276,6 +288,38 @@ void CAddonDll::Destroy()
...
+  /* check the API version */
+  const char* kodiVersion = "0.0.0";
+  const char* kodiMinVersion = "0.0.0";
+  kodi::addon::GetInstanceVersions(instanceType, &kodiVersion, &kodiMinVersion);
+
+  const char* addonVersion = "0.0.0";
+  const char* addonMinVersion = "0.0.0";
+  m_pDll->GetAddonInstanceVersions(instanceType, &addonVersion, &addonMinVersion);
+
+  if (AddonVersion(addonVersion) < AddonVersion(kodiMinVersion) ||
```
AlwinEsch: I remove the min version, you are right :)

xbmc/addons/kodi-addon-dev-kit/include/kodi/versions.h
```diff
@@ -0,0 +1,215 @@
...
+
+/*
+ * Versions of all add-on instances are defined below.
+ *
+ * This is added here and not in related header to prevent not
+ * needed includes during compile. Also have it here a better
+ * overview.
+ */
+
+#define INSTANCE_VERSION_MAIN                   "1.0.0"
```
AlwinEsch: This was mainly to prevent dependencies!

In the coming system becomes a add-on type opened from the base add-on system, this defines there is the only part who is needed from them.

Also if a add-on want to use a inline function from there, prevent this not needed includes to add-on type instances who never used.

To this `GetInstanceVersions` function are them needed.

PR #11213, opened 2016-12-18
WIP: [cmake] linux: Add DISTRIBUTION option

cjmayo:
## Description
Skip tasks better run by package manager

## Motivation and Context
When building on Gentoo Linux updating the icon cache cannot be done within the sandbox that a package is being built in and binary stripping is something that should be left to the package manager.

I'd be quite happy to see both these dropped from Kodi but I guess they were consciously included to help an ordinary build - therefore this patch maintains the current behaviour but adds a CMake option a package manager can use to skip both these steps.

For now I've called the option DISTRIBUTION but better ideas more than welcome. LINUX_DISTRIBUTION? PACKAGE something or other? MANAGER?

Happy to update and also add any documentation required.

## Types of change

## Checklist:

### DIFF
--- a/project/cmake/CMakeLists.txt
+++ b/project/cmake/CMakeLists.txt
@@ -39,6 +39,7 @@ set(CMAKE_ASM_FLAGS "${CMAKE_C_FLAGS} -x assembler-with-cpp")
 
 # general
 option(VERBOSE            "Enable verbose output?" OFF)
+option(DISTRIBUTION       "Building for distribution?" OFF)
 option(ENABLE_DVDCSS      "Enable libdvdcss support?" ON)
 option(ENABLE_UPNP        "Enable UPnP support?" ON)
 option(ENABLE_NONFREE     "Enable non-free components?" ON)
--- a/project/cmake/scripts/linux/ArchSetup.cmake
+++ b/project/cmake/scripts/linux/ArchSetup.cmake
@@ -24,9 +24,11 @@ else()
   endif()
 endif()
 
-# Make sure we strip binaries in Release build
-if(CMAKE_BUILD_TYPE STREQUAL Release AND CMAKE_COMPILER_IS_GNUCXX)
-  set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -s")
+if(NOT DISTRIBUTION)
+  # Make sure we strip binaries in Release build
+  if(CMAKE_BUILD_TYPE STREQUAL Release AND CMAKE_COMPILER_IS_GNUCXX)
+    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -s")
+  endif()
 endif()
 
 find_package(CXX11 REQUIRED)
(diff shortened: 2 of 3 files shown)


hudokkow:
I don't dislike the idea but I'm not sold on the `DISTRIBUTION` name. There's probably other stuff that could be under the same umbrella. TexturePacker? CMake add-on helper files, etc.?

fetzerch:
For the stripped binary: In my opinion we should just make sure that the CMAKE_BUILD_TYPE RelWithDebInfo works as designated (and we really should set it as default).
Then ppl that want stripped binaries can just build with CMAKE_BUILD_TYPE=Release, others can use the default or RelWithDebInfo.

For the icon cache, i'm not sure. Maybe @wsnipex has an idea.

wsnipex:
CMAKE_BUILD_TYPE=RelWithDebInfo works as intended and should be used when distros want debug symbols as well as compiler optimizations.

Re Gtk cache: We can probably drop this, since it should work fine without the cache, package managers should do this and one can always do it manually if really needed.

hauihau:
@hudokkow cmake in contrast to autotools now also installs /usr/bin/TexturePacker. Correct me if I'm wrong, but texturepacker is only used while building. Wouldn't it make sense to skip it from installing?

wsnipex:
texturepacker is also a tool skinners need, so we ship install and also ship it as extra package on our ubuntu ppa

hauihau:
but it has never been installed in an autotools build. skinners can still use the extra package from your ppa. if gentoo users want to create a new skin, they can request a texturepacker ebuild.

wsnipex:
feel free to remove it from your ebuild

hauihau:
How was texturepacker installed when autotools were used? I checked the history of tools/depends/native/TexturePacker, cmake was added 10 month ago. So I don't get the point why texturepacker has to be installed now, while it hasn't all the time before...

cjmayo:
`CMAKE_BUILD_TYPE=RelWithDebInfo` does stop me getting the Gentoo QA warnings about pre-stripped files, thanks.

Closing this one and instead submitted the patch for gtk-update-icon-cache we have been using.

hauihau:
Don't know wether some of you guys are using GCC LTO, but `CMAKE_BUILD_TYPE=RelWithDebInfo` added `-gsplit-dwarf` to my FLAGs which resulted in a GCC ICE :|

```
/home/gentoo/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/Util.cpp: In function ‘GetHomePath’:
/home/gentoo/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/Util.cpp:522:0: internal compiler error: in lhd_decl_printable_name, at langhooks.c:222
   return ::GetHomePath(strTarget, strPath); 
```

I thought this could be interesting four you, the corresponding GCC bug can be found here https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78871

fetzerch:
@hauihau: If available we're using gold and split dwarf for faster link performance. You can use `-DENABLE_LDGOLD=Off` to disable that (https://github.com/xbmc/xbmc/blob/master/project/cmake/modules/LDGOLD.cmake).

hauihau:
I'm also using gold. Lets see what happens to the GCC bug. As long it's not fixed I either disable LTO or patch the LDGOLD.cmake.

PR #11218, opened 2016-12-18
[gui] streamline shuffle infolabel

phil65:
Makes it **On / Off** instead off **Random / Off**


phil65:
trivial, no need for jenkins.

PR #11222, opened 2016-12-19
[cec] Add settings for configuring button repeats

popcornmix:
## Description
Add settings for configuring CEC button repeats

## Motivation and Context
CEC remotes repeat at the rate the TV vendor chooses. This may be just a couple of times a second.
This is slow compared to other kodi controllers (lirc, keyboard, game controller) and makes kodi seem laggy.

I have added some changes that are now present in libCEC 4 that allow the CEC repeat messages to be ignored and replaced with messages more like a keyboard generates. The controls are:
* Remote button press delay before repeating (ms)
* Remote button press repeat rate (ms)
* Remote button press release time (ms)

CEC is a poorly implemented standard, and behaviour between different vendors is not consistent which makes these settings necessary. A CEC source may:

Send pressed on press and released on release.
Send pressed on press and periodically and released on release.
Send pressed on press and periodically and nothing on release.

The "release" setting can allow a timeout to be set to assume button is released after, for sources in the third category. This is not ideal as it means the release lags by the amount set. If the periodic repeat rate is high, then the release time can be set lower to reduce the lag.

If the repeat delay is left at 0 (the default) then the behaviour is unchanged.
Typically just changing the repeat delay to, say 150ms will enable this feature and scrolling becomes much quicker. The other options only need adjusting in a minority of cases where source doesn't behave as desired.

## How Has This Been Tested?
Has been present in Raspberry Pi builds for over a year.

## Types of change

## Checklist:

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -17859,7 +17859,22 @@ msgctxt "#36046"
 msgid "Force AVR to wake up when Kodi is activated"
 msgstr ""
 
-#empty strings from id 36047 to 36098
+#: system/peripherals.xml
+msgctxt "#36047"
+msgid "Remote button press delay before repeating (ms)"
+msgstr ""
+
+#: system/peripherals.xml
+msgctxt "#36048"
+msgid "Remote button press repeat rate (ms)"
+msgstr ""
+
+#: system/peripherals.xml
+msgctxt "#36049"
+msgid "Remote button press release time (ms)"
+msgstr ""
+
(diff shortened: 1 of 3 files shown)


popcornmix:
Ping @opdenkamp

wrxtasy:
I've actually back ported this PR into AMLogic LibreELEC 8.0 Krypton that has LibCEC 4 support and results are noticeable improvements in CEC remote response times. 👍

Rechi:
@popcornmix this needs a rebase

Rechi:
@popcornmix @opdenkamp ping

PR #11229, opened 2016-12-20
[gui] improve scaling support for container layouts.

phil65:
improve scaling support by inheriting listitemlayout width/height from container control width/height in case nothing is explicitely set.
In other words: the width/height of a layout will get set to the container width/height in case it is not defined explicitely via "width"/"height" attribute.
This will allow me to drop largest parts of constants_1920.xml and constants_2560.xml in estuary.

@ronie @BigNoid for testing, @ksooo perhaps for review?

### DIFF
--- a/xbmc/guilib/GUIListItemLayout.cpp
+++ b/xbmc/guilib/GUIListItemLayout.cpp
@@ -167,7 +167,7 @@ void CGUIListItemLayout::LoadControl(TiXmlElement *child, CGUIControlGroup *grou
   }
 }
 
-void CGUIListItemLayout::LoadLayout(TiXmlElement *layout, int context, bool focused)
+void CGUIListItemLayout::LoadLayout(TiXmlElement *layout, int context, bool focused, float maxWidth, float maxHeight)
 {
   m_focused = focused;
   layout->QueryFloatAttribute("width", &m_width);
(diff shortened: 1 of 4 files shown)


phil65:
this is the corresponding estuary change (not 100% complete yet)
https://github.com/phil65/skin.estuary/commit/c2d8179873181eca8e7f140e4e69980e9a167c21

ronie:
works fine on my end. can't spot any issues.

BigNoid:
Yup works afaict. The control inside the item/focusedlayout also inherent the parent width/height now so that is great.

xbmc/guilib/GUIListItemLayout.cpp
```diff
@@ -176,17 +176,20 @@ void CGUIListItemLayout::LoadLayout(TiXmlElement *layout, int context, bool focu
...
+
+  TiXmlElement *child = layout->FirstChildElement("control");
   while (child)
   {
     LoadControl(child, &m_group);
     child = child->NextSiblingElement("control");
   }
-  // ensure width and height are valid
-  m_width = std::max(1.0f, m_width);
-  m_height = std::max(1.0f, m_height);
```
ksooo: These two line imo should stay as maxWidth and/or maxHeight could still be zero, this m_width/m_height could also be. But i would move these two lines to line 184, above m_group.SetWidth(m_widgth).

phil65: wouldnt it make more sense to leave m_width/m_height at "0" for those? we could completely skip rendering for containers with width/height = 0 then for example..

phil65:
updated. jenkins build this please.

xbmc/guilib/GUIListItemLayout.cpp
phil65: question still valid, i will merge this without behaviour change for now though.

xbmc/guilib/GUIListItemLayout.cpp
```diff
@@ -176,17 +176,22 @@ void CGUIListItemLayout::LoadLayout(TiXmlElement *layout, int context, bool focu
...
     m_condition = g_infoManager.Register(condition, context);
   m_isPlaying.Parse("listitem.isplaying", context);
-  TiXmlElement *child = layout->FirstChildElement("control");
+  // ensure width and height are valid
+  if (!m_width)
+    m_width = maxWidth;
+  if (!m_height)
+    m_height = maxHeight;
+- m_width = std::max(1.0f, m_width);
+- m_height = std::max(1.0f, m_height);
```
Rechi: I think `-` at the beginning went in by accident.

phil65:
jenkins build this please.

PR #11230, opened 2016-12-20
[python] slider - add int/float support

ronie:
for slider controls, currently only the get/set percentage method is exposed to the python api.

this adds support for setting / getting float and int as well.
this allows addons to create a slider with (for instance) a range from -10 to 10.

@tamland / @phil65

### DIFF
--- a/xbmc/interfaces/legacy/Control.cpp
+++ b/xbmc/interfaces/legacy/Control.cpp
@@ -451,6 +451,34 @@ namespace XBMCAddon
         ((CGUISliderControl*)pGUIControl)->SetPercentage(pct);
     }
 
+    int ControlSlider::getInt()
+    {
+      return (pGUIControl) ? static_cast<CGUISliderControl*>(pGUIControl)->GetIntValue() : 0;
+    }
+
+    void ControlSlider::setInt(int value, int min, int delta, int max)
+    {
+      if (pGUIControl)
+        static_cast<CGUISliderControl*>(pGUIControl)->SetType(SLIDER_CONTROL_TYPE_INT);
+        static_cast<CGUISliderControl*>(pGUIControl)->SetRange(min, max);
+        static_cast<CGUISliderControl*>(pGUIControl)->SetIntInterval(delta);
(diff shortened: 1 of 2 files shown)


xbmc/interfaces/legacy/Control.h
```diff
@@ -2833,6 +2833,112 @@ namespace XBMCAddon
...
+#else
+      virtual int getInt();
+#endif
+
+#ifdef DOXYGEN_SHOULD_USE_THIS
+      ///
+      /// \ingroup python_xbmcgui_control_slider
+      /// @brief \python_func{ setInt(value, min, delta, max) }
+      ///-----------------------------------------------------------------------
+      /// Sets the value of the slider.
```
phil65: nitpick: setXX() not only sets the value, but also borders and delta and type. So better sth like "Set slider to integer mode and configure its min / max, delta and initial value"

ronie:
thx! updated as suggested.

ronie:
jenkins build this please

ronie:
jenkins build this please

xbmc/interfaces/legacy/Control.cpp
```diff
@@ -451,6 +451,34 @@ namespace XBMCAddon
...
+    int ControlSlider::getInt()
+    {
+      return (pGUIControl) ? static_cast<CGUISliderControl*>(pGUIControl)->GetIntValue() : 0;
+    }
+
+    void ControlSlider::setInt(int value, int min, int delta, int max)
+    {
+      if (pGUIControl)
+        static_cast<CGUISliderControl*>(pGUIControl)->SetType(SLIDER_CONTROL_TYPE_INT);
+        static_cast<CGUISliderControl*>(pGUIControl)->SetRange(min, max);
```
tamland: @ronie did you see the coverity report? c++ is a curly-bracket language...

ronie: thanx! https://github.com/xbmc/xbmc/pull/11441

PR #11225, opened 2016-12-20
[PVR] Fix crashes on pvr manager deinit/reinit - part 2

ksooo:
This is a followup to #11143 

Now, that first chunk of fixes is in master/Krypton, next (and hopefully the last) wave of followup issues arrives. First wave of fixes was definitely okay, but formerly, we did not get that far... ;-)

Crashes with stacks very similar to the following were reported on the forum (http://forum.kodi.tv/showthread.php?tid=298461&pid=2481973#pid2481973) and by @peak3d:

<pre>
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x006c98f8 in CDatabase::InitSettings (this=this@entry=0x5c12c500, dbSettings=...) at /home/neil/projects/LibreELEC.tv/build.LibreELEC-RPi2.arm-8.0-devel-debug/kodi-17.0-beta7-7581c4a/xbmc/dbwrappers/Database.cpp:404
[Current thread is 1 (Thread 0x56e233a0 (LWP 831))]

Thread 52 (Thread 0x66f023a0 (LWP 815)):
...
#29 0x007d0d28 in PVR::CPVRRecordings::CPVRRecordings (this=0x721141e8) at /home/neil/projects/LibreELEC.tv/build.LibreELEC-RPi2.arm-8.0-devel-debug/kodi-17.0-beta7-7581c4a/xbmc/pvr/recordings/PVRRecordings.cpp:51
=> PVR::CPVRManager::ResetProperties called PVR::CPVRManager::Clear() which deleted m_database and has set it to nullptr!
#30 0x007fb938 in PVR::CPVRManager::ResetProperties (this=this@entry=0x2ec7f50) at /home/neil/projects/LibreELEC.tv/build.LibreELEC-RPi2.arm-8.0-devel-debug/kodi-17.0-beta7-7581c4a/xbmc/pvr/PVRManager.cpp:281
#31 0x007fbb18 in PVR::CPVRManager::Start (this=0x2ec7f50) at /home/neil/projects/LibreELEC.tv/build.LibreELEC-RPi2.arm-8.0-devel-debug/kodi-17.0-beta7-7581c4a/xbmc/pvr/PVRManager.cpp:322
[...]

PR #11231, opened 2016-12-20
[PVR] fix crash stopping manager: addons should be destroyed as soon as possible

janbar:
Fix crash every time kodi stop running pvr manager.

Program terminated with signal SIGSEGV, Segmentation fault.
0  __GI___pthread_mutex_lock (mutex=0x0) at ../nptl/pthread_mutex_lock.c:66
1  0x0000000001085ad3 in PVR::CPVRManager::QueueJob(CJob*) ()
2  0x00007f5328529794 in PVRClientMythTV::RunHouseKeeping() ()
3  0x00007f5328529968 in PVRClientMythTV::HandleBackendMessage(Myth::shared_ptr<Myth::EventMessage>)

Addons callback have to be closed before freeing of PVRManager resources.

### DIFF
--- a/xbmc/pvr/PVRManager.cpp
+++ b/xbmc/pvr/PVRManager.cpp
@@ -347,6 +347,10 @@ void CPVRManager::Stop(void)
     CApplicationMessenger::GetInstance().SendMsg(TMSG_MEDIA_STOP);
   }
 
+  /* destroy addons */
+  m_addons->Unload();
+  CLog::Log(LOGNOTICE, "PVRManager - clients unloaded");
+
   /* stop all update threads */
   SetState(ManagerStateInterrupted);
 


ksooo:
@FernetMenta i think this is wrong, but you are the expert.

janbar:
Hi @ksooo, I have a full bin of crash log since few months ago, and I decided to close the filler ;-). This commit fixes the issue, maybe the design is not as you are awaiting.

ksooo:
> This commit fixes the issue, maybe the design is not as you are awaiting.

 pvr manager does not own the addons, thus the pvr manager  is not allowed to unload the addons. solution is imo to stop accepting callbacks early.

ksooo:
> I have a full bin of crash log since few months ago

 good to know that the issue is not caused by my latest pvr manager changes, then. Does the problem still occur with very latest master / krypton build (without your patch, ofc), btw?

janbar:
Yes I fixed it since a couple of hours

ksooo:
Try this instead of your fix (untested):

<pre>
diff --git a/xbmc/pvr/PVRManager.cpp b/xbmc/pvr/PVRManager.cpp
index 412f9dc..036642d 100644
--- a/xbmc/pvr/PVRManager.cpp
+++ b/xbmc/pvr/PVRManager.cpp
@@ -268,9 +268,12 @@ void CPVRManager::Clear(void)
   m_bIsSwitchingChannels  = false;
   m_bEpgsCreated = false;
 
-  for (unsigned int iJobPtr = 0; iJobPtr < m_pendingUpdates.size(); iJobPtr++)
-    delete m_pendingUpdates.at(iJobPtr);
-  m_pendingUpdates.clear();
+  {
+    CSingleLock lock(m_critSectionTriggers);
+    for (unsigned int iJobPtr = 0; iJobPtr < m_pendingUpdates.size(); iJobPtr++)
+      delete m_pendingUpdates.at(iJobPtr);
+    m_pendingUpdates.clear();
+  }
 
   /* unregister application action listener */
   {
</pre>

janbar:
The crash is caused by `CSingleLock lock(m_critSectionTriggers)`. I try it but I have a bad feeling.

ksooo:
> The crash is caused by CSingleLock lock(m_critSectionTriggers). I try it but I have a bad feeling.

It crashes on exit of kodi, not on enabling/disabling or installing/uninstalling pvr client addon, right? Problem is that the pvr manager instance is already destroyed when the addon callback arrives. Then you might be right and my fix suggestion does not help. Unfortunately I'm convinced that your fix approach (although it fixes the crash) is also not correct as it imo violates addon system software architecture principle that only the addon manager loads/unloads addons and because it has unwanted side effects when stopping the pvr manager to restart it (means not on kodi app exit).

janbar:
In fact , your fix works too. Maybe empty the queue could be enougth to fix it.

janbar:
Let me know if you prefer your solution to close it.

ksooo:
Yes, I'd prefer my solution (added to #11225 now). Please close this PR.

FernetMenta:
@ksooo thanks, your fix look good.

janbar:
I closed it. But IMO it is a chance the addon callback hold the lock just before the clear. In others circumstances I haven't yet fully tested. I will let you know and I keep it warm ;).

PR #11232, opened 2016-12-20
[PVR] fix crash stopping manager: addons should be destroyed as soon as possible

janbar:
Fix crash every time kodi stop running pvr manager.

Program terminated with signal SIGSEGV, Segmentation fault.
0  __GI___pthread_mutex_lock (mutex=0x0) at ../nptl/pthread_mutex_lock.c:66
1  0x0000000001085ad3 in PVR::CPVRManager::QueueJob(CJob*) ()
2  0x00007f5328529794 in PVRClientMythTV::RunHouseKeeping() ()
3  0x00007f5328529968 in PVRClientMythTV::HandleBackendMessage(Myth::shared_ptr<Myth::EventMessage>)

Addons callback have to be closed before freeing of PVRManager resources.

### DIFF
--- a/xbmc/pvr/PVRManager.cpp
+++ b/xbmc/pvr/PVRManager.cpp
@@ -348,6 +348,10 @@ void CPVRManager::Stop(void)
     CApplicationMessenger::GetInstance().SendMsg(TMSG_MEDIA_STOP);
   }
 
+  /* destroy addons */
+  m_addons->Unload();
+  CLog::Log(LOGNOTICE, "PVRManager - clients unloaded");
+
   /* stop all update threads */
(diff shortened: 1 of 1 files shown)


janbar:
I created the backport with PR#11231

ksooo:
As discussed in the original PR, I'd prefer my solution (added to #11225 now, will be backported when the original PR is ready and well tested). Please close this PR.

janbar:
Okay I closed it. But IMO it is a chance the addon callback hold the lock just before the clear. In others circumstances I haven't yet fully tested. I will let you know and I keep it warm ;).

PR #11236, opened 2016-12-21
[gui] add option to only show full lines for textboxes

phil65:
Next scaling-related one:
Add an option for skinners to only show full lines for textboxes. Esp. useful when havin several fontsets with different sizing.
` <showfulllines>true</showfulllines>`

I would even be tempted to make this behaviour default since in 95% of cases cut-off textboxes are not wanted I think.

### DIFF
--- a/xbmc/guilib/GUITextBox.cpp
+++ b/xbmc/guilib/GUITextBox.cpp
@@ -216,7 +216,7 @@ void CGUITextBox::Render()
       {
         float textHeight = m_font->GetTextHeight(std::min((unsigned int)m_lines.size(), m_itemsPerPage));
 
-        if (textHeight <= m_renderHeight)
+        if (textHeight < m_renderHeight)
           posY += (m_renderHeight - textHeight) * 0.5f;
       }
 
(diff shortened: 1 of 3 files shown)


BigNoid:
+1 for default to true

ronie:
agreed, no need to be optional imo.

phil65:
changed default. jenkins build this please.

phil65:
updated PR because first try broke vertically centered textboxes in some cases. For now I disabled showFullLines for vertically centered textboxes because that requires quite some more work.

MartijnKaijser:
ping

ruuk:
+1 and optional should remain. For instance, there are designs where the text is bounded on the bottom by a window element or the screen edge where it is better to have the partial text cut right at the boundary.

Rechi:
@phil65 this needs a rebase

MartijnKaijser:
Who can we ask for review? @xhaggi perhaps?

xbmc/guilib/GUITextBox.cpp
```diff
@@ -108,8 +110,12 @@ void CGUITextBox::UpdateInfo(const CGUIListItem *item)
   float textHeight = m_font ? m_font->GetTextHeight(m_lines.size()) : m_itemHeight * m_lines.size();
   float maxHeight = m_height ? m_height : textHeight;
   m_renderHeight = m_minHeight ? CLAMP(textHeight, m_minHeight, maxHeight) : m_height;
-  m_itemsPerPage = (unsigned int)(m_renderHeight / m_itemHeight);
-
+  m_itemsPerPage = static_cast<unsigned int>(m_renderHeight / m_itemHeight);
+  if (m_showFullLines && !(m_label.align & XBFONT_CENTER_Y))
+  {
+	unsigned int visibleLines = std::min(static_cast<unsigned int>(m_lines.size()), m_itemsPerPage);
```
xhaggi: wrong indentation?

xhaggi:
I tend to rename the property to `cutlastline` default: false. After that I'll take another look.

phil65:
updated.

phunkyfish:

PR #11237, opened 2016-12-21
[cmake|win32] allow webserver to be disabled on windows

Rechi:
## Description

The defines `HAS_WEB_SERVER` and `HAS_WEB_INTERFACE` are already defined if `HAVE_LIBMICROHTTPD` is defined. There is no need to generally set it for windows.
`TestWebServer` must only be compiled if the HTTPRequestHandlers are build.

## How Has This Been Tested?

Compiled on Windows with disabled microhttpd.

## Checklist:


fetzerch (approved):
Thanks! When I started with cmake on windows, I found a few places with #pragma lib that might have become obsolete now.

fetzerch:
jenkins build and merge

PR #11239, opened 2016-12-21
[keyboard] improvements to German keyboard layout

da-anda:
## Description
Improves the layout of the German OSD keyboard by adding often used special chars to the default letter views like it has already been done for the English layout.

In addition, it's adding a ABC layout for German.

One change that is subject to discussion is, that I changed the number row to switch to special chars when using the SHIFT modifier, just like on regular keyboards. I'm not aware of another layout that does it, but felt it would make sense.

## Motivation and Context
current layout was annoying when typing mail-addresses or UNC paths

## How Has This Been Tested?
runtime tested

## Checklist:

### DIFF
--- a/system/keyboardlayouts/german.xml
+++ b/system/keyboardlayouts/german.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+﻿<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <!--
 Please use English language names instead.
 Default font lacks support for all characters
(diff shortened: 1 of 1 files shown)


da-anda:
ping @ksooo @phil65 - what do you think?

ksooo:
Nice. Thanks.

system/keyboardlayouts/german.xml
```diff
@@ -9,13 +9,33 @@ Default font lacks support for all characters
       <row>1234567890ß</row>
       <row>qwertzuiopü</row>
```
Rechi: maybe add `+` as last button (and the other ones in the rows above and below), than it looks more like the real keyboard

da-anda: I think an additional column could break certain skins

PR #11240, opened 2016-12-21
[keyboard] improvements to German keyboard layout

da-anda:
backport of #11239


da-anda:
good to go for Krypton @MartijnKaijser ?

PR #11242, opened 2016-12-21
[GUI] remember last used keyboard layout

da-anda:
When having multiple keyboard layouts selected, the OSD keyboard will always show the first layout in the list. With this change, Kodi will remember the last used keyboard layout and will use that instead.

## Description
A hidden setting has been added which will be updated each time the layout is being changed.

## Motivation and Context
There is currently no way to define a preferred keyboard layout. Even when you select the keyboard layouts in a specific order, they will always be stored in alphabetical order. Remembering the last used layout makes a custom order superflous and is even more flexible.

## How Has This Been Tested?
runtime tested

## Types of change
added hidden setting to store the last active keyboard layout

### DIFF
--- a/xbmc/dialogs/GUIDialogKeyboardGeneric.cpp
+++ b/xbmc/dialogs/GUIDialogKeyboardGeneric.cpp
@@ -133,12 +133,17 @@ void CGUIDialogKeyboardGeneric::OnInitWindow()
   m_layouts.clear();
   const KeyboardLayouts& keyboardLayouts = CKeyboardLayoutManager::GetInstance().GetLayouts();
   std::vector<CVariant> layoutNames = CServiceBroker::GetSettings().GetList(CSettings::SETTING_LOCALE_KEYBOARDLAYOUTS);
+  std::string activeLayout = CServiceBroker::GetSettings().GetString(CSettings::SETTING_LOCALE_ACTIVEKEYBOARDLAYOUT);
 
   for (std::vector<CVariant>::const_iterator layoutName = layoutNames.begin(); layoutName != layoutNames.end(); ++layoutName)
   {
     KeyboardLayouts::const_iterator keyboardLayout = keyboardLayouts.find(layoutName->asString());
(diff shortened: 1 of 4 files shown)


xbmc/dialogs/GUIDialogKeyboardGeneric.cpp
```diff
@@ -133,12 +133,17 @@ void CGUIDialogKeyboardGeneric::OnInitWindow()
...
+  std::string activeLayout = CServiceBroker::GetSettings().GetString(CSettings::SETTING_LOCALE_ACTIVEKEYBOARDLAYOUT);
 
   for (std::vector<CVariant>::const_iterator layoutName = layoutNames.begin(); layoutName != layoutNames.end(); ++layoutName)
   {
     KeyboardLayouts::const_iterator keyboardLayout = keyboardLayouts.find(layoutName->asString());
     if (keyboardLayout != keyboardLayouts.end())
+    {
       m_layouts.push_back(keyboardLayout->second);
+      if (layoutName->asString() == activeLayout)
+        m_currentLayout = m_layouts.size() - 1;
```
da-anda: I suppose there is a nicer way to get the index?

Memphiz:
I think its fine as is. There is no other way to get the index as long as those are vectors and not dictionaries.

da-anda:
thanks for review. I'll wait until sunday, and will merge then if nobody objects.

@MartijnKaijser trivial enough for a backport?

phil65:
@Memphiz Wouldn't std::distance() work here ?

MartijnKaijser:
ping @Montellese if he knows another way

Memphiz:
@phil65 size() is already a shortcut to distance() as of the vector.size() docu...

Memphiz:
Also there is nothing wrong with that way - the for loop is already there - size has constant complexity - i find no reason why this should be unclean...

phil65:
sure. @Montellese pointed me to std::distance() some weeks ago for very similar use case, that´s why I was askin.
All fine then, feature makes sense.

PR #11247, opened 2016-12-22
FFmpeg: Bump to final 3.1.6-Krypton

fritsch:
This should be the version we ship with Krypton. Needs backport. After that master should go into 3.2 direction as we will hit 3.3 soon.


fritsch:
jenkins build this please

PR #11183, opened 2016-12-15
misc cmake fixes

stefansaraev:
## Description

as title says.

## Motivation and Context

I needed this to compile kodi for my OE fork

## Types of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [x] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #11250, opened 2016-12-22
[cmake] Fix packaging of binary addons (324325th edition)

fetzerch:
Not sure how this got introduced, but I want to rework the packaging either way now that autotools is gone.

Krypton is fine, no backport needed.

Jenkins build was fine as well: http://jenkins.kodi.tv/job/Android-ARM/10424/


MartijnKaijser:
@fetzerch that line was removed on purpose before in https://github.com/xbmc/xbmc/pull/11183 cause android failed to build.
Now you see it failing again with no addons build (like jenkins does by default)
http://jenkins.kodi.tv/job/Android-ARM/10431/

PR #11254, opened 2016-12-23
[gui] fix crash when using container.property() in non-mediawindows

phil65:
Quite an old bug: Kodi completely crashes when using container(xx).property(yy) in non-mediawindows (root cause: non-mediawindow is casted to CGUIMediaWindow)

Example crash trigger: https://github.com/phil65/skin.estuary/commit/e259a49052c6e7db8a6a9164c1bbe7a98b05a658

Atm container.property() can only work for media windows because our CGUIBaseContainer class only carries a vector of listitems, and not a CFileItemList object.

EDIT: appended a second commit for container(xx).content crash, same reason as above.

### DIFF
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -8076,17 +8076,9 @@ std::string CGUIInfoManager::GetMultiInfoLabel(const GUIInfo &info, int contextW
     return GetMusicPlaylistInfo(info);
   else if (info.m_info == CONTAINER_PROPERTY)
   {
-    CGUIWindow *window = NULL;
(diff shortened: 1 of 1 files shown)


BigNoid:
I don't think there's ```Container(xx).Property(yy)``` functionality? In that example commit you replace ```Container.ListItem.Property``` with ```Container.Property```. I thought the recent additions allowed you to omit the container part, not the listitem part.

phil65:
Yeah that example is just there to have a test case,  it doesn't make any sense in terms of functionality.

phil65:
jenkins build this please.

PR #11257, opened 2016-12-23
[cmake/depends] Add info about Android SDK/NDK to Toolchain_binaddons…

fetzerch:
….cmake.in

Copied from Toolchain.cmake.in. This is necessary to build libretro cores with the Android NDK where we need access to NDKROOT/ndk-build.

@garbear: please pick


wsnipex (approved):
lgtm

PR #11258, opened 2016-12-23
[Python] add setContent for mediawindows

ronie:
allow addons that use WindowXML to create a mediawindow to set the content type of the container.

@tamland / @phil65

### DIFF
--- a/xbmc/interfaces/legacy/WindowXML.cpp
+++ b/xbmc/interfaces/legacy/WindowXML.cpp
@@ -271,6 +271,13 @@ namespace XBMCAddon
       A(m_vecItems)->SetProperty(key, value);
     }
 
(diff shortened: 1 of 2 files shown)


xbmc/interfaces/legacy/WindowXML.cpp
```diff
@@ -271,6 +271,12 @@ namespace XBMCAddon
       A(m_vecItems)->SetProperty(key, value);
     }
 
+    void WindowXML::setContent(const String& value)
+    {
+      XBMC_TRACE;
```
phil65: also LOCKGUI;

ronie: done, thx!

ronie:
jenkins build this please

PR #11263, opened 2016-12-24
[PVR] Fix guide window: do not jump to grid start on channel group change; go to 'now' instead

ksooo:
Reported in the forum (http://forum.kodi.tv/showthread.php?tid=301765). Annoying bug. 

Additionally, I added a keyboard shortcut key for switching channel groups as it is very complicated to do otherwise (scroll left to begin of grid, open side blade, select new channel group, close side blade).

@Jalle19 mind doing the review


Jalle19 (approved):

ksooo:
jenkins build this please

PR #11267, opened 2016-12-24
Websocket connection header is case-insensitive

emlove:
## Description
This change makes the websocket connection header check case insensitive. Per [RFC 6455, 4.2.1.4](https://tools.ietf.org/html/rfc6455#section-4.2.1), this check should be case insensitive.

## Motivation and Context
Not all websocket clients will send the connection header using the same case that Kodi is expecting. See KeepSafe/aiohttp#1495.

## How Has This Been Tested?
This has been tested by connecting to Kodi using Python's aiohttp v1.2.0, using the following test script:

```python
import aiohttp
import asyncio

@asyncio.coroutine
def main(loop):
    client = aiohttp.ClientSession(loop=loop)
    ws = yield from client.ws_connect('ws://192.168.1.193:9090/jsonrpc')
    while True:
        msg = yield from ws.receive()
        if msg.type == aiohttp.WSMsgType.TEXT:
            print(msg)
        elif msg.type == aiohttp.WSMsgType.CLOSED:
            break
        elif msg.type == aiohttp.WSMsgType.ERROR:
            break

loop = asyncio.get_event_loop()
loop.run_until_complete(main(loop))
```

## Types of change

## Checklist:

### DIFF
--- a/xbmc/network/websocket/WebSocketV13.cpp
+++ b/xbmc/network/websocket/WebSocketV13.cpp
@@ -102,7 +102,7 @@ bool CWebSocketV13::Handshake(const char* data, size_t length, std::string &resp
 
   // There must be a "Connection" header with the value "Upgrade"
   value = header.getValue(WS_HEADER_CONNECTION_LC);
-  if (value == NULL || strstr(value, WS_HEADER_UPGRADE) == NULL)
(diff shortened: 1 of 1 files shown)


Montellese (approved):

Montellese:
jenkins build and merge

MartijnKaijser:
jenkins build this please

PR #11270, opened 2016-12-25
[PVR] PVR windows: Fix item selection on window update.

ksooo:
After deleting a recording/timer/timer rule/channel using the respective pvr window, the first item in the item list was selected. This is very annoying behavior, for instance if you want to delete some items in a row. Example: you have ten items and want to delete item 5 to 10:

Old behavior:
- travel to item 5
- delete item 5 => item disappears, item 1 gets selected
- travel down from item 1 to the new item 5
- delete item 5 => item disappears, item 1 gets selected
- travel down from item 1 to the new item 5
...

New behavior:
- travel to item 5
- delete item 5 => item disappears, new item 5 gets selected
- delete new item 5 => item disappears, new item 5 gets selected
...
- delete item 5 that was originally item 10 => item disappears, item 4 gets selected

Runtime-tested on latest linux and macOS master.

@Jalle19 mind doing the review

### DIFF
--- a/xbmc/pvr/windows/GUIWindowPVRBase.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRBase.cpp
@@ -348,7 +348,29 @@ bool CGUIWindowPVRBase::Update(const std::string &strDirectory, bool updateFilte
     return false;
   }
 
-  return CGUIMediaWindow::Update(strDirectory, updateFilterPath);
+  int iOldCount = m_vecItems->GetObjectCount();
+  int iSelectedItem = m_viewControl.GetSelectedItem();
+  const std::string oldPath = m_vecItems->GetPath();
+
+  bool bReturn = CGUIMediaWindow::Update(strDirectory, updateFilterPath);
+
+  if (bReturn &&
+      iSelectedItem != -1 && // something must have been selected
+      iOldCount > 0) // more than only ".." item or nothing in previous item list
+  {
+    int iNewCount = m_vecItems->GetObjectCount();
+    if (iOldCount > iNewCount && // at least one item removed by Update()
+        oldPath == m_vecItems->GetPath()) // update not due changing into another folder
+    {
+      // restore selected item if we just deleted one or more items.
+      if (iSelectedItem >= iNewCount)
+        iSelectedItem = iNewCount - 1;
+
+      m_viewControl.SetSelectedItem(iSelectedItem);
+    }
+  }
+
+  return bReturn;
 }
 
 void CGUIWindowPVRBase::UpdateButtons(void)


ksooo:
ping @Jalle19

xhaggi:
@ksooo could you please check why it works on our movie window. If you delete a movie from your database the selected position will stay the same. We should check it there before we add something like that. Thanks.

Jalle19:
Seems I forgot to actually submit my review

ksooo:
@xhaggi I have checked that. If I got it right, it works for movies because there CGUIMediaWindow::OnDeleteItem (https://github.com/xbmc/xbmc/blob/master/xbmc/windows/GUIMediaWindow.cpp#L1440) is used, which only works for real files. 

PVR is imo really different, not just because timers and recordings are no files on local disk, but also because changes are also triggered by the backend. There is not always a UI action triggering the action. Thus i consider it perfectly okay to check item count before and after refresh and to set the selection accordingly. If not this logic belongs to CMediaWindow.

xbmc/pvr/windows/GUIWindowPVRRecordings.cpp
```diff
@@ -144,13 +145,25 @@ bool CGUIWindowPVRRecordings::Update(const std::string &strDirectory, bool updat
     }
   }
 
-  if (bReturn && iOldCount > 0 && m_vecItems->GetObjectCount() == 0 && oldPath == m_vecItems->GetPath())
```
ksooo: Sorry, what? GetSize() where?

xbmc/pvr/windows/GUIWindowPVRChannels.cpp
```diff
@@ -99,6 +103,16 @@ bool CGUIWindowPVRChannels::Update(const std::string &strDirectory, bool updateF
       Update(GetDirectoryPath());
     }
   }
+
+  if (bReturn && iOldCount > m_vecItems->GetObjectCount() && oldPath == m_vecItems->GetPath())
+  {
+    // restore selected item if we just deleted one or more items.
```
ksooo: it's just three lines, but np, can do.

xhaggi: Why not move that code into the base class?

xhaggi:
@ksooo thanks for taking a look at the movie window. We should think about a store/restore position logic in the base class. Two nethods like we have for the view stuff.

xbmc/pvr/windows/GUIWindowPVRRecordings.cpp
phil65: @Jalle19 	means m_vecItems->Size() I think.

xbmc/pvr/windows/GUIWindowPVRChannels.cpp
ksooo: Already done (locally) as @Jalle19 suggested the same. :-)

xbmc/pvr/windows/GUIWindowPVRRecordings.cpp
ksooo: GetObjectCount() is imo the right choice here, as it in opposite to Size() subtracts parent folder items (".." items) from the count returned.

ksooo:
refactored: added CGUIWindowPVRBase::SetSelectedItem()

ksooo:
jenkins build this please

ksooo:
@Jalle19 @xhaggi Completely reworked. Now everything is in base class. Good to go now?

ksooo:
jenkins build this please

xhaggi:
Looks better now, thanks.

ksooo:
jenkins build this please

PR #11273, opened 2016-12-26
Recalculate SyncClock after reconfigure

peak3d:
Recalculate SyncClock after reconfigure

## Description
On Streamchange the Sync-Clock stays disabled, if the frame rate does not change.
This PR checks at configure time, if the sync clock must be setup.


FernetMenta:
thanks

PR #11275, opened 2016-12-26
System > Audio option "Maintain original volume on downmix" wording change

joethefox:
Freely taken from the ATV4 settings, I thought that this option can be worded in a less technical form as the ATV4 does.

## Motivation and Context
reduce technical jargon where we can

## Types of change

## Checklist:

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -1590,7 +1590,7 @@ msgstr ""
 
 #: system/settings/settings.xml
 msgctxt "#346"
-msgid "Maintain original volume on downmix"
+msgid "Reduce loud sounds"
 msgstr ""
 
 #: system/settings/settings.xml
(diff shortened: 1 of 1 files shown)


FernetMenta:
that's absolutely not what it does, you even inverted the meaning. the description "Reduce loud sounds" does not even come close the function "Maintain original volume on downmix"

note: originally this was called "bump volume on downmix". you want to rename it to "reduce lound sounds" :)

joethefox:
ok, this shows that I didn't understand the option 😳 

I analized the results -> if enabled then the voices are at the same level of the sound effect, if disabled the sound effects are louder than the voices. So without looking what happens under the hood, from my understand, it level the sound effects to the same level of the voices without considering who levels with who.

joethefox:
It's clear that if this change makes more confusion then the goal is not reached

zag2me:
This setting is one of my hates with Kodi, I don't think it has any use, I remember trying it thinking it would boost the center channel speech which is what 99% of people will want to do with a 2.0 setup, but it seems far more complicated beyond any normal users understanding. I tried it on and off and couldn't tell the difference.

FernetMenta:
this setting is absolutely useful, in particular for tv where audio often changes from 5.1 to 2.0 and back. without maintaining the volume on 5.1, you notice a bump of volume on commercial breaks.

garbear:
If you change it to "Keep original volume on audio change" it would be more clear and less technical

joethefox:
After this commit https://github.com/xbmc/xbmc/commit/fd1825c6cf2b5186204fb93e03f8311f48e07963 the logic has been corrected (it was inverted previously) and now when the option is disabled the sound is softer when playing a 5.1 audio and louder when playing a 2.0 (as fernetmenta says), when enabled both 5.1 and 2.0 are at the same volume level. 
Now I'm convinced that this option has nothing to do with my proposal of change.

da-anda:
> this setting is absolutely useful, in particular for tv where audio often changes from 5.1 to 2.0 and back. without maintaining the volume on 5.1, you notice a bump of volume on commercial breaks.

this is the first time I really understood what this setting is supposed to do, thanks. :)

PR #11277, opened 2016-12-27
wait for network after resuming from suspend

verybadsoldier:
Makes Kodi wait for network connection after resuming from suspend.

## Description
This introduces a function `WaitForNet` in Kodi's power manager which will block (with a timeout) until at least one network interface is initialized and connected. This function is invoked in `OnWake` so it is triggered on every wake up.
This behavior is disabled by default and has to be activated by a new advanced setting:
```
    <suspend>
        <waitnetonwakeup>true</waitnetonwakeup>
    </suspend>
```

This new setting would have to be documented on the Wiki I guess.

## Motivation and Context
When waking the machine from sleep/suspend and directly trying to access network resources (e.g. playing a video or reading from the MySQL DB) then Kodi will fail cause the network has not been initialized yet.
I have experienced and described this issue here:
http://forum.kodi.tv/showthread.php?tid=301976

## How Has This Been Tested?
I tested it on a Intel NUC i5 using LibreELEC 7.02 using a wired network. I tested it by suspending the machine and waking it up again. Then I observed the behavior with and without this setting enabled. With the setting enabled Kodi blocked for several seconds and put something like this in the logfile:

```
12:51:10 T:140441393608768  NOTICE: OnWake: Running resume jobs
12:51:10 T:140441393608768  NOTICE: WaitForNet: Waiting for first network interface to come up
12:51:18 T:140441393608768  NOTICE: WaitForNet: network interface 'eth0' is up after waiting 7400 ms
12:51:18 T:140441393608768  NOTICE: OnWake: Restarting lirc
```

After this I could use Kodi and directly access network resources without issues.

## Types of change

## Checklist:

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -20394,3 +20394,14 @@ msgstr ""
 msgctxt "#39010"
 msgid "Select sort method"
 msgstr ""
+
+#: system/settings/settings.xml
+msgctxt "#39011"
+msgid "Maximum wait time for network"
+msgstr ""
+
+#. Description of setting with label #39011 "Maximum wait time for network"
+#: system/settings/settings.xml
+msgctxt "#39012"
+msgid "Set the maximum time to wait for the network to come up after waking up. When time has passed before network is up startup will continue."
+msgstr ""
--- a/xbmc/powermanagement/PowerManager.cpp
+++ b/xbmc/powermanagement/PowerManager.cpp
@@ -32,6 +32,7 @@
 #include "guilib/LocalizeStrings.h"
 #include "interfaces/AnnouncementManager.h"
 #include "interfaces/builtins/Builtins.h"
+#include "network/Network.h"
 #include "pvr/PVRManager.h"
 #include "settings/lib/Setting.h"
 #include "settings/Settings.h"
@@ -55,6 +56,10 @@
 #elif defined(TARGET_WINDOWS)
 #include "powermanagement/windows/Win32PowerSyscall.h"
 extern HWND g_hWnd;
+#endif
+
+#if defined(TARGET_POSIX)
+#include "linux/XTimeUtils.h"
 #endif
 
 using namespace ANNOUNCEMENT;
@@ -270,6 +275,12 @@ void CPowerManager::OnWake()
 {
   CLog::Log(LOGNOTICE, "%s: Running resume jobs", __FUNCTION__);
 
+  {
+    const int waitNetTimeout = CServiceBroker::GetSettings().GetInt(CSettings::SETTING_POWERMANAGEMENT_WAITNETAFTERWAKEUP);
+    if (waitNetTimeout > 0)
+      WaitForNet(waitNetTimeout);
+  }
+
   // reset out timers
   g_application.ResetShutdownTimers();
 
@@ -326,4 +337,32 @@ void CPowerManager::SettingOptionsShutdownStatesFiller(const CSetting *setting, 
     list.push_back(make_pair(g_localizeStrings.Get(13014), POWERSTATE_MINIMIZE));
 #endif
   }
+}
+
+void CPowerManager::WaitForNet(int timeout) const
+{
+  CNetwork& net = g_application.getNetwork();
+
+  // check if we have at least one network interface to wait for
+  if (!net.IsAvailable())
+    return;
+
+  CLog::Log(LOGNOTICE, "%s: Waiting for a network interface to come up (Timeout: %d s)", __FUNCTION__, timeout);
+
+  const static int intervalMs = 200;
+  const int numMaxTries = (timeout * 1000) / intervalMs;
+
+  for(int i=0; i < numMaxTries; ++i)
+  {
+    if (i > 0)
+      Sleep(intervalMs);
+
+    if (net.IsConnected())
+    {
+      CLog::Log(LOGNOTICE, "%s: A network interface is up after waiting %d ms", __FUNCTION__, i * intervalMs);
+      return;
+    }
+  }
+
+  CLog::Log(LOGNOTICE, "%s: No network interface did come up within %d s... Giving up...", __FUNCTION__, timeout);
 }
--- a/xbmc/powermanagement/PowerManager.h
+++ b/xbmc/powermanagement/PowerManager.h
@@ -93,6 +93,8 @@ private:
 
   void OnLowBattery();
 
+  void WaitForNet(int timeout) const;
+
   IPowerSyscall *m_instance;
 };
 
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -2890,6 +2890,18 @@
           </constraints>
           <control type="list" format="string" />
         </setting>
+        <setting id="powermanagement.waitnetafterwakeup" type="integer" label="39011" help="39012">
+          <level>2</level>
+          <default>0</default>
+          <constraints>
+            <minimum label="351">0</minimum>
+            <step>1</step>
+            <maximum>30</maximum>
+          </constraints>
+          <control type="spinner" format="string">
+            <formatlabel>14045</formatlabel>
+          </control>
+        </setting>
       </group>
       <group id="2" label="14256">
         <setting id="powermanagement.wakeonaccess" type="boolean" label="13026" help="36350">
--- a/xbmc/settings/Settings.cpp
+++ b/xbmc/settings/Settings.cpp
@@ -392,6 +392,7 @@ const std::string CSettings::SETTING_POWERMANAGEMENT_DISPLAYSOFF = "powermanagem
 const std::string CSettings::SETTING_POWERMANAGEMENT_SHUTDOWNTIME = "powermanagement.shutdowntime";
 const std::string CSettings::SETTING_POWERMANAGEMENT_SHUTDOWNSTATE = "powermanagement.shutdownstate";
 const std::string CSettings::SETTING_POWERMANAGEMENT_WAKEONACCESS = "powermanagement.wakeonaccess";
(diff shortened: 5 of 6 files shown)


FernetMenta:
no advanced settings. those are only valuable for the very few who know about them. mostly only the developer who implemented it.

verybadsoldier:
@FernetMenta 
Okay, so it should be a regular setting configurable in the UI? I will try to find out how to do this.

So the general concept is accetable and might get merged if the setting is changed?

xbmc/powermanagement/PowerManager.cpp
```diff
@@ -327,3 +333,25 @@ void CPowerManager::SettingOptionsShutdownStatesFiller(const CSetting *setting,
...
+void CPowerManager::WaitForNet()
+{
+  CLog::Log(LOGNOTICE, "%s: Waiting for first network interface to come up", __FUNCTION__);
+
+  const static int maxLoopCount = 50;
+  const static int sleepTimeMs = 200;
+
+  for(unsigned i=0; i < maxLoopCount; ++i)
+  {
+    CNetworkInterface* pIface = g_application.getNetwork().GetFirstConnectedInterface();
```
FrankHeimes: Consider replacing `CNetworkInterface*` with `const auto * const` if `GetName()` is `const` or `const auto` otherwise.

xbmc/powermanagement/PowerManager.h
```diff
@@ -93,6 +93,8 @@ class CPowerManager : public IPowerEventsCallback
 
   void OnLowBattery();
 
+  void WaitForNet();
```
FrankHeimes: Consider declaring `WaitForNet() const`.

xbmc/settings/AdvancedSettings.h
```diff
@@ -383,6 +383,8 @@ class CAdvancedSettings : public ISettingCallback, public ISettingsHandler
 
     std::string m_userAgent;
 
+    bool m_waitForNetAfterWakeUp;
```
FrankHeimes: Consider initializing member variable when declaring it: `bool m_waitForNetAfterWakeUp = true;`

xbmc/settings/AdvancedSettings.cpp
```diff
@@ -1199,6 +1199,12 @@ void CAdvancedSettings::ParseSettingsFile(const std::string &file)
       m_seekSteps.push_back(atoi((*it).c_str()));
   }
 
+  pElement = pRootElement->FirstChildElement("suspend");
+  if (pElement)
+  {
+    XMLUtils::GetBoolean(pElement, "waitnetonwakeup", m_waitForNetAfterWakeUp);
+  }
```
FrankHeimes: `CAdvancedSettings::ParseSettingsFile()` is ridiculously long and should be refactored into local free functions.
That is too much work for a single feature implementation, but not adding more inline code would be a start.

Consider putting lines 1202 - 1206 into a free function
```
namespace {
	bool GetWaitNetOnWakeupFlag(TiXmlElement *rootElement)
	{
		if (!rootElement)
			throw std::invalid_argument("rootElement");

		auto firstChild = rootElement->FirstChildElement("suspend");
		if (!firstChild)
			return true; // Default behavior is to enable waitnetonwakeup

		bool waitForNetAfterWakeUp = true;
		XMLUtils::GetBoolean(firstChild, "waitnetonwakeup", waitForNetAfterWakeUp);
		return waitForNetAfterWakeUp;
	}
}
```
The line `1202: m_waitForNetAfterWakeUp = GetWaitNetOnWakeupFlag();` makes clear that the member is always well defined. Now, this is a lot more code than the original, but it is self-contained. `pRootElement` is checked for NULL, but this happens in line 444, i.e. 750+ lines above. Any (future changes) in between could invalidate it.

BTW.: IMHO, `XMLUtils` should provide function-style Getters, too, that don't require handing over a reference variable.

Yes, I'm new to this group and I do the code review like I do at work. Am I pushing too hard?

FrankHeimes (changes requested):

xbmc/settings/AdvancedSettings.cpp
MartijnKaijser: We welcome constructive reviews 👍

FernetMenta:
@verybadsoldier for what reason should we wait not always for network to be ready?

verybadsoldier:
@FernetMenta
For me it would be fine if it was just hardwired but I am afraid it could be a problem for systems without network (e.g. no NIC at all or no cable connected). Also there might be users who run Kodi exclusively from local storage and do not care for network connections. 
In both cases an unwanted delay would be introduced on wakeup. In my case it takes about 7 seconds for the network to beccome available.

verybadsoldier:
Reworked so that this only kicks in if there is at least one network interface. This will prevent waiting for network when there are no network Interfaces in the system. So maybe I really could just remove the advanced setting and make waiting for network after wakeup the standard behavior.

MartijnKaijser:
Regardless an advancedsettings is out of the question here

t4-ravenbird:
The original problem that this PR seeks to solve is already possible to fix by enabling the built-in [wake-on-lan](http://kodi.wiki/view/wake%20on%20lan)

That solution works the same when starting up, and when resuming from suspend. There have been MySQL issues in both cases and they are both solved when using wake-on-lan, for instance ;
http://forum.kodi.tv/showthread.php?tid=213321&pid=1879249#pid1879249
http://forum.kodi.tv/showthread.php?tid=173267&pid=1820808#pid1820808

verybadsoldier:
Okay, removed the advanced settings.

@t4-ravenbird 
Yes, the wake-on-lan feature can help here but to be honest I see it only as a workaround:
- the original problem is not related to WOL, so having to enable the WOL-feature feels a bit "hackish"
- the WOL-wait does not really wait until the network connection is established, according to the code it seems to just wait the configured period of time regardless of the network state
- it only kicks in when the network is actually accessed (which makes sense in the WOL context), but in my opinion Kodi's network itself should be up and working after resuming

t4-ravenbird:
WOL-wait waits until the correct network interface is available before proceeding (As opposed to just waiting for 1st interface)
Any thread accessing the network resource is held back until it is ready - any other threads are free to run.

verybadsoldier:
I fiddled around lots with this WOL-wait today and it does not really fix the problem. In some cases it did help but in most cases it did not get triggered when accessing network resources after resuming.

xbmc/powermanagement/PowerManager.cpp
```diff
@@ -327,3 +331,30 @@ void CPowerManager::SettingOptionsShutdownStatesFiller(const CSetting *setting,
...
+{
+  CNetwork& net = g_application.getNetwork();
+
+  // check if we have at least one network interface to wait for
+  if (!net.IsAvailable())
+    return;
+
+  CLog::Log(LOGNOTICE, "%s: Waiting for a network interface to come up", __FUNCTION__);
+
+  const static int numMaxTries = 50;
```
FernetMenta: those are kind of ugly. I would make them configurable as max waiting time. a value of zero disables this feature.

xbmc/settings/AdvancedSettings.cpp
Jalle19: I thought you were a bot at first since the comments were so methodical and clean. +1

verybadsoldier:
Agreed, I changed the loop and also added a setting which is configurable in the UI under "System -> Power saving".

image

MartijnKaijser:
I think a max of 120 seconds is a bit much. By that time i would have throw it out the window. 30 might be more realistic although still a lot

stefansaraev:
there is an unwanted mode (0644 -> 755) change in this PR. please chmod everything back to 0644

verybadsoldier:
Reduced maximum timeout value to 30.

Thanks for pointing out the mode changes, fixed. Sorry for that.

FernetMenta:
looks good to me
thanks!

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -20394,3 +20394,14 @@ msgstr ""
...
+
+#: system/settings/settings.xml
+msgctxt "#39011"
+msgid "Wait for network after waking up"
+msgstr ""
+
+#. Description of setting with label #39011 "Wait for network after waking up"
+#: system/settings/settings.xml
+msgctxt "#39012"
+msgid "Set the maximum time to wait for the network to come up after waking up."
```
MartijnKaijser: "Set the maximum time to wait for the network to come up after starting or waking up. When time has past before network is up startup will continue."
If it indeed also handles a startup (which it should imo)

MartijnKaijser:
See my comments about the text. Once those are done we can let it build and merge it in.

chewitt:
@verybadsoldier Does this also handle "wait on startup" ?? if not, could it do both? There's been a long-standing issue for MySQL users where Kodi starts and the network stack isn't ready. LE has a delayed start workaround function that works, but it would be cleaner if Kodi was able to handle it internally.

verybadsoldier:
Ok, I have changed the wording, thanks for the feedback.

Currently it does only handle the wait when waking up. I will take a look to reuse this feature also on startup. But if possible I would like to finish this first and then do it in a separate PR if that would be ok for. I promise to come back :)

MartijnKaijser:
👍 might want to remove "starting or" part of the strin then and we're ready to go

verybadsoldier:
Oh, good point, fixed :D Thanks!

MartijnKaijser:
jenkins build this please

MartijnKaijser:
Jenkins build and merge

MartijnKaijser:
Jenkins build this please

verybadsoldier:
(Hopefully) fixed header issue for Android/iOS. Would you please trigger Jenkins again to see if it works?
Sorry I cannot test this on my end :(

MartijnKaijser:
No problem. That why we have jenkins build this please

PR #11280, opened 2016-12-28
VideoPlayer: fix unwanted kick-in of dirty regions

FernetMenta:
fix for http://trac.kodi.tv/ticket/17100

scenario:
rendering is slightly late and frames are skipped, last buffer is rendered, swapBuffers does not block   and render threads spins through without giving vp a chance to deliver a new frame. present state is set to idle and dirty regions kick in. render thread sleeps for 40ms and story continues.

now we go to idle one second after last frame was submitted to renderer


fritsch:
jenkins build this please (thx once again for solving the unsolvable things ;-))

PR #11282, opened 2016-12-28
FIX: [amcs] Handle 3D -> 2D Mono

koying:


koying:
jenkins build this please

da-anda:
tested on the SHIELD and fixes the issue only partially. 3D movies are now always 2D, regardless of setting ;)

da-anda:
tested some more and have a better picture of the bug now. It seems as if the surface won't scale back to regular size once you turned on 2D rendering. So if you start in 3D mode, everything is fine, if you then switch to 2D mode, everything is still fine, but if you then switch back to 3D, picture will stay 2D.

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
```diff
@@ -143,6 +143,11 @@ bool CRendererMediaCodecSurface::RenderHook(int index)
         srcRect.x2 *= 2.0;
       break;
 
+      case RENDER_STEREO_MODE_MONO:
+        dstRect.y2 = dstRect.y2 * (dstRect.y2 / m_sourceRect.y2);
+        dstRect.x2 = dstRect.x2 * (dstRect.x2 / m_sourceRect.x2);
```
FrankHeimes: This is: `dstRect.x2 *= dstRect.x2 / m_sourceRect.x2`.
If `dstRect` is narrow and `m_sourceRect` is wide, then `dstRect` becomes more narrow?
All instructions in this method silently assume that `.x1 = 0` and `.y1 = 0`. Is this correct?

koying: This assumes Mono uses top image of TAB and left image of SBS, indeed.
Not sure about the "narrow" comment. I don't know of a use case where TAB/SBS is not full resolution, really.

fritsch:
The second commit is not on topic but looks correct.

MartijnKaijser:
jenkins build this please

PR #11284, opened 2016-12-28
[PVR] Krypton: Fix PVR Manager crash on startup

ksooo:
Backport of #11283 

@Jalle19 for review?

### DIFF
--- a/xbmc/pvr/PVRManager.h
+++ b/xbmc/pvr/PVRManager.h
@@ -706,6 +706,8 @@ private:
     ManagerState                    m_managerState;
     std::unique_ptr<CStopWatch>     m_parentalTimer;
 
+    CCriticalSection                m_startStopMutex; // mutex for protecting pvr manager's start/restart/stop sequence */
+
     std::atomic_bool m_isChannelPreview;
     CEventSource<PVREvent> m_events;
   };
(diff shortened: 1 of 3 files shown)


ksooo:
jenkins build this please

xbmc/pvr/PVRManager.h
```diff
@@ -706,6 +706,8 @@ namespace PVR
     ManagerState                    m_managerState;
     std::unique_ptr<CStopWatch>     m_parentalTimer;
 
+    CCriticalSection                m_initMutex; // mutex for protecting pvr manager's start/restart sequence */
```
FrankHeimes: `m_startStopMutex` would be a more appropriate name as it serializes only these two function.

FrankHeimes (changes requested):
Acquiring two locks in a function is error-prone. It can work if the order of aquisition is always the same.
Can you guarantee that neither `Start()` nor `Stop()` is ever called while `m_critSection` is held?

ksooo:
> Can you guarantee that neither Start() nor Stop() is ever called while m_critSection is held?

 As long as I'm maintaining this code and reviewing the patches, yes, because I'm very well aware of this problem.

 But, you are right, this code is not in good shape. The whole pvr manager start/stop logic is subject to complete rework, but not for Krypton.

xbmc/pvr/PVRManager.h
ksooo: changed. thx

ksooo:
jenkins build this please

ksooo:
I think this one is good to go. Well tested etc pp blah blah... (would have merged the respective master PR first, if jenkins would not act up on me currently)

PR #11285, opened 2016-12-28
[guiinfo][PVR] Fix 'player.starttime' and 'player.finishtime'

ksooo:
Fix 'player.starttime' and 'player.finishtime' info labels to properly support live tv timeshifting.

These info labels did not take live tv timeshifting into account as they always took static start and end time from the epg event currently played which is not correct when timeshifting live tv.

screenshot002

 Fix was pretty easy, the special epg tag handling for these info labels could just be removed, as player (nowadays) handles this internally and correct.

 Runtime tested on latest master on macOS and Linux.

@phil65 fyi
@Jalle19 i would like to ask you to do the code review

### DIFF
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -7941,32 +7941,18 @@ std::string CGUIInfoManager::GetMultiInfoLabel(const GUIInfo &info, int contextW
   }
   else if (info.m_info == PLAYER_FINISH_TIME)
   {
-    CDateTime time;
-    CEpgInfoTagPtr currentTag(GetEpgInfoTag());
-    if (currentTag)
-      time = currentTag->EndAsLocalTime();
-    else
-    {
-      time = CDateTime::GetCurrentDateTime();
-      int playTimeRemaining = GetPlayTimeRemaining();
-      float speed = g_application.m_pPlayer->GetPlaySpeed();
-      if (speed >= 0.75 && speed <= 1.55)
-        playTimeRemaining /= speed;
-      time += CDateTimeSpan(0, 0, 0, playTimeRemaining);
-    }
+    CDateTime time(CDateTime::GetCurrentDateTime());
+    int playTimeRemaining = GetPlayTimeRemaining();
+    float speed = g_application.m_pPlayer->GetPlaySpeed();
+    if (speed >= 0.75 && speed <= 1.55)
+      playTimeRemaining /= speed;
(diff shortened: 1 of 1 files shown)


FernetMenta:
player does not handle this correctly at the moment. please keep as is until I did the necessary changes to player

FernetMenta:
disregard my comment, after discussion with @ksooo on slack, I agree to this

ksooo:
jenkins build this please

xbmc/GUIInfoManager.cpp
```diff
@@ -7941,32 +7941,18 @@ std::string CGUIInfoManager::GetMultiInfoLabel(const GUIInfo &info, int contextW
...
-      int playTimeRemaining = GetPlayTimeRemaining();
-      float speed = g_application.m_pPlayer->GetPlaySpeed();
-      if (speed >= 0.75 && speed <= 1.55)
-        playTimeRemaining /= speed;
-      time += CDateTimeSpan(0, 0, 0, playTimeRemaining);
-    }
+    CDateTime time(CDateTime::GetCurrentDateTime());
+    int playTimeRemaining = GetPlayTimeRemaining();
+    float speed = g_application.m_pPlayer->GetPlaySpeed();
+    if (speed >= 0.75 && speed <= 1.55)
```
FrankHeimes: This test is implemented at 7 locations. Consider moving it into a new property `CApplicationPlayer::GetIsPlaying() const`. This makes it easier to guarantee consistent changes of 0.75 and 1.55 and removes necessity for temporary variables in 3 cases.

FrankHeimes (approved):
Looks good

xbmc/GUIInfoManager.cpp
ksooo: PR welcome, not really my playground.

phil65:
Would be cool if you could add a check whether something with a valid finishtime is playing at the moment so that the "finishtime" infolabel stays empty when no reasonable finish time is available. (perhaps check whether CApplication::GetTotalTime() == 0.0 ? )

ksooo:
> Would be cool if you could add a check whether something

 Yep, but this is something for another PR.

ksooo:
jenkins build this please

ksooo:
jenkins build this please

ksooo:
jenkins build this please

Jalle19:
Also forgot my review in pending state on this one... I was gonna comment on those speed numbers too just like @FrankHeimes did.

PR #11289, opened 2016-12-28
[ScriptInvocationManager] make ExecuteAsync() thread-safe

phil65:
Lock is left too early.
Was mentioned on Slack some time ago.


FrankHeimes (approved):

arnova:
Looks sane to me.

MartijnKaijser:
Jenkins build this please 
Is this something to backport as well?

fritsch (approved):
If we want to be more popely than the pope. The lock could be taken one line further down and we could use the dtor to unlock it by introducing a scope, but we are not the pope :-)

fritsch:
@MartijnKaijser yes.

PR #11295, opened 2016-12-29
context menu - fix positioning

ronie:
the context menu was positioned incorrectly when called from a dialog.

(we don't need to take the render position offset of the window into account,
as that's already included in the offset of the focused control)

before:
before

after:
after

@HitcherUK @BigNoid @phil65


BigNoid:
Yup the offset should not be considered in this case, nice find.

phil65:
makes sense. 👍

HitcherUK:
Looks good to me.

PR #11298, opened 2016-12-29
[PVR][music][video] Video Fullscreen/Music Visualisation: Fix hiding of player info

ksooo:
Ever wondered why the normal way of closing dialogs (using esc key or back button on your remote) does not close the player info in video fullscreen or music visualisation, but exits fullscreen to previous window? This is because player info internally is 'something else' than most other dialogs and closing it technically is different form closing a dialog.

This PR fixes this. Now esc key, back button on remote closes the player info if open and exits fullscreen if player info is not open. Feels 'naturally' and consistent now, imo. 

This was tested on latest master on LInux and macOS.

For those not knowing what 'player info' is - for live tv it's this:

screenshot007

It pops for instance up, if you press up/down while in fullscreen live tv window. ;-)

@Jalle19 for review, please?

### DIFF
--- a/system/keymaps/remote.xml
+++ b/system/keymaps/remote.xml
@@ -188,7 +188,7 @@
       <right>StepForward</right>
       <up>ChapterOrBigStepForward</up>
       <down>ChapterOrBigStepBack</down>
-      <back>Back</back>
+      <back>Fullscreen</back>
       <menu>OSD</menu>
       <contentsmenu>OSD</contentsmenu>
       <rootmenu>OSD</rootmenu>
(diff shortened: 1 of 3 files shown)


ronie:
i'm not a big fan of the changes to the music part of this PR.
you now would have to press esc/back twice in order to exit the fullscreen visualization window.
that doesn't feel like an improvement to me?

phil65:
I think that this change also doesnt take into account that the ShowInfo can mean two diffrent things for Video section depending on how the skinner implements it: Either open/close a modal dialog or toggle Player.ShowInfo infobool. See https://github.com/xbmc/xbmc/blob/master/xbmc/Application.cpp#L1721

ksooo:
Okay. Seems that this was not one of my best ideas. Thanks for the feedback @phil65 @ronie Closing this PR, then.

phil65:
idea was good, sadly our code doesnt allow a straightforward fix for that mess. :)

ksooo:
> idea was good, sadly our code doesnt allow a straightforward fix for that mess. :)

 Yep, that's what I ment.

PR #8925, opened 2016-01-20
Add opus tags

notspiff:
This
- improves the FFmpeg tag loader. Metadata can be attached to streams as well.
- Enables the FFmpeg tag loader for opus files.

PR #11300, opened 2016-12-29
Tag loading  for dsf, opus etc. music files extracted using ffmpeg

DaveTBlake:
Extend tag parsing done by the ffmpeg tag loader to include more ID3v2 tags in line with those loaded by Taglib for other format music files.  Most significant is album artist, without it Kodi was incorrectly interpreting albums with varied song artists as always being a compilation.

I got the key names ffmpeg uses by experimenting with a fully tagged dsf file.  The key thing to note is that ffmpeg does not return multi-value keys, just the first value for a key. Hence users will need to treat in the same way as using ID3v2.3 tag, using only single frame values and an item separator ( " / " by default) to indicate the individual artists, album artists, genres etc.

A delayed followup to #8925 and as such could be backported
@fritsch  and maybe @notspiff 

To test this you need fully tagged dsf or opus etc. music files.

### DIFF
--- a/xbmc/music/tags/MusicInfoTagLoaderFFmpeg.cpp
+++ b/xbmc/music/tags/MusicInfoTagLoaderFFmpeg.cpp
@@ -84,17 +84,71 @@ bool CMusicInfoTagLoaderFFmpeg::Load(const std::string& strFileName, CMusicInfoT
     return false;
   }
 
+  /* ffmpeg supports the return of ID3v2 metadata but has its own naming system 
+     for some, but not all, of the keys. In particular the key for the conductor 
+     tag TPE3 is called "performer". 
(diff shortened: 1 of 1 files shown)


DaveTBlake:
jenkins build this please

notspiff:
for what it's worth it looks all good to me.

notspiff:
you can use https://github.com/xbmc/FFmpeg/blob/master/libavformat/id3v2.c#L43 to sanity check a little.

DaveTBlake:
Thanks @notspiff that was the code reference I needed before but failed to find,  I just used an experimental approach instead.

The code layout, with { } for tags we don't process yet etc., is an attempt to mimic what is in the TagLib eqquivalent, just to make it obvious if `CTagLoaderTagLib` is updated then this needs updating too.

DaveTBlake:
Checked against ffmpeg source, and code comment updated

PR #11301, opened 2016-12-29
VideoPlayer + win32: fix video stuttering 

FernetMenta:
1) fix http://trac.kodi.tv/ticket/17100 and forum
http://forum.kodi.tv/showthread.php?tid=299115&pid=2488633#pid2488633

2) allow dr kicking in when video player is paused

### DIFF
--- a/xbmc/rendering/dx/RenderSystemDX.cpp
+++ b/xbmc/rendering/dx/RenderSystemDX.cpp
@@ -1189,6 +1189,17 @@ void CRenderSystemDX::PresentRenderImpl(bool rendered)
     CD3DHelper::PSClearShaderResources(m_pContext);
   }
 
+  // time for decoder that may require the context
+  {
+    CSingleLock lock(m_decoderSection);
+    XbmcThreads::EndTime timer;
+    timer.Set(5);
+    while (!m_decodingTimer.IsTimePast() && !timer.IsTimePast())
+    {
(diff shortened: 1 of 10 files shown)


FernetMenta:
Note that calling decode for dxva does not take long, it just queues the commands for the GPU. That means that we don't really block the render tread with this change.

xbmc/rendering/dx/RenderSystemDX.cpp
```diff
@@ -1189,6 +1189,17 @@ void CRenderSystemDX::PresentRenderImpl(bool rendered)
...
+    XbmcThreads::EndTime timer;
+    timer.Set(5);
+    while (!m_decodingTimer.IsTimePast() && !timer.IsTimePast())
+    {
+      m_decodingEvent.wait(lock, 1);
+    }
+  }
+
   FinishCommandList();
   m_pImdContext->Flush();
```
afedchin: you were going to drop this. don't you?

FernetMenta: drop flush? you said that it does not harm so I kept it.

afedchin: ok. keep it as is.

afedchin (approved):

FernetMenta:
jenkins build this please

PR #11305, opened 2016-12-30
[PVR] PVR windows: Fix channelgroup listener registration.

ksooo:
This fixes a regression that slipped in due to the pvr window listener logic changes I did some months ago.

Visual effects of the issue are for instance an empty channel window after changing kodi's ui language and no live update of the channel or guide window in case channels get added or removed in the backend.

Fix has been runtime tested on Linux and macOS on latest Kodi master.

@Jalle19 mind doing the review.

### DIFF
--- a/xbmc/pvr/windows/GUIWindowPVRBase.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRBase.cpp
@@ -300,9 +300,10 @@ bool CGUIWindowPVRBase::InitChannelGroup()
     if (m_channelGroup != group)
     {
       m_viewControl.SetSelectedItem(0);
-      m_channelGroup = group;
-      m_vecItems->SetPath(GetDirectoryPath());
+      SetChannelGroup(group, false);
     }
+    // Path might have changed since last init. Set it always, not just on group change.
+    m_vecItems->SetPath(GetDirectoryPath());
     return true;
   }
   return false;
@@ -314,7 +315,7 @@ CPVRChannelGroupPtr CGUIWindowPVRBase::GetChannelGroup(void)
   return m_channelGroup;
 }
 
-void CGUIWindowPVRBase::SetChannelGroup(const CPVRChannelGroupPtr &group)
+void CGUIWindowPVRBase::SetChannelGroup(const CPVRChannelGroupPtr &group, bool bUpdate /* = true */)
 {
   if (!group)
     return;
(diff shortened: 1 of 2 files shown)


xbmc/pvr/windows/GUIWindowPVRBase.h
```diff
@@ -84,7 +84,7 @@ namespace PVR
 
     bool InitChannelGroup(void);
     virtual CPVRChannelGroupPtr GetChannelGroup(void);
-    virtual void SetChannelGroup(const CPVRChannelGroupPtr &group);
+    void SetChannelGroup(const CPVRChannelGroupPtr &group, bool bUpdate = true);
```
Jalle19: Do you mind adding a docblock so the purpose of the new boolean becomes clearer?

xbmc/pvr/windows/GUIWindowPVRBase.cpp
```diff
@@ -300,9 +300,10 @@ bool CGUIWindowPVRBase::InitChannelGroup()
     if (m_channelGroup != group)
     {
       m_viewControl.SetSelectedItem(0);
-      m_channelGroup = group;
-      m_vecItems->SetPath(GetDirectoryPath());
+      SetChannelGroup(group, false);
```
Jalle19: Can't see the full context on the phone so are you sure that both this method and SetChannelGroup() don't attempt to grab the same lock?

ksooo: > that both this method and SetChannelGroup() don't attempt to grab the same lock

 they do, but whats the problem with this? performance? no. these methods are not getting called very often.

xbmc/pvr/windows/GUIWindowPVRBase.h
ksooo: Sure, can do. Sigh, I wish my predecessors would have been more professional on this. But ofc this is not excuse for not adding docs for new stuff.

ksooo: doxy added.

ksooo:
@Jalle19 good to go now?

Jalle19 (approved):

ksooo:
jenkins build this please

ksooo:
jenkins errors are not related to this PR.

PR #11306, opened 2016-12-30
[gui] Bump GUI version ABI as we start to break backwards compatibility.

MartijnKaijser:
We start to break the GUI API so let's bump it to prevent user/dev headaches of installing incompatible skins


MartijnKaijser:
@ronie @phil65

ronie:
let's merge this once our L* repo is set up.
then v18 users would at least have something to install.

MartijnKaijser:
Yep that was the plan. Only merge some time after we had the new repo set up and we have some skins thrown in.

PR #11281, opened 2016-12-28
[cmake] Add FindMDNSEmbedded module (for Zeroconf on Android)

fetzerch:
Backport of https://github.com/xbmc/xbmc/pull/11243

Jenins build: http://jenkins.kodi.tv/view/Helpers/job/BuildMulti-Krypton/39/

PR #11308, opened 2016-12-30
Revert "[cmake] Add FindMDNSEmbedded module (for Zeroconf on Android)"

MartijnKaijser:
Reverts xbmc/xbmc#11281

This seems to cause havoc on certain devices at random. Let's just revert for now and find out what the issue is.


MartijnKaijser:
jenkins build this please

PR #11310, opened 2016-12-30
[cmake] Fix make binary-addons when crosscompiling

fetzerch:
@wsnipex: This works as long as you have the toolchain file in tools/depends.
I'd prefer if it would work also with the installed one so that we can have a clean source dir.
But at least it doesn't use the wrong file anymore.


wsnipex (approved):
thanks. As you said, still not perfect, but at least it should work.

PR #11311, opened 2016-12-30
[estuary] sync

phil65:
@ronie
Pngs were batch-converted with pngcrush to get rid of some warnings during skin packaging.
See https://stackoverflow.com/questions/22745076/libpng-warning-iccp-known-incorrect-srgb-profile


ronie:
not much to review :-)

PR #11314, opened 2016-12-30
[depends] explictly disable building curl with rtmp support

MartijnKaijser:
If build environment wasn't cleaned yet curl might be build with rtmp support. Explicitly disable it.


FernetMenta:
thanks

MartijnKaijser:
No need to build this as we know it worked

PR #11319, opened 2016-12-31
Revert "Speed up startup time for movies in split rar archives."

fritsch:
This causes issues, especially on Android in combination with mediacodec: http://forum.kodi.tv/showthread.php?tid=300108

Revert for now.


fritsch:
jenkins build this please

PR #11321, opened 2016-12-31
[cmake] Fixes cpack support and WiiRemote packaging for rpi

nsenica:
## Description
This pull request provides two fixes:
- Prevents cmake from not allowing to create DEBs due to treating rbpi differently to linux.
- WiiRemote was not build and thus packaging was failing, due to not having the WiiRemote target defined. 

This will also require a backport to v17. PR: #11332

Forum topic: http://forum.kodi.tv/showthread.php?tid=302406

## Motivation and Context
See above

## How Has This Been Tested?
Built kodi with CPACK_GENERATOR=DEB option.

## Types of change

## Checklist:

### DIFF
--- a/cmake/scripts/linux/Install.cmake
+++ b/cmake/scripts/linux/Install.cmake
@@ -352,7 +352,7 @@ endforeach()
 
 # generate packages? yes please, if everything checks out
 if(CPACK_GENERATOR)
-  if(CPACK_GENERATOR STREQUAL DEB AND CORE_SYSTEM_NAME STREQUAL linux)
+  if(CPACK_GENERATOR STREQUAL DEB AND ( CORE_SYSTEM_NAME STREQUAL linux OR CORE_SYSTEM_NAME STREQUAL rbpi ) )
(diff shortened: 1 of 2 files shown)


wsnipex:
thanks, looks good. Surprised that not more is needed :)
@popcornmix this should allow us to provide our own debs for rpi on raspbian and ubuntu.

popcornmix:
So Jenkins couldbe made to produce a deb file?Sounds great.

nsenica:
Indeed :)

I think that the architecture (ARCH) on ArchSetup for rbpi will need to be changed to "arm" instead of a "arm-linux-gnueabihf" for it to work properly without any other change. I'll just need to test if everything is fully working.

There are a couple more optimisations flags that will be needed for performance reasons mainly, which I'll create a separate PR later on, but to generate the deb packages this PR is enough.

wsnipex:
jenkins build and merge

nsenica:
@wsnipex Just a minor note, to effectively build the .debs at the end of the compile phase, it's needed to call "cpack" on the build dir, so the packages are generated.

PR #11323, opened 2016-12-31
initialize anondata

peak3d:
Crashfix at least for android, probably for other platforms as well

## Description
zeroconf / mdns tries a strlen on anaondata wich address is not initialized. 

## Types of change

### DIFF
--- a/tools/depends/target/mdnsresponder/02-initialize_anondata.patch
+++ b/tools/depends/target/mdnsresponder/02-initialize_anondata.patch
@@ -1,0 +1,10 @@
+--- mDNSShared/dnssd_clientshim_orig.c	2016-12-31 13:31:09.687948146 +0100
++++ mDNSShared/dnssd_clientshim.c	2016-12-31 13:05:12.159906389 +0100
+@@ -271,6 +271,7 @@
+     x->autorename = mDNSfalse;
+     x->name = n;
+     x->host = h;
++    x->s.AnonData = 0;
+ 
+     // Do the operation
+     err = mDNS_RegisterService(&mDNSStorage, &x->s,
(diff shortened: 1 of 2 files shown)


FernetMenta:
I suggest to ask upstream if this is correct

peak3d:
Who is upstream? I have zeroconf clue where these things are coming from....

MartijnKaijser:
This is upstream  https://opensource.apple.com/tarballs/mDNSResponder/

FernetMenta:
@Memphiz you may have most experience with open source @ Apple. Can you advice?

peak3d:
Sorry, no idea a.) if our version (625.41.2) is up-to-date and b) how to pull latest sources

MartijnKaijser:
I've uploaded "latest" to our mirror now http://mirrors.kodi.tv/build-deps/sources/mDNSResponder-765.20.4.tar.gz
I checked the file and proposed patch is not in yet.
At the bottom of the page there's an email address and a mailinglist link so guess just mailing will be an option?
We also have that other quite huge patch which would be something to upstream? I read something that zeroconf was somewhat improved on Android 5 and higher but most still recommended external libraries.

FernetMenta:
Bonjour-dev should be the right mailing list for this.

Memphiz:
well no real idea either. mdnsresponder is not used by apple anymore in latest osx versions. It feels that there are a lot of custom patched sources of this lib are floating around because the original apple source doesn't compile for some other platforms by default. i wouldn't bother to upstream. tbh i wouldn't have touched that lib in the first place. Imo just fix the crash as proposed in this pr and be done?

peak3d:
Will be done somehow different....

PR #11324, opened 2016-12-31
[osx/windowing] - fix regression introduced in 4b9126649940de22e89619…

Memphiz:
…cbbe8ed5d64c5411a9 - thx to fernetmenta - fixes trac #17160

@FernetMenta was able to reproduce now - now clue why i couldn't before. Works for me with this change now. Thx for giving the hint :)


FernetMenta:
thanks

MartijnKaijser:
jenkins build this please

FernetMenta:
maybe also the reason why gfx co text is always at 60 fps currently

MartijnKaijser:
Merge?

Memphiz:
@FernetMenta - i doubt it - the regression was only there for a couple of weeks - the 60fps thingy is older iirc...

FernetMenta:
@Memphiz I am not saying that this is the root cause but it defi itely ruined your test build.

PR #11329, opened 2016-12-31
Hide notifications shown on device connect

garbear:
Follow up to #11279

Device notifications were only disabled on disconnect events, and connect events were still being shown. Fixed by using `#ifdef 0` to avoid Coverty warning:

CID 1398191: Control flow issues (DEADCODE)

PR #11290, opened 2016-12-29
Backport: Hide device event notifications

garbear:
Backport of #11279

PR #11331, opened 2016-12-31
Backport: Hide notifications shown on device connect

garbear:
Backport of #11329

Follow up to backport #11290

Device notifications were only disabled on disconnect events, and connect events were still being shown. Fixed by using `#ifdef 0` to avoid Coverty warnings.


fritsch:
In 2k17 we have a code versioning system :-) There is no need for #if 0 - as we can simply revert and no code gets lost therefore.

MartijnKaijser:
However that won't show todo in the code ;)

garbear:
and removing the code will break `git blame` without tracing down the file's history. I'm OK either way, as I plan to fix this for v18 and possibly v17.1

MartijnKaijser:
Jenkins build this please

PR #11334, opened 2017-01-01
[RetroPlayer] Fix crash when refresh rate changes

fetzerch:
## Description
RetroPlayer didn't unregister itself from the windowing system on destruction leading into a crash whenever a display lost event occurs (for example on refresh rate change).

Steps to reproduce:
- Set 'Settings->Playback->Video->Adjust Refresh Rate' to 'On Start/Stop'
- Load a ROM with a different refresh rate than the current one
- Press stop
- Try to load the ROM again

Back trace:

    Segmentation fault at xbmc/windowing/X11/WinSystemX11.cpp:628
        628           (*i)->OnLostDisplay();

Reported in the forum: http://forum.kodi.tv/showthread.php?tid=257904&pid=2488615#pid2488615
EDIT: reporter confirmed fix.

## Motivation and Context
Fix crash.

## How Has This Been Tested?
Started a few roms with different refresh rates.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/RetroPlayer/RetroPlayer.cpp
+++ b/xbmc/cores/RetroPlayer/RetroPlayer.cpp
@@ -46,6 +46,8 @@ CRetroPlayer::CRetroPlayer(IPlayerCallback& callback) :
 
 CRetroPlayer::~CRetroPlayer()
 {
+  g_Windowing.Unregister(this);
+
   CloseFile();
 }
 
--- a/xbmc/cores/RetroPlayer/RetroPlayer.cpp
+++ b/xbmc/cores/RetroPlayer/RetroPlayer.cpp
@@ -30,7 +30,6 @@
 #include "utils/log.h"
 #include "utils/MathUtils.h"
 #include "utils/StringUtils.h"
-#include "windowing/WindowingFactory.h"
 #include "FileItem.h"
 #include "URL.h"
 
@@ -41,13 +40,10 @@ CRetroPlayer::CRetroPlayer(IPlayerCallback& callback) :
   m_renderManager(m_clock, this),
   m_processInfo(CProcessInfo::CreateInstance())
 {
-  g_Windowing.Register(this);
 }
 
 CRetroPlayer::~CRetroPlayer()
 {
-  g_Windowing.Unregister(this);
-
   CloseFile();
 }
 
--- a/xbmc/cores/RetroPlayer/RetroPlayer.h
+++ b/xbmc/cores/RetroPlayer/RetroPlayer.h
@@ -36,8 +36,7 @@ namespace GAME
   class CRetroPlayerVideo;
 
   class CRetroPlayer : public IPlayer,
-                       public IRenderMsg,
-                       public IDispResource
+                       public IRenderMsg
   {
   public:
     CRetroPlayer(IPlayerCallback& callback);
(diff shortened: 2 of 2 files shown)


a1rwulf (approved):
Looks good to me.

garbear:
Line should go after CloseFile() so that resources are released in the opposite order

fetzerch:
Should that be changed in VideoPlayer as well then? https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/VideoPlayer.cpp#L677

garbear:
Yes, it looks like `g_Windowing.Unregister(this);` should go between the other two commands

FernetMenta:
vp has it at the location where it should be. you don't want to receive a signal from windowing in closefile.

garbear:
@fetzerch like this: https://github.com/garbear/xbmc/compare/af47d8331~2...af47d8331

That ensures that initialization / deinitialization happen in the opposite order.

FernetMenta:
who said that it needs to be in reverse order. it is more important to know that the code does. @garbear the way you have it in your linked patch is definitely wrong.

you don't really know what that code does, do you?

garbear:
All I know of the code is from the undocumented public interface. I've meagerly tried understanding the code, but this system is really too large to understand everything going on without clear guidance.

FernetMenta:
RetroPlayer is better off if it does not register to windowing at all.

FernetMenta:
to be more verbose. you did not implement the callback interface IDispResource and can't be notified about display lost events anyway.
display lost events are generated mainly when refresh rate gets changed. a less often scenario is moving the window from one monitor to another where both are controlled by different graphic cards.
Is that somerhing RP wants to handle?

garbear:
Thanks for the explanation. RetroPlayer renders frames as they're available to minimize latency, so it doesn't depend on refresh rate. Is handling graphics card changes something we should do in RetroPlayer, or in the core, or in a new layer between RetroPlayer and the core?

FernetMenta:
In general you won't move the app window around while playing a game, right? I don't see any case that has to be handled, at least currently.

garbear:
Ok, then this code isn't needed. It was left over from the initial version of RetroPlayer, and now I distinctly remember not finishing this because the interface was unclear.

There were a couple of other interactions with VideoPlayer that I couldn't figure out due to lack of documentation. The clock, for instance. Does VideoPlayer have any timing code we could factor out and reuse?

FernetMenta:
There's quite a few things that should be changed in RP, i.e. it uses internal classes of VP that are not exposed to the outside. That makes my life harder transforming VP because internal changes to VP now break RP.

a1rwulf:
@FernetMenta any chance that you can name some of those changes you would like to see, so that I can have a look at it?
Or at least what internal class usages you refer to, so that we can make your life easier?

FernetMenta:
@garbear @a1rwulf in other words: cut all dependencies to VP. just copy the code you need from VP's subcomponents. this is much better than this unwanted dependencies.

fetzerch:
Since further improvements can be done as separate PRs,
any objections to merge this fix in the meantime? I've removed the registration to windowing.

a1rwulf:
@fetzerch ok for me, but let's wait for @garbear.
@FernetMenta thanks for pointing that out. You mean something similar to what I did for the libretro-gl stuff? https://github.com/a1rwulf/xbmc/commit/7daafc2388623ed6781f5346f7e45cc041b225f1

FernetMenta:
@a1rwulf not really. the outside world must not access any subfolder in VideoPlayer. Those are private components owned by VP. neither the outside world is allowed to create any files in folders owned by VP. this is a very basic principle of software design. you do access private API's.

garbear:
OK if this code goes in. I sent @FernetMenta a PR to begin the process of separating RetroPlayer and VP

garbear (approved):

fetzerch:
jenkins build this please

PR #11335, opened 2017-01-01
add option to wait for network on startup

verybadsoldier:
Makes Kodi wait for a network interface to become available on startup.

## Description
This is a successor to https://github.com/xbmc/xbmc/pull/11277 which made Kodi wait for network after waking up. As requested, this PR also makes Kodi wait right on startup for a network interface.

Please review especially:
* Is the invocation of `CNetwork::WaitForNet` in `CApplication::Initialize` at the right place? It is important to do before the database initialization I think, but should it be after `g_peripherals.Initialise()`?
* The setting to configure this is `powermanagement.waitfornetwork`. Maybe it should better be now something like `network.waitfornetwork` as it is not specifc to waking up now anymore?

## Motivation and Context
When using MySQL it often is a problem that the network interface is not available yet when Kodi starts and then fails to initialize the database because of the missing connection to MySQL.

## How Has This Been Tested?
I started Kodi several times and observed the log outputs.

## Types of change

## Checklist:


t4-ravenbird:
Again, this is redundant functionality that is already present in the wake-on-access module - https://github.com/xbmc/xbmc/blob/master/xbmc/network/WakeOnAccess.cpp#L242
(Same situation as https://github.com/xbmc/xbmc/pull/11277 )
The solutions are not identical of course, but overlapping, and it is strange to reinvent the wheel instead of adapting the existing.

verybadsoldier:
In my opinion there really is not so much similarity between the two. The only overlapping functionality is that both things are waiting for some kind of network connection. But even this differs as one is waiting for an IP being able to reach another IP taking the subnet into account while the other one is just waiting for an interface being connected.

FernetMenta:
I agree that this functionality here is different to wake-on-lan. There may be some duplicate or similar loc but this can be cleaned up at a later time if it really makes sense.

jenkins build this please

PR #11336, opened 2017-01-01
[PVR] PVR windows: Fix subfolder item selection on window update.

ksooo:
Followup to #11270 , where I forgot to handle subfolders (e.g. children of time rules, recordings subfolders) correctly.

@Jalle19 for review?


Jalle19 (commented):
Can't say I really understand the fix but +1 if it works

ksooo:
jenkins build this please

PR #11339, opened 2017-01-01
[IMXEGL] reconfigure fb1 & unblank (backport of #11338)

mk01:
(this was forgotten from the previous commit)

## Description
reconfigure fb1 & unblank before (hdmi)sound is resumed in WinSystem

## Motivation and Context
this should prevent resuming audio while hdmi system is still blanked (this can happen depending on order of windowing.register() calls after commit 21671109b53e714e5eea1fa1c29da39f6e8bf518)
also it makes m_delayDispReset working while keeping the proper order conf fb0 -> conf fb1 -> unblank -> resume audio 

## Types of change

## Checklist:

### DIFF
--- a/xbmc/windowing/egl/EGLNativeTypeIMX.cpp
+++ b/xbmc/windowing/egl/EGLNativeTypeIMX.cpp
@@ -263,6 +263,8 @@ bool CEGLNativeTypeIMX::SetNativeResolution(const RESOLUTION_INFO &res)
   CreateNativeDisplay();
   CreateNativeWindow();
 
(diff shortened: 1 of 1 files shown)


fritsch:
jenkins build and merge

FernetMenta:
jenkins does not build for IMX, why do you trigger him? does not comply with green IT :)

fritsch:
Hehehe. I trigger it to make sure (as in the past) no introduced ifdef - here not the case - introduces issues for other platforms. So, just to make sure for the record, that build breakage was not introduced by this PR - for others not having to look at the code when searching issues.

PR #11337, opened 2017-01-01
reADD: [amcs] Handle Rotation

koying:
Regression from Jarvis

PR #11342, opened 2017-01-02
reADD: [amcs] Handle Rotation

MartijnKaijser:
Backport from https://github.com/xbmc/xbmc/pull/11337


koying:
jenkins build this please

koying (approved):
There doesn't seem to be any modification from K to Master, so should be fine

PR #11343, opened 2017-01-02
[cmake] Fixes rpi ARCH

nsenica:
## Description
If the ARCH is set to arm-linux-gnueabihf kodi will be looking for libX-arm-linux-gnueabihf.so's but they don't exist, since the architecture libs are compiled with is arm.

## Motivation and Context
See above

## How Has This Been Tested?
Built kodi.
This probably will require a bit more testing just to be sure that it doesn't break anything.

## Types of change

## Checklist:


popcornmix:
I've been building with ``-DWITH_ARCH=arm`` to fix loading of binary addons.
Assuming this is an equivalent fix, then +1 from me.

nsenica:
@popcornmix That's exactly what this fixes :)

wsnipex:

wsnipex:
@MilhouseVH could you include this in your builds please?

lrusak:
@wsnipex It won't matter. This change is only for compiling on an arm host it seems. we set `-DWITH_ARCH=arm` because the arm cpu will never be detected when building on x86_64.

see, https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/mediacenter/kodi/package.mk#L188

wsnipex:
this will trigger whenever you build for rbpi (CORE_SYSTEM_NAME=rbpi) and CPU=arm1176jzf-s or cortext-a7 or cortex-a53

PR #11352, opened 2017-01-03
[cmake] Move NEON flags to platform specific ArchSetup so its easily …

nsenica:
## Description
NEON/CPU compiler flags were being set in common ArchSetup.cmake, which made it impossible to customize for each platform, specially because as this comes in after each platform ArchSetup is evaluated, the setting in common would override the same settings set by each platform.
This fix keeps functionality and will allow platform customisations possible.

For Rpi it also uses the same compiler flags that were used in https://github.com/xbmc/xbmc/blob/master/tools/depends/configure.ac#L394-L411

@popcornmix For RBPi2 platform, would setting "-march=armv7ve -ffast-math -O3" increase performance ?
 
Backport is needed, I'll provide it if this PR is accepted.

## Motivation and Context
Let platforms specify their needs in terms of compiler flags and not set them globally for all.

## How Has This Been Tested?
Building Kodi

## Types of change

## Checklist:


stefansaraev:
> @popcornmix For RBPi2 platform, would setting "-march=armv7ve -ffast-math -O3" increase performance ?

none of -ffast-math -O3 should be enabled

popcornmix:
armv7ve includes virtualisation extensions? Not sure if that would provide anything useful, but feel free to build it both ways and see if you can detect any performance differences.

As @stefansaraev says, -ffast-math and -O3 frequently get suggested and tried, but there are a lot of libraries that kodi uses and in general something subtle breaks when you try. The performance improvement (probably only a couple of %) doesn't really compensate for the decrease in stability.

Again feel free to test in your local builds to see if you can measure any greater gains but kodi will be very cautious about accepting them as defaults.

nsenica:
Your comments do make sense, I don't think the performance improvement is that significant (if at all noticeable) to include anything else. I'll just leave the ones that were in the configure.ac, if anyone wants to try different configuration, they can be set directly from command line with cmake. 

Thanks for your input.

wsnipex:
argh, disregard please. Shouldn't review with a headache

wsnipex:
I'm not sure I understand your use case. Afaict all this PR does is hardcoding neon flags whenever the cpu is arm. Before it was possible to simply set -DENABLE_NEON=OFF and put your desired compiler flags in CMAKE_CXX_FLAGS.

nsenica:
I understand your point, it does make sense to some extent.

There are two sides of this commit:
1 - If we do as you say, then we'll have to remove [this line](https://github.com/xbmc/xbmc/blob/master/cmake/scripts/common/ArchSetup.cmake#L149) as well. It sets these flags and overrides any common flag passed on CMAKE_CXX_FLAGS from command line. So either we remove all references to adding specific compiler flags and we pass them all from command line as we ned or we allow specific settings that we understand to be beneficial do be set by default. So we can already provide a build with some optimisations from source.
2 - The other side is to mimic the behaviour we have in the [previous system](https://github.com/xbmc/xbmc/blob/master/tools/depends/configure.ac#L394-L411) (Makefile, autobuild, ?? not sure about the correct name now), where specifying --target_platform=raspberry-pi or --target_platform=raspberry-pi2 would set these exact flags depending on the target platform.
Which for a matter of consistency across build systems, they should also be removed if it is decided to go as you suggest.

I'd personally prefer to the approach I suggested in this PR, however, I'm fine going with your suggestion as well, as long as we keep it consistent across platforms and architectures.

I hope this clarifies it a bit. :)

wsnipex:
ad 1.) No. The flags are only set if ENABLE_NEON=ON, which is an option that you can override.
Its on by default for arm cpus, so the average user doesn't have to know exactly what he is doing.
Advanced users, packagers and devs can set it to OFF and set whatever cflags they wish instead.
Your PR makes this impossible, as it hardcodes the flags.

ad 2.) I agree that we should bring re implement  --target_platform=raspberry-pi/2/3 in cmake, maybe also for some other very well defined embedded systems.

nsenica:
Regarding 1), you're right, I didn't realise my changes were breaking it, let me think about it and I'll come up with a solution. Maybe follow your original suggestion and move this block inside each platform that deals with NEON, I'll have a look.. thanks :)

Regarding 2) we could eventually split cortex-a7 from cortex-a53 and add their specific compiler flags, the amount of work is not that much and we could do it. However I'm not sure if the improvement is noticeable at all and that would require packagers to create three different versions one for each RPi :)

nsenica:
I've pushed a new attempt to fix 1).

I very much appreciate feedback and the idea is to make it work properly, so if you have any feedback I'm very happy to hear it and incorporate it in the final version.

cmake/scripts/rbpi/ArchSetup.cmake
```diff
@@ -13,9 +13,12 @@ else()
   if(CPU STREQUAL arm1176jzf-s)
     set(ARCH arm)
     set(NEON False)
+    set(NEON_FLAGS "-mcpu=arm1176jzf-s -mtune=arm1176jzf-s -mfloat-abi=hard -mfpu=vfp")
   elseif(CPU MATCHES "cortex-a7" OR CPU MATCHES "cortex-a53")
     set(ARCH arm)
     set(NEON True)
+    set(NEON_FLAGS "-fPIC -mcpu=cortex-a7 -mfloat-abi=hard -mfpu=neon-vfpv4 -mvectorize-with-neon-quad")
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -lpthread")
```
wsnipex: why is this needed? and why not for rpi1?

wsnipex (commented):
thanks now it looks good to me, besides the question above

cmake/scripts/rbpi/ArchSetup.cmake
nsenica: As explained before, these flags copy the existing ones from https://github.com/xbmc/xbmc/blob/master/tools/depends/configure.ac#L394-L411 for both platforms. 
If you look a bit above (on line 16) you'll see that there are different flags, more specific to RPi 0/1 CPU.

I'm not sure if this answers your question, but if not please let me know.

wsnipex: then those linker flags are a relic workaround for missing checks somewhere in our depends.
We build perfectly fine without that, so please remove that line.

nsenica: Done :) 

Thanks for taking the time to review this PR.

wsnipex:
jenkins build this please

PR #11353, opened 2017-01-03
android:debuggable support

peak3d:
Set android:debuggable=[true/false] regarding the cmake build type

## Description
There are some operations on android for that the ap has to be debugable.
For example run-as commands in adb shell on devices without root access.

This PR sets the application::android:debuggable value regarding the cmake build type

## Motivation and Context
We need run-as org.xbmc.kodi kill -6 [pid] to generate tombstone.

### DIFF
--- a/cmake/scripts/android/Install.cmake
+++ b/cmake/scripts/android/Install.cmake
@@ -13,6 +13,12 @@ if(NOT ZIPALIGN_EXECUTABLE)
   message(FATAL_ERROR "Could NOT find zipalign executable")
 endif()
 
+if(CMAKE_BUILD_TYPE STREQUAL Debug)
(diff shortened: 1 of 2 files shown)


cmake/scripts/android/Install.cmake
```diff
@@ -13,6 +13,12 @@ if(NOT ZIPALIGN_EXECUTABLE)
   message(FATAL_ERROR "Could NOT find zipalign executable")
 endif()
 
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
```
fetzerch: no need for quotes around Debug

fetzerch (approved):
lgtm

PR #11356, opened 2017-01-03
PAPlayer: Explicitely cast to 64 bit before doing large multiplications

fritsch:
Fixes playback of 8 hour hearbooks.


Jalle19 (commented):
Wouldn't it be better to change the type of the members instead of casting?

fritsch:
Did you check where the members are coming from? :-) File.h - I don't really feel like changing these, cause that will introduce issues.

FernetMenta:
jenkins build and merge

PR #11357, opened 2017-01-03
fix invalid output from JsonSchemaBuilder

Jalle19:
## Description

The scanner scans up until an `\n` character, which means a potential `\r` might sneak in. When each character is looped over there was no handling for `\r` so they got streamed to the output, resulting in bogus new lines in the output file.

## Motivation and Context

Without this I can't build Kodi, the build will fail with thousands of lines of `missing termination character "`

## How Has This Been Tested?

It builds

## Checklist:

ping @wsnipex since I believe you wrote this?

### DIFF
--- a/tools/depends/native/JsonSchemaBuilder/src/JsonSchemaBuilder.cpp
+++ b/tools/depends/native/JsonSchemaBuilder/src/JsonSchemaBuilder.cpp
@@ -64,6 +64,11 @@ void print_json(ifstream &in, ofstream &out)
     closing = false;
     for (string::iterator itr = line.begin(); itr != line.end(); itr++)
     {
+      // Skip \r characters
+      if (*itr == '\r') {
+        break;
+      }
+
       // Count opening { but ignore the first one
       if (*itr == '{')
       {


Jalle19:
jenkins build this please

Jalle19:
@Montellese maybe you know something about this too?

wsnipex:
iirc the idea was to set your git client to auto correct "bogus" windows EOLs

Jalle19:
@wsnipex I don't think requiring specific git preferences in order to build the project is a good idea. With this fix it doesn't matter.

wsnipex:
I'm not against this change, but I'm against ppl checking in code with windows line endings.

Jalle19:
@wsnipex me too. Okay to merge then?

Montellese:
@Jalle19: on what platform does it not work for you? Works fine for me on win32. But I don't have anything against this fix.

Concerning CRLF handling, there are lots of projects which enforce auto crlf handling through git options so IMO that's not unreasonable.

Jalle19:
@Montellese Linux

Montellese (approved):
Sure, go ahead.

PR #11358, opened 2017-01-03
librarynodes - fix invalid xml

ronie:
this used to be part of https://github.com/xbmc/xbmc/pull/11017 but was dropped for the wrong reasons.

the '&' character needs to be escaped in xml documents.

@DaveTBlake

### DIFF
--- a/system/library/music/musicroles/Arrangers.xml
+++ b/system/library/music/musicroles/Arrangers.xml
@@ -2,5 +2,5 @@
 <node order="5" type="folder" visible="Library.HasContent(Role, Arranger)">
 	<label>29988</label>
 	<icon>DefaultMusicGenres.png</icon>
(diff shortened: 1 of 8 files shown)


DaveTBlake:
Thanks. Not sure why my testing failed that time, but this does work and meets the xml standard (although if users create nodes without escaping the & that works too).

ronie:
the reason it works in kodi is we've written our own function to handle these errors:
https://github.com/xbmc/xbmc/blob/master/xbmc/utils/XBMCTinyXML.cpp#L239
(imo not a good idea as it masks actual errors in the code)

DaveTBlake:
Thanks for the explanation, and I agree masking errors is not good. I will endevour to escape & in any folder type custom node examples I give to users. May even get to update (possibly create) the wiki one day.

ronie:
ping @Ignoble61 
just a heads-up, in case this affects your addon.

PR #11360, opened 2017-01-03
[StringUtils] Fix SecondsToTimeString() for negative values 

phil65:
Atm Player.SeekTime can show really weird values when seeking "behind" 0 or totalplaytime.
This commit makes sure that 0 <= Player.Seektime <= totalplaytime.
@FernetMenta

### DIFF
--- a/xbmc/utils/StringUtils.cpp
+++ b/xbmc/utils/StringUtils.cpp
@@ -933,6 +933,8 @@ long StringUtils::TimeStringToSeconds(const std::string &timeString)
 
 std::string StringUtils::SecondsToTimeString(long lSeconds, TIME_FORMAT format)
 {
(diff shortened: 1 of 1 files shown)


FernetMenta:
the assuption that minimum displayed time is zero is not correct. the issue is caused by seekhandler and has to be fixed there.

phil65:
Agreed, should be fixed at the root.
So I think we basically need two fixes:
- SeekHandler method GetSeekSize() should only return values within the appropriate range
- StringUtils::SecondsToTimeString() needs to deal correctly with negative values.

Changed this PR to fix the second part, SeekHandler is not my territory :)

@FernetMenta

FernetMenta:
looks ok to me

phil65:
jenkins build this please.

PR #11367, opened 2017-01-04
vtb: fix inverted logic for fence

FernetMenta:
the old code did hold a buffer if there was no fence. that lead to video stuttering


FernetMenta:
jenkins build this please

Memphiz:
thx - i have added this to my test build and let the users double check :)

PR #11370, opened 2017-01-04
[cmake] fixes cpack dependency list generation

nsenica:
## Description
While trying to generate deb packages from cpack I realised that the dependencies where not properly generated for binary packages. I also realised that the current implementation goes a bit about the [cpack specification](https://cmake.org/cmake/help/v3.6/module/CPackDeb.html#variable:CPACK_DEBIAN_%3CCOMPONENT%3E_PACKAGE_DEPENDS) (Check the Note), which states that if SHLIBDEPS is ON the found dependencies will be _appended_ to the DEPENDS of each package. Which was not what was happening with the [kodi's implemention](https://github.com/xbmc/xbmc/blob/c4da9e834abc7e4a208e818e7390d8d1283f6ee7/cmake/cpack/CPackConfigDEB.cmake#L267-L273) - it was overriding as the information within the txts states, which needs to be changed as well.
This PR enables PACKAGE_SHLIBDEPS in the needed packages and allows it to be appended to the PACKAGE_DEPEND list.

Documentation in each .txt template was updated to reflect the above info.

Backport will be provided when the final version of this PR is agreed upon.

## Motivation and Context
Create .deb with the correct dependencies.

## How Has This Been Tested?
Building .deb packages using cpack.

## Types of change

## Checklist:


hudokkow:
Looks good, thx.

Are you sure it's picking **all** the deps? I have a vague memory about problems with this approach but can't test now.

I couldn't find the note in the link you posted above. The correct link is https://cmake.org/cmake/help/v3.6/module/CPackDeb.html#variable:CPACK_DEBIAN_%3CCOMPONENT%3E_PACKAGE_DEPENDS

btw, we're missing a `kodi-game-dev.txt.in` since https://github.com/xbmc/xbmc/commit/60cc35deb65558ce3c793c702ae9d481b0f46173. Care to add it?

nsenica:
@hudokkow Thanks for your comments.

I've updated the original post with the correct link, thanks for pointing that out.

About the dependencies, the resulting dependencies after generating the .debs match the required ones that @wsnipex defined in [debian/control](https://launchpad.net/~team-xbmc/+archive/ubuntu/unstable/+files/kodi_17.0~git20161230.1859-rc2-0trusty.debian.tar.xz) file for trusty. The only things I had to leave out, as I didn't find a way to add it, were the "misc:Depends" and "python:Depends" but I believe that those do not add any extra dependency.
The testing I did, provided me with enough confidence that we are not missing out any (at least) important dependency, but with this change, if we figure out that it's missing we can always add it as it now give us with that flexibility.

I'll add kodi-game-dev.txt.in, sure. :)

Once this is done, I can build and generate deb files so you can have a look at them, or I can post here the result of `dpkg -I *.deb` of the generated packages, as you prefer.

nsenica:
Added kodi-game-dev.txt.in

nsenica:
@wsnipex those are untouched and are defined [here](https://github.com/nsenica/xbmc/blob/f8e45442769648e2faf2d7a1389d8442468cf803/cmake/cpack/deb/packages/kodi.txt.in#L18) and they're exacly as it is in your debian/control file, if I'm not mistaken, except for librtmp and libmad which are not required anymore.

Once again, if we realise that they're missing, we can always add them later.

wsnipex:
ok, thanks :)

wsnipex:
@hudokkow good to go?

hudokkow:
Thx.
Please backport without the kodi-game-dev.txt.in commit.

PR #11374, opened 2017-01-05
VideoPlayer: dxva - only retain buffers for deinterlacing

FernetMenta:
fix http://trac.kodi.tv/ticket/17100

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoRenderers/WinRenderer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/WinRenderer.cpp
@@ -262,6 +262,7 @@ void CWinRenderer::AddVideoPictureHW(DVDVideoPicture &picture, int index)
     SAFE_RELEASE(buf->pic);
     buf->pic = m_processor->Convert(picture);
     buf->frameIdx = m_frameIdx;
+    buf->pictureFlags = picture.iFlags;
     m_frameIdx += 2;
   }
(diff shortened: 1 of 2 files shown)


FernetMenta:
@afedchin back-refs are only required for deinterlacing, right?

afedchin:
right. but I can't understand how it may cause skips/drops?

FernetMenta:
because it runs out of free buffers.

FernetMenta:
to be more verbose: the log shows 5 render buffers. with current render and two taken by backrefs, there are only 2 left. Most of the time there is only one queued and renderManager can't skip a late frame. if latency can''t be corrected by skipping a frame, VP will ask decoder for a drop. This is more noticeable than a skip.

FernetMenta:
@afedchin ok to merge?

afedchin:
it's ok to merge. but I still dont' understand how it worked before v17?

FernetMenta:
biggest difference to earlier versions is that we don't block the render thread in RM anymore. 

jenkins build this please

afedchin:
my point of view is we just moved the blocking to another place i.e. to the render system. that's why I didn't understand why this causes the issue in v17 and doesn't do the same in v16

FernetMenta:
we did not just move the blocking. blocking in render system requires more elasticity because the render thread can block there for longer than frametime or it may not block at all, if internal buffers are not full. This behaviour is less noticeable on Intel systems.

PR #11378, opened 2017-01-05
[cmake] don't use hardcoded 'kodi' paths

ronie:
currently building binary addons fails / get installed in the wrong directory if you rebrand kodi.

@fetzerch @wsnipex

### DIFF
--- a/cmake/addons/CMakeLists.txt
+++ b/cmake/addons/CMakeLists.txt
@@ -138,7 +138,7 @@ if(ADDON_SRC_PREFIX)
 endif()
 
 if(NOT APP_LIB_DIR)
-  set(APP_LIB_DIR "${ADDON_DEPENDS_PATH}/lib/kodi")
+  set(APP_LIB_DIR "${ADDON_DEPENDS_PATH}/lib/${APP_NAME_LC}")
 else()
   file(TO_CMAKE_PATH "${APP_LIB_DIR}" APP_LIB_DIR)
 endif()
--- a/cmake/scripts/common/CheckTargetPlatform.cmake
+++ b/cmake/scripts/common/CheckTargetPlatform.cmake
@@ -47,8 +47,8 @@ function(check_install_permissions install_dir have_perms)
   # param[in] install_dir directory to check for write permissions
   # param[out] have_perms wether we have permissions to install to install_dir
 
-  set(testfile_lib ${install_dir}/lib/kodi/.cmake-inst-test)
-  set(testfile_share ${install_dir}/share/kodi/.cmake-inst-test)
+  set(testfile_lib ${install_dir}/lib/${APP_NAME_LC}/.cmake-inst-test)
+  set(testfile_share ${install_dir}/share/${APP_NAME_LC}/.cmake-inst-test)
   get_filename_component(testdir_lib ${testfile_lib} DIRECTORY)
   get_filename_component(testdir_share ${testfile_share} DIRECTORY)
 


hudokkow (approved):

hudokkow:
Thx.
Needs backport.

Kwiboo:
Not sure how much master and Krypton have diverged in this matter but I must use https://github.com/Kwiboo/plex-home-theater/commit/136ffa7d4242d761007a785f889fdc97833632e7 to be able to build binary addons with a different app name with Krypton and LibreELEC.
All kodi binary addons I have come across look for a `Kodi` package and not a `<app-name>` package, thus would fail to find the `KodiConfig.cmake` and end up using wrong include/install paths.

wsnipex:
right, those are valid concerns. All current addons look for KodiConfig.cmake, many still even wrongly use
\#include "kodi/something.h"

So we have the problem that we either fix kodi to be completely APP_NAME agnostic and break all binary addons or we keep hardcoding some of our install paths. Both is ugly.
We could argue that whoever rebrands kodi should also rebrand addons, but thats a long stretch....

Kwiboo:
Sorry for hijacking this with comments not directly related to this PR.

As someone who is experimenting with Kodi Krypton app name re-branding for a rewrite of my openpht fork I think it is important that paths is different to allow side-by-side installations and this PR seems correct.
But the current renaming of `KodiConfig.cmake` and the variables it contains makes re-branding harder then it needs to be. I will send a PR with my findings at a later date once I have solved my other path and app-name related issues for non-LibreELEC platforms.

AlwinEsch:
Hard to bring a good answer from my side :), I understand completely the reason to change the include to a variable name. But for me was the way of `#include` with a **`"`** always wrong. The add-on headers are for a add-on external parts where normally a **`<..>`** becomes used.

On coming rework has it maybe over 50 include files and to have them without a folder name on begin can confuse someone or system (addon interface `File.h` or system `file.h`), the other way with a add of kodi_*.h to every file looks then a bit to crazy for me (is then also like a directory).

The rework from me has as main goals:
- Use add-on instances (e.g. PVR) multiple times together
- Use different instances together (e.g. PVR together with Inputstream)
- Make the functionality of Python also for Binary add-ons
and the hardest:
- Remove old Python way from Kodi itself and do Python over the "C++" add-on interface (This is the main goal to add also other languages like JavaScript or Java itself as languages.
- Bring in add-on's as own EXE files where Kodi is then like a OS.

You know how much functions and parts are on Python. Everything whats there comes also to binaries!

I thought to find maybe a special development kit name who can hold independent from Kodi, so that add-on must always use a `#include <how-is-your-kit-name/...h>`.

AlwinEsch:
Here is a part to see about the coming includes for Kodi: https://github.com/AlwinEsch/kodi/tree/add-addon-instance-handling/xbmc/addons/kodi-addon-dev-kit/include/kodi
Them there are from first round and becomes currently revised from me and some other new.

ronie:
is there anything left in this PR we'd like to see merged?
otherwise it can be closed as obsolete/superseded by #11452 

ping @Kwiboo

Kwiboo:
This will probably only work if the addon is looking for `<app-name>` and not `Kodi` or `<app-name>Config.cmake` is renamed to `KodiConfig.cmake` when building an addon out of tree, otherwise `${APP_NAME_LC}` wont be known (the chicken-egg problem)

ronie:
ok, thx!

PR #11379, opened 2017-01-05
[application] load includes before fonts

BigNoid:
## Description
Move up the loading of includes before the fonts are loaded so Fonts.xml can have includes too.

## Motivation and Context
See http://forum.kodi.tv/showthread.php?tid=303121
While this is probably an edge case, the code change looks trivial enough to implement it.

## How Has This Been Tested?
Runtime tested, everything seemed normal.

## Types of change

## Checklist:

@ronie

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -1685,6 +1685,8 @@ bool CApplication::LoadSkin(const std::string& skinID)
 
   g_colorManager.Load(m_ServiceManager->GetSettings().GetString(CSettings::SETTING_LOOKANDFEEL_SKINCOLORS));
 
+  g_SkinInfo->LoadIncludes();
+  
   g_fontManager.LoadFonts(m_ServiceManager->GetSettings().GetString(CSettings::SETTING_LOOKANDFEEL_FONT));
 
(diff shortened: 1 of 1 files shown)


gade01:
This could save a lot of code.

Would it be possible to merge it?

phil65:
Not sure if there might cause regressions because now you load the includes before loading the skin strings? (sth like include condittion="String.Equals(Skin.String(xx),$LOCALIZE[31123])")

Razzeee:
@BigNoid can you check the case phil mentions?

BigNoid:
It doesn't work indeed, but it doesn't work on master either
```<include file="View_57_Shift.xml" condition="String.IsEqual(Skin.String(Nox.Test),$LOCALIZE[31037])"/>```
for example.

But it might be better to move the font loading down the include loading to be safe?

phil65:
Hmm or perhaps just ignore my comment, I think the case I mentioned is really an ugly way to solve things anyways. I guess the pros outweight the cons here, so from my side, feel free to merge.

Razzeee:
jenkins build this please

PR #11034, opened 2016-12-02
Savestate manager

garbear:
This adds a savestate manager. The underlying database is from the picture library back I wrote back in 2012 (#1562)

## Description
This includes three commits:

* Denormalized database ORM - an ORM I wrote specifically for our dual SQLite/MySQL database system. Provides a denormalized database (via JSON serialization, and optionally BSON) with dynamic normalization whenever fields are added/removed
* Savestate database - built on the denormalized database implementation
* Savestate manager - pictured below

## TODO
Before this is merged to master, the custom ORM needs to be replaced by flat XML files for savestates.

## How Has This Been Tested?
Included in my RP test builds since May 2016

## Screenshots (if appropriate):
savestate manager

## Types of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #11380, opened 2017-01-05
[games] Implement in-game saves

fetzerch:
Implement in-game (save ram) saves.

## Description
This implements in-game saves where the frontend accesses the game's memory and saves/loads it to/from disk. This is mostly used in emulators for SRAM (battery backed up ram on cartridges) or memory cards.

For reference, these are the other ways how saving is implemented in the Game API:
- Savestates (https://github.com/xbmc/xbmc/pull/11034)
- The game saves it directly (already implemented and used for example by ScummVM)

Differences to save states:
- Works only for supported games (e.g. emulated games with SRAM support)
- Often works emulator independent (and can be used to start a game with one emulator and continue with another)
- Visible in-game (e.g. in-game save game selection menus)
- Closer to how saving in the good old days worked where one could only save at certain points where the game allowed it.

Previously PRed to the retroplayer banch in https://github.com/garbear/xbmc/pull/73
*API-Change*: Merging has to be synced with https://github.com/kodi-game/game.libretro/pull/10.

## Motivation and Context
Save games.

## How Has This Been Tested?
Best tested with 2048 or SNES/GBA emulators. 

## Types of change

## Checklist:

### DIFF
--- a/xbmc/games/addons/GameClient.cpp
+++ b/xbmc/games/addons/GameClient.cpp
@@ -20,6 +20,7 @@
 
 #include "GameClient.h"
 #include "GameClientCallbacks.h"
+#include "GameClientInGameSaves.h"
 #include "GameClientInput.h"
 #include "GameClientKeyboard.h"
 #include "GameClientMouse.h"
@@ -276,6 +277,9 @@ bool CGameClient::OpenFile(const CFileItem& file, IGameAudioCallback* audio, IGa
   if (!InitializeGameplay(file.GetPath(), audio, video))
     return false;
 
+  m_inGameSaves.reset(new CGameClientInGameSaves(this, &m_struct));
+  m_inGameSaves->Load();
+
   return true;
 }
 
@@ -486,6 +490,9 @@ void CGameClient::CloseFile()
 
   if (m_bIsPlaying)
   {
+    m_inGameSaves->Save();
+    m_inGameSaves.reset();
+
     try { LogError(m_struct.UnloadGame(), "UnloadGame()"); }
     catch (...) { LogException("UnloadGame()"); }
   }
--- a/xbmc/games/addons/GameClient.h
+++ b/xbmc/games/addons/GameClient.h
@@ -39,6 +39,7 @@ class CFileItem;
 namespace GAME
 {
 
+class CGameClientInGameSaves;
 class CGameClientInput;
 class CGameClientKeyboard;
 class CGameClientMouse;
@@ -160,6 +161,9 @@ private:
   PERIPHERALS::EventRateHandle m_inputRateHandle; // Handle while keeping the input sampling rate at the frame rate
   std::unique_ptr<IGameClientPlayback> m_playback; // Interface to control playback
   GAME_REGION           m_region;              // Region of the loaded game
+
+  // In-game saves
+  std::unique_ptr<CGameClientInGameSaves> m_inGameSaves;
 
   // Input
   std::map<int, std::unique_ptr<CGameClientInput>> m_ports;
--- a/xbmc/games/addons/GameClientInGameSaves.cpp
+++ b/xbmc/games/addons/GameClientInGameSaves.cpp
@@ -1,0 +1,162 @@
+/*
+ *      Copyright (C) 2016-2017 Team Kodi
+ *      http://kodi.tv
+ *
+ *  This Program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2, or (at your option)
+ *  any later version.
+ *
+ *  This Program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this Program; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "GameClientInGameSaves.h"
+
+#include "GameClient.h"
+#include "GameClientTranslator.h"
+#include "filesystem/File.h"
+#include "filesystem/Directory.h"
+#include "profiles/ProfilesManager.h"
+#include "utils/URIUtils.h"
+#include "utils/log.h"
+
+#include <assert.h>
+
+using namespace GAME;
+
+#define INGAME_SAVES_DIRECTORY          "InGameSaves"
+#define INGAME_SAVES_EXTENSION_SAVE_RAM ".sav"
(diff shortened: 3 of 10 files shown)


xbmc/games/addons/GameClient.h
```diff
@@ -161,6 +162,9 @@ class CGameClient : public ADDON::CAddonDll
   std::unique_ptr<IGameClientPlayback> m_playback; // Interface to control playback
   GAME_REGION           m_region;              // Region of the loaded game
 
+  // In-game saves
+  std::unique_ptr<CGameClientInGameSaves> m_inGameSaves;
```
notspiff: why a pointer? it's unconditionally allocated.

xbmc/games/addons/GameClientInGameSaves.cpp
```diff
@@ -0,0 +1,165 @@
...
+  m_gameClient(addon),
+  m_dllStruct(dllStruct)
+{
+  assert(m_gameClient != nullptr);
+  assert(m_dllStruct != nullptr);
+}
+
+void CGameClientInGameSaves::Load()
+{
+    Load(GAME_MEMORY_SAVE_RAM);
```
notspiff: indent

xbmc/games/addons/GameClientInGameSaves.cpp
```diff
@@ -0,0 +1,165 @@
...
+    Load(GAME_MEMORY_RTC);
+}
+
+void CGameClientInGameSaves::Save()
+{
+    Save(GAME_MEMORY_SAVE_RAM);
```
notspiff: indent

xbmc/games/addons/GameClientInGameSaves.cpp
```diff
@@ -0,0 +1,165 @@
...
+  }
+  catch (...)
+  {
+    CLog::Log(LOGERROR, "GAME: %s: Exception caught in GetMemory()", m_gameClient->ID().c_str());
+  }
+
+  if (size > 0)
+  {
+    const std::string path = GetPath(memoryType);
+
```
notspiff: why the spacefest?

xbmc/games/addons/GameClientInGameSaves.h
```diff
@@ -0,0 +1,73 @@
...
+struct GameClient;
+
+namespace GAME
+{
+  class CGameClient;
+
+  /*!
+   * \brief This class implements in-game saves.
+   *
+   * Some games do not implement state persistence on their own, but rely on the frontend for saving their current
```
notspiff: i'd do \details here. not exactly a brief.

fetzerch: brief ends when starting a new paragraph (see https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmddetails and http://www.stack.nl/~dimitri/doxygen/manual/docblocks.html#cppblock). but why not be more specific

xbmc/games/addons/GameClient.h
fetzerch: I tried that at first, but CGameClientInGameSaves needs a valid pointer to the dll struct which is initialized only in CGameClient::Initialize so I have to delay the initialization.

xbmc/games/addons/GameClientInGameSaves.h
akva2: yeah i know but i like being specific as you say (every time i encountered it i had to look up in the manual - has started to stick now though ;))

xbmc/games/addons/GameClientInGameSaves.cpp
fetzerch: sorry that one i don't get. any more indentation errors, or did i mess up with the size somehow?

akva2: 3 newlines seems excessive.

fetzerch: That is very true ;) sry i looked at GetPath() in Load()

akva2:
https://imgflip.com/i/1h66wu

akva2:
(and it looks all good to me now).

fetzerch:
Thanks!

Does it make sense to save periodically (maybe every minute)? We don't control cores and they might crash before we're able to stop a game. Guess that could easily be implemented with a CTimer in CGameClient and could even be reused later for savestates. If it makes sense this can however come in a separate PR.

akva2:
maybe, but then we'd definitely need to ping-pong on the sram file, or you may end up dumping corrupted data.

garbear:
I would like to abstract the various software and hardware details of managing one's saved data.

I think the best interface would be a "Saved game" concept associated with a single game/file. The saved game would include the player's current progress, including awards/achievements (scraped from a ROM achievements service) and a history of saved data. The saved data would include all forms of the game's storage at a point in time, and we could eventually allow more fine-grained memory management like loading/saving/importing/exporting SRAM saves (per-file) or savestates (per-emulator).

xbmc/games/addons/GameClientInGameSaves.cpp
```diff
@@ -0,0 +1,162 @@
...
+#include "utils/URIUtils.h"
+#include "utils/log.h"
+
+#include <assert.h>
+
+using namespace GAME;
+
+#define INGAME_SAVES_DIRECTORY          "InGameSaves"
+#define INGAME_SAVES_EXTENSION_SAVE_RAM ".sav"
+#define INGAME_SAVES_EXTENSION_RTC      ".rtc"
```
thebrid: Would you consider replacing these with typed constants?

garbear:
@fetzerch is this ready to go in?

xbmc/games/addons/GameClient.cpp
```diff
@@ -276,6 +277,9 @@ bool CGameClient::OpenFile(const CFileItem& file, IGameAudioCallback* audio, IGa
   if (!InitializeGameplay(file.GetPath(), audio, video))
     return false;
 
+  m_inGameSaves.reset(new CGameClientInGameSaves(this, &m_struct));
```
garbear: Move this to the constructor. I was just bit by a class that got the order of closing the resource wrong, so it's best to enforce the condition that it's always allocated.

fetzerch: See the comment above from from spiff, I would have done it if it was possible, but m_struct is only valid after CGameClient::Initialize.

fetzerch:
From my side it's ready - but I thought you don't want it in its current from (without the abstraction).
Imho it would be nice to have save games working in nightly-builds and we can refactor it to a more fancy solution afterwards, but then at least the code doing the actual saving/loading is there.
Another advantage would be that we don't have to sync the game.libretro API change any more. It's always something you have to remember when rebasing the retroplayer branches.

Therefore: Your button, (needs a merge of kodi-game/game.libretro#10 as well).

garbear (approved):

garbear:
jenkins build this please

PR #11381, opened 2017-01-05
[android] Enable DisplayMode API on all devices with Nougat

MartijnKaijser:
It works :)


CiNcH83:
Display mode API has been added with Marshmallow though.

MartijnKaijser:
Have you actually read the code?

koying:
+1. 
Maybe just remove the whole block for Leia, as we can assume all Shield's will be N by then

MartijnKaijser:
That was certainly my intention in several months :)

CiNcH83:
> Have you actually read the code?

OK, got it. Rule only applies to SHIELD...

PR #11384, opened 2017-01-06
Quell warnings

notspiff:
For non-hostage code.

### DIFF
--- a/xbmc/guilib/GUIFontTTF.cpp
+++ b/xbmc/guilib/GUIFontTTF.cpp
@@ -732,7 +732,7 @@ bool CGUIFontTTFBase::CacheCharacter(wchar_t letter, uint32_t style, Character *
       m_posX += -bitGlyph->left;
 
     // check we have enough room for the character
-    if ((m_posX + bitGlyph->left + bitmap.width) > static_cast<int>(m_textureWidth))
+    if ((m_posX + bitGlyph->left + bitmap.width) > m_textureWidth)
     { // no space - gotta drop to the next line (which means creating a new texture and copying it across)
       m_posX = 0;
       m_posY += GetTextureLineHeight();
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -8517,7 +8517,7 @@ std::string CGUIInfoManager::GetMusicLabel(int item)
     {
       std::string strBitrate = "";
       if (m_audioInfo.bitrate > 0)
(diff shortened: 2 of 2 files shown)


Paxxi:
Nice!

xbmc/guilib/GUIFontTTF.cpp
```diff
@@ -732,7 +732,7 @@ bool CGUIFontTTFBase::CacheCharacter(wchar_t letter, uint32_t style, Character *
       m_posX += -bitGlyph->left;
 
     // check we have enough room for the character
-    if ((m_posX + bitGlyph->left + bitmap.width) > static_cast<int>(m_textureWidth))
+    if ((m_posX + bitGlyph->left + bitmap.width) > m_textureWidth)
```
Rechi: this adds a warning (not removes one), because the left side is signed and the right side is unsigned

akva2: hmm. then there are platform / version  / compiler differences. this definitely quells the warning on my machine, the lhs is unsigned here.
with cast:
```
/home/akva/kode/xbmc-cmake/xbmc/guilib/GUIFontTTF.cpp: In member function ‘bool CGUIFontTTFBase::CacheCharacter(wchar_t, uint32_t, CGUIFontTTFBase::Character*)’:
/home/akva/kode/xbmc-cmake/xbmc/guilib/GUIFontTTF.cpp:735:50: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     if ((m_posX + bitGlyph->left + bitmap.width) > static_cast<int>(m_textureWidth))
```
without cast it's silent. fun! ubuntu 16.04 / gcc 5.4 / libfreetype6-dev=2.6.1-0.1ubuntu2

akva2: and the result being unsigned is the correct interpretation of c(++) rules. if you add signed and unsigned numbers, the result is unsigned. and bitmap.width is unsigned. is it signed in your freetype (ftimage.h, struct FT_Bitmap_) ?

Rechi: I'm building with depends where freetype 2.4.7 is used. `width` is signed in this version.

akva2: right. i guess we need to explicitly cast both sides then to have it work everywhere.

Rechi: I just checked and it is unsigned since version 2.5.4. Bumping to a version equal or greater maybe another option.

akva2: opened https://github.com/xbmc/xbmc/pull/11422 up to others to decide what to do.

PR #11385, opened 2017-01-06
Kryptonfail

fritsch:
Krypton only.

This stops MediaCodec from crashing when an instance is already running. It won't help to open a second decoder properly. For LiveTV that means: HW -> switch -> SW -> switch -> HW. Better than HW -> switch -> crash -> kodi exit.

The second one was asked for by peak3d - that thingy is racy as hell anyways. Removing another direct call into the App can only make it better.

For master and v18 the proper interface change is already available. Krypton on Android is in very bad shape, but we don't need to release it crashing.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.cpp
@@ -675,8 +675,6 @@ void CDVDVideoCodecAndroidMediaCodec::Dispose()
       xbmc_jnienv()->ExceptionClear();
   }
   ReleaseSurfaceTexture();
-  if (m_render_surface)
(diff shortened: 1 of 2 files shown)


peak3d:
After looking / testing we have decided against dirty hacks to get mediacodec streamchange working on K17. To get it halfways running we would need changes in core.

K18's API changes in VideoPlayer allow an implementation wich works.
Lets push K18 / kodi-agile to get it live.

fritsch:
jenkins build this please

fritsch:
Reports from http://trac.kodi.tv/ticket/16578#comment:40 tell us that the additional commit: https://github.com/xbmc/xbmc/pull/11385/commits/79ea41eb0f77508759dbc9e7e9d4e9406eb36296 makes livetv with hw acceleration working.

On my list: CMake / Define changes to reduce the AC3 Encoder quality for android to 384 which is currently done by @fetzerch

PR #11387, opened 2017-01-06
Replace raw new/delete in Util with stack object

thebrid:
## Description
Replace a single heap object allocated using new/delete with a stack-allocated object.

## Motivation and Context
Raw new/delete may lead to inadvertent memory leaks if we return early / throw an exception. I considered using a smart pointer but the object we're constructing seemed relatively small so it seemed fine to just construct it on the stack.

## How Has This Been Tested?
This concerns loading of SMI subtitles. I compiled & ran my changes on x86-64 Ubuntu and loaded a sample SMI subtitle file. KODI ran fine & displayed the expected subtitles.

The changes should be localised in this part of the code.

I considered adding unit tests for this change but this function has external dependencies on the filesystem which seem non-trivial to isolate.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/Util.cpp
+++ b/xbmc/Util.cpp
@@ -2056,11 +2056,11 @@ void CUtil::ScanForExternalSubtitles(const std::string& strMovie, std::vector<st
     if (URIUtils::HasExtension(vecSubtitles[i], ".smi"))
     {
       //Cache multi-language sami subtitle
-      CDVDSubtitleStream* pStream = new CDVDSubtitleStream();
-      if (pStream->Open(vecSubtitles[i]))
+      CDVDSubtitleStream stream;
+      if (stream.Open(vecSubtitles[i]))
       {
         CDVDSubtitleTagSami TagConv;
-        TagConv.LoadHead(pStream);
+        TagConv.LoadHead(&stream);
         if (TagConv.m_Langclass.size() >= 2)
         {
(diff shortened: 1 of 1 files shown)


FernetMenta:
@ace20022 ok with you?

ace20022:
+1 in general.

>if we return early / throw an exception

I have not checked it, but where can this happen?

thebrid:
@ace20022 , I didn't have a specific case in mind, more general exception safety, but we could emit an exception here:

[CUtil::ScanForExternalSubtitles](https://github.com/xbmc/xbmc/blob/master/xbmc/Util.cpp#L2060)
[CDVDSubtitleStream::Open](https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDSubtitles/DVDSubtitleStream.cpp#L45)
[CFileItem::CFileItem](https://github.com/xbmc/xbmc/blob/master/xbmc/FileItem.cpp#L296)
std::string::operator=

Paxxi:
Even if we don't care about exceptions we should not use raw pointers anymore except for non-owning use or possibly special case. This is cleanup that we could benefit from throughout our code base. 
Good initiative @thebrid

ace20022:
jenkins build this please

ksooo:
@thebrid nice. thanks for your contribution.

thebrid:
Thanks :smile:

PR #11393, opened 2017-01-07
[Android] disable zeroconf

peak3d:
Zeroconf for android runs on 1 kernel with 100% CPU, the code has bugs (unitialized values) and someone told that it is not working. Remove.


fritsch:
jenkins build this please

fetzerch:
why not just disable it instead of removing all the buildsys bits?

peak3d:
how?

PR #11395, opened 2017-01-07
[backport] [cmake] Platform specific overrides

fetzerch:
Backport of https://github.com/xbmc/xbmc/pull/10918

### DIFF
--- a/project/cmake/scripts/common/Macros.cmake
+++ b/project/cmake/scripts/common/Macros.cmake
@@ -51,6 +51,53 @@ function(source_group_by_folder target)
   endif()
 endfunction()
 
+# Marks header file as being overridden on a certain list of platforms.
+#
+# Explicitly marking a file as overridden on specific platforms avoids issues with globbing where
+# CMake would have to be called manually when overriding for a new platform.
+#
+# Usage: add_platform_override(${PROJECT_NAME} settings.h PLATFORMS android linux osx)
+function(add_platform_override target filename)
+  cmake_parse_arguments(ARG "" "" "PLATFORMS" ${ARGN})
+  if(NOT ARG_PLATFORMS)
+    message(FATAL_ERROR "Missing parameter PLATFORMS")
+  endif()
+
+  # Generate an _override.h header that is either empty (platform doesn't define overrides)
+  # or includes the corresponding platform override header.
+  # This _override.h has to be included by the generic header.
+
+  # Determine filename of override header.
+  string(REPLACE ".h" "_override.h" override_file ${filename})
+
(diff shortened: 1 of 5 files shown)


fritsch:
jenkins build this please please

FernetMenta:
@fritsch see my comments on the PR to master

fritsch:
@FernetMenta seen it, though wondering as the proposed idea is the same currently already used for the Factory stuff.

FernetMenta:
@fritsch right, it does not fit there either and it does not fit for other things, see my comment on the Mir PR. "linux" does not sufficiently define a platform in the context of the requirements we have here. X11, Mir, Wayland define the platform

fetzerch:
I don't see the relation to this PR. This one is if you have a header with defines specific for a platform.

The other problem is that you want to select a certain cpp depending on a variable. And that we have in several places with different variables. X11, Mir, Wayland is one part of the equation. Just 2 other random examples: https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/CMakeLists.txt and https://github.com/xbmc/xbmc/blob/master/xbmc/cores/AudioEngine/CMakeLists.txt. Not sure if this problem is really generic enough to handle that in a cmake macro. Also I think we're missing a generic concept of options that are exclusive (such as wayland, x11) or where you can have multiple options enabled (such as VDPAU/VAAPI, or ALSA/PA).

The impact for this PR is that we need to think what to do if we have several vars instead of just the platform.

FernetMenta:
Your links make the problem obvious too. X11, vaapi, vdpau are no options but currently they are treated as options by macros like FOUND_X11. Those components do define a platform/port.

fritsch:
Do we want to split? Krypton - which is "gone" anyways and future master?

FernetMenta:
I don't think so, otherwise we cold have gone with the ifdef from the beginning

fetzerch:
The "platform" variable in cmake can't be changed without rewriting huge parts. They define for example how installation or packaging is done. We can have (a) new variable(s) for this. But we have to sort out all the 'options' before and that's going to take a while.

FernetMenta:
ok, then go ahead with this for k

fritsch:
jenkins build this please

fritsch:
Thanks very much.

stefansaraev:
the backport PR is exactly 2 days old. not 2 months. if you disagree to the ifdef, this can be reverted.

FernetMenta:
sure, it has to be reverted. I just commented on the ifdef.

PR #11396, opened 2017-01-07
Replace raw new/delete inside CDVDFactorySubtitle 

thebrid:
## Description
Replaced use of raw new/delete inside CDVDFactorySubtitle. Previous code is creating a CDVDSubtitleStream in CDVDFactorySubtitle then passing ownership to the relevant CDVDSubtitleParser for each subtitle format. I've replaced the raw new/delete with unique_ptr and moved ownership into the parsers. This means we can simplify the factory function a lot.

I've also deleted some commented out code that looked like it wasn't serving a purpose any more.

As this change is already getting a bit big, I deliberately avoided changing the return values of the function over to unique_ptrs as this would expand the size of the PR.

What is here I've separated out into a few commits to make it easier to read.

## Motivation and Context
The main goal is improving general exception safety. Hopefully the factory function should be a bit easier to read now.

## How Has This Been Tested?
Before starting, I added tests for CDVDFactorySubtitle to ensure I didn't break anything. I ran these both before & after making the change.

I also tested manually by loading subtitles for all of the affected subtitle formats. I ran this on x86-64 Ubuntu.

The changes should be isolated to the classes touched here.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDSubtitles/DVDFactorySubtitle.cpp
+++ b/xbmc/cores/VideoPlayer/DVDSubtitles/DVDFactorySubtitle.cpp
@@ -34,42 +34,41 @@ CDVDSubtitleParser* CDVDFactorySubtitle::CreateParser(std::string& strFile)
 {
   char line[1024];
   int i;
-  CDVDSubtitleParser* pParser = NULL;
 
   std::unique_ptr<CDVDSubtitleStream> pStream(new CDVDSubtitleStream());
   if(!pStream->Open(strFile))
   {
-    return NULL;
+    return nullptr;
   }
 
-  for (int t = 0; !pParser && t < 256; t++)
+  for (int t = 0; t < 256; t++)
   {
     if (pStream->ReadLine(line, sizeof(line)))
     {
       if ((sscanf (line, "{%d}{}", &i)==1) ||
           (sscanf (line, "{%d}{%d}", &i, &i)==2))
       {
-        pParser = new CDVDSubtitleParserMicroDVD(std::move(pStream), strFile.c_str());
+        return new CDVDSubtitleParserMicroDVD(std::move(pStream), strFile.c_str());
       }
       else if (sscanf(line, "[%d][%d]", &i, &i) == 2)
       {
-        pParser = new CDVDSubtitleParserMPL2(std::move(pStream), strFile.c_str());
+        return new CDVDSubtitleParserMPL2(std::move(pStream), strFile.c_str());
       }
       else if (sscanf(line, "%d:%d:%d%*c%d --> %d:%d:%d%*c%d", &i, &i, &i, &i, &i, &i, &i, &i) == 8)
       {
-        pParser = new CDVDSubtitleParserSubrip(std::move(pStream), strFile.c_str());
+        return new CDVDSubtitleParserSubrip(std::move(pStream), strFile.c_str());
       }
       else if (sscanf(line, "%d:%d:%d:", &i, &i, &i) == 3)
       {
-        pParser = new CDVDSubtitleParserVplayer(std::move(pStream), strFile.c_str());
+        return new CDVDSubtitleParserVplayer(std::move(pStream), strFile.c_str());
       }
       else if ((!memcmp(line, "Dialogue: Marked", 16)) || (!memcmp(line, "Dialogue: ", 10)))
       {
-        pParser =  new CDVDSubtitleParserSSA(std::move(pStream), strFile.c_str());
+        return new CDVDSubtitleParserSSA(std::move(pStream), strFile.c_str());
       }
       else if (strstr (line, "<SAMI>"))
       {
-        pParser = new CDVDSubtitleParserSami(std::move(pStream), strFile.c_str());
+        return new CDVDSubtitleParserSami(std::move(pStream), strFile.c_str());
       }
     }
     else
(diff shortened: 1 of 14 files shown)


cmake/treedata/common/tests.txt
```diff
@@ -8,3 +8,5 @@ xbmc/threads/test                 test/threads
 xbmc/utils/test                   test/utils
 xbmc/video/test                   test/video
 xbmc/cores/AudioEngine/Sinks/test test/audioengine_sinks
+xbmc/cores/VideoPlayer/DVDSubtitles/test test/dvdsubtitles
```
FernetMenta: no tests in core

xbmc/cores/VideoPlayer/DVDSubtitles/DVDFactorySubtitle.cpp
```diff
@@ -1,6 +1,6 @@
 /*
- *      Copyright (C) 2005-2013 Team XBMC
- *      http://xbmc.org
+ *      Copyright (C) 2005-2017 Team Kodi
+ *      http://kodi.tv
```
FernetMenta: please leave headers

xbmc/cores/VideoPlayer/DVDSubtitles/test/TestDVDFactorySubtitle.cpp
```diff
@@ -0,0 +1,80 @@
+/*
```
FernetMenta: see above, no tests here

thebrid:
@FernetMenta - done!

xbmc/cores/VideoPlayer/DVDSubtitles/DVDFactorySubtitle.cpp
```diff
@@ -21,132 +21,62 @@
...
+  for (int t = 0; t < 256; t++)
   {
     if (pStream->ReadLine(line, sizeof(line)))
     {
       if ((sscanf (line, "{%d}{}", &i)==1) ||
           (sscanf (line, "{%d}{%d}", &i, &i)==2))
       {
-        pParser = new CDVDSubtitleParserMicroDVD(pStream, strFile.c_str());
-        pStream = NULL;
+        return new CDVDSubtitleParserMicroDVD(std::move(pStream), strFile.c_str());
```
FernetMenta: how about pStream.release()?

thebrid: There are a lot of benefits in keeping the pointer owned by a unique pointer. If we call release, the pointer is effectively unowned so any exceptions which occur before we assign it to another smart pointer could cause us to leak resources.

We are doing a bunch of operations here which may throw - constructing CDVDSubtitleParserMicroDVD and its ancestor classes, copy constructing a string etc. so there's the possibility it could go wrong now or in the future.

FernetMenta: move does transfer the ownership as well. what do you think is the difference to?

return new CDVDSubtitleParserMicroDVD(pStream.release(), strFile.c_str());

Paxxi: Release is called before new ownership is taken so errors in constructor will leak memory. Low risk but best to not get in the habit of doing it that way

xbmc/cores/VideoPlayer/DVDSubtitles/DVDSubtitleParser.h
```diff
@@ -64,16 +65,14 @@ class CDVDSubtitleParserText
...
 public:
-  CDVDSubtitleParserText(CDVDSubtitleStream* stream, const std::string& filename)
+  CDVDSubtitleParserText(std::unique_ptr<CDVDSubtitleStream> && stream, const std::string& filename)
     : CDVDSubtitleParserCollection(filename)
+		, m_pStream(std::move(stream)) 
   {
-    m_pStream  = stream;
   }
 
   virtual ~CDVDSubtitleParserText()
```
Paxxi: Can be replaced with `=default ` if it's now empty

xbmc/cores/VideoPlayer/DVDSubtitles/DVDFactorySubtitle.cpp
thebrid: Here's the process if we use std::move and move construct another unique_ptr:

1. Managed by source unique_ptr
2. Managed by destination unique_ptr

If we call release(), we insert another step into the process:
1. Managed by source unique_ptr
2. Unmanaged raw pointer
3. Managed by destination unique_ptr

If any exception is thrown between 1 & 3, we could leak resources. Move construction gives us a guarantee that we won't.

xbmc/cores/VideoPlayer/DVDSubtitles/DVDSubtitleParser.h
thebrid: Good shout

thebrid:
@Paxxi , updated now

xbmc/cores/VideoPlayer/DVDSubtitles/DVDFactorySubtitle.cpp
FernetMenta: thanks a lot!

FernetMenta:
jenkins build this please

thebrid:
Jenkins failed on WIN-32 build:

CMake Error at C:/Program Files (x86)/CMake/share/cmake-3.6/Modules/FindPackageHandleStandardArgs.cmake:148 (message):
  Could NOT find Java (missing: Java_JAVA_EXECUTABLE Runtime)

PR #11397, opened 2017-01-08
[PVR] PVRMananger::m_currentFile: CFileItem* -> CFileItemPtr.

ksooo:
Cleanup: Replace raw pointer with smart pointer.

@Jalle19 for review?

### DIFF
--- a/xbmc/pvr/PVRManager.cpp
+++ b/xbmc/pvr/PVRManager.cpp
@@ -157,7 +157,6 @@ bool CPVRManagerJobQueue::WaitForJobs(unsigned int milliSeconds)
 CPVRManager::CPVRManager(void) :
     CThread("PVRManager"),
     m_addons(new CPVRClients),
-    m_currentFile(NULL),
     m_bFirstStart(true),
     m_bIsSwitchingChannels(false),
     m_bEpgsCreated(false),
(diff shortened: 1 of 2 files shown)


xbmc/pvr/PVRManager.cpp
```diff
@@ -1579,9 +1571,8 @@ bool CPVRManager::PerformChannelSwitch(const CPVRChannelPtr &channel, bool bPrev
 
   CLog::Log(LOGDEBUG, "PVRManager - %s - switching to channel '%s'", __FUNCTION__, channel->ChannelName().c_str());
 
-  // will be deleted by CPVRChannelSwitchJob::DoWork()
```
Jalle19: Did you change this usage too? Can't really see from the diff.

Jalle19 (approved):
Apart from the comment, +1

xbmc/pvr/PVRManager.cpp
ksooo: Hehe, good catch, <code>CPVRChannelSwitchJob::DoWork</code> actually did not (!) delete it, this was a bug, which now is fixed.

ksooo: CPVRChannelSwitchJob now also uses shared pointers.

ksooo:
jenkins build this please

ksooo:
WIN-32 Jenkins error is not related to this PR.

PR #11399, opened 2017-01-08
Krypton: Keep Autotools working

fritsch:
This keeps autotools alive.

### DIFF
--- a/xbmc/cores/AudioEngine/AEDefines.h
+++ b/xbmc/cores/AudioEngine/AEDefines.h
@@ -22,5 +22,7 @@
 #define AE_AC3_ENCODE_BITRATE 640000
 #define AE_DTS_ENCODE_BITRATE 1411200
 
+#ifdef BUILDING_WITH_CMAKE
 // Enable platform specific overrides
 #include "AEDefines_override.h"
+#endif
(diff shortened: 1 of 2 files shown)


fritsch:
jenkins build this please (just want to see if it still does what it should)

xbmc/cores/AudioEngine/AEDefines.h
```diff
@@ -22,5 +22,7 @@
 #define AE_AC3_ENCODE_BITRATE 640000
 #define AE_DTS_ENCODE_BITRATE 1411200
 
+#ifdef BUILDING_WITH_CMAKE
```
FernetMenta: what is this, still an ugly ifdef hack

fritsch: It's a cmake define, like the one for SSE the one for X11, used e.g. here: https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp#L49 the one for MIR, the one for whatever ... can you explain me the difference between your problem with the ifdef in this place and the ifdef that you pushed some weeks ago into VAAPI?

I think I have a problem calibrating towards your wishes.

fritsch:
Oki - next try on the way. Calibration ongoing ...

xbmc/cores/AudioEngine/AEDefines.h
FernetMenta: put it in words: why should the code know if its built with cmake or whatever system else?

stefansaraev: because it is a header, generated only by the cmake buildsystem ?

PR #11416, opened 2017-01-10
VC-1 initial support in BitstreamParser

peak3d:
Implement VC1-Packet Bitstream parser

## Motivation and Context
VC-1 in mkv container sometimes have only DTS but no PTS values.
AML decoder needs from time to time a secure PTS value wich will be set on I-Frames.

This PR does currently NOT return every I-Frame, but it returns true on recovery points.

## How Has This Been Tested?
Separate test environment

## Types of change

### DIFF
--- a/xbmc/utils/BitstreamConverter.cpp
+++ b/xbmc/utils/BitstreamConverter.cpp
@@ -72,6 +72,34 @@ enum {
     HEVC_NAL_SEI_SUFFIX = 40
 };
 
+enum {
+  VC1_PROFILE_SIMPLE,
+  VC1_PROFILE_MAIN,
+  VC1_PROFILE_RESERVED,
+  VC1_PROFILE_ADVANCED
+};
+
+enum {
+  VC1_END_OF_SEQ = 0x0A,
+  VC1_SLICE = 0x0B,
+  VC1_FIELD = 0x0C,
+  VC1_FRAME = 0x0D,
+  VC1_ENTRYPOINT = 0x0E,
+  VC1_SEQUENCE = 0x0F,
+  VC1_SLICE_USER = 0x1B,
+  VC1_FIELD_USER = 0x1C,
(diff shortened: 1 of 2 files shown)


xbmc/utils/BitstreamConverter.cpp
```diff
@@ -72,6 +72,34 @@ enum {
     HEVC_NAL_SEI_SUFFIX = 40
 };
 
+enum {
```
FernetMenta: wrong coding style

FernetMenta:
BitstreamParser was originally borrowed code to parse h264. It was already ugly hacking hevc into it, but vc-1 should really go into its own class and file.

xbmc/utils/BitstreamConverter.cpp
```diff
@@ -303,6 +400,90 @@ bool CBitstreamParser::FindIdrSlice(const uint8_t *buf, int buf_size)
   return rtn;
 }
 
+bool CBitstreamParser::VC1RecoveryPoint(const uint8_t *buf, const uint8_t *buf_end)
+{
+  bool rtn = false;
```
thebrid: This variable is never written to. Do we need it?

peak3d:
Sry, no time to clean up all the mess.

xbmc/utils/BitstreamConverter.cpp
peak3d: thx

xbmc/utils/BitstreamConverter.cpp
peak3d: ?????? The original file is completely in this style

FernetMenta: right, but the fact that the old code does not comply to our coding guideline does not justify that new code does not have to do so either. what do you think?

PR #11418, opened 2017-01-10
Add the missing bits to fully add details to a video ListItem

phate89:
## Description
While I was looking into python scraper I noticed some missing info in our api and several fields that we load with a string that we then split when in python we could provide directly a list.
I'm not 100% sure if it's enough to set a function private to hide it from swig

## Motivation and Context
Add missing python labels info

## How Has This Been Tested?
I tested it with a fake addon

## Types of change

## Checklist:

### DIFF
--- a/xbmc/interfaces/legacy/ListItem.cpp
+++ b/xbmc/interfaces/legacy/ListItem.cpp
@@ -139,14 +139,14 @@ namespace XBMCAddon
       if (!item) return;
       {
         LOCKGUIIF(m_offscreen);
-        for (Properties::const_iterator it = dictionary.begin(); it != dictionary.end(); ++it)
+        for (const auto& it: dictionary)
         {
-          std::string artName = it->first;
+          std::string artName = it.first;
           StringUtils::ToLower(artName);
           if (artName == "icon")
-            item->SetIconImage(it->second);
+            item->SetIconImage(it.second);
           else
-            item->SetArt(artName, it->second);
+            item->SetArt(artName, it.second);
         }
       }
     }
@@ -157,8 +157,8 @@ namespace XBMCAddon
 
       LOCKGUI;
       CVideoInfoTag& vtag = *item->GetVideoInfoTag();
-      for (Properties::const_iterator it = dictionary.begin(); it != dictionary.end(); ++it)
-        vtag.SetUniqueID(it->second, it->first);
+      for (const auto& it : dictionary)
+        vtag.SetUniqueID(it.second, it.first);
     }
 
     void ListItem::setRating(std::string type, float rating, int votes /* = 0 */, bool defaultt /* = false */)
@@ -217,13 +217,13 @@ namespace XBMCAddon
       else if (lowerKey == "totaltime")
       {
         CBookmark resumePoint(item->GetVideoInfoTag()->GetResumePoint());
-        resumePoint.totalTimeInSeconds = (float)atof(value.c_str());
+        resumePoint.totalTimeInSeconds = static_cast<float>(atof(value.c_str()));
         item->GetVideoInfoTag()->SetResumePoint(resumePoint);
       }
       else if (lowerKey == "resumetime")
       {
         CBookmark resumePoint(item->GetVideoInfoTag()->GetResumePoint());
-        resumePoint.timeInSeconds = (float)atof(value.c_str());
+        resumePoint.timeInSeconds = static_cast<float>(atof(value.c_str()));
         item->GetVideoInfoTag()->SetResumePoint(resumePoint);
       }
       else if (lowerKey == "specialsort")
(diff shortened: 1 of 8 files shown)


phil65:
There are some more missing I think:
We need ability to set these VideoInfoTag members via setInfo(), too:

m_iSpecialSortEpisode
m_iSpecialSortSeason
m_showLink
m_iTrack

phate89:
I know. I started adapting scrapers from movie part so that's the ones that popped first..But it's useful to have a list.. thanks
Also we still miss the xml fanart that I'm struggling to find a good way to add it..

phil65:
Another one to have a complete list here: m_strEpisodeGuide (I have no idea what that is used for though :) )
Those are the easy ones, m_fanart and m_namedSeasons are the only ones left then which are a bit more tricky.

xbmc/interfaces/legacy/ListItem.h
```diff
@@ -654,6 +656,8 @@ namespace XBMCAddon
       /// Added labels **setid**, **set**, **imdbnumber**, **code**, **dbid**, **path** and **userrating**.
       /// Expanded the possible infoLabels for the option **mediatype**.
       /// @python_v18 Added new **game** type and associated infolabels.
+      /// Added labels **setoverview**, **tag**.
```
phil65: @AlwinEsch does every line here need a separate python_v18 tag so that they get indexed correctly?

phil65:
> I'm not 100% sure if it's enough to set a function private to hide it from swig

Since that method is static, perhaps move it to StringUtils?

phate89:
Ok i added sortepisode, sortseason, episodeguide and showlink... tracknumber already existed..
Now we just miss m_strPictureURL, m_fanart and m_namedSeasons..
Namedseasons I could add a setinfo field where we could pass `[( "Murder House", 1),( "Asylum", 2)]` but seems counterintuitive to have the season number after the season name.
I checked and swig didn't create the binding for the private function so should be ok

phate89:
I also added an "AddSeason" method to fill the namedseasons and a commit to modernize ListItem code.. The missing functions will have to wait a bit..

phil65:
nice work, thx!

phil65:
this here https://github.com/phil65/xbmc/commit/3c8378edfa271a69bdc558c93ec388c3fae12741 should add support for setting m_fanart (untested yet)

xbmc/interfaces/legacy/ListItem.h
```diff
@@ -1032,6 +1033,9 @@ namespace XBMCAddon
       /// Function completely removed and replaced with context menu add-ons.
       ///
 #endif
+
+private:
+  std::vector<std::string> getStringArray(const InfoLabelValue& alt, const std::string& tag, std::string value = "");
```
akva2: wrong indent.

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -169,6 +169,14 @@ namespace XBMCAddon
       item->GetVideoInfoTag()->SetRating(rating, votes, type, defaultt);
     }
 
+    void ListItem::AddSeason(int number, std::string name /* = "" */)
+    {
+      if (!item) return;
+
+      LOCKGUI;
```
akva2: LOCKGUIIF(m_offscreen) like the other stuff or this will deadlock scrapers.

phate89: @akva2 Fixed it. thanks. So also SetRatings() and all the setters needs this? because I saw some other that didn't have it

phate89:
Ok now we have full coverage for video stuff. Not sure if "addAvailableThumbs" is a good name though..

xbmc/interfaces/legacy/ListItem.cpp
akva2: yeah, they certainly will. the locks are there because these callbacks were meant to modify items on screen (in a container), and in that case we can grab the gui lock. i do not recall all the gritty details, but the scrapers are called with another lock holding the thread at some point, resulting in deadlocks. i think it was when you run a 'manual' scan (i.e. update from info dialog, or 'i' on a unscanned item).

xbmc/utils/ScraperUrl.cpp
```diff
@@ -339,6 +339,15 @@ bool CScraperUrl::ParseEpisodeGuide(std::string strUrls)
   return true;
 }
 
+bool CScraperUrl::AddElement(std::string url, std::string aspect /* = "" */)
+{
+  TiXmlElement thumb("thumb");
```
phil65: I think it´s better to directly create an SUrlEntry  object here and append it to m_url instead of goin through xml.

xbmc/utils/Fanart.cpp
```diff
@@ -56,6 +55,21 @@ void CFanart::Pack()
   m_xml << fanart;
 }
 
+void CFanart::AddFanart(std::string image, std::string preview, std::string colors)
```
thebrid: Consider passing these std::string arguments by const reference to prevent unnecessary copies.

phate89: thanks fixed it!

phate89:
@phil65 @tamland 
Can you do a check to see if there's some silly mistake in these changes?

xbmc/utils/Fanart.cpp
Paxxi: If we're super strict here passing as a const ref can lead to worse performance. It will disable some optimization opportunities and the string is copied anyway because it's the only use of it. 

Const ref isn't necessarily a good default after c++11 because move semantics totally changed the rules. 

This method isn't called in a hot path so this comment is only for information, no changes requested

xbmc/interfaces/legacy/ListItem.h
```diff
@@ -704,6 +741,69 @@ namespace XBMCAddon
...
+#else
+      void setAvailableFanart(const std::vector<Properties>& images);
+#endif
+
+#ifdef DOXYGEN_SHOULD_USE_THIS
+      ///
+      /// \ingroup python_xbmcgui_listitem
+      /// @brief \python_func{ addAvailableThumb(images) }
+      ///-----------------------------------------------------------------------
+      /// @brief Add a thumb yo available thumbs (needed for scrapers)
```
ronie: yo typo

xbmc/interfaces/legacy/ListItem.h
```diff
@@ -704,6 +741,69 @@ namespace XBMCAddon
...
+      ///
+      /// @param url            string (image path url)
+      /// @param aspect         [opt] string (image type)
+      /// @param spoof          [opt] string
+      /// @param cache          [opt] string
+      /// @param post           [opt] bool
+      /// @param isgz           [opt] bool
+      /// @param season         [opt] integer
```
ronie: perhaps a brief explanation of the params?
i have no idea what 'spoof' / 'isgz' / 'post' mean...

phate89: The problem is that I don't know too. I think they are the parameters to use to get the image (need use post or get, need to use gzip, referrer) but I'm not 100% sure and  I didn't found a scraper that uses them..They just pass XML with aspect and url

xbmc/interfaces/legacy/ListItem.h
```diff
@@ -272,11 +272,13 @@ namespace XBMCAddon
 #ifdef DOXYGEN_SHOULD_USE_THIS
       ///
       /// \ingroup python_xbmcgui_listitem
-      /// @brief \python_func{ setUniqueIDs(values) }
+      /// @brief \python_func{ setUniqueIDs(values, defrating) }
```
garbear: Can we use `defaultrating` here? I found this abbreviated name confusing.

xbmc/interfaces/legacy/ListItem.h
ronie: yup, i had a brief look at xbmc/utils/ScraperUrl.cpp and it looks like you're correct.

notspiff: the only one of these that are useful in the context is referrer (for thumb urls). the rest i added as props in the other version without applying brain.

phate89:
Updated with the latest changes requested..

phate89:
jenkins build this please

phate89:
jenkins build this please

PR #11420, opened 2017-01-11
[python] browse dialog - list network and/or local paths

ronie:
if you open the browse dialog from addon settings, it can list both local drives and network shares.
but if you call the browse dialog from the addon itself, no such option existed.

this to great annoyances of several (including me) python devs.
http://forum.kodi.tv/showthread.php?tid=290651

this pr unifies things by adding the same local/network drives option to xbmgui.Dialog.browse()

@phil65 @tamland

### DIFF
--- a/xbmc/interfaces/legacy/Dialog.cpp
+++ b/xbmc/interfaces/legacy/Dialog.cpp
@@ -31,6 +31,7 @@
 #include "video/dialogs/GUIDialogVideoInfo.h"
 #include "music/dialogs/GUIDialogMusicInfo.h"
 #include "settings/MediaSourceSettings.h"
+#include "storage/MediaManager.h"
 #include "dialogs/GUIDialogKaiToast.h"
 #include "ModuleXbmcgui.h"
 #include "guilib/GUIKeyboardFactory.h"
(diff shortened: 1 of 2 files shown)


xbmc/interfaces/legacy/Dialog.cpp
```diff
@@ -238,19 +239,30 @@ namespace XBMCAddon
...
       std::string mask = maskparam;
       VECSOURCES *shares = CMediaSourceSettings::GetInstance().GetSources(s_shares);
-      if (!shares) 
-        throw WindowException("Error: GetSources given %s is NULL.",s_shares.c_str());
+
+      VECSOURCES localShares;
+      if (!shares)
+      {
+        g_mediaManager.GetLocalDrives(localShares);
+        if (!s_shares.c_str() || strcmpi(s_shares.c_str(), "local") != 0)
```
akva2: very minor but .c_str() is guaranteed non-null (data + 0 terminator).

akva2: also strcmpi is a msft-ism, strcasecmp is the posix construct (kodi has defines so it works fine of course so mostly a FYI).

Rechi:
@tamland ping

PR #11421, opened 2017-01-11
[FavouritesDirectory] fix for adding addons to favourites via addonbrowser

phil65:
Fixes some scenarios where adding addons as favourites doesnt work properly.
@tamland

### DIFF
--- a/xbmc/filesystem/FavouritesDirectory.cpp
+++ b/xbmc/filesystem/FavouritesDirectory.cpp
@@ -197,8 +197,13 @@ std::string CFavouritesDirectory::GetExecutePath(const CFileItem &item, const st
       execute = StringUtils::Format("ActivateWindow(%s,%s,return)", contextWindow.c_str(), StringUtils::Paramify(item.GetPath()).c_str());
   }
   //! @todo STRING_CLEANUP
-  else if (item.IsScript() && item.GetPath().size() > 9) // plugin://<foo>
+  else if (item.IsScript() && item.GetPath().size() > 9) // script://<foo>
     execute = StringUtils::Format("RunScript(%s)", StringUtils::Paramify(item.GetPath().substr(9)).c_str());
+  else if (item.IsAddonsPath() && item.GetPath().size() > 9) // addons://<foo>
+  {
+    CURL url(item.GetPath());
+    execute = StringUtils::Format("RunAddon(%s)", url.GetFileName().c_str());
+  }
   else if (item.IsAndroidApp() && item.GetPath().size() > 26) // androidapp://sources/apps/<foo>
     execute = StringUtils::Format("StartAndroidActivity(%s)", StringUtils::Paramify(item.GetPath().substr(26)).c_str());
   else  // assume a media file


xbmc/filesystem/FavouritesDirectory.cpp
```diff
@@ -197,8 +197,15 @@ std::string CFavouritesDirectory::GetExecutePath(const CFileItem &item, const st
       execute = StringUtils::Format("ActivateWindow(%s,%s,return)", contextWindow.c_str(), StringUtils::Paramify(item.GetPath()).c_str());
   }
   //! @todo STRING_CLEANUP
-  else if (item.IsScript() && item.GetPath().size() > 9) // plugin://<foo>
+  else if (StringUtils::StartsWith(item.GetPath(), "script://"))
```
Montellese: This is not identical because this also allows to call the builtin with just `script://` which will not result in anything useful. The same applies for the other cases you added.

Furthermore I'd prefer to keep using `item.IsScript()` because that way we don't clutter the code with static `script://` strings all over the place.

xbmc/filesystem/FavouritesDirectory.cpp
```diff
@@ -197,8 +197,15 @@ std::string CFavouritesDirectory::GetExecutePath(const CFileItem &item, const st
...
     execute = StringUtils::Format("RunScript(%s)", StringUtils::Paramify(item.GetPath().substr(9)).c_str());
+  else if (StringUtils::StartsWith(item.GetPath(), "plugin://"))
```
Montellese: Please use `item.IsPlugin()` and also make sure that it has a path after the protocol by checking the length.

xbmc/filesystem/FavouritesDirectory.cpp
```diff
@@ -197,8 +197,15 @@ std::string CFavouritesDirectory::GetExecutePath(const CFileItem &item, const st
...
+    execute = StringUtils::Format("PunPlugin(%s)", StringUtils::Paramify(item.GetPath().substr(9)).c_str());
+  else if (item.IsAddonsPath())
```
Montellese: Please also check that there is a path after the protocol.

Montellese (changes requested):
Looks good apart from the minors.

phil65:
@Montellese updated.

phil65:
jenkins build this please.

xbmc/filesystem/FavouritesDirectory.cpp
```diff
@@ -197,8 +197,15 @@ std::string CFavouritesDirectory::GetExecutePath(const CFileItem &item, const st
       execute = StringUtils::Format("ActivateWindow(%s,%s,return)", contextWindow.c_str(), StringUtils::Paramify(item.GetPath()).c_str());
   }
   //! @todo STRING_CLEANUP
-  else if (item.IsScript() && item.GetPath().size() > 9) // plugin://<foo>
+  else if (item.IsScript() && item.GetPath().size() > 9) // script://<foo>
     execute = StringUtils::Format("RunScript(%s)", StringUtils::Paramify(item.GetPath().substr(9)).c_str());
+  else if (item.IsPlugin() && item.GetPath().size() > 9) // plugin://<foo>
+    execute = StringUtils::Format("PunPlugin(%s)", StringUtils::Paramify(item.GetPath().substr(9)).c_str());
```
tamland: RunPlugin does not do what you'd expect.. Plugins should use RunAddon

akva2: well, he's call *P*unPlugin in any case ;p

phil65: ah ok, now i got it, plugins get translated to ActivateWindow(xx,xx).... will fix.
(I like PunPlugin though, perhaps I will propose that in a follow-up PR)

phil65:
updated.

phil65:
jenkins build this please.

phil65:
worth a backport?

xbmc/filesystem/FavouritesDirectory.cpp
```diff
@@ -197,8 +197,13 @@ std::string CFavouritesDirectory::GetExecutePath(const CFileItem &item, const st
       execute = StringUtils::Format("ActivateWindow(%s,%s,return)", contextWindow.c_str(), StringUtils::Paramify(item.GetPath()).c_str());
   }
   //! @todo STRING_CLEANUP
-  else if (item.IsScript() && item.GetPath().size() > 9) // plugin://<foo>
+  else if (item.IsScript() && item.GetPath().size() > 9) // script://<foo>
     execute = StringUtils::Format("RunScript(%s)", StringUtils::Paramify(item.GetPath().substr(9)).c_str());
+  else if (item.IsAddonsPath() && item.GetPath().size() > 9) // addons://<foo>
+  {
+    CURL url(item.GetPath());
+    execute = StringUtils::Format("RunAddon(%s)", url.GetFileName());
```
Rechi: `url.GetFileName().c_str()`

phil65:
jenkins build this please.

PR #11422, opened 2017-01-11
follow up to 4cf4253b1fd3e65d1efe8d15cf86489b0738e75d

notspiff:
turns out signedness of width varies across freetype differences.
explicitly cast both sides of comparison to handle both cases.

it's either this or you bump freetype to >= 2.5.4 in depends. ref discussion in https://github.com/xbmc/xbmc/pull/11384


wsnipex:
I'm for this, since there are many old libs out in the (linux) wild

PR #11424, opened 2017-01-11
unify library node names

ronie:
use 'Years' / 'Titles' in library listings, as we use plurals for all other node names as well.

@HitcherUK @BigNoid @phil65


BigNoid:
This might be a good time to think of a new name for ```Titles```.  It's the only node that isn't named after what it contains.
How about ```All movies```, ```All tvshows```, etc?

HitcherUK:
Deja vu http://trac.kodi.tv/ticket/11991

ronie:
doesn't seem to be much interest in renaming the titles node?

if you feel it's needed, feel free to create a follow-up pr.

PR #11427, opened 2017-01-11
[CEC] dll name has changed to libcec.dll in 4.0.1

MartijnKaijser:


MartijnKaijser:
@opdenkamp ping

opdenkamp:
@opdenkamp just got another baby and didn't really follow things much recently ;) looks like the zip the wrong filename in if you need this. the old name (pre 4.0) was libcec.dll and it's now cec.dll, but for backwards compat, the zip used to contain both names. files are exactly the same, just with another name, so you can just copy it and repackage it if you have time before me.

MartijnKaijser:
congrats :)
I'll fix the package on the server

PR #11434, opened 2017-01-12
[PVR][guiinfo] Videofullscreen/Musicvisualisation: Direct channel number input

ksooo:
Instead of opening the numeric input dialog when pressing 0 - 9 while in PVR fullscreen or PVR music visualisation, now only the entered number appears on screen, like this works for almost all other tv sets and settop boxes.

2 seconds (existing advancedsettings value!) after the last digit was entered, Kodi will automatically switch to the channel with the entered number.

Entering an invalid number will be ignored and nothing will happen.

Entering more than 4 digits results in left-shifting the previously entered digits, discarding the oldest digit, and appending the just entered number.

Largest supported channel number is 9999.

Entering 4 zeros in a row will immediately end direct channel number input.

This PR was tested on latest Kodi master on Linux and macOS.

screenshot000

@da-anda fyi

@phil65 already reviewed the skin change

@ronie: skins must be adjusted to use the new info label PVR.ChannelNumberInput. Otherwise the feature will also work, but nothing will be seen on the screen while entering the number.

@Jalle19 for review?


ronie:
thanx for the heads-up!

system/settings/settings.xml
```diff
@@ -1411,7 +1411,7 @@
         </setting>
         <setting id="pvrplayback.confirmchannelswitch" type="boolean" label="19281" help="36231">
           <level>1</level>
-          <default>false</default>
+          <default>true</default>
```
afedchin: disagree with this.

ksooo: why?

ksooo: you know where this is used, do you?

ksooo: It is used exclusively with the channel surfing feature (pressing up/down key while in fullscreen PVR / music vis PVR). If you want channelchange without confirmation you can always use channel up/channel down keys on your remote.

Jalle19:
Isn't there a button in the OSD during full screen playback which used to open the numpad? AFAIK it was only for users with limited remotes.

ksooo:
> Isn't there a button in the OSD during full screen playback which used to open the numpad? AFAIK it was only for users with limited remotes.

The button was "1" to "9", which does not help for remotes without these keys.

Jalle19:
@ksooo I mean a GUI button in the OSD.

system/settings/settings.xml
afedchin: I don't like this because this setting require a confirmation for switching channel with numeric input. If it is changed by this PR, I don't mind

ksooo:
> @ksooo I mean a GUI button in the OSD.

 if there is such a GUI button i don't know it.

system/settings/settings.xml
ksooo: There is no switching channels with numeric input dialog anymore with this PR.

xbmc/pvr/PVRManager.h
```diff
@@ -603,6 +604,12 @@ namespace PVR
      */
     void PublishEvent(PVREvent state);
 
+    /*!
+     * @brief Get the nmeric channel input handler.
```
Jalle19: Typo (nmeric)

system/settings/settings.xml
ksooo: and the new feature introduced by this PR always works without confirmation.

xbmc/pvr/PVRNumericChannelInputHandler.cpp
```diff
@@ -0,0 +1,159 @@
...
+}
+
+void CPVRNumericChannelInputHandler::AppendChannelDigit(int iDigit)
+{
+  if (iDigit < 0 || iDigit > 9)
+    return;
+
+  CSingleLock lock(m_critSection);
+
+  if (iDigit == 0 && m_digits.size() == 0)
```
Jalle19: `m_digits.empty()`

xbmc/pvr/PVRNumericChannelInputHandler.cpp
```diff
@@ -0,0 +1,159 @@
...
+      m_digits.pop_front();
+
+    m_digits.emplace_back(iDigit);
+
+    // recalc channel string
+    m_strChannel.erase();
+    if (m_digits.size() != m_iDigits || GetChannel() > 0)
+    {
+      for (int digit : m_digits)
+        m_strChannel.append(std::to_string(digit));
```
Jalle19: I have a feeling `to_string` won't work on Android

system/settings/settings.xml
afedchin: ok then. sry for noise.

Jalle19 (changes requested):

system/settings/settings.xml
ksooo: okay, i try to by more clear:

before:
- 1-9 opens numeric input dialog
- setting influences whether dialog auto closes or not

after:
- 1-9 does not open dialog, only dispalys the entered number
- setting does not have an affect now
- always auto-switching

short: the setting we're discussing here has nothing to do with channel switching via numeric input anymore

so, no problem.

xbmc/pvr/PVRManager.h
ksooo: nitpicker nitpicker ... okay. :-)

xbmc/pvr/PVRNumericChannelInputHandler.cpp
ksooo: yep. will change.

xbmc/pvr/PVRNumericChannelInputHandler.cpp
ksooo: ugh. will jenkins let find out and find an alternative if it turns ot that you are right.

ksooo:
jenkins build this please

xbmc/pvr/PVRNumericChannelInputHandler.cpp
ksooo: Damn, you are right. Does not compile on Android.

system/settings/settings.xml
stefansaraev: > short: the setting we're discussing here has nothing to do with channel switching via numeric input anymore

it changes a well known behaviour. if not required for this PR, and if you really want to change that behaviour, it should be sent as a separate PR

ksooo:
@Jalle19 all changes you requested are in. good to go?

xbmc/pvr/PVRNumericChannelInputHandler.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include "pvr/channels/PVRChannelGroupsContainer.h"
+
+#include "PVRNumericChannelInputHandler.h"
+
+using namespace KODI::MESSAGING;
+
+namespace PVR
+{
+
+CPVRNumericChannelInputHandler::CPVRNumericChannelInputHandler(int iDigits, int iSwitchDelay)
```
xhaggi: I small question about the wording. Does ChannelNumberInputHandler fits better here?

xbmc/pvr/PVRNumericChannelInputHandler.cpp
```diff
@@ -0,0 +1,160 @@
...
+  for (std::deque<int>::const_reverse_iterator it = m_digits.rbegin(); it != m_digits.rend(); ++it)
+  {
+    iNumber += (*it * iDigitMultiplier);
+    iDigitMultiplier *= 10;
+  }
+
+  return iNumber;
+}
+
+std::string CPVRNumericChannelInputHandler::GetChannelAsString() const
```
xhaggi: Same for all the getter, GetChannelNumberAsString

xhaggi:
Looks good and makes sense because the numric input dialog can only be reached by the numpad of a remote and  then you will never use the numeric controls of the dialog.

xbmc/pvr/PVRNumericChannelInputHandler.cpp
ksooo: Yeah, like it. Will change the naming.

ksooo:
@xhaggi requested changes are in.

ksooo:
jenkins build this please

xhaggi:
Thanks, that was fast :)

xbmc/pvr/PVRChannelNumberInputHandler.cpp
```diff
@@ -0,0 +1,155 @@
...
+      m_digits.pop_front();
+
+    m_digits.emplace_back(iDigit);
+
+    // recalc channel string
+    m_strChannel.erase();
+    if (m_digits.size() != m_iDigits || GetChannelNumber() > 0)
+    {
+      for (int digit : m_digits)
+      m_strChannel.append(StringUtils::Format("%d", digit));
```
Jalle19: Missing indentation

xbmc/pvr/PVRChannelNumberInputHandler.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+#include "PVRChannelNumberInputHandler.h"
+
+using namespace KODI::MESSAGING;
+
+namespace PVR
+{
+
+CPVRChannelNumberInputHandler::CPVRChannelNumberInputHandler(int iDigits, int iSwitchDelay)
+: m_iDigits(iDigits),
```
Jalle19: I would rename this to m_iMaxDigits to better reflect its purpose

xbmc/pvr/PVRManager.cpp
```diff
@@ -162,7 +163,8 @@ CPVRManager::CPVRManager(void) :
     m_bEpgsCreated(false),
     m_progressHandle(NULL),
     m_managerState(ManagerStateStopped),
-    m_isChannelPreview(false)
+    m_isChannelPreview(false),
+    m_channelNumberInputHandler(4 /* digits */, g_advancedSettings.m_iPVRNumericChannelSwitchTimeout)
```
Jalle19: Maybe add a constant for 4?

Jalle19 (commented):
Apart from these +1, I don't really have an opinion on the setting default change

ksooo:
@Jalle19 all of the changes you requested are now in. thanks for the review.

ksooo:
jenkins build this please

ksooo:
jenkins build this please

da-anda:
ok, I tested this PR a bit with latest Milhouse build and found some issues.

- the number input didn't seem to work when playing a radio channel. All I could enter was 1, all other keys where ignored
- when in any PVR window (like the channels list) pressing a number still trigger the numpad OSD. We probably should get rid of it and instead jump to the related channel in the list

Then I was wondering about if we could probably use the channel switch delay setting for the delay of the numeric input instead of an advanced setting. IIRC this setting is currently only used when pressing up/down in the OSD and not for channelUp/channelDown actions. Just a thought.

ksooo:
Confirming the radio channel bug. Will have look.

The other pvr windows suggestion is for another PR. This is simply not part of this iteration.

Using the official setting instead of the advanced setting (that we were also using here before for whatever reason, btw) makes sense.

da-anda:
btw, thanks sooooo much for this PR. I think I owe you a brewery by now ;)

ksooo:
>> ok, I tested this PR a bit with latest Milhouse build and found some issues.
>> the number input didn't seem to work when playing a radio channel. All I could enter was 1, all other keys where ignored

> Radio channel input (other numbers than 1) works fine on my Mac, but not on my custom LibreELEC build.

Seems to be a remote vs. keyboard config issue problem. When using a keyboard, numbers other than 0 and 1 also work on LibreELEC.

Strange.

da-anda:
thought about this feature some more. I think it might be good if the "back" action could also abort the number input process. What do you think?

ksooo:
> thought about this feature some more. I think it might be good if the "back" action could also abort the number input process. What do you think?

 maybe later.

> edit: as for the numeric input. I just saw your fix by also listening to JUMPSMS. The correct fix IMO would be to adjust the remote.xml keymap 

 both is needed, to be on the "save" side.

ksooo:
jenkins build this please

ksooo:
Well, reworked everything and now it is also possible to use direct input in channel and guide window, @da-anda 

@Jalle19 could you please re-review?

xbmc/pvr/windows/GUIWindowPVRGuide.cpp
```diff
@@ -45,6 +44,7 @@ using namespace EPG;
 
 CGUIWindowPVRGuide::CGUIWindowPVRGuide(bool bRadio) :
   CGUIWindowPVRBase(bRadio, bRadio ? WINDOW_RADIO_GUIDE : WINDOW_TV_GUIDE, "MyPVRGuide.xml"),
+  CPVRChannelNumberInputHandler(1000),
```
Jalle19: Maybe a constant here?

Jalle19 (approved):
More complicated than the previous implementation, but hey, if it works...

da-anda:
I think we should drop the JumpSMS stuff now that we found the root cause of it not working on radio channels. What if users would prefer to jump to a certain channel in the list instead of switching channels by adjusting the keymap?

addons/skin.estuary/xml/DialogSeekBar.xml
```diff
@@ -1,11 +1,23 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <window>
-	<visible>[[Player.Seeking | Player.DisplayAfterSeek | [Player.Paused + !Player.Caching] | Player.Forwarding | Player.Rewinding | Player.ShowInfo | Window.IsActive(fullscreeninfo) | Window.IsActive(videoosd) | Window.IsActive(playerprocessinfo)] + Window.IsActive(fullscreenvideo)] | Window.IsActive(visualisation) | !String.IsEmpty(Player.SeekNumeric)</visible>
+	<visible>[[Player.Seeking | Player.DisplayAfterSeek | [Player.Paused + !Player.Caching] | Player.Forwarding | Player.Rewinding | Player.ShowInfo | Window.IsActive(fullscreeninfo) | Window.IsActive(videoosd) | Window.IsActive(playerprocessinfo)] + Window.IsActive(fullscreenvideo)] | Window.IsActive(visualisation) | !String.IsEmpty(Player.SeekNumeric) | !String.IsEmpty(PVR.ChannelNumberInput)</visible>
```
da-anda: this condition needs to be changed, because if TV or radio is playing in the background and you do a numeric input f.e. in channel list, the seekbar is also popping up. I think the correct condition to be added would be:
`[Window.IsActive(fullscreenvideo) | Window.IsActive(visualisation)] + !String.IsEmpty(PVR.ChannelNumberInput)`

ksooo:
> I think we should drop the JumpSMS stuff now that we found the root cause of it not working on radio channels.

 Nope. Currently not needed, but correct nevertheless.

ksooo:
> More complicated than the previous implementation,

 Not really; just factored out a base class of what we had before to reuse the code for three different use cases (before it was only one).

da-anda:
>> I think we should drop the JumpSMS stuff now that we found the root cause of it not working on radio channels.
> Nope. Currently not needed, but correct nevertheless.
it's two different things, isn't it? JumpSMS is about jumping by name while numeric input is jumping by number. But up to you.

addons/skin.estuary/xml/DialogSeekBar.xml
ksooo: Good catch. Will change. Thanks.

phil65: that vis condition is getting much too complex for my taste :) Will fix that in a follow-up though, lets get this PR in first.

ksooo: Sounds like a plan. Thanks & let's make it so, @phil65

ksooo:
jenkins build this please

ksooo:
jenkins build this please

ksooo:
jenkins build this please

ksooo:
jenkins error is unrelated

da-anda:
thanks a lot, huge improvement

PR #11437, opened 2017-01-13
[win][platinum] Change GWL_USERDATA to GWLP_USERDATA

Rechi:
## Description

Replace `GWL_USERDATA` with `GWLP_USERDATA`. They both are defined to the same value on Win32 so no real change on Win32.
Cherry picked witch permission from @Paxxi.

## Motivation and Context

On Win64 `GWL_USERDATA` is undefined and `GWLP_USERDATA` must be used.

## How Has This Been Tested?

Compiled on Windows (Win32 & x64).

## Types of change

## Checklist:


hudokkow:
If you're up to it, a patch file is welcome in https://github.com/xbmc/xbmc/tree/master/lib/libUPnP/patches

Montellese:
A patch file is not only welcome it's mandatory. Otherwise we cannot keep track of changes to external libraries.

Montellese (changes requested):
A patch file is not only welcome it's mandatory. Otherwise we cannot keep track of changes to external libraries.

Rechi:
@Montellese added patch file

Montellese (approved):

PR #11438, opened 2017-01-13
[peripherals] fix screensaver shot away

AlwinEsch:
## Description

The Joystick monitor triggers always a "OnAxisMotion(...)" also without a
real axis change.

This fix the always direct cone away mouse pointer after mouse move and the
screensavers who are startet and immediately stopped.

@garbear can you check this is OK

## Types of change

## Checklist:

### DIFF
--- a/xbmc/input/joysticks/JoystickMonitor.cpp
+++ b/xbmc/input/joysticks/JoystickMonitor.cpp
@@ -23,6 +23,10 @@
 #include "input/InputManager.h"
 #include "ServiceBroker.h"
 
+#include <cmath>
+
+#define AXIS_THRESHOLD       0.75f // Axis must exceed this value to be mapped
+
 using namespace KODI;
 using namespace JOYSTICK;
 
@@ -50,7 +54,9 @@ bool CJoystickMonitor::OnHatMotion(unsigned int hatIndex, HAT_STATE state)
 
 bool CJoystickMonitor::OnAxisMotion(unsigned int axisIndex, float position, int center, unsigned int range)
 {
-  if (position)
+  const bool bIsActive = (std::abs(position) >= AXIS_THRESHOLD);
+
+  if (bIsActive)
   {
     CServiceBroker::GetInputManager().SetMouseActive(false);
     return ResetTimers();


xbmc/input/joysticks/JoystickMonitor.cpp
```diff
@@ -22,6 +22,8 @@
 #include "Application.h"
 #include "input/InputManager.h"
 
+#define AXIS_THRESHOLD       0.75f // Axis must exceed this value to be mapped
```
AlwinEsch: This is taken from ./xbmc/input/joysticks/generic/ButtonMapping.cpp,  makes it sense to define in header to have global?

garbear:
on a snowboarding trip, I can review when I have some down time

AlwinEsch:
@garbear still on snowboarding trip? :-D ;-)

The problem is still present in Kodi, what do you think about them here?

xbmc/input/joysticks/JoystickMonitor.cpp
garbear: No, the threshold is application-dependent. For example, I chose 0.75f in the button mapper because pressure-sensitive buttons send both an analog and a digital button press, and this gives more weight to the digital button press. Here, this threshold determines the sensitivity to awakening the screen saver. 0.75f is a good value, though I don't see any reason it shouldn't be 0.5f.

xbmc/input/joysticks/JoystickMonitor.cpp
```diff
@@ -49,7 +51,9 @@ bool CJoystickMonitor::OnHatMotion(unsigned int hatIndex, HAT_STATE state)
 
 bool CJoystickMonitor::OnAxisMotion(unsigned int axisIndex, float position, int center, unsigned int range)
 {
-  if (position)
+  const bool bIsActive = (std::abs(position) >= AXIS_THRESHOLD);
```
garbear: This calculation needs to change since the Peripheral API bump to 1.3.0. See https://github.com/xbmc/xbmc/blob/c80e266/xbmc/input/joysticks/generic/InputHandling.cpp#L69 for example.

This should be:

```c++
const float magnitude = std::abs(position - center) / range;

if (magnitude >= AXIS_THRESHOLD)
{
  ...
```

garbear:
Unfortunately the solution is more complicated than this. Accelerometers and unmapped buttons will send erratic positions, and we also don't know the true center and range. This class should only interrupt the screen saver if the axis is mapped to a button on the controller.

AlwinEsch:
You are right there is a bit more needed, I close this them.

garbear:
Is this still an issue, or has it been fixed in one of my many joystick PRs?

AlwinEsch:
Was more a system fault and fixed here with https://askubuntu.com/questions/585706/deactivate-accelerometer

With them active are also not working games.

There was a fix from @wsnipex https://github.com/xbmc/xbmc/commit/d8d6f0842a0ef1f325b94d7c31d9ff3829905732 but was gone on other changes

garbear:
What if we only allow mapped axes to wake from screen saver?

AchimTuran:
@AlwinEsch ist this issue still present or should we close this PR?

Rechi:
@AlwinEsch this needs a rebase

Rechi:
@garbear is this obsolete with #12892?

garbear:
Yes

PR #11444, opened 2017-01-14
cmake: Fix an issue with path preventing passing multiple addons to ADDONS_TO_BUILD

samnazarko:
## Description
For a while now, passing multiple binary addons to -DADDONS_TO_BUILD was not working correctly, and the following error would occur:

```
-- Bootstrapping following addons: pvr.demo pvr.wmc
[100%] Completed 'binary-addons'
[100%] Built target binary-addons
CMake Error at CMakeLists.txt:220 (message):
No addons available to be built
```

Some improvements were made in 897094feeef8357eaf7ce2a4328a062b1af9d151, which means we no longer need to fix the separate_arguments. I however still found one small change necessary when building Kodi binary add-ons out of source. Here is a basic example on how to reproduce, after building Kodi. 

```
pushd project/cmake/addons
mkdir build
cd build
ADDONS_AUDIO_DECODERS="audiodecoder.modplug audiodecoder.nosefart"
ADDONS_TO_BUILD="${ADDONS_AUDIO_DECODERS}"
PLATFORM="-DCMAKE_INCLUDE_PATH=/opt/vc/include:/opt/vc/include/interface:/opt/vc/include/interface/vcos/pthreads:/opt/vc/include/interface/vmcs_host/linux -DCMAKE_LIBRARY_PATH=/opt/vc/lib"
cmake -DOVERRIDE_PATHS=1 -DCMAKE_INSTALL_PREFIX=${out}/usr/ -DBUILD_DIR=$(pwd) -DADDONS_TO_BUILD="${ADDONS_TO_BUILD}" "$PLATFORM" ../
$BUILD kodiplatform_DIR=$(pwd) build/
```

This needs a backport (note the path different to project/cmake vs cmake in master)

## Motivation and Context

It is not always appropriate to build all add-ons for a platform. This can be overcome by forking the Kodi binary add-ons repo and maintaining a branch for each platform, but this is time consuming when sometimes all you want to do is exclude some addon (i.e. ADSP).

## How Has This Been Tested?

This has been tested on both Krypton (producing test builds for OSMC) and master since Krypton was branched.

## Types of change

## Checklist:

### DIFF
--- a/cmake/addons/CMakeLists.txt
+++ b/cmake/addons/CMakeLists.txt
@@ -214,7 +214,7 @@ if(NOT addons)
                   WORKING_DIRECTORY ${BOOTSTRAP_BUILD_DIR})
 
   # now look for all the addons to be built again
-  file(GLOB_RECURSE addons ${ADDONS_DEFINITION_DIR}/*.txt)
+  file(GLOB_RECURSE addons ${BOOTSTRAP_BUILD_DIR}/binary-addons/src/binary-addons/*.txt)
 
   if(NOT addons)
     message(FATAL_ERROR "No addons available to be built")


wsnipex:
does running cmake . && make  in cmake/addons/bootstrap  before building the addons fix your issue?

samnazarko:
It did not fix it with local testing, no, but I have only tested quickly on Krypton. 

Please clarify whether the above commands in the first post should work or not without needing to run the bootstrap.

wsnipex:
yes it should work, but it doesn't. Your fix is not correct though, since trying to build other addons after the initial run will still fail.
Please try https://github.com/wsnipex/xbmc/commit/8a58663f0c0f0a0feaf4a8ccc727b4a527bb10a4

samnazarko:
Works

+1

PR #11445, opened 2017-01-14
Add support for SMB File Rename if same share

fedux:
## Description
If the file to be moved in ActionMove is in the same SMB share it can be renamed.

## Motivation and Context
Copying a file instead of renaming is inefficient as it has to be transferred twice over the network. This makes move operations a server side operation.

## How Has This Been Tested?
Tested in my own setup.

## Types of change

## Checklist:


Paxxi:
Thank you. Krypton is feature complete at this point so I don't think it's a good idea to merge this. 

We rely on several different smb implementations so ensuring that this does not introduce any bugs takes a lot of effort. 
We currently build against, Native windows, smb pre-gpl3, smb gold, libdsm that I know of. 

Seems like a good candidate to include in master though where it has time to get proper testing.

fedux:
Fair enough, rebased onto master.

PR #11386, opened 2017-01-06
[settings] move subtitle charset to standard level

stefansaraev:
move settings -> player -> language -> subtitle character set to basic level 

## Motivation and Context

this was requested in irc. let's make setting subtitle charset easier to newcomers

## Types of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [x] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #11449, opened 2017-01-14
Align the settings level for the 2 different character set settings

jjd-uk:
Changing the Player -> Language -> Subtitles -> "Character set" setting level to Standard (this also required changing the parent "subtitles.font" setting) to match the setting level of the Interface -> Regional -> Language -> "Character set" setting. Help string is also updated to inform users there is another "character set" setting for subtitles.

This is being changed after user feedback that they saw the "character set" setting in Interface -> Regional and thought it would apply to both the UI and to subtitles. 

@MartijnKaijser Your opinion? is it too late for for Krypton if it is an acceptable change?

Note that @stefansaraev has https://github.com/xbmc/xbmc/pull/11386 also open which in my opinion doesn't go far enough.

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -17326,7 +17326,7 @@ msgstr ""
 #. Description of setting with label #14091 "Character set"
 #: system/settings/settings.xml
 msgctxt "#36116"
-msgid "Choose which character set is used for displaying text in the user interface."
+msgid "Choose which character set is used for displaying text in the user interface. This doesn't change the character set used for subtitles, for that go to Player > Language."
 msgstr ""
 
 #. Description of setting with label #14079 "Timezone country"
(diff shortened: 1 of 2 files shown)


MartijnKaijser (approved):

MartijnKaijser:
OK for mr

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -17326,7 +17326,7 @@ msgstr ""
 #. Description of setting with label #14091 "Character set"
 #: system/settings/settings.xml
 msgctxt "#36116"
-msgid "Choose which character set is used for displaying text in the user interface."
+msgid "Choose which character set is used for displaying text in the user interface. This does change the character set used for subtitles, for that go to Player > Language."
```
Rechi: I think you meant `This doesn't change` ...

jjd-uk: Yes you're right, will change

jjd-uk:
Minor mistake as pointed out by Rechi is corrected.

PR #11447, opened 2017-01-14
App-name fixes for Krypton

Kwiboo:
## Description

This PR contains fixes to make Kodi Krypton more app-name friendly:
- Partial backport of #10946 to fix a target name on windows
- Includes a backport of #11085 to use `%APPDATA%\<app-name>` as userdata folder
- Installs `<app-name>.desktop` into `usr/share/applications/` instead of `kodi.desktop`
- Uses `${APP_NAME_LC}` in `PathSetup.cmake` instead of `kodi`
- ~Generates `KodiConfig.cmake` instead of `${APP_NAME}Config.cmake` as all binary addons looks for a `Kodi` package and `KODI_INCLUDE_DIR`~

## Motivation and Context

Building Kodi Krypton using a different app-name (OpenPHT in my case) leaves some parts with kodi in the name or path.

## How Has This Been Tested?

Clean builds of Kodi and binary addons with this PR and my [app-name-openpht](https://github.com/xbmc/xbmc/compare/Krypton...Kwiboo:app-name-openpht) branch on Windows, macOS and Ubuntu Xenial.
Autotools build have not been tested.

## Types of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [x] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #11452, opened 2017-01-14
App-name fixes for master

Kwiboo:
## Description

This PR contains fixes to make Kodi more app-name friendly:
- Installs `<app-name>.desktop` into `usr/share/applications/` instead of `kodi.desktop`
- Uses `${APP_NAME_LC}` in `PathSetup.cmake` instead of `kodi`
- ~~Generates `KodiConfig.cmake` instead of `${APP_NAME}Config.cmake` as all binary addons looks for a `Kodi` package and `KODI_INCLUDE_DIR`~~

This is the same as #11447 but for master.

## Motivation and Context

Building Kodi using a different app-name (OpenPHT in my case) leaves some parts with kodi in the name or path.

## How Has This Been Tested?

Clean builds of Kodi and binary addons with this PR on Windows, macOS and Ubuntu Xenial.

## Types of change

## Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,7 @@ include(cmake/scripts/common/AddOptions.cmake)
 include(cmake/scripts/common/ArchSetup.cmake)
 include(cmake/scripts/common/Macros.cmake)
 include(cmake/scripts/common/ProjectMacros.cmake)
+core_find_versions()
 include(cmake/scripts/${CORE_SYSTEM_NAME}/PathSetup.cmake)
 include(ExternalProject)
 
@@ -76,7 +77,6 @@ else()
 endif()
 
 core_find_git_rev(APP_SCMID FULL)
-core_find_versions()
 
 # Dynamically loaded libraries built with the project
 add_custom_target(${APP_NAME_LC}-libraries)
--- a/cmake/scripts/android/PathSetup.cmake
+++ b/cmake/scripts/android/PathSetup.cmake
@@ -29,5 +29,5 @@ list(APPEND final_message "Includedir: ${includedir}")
 list(APPEND final_message "Datarootdir: ${datarootdir}")
 list(APPEND final_message "Datadir: ${datadir}")
 
-set(PATH_DEFINES -DBIN_INSTALL_PATH=\"${libdir}/kodi\"
-                   -DINSTALL_PATH=\"${datarootdir}/kodi\")
+set(PATH_DEFINES -DBIN_INSTALL_PATH=\"${libdir}/${APP_NAME_LC}\"
+                   -DINSTALL_PATH=\"${datarootdir}/${APP_NAME_LC}\")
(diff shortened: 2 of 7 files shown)


MartijnKaijser:
@wsnipex @fetzerch @hudokkow ping

CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ include(cmake/scripts/common/AddOptions.cmake)
 include(cmake/scripts/common/ArchSetup.cmake)
 include(cmake/scripts/common/Macros.cmake)
 include(cmake/scripts/common/ProjectMacros.cmake)
+core_find_versions()
```
hudokkow: Any reason to move this?

hudokkow:
I can't help feeling that this is going in the wrong direction.
Apart from the s/kodi/${APP_NAME}/ and innocuous desktop file, the rest is just taking us where we don't want to be: hardcoding file and path names.
In this case, add-ons are the real problem. We should be able to pass ${APP_NAME} and variants to addons build env.

Ideas are welcome.

CMakeLists.txt
Kwiboo: ${APP_NAME_LC} is used in PathSetup.cmake with this commit

Kwiboo:
As a third-party developer I see the Kodi-addons config/includes more like an SDK for binary addons and because of that a hardcoded name makes more sense to me and is what all/most binary addons is currently looking for in their cmake-files.

Since I will only be developing on top on the Krypton branch for next few months and there is other ongoing work in the addons system I will drop the `KodiConfig.cmake instead of ${APP_NAME}Config.cmake` commit in this PR.

MartijnKaijser:
ping @wsnipex

wsnipex:
I agree with @hudokkow 
this reverts changes done lately and is going backwards. Passing the APP_NAME to addons would be trivial, but would require a little change in every addon :(
@notspiff, what do you think?

notspiff:
We cannot really pass this to addons in a fully automated fashion due to chicken-egg problems. We can add a parameter in each buildsys to specify the cmake package name to use in place of Kodi. I agree that is probably something that should be done (over time..)

Kwiboo:
@wsnipex @notspiff unsure if you are referring to the dropped commit or the change to `PathSetup.cmake` in this PR.

To solve the chicken-egg problem @AlwinEsch had a good idea in https://github.com/xbmc/xbmc/pull/11378#issuecomment-270756448
"I thought to find maybe a special development kit name who can hold independent from Kodi, so that add-on must always use a `#include <how-is-your-kit-name/...h>`."

The complete set of changes in #11447 is what I currently need to build Krypton and all binary addons using a different app-name (OpenPHT) without having to patch all binary addons looking for another kit-name, other suggestions are welcomed.
This PR have already dropped all changes related to the binary addon kit-name.

wsnipex:
yes, I was referring to the dropped commit. Whats left here is ok.
jenkins build this please

hudokkow:
jenkins build this please

PR #11453, opened 2017-01-14
Revert "DVDVideoCodecAndroidMediaCodec: Don't call into App while oth…

fritsch:
This will be needed when decoder reconfiguring is implemented, but for now it keeps the last frame on screen which we don't want. So plain revert it.


fritsch:
jenkins build this please

PR #11454, opened 2017-01-14
[pydocs] Corrections on the xbmcvfs/file examples

enen92:
This PR applies some fixes to the examples provided in the documentation for the `xbmcvfs` module / `File`.

## Description
`xbmcvfs.File()` receives only 2 arguments: the file (string) and the mode which is optional ('w'). Examples provided in the documentation are not in agreement with the method definition. One of the examples use a list for the argument `mode`. The other uses a third argument that does not exist.

## How Has This Been Tested?

Tested the provided examples in Python (they lead to errors) and also checked the code in the header file.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/interfaces/legacy/File.h
+++ b/xbmc/interfaces/legacy/File.h
@@ -43,11 +43,10 @@ namespace XBMCAddon
     /// \python_class{ xbmcvfs.File(filepath, [mode]) }
     ///
     /// @param filepath             string Selected file path
-    /// @param mode                 [opt] string Additional mode options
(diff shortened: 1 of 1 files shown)


ronie:
the brackets mean the arg is optional, not that the arg should be a list ;-)

enen92:
hum ok. I found it misleading on the example provided:

```python
f = xbmcvfs.File(file, ['w'])
```

I'll change it back. Is the other one correct?

phil65:
Changes look fine, we dont mark optional arguments with brackets in the examples.

PR #11462, opened 2017-01-15
move the touch command to the codegenerator makefile which runs on all platforms

Memphiz:
this should fix building with autotools/xcode again after #11400

### DIFF
--- a/codegenerator.mk
+++ b/codegenerator.mk
@@ -69,3 +69,4 @@ else
 #build json builder - ".." because makefile is in the parent dir of "bin"
 	$(MAKE) -C $(abspath $(@D)/..)
 endif
+$(shell touch $(TOPDIR)/xbmc/cores/AudioEngine/AEDefines_override.h)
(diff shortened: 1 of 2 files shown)


Memphiz:
jenkins build this please

wsnipex:
while at it, can you also add the override headers for droid? https://github.com/xbmc/xbmc/blob/Krypton/xbmc/cores/AudioEngine/overrides/android/AEDefines.h
Since we don't know the target OS at this stage, I'd simply ifdef them

Memphiz:
@wsnipex i only added the missing file I have no clue at all what about this PR is or what it should do ... if you really wanna drag me into it i can try during the week sometime ... busy ATM

wsnipex:
then just leave as is and be done with it. It shouldn't have been merged like this in the first place, so at least it will build.

MartijnKaijser:
@Memphiz can this be merged?

Memphiz:
Jep

PR #11467, opened 2017-01-15
restore the ability to build without python and add cmake option

notspiff:
Sure, a build without python is of little interest in itself. However, it restores the ability to use valgrind with kodi. python 2.7 does lots of naughty things that screws up inside the valgrind vm, rendering it unusable (lots and lots of wolf-wolf). 

I'm sure the use of add_defintions() isn't kosher, so @fetzerch please instruct.


AlwinEsch:
From my side also OK, it can help on some places.

There I hope to finish my big addon system rework soon where Python becomes complete removed and handled with another way. No fear, Kodi can then still use all Python add-ons but then from external part who do his work for Kodi and what allow also a way for other programming languages.

CMakeLists.txt
```diff
@@ -225,6 +226,10 @@ elseif(CORE_SYSTEM_NAME STREQUAL rbpi)
   core_require_dep(MMAL)
 endif()
 
+if(PYTHON_FOUND)
```
fetzerch: Usually we do this in the find modules via XX_DEFINITIONS

CMakeLists.txt
```diff
@@ -307,7 +312,9 @@ set_target_properties(lib${APP_NAME_LC} PROPERTIES PREFIX "")
 add_dependencies(lib${APP_NAME_LC} libcpluff ffmpeg dvdnav crossguid)
 
 # Codegen
-add_subdirectory(${CMAKE_SOURCE_DIR}/xbmc/interfaces/swig build/swig)
+if(ENABLE_PYTHON)
```
fetzerch: Can't you move it cmake/treedata/optional/common/python.txt or is it executed too late then?

CMakeLists.txt
notspiff: yes i know, but i found it a bit tedious (and possibly confusing) to introduce a layer on top of the built-in python find module just to set the definition. but if that's what you want..

CMakeLists.txt
notspiff: sure i can, but i thought you had it explicitly in the buildsystem for a reason (other than technical).

CMakeLists.txt
notspiff: ah i see we already override the built-in module. then it's a no brainer.

CMakeLists.txt
fetzerch: The only thing I could imagine is that it either needs some vars from the main cmakelists or would get introduced too late otherwise. Maybe you can just quickly try if it compiles, otherwise we can also leave it here. Maybe you had it explicitly there for a reason: https://github.com/fetzerch/xbmc/commit/538ef57c8b78b85ac592fa9caa1a55f7d991378d :D

CMakeLists.txt
fetzerch: Yep, just because we already have the module

CMakeLists.txt
notspiff: aah it came in from there. in that case there used to be a reason, but there is none now :)

notspiff:
requested changes applied.

MartijnKaijser:
jenkins build this please

cmake/modules/FindPython.cmake
```diff
@@ -29,11 +29,12 @@ else()
   list(APPEND PYTHON_LIBRARIES ${PC_PYTHON_STATIC_LIBRARIES})
 endif()
 
+list(APPEND PYTHON_DEFINITIONS -DHAS_PYTHON=1)
 
 include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(PYTHON REQUIRED_VARS PYTHON_INCLUDE_DIR PYTHON_LIBRARY PYTHON_LIBRARIES)
+find_package_handle_standard_args(PYTHON REQUIRED_VARS PYTHON_INCLUDE_DIR PYTHON_LIBRARY PYTHON_LIBRARIES PYTHON_DEFINITIONS)
```
fetzerch: is there a reason why you add it here? Usually we only add vars that are the result of find_* (and that have a chance to be set to NOT-FOUND)

cmake/modules/FindPython.cmake
```diff
@@ -29,11 +29,12 @@ else()
...
 if(PYTHON_FOUND)
   set(PYTHON_INCLUDE_DIRS ${PYTHON_INCLUDE_DIR})
 endif()
 
-mark_as_advanced(PYTHON_EXECUTABLE PYTHON_INCLUDE_DIRS PYTHON_INCLUDE_DIR PYTHON_LIBRARY PYTHON_LIBRARIES PYTHON_LDFLAGS FFI_LIBRARY EXPAT_LIBRARY INTL_LIBRARY GMP_LIBRARY)
+mark_as_advanced(PYTHON_EXECUTABLE PYTHON_INCLUDE_DIRS PYTHON_INCLUDE_DIR PYTHON_LIBRARY PYTHON_LIBRARIES PYTHON_LDFLAGS FFI_LIBRARY EXPAT_LIBRARY INTL_LIBRARY GMP_LIBRARY PYTHON_DEFINITIONS)
```
fetzerch: PYTHON_DEFINITIONS is not a cached variable (and shouldn't be one), so it doesn't make sense to mark it as advanced.

fetzerch (commented):
Almost ;)

cmake/modules/FindPython.cmake
notspiff: since we already require python(libs) by the time we hit this, i figured the additional sanity check wouldn't hurt in case some enterprising individual removed it by accident/stupidity. removed.

cmake/modules/FindPython.cmake
notspiff: meh. wrong context. working on a project where all things that come out of find rules have to be cached to ensure gui visibility (some people prefer to hack the cache instead of installing shit sanely). you are of course correct, removed.

cmake/modules/FindPython.cmake
```diff
@@ -29,6 +29,7 @@ else()
   list(APPEND PYTHON_LIBRARIES ${PC_PYTHON_STATIC_LIBRARIES})
 endif()
 
+list(APPEND PYTHON_DEFINITIONS -DHAS_PYTHON=1)
```
wsnipex: please move to L38

notspiff: moved inside the if(true) as requested

wsnipex:
thanks, looks fine now
jenkins build this please

stefansaraev:
@notspiff you also need http://sprunge.us/SfVL

notspiff:
just in case u didn't get my point: since we pass REQUIRED to all the find_package calls unconditionally, we will never reach the if() without python being found. you might want to use Python_FIND_REQUIRED if the rule is intended to support the flag. thanks @stefansaraev picked and squashed.

fetzerch:
looks good now, thanks much!

tamland:
There is a file that lets your suppress python errors in valgrind.

MartijnKaijser:
Jenkins build and merge

PR #11471, opened 2017-01-16
FIX: [droid] Copy splash to xxxhdpi drawable

koying:
Prevents crash on very large dpi devices, as a "plain" drawable is assumed mdpi

Not very bright, I must say, but I see no alternative.

@da-anda we should probably have a 720p and 1080p splash for droid, or we end up adding 1.8Mb to the apk for not much. At least a 720p for lower dpi would limit the damages.


koying:
jenkins build this please

da-anda:
a JPG would save even more space than sticking to the PNG. Yes, PNG is lossless, but with our splash screens becoming more complex compared to the old XBMC ones, filesize is increasing quite a lot. I can prepare a 720p splash though if needed.

koying:
I assume you want to keep a PNG for our actual Splash?
If so, yes, a 720p JPG will help

da-anda:
I'm not set on PNG for our main Splash. For a complex splash like the Krypton one PNG doesn't have much benefit, so I'd switch to JPG entirely, at least for this release

MartijnKaijser:
So just keep bigger splash for now or take a 720p one?

da-anda:
I had prepared a 720p PNG splash as well as a 1080p JPG. Both are on our google drive.

PR #11475, opened 2017-01-16
Crash on view current playlist with many items Ticket #17188 

DaveTBlake:
This fixes the RC3 crash that happens on trying to view the current playlist when it contains a large number of items. See Trac Ticket http://trac.kodi.tv/ticket/17188 and discussed on the forum http://forum.kodi.tv/showthread.php?tid=304133

If the current playlist was large enough (relative to processor power) to need to show the progress dialog then Kodi would crash shortly after this was closed. Reported on Windows as a music library issue, but probably would effect Video too if the playlst was large. Also seem in testing on Linux (LE) - I think the same issue but have been unable to prove it. The issue was a zero divide.

On GUI_MSG_WINDOW_DEINIT CGUIDialogProgress::Reset sets m_iMax = 0. But subsequently
CMusicInfoLoader::LoadItemLookup calls CGUIDialogProgress::SetProgressAdvance causing a zero over zero division.

Crash not always easy to reproduce as seems to be compiler optimisation dependant (0 / 0 being handled differently?). While I could catch the issue while debugging the Jenkins build of RC3, I could not repeat it in any of my local builds. It also seems to ocurrs in v16.1 on LE, but much harder to reproduce and I could not catch in the debug.

Too test queue a large number of songs to play e.g. ~4000 songs on Windows, say from Genres node, or even unscanned music files from file view, and then click "Goto playlist" on the sideblade. If you have selected a sufficient number of tracks the progress dialog will be displayed until the current playlist is on view.

Has been backported as #11474

### DIFF
--- a/xbmc/dialogs/GUIDialogProgress.cpp
+++ b/xbmc/dialogs/GUIDialogProgress.cpp
@@ -160,7 +160,8 @@ void CGUIDialogProgress::SetProgressAdvance(int nSteps/*=1*/)
   if (m_iCurrent>m_iMax)
     m_iCurrent=0;
 
-  SetPercentage((m_iCurrent*100)/m_iMax);
+  if (m_iMax > 0)
+    SetPercentage((m_iCurrent*100)/m_iMax);
 }
 
 bool CGUIDialogProgress::Abort()


jjd-uk:
Replicate crash in Windows and tested with KodiSetup-20170116-f25c7e2-HEAD.exe and confirmed crash does not happen in that build.

jenkins build this please

Paxxi (dismissed):
The check should be where m_iMax is set and also the constructor/reset method should set it to 1.
This solution will avoid the crash but you get a progress bar that does not update itself.

DaveTBlake:
@Paxxi not so sure that iMax set to 1 at reset (deinitialise) is right either.  The bigger question is why is Kodi trying to advance the progress on a dialog that has been deinitialised.  Unfortunately I can't work that out.

DaveTBlake:
>This solution will avoid the crash but you get a progress bar that does not update itself.

Having looked at this more I see that `iMax=0` is the way we indicate that the progress bar is not in use and so do not want it to update. `CMusicInfoLoader::LoadItemLookup` attempts to advance progress bar every time, it does not check to see if the bar has been opened (because 1.5s has passed).

PR #11477, opened 2017-01-17
add pycryptodome python module

peak3d:
Add PyCryptoDome python cryptography module

## Description
see title

## Motivation and Context
Add crypto to python addon developing

## Types of change

### DIFF
--- a/system/addon-manifest.xml
+++ b/system/addon-manifest.xml
@@ -29,6 +29,7 @@
   <addon>screensaver.xbmc.builtin.black</addon>
   <addon>screensaver.xbmc.builtin.dim</addon>
   <addon>script.module.pil</addon>
+  <addon>script.module.pycryptodome</addon>
   <addon>service.xbmc.versioncheck</addon>
   <addon>skin.estuary</addon>
   <addon>skin.estouchy</addon>
--- a/addons/script.module.pycryptodome/addon.xml
+++ b/addons/script.module.pycryptodome/addon.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<addon id="script.module.pycryptodome"
+       name="Python Crypto Library"
+       version="3.4.3"
+       provider-name="Legrandin">
+  <requires>
+    <import addon="xbmc.python" version="2.7.0"/>
+  </requires>
+  <extension point="xbmc.python.module" library="lib" />
+  <extension point="xbmc.addon.metadata">
+    <platform>all</platform>
+  </extension>
+</addon>
--- a/cmake/installdata/common/addons.txt
+++ b/cmake/installdata/common/addons.txt
@@ -23,6 +23,7 @@ addons/webinterface.default/*
 addons/screensaver.xbmc.builtin.dim/*
 addons/screensaver.xbmc.builtin.black/*
 addons/script.module.pil/*
+addons/script.module.pycryptodome/*
 addons/script.module.pysqlite/*
 addons/resource.language.en_gb/*
 addons/resource.uisounds.kodi/*
--- a/project/BuildDependencies/scripts/0_package.list
+++ b/project/BuildDependencies/scripts/0_package.list
@@ -37,6 +37,7 @@ mysql-connector-c-6.1.6-win32-vc140-v2.7z
 openssl-1.0.2g-win32-vc140-v2.7z
 pcre-8.37-win32-vc140-v3.7z
 pillow-3.1.0-win32-vc140.7z
+pycryptodome-3.4.3-win32.7z
 python-2.7.11-win32-vc140-v2.7z
 shairplay-0.9.0-win32-vc140-v2.7z
 sqlite-3.10.2-win32-vc140.7z
--- a/tools/android/packaging/Makefile.in
+++ b/tools/android/packaging/Makefile.in
@@ -88,7 +88,7 @@ shared:
 	mkdir -p assets
 	cp -rfp $(PREFIX)/share/@APP_NAME_LC@/* ./assets
 	find `pwd`/assets/ -depth -name ".git" -exec rm -rf {} \;
-	find `pwd`/assets/ -name "*.so" -not -name "*imaging*.so" -exec rm {} \;
+	find `pwd`/assets/ -name "*.so" -exec rm {} \;
 	find `pwd`/assets/addons/skin.*/media/* -depth -not -iname "Textures.xbt" -exec rm -rf {} \;
 	cd `pwd`/assets/addons; rm -rf $(EXCLUDED_ADDONS)
 	mkdir -p assets/system/certs
@@ -133,6 +133,7 @@ libs: $(PREFIX)/lib/@APP_NAME_LC@/lib@APP_NAME_LC@.so
 	find $(PREFIX)/share/@APP_NAME_LC@/addons -name "*.so" -exec cp -fp {} xbmc/obj/local/$(CPU)/ \;
 	find $(PREFIX)/lib/@APP_NAME_LC@/addons -name "*.so" -exec cp -fp {} xbmc/obj/local/$(CPU)/ \; || true
 	find $(PREFIX)/lib/@APP_NAME_LC@/system -name "*.so" -exec cp -fp {} xbmc/obj/local/$(CPU)/ \;
+	DIR=${CURDIR}; cd $(PREFIX)/lib/python2.7/site-packages; for i in `find Crypto -name \*.so` ; do FN=`echo $$i | cut -c1- | tr "/" "_"` ; mv $$i $$DIR/xbmc/obj/local/$(CPU)/$$FN ; done
 	cd xbmc/obj/local/$(CPU)/; find . -name "*.so" -not -name "lib*.so" | sed "s/\.\///" | xargs -I@ mv @ lib@
 	cp -fp xbmc/obj/local/$(CPU)/*.so xbmc/lib/$(CPU)/
 	$(STRIP) --strip-unneeded xbmc/lib/$(CPU)/*.so
--- a/tools/depends/target/Makefile
+++ b/tools/depends/target/Makefile
@@ -13,7 +13,7 @@ DEPENDS = \
 	python27 libshairplay \
 	libplist libcec libbluray tinyxml dummy-libxbmc \
 	libssh taglib libusb libnfs \
-	pythonmodule-pil pythonmodule-setuptools \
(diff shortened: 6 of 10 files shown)


stefansaraev:
not agains this but here my 2 cents if it matters:

we should avoid adding any native python modules, addons should use as much from python stdlib as they can. pil is more than enough pain, already.

Paxxi:
@stefansaraev agreed but unfortunately python does not expose the crypto library from openssl, maybe gnutls? So this is one of the bits that makes sense to ship as a binary. 

Writing this it just hit me that maybe we can expose these through Kodi and avoid the python build pain? 

Same goes for pillow but that's a rather large API surface

stefansaraev:
yea. sure.

anyhow. this should be done as script.module.pil is then - make a new addon, let's say script.module.pycryptodome, add it to addon-manifest.xml, and let 3rdparty addons do ```<import addon="script.module.pycryptodome" version="x.y.z">```

peak3d:
1.) @stefansaraev looked at the PIL patches and decided aginst such a thing (beside the fact that I don't understand all the mess to move / copy things into an addon-folder)
2.) The current PR uses the concept how python modules are normally installed. The dependency is installed with python --install in site-packages folder (and yes, it feels correct)
3.) I have removed the SelfTest for kod (wich saves 15MB of things wedon't use)
4.) For android packaging this PR renames the .so files using the folder names (Crypto/Hash/test.so will become Crypto_Hash_test.so and resides in androids system/lib path.

@wsnipex any objections?

wsnipex:
lets see if it builds:
jenkins build this please

peak3d:
It's at no place triggered, means every build comes for now without crypto stuff

wsnipex:
you have to add it to depens/target/Makefile

peak3d:
@stefansaraev what is the difference / benefit to make it "same as PIL" ? Seems that you have at least more background than me about those py things.

peak3d:
jenkins build this please

stefansaraev:
@peak3d it has few benefits to not ship in site-packages/ but as a separate addon, and require 3rdparties to explicitly import it:

- that's how it works with all other (non std) python stuff, example: addons dont just import beautifulsoup. they are required to import script.module.beautifulsoup4 in their addon.xml

- it would be clear if a specific addon needs (and uses) pycryptodome, you can identify those addons by simple grep in repo's addons.xml in case you want to remove pycryptodome in future :)

also, there was an effort to force addons not use import from site-packages, but use only modules that are already in the repo (ones we can control), see #5127. there are good reasons to do so. pil (and now pycryptodome) makes this task harder

peak3d:
Thanks a lot for making this clear, @stefansaraev. Is PIL part of the kodi package?
I haven't found script addons in repo.binary.addon, but PIL is binary, isn't it?

stefansaraev:
https://github.com/xbmc/xbmc/tree/master/addons/script.module.pil

on linux, it's shipped as addon.xml only, and Pillow package from system is used (example: ubuntu, python-pil is a required runtime dependency, it is automaticaly pulled by apt-get when you install kodi). embedded distros (OE, LE, buildroot based) take care to have it cross-compiled and installed. on non-linux - it is part of the kodi package.

ios
```
seo@x230 ~/Desktop/ios $ ar x ../org.xbmc.kodi-ios_17.0-0-rc3_iphoneos-arm.deb 
seo@x230 ~/Desktop/ios $ tar tf data.tar.lzma | grep Pillow
./Applications/Kodi.app/Frameworks/lib/python2.7/site-packages/Pillow-3.0.0-py2.7-macosx-10.4-x86_64.egg
```

osx
```
seo@x220 ~/Desktop $ 7z l kodi-17.0-Krypton_rc3-x86_64.dmg  | grep Pillow
2017-01-12 18:51:01 .....      1118086      1118208  Kodi/Kodi.app/Contents/Libraries/lib/python2.7/site-packages/Pillow-3.0.0-py2.7-macosx-10.4-x86_64.egg
```

android
```
seo@x230 ~/Desktop $ unzip -l kodi-17.0-Krypton_rc3-armeabi-v7a.apk  | grep -i script.module.pil | pastebinit 
http://sprunge.us/TjIj
```
windows
```
seo@x230 ~/Desktop $ find kodi-17.0-Krypton_rc3/uni_pfatxtv/script.module.pil/ | pastebinit 
http://sprunge.us/hdQf
```

ok. on osx/ios it is in site-packages/ no idea why

peak3d:
would be  a simple dummy-addon (script.module.pycryptodome) enough?
Wich is generated during the build process currently used?

If pycrypto is not build, the addon is not generated.

Ok, for linux we have to set the same dependencies for kodi install, but to be honest, I don't see the reason why copying (for windows) all the files from a to b. let the things in site_packages, write the dummy-addon and it should work like PIL.

Pls. correct me if it's nit correct.

stefansaraev:
a dummy addon (on all platforms) + runtime dependency on linux (ubuntu) would be enough yes. on non linux.. honestly, I have no idea if it should be in site-packages or an .egg file inside addon's dir. IMO it should be an .egg in addon's dir, but maybe there are platform limitations, idk, platform devs know better

peak3d:
The complete addon system is currently reworked by @alwinus, I want to keep work as small as possible ( I hate wasting time for nothing).

My suggestion:
- I'll add an dummy addon
- I'll look why pycryptodome is currently not build for win (is there an other location for deps??)
- I try to figure out where the runtime dep is located (do you have a tipp)??

Then we wait for alwinus to finish his work and figure out a solution wich fits best for both PIL / crypto.
Ok with that?

stefansaraev:
> The complete addon system is currently reworked by @alwinus, I want to keep work as small as possible ( I hate wasting time for nothing).

afaik, he is working on binary addons, not python suff..

> My suggestion:

> I'll add an dummy addon

a right thing to do, imo.

> I'll look why pycryptodome is currently not build for win (is there an other location for deps??)

idk, cc @Paxxi and other windows people

> I try to figure out where the runtime dep is located (do you have a tipp)??

uh.. none. as said, ios/osx seems to be different than others and I have no idea why.

> Then we wait for alwinus to finish his work and figure out a solution wich fits best for both PIL / crypto.

> Ok with that?

sure, I am ok with everything, I just express an opinion here, before it's too late, my comments are not a requests for change ;)

peak3d:
Your experience is far better than mine in this area - I appreciate your input a lot!

> afaik, he is working on binary addons, not python suff..
alwinus is also changing much / all of the python concept

stefansaraev:
btw, if script.module.pycryptodome is not added to addon-manifest.xml, it will (unless somthing has changed recently and I missed it) be added to addondb as "disabled" addon on first run.

peak3d:
thx,!

Rechi:
Windows doesn't build any deps under tools/depends. Most of them get downloaded as binary.
Someone needs to compile it for windows, pack it and upload to the mirrors.
>I'll look why pycryptodome is currently not build for win (is there an other location for deps??)

peak3d:
ok, @Rechi just looked into pillow 7z file, I'll try to prepare such a thing

peak3d:
jenkins build this please

peak3d:
@wsnipex there are 2 red once in jenkins build but for me they don't seem to be PR related (tests fail)
If you agree, I'll merge...

Rechi:
Maybe the change in `tools/depends/target/python27/modules.setup`  causes the Segmentation fault.

system/addon-manifest.xml
```diff
@@ -29,6 +29,7 @@
   <addon>screensaver.xbmc.builtin.black</addon>
   <addon>screensaver.xbmc.builtin.dim</addon>
   <addon>script.module.pil</addon>
+  <addon>script.module.pycryptodome</addon>
```
MartijnKaijser: This should not be there (PIL shouldn't either). It is not a mandatory addon and any addon that actually requires it should import it like normal dependency which should auto enable it.

tools/depends/target/pythonmodule-pycryptodome/addon.xml
```diff
@@ -0,0 +1,13 @@
...
+<addon id="script.module.pycryptodome"
+       name="Python Crypto Library"
+       version="3.4.3"
+       provider-name="Legrandin">
+  <requires>
+    <import addon="xbmc.python" version="2.7.0"/>
+  </requires>
+  <extension point="xbmc.python.module" library="lib" />
+  <extension point="xbmc.addon.metadata">
+    <platform>all</platform>
```
MartijnKaijser: This is false cause it's not there on Windows

MartijnKaijser (changes requested):

system/addon-manifest.xml
MartijnKaijser: Also this is not packaged on Windows so it would break starting Windows.

MartijnKaijser (changes requested):

tools/depends/target/pythonmodule-pycryptodome/addon.xml
peak3d: its available on windows, just finished the 7z

system/addon-manifest.xml
ronie: i thought the manifest file decides which addons are enabled by default?
iirc if it's not listed the script.module will be installed, but disabled.

peak3d:
@ronie afaiu @MartijnKaijser  correct, it will be enabled at the time the first 3rd-party addon requests it.

peak3d:
jenkins build this please

system/addon-manifest.xml
stefansaraev: yes, that's exactly what it is doing (and you can not disable/unintall those addons), @ronie

peak3d:
@stefansaraev i read from the (outdated) comment to ronie that we should add pycryptodome to addon-manifest.xml, right?

stefansaraev:
@peak3d yes, I commented on that yesterday.

agian, in short: every addon that is shipped with the kodi distributionn MUST be added to addon-manifest.xml. it is NOT an optional thing. otherwise addoninstaller would add that addon to addondb in DISABLED state. this was a well-discussed DESIGN decision and I doubt it is going to be changed.

stefansaraev:
one more thing: it would not only be added as disabled to addondb, but addons of type "xbmc.python.module" are not shown in the addonmanager GUI (addons -> my addons -> all), so user has NO chance to manualy enable it

Rechi:
`addons/script.module.pycryptodome/*` must be added in `cmake/installdata/common/addons.txt`

peak3d:
jenkins build this please

PR #11478, opened 2017-01-17
[Fix] Restore caching for slow plugins Ticket #17211

DaveTBlake:
This effectively reverts #10394 that removed caching for all plugins, and fixes ticket http://trac.kodi.tv/ticket/17211 It needs to be backported.

As discussed on the internal forum, the radical change in caching policy is a regression for those plugins that fetch data online. The user temporarily switching to say player OSD or current playlist should not require the plugin to requery the online data.  The radio addon is a clear example, populating the list of online radio stations available is a slow query. 

Like that example, much plugin content is pretty static, so it is unnecessary to keep fetching the same listings over and over again. It is also unfriendly, putting an additional load on the servers of the source sites, and where the API keys are limited by the amount of call you're allowed to make will run the risk of running out of credits.

Any future changes to the plugin caching policy need to be done in a measured way, not just dropped.

This does leave the issue with TV shows added to favorites http://trac.kodi.tv/ticket/16560 unresolved (this is what #9898 and #10394 were attempting to solve).

### DIFF
--- a/xbmc/filesystem/PluginDirectory.cpp
+++ b/xbmc/filesystem/PluginDirectory.cpp
@@ -228,7 +228,7 @@ void CPluginDirectory::EndOfDirectory(int handle, bool success, bool replaceList
     return;
 
   // set cache to disc
-  dir->m_listItems->SetCacheToDisc(CFileItemList::CACHE_NEVER);
+  dir->m_listItems->SetCacheToDisc(cacheToDisc ? CFileItemList::CACHE_IF_SLOW : CFileItemList::CACHE_NEVER);
 
   dir->m_success = success;
   dir->m_listItems->SetReplaceListing(replaceListing);
--- a/xbmc/interfaces/legacy/ModuleXbmcplugin.h
+++ b/xbmc/interfaces/legacy/ModuleXbmcplugin.h
@@ -131,8 +131,8 @@ namespace XBMCAddon
     /// @param updateListing        [opt] bool - True=this folder should
     ///                             update the current listing/False=Folder
     ///                             is a subfolder(Default).
-    /// @param cacheToDisc          [opt] bool - True=Allow folder to be cached
-    ///                             (default)/False=this folder
+    /// @param cacheToDisc          [opt] bool - True=Folder will cache if
+    ///                             extended time(default)/False=this folder
     ///                             will never cache to disc.
     ///
     ///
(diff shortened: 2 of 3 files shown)


DaveTBlake:
Just a heads up that I've done this @MartijnKaijser , @tamland, @popcornmix , @phil65, @BigNoid, @arnova, @koying, @dentaku65, @romanvm ... and anyone I have missed.

Been trying to chase another regerssion and got so sick of radio addon being unusably slow. If someone else would like to look at the unresolved TV Show favorites issue, I'm not sure I understand that.

BigNoid:
The trac ticket is a very specific use case, but all plugin (no matter the plugin url) started from favourites or any other skin based shortcut use the cached results instead of fresh ones. This means the user sees no new content until the cache is removed if all they do is start it from favourites or home screen.

DaveTBlake:
Both tickets - what happens without caching and the favorites issue - give specific use cases. I realsie that the issues are wider, but the consensus does seem that dropping caching is not the way to solve the favorites problem. 

@BigNoid perhaps you understand the suggestions on the forum for the favorites fix better than I do? Oh yes @coolllman heads up that this is revert is proposed, so favorites will still be an issue.

koying:
I suspect that ```Refresh(true)``` in https://github.com/xbmc/xbmc/blob/2966375f853bb01c3be571a892190f4a128c224b/xbmc/windows/GUIMediaWindow.cpp#L501 fixes the "plugin from favourite" issue

Paxxi:
Seems to me that the onus to cache data in the scenario you describe should be on the addon author. This would ofc require planning and communication with addon authors so not a quick fix.

DaveTBlake:
jenkins build this please

BigNoid:
@koying your proposed fix works fine afaict, thx!
@DaveTBlake if you want you can pick https://github.com/BigNoid/xbmc/commit/cf23b94dd28b9ca654d7d8e43c6f289c3484a414 in this PR, or I'll do a separate PR.

arnova:
Looks good

DaveTBlake:
@BigNoid glad you have looked at the favorites fix. I see no reason not to include the one liner with this, so will add a commit to this PR.

@Paxxi if the team wants to review caching policy for v18 that's fine by me. This is just a quick regression fix for v17

BigNoid:
@DaveTBlake Wasn't the second commit also supposed to be added to the backport in that case? I'll put up a PR for that for Krypton.

DaveTBlake:
Yes @BigNoid , I cocked up :( Fingers too fast!
Just doing a separate PR

PR #11480, opened 2017-01-17
[win32][fix] Quick test to fix one addon installer issue

Paxxi:
This maybe can fix the addon installer issue? Or maybe not, doesn't seem to have broken my local test so yay for that

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/FilesystemInstaller.cpp
+++ b/xbmc/addons/FilesystemInstaller.cpp
@@ -26,7 +26,34 @@
 #include "utils/StringUtils.h"
 #include "utils/URIUtils.h"
 
+#ifdef TARGET_POSIX
+#include "XTimeUtils.h"
+#endif
+
 using namespace XFILE;
+
+namespace
+{
+
+bool renameOrRetry(const std::string & source, const std::string & dest, const char * description)
+{
+  auto count = 1;
+  auto result = false;
+  do
+  {
+    result = CFile::Rename(source, dest);
+    if (!result)
+    {
+      CLog::Log(LOGERROR, "Failed to move %s addon files from '%s' to '%s', retrying in 500ms",
+			          description, source.c_str(), dest.c_str());
+      Sleep(500);
+    }
+  } while (!result && count++ < 4);
+
+  return result;
+}
+
+} // end namespace unnamed
 
 CFilesystemInstaller::CFilesystemInstaller()
 {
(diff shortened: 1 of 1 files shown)


xbmc/addons/FilesystemInstaller.cpp
```diff
@@ -52,18 +52,36 @@ bool CFilesystemInstaller::InstallToFilesystem(const std::string& archive, const
...
+      }
+    } while (!result && count++ < 4);
+
+    if (!result)
       return false;
-    }
   }
 
-  if (!CFile::Rename(newAddonData, addonFolder))
+  auto count = 1;
```
thebrid: These 12 lines are near-duplicates of the 12 lines starting on line 55. Consider refactoring them into a function.

Paxxi: They are but slightly different so I decided against making it a function as it wouldn't be much better for just these two places

thebrid: The 2 pieces perform the same operation on different inputs. There are a lot of advantages to eliminating duplication. If someone finds a bug in one, they may correct it there but not in the other. Similarly if someone makes an enhancement - they may apply it in one copy but not the other.

MartijnKaijser: @thebrid if you are willing to do so feel free to make a PR. We have many areas that need work and we lack enough developers.

thebrid: Coming up :smile_cat:

xbmc/addons/FilesystemInstaller.cpp
```diff
@@ -52,18 +52,36 @@ bool CFilesystemInstaller::InstallToFilesystem(const std::string& archive, const
...
-    if (!CFile::Rename(addonFolder, oldAddonData))
+    auto count = 1;
+    auto result = false;
+    do
     {
-      CLog::Log(LOGERROR, "Failed to move old addon files from '%s' to '%s'", addonFolder.c_str(), oldAddonData.c_str());
+      result = CFile::Rename(addonFolder, oldAddonData);
+      if (!result)
+      {
+        CLog::Log(LOGERROR, "Failed to move old addon files from '%s' to '%s', reytring in 500ms", addonFolder.c_str(), oldAddonData.c_str());
```
thebrid: Also typo here, will fix.

xbmc/addons/FilesystemInstaller.cpp
```diff
@@ -52,18 +52,36 @@ bool CFilesystemInstaller::InstallToFilesystem(const std::string& archive, const
...
+        Sleep(500);
```
thebrid: Build is complaining of missing declaration for Sleep(). Should this be including threads/Thread.h?

xbmc/addons/FilesystemInstaller.cpp
thebrid: @Paxxi , sent a PR to your branch.

xbmc/addons/FilesystemInstaller.cpp
FernetMenta: XTimeUtils.h for posix

thebrid: :+1:

Paxxi:
jenkins build this please

PR #11205, opened 2016-12-18
[guilib] implicit referencing for listitem bools / ints

phil65:
some code refactoring + also allow implicit referencing for listitem bools, listitem ints and container bools.

@ronie @xhaggi

PR #11484, opened 2017-01-18
[gui] fix container bools

phil65:
Should fix regression mentioned in https://github.com/xbmc/xbmc/pull/11205.
@ronie for a quick test.


ronie:
thx! fixes the issue indeed.

phil65:
jenkins build this please.

PR #11486, opened 2017-01-18
[python] add setType() method for edit controls

ronie:
allow addons to define the edit control type.

@tamland @phil65


tamland:
The types should be mapped to a constants using the SWIG_CONSTANT2 macro (so they can be read by humans). Also, I would add this parameter to the constructor instead as I don't see why you would want to change the type after creating it.

ronie:
@tamland i indeed need the functionality to change the input type on the fly.
similar to our smartplaylist editor, where input type depends on the selection made by another control in the same window:
smartplaylist-rule


i've added the constants, as requested

xbmc/interfaces/legacy/Control.cpp
```diff
@@ -1055,6 +1055,16 @@ namespace XBMCAddon
       return msg.GetLabel();
     }
 
+    void ControlEdit::setType(int type, const String& heading)
+    {
+      if (pGUIControl)
+      {
+        LOCKGUI;
+
+        ((CGUIEditControl *) pGUIControl)->SetInputType((CGUIEditControl::INPUT_TYPE)type, CVariant{heading});
```
thebrid: Consider using C++ style casts over C-style.

tamland:
ok, makes sense (and changes looks good!)

MartijnKaijser:
ping @ronie :) Not sure if you can also take care of the comment

ronie:
i can, and i will. just not sure when :-)

xbmc/interfaces/legacy/Control.cpp
Paxxi: I agree with you on this one.
this would be `static_cast<CGUIEditControl*>(pGUIControl)->` and same for the type

Paxxi (changes requested):
can you make the changes and we'll get this merged

xbmc/interfaces/legacy/Control.h
```diff
@@ -1056,6 +1057,43 @@ namespace XBMCAddon
...
+      /// @param type              integer - type of the edit control.
+      /// | Param                            | Definition                                  |
+      /// |----------------------------------|:--------------------------------------------|
+      /// | xbmcgui.INPUT_TYPE_TEXT          | (standard keyboard)
+      /// | xbmcgui.INPUT_TYPE_NUMBER        | (format: #)
+      /// | xbmcgui.INPUT_TYPE_DATE          | (format: DD/MM/YYYY)
+      /// | xbmcgui.INPUT_TYPE_TIME          | (format: HH:MM)
+      /// | xbmcgui.INPUT_TYPE_IPADDRESS     | (format: #.#.#.#)
+      /// | xbmcgui.INPUT_TYPE_PASSWORD      | (input is masked)
+      /// | xbmcgui.INPUT_TYPE_PASSWORD_MD5  | (input is masked, return md5 hash of input)
```
romanvm: Maybe it's worth using a more secure hash (if possible)?

Rechi:
@Paxxi ping

Paxxi (approved):
I don't really remember what I wanted changed but I think it looks fine

Rechi:
@ronie ready to get merged?

ronie:
yup

PR #11488, opened 2017-01-18
[python] add input validation for edit controls

ronie:
allow addons to use input validation for edit controls.

@tamland @phil65 @Montellese


tamland:
same here: the rule numbers should be mapped to readable constants

xbmc/interfaces/legacy/Control.cpp
```diff
@@ -1059,6 +1060,38 @@ namespace XBMCAddon
...
+
+        if (rule == 0)
+          validator = StringValidation::NonEmpty;
+        else if (rule == 1)
+          validator = StringValidation::IsInteger;
+        else if (rule == 2)
+          validator = StringValidation::IsPositiveInteger;
+        else if (rule == 3)
+          validator = StringValidation::IsTime;
+        ((CGUIEditControl *) pGUIControl)->SetInputValidation(validator);
```
thebrid: Consider using C++-style casts.

PR #11490, opened 2017-01-19
fixed: context menu manager mess with profiles

notspiff:
killed e.g. 'mark as watched'.
ref http://forum.kodi.tv/showthread.php?tid=290858

### DIFF
--- a/xbmc/ContextMenuManager.cpp
+++ b/xbmc/ContextMenuManager.cpp
@@ -45,7 +45,13 @@ CContextMenuManager::CContextMenuManager(CAddonMgr& addonMgr)
 
 CContextMenuManager::~CContextMenuManager()
 {
+  Deinit();
+}
+
+void CContextMenuManager::Deinit()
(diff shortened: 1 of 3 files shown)


phil65:
Thx for the fix! this was bugging me too.

MartijnKaijser:
jenkins build this please

fergalmoran:
Do we really have to wait until v18 before we see this fix? It's a SERIOUSLY frustrating bug.

stefansaraev:
no need to be angry, it was backported to Krypton branch few hours ago. but you are on github so you should know that already.

MartijnKaijser:
@fergalmoran I would even revert the fix just because of your response!

fergalmoran:
@stefansaraev I'm not angry, I'm just expressing how frustrating the bug is. I hadn't noticed it was backported when I responded.

@MartijnKaijser If you think that's proportionate, fair enough.

PR #11492, opened 2017-01-19
[fix][audio] quick fix for some decoder that cannot accept empty package

zeas2:
## Description
Some .wmv video with audio in WMALossLess may decode fail.
It seems that ffmpeg's wmalossless decoder cannot accept empty (or zero sized) package data.

## Motivation and Context
This quick fix solved that.

## How Has This Been Tested?
Test environment: Win10x64, VS2017RC
I tested with some local media files and nothing bad caused.
WMV with WMALossLess now sounds OK.
Tested other media files like avi with mp3, standalone WMALossLess and normal wma, FLAC file, and working fine.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayerAudio.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayerAudio.cpp
@@ -503,6 +503,9 @@ void CVideoPlayerAudio::Process()
         // guess next pts
         m_audioClock += audioframe.duration;
 
+		if (consumed >= pPacket->iSize) // quick fix for some decoder like wmalossless that cannot accept empty package
+			break;
+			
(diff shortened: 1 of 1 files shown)


xbmc/cores/VideoPlayer/VideoPlayerAudio.cpp
```diff
@@ -503,6 +503,9 @@ void CVideoPlayerAudio::Process()
         // guess next pts
         m_audioClock += audioframe.duration;
 
+		if (consumed >= pPacket->iSize) // quick fix for some decoder like wmalossless that cannot accept empty package
```
FernetMenta: apart from wrong indentation this is not correct either. decoder may produce more than a single output packet for a packet that goes in.
can you provide a sample that causes issues?

Rechi:
@zeas2 is your issue still present after #11946?

Rechi:
If this issue is still present in current master, please provide an example.

PR #11500, opened 2017-01-20
[PVR][Settings] Remove 'hide connection lost warning' setting and code.

ksooo:
Cleanup. We do not need this setting anymore since we have "async connect" for pvr client addons (introduced with krypton), because we do no longer try to load addons in a (potential endless) loop.

Tested on latest master on macOS.

@Jalle19 like always, for the review? ;-)


Jalle19 (approved):

ksooo:
jenkins build this please

ksooo:
jenkins build this please

PR #11501, opened 2017-01-20
fix skip specified amount

ronie:
as reported on the forum: http://forum.kodi.tv/showthread.php?tid=304515
forward/backward skipping a specified amount of time is currently broken
(http://kodi.wiki/view/Skip_steps#Skip_the_specified_amount)

i'm unsure if this is the correct way to fix it, as we never needed these key mappings before.

@afedchin

### DIFF
--- a/xbmc/utils/SeekHandler.cpp
+++ b/xbmc/utils/SeekHandler.cpp
@@ -361,6 +361,7 @@ bool CSeekHandler::SeekTimeCode(const CAction &action)
     case ACTION_STEP_BACK:
     case ACTION_BIG_STEP_BACK:
     case ACTION_CHAPTER_OR_BIG_STEP_BACK:
+    case ACTION_MOVE_LEFT:
     {
       SeekSeconds(-GetTimeCodeSeconds());
       return true;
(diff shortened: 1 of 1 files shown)


jjd-uk:
@xhaggi may also be able to advise.

xhaggi:
Sorry for the late response but all the GitHub mails were marked as spam on my end.

After a deeper look at this, I'm not sure how this worked before ;)

The numeric seek stuff was removed by FernetMenta  and later re-added by afedchin.

https://github.com/xbmc/xbmc/pull/10706/files#diff-a14477a07f11f1e01caa83e24c7794a6L160
https://github.com/xbmc/xbmc/pull/10769/files#diff-c647c68d33c198547ac05b3c7e0eafc4R352

Nothing was changed. The keys to trigger the seek actions are the same.

And the keymaps for `<VideoTimeSeek>` were not changed since 7 years.
https://github.com/xbmc/xbmc/blame/601c7a490c5a6e0fbb914b63c93d6c8b3c756a27/system/keymaps/keyboard.xml#L404

Your fix looks valid but why did it work in Kodi 16.1?

xhaggi:
After a second look I would suggest to add `ACTION_MOVE_LEFT` and `ACTION_MOVE_RIGHT` to the switch case at https://github.com/xbmc/xbmc/pull/10769/files#diff-c647c68d33c198547ac05b3c7e0eafc4R352 instead of map left/right to one of the valid actions.

ronie:
@xhaggi done (and sorry for the late response as well)

jjd-uk:
Did a quick test and it looks to be working exactly how it did previously in v16 and before.

PR #11502, opened 2017-01-21
Resolution: Do not switch resolutions when in 3D (only refreshrates)

fritsch:
Since my updates to the Adjust Refreshrate to match video logic I received not too much negative feedback, besides some driver bugs and AVRs suddenly confronted with 4k there was only this one thing.

For 3D if the resolution is given with computed SAR, e.g. 2:1 for a 1920 source would end up with 3840 but TVs most likely only do their 3D mode in 1920 x 1080 I workaround this.

I don't have a 3D TV and cannot test it, this might fix:  http://trac.kodi.tv/ticket/17251 - as the user uses LibreELEC, he cannot verify or help me testing. Perhaps @lrusak @MilhouseVH can provide help there.

### DIFF
--- a/xbmc/guilib/Resolution.cpp
+++ b/xbmc/guilib/Resolution.cpp
@@ -224,11 +224,13 @@ RESOLUTION CResolutionUtils::FindClosestResolution(float fps, int width, bool is
       // evaluate all higher modes and evaluate them
       // concerning dimension and refreshrate weight
       // skip lower resolutions
+      // don't change resolutions when 3D is wanted
       if ((width < orig.iScreenWidth) || // orig res large enough
          (info.iScreenWidth < orig.iScreenWidth) || // new res is smaller
          (info.iScreenHeight < orig.iScreenHeight) || // new height would be smaller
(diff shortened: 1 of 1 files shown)


fritsch:
On my system: Before the patch -> Sample ends up in 3840, after the patch sample ends up in 1920x1080.

da-anda:
@popcornmix mind having a look? As the PI does support native 3D resolutions you can probably tell best if TVs would detect 3840x1080 as 3D or not.

fritsch:
For PI it was not a problem. Cause they run in 1080p and won't switch that up - as no larger resolution is available.

da-anda:
ah, with 3840 you didn't mean 3840x1080 but the actual 4K resolution? If so, then nvm

FernetMenta:
this would break working systems, by i.e. not switching to 24hz on a non-4k tv, right?

fritsch:
Yes, just related to the resolution. Meaning: Don't switch resolutions if a 3D mode is asked for.

fritsch:
jenkins build this please

da-anda:
not sure where this logic is best suited, but a user reported that it would be necessary to switch resolution down to 1080p for 3D movies in case UI etc is in 4K

fritsch:
He should run kodi in 1080p - looks much better than in 4k with our bilinear / nearest neighbour like gui scalers.

PR #11504, opened 2017-01-21
cec: Do not use tv language by default

LipkeGu:
## Description
This PR disables the option "use_tv_menu_language".
## Motivation and Context

TVs from manufacturers like LG sends "en_us" also when the TV language
is "de_de". Users should enable it when they want to use the TV menu
language but not by default. Also kodi does not show any text
when the system or kodi does not have the language installed.

## How Has This Been Tested?
Compiled a image from source with this PR. Kodi is using its default language.
The cec setting is disabled.

## Types of change

## Checklist:


opdenkamp:
This is an LG problem, which completely ignores the spec and says the language is Korean. Since the percentage of people using this code who speak Korean is pretty small, hardcoding it to English was the best work around.

There's no problem with the feature itself, and I've heard from lots of people that it's very nice that Kodi switches to their native language after installing it, so I don't see any reason to disable this for everyone who bought a TV from a vendor who doesn't ignore the specification (anyone other than LG). If you have an LG and don't want English, then I suggest you disable this.

So nack from my end.

LipkeGu:
OK :)

PR #11505, opened 2017-01-21
add another overload for begin()/end()

notspiff:
msvc needs this one as well

### DIFF
--- a/xbmc/FileItem.h
+++ b/xbmc/FileItem.h
@@ -755,6 +755,8 @@ public:
 
   VECFILEITEMS::const_iterator begin() { return m_items.cbegin(); }
   VECFILEITEMS::const_iterator end() { return m_items.cend(); }
+  VECFILEITEMS::const_iterator begin() const { return m_items.begin(); }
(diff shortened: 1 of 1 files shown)


MartijnKaijser:
@Montellese ping
jenkins build this please

Paxxi:
Are the two lines above this change actually needed? Seems like this should replace those

Montellese:
@Paxxi: in #11003 @notspiff stated that the compiler didn't pick up on them the other way around. But I haven't tested that myself.

Paxxi:
Oh OK. Didn't mean for notspiff to change it but mostly wondering why, no objections from me on merging.

notspiff:
For reasons that probably makes sense if you investigate details it does
not consider const begin/end in a mutable context. Gcc considers
cbegin/cend for const context but it seems msvc does not. We could perhaps
get away without cbegin/cend but we need const and nonconst begin/end
explicitly

21. jan. 2017 22:45 skrev "Pär Björklund" <notifications@github.com>:

Montellese:
Gave it a quick test with MSVC++ and works fine with this. jenkins build and merge

PR #9094, opened 2016-02-08
Userrating in MusicOSD

razzeee:
This adds our 1-10 userrating dialogselect as a bindable key. So skins can call "showrating" and it will show up. Should also work for keymaps.

It also changes the onscreen rating display as you change it. 
The only drawback is, that as long as the song you just rated is playing, you can reopen the rating dialog and won't see the rating preselected. If it was rated previous to your play, this will work.

I also need a skinner to change redo my skin work, as mine was just for tests.
So any takers? @phil65 @HitcherUK ?

PR #11513, opened 2017-01-21
[Fix] Multiple changes of song rating from player OSD (ticket #17190)

DaveTBlake:
This fixes trac http://trac.kodi.tv/ticket/17190,  that was actually a known design issue since implementation of update of song user rating from music OSD  in #9094

As @Razzeee said about this new feature at the time "The only drawback is, that as long as the song you just rated is playing, you can reopen the rating dialog and won't see the rating preselected. "
To reproduce some of the symptoms:

Start playback of a track that is currently unrated
Set the rating to, let's say, 3
Click the rating button **again** and you'll notice the selection dialog does not represent the current rating 
Select 'no rating' and your rating will still be 3

The issue was that whilst the db was being correcly updated, as were any other windows, the rating value was not set in `g_application.CurrentFileItem`.  Using  std::make_shared<CFileItem> this could never work.

@Paxxi and @fritsch kindly helped with a simple solution, adding a method to CApplication, thank you both. 
@ronie it seems I got to fix this anyway, perhaps you and @Razzeee could test.

I have tested setting rating during playback in as many ways as I can, but it helps to have more.  It is even happy if the playback has moved to a new song before the rating is selected.

If it looks good then I will backport soonest

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -4685,6 +4685,11 @@ const std::string& CApplication::CurrentFile()
   return m_itemCurrentFile->GetPath();
 }
 
+std::shared_ptr<CFileItem> CApplication::CurrentFileItemPtr()
+{
+  return m_itemCurrentFile;
+}
+
 CFileItem& CApplication::CurrentFileItem()
 {
   return *m_itemCurrentFile;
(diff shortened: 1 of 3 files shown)


fritsch:
Much more safe than: ```CFileItem& CurrentFileItem()``` - fine with me.

ronie:
tested and works nicely. thx!

Paxxi (approved):

PR #11532, opened 2017-01-23
AESinkAUDIOTRACK: Be more explicit when logging PT method

fritsch:
Print which PT method we use in more detail.


fritsch:
jenkins build this please

PR #11536, opened 2017-01-24
[PVR] Trac 17246: Fix deadlock that might occur during initial channel icon search.

ksooo:
This user reported deadlock might occur on kodi startup when switching to the channels window while user supplied channel icons are still read and set.

Fixes trac 17246: http://trac.kodi.tv/ticket/17246

Stack traces (supplied by bug reporter):
kodi-frozen

Ghist of the fix is to put the problematic long running and guilib-calling (g_graphicsContext locking) channel icon search and set code into a  CJob implementation.

Fix has been tested on latest master on macOS by me and using test build on Windows by the bug reporter.

@Jalle19 for review?

### DIFF
--- a/xbmc/pvr/channels/PVRChannelGroup.cpp
+++ b/xbmc/pvr/channels/PVRChannelGroup.cpp
@@ -33,6 +33,7 @@
 #include "settings/lib/Setting.h"
 #include "settings/Settings.h"
 #include "threads/SingleLock.h"
+#include "utils/JobManager.h"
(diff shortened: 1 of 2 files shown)


Paxxi:
Oh that looks like an awesome visual Studio feature, must try it out. 

Fix looks sane to me

Jalle19 (approved):
Looks good to me

ksooo:
jenkins build this please

PR #11539, opened 2017-01-24
[gitignore] add /addons/script.module.pycryptodome/

Rechi:
This was forgotten in #11477.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,6 +125,7 @@ cmake_install.cmake
 /addons/skin.estouchy/media/Textures.xbt
 /addons/skin.pm3-hd/media/Textures.xbt
 /addons/script.module.pil/
+/addons/script.module.pycryptodome/
 /addons/audioencoder.*
 /addons/pvr.*
 /addons/adsp.*


alexhass:
Netflix addon require this pycryptodome. How can users install it now?

Rechi:
Github is for development only and not for user support. Please use the Kodi forum next time.

Just install a recent master nightly and pycryptodome will be include.

alexhass:
Thanks for feedback Rechi. I'm not asking for support here, it's more for missing references. Forums are black holes and the issue is a specific result of exacly this commit. So 100% in this context.

I have Milhouse 9.0 Build 1208, LibreELEC (community): devel-20171208232721-#1208-g712987f (Generic.x86_64) and it is missing. That build is only 2 days old. When has it added back? So I can ping milhouse maybe.

MilhouseVH:
script.module.pycryptodome is present in the LE you mention:
```
LibreELEC (community): devel-20171208232721-#1208-g712987f (Generic.x86_64)
NUC:~ # ls -la /usr/share/kodi/addons/script.module.pycryptodome
total 1
drwxrwxr-x    2 root     root            32 Dec  8 23:42 .
drwxrwxr-x   89 root     root          2661 Dec  8 23:46 ..
-rw-r--r--    1 root     root           416 Dec  8 23:42 addon.xml
NUC:~ # python
Python 2.7.14 (default, Dec  8 2017, 20:33:29)
[GCC 7.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import Cryptodome
>>>
```
Do you say it's not present because the Kodi add-on system claims it is a missing dependency? 
s1

Unfortunately this is (most probably) a Kodi issue (due to the built-in nature of the addon confusing Kodi), and is just cosmetic - Python works so ignore it.

As @Rechi says this is an issue for the forum so I'll reply again to your there, and keep the conversation there.

PR #11541, opened 2017-01-24
[PVR][application] Fix update of PVR client addons

ksooo:
Update of PVR client addons did only work if the shared library name and or path to the shared library did not change during update, which is usually not the case as most (if not all?) shared libraries for PVR client addons have the version number in it's name.

This PR contains two commits. First commit is specific for macOS/iOS, second commit is platform independent.

First commit fixes a crash that occured on macOS while updating a pvr client addon. Due to the bug fixed by this PR, addon manager tried to load the old, already deleted shared library. This failed and led to a crash in <code>CAddonDll::LoadDll()</code> (https://github.com/xbmc/xbmc/blob/master/xbmc/addons/AddonDll.cpp#L116). It seems, that support for "alternate binary addon path" was simply forgotten for macOS/iOS, I added it, @Memphiz

Second commit fixes the actual addon update bug. Fix is in PVR code, addon manager is not affected.

Fix has been runtime tested on latest master on macOS and latest Krypton on Linux.

@Jalle19 for review?

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -951,6 +951,10 @@ bool CApplication::InitDirectoriesOSX()
   else
     userHome = "/root";
 
+  std::string binaddonAltDir;
+  if (getenv("KODI_BINADDON_PATH"))
+    binaddonAltDir = getenv("KODI_BINADDON_PATH");
+
   std::string appPath = CUtil::GetHomePath();
   setenv("KODI_HOME", appPath.c_str(), 0);
 
@@ -969,6 +973,7 @@ bool CApplication::InitDirectoriesOSX()
   {
     // map our special drives
     CSpecialProtocol::SetXBMCBinPath(appPath);
+    CSpecialProtocol::SetXBMCAltBinAddonPath(binaddonAltDir);
     CSpecialProtocol::SetXBMCPath(appPath);
     #if defined(TARGET_DARWIN_IOS)
       std::string appName = CCompileInfo::GetAppName();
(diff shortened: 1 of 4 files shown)


tamland:
I think I've said this a few times now but to repeat: don't use the On* instance hooks. They are fundamentally broken.

What should happen to make pvr client updating work is simply removing the custom pvr-only behaviour it does and follow what everything else is doing. That means:
1. removing the 'running instance' nonsense
2. the thing that is responsible for running pvr clients should listen for addon manager changes; then destroy+reload to get the new addon.

ksooo:
@tamland fair enough, let me suggest to do this step-by-step. Especially for Krypton we need a pragmatic, low-risk fix for a serious bug. So, I want to merge this (and the master PR) as is for now to restore broken basic functionality. For Leia, in the next step, we will see what we can do to get it "right".

BTW: I'm new to all the addon managenent stuff, never touched this code before, just had the feeling that if not I fix it nobody will. If you told several times, that the current implementation is "nonsense", this message did not reach me so far. I neither did that implementation nor did I ever care so far.

tamland:
@ksooo Make sense. For krypton at least anything goes of course. Just mentioning this because I've seen there's bee several patches to pvr installing (And the reason for that is imo because it does it's own weird things that nothing else does). This wasn't directed at you specifically in any way ksooo. I've just been talking about these hooks and the 'running instance' it in general too many times;) And I know nobody have touched this code for many years (me included!).

ksooo:
@tamland all good. :-)

ksooo:
jenkins build this please

PR #11522, opened 2017-01-22
[win32][depends]Patch TagLib to v1.11.1 + d2e0e55 to fix regression causing hangs

DaveTBlake:
For win32 update TagLib to v1.11.1 with  https://github.com/taglib/taglib/commit/d2e0e5522308ce4c905c205c7bf5972ea71cd6e5 as a patch to fix regression causing hangs with (Shoutcast) internet radio streams. See ticket http://trac.kodi.tv/ticket/17210 .
(Shoutcast) internet radio streams ending ".mp3" get passed to TabLib as if they have ID3 tags, and TabLib was spending upto 10 mins looking for them, looking for the end of the file incase the tags were there. 

The step to v1.11.1 brings win32  in line with the package changes for other platforms made by #11101
and the patch to match  #11543 

This change has already been backported to v17 by #11547 

Note for future update to TagLib - it is worth testing with Shoutcasts, as well as other music file formats.

PR #11543, opened 2017-01-24
[depends] patch taglib: Efficient lookup for an ID3v2 tag in MPEG fil…

Rechi:
…es with garbage.

## Description
add https://github.com/taglib/taglib/commit/d2e0e5522308ce4c905c205c7bf5972ea71cd6e5 as a patch

## Types of change

## Checklist:

### DIFF
--- a/tools/depends/target/taglib/ID3v2Lookup.patch
+++ b/tools/depends/target/taglib/ID3v2Lookup.patch
@@ -1,0 +1,72 @@
+From d2e0e5522308ce4c905c205c7bf5972ea71cd6e5 Mon Sep 17 00:00:00 2001
+From: Tsuda Kageyu <tsuda.kageyu@gmail.com>
+Date: Fri, 20 Jan 2017 21:14:38 +0900
+Subject: [PATCH] Efficient lookup for an ID3v2 tag in MPEG files with garbage.
+
+This looks for an ID3v2 tag until reaching the first valid MPEG frame in O(n) time.
+---
+ taglib/mpeg/mpegfile.cpp | 45 +++++++++++++++++++++++++++++----------------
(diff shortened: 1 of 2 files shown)


DaveTBlake:
Many thanks for this.
Does something else need to be done for win32? And either way is it not best to apply #11522 the update v.1.11 on win32 first?

popcornmix:
> I still think it crazy to even try scanning ID3 tags from a Shoutcast, but always good to have a TagLib that can cope.

But is there evidence this only affects Shoutcast?
It sounds like any mp3 file without ID3 tags would provoke the bug.
It seems likely that any live streaming site that uses mp3 format wouldn't have ID3 tags, so a generic fix seems much better than a Shoutcast specific hack.

Rechi:
@DaveTBlake this doesn't affect windows, it's for all platforms expect Windows.
I think @Paxxi will do something similar for Windows.

DaveTBlake:
Good point @popcornmix. The absense of tags from an mp3 file is fine (as best I can test). The issue here is giving a continuous stream of untagged data to something that is looking for ID3 tags anywhere in a file - it takes a long time to  stop looking! Shoutcasts are the only example of this I have found, but I guess there could be other continuous streams that end ".mp3".

Yes there has been a regerssion in TagLib, they are fixing it, and good for us to use that fix. But this bug  has revealed that we are treating their utility in an incorrect way. I think we should take more care over what we throw at a tag scanner. Calling a routine designed to scan ID3 tags from a finite file with a stream of untagged data is wrong, it is beyond its scope.

MartijnKaijser:
jenkins build this please

DaveTBlake:
Same patch implemented for win32 by  #11547.  Thanks all.

PR #11545, opened 2017-01-25
Update movie and TV library nodes

Hitcher:
## Description
Re-order and re-label the nodes so 'Recently added' and 'In progress'
are at the top and drop 'movies' and 'TV shows' from the label.

## Motivation and Context
Puts the most used items at the forefront.

## How Has This Been Tested?
Tested with Confluence and Estuary.

screenshot000
screenshot001
screenshot002
screenshot003

## Types of change

## Checklist:

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -12638,7 +12638,10 @@ msgctxt "#20381"
 msgid "Specials"
 msgstr ""
 
-#empty string with id 20382
+#: xbmc/filesystem/VideoDatabaseDirectory/DirectoryNodeOverview.cpp
+msgctxt "#20382"
+msgid "Recently added"
+msgstr ""
 
 msgctxt "#20383"
 msgid "Selected folder contains a single video"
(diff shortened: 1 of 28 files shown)


HitcherUK:
@ronie @phil65 @BigNoid @jjd-uk

ronie:
looks good to me.

i'd say make the same changes to the musicvideos node as well.

jjd-uk:
Looks good to me too.

da-anda:
I like the label change, but on Estuary having the "recently added" node right above the widget really seems a bit odd.

HitcherUK:
> could we probably get rid of the old labels or simply rename them instead of adding a new label?

Keeping them would be better as skinners can still use them _and_ have them translated.

da-anda:
sure, but skinners could also use the new label, or in case we simply change the old labels, keep using them. I'd just like to avoid to keep our labels list growing and growing without proof that skins actually use these. I'm pretty sure we already have dead labels. That's why I mentioned it.

HitcherUK:
Can someone please merge? Thanks.

ronie:
> i'd say make the same changes to the musicvideos node as well.

no interest ?

HitcherUK:
Forgot, sorry. Will look at the weekend.

HitcherUK:
@ronie Added music nodes

ronie:
thx!

PR #11546, opened 2017-01-25
ADD: "Clean thumbnails" functionality

razzeee:
## Description

Picked from spmc https://github.com/koying/SPMC/commit/3f255975618e697d018d74feb2e19e9624c72be2#diff-92b91a5c902936f6e9a03f26b054d7d6

Haven't really looked at the code yet.

I made some small string changes and moved it into a category that makes more sense, I think.

## How Has This Been Tested?

Compiles and seem to work fine, I just don't have any thumbs on my dev pc right now. But I think it's in spmc for a while now already.

image

## Types of change

## Checklist:

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -4619,7 +4619,12 @@ msgctxt "#1450"
 msgid "Put display to sleep when idle"
 msgstr ""
 
-#empty strings from id 1451 to 2049
+#: system/settings/settings.xml
+msgctxt "#1451"
+msgid "Clean thumbnails"
+msgstr ""
+
+#empty strings from id 1452 to 2049
 #strings through to 1470 reserved for power-saving
 
 msgctxt "#2050"
@@ -20037,7 +20042,12 @@ msgctxt "#37045"
 msgid "Extract chapter thumbnails for presentation in the chapters / bookmarks dialogue. This might increase CPU load."
 msgstr ""
 
-#empty strings from id 37046 to 38009
+#: system/settings/settings.xml
+msgctxt "#37046"
+msgid "Remove dangling thumbnails."
+msgstr ""
+
+#empty strings from id 37047 to 38009
 
 #: system/settings/rbp.xml
 msgctxt "#38010"
--- a/xbmc/settings/Settings.h
+++ b/xbmc/settings/Settings.h
@@ -377,6 +377,7 @@ public:
   static const std::string SETTING_ADDONS_ALLOW_UNKNOWN_SOURCES;
   static const std::string SETTING_GENERAL_ADDONFOREIGNFILTER;
   static const std::string SETTING_GENERAL_ADDONBROKENFILTER;
+  static const std::string SETTING_THUMBNAILS_CLEANUP;
   static const std::string SETTING_SOURCE_VIDEOS;
   static const std::string SETTING_SOURCE_MUSIC;
   static const std::string SETTING_SOURCE_PICTURES;
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -863,6 +863,10 @@
           <level>2</level>
           <control type="button" format="action" />
         </setting>
+        <setting id="thumbnails.cleanup" type="action" label="1451" help="37046">
+          <level>2</level>
+          <control type="button" format="action" />
+        </setting>
         <setting id="videolibrary.export" type="action" label="14248" help="36149">
           <level>2</level>
           <control type="button" format="action" />
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -5060,6 +5060,11 @@ void CApplication::StartVideoScan(const std::string &strDirectory, bool userInit
   CVideoLibraryQueue::GetInstance().ScanLibrary(strDirectory, scanAll, userInitiated);
 }
 
+void CApplication::StartThumbnailsCleanup(bool userInitiated /* = true */)
+{
+  CTextureCache::GetInstance().AddJob(new CTextureCleanupJob());
+}
+
 void CApplication::StartMusicCleanup(bool userInitiated /* = true */)
 {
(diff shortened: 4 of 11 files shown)


addons/resource.language.en_gb/resources/strings.po
```diff
@@ -20419,3 +20429,4 @@ msgstr ""
 msgctxt "#38208"
 msgid "If EXIF information exists (date, time, camera used, etc.), it will be displayed."
 msgstr ""
+
```
popcornmix: Remove this

xbmc/settings/Settings.h
```diff
@@ -393,6 +394,7 @@ class CSettings : public CSettingCreator, public CSettingControlCreator
   static const std::string SETTING_GAMES_ENABLEREWIND;
   static const std::string SETTING_GAMES_REWINDTIME;
 
+
```
popcornmix: Remove this

MilhouseVH:
If I've read the code correctly, this is removing all thumbnails from `special://thumbnails` that have a row in TexturesXX.db, but it doesn't appear to be removing the corresponding row from the database once each file is deleted. I think that will cause problems when Kodi tries to display a thumbnail it believes to be already cached (as there's a row in the db) but the thumbnail file no longer exists in `special://thumbnails`. I guess you could just add `DELETE * from texture` once all the files have been deleted.

Removing all files from the texture cache is also a pretty extreme option/solution... it's quite unlike the "Clean library" option which only removes media items that no longer exist - this option is more like "Remove all thumbnails", and perhaps shouldn't be confused with the existing "Clean" functions.

A "Clean thumbnails" option that "prunes" the texture cache might be more appropriate/beneficial (ie. removing only those thumbnails that aren't referenced in the media library) however this is likely to be a much more involved/lengthy solution.

MilhouseVH:
On second reading (of the code) I think it is only meant to delete the thumbnail files that don't exist in the Textures database, so in that sense it is a valid "clean" option. Although as very few processes actually remove rows from the Textures database - I don't think "Clean library" does - I'm not sure how many thumbnails will actually be removed by this.

da-anda:
do we really need a setting for this? Shouldn't Kodi rather do is homework automatically? Like all X weeks when idle?

Razzeee:
Well ideally we shouldn't even have to have a job.

In the best case (local library) whatever removes the movie/item should remove the thumb.
It becomes tricky when we're talking thumbs for plugins that grab stuff from the net (not sure if these also get cached here). Those would probably benefit from something like a lifetime and something like a last seen.

arnova:
Well, at least it's a good start. We don't do this for music/video libraries as well, so the same thing could be said for those. Having this would be an improvement over what we already have.

MartijnKaijser:
Yes plugins images also get cached. The thumbnails table has a " last used" column which you could use for the lifetime

arnova:
I know when JM implemented this, it was planned to use the last used column at some point, but this never happened (probably because JM left before he could implement it).

arnova:
Btw. note that this is used for all art, not just video/music library items so only cleaning the ones related to those isn't enough anyway.

arnova:
Btw. note that this is used for all art, not just video/music library items so only cleaning the ones related to those isn't enough anyway.

MilhouseVH:
As this will only clean thumbnails that don't have a row in the TexturesXX.db the likelihood is that this will remove/clean very few - if any - thumbnails.

Unless the rows are removed from TexturesXX.db (for example, during a clean library, or when uninstalling an add-on) this PR will unfortunately have very little impact (and may lead to false expectations among users, although the help text is relatively clear). However if Kodi is removing the rows from the database it might as well remove the files also, again leaving little for this PR to do.

And even with this PR, and Kodi removing rows and files from the texture cache during a clean library etc., it still leaves the problem of centralised metadata: using a shared database, you run a clean library on clientA which removes the thumbnail row from it's local TexturesXX.db (and might even remove the corresponding thumbnail from `special://thumbnails`). Next, you run "Clean thumbnails" on clientB and it will remove nothing as all the thumbnail rows of the just cleaned movies are still present in the local TexturesXX.db on this client - the only place these thumbnails are no longer present (as far as this client is concerned) is in the shared database (ie. artwork urls).

So while this PR looks fine for deleting "dangling" thumbnails that's all it does and I'm not sure this is really a problem worth solving with a new GUI setting (it will cause confusion, while most users should have hardly any dangling thumbnails, and those they have are unlikely to be a problem other than a minimal amount of bloat). If dangling thumbnails is a problem worth solving then we should be identifying the cause of the dangling thumbnails and fixing that...

By the way, my [texturecache.py](https://github.com/MilhouseVH/texturecache.py) script can identify dangling thumbnails with: `texturecache.py r` - I have 9 such files out of 23619, and all but 1 are 0 bytes in size (the other is 5818 bytes). `texturecache.py R` can be used to remove these dangling files, which is the equivalent of this PR.

There's also `texturecache.py X` which will identify rows in the TexturesXX.db database that point at non-existent thumbnail files (`texturecache.py Xd` to remove these rows).

Finally, `texturecache.py p` will list thumbnails that do not exist in the media library (`texturecache.py P` to "prune" and remove these entries). These cached artworks suitable for pruning are often temporary in nature (plugin/add-on related, eg. youtube video thumbnail previews etc.) or related to removed media (movies, tv shows/seasons/episodes, albums/artists/songs, add-ons), and can result in significant bloat if not removed. I've had reports of this removing up to 2GB of unnecessary thumbnail files.

Razzeee:
Good reasoning thank you :)

MartijnKaijser:
I have zero interest in some python file that cleans stuff up. This should be a core feature and nothing else

MilhouseVH:
> I have zero interest in some python file that cleans stuff up.

So don't use it - it exists because there is nothing better and I only mentioned it to highlight the kind of functionality that may be useful.

> This should be a core feature and nothing else. So it must be implemented in core either automatic or on clean library or whatever.

Looking forward to your PR.

PR #11547, opened 2017-01-25
[fix][win32][depends] Patch taglib to fix regression causing hangs

Paxxi:
should fix #11530 without any complications

## Types of change

## Checklist:

### DIFF
--- a/project/BuildDependencies/scripts/0_package.list
+++ b/project/BuildDependencies/scripts/0_package.list
@@ -41,7 +41,7 @@ python-2.7.11-win32-vc140-v2.7z
 shairplay-0.9.0-win32-vc140-v2.7z
 sqlite-3.10.2-win32-vc140.7z
 swig-3.0.10-win32.7z
-taglib-1.11-win32-vc140-v2.7z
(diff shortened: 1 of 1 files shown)


arnova:
Are you sure this is a 100% fix? I believe I previously reproduced the issue on Linux with 1.11.1

MartijnKaijser:
It's a patched version with the fix on top

Paxxi:
Yep, that's why the weird name to mean it's 11.1 with that extra commit

arnova:
@Paxxi : Ok, missed that, see it now :)

DaveTBlake:
@Paxxi after testing I can confirm this patch fixes regerssion http://trac.kodi.tv/ticket/17210, and tag scanning of ordinary mp3 files generally seems to be fine as before. Thanks.

Rechi:
shouldn't this also be applied to master?

DaveTBlake:
Pending #11522 (win32 to 1.11.1) could be amended to do that

MartijnKaijser:
Please do so (or create new one)

PR #11550, opened 2017-01-26
[Win32] D3DResource improvements.

afedchin:
- [x] re-factor ID3DResource to make possible tell to a resource that device is not valid anymore.
- [x] made a bit more safest OverlayRendererDX
- [x] reload a skin after a driver failure.

## Description
The most of DirectX resources like textures and buffers try to receive and save their status from current device before destroying device completely. A driver failure makes this approach impossible because a driver doesn't service device anymore that may cause a segfault inside a bad designed driver (you know). This fix implements a possibility to tell a resource to do not use device that makes a resurrection faster and safest.

Also after a driver failure Kodi's UI become unusable till skin reload. This adds reloading a skin after a driver failure.

## How Has This Been Tested?
It was tested localy on unstable driver.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/DXVA.h
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/DXVA.h
@@ -147,7 +147,7 @@ protected:
 
   // ID3DResource overrides
   void OnCreateDevice() override  {}
-  void OnDestroyDevice() override { CSingleLock lock(m_section); m_state = DXVA_LOST;  m_event.Reset(); }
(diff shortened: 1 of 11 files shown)


afedchin:
jenkins build this please

Paxxi:
Can't comment on functionality but code looks sensible to me.

MartijnKaijser:
Is this something you'd say is "needed" as well for v17 (or perhaps delay till v17.1)?

afedchin:
Yes, it's needed. But I’m not sure is there a window before release?

MartijnKaijser:
@afedchin RC4 is up for this weekend so we can still get it in and tested by the masses

afedchin:
ok. will prepare a PR

PR #11554, opened 2017-01-27
[pvr] refactor guide osd to be channel guide dialog and use it in guide info

xhaggi:
This PR refactor the PVR dialog guide OSD to supports listing of all EPG items for a supplied channel instead of just the current playing channel in fullscreen mode. Therefore the dialog is renamed to `GUIDialogPVRChannelGuide`.

I also added a new button _Channel guide_ to the guide info dialog from where you can open the full EPG list for this channel.

screenshot003

@ksooo mind taking a look. I hope you remember that we talked about it some times ago 😄 
@phil65 or @ronie It would be nice if you could adjust the skins.


addons/skin.estuary/xml/DialogPVRChannelGuide.xml
```diff
@@ -4,7 +4,9 @@
 	<controls>
 		<control type="group">
 			<depth>DepthOSD</depth>
+			<!--
 			<animation effect="fade" start="100" end="0" time="200" tween="sine" condition="$EXP[infodialog_active]">Conditional</animation>
+			-->
```
ksooo: I suggest to remove the code instead of only commenting it.

xbmc/pvr/PVRGUIActions.h
```diff
@@ -68,6 +68,13 @@ namespace PVR
     bool ShowEPGInfo(const CFileItemPtr &item) const;
 
     /*!
+     * @brief Open a dialog with the epg list for a given item.
+     * @param item containing the channel. item must a channel.
```
ksooo: "must a " -> "must be a"

EDIT: despite from this, item can be either a channel, an epg tag or a timer. CPVRItem::GetChannel() takes care of the conversion. You are actually feeding this method with an epg tag from CGUIDialogPVRGuideInfo.

xbmc/pvr/dialogs/GUIDialogPVRChannelGuide.h
```diff
@@ -46,5 +47,8 @@ namespace PVR
 
     CFileItemList    *m_vecItems;
```
ksooo: While you are at refactoring this, would it make sense to use a std::unique_ptr instead of a raw pointer here?

xbmc/pvr/dialogs/GUIDialogPVRChannelGuide.cpp
```diff
@@ -70,23 +70,33 @@ bool CGUIDialogPVRGuideOSD::OnMessage(CGUIMessage& message)
...
   g_graphicsContext.Lock();
   m_viewControl.SetCurrentView(DEFAULT_VIEW_LIST);
 
   // empty the list ready for population
   Clear();
 
-  g_PVRManager.GetCurrentEpg(*m_vecItems);
+  // no user-specific channel set use current playing channel
+  if (!m_channel)
+    m_channel = g_PVRManager.GetCurrentChannel();
```
ksooo: I know it was like this before, but it is no good idea to call into g_PVRManager with locked g_graphicsContext. May I suggest to move the two lines above to the very beginning of this method, before locking the global graphics mutex?

xbmc/pvr/dialogs/GUIDialogPVRChannelGuide.cpp
```diff
@@ -108,41 +118,42 @@ void CGUIDialogPVRGuideOSD::OnInitWindow()
   m_viewControl.SetSelectedItem(iSelectedItem);
 }
 
-void CGUIDialogPVRGuideOSD::OnDeinitWindow(int nextWindowID)
+void CGUIDialogPVRChannelGuide::OnDeinitWindow(int nextWindowID)
 {
   CGUIDialog::OnDeinitWindow(nextWindowID);
+  m_channel = nullptr;
```
ksooo: m_channel.reset() is more efficient.

xbmc/pvr/dialogs/GUIDialogPVRGuideInfo.cpp
```diff
@@ -72,6 +73,26 @@ bool CGUIDialogPVRGuideInfo::OnClickButtonOK(CGUIMessage &message)
...
+bool CGUIDialogPVRGuideInfo::OnClickButtonChannelGuide(CGUIMessage &message)
+{
+  bool bReturn = false;
+
+  if (message.GetSenderId() == CONTROL_BTN_CHANNEL_GUIDE)
+  {
+    if (!m_progItem || !m_progItem->HasPVRChannel())
+    {
+      /* invalid channel */
+      CGUIDialogOK::ShowAndGetInput(CVariant{19033}, CVariant{19067});
```
ksooo: Could you please add the text for the two resource identifiers as a comment, like we usually do?

ksooo (changes requested):
Nice one, like it. Despite from my minor nitpicking it is fine already.

addons/skin.estuary/xml/DialogPVRChannelGuide.xml
xhaggi: it's part of the skip commit. I don't want to merge it as it is, because it needs a bit more love from @ronie or @phil65

xbmc/pvr/PVRGUIActions.h
xhaggi: will adjust the doc

xbmc/pvr/dialogs/GUIDialogPVRChannelGuide.h
xhaggi: sure could be done in a separate commit

xbmc/pvr/dialogs/GUIDialogPVRChannelGuide.cpp
xhaggi: ok, will move it up.

xbmc/input/ButtonTranslator.cpp
```diff
@@ -324,7 +324,7 @@ static const ActionMapping windows[] =
     { "pvrchannelscan"           , WINDOW_DIALOG_PVR_CHANNEL_SCAN },
     { "pvrupdateprogress"        , WINDOW_DIALOG_PVR_UPDATE_PROGRESS },
     { "pvrosdchannels"           , WINDOW_DIALOG_PVR_OSD_CHANNELS },
-    { "pvrosdguide"              , WINDOW_DIALOG_PVR_OSD_GUIDE },
```
da-anda: I think it would be nice if we'd keep this ID in addition to the new one for backwards compat reasons to not kill all custom user keymaps. We could add a comment to remove it in like 2 versions from now.

xbmc/pvr/dialogs/GUIDialogPVRChannelGuide.cpp
xhaggi: it does the same IIRC, but sure i can use `reset()` instead

xbmc/pvr/dialogs/GUIDialogPVRGuideInfo.cpp
xhaggi: c/p from another method. will change it.

xbmc/input/ButtonTranslator.cpp
xhaggi: sure, will re-add the mapping for backward compat

phil65:
Cool. I'm still on holidays, so give me few more days.

xhaggi:
addressed your comments.

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -11261,7 +11261,11 @@ msgctxt "#19685"
 msgid "Confirm shutdown"
 msgstr ""
 
-#empty string with id 19686
+#: Label for buttons to open channel guide dialog
+#: addons/skin.estuary/1080i/DialogPVRInfo.xml
```
ksooo: “/1080i/“ => “/xml/“

ksooo (approved):
just found another small glitch...

addons/resource.language.en_gb/resources/strings.po
xhaggi: There are tons of other places references 1080i, we should change it at all.

ksooo: Right, but what speaks against doing it right in the first place for new stuff we're adding? ;-)

xhaggi: Nothing

ksooo: :-)

IchabodFletchman:
I've made the changes to Estuary including the new icon. @phil65 can cherry pick it once the PR is merged

xhaggi:
@IchabodFletchman thanks for your contribution but we should think about changing the dialog to fit both, being an OSD dialog and a normal dialog laying on top of other dialogs. IMO the dialog looks a bit odd on top of the guide info dialog.

xhaggi:
@phil65  do have some spare time to do the necessary skin changes?

xhaggi:
I've dropped the skin changes for now because we can add it later.

@phil65 if you want to adapt something 8cf27edc39ad05da07be85497344de4b0ff2f41f or IchabodFletchman/skin.estuary@58333df

xhaggi:
@ksooo I'll merge this after the build is fine.

xhaggi:
jenkins build and merge

xhaggi:
jenkins build and merge

Leatherface75:
Would be nice with IMDB support too there.

PR #11557, opened 2017-01-28
Forced full tag rescans of music files

DaveTBlake:
Improved management of the forced full tag rescans of music files. This addresses two issues:
1) When upgrading Kodi version for a client/server system, storing the need for a forced rescan in the GUI settings means that every client attempts to force a rescan, while actually the db on the server only needs the tags rescanning once.

2) The way the music database is populated from music file tags depends upon system settings. The user can change these, but has no way to apply them. On library update, or scan to library of a source from file view, the tags are only rescanned using the new settings, if the music files have been changed. So this means either dropping the source, and thus loosing all the playcount and rating data as well as needing to rescrape, or having to make a timestamp change to all your music files.

To those ends this PR 
a) Adds a versiontagscan table to track the version of tag processing used to populate the library (effectively the db version at the time the tags were scanned). This replaces the  use of the MusicNeedsUpdate GUI setting with read/write to database.

b) Adds an advancedsetting to optionally prompt for full tag rescan, regardless of files being unchanged, every time scanning is initiated i.e. from "Update Library" and when "Scan to library" from file view

To test attempt to upgrade from earlier versions of Kodi - from v17 no forced rescan is necessary, from v16 or before one is. Try an upgrade on a client/server setup.

Subsequently add `<promptfulltagscan>true</promptfulltagscan>` to the  `<musiclibrary>` section of advancesettings.xml, and on library update you will be given the option of a full rescan regardless of files being unchanged.

### DIFF
--- a/xbmc/music/MusicDatabase.cpp
+++ b/xbmc/music/MusicDatabase.cpp
@@ -192,6 +192,10 @@ void CMusicDatabase::CreateTables()
 
   CLog::Log(LOGINFO, "create cue table");
   m_pDS->exec("CREATE TABLE cue (idPath integer, strFileName text, strCuesheet text)");
+
+  CLog::Log(LOGINFO, "create versiontagscan table");
+  m_pDS->exec("CREATE TABLE versiontagscan (idVersion integer, iNeedsScan integer)");
+  m_pDS->exec(PrepareSQL("INSERT INTO versiontagscan (idVersion, iNeedsScan) values(%i, 0)", GetSchemaVersion()));
 }
 
 void CMusicDatabase::CreateAnalytics()
@@ -4662,8 +4666,6 @@ void CMusicDatabase::UpdateTables(int version)
   if (version < 53)
   {
     m_pDS->exec("ALTER TABLE song ADD dateAdded text");
-    CMediaSettings::GetInstance().SetMusicNeedsUpdate(53);
-    CServiceBroker::GetSettings().Save();
   }
   if (version < 54)
   {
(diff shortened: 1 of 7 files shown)


FernetMenta:
why don't you create a gui settings page for all the music library related settings? this is imo much better than adding another advanced settings.

DaveTBlake:
A music library GUI setting page already exists, but does not include any settings relating to scanning tags. The tag scanning options are historically in advancedsettings.xml, I guess because some are specialised but some are also easier to edit in a file e.g. list of  file extensions.

From a development view point it is much easier to add a facility initially via advancedsettings.xml, than start messing with the GUI, so that is what I have done here **as an initial step** at least. Given the difficulty in getting peer review for music changes I am trying to break any changes into stages. 

I would prefer moving options from advancedsettings.xml to settings GUI as a separate PR

FernetMenta:
> I would prefer moving options from advancedsettings.xml to settings GUI as a separate PR

Fine, but please don't forget it :)

> From a development view point it is much easier to add a facility initially via advancedsettings.xml, than start messing with the GUI

Sure, this is why we have this mess. In general devs never come the the second part.

DaveTBlake:
No worries, I will set myself a settings review task for the end of a number of music improvements I have in mind, best setting done coherently all together.

xbmc/music/MusicDatabase.cpp
```diff
@@ -4958,17 +4960,78 @@ void CMusicDatabase::UpdateTables(int version)
...
-    CServiceBroker::GetSettings().Save();
   }
+  if (version < 61)
+  {
+    // Create versiontagscan table
+    m_pDS->exec("CREATE TABLE versiontagscan (idVersion integer, iNeedsScan integer)");
+    m_pDS->exec("INSERT INTO versiontagscan (idVersion, iNeedsScan) values(0, 0)");
+  }
+
+  // Set the verion of tag scanning required.
```
da-anda: typo `verion`

xbmc/music/MusicDatabase.cpp
```diff
@@ -4958,17 +4960,78 @@ void CMusicDatabase::UpdateTables(int version)
...
+{
+  m_pDS->exec(PrepareSQL("UPDATE versiontagscan SET iNeedsScan=%i", version));
+}
+
+void CMusicDatabase::SetMusicTagScanVersion(int version /* = 0 */)
+{
+  if (version == 0)
+    m_pDS->exec(PrepareSQL("UPDATE versiontagscan SET idVersion=%i", GetSchemaVersion()));
+  else
+    m_pDS->exec(PrepareSQL("UPDATE versiontagscan SET idVersion=%i", version));
```
da-anda: I'm not really into our DB code, but doesn't a UPDATE query usually also require a WHERE part to specify which row to update?

da-anda:
if I got this right, this new table will only contain one row for the pure sake of tracking if tags need to be rescanned for a specific DB version? If so, I'm against it. The usage of MySQL is not recommended in any way, and adding a workaround for this particular usage scenario is IMO the wrong way. People that use MySQL know that there are pitfalls with it and also how to mess with Kodi settings files and are for sure also capable of setting a certain settings value in GUI settings if required. I know that I don't really have to say anything in regards to our codebase, but I really don't like this approach and I'm sure there is a better solution to this, or at least a more generic one that can be reused for other multiclient settings that would make sense.

xbmc/music/MusicDatabase.cpp
DaveTBlake: This table only has one row, so no WHERE is needed

DaveTBlake:
@da-anda this isn't a "MySQL work around", but from a pure design viewpoint the version of processing used to convert music file tag into library data belongs to the library, with the data, not off in the both user  and client specific GUIsettings.xml files

It is much like the db version itself, which is also stored in a single rowed table, and there is nothing wrong with that.

Montellese:
Is a single row enough in case of multiple profiles using the same library but with restrictions on certain profiles? In that case it might happen that not all songs/tags are rescanned. But since our profile support sucks anyway this is a corner case.

DaveTBlake:
The first commit/part of the this PR really is a simple attempt to preserve the originating db version after an upgrade of Kodi on an existing installation, so that it can be used to make decisions such as initiating rescanning etc. while  Kodi running. Using GUIsettings.xml files as currently is totally inapproriate.

Profile support does not come into this, certainly with the way profiles are now.

da-anda:
sorry for the little overreaction above. All sorted out on Slack. Thanks for your patience.

DaveTBlake:
Discussed fully on Slack, think we are good with this. I intend to merge early next week unless any one objects, so that I can submit my next music PR without any confusion.

jenkins build this please

PR #11559, opened 2017-01-28
VideoPlayer: fix audio stream not started

FernetMenta:
fixes in particular some DVD menus where an audio stream gets started while video is already running


FernetMenta:
jenkins build this please

b-pass:
This seems to have broken Edl commercial skipping for me in Krypton HEAD.  When it goes to do an Edl skip (VideoPlayer.cpp Line ~2421) video and audio both hang.  Doing a manual seek forward or backward fixes it.

I confirmed it was this commit since this is a one line change, commenting it out fixes the problem for me.

I tried two different videos and several different Edl commercials in each one and they all had this problem, I don't think it is specific to my recordings or my EDLs.

FernetMenta:
I see the issue which is a design flaw of EDL skipping. Instead of looking the the actual playing time it looks at the other end of the demuxer queue. As a result a cumbersome delayed flushing process was implemented that blocks the player thread for up to 8 seconds.
I rarely touched edl so far but now its time to do some rework.

PR #11565, opened 2017-01-29
[cmake] don't abort building binary addons if an addon cannot be downloaded

Rechi:
## Description
Don't abort building of binary addons if an addon cannot be dowloaded.
Instead mark this addon as failed and continue.

## Motivation and Context
Jenkins had no binary addons in the last days.

## Types of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [x] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #11566, opened 2017-01-29
[cmake] don't abort building binary addons if an addon cannot be downloaded

Rechi:
backport of #11565


MartijnKaijser:
Jenkins build this please

wsnipex (commented):
Looks good. Adding a retry for failed downloads would be great as well

PR #11570, opened 2017-01-29
[EPG] Refactor and complete implementation of EpgSearchFilter

ksooo:
There are reports in the forum that some controls in our EPG search dialog seem to have no affect. 

http://forum.kodi.tv/showthread.php?tid=304221
http://forum.kodi.tv/showthread.php?tid=298556

That's right. :-) According to the commit history it seems that the code that is responsible for that, <code>EpgSearchFilter</code>, simply never got implemented completely.

screenshot000

This PR adds the missing code pieces to <code>EpgSearchFilter</code> and I took the chance to refactor it.
 
The change has been runtime-tested on latest master on macOS.

@Jalle19 for the review?

### DIFF
--- a/xbmc/epg/EpgSearchFilter.h
+++ b/xbmc/epg/EpgSearchFilter.h
@@ -31,30 +31,94 @@ namespace EPG
 
   /** Filter to apply with on a CEpgInfoTag */
 
-  struct EpgSearchFilter
+  class CEpgSearchFilter
   {
+  public:
+    CEpgSearchFilter();
+
     /*!
      * @brief Clear this filter.
      */
-    virtual void Reset();
+    void Reset();
 
     /*!
      * @brief Check if a tag will be filtered or not.
      * @param tag The tag to check.
      * @return True if this tag matches the filter, false if not.
      */
-    virtual bool FilterEntry(const CEpgInfoTag &tag) const;
+    bool FilterEntry(const CEpgInfoTag &tag) const;
 
-    virtual bool MatchGenre(const CEpgInfoTag &tag) const;
-    virtual bool MatchDuration(const CEpgInfoTag &tag) const;
-    virtual bool MatchStartAndEndTimes(const CEpgInfoTag &tag) const;
-    virtual bool MatchSearchTerm(const CEpgInfoTag &tag) const;
-    virtual bool MatchChannelNumber(const CEpgInfoTag &tag) const;
-    virtual bool MatchChannelGroup(const CEpgInfoTag &tag) const;
-    virtual bool MatchBroadcastId(const CEpgInfoTag &tag) const;
-    virtual bool MatchChannelType(const CEpgInfoTag &tag) const;
-
+    /*!
+     * @brief remove duplicates from a list of epg tags.
+     * @param results the list of epg tags.
+     * @return the number of items in the list after removing duplicates.
+     */
     static int RemoveDuplicates(CFileItemList &results);
+
+    const std::string &GetSearchTerm() const { return m_strSearchTerm; }
+    void SetSearchTerm(const std::string &strSearchTerm) { m_strSearchTerm = strSearchTerm; }
+    void SetSearchPhrase(const std::string &strSearchPhrase);
+
+    bool IsCaseSensitive() const { return m_bIsCaseSensitive; }
+    void SetCaseSensitive(bool bIsCaseSensitive) { m_bIsCaseSensitive = bIsCaseSensitive; }
+
+    bool ShouldSearchInDescription() const { return m_bSearchInDescription; }
+    void SetSearchInDescription(bool bSearchInDescription) {m_bSearchInDescription = bSearchInDescription; }
+
+    int GetGenreType() const { return m_iGenreType; }
+    void SetGenreType(int iGenreType) { m_iGenreType = iGenreType; }
+
+    int GetGenreSubType() const { return m_iGenreSubType; }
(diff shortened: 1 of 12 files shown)


ksooo:
@Jalle19 the 'is' -> 'should' change you suggested is now in.

xbmc/epg/EpgSearchFilter.h
```diff
@@ -20,42 +20,107 @@
...
+    bool ShouldRemoveDuplicates() const { return m_bRemoveDuplicates; }
+    void SetRemoveDuplicates(bool bRemoveDuplicates) { m_bRemoveDuplicates = bRemoveDuplicates; }
+
+    bool IsRadio() const { return m_bIsRadio; }
+    void SetIsRadio(bool bIsRadio) { m_bIsRadio = bIsRadio; }
+
+    int GetChannelNumber() const { return m_iChannelNumber; }
+    void SetChannelNumber(int iChannelNumber) { m_iChannelNumber = iChannelNumber; }
+
+    bool IsFTAOnly() const { return m_bFTAOnly; }
```
Jalle19: Since you've used `MatchFreeToAir`, maybe these should be `FreeToAir` too? Including the member variable.

xbmc/epg/EpgSearchFilter.h
```diff
@@ -66,7 +131,7 @@ namespace EPG
     CDateTime     m_startDateTime;            /*!< The minimum start time for an entry */
     CDateTime     m_endDateTime;              /*!< The maximum end time for an entry */
     bool          m_bIncludeUnknownGenres;    /*!< Include unknown genres or not */
-    bool          m_bPreventRepeats;          /*!< True to remove repeating events, false if not */
+    bool          m_bRemoveDuplicates;          /*!< True to remove duplicate events, false if not */
```
Jalle19: White-space alignment in the comment

xbmc/pvr/windows/GUIWindowPVRSearch.cpp
```diff
@@ -75,23 +75,23 @@ bool CGUIWindowPVRSearch::FindSimilar(const CFileItemPtr &item)
 
   // construct the search term
   if (item->IsEPG())
-    m_searchfilter.m_strSearchTerm = "\"" + item->GetEPGInfoTag()->Title() + "\"";
+    m_searchfilter.SetSearchTerm("\"" + item->GetEPGInfoTag()->Title() + "\"");
```
Jalle19: Since all uses of this method pass the parameter wrapped inside `""`, maybe the setter itself should handle that. At least to me it is not at all obvious that the search term must be wrapped like this.

Jalle19 (changes requested):
Some minors only. Good job!

xbmc/epg/EpgSearchFilter.h
MartijnKaijser: Shouldn't we just stop with this alignment in general?

Jalle19: Maybe, but in that case the comments should be blocks above the line, not on the same line like this

xbmc/pvr/windows/GUIWindowPVRSearch.cpp
ksooo: TBH, I have no idea why this wrapping is there. According to commit history, @xhaggi introduced this lines of code. Maybe he can explain?

xbmc/epg/EpgSearchFilter.h
ksooo: will fix this.

xbmc/epg/EpgSearchFilter.h
ksooo: yep, will change this.

xbmc/pvr/windows/GUIWindowPVRSearch.cpp
ksooo: ... and there is one place where SetSearchTerm is called without that wrapping...

xhaggi: I have no clue why it should be necessary. This code was there before I refactored the PVR windows. The history says I've recovered that code not that I've introduced it.

ksooo: I guess it has something to with "match every single word" versus "match the whole string". Will take a look into the impementation of <code>CTextSearch</code> (which is used internally by EpgSearchFilter) and find out.

ksooo: Could be something with "match string as a whole" versus "match every single word of the string"... I will know later tonight.

Jalle19: Okay. What an interface :)

ksooo:
@Jalle19 all requested changes are in.

xbmc/pvr/windows/GUIWindowPVRSearch.cpp
xhaggi: Should we introduce a flag + method to control this behavior? Like `matchTerms` if a term base matching should be done.

Jalle19:
+1 from me, jenkins build this please

ksooo:
@Jalle19 added another commit with a bug fix.

Jalle19 (approved):

ksooo:
jenkins build this please

ksooo:
Jenkins build this please

ksooo:
jenkins build this please

PR #11571, opened 2017-01-29
Fix analog stick/trigger "sticking" bug

garbear:
Some triggers are "anomalous" in that their center isn't 0. For example, on OSX, triggers at rest are -1, and fully pressed are +1. The author of SDL has observed triggers that start at +1 and travel to 0 when fully pressed.

I noticed this issue in mid-2015 and "fixed" it by detecting these offset triggers at startup. However, this fix was at the driver level, and the driver (purposefully) doesn't have access to button maps, so it has to apply this filter to all triggers.

The problem occurs when axes are falsely detected as anomalous triggers. This causes analog sticks to appear "stuck" at 0.5 when centered. This rarely happens, but is extremely annoying when it does and can only be fixed by unplugging/replugging controllers.

This PR drastically improves the situation by only detecting anomalous triggers in the controller dialog when they're being mapped. The Peripheral API has been expanded to make button maps aware of the center and range of axes.

The corresponding peripheral.joystick PR is https://github.com/xbmc/peripheral.joystick/pull/80

The resulting button map with anomalous triggers looks like:

```xml
<?xml version="1.0" ?>
<buttonmap>
    <device name="Xbox 360 Wired Controller" provider="cocoa" vid="046D" pid="C21D" buttoncount="15" axiscount="6">
        <configuration>
            <axis index="4" center="-1" range="2" />
            <axis index="5" center="-1" range="2" />
        </configuration>
        <controller id="game.controller.default">
            <feature name="lefttrigger" axis="+4" />
            <feature name="righttrigger" axis="+5" />
            ...
        </controller>
    </device>
</buttonmap>
```

Also, the "Fix skipping" button has been renamed to "Ignore input" because this PR drastically fixes skipping and it barely occurs anymore.

ignore_input

## Types of change
- [x] Bug fix with breaking change

### DIFF
--- a/addons/kodi.peripheral/addon.xml
+++ b/addons/kodi.peripheral/addon.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<addon id="kodi.peripheral" version="1.2.1" provider-name="Team-Kodi">
-	<backwards-compatibility abi="1.2.0"/>
+<addon id="kodi.peripheral" version="1.3.0" provider-name="Team-Kodi">
+	<backwards-compatibility abi="1.3.0"/>
 	<requires>
 		<import addon="xbmc.core" version="0.1.0"/>
 	</requires>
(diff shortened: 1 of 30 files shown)


garbear:
Updated with some fixes and a string change - see screenshot in PR description

garbear:
does anyone have an opinion on what the "Fix skipping" or "Ignore input" button should be labeled?

The feature is described in the wiki: [HOW-TO:Fix_buttons_skipped_while_mapping](http://kodi.wiki/view/HOW-TO:Fix_buttons_skipped_while_mapping). The help text looks like:

buttons_were_ignored

garbear:
jenkins build this please

PR #11573, opened 2017-01-29
[OverlayRenderer] fix renderer after #11550

afedchin:
fix regression of #11550


afedchin:
jenkins build and merge

PR #11580, opened 2017-01-30
[lang] update skin path of estuary in language key usage docs

xhaggi:
Small search/replace of `1080i` with `xml` in the language key docs for our default language file.

@ronie or @MartijnKaijser mind taking a look.


ronie:
yup, should be fine.

xhaggi:
ok will be merge after build is succeeded.

MartijnKaijser:
Build doesn't care for string files so just merge

PR #11561, opened 2017-01-28
create archive_cache if not exists

ghost:
Its never was created though used for caching which fails nonexistent directory.

## Description

The archive_cache to cache addons/misic etc files but this directory doesn't exits on vanilla installation.
## Motivation and Context

To make sure that the directory exists and caching improves the speed.

## How Has This Been Tested?

Tested on windows, directory created at launch.
## Types of change

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ x] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [x ] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ x] All new and existing tests passed

PR #11489, opened 2017-01-19
[Fix]Clear plugin directory cache on window init Ticket #16560

DaveTBlake:
Backport of the second commit of #11478, see discussion there. It fixes ticket http://trac.kodi.tv/ticket/16560 the issue with TV shows etc. added to favorites (what #9898 and #10394 were attempting to solve).

PR #11582, opened 2017-01-30
Addon caching enabled

ghost:
Caching in Addon was enabled back in Krypton RC4 however this change is required for it to function properly.
## Description

Fix to bring back the caching in Addon as it was supported in previous version.
## Motivation and Context

This change is required for caching in Addon to work. it also required the PR #11561
## How Has This Been Tested?

Tested several addons (Youtube etc), play a video and after verified that it doesn't refresh the screen.
## Types of change

## Checklist:

### DIFF
--- a/xbmc/windows/GUIMediaWindow.cpp
+++ b/xbmc/windows/GUIMediaWindow.cpp
@@ -503,7 +503,10 @@ bool CGUIMediaWindow::OnMessage(CGUIMessage& message)
           resetHistory = true;
         }
         if (resetHistory)
+        {  
+          m_vecItems->RemoveDiscCache(GetID());
           SetHistoryForPath(m_vecItems->GetPath());
+        }
       }
       if (message.GetParam1() != WINDOW_INVALID)
       { // first time to this window - make sure we set the root path
--- a/xbmc/windows/GUIMediaWindow.cpp
+++ b/xbmc/windows/GUIMediaWindow.cpp
@@ -511,7 +511,7 @@ bool CGUIMediaWindow::OnMessage(CGUIMessage& message)
       }
       if (message.GetParam2() == PLUGIN_REFRESH_DELAY)
       {
-        Refresh(true);
+        Refresh();
         SetInitialVisibility();
         RestoreControlStates();
         SetInitialVisibility();


BigNoid:
This brings back the stale plugins if started from a shortcut. To me that is a bigger issue than the list refreshing after playback, which might even be a good thing as you get to see the latest content.

Shani-08:
Not really, most of the first pages in plugins are static anyways. For play issue, say if you open youtube addon and search something. When the play is finished or stopped, it does the search again.

BigNoid:
People also add shortcuts to specific plugin content. Also for search results the same argument goes; it may have been changed in the time the video has played, so it's not necessarily a bad thing.

DaveTBlake:
This reverts #11489 and so the issue with TV shows etc. added to favorites ticket http://trac.kodi.tv/ticket/16560 remains unsolved. Can we not find a way to fix that without breaking something else?

If this accepted we also need PRs for master to do this change, and the supporting addition of archive cache, there as well.

koying:
I came up with https://github.com/koying/SPMC/commit/5ffbd38a5af5aeebb83e55d981f6f4e1c965fffd after some debugging to find the relevant possible pathways to the GUIMediaWindow.

AFAICT, that clears the cache any time we enter with a non-returning path (e.g. when coming from Favourites), but not when returning (e.g. from a playback).

Shani-08:
Nice, i will test your change.

popcornmix:
@Shani-08 Koying's patch was for Jarvis. A rebased version should look like:
https://github.com/popcornmix/xbmc/commit/2306497f94b1775170c18069d89f993fe10869a3

Shani-08:
Yes, i have done exactly that and compiling/testing.

Shani-08:
@koying change covers the fav. issue as well as the Addon refresh from play.

BigNoid:
I still don't see the issue with reloading after playback. Imo that is the preferred way. I haven't noticed any difference with this PR in terms of speed of the lists that are presented. Could you point me to an addon where loading is so slow that it's preferred to have cached results instead of another query?

Shani-08:
Two example, use Vine addon from Kodi repo, When you search for something, the list is displayed but when you play something/stop it ask for search again. with the above change it maintains the results.
Use youtube, search and play something and it will search again. (more visible on the slow connections), however users dont want to keep jumping up and down as they would like to play through those videos in the list instead of it getting refreshed. They can always search again for latest list.

mortael:
Also, https://www.dropbox.com/s/oko56xx0lyfg1i0/plugin.video.refresh-example.zip?dl=0 shows an example of a possible addon :)

BigNoid:
> Also, https://www.dropbox.com/s/oko56xx0lyfg1i0/plugin.video.refresh-example.zip?dl=0 shows an example of a possible addon :)

lol, yeah waiting half a second for adding 1 result to the list is hardly a real life situation. In my testings almost all results are fetched within that timeframe.



> Two example, use Vine addon from Kodi repo, When you search for something, the list is displayed but when you play something/stop it ask for search again. with the above change it maintains the results.
> Use youtube, search and play something and it will search again. (more visible on the slow connections), however users dont want to keep jumping up and down as they would like to play through those videos in the list instead of it getting refreshed. They can always search again for latest list.

so it's basically a problem with that addon then. When I search in YouTube the results are kept when stopping playback and I return to the same item on the list.

Shani-08:
When  you search in Youtube, the result is refreshed however it keeps the last played item selected. But due to refresh the listing could be changed or moved all the time or your played item might not be there anymore if you played a longer video. The point is, this bring a lot of issues in existing addons while the fix works for most of the issues including the launch from fav.

BigNoid:
I agree that keeping the addons compatible is preferable at this point.

BigNoid:
jenkins build this please

MartijnKaijser:
@Shani-08 can you please also create a PR to master branch?

Shani-08:
sure but tomorrow. Thanks.

PR #11583, opened 2017-01-30
[cmake] [linux] [backport] Don't compile/add WiiRemote if libcwiid-dev is not present

nsenica:
## Description
WiiRemote will fail if libcwiid-dev is not present in the system, so don't add it to the build.

Topic: http://forum.kodi.tv/showthread.php?tid=305169 

## How Has This Been Tested?
Building kodi

## Types of change

## Checklist:


nsenica:
Generation of wiiremote deb via cpack is also affected by the same problem, I'll provide fixes in a separate PR later on.

PR #11585, opened 2017-01-30
[cmake] Tidy up ffmpeg in order to be more flexible and versatile. Also includes switches for Raspberry Pi.

nsenica:
## Description
ffmpeg options were not totally correct, in my point of view, some settings like compiler flags are useful even when not cross compiling. Also there were choices made based on the CPU which was not passed on to the project thus being useless. Also Raspberry Pi didn't have any switches on its own, which are also needed.
I believe this provides more flexibility and the output is more inline with kodi's configuration.

This is my first approach to the problem, so feedback is welcome :)

## Motivation and Context
See above.

## How Has This Been Tested?
By building kodi.

## Types of change

## Checklist:

### DIFF
--- a/tools/depends/target/ffmpeg/CMakeLists.txt
+++ b/tools/depends/target/ffmpeg/CMakeLists.txt
@@ -57,15 +57,10 @@ elseif(CORE_SYSTEM_NAME STREQUAL osx)
                           --target-os=darwin
                           --disable-securetransport)
 elseif(CORE_SYSTEM_NAME STREQUAL rbpi)
-  if(CPU STREQUAL "cortex-a7" OR CPU STREQUAL "cortex-a53")
-    list(APPEND ffmpeg_conf --cpu=${CPU} --enable-pic --disable-armv5te --disable-armv6t2)
-  endif()
-  list(APPEND ffmpeg_conf --enable-hardcoded-tables --disable-vaapi --disable-vdpau --enable-mmal
-                          --enable-omx-rpi)
-
+  list(APPEND ffmpeg_conf --cpu=${CPU} --disable-vaapi --disable-vdpau)
 endif()
 
-if(CPU MATCHES arm)
+if(CPU MATCHES arm OR CORE_SYSTEM_NAME STREQUAL rbpi)
   list(APPEND ffmpeg_conf --enable-pic --disable-armv5te --disable-armv6t2)
 elseif(CPU MATCHES mips)
   list(APPEND ffmpeg_conf --disable-mips32r2 --disable-mipsdspr1 --disable-mipsdspr2)
--- a/tools/depends/target/ffmpeg/CMakeLists.txt
+++ b/tools/depends/target/ffmpeg/CMakeLists.txt
@@ -11,8 +11,27 @@ if(CROSSCOMPILING)
   list(APPEND ffmpeg_conf --pkg-config=${PKG_CONFIG_EXECUTABLE} --pkg-config-flags=--static)
   list(APPEND ffmpeg_conf --enable-cross-compile --cpu=${CPU} --arch=${CPU} --target-os=${OS})
   list(APPEND ffmpeg_conf --cc=${CMAKE_C_COMPILER} --cxx=${CMAKE_CXX_COMPILER} --ar=${CMAKE_AR})
-  list(APPEND ffmpeg_conf --extra-cflags=${CMAKE_C_FLAGS} --extra-cxxflags=${CMAKE_CXX_FLAGS} --extra-ldflags=${CMAKE_EXE_LINKER_FLAGS})
   message(STATUS "CROSS: ${ffmpeg_conf}")
+endif()
+
+if(CMAKE_C_FLAGS)
+  list(APPEND ffmpeg_conf --extra-cflags=${CMAKE_C_FLAGS})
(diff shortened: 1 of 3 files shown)


notspiff:
Seems sensible. Note i can only offer opinion, not a team member. Can somebody ask jenkins for his opinion please?

popcornmix:
jenkins build this please

nsenica:
I'm not sure why jenkins failed for some builds, but it doesn't look like it is related with this change.

Rechi:
jenkins build this please

Paxxi:
Jenkins fails a bit randomly. 
I would say wait until @wsnipex has had a chance to look at it, I think he's on vacation so might take a little while.

nsenica:
Not a problem, we'll wait :) Thanks.

tools/depends/target/ffmpeg/CMakeLists.txt
```diff
@@ -37,6 +56,13 @@ elseif(CORE_SYSTEM_NAME STREQUAL osx)
                           --disable-decoder=mpeg_xvmc --disable-vda --disable-crystalhd --enable-videotoolbox
                           --target-os=darwin
                           --disable-securetransport)
+elseif(CORE_SYSTEM_NAME STREQUAL rbpi)
+  if(CPU STREQUAL "cortex-a7" OR CPU STREQUAL "cortex-a53")
+    list(APPEND ffmpeg_conf --cpu=${CPU} --enable-pic --disable-armv5te --disable-armv6t2)
+  endif()
+  list(APPEND ffmpeg_conf --enable-hardcoded-tables --disable-vaapi --disable-vdpau --enable-mmal
+                          --enable-omx-rpi)
+
```
popcornmix: I don't think mmal and omx-rpi are useful here. Hardware decode is handled at a higher level in kodi (much more efficient than decoding to arm-side pixel buffers in ffmpeg).

Any evidence on the benefit of ``hardcoded-tables``?

nsenica: If mmal and omx-rpi won't help, I can remove them no problem.

About `hardcoded-tables` I read its description:
`Depending on your use case, --enable-hardcoded-tables may be a useful option. It results in an increase of approximately 15% in the size of libavcodec, the main library impacted by this change. It enables savings in table generation time, done once at codec initialization, since by hardcoding the tables, they do not need to be computed at runtime. However, the savings are often negligible (~100k cycles is a typical number) especially when amortized over the entire encoding/decoding operation. By default, this is not enabled. Improvements are being made to the runtime initialization, and so over time, this option will have an impact on fewer and fewer codecs.``

and I felt that it could make a difference on the raspberry pi because less computation would be needed in runtime. If you think this is not needed, I can remove it as well.

popcornmix: It would be best if you could build with and without ``hardcoded-tables`` and report any measured change in CPU usage or startup time of file playback (it may be measurable from debug log messages close to the "Opening:" message).
I think without a measurable performance improvement, and given it increases library size, then it's best to leave it in the default state as disabled.
But if you could show it saves something noticeable (e.g. 500ms) then we may want it.

nsenica: I think I can try it yes :) Do you have any suggestion on the type of codec that could be ideal to do this test ? Is it worth trying with your backports branch or regular kodi is enough ?

wsnipex:
looks ok for me.
Just for claification though. This is change rounds out a few corner cases, since it will only make a difference if:
1.) building without unified depends
2.) not having a suitable ffmpeg installed system wide

jenkins build this plese

nsenica:
Yes, you're right, and this mainly targets Raspberry Pi while making it more uniform for other platforms as well. This also has the potential to increase performance of ffmpeg if compiled together with Kodi.

I plan to revisit this PR during this weekend, specially because there's a small redundancy on some options for RPi platform, and remove the options that popcornmix mentioned.

tools/depends/target/ffmpeg/CMakeLists.txt
popcornmix: Shouldn't be any different with master branch of kodi or my branch.
Grepping ffmpeg for ``CONFIG_HARDCODED_TABLES`` the only code we are likely to use is mpegaudio (i.e. mp3 audio) and fft (could be used in a number of places). I'd just try a few common video formats and see if you spot anything. My guess would be that you won't.

nsenica: FFT is widely used on decompressing, so there might be a benefit there. If it is noticeable, I also have some doubts. 
I'll do the following, I'll edit this PR and remove the hardcoded-tables (along with the others you mentioned previously), and I'll do some testing. If it proves to be noticeable I'll open a new PR to add that option.

Thanks for taking the time to review this PR.

nsenica:
I've done the changed I mentioned earlier :) I think it is better and less confusing this way.

nsenica:
I'll be away for the next days, feel free to do as you see fit. :) 

Once/if this is merged, I'll provide a backport for krypton as well.

wsnipex:
jenkins build and merge

MartijnKaijser:
jenkins build this please

PR #11587, opened 2017-01-31
FIX: [regression] URL encoding is case-insentive

koying:
Partial revert of 932b0d2cec13970bf1f1e78cb6642e3890ce9d68
URL encoding is case-insentive according to https://tools.ietf.org/html/rfc3986#section-2.1 , so the move from lower to upper case is not required, but causes regression
Fixes #17266

This does nothing to fix #17266 for alpha/beta/RC users who had their db polluted by the issue


hudokkow (approved):

arnova (approved):

MartijnKaijser:
So what is next step? Mandatory database upgrade cleaning it up?

koying:
This was never officially released and probably pretty limited scope.
Db upgrade seems overkill...

PR #11592, opened 2017-01-31
VideoPlayer: VideoReferenceClock to the right place

FernetMenta:
see title. cleans up the mess in videoRefClock. Implementations of videoSync belong to the windowing systems.

@popcornmix could you please care about pi?
@peak3d could you do adjustments for Android?
@mk01  IMX also needs rework


xbmc/cores/VideoPlayer/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 set(SOURCES DVDAudio.cpp
             DVDClock.cpp
+            VideoReferenceClock.cpp
```
stefansaraev: the list should be kept sorted, so move this at the end. same for HEADERS below.

xbmc/windowing/X11/WinSystemX11.cpp
```diff
@@ -20,8 +20,6 @@
 
 #include "system.h"
 
-#if defined(HAVE_X11)
```
popcornmix: This breaks build on Pi. I assume this is being built unconditionally and using HAVE_X11 to avoid build issues on platforms without X.
```
/home/dc4/xbmc_holder/xbmc/xbmc/windowing/X11/WinSystemX11.h:31:22: fatal error: X11/Xlib.h: No such file or directory
 #include "X11/Xlib.h"
```

FernetMenta: if this is included for pi builds, something is going wrong

FernetMenta: what src file includes this header?

popcornmix: WinSystemX11.cpp  includes WinSystemX11.h which includes X11/Xlib.h

FernetMenta: The build system is weak in this regard because it includes files if it finds packages instead of what it really should include. @fetzerch works on a solution. In the meantime you could add another hack to this weird logic there:
https://github.com/xbmc/xbmc/blob/master/xbmc/windowing/CMakeLists.txt#L15

afedchin:
no changes for windows required?

FernetMenta:
@afedchin I would have done them myself but if you like to do so, I'd appreciate it

FernetMenta:
jenkins build this please

FernetMenta:
@Memphiz could you do the changes required for ios. I have XCode 8.2 and can't build ios anymore.

MartijnKaijser:
@peak3d could you look at the Android part?

peak3d:
aml / android implemented

FernetMenta:
jenkins build this please

PR #11537, opened 2017-01-24
 [PVR] Krypton: Trac 17246: Fix deadlock that might occur during initial channel icon search.

ksooo:
Backport of #11536 

@Jalle19 for review?

PR #11593, opened 2017-01-31
Revert "[PVR] Fix deadlock that might occure during initial channel icon search."

ksooo:
This reverts commit df457705c57aabee2c987f783b912b23fe12c2ca.

The deadlock fixed by #11537 is a very special case, but the crash on kodi startup caused by this change happens to quite some people. Reported in the forum and kodi internally.

I will try to come up with a proper fix for 17.1

@MartijnKaijser your button

### DIFF
--- a/xbmc/pvr/channels/PVRChannelGroup.cpp
+++ b/xbmc/pvr/channels/PVRChannelGroup.cpp
@@ -31,7 +31,6 @@
 #include "settings/lib/Setting.h"
 #include "settings/Settings.h"
 #include "threads/SingleLock.h"
-#include "utils/JobManager.h"
 #include "utils/log.h"
 #include "utils/StringUtils.h"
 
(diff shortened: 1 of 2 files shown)


ksooo:
jenkins build this please

Jalle19:
Why does it crash?

ksooo:
CPVRDatabase is in no way thread safe. My change leads to concurrent calls to PVRDatabase - one from PVR manager thread, the other one from the channel icon search job thread.

I wonder whether the 'other' db crash we are trying to fix for ages is actually a concurrency problem? I think the stack traces look similar, but I'm not sure.

ksooo:
@Jalle19 http://pastebin.com/Kp1taJmb <= thread 1 vs thread 2

MartijnKaijser:
Jenkins build this please

PR #11594, opened 2017-01-31
[PVR][jsonrpc] CPVROperations::ToggleTimer: fix crash

ksooo:
CPVROperations::ToggleTimer: fix crash due to invalid epg tag given (e.g. event end time is in the past, thus no timer can be created)

@joethefox could you please runtime test and report back?

@Jalle19 for review?


joethefox:
tested and I can confirm that the issue is solved

ksooo:
jenkins build this please

Jalle19 (approved):

PR #11596, opened 2017-01-31
[backport] Fix freezing when some buttons are held

garbear:
Long presses aren't implemented for joysticks, so holding a button will send the action in rapid succession. If the action is select, this will queue hundreds of select commands and freeze Kodi.

In the long run, I plan to implement long presses for joysticks. For Krypton, however, this simple fix should suffice.

Backport of #11586


MartijnKaijser:
jenkins build this please

MartijnKaijser:
Jenkins build this please

PR #11599, opened 2017-02-01
[estuary] three minor fixes

ronie:
- add vc1 flag (no idea if the vc-1 flag is valid, but vc1 is)
- fix missing delimiter in rss feeds

@phil65


phil65:
looks good, +1

ronie:
added one more for your pleasure :-)

PR #11600, opened 2017-02-01
[Krypton] Controller dialog: Update button label

garbear:
After #11571 goes in, skipping during button-mapping is greatly reduced. Therefore, the "Fix skipping" title becomes largely useless. "Ignore input" is a much more appropriate name.

I have already included this title in the documentation: http://kodi.wiki/view/HOW-TO:Fix_buttons_skipped_while_mapping

This should go in before final release to avoid issues with translations.

After the rename:

ignore_input 5 44 07 pm


garbear:
wasn't expecting Krypton to get tagged today. I'll merge this with #11571

PR #11603, opened 2017-02-02
[Fix] Use lower case table names in query

DaveTBlake:
As MySQL is case sensitve to table name a couple of queries added by  #9300 need correction to lower case.

@MilhouseVH thanks for spotting this


MartijnKaijser:
Jenkins build this please

DaveTBlake:
This really is just a typo, with bad consequnces for MySQL, so going to merge

PR #11604, opened 2017-02-02
[Fix] [Backport]Use lower case table names in query

DaveTBlake:
As MySQL is case sensitve to table name a couple of queries added by #9300 need correction to lower case.

Backport of #11603


MartijnKaijser:
Jenkins build this please

PR #11606, opened 2017-02-02
[PVR] Fix crash in CPVRManager::UpdateItem

ksooo:
Crash reported in the forum: http://forum.kodi.tv/showthread.php?tid=298461&pid=2513928#pid2513928 => http://sprunge.us/cRSg

@Jalle19 for review?


Jalle19 (commented):
Can you have the reporter test this fix?

ksooo:
I experienced this crash as well and I have runtime-tested the fix. It's very hard to reproduce. So I doubt the submitter is able to reproduce it.

But I have this patch applied on my private build for 3 weeks now and since than the crash did not occur any more.

ksooo:
jenkins build this please

Jalle19:
@ksooo great, merge away in that case!

PR #11609, opened 2017-02-03
Fix cleanonupdate to clean music library

DaveTBlake:
The music library advanced setting <cleanonupdate> has not been working properly since #5735 in Helix.  When enabled songs on any offline sources are removed during library update as intended. However it also accidentally prevents  any subsequent scraping additional album or artist information, and fails to clean the library when no files are being scanned for tags.

This brings processing in line with how it works for the video library.

To see this fault enable <cleanonupdate> for the music library in advancedsettings.xml and having done a library update, go to album or artists node and click on "Query Info for all" on the context menu. No scraping is ever attempted.

This could be backported, or you could argue that broken since Helix there is no rush.

@Montellese perhaps you have an interest in this?
@da-anda you almost volunteered :)

### DIFF
--- a/xbmc/music/infoscanner/MusicInfoScanner.cpp
+++ b/xbmc/music/infoscanner/MusicInfoScanner.cpp
@@ -92,18 +92,6 @@ void CMusicInfoScanner::Process()
   ANNOUNCEMENT::CAnnouncementManager::GetInstance().Announce(ANNOUNCEMENT::AudioLibrary, "xbmc", "OnScanStarted");
   try
   {
-    if (m_bClean)
-    {
-      CleanDatabase(false);
-      m_bRunning = false;
-
-      return;
-    }
-
-    unsigned int tick = XbmcThreads::SystemClockMillis();
-
-    m_musicDatabase.Open();
-
     if (m_showDialog && !CServiceBroker::GetSettings().GetBool(CSettings::SETTING_MUSICLIBRARY_BACKGROUNDUPDATE))
     {
       CGUIDialogExtendedProgressBar* dialog =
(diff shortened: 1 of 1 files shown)


DaveTBlake:
Anyone care to review this? Looking to merge next week regardless.

jenkins build this please

xbmc/music/infoscanner/MusicInfoScanner.cpp
```diff
@@ -112,7 +104,14 @@ void CMusicInfoScanner::Process()
...
     }
 
-    m_bClean = g_advancedSettings.m_bMusicLibraryCleanOnUpdate;
+    // check if we only need to perform a cleaning
+    if (m_bClean && m_pathsToScan.empty())
+    {
+      CleanDatabase(false);
+      m_bRunning = false;
+
+      return;
```
garbear: Could you remove the early return?

DaveTBlake: Why do that? It is within an If statement where we have `m_pathsToScan.empty()`, so there are no items to scan, nothing to do but clean. Hence we clean and return. Or are you seeing something else?

BTW thnaks for looking.

garbear: In general, early returns are bad. Here, there are side effects from the other code path if we return half way through the function. This branch should be taken at the beginning of the function.

DaveTBlake: Sorry I don't see any "side effects"? Getting "tick" I guess? The db does need to be open for just cleaning as well as the scanning, although it will open it if it isn't. I can move those, but we do want the progress display when not being done background in case delete takes noticeable time, so branch is not at beginning.

garbear: side effect - m_handle is set. It's inconsequential, I was just pointing out an example of why early returns should be avoided. And it looks like it never actually opens the progress display before returning?

DaveTBlake: I'm happy to have things pointed out when I am being blind. :)

`CVideoInfoScanner` does similar with progress, but cleaning branch does set  `m_handle = NULL`
https://github.com/xbmc/xbmc/blob/efe979aec099a7fb2bb13ff235280ba1fdf55cb8/xbmc/video/VideoInfoScanner.cpp#L92

What do you think?

garbear: sure, cleanup commits are always welcome. just make sure they're separated from any fixes.

DaveTBlake:
Tidy up, thanks @garbear 
jenkins build this please

DaveTBlake:
jenkins build this please

PR #11612, opened 2017-02-04
Don't assign wrong InfoTags to items in the JSON handler

dgam1:
In interfaces/json-rpc/FileItemHandler.cpp method FillFileItemList in some cases assigns wrong InfoTags to FileItems. One way this becomes apparent is when you add to the picture playlist and then try to open it using JSON queries like these:

`{"params": {"item": [{"file": "/pics/pic1.jpg"}, {"file": "/pics/pic2.jpg"}, {"file": "/pics/pic3.jpg"}], "playlistid": 2}, "jsonrpc": "2.0", "id": 1, "method": "Playlist.Add"}`

`{"params": {"item": {"position": 1, "playlistid": 2}}, "jsonrpc": "2.0", "id": 1, "method": "Player.Open"}`

With the pictures having VideoInfoTags the slideshow will not work. This patch will have the method create InfoTags only as needed, which in some cases is not at all where the players can handle FileItems without them.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/interfaces/json-rpc/FileItemHandler.cpp
+++ b/xbmc/interfaces/json-rpc/FileItemHandler.cpp
@@ -387,41 +387,26 @@ void CFileItemHandler::HandleFileItem(const char *ID, bool allowFile, const char
 
 bool CFileItemHandler::FillFileItemList(const CVariant &parameterObject, CFileItemList &list)
 {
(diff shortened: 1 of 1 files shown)


da-anda:
@Montellese mind also having a look? :)

MartijnKaijser:
@dgam1 we'll close it. Please let us know if we misjudged the intend of the PR

PR #11613, opened 2017-02-04
Have the JSON handler sent a custom CEvent to the slideshow class

dgam1:
When you successfully start a picture slideshow with a JSON query you won't receive a response until the slideshow has been closed. To reproduce send a query like this:

`{"params": {"item": {"path": "/path/to/pictures/"}}, "jsonrpc": "2.0", "id": 1, "method": "Player.Open"}`

The reason being that the slideshow runs a new nested main loop and so, until it's closed, doesn't release the lock of the ThreadMessage the JSON handler is waiting at. With this patch a custom CEvent will be attached to the GUIMessage, which the slideshow class will temporarily set as a member variable to be pulsed as soon as processing of the slideshow begins. All of this takes place before the new main loop processes other messages.
A simpler fix would be to send a non-waiting GUIMessage, but then the caller would receive an ACK before the slideshow begins handling the request.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/interfaces/json-rpc/PlayerOperations.cpp
+++ b/xbmc/interfaces/json-rpc/PlayerOperations.cpp
@@ -1156,7 +1156,7 @@ JSONRPC_STATUS CPlayerOperations::StartSlideshow(const std::string& path, bool r
   g_application.WakeUpScreenSaverAndDPMS();
   CGUIMessage msg(GUI_MSG_START_SLIDESHOW, 0, 0, flags);
   msg.SetStringParams(params);
-  CApplicationMessenger::GetInstance().SendGUIMessage(msg, WINDOW_SLIDESHOW, true);
+  CApplicationMessenger::GetInstance().SendGUIMessage(msg, WINDOW_SLIDESHOW);
 
   return ACK;
 }


da-anda:
@Montellese mind having a look?

dgam1:
I updated the pull request.
The idea behind using a shared pointer for the CEvent is for the main thread to be able to handle the possibility of the JSON thread dying without taking the CEvent with it. Sending over the reference to one single shared pointer obviously doesn't do the job so I changed it so that a second one gets sent over for the slideshow to take care of.

Paxxi (changes requested):
Appropriate way to solve this is to use PostMsg instead of SendGuiMessage 

It will return instantly without blocking

dgam1:
@Paxxi Sending a non-waiting GUIMessage is the same as using PostMsg. And that would indeed be a simple fix for the blocking issue. But then the behavior for slideshows would be different than for (non-PVR) audio and video where JSON clients receive a response only after the request has been processed. A client would then have to repeatedly poll the player status to see if the slideshow started.

Paxxi:
The difference is that SendGuiMessage is basically a legacy method intended for use by CGUI* classes. No new code outside of those should know what a CGuiMessage is or use it, therefore SendMsg or PostMsg should be used. 

I don't like polling in general but waiting on an event like this can lead to unpleasant issues.  One example is that when using Kore starting a movie, if a dialog opens, e. G.

dgam1:
I reduced the change to sending a non-waiting GUIMessage, which resolves the blocking issue for the moment. Right now there is no abstraction when it comes to dealing with slideshows as the JSON classes directly access the GUIWindow. That's a bigger task I might work on at a later time.

Paxxi:
Help in that area is certainly welcome, don't hesitate to ask if there's anything.

PR #11614, opened 2017-02-04
CAddonInstallJob: disable add-on after update, if it is marked as broken

Rechi:
## Description
Disable broken add-ons after update.

## Motivation and Context
Add-ons which are marked as broken can't be installed, but they don't get disabled if they were already installed before getting marked as broken.

## How Has This Been Tested?
Runtime tested

## Types of change

## Checklist:

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -14783,7 +14783,10 @@ msgctxt "#24093"
 msgid "Checking %s..."
 msgstr ""
 
-#empty string with id 24094
+#: xbmc/addons/AddonInstaller.cpp
+msgctxt "#24094"
+msgid "Add-on has been marked broken in repository, so it will get disabled."
+msgstr ""
 
 #: xbmc/addons/GUIDialogAddonInfo.cpp
 msgctxt "#24095"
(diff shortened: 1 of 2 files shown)


Rechi:
@tamland can you have a look?

tamland:
Hm, I was just working on the same thing (I think), but on sort of a different solution. The thing is, this functionality already exists here: https://github.com/xbmc/xbmc/blob/master/xbmc/addons/Repository.cpp#L261 (except it doesn't work currently due to refactoring) which needs to be take care of as well.

tamland:
Auto-disabling is a good idea though.. Definitely liking this better that the old popup flooding which is just horrible. *BUT*, it's a quite intrusive thing to do so I don't think we should forcefully do that on everything in the installer, but rather (e.g.) only when it's an auto install and not overridden by user in any way.

MartijnKaijser:
The broken tag usually contain information why it is broken so we should at least show that message.

Rechi:
I can add it to EventLog and also show a notifictaion message when an addon gets disabled

Rechi:
updated

tamland:
@Rechi Not sure if you wanted to finish this up or not, but couldn't wait so I stole this code to->#11664

Rechi:
@tamland I thought already all changes were done, but I have no problem you took the code.
can be closed if it isn't needed anymore.

tamland:
Ok, sorry for not being very clear. It needed to take care of handling users manually installing and not disable those (and cleaning up the old code). Which is essentially what I've added to #11664.

PR #11615, opened 2017-02-04
[PVR] Fix trac17246 - another approach

ksooo:
This is a fix for trac 17246: http://trac.kodi.tv/ticket/17246

Completely new approach this time, which should no longer lead to crashes the previous approach has caused. ;-)

The gist of the fix is to obtain and remember the pointer to the extended progress bar dialog instance early (PVR Manager init), avoiding the deadly <code>g_WindowManager.GetWindow()</code> call.

I discovered other scenarios which could lead to the same deadlock (in <code>CPVRMananger::ShowProgressDialog(string,int)</code>)which are are also fixed now.

The stack trace of one of the possible deadlocks:
kodi-frozen

@Jalle19 for review?

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -10619,13 +10619,13 @@ msgctxt "#19285"
 msgid "Browse for icon"
 msgstr ""
 
-#. unused?
+#. Label for channel icon search progress dialog
+#: xbmc/pvr/channels/PVRChannelGroup.cpp
 msgctxt "#19286"
 msgid "Searching for channel icons"
 msgstr ""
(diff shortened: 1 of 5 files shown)


Jalle19 (commented):
If it works, sure. I'm wondering if it would make sense to remove this feature altogether since most backends AFAIK are able to supply channel icons on their own.

ksooo:
jenkins build this please

ksooo:
We have positive feedback from the reporter of the crashes the first approach introduced.

ksooo:
The submitter of the trac ticket got a a test build, did runtime testing, and confirmed that the original issue is solved, no crashes.

PR #11616, opened 2017-02-05
Fix analog sticks

garbear:
The first fix is for a logic error in #11571 that breaks analog sticks.

Also included is a fix for an edge case documented in bd6697b.

Another two fixes have been PRed to https://github.com/xbmc/peripheral.joystick/pull/84


garbear:
I split the cleanup commits into #11621. I also simplified the detection code by removing checks for the quirk that a following PR will addresses.

jenkins build this please

PR #11617, opened 2017-02-05
[PVR] Krypton: Fix trac17246 - another approach

ksooo:
Backport of #11615 

@Jalle19 for review?


ksooo:
jenkins build this please

ksooo:
Both the submitter of the trac ticket and the reporter of the crashes the first fix attempt had introduced confirmed that issues are fixed now.

Jalle19 (approved):
As long as it works :-)

PR #11618, opened 2017-02-05
highlight programinfo instead of channelname

hamphh73:
Highlight programinfo instead of channelname

## Description
In the channellist the programinfo is the more important information compared to the channelname. So the programinfo should be at the top and use the big font.

## Motivation and Context
Better overview in channellist.

## How Has This Been Tested?
Tested in current krypton release

## Types of change

## Checklist:
- [x ] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 


ksooo:
Sorry, I do see it exactly the other way around, namely like it currently is. -1 from me on this.

PR #11619, opened 2017-02-05
FileCache: Increase CHUNK_SIZE to 128 KB

fritsch:
I need that to play highbitrate videos over my wireless bridge.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
@@ -295,10 +295,10 @@ bool CDVDDemuxFFmpeg::Open(CDVDInputStream* pInput, bool streaminfo, bool filein
       if (trySPDIFonly || (iformat && strcmp(iformat->name, "wav") == 0))
       {
         AVProbeData pd;
-        uint8_t probe_buffer[FFMPEG_FILE_BUFFER_SIZE + AVPROBE_PADDING_SIZE];
+        std::unique_ptr<uint8_t[]> probe_buffer (new uint8_t[FFMPEG_FILE_BUFFER_SIZE + AVPROBE_PADDING_SIZE]);
 
         // init probe data
-        pd.buf = probe_buffer;
+        pd.buf = probe_buffer.get();
         pd.filename = strFile.c_str();
 
         // av_probe_input_buffer might have changed the buffer_size beyond our allocated amount
--- a/xbmc/cores/FFmpeg.h
+++ b/xbmc/cores/FFmpeg.h
@@ -65,4 +65,4 @@ public:
   int level;
 };
 
-#define FFMPEG_FILE_BUFFER_SIZE   32768 // default reading size for ffmpeg
+#define FFMPEG_FILE_BUFFER_SIZE   131072 // default reading size for ffmpeg
--- a/xbmc/filesystem/FileCache.cpp
+++ b/xbmc/filesystem/FileCache.cpp
@@ -43,7 +43,7 @@
 
 using namespace XFILE;
 
-#define READ_CACHE_CHUNK_SIZE (64*1024)
+#define READ_CACHE_CHUNK_SIZE (128*1024)
 
 class CWriteRate
 {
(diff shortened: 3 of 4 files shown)


fritsch:
@popcornmix @Memphiz @FernetMenta as we talked before about that. Okay?

popcornmix:
Does it need to be that high? We've been testing 128K in milhouse builds.
We know 1M causes very slow opening of ISO files (from below 10s to over 30s)
http://forums.kodi.tv/showthread.php?tid=250817&pid=2307214#pid2307214
so I'm a bit nervous about 512K.

fritsch:
Can't we let this user test a build with 512 and one with 128? And ask which one is better? (Without telling in advance?)

fritsch:
Btw. I chose this 512 cause I have seen it in a @memphiz thread.  I am happy with 128, too.

popcornmix:
I've pinged the original guy from forum.
Memphiz did say that libnfs should use the chunksize the server reports:
http://forum.kodi.tv/showthread.php?tid=298461&pid=2503108&highlight=memphiz#pid2503108
Can you confirm if that works for you?

fritsch:
I think you asked me to test that the last time, basically: https://github.com/fritsch/xbmc/blob/cc8ee867d472cbe7991fb06452b49c77e39006b1/xbmc/filesystem/FileCache.cpp#L177 would end up with 64 KB as NFS returns just 1.

fritsch:
Btw. this would be:
```
20:04:06.593 T:140226060084992   DEBUG: NFS: chunks: r/w 1048576/1048576
```

1 MB.

fritsch:
Btw. currently the write is clamped to 32 KB also - with that change I have a bad feeling :-)

arnova:
I'm wondering: Why does a bigger chunk-size help here? Is the NFS implementation that crappy? And if it's an NFS-problem, why increase it for all protocols?

MrMC:
FYI: https://lists.samba.org/archive/samba-technical/2014-February/097855.html

fritsch:
@arnova requesting small chunks in a row over weak connection with varying
response time is really bad in general no matter what protocol.

Am 07.02.2017 2:43 nachm. schrieb "Scott Davilla" <notifications@github.com

Memphiz (commented):
I think this change might be correct regardless of the cache size...

fritsch:
Rebased and added a commit by @popcornmix

popcornmix:
These commits were in last Milhouse build.
Previously I could only play live HD TV recordings over wifi with buffermode=1.
Now I can play them without that setting.
So +1 from me.

fritsch:
@FernetMenta is the ffmpeg change fine with you?

MrMC:
uint8_t probe_buffer[FFMPEG_FILE_BUFFER_SIZE + AVPROBE_PADDING_SIZE];

be careful you don't blow up your local stack.

fritsch:
@MrMC okay with a smartptr?

MrMC:
smartptr is ok

fritsch:
jenkins build and merge

(got fernet's OK on IRC).

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
```diff
@@ -295,10 +295,10 @@ bool CDVDDemuxFFmpeg::Open(CDVDInputStream* pInput, bool streaminfo, bool filein
       if (trySPDIFonly || (iformat && strcmp(iformat->name, "wav") == 0))
       {
         AVProbeData pd;
-        uint8_t probe_buffer[FFMPEG_FILE_BUFFER_SIZE + AVPROBE_PADDING_SIZE];
+        std::unique_ptr<uint8_t[]> probe_buffer (new uint8_t[FFMPEG_FILE_BUFFER_SIZE + AVPROBE_PADDING_SIZE]);
```
FernetMenta: I don't think a smart pointer is a good idea for a buffer handed to ffmpeg. In this case the variable goes out of scope while ffmped still has a ref to the buffer

popcornmix: But is this any different to previous code? There, a buffer on stack was passed in, which will disappear when scope is left.

fritsch:
Yeah.

Am 26.04.2017 2:50 nachm. schrieb "popcornmix" <notifications@github.com>:

FernetMenta:
I am not sure. local variable are in general allocated at the start of a function.

popcornmix:
I'm pretty sure C/C++ only guarantees allocation within the closest scope of a block.
If it is allocated at start of function that is purely a compiler choice (that may change with versions/optimisation level).
See: http://stackoverflow.com/questions/30702717/memory-allocation-of-local-variables-within-nested

Possibly the code has always been broken and we are only now getting hit.
When would be best to allocate and free?
Make probe_buffer a member variable, allocate in the same local scope and free in destructor?

PR #11620, opened 2017-02-05
jemalloc depend build

peak3d:
Use jemalloc for 64-bit linux builds

## Description
This PR links jemalloc static library to kodi wich hooks / overrides all memory functions from stdlib.
This step is necessary because linux memory allocator fails for certain situations: keyword: milhouse script with 16 threads importing a fake library with 1000 movies

## Motivation and Context
Reduce / speedup memory allocation

## How Has This Been Tested?
odroid-C2 / milhouse script / nedscotts fake library.
RES value changes from 800MB down to 90 MB

## Types of change

## Checklist:

### DIFF
--- a/cmake/scripts/linux/ArchSetup.cmake
+++ b/cmake/scripts/linux/ArchSetup.cmake
@@ -44,3 +44,7 @@ endif()
 if(ENABLE_MIR)
   set(ENABLE_VDPAU OFF CACHE BOOL "Disabling VDPAU since no Mir support" FORCE)
 endif()
+
+if(CPU MATCHES 64 AND (NOT (DEFINED ENABLE_JEMALLOC) ) )
+  set(ENABLE_JEMALLOC ON CACHE BOOL "Enable jmalloc allocator for 64-bit linux builds" FORCE)
+endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,7 @@ if(UNIX)
     option(ENABLE_X11         "Enable X11 support?" ON)
     option(ENABLE_AML         "Enable AML?" OFF)
     option(ENABLE_IMX         "Enable IMX?" OFF)
+    option(ENABLE_JEMALLOC    "Enable JEMALLOC?" OFF)
   endif()
 endif()
 # System options
@@ -115,7 +116,7 @@ if(CORE_SYSTEM_NAME STREQUAL android)
 endif()
 
 # Optional dependencies
-set(optional_deps MicroHttpd MySqlClient SSH XSLT
+set(optional_deps MicroHttpd MySqlClient SSH XSLT JEMALLOC
                   Alsa UDEV DBus Avahi MDNS SmbClient CCache
                   PulseAudio VDPAU VAAPI Bluetooth CAP Python)
 
--- a/cmake/cpack/deb/packages/kodi.txt.in
+++ b/cmake/cpack/deb/packages/kodi.txt.in
@@ -16,7 +16,7 @@ PACKAGE_SECTION video
 PACKAGE_PRIORITY optional
 PACKAGE_SHLIBDEPS
 PACKAGE_DEPENDS @APP_NAME_LC@-bin (>= @CPACK_DEBIAN_PACKAGE_VERSION@), @APP_NAME_LC@-bin (<< @CPACK_DEBIAN_PACKAGE_VERSION@.1~), curl, libcurl3, mesa-utils, x11-utils, fonts-liberation | ttf-liberation, fonts-dejavu-core | ttf-dejavu-core, python-bluez | python-lightblue, python-imaging, python-simplejson, libass5 | libass4, libgif5 | libgif7, libssh-4 | libssh2-1, libnfs8 | libnfs4 | libnfs1, libbluray1, libshairplay0, libvorbisfile3, libaacs0, libcec4, libgnutls30 | libgnutls-deb0-28 | libgnutls28 | libgnutls26, libxslt1.1, libyajl2
-PACKAGE_RECOMMENDS libvdpau1, libva-intel-vaapi-driver, libva1
+PACKAGE_RECOMMENDS libvdpau1, libva-intel-vaapi-driver, libva1, libjemalloc-dev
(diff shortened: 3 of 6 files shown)


cmake/scripts/linux/ArchSetup.cmake
```diff
@@ -44,3 +44,7 @@ endif()
 if(ENABLE_MIR)
   set(ENABLE_VDPAU OFF CACHE BOOL "Disabling VDPAU since no Mir support" FORCE)
 endif()
+
+if(CPU MATCHES 64)
+  set(ENABLE_JEMALLOC ON CACHE BOOL "Enable jmalloc allocator for 64-bit linux builds" FORCE)
```
stefansaraev: jemalloc should be completely optional

lrusak:
Filename should be FindJEMALLOC.cmake otherwise cmake can't find it when it is named Findjemalloc.cmake

fritsch:
Because you prefer the oom killer? From my testing an optional disable is
okay, so the other way round.

Am 05.02.2017 11:31 nachm. schrieb "Stefan Saraev" <notifications@github.com

cmake/scripts/linux/ArchSetup.cmake
peak3d: It is an cmake option. You can pass -DENABLE_JMALLOC=OFF in the cmake step

stefansaraev:
IMO replacing the system malloc implementation is dangerous. it may affect shitload of library code (see how many dependencies kodi is using) that you dont have of.

as you seem to think it's ok to force it on everyone, I would like to have few questions answered:

- are you sure it is malloc that fails. how exactly does it fail.
- are you sure it is glibc's ptmalloc that fails, and it's not glibc / kernel bug or mis-configuration
- does the failure on everything linux 64bit on only specific board / distro / configuration
- does the failure happen on low-mem devices only, in that case, does changing kernel allocator make a difference
- are you sure it is not oom killer bug. iirc, there was a regression in 4.4 (ubuntu xenial) kernel recently
- in case it fails only with odroid/LE, is it related to sqlite's default mmap size?

cmake/scripts/linux/ArchSetup.cmake
stefansaraev: that didn't work for me, so I assumed you wanted it forced-enabled on aarch64/x86_64.

fritsch:
Currently busy. Will answer in 2 hours. All valid questions. I debugged the
issue for two days and asked most of them to myself. Thanks for asking,
will come back.



Am 06.02.2017 10:14 vorm. schrieb "Stefan Saraev" <notifications@github.com

peak3d:
@stefansaraev would be great if you can give us the answers to your questions.

fritsch:
After a whole weekend of debugging my answer and theory is as follows:
- malloc does not fail. It allocates and frees the memory correctly
- See first line
- No failure happens. We have an issue depending on how memory is allocated. From my testing it is a fragmentation issue.
Use case: 16 threads for artwork caching.
Every thread does: 
a) Allocate memory for texture (_aligned_malloc)
b) Create an FFmpegImage with one outputbuffer (av_malloc) and one intermediate buffer (av_malloc)
c) m_outputbuffer data is copied into texture
d) outputuffer and intermediate buffer is freed
e) Texture is cached.
No the following happens (you can verify with memcheck of valgring, by telling it to write e.g. 0x21 into the memory it frees):
[T][F][F][T][F][F][T][F][F]... <- a, b
[T][0x21][0x21][T][0x21][0x21] ...

Now we have memory fragmentation, in the case that some static (Texture) and some dynamic, short living stuff (FFmpegImage) is interleaved. Now consider memory is needed larger than the freed [F]. It gets it at the end. So your memory grows and grows cause of the fragmentation.

- no oom killer bug. Not 4.4 regression. Verified on 4.10, 4.9
- Not only related to LE. Happens on Ubuntu 16.10, Ubuntu 16.04

jemalloc's better allocating / deallocating strategy "worksaround" this fragementation and we properly stay with low RES.

On my Apollo Lake System with 1500 movies: 0.4gb vs. 1.2 gb (RES), VIRT obviously much more.
If you have time. Please get the millhouse texture script, set it to 16 threads and tell us your findings. With and without jemalloc. If you have more time, use valgrind to set the memory it frees to 0x21 or something else, dump the core and grep over the image to see the fragementation.

Is this explanation reasonable to you?

FernetMenta:
> Now we have memory fragmentation, in the case that some static (Texture) and some dynamic, short living stuff (FFmpegImage) is interleaved

the test case with the given script did not produce any static textures. how do you explain the memory consumption?

popcornmix:
I wonder if this bug report is the same issue?
https://sourceware.org/bugzilla/show_bug.cgi?id=14581

fritsch:
@FernetMenta What I meant is: the allocation prior to the dynamic ffmpeg allocation rests longer than the ffmpeg allocation, therefore the 16 threads cause x00y00z00a00b00

It's a theory by now. Really happy for input.

What do you guys get in your coredumps with valgrind's memcheck and the free pattern?

fritsch:
@popcornmix that looks exactly like our issue.

popcornmix:
Does:
```export MALLOC_MMAP_THRESHOLD_=131072```
make any difference to the glibc memory usage?
Ref: https://bugzilla.redhat.com/show_bug.cgi?id=843478

fritsch:
@popcornmix yes - that helps! it stays down.

@peak3d can you also check? (Make really sure it's exported)

fritsch:
@stefansaraev after all those explanations. Can you reproduce?

stefansaraev:
@fritsch yes. thank you, I know how it's supposed to work, in theory...

yes, I can reproduce. testing on wetek play 2. kodi master. glibc 2.25. ffmpeg 3.2.3. texturecache.py 16 threads.

without this PR:
- MALLOC_MMAP_THRESHOLD_ not set, - ~900MB used
- MALLOC_MMAP_THRESHOLD_=131072 - memory usage varies, but it's relatively low - 150 - 300MB

with this PR (jemalloc cross compiled and linked into final executable)
- with or without MALLOC_MMAP_THRESHOLD_ set, it's even worse - ~1500MB. 

jemalloc is pain to cross compile. autoreconf is broken. fails to intall doc/ and so on. but I got it ready in 5min. compiled with --enable-static --disable-shared --with-pic. 

does increased memory usage with jemalloc on my setup make sense to you ?

peak3d:
@popcornmix nice finding 👍 ! Thanx!
Time to close this PR, ok?

MilhouseVH:
> does increased memory usage with jemalloc on my setup make sense to you ?

That isn't what I've seen on x86 (Skylake i5 NUC with 8GB RAM). Here are the results of texturecache.py 16 threads (repeated 3 times) for plain Kodi (regular malloc), kodi + jemalloc as a shared library, kodi + jemalloc static link:
http://pastebin.com/raw/2RiG8dzA

With the switch to the EXPORT, the results are comparable to jemalloc shared/static.

PR #11622, opened 2017-02-06
Fix key length in UNIQUE INDEX idxArtist1

zuckschwerdt:
This patch changes the key length on the idxArtist1 index from 36 to 255 chars.

## Description

This changes a textual SQL command in xbmc/music/MusicDatabase.cpp
void CMusicDatabase::CreateAnalytics()

## Motivation and Context

The strMusicBrainzArtistID is usually a simple UUID (36 characters).
Collaborations ("feat.") are given as UUID"/"UUID (73 characters) though.

E.g.

```
The Crystal Method  aaf09f31-bb5c-43e5-9f54-bb6554c33a71
```
and
```
The Crystal Method feat. Justin Warfield  aaf09f31-bb5c-43e5-9f54-bb6554c33a71/2a5b713c-839e-4392-b344-59829ec77613
```

Creating the idxArtist1 index with just 36 characters will result in collisions.
This is especially bad when trying to upgrade the library to a newer version:

## How Has This Been Tested?

Tested with an upgrade install of Kodi 17.0 Krypton

```
NOTICE: Old database found - updating from version 56 to 60
[MyMusic60] Undefined MySQL error: Code (1062)
Query: CREATE UNIQUE INDEX idxArtist1 ON artist(strMusicBrainzArtistID(36))
```

(the actual error is: `ERROR 1062 (23000): Duplicate entry 'NULL' for key 'idxArtist1'`)

## Types of change

## Checklist:

### DIFF
--- a/xbmc/music/MusicDatabase.cpp
+++ b/xbmc/music/MusicDatabase.cpp
@@ -210,7 +210,7 @@ void CMusicDatabase::CreateAnalytics()
   m_pDS->exec("CREATE INDEX idxGenre ON genre(strGenre(255))");
 
   m_pDS->exec("CREATE INDEX idxArtist ON artist(strArtist(255))");
-  m_pDS->exec("CREATE UNIQUE INDEX idxArtist1 ON artist(strMusicBrainzArtistID(36))");
+  m_pDS->exec("CREATE UNIQUE INDEX idxArtist1 ON artist(strMusicBrainzArtistID(255))");
 
   m_pDS->exec("CREATE INDEX idxPath ON path(strPath(255))");
 


phate89:
Changing the createanalytics function isn't enough. You also need to increase database version to trigger recreation ofindices
@DaveTBlake for better review

zuckschwerdt:
I encountered this problem while upgrading to 17.0, the upgrade was unsuccessful. The database upgrade would rerun each startup and fail at `CREATE UNIQUE INDEX idxArtist1 ON artist(strMusicBrainzArtistID(36))`. Applying this fix then enabled the upgrade to run successfully. I guess I fixed this manually when migrating to mysql on the last upgrade.
If the database is already current and working this change will indeed do nothing.
I'm hesitant to effect a database upgrade for such an insignificant fix. It might be better to hold off on this until the next db version increase then.

DaveTBlake:
Sorry but this PR is wrong in many ways.
The index is correct as it is. Collaborations should be split into separate artists matched with their correct UUID. 

The issue is that your original v56 db that you are trying to upgrade has data from tags that were incorrectly processesed.  "The Crystal Method feat. Justin Warfield" is 2 artists with separate MBIDs aaf09f31-bb5c-43e5-9f54-bb6554c33a71 and 2a5b713c-839e-4392-b344-59829ec77613.

I do not know how they got into your db, but scanning that music file with v17.0 will correctly split the artists and MBIDs into separate records.

For future Kodi versions it could be worth considering checking for what are in effect corrupted records as part of the upgrade process.

DaveTBlake:
Cross posted!
You just need to rescan your music rather than upgrade and old db.

phate89:
@DaveTBlake no idea about music part as you know so you know better about this pr but if strMusicBrainzArtistID can't be bigger than 36 characters it should be of type varchar(36) and not text...

DaveTBlake:
Yes that is true, an historic oversight. Although with SQLite there is no advantage to varchar(x) over text, they take the same space.

phate89:
Excluding the advantage in the database itself setting the right size limit allows to have an additional check to avoid wrong data in the db like in zuckschwerdt's db..

DaveTBlake:
A db uprade would have to handle finding and fixing that wrong data too. The need for that was the useful part of this user feedback.

zuckschwerdt:
I just checked the backups and the composite artist ids a present with version 46 (Oct 2013). My version before that was 37 (May 2013) and doesn't have those entries.
Thanks for the help and insight, and sorry for running in the wrong direction so hastily ;)

DaveTBlake:
You will really benefit from a fresh start with the improved tag processing in v17, scanning from scratch. Unfortunately that means loosing playcounts etc.  Any future issues checking in at the Music Support forum first could be worthwhile.

I'm going to close this PR, but it has raised some ideas for future improvements which I can incorporate.

Montellese:
@DaveTBlake: IMO expecting users with an existing music database to know that they have to drop their entire music database because there might be a conflict with an index during the upgrade process is really bad.
The way it is now any user with an existing music db which runs into this problem will start kodi, the music db upgrade will fail leaving the user with a partially upgraded database and most likely with a bad SQL schema. In that situation it's no use to do a full rescan of your music because you will try to write it into a broken music db.

DaveTBlake:
@Montellese I agree, I do not have that expectation. The problem this user hit has been waiting to happen for several versions at least, nothing I have added has made it any worse. But as I said it has raised some issues that I now realise need improvement and will address for v18.

DaveTBlake:
A postscript: this unique index has existsed on strMusicBrainzArtistID
`CREATE UNIQUE INDEX idxArtist1 ON artist(strMusicBrainzArtistID(36)`
since Frodo, db version 32, when Mbids were added to the schema. 

This should prevent any records being inserted or the artist table updated in any way that violates that index.  So I do not understand how @zuckschwerdt has old MySQL database(s) with both `aaf09f31-bb5c-43e5-9f54-bb6554c33a71` and `aaf09f31-bb5c-43e5-9f54-bb6554c33a71/2a5b713c-839e-4392-b344-59829ec77613`. Was that a real example from the db that failed to migrate?

(But worth noting that SQLite does not support length parameters on indexes, so a SQLite db will allow those records to coexist, but equally they would not violate the creation of analytics on migration to a new db schema)

The actual error was 
`ERROR 1062 (23000): Duplicate entry 'NULL' for key 'idxArtist1'`
this says that MySQL is rejecting a repeated **NULL**, not some duplicated mbid,  and yet it normally allows unique indexes to have NULL values. This is very common in the artist table.

Note that changing the index to 255 would not have avoided rejection of duplicate NULL.

The only mention I could find on Google was that a few other applications had MySQL randomly complain about duplicate NULLs, but all were fine on retry. The suggestion is db corruption of some kind solved by export, recreate, and import. It seems we have to write this up as yet another random MySQL querk. But I will see if I can change things in v18 so that the failed migration is more obvious.

zuckschwerdt:
To explain a bit further: I migrated the sqlite db to mysql for v17 by hand. The db was about 10 years old (Dharma at least). I guess the index was missing or manually fixed back then. The example is real, I had about 50 compound entries, i.e. longer than 36 bytes.
The error message is surprising and misleading. My best guess is that the truncated key maps to something that is reported as NULL. I ran the query by hand and tested things, NULL entries are not a problem. I did manage to import the db with the longer index setting. Also manually adding a longer index on the broken db worked. And removing/nulling the long ids is also a solution. 
However, as you recommended, I nuked the db an started fresh now.

PR #11623, opened 2017-02-06
Add dialog for delayed axis detection

garbear:
This is a follow-up to #11616.

On some platforms, like OSX, joystick drivers send events instead of polling for the joystick's state. If an axis has sent no events, its value won't be known initially. This breaks button mapping, because we need the initial value to detect the axis's center. For example, if a trigger is pushed really fast before being seen by Kodi, the first value could arrive after the trigger is fully pressed. It would look like the fully-pressed position is the center, and when the trigger is unpressed it seems "stuck".

To solve this, if an axis is discovered by Kodi during button mapping, a dialog is shown to the user. This dialog absorbs all button presses, which allows all axes to be moved so that they are detected by Kodi. In future button mapping sessions, the initial value of the axes will be known.

This dialog only needs to be shown once per restart. I must admit, it is very annoying have this error pop up during button mapping. However, I can't think of any way to detect the center if the axis is discovered in any possible state.

I'm open to suggestions for a better help message.

screen shot 2017-02-06 at 1 57 24 am

## Types of change

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -16465,16 +16465,19 @@ msgstr ""
 #empty string with id 35013
 
 #. Help text of the button to fix the bug where buttons are skipped in the controller dialog. %s - list of disabled buttons and axes
+#: xbmc/games/controllers/dialogs/GUIDialogButtonCapture.cpp
 msgctxt "#35014"
 msgid "Some controllers have buttons and axes that interfere with mapping. Press these now to disable them:[CR]%s"
 msgstr ""
 
-#. Name of a controller button, e.g. Button 1. %d - button index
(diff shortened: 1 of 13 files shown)


garbear:
Rebased after merge of #11641. Still needs some work to reduce code.

garbear:
Work finished - common code has been refactored into a base class. Also improved some text and updated the screenshot

NOTE: Contains the two codebase improvement commits found in #11645

jenkins build this please

garbear:
jenkins build this please

garbear:
jenkins build this please

garbear:
Rebased on #11645. jenkins build this please

garbear:
jenkins build this please

garbear:
jenkins build this please

garbear:
build failures look unrelated, but just in case.. jenkins build and merge

PR #11625, opened 2017-02-06
[touchscreen/pictures] - allow to back out from fullscreen picture mo…

Memphiz:
…de by mapping longpress
- straight forward - currently the touchscreen.xml prevents anybody to escape from picture fullscreen. This adds the ability to escape via single finger longpress

### DIFF
--- a/system/keymaps/touchscreen.xml
+++ b/system/keymaps/touchscreen.xml
@@ -88,6 +88,7 @@
       <swipe direction="left" pointers="2">NextPicture</swipe>
       <tap pointers="2">Pause</tap>
       <tap pointers="3">Info</tap>
+      <longpress>Back</longpress>
     </touch>
   </SlideShow>
   <ScreenCalibration>


system/keymaps/touchscreen.xml
```diff
@@ -84,8 +84,8 @@
     <touch>
       <zoom>ZoomGesture</zoom>
       <rotate>RotateGesture</rotate>
-      <swipe direction="right" pointers="2">PreviousPicture</swipe> <!-- right/left are flipped here to make picture flipping feel more natural on touch screens. Two pointers are used in order to avoid conflicting with the panning gesture on zoomed-in images. -->
-      <swipe direction="left" pointers="2">NextPicture</swipe>
+      <swipe direction="right">PreviousPicture</swipe> <!-- right/left are flipped here to make picture flipping feel more natural on touch screens. Two pointers are used in order to avoid conflicting with the panning gesture on zoomed-in images. -->
```
Rechi: `Two pointers are used in order to avoid conflicting with the panning gesture on zoomed-in images.` should also be removed from the comment

Memphiz: Thx for the hint - there is indeed a conflict ... will think About something else ...

PR #11631, opened 2017-02-07
Update SSE3 check to match with the others

anthonyryan1:
SSE3 is in /proc/cpuinfo, but under an alternative name

## Description
One of the comments in FindSSE was incorrect, this corrects it and explains why it's unusual.

## Motivation and Context
This doesn't change anything meaningful. Some systems (including my own) will now have SSE3 properly enabled in the build system when it was disabled previously. This is mostly to correct a
misconception within the code.

## How Has This Been Tested?
Ran cmake to confirm that SSE3 was still properly enabled on my Linux based system while previously it was not. Ensured kodi built and ran properly in 2 unique configurations and systems.

## Types of change

## Checklist:

### DIFF
--- a/cmake/modules/FindSSE.cmake
+++ b/cmake/modules/FindSSE.cmake
@@ -14,13 +14,10 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
      string(COMPARE EQUAL "sse2" "${_SSE_THERE}" _SSE2_TRUE)
      CHECK_CXX_ACCEPTS_FLAG("-msse2" _SSE2_OK)
 
-     # /proc/cpuinfo apparently omits sse3 :(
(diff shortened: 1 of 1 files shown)


hudokkow:
Thanks.
jenkins build this please

koenkooi:
Does this (still?) allow setting it manually when cross-compiling to x86_64?

anthonyryan1:
I expect it should. The code is actually more in line with how SSE2, SSE4.1 etc are enabled so if they can currently be manually enabled via cmake, this will behave identically.

If you'd like me to test this, I can, but the only non-x86 machine I have is rather underpowered, so between setting up the cross-compile environment and building Kodi it'll probably need to run for 4 or 5 days to build and test.

PR #11632, opened 2017-02-07
Fix musicbrainz resolution with python based scrapers

notspiff:
This fixes mbrainz resolution using python based scrapers. Previously it tried to load the py file as xml, then crashed and burned.

@ronie

### DIFF
--- a/addons/metadata.demo.artists/demo.py
+++ b/addons/metadata.demo.artists/demo.py
@@ -46,6 +46,9 @@ if action == 'find':
     liz.setProperty('artist.genre', 'classical / jazz')
     liz.setProperty('artist.born', '2012')
     xbmcplugin.addDirectoryItem(handle=int(sys.argv[1]), url="/path/to/artist2", listitem=liz, isFolder=True)
+elif action == 'resolveid':
+    liz=xbmcgui.ListItem(path='/path/to/artist2', offscreen=True)
+    xbmcplugin.setResolvedUrl(handle=int(sys.argv[1]), succeeded=True, listitem=liz)
 elif action == 'getdetails':
     url=urllib.unquote_plus(params["url"])
     print 'Artist with url %s' %(url)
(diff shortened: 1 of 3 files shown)


ronie:
thx! works as expected.

i do have some additional questions on the scraping process:
http://forum.kodi.tv/showthread.php?tid=306218
^ ping @DaveTBlake as well

ronie:
jenkins build this please

@notspiff is that unrelated commit supposed to go in as well?

notspiff:
it quells yet another warning introduced in the pvr code. if you insist i
can submit it separately but seemed like more hassle than it was worth.

>

MartijnKaijser:
jenkins build this please

DaveTBlake:
Late in the day I know but the `CScraper::ResolveIDToUrl` change seems an odd thing to do.

For the XML music scrapers  the "ResolveIDToUrl" function called by `CScraper::ResolveIDToUrl` is used to create the URL for that artist or album mbid, it gets put into the scraper list of albums or artists, and is subsequently used to get the details. It does not make any server requests, or validate the mbid in any way, it just does some string manipulation.

The Python music scrapers "getdetails" action, called from PythonDetails, is simply passed the mbid. So why need the Python scraper to have a "resolveid" action to put a fake album or artist into the list so that a count check avoids lookup by name etc. ?

It makes more sense to me to change `CMusicInfoScanner::DownloadAlbumInfo` and `DownloadArtistInfo` to check `preferredScraper->IsPython()` and act accordingly. Or am I missing something?

akva2:
as i said in the description - i have gone through stupid hoops to operate within the current scraper system. it does not make sense, it's just to map to the old.

DaveTBlake:
Fair enough :)

I have also discovered that XML music scrapers "ResolveIDToUrl" is not just string parsing. It takes the Musicbrainz *release* id that Kodi has from music file tags (strMusicBrainzAlbumID in the album table), and looks up  the Musicbrainz *release group* id that is needed to look up the album in Fanart.TV and TADB. Python scraper needs to do the same.

PR #11638, opened 2017-02-08
refactor speed up/down range

Rechi:
## Description
move the hardcoded values for minimal speed and maximal speed into functions

## Motivation and Context
The same two float values were hardcoded in 11 different lines. 

## How Has This Been Tested?
Runtime tested

## Types of change

## Checklist:

### DIFF
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -7164,7 +7164,7 @@ bool CGUIInfoManager::GetBool(int condition1, int contextWindow, const CGUIListI
     case PLAYER_PLAYING:
       {
         float speed = g_application.m_pPlayer->GetPlaySpeed();
-        bReturn = (speed >= 0.75 && speed <= 1.55);
+        bReturn = (speed >= IPlayer::MinTempo() - 0.05 && speed <= IPlayer::MaxTempo() + 0.05);
       }
       break;
     case PLAYER_PAUSED:
@@ -7174,7 +7174,7 @@ bool CGUIInfoManager::GetBool(int condition1, int contextWindow, const CGUIListI
       bReturn = g_application.m_pPlayer->GetPlaySpeed() < 0;
       break;
     case PLAYER_FORWARDING:
-      bReturn = g_application.m_pPlayer->GetPlaySpeed() > 1.5;
+      bReturn = g_application.m_pPlayer->GetPlaySpeed() > IPlayer::MaxTempo() + 0.05;
       break;
     case PLAYER_REWINDING_2x:
       bReturn = g_application.m_pPlayer->GetPlaySpeed() == -2;
@@ -7221,7 +7221,7 @@ bool CGUIInfoManager::GetBool(int condition1, int contextWindow, const CGUIListI
     case PLAYER_IS_TEMPO:
       {
         float speed = g_application.m_pPlayer->GetPlaySpeed();
-        bReturn = (speed >= 0.75 && speed <= 1.55 && speed != 1);
+        bReturn = (speed >= IPlayer::MinTempo() - 0.05 && speed <= IPlayer::MaxTempo() + 0.05 && speed != 1);
       }
       break;
     case PLAYER_RECORDING:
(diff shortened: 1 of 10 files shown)


xbmc/GUIInfoManager.cpp
```diff
@@ -7174,7 +7174,7 @@ bool CGUIInfoManager::GetBool(int condition1, int contextWindow, const CGUIListI
       bReturn = g_application.m_pPlayer->GetPlaySpeed() < 0;
       break;
     case PLAYER_FORWARDING:
-      bReturn = g_application.m_pPlayer->GetPlaySpeed() > 1.5;
+      bReturn = g_application.m_pPlayer->GetPlaySpeed() > IPlayer::MaxPlaySpeed();
```
FernetMenta: this is a change of behavior

xbmc/GUIInfoManager.cpp
```diff
@@ -7954,7 +7954,7 @@ std::string CGUIInfoManager::GetMultiInfoLabel(const GUIInfo &info, int contextW
   {
     std::string strTime;
     float speed = g_application.m_pPlayer->GetPlaySpeed();
-    if (speed < 0.8 || speed > 1.5)
```
FernetMenta: changes behavior

xbmc/GUIInfoManager.cpp
Rechi: should I replace it with `IPlayer::MaxPlaySpeed() + 0.05`?

FernetMenta: better change a couple of things:

- this feature is referred to as tempo, not playspeed
- min / max are currently 0.8 / 1.5
- the diff of .5 is there in order to deal with rounding errors, would be nice if the code reflects this

Rechi:
updated

xbmc/GUIInfoManager.cpp
Rechi: I know setting this to `1.55` would changed behavior, but I don't understand why rounding errors must not be considered here.

FernetMenta: > but I don't understand why rounding errors must not be considered here.

me neither. I think they should. so it changes behaviour into the right direction :)

Rechi: actually I've update `IPlayer::MaxPlaySpeed()` to return `1.5`, so should I add `+ 0.05` like I've done at the other places?

xbmc/GUIInfoManager.cpp
Rechi: should rounding errors also be considered here?

xbmc/GUIInfoManager.cpp
FernetMenta: sounds reasonable. the difference between 1.5x and ff is that on ff the indicators stay on screen.

xbmc/GUIInfoManager.cpp
FernetMenta: yes, I think so

FernetMenta:
cleanup brings some issues to surface. I think you can add the label "fix" to this pr. well done

Paxxi:
Mark the methods constexpr as well, compiler should basically treat them as defines in that case

Rechi:
must there still be something marked with `const` when marking the function as `constexpr`?

da-anda:
hmm, instead of hardcoded values in 11 places we now have the same logic duplicated in 11 different places (`IPlayer::MinTempo() - 0.05`). Sorry for the probably dump question, but if every usecase of `IPlayer::MinTempo()` always will get 0.05 substracted, why not simply change the initial value of it? Same ofc with MaxTempo().
Also, what about having a function like "IPlayer::isWithinPlaySpeedBoundaries(playSpeed)" instead of having to duplicate this `playSpeed >= IPlayer::MinTempo() - 0.05 && playSpeed <= IPlayer::MaxTempo() + 0.05` logic in dozens of places?

FernetMenta:
> Also, what about having a function like "IPlayer::isWithinPlaySpeedBoundaries(playSpeed)"

not a bad idea, but

- the correct term is tempo not playspeed
- should not be a static method of IPlayer, paplayer only supports tempo 1, external player does not support tempo either

Rechi:
have update with 3 new functions
- IsPlayingTempo
- MinTempo
- MaxTempo

Now I've an issue with getting the Min/MaxTempo into this line https://github.com/xbmc/xbmc/blob/958d453/xbmc/cores/VideoPlayer/VideoPlayerAudio.cpp#L41. Maybe anyone has an idea.

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3672,6 +3677,16 @@ bool CVideoPlayer::SupportsTempo()
   return m_canTempo;
 }
 
+const float CVideoPlayer::MinTempo() const
```
FernetMenta: this is not const, actually this can change during playback

FernetMenta:
Now this gets beyond a simple refactoring so we can change more things to get this right. VP moves away from the old polling model. That is that GUI bothers player with many calls and locks every cycle to get data that have not changed. playspeed does even lack proper locking.

Add playspeed to CProcessInfo and push updates also to CDataCacheCore. GUI can poll CDataCacheCore instead of player. VIdeoPlayerAudio has a ref to ProcessInfo so it gets access to required info.

Rechi:
@FernetMenta is this superseded by #12470?

FernetMenta:
yes, it is

PR #11641, opened 2017-02-08
[controller dialog] Ignore delayed axes

garbear:
I've identified some code that can be removed from #11623. Until that's ready, this fix just ignores the erratic input.


garbear:
jenkins build this please

PR #11645, opened 2017-02-09
Improve keyboard handling during button mapping

garbear:
Right now, if Kodi receives a keyboard press while mapping a button, it will unconditionally abort the prompt. Empirically, I've observed users becoming confused when they press a direction but the focus doesn't move. Also, I've seen users press A and become confused as to why a "select" button caused a "cancel" action.

It feels like a sin including Key.h every time I want a couple actions, so I also included a commit separating CAction from CKey.

## Types of change


garbear:
@MartijnKaijser would you consider UX improvements a fix to be backported?

Paxxi:
Can't use line comments on the phone. 
Looks good overall. 
Namespace KEYBOARD should be in namespace KODI, many older don't do this but new namespace should so we get a proper hierarchy. 

Cpp file shouldn't use namespace KEYBOARD but also be wrapped, same as header file.

garbear:
> Cpp file shouldn't use namespace KEYBOARD but also be wrapped, same as header file.

No symbols are declared in the .cpp file, so this shouldn't matter, right?

Paxxi:
No I don't know of a compile time reason to require it really. Part style and afaik it can mess with static analysis tooling. 

But we specifically require no namespace indentation to make this scenario simple. I leave it up to you to decide. 

Sticking it into the Kodi namespace is a requirement though 😀

garbear:
KODI namespace now has a few new friends to play with

Paxxi:
Thanks, I'm happy :)

garbear:
jenkins build this please

LS80:
@garbear This appears to have inadvertently removed the action ids from the Python API xbmcgui module (e.g. `xbmcgui.ACTION_MOVE_DOWN`).

I think the new `input/ActionIDs.h` needs to be included in `interfaces/swig/AddonModuleXbmcgui.i`.

PR #11648, opened 2017-02-09
Simplify new controller dialog text

garbear:
## Description

When an unknown controller is first detected, a dialog is shown. ATM this dialog is an ugly wall of text. This PR simplifies the dialog and adds useful information - the controller name.

Before:

new controller - complicated

The "No" button is highlighted because I just want the ugly text to go away.

After:

new controller - simple

The "Yes" button is highlighted because yes, I do want to setup my controller.

## Motivation and Context
Observation of a non technical user.

## Types of change

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -16754,13 +16754,13 @@ msgstr ""
 #. Title of dialog to ask users if they would like to configure a new controller
 #: xbmc/input/joysticks/dialogs/GUIDialogNewJoystick.cpp
 msgctxt "#35011"
-msgid "New controller detected"
+msgid "Unknown controller detected"
 msgstr ""
 
-#. Text of dialog to ask users if they would like to configure a new controller
+#. Text of dialog for new controllers. %s - controller name
 #: xbmc/input/joysticks/dialogs/GUIDialogNewJoystick.cpp
 msgctxt "#35012"
-msgid "A new controller has been detected. Configuration can be done at any time in \"Settings -> System Settings -> Input\". Would you like to configure it now?"
+msgid "Would you like to setup \"%s\"?"
 msgstr ""
 
 #empty string with id 35013
--- a/xbmc/input/joysticks/dialogs/GUIDialogNewJoystick.cpp
+++ b/xbmc/input/joysticks/dialogs/GUIDialogNewJoystick.cpp
@@ -22,9 +22,11 @@
 #include "ServiceBroker.h"
 #include "guilib/GUIComponent.h"
 #include "guilib/GUIWindowManager.h"
+#include "guilib/LocalizeStrings.h"
 #include "guilib/WindowIDs.h"
 #include "messaging/helpers/DialogHelper.h"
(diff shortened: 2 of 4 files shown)


da-anda:
hmm, didn't we have a discussion back then that resulted in this long version? I personally prefer having too much information instead of too less. And because the mapping is quite a bit hidden in our settings.

Btw, what was the reason again why we even need this dialog? I thought we ship with default mappings, so that it's plug'n'play anyways? If the main reason for this dialog is to configure a controller for all the different gaming systems, then it's IMO the wrong approach.

garbear:
any objections? jenkins build this please

da-anda:
if this dialog would be the first thing I see after startup, regardless of how long the message is, I'd already question the software. So if this dialog is not mandatory to use Kodi itself with that controller, we should not display it and only bring it up when a user actually needs a platform specific mapping and doesn't have it. Or after the first X button presses after startup of Kodi.

Maybe the following would be a sane approach:
- on startup don't show this dialog at all
- only show dialog if user is pressing X number of buttons in a row (detecting user is testing functionality) or when controller is plugged while Kodi is running
- also show dialog if a user tries to play a game and has not mapping for this platform

This would have the benefit that users only get prompted with this dialog when needed and also eliminate the requirement for the long dialog message, because users no longer have to go to the according rather hidden settings section to manually start a mapping in case they dismissed the dialog before.
What do you think?

garbear:
haha sorry I clicked "edit" to quote your comment... and then pasted my reply into your comment box! at least the points I address are preserved.

garbear:
> if this dialog would be the first thing I see after startup, regardless of how long the message is, I'd already question the software.

A very good point, fortunately this is not the case.

> So if this dialog is not mandatory to use Kodi itself with that controller, we should not display it and only bring it up when a user actually needs a platform specific mapping and doesn't have it. Or after the first X button presses after startup of Kodi.

Currently, the dialog is only shown when a (digital) button is pressed and the physical controller isn't mapped to ANY controller profiles. If the user doesn't select Yes, then the dialog isn't shown again.

> - only show dialog if user is pressing X number of buttons in a row (detecting that user is testing if controller is working by hammering random buttons)

Currently, this count is 1 button. How many buttons in a row do you think should bring up the dialog?

> - also show dialog if a user tries to play a game and has no mapping for this platform

As the button mapper approaches 100% coverage, this won't be an issue. Currently, if the controller can navigate the GUI, the button mapper is smart enough to map it to any game.

da-anda:
thanks for the explanations and sorry for asking so many questions. Seems a lot already changed since the last time I tested the controller stuff. In this case, change seems fine and sorry for the noise.

As for the button press threshold to trigger the dialog: I leave this up to you to change it. My idea with it was to avoid showing the dialog if a user accicently hits a button when putting the controller away, sits down on it or whatever. And because I think that every user testing functionality will hit more buttons and axis than just one, I had probably set it to something like 3, which seems to be low enough to trigger early but should also prevent the dialog to pop up on accidental key presses. But as said, up to you.

garbear:
Closing because meh

PR #11649, opened 2017-02-10
Add browse for subtitle action

notspiff:
This allows mapping a key to browse for subtitle in video fullscreen window. Ref forum.kodi.tv/showthread.php?tid=306428&pid=2521764

### DIFF
--- a/xbmc/input/ActionIDs.h
+++ b/xbmc/input/ActionIDs.h
@@ -55,6 +55,7 @@
 #define ACTION_SHOW_OSD               24 //!< show/hide OSD. Can b used in videoFullScreen.xml window id=2005
 #define ACTION_SHOW_SUBTITLES         25 //!< turn subtitles on/off. Can b used in videoFullScreen.xml window id=2005
 #define ACTION_NEXT_SUBTITLE          26 //!< switch to next subtitle of movie. Can b used in videoFullScreen.xml window id=2005
+#define ACTION_BROWSE_SUBTITLE       247 //!< Browse for subtitle. Can be used in videofullscreen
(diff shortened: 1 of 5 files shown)


da-anda:
would it make sense to also expose this via JSON-RPC? Might also be useful for smartphone remotes f.e.. What do you think?

notspiff:
I don't see how that is related to this pr. This adds a mappable action so you do not have to hop through osd to browse.

Also i think json already supports pushing subtitles to the player.

da-anda:
from what I have seen in the JSON-RPC methods schema you can only set a specific subtitle, but not trigger the builtin search dialog. I only mentioned that it also might be nice to have the option to trigger this builtin search dialog via JSON-RPC, which would expose the same functionality for physical remotes (via keymaps) to smartphone remotes/webinterfaces. It's no blocker, but would IMO fit within this PR. The only thing to consider when exposing this to JSON-RPC is that this method needs to be called from GUI thread I suppose (I'm not that experienced with threading).

PS: there seems to be a merge conflict atm

notspiff:
so you want a json action to trigger the dialog on screen? this should work just fine through the generic Action input operation.

have rebased (a refactoring pr from garbear landed and caused the conflict).

da-anda:
ah, you mean via `Input.ExecuteAction`? Totally missed that. Great. Thanks.

gade01:
Is it on purpose this function can't be called from VideoOSD.xml?

I can only get it working if mapped to a keymap and called in VideoFullScreen.xml

notspiff:
yes.

gade01:
Ok, thanks.

PR #11650, opened 2017-02-10
explicitly pass rrefs for move semantics

notspiff:
should quell PASS_BY_VALUE warnings from coverity

also zero initialize member struct to quell a (harmless) UNINIT_CTOR warning
requested by @fritsch. note that these issues exist in every other add-on class as well.


fritsch:
as mentioned yesterday: that was no request. Thanks for this fixup.

jenkins build this please

notspiff:
jenkins seems to be out of order due to issues on the windows slave. been spinning for half a day

PR #11651, opened 2017-02-10
Improve performance for python addons using xbmcgui.ListItem

popcornmix:
## Description
Improve performance for python addons using xbmcgui.ListItem

## Motivation and Context
See: http://trac.kodi.tv/ticket/17304#comment:28

## How Has This Been Tested?
In last two Milhouse nightly builds.
Myself and trac ticket reported have measured significant performance increases

## Types of change

## Checklist:

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -2745,7 +2745,9 @@ void CApplication::FrameMove(bool processEvents, bool processGUI)
       CSingleExit ex(g_graphicsContext);
       m_frameMoveGuard.unlock();
       // Calculate a window size between 2 and 10ms, 4 continuous requests let the window grow by 1ms
-      unsigned int sleepTime = std::max(static_cast<unsigned int>(2), std::min(m_ProcessedExternalCalls >> 2, static_cast<unsigned int>(10)));
+      // WHen not playing video we allow it to increase to 80ms
+      unsigned int max_sleep = m_pPlayer->IsPlayingVideo() && !m_pPlayer->IsPausedPlayback() ? 10 : 80;
+      unsigned int sleepTime = std::max(static_cast<unsigned int>(2), std::min(m_ProcessedExternalCalls >> 2, max_sleep));
       Sleep(sleepTime);
       m_frameMoveGuard.lock();
       m_ProcessedExternalDecay = 5;
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -2748,8 +2748,9 @@ void CApplication::FrameMove(bool processEvents, bool processGUI)
       unsigned int sleepTime = std::max(static_cast<unsigned int>(2), std::min(m_ProcessedExternalCalls >> 2, static_cast<unsigned int>(10)));
       Sleep(sleepTime);
       m_frameMoveGuard.lock();
+      m_ProcessedExternalDecay = 5;
     }
-    else
+    if (m_ProcessedExternalDecay && --m_ProcessedExternalDecay == 0)
       m_ProcessedExternalCalls = 0;
   }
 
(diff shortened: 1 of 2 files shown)


popcornmix:
Ping @FernetMenta 
https://github.com/xbmc/xbmc/pull/9748 introduced the performance regression.

I suspect a better solution to this issue will be found for Leia, but this is a low risk fix that could be backported to Krypton.

xbmc/Application.cpp
```diff
@@ -296,6 +296,7 @@ CApplication::CApplication(void)
   , m_fallbackLanguageLoaded(false)
   , m_WaitingExternalCalls(0)
   , m_ProcessedExternalCalls(0)
+  , m_ProcessedExternalDecay(0)
```
FernetMenta: please use c++11 style for init of members

popcornmix: I've switched them all. Is that okay?

FernetMenta: I meant doing init by assignment in the header. that is much better than having to keep the right order in the constructor.

popcornmix: Ah like this?

FernetMenta: yes, thanks

DaveTBlake:
Do you think the Python regression could also be the cause of music playback stuttering when returning back to the visualisation screen? User reports it clearly here http://forum.kodi.tv/showthread.php?tid=306568.  Happens with visualisation, but not genreal lib navigation.

I going to set  a build going for him test

FernetMenta:
Completely different issue. Most likely related to the visualization. Note that what you call python regression would show effect in the opposite direction, means cure stuttering in a/v playback. It's the outside world like python and addons that disturbs the core.

FernetMenta:
jenkins build and merge

FernetMenta:
jenkins is out of order

basrieter:
Will this also be merged into the Krypton branch and be in 17.1 ?

popcornmix:
Hopefully. Keep an eye on https://github.com/xbmc/xbmc/pull/11660

basrieter:
I guess I should wake up and should have looked better at the text 

> popcornmix referenced this pull request 6 hours ago
Open
Improve performance for python addons using xbmcgui.ListItem #11660 

Sorry for that.

posti85:
We all hope is merged soon xD it's known our kodi addon are running very slow in our Raspberries. Thank popcornmix for the code and basrieter for the research!

MartijnKaijser:
@popcornmix could you make a backport PR?

rembound:
I have a twitch.tv plugin that displays the video stream in a windowed videowindow `xbmc.Player().play(streamurl, windowed=True)` and the chat in a list as an overlay. This list is continually updated with the latest chat messages `messagelist.addItem(xbmcgui.ListItem(line))`.

It always worked flawlessly on my RPI3, but since the Kodi 17 update, the video stutters when adding items to the list.

The following code is run in a thread and is a minimal case. Every second, a message is added to the list. Every few seconds, the video that is playing will stutter.
```
messagelist.addItem(xbmcgui.ListItem("the quick brown fox jumps over the lazy dog"))
listsize = messagelist.size()
if listsize > CHAT_MAXLINES:
    messagelist.removeItem(0)
messagelist.selectItem(messagelist.size()-1)

time.sleep(1)
```

Could this patch also be applied to fix performance issues during video playback?

popcornmix:
As this has only just been merged then I assume your issue has nothing to do with this PR.
(this PR actually does nothing different when video is playing).

Open a [trac](http://trac.kodi.tv) ticket if you have found a regression on Krypton compared to Jarvis.

PR #11654, opened 2017-02-10
Controller dialog: Rename to Controller setup

garbear:
From a UX perspective, the shorter synonym is easier to read when there's small text on a large screen.

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -16792,7 +16792,7 @@ msgstr ""
 #. Error dialog text when the controller dialog is opened but peripheral add-ons are disabled
 #: xbmc/peripherals/bus/virtual/PeripheralBusAddon.cpp
 msgctxt "#35018"
-msgid "Controller configuration depends on a disabled add-on. Would you like to enable it?"
+msgid "Controller setup depends on a disabled add-on. Would you like to enable it?"
 msgstr ""
 
 #. Label of the button to fix the bug where buttons are skipped in the controller dialog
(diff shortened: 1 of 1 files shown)


garbear:
Any objections?

MartijnKaijser:
do we use "configuration" in other places as well cause we should keep consistency

MartijnKaijser:
@jjd-uk or any other native english speaker what's your take on this to make it all "setup"?

jjd-uk:
Seems perfectly ok to use "setup" instead of "configuration" to me. Also as far as I can see we don't use "configuration" in a similar way anywhere else, so changing to "setup" shouldn't introduce any language use inconsistency.

garbear:
I passed out after 5 hours of bugfixing on the plane, so didn't get to this. Remind me to address by beta 4.

phunkyfish:
@garbear you still want to merge this or put it on the backburner?

garbear:
Let's keep it open, I reconsider finishing this every once in a while

howie-f:
@garbear is this meant to change any "configuration" anywhere in the gui to "setup" ? tell me if i should deliver a shiny commit. actually i'm bored from not hangin out in the park.

garbear:
Codewise, yes this is the way forward, but I don't know how to reach consensus on GUI wording. The authority to edit large amounts of existing strings should stem from a comprehensive view of the UX we're trying to accomplish. Before changing code then, we should define the UX goals we want to achieve, and then proceed with this and possibly other PRs that tackle these goals.

howie-f:
ok, then feel free to ping of there's news on that somehow ;)

PR #11317, opened 2016-12-30
Don't require an optical drive for stub file handling

Woerd88:
Partially Reverted: xbmc@731a7b1
Forum discussion: http://forum.kodi.tv/showthread.php?tid=178014&page=2&highlight=stub

With this change ppl can check their stub information on systems that don't have an optical drive.
(Raspberry PI's, Android Tablets etc.)

Edit: Sorry about the lame default title, i did not create it properly.
I am still learning to work with git.

## Description

Removed HasOpticalDrive check so we can see the custom information from the stub file.

## Motivation and Context

If fixes the issue where you can't "play" a stub file on a device where no optical drive is present.
Therefor you can't view custom stubfile messages, e.g. "Enter disc 46".

## How Has This Been Tested?

Not at all! I am a developer but have no expierence with building kodi (so far)
As @da-anda proposed, maybe someone can launch an automatic build?
I am having difficulty reading/following the checklist because of the server/wiki issues. 
(Cloudfare switch)

## Types of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [ ] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #11655, opened 2017-02-10
An optical drive is not required for proper stub file handling

dml33:
Supersedes: https://github.com/xbmc/xbmc/pull/11317
Forum discussion: http://forum.kodi.tv/showthread.php?tid=178014&page=2&highlight=stub
Media stub Wiki: http://kodi.wiki/view/media_stubs

## Description
This changes allows stub information (title and message) given by a media stub file be also displayed in those systems which do not have optical drives, such as Raspberry PI's, Android Tablets, NUC, etc.

## Motivation and Context
Since Kodi v12 Frodo, if Kodi detects that the platform where it is running does not have an optical drive, the information of the stub file is not displayed. Instead, a dialog box with only a "You need an optical disc drive to play this video" message was shown.

This patch ensures that the specific instructions that the user has already written in the stub file in order to find the offline media are displayed in any case.

## How Has This Been Tested?
I built a win32 Kodi v18 version with this patch.
I tested it ina laptop with an optical drive and a NUC, without it.
The content of the dialogs were agreed among the people in the Feature Request forum.

- Dialog when the platform has an optical drive (no change):
[With optical drive](http://i.imgur.com/KsZqgeX.jpg)
- Dialog when the platform does not have an optical drive (new):
[Without optical drive](http://i.imgur.com/EJG3jJd.jpg)

## Types of change

## Checklist:


Rechi:
Is there any reason you don't move the logic of creating the dialog into https://github.com/xbmc/xbmc/blob/cbfa964/xbmc/dialogs/GUIDialogPlayEject.cpp#L89-L135 and do the check ` if (g_mediaManager.HasOpticalDrive())` there?

da-anda:
@Rechi dml33 is all new to this and suggestions are welcome.

dml33:
Hi.

Thanks for the comments.

The main reason is that a Play/Eject dialog box is not suitable for this case. I mean, there is no optical drive to eject its tray and the movie/video/music cannot be played, since it is actually offline. I think that an informative dialog with the custom message and an OK button is much more suitable.

Besides, IMHO, the code for the dialogs should be as much standard/simple as possible (just message + buttons), so customizations/particular cases should be avoided in such a low level layer, but be managed at a higher level.

As da-anda says, any suggestions are welcome.

Thanks.

Rechi:
A more detailed instruction of my suggestion
- remove https://github.com/xbmc/xbmc/blob/cbfa964/xbmc/Application.cpp#L3257-L3270, except line 3264
- move https://github.com/xbmc/xbmc/blob/cbfa964/xbmc/dialogs/GUIDialogPlayEject.cpp#L96-L100 to line 121
- surround https://github.com/xbmc/xbmc/blob/cbfa964/xbmc/dialogs/GUIDialogPlayEject.cpp#L121-L134 inclusive the moved lines into `if (g_mediaManager.HasOpticalDrive())`
- maybe also surround it with an `#ifdef HAS_DVD_DRIVE` like it is done in xbmc/Application.cpp, but I'm not sure if that is needed
- add an else block to the new if with `CGUIDialogOK::ShowAndGetInput(...)` & `return true`

I hope this is understandable. If you have further questions just ask.

dml33:
Hi.

From your explanation, I understand that the `if` of https://github.com/xbmc/xbmc/blob/cbfa964/xbmc/Application.cpp#3621 should stay. It will return true if there is an optical drive, but it will return false if not. Then, the else with have the block I wrote with the `CGUIDialogOK::ShowAndGetInput`.

This may simplify a little the `#ifdef HAS_DVD_DRIVE` section of https://github.com/xbmc/xbmc/blob/cbfa964/xbmc/Application.cpp#L3257-L3267 but if there is no optical drive, the stub file information will be obtained twice: firstly in https://github.com/xbmc/xbmc/blob/cbfa964/xbmc/dialogs/GUIDialogPlayEject.cpp#L105-L115, which will not be useful, and later in the block of my PR for the `CGUIDialogOK`.

Rechi:
> From your explanation, I understand that the `if` of https://github.com/xbmc/xbmc/blob/cbfa964/xbmc/Application.cpp#L3621 should stay. It will return true if there is an optical drive, but it will return false if not. Then, the else with have the block I wrote with the `CGUIDialogOK::ShowAndGetInput`.

I think you wanted to point me to a different line. The content of the line you pointed me is ` if(m_bPlaybackStarting)`.
Maybe you have understand something wrong, but there won't show up a dialoge twice.


I was thinking of something like this.
Application.cpp
```
    if (item.IsVideo())
      CUtil::ClearSubtitles();
  }

  if (item.IsDiscStub())
    return MEDIA_DETECT::CAutorun::PlayDiscAskResume() ? PLAYBACK_OK : PLAYBACK_FAIL;

  if (item.IsPlayList())
    return PLAYBACK_FAIL;
```

GUIDialogPlayEject.cpp
```
  // Use the label for Line 1 if not defined
  if (strLine1.empty())
    strLine1 = item.GetLabel();

#ifdef HAS_DVD_DRIVE
  if (g_mediaManager.HasOpticalDrive())
  {
    // Create the dialog
    CGUIDialogPlayEject * pDialog = (CGUIDialogPlayEject *)g_windowManager.
      GetWindow(WINDOW_DIALOG_PLAY_EJECT);
    if (!pDialog)
      return false;

    // Setup dialog parameters
    pDialog->SetHeading(CVariant{219});
    pDialog->SetLine(0, CVariant{429});
    pDialog->SetLine(1, CVariant{std::move(strLine1)});
    pDialog->SetLine(2, CVariant{std::move(strLine2)});
    pDialog->SetChoice(ID_BUTTON_PLAY - 10, 208);
    pDialog->SetChoice(ID_BUTTON_EJECT - 10, 13391);
    if (uiAutoCloseTime)
      pDialog->SetAutoClose(uiAutoCloseTime);

    // Display the dialog
    pDialog->Open();

    return pDialog->IsConfirmed();
  }
  else
#endif
  {
    CGUIDialogOK::ShowAndGetInput(...);
    return true;
  }
}
```

dml33:
Hi. 

Sorry, I wanted to point to you to https://github.com/xbmc/xbmc/blob/cbfa964/xbmc/Application.cpp#L3261, instead.

If you want to keep the logic at `GUIDialogPlayEject.cpp`, is it not strange/incoherent to call a Dialog Play/Eject, but at the end, the result could be just a Dialog OK, which has a completely different behaviour?

If it as you suggest, the `MEDIA_DETECT::CAutorun::PlayDiscAskResume()` in the `Application.cpp` cannot be called if there is no optical drive. Then, it should go inside `GUIDialogPlayEject.cpp` as follows:

`Application.cpp`

```
if (item.IsDiscStub())
  return (CGUIDialogPlayEject::ShowAndGetInput(item))
```

`GUIDialogPlayEject.cpp`

```
  // Use the label for Line 1 if not defined
  if (strLine1.empty())
    strLine1 = item.GetLabel();

#ifdef HAS_DVD_DRIVE
  if (g_mediaManager.HasOpticalDrive())
  {
    // Create the dialog
    CGUIDialogPlayEject * pDialog = (CGUIDialogPlayEject *)g_windowManager.
      GetWindow(WINDOW_DIALOG_PLAY_EJECT);
    if (!pDialog)
      return false;

    // Setup dialog parameters
    pDialog->SetHeading(CVariant{219});
    pDialog->SetLine(0, CVariant{429});
    pDialog->SetLine(1, CVariant{std::move(strLine1)});
    pDialog->SetLine(2, CVariant{std::move(strLine2)});
    pDialog->SetChoice(ID_BUTTON_PLAY - 10, 208);
    pDialog->SetChoice(ID_BUTTON_EJECT - 10, 13391);
    if (uiAutoCloseTime)
      pDialog->SetAutoClose(uiAutoCloseTime);

    // Display the dialog
    pDialog->Open();
    if (pDialog->IsConfirmed())
      return (MEDIA_DETECT::CAutorun::PlayDiscAskResume() ? PLAYBACK_OK : PLAYBACK_FAIL);
    else
      return false;
  }
  else
#endif
  {
    CGUIDialogOK::ShowAndGetInput(...);
    return true;
  }
}
```

`

Rechi:
sorry had a wrong assumption

Application.cpp
```
    if (item.IsVideo())
      CUtil::ClearSubtitles();
  }

  if (item.IsDiscStub())
  {
    if (CGUIDialogPlayEject::ShowAndGetInput(item))
    {
#ifdef HAS_DVD_DRIVE
      // PlayDiscAskResume takes path to disc. No parameter means default DVD drive.
      // Can't do better as CGUIDialogPlayEject calls CMediaManager::IsDiscInDrive, which assumes default DVD drive anyway
      return MEDIA_DETECT::CAutorun::PlayDiscAskResume() ? PLAYBACK_OK : PLAYBACK_FAIL;
#endif
    }
    return PLAYBACK_OK;
  }

  if (item.IsPlayList())
    return PLAYBACK_FAIL;
```
 
and don't move `MEDIA_DETECT::CAutorun::PlayDiscAskResume` to GUIDialogPlayEject.cpp.

da-anda:
wouldn't a helper function "playStubFile" be more apropriate and have it contain the necessary logic? I'm also not a huge fan of adding stub logic into a generic dialog

dml33:
> and don't move MEDIA_DETECT::CAutorun::PlayDiscAskResume to GUIDialogPlayEject.cpp.

I do not like this, since CGUIDialogPlayEject::ShowAndGetInput returns false when there is no optical drive and there is no view of that situation in Application.cpp. In my opinion, it is confusing.

> wouldn't a helper function "playStubFile" be more apropriate and have it contain the necessary logic? I'm also not a huge fan of adding stub logic into a generic dialog.
> 
> edit: also, stub files might not be related to a disc at all, as you can also have stub files for your home video VHS collection

I agree. I like it. Let's do it this way, an auxiliary function in Application.cpp which contains the whole logic and branches to GUIDialogPlayEject.cpp or CGUIDialogOK, which will only have to open the proper dialog and display the lines which have been previously obtained.

@Rechi, what do you think?

I will prepare the commit and update the PR with it.

Rechi:
My previous goal was only not duplicating the read logic.
A helper function to achieve this would be a good idea, but I'm not sure where in which file it should located.

da-anda:
yes, cApplication is not the best place for such a function. The main class is way to overloaded already. But I also don't have a location for it at hand out of memory.

dml33:
Hi.

What about placing the auxiliary function 'playStubFile' in `storage/MediaManager.cpp`? There is a lot of stuff there related to DVD/CD/removable devices and it could fit in.

What do you think?

I cannot clean the mess I made with the commits. I am very sorry. I am new using git and I used it very much like SVN, making the final commits very unreadable and strange. I am planning to close/abandon this and submit a new and clean pull request with the final code, when we all agree with it.

Thanks.

Rechi:
At least storage/MediaManager.cpp is better then Application.cpp

If it's easier for you, create a new PR and link to this one.

Rechi:
@dml33 are you still planing to updated this PR / create a new one?

dml33:
Hi.

I am planning to create a new one and use this as a reference. I hope I can
find some free time soon.

Thanks

Razzeee:
@dml33 Could we close this for now then?

Rechi:

apostolostopalidis:
Any chance for this to revive?

PR #11656, opened 2017-02-11
[interfaces] legacy: fix format error when building with -Werror=form…

wsnipex:
…at-security

fixes PPA builds. which use -Werror=format-security by default

>‘std::vector<std::__cxx11::basic_string<char> > XBMCAddon::xbmcgui::ListItem::getStringArray(const InfoLabelValue&, const string&, std::__cxx11::string)’:
/data/src/xbmc/xbmc/interfaces/legacy/ListItem.cpp:860:176: error: format not a string literal and no format arguments [-Werror=format-security]
           throw WrongTypeException(StringUtils::Format("When using \"%s\" you need to supply a string or list of strings for the value in the dictionary", tag.c_str()).c_str());
                                                                                                                                                                                ^
cc1plus: some warnings being treated as errors

## Types of change

## Checklist:

### DIFF
--- a/xbmc/interfaces/legacy/ListItem.cpp
+++ b/xbmc/interfaces/legacy/ListItem.cpp
@@ -857,7 +857,7 @@ namespace XBMCAddon
       for (const auto& el : alt.later())
       {
         if (el.which() == second)
-          throw WrongTypeException(StringUtils::Format("When using \"%s\" you need to supply a string or list of strings for the value in the dictionary", tag.c_str()).c_str());
+          throw WrongTypeException("When using \"%s\" you need to supply a string or list of strings for the value in the dictionary", tag.c_str());
         els.emplace_back(el.former());
       }
       return els;


wsnipex:
jenkins build this please

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -857,7 +857,7 @@ namespace XBMCAddon
       for (const auto& el : alt.later())
       {
         if (el.which() == second)
-          throw WrongTypeException(StringUtils::Format("When using \"%s\" you need to supply a string or list of strings for the value in the dictionary", tag.c_str()).c_str());
+          throw WrongTypeException("%s", StringUtils::Format("When using \"%s\" you need to supply a string or list of strings for the value in the dictionary", tag.c_str()).c_str());
```
Rechi: Is there a benefit from formatting with `StringUtils::Format` and pass the result in a new format string instead of just calling `WrongTypeException` with the format string that is passed to `StringUtils::Format`?

wsnipex: I have no idea, but Stringutils has some platform specific handling and was used here before, so I didn't want to unwillingly change behavior.
I can change it to remove its use if preferred.

wsnipex:
updated to drop the stringutils call.
jenkins build this please

wsnipex:
jenkins build this please

PR #11657, opened 2017-02-11
JSON: Handle option "resume" when opening playlists

dgam1:
The issue I'm trying to fix here is that when you send a JSON query of type "Player.Open", using playlistid and position as parameters, the option "resume" gets silently ignored. To reproduce it, send these queries (with a loaded video playlist):

`{"params": {"item": {"position": 0, "playlistid": 1}, "options": {"resume": false}}, "jsonrpc": "2.0", "id": 1, "method": "Player.Open"}`

`{"params": {"item": {"position": 0, "playlistid": 1}, "options": {"resume": 50.0}}, "jsonrpc": "2.0", "id": 1, "method": "Player.Open"}`

`{"params": {"item": {"position": 0, "playlistid": 1}, "options": {"resume": {"hours": 0, "seconds": 10, "minutes": 10, "milliseconds": 10}}}, "jsonrpc": "2.0", "id": 1, "method": "Player.Open"}`

The changes I implemented are:
xbmc/interfaces/json-rpc/PlayerOperations.cpp:
In CPlayerOperations::Open when opening a playlist by id and position, it doesn't send message TMSG_MEDIA_PLAY anymore but TMSG_PLAYLISTPLAYER_PLAY, attaching as payload a pointer to a CVariant containing either a double (percentage) or an int (-1 for STARTOFFSET_RESUME or number of seconds offset).

xbmc/PlayListPlayer.cpp:
PLAYLIST::CPlayListPlayer::OnApplicationMessage, handling the message TMSG_PLAYLISTPLAYER_PLAY, now reads param2 as the playlistid and the payload pointer as the pointer to the CVariant instance containing the resume value. It will set the current playlist, if needed. Before this patch, this was done in TMSG_MEDIA_PLAY. I removed that part from the handler of the latter message since the only occurrence in the code base of using TMSG_MEDIA_PLAY to open a specific playlist at a specific position was in the JSON handler that I changed above.
Also, in CPlayListPlayer::Play the start offset and percentage both always get reset.

xbmc/Application.cpp:
Reset the property "StartPercent" from m_itemCurrentFile as otherwise the property will be set again on the original item in the playlist, once the playback ends.

With these changes resume works as expected on videos. With music it works when you supply a specific time (but the display will act as if the file is being played from the beginning), percentage and STARTOFFSET_RESUME currently get ignored by the music player, which is outside of the scope of this PR.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -3093,6 +3093,9 @@ bool CApplication::PlayFile(CFileItem item, const std::string& player, bool bRes
     if(item.GetProperty("StartPercent").isString())
       fallback = (double)atof(item.GetProperty("StartPercent").asString().c_str());
     options.startpercent = item.GetProperty("StartPercent").asDouble(fallback);
+    // Clear the property from m_itemCurrentFile so that it doesn't get set
+    // again on the original FileItem when the playback ends
+    m_itemCurrentFile->ClearProperty("StartPercent");
   }
 
   options.starttime = CUtil::ConvertMilliSecsToSecs(item.m_lStartOffset);
--- a/xbmc/PlayListPlayer.cpp
+++ b/xbmc/PlayListPlayer.cpp
@@ -361,9 +361,9 @@ bool CPlayListPlayer::Play(int iSong, std::string player, bool bAutoPlay /* = fa
     }
   }
 
-  // reset the start offset of this item
-  if (item->m_lStartOffset == STARTOFFSET_RESUME)
-    item->m_lStartOffset = 0;
+  // reset the start offset and start percent of this item
+  item->m_lStartOffset = 0;
+  item->ClearProperty("StartPercent");
 
   //! @todo - move the above failure logic and the below success logic
   //!        to callbacks instead so we don't rely on the return value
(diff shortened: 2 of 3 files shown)


Paxxi (commented):
Looks good to me
Any opinion @montellese?

Montellese:
I'm not sure I understand how the resume value is interpreted. Is it the item's index in the playlist? Or do you just resume within the first item or can it handle the fact where the given resume value exceeds the first item(s) in the playlist?

Thanks for the detailed description of the change. Makes it much easier to review.

dgam1:
I changed the code in the PlayListPlayer to improve readability.
The resume value will be set only on the one item at the specified position of the specified playlist. Then playback will start at that position ( [L760](https://github.com/dgam1/xbmc/blob/c667200cee38f0bd6998e48e2df6104a70abaa34/xbmc/PlayListPlayer.cpp#L760) ). If the value is higher than the length of the item, the player will take care of it by seeking to the end of the file. The next item in the playlist will be played from the start.

What I didn't know when I submitted this PR is that FileItem's member m_lStartOffset has a completely different meaning for audio files. There it is set for the likes of cue sheet items with several songs in one file, indicating where a song begins within that file. The audio player will thus cut off everything before that value. Resuming is not provided for audio items and changing the offset will break cue sheet items. That means this PR shouldn't be merged at this time.

Having said that, I am currently working on implementing resuming on audio files where m_lStartOffset will have the same function for both video and audio. The PR for that should be ready within the next couple of days.

Rechi:
@dgam1 this needs a rebase

phunkyfish:
@dgam1 are you still interested in getting this merged?

phunkyfish:

phunkyfish:
Going to close this, can be reopened at a later date if required.

PR #11658, opened 2017-02-11
allow tag rev to be defined for rebuilds

mdeguzis:
## Description
allow tag rev to be defined for rebuilds

## Motivation and Context
If someone packaging Kodi makes a change in their debian packaging, the pkg rev always stays 0. The build script checks if TAGREV is set, but if you use:

```
PKGREV="2"
...
RELEASEV="${KODI_TAG}" \
TAGREV="${PKGREV}"
DISTS="${DIST}" \
ARCHS="${ARCH}" \
BUILDER="${BUILDER}" \
PDEBUILD_OPTS="${BUILDOPTS}" \
PBUILDER_BASE="${PBUILDER_BASE}" \
tools/Linux/packaging/mk-debian-package.sh
```
...Tag rev is not picked up. The current workaround for me, it to export TAGREV above this piece.

## How Has This Been Tested?
Tested on Arch Linux and SteamOS. Took the existing xbmc-packaging tarball, made the change and repacked it. Did an ls of the build directly as it began to build to see that the full pkg text was "kodi_17.0-Krypton~git20170211.1829-a10c5048f2-2"

## Checklist:

### DIFF
--- a/tools/Linux/packaging/mk-debian-package.sh
+++ b/tools/Linux/packaging/mk-debian-package.sh
@@ -22,6 +22,7 @@
 RELEASEV=${RELEASEV:-"auto"}
 VERSION_PREFIX=${VERSION_PREFIX:-""}
 TAG=${TAG}
+TAGREV=${TAGREV:-""}
 REPO_DIR=${WORKSPACE:-$(cd "$(dirname $0)/../../../" ; pwd)}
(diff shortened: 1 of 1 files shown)


MartijnKaijser:
@wsnipex this seems ok?

PR #11662, opened 2017-02-12
[settings] change allowed remotedelay range to inlcude zero

Kwiboo:
## Description

This change allows to set the advanced settings remotedelay value to 0. Changes the allowed range from 1-20 to 0-20.

## Motivation and Context

The current range limit of 1-20 forces [LIRC.cpp](https://github.com/xbmc/xbmc/blob/master/xbmc/input/linux/LIRC.cpp#L271-L279) to always skip at minimum one repeat event, something that is undesired when using lircd-uinput together with eventlircd.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/settings/AdvancedSettings.cpp
+++ b/xbmc/settings/AdvancedSettings.cpp
@@ -1011,7 +1011,7 @@ void CAdvancedSettings::ParseSettingsFile(const std::string &file)
     }
   }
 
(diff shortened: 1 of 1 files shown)


MartijnKaijser:
@garbear ping. You have any knowledge in this area (or some one else)?

garbear:
I checked the LIRC code and it indeed skips a repeat event if range is 1. This change looks OK.

PR #11665, opened 2017-02-12
[cmake] iconv is a required dependency

bkuhls:
This patch adds support for libiconv on non-WIN32 platforms  currently
only provided by the autoconf-based build system:
https://github.com/xbmc/xbmc/blob/Krypton/configure.ac#L1172

This commit fixes an error during linking with an uClibc-based
buildroot toolchain:

[100%] Linking CXX executable kodi.bin
/home/buildroot/br8_ffmpeg3_kodi17_github/output/host/usr/lib/gcc/i586-buildroot-linux-uclibc/6.3.0/../../../../i586-buildroot-linux-uclibc/bin/ld: build/utils/utils.a(CharsetConverter.cpp.o): undefined reference to symbol 'libiconv_open'
/home/buildroot/br8_ffmpeg3_kodi17_github/output/host/usr/i586-buildroot-linux-uclibc/sysroot/usr/lib32/libiconv.so.2: error adding symbols: DSO missing from command line

Signed-off-by: Bernd Kuhls <bernd.kuhls@t-online.de>

## Types of change

## Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,6 +112,7 @@ set(required_deps Cdio
                   Fmt
                   FreeType
                   FriBidi
+                  Iconv
                   LibDvd
                   Lzo2
                   PCRE
--- a/cmake/modules/FindIconv.cmake
+++ b/cmake/modules/FindIconv.cmake
@@ -15,11 +15,19 @@
 
 find_path(ICONV_INCLUDE_DIR NAMES iconv.h)
 
-find_library(ICONV_LIBRARY NAMES iconv libiconv)
+find_library(ICONV_LIBRARY NAMES iconv libiconv c)
+
+set(CMAKE_REQUIRED_LIBRARIES ${ICONV_LIBRARY})
+check_function_exists(iconv HAVE_ICONV_FUNCTION)
+if(NOT HAVE_ICONV_FUNCTION)
+  check_function_exists(libiconv HAVE_LIBICONV_FUNCTION2)
+  set(HAVE_ICONV_FUNCTION ${HAVE_LIBICONV_FUNCTION2})
+  unset(HAVE_LIBICONV_FUNCTION2)
+endif()
 
 include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(iconv
-                                  REQUIRED_VARS ICONV_LIBRARY ICONV_INCLUDE_DIR)
+find_package_handle_standard_args(Iconv
+                                  REQUIRED_VARS ICONV_LIBRARY ICONV_INCLUDE_DIR HAVE_ICONV_FUNCTION)
 
 if(ICONV_FOUND)
   set(ICONV_LIBRARIES ${ICONV_LIBRARY})
@@ -33,4 +41,4 @@ if(ICONV_FOUND)
   endif()
 endif()
 
-mark_as_advanced(ICONV_INCLUDE_DIR ICONV_LIBRARY)
+mark_as_advanced(ICONV_INCLUDE_DIR ICONV_LIBRARY HAVE_ICONV_FUNCTION)
--- a/cmake/platform/windowsstore/windowsstore.cmake
+++ b/cmake/platform/windowsstore/windowsstore.cmake
@@ -1,1 +1,1 @@
-set(PLATFORM_REQUIRED_DEPS zlib iconv)
+set(PLATFORM_REQUIRED_DEPS zlib)


cmake/modules/FindIconv.cmake
```diff
@@ -0,0 +1,138 @@
...
+find_path(ICONV_INCLUDE_DIR iconv.h
+    HINTS
+    ${CMAKE_PREFIX_PATH}
+    ${ICONV_DIR}
+    $ENV{ICONV_DIR}
+    PATH_SUFFIXES include
+)
+
+if(NOT ICONV_INCLUDE_DIR STREQUAL "ICONV_INCLUDE_DIR-NOTFOUND")
+    set(CMAKE_REQUIRED_INCLUDES ${ICONV_INCLUDE_DIR})
```
wsnipex: duplicated further down, so can be removed here

cmake/modules/FindIconv.cmake
```diff
@@ -0,0 +1,138 @@
...
+else()
+    set(ICONV_TEST "In glibc")
+endif()
+
+set(CMAKE_REQUIRED_INCLUDES ${ICONV_INCLUDE_DIR})
+set(CMAKE_REQUIRED_LIBRARIES ${ICONV_LIBRARY})
+
+if(MSVC_VERSION GREATER 1800)
+    set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} legacy_stdio_definitions.lib)
+endif()
```
wsnipex: since we call this module only for non-windows, this block is superfluous

cmake/modules/FindIconv.cmake
```diff
@@ -0,0 +1,138 @@
...
+     }"
+    ICONV_COMPILES)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(ICONV DEFAULT_MSG ICONV_TEST ICONV_INCLUDE_DIR ICONV_COMPILES)
+
+if(ICONV_FOUND)
+  set(ICONV_LIBRARIES ${ICONV_LIBRARY})
+else(ICONV_FOUND)
+  set(ICONV_LIBRARIES)
```
wsnipex: not needed

cmake/modules/FindIconv.cmake
```diff
@@ -0,0 +1,138 @@
...
+endif(ICONV_FOUND)
+
+if(ICONV_FOUND)
```
wsnipex: please merge with the if block above

cmake/modules/FindIconv.cmake
```diff
@@ -0,0 +1,138 @@
...
+    set(CMAKE_REQUIRED_INCLUDES ${ICONV_INCLUDE_DIR})
+    set(CMAKE_REQUIRED_LIBRARIES ${ICONV_LIBRARIES})
+
+    if(MSVC_VERSION GREATER 1800)
+        set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} legacy_stdio_definitions.lib)
+    endif()
```
wsnipex: again, no win32 stuff needed

wsnipex (changes requested):
thanks. Only a few minor cleanups please.

bkuhls:
@wsnipex : done

wsnipex:
jenkins build this please

bkuhls:
@wsnipex: updated patch to fix rbpi build error

wsnipex:
I'm not sure we should skip the check on rpbi, @popcornmix?
jenkins build this please

popcornmix:
I would have thought rbpi should behave the same as linux so should be enabled.
What is the problem? Does this not handle cross-compiles?

bkuhls:
@wsnipex @popcornmix: The previous patch failed on rbpi: http://jenkins.kodi.tv/job/LINUX-RBPI/9535/console

MartijnKaijser:
ping

bkuhls:
Hi, I tried to reproduce the build error caught by jenkins in rbpi but I am having problems to find the correct instructions for setting up the cross-compile env. When adding the patch to Libreelec master it works when build with "PROJECT=RPi DEVICE=RPi2 ARCH=arm make":
```
-- Core system type: rbpi
-- Platform: 
-- CPU: arm, ARCH: arm
-- Cross-Compiling: TRUE
[...]
-- Found ICONV: In glibc  
-- One (and only one) of the ICONV_ACCEPTS_... tests must pass
```

CMakeLists.txt
```diff
@@ -128,6 +128,9 @@ endif()
 if(CORE_SYSTEM_NAME STREQUAL android)
   list(APPEND required_deps Zip)
 endif()
+if(NOT WIN32 AND NOT CORE_SYSTEM_NAME STREQUAL rbpi)
```
Rechi: `... AND NOT CORE_PLATFORM_NAME_LC STREQUAL rbpi`

bkuhls: fixed, thx!

Rechi:
Was there already a final decision if finding iconv on rbpi can be skipped?
Windows also needs iconv so why is the check skipped for it too?

bkuhls:
Let's try a different approach, I found a FindIconv.cmake file which looks more promising w.r.t. cross-platform compatibility. Unfortunately I can not test win/mac builds myself so I have to wait for Jenkins results, this may take several iterations in the worst case. Cross-compiling for "Core system type: linux" with glibc and uClibc using buildroot worked however.

wsnipex:
this FindIconv is worse then the old one.

cmake/modules/FindIconv.cmake
```diff
@@ -0,0 +1,43 @@
...
+#
+# - Try to find Iconv
+# Once done this will define
+#
+#  ICONV_FOUND - system has Iconv
+#  ICONV_INCLUDE_DIR - the Iconv include directory
+#  ICONV_LIBRARIES - Link these to use Iconv
+#  ICONV_SECOND_ARGUMENT_IS_CONST - the second argument for iconv() is const
+#
+include(CheckCXXSourceCompiles)
```
Rechi: not needed

cmake/modules/FindIconv.cmake
```diff
@@ -0,0 +1,43 @@
...
+
+IF (ICONV_INCLUDE_DIR AND ICONV_LIBRARIES)
+  # Already in cache, be silent
+  SET(ICONV_FIND_QUIETLY TRUE)
```
Rechi: why is this needed

cmake/modules/FindIconv.cmake
```diff
@@ -0,0 +1,43 @@
...
+ENDIF (ICONV_INCLUDE_DIR AND ICONV_LIBRARIES)
+
+FIND_PATH(ICONV_INCLUDE_DIR iconv.h)
+
+FIND_LIBRARY(ICONV_LIBRARIES NAMES iconv libiconv libiconv-2 c)
+
+IF(ICONV_INCLUDE_DIR AND ICONV_LIBRARIES)
+  SET(ICONV_FOUND TRUE)
```
Rechi: this has to be done with `find_package_handle_standard_args`

cmake/modules/FindIconv.cmake
```diff
@@ -0,0 +1,43 @@
...
+ENDIF(ICONV_INCLUDE_DIR AND ICONV_LIBRARIES)
+
+set(CMAKE_REQUIRED_INCLUDES ${ICONV_INCLUDE_DIR})
+set(CMAKE_REQUIRED_LIBRARIES ${ICONV_LIBRARIES})
```
Rechi: why is this needed?

cmake/modules/FindIconv.cmake
```diff
@@ -0,0 +1,43 @@
...
+
+IF(ICONV_FOUND)
+  IF(NOT ICONV_FIND_QUIETLY)
+    MESSAGE(STATUS "Found Iconv: ${ICONV_LIBRARIES}")
```
Rechi: `find_package_handle_standard_args` handles this whole logic

Rechi (changes requested):
This doesn't fit the style we use for the other find modules.

bkuhls:
Pushed another try with a simplified FindIconv.cmake based on the review by @Rechi.

I moved Iconv to optional_deps now because
- the current situation assumes that iconv functions are provided by libc, which is used for linking anyway
- my patch tries to solve a link error for (those rare) toolchains where libc does not provide iconv functions
- properly checking for Iconv using cmake is not easy (cmake itself does not provide a module for this purpose (an upstream bug ( https://cmake.org/Bug/view.php?id=13517 ), dating back to 2012, was closed last year ( https://gitlab.kitware.com/cmake/cmake/issues/13517 ) without an solution, the other FindIconv.cmake files I tested here did not work for all platforms
- when libiconv is found we can assume we should use it, if it was not found we should not throw an error because building Kodi worked without it with the exception of toolchains lacking iconv functions (buildroot only allows the usage of libiconv when the toolchain does not provide locale functions itself: https://github.com/buildroot/buildroot/blob/master/package/libiconv/Config.in#L3 ). This is why I moved Iconv to optional_deps and hope that Jenkins is happy with it ;)

bkuhls:
Analysis of Jenkins errors:

Win-64, LINUX-64 & IOS-ARM64: did not even start building because of a git error

LINUX-RBPI: did Jenkins use the correct patch? Iconv is not part of core_require_dep anymore
```  Could not find Iconv
  cmake/scripts/common/Macros.cmake:340 (find_package)
  CMakeLists.txt:167 (core_require_dep)```

**Comment by Rechi** [comment]
> I moved Iconv to optional_deps now [...]

This is wrong, iconv is a hard requirement for all platforms.

**Comment by bkuhls** [comment]
Ok, next try: This time FindIconv.cmake is meant to find the function iconv_open as hard dependency, not libiconv.
First we check for function iconv_open with the standard set of libraries, this means we assume that iconv functions are provided by libc. Only if they are not found we check for an iconv library containing iconv_open
```-- Found Iconv: /home/buildroot/br8/output/host/x86_64-buildroot-linux-uclibc/sysroot/usr/lib/libiconv.so```

If all of this fails:
```CMake Error at /usr/share/cmake-3.7/Modules/FindPackageHandleStandardArgs.cmake:138 (message):
  Could NOT find Iconv (missing: ICONV_LIBRARY)
  /usr/share/cmake-3.7/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  modules/FindIconv.cmake:35 (find_package_handle_standard_args)
  scripts/common/Macros.cmake:354 (find_package)
  CMakeLists.txt:140 (core_require_dep)```

File: cmake/modules/FindIconv.cmake [hunk 12]
```diff
@@ -0,0 +1,40 @@
...
+
+  if(ICONV_LIBRARY)
+    set(CMAKE_REQUIRED_LIBRARIES ${ICONV_LIBRARY})
+    check_function_exists(iconv_open ICONV)
+  endif()
+
+  if(ICONV)
+    set(ICONV_INCLUDE_DIRS ${ICONV_INCLUDE_DIR})
+    set(ICONV_LIBRARIES ${ICONV_LIBRARY})
+  endif()
```
wsnipex: please move this block down after find_package_handle_standard_args, but make it if(ICONV_FOUND)

cmake/modules/FindIconv.cmake
```diff
@@ -0,0 +1,40 @@
...
+
+  include(FindPackageHandleStandardArgs)
+  find_package_handle_standard_args(Iconv
+                                  REQUIRED_VARS ICONV_LIBRARY ICONV_INCLUDE_DIR)
```
wsnipex: and add ICONV here. It might also make sense to rename Iconv to something not possibly clashing with the upper case var ICONV. In that case adjust the ICONV_FOUND block mentioned in my comment above.

wsnipex (dismissed):
looks much better, just a minor change please

bkuhls:
Pushed new version, tested with buildroot git master:
```
Krypton - uclibc
-- Found Iconv: /home/buildroot/br8_ffmpeg3_kodi17_github/output/host/i686-buildroot-linux-uclibc/sysroot/usr/lib/libiconv.so  

Krypton - glibc
-- Found Iconv: 1  

Leia - uclibc
-- Found Iconv: /home/buildroot/br8_ffmpeg3_kodi17_github/output/host/i686-buildroot-linux-uclibc/sysroot/usr/lib/libiconv.so  

Leia - glibc
-- Found Iconv: 1  ```

File: cmake/modules/FindIconv.cmake [hunk 14]
```diff
@@ -0,0 +1,36 @@
...
+  if(ICONV_LIBRARY)
+    set(CMAKE_REQUIRED_LIBRARIES ${ICONV_LIBRARY})
+    set(CMAKE_REQUIRED_INCLUDES iconv.h)
+    check_library_exists(${ICONV_LIBRARY} libiconv_open "iconv.h" HAVE_ICONV_IN_LIBICONV)
+  endif()
+
+  find_package_handle_standard_args(Iconv
+                                    REQUIRED_VARS ICONV_LIBRARY HAVE_ICONV_IN_LIBICONV)
+
+  if(HAVE_ICONV_IN_LIBICONV)
```
Rechi: `if(ICONV_FOUND)`
`ICONV_FOUND` gets set by `find_package_handle_standard_args`

wsnipex:
jenkins build this please

Rechi (changes requested):
`find_path` for iconv.h is now missing for the non HAVE_ICONV_IN_LIBC case
everything else looks good

cmake/modules/FindIconv.cmake
bkuhls: ok

bkuhls:
In the next iteration of this patch I removed
```
set(CMAKE_REQUIRED_INCLUDES iconv.h)
```
so I think the find_path for iconv.h is not needed anymore, check_library_exists uses "iconv.h", no additional path needed imho.

cmake/modules/FindIconv.cmake
```diff
@@ -0,0 +1,35 @@
...
+# Finds the Iconv library
+#
+# This will define the following variables:
+#
+#  ICONV_FOUND - system has Iconv
+#  ICONV_LIBRARIES - Link these to use Iconv
+
+include(FindPackageHandleStandardArgs)
+
+check_library_exists("" iconv_open "iconv.h" HAVE_ICONV_IN_LIBC)
```
wsnipex: this doesn't work like you think. LOCATION is intended to be a directory, not a header file.
see https://github.com/Kitware/CMake/blob/master/Modules/CheckLibraryExists.cmake#L61

bkuhls:
@wsnipex: Done.
Sorry for using again a totally new FindIconv.cmake file, I never coded for CMake in the past, so I have to search the internet for ideas. While searching for the correct syntax for check_library_exists I came across https://github.com/owncloud/client/blob/fe0de111fdd4eef1c803cc3aca13d4f70c12b7ce/cmake/modules/FindIconv.cmake which looks as it contains everything needed for our purpose.
Let's see if Jenkins is happy with it ;)

wsnipex:
I'm not going to bother reviewing anymore, since you keep ignoring it and replace the file completely instead.

Rechi (changes requested):
This is getting too complex again.
Restore the previous version of your patch and just do the requested changes.

Rechi:
@bkuhls you still plan to do the requested changes?

bkuhls:
Yes, but after PR 12942 landed, which added FindIConv.cmake, I have to rework this PR.

bkuhls:
Done, it seems the new FindIConv.cmake also works fine with uClibc, only required_deps needed to be changed. Tested with glibc & uClibc.

Rechi:
One last thing, iconv has to be removed from
https://github.com/xbmc/xbmc/blob/5837475d/cmake/platform/windowsstore/windowsstore.cmake

bkuhls:
@Rechi: Done.
I just tested a backport of this PR together with FindIConv.cmake from PR 12942 successfully with Krypton HEAD. If this PR is accepted does it make sense that I send an update to PR 11665?

Rechi:
wrong revision was build
jenkins build this please

Rechi (approved):

wsnipex:
the findiconv modules from the UWP pull is just as broken.
It fails when iconv is built-in:

-- Could NOT find iconv (missing:  ICONV_LIBRARY)

Rechi:
@bkuhls is https://github.com/xbmc/xbmc/commit/a78fc281229c62f76c85f61997fb30d50a3cd016 working for you?

bkuhls:
@Rechi: Yes, Kodi git master with your patch still works fine with libiconv:
```
-- Found Iconv: /home/buildroot/br8/output/host/x86_64-buildroot-linux-uclibc/sysroot/usr/lib/libiconv.so
[...]
$ output/host/bin/x86_64-buildroot-linux-uclibc-readelf -a  output/target/usr/lib/kodi/kodi.bin | grep NEEDED | grep iconv
 0x0000000000000001 (NEEDED)             Shared library: [libiconv.so.2]
$ output/host/bin/x86_64-buildroot-linux-uclibc-g++ -v
[...]
gcc-Version 6.4.0 (Buildroot 2017.11-git-00994-ge467773c29-dirty) 
```

Rechi:
@bkuhls please update this PR with my patch

Rechi:
wrong revision was build
jenkins build this please

bkuhls:
@Rechi: The same problem I saw back in February ( https://github.com/xbmc/xbmc/pull/11665#issuecomment-279990639 ) again occurs with your patch on rbpi. We should treat libiconv as an optional dependency and do not bother to know which target library contains iconv functions if libiconv is not present.
@wsnipex: Regarding https://github.com/xbmc/xbmc/pull/11665#issuecomment-341655778 I would like to point out that this message is just an information that libiconv could not be found, the resulting build worked fine.

Rechi:
@bkuhls please leaf the PR as it currently is.
@wsnipex and I found the real reason why rbpi currently fails. A fix is very likely coming soon.

wsnipex:
jenkins build this please

wsnipex:
jenkins build this please

bkuhls:
Considering the fact that this PR needed changes for the rbpi toolchain do you think a backport of this PR to Krypton will be accepted?

Rechi:
The rbpi toolchain change won't be backported, so this also can't be backported as it would break rbpi.

PR #11668, opened 2017-02-13
FFmpegImage: Document and change alignment

fritsch:
This makes the intermediate buffer length aligned to 32, so that intrinsics can properly work with this intermediate buffer.

Change frame's memory to av_malloc / av_free allocated

### DIFF
--- a/xbmc/guilib/FFmpegImage.cpp
+++ b/xbmc/guilib/FFmpegImage.cpp
@@ -400,6 +400,7 @@ bool CFFmpegImage::DecodeFrame(AVFrame* frame, unsigned int width, unsigned int 
     return false;
   }
 
+  // we align on 16 as the input provided by the Texture also aligns the buffer size to 16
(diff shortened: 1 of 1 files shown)


fritsch:
@popcornmix @MilhouseVH for testing.

xbmc/guilib/FFmpegImage.cpp
```diff
@@ -53,14 +53,14 @@ Frame::Frame(const Frame& src) :
 {
   if (src.m_pImage)
   {
-    m_pImage = new unsigned char[m_imageSize];
+    m_pImage = (unsigned char*) av_malloc(m_imageSize);
```
fritsch: this allocation sucks!

MilhouseVH:
@fritsch is this happening? I've had it in my RPi/Generic builds almost 3 months without issue, but will be dropping it now.

fritsch:
Thx for shipping it. It could go in.

Am 06.04.2017 12:12 vorm. schrieb "MilhouseVH" <notifications@github.com>:

MartijnKaijser:
jenkins build this please

PR #11670, opened 2017-02-14
fix BitstreamConverter missing declarations

peak3d:
[AML] Compile Fix BitsreamConverter

## Description
see title

### DIFF
--- a/xbmc/utils/BitstreamConverter.h
+++ b/xbmc/utils/BitstreamConverter.h
@@ -87,8 +87,8 @@ public:
   CBitstreamParser();
   ~CBitstreamParser();
 
-  static bool Open();
-  static void Close();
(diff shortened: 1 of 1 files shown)


xbmc/utils/BitstreamConverter.h
```diff
@@ -87,8 +87,8 @@ class CBitstreamParser
   CBitstreamParser();
   ~CBitstreamParser();
 
-  static bool Open();
-  static void Close();
+  static bool Open(){};
```
tamland: @peak3d You forgot to return. If called, this is undefined behaviour. It's a good idea enable -Werror=return-type to catch these types of errors..

peak3d:
Thank you, will fix, @tamland

PR #11672, opened 2017-02-14
Backport: Improve keyboard handling during button mapping

garbear:
Backport of #11645


MartijnKaijser:
Sorry but isn't this a bit to big for a backport? It touches so many things

garbear:
your call. I can drop the first two commits, which will leave the patch at +200/-3, which would be a lot safer as it doesn't touch existing code

MartijnKaijser:
i'd say keep it simple

garbear:
K, it's been updated to a small addition. It shouldn't disturb any existing code.

MartijnKaijser:
Jenkins build this please

PR #11673, opened 2017-02-14
always scrape artist data

ronie:
currently we skip the artist metadata scraper if we did not mange to scrape an album by this artist.

this PR changes that by always scraping the artist data.

@DaveTBlake

### DIFF
--- a/xbmc/music/infoscanner/MusicInfoScanner.cpp
+++ b/xbmc/music/infoscanner/MusicInfoScanner.cpp
@@ -840,45 +840,42 @@ int CMusicInfoScanner::RetrieveMusicInfo(const std::string& strDirectory, CFileI
       if (!m_musicDatabase.HasAlbumBeenScraped(album->idAlbum))
         albumScrapeStatus = UpdateDatabaseAlbumInfo(*album, albumScraper, false);
 
-      if (albumScrapeStatus == INFO_ADDED)
+      for (VECARTISTCREDITS::const_iterator artistCredit  = album->artistCredits.begin();
+                                            artistCredit != album->artistCredits.end();
+                                          ++artistCredit)
       {
-        for (VECARTISTCREDITS::const_iterator artistCredit  = album->artistCredits.begin();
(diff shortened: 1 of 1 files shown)


DaveTBlake:
Thanks for this @ronie although I won't get a chance to test fully with some trickier data until the weekend.

But just thinking about it something is still niggling me about whether there are reasons to skip the artists if we don't get an album result.  Must pin it down !

DaveTBlake:
What do you think about scraping albums and artists after the tag processing for all the song files in the library has been done, rather than as each album as it is added? More about this on the forum thread.

jjd-uk:
@DaveTBlake think this is also waiting further feedback from you.

DaveTBlake:
I don't think that we should do what this PR does, and having realised that I have been working on an alternative (as yet incomplete because I haven't had the time to do it). 

When we don't have mbids scraping artist and album together helps to identify things correctly, but the changes here to always try the artist can end up with repeated futile attempts and needless server traffic.

DaveTBlake:
@ronie @notspiff  my take on better scraping of artist and album info on library update is here
https://github.com/DaveTBlake/xbmc/tree/KeepScrapedMBID

While it does scrape the album artist even when scraping the album fails, as this PR wanted, it keeps a list to avoid repeatedly requesting artists that are unknown (could happen every song, and for every album otherwise). 

It also stores the scraped mbids in the db for later use by cdArt addon etc., but flags them as scraped so that any mis-identified artists or albums can be corrected manually by refresh from the info dialog (it doesn't use any scraped mbids, only those from tags, but does a name lookup again instead).

Need to do some more testing before raising PR, and the db skips from v64 to v67 because I have other db changes in progress, but input very welcome.

DaveTBlake:
Have raised #12120 that I think should replace this.

Rechi:
@DaveTBlake is this now obsolete with #12120 merged?

DaveTBlake:
Yes it is obsolete, so closing. Thanks @ronie for getting me to think about this and improve scraping.

PR #11674, opened 2017-02-15
[linux] systemd service support for add-ons and timedated support for timezone

lrusak:
## Description

This change is pretty specific to LibreELEC however it may be something that upstream is interested so I am making a PR. This will probably replace our current patches in LibreELEC regardless of if they are accepted here or not. If this isn't something that we want in Kodi then it would still be nice to get some comments about the code quality (I'm no c++ dev). 

This allows running systemd services that are inside the add-on. On LibreELEC because we use Kodi's add-on system as the package manager. We provide binary add-ons that can contain services (like tvheadend, docker, etc). These services are started via systemd service files. We previously maintained an ugly patch to start these services. see, https://github.com/xbmc/xbmc/commit/aadeef8bdff40ae19cb61f1206ff30ebbaebb8bb and [here](https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/mediacenter/kodi/scripts/systemd-addon-wrapper)

I wanted to make something that is cleaner and using proper methods. Since Kodi has a DBus method already it was rather simple to use that. So when add-ons are enabled/disabled dbus calls the systemd manager to enable/disable and start/stop service units.

The second commit extends this to add timedated support to allow setting the host OS timezone from within kodi. LibreELEC again maintained a patch for this. see, https://github.com/xbmc/xbmc/commit/fb5c7a43280419bfa464b945da3b46f74782f0ad

I decided to only depend on dbus because that's what is done in [LogindUPowerSyscall.cpp](https://github.com/xbmc/xbmc/blob/master/xbmc/powermanagement/linux/LogindUPowerSyscall.cpp) as it doesn't look for systemd or upower.

I'll probably add some logging to SystemdUtils.cpp eventually.

## Motivation and Context

see description above

## How Has This Been Tested?

I tested this in LibreELEC. It works just like before with the old patches.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/platform/linux/LinuxTimezone.cpp
+++ b/xbmc/platform/linux/LinuxTimezone.cpp
@@ -13,6 +13,9 @@
 #include "PlatformDefs.h"
 #include "LinuxTimezone.h"
 #include "utils/SystemInfo.h"
+#ifdef HAS_DBUS
+#include "SystemdUtils.h"
+#endif
 
 #include "ServiceBroker.h"
 #include "Util.h"
@@ -141,6 +144,9 @@ void CLinuxTimezone::OnSettingChanged(std::shared_ptr<const CSetting> setting)
   const std::string &settingId = setting->GetId();
   if (settingId == CSettings::SETTING_LOCALE_TIMEZONE)
   {
+#ifdef HAS_DBUS
+    CSystemdUtils::SetTimezone(std::static_pointer_cast<const CSettingString>(setting)->GetValue());
+#endif
     SetTimezone(std::static_pointer_cast<const CSettingString>(setting)->GetValue());
 
     CDateTime::ResetTimezoneBias();
--- a/xbmc/platform/linux/SystemdUtils.cpp
+++ b/xbmc/platform/linux/SystemdUtils.cpp
@@ -1,0 +1,82 @@
+/*
+ *      Copyright (C) 2005-2013 Team XBMC
+ *      http://kodi.tv
+ *
+ *  This Program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2, or (at your option)
+ *  any later version.
+ *
+ *  This Program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with XBMC; see the file COPYING.  If not, write to
+ *  the Free Software Foundation, 675 Mass Ave, Cambridge, MA 02139, USA.
+ *  http://www.gnu.org/copyleft/gpl.html
+ *
+ */
+
+#include "SystemdUtils.h"
+
+#include "DBusUtil.h"
+#include "utils/log.h"
+
+// systemd DBus interface specification:
+// https://www.freedesktop.org/wiki/Software/systemd/dbus
+
+static const std::string SYSTEMD_DEST = "org.freedesktop.systemd1";
+static const std::string SYSTEMD_PATH = "/org/freedesktop/systemd1";
+static const std::string SYSTEMD_IFACE = "org.freedesktop.systemd1.Manager";
+
+bool CSystemdUtils::EnableUnit(const char **units, unsigned int length)
+{
+  CDBusMessage message(SYSTEMD_DEST, SYSTEMD_PATH, SYSTEMD_IFACE, "EnableUnitFiles");
+
+  // It takes a list of unit files to enable (either just file names or full absolute paths)
+  message.AppendArgument(units, length);
+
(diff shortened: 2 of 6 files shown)


hudokkow:
ping @tamland, @Montellese and @notspiff

notspiff:
two comments from me: a lot of unncessary pass by const char * (inherited from style in surrounding code) and i thought we had stopped with the ifdef around entire files nonsense by now (should be handled in buildsystem). other than that, good stuff.

tamland:
You'll need to either make this into a new addon type or extend the current service addon. Not just "attach" some systemd logic onto the service addon type, which are currently for something totally different. The path (and other things) should then be configurable similar to python, not hardcoded.

Also, there seems to be no:
-synchronisation between systemd states and addon enabled/disabled state 
-profiles handling

lrusak:
Thanks for the comments guys

@notspiff I'm not really sure about the implications of `const char *` and when/why it should be used/avoided.

The ifdef stuff was just copied as I said. If it's not correct I can look into changing it but will probably have to change it for similar code paths as well. fun. :smile: 

@tamland that sounds like a pretty sane idea to me however I really don't know how involved it is to create a new add-on type. I assume you mean configurable like how we allow passing the path to the included python library in the addon.xml? [Like this?](https://github.com/xbmc/xbmc/blob/master/addons/script.module.pil/addon.xml#L10)

In terms of synchronization systemd exposes `ActiveState` and `UnitFileState` to check if the service is running/dead and enabled/disabled. I guess checking against this before the services are run is a good idea.

Profiles I don't know much about and never use them so I don't know about any special add-on handling that they use.

notspiff:
@lrusak the const char* has only a syntactic cost in this case. i like to delay it to the very end and pass as (const) std::string(&) as long as possible, but as i said it's caused by the surrounding code so you cannot address it without a huge refactor.

the ifdef stuff is simple to address: you remove the #ifdef HAS_DBUS surrounding SystemdUtils.h/cpp , then only add the files to the build system if DBUS_FOUND is true, like https://github.com/xbmc/xbmc/blob/master/xbmc/linux/CMakeLists.txt#L35

tamland:
Totally forgot about this.. @lrusak can you rebase? If you simply move the start/stop code for the .service type to [here](https://github.com/xbmc/xbmc/blob/7005e4f7383b3cf6f9413a3bab4f2bf6d689b223/xbmc/addons/Service.cpp#L133) and [here](https://github.com/xbmc/xbmc/blob/7005e4f7383b3cf6f9413a3bab4f2bf6d689b223/xbmc/addons/Service.cpp#L163), it should behave the same as python services.

xbmc/platform/linux/SystemdUtils.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+#include "DBusUtil.h"
+#include "utils/log.h"
+
+// systemd DBus interface specification:
+// https://www.freedesktop.org/wiki/Software/systemd/dbus
+
+#define SYSTEMD_DEST  "org.freedesktop.systemd1"
+#define SYSTEMD_PATH  "/org/freedesktop/systemd1"
+#define SYSTEMD_IFACE "org.freedesktop.systemd1.Manager"
```
pkerling: Prefer `const std::string` over `#define` whenever possible

And yeah, I know it's c&p from the other code, just saying :-)

pkerling:
General comment: I think you *should* be able to use `AppendArgument` with strings directly (and not call `c_str()`) after my last refactor of the d-bus stuff. If not we should totaly make that possible :-)

xbmc/platform/linux/LinuxTimezone.cpp
```diff
@@ -153,6 +156,9 @@ void CLinuxTimezone::OnSettingChanged(std::shared_ptr<const CSetting> setting)
   const std::string &settingId = setting->GetId();
   if (settingId == CSettings::SETTING_LOCALE_TIMEZONE)
   {
+#ifdef HAS_DBUS
+    CSystemdUtils::SetTimezone(std::static_pointer_cast<const CSettingString>(setting)->GetValue());
```
pkerling: I'm not sure we always want to do that. On Desktop systems, users usually don't have the right to call `SetTimezone` on timedated, so you'll get error log spam. And even if they have the right, it's totally unexpected (to me at least) that changing the time zone in Kodi would change the system time zone.

lrusak:
@pkerling thanks! I'll revisit this eventually :+1:

xbmc/platform/linux/LinuxTimezone.cpp
lrusak: I totally agree. The PR was done completely with LibreELEC in mind

pkerling: I know that on LibreELEC, Kodi runs as root. Is that expected to change at some point or will you continue to have that?

Also, do we have sth like a "standalone/management" mode for boxes only running Kodi (like LibreELEC basically...) that we could use for enabling this feature?

phunkyfish:
@lrusak close or backburner?

lrusak:
The idea is still valid but the implementation not so much. The idea behind this was simply to reduce the patchset that LibreELEC carries. I'll look into doing something different down the road.

PR #11675, opened 2017-02-15
control horizontal sliders by up/down to right/left GUISliderControl.cpp

harryberlin:
lets control horizontal sliders by up/down to right/left, if itself is active.

## Description

deleted the vertical condition, to control horizontal sliders too.

## Motivation and Context

very helpful for rotary controls.
works as a real active flag.

## How Has This Been Tested?

On OSMC from today

1. activate the horizontal slider with press
2.a) moves left with left
2.b) moves left with down
2.c) moves right with up
2.d) moves right with right
3. deactivate with press

## Types of change

## Checklist:

### DIFF
--- a/xbmc/guilib/GUISliderControl.cpp
+++ b/xbmc/guilib/GUISliderControl.cpp
@@ -212,7 +212,7 @@ bool CGUISliderControl::OnAction(const CAction &action)
     break;
 
   case ACTION_MOVE_UP:
-    if (IsActive() && m_orientation == VERTICAL)
+    if (IsActive())
     {
       Move(1);
       return true;
@@ -220,7 +220,7 @@ bool CGUISliderControl::OnAction(const CAction &action)
     break;
 
   case ACTION_MOVE_DOWN:
-    if (IsActive() && m_orientation == VERTICAL)
+    if (IsActive())
     {
       Move(-1);
       return true;


ronie:
over the last few years we've been trying to make navigation in kodi more intuitive.
pressing 'up' to move 'left' doesn't fit that bill imo.

harryberlin:
i mean, it is not intuitive to select a slider to activate, and you can leave it before deactivating (select again).
for example: the horizontal slider, can be left with up and down.
so my aim don't ignore the directions, better to use them for moving.

ronie:
requiring a click to deactive a slider is one thing... and i'm open for discussion.
but using up/down to navigate left/right is not something i would prefer to add.

harryberlin:
i'm using a rotary encoder. on clockwise it returns up. on counterclockwise it returns down.
exist actions like clockwise and couterclockwise for kodi to handle these things?
i think no. would be cool to get something.

harryberlin:
and it is only for the slider. the other navigations stays untouched.

BigNoid:
I'm also against this. We shouldn't implement awkward navigation schemes. The way the sliders are now are how it should be; normal navigation to them and when you want to adjust, select it first. Sure, if you are used to the previous way of immediately being able to adjust them it takes getting used to, but for new users this is far more intuitive.

phil65:
Also -1 from my side.
In addition, this change would only make sense somehow when also adjusting line 199 and 207, otherwise things get inconsistent.

Hedda:
For reference; such type of rotary controller input device with a control knob (a.k.a. command wheel) that harryberlin mentions he uses for navigation is quite common input type to control multimedia and HVAC in luxury cars (and modern sedans).

Audi has something they call "MMI" (which in their case stands for 'Multi Media Interface')

https://en.wikipedia.org/wiki/Multi_Media_Interface

BMW has a rotary knob interface called "iDrive" (which is also used in current Rolls-Royce models, as Rolls-Royce is owned by BMW)

https://en.wikipedia.org/wiki/IDrive

Mercedes-Benz has something they call "COMAND" (Cockpit Management and Data) system

https://en.wikipedia.org/wiki/Comand_APS

Allowing for rotary controller navigation might not be a bad thing to enable compatibility with that type of input interface, which can achieve an convenient interface with a central/single/one-knob that in some cases also acts as a joystick.

To quote wikipedia: "The design rationale of iDrive is to replace an array of controls for the above systems with an all-in-one unit."

phil65:
I dont see how that explanation is connected to this PR. A rotary knob would be fine already today in case it sends left/right commands when rotating.

harryberlin:
e39 bmw bordscreen with kodi on rasp:
https://www.youtube.com/watch?v=PkdtcNzpZ7Y

PR #11679, opened 2017-02-16
[PVR] Timer settings dialog: Prefill recordings folder with timer title

ksooo:
This is a small change that imo significantly improves user experience. 

For example, when creating an epg based timer rule, which shall put recordings in a certain directory (recording folder setting), in most cases the user wants to use the name of epg event for the recording folder name. With this change the user must no longer type in that name as it is prefilled by the timers settings dialog.

Runtime tested on latest master.

@Jalle19 for review?


Jalle19 (approved):
One potential issue with this is that backends/addons will have to sanitize the directory name (see e.g. the multitude of options regarding this when creating a recording profile in tvheadend).

I think it's safe to say that it's not Kodi's problem though, so +1.

ksooo:
> I think it's safe to say that it's not Kodi's problem though

I agree. 

jenkins build this please

ksooo:
jenkins build this please

PR #11680, opened 2017-02-17
[Fix]Limited number of Random songs apply sort and limit in DB

DaveTBlake:
All music library sorting, including random order, is currently done in the list of file items after having fetched the data, not at the database. This is particularly inefficient when the query is say for a limited number of random songs - **all** the songs are fetched along with their artist data, this is manipulated from the results set into a CFileItemList (the reason songs node is so slow) and only then most of the data disgarded. Loading say 100k of songs just to see 20 of them!!!

The Library Data Provider addon, used by many skins, requests 20 random unplayed songs every 10 mins, a reasonable request. On large music libraries when this happens during initial scanning (all the songs are unplayed) it has caused crashes.

This is a simple fix of the limited random song query, it applies the random sort at the db i.e. can get  X random songs without processing the entire table into local memory. I hope that it can be backported.

I am looking at refactoring the way sorting is done entirely, but I think we need to apply this limited fix for immediate release.


DaveTBlake:
Have tested with both SQLite and MySQL databases, using smart playlists and JSON API calls

jenkins build this please

PR #11683, opened 2017-02-17
[Fix][Backport]Limited number of Random songs apply sort and limit in DB

DaveTBlake:
Backport of #11680 

This is a simple fix of the limited random song query, it applies the random sort at the db i.e. can get X random songs without needlessly processing the entire song (and artists) into local memory. 

As well as being much more efficient it will avoid "out of memory" crashes that can be caused on large music libraries by the Library Data Provider addon, used by many skins, when it requests 20 random unplayed songs.


MartijnKaijser:
Jenkins build this please

PR #11685, opened 2017-02-17
Settings: Disable dithering by default - cause 100% load on old rigs

fritsch:
See title.

Background: http://forum.kodi.tv/showthread.php?tid=231955&pid=2528669#pid2528669


fritsch:
@laurimyllari For information. Yeah - there are still a whole lot old hw rigs out there - so we need to carefully think of the defaults, sadly.

PR #11692, opened 2017-02-19
[Darwin] Separate Apple Remote settings into new group

garbear:
This places the Apple Remote settings into a new group named "Apple Remote".

Previously, they were part of the Devices group, which led to poor UX by having 7 mostly unrelated settings in the same group.

ping @jjd-uk, this is your territory

Before:
screen shot 2017-02-18 at 10 59 08 pm

After:
screen shot 2017-02-18 at 10 36 11 pm

### DIFF
--- a/system/settings/darwin.xml
+++ b/system/settings/darwin.xml
@@ -23,7 +23,7 @@
       </group>
     </category>
     <category id="input">
-      <group id="2">
+      <group id="3">
         <setting id="input.appleremotemode" type="integer" label="13600" help="36416">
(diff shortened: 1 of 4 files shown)


jjd-uk:
Looks sensible to me.

I also wonder with your Controller additions whether it might make sense to split those out to their own grouping to have something like:

```
**General**
Enable mouse and touch screen support
Peripherals

**Controllers**
Configure attached controllers
Test rumble
Power off controllers on exit
```

PR #11693, opened 2017-02-19
[estuary] added: Audio/video bitrate info for video process dialog

arnova:
This adds the use of the new labels for video/audio bitrate info. I think we should backport this to Confluence as well.

## Types of change

## Checklist:

### DIFF
--- a/addons/skin.estuary/xml/DialogPlayerProcessInfo.xml
+++ b/addons/skin.estuary/xml/DialogPlayerProcessInfo.xml
@@ -163,7 +163,7 @@
 					<width>1600</width>
 					<height>50</height>
 					<aligny>bottom</aligny>
-					<label>$INFO[Player.Process(videowidth),[COLOR button_focus]$LOCALIZE[38031]:[/COLOR] ,x]$INFO[Player.Process(videoheight),, px]$INFO[Player.Process(videodar),$COMMA , AR]$INFO[Player.Process(videofps),$COMMA , FPS]</label>
+					<label>$INFO[Player.Process(videowidth),[COLOR button_focus]$LOCALIZE[38031]:[/COLOR] ,x]$INFO[Player.Process(videoheight),, px]$INFO[Player.Process(videodar),$COMMA , AR]$INFO[Player.Process(videofps),$COMMA , FPS]$INFO[VideoPlayer.VideoBitrate,$COMMA , kb/s]</label>
 					<font>font14</font>
 					<shadowcolor>black</shadowcolor>
 					<visible>Player.HasVideo</visible>
@@ -172,7 +172,7 @@
 					<width>1600</width>
 					<height>50</height>
 					<aligny>bottom</aligny>
-					<label>[COLOR button_focus]$LOCALIZE[460]:[/COLOR] $INFO[Player.Process(audiochannels),,$COMMA ]$INFO[Player.Process(audiodecoder)]$INFO[Player.Process(audiobitspersample),$COMMA , bits]$INFO[Player.Process(audiosamplerate),$COMMA , Hz]</label>
+					<label>[COLOR button_focus]$LOCALIZE[460]:[/COLOR] $INFO[Player.Process(audiochannels),,$COMMA ]$INFO[Player.Process(audiodecoder)]$INFO[Player.Process(audiobitspersample),$COMMA , bits]$INFO[Player.Process(audiosamplerate),$COMMA , Hz]$INFO[VideoPlayer.AudioBitrate,$COMMA , kb/s]</label>
 					<font>font14</font>
 					<shadowcolor>black</shadowcolor>
 				</control>


jjd-uk:
For Confluence that would be @HitcherUK

arnova:
ping @ronie / @FernetMenta

FernetMenta:
This value does not fit into this dialog because it is metadata of the stream. It should be displayed with the other info that comes up when you press "i"

arnova:
@FernetMenta : IMHO it doesn't make sense to clutter the video info OSD with this info (how do you see this fitted in there?). I think from the user's perspective it would make much more sense to have all codec/mux/process info in a single dialog. I think the enduser doesn't really think about where the info came from.

FernetMenta:
Well, if you mix up the info you'll get the same crap as we had before. Means that nobody on the planet knows exactly what those values mean. 
If you what to have those on the same page, you should at least add a separator for the different categories. You also should change the name of the dialog form ProcessInfo to something else.

You get the most from those values if you make it clear:
- we have input to the process, the input is defined my i.e. metadata of a stream
- process parameters, that is how player configures itself to deal with the input
- output, i.e. audio was downmixed to 2.0 or video was downscaled

arnova:
@FernetMenta : Actually I roughly had the same thoughts
@ronie: Can you look into this?

ronie:
i can, if someone could make a list of which infolabels belong to which category.

FernetMenta:
@ronie @arnova all PLAYER_PROCESS_* belong to category process. PLAYER_PROCESS_AUDIOCHANNELS i.e. is not the channels of the stream but could be transport channels needed for passthrough.

arnova:
@ronie: Any update on this? Do you need any additional info? Having this is still high on my wishlist :)

phunkyfish:
@arnova would you still like this PR merged?

phunkyfish:

arnova:
@phunkyfish : As far as I'm concerned this could be merged as-is. But back then there were people against it. I still believe this is a nice improvement but we need to agree whether it's ok as-is or there's another way to implement it in skins (which is not my turf).

arnova:
Btw. I would simply rename the dialog from process info to player info, merge this as-is and be done with it.

phunkyfish:
@ronie what do you think?

ronie:
skin changes itself look ok to me.

for the rest, i'd rather stay out of technical wrong/right discussions between devs.

phunkyfish:
Thanks for looking @ronie.

@arnova not sure who is best to ask about this.

arnova:
> Thanks for looking @ronie.
> 
> @arnova not sure who is best to ask about this.

@phunkyfish : Not sure either but I opt for simply changing the headers in the dialog from "VIdeo Process Information" and "Audio Process Information" to "Video Information" and "Audio Information". I think that's a good compromise.

If anyone is against this, speak up (and suggest a proper alternative), else we'll merge it this way.

phunkyfish:
@arnova do you have before/after screenshots?

arnova:
image

After:

image

phunkyfish:
Screenshot 2020-04-28 at 08 58 55

In the above screenshot was the other thought to have as one of the white boxes? With something like `V: 1050kb/s` and `A: 130 kb/s`?

arnova:
> Screenshot 2020-04-28 at 08 58 55
> 
> In the above screenshot was the other thought to have as one of the white boxes? With something like `V: 1050kb/s` and `A: 130 kb/s` alongside h264, 1080HD etc.?
> 
> If that is the right place is there an issue with putting it there? I'd say the skinner can probably come up with a fancier way of doing it than just pure text. Or are we short on real estate there? @ronie thoughts?

I think media flags are always icons/bitmaps, so I don't think that can be easily implemented that way. But indeed: @ronie may be able to confirm this.

jjd-uk:
@arnova I implemented the music player infolabels for SampleRate, BitRate, & BitsPerSample into the Viz window media flags here https://github.com/xbmc/xbmc/pull/17183 so any additional media flags for video properties could be done in a similar way.

jjd-uk:
Personally I think this is fine as is from a cosmetic perspective, so from that point of view is ok to be merged as is. For video the space for additional media flags might be more limited, I'm working on redoing the info shown at the bottom while either seeking or info is pressed so I could look at adding these. However even if media flags were to be added I believe it's still worth having the info added as in this PR to DialogPlayerProcessInfo so all the detailed stream info is together in one place.

phunkyfish:
So we bring this in. Will merge later today.

phunkyfish (approved):

ksooo:
> Will merge later today.

 Hmmm. If we want it not where this PR introduces this information, why should we merge this?

ksooo:
At least we should make our homework before merging. As discussed in this PR, with this PR the dialog is no longer a "video process" dialog and that should reflect at the GUI and in the source. Examples:

* Change GUI string "Player process info" (see screen shots) to "Player info"
* Change dialog impl source name from CGUIDialogPlayerProcessInfo to CGUIDialogPlayerInfo
* Change window id define name WINDOW_DIALOG_PLAYER_PROCESS_INFO
* Change skin xml file name "DialogPlayerProcessInfo.xml" 
* ...?

phunkyfish:
> > Will merge later today.
> 
> Hmmm. If we want it not where this PR introduces this information, why should we merge this?

Fair

phunkyfish:
@arnova do you want to implement the naming suggestions or wait until @jjd-uk can work this into the media flags?

arnova:
> @arnova do you want to implement the naming suggestions or wait until @jjd-uk can work this into the media flags?

Currently I don't have much spare time on my hands so if @jjd-uk could take care of this, that would be much appreciated.

DaveTBlake:
Has this been superseded @arnova @jjd-uk ?
If so please close and set milestone "abandoned"

jjd-uk:
Using media flags is not an option die to lack of space in the OSD.

I'm not sure why C++ changes are being requested for a Estuary only change to simply add a couple of infolabels, so I'm not willing to touch that code and I'm not even sure it's within my capability.

I think arnova has no time at the moment so I'm labelling Backburner until we hear something from @arnova

PR #11704, opened 2017-02-19
[PVR][guiinfo][Estuary] Recordings window: Introduce a content view filter

ksooo:
Recordings window: Introduce a content view filter (all|watched|unwatched), like already existing for videos and tv shows.

screenshot000
screenshot001

This was runtime tested on latest kodi master on macOS.

@Jalle19 for review, please?


addons/resource.language.en_gb/resources/strings.po
```diff
@@ -10671,7 +10671,25 @@ msgctxt "#19296"
...
+#. Label for filtered recordings view value: 'show only unwatched'
+#: xbmc/pvr/windows/GUIWindowPVRRecordings.cpp
+msgctxt "#19298"
+msgid "Unwatched"
+msgstr ""
+
+#. Label for filtered recordings view value: 'show only watched'
+#: xbmc/pvr/windows/GUIWindowPVRRecordings.cpp
+msgctxt "#19299"
+msgid "Watched"
```
MartijnKaijser: Aren't these labels available somewhere else?

ksooo: Yep, but the implementation (I just reused) relies on three strings with numeric values in a row, and because one of the strings for videos ("All videos") did not fit, I needed to reinvent three new strings. Not a problem, if you ask me. The context is another, this it even cn be dangerous to reuse the strings, we had that problem already..

ksooo: @MartijnKaijser on second thought, this is bullsh**. I will improve the implementation not to rely any longer on three num values in a row. ;-)

MartijnKaijser: Hahaha

ksooo:
@ronie headups: new infolabel: Container(id).NonFolderItems
@phil65, @ronie are the skin changes okay?

ronie:
thx for the headsup!
skin changes look sane to me.

ksooo:
jenkins build this please

xbmc/guilib/GUIBaseContainer.cpp
```diff
@@ -1306,10 +1306,21 @@ std::string CGUIBaseContainer::GetLabel(int info) const
     break;
   case CONTAINER_ACTUAL_ITEMS:
   case CONTAINER_NUM_ITEMS:
+  case CONTAINER_NONFOLDER_ITEMS:
     {
       unsigned int numItems = GetNumItems();
       if (info == CONTAINER_NUM_ITEMS && numItems && m_items[0]->IsFileItem() && (std::static_pointer_cast<CFileItem>(m_items[0]))->IsParentFolder())
         label = StringUtils::Format("%u", numItems-1);
+      else if (info == CONTAINER_NONFOLDER_ITEMS)
```
xhaggi: This looks a bit odd to me. Why did you add CONTAINER_NONFOLDER_ITEMS to this case and do a second and third check? Each call will check for that case, then it'll do the unnecessary check (first if condition) before it'll check CONTAINER_NONFOLDER_ITEMS again. Why not add a separate case?

xbmc/GUIInfoManager.cpp
```diff
@@ -2400,6 +2407,7 @@ const infomap container_bools[] ={{ "onnext",           CONTAINER_MOVE_NEXT },
                                   { "numpages",         CONTAINER_NUM_PAGES },
                                   { "numitems",         CONTAINER_NUM_ITEMS },
                                   { "actualitems",      CONTAINER_ACTUAL_ITEMS },
+                                  { "nonfolderitems",   CONTAINER_NONFOLDER_ITEMS },
                                   { "currentpage",      CONTAINER_CURRENT_PAGE },
```
xhaggi: Why not call it `numfileitems`, instead of negate folder items ;)

xhaggi:
A bit late but you are too fast in merging new stuff.

xhaggi:
Next time it would be good to separate gui related changes from skin changes. It will make things a bit easier to spot.

ksooo:
I'm not too fast. If I'm not allowed to work with that speed I will stop working at all. The current speed is what keeps me happy with the project. ;-)

And honestly, what you are saying is not wrong, but as I always use the subject line to tag what is effected and I always use the respective tags I do not see a real benefit in splitting a few lines PR in separate commits - this is, well, bureaucracy.

xhaggi:
Any intention to address my comments? If not I'll do it on my own.

xbmc/guilib/GUIBaseContainer.cpp
ksooo: Good point.

xbmc/GUIInfoManager.cpp
ksooo: Because not every non-folder item is a file item, imo. What about channels games, epg events? Are those file items? I don't think so.

ksooo:
> Any intention to address my comments? If not I'll do it on my own.

 Sorry, I simply overlooked your comments. Please beautify the logic in GUIBaseContainer if you feel like this, but imo it is not okay to change the name of the GUI label, because "numfileitems" does not describe precisely the semantics of the new  label.

xhaggi:
I'm currently working on it. 

> but imo it is not okay to change the name of the GUI label, because "numfileitems" does not describe precisely the semantics of the new label.

Agree, but I'll add the `num` to the GUI info label so it becomes `NumNonFolderItems`.

PR #11707, opened 2017-02-19
Fix rumble always enabled

garbear:
This adds a setting to disable rumble for notifications. The default is to not rumble.

Reported here: http://forum.kodi.tv/showthread.php?tid=296564

This also improves the visibility conditions for rumble and power-off settings. Now, they are only shown if the underlying joystick driver supports each.

The corresponding peripheral.joystick commit is https://github.com/xbmc/peripheral.joystick/pull/86

screen shot 2017-02-19 at 12 09 03 pm

## Types of change

### DIFF
--- a/addons/kodi.peripheral/addon.xml
+++ b/addons/kodi.peripheral/addon.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<addon id="kodi.peripheral" version="1.3.0" provider-name="Team-Kodi">
-	<backwards-compatibility abi="1.3.0"/>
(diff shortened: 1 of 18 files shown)


jjd-uk:
After seeing your Apple Remote PR, I wonder with this addition and together with the other Controller stuff you've already added, whether it might make sense to split the Controller settings out to their own grouping to have something like:

```
**General**
Enable mouse and touch screen support
Peripherals

**Controllers**
Configure attached controllers
Enable rumble for notifications
Test rumble
Power off controllers on exit
```

garbear:
Good idea, I'll follow up with the new settings group when I get home from ELC.

jenkins build this please

garbear:
errors look unrelated, but just in case, jenkins build this please

PR #11713, opened 2017-02-21
Search backwards for keyframe when resuming playback

dgam1:
Currently, when resuming playback of a video, a human can easily notice how the player will jump up to several seconds ahead of the resume point. The reason being that the player will seek to the closest keyframe _after_ the position one wants to jump to.
This patch sets the backward flag so that the player jumps to the closest keyframe _before_ the resume point. (See CDVDDemuxFFmpeg::SeekTime, FFmpeg flag AVSEEK_FLAG_BACKWARD)

Instead of skipping a few seconds after resuming, the last couple of seconds will be repeated and the user will not miss content.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -1323,7 +1323,7 @@ void CVideoPlayer::Process()
     double startpts = DVD_NOPTS_VALUE;
     if (m_pDemuxer)
     {
-      if (m_pDemuxer->SeekTime(starttime, false, &startpts))
+      if (m_pDemuxer->SeekTime(starttime, true, &startpts))
         CLog::Log(LOGDEBUG, "%s - starting demuxer from: %d", __FUNCTION__, starttime);
       else
         CLog::Log(LOGDEBUG, "%s - failed to start demuxing from: %d", __FUNCTION__, starttime);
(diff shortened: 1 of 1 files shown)


FernetMenta (approved):
looks good

Rechi:
jenkins build this please

ksooo:
@FernetMenta should we backport this to Krypton? Seems low risk, but nice improvement.

FernetMenta:
Looks like a trivial change on first glance but we rely heavily on ffmpeg here. Setting the backward flag on first seek is probably not the normal case. There are many different containers with their own seek functions. To be save all those needs to be tested with this change.

dgam1:
@FernetMenta From what I can tell, every implementation of CDVDDemux's SeekTime() in xbmc/cores/VideoPlayer/DVDDemuxers/ ignores the backwards parameter, except the one in CDVDDemuxFFmpeg.cpp. Both in the master and the Krypton branch.

FernetMenta:
@dgam1 I was not referring to implementations of CDVDDemux but to containers like mkv, avi, mpegts, etc. Every container (ffmpeg calls it a format) has its own way of seeking. In particular formats without an index like mpegts may do harder in finding a key frame before the seek pos.

Not long ago I fixed an issue in ffmpeg that made seeking in mkv containers stall up to 10 seconds. This just should underpin that seeking in general is not a trivial process and even a tiny change should not be underestimated.

MartijnKaijser:
let's first test this properly on master

PR #11666, opened 2017-02-12
OMXPlayer: Fixup after DVDMessage changes

popcornmix:
## Description
The changes in DVDMessage in https://github.com/xbmc/xbmc/pull/11646 broke omxplyer.
This fixes it up.

## Types of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [X] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #11714, opened 2017-02-21
[Backport] VideoPlayer: fix edl skipping

kszaq:
This is a backport of https://github.com/xbmc/xbmc/pull/11646 and https://github.com/xbmc/xbmc/pull/11666.

Confirmed by user to fix EDL skipping in Krypton, no side effects seen in my community builds for Amlogic S805 and S905.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDMessage.h
+++ b/xbmc/cores/VideoPlayer/DVDMessage.h
@@ -87,10 +87,6 @@ public:
     VIDEO_SET_ASPECT,               // set aspectratio of video
     VIDEO_DRAIN,                    // wait for decoder to output last frame
 
-    // audio related messages
-
-    AUDIO_SILENCE,
-
     // subtitle related messages
     SUBTITLE_CLUTCHANGE,
     SUBTITLE_ADDFILE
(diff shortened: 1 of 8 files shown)


MartijnKaijser:
@FernetMenta ping

FernetMenta:
@MartijnKaijser your decision. I created this patch and I think it is ok but it is not a trivial fix and I recommend thorough testing before this goes into a final release.

MartijnKaijser:
upcoming RC would a wide test and if it gives issues we can revert. Option two i put it in 17.2

MartijnKaijser:
Jenkins build this please

lilgandhi1199:
> upcoming RC would a wide test and if it gives issues we can revert. Option two is to put it in 17.2 as that should have more nightly testing.

Is it too late to reverse this?
I'm 80% sure this has caused the audio de-sync issues I've experienced with EDL files since V 17.1 outlined here
https://github.com/xbmc/xbmc/issues/17328

lilgandhi1199:
@FernetMenta Can you reverse this patch you made in the current version of Kodi. Since 17.1 it's produced the above mentioned bug ^^^

lilgandhi1199:
@kszaq Can you tell me what this supposedly fixed?
I've only found it to make everything muting related broken

PR #11678, opened 2017-02-16
changed: disable hardcoded musicbrainz rate limiting for python scrapers

notspiff:
Python scrapers should handle this on their own.

@ronie

PR #11720, opened 2017-02-21
properly exclude musicbrainz timeout for python scrapers

ronie:
due to the change in https://github.com/xbmc/xbmc/pull/11678, bMusicBrainz now always returns false.

the issue i see with this is that kodi now ignores the scraper response to the 'resolveid' call
and continues with a 'find' call instead of 'getdetails'.

i think we only need to exclude the sleep timeout here.

@notspiff


notspiff:
doh

PR #11721, opened 2017-02-22
[Breaking Change]JSON: Do not respond with values for query "Player.Seek"

dgam1:
When sending a JSON query "Player.Seek", the return values are subject to a race condition.
In xbmc/interfaces/json-rpc/PlayerOperations.cpp, function Seek(), the command to seek will end up in a queue of the player and then the current percentage and time will be returned. Sometimes the seeking will not have concluded by that time so that the pre-seek values get returned. One way to see the unreliability of the response in action is to seek on slow internet streams. For me the ServusTV Video Addon is slow enough. But I have also witnessed this for videos on my HD (I'm currently working on something that has made me look at those numbers a lot).

At first I tried to find a fix for this problem other than removing the return values but I can't come up with a non-convoluted solution. I'm also wondering if there are many people who really want those return values anyway.

## Types of change

## Checklist:


da-anda:
ping @Montellese

Paxxi:
looks sane to me, blocking there to make sure values are correct seems like a worse solution. Caller should be able to just assume the seek happened and update it's local state directly and then resyncing.

Not my area of expertise so let's try pinging @Montellese  again.

Tolriq:
Since @Montellese is still MIA for JSON.

Any change to JSON needs adapting of https://github.com/xbmc/xbmc/blob/master/xbmc/interfaces/json-rpc/schema/version.txt

Since this is a breaking change major version must be bumped to 9.X.X and commit updating the version.txt must contains [Breaking Change] in title see https://github.com/xbmc/xbmc/commits/master/xbmc/interfaces/json-rpc/schema/version.txt for history.

--

About this PR if I may, this is a major breaking change for no real gain. Most people does use object mapping to unwrap Kodi returns and would break with this change for nothing gained.
Adding a warning in description to warn about potential unreliable value seems enough way safer seeing no gain.

See for example https://github.com/xbmc/Kore/blob/master/app/src/main/java/org/xbmc/kore/jsonrpc/method/Player.java#L270

Montellese:
I'm also wondering if this warrants a breaking change. Wouldn't it be "easier" to just add a description to the return value which states that the values can not be trusted and are deprecated? That way we don't break any clients which use the return value.

As @Tolriq mentioned when changing something in JSON-RPC we always have to update the JSON-RPC API version in `xbmc/interfaces/json-rpc/schema/version.txt`.

dgam1:
Thank you two for the info. I had assumed that the JSON version number would be changed only for releases. The change has been added to this PR for the sake of completeness.

MartijnKaijser:
ping @Montellese

Montellese:
Well as mentioned before I'm still not convinced that this change is worth a backwards incompatible change to the JSON-RPC API. Adding a description describing the potential misunderstanding of the return value would IMO be enough.

I'm fine with keeping this PR around in case there will be other backwards incompatible changes to the API until the release of Leia but just this PR doesn't warrant the extra work it will cause for every JSON-RPC client developer.

Rechi:
@dgam1 this needs a rebase

Rechi:
@Montellese after #13051 (which bumped major json-rpc version) got merged we might want to consider this again.

DaveTBlake:
Although #13051 made breaking changes to the API, hence bumped version, they were changes to lesser used aspects of the API anyway. Backwards incompatibility would only be to those using fields by position, and the changes brought more consistency and fixed broken functionality.

The change proposed here would have a far wider impact to JSON RPC clients, and for little gain. I think @Montellese is right, adding a description describing the potential misunderstanding of the return value would be enough.

Montellese:
I fully agree with @DaveTBlake. Backwards incompatible changes need to be weighed based on how big the impact is.

MartijnKaijser:
Option could always be to add it as completely new option so both can live next to each other and then jsonrpc users can slowly migrate

DaveTBlake:
@dgam1 I have just noticed that you keep rebasing this PR as it stands, but from the previous critical feedback it has received there really is little point. The change you propose is potentially disruptive for very little practical gain.  I am sorry that no one on the team has noticed this before. You may not realise it but the "this needs a rebase" messages you received  comes from an automated process, and should not be taken as encouragement or any indication that a PR is likely to be  approved or merged.

Thank you for your interest in contributing to the JSON API for Kodi, but I suggest that you close this PR because as far as I can tell it is not going to be approved in the current form.

DaveTBlake:
Previous post is an automated message, @dgam1 don't bother rebasing, spend your time more usefully

MartijnKaijser:
I'll close this PR and hopefully you are willing to do the suggested changes.

PR #11726, opened 2017-02-22
[Fix][Backport]REMIXER tag as alternative to MIXARTIST for FLAC and APE tags

DaveTBlake:
Backport of  #11725 

Add REMIXER as alternative name for MIXARTIST tag in Vorbis and APE tags. Some tagging software uses this tag name for remixer, consistent with other tag formats, so cover the alternative value.


DaveTBlake:
jenkins build this please

PR #11728, opened 2017-02-23
AESinkPULSE: Avoid deadlock

fritsch:
Fixes: http://forum.kodi.tv/showthread.php?tid=307585

deadlock between mainloop and local guard.


fritsch:
jenkins build and merge

PR #11731, opened 2017-02-23
[Fix] VideoPlayer: Delete vob sub demuxer before opening a new one (memory leak).

ace20022:


xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3599,7 +3598,8 @@ bool CVideoPlayer::OpenStream(CCurrentStream& current, int64_t demuxerId, int iS
       std::unique_ptr<CDVDDemuxVobsub> demux(new CDVDDemuxVobsub());
       if(!demux->Open(st.filename, source, st.filename2))
         return false;
-      m_pSubtitleDemuxer = demux.release();
+
+      m_pSubtitleDemuxer = std::move(demux);
```
FernetMenta: not sure about this but IMO this was wrong before. I don't think we want to keep the old demuxer in case of opening the new one fails.

ace20022:
@FernetMenta I agree, the last commit should suffice, right?

FernetMenta:
@ace20022 yes, looks good

FernetMenta (approved):

PR #11738, opened 2017-02-24
Sorting using artist sort name tags

DaveTBlake:
Add processing of ARTISTSORT (TSOP), ALBUMARTISTSORT (TS02) and COMPOSERSORT (TSOC) tags to provide a sort name for artists. Optionally use this field when sorting artists, albums or songs by artist name. 

This bumps the version of the music database to MyMusic63 adding fields to the album, artstist and song tables, and does some tidy up of redundant fields at the same time.

To see the use you need to (re)scan music files into the library that have at least some of these tags embedded, and then add `<useartistsortname>true</useartistsortname>` to the <musiclibrary> section of advancesettings.xml. When the artist, album or songs nodes is sorted by artist name it is the sort name that will be used e.g. "Eric Clapton" with sortname "Clapton, Eric" can be seen under "C" for Clapton rather than "E" for Eric.

This is the first phase of implementation, and makes the basic feature available without any impact to current use.  To follow is adding a UI setting for this option, an info label (so sort name could also be used as a latin script alias name for non-western artists), and to make available to JSON API.  I hope to discover from initial use the best way to make this available via the UI.


xbmc/music/Album.cpp
```diff
@@ -54,10 +54,13 @@ CAlbum::CAlbum(const CFileItem& item)
   genre = tag.GetGenre();
   std::vector<std::string> musicBrainzAlbumArtistHints = tag.GetMusicBrainzAlbumArtistHints();
   strArtistDesc = tag.GetAlbumArtistString();
+  strArtistSort = tag.GetAlbumArtistSort();
+  //Split the artist sort string to try and get sort names for individual artists
+  std::vector<std::string> artistSort = StringUtils::Split(strArtistSort, g_advancedSettings.m_musicItemSeparator);
```
Paxxi: I would use `auto` here to avoid some typing.

xbmc/music/Album.cpp
```diff
@@ -135,14 +138,19 @@ CAlbum::CAlbum(const CFileItem& item)
         musicBrainzAlbumArtistHints = tag.GetAlbumArtist();
     }
 
+    // Try to get number of artist sort names and musicbrainz ids to match. Split sort names 
+    // further using multiple possible delimiters, over single separator applied in Tag loader
+    if (artistSort.size() != tag.GetMusicBrainzAlbumArtistID().size())
+      artistSort = StringUtils::SplitMulti(artistSort, { ";", ":", "|", "#" });
+
     for (size_t i = 0; i < tag.GetMusicBrainzAlbumArtistID().size(); i++)
```
Paxxi: use `for ( auto& id : tag.GetMusicBranzAlbumArtistID())` instead

xbmc/music/Album.cpp
```diff
@@ -166,9 +179,17 @@ CAlbum::CAlbum(const CFileItem& item)
       // Split album artist names further using multiple possible delimiters, over single separator applied in Tag loader
       albumArtists = StringUtils::SplitMulti(albumArtists, g_advancedSettings.m_musicArtistSeparators);
 
-    for (auto artistname : albumArtists)
+    if (artistSort.size() != albumArtists.size())
+      // Split artist sort names further using multiple possible delimiters, over single separator applied in Tag loader
+      artistSort = StringUtils::SplitMulti(artistSort, { ";", ":", "|", "#" });
+
+    for (size_t i = 0; i < albumArtists.size(); i++)
```
Paxxi: same about the for auto loop

xbmc/music/MusicDatabase.cpp
```diff
@@ -1194,6 +1228,53 @@ bool CMusicDatabase::UpdateArtist(const CArtist& artist)
...
+    if (NULL == m_pDS.get()) return -1;
+
+    strSQL = PrepareSQL("SELECT strArtist, strSortName FROM artist WHERE idArtist = %i", idArtist);
+    m_pDS->query(strSQL);
+    if (m_pDS->num_rows() != 1)
+    {
+      m_pDS->close();
+      return -1;
+    }
+    strArtistName = m_pDS->fv("strArtist").get_asString();
```
Paxxi: I would move the declarations here as they aren't used outside the try block there's no real reason to have a separate declaration for it. Makes a tiny bit harder to know if they're actually empty or something is overwritten.

Paxxi (commented):
Looks good to me, the comments are suggestions, can be merged as is.

xbmc/music/Album.cpp
DaveTBlake: Since it is checking the element of one array against the element in same position in another using the array index makes more sense to me. Or is there another more elegent way to do this?

xbmc/music/Album.cpp
DaveTBlake: Same reason for index - manipulating two arrays of the same size

xbmc/music/Album.cpp
Paxxi: oh didn't notice that. I caught it in another place where I was starting to comment and cancelled it. Yeah ignore my comments about the for loops.

DaveTBlake:
jenkins build this please

MartijnKaijser:
[  FAILED  ] 1 test, listed below:
[  FAILED  ] TestSortUtils.GetFieldsForSorting

DaveTBlake:
Tests adjusted.
jenkins build this please

Rechi:
jenkins build this please

Rechi:
jenkins build this please

Rechi:
jenkins build this please

MartijnKaijser:
Jenkins build this please

MartijnKaijser:
jenkins build this please

PR #11741, opened 2017-02-24
[xbmc] Start work on getting rid of our emulated GetDiskFreeSpaceEx method

Paxxi:
Decided to unify some of our low level file handling
methods on the std::filesystem idea. As it's not standardized
yet and implementations are experimental this is a simple
version that suits our current needs. The idea is that once platforms
start offering non-experimental versions of std::filesytem we can
easily move over to using them.

## How Has This Been Tested?

runtime tested on windows

## Types of change

## Checklist:

### DIFF
--- a/xbmc/platform/posix/Filesystem.cpp
+++ b/xbmc/platform/posix/Filesystem.cpp
@@ -22,17 +22,12 @@
 #include "system.h"
 #include "filesystem/SpecialProtocol.h"
 
-#ifdef TARGET_POSIX
-#include <sys/stat.h>
-#endif
-#if !defined(TARGET_DARWIN) && !defined(TARGET_FREEBSD) && !defined(TARGET_ANDROID)
-#include <sys/vfs.h>
-#else
+#if defined(TARGET_LINUX)
+#include <sys/statvfs.h>
+#elif defined(TARGET_DARWIN) || defined(TARGET_FREEBSD)
 #include <sys/param.h>
 #include <sys/mount.h>
-#endif
-
-#if defined(TARGET_ANDROID)
+#elif defined(TARGET_ANDROID)
 #include <sys/statfs.h>
 #endif
 
@@ -47,13 +42,13 @@ space_info space(const std::string& path, std::error_code& ec)
 {
   ec.clear();
   space_info sp;
-#if defined(TARGET_ANDROID) || defined(TARGET_DARWIN)
+#if defined(TARGET_LINUX)
+  struct statvfs64 fsInfo;
+  auto result = statvfs64(CSpecialProtocol::TranslatePath(path).c_str(), &fsInfo);
+#else
   struct statfs fsInfo;
   // is 64-bit on android and darwin (10.6SDK + any iOS)
   auto result = statfs(CSpecialProtocol::TranslatePath(path).c_str(), &fsInfo);
-#else
-  struct statfs64 fsInfo;
-  auto result = statfs64(CSpecialProtocol::TranslatePath(path).c_str(), &fsInfo);
 #endif
 
   if (result != 0)
--- a/xbmc/platform/posix/Filesystem.cpp
+++ b/xbmc/platform/posix/Filesystem.cpp
@@ -1,0 +1,75 @@
+/*
+ *      Copyright (C) 2005-2017 Team Kodi
+ *      http://xbmc.org
+ *
+ *  This Program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2, or (at your option)
+ *  any later version.
+ *
+ *  This Program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with XBMC; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "platform/Filesystem.h"
+#include "system.h"
+#include "filesystem/SpecialProtocol.h"
+
+#ifdef TARGET_POSIX
+#include <sys/stat.h>
+#endif
+#if !defined(TARGET_DARWIN) && !defined(TARGET_FREEBSD) && !defined(TARGET_ANDROID)
+#include <sys/vfs.h>
+#else
+#include <sys/param.h>
+#include <sys/mount.h>
+#endif
+
+#if defined(TARGET_ANDROID)
+#include <sys/statfs.h>
+#endif
+
+namespace KODI
+{
+namespace PLATFORM
+{
+namespace FILESYSTEM
+{
+
+space_info space(const std::string& path, std::error_code& ec)
+{
+  ec.clear();
+  space_info sp;
+#if defined(TARGET_ANDROID) || defined(TARGET_DARWIN)
+  struct statfs fsInfo;
+  // is 64-bit on android and darwin (10.6SDK + any iOS)
+  auto result = statfs(CSpecialProtocol::TranslatePath(path).c_str(), &fsInfo);
+#else
+  struct statfs64 fsInfo;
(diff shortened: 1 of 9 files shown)


Paxxi:
jenkins build this please

MartijnKaijser:
jenkins build this please

Paxxi:
@Montellese thoughts?
maybe @wsnipex for linux and @koying for android and @Memphiz for osx

xbmc/platform/posix/Filesystem.cpp
```diff
@@ -0,0 +1,75 @@
...
+{
+  ec.clear();
+  space_info sp;
+#if defined(TARGET_ANDROID) || defined(TARGET_DARWIN)
+  struct statfs fsInfo;
+  // is 64-bit on android and darwin (10.6SDK + any iOS)
+  auto result = statfs(CSpecialProtocol::TranslatePath(path).c_str(), &fsInfo);
+#else
+  struct statfs64 fsInfo;
+  auto result = statfs64(CSpecialProtocol::TranslatePath(path).c_str(), &fsInfo);
```
wsnipex: quote from the linux standards base: https://refspecs.linuxbase.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/baselib-statfs64.html
>Note: Application developers should use the statvfs64() function to obtain general file system information. Applications should only use the statfs64() function if they must determine the file system type, which need not be provided by statvfs64().

xbmc/utils/SystemInfo.cpp
```diff
@@ -451,67 +453,64 @@ const std::string& CSysInfo::GetAppName(void)
...
+      } while (strlen(pcBuffer + iPos) > 0);
     }
     delete[] pcBuffer;
 #else // for linux and osx
-    if( GetDiskFreeSpaceEx( "/", NULL, &ULTotal, &ULTotalFree ) )
-    {
-      ULTotalTmp.QuadPart+= ULTotal.QuadPart;
-      ULTotalFreeTmp.QuadPart+= ULTotalFree.QuadPart;
-    }
+    total = space("/", ec);
```
koying: This actually makes no sense on unix-derived platforms ;)
Not your fault, obviously

koying:
Wouldn't hooking into our StorageProvider a valid alternative?

xbmc/platform/posix/Filesystem.cpp
Paxxi: I just copied the code from current implementation, will fix

xbmc/utils/SystemInfo.cpp
Paxxi: Yeah it should probably fetch the mounts and iterate over them as on windows?

Paxxi:
Our storage provider is a bit higher in the hierarchy for what I want to achieve with the filesystem methods, intending to implement stat and possibly the directory iterators so we have a common low level abstraction and can reduce some ifdefs. 

Using storage manager for the sysinfo might be a good idea, will have a look at it and see.

xbmc/utils/SystemInfo.cpp
koying: Yeah, but even that doesn't make much sense to get "Free Disk Space"...
What is Windows doing when not passed an actual path?

Paxxi: Windows only cares about a disk or mount so we pass it c: d: and so on. 
Possible to mount in directories like Linux but afaik I've never seen it used on windows so we don't support it.  

Suggestions on what would make sense?

Paxxi: Right, looking at IStorageProvider the GetDiskUsage method might be suitable to use but maybe add an overloaded version that returns the numbers and not formatted strings

koying: > Suggestions on what would make sense

No, really...
Seems only used from https://github.com/xbmc/xbmc/blob/61f212f75c69a7478b93cf87075dc305734320b4/xbmc/utils/SystemInfo.cpp#L1045 and the whole concept of "GetHddSpaceInfo" seems bogus to me ;)

koying: The thing is untouched since years. I bet it still dates from single disk / single partition xbox ;)

Paxxi: and that method is called from CGUIInfoManager so it's likely that some random skin out there uses it, I'm all for dropping it but maybe @phil65 or @ronie has something to say about it.

ronie: correct, there are a number of skins that use the System.FreeSpace / System.TotalSpace infolabels.

they are either shown in some home screen widget or in the filemanager window.

xbmc/platform/posix/Filesystem.cpp
```diff
@@ -0,0 +1,75 @@
...
+#endif
+
+  if (result != 0)
+  {
+    ec.assign(result, std::system_category());
+    sp.available = static_cast<uintmax_t>(-1);
```
thebrid: Consider using std::numeric_limits<uintmax_t>::max()

Paxxi: standards draft uses static_cast http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0218r0.html#space_info

Doesn't make any real difference but avoids an include :)

xbmc/platform/posix/Filesystem.cpp
```diff
@@ -0,0 +1,70 @@
...
+    sp.capacity = static_cast<uintmax_t>(-1);
+    sp.free = static_cast<uintmax_t>(-1);
+    return sp;
+  }
+  sp.available = static_cast<uintmax_t>(fsInfo.f_bavail * fsInfo.f_bsize);
```
wsnipex: note: while it's probably negligible for our purpose, but this is incorrect when kodi is run as root, because  file systems reserve a certain percentage of the disk space for root(usually 5%-10%)
I'd still keep it as is, because distros should really stop running kodi as root

xbmc/platform/posix/Filesystem.cpp
thebrid: :+1:

xbmc/platform/posix/Filesystem.cpp
Paxxi: Reading https://www.ibm.com/support/knowledgecenter/ssw_i5_54/apis/statvf64.htm it seems like a strange design that it specifically mentions non-privileged users. I'm thinking it should report the correct amount depending on user and quotas but maybe that's left to another api?

The draft standards mentions it should report the f_bavail * f_bsize so we should be good in that regard http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0218r0.html#space_info

stefansaraev: first of all, not all filesystem reserve root-only blocks. and second, you dont have to care about that. no matter if kodi is run as root or not. those blocks are not to be reported in a different way depending on current uid. even fs specific tools like "df" dont care, you should not. too.

Paxxi: Thank you for the explanation, at least we're all in agreement on this line of code. Let the battle over root vs nonroot continue in all its glory 😊

xbmc/platform/posix/Filesystem.cpp
```diff
@@ -0,0 +1,70 @@
...
+ *
+ */
+
+#include "platform/Filesystem.h"
+#include "system.h"
+#include "filesystem/SpecialProtocol.h"
+
+#if defined(TARGET_LINUX)
+#include <sys/statvfs.h>
+#elif defined(TARGET_DARWIN) && defined(TARGET_FREEBSD)
```
Memphiz: Should this be an "or" instead?

Paxxi: It certainly should!

Memphiz:
Looks straight forward to me

Paxxi:
Fixed the invalid test for TARGET_BSD / TARGET_DARWIN

Dropped the counting of "total" space on win32 as it's a useless metric, added some comments but kept the GetDiskspaceInfo method for backward compat with skins that rely on it.

jenkins build this please

Paxxi:
jenkins build this please

Paxxi:
build error unrelated.

PR #11734, opened 2017-02-24
[estuary] sync pt2

phil65:
@ronie

PR #11743, opened 2017-02-25
[estuary] re-add PVRChannelNumberInput to fullscreenvideo

ronie:
fix regression introduced by https://github.com/xbmc/xbmc/pull/11734

### DIFF
--- a/addons/skin.estuary/xml/DialogSeekBar.xml
+++ b/addons/skin.estuary/xml/DialogSeekBar.xml
@@ -7,6 +7,7 @@
 	<depth>DepthOSD</depth>
 	<zorder>0</zorder>
 	<controls>
+		<include>PVRChannelNumberInput</include>
 		<control type="group">
 			<bottom>0</bottom>
 			<height>190</height>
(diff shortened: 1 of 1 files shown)


ksooo:
@ronie Thanks for taking care, but this is only half of the fix. Another regression introduced by phil's optimization is:

- Tune to a channel, when in fullscreen livetv, open the channel OSD
- now use numeric keys to enter a number
- the number appears on screen, that good
- but together with the number, the seekbar appears (behind the channel dialog), which is wrong

screenshot001

To address this, my original PR contained a visibility condition which is now gone.

> https://github.com/xbmc/xbmc/pull/11434/files#diff-ead36a199046333224f68d2eac9f4d5fR3

This part I mean: <code>| [[Window.IsActive(fullscreenvideo) | Window.IsActive(visualisation)] + !Window.IsActive(pvrosdchannels) + !String.IsEmpty(PVR.ChannelNumberInput)]</code>

ronie:
thanx for testing!

fix here: https://github.com/xbmc/xbmc/pull/11745

PR #11745, opened 2017-02-25
[estuary] hide seekbar when pvr osd channels is visible

ronie:
fix regression introduced by latest skin sync.

@ksooo


ksooo:
Runtime-tested. Works.

PR #11746, opened 2017-02-25
[PVR] Trac 17339: Remove non-working recording settings defaults.

ksooo:
Fixes http://trac.kodi.tv/ticket/17339

This PR removes the following settings, as they as explained by me in the ticket, can only work by chance for conceptional reasons (there are no defaults across pvr client addons, every client can define its own set of recordings values and defaults...):

screenshot000

Please note that start and end time padding defaults are intentionally still available, as for those there is a semantics defined across all addons.

This change has been runtime-tested on latest master on macOS.

At our skinners: Yeah, three spinners less. ;-)

@Jalle19 for review?


Jalle19 (approved):

ksooo:
jenkins build this please

PR #5958, opened 2014-12-16
Fix state not saved after #5842 if playing from "Recently added".

ghost:
See:
- http://forum.kodi.tv/showthread.php?tid=210479
- https://github.com/xbmc/xbmc/pull/5842#issuecomment-67107411
- http://trac.kodi.tv/ticket/15603
- http://trac.kodi.tv/ticket/15660
## 

Playing something from "Recently added" creates a play list with a single item.  Oops.
## 

**Needs backport to Helix**

PR #11747, opened 2017-02-25
PlayListPlayer: New function the ensure items are played with a playlist

dgam1:
There are cases where playback of files doesn't create a playlist. @Montellese made a comment about that here: https://github.com/xbmc/xbmc/pull/5958#issuecomment-67295177

One of those cases is when you start playback of an item over JSON like this:
`{"params": {"item": {"file": "/videos/video.mp4"}}, "jsonrpc": "2.0", "id": 1, "method": "Player.Open"}`
(Or, more generally, every time TMSG_MEDIA_PLAY gets sent with a single item attached)

Another is where you start playback of a file over the Kodi file manager. What's worse here is that the old playlist doesn't get deleted before playback.

To take care of that I added a new overloaded function Play() to the PlayListPlayer that takes an item as argument, creates a new playlist for it and starts the playback.

For the second commit I found two instances that can use the new function:
1) In CGUIWindowVideoBase::PlayMovie(), when opening a file from the video browser
2) In CGUIWindowMusicBase::OnPlayMedia(), when opening a file from the music browser while the setting to automatically queue the next item is disabled

## Types of change

## Checklist:

### DIFF
--- a/xbmc/PlayListPlayer.cpp
+++ b/xbmc/PlayListPlayer.cpp
@@ -256,6 +256,24 @@ bool CPlayListPlayer::PlaySongId(int songId)
   return Play();
 }
 
+bool CPlayListPlayer::Play(const CFileItemPtr &pItem, std::string player)
+{
+  int playlist;
+  if (pItem->IsAudio())
+    playlist = PLAYLIST_MUSIC;
+  else if (pItem->IsVideo())
+    playlist = PLAYLIST_VIDEO;
+  else
+    return false;
+
+  ClearPlaylist(playlist);
+  Reset();
+  SetCurrentPlaylist(playlist);
+  Add(playlist, pItem);
+
+  return Play(0, player);
+}
+
 bool CPlayListPlayer::Play(int iSong, std::string player, bool bAutoPlay /* = false */, bool bPlayPrevious /* = false */)
 {
   if (m_iCurrentPlayList == PLAYLIST_NONE)
@@ -855,10 +873,14 @@ void PLAYLIST::CPlayListPlayer::OnApplicationMessage(KODI::MESSAGING::ThreadMess
 
         ClearPlaylist(playlist);
         SetCurrentPlaylist(playlist);
-        //For single item lists try PlayMedia. This covers some more cases where a playlist is not appropriate
-        //It will fall through to PlayFile
         if (list->Size() == 1 && !(*list)[0]->IsPlayList())
-          g_application.PlayMedia(*((*list)[0]), pMsg->strParam, playlist);
+        {
+          CFileItemPtr item = (*list)[0];
+          if (item->IsAudio() || item->IsVideo())
+            Play(item, pMsg->strParam);
+          else
+            g_application.PlayMedia(*item, pMsg->strParam, playlist);
+        }
         else
         {
           // Handle "shuffled" option if present
--- a/xbmc/PlayListPlayer.h
+++ b/xbmc/PlayListPlayer.h
@@ -72,6 +72,11 @@ public:
   bool PlaySongId(int songId);
   bool Play();
 
+  /*! \brief Creates a new playlist for an item and starts playing it
+   \param pItem The item to play.
+   */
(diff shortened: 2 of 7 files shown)


rmrector:
Nice. [This block](https://github.com/xbmc/xbmc/blob/a077d7b954189dc506231eca504baf9cbc8f810a/xbmc/interfaces/builtins/PlayerBuiltins.cpp#L474) currently plays a single video without a playlist, can you rig the new function up here as well? This is the built-in PlayMedia action, and it's also used for skin containers not on a media window, like Home widgets.

dgam1:
This ought to do it.

dgam1:
Rebased. As of PR #10857 there is another use case for the new function.

Razzeee:
jenkins build this please

Razzeee:
Reads fine to me and also builds fine.

I would like to merge this in the coming days.

Razzeee:
Thanks @dgam1 

jenkins build and merge

ksooo:
@dgam1 , @Razzeee This PR breaks PVR. Example: Open Channel window, select a channel, open its context menu. Select "switch to channel" from context menu. Nothing happens. 

functionality is triggered from here: https://github.com/xbmc/xbmc/blob/master/xbmc/video/ContextMenus.cpp#L168

dgam1:
Ah, I don't have PVR hardware and so hadn't tested that. Should have mentioned it.

If FileItem's IsVideo returns false for LiveTV, does IsAudio also return false for PVR radio?

ksooo:
IsAudio() returns true for PVR radio.

dgam1:
Thanks. Now I have to assume that if you open a LiveTV channel over a home widget, it will not create a playlist for that item, right?

If my assumption is correct, this commit should fix that (short of having IsVideo return true for LiveTV): https://github.com/dgam1/xbmc/commit/e982422d1e416dec2443ef91981164256dd53cb0

Would you mind testing that?

ksooo:
@dgam1 have you seen #11981 ? Or are you trying to fix something else with the above mentioned commit?

dgam1:
@ksooo Yes, I have have seen it and based on it I assume this PR missed another case where playback of an item doesn't create a playlist. That case being when you open a LiveTV item over a home widget. So my new commit not only checks IsAudio and IsVideo but also IsLiveTV, which can now be handled thanks to your PR.

ksooo:
Okay, now I got it. Yes, your changes, in addition to #11981 do fix the home screen widget in the sense that now a playlist for Live TV gets created.

But, does it actually make sense to create a playlist for Live TV? I know this has been the case before, but ... I suddenly feel like this could be nonsense. ;-)

ksooo:
Maybe we should try to play Live TV without creating a playlist?

ksooo:
I think this is the right way to do it: https://github.com/ksooo/xbmc/commit/dac2902c7d7b0a8ee0ec54a79ec0599b4c0d5f5e

ksooo:
^^^ The actual bug was to create a playlist for Live TV in <code>static void SetPathAndPlay(CFileItem& item)</code> ... which at the end led to the confusion (on my end).

dgam1:
It seems to me that "everything that is being played is in a playlist" is the overall easier route to go but it appears like that's something on which you devs haven't fully come to an agreement yet :)

Regarding the commit you just posted: Should an already existing playlist be cleared before starting LiveTV? Is radio another case that requires special handling? And for that matter, what about internet live streams in general, do they belong in a playlist?

ksooo:
> It seems to me that "everything that is being played is in a playlist" is the overall easier route to go

 I generally agree with you. It seems to be quite a mess currently. But I have to admit that I'm not very deep inside the playlist stuff. 

 Why is a playlist not just a list containing arbitrary playable items, for example a mix of two videos, a Live TV channel, then two songs. Why is there a separation between music and video? But I guess this is something we should discuss in the Kodi forum.

 For now, I would be happy to have it consistent - no playlist for Live TV.

PR #11753, opened 2017-02-25
[PVR] fix timer type display for 'invalid' types

metaron-uk:
Allows 'invalid' timer types to be correctly displayed in the Timer Settings dialog
(See forum http://forum.kodi.tv/showthread.php?tid=307538)

## Description
Currently a timer type can be 'skipped' when building the list of valid timer types to present in the Timer Settings dialog. The result is that not all timer types are displayed properly.
This change adds the current type to the list if it would otherwise be skipped so the settings can be displayed correctly.

## Motivation and Context
Ensures timers are always displayed correctly.

## How Has This Been Tested?
On x86 (linux) with pvr.mythtv client.

## Screenshots:
Display of an 'Unhandled' timer type
screenshot062
After (How this type should be presented):
screenshot063
screenshot064
Before (How this type was previously presented):
screenshot065
screenshot066

## Types of change

## Checklist:

### DIFF
--- a/xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
+++ b/xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
@@ -735,7 +735,8 @@ void CGUIDialogPVRTimerSettings::InitializeTypesList()
     return;
   }
 
-  int idx = 0;
+  bool bFoundThisType(false);
+  int idx(0);
   const std::vector<CPVRTimerTypePtr> types(CPVRTimerType::GetAllTypes());
   for (const auto &type : types)
   {
@@ -750,11 +751,11 @@ void CGUIDialogPVRTimerSettings::InitializeTypesList()
       continue;
 
     // Drop TimerTypes that require EPGInfo, if none is populated
-    if (m_bIsNewTimer && type->RequiresEpgTagOnCreate() && !m_timerInfoTag->GetEpgInfoTag())
+    if (type->RequiresEpgTagOnCreate() && !m_timerInfoTag->GetEpgInfoTag())
       continue;
 
     // Drop TimerTypes without 'Series' EPG attributes if none are set
-    if (m_bIsNewTimer && type->RequiresEpgSeriesOnCreate())
+    if (type->RequiresEpgSeriesOnCreate())
     {
       const EPG::CEpgInfoTagPtr epgTag(m_timerInfoTag->GetEpgInfoTag());
       if (epgTag && !epgTag->IsSeries())
@@ -762,15 +763,21 @@ void CGUIDialogPVRTimerSettings::InitializeTypesList()
     }
 
     // Drop TimerTypes that forbid EPGInfo, if it is populated
-    if (m_bIsNewTimer && type->ForbidsEpgTagOnCreate() && m_timerInfoTag->GetEpgInfoTag())
+    if (type->ForbidsEpgTagOnCreate() && m_timerInfoTag->GetEpgInfoTag())
       continue;
 
     // Drop TimerTypes that aren't rules if end time is in the past
     if (!type->IsTimerRule() && m_timerInfoTag->EndAsLocalTime() < CDateTime::GetCurrentDateTime())
       continue;
 
+    if (!bFoundThisType && *type == *m_timerType)
+      bFoundThisType = true;
+
     m_typeEntries.insert(std::make_pair(idx++, type));
   }
+
+  if (!bFoundThisType)
+    m_typeEntries.insert(std::make_pair(idx++, m_timerType));
 }
 
 void CGUIDialogPVRTimerSettings::InitializeChannelsList()


metaron-uk:
@ksooo - if you like this, let me know and I'll create a backport PR as I think it ought to go in the next Krypton point release as well.
@janbar - have you come across this problem yourself? If so, what do you think of my proposed solution.

ksooo:
Yes, your patch ensures that it cannot happen anymore that the timer type of the current timer tag is omitted from the timer types list, but could you please explain how it can happen that a tag is given to the dialog that leads to this problem?

metaron-uk:
* if (type->ForbidsNewInstances()) because only the backend has the necessary UI to create timers of this type, but Kodi can still tweak an existing one (as in the screeshot  example)
* if for some other reason the conditions to create a new rule timer of this type aren't met (although you seem to have fixed a lot of these with your latest PR :-)

ksooo:
The correct fix for the "forbidsnewinstances" case would be to add  "m_bIsNewTimer &&" in front of the respective if clause in the timer types init member function. Seems I forgot this with my latest PR.

Are there any other real cases? Then we should fix those one by one.

Adding the "golden hammer" like you suggest only hides the actual bugs leading to filtering out a type, imo.

metaron-uk:
Adding "m_bIsNewTimer &&" in front covers the 'new timer' case, but what about modifying an existing timer rule of another type?
If the client 'ForbidsNewInstances()' of that timer type, then surely it shouldn't be available for selection as an alternative.
In the same way, if type->RequiresEpgTagOnCreate() && !m_timerInfoTag->GetEpgInfoTag() etc... those wouldn't be valid options to select either.
imo, the only time that an 'invalid' type should appear in the list is when the timer already has that type (aka the "golden hammer").

ksooo:
So, basically you are saying to revert the other "m_bIsNewTimer &&" changes a did in my latest PR and to use the "golden hammer" instead?

metaron-uk:
That would be my recommendation, yes.

janbar:
@metaron-uk , I thought "unhandled" was treated as presented here, regarding its declaration:

    PVR_TIMER_TYPE_IS_REPEATING |
    PVR_TIMER_TYPE_FORBIDS_NEW_INSTANCES |
    PVR_TIMER_TYPE_SUPPORTS_PRIORITY |
    PVR_TIMER_TYPE_SUPPORTS_LIFETIME |
    PVR_TIMER_TYPE_SUPPORTS_RECORDING_GROUP

Seems that isn't the case. I have never been able to test it. But your first screenshot "After (How this type should be presented)" is right for me and this is what I expected.

ksooo:
@metaron-uk you are right. Could you please change this PR to revert my last changes and to implement the "golden hammer"?

xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
```diff
@@ -769,8 +770,13 @@ void CGUIDialogPVRTimerSettings::InitializeTypesList()
     if (!type->IsTimerRule() && m_timerInfoTag->EndAsLocalTime() < CDateTime::GetCurrentDateTime())
       continue;
 
+    if (type == m_timerType)
```
ksooo: Please change to <code>if (!foundThisType && *type == *m_timertype)</code>.

First is for efficiency, to only compare types if actually needed.
Second is to not compare the timer type pointers but the actual objects.

xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
```diff
@@ -735,7 +735,8 @@ void CGUIDialogPVRTimerSettings::InitializeTypesList()
     return;
   }
 
-  int idx = 0;
+  bool foundThisType(false);
```
ksooo: I would prefer if you use a type prfix for the variable name => bFoundThisType
We do not do this in a consistent way in our code base, but to follow our coding guidelines we should obey the guideline for new code.

xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
```diff
@@ -769,8 +770,13 @@ void CGUIDialogPVRTimerSettings::InitializeTypesList()
     if (!type->IsTimerRule() && m_timerInfoTag->EndAsLocalTime() < CDateTime::GetCurrentDateTime())
       continue;
 
+    if (type == m_timerType)
+      foundThisType = true;
+
     m_typeEntries.insert(std::make_pair(idx++, type));
   }
+  if (!foundThisType)
```
ksooo: Pleaes insert a blank line above this line.

ksooo (changes requested):

metaron-uk:
> @metaron-uk , I thought "unhandled" was treated as presented here, regarding its declaration:
> 
> PVR_TIMER_TYPE_IS_REPEATING |
> PVR_TIMER_TYPE_FORBIDS_NEW_INSTANCES |
> PVR_TIMER_TYPE_SUPPORTS_PRIORITY |
> PVR_TIMER_TYPE_SUPPORTS_LIFETIME |
> PVR_TIMER_TYPE_SUPPORTS_RECORDING_GROUP
>
> Seems that isn't the case. I have never been able to test it. But your first screenshot "After (How this type should be presented)" is right for me and this is what I expected.

@janbar You are right, I have been experimenting...

xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
metaron-uk: OK

xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
metaron-uk: Good catch.

xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp
metaron-uk: done

metaron-uk:
@ksooo not sure why github is showing both the commits in your last pull request - I only reverted one of them. Let me know if you want this squashed into one.

ksooo:
Please rebase to latest master and then squash the changes of the PR to a single commit.

metaron-uk:
@ksooo - rebased as requested and checked on x86 linux - good to go.

ksooo:
jenkins build this please

ksooo (approved):
thanks

PR #11754, opened 2017-02-25
store arbitrary artist art in art table

ronie:
store artist artwork (artist logo, artist banner, etc..) in the art table.
skins will be able to display this artwork using ListItem.Art()

we already have this functionality for all other media types (movies, tvshows, albums, etc..)
but it was never hooked up to artists.

@DaveTBlake

### DIFF
--- a/xbmc/music/Artist.cpp
+++ b/xbmc/music/Artist.cpp
@@ -49,6 +49,11 @@ void CArtist::MergeScrapedArtist(const CArtist& source, bool override /* = true 
   thumbURL = source.thumbURL;
   fanart = source.fanart;
   discography = source.discography;
+  for (std::map<std::string, std::string>::const_iterator i = source.art.begin(); i != source.art.end(); ++i)
+  {
+    if (override || art.find(i->first) == art.end())
+      art[i->first] = i->second;
+  }
 }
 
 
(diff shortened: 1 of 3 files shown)


DaveTBlake:
>it was never hooked up to artists.

But we do already store scraped artist art in the art table (from CMusicInfoScanner). Also there is an user reported issue of artist art getting stuck as the wrong image , so I would like to check this out a bit more.

jjd-uk:
@DaveTBlake I believe ronie is waiting on your approval for this one.

DaveTBlake:
Sorry for delay @ronie I lost sight of this, thanks for reminder @jjd-uk 

On the one hand it makes sense to make artist art consistent with how album art is handled, so thanks for this.  However I am having difficulty actually invoking this code in testing, and I'm not sure how to cause the  album art processing it replicates either.  Perhaps you can help me out? Maybe I'm holding it wrong? 

How does any additional arbitrary art (banners, logos etc.) _for music_ get into Kodi?

DaveTBlake:
> How does any additional arbitrary art (banners, logos etc.) for music get into Kodi?

Is the answer that currently they don't, but that #11755 would make it possible via the new Python scrapers? So does that also mean that the existing CAlbum implementation is redundant until then too, or am I missing something?

Rechi:
@ronie this needs a rebase

DaveTBlake:
Closing as changes included in #12891, thanks for the inspiration @ronie

PR #11755, opened 2017-02-26
python music scrapers - art

ronie:
this is my clumsy attempt to let the python music scrapers set some additional artwork.

(needs https://github.com/xbmc/xbmc/pull/11754)

@notspiff

### DIFF
--- a/xbmc/addons/Scraper.cpp
+++ b/xbmc/addons/Scraper.cpp
@@ -773,6 +773,13 @@ void DetailsFromFileItem<CAlbum>(const CFileItem &item, CAlbum &album)
 
   int nThumbs = item.GetProperty("album.thumbs").asInteger32();
   ParseThumbs(album.thumbURL, item, nThumbs, "album.thumb");
+
+  std::map<std::string, std::string> artwork;
+  std::string arttype = FromString(item, "album.cdart.url");
+  if (!arttype.empty())
+    artwork.insert(std::make_pair("cdart", arttype));
+  if (!artwork.empty())
+    album.art = artwork;
 }
 
 template<>
@@ -807,6 +814,16 @@ void DetailsFromFileItem<CArtist>(const CFileItem &item, CArtist &artist)
   int nFanart = item.GetProperty("artist.fanarts").asInteger32();
   artist.fanart.m_xml = ParseFanart(item, nFanart, "artist.fanart");
   artist.fanart.Unpack();
+
+  std::map<std::string, std::string> artwork;
+  std::string arttype = FromString(item, "artist.clearlogo.url");
+  if (!arttype.empty())
+    artwork.insert(std::make_pair("clearlogo", arttype));
+  arttype = FromString(item, "artist.banner.url");
+  if (!arttype.empty())
+    artwork.insert(std::make_pair("banner", arttype));
+  if (!artwork.empty())
+    artist.art = artwork;
 }
 
 template<>


xbmc/addons/Scraper.cpp
```diff
@@ -751,6 +751,13 @@ void DetailsFromFileItem<CAlbum>(const CFileItem& item, CAlbum& album)
 
   int nThumbs = item.GetProperty("album.thumbs").asInteger();
   ParseThumbs(album.thumbURL, item, nThumbs, "album.thumb");
+
+  std::map<std::string, std::string> artwork;
+  std::string cdart = FromString(item, "album.cdart.url");
+  if (!cdart.empty())
+    artwork.insert(std::pair<std::string, std::string>("cdart", cdart));
```
notspiff: std::make_pair("cdart", cdart)

notspiff:
looks fine, better with make_pair

xbmc/addons/Scraper.cpp
```diff
@@ -785,6 +792,16 @@ void DetailsFromFileItem<CArtist>(const CFileItem& item, CArtist& artist)
...
   artist.fanart.Unpack();
+
+  std::map<std::string, std::string> artwork;
+  std::string logo = FromString(item, "artist.clearlogo.url");
+  if (!logo.empty())
+    artwork.insert(std::pair<std::string, std::string>("clearlogo", logo));
+  std::string banner = FromString(item, "artist.banner.url");
+  if (!banner.empty())
+    artwork.insert(std::pair<std::string, std::string>("banner", banner));
+  if (!artwork.empty())
```
xhaggi: Minor, but you could re-use the same var for every value if you only want to transfer it into the map.

```
std::string value = FromString(...
if (!value.empty())
  // transfer value into map
value = FromString(...
if (!value.empty())
  // transfer value into map

```

Razzeee:
@ronie 
do you want to adapt this before we merge?

ronie:
sure, i'll look into it asap.. hopefully next week

jjd-uk:
Does this still need changes?

@DaveTBlake for review as well

ronie:
> Does this still need changes?

updated now with the suggestions made above

DaveTBlake:
Any reason for adding "cdart" for albums, and "banner" and "clearlogo" for artists?

zag2me:
Back Album cover and Spine album cover would also be nice. Both are supported by TheAudioDB and have already been integrated into at least one skin by hardcoding the filenames.

See: http://forum.kodi.tv/showthread.php?tid=313352&pid=2586744#pid2586744

DaveTBlake:
Is there an easy way to make the suppported artwork pairs configurable rather than hard coded?

zag2me:
This discussion has come up before and its perfectly legitimate and sensible request. But I have always believed that artwork types should be hard coded to provide consistency through different skins and Add-ons.

phil65:
They shouldnt be hardcoded, but we should have some conventions which ones are used commonly.

DaveTBlake:
Fair enough @zag2me I can see your point, but at the moment it all seems very arbitrary and inconsistent to me.  @phil65 a convention would be welcome.

I guess "cdart" is general for albums, or is it an image of the disk? In comparison "cover" (front), "back" and "spine" are clearly specific. Then there are also multiple images from any  liner booklet art, do we go that far?

For artist how are "banner" and "clearlogo" used, or compare to fanart?

phil65:
banner and clearlogo are the same as for movies. cdart is an image of the cd/dvd itself.

MartijnKaijser:
See https://fanart.tv/  how they look.
Videos database support the standard poster/fanart as hardcoded and with JSON-RPC you can additionally add whatever image names you want by simply defining them.

DaveTBlake:
Ah much clearer, thanks. I tend to listen to music with TV off so oblivious to all this glorious art!

rmrector:
`ParseThumbs` just above these can already pull in all freely named artwork for Python scrapers, by setting "aspect" to "clearlogo" or whatever. "aspect" is a bit of a misnomer but in the video library it is directly translated to "art type" when it is added (see #12612 for the missing artwork there), and `ParseThumbs` is shared between video and music, so extended artwork in the music library isn't being held back here.

ronie:
superseded by https://github.com/xbmc/xbmc/pull/13848

PR #11756, opened 2017-02-26
drop support for 'old' static listitem format

ronie:
there are two possible formats to define a static listitem, but i've never seen any skin using this format:
https://github.com/xbmc/xbmc/blame/a005e3f5e68d07f28d642bbc2e988ace72499a9e/guilib/GUIBaseContainer.cpp#L886-L889

the main purpose of this pr is to get rid of the requirement to define `<onclick>noop</onclick` for static listitems that have no action.
the presence of an  'onclick' tag is currently used to differentiate between the two formats.

ping @HitcherUK @BigNoid @phil65


HitcherUK:
I never knew that way even existed until now.

ronie:
jenkins build this please

PR #11761, opened 2017-02-26
PVR: never cache pvr directories

FernetMenta:
pvr items are not cached, hence the directory must be be cached either. currently calling Exists on pvr recordings always fails because the cached dir is empty.

@ksooo I am investigating an issue with scanning pvr recordings to tv shows. there is more broken.

### DIFF
--- a/xbmc/filesystem/PVRDirectory.h
+++ b/xbmc/filesystem/PVRDirectory.h
@@ -32,8 +32,10 @@ public:
   CPVRDirectory();
   virtual ~CPVRDirectory();
 
-  virtual bool GetDirectory(const CURL& url, CFileItemList &items);
-  virtual bool AllowAll() const { return true; }
+  virtual bool GetDirectory(const CURL& url, CFileItemList &items) override;
+  virtual bool AllowAll() const override { return true; }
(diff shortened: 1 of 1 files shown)


ksooo:
Never thought about this, but sounds reasonable.

metaron-uk:
@FernetMenta I've been doing this for a while. 
You have to disable 're-scan library on startup' and do periodic manual updates once PVR has settled (pvr.mythtv returns an empty recordings list to start with, then fills it in the background while the guide data loads).
Unless you have good season/episode data and a 'correct' show title provided by the backend for your recordings it won't work reliably. The UK Radio times data used to be good for this (what I was using originally). Schedules Direct data should also be OK.

metaron-uk:
@FernetMenta if you do this, I hope the code that generates the recordings directory gets optimized at the same time, otherwise browsing the recordings directory on RPi could become unusable. When you delete a recording and it has to re-build the list things can slow down rather a lot while it re-caches the directories

FernetMenta:
@metaron-uk pvr items are not cached, hence this pr should not change what you mentioned.

FernetMenta:
jenkins build this please

PR #11763, opened 2017-02-26
[depends] install pycryptodome under the alternative Cryptodome names…

wsnipex:
…pace instead of Crypto

Under linux the standard python-crypto module is usually already installed as Crypto and must not be overwritten by cryptodome.
Using the alternative namespace Cryptodome avoids clashes with the original module since both are not compatible with each other.

Addons using this module are advised to change their import statements to import Cryptodome

@peak3d  ping


wsnipex:
jenkins build this please

peak3d:
thanks, @wsnipex . 👍

PR #11771, opened 2017-02-27
[Fix]Skip trying to use TagLib to read tags from Shoutcasts

DaveTBlake:
Stop trying to get TabLib to read metadata from Shoutcasts, but not excluding other internet streams.
An alternative to #11759 

TabLib is not designed to cope with Shoutcasts, which while being an MPEG audio stream have a bespoke header and metadata format, and some streams cause TabLib to hang (or actually search indefinitely).

@fritsch @arnova @phate89 @wsnipex 

I realise that a future refactoring wants to make TagLibVFSStream more controlled, and protect from unlimited reading. This is not an attempt at that, just a modest change that could be backported, that will avoid hangups caused by passing tagLib streams that we know do not contain tags it can read.

### DIFF
--- a/xbmc/music/tags/TagLibVFSStream.cpp
+++ b/xbmc/music/tags/TagLibVFSStream.cpp
@@ -20,6 +20,8 @@
 #include "limits.h"
 #include "TagLibVFSStream.h"
 #include "filesystem/File.h"
+#include "filesystem/FileCache.h"
+#include "filesystem/ShoutcastFile.h"
 #include <taglib/tiostream.h>
 
 using namespace XFILE;
@@ -37,6 +39,17 @@ TagLibVFSStream::TagLibVFSStream(const std::string& strFileName, bool readOnly)
   {
     if (!m_file.Open(strFileName))
       m_bIsOpen = false;
+    else
+    { // Check if this stream is a Shoutcast, and skip scanning tags if it is.
+      // Shoutcast has its own metadata format that TagLib does not read
+      CFileCache* fileCacheItem = dynamic_cast<CFileCache*>(m_file.GetImplementation());
+      if (fileCacheItem)
+      {
+        CShoutcastFile* item = dynamic_cast<CShoutcastFile*>(fileCacheItem->GetFileImp());
+        if (item)
+          m_bIsOpen = false;
+      }
+    }
   }
   else
   {


wsnipex:
looks good to me

xbmc/music/tags/TagLibVFSStream.cpp
```diff
@@ -37,6 +39,17 @@ TagLibVFSStream::TagLibVFSStream(const std::string& strFileName, bool readOnly)
   {
     if (!m_file.Open(strFileName))
       m_bIsOpen = false;
+    else
+    { // Check if this stream is a Shoutcast, and skip scanning tags if it is.
+      // Shoutcast has its own metadata format that TagLib does not read
+      CFileCache* fileCacheItem = dynamic_cast<CFileCache*>(m_file.GetImplementation());
```
FernetMenta: this is not reliable because cache can be deactivated

DaveTBlake: How can I deactivate cache to test?
And do you have any other suggestions how to burrow reliably into the IFile sequence of items to check for the Shoutcast? I am at the limit of my knowledge on this

FernetMenta: see http://kodi.wiki/view/advancedsettings.xml#cache

maybe this could work:
try cast to filecache, if that works cast this into shoutcast, else cast original into shoutcast

fritsch:
If you wanted a timeout instead: https://github.com/fritsch/xbmc/commit/a847221de022e0cc1df59cda90814793170b7522

xbmc/music/tags/TagLibVFSStream.cpp
DaveTBlake: Deactivating cache I still get the first file in the sequence as a CFileCache.
@Paxxi suggested a looping variant with IFile, but the inheritance chain is broken so I can't get that to work either.

DaveTBlake:
Thanks for the timeout approach @fritsch, a solution to the general problem of TagLibVFSStream running unlimited, but I would want do to lot more testing before applying an arbitrary time limit on all tag scanning. Here I'm looking for a fix to backport at this late stage in the cycle.

Now I believe that #11759 is a valid backport solution - even if it scans tag data from an internet stream Kodi does not display it, so no one will scream about lost features.

It is future work to add control of TagLibVFSStream, *and* if we want to implement the display of metadata from true mp3 internet streams that entails implementing accurate filtering that skips the Shoutcasts before calling TagLib.

PR #11772, opened 2017-02-27
[cmake/windowing] add initial rockchip support

mac-l1:
## Description

This new feature adds basic initial support for rockchip devices. Initial support means that kodi is built with minimal changes and is runnable on linux distributions publicly supported by rockchip - just kodi as-is. The commit consists of following changes:
- [cmake] add an enable rockchip switch
- [cmake] this switch excludes VAAPI and VDPAU as these are based on OpenGL that is not supported on rockchip devices
- [cmake] find ROCKCHIP include and library directories and libraries; furthermore use compiler optimization option O2 over O3 as arm gcc assembler errors out
- [cmake] for ROCKCHIP don't use OpenGL and use OpenGLES, XRandR, X11 and LibDRM
- [windowing/egl] for ROCKCHIP re-use Amlogic native egl handling as these are similar

## Motivation and Context
The new feature is needed to establish an initial version that runs on rockchip's open source linux distributions on their platforms as they are recently opening up to the linux open source community (they recently open sourced their hardware video acceleration libraries and their hardware accelerated 2D graphics library). Hence kodi cannot be missed anymore in these linux distributions and on this hardware.

This change is implemented so that it minimizes impact on other software parts and minimizes the chances to break any non-rockchip-related build. The same way it also establishes a sound base to do develop further features like hardware video decoding and related zero-copy video rendering.

## How Has This Been Tested?

The feature (just a running kodi, as-is) has been tested on latest debian linux distribution (stretch) and kernel (4.4) that is open sourced and provided by rockchip. It has been tested on firefly and miqi rk3288 devices.
Tests consisted of basic kodi browsing the menus, playing h264-videos and some youtube videos using the add-on.

## Types of change

## Checklist:


CMakeLists.txt
```diff
@@ -192,8 +198,12 @@ else()
...
 
 if(NOT WIN32)
-  core_optional_dep(OpenGl)
-  if(OPENGL_FOUND AND NOT APPLE)
+  if(NOT ROCKCHIP_FOUND)
+    core_optional_dep(OpenGl)
+  else()
+    core_optional_dep(XRandR ENABLE_X11)
+  endif()
+  if(OPENGL_FOUND AND NOT APPLE AND NOT ROCKCHIP_FOUND)
```
FernetMenta: I don't think the already existing nonsense should be extended here. even a blind man can see that this leads to nowhere

hudokkow: Well, that was constructive...

lrusak:
Thanks for the PR @mac-l1 however I don't think this implementation is quite right. A couple things:

1. You are basically piggybacking on amlogic. In that case why not just extend the amlogic codebase?
2. We want to move away from vendor specific code. This is the biggest thing. We want to move to a common codebase (as much as possible) instead of diverging from one. I don't know much about rockchip hardware (I don't own any) but I hope they are aiming for mainline kernel support and more (v4l2, cec, etc).

So in my opinion I don't think we can include this PR, however, it would be nice if others could chime in about what they think.

CMakeLists.txt
mac-l1: not that constructive but you are right: it was a mess and now is even more a mess. i included and excluded the libs for rockchip so that i was sure it didnt break any possible other system (which i dont know and cant test) in line with the already existing spaghetti flow. so it just works without interfering with other any other possible display system builds.

my proposal would be to add a dedicated rockchip block just after "if(NOT WIN32)" that clearly and only includes all necessary libs for rockchip and doesnt enter the spaghetti flow further.

something like this:
```
if(NOT WIN32)
  if(ROCKCHIP_FOUND)
    core_require_dep(EGL)
    core_optional_dep(OpenGLES)
    core_optional_dep(X ENABLE_X11)
    core_optional_dep(XRandR ENABLE_X11)
    core_optional_dep(LibDRM ENABLE_X11)
  else()
    ... original code
```

would that be acceptable?

FernetMenta: This entire thing needs a redesign that makes it clear what modules are required for what platforms. X11, XrandR for example are no optional modules. Those are requirements for X11, that is a platform by itself. OpenGLES is not optional for Rockchip, it is a requirement. If you wnat to build for X11, Rockchip, or other platforms, build needs to fail if required modules are not found. Going an alternative branch makes no sense.

mac-l1: i agree: in general it needs a redesign to get out of this mess, enabling clear and straightforward platform based module selection. it is exactly what i bumped into while my goal primary is to give working hardware acceleration and rendering for current and coming rockchip devices back to the kodi and rockchip community.

is this redesign already scheduled and if so when and by who? i would be happy to help but platform display modules are not my expertise; rockchip hardware decoding and rendering are.

mac-l1:
@FernetMenta: in order to proceed i reworked/simplified the change as not to add more complexity by adding alternative flows or so. could you review this?

in essence:
- i disabled opengl for rockchip based on th enable rockchip switch
- i added the XrandR module just after the X module in the OpenGLES case as it was missing (defect?)

xbmc/windowing/X11/GLContextEGL.h
```diff
@@ -23,6 +23,7 @@
 #if defined(HAVE_X11)
 #include "GLContext.h"
 #include "EGL/egl.h"
+#include <X11/Xutil.h>
```
FernetMenta: this is not related to rockchip, please drop

CMakeLists.txt
```diff
@@ -210,6 +217,7 @@ if(NOT WIN32)
         core_optional_dep(LibDRM ENABLE_MIR)
       else()
         core_optional_dep(X ENABLE_X11)
+        core_optional_dep(XRandR ENABLE_X11)
```
FernetMenta: how is this related to rockchip?

Hedda:
@mac-l1 

Why can't Rockchip (or the Rockchip community) add a support library for VAAPI or VDPAU APIs instead?

Supporting VAAPI or VDPAU API means support in all apps, not just Kodi, without changes needed to them.

See link below for a proof-of-concept how it was done like that for Allwinner:

http://forum.kodi.tv/showthread.php?tid=254202

Maybe this could be dicussed in a new thread Kodi's development forum?

http://forum.kodi.tv/forumdisplay.php?fid=93

koenkooi:
@Hedda v4l2 would be a better fit for most SoC architectures.

xbmc/windowing/X11/GLContextEGL.h
mac-l1: if <X11/Xutil.h> is not included then i get the following build error when compiling:
`In file included from /home/linaro/kodi/xbmc/windowing/X11/GLContextEGL.cpp:37:0:
/home/linaro/kodi/xbmc/windowing/X11/GLContextEGL.h:39:3: error: ‘XVisualInfo’ does not name a type` and subsequent errors

this is related to rockchip as they specifically use ARMs mali GPU that implements OpenGLES (midgard). the corresponding header file regarding egl <EGL/eglplatform.h> of ARM differs from those of other GPU vendors, and also from the mostly used mesa compatible implementations.

in this case the <EGL/egplatform.h> from the usual x86* unix environments implicitly include the mesa compliant implementation of <EGL/platform.h>, and this one automatically includes <X11/Xutil.h> as it assumes X11 automatically for unix. see [https://cgit.freedesktop.org/mesa/mesa/tree/include/EGL/eglplatform.h#n120](url)

ARMs <EGL/egplatform.h> does not automatically include <X11/Xutil.h>. See [https://github.com/rockchip-linux/libmali/blob/rockchip/include/EGL/eglplatform.h](url)

kodi is probably not much used for the platform combination of linux and X11 and ARM GPU OpenGLES.
however kodi probably is very familiar with the combination of android, surfaceflinger and ARMs GPU OpenGLES but then this X11/Xutil.h dependency doesnt exist.

so for rockchip the X11/Xutil.h dependency needs to be specified. so please merge.

FernetMenta: I don't understand why you add any files from X11 to this platform.

CMakeLists.txt
mac-l1: not directly but more by the way the platform display module selection flow is implemented (the messy one).

if i dont add the XRandR module then i get the error:
`build/windowing/windowing.a(WinEventsX11.cpp.o):WinEventsX11.cpp:function CWinEventsX11Imp::MessagePump(): error: undefined reference to 'XRRUpdateConfiguration'
` and two more XR* undefined references due to missing 

rockchip doesnt support opengl but only opengles and this is probably not usual. i guess most kodi platforms that have opengles also have opengl (e.g. most mesa based implementations do) and then according the selection flow XRandR is automatically added (about 10 lines before, line #209) or the old ./configure way is used to specify this explicitly when building (i used to do this before using cmake)

it seems strange to add the X module and not the XRandR module. might actually be best to merge the FindX.cmake and FindXRandR.cmake into one.

so please merge.

FernetMenta: I certainly won't merge this in this state.

1)
X11, AML, IMX are platforms, means they define their own windowing system. I don't see what you intend Rockchip to be. Without supporting any of the hw accelerators APIs we have avaialbe for X11, it makes zero sense running a kodi build for X11 on that processor because the CPU is not powerful enough.

2)
If rockchip intends to be a platform like AML, it has to provide its own implementeations for the egl windowing system. Assuming that the port for AML will always do is wrong.

mac-l1: you have a point asking wheter rockchip is a platform or not. i was wondering too.

1. X11 platform
if the definition is that a platform defines its own windowing system than rockchip is not a platform here. this commits makes kodi run on latest linaro debian 9 distribution using ordinary X11 API and OpenGLES compliant API. the X server runs with glamor x11 display driver (hw accelerated OpenGLES rendering). this commit also uses the HW accelerator api of OpenGLES in X11 that you provide because it makes sense and just works.
the only essential issue is that instead of OpenGL only OpenGLES needs to be used and that seems something that current kodi builds arent familiar with in combination with X11 and XRandR. it seems strange that nobody experienced this before.

2. EGL WINDOWING PLATFORM
on the other hand, regarding the egl windowing system i now reuse AML logic for resolution selection just to avoid to implement an additional, similar one - its simply scanning standard linux framebuffers to get/set its modes. however the AML one offers much more that is simply not used for rockchip so it's a bit too much but its just there and already implemented.

3. EGL DRM WINDOWING
it would probably be best to implement a new variant here: however not dedicated for rockchip but for DRM in general as that would be the generic way for linux (could be even independent of X). in the end that would be best i guess. for playing videos it could offer a clean, generic way to get the frame from the hw decoder rendering it directly to drm planes / display without the need of any processing through GPU in between, all linux generic. by bypassing the GPU you would save memory bandwidth that is needed for 4K / 8K frames at high frame rates. however would be quite some effort.

anyway, right now i dont see a way to proceed so i guess for now we are stuck.

xbmc/windowing/X11/GLContextEGL.h
mac-l1: that is because kodi for rockchip linux is just an X11 application that uses OpenGLES to render it graphics into its window. so it isnt really a platform actually.

however it seems that with the graphical GPU driver of the rockchip SoC that implements OpenGLES (ARM mali GPU) has a different egl header file than usual for kodi (probably mesa based). 

so the include files in mali egl package (/usr/lib/arm-linux-gnueabihf/pkgconfig/egl.pc) dont include any <X11/Xutil.h> inside. but for instance the include files in the mesa egl package on ordinary pc do (/lib/x86_64-linux-gnu/pkgconfig/egl.pc) .

this implicit inclusion of XUtil.h in <egl/platform.h> is used in GLContextEGL.cpp to define the XVisualInfo type that is used in its X11 calls. this doesnt work for mali egl package and hence Xutil.h needs to be included explicitly.

hope things are a bit clearer now?

CMakeLists.txt
FernetMenta: @mac-l1 thanks much, things are getting clearer now.

1. Could you please separate X11 related changes into a new PR? I tihnk those are ok and can go in.

2. Are you saying there is nother AML specific in EGLNativeTypeAmlogic? If this is the case we maybe should rename it to something more generic.

mac-l1: @FernetMenta things are indeed getting more clear now, thanks, learning everyday. i now got it working by completely removing the AML dependency so only the X11 platform stuff is used without using egl native type EGLNativeTypeAmlogic anymore.

mac-l1:
@FernetMenta just commited the removal of all AML / EGLNativeType dependencies. to be clear how to proceed best:

1) is this enough? or

2) should i make another PR like you mentioned: one rockchip independent PR for X11/XRandR changes and then proceed with current PR with only rockchip build stuff like excluding OpenGL, VDPAU and VAAPI and so? or

3) close current one and make two new other PRs with the same result as 2)

can you please advise?

CMakeLists.txt
FernetMenta: for X11 builds the module FindROCKCHIP should not be needed. X11 is either build with OpenGL or GLES but not for a specific hardware vendor.

FernetMenta:
@mac-l1 please go for 2) and open a new PR for X11 with GLES fixes. There is no need to exclude VAAPI or VDPAU, those are runtime detected. (that will be reflected in GUI settings soon)

mac-l1:
This pull request can be closed as it is obsolete afte the merge of pull [https://github.com/xbmc/xbmc/pull/11805](url)

Building kodi for rockchip linux devices can be done by:
`cmake .. -DCMAKE_INSTALL_PREFIX=/usr/local -DENABLE_OPENGL=OFF -DENABLE_VDPAU=OFF -DENABLE_VAAPI=OFF -DCMAKE_BUILD_TYPE=RelWithDebInfo '-DCMAKE_CXX_FLAGS:STRING=-DGL_GLEXT_PROTOTYPES -DEGL_EGLEXT_PROTOTYPES -fpermissive'`

PR #11780, opened 2017-02-28
[xbmc][win32] Move to using unicode win32 api methods

Paxxi:
This should have been done a long time ago but
with UWP no longer supporting the ascii versions it's
time to get it done.

The changes in UnrarXLib isn't really tested but it's
about to get dropped from our tree anyway.

## Types of change

## Checklist:


Paxxi:
jenkins build this please

maybe @Rechi for review?

cmake/platform/windows/defines.txt
```diff
@@ -1 +1 @@
--DTARGET_WINDOWS -DNOMINMAX -D_CRT_SECURE_NO_WARNINGS -D_WINSOCKAPI_
\ No newline at end of file
+/DTARGET_WINDOWS /DNOMINMAX /D_CRT_SECURE_NO_WARNINGS /D_WINSOCKAPI_
```
Rechi: don't think this is needed

lib/UnrarXLib/suballoc.cpp
```diff
@@ -92,7 +92,7 @@ bool SubAllocator::StartSubAllocator(int SASize)
   }
 
   if(AllocSize != AllocSize2)
-    OutputDebugString("ERROR - had to allocate smaller data than required, extract can very well fail");
+    OutputDebugString(L"ERROR - had to allocate smaller data than required, extract can very well fail");
```
Rechi: with this change every target expect windows fails

lib/UnrarXLib/win32acl.cpp
```diff
@@ -7,9 +7,8 @@ static bool ReadSacl=false;
 #ifndef SFX_MODULE
 void ExtractACL(Archive &Arc,char *FileName,wchar *FileNameW)
 {
-#if defined(_XBOX) || defined (_LINUX)
   return;
```
Rechi: function returns now on every platform without any code executed
If this is relay wanted remove code below.

xbmc/platform/win32/CharsetConverter.cpp
```diff
@@ -18,6 +18,7 @@
  *
  */
 
+#if defined(TARGET_WINDOWS) || defined(TARGET_WIN10)
```
Rechi: shouldn't this be handled by the build system?

xbmc/platform/win32/CharsetConverter.h
```diff
@@ -22,6 +22,7 @@
 
 #include <string>
 
+#if defined(TARGET_WINDOWS) || defined(TARGET_WIN10)
```
Rechi: shouldn't this be handled by the build system?

Rechi (commented):
haven't tested this yet, only looked at the code changes

lib/libUPnP/Neptune/Source/System/Win32/NptWin32SerialPort.cpp
```diff
@@ -1,337 +1,334 @@
...
+    return NPT_SUCCESS;
+}
+
+/*----------------------------------------------------------------------
+|   NPT_SerialPort::NPT_SerialPort
++---------------------------------------------------------------------*/
+NPT_SerialPort::NPT_SerialPort(const char* name)
+{
+    m_Delegate = new NPT_Win32SerialPort(name);
+}
```
Montellese: Something went wrong here. The whole file is shown as changed but you probably only changed a few lines.

Montellese (changes requested):
Two general observations:
* What is `TARGET_WIN10`? You introduce it in some places but by far not in all.
* Please move the changes to libUPnP into a separate commit and add another commit with the patch of the separate commit to the libUPnP patch set.

xbmc/platform/win32/CharsetConverter.cpp
Paxxi: I want to avoid having ifdefs in every file that includes this file so I chose to stick them in here to make calling code a bit cleaner

xbmc/platform/win32/CharsetConverter.h
Paxxi: I want to avoid having ifdefs in every file that includes this file so I chose to stick them in here to make calling code a bit cleaner

lib/libUPnP/Neptune/Source/System/Win32/NptWin32SerialPort.cpp
Paxxi: had some issues with gits autocrlf, thought I caught all of it, redoing the neptune stuff now with the patch file.

Paxxi:
the introduction of `TARGET_WIN10` here was poorly documented(not at all).
This is the define that will be used for UWP builds, the reason that it's not everywhere is that I've added it only in places I'm quite certain will run on UWP as well.
Discussion about it has been mostly in Slack afaik so easy to miss.

I could as well drop them here as it will be added in a later PR once UWP is nearer anything usable, just thought I'd sneak a few in to make it easier for the people working on the port.

cmake/platform/windows/defines.txt
Paxxi: reverted this change

Paxxi:
jenkins build this please

Rechi:
jenkins build this please

Paxxi:
the build failures was unrelated, all platforms built fine except rpi which failed to find the git hash

Paxxi:
@montellese good to go?

Montellese:
libUPnP looks fine now thanks.

Montellese (approved):

PR #11782, opened 2017-02-28
[PVR] Guide window: Feature: 'Smart selection' of epg events."

ksooo:
This is a nice little feature that has been requested a couple of times - "smart selection" of events in the Guide window.

This feature can be activated here: 
screenshot000

How does it work? The action taken when selecting an epg event is nort fixed, but dependent on the current date and time.

Selecting an event
* with end time in the past, select will stat playing the recording of the the selected event if there is one, or it will bring up the event  information dialog.
* which is currently active will switch to the respective channel.
* with start time in the future will either program a recording timer for the event or show timer settings dialog if there already is a timer set for this event.

This PR also changes the default setting value to "smart selection" (as suggested by @da-anda).

@Jalle19 for code review.

### DIFF
--- a/xbmc/pvr/windows/GUIWindowPVRGuide.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRGuide.cpp
@@ -280,6 +280,40 @@ bool CGUIWindowPVRGuide::OnMessage(CGUIMessage& message)
                   CPVRGUIActions::GetInstance().ToggleTimer(pItem);
                   bReturn = true;
                   break;
+                case EPG_SELECT_ACTION_SMART_SELECT:
+                {
+                  const CEpgInfoTagPtr tag(pItem->GetEPGInfoTag());
+                  if (tag)
+                  {
+                    const CDateTime start(tag->StartAsUTC());
+                    const CDateTime end(tag->EndAsUTC());
+                    const CDateTime now(CDateTime::GetUTCDateTime());
+
+                    if (start <= now && now <= end)
+                    {
+                      // current event
+                      CPVRGUIActions::GetInstance().SwitchToChannel(pItem, true);
+                    }
+                    else if (now < start)
+                    {
+                      // future event
+                      if (tag->HasTimer())
+                        CPVRGUIActions::GetInstance().EditTimer(pItem);
+                      else
+                        CPVRGUIActions::GetInstance().AddTimer(pItem, false);
+                    }
+                    else
+                    {
+                      // past event
+                      if (tag->HasRecording())
+                        CPVRGUIActions::GetInstance().PlayRecording(pItem, true);
+                      else
+                        CPVRGUIActions::GetInstance().ShowEPGInfo(pItem);
+                    }
+                    bReturn = true;
+                  }
+                  break;
+                }
               }
               break;
             case ACTION_SHOW_INFO:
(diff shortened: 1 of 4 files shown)


ksooo:
@da-anda fyi

xbmc/pvr/windows/GUIWindowPVRGuide.cpp
```diff
@@ -280,6 +280,37 @@ bool CGUIWindowPVRGuide::OnMessage(CGUIMessage& message)
...
+
+                    if (start <= now && now <= end)
+                    {
+                      // current event
+                      CPVRGUIActions::GetInstance().SwitchToChannel(pItem, true);
+                    }
+                    else if (now < start)
+                    {
+                      // future event
+                      CPVRGUIActions::GetInstance().ToggleTimer(pItem);
```
Jalle19: I'm not sure if it's intuitive to delete a timer by pressing OK. Maybe instead we could add a timer if one doesn't exist and open the info dialog if one does exist?

ksooo: Good point.

And, would you ask the user whether he wants to add a timer or just do it? Currently there is no prompt.

Jalle19: Just add it, after all the mode is called "smart" select :-)

da-anda:
nice addition, thanks. We should probably make this the new default if there are no objections from others.

Btw, I think we could probably get rid of the "context menu" option, now that we have longpress to trigger it. I think that any limited button remote these days can trigger a long press, and regular IR remotes usually have enough buttons for a dedicated context menu key. Not related to this change, it just came to my mind.

ksooo:
CEC remotes do not support long press (afaik) and many VEG remotes have no "menu" key, thus we should not remove "context menu"

I'm fine with changing the default to "smart select".

da-anda:
well, those remotes will have the same issue with the context menu anywhere in Kodi, but that's not related to this change. Let's discuss it on the forum or on Slack

ksooo:
@da-anda, @Jalle19 the changes you suggested are now in.

ksooo:
jenkins build this please

ksooo:
Jenkins WIN-32 error is unrelated.

<pre>
library.mak:99: recipe for target 'libavcodec/avcodec-57.dll' failed
make: *** [libavcodec/avcodec-57.dll] Error 1
</pre>

xhaggi:
Sorry for the late response, but changing the default to that option will confuse most of the user IMO. All the stuff which is now hidden behind "smart select" can be triggered in the info dialog. Not really intuitive if you ask me. I am not against this feature, only against changing the default.

ksooo:
> Sorry for the late response, but changing the default to that option will confuse most of the user IMO

 There will always be people crying if some behavior the are used to changes.

 I share @da-anda's pow:

> Reason: Kodi tries to be smart wherever it can to improve user experience, while giving pros the ability to modify behavior.

 Info dialog is two clicks where smart select ideally is just one click.

 OTOH, I will not fight for the new default. This is not worth the time. At the end it's just a setting. If you feel the new default is wrong, get an agreement with @da-anda and change it.

xhaggi:
From a user perspective the best default is the "show info" option, because everything else is clear, no matter if you need to click twice on your remote or keyboard. That's the reason we should not change to something which needs a bit more explanation to understand.

Next time we should ask more than just @da-anda if we want to decide to change a setting default.
Every change has a huge impact to our support team etc. You know what's changed, a normal user doesn't.

I'll create a new PR to revert the default.

PR #11000, opened 2016-11-27
[Estury] Add Season/Episode to PVR Recordings

metaron-uk:
Adds Season.Episode to PVR recordings in Estuary Skin

## Description
If a PVR client provides Season/Episode information, display it in the PVR recordings list.
Format = Standard kodi Season/Episode format as used in Videos section: 
- 1.2 = Season 1, Episode 2
- S1 = Special 1

## Motivation and Context
If you record a whole series, you may miss one, or they may be recorded out of order. This allows you to see which episode comes next. 
Couple this with sort by 'File' (https://github.com/xbmc/xbmc/pull/10999) and you can sort in season/episode order.

## How Has This Been Tested?
Used for several months (>12) on a confluence based skin with uk-rt, then uk-atlas and now schedules-direct as a listings source.
Tested in this form using pvr.mythtv with a 0.27 backend on x86 linux

## Screenshots (if appropriate):
screenshot053
screenshot051
screenshot055

## Types of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

@ksooo @phil65 something for L??? I guess as I seem to have missed the boat for Krypton

PR #10999, opened 2016-11-27
[PVR] Reintroduce filename sort for PVR recordings

metaron-uk:
Re-introduce filename sorting for PVR recordings which was removed by https://github.com/xbmc/xbmc/pull/10907 as requested by @ksooo 

## Description
Allows recordings to be sorted by filename again - which is by season/episode when that information is provided by the PVR client.

## How Has This Been Tested?
Tested on x86 using default estuary skin and modified version of confluence.

## Types of change
- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #11783, opened 2017-02-28
[PVR] Standard list items for EPG, timers & recordings [WIP]

metaron-uk:
Adds Title (Subtitle)  SeasonxEpisode to PVR EPG entries, timers and recordings
(As per suggestion in https://github.com/xbmc/xbmc/pull/11000 )

## Description
If a PVR client provides EpisodeName, Season and/or Episode information, display this in lists of EPG entries, timers and recordings:

Format = Standard kodi Season/Episode format as used in Videos section and Estuary skin: 
Title (EpisodeName) 1.2 = Season 1, Episode 2
Title (EpisodeName) S1 = Special 1
Title (EpisodeName) = no Season or Episode, but an EpisodeName
Title = no Season, Episode or EpisodeName

## Motivation and Context
If you record a whole series, you may miss one, or they may be recorded out of order. This allows you to see which episodes you missed. 
Couple this with sort by 'File' (https://github.com/xbmc/xbmc/pull/10999) and you can sort in season/episode order.

## How Has This Been Tested?
Tested in this form using pvr.mythtv with a 0.27 backend on x86 linux using Estury and Confluence

As per https://github.com/xbmc/xbmc/pull/11000 but with x instead of .

## Types of change

I would say it is non-breaking but skins like confluence and Estuary which use <label>$INFO[ListItem.Label]$INFO[ListItem.EpisodeName, (,)]</label> will need to change this to
<label>$INFO[ListItem.Label]</label> to avoid getting duplicate 'EpisodeName' entries

Several other non-core skins will just work :)

Let me know if I should include Estuary changes in a separate commit (to be done later) or maybe I should re-purpose https://github.com/xbmc/xbmc/pull/11000 to accompany it. (Failing that, feel free to have a go yourselves!)

@phil65 @ronie - hope you like it
@ksooo - OK, what's wrong with my coding style ;-)

### DIFF
--- a/xbmc/FileItem.cpp
+++ b/xbmc/FileItem.cpp
@@ -133,7 +133,34 @@ CFileItem::CFileItem(const CEpgInfoTagPtr& tag)
   m_bIsFolder = false;
   m_epgInfoTag = tag;
   m_strPath = tag->Path();
-  SetLabel(tag->Title());
+
+  std::string strLabel("");
+  std::string strTitleEpisodeName("");
+  //Title and Episode
+  if (tag->EpisodeName().empty())
+    strTitleEpisodeName = tag->Title();
+  else
+    strTitleEpisodeName = StringUtils::Format("%s (%s)", tag->Title().c_str(), tag->EpisodeName().c_str());
+  //Season + Episode
+  int iSeason(-1);
+  int iEpisode(-1);
+  if (tag->SeriesNumber() > 0)
+    iSeason = tag->SeriesNumber();
+  if (tag->EpisodeNumber() > 0)
+    iEpisode = tag->EpisodeNumber();
+  if (iEpisode >= 0)
+  {
+    if (iSeason == 0) // prefix episode with 'S'
+      strLabel = StringUtils::Format("%s S%d", strTitleEpisodeName.c_str(), iEpisode);
+    else
+      strLabel = StringUtils::Format("%s %dx%d", strTitleEpisodeName.c_str(), iSeason, iEpisode);
+  }
+  else
+  {
+    strLabel = StringUtils::Format("%s", strTitleEpisodeName.c_str());
+  }
+  SetLabel(strLabel);
+
   m_strLabel2 = tag->Plot();
   m_dateTime = tag->StartAsLocalTime();
 
@@ -198,7 +225,34 @@ CFileItem::CFileItem(const CPVRRecordingPtr& record)
   m_bIsFolder = false;
   m_pvrRecordingInfoTag = record;
   m_strPath = record->m_strFileNameAndPath;
-  SetLabel(record->m_strTitle);
+
+  std::string strLabel("");
+  std::string strTitleEpisodeName("");
+  //Title and Episode
+  if (record->EpisodeName().empty())
+    strTitleEpisodeName = record->m_strTitle;
+  else
+    strTitleEpisodeName = StringUtils::Format("%s (%s)", record->m_strTitle.c_str(), record->EpisodeName().c_str());
+  //Season + Episode
+  int iSeason(-1);
+  int iEpisode(-1);
+  if (record->m_iSeason > 0)
+    iSeason = record->m_iSeason;
+  if (record->m_iEpisode > 0)
+    iEpisode = record->m_iEpisode;
+  if (iEpisode >= 0)
+  {
+    if (iSeason == 0) // prefix episode with 'S'
+      strLabel = StringUtils::Format("%s S%d", strTitleEpisodeName.c_str(), iEpisode);
+    else
+      strLabel = StringUtils::Format("%s %dx%d", strTitleEpisodeName.c_str(), iSeason, iEpisode);
+  }
+  else
+  {
+    strLabel = StringUtils::Format("%s", strTitleEpisodeName.c_str());
+  }
+  SetLabel(strLabel);
+
   m_strLabel2 = record->m_strPlot;
   FillInMimeType(false);
 }
@@ -212,7 +266,36 @@ CFileItem::CFileItem(const CPVRTimerInfoTagPtr& timer)
   m_bIsFolder = timer->IsTimerRule();
   m_pvrTimerInfoTag = timer;
   m_strPath = timer->Path();
-  SetLabel(timer->Title());
+
+  std::string strLabel("");
+  const CEpgInfoTagPtr tag(timer->GetEpgInfoTag());
+  if (tag)
+  {
+    //Season + Episode
+    int iSeason(-1);
+		int iEpisode(-1);
+    if (tag->SeriesNumber() > 0)
+      iSeason = tag->SeriesNumber();
+    if (tag->EpisodeNumber() > 0)
+      iEpisode = tag->EpisodeNumber();
+    if (iEpisode >= 0)
+    {
+      if (iSeason == 0) // prefix episode with 'S'
+        strLabel = StringUtils::Format("%s S%d", timer->Title().c_str(), iEpisode);
+      else
+        strLabel = StringUtils::Format("%s %dx%d", timer->Title().c_str(), iSeason, iEpisode);
+    }
+    else
+    {
+      strLabel = timer->Title();
+    }
+  }
+  else
+  {
+    strLabel = timer->Title();
+  }
+  SetLabel(strLabel);
+
   m_strLabel2 = timer->Summary();
   m_dateTime = timer->StartAsLocalTime();
 
--- a/xbmc/FileItem.cpp
+++ b/xbmc/FileItem.cpp
@@ -133,33 +133,7 @@ CFileItem::CFileItem(const CEpgInfoTagPtr& tag)
   m_bIsFolder = false;
   m_epgInfoTag = tag;
   m_strPath = tag->Path();
-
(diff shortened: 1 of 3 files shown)


FernetMenta:
FileItem is a generic class and should not be aware of all this pvr specific things. Sure, FileItem is already a mess but that does not mean that we should continue making it worse.

ksooo:
Thanks for your contribution.

I generally like the idea of consistency across all typos if pvr items, but I think you took the wrong implementation approach. CFileItem should stay what it is, a dumb data container. Formatting data for UI (skin) usage should imo always happen in CGUIInfoManager.

With your approach, a skinned does not have the chance to customize the labels for pvr list items to the extend he can do today. Not everyone may like the formatting you have chosen.

metaron-uk:
> @FernetMenta FileItem is a generic class and should not be aware of all this pvr specific things. Sure, FileItem is already a mess but that does not mean that we should continue making it worse.

> @ksooo CFileItem should stay what it is, a dumb data container. Formatting data for UI (skin) usage should imo always happen in CGUIInfoManager.

Aaah a software architecture question. I agree this is a mess. It took me ages to find all this last time I looked for it and just as long to find it this time.
The place to look seems to be https://github.com/xbmc/xbmc/blob/master/xbmc/GUIInfoManager.cpp#L9603
Would you propose that I strip out the formatting and the call to item->SetLabel() from the FileItem section and put it here instead?
Although this file is already *massive* at least all the LISTITEM management would then be in one place.

PS: I noticed that Label2 is set elsewhere, during sorting / filtering and this over-writes whatever it gets set to in FileItem before it is used anyway. I was going to submit a PR to delete it a year or so ago, but it didn't seem worth the trouble.

> @ksooo With your approach, a skinned does not have the chance to customize the labels for pvr list items to the extend he can do today.

That's not actually true. ListItem.Label doesn't have to be used. You could construct an alternative using ListItem.Title:
<code>$INFO[ListItem.Label] </code> == <code>$INFO[ListItem.Title,, ]$INFO[ListItem.EpisodeName,(,) ]$INFO[Listitem.Season,,x]$INFO[Listitem.Episode]</code>

Or if you want something shorter without the season numbering:
<code>$INFO[ListItem.Title,, ]$INFO[ListItem.EpisodeName,(,) ]</code>

Estuary already does this in several places, as you can see from these screenshots taken with the current version of this PR included:
screenshot067
Here ListItem.Label is used by the main timeline, but ListItem.Title (white) and ListItem.EpisodeName (grey) is used in the detailed display below (I don't get season/episode information in my guide data - I'm using an old mythtv backend)

screenshot068
Here is an example of my recordings showing the existing <code>$INFO[ListItem.Label,, ]$INFO[ListItem.EpisodeName,(,) ]</code> in the list and <code>ListItem.Title</code> <code> ListItem.EpisodeName </code> combination in the sidebar detail.

I agree that not everyone will like the 'defaults'. Personally I prefer <season>.<episode> rather than the <episode>x<season> @ksooo suggested based on what was used in the Video section on confluence. I haven't checked if it was skin driven or hard coded yet...
In any case, the point is this - they are just defaults. If a skinner wants something else, they can code it. It just means they get sane defaults to start from.

My personal preference is for the style of numbering S1 (Special1) or 1.2 (Season1 Episode2) rather than 5x1 (Season1 Episode5), but honestly having a standard season/episode display format is more important than what it is imho. If the 5x1 format is already in common usage elsewhere, we'll go with that.

Post your opinions here and I'll code something up. Defaults can always be changed....

xbmc/FileItem.cpp
```diff
@@ -133,7 +133,34 @@ CFileItem::CFileItem(const CEpgInfoTagPtr& tag)
...
+  if (tag->SeriesNumber() > 0)
+    iSeason = tag->SeriesNumber();
+  if (tag->EpisodeNumber() > 0)
+    iEpisode = tag->EpisodeNumber();
+  if (iEpisode >= 0)
+  {
+    if (iSeason == 0) // prefix episode with 'S'
+      strLabel = StringUtils::Format("%s S%d", strTitleEpisodeName.c_str(), iEpisode);
+    else
+      strLabel = StringUtils::Format("%s %dx%d", strTitleEpisodeName.c_str(), iSeason, iEpisode);
```
metaron-uk: This should actually be StringUtils::Format("%s %dx%d", strTitleEpisodeName.c_str(),  iEpisode, iSeason); (swap season/episode) to match the format proposed by @ksooo - same issues applies to the other places

metaron-uk: I've done some more digging and this format in confluence at least is actually set by the skin (ViewsVideoLibrary.xml) <code>\<label fallback="416"\>$INFO[listitem.Season,,x]$INFO[listitem.Episode]\</label\></code>

metaron-uk:
OK, here's what I've found out.
1) GUIInfoManager calls item->GetLabel (https://github.com/xbmc/xbmc/blob/master/xbmc/GUIInfoManager.cpp#L9604) which doesn't exist in the CFileItem class defined in FileItem.h?? - Is this inherited from somewhere??
2) Each type of FileItem (e.g. https://github.com/xbmc/xbmc/pull/11783/files#diff-6db2f8726d6229b9edcb6ceef3172652R127) calls SetLabel to set its label to whatever (previously 'Title' now my extended version during initialization (?).
3) CFileItem::SetLabel (https://github.com/xbmc/xbmc/blob/master/xbmc/FileItem.cpp#L1391) does some special case '..' list sorting stuff then sets the item label using CGUIListItem::SetLabel
4) CGUIListItem::SetLabel (https://github.com/xbmc/xbmc/blob/master/xbmc/guilib/GUIListItem.cpp#L69) sets the label and invalidates the list and sort order if the label has changed

Currently the actual setting of the label to a text string happens in 2) FileItem (e.g. https://github.com/xbmc/xbmc/pull/11783/files#diff-6db2f8726d6229b9edcb6ceef3172652R127

If we bypass SetLabel and GetLabel by directly referencing the underlying data in GUIInfoManager the '..' and list sorting stuff in 3) won't be consistent with what is presented in the list. This stuff undoubtedly belongs in a place that does file item list sorting (FileItem?) , but having tried to understand part of the list sorting and '..' mechanism before, it is a complex subject.

Conclusion:
* If we move the setting of Label to anywhere other than the FileItem initialisation routine without understanding why it was put there, we are likely to break list sorting.

@ksooo (2yrs) / @Montellese (5yrs) / @opdenkamp (5yrs), you were the last people in CFileItem::SetLabel and the FileItem initialization routines. Any comments?

ksooo:
Sorry, this currently has low prio for me. Not enough time to think about the changes and consequences. Not really interested in the topic, tbh. Maybe someone else wants to take this.

metaron-uk:
Regarding the question of list defaults, what about giving the user some options about how items are displayed in the list below 'ignore articles when sorting':
screenshot069
Say two options: 
* Lists Content; Title, Title (EpisodeName), Title (EpisodeName) SeasonEpisode
* List SeasonEpisode format; 1.23, 1x23, S1E23, s1e23

This could then be used across PVR & Video items.

IMO the settings screens are a little scattered at the moment from an end user's perspective.
Should you change this type of thing in 'Media', 'Video' 'PVR' or 'Interface' ... What if its a list of Music instead?

@ronie / @phil65 - any suggestions / comments?

metaron-uk:
> Sorry, this currently has low prio for me. Not enough time to think about the proposed changes and consequences. Not really interested in the topic, tbh. Maybe someone else wants to take this.

Well, there's no point in me spending more time on this if I haven't got the backing of a team kodi member. I understand that @ksooo's backend of choice (tvheadend) doesn't support season/episode yet.

I already have what I want in my skin of choice anyhow (my original change was merged into confluence).
If anyone still thinks it is worth progressing, let me know and I'll carry on.

ksooo:
>  I understand that @ksooo's backend of choice (tvheadend) doesn't support season/episode yet.

 Well, that's not correct. tvheadend does support this for a long time. Anyway, you are right with For me, this feature is not very important. Please understand that I have limited spare time and focus on (the many other) things I'm personally more interested in. That might sound egoistic, but sorry, that's life.

metaron-uk:
No problem. We all have to make priority calls. Keep up the good work!

Sent with AquaMail for Android
http://www.aqua-mail.com

metaron-uk:
Looks like I'm not the only person who thinks Season/Episode is worth having http://forum.kodi.tv/showthread.php?tid=175135&pid=2546853#pid2546853
Working on making this more generic and also applicable to video list entries.

metaron-uk:
* Experimenting with my guide and recording data at least 'year' needs some more tweaks / an option to disable using year (even schedules direct data is patchy in quality!). Still very-much WIP.
* Adding FormatListLabel makes it generic but FileItem.h change requires extensive re-compile. Suggestions about where to put this function (serious ones!) gratefully received.

Rechi:
@metaron-uk this needs a rebase

metaron-uk:
I don't have the skills or time to re-write the fileitem related classes. I've been using the alternative solution (https://github.com/xbmc/xbmc/pull/11969) in a local copy of estuary for over a year now without troubles.
Let's have one less open pr that's going nowhere.

ksooo:
Thanks anyway for your contributions.

PR #8011, opened 2015-09-10
Unify music library and files view

Montellese:
This is @notspiff's take at unifying the music library and files view like we have been using it for the video library for quite a while now.

When reviewing please keep in mind that over 90% of the changes consist of code being moved around between classes without applying any cosmetic fixes to them. So please hold off on the cosmetic comments ;-)

I'm not really much of a music library user so this is not well tested from my side. I'll trigger test builds and post the links to them here for people to be able to test this.

**Test builds**
- win32: http://mirrors.xbmc.org/test-builds/win32/KodiSetup-20150910-9b66e67-unify_music.exe
- android x86: http://mirrors.xbmc.org/test-builds/android/x86/kodi-20150910-9b66e67-unify_music-x86.apk
- android arm: http://mirrors.xbmc.org/test-builds/android/arm/kodi-20150910-9b66e67-unify_music-armeabi-v7a.apk
- OSX-64: http://mirrors.xbmc.org/test-builds/osx/x86_64/kodi-20150910-9b66e67-unify_music-x86_64.dmg
- iOS: http://mirrors.xbmc.org/test-builds/darwin/ios/kodi-20150910-9b66e67-unify_music-ios.deb

PR #11785, opened 2017-03-01
[Fix]Speedup Songs node by skipping pointless fetch of cuesheet data

DaveTBlake:
Chipping away at what makes songs node on large library so slow. 

The way embedded cueshseets are implemented is taking *7%* of the total time taken to process the songs node, even when there are no embedded cuesheets in the music collection. Even worse the cuesheet processing in `GetSongsFullByWhere()` is only to get replay gain data for any embedded cuesheets, yet other design flaws means that this replay gain is not actually applied on playback. Not only is the way the data is fetched inefficient, but it is totally pointless!

It is a minor change to stop fetching embedded cuesheet replay gain data by default, and I suggest we backport it for an immediate performance gain.

I am looking at fixing the cuesheet data design and replaygain issues for Leia in a separate PR.


Paxxi (approved):

DaveTBlake:
jenkins build this please

DaveTBlake:
Jenkins fail is libavcode issue, nothing to do with this PR.

Looks like embedded cuesheets have  been broken by the unification of music lib and file views #8011 back in Jarvis. Not only is the lib data ignored, but when not in the lib play from file view results in a corrupted current playlist

PR #11786, opened 2017-03-01
[Fix][Backport]Speed up Songs node by skipping pointless fetch of cuesheet data

DaveTBlake:
Stop fetching embedded cuesheet replay gain data in `GetSongsFullByWhere` by default, it make songs node slow and the data is not used in playback anyway.

The way embedded cueshseets are implemented is taking 7% of the total time taken to process the songs node, even when there are no embedded cuesheets in the music collection. Even worse the cuesheet processing in GetSongsFullByWhere() is only to get replay gain data for any embedded cuesheets, yet other design flaws means that this replay gain is not actually applied on playback. Not only is the way the data is fetched inefficient, but it is totally pointless!

Backport of  #11785 for an immediate 7% performance gain in songs node.


DaveTBlake:
jenkins build this please

PR #11791, opened 2017-03-01
[Fix] XBTFReader: fix loading images from xbt files.

ace20022:
fix regression introduced by 06a781a1b4d1ee715887940e6c915a9c214a1fdc.


Paxxi (approved):
looks sane to me, maybe a comment as to why it's adding one and reading one less than sizeof but that's a minor thing

ace20022:
Comment added, thx Paxxi!

PR #9559, opened 2016-04-05
add support for external PGS/.sup subtitles

taxigps:
use multi-source stream to load external PGS/.sup subtitles.
issue: times of external .sup subtitle stream not sync with video/audio stream, need set the subtitle time delay 4 second.

PR #11795, opened 2017-03-02
[Fix] VideoPlayer: Don't correct start time external pgs subtitle streams.

ace20022:
This fixes the problem described in https://github.com/xbmc/xbmc/pull/9559.


ace20022:
@FernetMenta it feels a bit hacky. It might be better to do this in the multi demuxer. But then I would need to change the the demuxer interface and add more code. What do you think?

FernetMenta (approved):
It could be worse :) Go ahead

ace20022:
jenkins build and merge

PR #11796, opened 2017-03-03
[Win32] installer: added  checks for required updates and SPs during installation + fix

afedchin:
see title

### DIFF
--- a/project/Win32BuildSetup/genNsisInstaller.nsi
+++ b/project/Win32BuildSetup/genNsisInstaller.nsi
@@ -331,9 +331,44 @@ SectionEnd
 SectionGroupEnd
 
 Function .onInit
+  ; WinVista SP2 is minimum requirement
   ${IfNot} ${AtLeastWinVista}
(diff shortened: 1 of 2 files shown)


Paxxi (changes requested):
Looks good overall but I think the xp warning should remain

project/Win32BuildSetup/genNsisInstaller.nsi
```diff
@@ -331,9 +331,44 @@ SectionEnd
 SectionGroupEnd
 
 Function .onInit
+  ; WinVista SP2 is minimum requirement
   ${IfNot} ${AtLeastWinVista}
-    MessageBox MB_OK|MB_ICONSTOP|MB_TOPMOST|MB_SETFOREGROUND "Windows Vista or above required.$\nThis program can not be run on Windows XP"
```
afedchin: `${IfNot} ${AtLeastWinVista}` condition will be true on XP and then the message below will be displayed.

Paxxi: Oh right, then all good

Paxxi (approved):

PR #11798, opened 2017-03-03
[cmake][windows] use correct d3dcompiler depending on target architecture

Rechi:
## Description
Use the correct d3dcompiler depending on the target architecture.

## Motivation and Context
Without this change win32 version of d3dcompiler is also packed in x64 and therefore Kodi can't start.

## How Has This Been Tested?
Tested on win32 and x64.

## Types of change

## Checklist:

### DIFF
--- a/cmake/modules/FindD3DX11Effects.cmake
+++ b/cmake/modules/FindD3DX11Effects.cmake
@@ -7,10 +7,10 @@
 find_file(D3DCOMPILER_DLL
           NAMES d3dcompiler_47.dll d3dcompiler_46.dll
           PATHS
-            "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Wow6432Node\\Microsoft\\Microsoft SDKs\\Windows\\v10.0;InstallationFolder]/Redist/D3D/x86"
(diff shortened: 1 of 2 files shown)


Rechi:
jenkins build this please

Paxxi (changes requested):
Use https://cmake.org/cmake/help/v3.0/variable/CMAKE_SIZEOF_VOID_P.html to check the bitness instead of trying to compile

wsnipex:
32/64 bit checks should go into ArchSetup.cmake

Rechi:
@wsnipex updated
@Paxxi changed as you requested, but documentation for `CMAKE_SIZEOF_VOID_P` says `[...] is determined by a try compile`. So I don't see any advantage on this method.

Paxxi:
@Rechi it's always done by cmake so it's in reality free but the more important thing for me is that it's the correct way to handle it in cmake. Makes it easier to read and grasp if familiar with cmake from other projects

wsnipex:
why not move SDK_ARCH to ArchSetup as well? Sooner or later this var might be useful/needed in other places as well

Rechi:
updated

wsnipex (approved):

Paxxi (approved):

PR #303, opened 2011-07-25
Wifi improvements

garbear:
Contains the following 3 modifications:
- Class NetworkAccessPoint has been augmented to include macAddress and channel
- CNetworkInterfaceLinux::GetAccessPoints() adds these two new fields, and includes general readability fixups
- CNetworkInterfaceWin32::GetAccessPoints() has been implemented

topfs2 has made some progress on a network abstraction layer (https://github.com/topfs2/xbmc/tree/network_abstraction). This will probably obsolesce these changes, but in the meantime, it can't hurt to have better wifi support :)

PR #11801, opened 2017-03-03
[Fix][Backport]Music cleanup hanging on large MySQL databases

DaveTBlake:
Music library cleanup, both on library update, and the overall facility avilable from settings menu, is hanging on very large MySQL (or MariaDB) databases. Smaller db can benefit for a more efficient process too.

The issue is `CMusicDatabase::CleanupArtists` which creates a temp table tmp_delartists of artist ids to keep, but this table contains duplicates and does not have an index. If few artists need removing then on a large collection, tmp_delartists can be very large (e.g. ~12 x number of artists). The query
`DELETE FROM artist WHERE idArtist NOT IN (SELECT idArtist FROM tmp_delartists)`
then hangs.

It is unsurprizing that doing a "NOT IN" on an non-indexed table with 370k rows is unsuccessful!

In comparison testing has proven that an additional step of creating a  temp table of unique artist ids to keep with an index (primary key), and using that in the "NOT IN" works reasonably with both few and many records on both MySQL and SQLite.

Look at having a lot of artists, then removing a small number of them by changes to music files.
Creating a test build for additional testing.  Backport of #11803 

@MilhouseVH any change of testing this on your MySQL system?

### DIFF
--- a/xbmc/music/MusicDatabase.cpp
+++ b/xbmc/music/MusicDatabase.cpp
@@ -3049,8 +3049,14 @@ bool CMusicDatabase::CleanupArtists()
     m_pDS->exec("INSERT INTO tmp_delartists select idArtist from song_artist");
     m_pDS->exec("INSERT INTO tmp_delartists select idArtist from album_artist");
     m_pDS->exec(PrepareSQL("INSERT INTO tmp_delartists VALUES(%i)", BLANKARTIST_ID));
-    m_pDS->exec("delete from artist where idArtist not in (select idArtist from tmp_delartists)");
+    // tmp_delartists contains duplicate ids, and on a large library with small changes can be very large.
+    // To avoid MySQL hanging or timeout create a table of unique ids with primary key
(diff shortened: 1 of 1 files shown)


MilhouseVH:
Tested on my RPi2 with MySQL 5.5 and it's working OK (confirmed the new query is being executed by enabling the MySQL query log). Unfortunately my music library isn't particularly large (8,000 songs or so) and is not in need of cleaning so from a performance point of view I can't really give an indication of the benefit but it looks sane.

I'll include it in tonight's build. Thanks Dave.

DaveTBlake:
>I'll include it in tonight's build. Thanks Dave.

Thanks @MilhouseVH that will be really useful,  I have a test user with RPi and 150k of songs, 30k artists. The SQL works manually, but good to try it in Kodi. His big MySQL db has been turning up some issues for me to fix.

Uatschitchun:
I confirm this one working well with my large source. Tested with Milhouse #0303 on Rpi3 and a build from Daves branch. Without this no cleaning (neither after update nor explicit from settings) was successful!

PR #11802, opened 2017-03-03
VidoePlayer: make no-key-frame-fallback dependent on fps

FernetMenta:
fix for http://forum.kodi.tv/showthread.php?tid=308525


FernetMenta:
jenkins build this please

PR #11803, opened 2017-03-03
[Fix]Music cleanup hanging on large MySQL databases

DaveTBlake:
Music library cleanup, both on library update, and the overall facility avilable from settings menu, is hanging on very large MySQL (or MariaDB) databases. Smaller db can benefit for a more efficient process too.

The issue is `CMusicDatabase::CleanupArtists` which creates a temp table tmp_delartists of artist ids to keep, but this table contains duplicates and does not have an index. If few artists need removing then on a large collection, tmp_delartists can be very large (e.g. ~12 x number of artists). The query
DELETE FROM artist WHERE idArtist NOT IN (SELECT idArtist FROM tmp_delartists)
then hangs.

It is unsurprizing that doing a "NOT IN" on an non-indexed table with 370k rows is unsuccessful!

In comparison testing has proven that an additional step of creating a temp table of unique artist ids to keep with an index (primary key), and using that in the "NOT IN" works reasonably with both few and many records on both MySQL and SQLite.

Backported as #11801


DaveTBlake:
Backport tested out, so going to merge

PR #11804, opened 2017-03-04
language addons: remove kaitoast close hack 

tamland:
Should fix http://trac.kodi.tv/ticket/17341

ping @Montellese I'm unsure what the purpose of this toast closing was... 
It doesn't appear to make any difference

### DIFF
--- a/xbmc/addons/LanguageResource.cpp
+++ b/xbmc/addons/LanguageResource.cpp
@@ -26,6 +26,7 @@
 #include "utils/StringUtils.h"
 #include "utils/Variant.h"
 #include "messaging/helpers/DialogHelper.h"
+#include "Skin.h"
 
 using namespace KODI::MESSAGING;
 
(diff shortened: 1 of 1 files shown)


Montellese:
IIRC the problem was that after switching languages there was immediately a kai toast which was not displayed properly because the language strings weren't loaded yet. But it doesn't happen all the time so I wasn't able to reliably reproduce it. This seemed to solve the issue for other people but obviously it's not a clean solution.

chewitt:
@tamland LE 8.0.1 test builds with this change applied to 17.1-RC1 now complete upgrade from Jarvis and older Kodi versions properly. Thanks :)

wrxtasy:
Unfortunately this did not fix "endless Add-on migration in progress" Jarvis > Krypton upgrade bug for me.

I had no extra language Addons installed. One extra Skin, and a Weather Addon.

Only solution to the Addons.db migration bug was to delete all /Database/Addons2x.db and restart Kodi.

PR #11810, opened 2017-03-05
Revert "[filesystem/samba] - use vfs instead of $home"

koying:
Samba wants .smb in the HOME dir.
No way around it unless libsmbclient is patched so that it behaves differently.

Fixes the issue on Android where clean installs cannot browse SMB


Memphiz (approved):
Thx and sorry for the trouble

PR #11813, opened 2017-03-05
Fix CHANGELOG if not exist

cyr-ius:
CPACK - Variable CHANGELOG if not exist

## Description
When build old version , changelog is empty , because  git list since last month.

## Motivation and Context
Fix this.

## How Has This Been Tested?
CPack old version Krypton 17.0 is OK

## Types of change
- [X ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X ] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [] I have added tests to cover my change


wsnipex:
could you fix the git log --since so that it always returns something instead?
--since="30 days ago" should work

Cyr-ius:
OK , i changed. It works.

PR #11814, opened 2017-03-05
Fix allocation of 0-length array

thebrid:
## Description
Using Valgrind to troubleshoot a crash, came across this. Not the crash I was looking for but interesting nonetheless. If we're called with argc=0, we allocate a 0-length array, pass it to Python then we get a memory error as Python tries to access member 0 of the 0-length array.

## Motivation and Context
Fixes an invalid memory read.

## How Has This Been Tested?
I clicked around a bunch and didn't manage to get it to crash. Re-ran in valgrind and the invalid memory read is no longer there. Ran all existing tests.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/interfaces/python/PythonInvoker.cpp
+++ b/xbmc/interfaces/python/PythonInvoker.cpp
@@ -20,6 +20,7 @@
 
 // python.h should always be included first before any other includes
 #include <Python.h>
+#include <iterator>
 #include <osdefs.h>
 
 #include "system.h"
@@ -95,6 +96,22 @@ static const std::string getListOfAddonClassesAsString(XBMCAddon::AddonClass::Re
   return message;
 }
 
+static std::vector<std::vector<char>> storeArgumentsCCompatible(std::vector<std::string> const & input)
+{
+  std::vector<std::vector<char>> output;
+  std::transform(input.begin(), input.end(), std::back_inserter(output),
+                [](std::string const & i) { return std::vector<char>(i.c_str(), i.c_str() + i.length() + 1); });
+  return output;
+}
+
+static std::vector<char *> getCPointersToArguments(std::vector<std::vector<char>> & input)
+{
+  std::vector<char *> output;
+  std::transform(input.begin(), input.end(), std::back_inserter(output), 
+                [](std::vector<char> & i) { return &i[0]; });
+  return output;
+}
+
 CPythonInvoker::CPythonInvoker(ILanguageInvocationHandler *invocationHandler)
   : ILanguageInvoker(invocationHandler),
     m_threadState(NULL), m_stop(false)
@@ -140,18 +157,8 @@ bool CPythonInvoker::execute(const std::string &script, const std::vector<std::s
 
   // copy the arguments into a local buffer
   unsigned int argc = arguments.size();
-  char ** argv = nullptr;
-  
-  if (argc > 0)
-  {
-    argv = new char*[argc];
-  }
-
-  for (unsigned int i = 0; i < argc; i++)
-  {
-    argv[i] = new char[arguments.at(i).length() + 1];
-    strcpy(argv[i], arguments.at(i).c_str());
-  }
+  std::vector<std::vector<char>> argvStorage = storeArgumentsCCompatible(arguments);
+  std::vector<char *> argv = getCPointersToArguments(argvStorage);
 
   CLog::Log(LOGDEBUG, "CPythonInvoker(%d, %s): start processing", GetId(), m_sourceFile.c_str());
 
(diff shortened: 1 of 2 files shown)


Paxxi (approved):
Looks good to me

xbmc/interfaces/python/PythonInvoker.cpp
```diff
@@ -147,7 +147,12 @@ bool CPythonInvoker::execute(const std::string &script, const std::vector<std::s
 
   // copy the arguments into a local buffer
   m_argc = arguments.size();
-  m_argv = new char*[m_argc];
+
+  if(m_argc > 0)
```
garbear: missing whitespace

xbmc/interfaces/python/PythonInvoker.cpp
```diff
@@ -147,7 +147,12 @@ bool CPythonInvoker::execute(const std::string &script, const std::vector<std::s
...
+  {
+    m_argv = new char*[m_argc];
+  }
```
garbear: also the braces aren't required for only 1 line

garbear:
other than the code style comments it looks good

xbmc/interfaces/python/PythonInvoker.cpp
```diff
@@ -147,7 +147,12 @@ bool CPythonInvoker::execute(const std::string &script, const std::vector<std::s
 
   // copy the arguments into a local buffer
   m_argc = arguments.size();
-  m_argv = new char*[m_argc];
+
+  if(m_argc > 0)
+  {
+    m_argv = new char*[m_argc];
```
garbear: this will leak if CPythonInvoker::execute() is called twice

xbmc/interfaces/python/PythonInvoker.cpp
thebrid: Is it OK if I keep them? I've seen people in the past attempt to add new lines to conditional blocks without braces so I try to add braces as a rule.

Paxxi: Extra braces are fine unless they severely reduce readability so keep them

xbmc/interfaces/python/PythonInvoker.cpp
thebrid: I figured this wasn't a regression so I didn't change it. I don't mind making the change though. Will send another commit in a little while.

xbmc/interfaces/python/PythonInvoker.cpp
```diff
@@ -95,9 +97,27 @@ static const std::string getListOfAddonClassesAsString(XBMCAddon::AddonClass::Re
...
+	return output;
+}
+
+static std::vector<char *> convertVectorsToCharPtrs(std::vector<std::vector<char>> & input)
+{
+	std::vector<char *> output;
+	std::transform(input.begin(), input.end(), std::back_inserter(output), [](std::vector<char> & i) { return &i[0]; });
+	return output;
+}
+
```
wsnipex: indentation on the whole block is off. We use 2 spaces.

thebrid: :+1:

thebrid:
@garbear feedback implemented
@wsnipex indentation should be fixed now

xbmc/interfaces/python/PythonInvoker.cpp
```diff
@@ -147,7 +147,12 @@ bool CPythonInvoker::execute(const std::string &script, const std::vector<std::s
 
   // copy the arguments into a local buffer
   m_argc = arguments.size();
-  m_argv = new char*[m_argc];
+  
+  if(m_argc > 0)
+  {
+    m_argv = new char*[m_argc];
+  }
```
garbear: This block doesn't fix the cosmetic issues I mentioned earlier

xbmc/interfaces/python/PythonInvoker.cpp
```diff
@@ -114,12 +134,6 @@ CPythonInvoker::~CPythonInvoker()
   Stop(true);
   pulseGlobalEvent();
 
-  if (m_argv != NULL)
-  {
-    for (unsigned int i = 0; i < m_argc; i++)
-      delete [] m_argv[i];
-    delete [] m_argv;
-  }
```
garbear: Replacing the member variables with local ones is the correct fix. However, you replace some overly-complex c++ code with some overly-complicated std container code. I recommend splitting this into two commits, one that moves this block  into CPythonInvoker::execute() and replaces the members with locals (keeping the c++ memory management), and then one that converts from `std::vector<std::string>` to `char**`.

thebrid:
@garbear changed commits as you requested.

Paxxi:
jenkins build this please

wsnipex:
jenkins build this please

DaveTBlake:
@thebrid check forum PM for using valgrind on another Kodi issue.

Rechi:
If you refer to Travis CI, we don't care about it and it also fails a bit randomly. Only jenkins must be successful.

wsnipex:
@garbear ping

garbear (approved):

wsnipex:
thanks for the patience :)

MilhouseVH:
Error in youtube plugin: http://sprunge.us/UZBD

Reverting this PR fixes the error.

thebrid:
@MilhouseVH checking

PR #11816, opened 2017-03-06
Fix EGLNativeTypeAmlogic for probing resolutions

Stane1983:
I noticed issue with latest official Amlogic kernel releases (3.10.33)
where /sys/class/amhdmitx/amhdmitx0/disp_cap reports also 4k2k resolutions
on S805 boards which results in 4K resolutions showing as available.

That makes 4K resolutions available to be set which is misleading and in the
end user may actually select unsupported resolution.

Also my suggestion is to merge it to Krypton too.

### DIFF
--- a/xbmc/windowing/egl/EGLNativeTypeAmlogic.cpp
+++ b/xbmc/windowing/egl/EGLNativeTypeAmlogic.cpp
@@ -171,8 +171,11 @@ bool CEGLNativeTypeAmlogic::ProbeResolutions(std::vector<RESOLUTION_INFO> &resol
   RESOLUTION_INFO res;
   for (std::vector<std::string>::const_iterator i = probe_str.begin(); i != probe_str.end(); ++i)
   {
-    if(aml_mode_to_resolution(i->c_str(), &res))
-      resolutions.push_back(res);
+    if (((StringUtils::StartsWith(i->c_str(), "4k2k")) && (aml_support_h264_4k2k() > AML_NO_H264_4K2K)) || !(StringUtils::StartsWith(i->c_str(), "4k2k")))
+    {
+      if (aml_mode_to_resolution(i->c_str(), &res))
+        resolutions.push_back(res);
+    }
   }
   return resolutions.size() > 0;
 


wrxtasy:
Yes i've seen the same problem, I've been playing around with LE Krypton on an ODROID-C1+ / S805 this last week with my 4K connected TV.

I actually patched out those unsupported 4k2k resolutions in the 3.10.xx Kernel.

This would be a better solution as we all know how bad vendors are at Kernel bug fixes.

chewitt:
@wrxtasy @Stane1983 https://github.com/LibreELEC/linux-amlogic/pull/4 addresses this in the 3.10 kernel that existing LE devices are using.

wrxtasy:
Yes but not everyone is using LE linux-amlogic 3.10.xx Kernels. 
There are other distro's to consider - not just LE.

Stane's patch covers everyone downstream.

Stane1983:
@chewitt, please check which platforms are covered by commit you mentioned. S805 is meson8. And yes, I tested this with LE kernel. No offense, but wetek doesn't have S805 device so issue was not visible before?

MrMC:
sorry to hijack, @wrxtasy , best way to get in touch with you ?

MartijnKaijser:
@peak3d ping

xbmc/windowing/egl/EGLNativeTypeAmlogic.cpp
```diff
@@ -171,8 +171,10 @@ bool CEGLNativeTypeAmlogic::ProbeResolutions(std::vector<RESOLUTION_INFO> &resol
   RESOLUTION_INFO res;
   for (std::vector<std::string>::const_iterator i = probe_str.begin(); i != probe_str.end(); ++i)
   {
-    if(aml_mode_to_resolution(i->c_str(), &res))
-      resolutions.push_back(res);
+    if (((StringUtils::StartsWith(i->c_str(), "4k2k")) && (aml_support_h264_4k2k() > AML_NO_H264_4K2K)) || !(StringUtils::StartsWith(i->c_str(), "4k2k"))) {
```
peak3d: Questions:
- what is aml_support_h264_4k2k() requesting? hardware? A static list anywhere around?
- EGLNativeTypeAmlogic is used for Odroid C2 / S905 as well, where is this check for this?

Coding style:
- Pls put brackets in a new line, thx!

Stane1983: aml_support_h264_4k2k() from AMLUtils.cpp is checking video codec profile (basically parsing /sys/class/amstream/vcodec_profile) if h264 4K decoding is supported with codec or not. S805 cannot display or decode 4K so value of aml_support_h264_4k2k() is AML_NO_H264_4K2K (see https://github.com/xbmc/xbmc/blob/master/xbmc/utils/AMLUtils.h#L44) For 9xx devices there is no need for these checks because they all can display 4K and that's why I did not include 2160 resolution checks (S9xx have different resolution strings. Eg. M8 is 4k2k30hz while on 9xx it's 2160p60hz.

About brackets, will do it in cpl mins.

PR #11820, opened 2017-03-06
Fix compilation error: ‘sprintf’ is not a member of ‘fmt’

candrews:
In order to use fmt::sprintf, fmt/printf.h must be included. See http://fmtlib.net/dev/api.html#printf-formatting

## Description

Currently, compilation fails with this error:
```
xbmc/utils/StringUtils.h:73:16: error: ‘sprintf’ is not a member of ‘fmt’
        result = fmt::sprintf(fmt, std::forward<Args>(args)...);
```
I'm using fmt 3.0.1.

## Motivation and Context

There's a compilation failure :-)

## How Has This Been Tested?

Compiled, and this failure doesn't occur.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/utils/StringUtils.h
+++ b/xbmc/utils/StringUtils.h
@@ -37,6 +37,7 @@
 #include <locale>
 
 #include <fmt/format.h>
+#include <fmt/printf.h>
 
 #include "LangInfo.h"
 #include "XBDateTime.h"


Paxxi:
Thanks! I'm curious which platform/compiler did you notice the error on? Jenkins built fine so it's a bit odd.

Rechi:
version 3.0.1 from https://github.com/fmtlib/fmt has no file named `printf.h`

candrews:
I was testing with git master of libfmt not libfmt 3.0.1 (as I thought I was doing)... sorry about that.

PR #11821, opened 2017-03-07
Replace deprecated glibc aliases.

stintel:
## Description
MNTTAB and MOUNTED are deprecated aliases. Replace them with
_PATH_MNTTAB and _PATH_MOUNTED.

## Motivation and Context
Following https://www.gnu.org/software/libc/manual/html_node/Mount-Information.html, the deprecated names should not be used. This is also required for compatibility with musl libc.

## How Has This Been Tested?
On glibc and uClibc this change is a noop, but compile-tested against glibc anyway.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/DllLoader/exports/emu_msvcrt.cpp
+++ b/xbmc/cores/DllLoader/exports/emu_msvcrt.cpp
@@ -1219,8 +1219,8 @@ extern "C"
   {
     FILE* file = NULL;
 #if defined(TARGET_LINUX) && !defined(TARGET_ANDROID)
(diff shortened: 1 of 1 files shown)


Paxxi:
Thank you. @wsnipex your thoughts?

wsnipex:
looks good.
on an related note: I wonder if we'll ever get rid of the win emu layers..

wsnipex:
jenkins build this please

Paxxi:
I'm working on it, currently it's only used by libdvd, once that's fixed we can drop it all

PR #11824, opened 2017-03-07
[Peripherals] Fix segfault by adding some synchronization to peripheral add-ons

garbear:
This fixes a segfault at exit when peripheral add-ons misbehave. In particular, if the add-on hangs in another thread, then it can be destroyed while still in use. This was causing a segfault on exit.

Another segfault (or possibly the same one) has been fixed by moving g_peripherals to ServiceManager. Before, it tried to release resources when the static object was deconstructed after the app shut down.

## Types of change

### DIFF
--- a/xbmc/games/ports/PortMapper.cpp
+++ b/xbmc/games/ports/PortMapper.cpp
@@ -28,7 +28,8 @@ using namespace GAME;
 using namespace JOYSTICK;
 using namespace PERIPHERALS;
 
-CPortMapper::CPortMapper()
+CPortMapper::CPortMapper(PERIPHERALS::CPeripherals& peripheralManager) :
+  m_peripheralManager(peripheralManager)
 {
   CPortManager::GetInstance().RegisterObserver(this);
 }
@@ -56,7 +57,7 @@ void CPortMapper::ProcessPeripherals()
   auto& oldPortMap = m_portMap;
 
   PeripheralVector devices;
-  g_peripherals.GetPeripheralsWithFeature(devices, FEATURE_JOYSTICK);
+  m_peripheralManager.GetPeripheralsWithFeature(devices, FEATURE_JOYSTICK);
 
   std::map<PeripheralPtr, IInputHandler*> newPortMap;
   CPortManager::GetInstance().MapDevices(devices, newPortMap);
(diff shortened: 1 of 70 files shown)


garbear:
I've pushed a new commit and updated the PR description. It turns out, there were only 2 segfaults at work here, one from misbehaving peripheral add-ons, and one due to g_peripherals being a static variable and trying to release resources that had already disappeared.

MartijnKaijser:
Does this also affect v17? 
Jenkins build this please

garbear:
Yes, peripheral.joystick is still hanging on the Steam Link, so without this Kodi crashes on exit 100% of the time.

garbear:
This almost concludes my Krypton backports. The only other bugs I care to fix are the iMON incompatibility and better Shield controller support. If we're gearing up for 17.1 release I'm fine waiting for 17.2 or 18.0.

garbear:
Hold off on merging. The shutdown stuff still isn't happening in the right order.

garbear:
lol, peripherals shutdown was 1 line off. Also pushed a commit for a third segfault that arose from starting/stopping things in the proper order.

jenkins build this please

garbear:
jenkins build this please

xbmc/games/ports/PortMapper.cpp
```diff
@@ -23,13 +23,18 @@
...
 using namespace KODI;
 using namespace GAME;
 using namespace JOYSTICK;
 using namespace PERIPHERALS;
 
-CPortMapper::CPortMapper()
+CPortMapper::CPortMapper(PERIPHERALS::CPeripherals* peripheralManager) :
+  m_peripheralManager(peripheralManager)
 {
+  assert(m_peripheralManager != nullptr);
```
thebrid: If null doesn't make sense here, should we pass by reference rather than pointer?

garbear: Yes, passing by reference would be better. I'll update the PR.

garbear:
updated with @thebrid's comment. jenkins build this please

garbear:
Build errors look unrelated, but both OSX and linux64 failed, so again we try. Jenkins build this please.

eigendude:
third time's the charm. jenkins build and merge

garbear:
let's try again. jenkins build and merge

wsnipex:
jenkins build this please

garbear:
There's an error with the test cases. Moving around the order of services on startup typically breaks things. I'll address this when I get a chance. Until then, this problem only manifests with broken peripheral add-ons, which is only peripheral.joystick on the Steam Link (and I work around it). So not critical.

garbear:
test cases fixed! jenkins build this please

PR #11830, opened 2017-03-08
[depends] libfmt: remove BASE_URL pointing to github

Montellese:
I just realized that I uploaded the win32 deps archive but not the one for depends builds and that the `Makefile` still pointed to github. I have uploaded `fmt-3.0.1.tar.gz` to our mirrors.

### DIFF
--- a/tools/depends/target/libfmt/Makefile
+++ b/tools/depends/target/libfmt/Makefile
@@ -22,7 +22,7 @@ else
     TARBALLS_LOCATION = $(ROOT_DIR)
     RETRIEVE_TOOL := curl
     ARCHIVE_TOOL := tar --strip-components=1 -xf
-    CMAKE := cmake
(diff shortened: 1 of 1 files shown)


wsnipex:
I assume fmt-3.0.0.tar.gz is a typo, since you uploaded http://mirrors.kodi.tv/build-deps/sources/fmt-3.0.1.tar.gz

as noted in https://github.com/xbmc/xbmc/pull/11039#discussion_r104319995 can you please add CMAKE_INSTALL_PREFIX for the native compile case? e.g. on L25

Montellese:
Yeah sorry, `fmt-3.0.1.tar.gz` as also used in the `Makefile`. I've added the extra line. Is it ok like this?
jenkins build this please.

wsnipex (approved):

PR #11835, opened 2017-03-10
[jenkins][windows] don't remove previous compiled version of ffmpeg & libdvd if nothing changed

Rechi:
## Description
Copy the behaviour for depends on other platform for ffmpeg and libdvd.
They are treated independent so it could happen that non, only one or both are compiled.
If Configuration is set to Release they always get rebuild.

## Motivation and Context
speed up compile time on jenkins for windows

## How Has This Been Tested?
locally tested with different parameters

## Types of change

## Checklist:

### DIFF
--- a/tools/buildsteps/win32/buildffmpeg.sh
+++ b/tools/buildsteps/win32/buildffmpeg.sh
@@ -12,6 +12,17 @@ FFMPEG_TARGET_OS=mingw32
 do_loaddeps $FFMPEG_VERSION_FILE
 FFMPEGDESTDIR=/xbmc/lib/win32/$LIBNAME
 
+if [ "$(pathChanged $FFMPEGDESTDIR $FFMPEG_VERSION_FILE /xbmc/project/BuildDependencies/DownloadMingwBuildEnv.bat /xbmc/tools/buildsteps/win32)" == "0" ]
+then
+  cp $FFMPEGDESTDIR/bin/*.dll /xbmc/system/
+  if [ -f $BGPROCESSFILE ]; then
+    rm $BGPROCESSFILE
+  fi
+  exit
+else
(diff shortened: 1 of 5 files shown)


afedchin:
the ffmpeg depends on gnutls package if it's changed then ffmpeg must be recompiled.
see https://github.com/xbmc/xbmc/blob/master/project/BuildDependencies/DownloadMingwBuildEnv.bat#L32

Rechi:
@afedchin adding `project/BuildDependencies/DownloadMingwBuildEnv.bat` to the checked files/directories should be enough?

Rechi:
added `project/BuildDependencies/DownloadMingwBuildEnv.bat` to checked files/directories

Paxxi:
I can't really judge this, I like the idea but maybe @wsnipex is a better judge if the script is sound?

tools/buildsteps/win32/buildffmpeg.sh
```diff
@@ -12,6 +12,17 @@ FFMPEG_TARGET_OS=mingw32
 do_loaddeps $FFMPEG_VERSION_FILE
 FFMPEGDESTDIR=/xbmc/lib/win32/$LIBNAME
 
+if [ "$(pathChanged $FFMPEGDESTDIR $FFMPEG_VERSION_FILE /xbmc/project/BuildDependencies/DownloadMingwBuildEnv.bat /xbmc/tools/buildsteps/win32)" == "0" ]
```
wsnipex: pathChanged only tests the first param. So you effectively only check if $FFMPEGDESTDIR changed.
You either need a loop here on in the pathChanged function

wsnipex:
looks ok to me.

afedchin (approved):

Rechi:
jenkins build and merge

PR #11839, opened 2017-03-11
[xbmc] General cleanup of ApplicationMessenger

Paxxi:
Remove the circular dependency with CApplication
Move the DelayedThreadMessage to the cpp file as it's an implementation detail.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/messaging/ApplicationMessenger.cpp
+++ b/xbmc/messaging/ApplicationMessenger.cpp
@@ -23,14 +23,26 @@
 #include <memory>
 #include <utility>
 
-#include "Application.h"
 #include "guilib/GraphicContext.h"
+#include "guilib/GUIMessage.h"
+#include "messaging/IMessageTarget.h"
 #include "threads/SingleLock.h"
 
 namespace KODI
 {
 namespace MESSAGING
 {
+
+class CDelayedMessage : public CThread
+{
+  public:
+    CDelayedMessage(ThreadMessage& msg, unsigned int delay);
+    virtual void Process() override;
+
+  private:
+    unsigned int   m_delay;
+    ThreadMessage  m_msg;
+};
 
 CDelayedMessage::CDelayedMessage(ThreadMessage& msg, unsigned int delay) : CThread("DelayedMessage")
 {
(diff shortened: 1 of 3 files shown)


Paxxi:
jenkins build this please

xbmc/messaging/ApplicationMessenger.cpp
```diff
@@ -101,7 +118,7 @@ int CApplicationMessenger::SendMsg(ThreadMessage&& message, bool wait)
     message.result = std::make_shared<int>(-1);
     // check that we're not being called from our application thread, else we'll be waiting
     // forever!
-    if (!g_application.IsCurrentThread())
+    if (!CThread::IsCurrentThread(m_threadID))
```
da-anda: wouldn't it be better to call `m_threadID` something like `m_guiThreadID` so that the code is reading more obvious?

Paxxi: Sounds like a good idea

Paxxi:
jenkins build this please

xbmc/messaging/ApplicationMessenger.h
```diff
@@ -400,6 +382,18 @@ class CApplicationMessenger
...
+   * \brief Set the UI thread id to avoid messenger being dependent on
+   * CApplication to determine if marshaling is required
+   * \param thread The UI thread ID
+   */
+  void SetGUIThread(ThreadIdentifier thread) { m_guiThreadId = thread; }
+
+  /*
+   * \brief Signals the shutdown of the application and message processing
+   */
+  void Shutdown() { m_bStop = true; }
```
xhaggi: In most cases we use `Stop` or `Deinit` as method name. It would be nice if we could do the same here.

Paxxi: changed.

Paxxi:
jenkins build this please

xbmc/messaging/ApplicationMessenger.h
xhaggi: Thx.

PR #11842, opened 2017-03-11
Fix uninitialised memory read in RenderManager

thebrid:
## Description
We sometimes read members from SPresent before we've initialised them. Initialise them in a new constructor.

## Motivation and Context
The behaviour here is undefined. From my testing on Linux, it doesn't seem to cause a big problem, but we're not guaranteed good behaviour.

## How Has This Been Tested?
Ran existing unit tests. They passed. Also ran kodi manually on x86_64 Ubuntu Linux to trigger this codepath. Worked fine.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.h
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.h
@@ -205,9 +205,9 @@ protected:
 
   struct SPresent
   {
-    double         pts;
-    EFIELDSYNC     presentfield;
-    EPRESENTMETHOD presentmethod;
+    double         pts = 0.0;
+    EFIELDSYNC     presentfield = FS_NONE;
+    EPRESENTMETHOD presentmethod = PRESENT_METHOD_SINGLE;
   } m_Queue[NUM_BUFFERS];
 
   std::deque<int> m_free;


FernetMenta:
> We sometimes read members from SPresent before we've initialised them. Initialise them in a new constructor.

Please elaborate. Using this structure when it was not filled with real data from player makes no sense.

xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.h
```diff
@@ -222,6 +222,7 @@ class CRenderManager
 
   struct SPresent
   {
+    SPresent();
```
FernetMenta: Instead of using a constructor you can init the members by assignemnt directly below.

thebrid: @FernetMenta looks like a race condition particularly involving the presentmethod field. We set the value from one thread here:

https://github.com/xbmc/xbmc/blob/3a27e3ba321dacdf7af05a3bab6bc39de9635738/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp#L847

And we read it from a *different* thread here:

https://github.com/xbmc/xbmc/blob/3a27e3ba321dacdf7af05a3bab6bc39de9635738/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp#L909

The order in which the two are executed can differ. I've observed both the correct & incorrect order.

FernetMenta: Are you talking about unit tests or real runtime tests? CRenderManager::Render should not be called if nothing valid is in the render buffers.
If a SPresent structure is used that was not filled with valid data, something has gone wrong before.

thebrid: That was a real runtime test in valgrind.

FernetMenta: maybe initialise with something like PRESENT_METHOD_INVALID. that makes it clearer that the linked render buffer does not hold any valid frame.

MartijnKaijser:
@fernetmenta @thebrid a gentle reminder what needs to be done with this?

FernetMenta:
my question has not been answered so far and I am not aware of any issues this PR should fix.

PR #11847, opened 2017-03-13
[gui] follow up for PR #11704

xhaggi:
@ksooo as discussed at #11704.


ksooo:
You should change "actualitems" to "numactualitems" as well, then, for consistency reasons. Otherwise I see no point in the name change of "nonfolderitems" to "numnonfolderitems".

xhaggi:
Added the missing break, thanks.

> You should change "actualitems" to "numactualitems" as well, then, for consistency reasons. Otherwise I see no point in the name change of "nonfolderitems" to "numnonfolderitems".

That's the reason why I want to double check new GUI labels etc. If it was merged it becomes hard to change. @phil65 does the same mistake if you ask me. Nothing will point-out that this info label will return the size of items. That's the reason why `NumItem` is prefixed with `num`. We should use the same pattern for new info labels if it makes sense and this is the case here.

xhaggi:
I've added a separate commit which renames the info label `ActualItems` to `NumAllItems`. It leaves the old one in for backward compatibility.

xhaggi:
I've dropped the backward compat stuff because `ActualItems` was merged into Leia only. In that case we don't need it.

xhaggi:
Jenkins build this please

ksooo:
Have you seen my latest comments? I'm not sure the PR is ready to merge.

xhaggi:
Which one?

ksooo:
Seems I forgot to submit my review. ;-)

xbmc/GUIInfoManager.cpp
```diff
@@ -2407,7 +2407,7 @@ const infomap container_bools[] ={{ "onnext",           CONTAINER_MOVE_NEXT },
                                   { "numpages",         CONTAINER_NUM_PAGES },
                                   { "numitems",         CONTAINER_NUM_ITEMS },
                                   { "numnonfolderitems", CONTAINER_NUM_NONFOLDER_ITEMS },
-                                  { "actualitems",      CONTAINER_ACTUAL_ITEMS },
+                                  { "numallitems",      CONTAINER_NUM_ALL_ITEMS },
```
xhaggi: Okay, lets ping our skinners @ronie @phil65 any objections?

xbmc/guilib/GUIBaseContainer.cpp
```diff
@@ -1304,27 +1304,27 @@ std::string CGUIBaseContainer::GetLabel(int info) const
...
+      for (auto item : m_items)
+      {
+        if (!item->m_bIsFolder)
+          numItems++;
+      }
+      label = StringUtils::Format("%u", numItems);
+    }
+    break;
   default:
       break;
```
xhaggi: Sure, but in the same commit. I don't want to add a new one only for that.

xbmc/GUIInfoManager.cpp
xhaggi: Btw I don't have a strong opinion on that, but all means all :smile:

xbmc/guilib/GUIBaseContainer.cpp
ksooo: sure

xbmc/GUIInfoManager.cpp
ronie: no objections from my end

xhaggi:
jenkins build this please

PR #11849, opened 2017-03-13
[PVR] Fix trac17374: Settings deinitialized too early on app exit.

ksooo:
This fixes http://trac.kodi.tv/ticket/17374

Settings are cleared too early on app exit => https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/PVRManager.cpp#L531 does not get the right settings values on app shutdown => wake up command gets not executed.

Bug was introduced by a fix I did some time ago, to only call wake up command on app exit, not on every pvr manager restart (which can happen quite some time during app lifetime).

@Montellese @FernetMenta I moved deinit of settings in CApplication::Cleanup() to a later point in time, namely to CServiceManager::Deinit(), after PVR mananger shutdown. Is this okay? Runtime test went fine.

### DIFF
--- a/xbmc/pvr/PVRManager.cpp
+++ b/xbmc/pvr/PVRManager.cpp
@@ -159,13 +159,16 @@ CPVRManager::CPVRManager(void) :
     m_progressBar(nullptr),
     m_progressHandle(nullptr),
     m_managerState(ManagerStateStopped),
-    m_isChannelPreview(false)
+    m_isChannelPreview(false),
+    m_bSettingPowerManagementEnabled(CServiceBroker::GetSettings().GetBool(CSettings::SETTING_PVRPOWERMANAGEMENT_ENABLED)),
+    m_strSettingWakeupCommand(CServiceBroker::GetSettings().GetString(CSettings::SETTING_PVRPOWERMANAGEMENT_SETWAKEUPCMD))
 {
   CAnnouncementManager::GetInstance().AddAnnouncer(this);
 }
 
 CPVRManager::~CPVRManager(void)
 {
+  CServiceBroker::GetSettings().UnregisterCallback(this);
   CAnnouncementManager::GetInstance().RemoveAnnouncer(this);
   CLog::Log(LOGDEBUG,"PVRManager - destroyed");
 }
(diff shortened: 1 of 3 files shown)


FernetMenta (commented):
Not sure if this is safe. Kodi registers a lot of callbacks into settings and as a result settings has to be cleared down as long as those objects exists.

FernetMenta:
Those kind of issues occur because settings are used as global variables. Have a look at AE:
https://github.com/xbmc/xbmc/blob/master/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp#L2544

It loads all settings into a member every time OnChange is fired by settings.

ksooo:
> Those kind of issues occur because settings are used as global variables.

 These are actually very valid points. I will try to find a small pragmatic fix for Krypton and do it right (settings changes callback) for Leia.

ksooo:
@FernetMenta I now implemented the settings listener approach you suggested - for now for the settings used during app exit (and causing the bug due to this). Good to go now? I will take care of the other settings used by PVR in another PR.

FernetMenta:
@ksooo for v17 is may be ok but for master you might want to make it more robust:

- PVRManager is global and g_PVRManager.SetPowerManagementEnabled can be called from everywhere by any other component. By doing this sync to settings is lost. It is not clear that those are only supposed to be called on settings change
- thread safety? maybe not an issue for the wakeup related settings but should be considered when looking at the rest.

ksooo:
Fair enough. I will change the implementation to address your comments.

ksooo:
@FernetMenta I addressed your latest comments re thread safety and sync. Good to go now?

xbmc/pvr/PVRManager.cpp
```diff
@@ -249,6 +252,28 @@ CPVRClientsPtr CPVRManager::Clients(void) const
...
 
+void CPVRManager::OnSettingChanged(const CSetting *setting)
+{
+  if (setting == nullptr)
+    return;
+
+  const std::string &settingId = setting->GetId();
+  if (settingId == CSettings::SETTING_EPG_DAYSTODISPLAY)
+  {
+    m_addons->SetEPGTimeFrame(static_cast<const CSettingInt*>(setting)->GetValue());
```
FernetMenta: @ksooo don't get me wrong, I don't want to be picky and I don't block this in any way. Just sharing ideas. This has been like this before, so no regression.
Hijacking the cb thread (in most cases gui thread) having it call into addons is probbly not the safest way to handle callbacks. AE for example just sets an event on the callback and the AE thread does the actual work:
https://github.com/xbmc/xbmc/blob/master/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp#L2616

PVRManager has its own thread as well and can do this work by itself.

Again, just want to share ideas

ksooo: > @ksooo don't get me wrong, I don't want to be picky and I don't block this in any way. Just sharing ideas. This has been like this before, so no regression.
Hijacking the cb thread (in most cases gui thread) having it call into addons is probbly not the safest way to handle callbacks.

All fine. Thanks for the feedback. As you said, no regression here. I will address this in a nother PR. I guess there are even more places in PVR using this anti pattern. ;-)

ksooo:
jenkins build this please

PR #11851, opened 2017-03-14
Fix Path Traversal Described In CVE-2017-5982

jacob-baines:
## Description
I noticed that NVD had assigned [CVE-2017-5982 ](https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2017-5982)to Kodi. I also found that there was an [exploid-db entry](https://www.exploit-db.com/exploits/41312/) as well as a [metasploit module](https://github.com/rapid7/metasploit-framework/pull/7980/files). However, I couldn't find a patch in the code base.

After some searching, I found a brief discussion on the [forums](http://forum.kodi.tv/showthread.php?tid=144110&pid=2527773#pid2527773) in which @MartijnKaijser welcomed patches. This pull request introduces a simple fix.

## Motivation and Context
The bug is a local file inclusion due to path traversal when using image:// in conjunction with an HTTP request. The following request command will download /etc/passwd from a remote Kodi web server.

```sh
albinolobster@ubuntu:~$ wget http://192.168.1.40:8080/%69%6d%61%67%65%2f%69%6d%61%67%65%3a%2f%2f%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%36%35%25%37%34%25%36%33%25%32%66%25%37%30%25%36%31%25%37%33%25%37%33%25%37%37%25%36%34
```
The decoded request looks like this: "image/image:/../../../../../../../../../../etc/passwd"

In order to stop this behavior I've dropped a very simple check for ".." in the pathToURL value in CHTTPImageHandler's constructor. This might not be a perfect solution, but it does catch the problem.

As an aside, one might be interested why something like image/image:/etc/passwd doesn't work. This is simply because CTextureCache::IsCachedImage marks "full paths" as a cached image and /etc/passwd eventually fails image decoding.

## How Has This Been Tested?
Ensured that the provided wget no longer downloads /etc/passwd. Browsed around the site at :8080 without issue.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/network/httprequesthandler/HTTPFileHandler.cpp
+++ b/xbmc/network/httprequesthandler/HTTPFileHandler.cpp
@@ -61,6 +61,18 @@ void CHTTPFileHandler::SetFile(const std::string& file, int responseStatus)
   if (m_url.empty())
     return;
 
+  if (file.find("image://") == 0)
+  {
+    const CURL pathToUrl(file);
+    if (pathToUrl.GetHostName().find("../") != std::string::npos ||
+        pathToUrl.GetHostName().find("..\\") != std::string::npos)
+    {
+        m_response.type = HTTPError;
+        m_response.status = MHD_HTTP_NOT_FOUND;
+        return;
+    }
+  }
+
   // translate the response status into the response type
   if (m_response.status == MHD_HTTP_OK)
     m_response.type = HTTPFileDownload;
--- a/xbmc/network/httprequesthandler/HTTPImageHandler.cpp
+++ b/xbmc/network/httprequesthandler/HTTPImageHandler.cpp
@@ -36,12 +36,6 @@ CHTTPImageHandler::CHTTPImageHandler(const HTTPRequest &request)
 
     XFILE::CImageFile imageFile;
     const CURL pathToUrl(file);
-    if (pathToUrl.GetHostName().find("..") != std::string::npos)
-    {
-        responseStatus = MHD_HTTP_NOT_FOUND;
-        return;
-    }
-
     if (imageFile.Exists(pathToUrl))
     {
       responseStatus = MHD_HTTP_OK;
--- a/xbmc/network/httprequesthandler/HTTPImageHandler.cpp
+++ b/xbmc/network/httprequesthandler/HTTPImageHandler.cpp
@@ -36,6 +36,12 @@ CHTTPImageHandler::CHTTPImageHandler(const HTTPRequest &request)
 
     XFILE::CImageFile imageFile;
     const CURL pathToUrl(file);
+    if (pathToUrl.GetHostName().find("..") != std::string::npos)
+    {
+        responseStatus = MHD_HTTP_NOT_FOUND;
+        return;
+    }
+
     if (imageFile.Exists(pathToUrl))
     {
       responseStatus = MHD_HTTP_OK;


Jalle19 (commented):
Maybe add a link to the CVE so people don't accidentally remove this in the future?

Tolriq:
@jacob-baines Since your at it 

https://github.com/xbmc/xbmc/blob/master/xbmc/network/httprequesthandler/HTTPVfsHandler.cpp#L43

Same issue by using image:// combined with /vfs/ :)

About the fix technically a valid file can contain ".." but security > small issue that just need to be documented, and of course this still leave access to any image on filesystem but not sure there's a lot of possible attack vector with that.

Pinging @Montellese as the proper fix should probably be done at image:// vfs handler.

Jalle19:
@Tolriq a better solution would be to check for `../` since filenames cannot contain forward slashes.

jacob-baines:
For anyone interested, this request will go through HTTPVfsHandler:
```sh
wget http://192.168.1.40:8080/vfs/image%3a%2f%2f%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%32%65%25%32%65%25%32%66%25%36%35%25%37%34%25%36%33%25%32%66%25%37%30%25%36%31%25%37%33%25%37%33%25%37%37%25%36%34 
```

@Tolriq To address the issue in /vfs/ I've moved the check down into HTTPFileHandler. This should also prevent any other HTTP*Handler that I haven't accounted for.

@Jalle19 I changed the check to look for "../" or "..\\". I believe Kodi supports Windows but I didn't explicitly test it either.

MrMC:
This and more are already fixed by the rejected zebralist check PR https://github.com/xbmc/xbmc/pull/10137. 

MrMC is running it, as is SPMC.

https://github.com/MrMC/mrmc/commit/955091c008e178e48aea737c891566c4e7e57359
https://github.com/MrMC/mrmc/commit/8d0a0b61a996c2f78f2b164fae60f1bb18eaaa3a

https://github.com/koying/SPMC/commit/6ea119023b402186cc8a5f75db89f4d1d22d261e

Tolriq:
@MrMC your patch disable some vital things from Kodi and SPMC implements only a small part since well SPMC is obviously not TARGET_DARWIN

There was already hundreds of discussion and I won't get back to them.

But if anyone take that road, do not forget access to shared sources, mandatory for Kore download / Chorus2 streaming and many other apps.

https://github.com/xbmc/xbmc/pull/10137 Have many details about impacts.

MrMC:
Leave the gaping holes open, totally up to kodi devs :) Got to love the whole head-in-sand approach. There is zero reason to have wide open external access to secure credentials that are stored in plain text.

Trapping just '..' is a epic fail because some assess to valid resources is by a '..' re-direct. One must check the actual resolved path as there are many ways to trick simple text parsers.

Tolriq:
For the last time as before : Yes you are 100% right and yes your solution is the way to go.

But no blocking everything including major Kodi features is not the solution, your hammer needs to be reduced to keep actual used functions. That's all I say since day 1 ;)

Jalle19:
If checking for `../` is not good enough then we should check the resolved absolute path to make sure it begins with the path to Kodi's directory. Then blacklist some files like #10137 already does. Shouldn't that prevent this?

Tolriq:
@Jalle19 Yes but only checking Kodi's directory is not enough.

Kodi sources can have the attribute allowSharing that allows access to their content via WebServer.

This is used and necessary for Kore download function, Chorus2 remote playing and any application that streams from Kodi.

https://github.com/xbmc/xbmc/pull/2764 explains how it works, and call to CFileUtils::RemoteAccessAllowed needs to be done to allow access when it returns true.

MrMC:
I would not describe gaping security holes as major Kodi features :) They are exploits that have been abused by other apps just because no one cared enough to create a proper API to provide those features in a safe and secure way. Lock them down and you can be sure some one will start to care.

Tolriq:
I understand better why some call you stubborn ;)
I have fixed other security issues that did triggers tons of support for apps because there was no other way.

The issue that your hammer does is known so can be anticipated and properly integrated.

And seeing how most does not care about security, I would not expect they care about users whining because of lost features.

Anyway I have given previous history and explanations to @Jalle19  and @jacob-baines that maybe did not know. I now disengage from this PR where I just came to help fixing another forgotten issue.

MrMC:
Sometimes, stubborn is what it takes to get the job done :)

Tolriq:
Job was started in 2013 with https://github.com/xbmc/xbmc/pull/2764 ;) 3 years earlier.

At a time where jmarshallnz was still around and contributing was less painful even if already complicated.

IMO solution would have been to update this function to add blacklist and call it earlier in the stack like you do, but it seems inventing a new wheel was chosen.

/me unsubscribe to be sure to disengage :p

wsnipex:
I have to agree with @MrMC here. We must take security issues seriously and the zebra list PR was a step in the right direction, albeit still incomplete.
just checking for "../" is not enough.

fritsch:
+1

Am 16.03.2017 4:26 nachm. schrieb "Wolfgang Schupp" <
notifications@github.com>:

garbear:
@jacob-baines I've updated the solution from the zebra list PR: https://github.com/garbear/xbmc/commit/CVE-2017-5982

The only remaining thing to address is the code quality in CFileUtils::ZebraListAccessCheck(). If you use this as a base, your solution should be much more acceptable.

tavoc:
just my two cents.
As far as I have seen, you are looking for "..". 

From an attacker view I can encode .. like U+002E or some other UTF style etc.
I had not a deep dive into this code, but are you "normalizing" all the user input to one specific encoding? Because only after this step a search for ".." makes sense.

jacob-baines:
This pull request has been closed.

PR #11852, opened 2017-03-15
Use invoke-id instead addon-id for selecting the monitor for abortrequest

peak3d:
Use invoker id instead addon-id for signaling python monitor

## Description
For addons wich come together with a service-addon the services are stopped if the main addon script is cancelled e.g. by using [ESC] or double [Baclspace].
Cancelling is done through Monitor::abortrequest signal.

Reason is that the Monitor to signal is selected by addon-id (plugin.video.netflix) which is shared for normal addon and service addon.
This PR selects the Monitor to signal by the invoker id wich is different for service-addon and normal addon.

## Motivation and Context
netflix addon service gets killed on ESC or fast double backspace

## How Has This Been Tested?
[Win32 / Linux / Mac] 
- install netflix addon
- start netflix addon
- press ESC
- look in the kodi log for canceled services.

## Types of change

### DIFF
--- a/xbmc/interfaces/legacy/LanguageHook.h
+++ b/xbmc/interfaces/legacy/LanguageHook.h
@@ -111,6 +111,7 @@ namespace XBMCAddon
      */
     virtual String GetAddonId() { return emptyString; }
     virtual String GetAddonVersion() { return emptyString; }
+    virtual long GetInvokerId() { return -1; }
 
     virtual void RegisterPlayerCallback(IPlayerCallback* player) = 0;
     virtual void UnregisterPlayerCallback(IPlayerCallback* player) = 0;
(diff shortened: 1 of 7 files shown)


Rechi (approved):

Rechi:
@tamland can you please review

tamland:
> Reason is that Addon and service share the same addon-id and the same Monitor.

Isn't this the real issue? Surely Monitor should be per interpreter, and not invoked base on addon id in the first place? Like the abortRequested module property

Rechi:
Issue is that aborting the main addon triggers https://github.com/xbmc/xbmc/blob/65fc516/xbmc/interfaces/python/XBPython.cpp#L654-L655.
In my opinion those lines should only be executed if an addon gets deactivated, uninstalled or maybe updated as the service needs to stop in these cases.
I cannot think of any reason that the main addon must be signalled it got aborted after a user presses escape while addon is running.

peak3d:
jenkins build this please

peak3d:
jenkins build this please

MartijnKaijser:
@peak3d could you make a backport PR?

PR #11856, opened 2017-03-15
AMLCodec: do not set blackout policy

kszaq:
Hardware decoders that use CMA decide on whether to release memory depending
on blackout policy. If we change the policy when resetting decoder CMA is
sometimes not released which leads to OOM. Although it's a kernel bug it would
be nice if Kodi did not trigger it.

## How Has This Been Tested?
Tested in my community S805 and S905(X) builds with Kodi Krypton.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
@@ -1701,11 +1701,6 @@ void CAMLCodec::Reset()
 
   SetPollDevice(-1);
 
-  // set the system blackout_policy to leave the last frame showing
-  int blackout_policy;
-  SysfsUtils::GetInt("/sys/class/video/blackout_policy", blackout_policy);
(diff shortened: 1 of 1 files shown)


peak3d:
For what reason is the blackout_policy switched in reset in current code?

kszaq:
As far as I can tell it's been there since the very first version of AMLCodec. It's supposed to not make screen black when we reset decoder but keep displaying the last frame instead.

peak3d:
Does this PR change any existing visible behaviour?
Seek forward / backward always forces a reset, is there now a black screen during the time we wait for the seek to finish?

kszaq:
It depends on system setting. If `blackout_policy` is `1`, the screen is black during seek. Otherwise the last frame is shown.

peak3d:
Whats about setting the policy on ctor / dtor instead every time on reset?

kszaq:
After consideration I'm going to drop this in favour of setting project-specific policy.

Thank you for your review @peak3d.

PR #11857, opened 2017-03-16
[win] use users profile folder as home folder

Rechi:
## Description
Use the users profile folder (normaly `C:\Users\<username>`) as home folder in the file browser.
Currently `special://home` is used which maps to `C:\Users\<username>\AppData\Roaming\Kodi`.

## Motivation and Context
Current mapping isn't useful if you want to add a source which is located in the profile folder.
Other platforms already behave like it is with this change on windows.

## How Has This Been Tested?
runtime tested

## Types of change

## Checklist:

### DIFF
--- a/xbmc/storage/windows/Win32StorageProvider.cpp
+++ b/xbmc/storage/windows/Win32StorageProvider.cpp
@@ -21,9 +21,12 @@
 #include "WIN32Util.h"
 #include "guilib/LocalizeStrings.h"
 #include "filesystem/SpecialProtocol.h"
+#include "platform/win32/CharsetConverter.h"
 #include "storage/MediaManager.h"
 #include "utils/JobManager.h"
 #include "utils/log.h"
+
+#include <ShlObj.h>
 
 bool CWin32StorageProvider::xbevent = false;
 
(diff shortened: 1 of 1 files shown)


xbmc/storage/windows/Win32StorageProvider.cpp
```diff
@@ -49,8 +52,13 @@ void CWin32StorageProvider::Initialize()
 
 void CWin32StorageProvider::GetLocalDrives(VECSOURCES &localDrives)
 {
+  using namespace KODI::PLATFORM::WINDOWS;
   CMediaSource share;
-  share.strPath = CSpecialProtocol::TranslatePath("special://home");
+  wchar_t profilePath[MAX_PATH];
+  if (SUCCEEDED(SHGetFolderPath(NULL, CSIDL_PROFILE, NULL, 0, profilePath)))
```
Paxxi: nitpick, use nullptr :) not necessary to change but remember for the future

xbmc/storage/windows/Win32StorageProvider.cpp
```diff
@@ -49,8 +52,13 @@ void CWin32StorageProvider::Initialize()
...
+    share.strPath = FromW(profilePath);
+  else
+    share.strPath = CSpecialProtocol::TranslatePath("special://home");
```
Paxxi: This could use the env variable `USER_PROFILE` as a fallback as that maps to the correct path and then if that doesn't exist fall back to using TranslatePath

Rechi:
@Paxxi updated

Paxxi (approved):

Rechi:
jenkins build and merge

Rechi:
only packing setup on windows failed, building kodi executable was successful.

PR #11836, opened 2017-03-10
[bluray] Improve Blu-ray playback

ace20022:

PR #11862, opened 2017-03-17
[bluray] allow do build without bluray support

Rechi:
## Description
libbluray is an optional dependency, so it should be possible to compile kodi without having libbluray.

## Motivation and Context
Tried to compile for windows x64 which has libbluray not packaged yet. It was working before #11836.

## How Has This Been Tested?
compiled without having libbluray

## Types of change

## Checklist:

@ace20022 is this change ok for you?


ace20022:
I have not changed anything in that regard with #11836

Rechi:
You include `DVDInputStreams/DVDInputStreamBluray.h` which should only be included if `HAVE_LIBBLURAY` is defined and use some values of an enum which is defined in `libbluray/bluray.h`.

jenkins build and merge

ace20022:
Correct, sorry

PR #11863, opened 2017-03-17
[PVR] GUI Actions: Execute recordings actions async

ksooo:
My recordings are located on a NAS which automatically spins down disks and falls asleep if NAS gets not accessed for a certain period of time and it automatically gets awakened on demand. The latter can take up to 20 secs.

Visual effect in Kodi is for example, a 20 secs blocking user interface after selecting "delete" on a recording in Kodi's PVR Recordings window.

This PR fixes this gui blocking by executing the functions in question asynchronously (and by giving visual feedback using the Kodi busy dialog).

This has been runtime tested on macOS and Linux on latest Kodi master.

@Jalle19 for review?

### DIFF
--- a/xbmc/pvr/PVRGUIActions.cpp
+++ b/xbmc/pvr/PVRGUIActions.cpp
@@ -19,6 +19,7 @@
  */
 
 #include "Application.h"
+#include "dialogs/GUIDialogBusy.h"
 #include "dialogs/GUIDialogKaiToast.h"
 #include "dialogs/GUIDialogNumeric.h"
 #include "dialogs/GUIDialogOK.h"
@@ -51,6 +52,7 @@
 #include "ServiceBroker.h"
 #include "settings/MediaSettings.h"
 #include "settings/Settings.h"
+#include "threads/Thread.h"
 #include "utils/StringUtils.h"
 #include "utils/URIUtils.h"
 #include "utils/log.h"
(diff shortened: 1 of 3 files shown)


xbmc/pvr/PVRGUIActions.cpp
```diff
@@ -710,16 +756,21 @@ namespace PVR
...
     {
       CGUIDialogOK::ShowAndGetInput(CVariant{257}, CVariant{19111}); // "Error", "PVR backend error. Check the log for more information about this message."
       return false;
     }
 
-    g_PVRManager.TriggerRecordingsUpdate();
     return true;
   }
 
+  class AsyncDeleteRecording : public AsyncRecordingAction
```
Jalle19: Why not define this class further up with the others?

xbmc/pvr/PVRGUIActions.cpp
```diff
@@ -750,15 +800,20 @@ namespace PVR
                                             CVariant{item->GetLabel()});
   }
 
+  class AsyncEmptyRecordingsTrash : public AsyncRecordingAction
```
Jalle19: Same here

xbmc/pvr/PVRGUIActions.cpp
```diff
@@ -768,19 +823,23 @@ namespace PVR
                                             CVariant{19293}); // "Remove all deleted recordings from trash? This operation cannot be reverted."
   }
 
+  class AsyncUndeleteRecording : public AsyncRecordingAction
```
Jalle19: Same here

xbmc/pvr/PVRGUIActions.cpp
```diff
@@ -700,6 +702,50 @@ namespace PVR
                                             CVariant{timer->Title()});
   }
 
+  class AsyncRecordingAction : private IRunnable
+  {
+  public:
+    bool Execute(const CFileItemPtr &item = CFileItemPtr());
```
Jalle19: A default value seems unnecessary?

Paxxi: Don't take a shared_ptr by reference, should be by value to ensure ref count is correctly incremented

ksooo: Ha? We're doing this all over the place.

xbmc/pvr/PVRGUIActions.cpp
ksooo: Intention was to have the implementation as close to the using code as possible. But no strong feelings about this here.

xbmc/pvr/PVRGUIActions.cpp
ksooo: > A default value seems unnecessary?

https://github.com/xbmc/xbmc/pull/11863/files#diff-e47a4eb266b6b884e4b669666a1cc89cR814

Alternatives would be Execute(CFileItemPtr()) or overloaded Execute().

Paxxi: Yeah, it's a broken pattern because const ref is so ingrained people just continued with it. It makes no sense to use it on shared_ptr and actually can make things worse as it gives the impression that things are fine but the ref count isn't incremented so things can blow up unexpectedly.

ksooo: Not sure whether is more a theoretical problem or something for people not knowing about the "unsafeness" here. But I see the performance impact and to some extend this makes it senseful to use refs.

Paxxi: The performance impact is allocation of a pointer and atomic increment so it's not really a worry. 

If you prefer const ref I suggest passing the CFileItem as const ref instead of the shared_ptr. It avoids the extra allocation and looks saner to the reader. 

Anyway, I'm not in the habit of blocking pull requests so this is merely advice, you're free to pick whatever method you prefer as all issues are as you said mostly theoretical.

ksooo:
@Jalle19 I've addressed your comments.

ksooo:
jenkins build this please

ksooo:
jenkins build this please

ksooo:
jenkins build this please

PR #11869, opened 2017-03-19
[guilib] add template for GetWindow() to enclose type cast

xhaggi:
This PR will introduce a template for `CGUIWindowManager::GetWindow(int id)` to get rid of all the different type casts we currently use and to save some dupe code if using C++ style type casts.

It also adds a convenient method `CGUIWindowManager::GetDialog(int id)` which check if the window for the supplied ID is a dialog and directly return a `CGUIDialog*`. So no need to cast `CGUIWindow*` to `CGUIDialog*` in that cases, just use the new method.

I've done all the work to use the new generic `GetWindow<T>()` and `GetDialog()` where possible.

Runtime tested on MacOS

@Paxxi mind taking a look.

### DIFF
--- a/xbmc/addons/RepositoryUpdater.cpp
+++ b/xbmc/addons/RepositoryUpdater.cpp
@@ -113,7 +113,7 @@ bool CRepositoryUpdater::CheckForUpdates(bool showProgress)
 
 static void SetProgressIndicator(CRepositoryUpdateJob* job)
 {
-  auto* dialog = static_cast<CGUIDialogExtendedProgressBar*>(g_windowManager.GetWindow(WINDOW_DIALOG_EXT_PROGRESS));
+  auto dialog = g_windowManager.GetWindow<CGUIDialogExtendedProgressBar>();
   if (dialog)
     job->SetProgressIndicators(dialog->GetHandle(g_localizeStrings.Get(24092)), nullptr);
 }
--- a/xbmc/guilib/GUIWindowManager.cpp
+++ b/xbmc/guilib/GUIWindowManager.cpp
@@ -1196,6 +1196,14 @@ CGUIWindow* CGUIWindowManager::GetWindow(std::type_index type) const
   return NULL;
 }
 
+CGUIDialog* CGUIWindowManager::GetDialog(int id) const
+{
+  CGUIWindow *window = GetWindow(id);
+  if (window->IsDialog())
+    return dynamic_cast<CGUIDialog*>(window);
+  return nullptr;
+}
+
 CGUIWindow* CGUIWindowManager::GetWindow(int id) const
 {
   CGUIWindow *window;
--- a/xbmc/guilib/GUIWindowManager.cpp
+++ b/xbmc/guilib/GUIWindowManager.cpp
@@ -572,7 +572,9 @@ void CGUIWindowManager::Add(CGUIWindow* pWindow)
                           "to the window manager", *idIt);
       return;
     }
+
     m_mapWindows.insert(std::pair<int, CGUIWindow *>(*idIt, pWindow));
+    m_mapWindowTypes.insert(std::make_pair(std::type_index(typeid(*pWindow)), pWindow));
   }
 }
 
@@ -602,15 +604,17 @@ void CGUIWindowManager::Remove(int id)
   WindowMap::iterator it = m_mapWindows.find(id);
   if (it != m_mapWindows.end())
   {
+    CGUIWindow *window = it->second;
     for(std::vector<CGUIWindow*>::iterator it2 = m_activeDialogs.begin(); it2 != m_activeDialogs.end();)
     {
-      if(*it2 == it->second)
+      if(*it2 == window)
         it2 = m_activeDialogs.erase(it2);
       else
         ++it2;
     }
 
     m_mapWindows.erase(it);
+    m_mapWindowTypes.erase(std::type_index(typeid(*window)));
   }
   else
   {
@@ -1179,6 +1183,17 @@ void CGUIWindowManager::FrameMove()
     (*it)->FrameMove();
 
   g_infoManager.UpdateAVInfo();
+}
+
+CGUIWindow* CGUIWindowManager::GetWindow(std::type_index type) const
+{
+  CSingleLock lock(g_graphicsContext);
+
+  auto it = m_mapWindowTypes.find(type);
+  if (it != m_mapWindowTypes.end())
+    return (*it).second;
+
+  return NULL;
 }
 
 CGUIWindow* CGUIWindowManager::GetWindow(int id) const
--- a/xbmc/guilib/GUIWindowManager.cpp
+++ b/xbmc/guilib/GUIWindowManager.cpp
@@ -158,7 +158,7 @@ using namespace KODI::MESSAGING;
 
 CGUIWindowManager::CGUIWindowManager(void)
 {
-  m_pCallback = NULL;
+  m_pCallback = nullptr;
   m_iNested = 0;
   m_initialized = false;
 }
@@ -458,17 +458,17 @@ bool CGUIWindowManager::SendMessage(CGUIMessage& message)
   if (message.GetMessage()==GUI_MSG_NOTIFY_ALL)
   {
     CSingleLock lock(g_graphicsContext);
-    for (rDialog it = m_activeDialogs.rbegin(); it != m_activeDialogs.rend(); ++it)
+
+    for (auto it = m_activeDialogs.rbegin(); it != m_activeDialogs.rend(); ++it)
     {
-      CGUIWindow *dialog = *it;
-      dialog->OnMessage(message);
+      (*it)->OnMessage(message);
     }
 
-    for (WindowMap::iterator it = m_mapWindows.begin(); it != m_mapWindows.end(); ++it)
+    for (const auto& entry : m_mapWindows)
     {
-      CGUIWindow *pWindow = (*it).second;
-      pWindow->OnMessage(message);
+      entry.second->OnMessage(message);
     }
+
     return true;
   }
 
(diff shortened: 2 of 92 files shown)


ksooo:
This is a nice one. Thanks.

xhaggi:
@Paxxi I've added a new commit (WIP) where I introduce a new map to receive the window id by type. With this new map and method it should be possible to get rid of the window id parameter. It would be nice if you could look at it and let me know if we're on the right direction.

xbmc/guilib/GUIWindowManager.h
```diff
@@ -227,6 +231,7 @@ class CGUIWindowManager : public KODI::MESSAGING::IMessageTarget
 
   typedef std::map<int, CGUIWindow *> WindowMap;
   WindowMap m_mapWindows;
+  std::unordered_map<std::type_index, CGUIWindow*> m_mapWindowTypes;
```
xhaggi: I use an unordered_map here to optimize map lookups. We don't need an ordered map and only insert/remove items on start and shutdown of the app.

Four years ago a cache `CGUIWindowManagerIdCache` was introduced to optimize the map lookup of std::map in `GetWindow(int id)`. Not sure if he checked the performance of an unordered_map.

https://github.com/xbmc/xbmc/commit/98b4f9c083fc7c6ec1549228ac59e9d3b6ad9bbc

xhaggi:
I added another commit with various refactorings / cleanups applied to GUIWindowManager

xbmc/addons/RepositoryUpdater.cpp
```diff
@@ -113,7 +113,7 @@ bool CRepositoryUpdater::CheckForUpdates(bool showProgress)
 
 static void SetProgressIndicator(CRepositoryUpdateJob* job)
 {
-  auto* dialog = static_cast<CGUIDialogExtendedProgressBar*>(g_windowManager.GetWindow(WINDOW_DIALOG_EXT_PROGRESS));
+  auto* dialog = g_windowManager.GetWindow<CGUIDialogExtendedProgressBar>();
```
Paxxi: no need for `auto*` plain `auto` is enough but it already was there so

xbmc/guilib/GUIWindowManager.cpp
```diff
@@ -458,17 +458,17 @@ bool CGUIWindowManager::SendMessage(CGUIMessage& message)
   if (message.GetMessage()==GUI_MSG_NOTIFY_ALL)
   {
     CSingleLock lock(g_graphicsContext);
-    for (rDialog it = m_activeDialogs.rbegin(); it != m_activeDialogs.rend(); ++it)
+
+    for (const auto& window : m_activeDialogs)
```
Paxxi: This changes behavior and may break in subtle ways, I'd leave it as is

xbmc/guilib/GUIWindowManager.cpp
```diff
@@ -561,18 +561,19 @@ void CGUIWindowManager::Add(CGUIWindow* pWindow)
...
   // push back all the windows if there are more than one covered by this class
   CSingleLock lock(g_graphicsContext);
-  m_idCache.Invalidate();
-  const std::vector<int>& idRange = pWindow->GetIDRange();
-  for (std::vector<int>::const_iterator idIt = idRange.begin(); idIt != idRange.end() ; ++idIt)
+
+  for (int id : pWindow->GetIDRange())
   {
-    WindowMap::iterator it = m_mapWindows.find(*idIt);
+    const auto& it = m_mapWindows.find(id);
```
Paxxi: no need for `const auto&` plain `auto` is all you need for the iterator returned

xbmc/guilib/GUIWindowManager.cpp
```diff
@@ -561,18 +561,19 @@ void CGUIWindowManager::Add(CGUIWindow* pWindow)
...
     if (it != m_mapWindows.end())
     {
       CLog::Log(LOGERROR, "Error, trying to add a second window with id %u "
-                          "to the window manager", *idIt);
+                          "to the window manager", id);
       return;
     }
-    m_mapWindows.insert(std::pair<int, CGUIWindow *>(*idIt, pWindow));
+
+    m_mapWindows.insert(std::pair<int, CGUIWindow*>(id, pWindow));
```
Paxxi: use `std::make_pair(id, pWindow)` instead of `std::pair<int, CGUIWindow*>(id, pWindow)`
and same goes for next line

xbmc/guilib/GUIWindowManager.cpp
```diff
@@ -598,18 +599,16 @@ void CGUIWindowManager::RegisterDialog(CGUIWindow* dialog)
 void CGUIWindowManager::Remove(int id)
 {
   CSingleLock lock(g_graphicsContext);
-  m_idCache.Invalidate();
-  WindowMap::iterator it = m_mapWindows.find(id);
+
+  const auto& it = m_mapWindows.find(id);
```
Paxxi: plain `auto` is enough

xbmc/guilib/GUIWindowManager.cpp
```diff
@@ -1174,32 +1173,45 @@ void CGUIWindowManager::FrameMove()
...
+    return nullptr;
 
   CSingleLock lock(g_graphicsContext);
 
-  window = m_idCache.Get(id);
-  if (window)
-    return window;
-
-  WindowMap::const_iterator it = m_mapWindows.find(id);
+  const auto& it = m_mapWindows.find(id);
```
Paxxi: plain `auto` is enough

Paxxi (commented):
I would go full auto wherever GetWindow or GetDialog is used to avoid having to repeat the type twice but not important enough for hundreds of changes that I require it 😄 

I don't think the use of const auto& in most for loops are actually of any use since the vectors contain pointers a plain auto should be perfectly fine to use.

Also for someone claiming not to be a c++ dev you sure know some stuff, I did not know about std::type_index 😄 

Overall really nice work, 👍

xbmc/addons/RepositoryUpdater.cpp
xhaggi: not added by me ;) but will change it.

xbmc/guilib/GUIWindowManager.cpp
xhaggi: overlooked that a reverse iterator is used here.

xbmc/guilib/GUIWindowManager.cpp
xhaggi: c/p ;) but you're right

xbmc/guilib/GUIWindowManager.cpp
xhaggi: after a second look: I don't know why we force reverse order to call OnMessage() .. it doesn't matter if you ask me. 2  lines below we iterate over all windows without a strict order ;)

xhaggi:
@Paxxi I've addressed your comments and squashed some commits.

xhaggi:
jenkins build this please

xbmc/guilib/GUIWindowManager.h
```diff
@@ -31,6 +31,8 @@
 #include <list>
 #include <utility>
 #include <vector>
+#include <typeindex>
```
Montellese: Can you add the includes in alphabetical order?

xhaggi: Sure

xhaggi:
@Montellese the includes are now in alphabetic order.

Paxxi:
We had a brief discussion about this on slack and it's all good with this approach

xhaggi:
Small update: I've dropped the is_pointer check from the template and instead I've added a check if the supplied type derives from `CGUIWindow`. `std::is_base_of` will guard against the pointer issue too. I updated the code doc for the template too.

xhaggi:
jenkins build and merge

Montellese:
I actually wanted to propose `std::is_base_of` but I figured it would get too complicated but great that you use it. The only thing I don't like about these type traits is that they produce error messages that nobody understands. That's why I usually combine them with a `static_assert()` like
```
template<typename T, typename std::enable_if<std::is_base_of<CGUIWindow,T>::value>::type* = nullptr>
T* GetWindow() const
{
  static_assert(std::is_base_of<CGUIWindow,T>::value, "Type not derived from CGUIWindow in CGUIWindowManager::GetWindow()");
  return dynamic_cast<T *>(GetWindow(std::type_index(typeid(T))));
};
```
That way the compiler will create an error with the message "Type not derived from CGUIWindow in CGUIWindowManager::GetWindow()" instead of a mumbo jumbo of template parameters and instantiations across multiple lines.

xhaggi:
@Montellese meh too late .. but I will do another PR with your suggestion.

xbmc/guilib/GUIWindowManager.cpp
```diff
@@ -1196,6 +1196,14 @@ CGUIWindow* CGUIWindowManager::GetWindow(std::type_index type) const
   return NULL;
 }
 
+CGUIDialog* CGUIWindowManager::GetDialog(int id) const
+{
+  CGUIWindow *window = GetWindow(id);
+  if (window->IsDialog())
```
popcornmix: I'm hitting a segfault here with window=null. Changing to 
```if (window && window->IsDialog())```
avoids the crash. See: http://forum.kodi.tv/showthread.php?tid=310869

Comment (xhaggi) [inline xbmc/guilib/GUIWindowManager.cpp]: You're right missed that.

**Comment by ksooo** [comment]
@xhaggi this unfortunately only works if there is at most one instance of the given window type, which is not always the case for some pvr windows. => https://github.com/xbmc/xbmc/blob/master/xbmc/guilib/GUIWindowManager.cpp#L260 For those windows, this PR leads to some serious regressions, because now a "random" window gets returned instead of the requested tv or radio instance of the window type.

As I never liked the strange reusage of the same window type for tv and radio windows (this violates OOP principles imo) I will fix this for the respective PVR windows.

But I want to give a heads-up here that the new template function must only be used if one can be sure that at a given point in time there is at least one instance of the requested type. If there are other reused window types, this PR might have also introduced regressions.

**Comment by xhaggi** [comment]
Thanks for pointing it out. Never thought of that while @Paxxi and me talked about a way to get rid of the window id in the template. The first approach was a template where you need to pass in the window id and the generic return type.

```
template<typename T, typename std::enable_if<std::is_base_of<CGUIWindow,T>::value>::type* = nullptr>
T* GetWindow(int id) const { return dynamic_cast<T *>(GetWindow(id)); };
```

IMO we should introduce this template too and use it where we can't use the other, or don't we?

xhaggi:
Or we drop the current template and pass in the id everywhere to be safe.

ksooo:
I don't think it is a good idea to forbid creation of multiple instances of a window class. This introduces less flexibility just to make this single template function work. But if we allow multiple instances, the template you introduced here is really dangerous to use. 

> Or we drop the current template and pass in the id everywhere to be safe.

 Yes, this is what needs to be done.

ksooo:
Unfortunately, this is some real monkey work, going to every change you made in this PR, reintroducing the id you removed before. :-/

xhaggi:
Will create a PR for it tomorrow if I found some spare time.

ksooo:
> Will create a PR for it tomorrow if I found some spare time.

 Thanks. Much appreciated.

PR #11873, opened 2017-03-19
[Fix]Default music artist separators remove colon and "|"  (ticket 17401)

DaveTBlake:
Simple modification that fixes Ticket http://trac.kodi.tv/ticket/17401

Both ":" and "|" appear in artist names, so having them as defualt can cause artists to be misidentified. Users can always configure these to be separators if they have used them in their tagging (and do not have any in the artist names)

A trivial change backported as #11874 .

### DIFF
--- a/xbmc/settings/AdvancedSettings.cpp
+++ b/xbmc/settings/AdvancedSettings.cpp
@@ -281,7 +281,7 @@ void CAdvancedSettings::Initialize()
   m_strMusicLibraryAlbumFormat = "";
   m_prioritiseAPEv2tags = false;
   m_musicItemSeparator = " / ";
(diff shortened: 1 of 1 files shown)


DaveTBlake:
@MilhouseVH think you were party to discussion of this on LE

jenkins build this please

MilhouseVH:
Thanks, I'll include this in the next build.

If I had any artists using the pipe separator, I'd just re-tag with ` / `. Same for colons. :)

HiassofT:
Looks good to me, and thanks a lot for your great support!

DaveTBlake:
@HiassofT anything music related just let me know and I'll see if I can get it sorted. 
This is trivial and with the RC2 deadline imminent my finger is poised over merge :)

PR #11874, opened 2017-03-19
[Fix][Backport]Default music artist separators remove colon and "|" (ticket 17401)

DaveTBlake:
Simple modification that fixes Ticket http://trac.kodi.tv/ticket/17401

Both ":" and "|" appear in artist names, so having them as defualt can cause artists to be misidentified. Users can always configure these to be separators if they have used them in their tagging (and do not have any in the artist names)

Backport of #11873 .


HiassofT (approved):
Tested on LibreELEC 8.0 branch with kodi 17.1rc1 and works fine. Thanks a lot!

PR #11877, opened 2017-03-19
[PVR][Esturay] Channelgroup manager dialog: Add support for radio channel groups

ksooo:
This PR adds support for radio channel groups to the PVR Channelgroups manager dialog. Stumbled over this missing feature today - could not believe that this was missing! ;-)

screenshot001
screenshot000

The changes were runtime tested an macOS on latest Kodi master.

@Jalle19 @xhaggi for core code review?
@phil65 @ronie for review of the skin changes?

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -9957,7 +9957,19 @@ msgctxt "#19172"
 msgid "Instant recording duration"
 msgstr ""
 
-#empty strings from id 19173 to 19174
+#. generic label for 'Live TV channel groups' used in different places
+#: addons/skin.estuary/xml/DialogPVRGroupManager.xml
+#: addons/skin.estuary/xml/Variables.xml
+msgctxt "#19173"
+msgid "TV channel groups"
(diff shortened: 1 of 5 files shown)


ronie (commented):
skin changes look ok to me

xhaggi:
Looks good. Never noticed that this is missing too ;)

stefansaraev:
nobody noticed, because nobody really uses it. nobody uses it, because for most users it is useless. and it will be useless as long it does not tell backends "hey, we've re-ordered some channels, change your channel order so next time I do pvr db reset we have the same order client-side". sorta ;)

if you ask me, this feature should be competely ripped out

ksooo:
@stefansaraev thanks for the feedback, although it is not really helpful and motivating.

jenkins build this please

ksooo:
jenkins build this please

xhaggi:
You're right but in case your backend does not support groups your are happy to do it here. Reset the PVR database is not an option for those guys. Along with this you can hide channel groups especially the internal "All channels" group. This option is only available in Kodi.

PR #11880, opened 2017-03-20
fixed: install into system-expected python library path

notspiff:
dist-packages is a debian-ism and does not play well on distros with other origins. ref http://forum.kodi.tv/showthread.php?tid=309969


wsnipex:
thanks!

PR #11881, opened 2017-03-20
depends: remove libsdl_image and libsdl2

stefansaraev:
libsdl_image is an orphan dependency, not used for very long
libsdl2 seems unused

## How Has This Been Tested?

not tested at all.

## Types of change

## Checklist:


xbmc/video/Teletext.cpp
```diff
@@ -32,11 +32,6 @@
 #include "filesystem/SpecialProtocol.h"
 #include "guilib/GraphicContext.h"
 
-#if HAVE_SDL_VERSION == 1
-#include <SDL/SDL_stdinc.h>
-#elif HAVE_SDL_VERSION == 2
-#include <SDL2/SDL_stdinc.h>
-#else
```
Rechi: maybe keep the SDL version and only remove SDL2 from here

FernetMenta: why should we keep SDL related code?

Rechi: no need to keep it but sdl doesn't get removed, only sdl2 will be removed by this PR

stefansaraev: btw, here https://www.libsdl.org/release/SDL-1.2.15/include/SDL_stdinc.h SDL_memcpy4 and SDL_memset4 are defined exacly the same. so I too dont see reason to keep using SDL_stdinc here. I've added a commit that does. can remove it later if you want me to

Rechi: ok if it is the exact same remove all sdl includes used only for `SDL_memcpy4`.

FernetMenta: great. 
maybe we can drop SDL_memcpy/memset completely on a next step. I would guess that replacing it is not too hard.

Rechi:
maybe https://github.com/Rechi/xbmc/commit/a795424 also isn't needed anymore

stefansaraev:
yep. added. thanks

wsnipex:
looks good to me. I wonder if OSX will ever get rid of it.

Rechi:
jenkins build this please

Rechi:
jenkins build and merge

PR #11885, opened 2017-03-21
FIX: [python] no abort + droid libs

koying:
Couple of python patches:

1) Do not allow python to abort Kodi
2) Re-apply 2,6 patch regarding libraries on droid

Not sure if 2) is still needed specifically with pillow, but definitely needed in general


wsnipex:
ad 1.) ok :) 
add 2) I wonder as well. Do you know of any current scenario where this is needed?

koying:
Re 2) It will be needed in the future for any PIL like python module
I backported your 2,7 bump to spmc16 and wondered why PIL stopped working again ;)

Even if not needed today, it would potentially prevent mucho headaches the next time we add a python module

wsnipex:
ok, fine for me, but should be tested. @MartijnKaijser ping
jenkins build this please

MartijnKaijser:
tested. Rest will be tested during nightlies

PR #11886, opened 2017-03-21
[guilib] add static_assert beside type trait to produce readable erro…

xhaggi:
@Montellese as you suggested in #11869

### DIFF
--- a/xbmc/guilib/GUIWindowManager.h
+++ b/xbmc/guilib/GUIWindowManager.h
@@ -152,7 +152,10 @@ public:
    * \return the window with for the given type \code{T} or null
    */
   template<typename T, typename std::enable_if<std::is_base_of<CGUIWindow,T>::value>::type* = nullptr>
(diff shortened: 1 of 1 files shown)


Montellese:
Thanks. Have you tried it out with out without the `static_assert()` assert to see the difference?

xhaggi:
I'm on Xcode and there is no difference.

xhaggi:
> No matching member function for call to 'GetWindow'

and some details:

> /xbmc/guilib/GUIWindowManager.h:154:48: Candidate template ignored: disabled by 'enable_if' [with T = CSettings]

xhaggi:
@Montellese good to go in?

Montellese:
Ah you're right I didn't consider the `enable_if`. In that case the `static_assert` isn't really necessary. Sorry about the noise :-/

xhaggi:
Np will close the PR

PR #11887, opened 2017-03-21
[pvr] cleanups in PVR manager

xhaggi:
Rewrite of announce logic used by PVR manager + some small refactorings.

@ksooo mind taking a look.


ksooo:
Well, this is not the right approach. @FernetMenta might explain this best, because he introduced the direct pvr manager calls in powermanager for a reason.

xhaggi:
Okay please let me know why, i didn't noticed it or point me to the discussion. IMO it makes sense to decouple the PVR manager from the PowerManager because the announce is trigger right next to that lines.

ksooo:
Ask fernet, please. I was not directly involved, and only remember that there was some discussion somewhere leading to a pvr api change and that fernet was involved. Sorry, seems I'm getting old. ;-)

xhaggi:
The discussion is in #9662 which introduce the sleep/wakeup pvr client stuff.

xhaggi:
The last commit will drop the announce logic for OnQuit, OnSleep, OnWake and use the methods `OnWake()`, `OnSleep()` and `Deinit()` instead. The methods are called directly like before, but I dropped the `SetWakeCommand()` call from the `PowerManager`. If good I'll squash the commit.

ksooo:
Yeah, this reads good. Will take a look at the code for the usual nitpicking later, when I'm back home.

xbmc/pvr/PVRManager.cpp
```diff
@@ -594,12 +573,31 @@ bool CPVRManager::SetWakeupCommand(void)
 
 void CPVRManager::OnSleep()
 {
+  SetWakeupCommand();
+  SaveLastPlayedChannel();
```
ksooo: Just cosmetics, but in CPVRManager::Deinit() you are calling these two methods in the the opposite order.

xbmc/pvr/PVRManager.h
```diff
@@ -561,6 +556,16 @@ namespace PVR
 
   private:
     /*!
+     * @brief Save the current playing channel as last played channel
```
ksooo: correct English is 'currently' here, imo.

xbmc/pvr/PVRManager.cpp
```diff
@@ -594,12 +573,31 @@ bool CPVRManager::SetWakeupCommand(void)
...
+
   g_PVRClients->OnSystemSleep();
 }
 
 void CPVRManager::OnWake()
 {
   g_PVRClients->OnSystemWake();
+
+  if (IsStarted())
```
ksooo: I just spotted, that CPVRManager::Announce always() does an IsStarted check before doing anything.

Okay, but then you should add the IsStarted check also to CPVRManager::OnSleep() and CPVRManager::Deinit(), not only to CPVRManager::OnWake()

Finally, the IsStarted check should be removed from CPVRManager::TriggerSearchMissingChannelIcons().

ksooo (changes requested):

xbmc/pvr/PVRManager.cpp
xhaggi: IMO every method should be self-contained. They need to validate the state if necessary. We should drop that check in OnWake.

xbmc/pvr/PVRManager.cpp
xhaggi: Sure will change the order.

xbmc/pvr/PVRManager.h
xhaggi: Okay

xbmc/pvr/PVRManager.cpp
ksooo: ok

xhaggi:
@FernetMenta dropped the commit which removes the switch channel job

@ksooo mind taking another look at it. The last commit drops the globals in PVR manager and use the CServiceBroker instead.

ksooo (commented):
No strong opinion on that, but what's wrong with g_* defines? The only "change" I see here is, that I now have to type more. For me, this is not an improvement, but a drawback.

xhaggi:
We want to get rid of those globals and it's done now. Later we could think about to detach recordings, clients, timers etc from the PVRManager and provide them by the ServiceManager/ServiceBroker too.

xhaggi:
jenkins build this please

ksooo:
Sorry, I don't see the point here. But no -1 from my side. Just asking myself what the benefit/improvement/cleanup is.

ksooo:
> Later we could think about to detach recordings, clients, timers etc from the PVRManager and provide them by the ServiceManager/ServiceBroker too.

 Haha. This could be done even more simple if we still had the defines.

old: #define g_PVRClients CServiceBroker::GetPVRMananger().Clients()
new: #define g_PVRClients CServiceBroker::GetPVRClients()

A one liner to change everything!

ksooo:
> We want to get rid of those globals and it's done now.

 My understanding of our goal to "get rid of the gloabls" is not to remove defines, but do architectural changes. Imo, a good approach to "get rid of the globals" is to pass the "owning component" (the PVR Manager for example) to the "owned components" (the PVR recordings for example), so that the owned component knows its owner and can access it, without calling a global method. @FernetMenta what's your take on this?

xhaggi:
> Imo, a good approach to "get rid of the globals" is to pass the "owning component" (the PVR Manager for example) to the "owned components" (the PVR recordings for example), so that the owned component knows its owner and can access it, without calling a global method.

That was my first thought, but passing the owner to each of them is not a nice approach for a service class. IMO all of them are services. My second thought was "why not detaching them from the PVR manager".  But I don't have a strong option on how to proceed here.

xhaggi:
BTW passing the owner to the owned services does not really solve your "we have to write more code" issue. Each access to Recordings, Timers etc outside will result in the same call `CServiceBroker::GetPVRManager().GetTimers()`.

xhaggi:
any objections or ready to merge?

After that I'll create another PR that drops the channel switch job as @FernetMenta suggested.

xhaggi:
Noticed that the PVRManager still expose a `GetInstance()` method which delegates the call to the `ServiceBroker`. I've added another commit to drop that and use the `ServiceBroker` directly

jenkins build this please

ksooo:
> Noticed that the PVRManager still expose a GetInstance() method which delegates the call to the ServiceBroker. I've added another commit to drop that and use the ServiceBroker directly

 Nice.

ksooo (approved):

xhaggi:
jenkins build this please

xhaggi:
WIN32 build errors are unrelated.

PR #11875, opened 2017-03-19
FIX: [cmake] use specific Android profile (so versions)

koying:
Use android cmake profile (which handles the android-specific non-versioned .so) and get-rid of last RPL hacks

PR #11888, opened 2017-03-22
[cmake] Fix compiling of Android-X86 after 742cc3c31b by checking SSE support again

Rechi:
## Description
Check SSE support on android again after #11875.
@koying FYI

## Types of change

## Checklist:


koying:
Mmm... Why are testing the target system to check if the build system support SSE?

Or, in other words, what is FindSSE for, exactly?

Rechi:
Checking `/proc/cpuinfo` may be wrong, but it also does `CHECK_CXX_ACCEPTS_FLAG(...)` which checks the build system.
FindSSE checks if SSE instructions can be used. Android-X86 needs them otherwise it won't compile.
I'm just restoring the behaviour before your PR was merged, because `CMAKE_SYSTEM_NAME` was set to Linux and the if block was executed. With this change the if block gets executed again.

Rechi:
jenkins build this please

koying:
Yeah, so your fix is correct, but checking /proc/cpuinfo is just not for cross-compiling.
If you happen to build on a machine not supporting SSE (ie any not x86 machine, afaik), it will fail the same way as currently.

ping @wsnipex

wsnipex:
this check was never intended for droid. I'd handle it just like windows, i.e. simply set the needed flags to true.

koying:
Actually, checking the build machine capabilities to set a target machine flags is always wrong, imho ;)
@wsnipex Do you agree removing the "/proc/cpuinfo" + regexp thingy, and only leaving the CHECK_CXX_ACCEPTS_FLAG part, is the way to go?

wsnipex:
nope, we want it for native builds. OSX can be considered native in this regard as well.
For cross we simply expect the platform and toolchain to support certain features, i.e. neon for arm or SSEx for droid x86

Rechi:
@koying @wsnipex  would be
```
elseif(CMAKE_SYSTEM_NAME MATCHES "Android")
  if(CPU MATCHES "x86_64" OR CPU MATCHES "i.86")
      CHECK_CXX_ACCEPTS_FLAG("-msse" _SSE_OK)
      CHECK_CXX_ACCEPTS_FLAG("-msse2" _SSE2_OK)
      CHECK_CXX_ACCEPTS_FLAG("-msse3" _SSE3_OK)
      CHECK_CXX_ACCEPTS_FLAG("-mssse3" _SSSE3_OK)
      CHECK_CXX_ACCEPTS_FLAG("-msse4.1" _SSE41_OK)
      CHECK_CXX_ACCEPTS_FLAG("-msse4.2" _SSE42_OK)
      CHECK_CXX_ACCEPTS_FLAG("-mavx" _AVX_OK)
      CHECK_CXX_ACCEPTS_FLAG("-mavx2" _AVX2_OK)
  endif()
elseif(CMAKE_SYSTEM_NAME MATCHES "Darwin")
```
ok?

koying:
> nope, we want it for native builds

The only way this is correct is when the build machine = the target machine currently.
In all cases, just doing the CHECK_CXX_ACCEPTS_FLAG is correct, so I fail to see what's the benefit of adding a potentially wrong test on "/proc/cpuinfo" or "/usr/sbin/sysctl"...

Besides, removing those will just leave:

```
if windows
  // hardcode
else
  CHECK_CXX_ACCEPTS_FLAG 
```

Simplifying the code.

@Rechi Yep, that's a solution.
You might need all the
```
    set(_SSE<*>_TRUE true)
```
to avoid bogus messages.

Rechi:
@koying updated

koying:
@wsnipex Even on pure linux, what if a distro builds on an old machine that do not support some newer SSE versions?

wsnipex:
1.) all those flags can be overridden
2.) we actually only use SSE1 and 2, so thats really all we need. Every CPU that kodi can be expected to run on supports this.
FFmpeg has runtime cpu detection

cmake/modules/FindSSE.cmake
```diff
@@ -40,6 +40,26 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
...
+elseif(CMAKE_SYSTEM_NAME MATCHES "Android")
+  if(CPU MATCHES "x86_64" OR CPU MATCHES "i.86")
+    set(_SSE_TRUE TRUE)
+    set(_SSE2_TRUE TRUE)
+    set(_SSE3_TRUE TRUE)
+    set(_SSSE3_TRUE TRUE)
+    set(_SSE41_TRUE TRUE)
+    set(_SSE42_TRUE TRUE)
+    set(_AVX_TRUE TRUE)
+    set(_AVX2_TRUE TRUE)
```
wsnipex: are you sure all CPUs in x86 based android devices support avx(2)?
I'd be much more conservative here and only enable what kodi requires to build by default. <= SSE2 should be enough.

Rechi: will remove `_SSE41_TRUE`, `_SSE42_TRUE`, `_AVX_TRUE`, and `_AVX2_TRUE`, because the were not set (/proc/cpuinfo check) in previous builds.
Previous jenkins builds reported Found SSE, SSE2, SSE3 and SSSE3.

Rechi:
@wsnipex updated
now it reports the same instructions as found as before 742cc3c.

wsnipex:
ok, jenkins build this please

koying:
Will take ma a while to get accustomed to those constructs ;)

So, e.g, SSE4_1_FOUND = _SSE41_TRUE & _SSE41_OK , correct?
If it is, I *think* doing the CHECK_CXX_ACCEPTS_FLAG for anything other that the ones we set _xxx_TRUE to true is unneeded, correct?

Rechi:
@koying correct
Some `CHECK_CXX_ACCEPTS_FLAG` are currently not need, but they were also execute previously.
I would like to get this in to get correct reports from jenkins.If you want to refactor this file you can do it in a separate PR.

koying:
Ok

PR #11891, opened 2017-03-22
Revert "[bluray] Early return if requested "dir" is in fact a file."

ace20022:
This reverts commit 462d64d9bbac5cde36c5e52efabe78e77260c2ac.

Our udf implementation does not distinguish between dirs and files inside an iso file.


ace20022:
jenkins build and merge

ace20022:
@popcornmix @MilhouseVH thx for the report!

PR #11893, opened 2017-03-23
[windows] use wmic qfe instead of powershell to check if hotfix is installed

Rechi:
## Description
Use `wmic qfe` instead of trying to execute a similar command with powershell to check if hotfix is installed.

## Motivation and Context
Some users didn't have the powershell directory in PATH and therefore weren't able to install Kodi, although they have the required hotfix installed.
Not having it in PATH seems strange, but they haven't removed it themself.

## How Has This Been Tested?
Running the installer

## Types of change

## Checklist:

### DIFF
--- a/project/Win32BuildSetup/genNsisInstaller.nsi
+++ b/project/Win32BuildSetup/genNsisInstaller.nsi
@@ -354,14 +354,17 @@ Function .onInit
     StrCpy $HotFixID ""
   ${Endif}
   ${If} $HotFixID != ""
-    SetOutPath "$TEMP\PS"
-    FileOpen $0 ps.ps1 w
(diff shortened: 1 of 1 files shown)


afedchin (approved):

Paxxi (approved):

Rechi:
added a check if `wmic` can be executed and a different error message if it isn't possible.

Rechi:
Wmic gets now called by the full path instead of relying on the PATH environment variable.
If SYSTEMROOT is wrong the system is really messed up and I think in that case the smallest problem is that Kodi can't be installed.
If the wording of the error message is fine, this can me merged.

PR #11895, opened 2017-03-24
[pvr] no need to run a job to announce channel switch

xhaggi:
As suggested by @FernetMenta here is the separate PR for removing the channel switch job.

@ksooo ping you too.


FernetMenta:
thanks! Let's try this. In case of any issues we can easily spot this.

ksooo:
Yep, let's make it so.

xhaggi:
Okay, jenkins build this please.

PR #11898, opened 2017-03-24
[PVR] Follow PR11887: Fix CPVRManager::ConnectionStateChange not to c…

ksooo:
…heck for started PVR manager.

This fixes non working PVR after #11887.

@xhaggi @stefansaraev @MilhouseVH  fyi


ksooo:
jenkins build and merge

xhaggi:
My mistake, thanks for pointing that out.

+1 if jenkins is happy

PR #11900, opened 2017-03-25
Remove '[CR]', formatting should be left to skins

BADMS:
These [CR] kind of dictate layout/formatting, this should be left as plain as possible for skins to manipulate as needed [imho] :)
Usage scenario: I want to place the text provided by kodi on one 'label' line, but the [CR] hardcoded means it has to be a 'textbox' or string of same height to match height of various linesnow forced by '[CR]', ie. if one straight line of text I can scroll it, I don't have to make a large box.
Thanks.

## Types of change

## Checklist:

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -497,7 +497,7 @@ msgstr ""
 
 #: xbmc/windows/GUIWindowFileManager.cpp
 msgctxt "#125"
-msgid "Would you like to delete the selected file(s)?[CR]Warning - this action can't be undone!"
+msgid "Would you like to delete the selected file(s)? Warning - this action can't be undone!"
 msgstr ""
 
 #: xbmc/peripherals/bus/PeripheralBus.cpp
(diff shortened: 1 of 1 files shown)


addons/resource.language.en_gb/resources/strings.po
```diff
@@ -16814,7 +16814,7 @@ msgstr ""
 #. This string is an early meme from the late 1990's that made fun of the poor translation in the 16-bit game Zero Wing from the late 1980's. DO NOT TRANSLATE!
 #: system/settings/settings.xml
 msgctxt "#35200"
-msgid "ALL YOUR BASE ARE BELONG[CR]TO US"
+msgid "ALL YOUR BASE ARE BELONG TO US"
```
garbear: Can the [CR] here remain? Part of the history of this joke involves the incongruous line break

da-anda:
certain linebreaks are intended for better readability of long descriptions

MartijnKaijser:
i think we can close this because as said certain linebreaks are intentional.

PR #11903, opened 2017-03-25
Fix crash due to exception when entering Games -> Keyboard settings

garbear:
More concerning is that a typo in strings.po can crash Kodi. We should probably add a try/catch block when calling the new format function.


garbear:
@Montellese do you think you could add a try/catch block in the format writer?

garbear:
jenkins build this please

PR #11905, opened 2017-03-26
[PVR] Trac #17374: enabled "PVR -> Power saving -> Wakeup command" setting is not called

ksooo:
This should finally fix http://trac.kodi.tv/ticket/17374. First fix attempt (#11849) was incomplete.

This is a more complete approach, which introduces a self contained auto updating settings container for PVR settings.

@FernetMenta for review, please.

### DIFF
--- a/xbmc/pvr/PVRGUIActions.cpp
+++ b/xbmc/pvr/PVRGUIActions.cpp
@@ -128,7 +128,15 @@ namespace PVR
   };
 
   CPVRGUIActions::CPVRGUIActions()
-  : m_bChannelScanRunning(false)
+  : m_bChannelScanRunning(false),
+    m_settings({
+      CSettings::SETTING_PVRRECORD_INSTANTRECORDTIME,
+      CSettings::SETTING_PVRRECORD_INSTANTRECORDACTION,
+      CSettings::SETTING_PVRPLAYBACK_SWITCHTOFULLSCREEN,
+      CSettings::SETTING_PVRPLAYBACK_STARTLAST,
+      CSettings::SETTING_PVRPARENTAL_PIN,
+      CSettings::SETTING_PVRPARENTAL_ENABLED
+    })
   {
   }
 
@@ -336,7 +344,7 @@ namespace PVR
     class InstantRecordingActionSelector
     {
     public:
-      InstantRecordingActionSelector();
+      InstantRecordingActionSelector(int iInstantRecordTime);
       virtual ~InstantRecordingActionSelector() {}
 
       void AddAction(PVRRECORD_INSTANTRECORDACTION eAction, const std::string &title);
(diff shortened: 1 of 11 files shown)


FernetMenta:
@ksooo I will have a look soon

ksooo:
@FernetMenta i have some ideas to dramatically simplify this. no need for a review atm. will ping you when I think I'm done.

ksooo:
Added some more commits:

1. get rid of CPVRGUIActions static singleton. Caused crashes on exit while destructing CPVRGUIAction's new CPVRSettings member while trying to deregister listener from already destructed settings manager

2. Factor out a minimal timers container and use it instead of CPVRTimers when updating timers, Saves resources and performs better (for example, no need to register as observer and setting listener etc while doing timers update).

3. Fix a memeory leak in CPVRTimers' timer tag map.

ksooo:
4. Fix destruction of windows on app exit (windows were queued for deletion, but dtors were never called, because async destruction does not work anymore on app exit).

ksooo:
@FernetMenta before this PR gets to big (stumbling over bugs and glitches again and again while working on the root cause of trac17374 ...) I want to hear your opinion.

xbmc/pvr/PVRGUIActions.cpp
```diff
@@ -450,9 +453,9 @@ namespace PVR
       if (bOnOff && !channel->IsRecording())
       {
         CEpgInfoTagPtr epgTag;
-        int iDuration = CServiceBroker::GetSettings().GetInt(CSettings::SETTING_PVRRECORD_INSTANTRECORDTIME);
+        int iDuration = m_settings.GetIntValue(CSettings::SETTING_PVRRECORD_INSTANTRECORDTIME);
```
xhaggi: Why do you introduce this and all the stuff in the PVRSettings. Why do we need a container for the PVR related settings? If you want to get the current setting value, all you need to do this calling this or the other type related setting methods. What do you want to achieve with it?

ksooo: Well, please read #11849 for the explanation.

xhaggi: Thx will do

ksooo:
I removed the not directly related commits from this PR and made separate PRs out of them:

- #11918
- #11919 
- #11920

FernetMenta:
@ksooo I only had time for a brief look at it. I think your changes are fine. Nice that you dropped another singleton that makes PVR a more self contained component.

ksooo:
rebased

ksooo:
jenkins build this please

ksooo:
Ay objctions? Otherwise I will merge this the next days.

ksooo:
jenkins build this please

PR #8008, opened 2015-09-10
JSON: replace YAJL with RapidJSON

Montellese:
This is an approach at replacing our JSON parsing/writing done with YAJL with RapidJSON which, according to a lot of benchmarks (https://code.google.com/p/rapidjson/wiki/Performance, https://github.com/mloskot/json_benchmark, https://github.com/miloyip/nativejson-benchmark, http://chadaustin.me/2013/01/json-parser-benchmarking/) is one of the fastest JSON libraries available and a lot faster than YAJL. We are not likely to notice this a lot because most of the JSON we process is rather small but it certainly won't hurt either.

I've also written a few unit tests for the JSON parsing/writing (which are independent of the JSON library we use).

rapidjson doesn't seem to be available for Ubuntu so it would need to be added to our PPA. I also haven't uploaded the source tarball and the win32 dependency archive yet.

TODOs:
- [x] upload `rapidjson-1.0.2.tar.gz` to mirrors
- [x] upload `rapidjson-1.0.2-win32.7z` to mirrors
- [x] add `rapidjson` to PPA

PR #11908, opened 2017-03-26
[cmake] fail if rapidjson is not found

wsnipex:
the packagename in find_package_handle_standard_args must be exactly the same as the one find_package has been called with.

fixes not erroring out in cmake when rapidjson is not found.

@Rechi sry, I requested the change, so I fix it ;)

### DIFF
--- a/cmake/modules/FindRapidJSON.cmake
+++ b/cmake/modules/FindRapidJSON.cmake
@@ -10,17 +10,22 @@
 #
 
 if(PKG_CONFIG_FOUND)
-  pkg_check_modules(PC_RapidJSON RapidJSON QUIET)
+  pkg_check_modules(PC_RapidJSON RapidJSON>=1.0.2 QUIET)
+endif()
+
+if(CORE_SYSTEM_NAME STREQUAL windows OR CORE_SYSTEM_NAME STREQUAL windowsstore)
+  set(RapidJSON_VERSION 1.0.2)
+else()
+  set(RapidJSON_VERSION ${PC_RapidJSON_VERSION})
 endif()
 
 find_path(RapidJSON_INCLUDE_DIR NAMES rapidjson/rapidjson.h
                                 PATHS ${PC_RapidJSON_INCLUDEDIR})
 
-set(RapidJSON_VERSION ${PC_RapidJSON_VERSION})
 
 include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(RAPIDJSON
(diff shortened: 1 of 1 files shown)


Rechi (approved):

MilhouseVH:
Native Linux builds now fail with this PR:
```
-- Found FriBidi: /usr/lib/x86_64-linux-gnu/libfribidi.so (found version "0.19.7")
-- Found PCRE: /usr/lib/x86_64-linux-gnu/libpcrecpp.so (found version "8.38")
CMake Error at /usr/share/cmake-3.5/Modules/FindPackageHandleStandardArgs.cmake:148 (message):
  Could NOT find RapidJSON (missing: RapidJSON_INCLUDE_DIR)
  /usr/share/cmake-3.5/Modules/FindPackageHandleStandardArgs.cmake:388 (_FPHSA_FAILURE_MESSAGE)
  cmake/modules/FindRapidJSON.cmake:22 (find_package_handle_standard_args)
  cmake/scripts/common/Macros.cmake:315 (find_package)
  CMakeLists.txt:169 (core_require_dep)


See also "/home/neil/projects/kodi.git/CMakeFiles/CMakeOutput.log".
See also "/home/neil/projects/kodi.git/CMakeFiles/CMakeError.log".
```
I do not have RapidJSON installed on the host.

Without this PR:
```
-- Could NOT find RAPIDJSON (missing:  RapidJSON_INCLUDE_DIR)
```
So for some reason it's not downloading from the mirror...

MilhouseVH:
Sorry, it doesn't build successfully when RapidJSON is not found (obviously):
```
[ 85%] Built target guilib
/home/neil/projects/kodi.git/xbmc/utils/JSONVariantWriter.cpp:23:36: fatal error: rapidjson/prettywriter.h: No such file or directory
compilation terminated.
```
My question is, should #8008 now be downloading the source when it's not installed on the host?

I'm building on Ubuntu 16.04 and can `apt-get install rapidjson-dev` which is v0.12, but would rather build with v1.0.2 if possible.

wsnipex:
it will not automatically download anything, we only do that for libdvd and ffmpeg(if needed).

you can find 1.0.2 in our build depends ppa

MilhouseVH:
Ah ok, thanks for the clarification - it's not often I build native. :)

wsnipex:
updated to check for min version 1.0.2
jenkins build this please

wsnipex:
meh, windows is special again.
jenkins build this please

PR #11910, opened 2017-03-26
Get ready for Mir 1.0 by removing deprecations

BrandonSchaefer:
## Description
Remove all deprecations and replace them with the new way to do it. Though the main diff change is the rename of surface --> window.

## Motivation and Context
So kodi can render on mir 1.0 once released.

## How Has This Been Tested?
Tested on ubuntu 17.04 working. This new change will require mir 0.26.2

## Types of change

## Checklist:

### DIFF
--- a/cmake/modules/FindMir.cmake
+++ b/cmake/modules/FindMir.cmake
@@ -11,7 +11,7 @@
 
 
 if(PKG_CONFIG_FOUND)
-  pkg_check_modules (PC_MIR mirclient QUIET)
+  pkg_check_modules (PC_MIR mirclient>=0.26.2 QUIET)
 endif()
 
 find_path(MIR_INCLUDE_DIR NAMES mir_toolkit/mir_client_library.h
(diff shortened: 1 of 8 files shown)


Rechi:
This won't build on older versions of Ubuntu anymore?

FernetMenta:
@Rechi it requires mir 0.26.2. Since 0.26.1 was available on 15.04 I would think that 0.26.2 will become available on older version of Ubuntu too. If that would make sense in the context of Kodi is a different question.

@BrandonSchaefer thanks!

BrandonSchaefer:
Yes this is true, and right now 0.26.2 is as far back as 16.04, through a stable phone ppa:
https://launchpad.net/~ci-train-ppa-service/+archive/ubuntu/stable-phone-overlay

If its wanted to support older versions of mir then there would need to be some if/defs which would lead to a messy port overall. Ideally I would like to attempt to stick to major releases such as mir 1.0 and further down the road mir 2.0. Though thats up to you guys! Im happy to keep a working version of this on w/e is wanted or needed.

FernetMenta:
As far as I am concerned I am happy to follow the advice of Canonical in this regard.

PR #11913, opened 2017-03-27
Disable gcc printf verifier for logging function

notspiff:
Now that the newfangled libfmt stuff is used, we get a warning from the verifier and it causes more harm than good.

### DIFF
--- a/xbmc/utils/log.h
+++ b/xbmc/utils/log.h
@@ -42,7 +42,7 @@ public:
   CLog();
   ~CLog(void);
   static void Close();
-  static void Log(int loglevel, PRINTF_FORMAT_STRING const char *format, ...) PARAM2_PRINTF_FORMAT;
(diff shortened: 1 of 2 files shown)


wsnipex:
@Montellese ping

xbmc/utils/log.h
```diff
@@ -42,7 +42,7 @@ class CLog
   CLog();
   ~CLog(void);
   static void Close();
-  static void Log(int loglevel, PRINTF_FORMAT_STRING const char *format, ...) PARAM2_PRINTF_FORMAT;
+  static void Log(int loglevel, PRINTF_FORMAT_STRING const char *format, ...);
```
Paxxi: This should probably use a variadic template now that it passes everything on to libfmt but that's a larger change.

Montellese:
Removing the verifier certainly makes sense. And what @Paxxi said makes sense as well but can be done in a separate PR.

wsnipex:
jenkins build and merge

PR #11915, opened 2017-03-28
Fix unhandled exception with invalid fmt arguments

garbear:
Follow up to #11903

## Types of change

### DIFF
--- a/xbmc/utils/StringUtils.h
+++ b/xbmc/utils/StringUtils.h
@@ -40,6 +40,7 @@
 
 #include "LangInfo.h"
 #include "XBDateTime.h"
+#include "utils/log.h"
 #include "utils/params_check_macros.h"
 
 class StringUtils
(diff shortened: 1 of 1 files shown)


Paxxi:
Don't log here, this introduces a circular dependency between clog and stringutils. Also invalid format should be a bug so letting the exception propagate and crash is a reasonable way to handle it. Possibly not for translations though

Paxxi:
There's little reason to specify a type in the format specifier unless you want an exception thrown or when specifying a certain formatting. Most uses should be fine with `{} ` or for positional `{#}`

Montellese:
Right now it actually doesn't behave any different than it did before with the added benefit that a lot of cases are already handled at compile-time which weren't before. With the old code you get a crash when you try to e.g. format an `std::string` instead of a `const char*`.

Maybe we can let it crash in debug mode and catch the exception in release mode? But I don't really like that either :-/

Paxxi:
Would be so much simpler if we had a higher level method for all translations where we could catch and log the error. Thinking about it, maybe it's time to add one.

garbear:
Comment from @FernetMenta:

> A wrong format is not a runtime error. Having the app crash is much better than hiding the error.

Closing.

PR #11916, opened 2017-03-28
[Estuary] PVR Channel list - Fix overlapping text.

FergusGeorge:
Move the top of the channel list, and adjust the movement and focusposition, so that it's contents don't overlap with the top and bottom breadcrumb trails.  When using PageUp and PageDown controls to scroll through the list, the top and bottom entries were rendered unreadable.

## Description
- Adjust the 'top' property of the PVR Channels List so that it doesn't clash with the upper TopBar.
- Adjust the 'movement' and 'focusposition' properties so that there's no gap at the top of the list when it's first viewed (and the 1st channel is highlighted) and to ensure that the bottom entry isn't fouled by the BottomBar. 

## Motivation and Context
- Unable to read top and bottom channel details when scrolling through list using PageUp and PageDown controls.

## How Has This Been Tested?
- Validated on a Raspberry Pi running OSMC with Kodi 17.1 at 1920 x 1080 resolution.
- Checked list scrolls normally with up/down/pageup/pagedown remote control buttons.
- Checked top and bottom entries aren't obscured by the TopBar and BottomBar.
- Checked top and bottom entries can be selected. 

Before:
before
After:
after

## Types of change
- [ X] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x ] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document

### DIFF
--- a/addons/skin.estuary/xml/MyPVRChannels.xml
+++ b/addons/skin.estuary/xml/MyPVRChannels.xml
@@ -12,11 +12,11 @@
 				<include>OpenClose_Left</include>
 				<control type="fixedlist" id="50">
 					<left>0</left>
-					<top>list_y_offset</top>
+					<top>140</top>
 					<right>918</right>
 					<bottom>list_y_offset</bottom>
-					<movement>3</movement>
-					<focusposition>5</focusposition>
+					<movement>5</movement>
+					<focusposition>4</focusposition>
 					<scrolltime tween="cubic" easing="out">500</scrolltime>
 					<orientation>vertical</orientation>
 					<onleft>9000</onleft>


ksooo:
Could you please update the commit message to something more meaningful (like you did with the PR title).

@phil65 ping for review.

FergusGeorge:
I've now added a second commit so that the commit message is more meaningful.

phil65:
This would only make sense when done globally, not just for PVR section.
Imo, a better solution would be to allow the skinner to set the jump size for scrollbars.

FergusGeorge:
I was tempted to tinker instead with the following entry in addons/skin.estuary/xml/Includes.xml:
`<constant name="list_y_offset">0</constant>` but I was wary of unintended consequences.

Is that worth a look, or am I better backing away and leaving this to the big boys?

For what it's worth, I'm using the above patch on my own system, alongside an equivalent one for MYPVRRecordings.xml.  The overlapping text isn't as obtrusive in the list and widelist views that I use for Movies, Music and TV Shows, as their list boxes aren't aligned to hard left of the screen.

Paxxi:
@FergusGeorge for future reference `git commit --amend` let's you edit the commit message and requires a force push

dkadioglu:
As I don't like the current behaviour of the channel list, too, I would like to add the values I use:
```--- MyPVRChannels.xml   2017-06-10 13:40:39.655950398 +0200
+++ MyPVRChannels.xml   2017-06-10 13:35:04.916516145 +0200
@@ -12,11 +12,11 @@
                                <include>OpenClose_Left</include>
                                <control type="fixedlist" id="50">
                                        <left>0</left>
-                                       <top>list_y_offset</top>
+                                       <top>120</top>
                                        <right>918</right>
-                                       <bottom>list_y_offset</bottom>
-                                       <movement>3</movement>
-                                       <focusposition>5</focusposition>
+                                       <bottom>40</bottom>
+                                       <movement>5</movement>
+                                       <focusposition>4</focusposition>
                                        <scrolltime tween="cubic" easing="out">500</scrolltime>
                                        <orientation>vertical</orientation>
                                        <onleft>9000</onleft>```
In my opinion the channel list behaves and looks way better with these. But I can understand that this should be fixed globally. How is the progress with that? Thanks in advance.

ksooo:
As @phil65 said: "This would only make sense when done globally, not just for PVR section.
Imo, a better solution would be to allow the skinner to set the jump size for scrollbars."

ksooo (commented):
-1 from my end.

FergusGeorge:
I revisited this on my own setup after the most recent OSMC update overwrote my changes.  For what's it's worth, I've now changed list_y_offset from 0 to 80 in the skin's Includes.xml file so that none of the lists foul against the top.  I've also changed every occurrence of `<bottom>list_y_offset</bottom>` to `<bottom>0</bottom>` across each of the skin's xml files so that it's only the top that's corrected, and there's no need to mess with the individual _movement_ and _focusposition_ tags.

Whilst it's not as complete as the full solution preferred by @phil65, it is a skin-wide and cleaner solution than my first attempt.

Any interest in this approach as an interim measure until someone smarter than me tackles the jumpsize stuff?

PR #11917, opened 2017-03-29
[PVR] Krypton: Quick and dirty fix for trac #17374.

ksooo:
This a quick and dirty fix for http://trac.kodi.tv/ticket/17374 
Proper fix for Leia will be provided by #11905

The original patch was supplied and tested by the trac ticket submitter.

@Jalle19 mind taking a look.

### DIFF
--- a/xbmc/pvr/PVRManager.cpp
+++ b/xbmc/pvr/PVRManager.cpp
@@ -493,9 +493,6 @@ void CPVRManager::Unload()
 
 void CPVRManager::Shutdown()
 {
-  // set system wakeup data
-  SetWakeupCommand();
-
   Unload();
 
   // release addons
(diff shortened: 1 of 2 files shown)


ksooo:
jenkins build this please

xbmc/pvr/PVRManager.cpp
```diff
@@ -493,8 +493,11 @@ void CPVRManager::Unload()
 
 void CPVRManager::Shutdown()
 {
+// quick and dirty Krypton-only fix for http://trac.kodi.tv/ticket/17374
+#if 0
   // set system wakeup data
   SetWakeupCommand();
```
ksooo: Why? I just was in the mood. And as this is krypton-only quick and dirty I don't really care, tbh. I don't think I will change this. I don't want to spend more time on this.

Jalle19 (approved):
I would also just remove the lines instead of using preprocessor directives, but it's not a big deal

ksooo:
jenkins build and merge

@Jalle19 @xhaggi I removed the <code>#if 0</code>

xhaggi:
Thanks

ksooo:
jenkins build this please

PR #11918, opened 2017-03-29
[guilib] CGUIWindowManager: Fix destruction of windows on Kodi exit

ksooo:
This PR fixes destruction of windows on Kodi exit (windows were queued for deletion, but dtors were never called, because async destruction does not work anymore on app exit).

I runtime-tested this on macOS, latest kodi master and verified using a debugger that now window destructors are actually called.

@Paxxi mind taking a look.


ksooo:
rebased.

jenkins build this please

frytek:
kodi death USA is fuckin crazy


1 kwi 2017 17:01 "Kai Sommerfeld" <notifications@github.com> napisał(a):

Paxxi:
Changes look sane so no objections. Don't think this actually matters as the dtor should just clean up resources and that happens anyway when the process exits. Never hurts to do the right thing though :)

ksooo:
jenkins errors are unrelated.

<pre
autoconf was not found and could not be built. Aborting.
./bootstrap: line 12: autoconf: command not found
</pre>

PR #11923, opened 2017-03-31
[rbpi] Make sure we strip binaries in Release build

nsenica:
This was missing from the rbpi build system.

Set the correct compiler flag so that the resulting binary can be stripped.


wsnipex:
@popcornmix ping
we already have this for x86

popcornmix:
Yes, I've noticed that kodi.bin isn't currently stripped so +1 from me.

wsnipex:
jenkins build and merge

PR #11924, opened 2017-03-31
[backport] [rbpi] Make sure we strip binaries in Release build

nsenica:
This was missing from the rbpi build system.

Set the correct compiler flag so that the resulting binary can be stripped.


MartijnKaijser:
Jenkins build this please

wsnipex (approved):

PR #11929, opened 2017-04-02
[filesystem][dialogs] Add context menus for favourites home screen widget

ksooo:
Subject says it all - this PR adds context menus to the favorite home screen widget items.

screenshot000

This change has been tested on macOS, latest kodi master.

@phil65 fyi, iirc you are in favor for this feature...

@Jalle19 @tamland mind doing the review.

### DIFF
--- a/xbmc/dialogs/GUIDialogFavourites.cpp
+++ b/xbmc/dialogs/GUIDialogFavourites.cpp
@@ -103,14 +103,10 @@ void CGUIDialogFavourites::OnClick(int item)
   if (item < 0 || item >= m_favourites->Size())
     return;
 
-  // grab our message, close the dialog, and send
-  CFileItemPtr pItem = (*m_favourites)[item];
-  std::string execute(pItem->GetPath());
-
   Close();
 
   CGUIMessage message(GUI_MSG_EXECUTE, 0, GetID());
-  message.SetStringParam(execute);
+  message.SetStringParam(CFavouritesDirectory::GetExecutePath(*(*m_favourites)[item], GetID()));
   g_windowManager.SendMessage(message);
 }
 
@@ -125,12 +121,12 @@ void CGUIDialogFavourites::OnPopupMenu(int item)
   CContextButtons choices;
   if (m_favourites->Size() > 1)
   {
-    choices.Add(1, 13332);
-    choices.Add(2, 13333);
+    choices.Add(1, 13332); // Move up
+    choices.Add(2, 13333); // Move down
   }
-  choices.Add(3, 15015);
-  choices.Add(4, 118);
-  choices.Add(5, 20019);
+  choices.Add(3, 15015); // Remove
+  choices.Add(4, 118); // Rename
+  choices.Add(5, 20019); // Choose thumbnail
 
   CFileItemPtr itemPtr = m_favourites->Get(item);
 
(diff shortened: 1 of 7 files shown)


Paxxi:
Not a fan of adding even more stuff to to CFileItem but since we don't have a plan for redesigning it I guess it's fine, certainly not worth blocking progress.

xbmc/dialogs/GUIDialogFavourites.cpp
```diff
@@ -254,3 +229,40 @@ CFileItemPtr CGUIDialogFavourites::GetCurrentListItem(int offset)
...
+  }
+  return false;
+}
+
+bool CGUIDialogFavourites::ChooseAndSetNewThumbnail(const CFileItemPtr &item)
+{
+  CFileItemList prefilledItems;
+  if (item->HasArt("thumb"))
+  {
+    const CFileItemPtr current(new CFileItem("thumb://Current", false));
```
Paxxi: Would be nice if you use make_shared here. I know we don't do that currently but trying to change that

ksooo: okay, will change

ksooo:
@Paxxi changes you requested are now in.

tamland:
That 'IsFavourite' had me puzzled for far too long, thinking it actually meant whether or not the item was favourites...
Imo we should use the `favourites://` scheme for this (if it isn't already?) since these are special fileitems that belongs to the favourites:// directory only. We already have to go via the favourites directory to translate path anyway. As a bonus we don't need to add more stuff to CFileItem.

ksooo:
@tamland currently, favourites items don't have a "favourites://" path. m_strPath contains something like "ActivateWindow(1052),\"musicdb:...." Thus, they cannot be identified using the protocol (CFileItem::m_strPath).  But yeah, I can try to change that.

Paxxi:
:thumbs_up:

ksooo:
@Paxxi @tamland here we go. Better this time?

ksooo:
@phil65 please note the skin change. This is removal of a (due to this PR) no longer needed workaround.

ksooo:
jenkins build this please

xbmc/filesystem/FavouritesDirectory.cpp
```diff
@@ -190,8 +192,12 @@ std::string CFavouritesDirectory::GetExecutePath(const CFileItem &item, int cont
 std::string CFavouritesDirectory::GetExecutePath(const CFileItem &item, const std::string &contextWindow)
 {
   std::string execute;
-  if (item.m_bIsFolder && (g_advancedSettings.m_playlistAsFolders ||
-                            !(item.IsSmartPlayList() || item.IsPlayList())))
+  if (item.GetPath().substr(0, 13) == "favourites://")
+  {
+    execute = CURL::Decode(item.GetPath().substr(13));
+  }
```
tamland: we have URIUtils::IsProtocol and ::GetFileName for this (no need for magic numbers)

ksooo: Thanks for the hint. Will change this ofc. Did  not know about URIUtils.

ksooo:
@tamland no more magic numbers. ;-) Good to go?

tamland:
yeah, looks good to me

ksooo:
jenkins build this please

PR #11932, opened 2017-04-04
[osx] - add the minimum required osx version to the info.plist - this…

Memphiz:
… will show a message box on startup when minimum osx version is not installed (also mentioning the needed minimum osx version)

## Description
Ensure that the bundle can't be started on to old osx versions

## Motivation and Context
To many users post crash reports due to unsupported osx versions. Its much better to use the osx feature here which will present the user with a message box that the minimum osx version is not present

## How Has This Been Tested?
This simply sets the osx deployment target as minimum osx version in the info.plist. Tested via generating the info.plist and adding it to a kodi bundle (editing to some very high osx version that is newer then the one i have installed). Results as expected in a dialog telling me i need a newer version for running Kodi.

## Screenshot
https://www.dropbox.com/s/fdgnxsbbtv4ln3a/osxminversion_dialog.png?dl=0

## Types of change

## Checklist:

### DIFF
--- a/xbmc/platform/darwin/osx/Info.plist.in
+++ b/xbmc/platform/darwin/osx/Info.plist.in
@@ -22,6 +22,8 @@
 	<string>@APP_VERSION_MAJOR@.@APP_VERSION_MINOR@.@APP_VERSION_TAG_LC@</string>
 	<key>CFBundleVersion</key>
 	<string>r####</string>
+	<key>LSMinimumSystemVersion</key>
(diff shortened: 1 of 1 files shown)


Memphiz:
Tested on jenkins - http://mirrors.kodi.tv/test-builds/osx/x86_64/kodi-20170404-d37c137-osx_minversion-x86_64.dmg (verified that Info.plist in the bundle has the correct minimum version of 10.8 set).

PR #11933, opened 2017-04-04
[osx] - add the minimum required osx version to the info.plist

Memphiz:
backport of #11932

@fetzerch do you see a problem with adding the deployment target in the ArchSetup? (i need it else the configure command won't replace the variable in Info.plist.in). Any side effects you can think of?


Memphiz:
Tested on jenkins - http://mirrors.kodi.tv/test-builds/osx/x86_64/kodi-20170404-935ac0c-osx_minversionbp-x86_64.dmg (verified that Info.plist in the bundle has the correct minimum version of 10.8 set).

PR #8133, opened 2015-09-25
[videodb] Fix TV shows cleaning after  5755573981ca706a82a7ecbbdf5de112a570377a.

ghost:
A regression in 5755573981ca706a82a7ecbbdf5de112a570377a meant removing a
deleted TV show needed multiple library cleanings (i.e. TV shows only got
removed when `filesToTestForDelete` was empty).  This is much more noticeable
after 5cae3b5bb79a0c7b1945297e54343349f78cf497.

After 5755573981ca706a82a7ecbbdf5de112a570377a:
- if we had files to be checked for deletion: `CleanMediaType` would return a
  decision for the TV show path (delete) and a decision for the parent path
  (keep), so the TV show wasn't removed from the database;
- if no files needed to be checked for deletion: `CleanMediaType` would return
  earlier and `pathsDeleteDecisions` was empty, so we'd have checked if the
  path existed and there was a decision for the parent path (delete) or no
  decision at all.

This commit checks if there's a decision to delete the parent path _or_ it
exists (we only end up there if there's a decision to delete the path or it
doesn't exist), so that you can delete a TV show with episodes (or other random
files).

See:
- http://forum.kodi.tv/showthread.php?tid=238881
## 

The 2 booleans are just to improve readability.

PR #11937, opened 2017-04-06
[videodb] Fix clean video library: no longer removes an entire parent path accidentally

eusi:
Fix clean video library: no longer removes an entire parent path accidentally. This issue only occurred on nested folder structures (e.g. Movies/Year/Genre/...).

## Description
There are two small changes:
1.) Fixed the issue (see "Motivation and Context").
2.) Made existing code more readable, inspired by [PR8133](https://github.com/xbmc/xbmc/pull/8133)

## Motivation and Context
Whenever I delete a movie on my NAS, a video library clean on Kodi removes not only the movie from the database, but also the entire path (folder and its movies)! 

I am using a movie structure like this:
SourceVideoFolder ("Movies")
---QualityFolder (like "HD", "SD", "4k" etc.)
------YearFolder (like "1900-1999", "2000-2010" etc)
---------Movies

This happens, because in function CleanMediaType() the pathsDeleteDecisions wrongly got only the bool "del" (corresponds if the missing movie should be deleted), instead of also checking if the folder still exists. The path shouldn't get deleted if it exists.

In [PR8133](https://github.com/xbmc/xbmc/pull/8133) anaconda also ran into a similar issue, as mentioned in the comments (on Oct 25, 2015), but they didn't solve the issue. My fix should solve this.

## How Has This Been Tested?
Tested it by my own via Win32. I just deleted movies on my NAS, executed a video clean and tested if it works as expected.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/video/VideoDatabase.cpp
+++ b/xbmc/video/VideoDatabase.cpp
@@ -8574,13 +8574,26 @@ void CVideoDatabase::CleanDatabase(CGUIDialogProgressBarHandle* handle, const st
     std::string strIds;
     while (!m_pDS->eof())
     {
-      auto pathsDeleteDecision = pathsDeleteDecisions.find(m_pDS->fv(0).get_asInt());
       // Check if we have a decision for the parent path
+      auto pathsDeleteDecision = pathsDeleteDecisions.find(m_pDS->fv(0).get_asInt());
       auto pathsDeleteDecisionByParent = pathsDeleteDecisions.find(m_pDS->fv(2).get_asInt());
-      if (((pathsDeleteDecision != pathsDeleteDecisions.end() && pathsDeleteDecision->second) ||
-           (pathsDeleteDecision == pathsDeleteDecisions.end() && !CDirectory::Exists(m_pDS->fv(1).get_asString(), false))) &&
-          ((pathsDeleteDecisionByParent != pathsDeleteDecisions.end() && pathsDeleteDecisionByParent->second) ||
-           (pathsDeleteDecisionByParent == pathsDeleteDecisions.end())))
+
+      bool hasDecision = pathsDeleteDecision != pathsDeleteDecisions.end();
+      bool hasDecisionForParent = pathsDeleteDecisionByParent != pathsDeleteDecisions.end();
+      
+      std::string itemPath = m_pDS->fv(1).get_asString();
+
+      // add the path to the list of paths to be deleted if
+      //   it has been handled before and should be deleted
+      //   OR
+      //   it hasn't been handled before and doesn't exist
+      // AND
+      //   its parent path has been handled before and should be deleted
+      //   OR
+      //   its parent path hasn't been handled before
+      if (((hasDecision && pathsDeleteDecision->second) || (!hasDecision && !CDirectory::Exists(itemPath, false)))
+          &&
+          ((hasDecisionForParent && pathsDeleteDecisionByParent->second) || !hasDecisionForParent))
         strIds += m_pDS->fv(0).get_asString() + ",";
 
       m_pDS->next();
@@ -8808,8 +8821,9 @@ std::vector<int> CVideoDatabase::CleanMediaType(const std::string &mediaType, co
                !sourcePathsDeleteDecision->second.second)
         del = false;
 
+      // delete parents folder (and all its items) only, if file needs to be deleted and folder does not exist
       if (scanSettings.parent_name)
-        pathsDeleteDecisions.insert(std::make_pair(m_pDS2->fv(2).get_asInt(), del));
+        pathsDeleteDecisions.insert(std::make_pair(m_pDS2->fv(2).get_asInt(), !CDirectory::Exists(parentPath, false) && del));
     }
 
     if (del)


MartijnKaijser:
ping @Montellese

eusi:
@Montellese @MartijnKaijser
Are there any further information needed from my side?
Can we get this into next release?

xbmc/video/VideoDatabase.cpp
```diff
@@ -8474,13 +8474,28 @@ void CVideoDatabase::CleanDatabase(CGUIDialogProgressBarHandle* handle, const st
...
+      //   it hasn't been handled before and doesn't exist
+      // AND
+      //   its parent path has been handled before and should be deleted
+      //   OR
+      //   its parent path hasn't been handled before
+      if( 
+          ( ( hasDecision && pathsDeleteDecision->second ) || ( !hasDecision && !CDirectory::Exists( itemPath, false ) ) ) 
+          &&
+          ( ( hasDecisionForParent && pathsDeleteDecisionByParent->second ) || !hasDecisionForParent ) 
+        )
```
a1rwulf: It's an edgecase and matter of taste, but given our coding guidelines, I'd prefer if you keep the format of the if statement like it was before.
Normally I would suggest to move it into it's own helper function to get rid of this big statements, but given the count of parameters you would need, it won't help much.

xbmc/video/VideoDatabase.cpp
```diff
@@ -8474,13 +8474,28 @@ void CVideoDatabase::CleanDatabase(CGUIDialogProgressBarHandle* handle, const st
...
       auto pathsDeleteDecisionByParent = pathsDeleteDecisions.find(m_pDS->fv(2).get_asInt());
-      if (((pathsDeleteDecision != pathsDeleteDecisions.end() && pathsDeleteDecision->second) ||
-           (pathsDeleteDecision == pathsDeleteDecisions.end() && !CDirectory::Exists(m_pDS->fv(1).get_asString(), false))) &&
-          ((pathsDeleteDecisionByParent != pathsDeleteDecisions.end() && pathsDeleteDecisionByParent->second) ||
-           (pathsDeleteDecisionByParent == pathsDeleteDecisions.end())))
+
+      bool hasDecision = pathsDeleteDecision != pathsDeleteDecisions.end();
+      bool hasDecisionForParent = pathsDeleteDecisionByParent != pathsDeleteDecisions.end();
+      
+      std::string itemPath = m_pDS->fv( 1 ).get_asString();
```
Montellese: The spaces around the `1` are unnecessary.

xbmc/video/VideoDatabase.cpp
```diff
@@ -8474,13 +8474,28 @@ void CVideoDatabase::CleanDatabase(CGUIDialogProgressBarHandle* handle, const st
...
+      // add the path to the list of paths to be deleted if
+      //   it has been handled before and should be deleted
+      //   OR
+      //   it hasn't been handled before and doesn't exist
+      // AND
+      //   its parent path has been handled before and should be deleted
+      //   OR
+      //   its parent path hasn't been handled before
+      if( 
+          ( ( hasDecision && pathsDeleteDecision->second ) || ( !hasDecision && !CDirectory::Exists( itemPath, false ) ) )
```
Montellese: All the extra spaces around the conditions and the parameters are unnecessary.

xbmc/video/VideoDatabase.cpp
```diff
@@ -8708,8 +8723,9 @@ std::vector<int> CVideoDatabase::CleanMediaType(const std::string &mediaType, co
                !sourcePathsDeleteDecision->second.second)
         del = false;
 
+      // delete parents folder (and all its items) only, if file needs to be deleted and folder does not exist
       if (scanSettings.parent_name)
-        pathsDeleteDecisions.insert(std::make_pair(m_pDS2->fv(2).get_asInt(), del));
+        pathsDeleteDecisions.insert(std::make_pair(m_pDS2->fv(2).get_asInt(), !CDirectory::Exists( parentPath, false ) && del));
```
Montellese: This extra call to `CDirectory::Exists()` might bring back a regression we had in the past where we try to check a path that is not online which takes a lot of time. That's the main reason why we do all this caching in `pathsDeleteDecisions` and `sourcePathsDeleteDecisions`. But it's very hard to say just from reading the code. It certainly doesn't make any sense to check the parent path's existance if we have already figured out that the source the path belongs to doesn't exist anymore.

Montellese (changes requested):
You're touching a very controversial and delicate part of the code here so kudos for that. I don't think that I have all the potential use cases in my head which we try (and fail) to cover with this code which makes it very hard to review :-/

xbmc/video/VideoDatabase.cpp
eusi: I needed more than an hour to understand how this code actual work (2 months ago). And for my scenario the code does not work at all (even worse: it deletes half my library every time I delete some movies). 
In my case the variable "del" decides that an entire path should be deleted, without checking if this is necessary (if path exists or contains movies that exist) or even wanted/desired from the user. So, there is something missing...

I guess some test cases would be very helpful for this part of code...
Test: Create Folder structure:
**Movies** (root folder, that should be added to the video library in kodi)
- **HD**
-- **2000** (into this folder add like 3 movies)
Then update database (3 movies found), then delete one of them on the hdd, execute a library clean and at the end you will see, that all 3 movies were deleted from kodi (instead of the one only).

My current workaround: 
Instead of adding **Movies** to the library, I added all quality level individual: **HD, SD, 4K** etc.
Because the folder level is now decreased by one, a lib clean works.

This issue needs to be fixed anyway someday. Because I have very little time atm, I will look into it again next months, so we can close this issue until Kodi 18 release hopefully ;-)

juliocesark:
It would be good to close the issue because it is already open a long time and the code can not be revised as reported Montellese.

phunkyfish:

PR #11946, opened 2017-04-09
video player updates

FernetMenta:
issues

no known issues


FernetMenta:
@garbear please adapt RP

FernetMenta:
jenkins build this please

Rechi:
@FernetMenta 
- some renames from `DVDVideoPicture` to `VideoPicture` are missing
- changes related to the union remove in `xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodec.h` are missing in some hardware decode renderers

FernetMenta:
@Rechi platform devs are invited to adapt to the changes. The source of the PR is a beanch on team kodi repo.

xbmc/cores/VideoPlayer/DVDCodecs/DVDFactoryCodec.cpp
```diff
@@ -55,73 +44,16 @@
...
-    CLog::Log(LOGDEBUG, "FactoryCodec - Overlay: %s - Failed", pCodec->GetName());
-    delete pCodec;
-  }
-  catch(...)
-  {
-    CLog::Log(LOGERROR, "FactoryCodec - Audio: Failed with exception");
-  }
-  return nullptr;
-}
+//#include "VPFactoryCodec_override.h"
```
ace20022: on purpose?

tools/depends/target/ffmpeg/FFMPEG-VERSION
```diff
@@ -1,5 +1,5 @@
 LIBNAME=ffmpeg
 BASE_URL=https://github.com/xbmc/FFmpeg/archive
-VERSION=3.1.6-Krypton
+VERSION=release/3.3-kodi
```
Rechi: This won't work, because of the slash

xbmc/cores/VideoPlayer/DVDCodecs/DVDFactoryCodec.cpp
FernetMenta: yes, it is planned to move the platform dependent code to separate files. waiting for @fetzerch to complete the cmake part.

tools/depends/target/ffmpeg/FFMPEG-VERSION
asavah: Yep, either cmake bump specifies too high (ffmpeg master?) versions for libraries or Kodi ffmpeg repo was not updated. https://github.com/xbmc/xbmc/pull/11946/commits/ffd5da1fb74db639f582e37ba89e80b6876b0e82#diff-ec68d1dfffe81bf01733e80ff85956e1R37 says _avcodec_ver ">=57.89.100" while version here https://github.com/xbmc/FFmpeg/blob/release/3.3-kodi/libavcodec/version.h#L30 is lower.

FernetMenta: oops, typo. corrected to minor version 87

FernetMenta:
jenkins build this please

tools/depends/target/ffmpeg/FFMPEG-VERSION
asavah: You need all of them, otherwise cmake won't pick up external ffmpeg and fall back to internal build
```
set(_avcodec_ver ">=57.87.100")
set(_avfilter_ver ">=6.80.100")
set(_avformat_ver ">=57.69.100")
set(_avutil_ver ">=55.55.100")
set(_swscale_ver ">=4.4.100")
set(_swresample_ver ">=2.5.100")
set(_postproc_ver ">=54.3.100")

```

FernetMenta: ffmpeg 3.3 is required. versions are here:
https://github.com/FFmpeg/FFmpeg/commit/e1cc7f83df001b471313fb7f03fd4b0fa1c5a747

FernetMenta: and 89 was correct for codec :)

FernetMenta: I will rebase our ffmpeg fork, that should fix it.

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/YUV2RGBShader.cpp
```diff
@@ -210,22 +209,21 @@ BaseYUV2RGBGLSLShader::BaseYUV2RGBGLSLShader(bool rect, unsigned flags, ERenderF
...
+  else if (m_format == SHADER_YUY2)
     m_defines += "#define XBMC_YUY2\n";
-  else if (m_format == RENDER_FMT_UYVY422)
+  else if (m_format == SHADER_UYVY)
     m_defines += "#define XBMC_UYVY\n";
-  else if (m_format == RENDER_FMT_VDPAU_420)
+  else if (m_format == SHADER_NV12_RRG)
     m_defines += "#define XBMC_NV12_RRG\n";
-  else if (m_format == RENDER_FMT_VAAPI)
+  else if (m_format == SHADER_NV12_RRG)
```
afedchin: SHADER_NV12_RRG exist already above.

FernetMenta: thanks! updated

Rechi:
@FernetMenta be aware that `ffmpeg-release/3.3-kodi.tar.gz` doesn't get re downloaded. Jenkins slaves won't get the changes as they have already the file and don't check if something has changed in the git repo.

cmake/modules/FindFFMPEG.cmake
```diff
@@ -33,14 +33,14 @@
...
-set(REQUIRED_FFMPEG_VERSION 3.1)
-set(_avcodec_ver ">=57.48.101")
-set(_avfilter_ver ">=6.47.100")
-set(_avformat_ver ">=57.41.100")
-set(_avutil_ver ">=55.28.100")
-set(_swscale_ver ">=4.1.100")
-set(_swresample_ver ">=2.1.100")
-set(_postproc_ver ">=54.0.100")
+set(REQUIRED_FFMPEG_VERSION 3.3)
+set(_avcodec_ver ">=57.87.100")
```
Rechi: version at https://github.com/xbmc/FFmpeg/tree/release/3.3-kodi is `57.89.100`

FernetMenta: I have not dropped the "squash commit" yet

FernetMenta:
jenkins build this please

FernetMenta:
@popcornmix any idea why ffmpeg fails to build for pi?

popcornmix:
Looks like a stale build to me. I'm not sure what cleaning is done with jenkins builds.
I suspect deleting ``/home/jenkins/workspace/LINUX-RBPI/tools/depends/target/ffmpeg/raspberry-pi-debug`` and trying again may help.

FernetMenta:
I cleared current workspace
jenkins build this please

Rechi:
The issue on rpi is that configure can't find `vdpau/vdpau.h` and `vdpau/vdpau_x11.h`.

popcornmix:
Is jenkins using make or cmake for ffmpeg? This looks wrong for pi:
https://github.com/xbmc/xbmc/blob/master/tools/depends/target/ffmpeg/Makefile#L29

I think ffmpeg 3.1 would ignore this on Pi, but ffmpeg 3.3 complains.
Removing ``--enable-vdpau --enable-vaapi`` does fix the ffmpeg configure when running locally.

FernetMenta:
root cause is that linux is not a platform we build for but X11 or Mir are ones.

Rechi:
On jenkins all (except windows) use this Makefile.
Cmake is only a fallback, if kodi gets build without depends.

xbmc/cores/VideoPlayer/DVDCodecs/Video/DXVA.h
```diff
@@ -73,7 +72,8 @@ class CRenderPicture
 public:
   CRenderPicture(CSurfaceContext *context);
   ~CRenderPicture();
-  ID3D11View*      view;
+  ID3D11View* view;
+  unsigned int extFormat;
```
afedchin: Can you please change `extFormat` type to DXGI_FORMAT and rename it to `format`? then it will be correspond to my current work on this part.

FernetMenta: not sure if I get this correctly. but are are invited to do the change yourself. the source of this pr is a branch in xbmc repo. push the change to this branch. we can do cleanup/squashing later

xbmc/cores/VideoPlayer/VideoRenderers/BaseRenderer.cpp
```diff
@@ -427,6 +427,28 @@ void CBaseRenderer::ManageRenderArea()
...
+  else if (m_format == RENDER_FMT_YUV420P16)
+    ret = SHADER_YV12_16;
+  else if (m_format == RENDER_FMT_NV12)
+    ret = SHADER_NV12;
+  else if (m_format == RENDER_FMT_YUYV422)
+    ret = SHADER_YUY2;
+  else if (m_format == RENDER_FMT_UYVY422)
+    ret = SHADER_UYVY;
+  else
+    CLog::Log(LOGERROR, "CLinuxRendererGL::GetShaderFormat - unsupported format %d", renderFormat);
```
afedchin: `CLinuxRendererGL` -> `BaseRenderer`

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/WinVideoFilter.cpp
```diff
@@ -261,24 +261,22 @@ bool CYUV2RGBShader::Create(unsigned int sourceWidth, unsigned int sourceHeight,
 
   switch (fmt)
   {
-  case RENDER_FMT_YUV420P:
-  case RENDER_FMT_YUV420P10:
-  case RENDER_FMT_YUV420P16:
```
afedchin: why you drop `SHADER_YV12_10` and `SHADER_YV12_16`?

FernetMenta: thosee are not dropped: https://github.com/xbmc/xbmc/pull/11946/commits/320c5735499c1325cce004dc45279918ff2def2a

xbmc/cores/VideoPlayer/DVDCodecs/Video/DXVA.h
afedchin: got it.

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/WinVideoFilter.cpp
FernetMenta: right, this is the shader here. removed by accident.

afedchin: np. will fix by myself with `DXVA::CRenderPicture` changes

FernetMenta:
jenkins build this please

xbmc/cores/VideoPlayer/DVDCodecs/Audio/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
-set(SOURCES DVDAudioCodecFFmpeg.cpp
+set(SOURCES DVDAudioCodecAndroidMediaCodec.cpp
```
Rechi: this file doesn't exist

xbmc/cores/VideoPlayer/DVDCodecs/Audio/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
...
+            DVDAudioCodecFFmpeg.cpp
             DVDAudioCodecPassthrough.cpp)
 
 set(HEADERS DVDAudioCodec.h
+            DVDAudioCodecAndroidMediaCodec.h
```
Rechi: this file doesn't exist

stefansaraev:
@peak3d fyi, livetv on amlogic fallbacks to software decoding at random pretty often. is there something I can help with debugging before this is merged?

peak3d:
@stefansaraev a debug log file without component logging would be great.

stefansaraev:
@peak3d here is one http://sprunge.us/AMJa

playback started at
```
21:52:29.211 T:547027980784   DEBUG: CGUIInfoManager::SetCurrentMovie(pvr://channels/tv/All channels/pvr.hts_539773674.pvr)
```

hardware decoding worked ok here, switched to another channel at

```
21:52:36.631 T:547027980784   DEBUG: CGUIInfoManager::SetCurrentMovie(pvr://channels/tv/All channels/pvr.hts_1716939035.pvr)
```

and it failed back to software decoding.

it seems to happen only on channel switch. every second channel switch triggers the sw decoding fallback. I can't reproduce when stopping playback then starting next channel.

peak3d:
thanx @stefansaraev will look at it

fritsch:
@peak3d you don't implement the ```Reconfigure``` method yet. Therefore opening a new hw decoding will fail, reconfigure will also fail and result is SW decoding, right?

peak3d:
@fritsch after talking to fernet some weeks ago it should work like it is implemented right now.

If a decoder is open, and fails, then it will be retried again after the old one is close.
Returning false on Reconfigure should work.

fritsch:
``` 21:52:37.243 T:546119660016   ERROR: CDVDVideoCodecAmlogic::Open - InstanceGuard locked
21:52:37.243 T:546119660016    INFO: CVideoPlayerVideo::OpenStream - could not open video codec ```

It retries two times and then uses SW.

FernetMenta:
jenkins build this please

peak3d:
@stefansaraev can you pls confirm that live-tv is working now?

stefansaraev:
@peak3d thanks for pinging - will test this evening. even if not fully working - IMO this should not be a blocker (we are on master) and can be fixed later.

ksooo:
well, I do not like this "break it, we are on master" approach. If for instance live tv is broken on master this might hinder me massively. If it just happens, okay, but not, if we know we will break stuff. Fix shit  before merging, please.

FernetMenta:
@ksooo no worries. what you are saying is one of the reasons why I decided to do my work not on kodi master branch.

stefansaraev:
@ksooo sorry I did not mean to provoke a discussion. what I meant is: if it works on everything else but only amlogic (and it was is actualy not a crash and burn bug on amlogic) - I wont object. and yes, minor breakage on specific platform (or feature) for few days in master is acceptable. it happens all the time.

while we are at it, you can test it too and raise your voice if not working for you ;)

@peak3d yes. after 20+ channel switches here it seems to be good. can not reproduce anymore. thanks.

Rechi:
@FernetMenta please rebase

hudokkow:
and if you don't mind, rename `fmpeg_options.txt`to `ffmpeg_options.txt`

FernetMenta:
rebased and renamed ffmpeg_options

afedchin (commented):
why you removed `ClearPicture`?

peak3d:
@afedchin because of issues (double release at the end of a stream) in mediacodec I had a small discussion with @FernetMenta. The removal is the result of this duscussion.
@FernetMenta already stated in the DVDVideoCodec interface description, that buffer refcounting has to be done in getPicture.

afedchin:
removing a method because it's used wrong is not a solution in my opinion.  I used `ClearPicture` in my decoder because the decoder wants to release resources asap when picture isn't needed anymore to use resources in decoding process and not in a time when player will request a new picture.

peak3d:
I removed the methode only in case its only called by VPV (hopefully) because it was dead code.
What decoder are you talking about?

afedchin:
I'm talking about my work on mvc decoder for windows. But still isn't good if decoder can't know that a resource can bee freed

> Edit: MMAL decoder for example still has ClearPicture because it is called internally

that means nothing, because a decoder doesn't know when resources can be freed till VP request a picture.

rhocheck:
ClearPicture is a relict and not needed anymore. Decoders can clear buffer itself on desctruction

afedchin:
on a destruction of a decoder? it too late. don't you? I'm talking not about all resources but only a resource which is provided to the player with a picture. When the picture isn't needed anymore then decoder wants to release the resource asap.  I still think that if the method is used in a wrong way it's not a reason to get rid of it.

FernetMenta:
ClearPicture originally was implemented to clear (memset) the old picture before GetPicture was called. Now with alignment to ffmpeg's new api this is history. you can't invalidate the last picture because you don't know if GetPicture will result in a new frame.
With new VideoBuffers this will get better. VideoBuffers will be refcounted like ffmpeg frames and VPV will release the ref when when it does not need it anymore. No need to keep the old crap alive. Look forward, things will get better :)

afedchin:
ok. but we are using not only ffmpeg decoder, right (at least in a theory)? VideoBuffer my ref count a frame (HW picture) and may make a copy of data from a picture and in second case the resource of the picture isn't needed anymore. let's allow to decoder decide. if the buffer is ref counted somewhere else it will not be released. This is how DXVA decoder works right now with DXVA renderer.

afedchin:
just to clarify my POV. I just want release a resource asap if it's not needed anymore. just to not produce a "buffer" resources which for mvc decoder very big

FernetMenta:
you are off topic here. currently a picture has to stay valid until VPV gets a new one from decoder. for this behaviour ClearPicture is not required.

FernetMenta:
> just to clarify my POV. I just want release a resource asap if it's not needed anymore

this is exaclty how it works, see ffmpeg, vaapi, vdpau for reference

popcornmix:
@peak3d the ClearPicture change causes a leak in mmal as we no longer release when VideoPlayerVideo is destroyed. This commit should fix it:
https://github.com/popcornmix/xbmc/commit/0dd1f5f6ce66ca72d4e05908fe0018976247ac42

afedchin:
> this is exaclty how it works, see ffmpeg, vaapi, vdpau for reference

it doesn't means that this is right. don't you? VPV requests a picture from a decoder when decoder reports that it has a picture. all time between each call GetPicture the picture actually is not needed. The picture is needed only till renderer copys or refs count it. resources which are holded by a picture may be used for decoding, but can't till VPV request new picture.

FernetMenta:
> all time between each call GetPicture the picture actually is not needed

this is wrong. if decoder produces no new picture, VPV submits the same frame to renderer again and again. otherwise still frames with overlays won't work.

FernetMenta:
@popcornmix in what branch do you have the mentioned fixup commit?

popcornmix:
newclock5 (although looks like you found it).

peak3d:
@FernetMenta can we be sure that the pVideoPicture passed in GetPicture contains the same information we have passed on the previous call? This would reduce development in each decoder (save the last given memorypool as it is now in MediaCodec and in @popcornmix 's PR)

FernetMenta:
@peak3d currently VPV does not touch VideoPicture, it just passes it to renderer. I don't know if the various hw renderes touch it.

This may change slightly with VideoBuffers. Unlike what we have today a VideoBuffer is always refcounted and VPV will likely request ref and release it when no longer required. This will even allow changing bufferpools with different types on the fly

FernetMenta:
@afedchin  for clarity: buffers are owned by the creater, this won't change. the creator can release a resource at any time, VPV does not touch the content.
this is i.e. a requirement for vdpau that invalidates all buffers after a change of refresh rate.

MartijnKaijser:
jenkins build this please
Can we get this moving to master?

Rechi:
Is just removing the function call to `ClearPicture` at https://github.com/xbmc/xbmc/blob/dde5764/xbmc/cores/RetroPlayer/RetroPlayerVideo.cpp#L195 enough or does this need more changes?

FernetMenta:
Dop it, if RP breaks, it is broken. I said this many times before that RP is not supposed to use VP's internal APIs

FernetMenta:
building Pi is still broken

MartijnKaijser:
@popcornmix can you get that fixed please?

popcornmix:
I can't see the jenkins error. Is this the ``--enable-vaapi`` issue or something else?

FernetMenta:
if I am not mistaken nothing has changed in regard to the vdpau build error

Rechi:
yes still the same error.
> The issue on rpi is that configure can't find `vdpau/vdpau.h` and `vdpau/vdpau_x11.h`.

FFmpeg configure script tells `--disable-vdpau` is autodetect, but maybe not if `--enable-vdpau` is specified.

popcornmix:
This line is wrong:
https://github.com/xbmc/xbmc/blob/master/tools/depends/target/ffmpeg/Makefile#L29
It needs to be conditional on X11/VAAPI/VDPAU but I'm not sure what the right test is.

FernetMenta:
The correct test is the actual target (windowing system) we build for, i.e. X11, Mir, GBM, etc.

popcornmix:
Yes, but what make variable is that in?

FernetMenta:
I don't know

peak3d:
There are 2 runs for ffmpeg (one in builddeps, and sometimes one during kodi build.
The one on kodi build time should have access to HAS_VDPAU, but it will likely not the case if we build dependencies.

This should be evaluated by @wsnipex / @Rechi . IMO this is something wich has to be placed in an arch / platform specific configuration.

Rechi:
The kodi build ffmpeg build disables vdpau depending on `CORE_SYSTEM_NAME STREQUAL rbpi` (https://github.com/xbmc/xbmc/blob/285e1ec/tools/depends/target/ffmpeg/CMakeLists.txt#L59-L61).
According to @FernetMenta that's not the correct way, just wanted to point it out.

So we should add a configure option for `tools/depends` which sets the windowing system?
Would it be enough to check if vdpau is available (like FFmpeg is doing it) and set a make variable for it?

peak3d:
If ffmpeg uses auto-detect we maybe want to remove this line completely?

FernetMenta:
IMO auto-detect does not work properly. The result should not be dependent on libs that are installed on a build systems.

wsnipex:
we do not know if we build for X11/GBM/Mir at the depends stage, so those cannot be used.

Just filter vdpau based on TARGET_PLATFORM=raspberry-pi

wsnipex:
an alternative would be to build ffmpeg at the kodi stage instead of in depends. At this point we do know which windowing system is being targeted.

peak3d:
IMO ffmpeg should stay in depends and should removed from kodi build.

wsnipex:
It's either or. The kodi ffmpeg build is only a fallback in case we do not find a proper version on the system(or in depends). If you see ffmpeg being built twice, you have a bug somewhere in this PR.

peak3d:
It always builds twice for all my crosscompile projects, but this is nothing new.
So my cmake command always includes the -DENABLE_INTERNAL_FFMPEG=OFF option

wsnipex:
then something is broken, but it wouldn't be the first time you have strange build issues ;)

wsnipex:
jenkins build this please

FernetMenta:
jenkins build this please

MartijnKaijser:
All platforms compile now. 
http://jenkins.kodi.tv/job/BuildMulti-All/685/
Go for merge?

FernetMenta:
I'll look at it this evening or tomorrow

FernetMenta:
jenkins build this please

garbear:
I rebased retroplayer on master post-merge and noticed that https://github.com/garbear/xbmc/commit/91d2a3d0 didn't make it into master. was this intentional?

FernetMenta:
@garbear seems to got lost somewhere. was this commit in vpupdates?

garbear:
yes, after the merge I rebased my dev branches on master. There were some conflicts due to squashing, which I `git rebase --skip`'ed. After rebasing, this commit was left over, meaning it didn't make it into the final merge.

FernetMenta:
I fixed it: https://github.com/xbmc/xbmc/commit/8405bbf242a73e7b88641c7db4a9808fd46a2f47

PR #11955, opened 2017-04-12
[Linux] Generic drm/kms implementation

lrusak:
## Description

This allows kodi to run directly on drm/kms without the need for any display manager. This implementation uses GBM to create the required surfaces.

One thing I don't really like is the naming here. @FernetMenta said that the windowing implementation defines the name so that's why everything is "gbm" because that's what actually creates the surface. I think drm or kms is a better naming scheme but I'm just bikeshedding.

This is mostly feature complete:
 - [x] Resolution switching
 - [x] Refresh rate switching
 - [ ] Multi monitor support
 - [ ] Multi GPU support

I hope to implement the two unfinished features in the future.

~~Hardware accelerated decoding of video is currently only possible on Intel hardware using VAAPI.~~ Hopefully in the future other platforms will have hardware accelerated decoding.

Please comment on any and all code quality, grammer, etc.

~~I've added some commits from @MaxKellermann to allow VAAPI to be used with GLES. @FernetMenta can you please take a look at this?~~

The rest is standalone other than some minor cmake changes. ~~I would like to get some help because I don't like the changes to FindVAAPI.cmake and I feel like it could be done better.~~

Big thanks to @FernetMenta and @anholt for all the help they provided.

PS. The 3D modes are completely untested (I can probably borrow a 3D tv to test with at some point).

A LibreELEC test build is available [here](http://lrusak.libreelec.tv/public/LibreELEC-Generic.x86_64-9.0-devel-20170411181905-r26003-g3cbaf6e.tar) 

## Motivation and Context

I was asked by @janesma at ELC in portland why kodi doesn't run on drm/kms. I didn't have an answer other than "I don't know". So I took it upon myself to do it.

## How Has This Been Tested?

I've built this with LibreELEC on RPi (using vc4 driver) and x86 hardware (AMD and Intel).

Running on Intel
Imgur
Imgur

Running on AMD
Imgur

Running on RPi2
Imgur

## Types of change

## Checklist:


MaxKellermann:
Big hooray for this code!

FernetMenta:
please drop all changes related to videoplayer from this pr. they need rebase on vpupdates and should be reviewed separately

MaxKellermann:
@FernetMenta: I submitted those videoplayer changes for separate review a long time ago, and you rejected them.

MaxKellermann:
@lrusak: I have included this in my Kodi fork, but I had to force-disable OpenGL (in favor of GLES, see https://github.com/MaxKellermann/xbmc/commit/1459ec1bb362445b17013bd9f9138f39752123a1) to make it build (on regular desktop Linux, not Raspi). There are still build failures, and I'm sorting them out.

MaxKellermann:
Had to disable VDPAU as well: https://github.com/MaxKellermann/xbmc/commit/92fb38ffed90bf54e491eb24f8726d2395e31d72

lrusak:
@FernetMenta should I PR the vp related commits to the vpupdates branch?

@MaxKellermann thanks! I'll pick the commits and add them!

FernetMenta:
@lrusak PRing to a PR on mainline repo is going to get interesting. I don't think we had this before. But why not, go ahead.

lrusak:
Dropped relevant VP commits. Disabled VAAPI support for now.

cmake/scripts/linux/ArchSetup.cmake
```diff
@@ -41,6 +41,6 @@ if(CMAKE_BUILD_TYPE STREQUAL Coverage)
   set(COVERAGE_EXCLUDES */test/* lib/* */lib/*)
 endif()
 
-if(ENABLE_MIR)
+if(ENABLE_MIR OR ENABLE_GBM)
```
FernetMenta: that is rather hacky for more than one reason. you print a message with "Mir" when not building for Mir. but the real problem is the inverted logic of the build system. it arbitrarily enables vdpau and the various platforms need to disable it. vdpau does only work with X11 and should only be enabled for this platform.

xbmc/windowing/WindowingFactory.h
```diff
@@ -29,6 +29,9 @@
 #elif defined(TARGET_WINDOWS) && defined(HAS_DX)
 #include "windows/WinSystemWin32DX.h"
 
+#elif defined(TARGET_LINUX) && defined(HAVE_GBM) && defined(HAS_GLES)
```
FernetMenta: what non linux systes has gbm?

xbmc/windowing/egl/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-if (NOT MIR_FOUND AND NOT X_FOUND)
+if (NOT MIR_FOUND AND NOT X_FOUND AND NOT GBM_FOUND)
```
FernetMenta: looking forward to extend this list with wayland :)

xbmc/windowing/gbm/GBMUtils.h
```diff
@@ -0,0 +1,69 @@
+#pragma once
+
+#include <xf86drm.h>
+#include <xf86drmMode.h>
+#include <gbm.h>
+#include <vector>
+
+#include "guilib/Resolution.h"
+
+struct gbm {
```
FernetMenta: we have curly bracket on new line

xbmc/windowing/gbm/GLContextEGL.cpp
```diff
@@ -0,0 +1,186 @@
+#include "GLContextEGL.h"
```
FernetMenta: our kodi header is missing

lrusak: fixed

xbmc/windowing/WindowingFactory.h
lrusak: what non linux system has mir?

cmake/scripts/linux/ArchSetup.cmake
lrusak: I understand that the inverted logic is ugly, but I am not sure how to fix it. How can we make it so that vdpau is disabled for mir and gbm but is optional for x11? Perhaps this can be cleaned up in a future PR.

xbmc/windowing/egl/CMakeLists.txt
lrusak: Cleanup for a future PR.

xbmc/windowing/WindowingFactory.h
FernetMenta: TARGET_LINUX is wrong for all ports here. Same issue again: LINUX is not a platform we build for. The actual targets are defined by GBM, X11, Mir, etc. Those are sufficient and only a single one of those must be defined for a build.

xbmc/windowing/gbm/GBMUtils.cpp
```diff
@@ -0,0 +1,517 @@
...
+void CGBMUtils::PageFlipHandler(int fd, unsigned int frame,
+		  unsigned int sec, unsigned int usec, void *data)
+{
+  (void)fd, (void)frame, (void)sec, (void)usec;
+
+  int *flip_happening = static_cast<int *>(data);
+  *flip_happening = 0;
+}
+
+bool CGBMUtils::WaitingForFlip()
```
FernetMenta: this method never returns true

xbmc/windowing/gbm/GBMUtils.cpp
```diff
@@ -0,0 +1,517 @@
...
+  }
+
+  return true;
+}
+
+void CGBMUtils::FlipPage()
+{
+  if (WaitingForFlip())
+  {
+    return;
```
FernetMenta: dead code

xbmc/windowing/gbm/GBMUtils.cpp
```diff
@@ -0,0 +1,517 @@
...
+
+  auto ret = drmModePageFlip(m_drm->fd,
+			     m_drm->crtc_id,
+			     m_drm_fb->fb_id,
+			     DRM_MODE_PAGE_FLIP_EVENT,
+			     &flip_happening);
+
+  if (ret == -1)
+  {
+    CLog::Log(LOGDEBUG, "CGBMUtils::%s - failed to queue DRM page flip", __FUNCTION__);
```
FernetMenta: only worth a debug log? I would not expect that we safely can ignore those errors

cmake/scripts/linux/ArchSetup.cmake
wsnipex: move this to the next block where you disable vaapi instead.

xbmc/windowing/gbm/GBMUtils.cpp
lrusak: thanks, I changed the errors in case statements to return true instead of break.

xbmc/windowing/gbm/GBMUtils.cpp
lrusak: fixed now that the method can return true

xbmc/windowing/gbm/GBMUtils.cpp
lrusak: Yea, I'm not sure _why_ a page flip would fail to take place, but we want to just warn about it.

[mpv](https://github.com/mpv-player/mpv/blob/master/video/out/vo_drm.c#L354) and [retroarch](https://github.com/libretro/RetroArch/blob/master/gfx/drivers_context/drm_ctx.c#L219) do the same thing

cmake/scripts/linux/ArchSetup.cmake
lrusak: done, I will squash the commits when rebasing next.

FernetMenta:
this PR is rather self-contained. as far as I am concerned feel free to merge at any time your are comfortable with it.

MartijnKaijser:
Jenkins build this please

lrusak:
rebased and updated tabs->spaces.

lrusak:
Objection, comments, concerns?

xbmc/windowing/gbm/GLContextEGL.cpp
```diff
@@ -0,0 +1,206 @@
...
+    EGL_SAMPLES,         0,
+    EGL_SURFACE_TYPE,    EGL_WINDOW_BIT,
+    EGL_RENDERABLE_TYPE, renderable_type,
+    EGL_NONE
+  };
+
+#ifdef EGL_EXT_platform_base
+  if (m_eglDisplay == EGL_NO_DISPLAY)
+  {
+    PFNEGLGETPLATFORMDISPLAYEXTPROC getPlatformDisplayEXT = (PFNEGLGETPLATFORMDISPLAYEXTPROC)eglGetProcAddress("eglGetPlatformDisplayEXT");
```
fooishbar: Doing this is not safe without first checking the EGL client extensions list (not the usual display extensions list!) for the platform_base extension.

xbmc/windowing/gbm/GBMUtils.cpp
```diff
@@ -0,0 +1,555 @@
...
+{
+  GetMode(res);
+
+  m_drm->gbm->width = m_drm->mode->hdisplay;
+  m_drm->gbm->height = m_drm->mode->vdisplay;
+
+  m_drm->gbm->surface = gbm_surface_create(m_drm->gbm->dev,
+                                           m_drm->gbm->width,
+                                           m_drm->gbm->height,
+                                           GBM_FORMAT_ARGB8888,
```
fooishbar: GBM_FORMAT_XRGB8888

xbmc/windowing/gbm/GBMUtils.cpp
```diff
@@ -0,0 +1,555 @@
...
+  }
+
+  m_drm->gbm->surface = nullptr;
+}
+
+bool CGBMUtils::SetVideoMode(RESOLUTION_INFO res)
+{
+  GetMode(res);
+
+  gbm_surface_release_buffer(m_drm->gbm->surface, m_bo);
```
fooishbar: This is too early; you only want to release the buffer after drmModeSetCrtc has returned.

xbmc/windowing/gbm/GBMUtils.cpp
```diff
@@ -0,0 +1,555 @@
...
+
+  uint32_t width = gbm_bo_get_width(bo);
+  uint32_t height = gbm_bo_get_height(bo);
+  uint32_t stride = gbm_bo_get_stride(bo);
+  uint32_t handle = gbm_bo_get_handle(bo).u32;
+
+  auto ret = drmModeAddFB(m_drm->fd,
+                          width,
+                          height,
+                          24,
```
fooishbar: If you use drmModeAddFB2, then you can pass the format directly and avoid the hardcoded 24/32. This would also allow you to support, e.g., 10bpc formats.

xbmc/windowing/gbm/GBMUtils.cpp
```diff
@@ -0,0 +1,555 @@
...
+    CLog::Log(LOGDEBUG, "CGBMUtils::%s - failed to find preferred mode", __FUNCTION__);
+    return false;
+  }
+
+  return true;
+}
+
+drm * CGBMUtils::InitDrm()
+{
+  const char *device = "/dev/dri/card0";
```
fooishbar: You'll want to stick this in a loop, iterating all the cards until one returns you a resources pointer with at least one CRTC/encoder/connector. Otherwise you can load virtual devices (like VGEM), or inert rendering-only devices (a discrete GPU on a multi-GPU system).

xbmc/windowing/gbm/GBMUtils.cpp
```diff
@@ -0,0 +1,555 @@
...
+bool CGBMUtils::RestoreOriginalMode()
+{
+  if(!m_orig_crtc)
+  {
+    return false;
+  }
+
+  auto ret = drmModeSetCrtc(m_drm->fd,
+                            m_orig_crtc->crtc_id,
+                            m_orig_crtc->buffer_id,
```
fooishbar: The original framebuffer might be long gone by the time you get here: failure should not be fatal.

xbmc/windowing/gbm/GBMUtils.cpp
```diff
@@ -0,0 +1,555 @@
...
+
+  m_drm_encoder = nullptr;
+  m_drm_connector = nullptr;
+  m_drm_resources = nullptr;
+
+  m_drm->connector = nullptr;
+  m_drm->connector_id = 0;
+  m_drm->crtc = nullptr;
+  m_drm->crtc_id = 0;
+  m_drm->crtc_index = 0;
```
fooishbar: 0 is a valid index, and also a valid fd; maybe try -1?

xbmc/windowing/gbm/GLContextEGL.cpp
lrusak: maybe @Kwiboo can comment on this, he added it.

xbmc/windowing/gbm/GBMUtils.cpp
lrusak: @Kwiboo needs the alpha layer

xbmc/windowing/gbm/GBMUtils.cpp
lrusak: it's not fatal, it simply returns false.

xbmc/windowing/gbm/GBMUtils.cpp
lrusak: something to be added in the future when I implement a multi gpu setting

xbmc/windowing/gbm/GBMUtils.cpp
lrusak: yes, something for the future

xbmc/windowing/gbm/GBMUtils.cpp
lrusak: This line can probably just be removed. I'll have to play with it.

xbmc/windowing/gbm/GBMUtils.cpp
lrusak: yes this seems reasonable

xbmc/windowing/gbm/GBMUtils.cpp
fooishbar: @lrusak it's not just multi-GPU. if you enable the 'VGEM' (virtual GEM device, used for testing) option in the kernel, it's probably going to be /dev/dri/card0, meaning that init will fail

lrusak: I'm not really worried about that right now. This has been tested successfully on a few different platforms. Extra features can be added in the future.

xbmc/windowing/gbm/GLContextEGL.cpp
Kwiboo: A fix to check the EGL client extensions list was added in https://github.com/Kwiboo/plex-home-theater/commit/59be410bb133d2dfb14bc4f410667683d67325a5, @lrusak feel free to squash that commit on next rebase (also fixed a build issue on rockchip)

xbmc/windowing/gbm/GBMUtils.cpp
Kwiboo: rockchip needs alpha enabled to allow rendering of a a video layer behind the gui layer

MartijnKaijser:
Any status change? Needs a rebase anyway :)

pkerling:
Just as a thought: Wayland has an unstable protocol for displaying surfaces with free mode setting, [fullscreen-shell](https://cgit.freedesktop.org/wayland/wayland-protocols/tree/unstable/fullscreen-shell/fullscreen-shell-unstable-v1.xml). It is implemented by weston. Using that would allow leaving all the KMS stuff etc. to weston and still be able to set mode, framerate etc. Do you think this would be a viable alternative and provide enough features?
I'm not saying a native KMS implementation isn't cool either way, just curious :) I haven't yet decided whether to support fullscreen_shell as part of the GSoC work and am a bit unsure. It's probably overkill to support both KMS and fullscreen_shell.

lrusak:
Rebased

jenkins build and merge

MartijnKaijser:
Jenkins build this please

lrusak:
build failure unrelated

MartijnKaijser:
jenkins build this please

PR #11957, opened 2017-04-12
[Depends] Replace distribute with setuptools

mrdominuzq:
## Description
Setuptools should be installed as native instead of target.  And now setuptools depends on things that depend on setuptools. So I think it's easier to install the whl file with pip

I also added openssl because pip needs it

## Motivation and Context
We use native python to compile so setuptools belongs there

## How Has This Been Tested?
build test

## Types of change

## Checklist:


Razzeee:
Would be cool, if you could fill your PR description with a little more life. Your code/reference might speak for itself, but it's really helping us to understand things.
And please set the checkboxes, that apply.

mrdominuzq:
Yes of course :)

wsnipex:
Please explain what issue this fixes, since it seems totally wrong to me. We certainly won't include openssl in native depends.

mrdominuzq:
Pycryptodome can't find setuptools, this fixes it. Pip needs openssl.

wsnipex (changes requested):
We build pycryptodome just fine, so this PR doesn't fix a real issue. Instead it requires adding native openssl without any benefit.
If you can make it work without native openssl, it'd be ok.

mrdominuzq:
I Think That if a patch is needed to build then there's a issue :) but minor detail. Another issue is the circular dependencies. Newer setuotools depends on pyparsing and pyparsing depends on setuptools. I will look into installation without openssl :)

Razzeee:
@mrdominuzq Did you come up with something?

mrdominuzq:
yes just didnt have the time. new job and all that :)

will see if i can do something today.

mrdominuzq:
opened a new PR #12804 
The other branch got fucked :P

PR #11964, opened 2017-04-13
[EPG] Get rid of epg container static singleton - hold it at service manager instead.

ksooo:
@FernetMenta mind doing the review.


FernetMenta (approved):
@ksooo looks good so far. Did you consider moving the entire egp component under pvr? I doubt that it will ever be used independently from pvr.

ksooo:
> Did you consider moving the entire egp component under pvr? I doubt that it will ever be used independently from pvr.

Yes, I thought about this and was hoping you would suggest this. So, the idea would be to make it like clients, timers etc, accessible via CServiceBrokerr::GetPVRMananager().GetEpgContainer(). I would also move all code from xbmc/epg to xbmc/pvr/epg. Sounds like a good plan?

FernetMenta:
Yes, sounds like a good plan.

ksooo:
I will come up with a PR that implements the new approach, making epg a subcomponent of pvr.

PR #11967, opened 2017-04-14
favourites refactor

tamland:
As discussed in #10881 this refactors CFavouritesDirectory to keep favourites in memory such that we can use it for context menus without having to load it from disk (on main thread) every time.


tamland:
jenkins build this please

xbmc/addons/AddonInstaller.cpp
```diff
@@ -882,5 +882,5 @@ void CAddonUnInstallJob::ClearFavourites()
   }
 
   if (bSave)
-    CFavouritesDirectory::Save(items);
+    CServiceBroker::GetFavouritesService().GetAll(items);
```
ksooo: Shouldn't this read <code>CServiceBroker::GetFavouritesService().Save(items);</code>?

xbmc/favourites/FavouritesService.cpp
```diff
@@ -0,0 +1,228 @@
...
+          item->SetArt("thumb", thumb);
+        items.Add(item);
+      }
+    }
+    favourite = favourite->NextSiblingElement("favourite");
+  }
+  return true;
+}
+
+CFavouritesService::CFavouritesService(std::string userDataFolder) : m_userDataFolder(std::move(userDataFolder))
```
ksooo: Are you passing <code>userDataFolder</code> by value by intention here?

xbmc/favourites/FavouritesService.cpp
```diff
@@ -0,0 +1,228 @@
...
+{
+  CFileItemList items;
+  std::string favourites = "special://xbmc/system/favourites.xml";
```
ksooo: Performance: standard ctor + assignment vs. copy ctor ?

xbmc/favourites/FavouritesService.cpp
```diff
@@ -0,0 +1,228 @@
...
+  favourites = URIUtils::AddFileToFolder(m_userDataFolder, "favourites.xml");
+  if(XFILE::CFile::Exists(favourites))
+    LoadFromFile(favourites, m_favourites);
+  else
+    CLog::Log(LOGDEBUG, "CFavourites::Load - no userdata favourites found, skipping");
+}
+
+CFavouritesService::~CFavouritesService()
+{
+}
```
ksooo: if the dtor is empty, I would declare it as <code>default</code> in the header file.

xbmc/favourites/FavouritesService.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+bool CFavouritesService::Persist()
+{
+  CXBMCTinyXML doc;
+  TiXmlElement xmlRootElement("favourites");
+  TiXmlNode *rootNode = doc.InsertEndChild(xmlRootElement);
+  if (!rootNode)
+    return false;
+
+  for (int i = 0; i < m_favourites.Size(); i++)
```
ksooo: Nowadays you can use std container iteration to iterate CFileItemList.

xbmc/favourites/FavouritesService.cpp
```diff
@@ -0,0 +1,228 @@
...
+    if (item->HasArt("thumb"))
+      favNode.SetAttribute("thumb", item->GetArt("thumb").c_str());
+
+    const CURL url(item->GetPath());
+    TiXmlText execute(CURL::Decode(url.GetHostName()));
+    favNode.InsertEndChild(execute);
+    rootNode->InsertEndChild(favNode);
+  }
+
+  auto path = URIUtils::AddFileToFolder(m_userDataFolder, "favourites.xml");
```
ksooo: I would not use <code>auto</code> for trivial types like std::string (and I would use copy ctor, see my note above).

xbmc/favourites/FavouritesService.cpp
```diff
@@ -0,0 +1,228 @@
...
+  { // create our new favourite item
+    const CFileItemPtr favourite(std::make_shared<CFileItem>(item.GetLabel()));
+    if (item.GetLabel().empty())
+      favourite->SetLabel(CUtil::GetTitleFromPath(item.GetPath(), item.m_bIsFolder));
+    favourite->SetArt("thumb", item.GetArt("thumb"));
+    favourite->SetPath(url.Get());
+    m_favourites.Add(favourite);
+  }
+  Persist();
+  lock.Leave();
```
ksooo: In other places in this file, a scoped <code>CSingleLock</code> instance is used, which I find more elegant and straight forward to read.

xbmc/favourites/FavouritesService.cpp
```diff
@@ -0,0 +1,228 @@
...
+  OnUpdated();
+  return true;
+}
+
+bool CFavouritesService::IsFavourited(const CFileItem& item, int contextWindow) const
+{
+  CSingleLock lock(m_criticalSection);
+  CURL url;
+  url.SetProtocol("favourites");
+  url.SetHostName(CURL::Encode(GetExecutePath(item, contextWindow)));
```
ksooo: It could make sense to factor out the three lines above as exactly this code sequence is used in <code>CFavouritesService::AddOrRemove</code>.

xbmc/favourites/GUIDialogFavourites.cpp
```diff
@@ -37,8 +38,9 @@ using namespace XFILE;
 
 #define FAVOURITES_LIST 450
 
-CGUIDialogFavourites::CGUIDialogFavourites(void)
-    : CGUIDialog(WINDOW_DIALOG_FAVOURITES, "DialogFavourites.xml")
+CGUIDialogFavourites::CGUIDialogFavourites(void) :
```
ksooo: I personally really dislike <code>void</code> in the parameter's declaration. Is there any good reason for doing this?

xbmc/filesystem/FavouritesDirectory.cpp
```diff
@@ -22,219 +22,31 @@
...
-#include "video/VideoInfoTag.h"
 #include "music/tags/MusicInfoTag.h"
-#include "URL.h"
 
 namespace XFILE
 {
 
+CFavouritesDirectory::CFavouritesDirectory()
+{
+}
```
ksooo: See my other comment. I would declare the ctor as <code>default</code> if the definition is actually emoty,

ksooo (commented):
Very nice, a view comments, though.

xbmc/favourites/FavouritesService.cpp
garbear: the assignment is generally elided and just the copy constructor is called

ksooo: I know, but there is no guarantee that the compiler actually does this. So, why not coding the copy construction explicitly? Then, there is guarantee. ;-)

xbmc/addons/AddonInstaller.cpp
tamland: good catch!

xbmc/favourites/FavouritesService.cpp
tamland: Yes. afaik by value is the most optimal for constructors (but it should be totally irrelevant here since this class is instantiated one time)

xbmc/favourites/FavouritesService.cpp
tamland: It says 'required' (first section): http://en.cppreference.com/w/cpp/language/copy_elision

xbmc/favourites/FavouritesService.cpp
tamland: I prefer auto everywhere, so I guess it's a tab v spaces thing..

xbmc/favourites/FavouritesService.cpp
ksooo: Ah, okay then. I did not know this. Thanks for clarifying.

tamland:
Think I covered *most* things now. Hopefully this is ok?

xbmc/favourites/FavouritesService.cpp
ksooo: auto everywhere? Doesn't  this  decrease code readability to s scripting language level? Hmm.

tamland: According to [others' opinions](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es11-use-auto-to-avoid-redundant-repetition-of-type-names) it "increases readability";)

ksooo:
Sure, I was only commenting, I did actually not request changes. Most of the comments were about personal preferences and optimizations anyway. ;-)

xbmc/favourites/FavouritesService.cpp
ksooo: As you said, there is no wrong and right here.

xbmc/favourites/FavouritesService.cpp
```diff
@@ -0,0 +1,226 @@
...
+    m_favourites.Copy(items);
+    Persist();
+  }
+  OnUpdated();
+  return true;
+}
+
+void CFavouritesService::OnUpdated()
+{
+  ANNOUNCEMENT::CAnnouncementManager::GetInstance().Announce(ANNOUNCEMENT::GUI, "xbmc", "OnFavouritesUpdated");
```
xhaggi: should we use `CServiceBroker::GetAnnouncementManager()`?

razzeee (commented):
LGTM

tamland:
jenkins build this please

xbmc/favourites/FavouritesService.cpp
tamland: I'll rather remove the whole dependency on announcement manager (and I will, in a followup pr as this is moved code)

ksooo: I introduced this dep. Purpose is to notify the directory provider that the favourites home screen widget needs to be refreshed.

=> https://github.com/xbmc/xbmc/blob/master/xbmc/listproviders/DirectoryProvider.cpp#L258

tamland: I meant *replace* ofc. Since favourites is a service now with its own lifetime, it can publish its own events using CEventStream

ksooo: Sure. Just wanted to give the background, so that it is clear that the functionality is needed.

DaveTBlake:
This PR has broken the JSON API `Favourites.GetFavourites` method -  "type", "window" and "windowparameters" are no longer correctly retruned. @tamland can you look into this?

The issue is related to this 
https://github.com/xbmc/xbmc/pull/11967/files#diff-4dd17326f32d07c71f9a7e00f25af0ddR64
replacing
https://github.com/xbmc/xbmc/blob/Krypton/xbmc/filesystem/FavouritesDirectory.cpp#L108

DaveTBlake:
Ping @tamland. This PR has broken the JSON API Favourites.GetFavourites method - "type", "window" and "windowparameters" are no longer correctly retruned. Can you look into this?

tamland:
Wow this API is nasty as hell. It's accessing internal details of how favorites are stored, no wonder it's breaking..

DaveTBlake:
Other JSON calls can use vfs path, so there is some consistency in "internal details" being used in the API, nut all done before my time. To be honest all I know is that it no longer works as expected and needs fixing for v18.  

Any reason path can't be like `ActivateWindow(10502,"musicdb://albums/2683/",return)` any more?
Not seeing why punctuation has to be in ascii.

tamland:
The encoded favourites:// path is the correct representations. If you pass it to the favourites service, it knows what to do. The leaky part is attempting to decode and extract information about "types" and "windows" from this uri.

Anyways does https://github.com/tamland/xbmc/commit/c3051d02c18ec80520df9af23e891489cfb6e218 fix it? I'm not able to reproduce this encoded favourites:// path. It's just empty here..

DaveTBlake:
Yes, just looking at some limited JSON output that seems to fix it, producing same JSON results as we had for Krypton. Raise a PR and ping me and I will do further testing to be sure.

The encoded favorites path was what I found in debug, not JSON output, sorry if it confued things. The JSON breakage was such that calling

```
{"jsonrpc": "2.0", "id": 0, "method": "Favourites.GetFavourites", 
"params":{"properties":["window","path","thumbnail","windowparameter"]}}
```
resulted in some properties not showing at all.

PR #11969, opened 2017-04-14
[Estuary] [PVR] Add Season/Episode to recordings

metaron-uk:
Adds season/episode display to Estuary skin for PVR recordings.
Scaled back version of a more ambitious change in https://github.com/xbmc/xbmc/pull/11783 which started to look like a step too far.

## Description
Adds season/episode display to Estuary skin in the \<season\>x\<episode\> format for recordings when provided by the PVR backend (or Videos database) and the previous behaviour when it doesn't.

Coded as a single $VAR so the format is consistent across all applications and easier to maintain/change

## Motivation and Context
Allows you to see at a glance if you have recorded all the episodes in a series, or if you are missing some.
(Sort your recordings by 'File' to see them in season/episode order)

## How Has This Been Tested?
Tested using current master on RPi (LibreElec Millhouse test builds) and on x86 against my mythtv backend.

Example recorded Series folder
screenshot070

Example recorded Series folder without episode names
screenshot071

Example movies (no episode name or season/episode information)
screenshot072

Example showing a mixture
screenshot073

Example showing 'recently added episodes' using the standard season/episode format
screenshot074

Similar format is used in the TV Series section, but this is hard coded (I haven't worked out where yet!)
screenshot075

## Types of change

## Checklist:

pings: @ksooo, @ronie @phil65

### DIFF
--- a/addons/skin.estuary/xml/Variables.xml
+++ b/addons/skin.estuary/xml/Variables.xml
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <includes>
+        <variable name="SeasonEpisode">
+	        <value>$INFO[ListItem.Season,,[COLOR grey]x[/COLOR]]$INFO[ListItem.Episode]</value>
+	</variable>
 	<variable name="PVRStatusImageVar">
 		<value condition="ListItem.IsRecording">windows/pvr/record.png</value>
 		<value condition="ListItem.HasTimer | ListItem.HasTimerSchedule">windows/pvr/timer.png</value>
@@ -8,7 +11,7 @@
 		<value condition="System.HasAddon(plugin.program.autocompletion) + !System.HasHiddenInput">plugin://plugin.program.autocompletion?info=autocomplete&amp;&amp;id=$INFO[Control.GetLabel(312).index(1)]&amp;&amp;limit=9</value>
 	</variable>
 	<variable name="PlaylistLabelVar">
-		<value condition="String.IsEqual(ListItem.DbType,episode)">$INFO[ListItem.TVShowtitle,,: ]$INFO[ListItem.Season,,x]$INFO[ListItem.Episode,,. ]$INFO[ListItem.Title]</value>
+		<value condition="String.IsEqual(ListItem.DbType,episode)">$INFO[ListItem.TVShowtitle,,: ]$VAR[SeasonEpisode,,. ]$INFO[ListItem.Title]</value>
 		<value condition="String.IsEqual(ListItem.DbType,musicvideo)">$INFO[ListItem.Artist,, - ]$INFO[ListItem.Title]</value>
 		<value>$INFO[ListItem.Label]</value>
 	</variable>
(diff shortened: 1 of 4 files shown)


Razzeee:
Thanks for your contribution and sorry for the late reply.
@phil65 and @ksooo for review.

addons/skin.estuary/xml/Variables.xml
```diff
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <includes>
+        <variable name="SeasonEpisode">
+	        <value>$INFO[ListItem.Season,,[COLOR grey]x[/COLOR]]$INFO[ListItem.Episode]</value>
```
phil65: why do you add a variable here? Variables only make sense if they contain several values.

metaron-uk:
I felt there were several other places that having season/episode information would/might be useful in future and having a variable seemed it would make this easier (simpler to code).
That's just my opinion however and getting season/episode information for PVR recordings is ultimately more important. I'll update the patch to not use a variable if that would make it acceptable for merge.

addons/skin.estuary/xml/Variables.xml
metaron-uk: OK I tried reverting to a scaled back version without using variables, but things get very complicated when you consider that EpisodeName, Season and Episode are all optional fields and you need a space separating Title or EpisodeName from the 'SeasonEpisode' combination.

Without a variable you have to do odd things which leave a trailing space in the list at all times and make the xml code hard to understand / maintain (something @ronie pointed out against one of my previous attempts)
<code>$INFO[ListItem.Label,, ]$INFO[ListItem.EpisodeName,(,) ]$INFO[ListItem.Season,,x]$INFO[ListItem.Episode,,]</code>

Otherwise you end up with no space shown between the title and episode for specials
HumansS1
Humans 1x2
Battlestar Galactical (Razor)S2
Battlestar Galactical (33) 1x1

Using a variable makes the code easier to read and the spacing just works:
<code><label>$INFO[ListItem.Label]$INFO[ListItem.EpisodeName, (,)]$VAR[SeasonEpisode, ,]</label></code>

Humans S1
Humans 1x2
Battlestar Galactical (Razor) S2
Battlestar Galactical (33) 1x1

I'm not really a skinner, so if there's a better way to do this without using variables please educate me!

HitcherUK:
I don't understand the whole PVR setup but wouldn't this be easier if SSxEE was injected into the label before it gets to the skin?

phil65:
Yeah I already suggested some time ago to do all formatting by core so that skinners only need to use listitem.label. Not sure if @metaron-uk was involved in that discussion.

ksooo:
As I said back then doing this in the core is fine. The only "problem" with this is that I cannot motivate myself to take an active stake in this story (I'm personally not very interested in all this season / episode stuff, tbh) and as I am (almost) alone when it comes to PVR core changes ... this is not progressing well, atm.

metaron-uk:
I did have a go at putting this into core (see  #11783 ) but the relatively 'easy' approach to set listitem.label I took unearthed a problem with the way the existing code was structured which I don't have the knowledge/skills/time to fix.

A compromise which replaces the variable I propose here with listitem.SeasonEpisode would be relatively easy to do (without needing to modify the code currently deemed unacceptable without a major re-write) but that still doesn't solve the desire to have a listitem.label include 'sane' defaults.

Unlike @ksooo, for me season/episode it pretty vital (it's essential for one of my primary use cases - recording episodes whenever they happen to be shown and then watching them in order / spotting when one is missing). 

This seemed like a good compromise until someone turns up with the knowledge and time to create a more maintainable FileItem class which @FernetMenta  + @ksooo would be prepared to sign off on.
I happen to agree with them by the way that FileItem is a mess and needs to be re-written, I just don't know enough to come up with a better solution myself :-(

jjd-uk:
Not sure who else might have the skills to do a core implementation if that's the preferred route, maybe @xhaggi he had been busy with real life so very little time to work on stuff but seems a bit more active recently so perhaps he can give guidance.

phunkyfish:

PR #11970, opened 2017-04-14
FIX: [droid;leanback] Various fixes

koying:
- Fixes Notifications not cleared
- Add xbmc_env parameters for refresh frequency
- Properly only instanciate XBMCJsonRPC if leanback is present
- Move initial refresh to Create rather than Resume


MartijnKaijser:
jenkins build and merge

koying:
`checking build system compiler gcc... ../libcrypto.so: file not recognized: File truncated` for aarch64

Same issue as before. Probably needs a workspace wipe

MartijnKaijser:
Jenkins build this please

PR #11958, opened 2017-04-12
[Estuary][PVR] Guide window: Add possibility to switch between horizontal and vertical epg grid view.

ksooo:
This PR enhances Estuary's PVR Guide window with the feature to be able to choose between two epg grid views - 'vertical' (already present before this PR) and 'horizontal' (new).

screenshot000
screenshot001
screenshot002

The changes have been runtime tested on macOS, latest Kodi master.

@phil65 @ronie might be interested in taking a look at the skin changes.

PR #11972, opened 2017-04-15
[EPG] Guide window: fix progress indicator rendering.

ksooo:
Spotted this after merging new horizontal epg grid layout (#11958), but this bug was there forever - just not as easy to spot with the vertical grid layout.

Pictures to illustrate the problem:

screenshot000
screenshot001

The change was runtime-tested on macOS, latest Kodi master.

@Jalle19 for code review?


ksooo:
jenkins build this please

PR #11978, opened 2017-04-16
ADD: [droid] speech-to-text in keyboard

koying:
## Description
Add speech-to-text in soft keyboard.
Triggered by voice button on Android TV, or `<menu>` on other droid flavours (if working)

## Types of change

### DIFF
--- a/xbmc/platform/android/activity/XBMCApp.cpp
+++ b/xbmc/platform/android/activity/XBMCApp.cpp
@@ -113,6 +113,7 @@ CCriticalSection CXBMCApp::m_applicationsMutex;
 std::vector<androidPackage> CXBMCApp::m_applications;
 CVideoSyncAndroid* CXBMCApp::m_syncImpl = NULL;
 CEvent CXBMCApp::m_vsyncEvent;
+std::vector<CActivityResultEvent*> CXBMCApp::m_activityResultEvents;
 
 
 CXBMCApp::CXBMCApp(ANativeActivity* nativeActivity)
@@ -865,6 +866,36 @@ void CXBMCApp::onNewIntent(CJNIIntent intent)
       CApplicationMessenger::GetInstance().PostMsg(TMSG_GUI_ACTIVATE_WINDOW, WINDOW_MUSIC_NAV, 0, nullptr, "", params);
     }
   }
+}
+
+void CXBMCApp::onActivityResult(int requestCode, int resultCode, CJNIIntent resultData)
+{
+  for (auto it = m_activityResultEvents.begin(); it != m_activityResultEvents.end(); ++it)
+  {
+    if ((*it)->GetRequestCode() == requestCode)
+    {
+      m_activityResultEvents.erase(it);
+      (*it)->SetResultCode(resultCode);
+      (*it)->SetResultData(resultData);
+      (*it)->Set();
+      break;
+    }
+  }
+}
+
+int CXBMCApp::WaitForActivityResult(const CJNIIntent &intent, int requestCode, CJNIIntent &result)
+{
+  int ret = 0;
+  CActivityResultEvent* event = new CActivityResultEvent(requestCode);
+  m_activityResultEvents.push_back(event);
+  startActivityForResult(intent, requestCode);
+  if (event->Wait())
+  {
+    result = event->GetResultData();
+    ret = event->GetResultCode();
+  }
+  delete event;
+  return ret;
 }
 
 void CXBMCApp::onVolumeChanged(int volume)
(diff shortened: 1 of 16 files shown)


MartijnKaijser:
Tested and works

xbmc/platform/android/activity/XBMCApp.cpp
```diff
@@ -867,6 +890,36 @@ void CXBMCApp::onNewIntent(CJNIIntent intent)
...
+      (*it)->Set();
+      break;
+    }
+  }
+}
+
+int CXBMCApp::WaitForActivityResult(const CJNIIntent &intent, int requestCode, CJNIIntent &result)
+{
+  int ret = 0;
+  CActivityResultEvent* event = new CActivityResultEvent(requestCode);
```
garbear: this variable can be placed on the stack:

```c++
CActivityResultEvent event(requestCode);
```

garbear:
Can we move the android code out of the generic keyboard class? Other than that, looks good to me

xbmc/platform/android/activity/XBMCApp.cpp
koying: Well, it could, but would result in copies when pushing/pulling to/from the vector, which is especially not free when manipulating jni objects..
Any specific reason you'd prefer stack?

koying:
> Can we move the android code out of the generic keyboard class?

What are you suggesting?
Keep in mind that all platforms have some form of speech-to-text that could be implemented.

garbear:
I suggest a new voice recognition abstraction with an android implementation

da-anda:
I have to agree that it makes sense to have voice recognition independet from the keyboard, as we might not want to show they keyboard everywhere voice commands might be nice to have in future (filtering/searching in list views, where the keyboard only should show up when "activating" the input field f.e.)

koying:
A keyboard abstraction makes plenty of sense if it (finally) allows an interface for native on-screen keyboards, but that is another scope for another time.

If it's just to not have 10 lines of ifdef'ed code, and that I have to have a README to find back my code, sorry but YOLO ;)

@da-anda This does speech-to-text, not voice command, and we always show the keyboard whenever text is to be entered. Furthermore, only relying on speech-to-text, without a keyboard to fix errors and have the final "OK" button is probably not a good idea.

xbmc/platform/android/activity/XBMCApp.cpp
```diff
@@ -867,6 +890,36 @@ void CXBMCApp::onNewIntent(CJNIIntent intent)
...
+  m_activityResultEvents.push_back(event);
```
tamland: I'm assuming there's some multithreading going on here, but m_activityResultEvents is not protected. Is this safe?

xbmc/platform/android/activity/XBMCApp.cpp
tamland: The real question is why use `new`, which has been deprecated for 6 years;)

xbmc/platform/android/activity/XBMCApp.cpp
koying: There is no actual multithreading happening, as Kodi is blocked waiting for the OS response.
I coded it that way for generalization and future improvements, but indeed, not thread-safe currently

xbmc/platform/android/activity/XBMCApp.cpp
koying: `new` has been deprecated? That's "new" to me ;)
Reference ?

Genuinely, what replaces it, then?

tamland: Yes, the "new" recommended way is std::unique_ptr;)

koying: You mean `std::unique_ptr<CActivityResultEvent>(` **new** ` CActivityResultEvent(requestCode))`;)

Guilty as charged, I'm an old school guy willing and able to manage the lifecycle of my objects by myself :)

tamland: I should probably have said "manually managing memory using new and delete has been deprecated for 6 year" (because nobody is that good), I stand corrected;)

koying: Well, still waiting for a reference that it has been "deprecated".
You're obviously welcome to use the wrappers if you like to, or push for a code guideline enforcement ;)

For the record, I'm the one who moved Kodi to c++11 just 2 years ago (https://github.com/xbmc/xbmc/commit/d3c46d246d932849ffb6641427b1278a9d4cd75b) so I'm not a complete dinosaur :)

MrMC: std::unique_ptr is the new new :)

MartijnKaijser:
jenkins build and merge

PR #11980, opened 2017-04-17
[addons] make versioncheck optional

tamland:
This isn't required to run and should be allowed to disable; if you want.


MartijnKaijser:
Is it still enabled by default?

tamland:
Yes, still enabled by default

PR #11983, opened 2017-04-17
[PVR] Fix 'Switch to channel' context menu action after #11747# - a b…

ksooo:
…etter approach (do not create a playlist for Livr TV).

Better than restoring previous behavior to create a playlist when selecting the "Switch to channel" context menu item is to not create a playlist for Live TV. We do not do this anywhere else.

This change was runtime tested on macOS, latest Kodi master.

@dgam1 fyi.


ksooo:
jenkins build this please

FernetMenta:
we should require a pvr test for every player related change

ksooo:
> we should require a pvr test for every player related change

sigh.

ksooo:
jenkins build this please

PR #11986, opened 2017-04-18
[Linux] EGL split up and cleanup

lrusak:
## Description

windowing/egl is a mess, let's split it up and clean it up. This basically just splits all existing code into a separate directory. This isn't meant to change the code in any way, only how it is organized. The code can be cleaned up later (which it needs IMO).

 - [x] RPi
 - [x] Android
 - [x] Amlogic
 - [ ] IMX6

My only concern here is the git history and how to properly preserve it.

Mainly I would like to preserve the EGLNativeTypeRaspberryPI.h -> RPIUtils.h change as github picks up the EGLNativeTypeRaspberryPI.cpp -> RPIUtils.c change automatically.
I'll see what I can do about this.

ping @popcornmix and @FernetMenta for review.

RPi is the most complicated of the bunch so I took that on first. The other targets should be relatively easy though I can only test amlogic (I have no ~~android~~ or imx6 devices available). If someone else wants to take one on that's great, otherwise I'll just try and work through it.

## Motivation and Context

windowing/egl is a confusing mess and should die in a fire

## How Has This Been Tested?

built on LibreELEC and tested to match all existing functionality.

## Types of change

## Checklist:


lrusak:
Added amlogic to the clean up. Tested on my wetek hub and seems to work fine.

In progress android cleanup here https://github.com/lrusak/xbmc/tree/egl-cleanup-android
I'm just working on setting up the android dev environment to test it.

hopefully imx6 after that.

hudokkow:
If you rename/move files around you'll lose (GitHub) history. There's no way around it currently. That happens because GitHub doesn't use `git log --follow`.

However, you can still take advantage of git blame if moving the file and changing its content is done in two separate commits. I suggest moving first. See https://github.com/xbmc/xbmc/pull/10446 for reference.

Right now you are losing everything. Test `Blame` and `History` in https://github.com/lrusak/xbmc/blob/a0353135e4576c70c32312ceee3f08e827462746/xbmc/windowing/amlogic/VideoSyncAML.cpp and you'll notice only one commit is shown.

For comparison, https://github.com/xbmc/xbmc/blob/master/CMakeLists.txt `History`only shows commits after the move to root but `Blame` shows all the commits.

Mettbrot:
Sorry to chime in, but this was a problem some time ago here https://github.com/xbmc/xbmc/pull/5481#issuecomment-67285571
I wonder what the status of this "feature" is at github 2.5 years later?

lrusak:
@hudokkow if you look at the commit changes at a whole you will see that github picks up the move and shows
`xbmc/windowing/egl/VideoSyncAML.cpp → xbmc/windowing/amlogic/VideoSyncAML.cpp`
quite a few files do this. Hopefully I can get all relevant files to show up like this.

stefansaraev:
github shows file renames as a "whole" just in this pr. this is not what we care about in general.

you should not delete and re-add renamed files in two commits. do everything in one (rename + your changes) commit per platform and you are done with it. git log --follow will work. git blame will work ;)

lrusak:
@stefansaraev thanks for the explanation. I'll leave the commits separate for now as it's easier for review, I can squash when ready. :+1:

FernetMenta:
+1
even if this broke history completly, it would be better than what we have currently

lrusak:
Pushed changes for android, It builds ok but I'm not sure if it runs yet because I have issues installing it on my Nexus 4.

lrusak:
Android isn't working as it crashes upon launch but I am having trouble debugging. Maybe someone can help me here.

via logcat all I see is
```
04-24 17:37:12.593   193   193 F DEBUG   : backtrace:
04-24 17:37:12.593   193   193 F DEBUG   :     #00 pc 024eccc6  /data/app/org.xbmc.kodi-1/lib/arm/libkodi.so
04-24 17:37:12.593   193   193 F DEBUG   :     #01 pc 01262ff8  /data/app/org.xbmc.kodi-1/lib/arm/libkodi.so (CAndroidUtils::CAndroidUtils()+88)
04-24 17:37:12.593   193   193 F DEBUG   :     #02 pc 0125f5ec  /data/app/org.xbmc.kodi-1/lib/arm/libkodi.so (CWinSystemAndroid::CWinSystemAndroid()+128)
04-24 17:37:12.593   193   193 F DEBUG   :     #03 pc 0071e6cc  /data/app/org.xbmc.kodi-1/lib/arm/libkodi.so (xbmcutil::GlobalsSingleton<CWinSystemAndroidGLESContext>::getInstance()+68)
04-24 17:37:12.593   193   193 F DEBUG   :     #04 pc 0071d3a4  /data/app/org.xbmc.kodi-1/lib/arm/libkodi.so
```

lrusak:
pushed a commit that to get android working. :+1: 

Now to move on to imx6.

lrusak:
Added imx6. It builds but it won't be working correctly yet as I'm not sure how to handle some stuff.

I also don't have any imx6 hardware to test with. Is it true that the entire display system has to be destroyed in order to change display modes? Can we just drop imx6 and move it to drm/kms? ;)

lrusak:
If I can get an ok by the platform devs for android, amlogic, and rpi, I can drop imx6 and leave that in the windowing/egl directory for now until I'm able to get a device and test the changes.

lrusak:
I've dropped the imx6 commits.

I consider this good to go unless we want to wait until imx6 can be included as well.

Platform devs please looks over:
Android: @koying @peak3d 
Amlogic: @peak3d @codesnake 
RPi: @popcornmix

lrusak:
note that `GLContextEGL.cpp` and `GLContextEGL.h` are the same for all three platforms. This file can be broken out and put into it's own egl folder after this PR.

xbmc/windowing/rpi/WinSystemRpi.cpp
```diff
@@ -0,0 +1,242 @@
...
+  m_nativeDisplay = EGL_NO_DISPLAY;
+  m_nativeWindow = EGL_NO_SURFACE;
+
+  m_displayWidth = 0;
+  m_displayHeight = 0;
+
+  m_stereo_mode = RENDER_STEREO_MODE_OFF;
+  m_delayDispReset = false;
+
+  m_rpi = new CRPIUtils();
```
popcornmix: m_rpi should be deleted in destructor

lrusak: thanks for the catch! fixed.

xbmc/windowing/rpi/WinSystemRpi.cpp
```diff
@@ -58,6 +58,9 @@ CWinSystemRpi::~CWinSystemRpi()
   {
     m_nativeWindow = nullptr;
   }
+
+  m_rpi = nullptr;
+  delete m_rpi;
```
popcornmix: I think these lines need to be swapped.

popcornmix:
+1 from me

jenkins build this please

koying:
Works for me.
+1

lrusak:
What is the consensus? Ok to merge before imx6 cleanup is finished?

Would still be nice to get an ok from an amlogic person @stefansaraev @peak3d @codesnake

stefansaraev:
worked for me last time I tested it. will re-test in 20 min and update this comment

MartijnKaijser:
jenkins build this please

PR #10892, opened 2016-11-08
Moved party mode, set default and clear default to new contextmenu

razzeee:
This is a weird one.

Everything seems to work, but not really. The new actions show up fine in the places where they should be, but:
- Partymode doesn't seem to do anything for me (same with master)
- Set/Clear default doesn't seem to be supported by estuary/confluence (yes I know we talked about this)

But well moved stuff to the new structure.

PR #11987, opened 2017-04-19
[music] Fix and optimize C(Set|Clear)Default:IsVisible

ksooo:
Fixes a regession introduced by #10892.
Details: https://github.com/xbmc/xbmc/pull/10892#issuecomment-293921723 ff

@Razzeee for code review and runtime test, please.
@tamland fyi.

### DIFF
--- a/xbmc/music/ContextMenus.cpp
+++ b/xbmc/music/ContextMenus.cpp
@@ -19,21 +19,36 @@
  */
 
 #include "ContextMenus.h"
-#include "guilib/GUIWindowManager.h"
 #include "music/dialogs/GUIDialogMusicInfo.h"
 #include "tags/MusicInfoTag.h"
 #include "PartyModeManager.h"
 #include "settings/Settings.h"
 #include "utils/LegacyPathTranslation.h"
+#include "utils/StringUtils.h"
+#include "utils/URIUtils.h"
 #include "utils/log.h"
-#include "filesystem/MusicDatabaseDirectory/DirectoryNode.h"
-#include "Util.h"
 #include "filesystem/MusicDatabaseDirectory.h"
 #include "ServiceBroker.h"
 
 
 namespace CONTEXTMENU
 {
+
+static bool IsMusicLibraryItem(const CFileItem& item)
+{
+  if (!item.IsParentFolder() && URIUtils::IsMusicLibraryContent(item.GetPath()))
+  {
+    XFILE::CMusicDatabaseDirectory dir;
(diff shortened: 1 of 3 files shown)


Razzeee:
Will test this tonight (no other opportunity any sooner, sorry)

xbmc/music/ContextMenus.cpp
```diff
@@ -35,6 +36,22 @@
 namespace CONTEXTMENU
 {
 
+static bool IsMusicDbItem(const CFileItem& item)
+{
+  if (URIUtils::IsMusicDb(item.GetPath()))
```
Razzeee: This if needs to go the `MUSICDATABASEDIRECTORY` will take care of it even if it's more expensive

razzeee (changes requested):

xbmc/music/ContextMenus.cpp
Razzeee: Only thing I can find wrong after that is that we need to handle `!item.IsParentFolder()` so that the `...` don't show the context menus.

ksooo: Hmmm, are these comments result of a runtime test or just out of theory? 

I thought parent folders do not have "musicdb" urls and I do not understand why the node type should be checked if the item has no "musicb" url. Doing the latter leads to the problem we're currently facing - namely "set as default" / "clear default" appearing at various wrong places.

ksooo: Okay, just found out that "musicdb" parent folders do havw "musicdb" URLs. I will change the code accordingly.

But I still cannot see why we always should check the node type, even if item has no music db url. Please elaboarte on this.

Razzeee: Result of a runtime test. 
All music nodes seem to have ``library:\\music\xxx` when I checked, so a check for `musicdb:\\` isn't helping us. We could do a new helper checking if the items are music library items.
There is a URIUtils class for checking library already, but it checks for both video, music and something else I've forgotten.
Se we could create `URIUtils::IsMusicLibrary`

ksooo:
@Razzeee I reintroduced the parentfolder check, as requested. Good to go now?

xbmc/music/ContextMenus.cpp
ksooo: Good idea. I will implement this.

ksooo:
@Razzeee should be really good to go now. I runtime tested this and could not find any "set as default" or "clear default" menu items.

Razzeee:
Will test tonight.

ksooo:
"superseeded" by #11990

PR #11989, opened 2017-04-20
FIXUP: select first unwatched broken after select by url

koying:


xbmc/video/windows/GUIWindowVideoNav.cpp
```diff
@@ -153,19 +158,24 @@ bool CGUIWindowVideoNav::OnMessage(CGUIMessage& message)
...
-          if (item.IsVideoDb())
+          SelectFirstUnwatched();
+
+          if (url.GetOption("showinfo") == "true")
           {
-            *(item.GetVideoInfoTag()) = XFILE::CVideoDatabaseFile::GetVideoTag(CURL(item.GetPath()));
-            if (!item.GetVideoInfoTag()->IsEmpty())
+            // We are here if the item is filtered out in the nav window
+            std::string path = message.GetStringParam(0);
+            CFileItem* item = new CFileItem(path, URIUtils::HasSlashAtEnd(path));
```
tamland: "I am able to manage the lifecycle of my objects by myself"-koying (Im sorry, I couldn't help it;))

I think I've commented on this exact code before..

koying: Hehe... Right. Not sure how it got back here...

xbmc/video/windows/GUIWindowVideoNav.cpp
```diff
@@ -153,19 +158,24 @@ bool CGUIWindowVideoNav::OnMessage(CGUIMessage& message)
...
           }
         }
-        if (i >= m_vecItems->Size() && url.GetOption("showinfo") == "true")
+        if (i >= m_vecItems->Size())
         {
-          // We are here if the item is filtered out in the nav window
-          std::string path = message.GetStringParam(0);
-          CFileItem item(path, URIUtils::HasSlashAtEnd(path));
-          if (item.IsVideoDb())
+          SelectFirstUnwatched();
```
Razzeee: Shouldn't this just live outside of the above if/else? It's now in both branches that can happen.

koying: Above which if/else?
If its `if (message.GetStringParam(0) != "")`, yes, it can, just that it would be executed unnecassirily more often...
If its `if (i >= m_vecItems->Size())` then no, because an item might have been select by the param

Razzeee: Sorry, ignore this, reviewing on github tricked me into believing the else was for another if then it really is...

razzeee (approved):

Razzeee:
jenkins build and merge

PR #11990, opened 2017-04-20
Revert "Moved party mode, set default and clear default to new contextmenu

ksooo:
This reverts commit 0c70c21c7f77acce0f9dd0c93506b816cee9abf4.

@Razzeee as we discussed on Slack, we want to revert this for now. The problem is more complex than simply copy & past over old code and logic. ;-)


ksooo:
jenkins build and merge

Razzeee:
We might want to talk about the party mode, should that be moved to the new system?

ksooo:
> We might want to talk about the party mode, should that be moved to the new system?

If this functionality makes sense not only for items displayed in music navigation window, but also for instance for certain home screen widget items, then yes. Otherwise, no.

Razzeee:
That's what we/I need to look at. Forgot that this is also in this PR.

PR #11992, opened 2017-04-20
[PVR] Search window: Fix very first search after kodi start not working.

ksooo:
This fixes a very special case with PVR search window:

1) Start Kodi
2) Open the PVR TV (or Radio, doesn't make a difference) window
3) Select "Search..." => Search dialog opens
==> bug: start and end date time both display 01/01/1970, 00:00:00
4)  do not change anything in the dialog; enter a search string - ideally a title of a show you know is in your current epg data
==> because of the nonsense start and end time (not matching epg data boundaries) nothing will be found.

This PR fixes start and end time mismatch by validating data against current epg data before opening the search dialog.

I runtime tested the changes on macOS, latest Kodi master.

@Jalle19 for review?

### DIFF
--- a/xbmc/pvr/epg/EpgSearchFilter.cpp
+++ b/xbmc/pvr/epg/EpgSearchFilter.cpp
@@ -48,8 +48,6 @@ void CPVREpgSearchFilter::Reset()
   m_iGenreSubType            = EPG_SEARCH_UNSET;
   m_iMinimumDuration         = EPG_SEARCH_UNSET;
   m_iMaximumDuration         = EPG_SEARCH_UNSET;
-  m_startDateTime.SetFromUTCDateTime(CServiceBroker::GetPVRManager().EpgContainer().GetFirstEPGDate());
-  m_endDateTime.SetFromUTCDateTime(CServiceBroker::GetPVRManager().EpgContainer().GetLastEPGDate());
   m_bIncludeUnknownGenres    = false;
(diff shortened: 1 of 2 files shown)


Jalle19:
Does this mean the search filter is constructed before the EPG is loaded? There is this in the constructor:

```
m_startDateTime.SetFromUTCDateTime(CServiceBroker::GetPVRManager().EpgContainer().GetFirstEPGDate());
```

ksooo:
> Does this mean the search filter is constructed before the EPG is loaded?

 Yes. CGUIWindowPVRSearch has a member of type CPVREpgSearchfilter. The window gets created very early during kodi startup, long before EPG data are loaded.

Jalle19:
Okay. Maybe the values set in the constructor could be removed and the proper values would be fetched when the search dialog is opened (kinda what you're doing now, but perhaps with a slightly different naming scheme).

If you don't want to bother then this is fine as it is now.

ksooo:
@Jalle19 I completely reworked s**t. What do you think about this approach. Much better, imo, because self contained.

Jalle19 (commented):
I would probably drop `const` and `mutable` since the members and methods aren't truly `const` now. Apart from that, much better, +1

ksooo:
@Jalle19 okay, I removed <code>const</code> and <code>mutable</code>, as you suggested.

jenkins build this please

PR #11994, opened 2017-04-21
FIX: [droid] remove dependency on android-support-v4.jar

koying:
The lib is not needed with a min api of 21.


MrMC:
death to the jar :) yea

koying:
Well, we might not be able to avoid them forever.
Next time we need one, we'll have to rewrite build & packaging ;)

garbear:
fixes my `make apk` error. thanks!

PR #11998, opened 2017-04-22
For addon created ListItems dont overwrite playcount data from DB

faush01:
When displaying ListItems from an addon dont overwite the playcount with the one stored in the DB if the addon has set a playcount in the video info tag data.
The addon should control what its play count is.

## Description
Default playcount in the video info tag data to -1 when creating ListItems
If the playcount is -1 update that with the value from the DB if there is one, otherwise set it to zero.

## Motivation and Context
I have an addon that wants to be able to set its own playcount and resume data and not have this overwritten by any local DB data Kodi has. The addon gets its play count and resume data from a remote source, in this case the www.emby.media server.

## How Has This Been Tested?
Tested with addon to make sure the playcount is no longer overwritten.
Tested:
 - if playcount is set by the addon it is not overwritten.
 - if an addon did not setting the playcount then it is correct set from the DB value.

## Types of change

- [X ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X ] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ X] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ X] All new and existing tests passed

### DIFF
--- a/xbmc/interfaces/legacy/ListItem.cpp
+++ b/xbmc/interfaces/legacy/ListItem.cpp
@@ -64,7 +64,7 @@ namespace XBMCAddon
       if (!path.empty())
         item->SetPath(path);
 
-	  item->GetVideoInfoTag()->SetPlayCount(-1);
+      item->GetVideoInfoTag()->SetPlayCount(-1);
     }
 
     ListItem::~ListItem()
--- a/xbmc/interfaces/legacy/ListItem.cpp
+++ b/xbmc/interfaces/legacy/ListItem.cpp
@@ -63,6 +63,8 @@ namespace XBMCAddon
         item->SetArt("thumb",  thumbnailImage );
       if (!path.empty())
         item->SetPath(path);
+
+	  item->GetVideoInfoTag()->SetPlayCount(-1);
     }
 
     ListItem::~ListItem()
--- a/xbmc/video/VideoDatabase.cpp
+++ b/xbmc/video/VideoDatabase.cpp
@@ -5266,13 +5266,6 @@ bool CVideoDatabase::GetPlayCounts(const std::string &strPath, CFileItemList &it
       m_pDS->next();
     }
 
-    // if playcount is still -1 set it to zero
-    for (CFileItemPtr pItem : items)
-    {
-      if (pItem->GetVideoInfoTag()->GetPlayCount() == -1)
-        pItem->GetVideoInfoTag()->SetPlayCount(0);
-    }
-
     return true;
   }
   catch (...)
--- a/xbmc/video/VideoDatabase.cpp
+++ b/xbmc/video/VideoDatabase.cpp
@@ -5255,7 +5255,7 @@ bool CVideoDatabase::GetPlayCounts(const std::string &strPath, CFileItemList &it
       CFileItemPtr item = items.Get(path);
       if (item)
       {
-        if (item->GetVideoInfoTag()->GetPlayCount() == -1)
+        if (!items.IsPlugin() || item->GetVideoInfoTag()->GetPlayCount() == -1)
         {
           item->GetVideoInfoTag()->SetPlayCount(m_pDS->fv(1).get_asInt());
         }
--- a/xbmc/video/VideoDatabase.cpp
+++ b/xbmc/video/VideoDatabase.cpp
@@ -5255,7 +5255,10 @@ bool CVideoDatabase::GetPlayCounts(const std::string &strPath, CFileItemList &it
       CFileItemPtr item = items.Get(path);
       if (item)
       {
-        item->GetVideoInfoTag()->SetPlayCount(m_pDS->fv(1).get_asInt());
+        if (item->GetVideoInfoTag()->GetPlayCount() == -1)
+        {
+          item->GetVideoInfoTag()->SetPlayCount(m_pDS->fv(1).get_asInt());
+        }
         if (!item->GetVideoInfoTag()->GetResumePoint().IsSet())
         {
           item->GetVideoInfoTag()->SetResumePoint(m_pDS->fv(2).get_asInt(), m_pDS->fv(3).get_asInt());
@@ -5263,6 +5266,17 @@ bool CVideoDatabase::GetPlayCounts(const std::string &strPath, CFileItemList &it
       }
       m_pDS->next();
     }
+
+    // if playcount is still -1 set it to zero
+    for (int i = 0; i < items.Size(); i++)
+    {
+      CFileItemPtr pItem = items[i];
+      if (pItem->GetVideoInfoTag()->GetPlayCount() == -1)
+      {
+        pItem->GetVideoInfoTag()->SetPlayCount(0);
+      }
+    }
+	
     return true;
   }
   catch (...)
--- a/xbmc/video/VideoDatabase.cpp
+++ b/xbmc/video/VideoDatabase.cpp
@@ -5256,9 +5256,8 @@ bool CVideoDatabase::GetPlayCounts(const std::string &strPath, CFileItemList &it
       if (item)
       {
         if (!items.IsPlugin() || item->GetVideoInfoTag()->GetPlayCount() == -1)
-        {
           item->GetVideoInfoTag()->SetPlayCount(m_pDS->fv(1).get_asInt());
-        }
+
         if (!item->GetVideoInfoTag()->GetResumePoint().IsSet())
         {
           item->GetVideoInfoTag()->SetResumePoint(m_pDS->fv(2).get_asInt(), m_pDS->fv(3).get_asInt());
@@ -5268,15 +5267,12 @@ bool CVideoDatabase::GetPlayCounts(const std::string &strPath, CFileItemList &it
     }
 
     // if playcount is still -1 set it to zero
-    for (int i = 0; i < items.Size(); i++)
+    for (CFileItemPtr pItem : items)
     {
-      CFileItemPtr pItem = items[i];
       if (pItem->GetVideoInfoTag()->GetPlayCount() == -1)
-      {
         pItem->GetVideoInfoTag()->SetPlayCount(0);
-      }
     }
-	
+
     return true;
   }
   catch (...)
--- a/xbmc/video/windows/GUIWindowVideoNav.cpp
+++ b/xbmc/video/windows/GUIWindowVideoNav.cpp
@@ -629,7 +629,7 @@ void CGUIWindowVideoNav::LoadVideoInfo(CFileItemList &items, CVideoDatabase &dat
       /* NOTE: Currently we GetPlayCounts on our items regardless of whether content is set
                 as if content is set, GetItemsForPaths doesn't return anything not in the content tables.
                 This code can be removed once the content tables are always filled */
-      if (!pItem->m_bIsFolder && !fetchedPlayCounts)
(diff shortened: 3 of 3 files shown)


Rechi:
I think getting playcount from the database shouldn't be excluded for all add-ons, only if the add-on provides a playcount.

Montellese:
I agree with @Rechi. There are a lot of add-ons relying on Kodi to store the playcount for their provided items because they don't have their own database.

faush01:
@Rechi 
Yes that would make sense, however the problem is playcount is an int, I can not think of a good way to work out if the addon set the playcount or not, i.e. if not set it is zero (0) and if it is set to zero (0) it is still zero (0).

How about -1? 
If playcount is -1 then the playcount from the DB if one is available is used.

This would require addons that want to have their playcount set form the DB value set the playcount to -1.
Addons that want to use their own playcount could set 0 or greater and it woul not be overwritten.

tamland:
Just make -1 the default for plugin items. Not need to require addons to set it.

faush01:
I can see two way to do this
in CVideoInfoTag::Reset()
set m_playCount = -1
This would start every CVideoInfoTag with a play count of -1
Or
In ListItem::setInfo
set the videotag playcount to -1
videotag.SetPlayCount(-1)
This would set just the playcount to -1 if not set by the plugin

I think the second way would be best

Rechi:
Third option
Call `item->GetVideoInfoTag()->SetPlayCount(-1)` at the end of the ListItem constructor at `xbmc/interfaces/legacy`

faush01:
should I cancel this pull request and submit a new one?

Rechi:
force push to your branch you created the PR from and the PR will get updated

faush01:
I have updated the pull request.

faush01:
Small fix, there was an edge cases where a file list was not having its playcount set.
Now it check and only bypasses the playcount set if the source of the items was an addon and the addon did not set the playcount itself.

Any feedback?

faush01:
Hi Guys,
I would be really awesome if I could get some feedback or sign off on this before my long weekend ends and I head back to work.
At the end of the day it is a reasonably small change however I can understand even small changes can have far reaching effects.
I anyone needs anything else done to help get the PR across the line please let me know.

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -63,6 +63,8 @@ namespace XBMCAddon
         item->SetArt("thumb",  thumbnailImage );
       if (!path.empty())
         item->SetPath(path);
+
+	  item->GetVideoInfoTag()->SetPlayCount(-1);
```
Rechi: indentation is 2 spaces, no tabs

xbmc/video/VideoDatabase.cpp
```diff
@@ -5255,14 +5255,28 @@ bool CVideoDatabase::GetPlayCounts(const std::string &strPath, CFileItemList &it
...
+    // if playcount is still -1 set it to zero
+    for (int i = 0; i < items.Size(); i++)
+    {
+      CFileItemPtr pItem = items[i];
+      if (pItem->GetVideoInfoTag()->GetPlayCount() == -1)
+      {
+        pItem->GetVideoInfoTag()->SetPlayCount(0);
+      }
+    }
+
```
Rechi: this is a tab, remove it

xbmc/video/VideoDatabase.cpp
```diff
@@ -5255,14 +5255,28 @@ bool CVideoDatabase::GetPlayCounts(const std::string &strPath, CFileItemList &it
       CFileItemPtr item = items.Get(path);
       if (item)
       {
-        item->GetVideoInfoTag()->SetPlayCount(m_pDS->fv(1).get_asInt());
+        if (!items.IsPlugin() || item->GetVideoInfoTag()->GetPlayCount() == -1)
+        {
+          item->GetVideoInfoTag()->SetPlayCount(m_pDS->fv(1).get_asInt());
+        }
```
Rechi: no need for curly brackets for a single statement

xbmc/video/VideoDatabase.cpp
```diff
@@ -5255,14 +5255,28 @@ bool CVideoDatabase::GetPlayCounts(const std::string &strPath, CFileItemList &it
...
     }
+
+    // if playcount is still -1 set it to zero
+    for (int i = 0; i < items.Size(); i++)
+    {
+      CFileItemPtr pItem = items[i];
+      if (pItem->GetVideoInfoTag()->GetPlayCount() == -1)
+      {
+        pItem->GetVideoInfoTag()->SetPlayCount(0);
+      }
```
Rechi: no need for curly brackets for a single statement

xbmc/video/VideoDatabase.cpp
```diff
@@ -5255,14 +5255,28 @@ bool CVideoDatabase::GetPlayCounts(const std::string &strPath, CFileItemList &it
...
         if (!item->GetVideoInfoTag()->GetResumePoint().IsSet())
         {
           item->GetVideoInfoTag()->SetResumePoint(m_pDS->fv(2).get_asInt(), m_pDS->fv(3).get_asInt());
         }
       }
       m_pDS->next();
     }
+
+    // if playcount is still -1 set it to zero
+    for (int i = 0; i < items.Size(); i++)
```
Rechi: can be replaced by a range-based for loop

faush01:
Excellent thanks @Rechi 
Changes made, committed and pushed

Rechi:
@Montellese @tamland please have a look again
For me it looks valid now.

xbmc/video/VideoDatabase.cpp
```diff
@@ -5255,14 +5255,24 @@ bool CVideoDatabase::GetPlayCounts(const std::string &strPath, CFileItemList &it
...
         if (!item->GetVideoInfoTag()->GetResumePoint().IsSet())
         {
           item->GetVideoInfoTag()->SetResumePoint(m_pDS->fv(2).get_asInt(), m_pDS->fv(3).get_asInt());
         }
       }
       m_pDS->next();
     }
+
+    // if playcount is still -1 set it to zero
+    for (CFileItemPtr pItem : items)
```
Montellese: Can this actually happen? I thought the default value of the playcount field in the files table is 0 but I might be wrong.

And btw you can use `for (auto pItem : items)` which is a bit more generic.

Montellese:
Commits need to be squashed.

Rechi:
@Montellese squashing can be done by the merger for some time already.
https://help.github.com/articles/merging-a-pull-request/

xbmc/video/VideoDatabase.cpp
faush01: It can happen if an addon did not set a playcount and there was no file entry for the plugin path to get a playcount for so at this point playcount will still be -1.

Montellese: Yes but I'm not sure if it should be the responsibility of `CVideoDatabase` to handle these cases because it shouldn't know what to do with files that it doesn't know about. IMO it should only touch files that it is tracking.

faush01: ok good point.
I can move it to where the GUIWindowVideoNav calls the GetPlayCounts, would that be better?

faush01:
I have moved the -1 check out of CVideoDatabase

xbmc/video/windows/GUIWindowVideoNav.cpp
```diff
@@ -642,6 +642,15 @@ void CGUIWindowVideoNav::LoadVideoInfo(CFileItemList &items, CVideoDatabase &dat
       if (!pItem->m_bIsFolder && !fetchedPlayCounts)
       {
         database.GetPlayCounts(items.GetPath(), items);
+        // for addons if playcount is still -1 set it to zero
+        if(items.IsPlugin())
+        {
+          for (auto pItem : items)
```
Montellese: `pItem` is already used as a variable in this scope so you're accidentally shadowing it which we should avoid to not cause any confusion.

xbmc/video/windows/GUIWindowVideoNav.cpp
```diff
@@ -642,6 +642,15 @@ void CGUIWindowVideoNav::LoadVideoInfo(CFileItemList &items, CVideoDatabase &dat
       if (!pItem->m_bIsFolder && !fetchedPlayCounts)
       {
         database.GetPlayCounts(items.GetPath(), items);
+        // for addons if playcount is still -1 set it to zero
+        if(items.IsPlugin())
```
Montellese: Space between `if` and `(`

Montellese:
Two minors and then it should be good to go.

xbmc/video/windows/GUIWindowVideoNav.cpp
faush01: absolute, good spot

faush01:
fix the variable shadowing and formatting.

faush01:
Do you guys need anything else done from my end?
If it is good to go do you merge this to master now or does that get done in a batch or something?

Rechi:
@Paxxi this would have been need to be squashed during merging

Paxxi:
ahh right, I totally missed that, so manually merge and squish then?

Rechi:
jenkins build this please

Rechi:
OSX failure is unrelated.

ksooo:
Damn! 23 seconds to late with my review requesting changes. I would recommend to introduce a constant instaed to use magic number -1 in several random places in different source files.

faush01:
I can still look at this, should I create another pull request for this?

Rechi:
If you could replace -1 with a constant, that would be great.
Yes just open a new PR.

notspiff:
this breaks all image addons. the reason is that the mere *existence* of a video info tag is used to flag this is a video.

faush01:
I think you have already ruled this change out for the issue you are seeing based on your other comment but just to confirm, this change does NOT remove the VideoInfoTab, it just works out where it sets the PlayCount from, either the video DB or the ListItem the pluggin supplied.

Montellese:
@faush01 the problem is that the call to `item->GetVideoInfoTag()->SetPlayCount(-1);` that you added to the `FileItem` constructor automatically adds a `CVideoInfoTag` object to the `CFileItem` independent of whether the item is a video, a music or an image object.

Montellese:
We could add a wrapper to `ListItem` for any internal calls to `item->GetVideoInfoTag()` which would basically do
```
CVideoInfoTag* GetVideoInfoTag()
{
  if (!item->HasVideoInfoTag())
    item->GetVideoInfoTag()->SetPlayCount(-1);

  return item->GetVideoInfoTag();
}
```
and use it everywhere in `ListItem` instead of directly calling `item->GetVideoInfoTag()`. That way we make sure that IF a plugin wants to access a video item property we first reset it's playcount to `-1`.

xhaggi:
Just a small hint. Please use `CVideoInfoTag::ResetPlayCount()` to reset the play count to `-1`. There is another convenient method `CVideoInfoTag::IsPlayCountSet()` to check if the play count is set. Introduced in fa61fe72b2d11c00505fd23494981052a214bc27.

Montellese:
@xhaggi thanks for the hint. I updated the fix and created a new PR at #12137.

PR #12000, opened 2017-04-23
[Android] Update AndroidKey.cpp

mihvel:
Extending support for the physical QWERTY keyboard in Android Kodi with processing of AKEYCODE_MOVE_HOME and AKEYCODE_MOVE_END key codes. Currently, these keyboard events are just ignored. In large Kodi list views jumping on FirstPage and LastPage is very convenient. I do not see any reason why it should not be supported in the Android Kodi, as it is on other OS platforms.

## Description
Enabling support for HOME/END physical keyboard keys for moving in Kodi lists on Android.

## Motivation and Context
There are more and more Android based media players on the market, not just tablets and phones. So, adding full support for the physical QWERTY keyboard in Android Kodi would be nice. Currently, I'm missing support and handling of the Home and End keyboard keys in the Android Kodi. Even my Android HTPC box handles Home/End keys properly, the Android Kodi does not process AKEYCODE_MOVE_HOME and AKEYCODE_MOVE_END key codes, so the keyboard events are just ignored. In large Kodi list views jumping on FirstPage and LastPage is very convenient and I used to use it on my previous Windows based HTPC. I do not see any reason why you should not support the keyboard in Kodi Android, as it is on other OS platforms.

http://forum.kodi.tv/showthread.php?tid=258808
http://forum.kodi.tv/showthread.php?tid=259037

## How Has This Been Tested?

I did my own build with these changes included, using Kodi Jarvis as starting point. I'm using it daily for almost a year, without any issues.

## Types of change

## Checklist:


Razzeee:
Thank you for your contribution and please excuse the late reply.
I'll ping @koying which should be able to process this PR!

koying:
+1
Just needs jenkins back to confirm it builds

MartijnKaijser:
jenkins build this please

PR #12001, opened 2017-04-23
[PVR] Search window improvements and fixes

ksooo:
Two changes related to the PVR search window :

1) Improvement: While searching epg events display standard "busy" dialog instaead of ugly modal dialog showing "searching..."

2) Fix: regression introduced with #11992. "Find similar" (triggers epg search from other pvr windows) did not work anymore.

The changes have been runtime tested on macOS, latest kodi master

@Jallle19 for code review?

### DIFF
--- a/xbmc/pvr/epg/EpgSearchFilter.cpp
+++ b/xbmc/pvr/epg/EpgSearchFilter.cpp
@@ -48,6 +48,15 @@ void CPVREpgSearchFilter::Reset()
   m_iGenreSubType            = EPG_SEARCH_UNSET;
   m_iMinimumDuration         = EPG_SEARCH_UNSET;
   m_iMaximumDuration         = EPG_SEARCH_UNSET;
+
+  m_startDateTime.SetFromUTCDateTime(CServiceBroker::GetPVRManager().EpgContainer().GetFirstEPGDate());
+  if (!m_startDateTime.IsValid())
+    m_startDateTime.SetFromUTCDateTime(CDateTime::GetUTCDateTime()); // default to 'now'
+
+  m_endDateTime.SetFromUTCDateTime(CServiceBroker::GetPVRManager().EpgContainer().GetLastEPGDate());
+  if (!m_endDateTime.IsValid())
+    m_endDateTime.SetFromUTCDateTime(m_startDateTime + CDateTimeSpan(10, 0, 0, 0)); // default to start + 10 days
+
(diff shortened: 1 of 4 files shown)


xbmc/pvr/epg/EpgSearchFilter.cpp
```diff
@@ -114,7 +117,7 @@ bool CPVREpgSearchFilter::MatchDuration(const CPVREpgInfoTagPtr &tag) const
 
 bool CPVREpgSearchFilter::MatchStartAndEndTimes(const CPVREpgInfoTagPtr &tag) const
 {
-  return (tag->StartAsLocalTime() >= m_startDateTime && tag->EndAsLocalTime() <= m_endDateTime);
+  return (tag->StartAsLocalTime() >= GetStartDateTime() && tag->EndAsLocalTime() <= GetEndDateTime());
```
Jalle19: Did you check if `m_startDateTime` and `m_endDateTime` are used anywhere else still?

ksooo: Yes, this time a took care. No other direct usage anymore. :-)

Jalle19: Thanks :)

Jalle19: I still don't like the `mutable` stuff but it's probably the easiest route for now. Ideally the start and end times should be set separately from the getters and setters, but at the right moment so that the values will be valid (unlike what happened in the constructor before).

ksooo: Maybe it's best to dump all the "cuteness" from the search filter class and to set the right start + stop values from outside (e.g. on search dialog init), before doing a search. What do you think?

Jalle19: Something like that yes

ksooo:
@Jalle19 what do you think about this new approach?

ksooo:
Answering my own question, I still do not like this fix. I think I will end up with something more pragmatic. CPVREpgSearchfilter::Reset will check the dates returned by the epg container and in case they are invalid it will init those values to "now" and "now + 10 days". This should solve all problems.

ksooo:
@Jalle19 I guess now I have it. ;-) What's your opinion on this?

Jalle19 (approved):

ksooo:
Jenkins build this please

PR #12002, opened 2017-04-24
AE: improve tempo filter

FernetMenta:
Improves accuracy of tempo filter

### DIFF
--- a/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEFilter.cpp
+++ b/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEFilter.cpp
@@ -55,7 +55,8 @@ void CActiveAEFilter::Init(AVSampleFormat fmt, int sampleRate, uint64_t channelL
   m_sampleRate = sampleRate;
   m_channelLayout = channelLayout;
   m_tempo = 1.0;
-  m_bufferedSamples = 0;
+  m_SamplesIn = 0;
+  m_SamplesOut = 0;
 }
 
 bool CActiveAEFilter::SetTempo(float tempo)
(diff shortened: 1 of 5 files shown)


FernetMenta:
jenkins build this please

xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEFilter.cpp
```diff
@@ -277,6 +280,8 @@ int CActiveAEFilter::ProcessFilter(uint8_t **dst_buffer, int dst_samples, uint8_
       return -1;
     }
 
+    m_SamplesOut = m_pOutFrame->pts;
```
popcornmix: I think this is responsible for the hang on Pi when pts=AV_NOPTS_VALUE.

FernetMenta: did you log pts? pts of outFrame is independent of input in this case. it should never be equal to AV_NOPTS_VALUE. further why should this behave different on Pi?

popcornmix: Yes, I added this logging:
http://paste.ubuntu.com/24450547/
and got this log:
http://paste.ubuntu.com/24450549/
Certainly looks like pts=AV_NOPTS_VALUE

FernetMenta: oh, this should be outFrame, not m_pOutFrame. if format conversion is required outFrame is set to m_pConvertFrame. that is the reason why the issue showed on Pi and not on Ubuntu

popcornmix:
Updated version doesn't hang any more on Pi.

FernetMenta:
jenkins build this please

PR #12004, opened 2017-04-24
changed: always expand iso images in music window

notspiff:
This makes iso images into filedirectories in the music window.

Currently, there has been no iso support for music (although we can do audio cd iso in the code i believe). Anyways, I want to add an add-on for SACD iso support. That requires this change.

It should do no harm, and by default .iso is not a music extension anyways.

### DIFF
--- a/xbmc/filesystem/Directory.h
+++ b/xbmc/filesystem/Directory.h
@@ -78,7 +78,10 @@ public:
 
   /*! \brief Filter files that act like directories from the list, replacing them with their directory counterparts
    \param items The item list to filter
-   \param mask  The mask to apply when filtering files */
-  static void FilterFileDirectories(CFileItemList &items, const std::string &mask);
+   \param mask  The mask to apply when filtering files
+   \param expandImages True to include disc images in file directory expansion
+  */
+  static void FilterFileDirectories(CFileItemList &items, const std::string &mask,
+                                    bool expandImages=false);
 };
 }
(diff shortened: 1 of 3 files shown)


xbmc/filesystem/Directory.h
```diff
@@ -79,6 +79,6 @@ class CDirectory
   /*! \brief Filter files that act like directories from the list, replacing them with their directory counterparts
    \param items The item list to filter
    \param mask  The mask to apply when filtering files */
```
Razzeee: Could you please add the documentation for the new parameter here? It's now incomplete

razzeee (approved):
Thank you! LGTM :)

xbmc/filesystem/Directory.h
notspiff: done. in order to make it possible to write sane doxy, i have flipped the bool as well ;)

Razzeee: Much better the variable name was not that good before :)

Razzeee:
jenkins build this please

Razzeee:
jenkins build this please

Razzeee:
Random build errors...

jenkins build this please

DaveTBlake:
@notspiff just discovered this change introduces a regression which I need some help to fix.
Before this change adding <playlistasfolders>false</playlistasfolders> to advancedsettings.xml allowed .m3u playlists to be listed in the playlists node as if they were a file (total size shown) and `enter` caused playback. After this ,m3u files are always treated as folders, `enter` expanding the folder etc. and <playlistasfolders> ignored.

PR #12005, opened 2017-04-24
changed: remove internal sftp support

notspiff:
This removes the internal sftp support, ref https://github.com/xbmc/repo-binary-addons/pull/82 and https://github.com/notspiff/vfs.sftp

Note this raises the question what to do with the network source setup dialog. I have opted to show the sftp entry unconditionally right now.


MartijnKaijser:
Perhaps it should be shown if the addon is present on the system. If the add-on is disabled it asks to enable it?

notspiff:
yes. that is certainly an option. but that requires some language for the add-ons to specify what its protocol support.

Rechi:
`tools/depends/target/libssh` can also be removed

notspiff:
good catch.

Rechi:
You also have to remove libssh from `tools/depends/target/Makefile` otherwise building depends will fail.

notspiff:
correct you are. using my left hand's left hand and 1% brain on this thanks for handholding ;)

candrews:
_bump_

Razzeee:
Rebase please :)

notspiff:
rebased. add-on still needs to be moved to new structure and synced with mainline code + we need to figure out what to do with the protocol support in the add network source dialog.

Rechi:
https://github.com/xbmc/xbmc/blob/45874ebd3e57a897b1435d9bcfecbce258eb9926/project/BuildDependencies/scripts/0_package.target-x64.list#L28 should also be removed

notspiff:
done. i assumed you wanted it removed for win32 as well

xhaggi:
@notspiff time to push this forward?

xbmc/network/GUIDialogNetworkSetup.cpp
```diff
@@ -203,9 +203,7 @@ void CGUIDialogNetworkSetup::InitializeSettings()
 #ifdef HAS_FILESYSTEM_NFS
   labels.push_back(std::make_pair(20259, NET_PROTOCOL_NFS));
 #endif
-#ifdef HAS_FILESYSTEM_SFTP
   labels.push_back(std::make_pair(20260, NET_PROTOCOL_SFTP));
```
xhaggi: should be dropped

notspiff:
need to update the add-on a little then yes. will get to asap.

xhaggi:
ok cool 👍

notspiff:
rebased, add-on updated, appeared to work fine in some quick tests.

notspiff:
@afedchin is there some (new) winstore file that needs cleaning ?

notspiff:
note: add-on still needs work for depends etc.

afedchin:
please remove libssh from dependencies list for uwp in the following files:
https://github.com/xbmc/xbmc/blob/master/project/BuildDependencies/scripts/0_package.target-win10-arm.list#L25
https://github.com/xbmc/xbmc/blob/master/project/BuildDependencies/scripts/0_package.target-win10-win32.list#L25
https://github.com/xbmc/xbmc/blob/master/project/BuildDependencies/scripts/0_package.target-win10-x64.list#L25

xhaggi:
> note: add-on still needs work for depends etc.

Thanks

xhaggi (approved):

candrews:
_bump_

candrews:
@notspiff just curious on the status of this one, I'm interested in seeing it in Kodi 18

notspiff:
i've gotten no indications this is wanted for v18 so it is in slumber mode.

MartijnKaijser:
If we can switch to addon we can add it.

Rechi:
Don't know if the following language strings should be kept in core
https://github.com/xbmc/xbmc/blob/dbb7f60b23bb10607fb868736588c160cca79e0d/addons/resource.language.en_gb/resources/strings.po#L6025-L6028
https://github.com/xbmc/xbmc/blob/dbb7f60b23bb10607fb868736588c160cca79e0d/addons/resource.language.en_gb/resources/strings.po#L12472-L12475

Rechi (approved):

PR #12011, opened 2017-04-25
[addons] remove not needed part and add new version check way

AlwinEsch:
First step of the big rework

## Description
This remove on first commit the never used GetSettings add-on function. This was thought to define addon settings by the code and not with xml, but nowhere used and becomes removed.

The other two commits add the first part of the new version way. This is thought to prevent e.g. the `GetVersion` function on PVR. There becomes everything added to one header (versions.h), reason to have in one is to read from them with cmake (on next changes in new coming request) and write them to a `addon.xml.in`.

## Motivation and Context
Big binary add-on system rework.

## Types of change

## Checklist:


addons/resource.language.en_gb/resources/strings.po
```diff
@@ -15232,7 +15232,13 @@ msgctxt "#29802"
 msgid "Passthrough audio in use"
 msgstr ""
 
-#empty strings from id 29803 to 29899
+#. The label in the add-ons dll class to report about outdated add-on api version
+#: xbmc/addons/AddonDll.cpp
+msgctxt "#29803"
+msgid "Using incompatible API version for type '%s'.\nKodi API version = '%s', add-on API version '%s'\n\nPlease check for updates or contact developer of add-on."
```
MartijnKaijser: Please do not use line breaks in strings if they are not really needed.

notspiff:
you want me to merge up front ?

AlwinEsch:
Want to create on all repo addons the request first, to reduce time for this changes.

xbmc/addons/AddonDll.cpp
```diff
@@ -461,5 +398,31 @@ ADDON_STATUS CAddonDll::TransferSettings()
...
+                            kodi::addon::GetTypeName(type),
+                            kodiVersion,
+                            addonVersion);
+
+    std::string heading = StringUtils::Format("%s: %s", TranslateType(Type(), true).c_str(), Name().c_str());
+    std::string text = StringUtils::Format(g_localizeStrings.Get(29803).c_str(),
+                                           kodi::addon::GetTypeName(type),
+                                           kodiVersion,
+                                           addonVersion);
+    CGUIDialogOK::ShowAndGetInput(CVariant{heading}, CVariant{text});
```
tamland: Don't make gui calls from deep down in the addons system please. We shouldn't have random functions adding popous. Instead this should be in the gui code tied to some user action (or notification if it's a background task)

garbear:
because it's an ABI change (break?) can we bumb the add-on versions? Otherwise an add-on compiled against a newer kodi will fail on an older Kodi build

ksooo:
> can we bumb the add-on versions?

 yes, please.

AlwinEsch:
Hi @garbear @ksooo this request is to allow it in future, currently is the add-on base interface not checked for version changes, that's the main reason to add `ADDON_GetTypeVersion` function here.

After them is in brings a old add-on a load failure, but no crashes:
bildschirmfoto vom 2017-04-27 00-46-35

The Version increase does not work for them, because the library is not loaded and his functions not available.

After them comes another request who bring the new version check way, as example see this https://github.com/FernetMenta/kodi-agile/blob/master/addons/kodi.global.main/addon.xml.in#L2. This is only the first step.

xbmc/addons/AddonDll.cpp
AlwinEsch: Have no idea for really good alternatives, want to have them only be available there, that's why I find a user message as not so good way. As alternative can I use this:
```cpp
CEventLog::GetInstance().AddWithNotification(EventPtr(new CNotificationEvent(heading, text, EventLevel::Error)));
```
who bring:
bildschirmfoto vom 2017-04-27 02-04-08

Here how it was with them here:
bildschirmfoto vom 2017-04-27 02-09-31

xbmc/addons/AddonDll.cpp
```diff
@@ -461,5 +400,31 @@ ADDON_STATUS CAddonDll::TransferSettings()
...
+                            Name().c_str(),
+                            kodi::addon::GetTypeName(type),
+                            kodiVersion,
+                            addonVersion);
+
+    std::string text = StringUtils::Format(g_localizeStrings.Get(29803).c_str(),
+                                           kodi::addon::GetTypeName(type),
+                                           kodiVersion,
+                                           addonVersion);
+    CEventLog::GetInstance().AddWithNotification(EventPtr(new CNotificationEvent(Name(), text, EventLevel::Error)));
```
AlwinEsch: @tamland have changed it to the `CEventLog::GetInstance().AddWithNotification` with:
bildschirmfoto vom 2017-04-27 02-04-08

This by start of screensaver with wrong version:
bildschirmfoto vom 2017-04-27 02-24-33

ksooo:
Not sure I got this right. But I guess you know best what you are doing. Maybe you should at least tell the addon maintainers that merging the addon PR you opened yesterday leads to a no longer functioning addon (at least until you merged this PR), because it is ABI incompatible?

AlwinEsch:
The function call tables (e.g. PVR) are not changed, this change "C" exported functions from the base addon functions.

Goal on end is to have only a `ADDON_Create` and `ADDON_GetTypeVersion` as exported functions, everything else becomes then handled with function tables. This allow then in future to check a several years old addon is compatible.

AlwinEsch:
jenkins build this please

AlwinEsch:
OK to bring in?

xbmc/addons/kodi-addon-dev-kit/include/kodi/libKODI_guilib.h
```diff
@@ -170,10 +170,10 @@ typedef struct CB_GUILib
 
 
 /* current ADDONGUI API version */
-#define KODI_GUILIB_API_VERSION "5.11.0"
+#define KODI_GUILIB_API_VERSION ADDON_GLOBAL_VERSION_GUI
 
 /* min. ADDONGUI API version */
-#define KODI_GUILIB_MIN_API_VERSION "5.10.0"
+#define KODI_GUILIB_MIN_API_VERSION ADDON_GLOBAL_VERSION_GUI
```
stefansaraev:
```
# cat /usr/share/kodi/addons/kodi.guilib/addon.xml 
<addon id="kodi.guilib" version="#define KODI_GUILIB_API_VERSION ADDON_GLOBAL_VERSION_GUI" provider-name="Team-Kodi">
  <backwards-compatibility abi="#define KODI_GUILIB_MIN_API_VERSION ADDON_GLOBAL_VERSION_GUI"/>
  <requires>
    <import addon="xbmc.core" version="0.1.0"/>
  </requires>
</addon>
```

any addon that depends on guilib is probably broken now. 
https://github.com/xbmc/xbmc/blob/master/cmake/scripts/common/Macros.cmake#L632-L634 should also be changed

PR #12013, opened 2017-04-26
Revert "DVDDemuxFFMPEG: Increase FFMPEG_FILE_BUFFER_SIZE to 128K"

fritsch:
This caused issues in the RPi forum threads. As we don't want to leave master broken we better revert until we find the time to investigate the root issue.

### DIFF
--- a/xbmc/cores/FFmpeg.h
+++ b/xbmc/cores/FFmpeg.h
@@ -65,4 +65,4 @@ public:
   int level;
 };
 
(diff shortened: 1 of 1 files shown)


MilhouseVH:
And x86_64.

http://forum.kodi.tv/showthread.php?tid=298462&pid=2576620#pid2576620

popcornmix:
@MilhouseVH no, you are correct. This revert is for part of #11619 so looks like it does cause problem for x86_64 too.

FernetMenta (approved):

fritsch:
jenkins build and merge

PR #12016, opened 2017-04-27
guilib: only update container after condition changed

FernetMenta:
fixes high CPU load caused by ever dirty controls

@Memphiz ping


Memphiz:
Thx - isn't this worth to blackport to krypton? I could do that ...

Memphiz:
I tested it and it's good to go - not sure but this is a bugfix and can go in right now with Jenkins approving?

FernetMenta:
it is good to go for master. no idea about v17

fritsch:
jenkins build and merge

PR #12018, opened 2017-04-27
VideoPlayer improvements

garbear:
Found another compiler warning when compiling on linux, and discovered an unused function


FernetMenta:
thanks

PR #12019, opened 2017-04-27
Fix compiler error on Windows

garbear:
This fixes a compiler error that I encountered on windows.


a1rwulf (approved):
Sorry @garbear I somehow missed your PR.
Thx @Razzeee for poking me.

PR #12026, opened 2017-04-28
DVDDemuxFFmpeg: max_packet_size - use GetBlockSize directly

fritsch:
This drops an ancient hack that was introduced for hdhomerun once upon a time. As this part of the code is quite fragile, let's get it tested first.


FernetMenta:
maybe this is still not correct: ffmpeg doc:

```
 * @param buffer_size The buffer size is very important for performance.
 *        For protocols with fixed blocksize it should be set to this blocksize.
 *        For others a typical size is a cache page, e.g. 4kb.
```

buffer_size should be set, not max_buffer_size

fritsch:
```
    unsigned char* buffer = (unsigned char*)av_malloc(FFMPEG_FILE_BUFFER_SIZE);
    m_ioContext = avio_alloc_context(buffer, FFMPEG_FILE_BUFFER_SIZE, 0, this, dvd_file_read, NULL, dvd_file_seek);
    m_ioContext->max_packet_size = m_pInput->GetBlockSize();
```

I can try to to set the buffer and allocate GetBlockSize and give that into avio_alloc_context.

fritsch:
This needs more thoughts, as it is used with the probe buffers again.

PR #12027, opened 2017-04-28
[ios]: adapt VTB renderer to VideoPicture

Rechi:
@FernetMenta 
The missing changes for ios and your vpupdates branch.
Not runtime tested, only compiled.


FernetMenta:
@Rechi thanks. I leave this to @Memphiz who is the iOS maintainer. IMO it is easier to catch up if compiler fails. Then you know where to look at.

Memphiz:
Looks good. Fine by me.

PR #12032, opened 2017-04-29
zlib configure modification to address cross-compile for android on OS X

kraka40:
## Description
Added CHOST environment variable to ./configure  to ensure correct cross-compiler picked up during target compilation.

## Motivation and Context
This was necessary when compiling depends library on OS X using NDK 12b.

## How Has This Been Tested?
tested on only one platform
OSX 
uname -a
16.5.0 Darwin Kernel Version 16.5.0: Fri Mar  3 16:52:33 PST 2017; root:xnu-3789.51.2~3/RELEASE_X86_64 x86_64

sdk
grep "Revision" source.properties 
Pkg.Revision=25.2.5

ndk
grep "Revision" source.properties 
Pkg.Revision = 12.1.2977051

## Types of change

## Checklist:

### DIFF
--- a/tools/depends/target/zlib/Makefile
+++ b/tools/depends/target/zlib/Makefile
@@ -7,7 +7,8 @@ VERSION=1.2.7
 SOURCE=$(LIBNAME)-$(VERSION)
 ARCHIVE=$(SOURCE).tar.gz
 # configuration settings
-CONFIGURE= CC="$(CC)" CFLAGS="$(CFLAGS)" ./configure --prefix=$(PREFIX) --static
+CONFIGURE= CHOST="$(TOOLCHAIN)/bin/$(HOST)" CC="$(CC)" CFLAGS="$(CFLAGS)" \
+	./configure --prefix=$(PREFIX) --static
 
 LIBDYLIB=$(PLATFORM)/$(LIBNAME).a
 


Memphiz:
Not sure if we really want this to be a seperate pull request or if we shouldn't all the other commits that are needed too. I wanted to wait on your reedback about my branch and would have done a pull request then myself - you've partly beat me to it ;)

As i am unsure how this CHOST thing affects the other platforms i let jenkins check already. I am still unsure why this works because of osx host the AR is hardcoded to /usr/bin/libtool which at least i don't have on my system. Do you have libtool in that place by any chance?

Jenkins build this please

kraka40:
I'm good waiting... working through your build and master with changes still.

I do have libtool in /usr/bin but that isn't the problem. It was picking up the cross compiler AR which was necessary on OS X. CHOST may or may not working on other platforms need to wait for more testing feedback. CHOST is being picked up by the zlib configuration script and being prepended on the AR, LIBTOOL, etc binaries. If not there it defaults. So thought it was better to try to use native zlib compiler options.

Memphiz:
No I meant from the Zlib changelog:
- Force the native libtool in Mac OS X to avoid GNU libtool [Beebe]

This hardcodes ar to /use/bin/libtool - which is not there in my mac ... no CHOST or other variable can overwrite that I think ... just try to rename your libtool and it should fail even with that CHOST change ... - not sure where that libtool comes from on your system (maybe brew or Ports or so?)

kraka40:
ah ok .. not sure .. I’ll look again.

 I just overwrote the target/Makefile not the native one so I’m sure zlib was compiled correctly for osx.  it was attempting to create the zlib.a library for android with OS X libtool, which wasn’t compatible with the cross-compiler gcc.    definitely need to see how this works with other builds.  I believe your fork is an older xbmc 15? set of libraries.

I was able to just finish compiling for android on OS X with 17.2.   Haven’t installed yet… coming soon.

Memphiz:
My fork is master from April 2017 ...

kraka40:
ok .. I must have pulled in incorrectly .. I'll look again sorry about that.

kraka40:
closing pull request .. CHOST is not the correct approach.

PR #12034, opened 2017-04-30
KODI: Update NEON support

C-Prime90:
## Description
This commit makes CCPUInfo::HasNeon() report NEON enabled on Aarch64 targets and updates the ifdef's for newer preprocessor.

## How Has This Been Tested?
This has bin tested locally on Amlogic S812/S905 running on EmbER firmware.

## Types of change

## Checklist:


stefansaraev:
are you 100% that enabling -funsafe-math-optimizations is safe for kodi ?

croniccorey:
I've had it enabled in my Linux builds since 16.0 without issues, i can remove that if you'd prefer.

cmake/scripts/linux/ArchSetup.cmake
```diff
@@ -16,10 +16,11 @@ else()
   elseif(CPU MATCHES arm)
     set(ARCH arm)
     set(NEON True)
-    set(NEON_FLAGS "-mfpu=neon -mvectorize-with-neon-quad")
+    set(NEON_FLAGS "-mfpu=neon -funsafe-math-optimizations -mvectorize-with-neon-quad")
   elseif(CPU MATCHES aarch64 OR CPU MATCHES arm64)
     set(ARCH aarch64)
-    set(NEON False)
+    set(NEON True)
+    set(NEON_FLAGS "-march=armv8-a -funsafe-math-optimizations")
```
stefansaraev: aarch64 is always neon. you dont need these, and you should not override user's specified -march= anyway

croniccorey:
I've updated the PR based on your comments.

MartijnKaijser:
@FernetMenta @peak3d ping as well

FernetMenta:
What non-NEON systems do we support? I am wondering if we can just drop the ifdef and make it default.

stefansaraev:
@FernetMenta rpi1/0

FernetMenta:
so we enable neon with a compiler switch, and then test for preprocessor macros? seems a bit combersome and error prone to me. finally results in this kind of issues.
why don't we user something like HAS_NEON that is set by defult and can be disabled for the outdated hardware

wsnipex:
agreed, we should be using the preprocessor defines here.

MartijnKaijser:
@croniccorey could you do the requested changes?

croniccorey:
so you would like to drop the compiler switch in favor of preprocessor defines?

wsnipex:
>You would like to drop the compiler switch in favor of preprocessor defines or drop preprocessor defines in favor of HAS_NEON set by cmake?
yes, this please

croniccorey:
Ok i will update PR to use HAS_NEON

croniccorey:
@MartijnKaijser @FernetMenta @wsnipex I have updated the PR based on your comments

wsnipex:
jenkins build this please

MartijnKaijser:
@croniccorey thx for sticking with us and for your contribution

PR #12035, opened 2017-04-30
[guilib] Evaluate skin expressions everywhere

rmrector:
## Description
Skin expressions $EXP are now evaluated for skin variable conditions, expressions inside of other expressions, and include conditions __*__. #12096 refactors custom window handling so their visible conditions can include skin expressions as well. Now expressions work everywhere you want them!

__*__ This works for conditions for the prolific named includes that are used in windows, but not conditions for nameless whole-file includes that can be used only in the main _includes.xml_ file. Allowing expressions there doesn't seem worth the hassle of all the code required to get that behavior just right.

## Motivation and Context
Even skinners deserve to be DRY.

http://forum.kodi.tv/showthread.php?tid=308385
http://forum.kodi.tv/showthread.php?tid=300636

## How Has This Been Tested?
Manually, by putting expressions in these newly supported places and ensure they do what I expect, and roll through all sorts of other windows checking that nothing else was affected.

## Types of change

- [X] Bug fix (non-breaking change which fixes an issue), or, depending on how you want to look at it,

## Checklist:

### DIFF
--- a/xbmc/guilib/GUIIncludes.cpp
+++ b/xbmc/guilib/GUIIncludes.cpp
@@ -367,7 +367,7 @@ void CGUIIncludes::ResolveIncludes(TiXmlElement *node, std::map<INFO::InfoPtr, b
     const char *condition = include->Attribute("condition");
     if (condition)
     {
-      INFO::InfoPtr conditionID = g_infoManager.Register(condition);
+      INFO::InfoPtr conditionID = g_infoManager.Register(ResolveExpressions(condition));
       bool value = conditionID->Get();
 
       if (xmlIncludeConditions)
--- a/xbmc/guilib/GUIIncludes.cpp
+++ b/xbmc/guilib/GUIIncludes.cpp
@@ -106,7 +106,15 @@ void CGUIIncludes::Clear()
   m_expressions.clear();
 }
 
-bool CGUIIncludes::Load(const std::string &file)
+void CGUIIncludes::Load(const std::string &file)
+{
+  if (!Load_Internal(file))
+    return;
+  FlattenExpressions();
+  FlattenSkinVariableConditions();
+}
+
+bool CGUIIncludes::Load_Internal(const std::string &file)
 {
   // check to see if we already have this loaded
   if (HasLoaded(file))
@@ -235,12 +243,59 @@ void CGUIIncludes::LoadIncludes(const TiXmlElement *node)
       if (condition)
       { // load include file if condition evals to true
         if (g_infoManager.Register(condition)->Get())
-          Load(file);
+          Load_Internal(file);
       }
       else
-        Load(file);
+        Load_Internal(file);
     }
     child = child->NextSiblingElement("include");
+  }
+}
+
+void CGUIIncludes::FlattenExpressions()
+{
+  for (auto& expression : m_expressions)
+  {
+    std::vector<std::string> resolved = std::vector<std::string>();
+    resolved.push_back(expression.first);
+    FlattenExpression(expression.second, resolved);
+  }
+}
+
+void CGUIIncludes::FlattenExpression(std::string &expression, const std::vector<std::string> &resolved)
+{
+  std::string original(expression);
+  CGUIInfoLabel::ReplaceSpecialKeywordReferences(expression, "EXP", [&](const std::string &expressionName) -> std::string {
+    if (std::find(resolved.begin(), resolved.end(), expressionName) != resolved.end())
+    {
+      CLog::Log(LOGERROR, "Skin has a circular expression \"%s\": %s", resolved.back().c_str(), original.c_str());
+      return std::string();
+    }
+    auto it = m_expressions.find(expressionName);
+    if (it == m_expressions.end())
+      return std::string();
+
+    std::vector<std::string> rescopy = resolved;
+    rescopy.push_back(expressionName);
+    FlattenExpression(it->second, rescopy);
+
+    return it->second;
+  });
+}
+
+void CGUIIncludes::FlattenSkinVariableConditions()
+{
+  for (auto& variable : m_skinvariables)
+  {
+    TiXmlElement* valueNode = variable.second.FirstChildElement("value");
+    while (valueNode)
+    {
+      const char *condition = valueNode->Attribute("condition");
+      if (condition)
+        valueNode->SetAttribute("condition", ResolveExpressions(condition));
+
+      valueNode = valueNode->NextSiblingElement("value");
+    }
   }
 }
 
--- a/xbmc/guilib/GUIIncludes.h
+++ b/xbmc/guilib/GUIIncludes.h
@@ -47,13 +47,13 @@ public:
   void Clear();
 
   /*!
-   \brief Load all include components (defaults, constants, variables, expressions and includes)
-    from the given \code{file}.
-   
+   \brief Load all include components(defaults, constants, variables, expressions and includes)
+   from the main entrypoint \code{file}. Flattens nested expressions and expressions in variable
+   conditions after loading all other included files.
+
    \param file the file to load
-   \return true if the file was loaded otherwise false
-   */
-  bool Load(const std::string &file);
+  */
+  void Load(const std::string &file);
 
   /*!
    \brief Resolve all include components (defaults, constants, variables, expressions and includes)
@@ -81,6 +81,15 @@ private:
     SINGLE_UNDEFINED_PARAM_RESOLVED
   };
 
+  /*!
+   \brief Load all include components (defaults, constants, variables, expressions and includes)
+   from the given \code{file}.
+
+   \param file the file to load
+   \return true if the file was loaded otherwise false
+  */
+  bool Load_Internal(const std::string &file);
+
   bool HasLoaded(const std::string &file) const;
 
   void LoadDefaults(const TiXmlElement *node);
@@ -88,6 +97,24 @@ private:
   void LoadVariables(const TiXmlElement *node);
   void LoadConstants(const TiXmlElement *node);
   void LoadExpressions(const TiXmlElement *node);
+
+  /*!
+   \brief Resolve all expressions containing other expressions to a single evaluatable expression.
+  */
+  void FlattenExpressions();
+
+  /*!
+   \brief Expand any expressions nested in this expression.
+
+   \param expression the expression to flatten
+   \param resolved list of already evaluated expression names, to avoid expanding circular references
+  */
+  void FlattenExpression(std::string &expression, const std::vector<std::string> &resolved);
+
+  /*!
+   \brief Resolve all variable conditions containing expressions to a single evaluatable condition.
+  */
+  void FlattenSkinVariableConditions();
 
   void SetDefaults(TiXmlElement *node);
   void ResolveIncludes(TiXmlElement *node, std::map<INFO::InfoPtr, bool>* xmlIncludeConditions = NULL);


gade01:
+1 for backporting this and https://github.com/xbmc/xbmc/pull/12012 to Krypton.

MartijnKaijser:
Not gonna happen

ronie:
ping @xhaggi for review

xhaggi:
Will take a look at it next week, thanks for mentioning me @ronie

MartijnKaijser:
@xhaggi you had any time?

xhaggi (changes requested):

xhaggi:
Forgot to submit my review ;(

xbmc/Application.cpp
```diff
@@ -1809,6 +1809,8 @@ bool CApplication::LoadUserWindows()
               id = atol(pType->FirstChild()->Value());
           }
           std::string visibleCondition;
+          for (TiXmlElement* node = pRootElement->FirstChildElement("visible"); node; node = node->NextSiblingElement("visible"))
+            g_SkinInfo->ResolveIncludes(pRootElement->FirstChildElement("visible"));
```
rmrector: This one is goofy, I'd rather not resolve the visible conditions here at all. I expected `CGUIWindow::Load(pRootElement)` to handle the window visibility for custom windows like it does regular windows, but I just could not get that to work. 

`CGUIIncludes::ResolveExpressions` is private and it didn't seem like a good idea to change that and add a `CSkinInfo::ResolveExpressions` to pass it through, not for this one goofy instance.

In all other instances `CGUIIncludes::ResolveIncludes` already expands expressions into the XML before `CGUIControlFactory::GetConditionalVisibility` is called so it would mostly be duplicated effort.

xhaggi: >I expected CGUIWindow::Load(pRootElement) to handle the window visibility for custom windows like it does regular windows

I thought the same and we definitely need to change something here. I've already started to work on it. After that every other comment should therefore become obsolete.

xbmc/guilib/GUIIncludes.cpp
```diff
@@ -513,13 +513,22 @@ std::string CGUIIncludes::ResolveConstant(const std::string &constant) const
...
-std::string CGUIIncludes::ResolveExpressions(const std::string &expression) const
+std::string CGUIIncludes::ResolveExpressions(const std::string &expression, const std::vector<std::string> &resolved /* = std::vector<std::string>() */) const
 {
   std::string work(expression);
   CGUIInfoLabel::ReplaceSpecialKeywordReferences(work, "EXP", [&](const std::string &str) -> std::string {
-    std::map<std::string, std::string>::const_iterator it = m_expressions.find(str);
-    if (it != m_expressions.end())
-      return it->second;
+    if (std::find(resolved.begin(), resolved.end(), str) != resolved.end())
+      CLog::Log(LOGWARNING, "Skin has a circular expression involving: %s", str.c_str());
```
xhaggi: We should change the log level to error and also log the input string to make it easier for skinners to spot such a circular expression.

rmrector:
I popped off the commit for custom windows as #12096 is a better solution.

xbmc/guilib/GUIIncludes.cpp
```diff
@@ -513,13 +513,22 @@ std::string CGUIIncludes::ResolveConstant(const std::string &constant) const
...
 {
-  std::string work(expression);
-  CGUIInfoLabel::ReplaceSpecialKeywordReferences(work, "EXP", [&](const std::string &str) -> std::string {
-    std::map<std::string, std::string>::const_iterator it = m_expressions.find(str);
-    if (it != m_expressions.end())
-      return it->second;
+  std::string work(strInput);
+  CGUIInfoLabel::ReplaceSpecialKeywordReferences(work, "EXP", [&](const std::string &expression) -> std::string {
+    if (std::find(resolved.begin(), resolved.end(), expression) != resolved.end())
+      CLog::Log(LOGERROR, "Skin has a circular expression \"%s\": %s", resolved.back().c_str(), strInput.c_str());
```
Paxxi: I would stick a `return std::string()` here and avoid the else clause and decrease nesting a bit.

xbmc/guilib/GUIIncludes.cpp
```diff
@@ -513,13 +513,22 @@ std::string CGUIIncludes::ResolveConstant(const std::string &constant) const
...
+    else
+    {
+      std::map<std::string, std::string>::const_iterator it = m_expressions.find(expression);
```
Paxxi: use `auto`

xbmc/guilib/GUIIncludes.cpp
```diff
@@ -530,6 +539,8 @@ const INFO::CSkinVariableString* CGUIIncludes::CreateSkinVariable(const std::str
 {
   std::map<std::string, TiXmlElement>::const_iterator it = m_skinvariables.find(name);
   if (it != m_skinvariables.end())
-    return INFO::CSkinVariable::CreateFromXML(it->second, context);
+    return INFO::CSkinVariable::CreateFromXML(it->second, context, [&](const std::string &expression) -> std::string {
```
Paxxi: no need to capture any variable here so you can use `[]` I would think?

wouldn't it be possible to pass ResolveExpressions directly instead of using a lamba?

Paxxi (commented):
Looks good to me. just some nitpicking that may or may not improve things :)

xbmc/guilib/GUIIncludes.cpp
```diff
@@ -513,13 +513,22 @@ std::string CGUIIncludes::ResolveConstant(const std::string &constant) const
   return StringUtils::Join(values, ",");
 }
 
-std::string CGUIIncludes::ResolveExpressions(const std::string &expression) const
+std::string CGUIIncludes::ResolveExpressions(const std::string &strInput, const std::vector<std::string> &resolved /* = std::vector<std::string>() */) const
```
rmrector: With it logging `strInput` when it encounters a circular expression, I want it to log the original string that will be easier for skinners to identify, rather than the possibly half-resolved `work`.

xhaggi: Agree with @rmrector

xbmc/guilib/GUIIncludes.cpp
rmrector: `ResolveExpressions` is a non-static member function, so...

 For usage in the lambda the current object has to be captured. `[this]` will work, though.

And to pass it directly it needs to be bound to the current object first, and I think
`std::bind(&CGUIIncludes::ResolveExpressions, this, std::placeholders::_1, std::vector<std::string>())`
is less readable than the lambda.

xhaggi: As expressions may not change until we reload the skin, we could first load expressions and resolve them directly for variables while we load them at https://github.com/xbmc/xbmc/blob/master/xbmc/guilib/GUIIncludes.cpp#L203

This way we would save some round trips here, or won't we?

xhaggi:
@rmrector I'm currently working on moving each step in `ResolveIncludesForNode` to it's own method and rename the method `ResolveIncludes` to a more generic method `Resolve` which calls the necessary new methods `SetDefaultsForNode`, `ResolveIncludesForNode`, `ResolveConstantsForNode` and `ResolveExpressionsForNode`. Next step could be renaming the whole class to fit to what it does, load/resolve various skin components.

xbmc/guilib/GUIIncludes.cpp
rmrector: Yup, `CreateSkinVariable` can be called multiple times for the same variable if it is used in more than one context.

xbmc/guilib/GUIIncludes.cpp
rmrector: Thinking on it further, it is probably a good idea to flatten nested expressions after they are loaded up here [L218](https://github.com/rmrector/xbmc/blob/ce501e0d1eb4aaaf979cef455269a3e736060a42/xbmc/guilib/GUIIncludes.cpp#L218), rather than recursing through expressions for every conditional statement in the skin.

rmrector:
@xhaggi sounds good. I need to spend a few days elsewhere, but next week I can work on cleanup and rebasing.

xbmc/guilib/GUIIncludes.cpp
xhaggi: Good idea, same like for variables.

rmrector:
Rebased on @xhaggi's recent changes and implemented other discussed changes. I added `CGUIIncludes::LoadEntrypoint` so that the `Flatten` methods can run after all include files are loaded so expressions/variable conditions can reference expressions in other include files, never mind the order they are loaded.

xbmc/guilib/GUIIncludes.h
```diff
@@ -47,13 +47,13 @@ class CGUIIncludes
...
-   \return true if the file was loaded otherwise false
-   */
-  bool Load(const std::string &file);
+  \brief Load all include components (defaults, constants, variables, expressions and includes)
+  from the main entrypoint \code{file}. Flattens nested expressions and expressions in variable
+  conditions after loading all other included files.
+
+  \param file the file to load
+  */
+  void LoadEntrypoint(const std::string &file);
```
xhaggi: IMO no need to introduce a new method, just put the flatten method calls at the bottom of the `Load` method.

xbmc/guilib/GUIIncludes.cpp
```diff
@@ -244,6 +251,53 @@ void CGUIIncludes::LoadIncludes(const TiXmlElement *node)
...
+
+    return it->second;
+  });
+}
+
+void CGUIIncludes::FlattenSkinVariableConditions()
+{
+  for (auto& variable : m_skinvariables)
+  {
+    TiXmlElement* valuenode = variable.second.FirstChildElement("value");
```
xhaggi: `valueNode` please.

xhaggi:
Thank you, I'll will take a look at it asap.

xhaggi (changes requested):

xhaggi:
Apart from my comments it looks good.

xbmc/guilib/GUIIncludes.h
rmrector: `Load` is called for every single included file, and flattening all previously loaded expressions/variables when loading each new file is just wasting cycles when they can all be flattened once after everything is loaded. That could add an unnecessary delay when loading and reloading a skin, depending on hardware and how skinners lay out their include files. The current code works no matter how many include files or in what order the skinner chooses to include them, simply and efficiently.

Imagine a skin with variables defined in _variables.xml_ and expressions in _expressions.xml_. If _variables.xml_ is included first then flatten when loading it will have to give  `CGUIInfoLabel::ReplaceSpecialKeywordReferences` the same `$EXP[whatever]` it is trying to replace (so it can try again after the next file is loaded), accomplishing nothing but wasting cycles. The same thing will happen when loading every included file between here and _expressions.xml_, compounding the issue.

I could move the flatten calls behind a bool parameter `flatten` on the `Load` method that would only be true for the main _include.xml_ file (and "file" includes from `ResolveIncludes`, which I've missed currently), if that is better than this new method.

xhaggi: Then we should name the current `Load` method `Load_Internal` and the `LoadEntrypoint` -> `Load` because we use this wording for most of these methods in our guilib. Sorry for my late response.

BTW could you also add some doxy method docs to the flatten methods to make it clear what they do.

rmrector: Done.

xhaggi (approved):

xhaggi:
Looks good and thanks for your contribution.

xbmc/guilib/GUIIncludes.cpp
```diff
@@ -106,7 +106,14 @@ void CGUIIncludes::Clear()
   m_expressions.clear();
 }
 
-bool CGUIIncludes::Load(const std::string &file)
+void CGUIIncludes::Load(const std::string &file)
+{
+  Load_Internal(file);
```
xhaggi: One minor if `Load` return false there is no need to call the flatten methods.

xhaggi:
Jenkins build this please

xhaggi:
Jenkins build this please

xbmc/guilib/GUIIncludes.cpp
```diff
@@ -235,15 +243,62 @@ void CGUIIncludes::LoadIncludes(const TiXmlElement *node)
...
+void CGUIIncludes::FlattenExpression(std::string &expression, const std::vector<std::string> &resolved)
+{
+  std::string original(expression);
+  CGUIInfoLabel::ReplaceSpecialKeywordReferences(expression, "EXP", [&](const std::string &expressionName) -> std::string {
+    if (std::find(resolved.begin(), resolved.end(), expressionName) != resolved.end())
+    {
+      CLog::Log(LOGERROR, "Skin has a circular expression \"%s\": %s", resolved.back().c_str(), original.c_str());
+      return std::string();
+    }
+    auto& it = m_expressions.find(expressionName);
```
Rechi: `auto it = ...`

xhaggi:
Jenkins build this please

rmrector:
Great! Thanks @xhaggi for all your help.

PR #12038, opened 2017-04-30
[guilib] Isolate included expressions $EXP

rmrector:
## Description
Backport of #12012. Isolate skin included expressions so that when their including condition is evaluated the order of operations matches expected behavior.

## Motivation and Context
The current behavior just replaces expressions inline, which can have unexpected results if it contains an or "|" not inside brackets.

## Types of change

This is technically a breaking change because someone could have designed their skin to take advantage of this behavior, but most likely they are just wrapping the condition in brackets themselves to work around it.

## Checklist:


MartijnKaijser:
jenkins build this please

PR #12040, opened 2017-05-01
[osx][droid] target zlib Makefile modified to ensure correct compiler

kraka40:
## Description
 Allowed for full compile of android apk on osx platform. The zlib library was incorrectly picking up the native archiving tools. 

The addition of CROSS_PREFIX to the Makefile configuration command ensures that the correct target binaries are utilized.  The zlib configuration script makes assumptions when using CHOST to guess target operating system (uname).  To get around this, --uname=$HOST is added to the configuration parameters.

## Motivation and Context
Allows for android cross compile of OSX

## How Has This Been Tested?
Tested on OSX only, produced a valid an functioning apk. apk side-loaded to firetv gen1 and operating.

Darwin Kernel Version 16.5.0
NDK 12.1.2977051
SDK 25.2.5
javac 1.8.0_111

## Types of change

## Checklist:

### DIFF
--- a/tools/depends/target/zlib/Makefile
+++ b/tools/depends/target/zlib/Makefile
@@ -7,8 +7,8 @@ VERSION=1.2.7
 SOURCE=$(LIBNAME)-$(VERSION)
 ARCHIVE=$(SOURCE).tar.gz
 # configuration settings
-CONFIGURE= CHOST="$(TOOLCHAIN)/bin/$(HOST)" CC="$(CC)" CFLAGS="$(CFLAGS)" \
-       ./configure --prefix=$(PREFIX) --static
+CONFIGURE= CROSS_PREFIX="$(TOOLCHAIN)/bin/$(HOST)-" CC="$(CC)" CFLAGS="$(CFLAGS)" \
+       ./configure --prefix=$(PREFIX) --static --uname=$(HOST)
 
 LIBDYLIB=$(PLATFORM)/$(LIBNAME).a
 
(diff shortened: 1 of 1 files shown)


wsnipex:
jenkins build this please

MartijnKaijser:
ping @Memphiz

Memphiz:
Not complete maybe -  I am working on it here https://github.com/xbmc/xbmc/pull/12059

Memphiz:
Correct me if i am wrong - but are native compilers not just called "gcc" without any prefix?

kraka40:
@Memphiz native compiler should be gcc, as to prefix not sure .. i'm sure a prefix could be added to pick up a specific gcc version locally if the system has multiple versions. 

As to the failed checks with my patch .. don't understand why they failed as they seem completely unrelated to zlib compilation or zlib library utilization
.

Rechi:
@Memphiz compiler is always the one specified as CC so that isn't an issue.
In my opinion #12059 is the way to go because it removes the hardcoded `AR` and `ARFLAGS` for darwin.
It should always use the values defined by the depends configure step.
Also I'm not sure if this breaks other platforms, although it doesn't look like it will break anything.

MartijnKaijser:
@Memphiz is this still relevant now your PR was merged?

Memphiz:
Not really - I would need to backport those changes from master instead of this PR

MartijnKaijser:
Let's close this. @kraka40 thx for helping out regardless

kraka40:
No problem. Change is pretty simple regardless should a backport be required.

PR #11845, opened 2017-03-13
[win] use 64-bit toolset if available

Rechi:
## Description
Use the 64-bit (crosscompiler) toolset on a 64-Bit OS if it is available, otherwise fall back to the 32-Bit toolset.

## Motivation and Context
Default toolset on 64-Bit OS for win32 are 32-Bit programs.

## How Has This Been Tested?
Compiled kodi and an addon. Checked used compiler during building.

## Types of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [x] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #12048, opened 2017-05-02
[buildsteps][windows] run-tests: use 64-bit toolset if available

Rechi:
## Description
Use the 64-bit (crosscompiler) toolset on a 64-Bit OS if it is available, otherwise fall back to the 32-Bit toolset for run-test.

## Motivation and Context
For compiling kodi 64-bit toolset is already used (#11845). If not using the same toolset for kodi-test as for kodi the whole Visual Studio Solution get recompiled which increases build time unnecessary.

## Types of change

## Checklist:


Rechi:
jenkins build this please

PR #12050, opened 2017-05-03
Set Param1 to WINDOW_INVALID for plugins when Init is called

faush01:
## Description

Small fix to send the correct WINDOW_INVALID in Param1 of the window init message that is send if the window is showing the content from a plugin.

## Motivation and Context

For opening addons with ActivateWindow the return param was being ignored, this was a bug.

## How Has This Been Tested?

Run all built time tests and built in tests.
Tested that when ActivateWindow is called with return the history is correctly set to just the root entry point.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/windows/GUIMediaWindow.cpp
+++ b/xbmc/windows/GUIMediaWindow.cpp
@@ -1490,8 +1490,8 @@ void CGUIMediaWindow::OnInitWindow()
   }
   else
   {
-    CGUIMessage msg(GUI_MSG_WINDOW_INIT, 0, 0, 0, PLUGIN_REFRESH_DELAY);
-    g_windowManager.SendThreadMessage(msg, m_controlID);
+    CGUIMessage msg(GUI_MSG_WINDOW_INIT, 0, 0, WINDOW_INVALID, PLUGIN_REFRESH_DELAY);
+    g_windowManager.SendThreadMessage(msg, GetID());
   }
 
   if (updateStartDirectory)


jjd-uk:
@xhaggi maybe one for you to review as you seem to be one of the few to have touched GUIMediaWindow.cpp in recent times.

jjd-uk:
For anyone reviewing the problem is discussed at http://forum.kodi.tv/showthread.php?tid=306301

The behaviour can easily be replicated with a Official Repo add-on such as Youtube.

xhaggi:
For me this fix looks invalid as it doesn't fix the root cause of the issue. First we need to find the changes which introduce this regression.

I've compared the current version of `CGUIMediaWindow` with Jarvis and for me it looks like 5220b49d4159f5ee62fd34409f5b811a16185b80 introduced some special handling for plugin media windows. It sends another GUI_MSG_WINDOW_INIT thread message to the current window but with `param1` = 0 instead of WINDOW_INVALID. In `CGUIMediaWindow::OnMessage` we check  `Param1` against WINDOW_INVALID and set the `m_startDirectory` depending if `returning` is true, which is not the case for this message (https://github.com/xbmc/xbmc/blame/master/xbmc/windows/GUIMediaWindow.cpp#L498).

IMO the following part should be changed from:
```
CGUIMessage msg(GUI_MSG_WINDOW_INIT, 0, 0, 0, PLUGIN_REFRESH_DELAY);
g_windowManager.SendThreadMessage(msg, m_controlID);
```
to
```
CGUIMessage msg(GUI_MSG_WINDOW_INIT, 0, 0, WINDOW_INVALID, PLUGIN_REFRESH_DELAY);
g_windowManager.SendThreadMessage(msg, GetID());
```

I don't have time to test it by myself so it would be nice if someone else could verify if this would fix the issue.

faush01:
@xhaggi 
I will look into this but I dont think this is the case.

The issue of when return is specified to the original window activate would still occur.

When you called set history it still builds the full path history which is what return should avoid doing.

faush01:
I have checked the changes you recommended and that did not fix the problem.

I have had a look through the history and I can not see how this ever worked. SetHistoryForPath has always been called and it has always rest the full path, I dont know how it work or even if it worked in the past.

I am happy to keep looking if people want but at the moment the only way I can see to fix this everywhere is to do what I have submitted.

If anyone has any better ideas I am all for it and willing to test it out.

xhaggi:
Do you mean that return won't work for non-addon media windows (e.g. VideoLibrary) too?

Okay, I'll have to debug it on my own. We need to know which `SetHistoryForPath` is called and why the history is not cleared in case return is passed in. BTW have a look at the first commit https://github.com/xbmc/xbmc/commit/9a921c39b8d0a366c7a52184cc86f8d47e685be5 which introduce the return param. There is a call to `m_history.ClearPathHistory();` on window init. This method should be used to clear the history.

https://github.com/xbmc/xbmc/blame/902f14b964adc88ff153476439d414370ddcd908/xbmc/windows/GUIMediaWindow.cpp#L493

faush01:
Just wont work, addon or media the full back path is set regardless if you specify return or not.

I am obviously missing something simple here, I can not see how the original code worked and I can not build that version so cant step though it.

I think if you step though it you will see what is happening, then SetHistoryForPath is called it sets the full history from the full path regardless of the return param.

jjd-uk:
Strange as the last time I checked this issue links to media items such as Videos in Library worked, it was only the links to Add-on items that wasn't working.

xhaggi:
I've done some debugging and this https://github.com/xhaggi/xbmc/commit/c00669a607bec7389841a411f31d2607bc65335f fixes the issue for me.

I added a Youtube path and an ARD medithek path to my favorites, without the fix it goes up in the directory path, with the fix it returned me to the home screen > favourites in Esturary.

xhaggi:
> Strange as the last time I checked this issue links to media items such as Videos in Library worked, it was only the links to Add-on items that wasn't working.

Same for me, only plugin directories are effected. Library directories worked for me too. You can easily check it if you use the home screen sub menu entries in Estuary e.g. under Movies > Genre, Years etc.

xhaggi:
@jjd-uk could you verify if my fix work for you?

jjd-uk:
@xhaggi if you can kick off a jenkins build then sure, on the way to devcon so probably won't be able to test until tomorrow

xhaggi:
For which OS?

jjd-uk:
Windows please

xhaggi:
Build is running http://jenkins.kodi.tv/view/Windows/job/WIN-32/12366/

faush01:
I just tried the suggested changes again and they do not work for me.
This is the Skin menu item I have on the home screen.

<onclick>ActivateWindow(Videos,plugin://plugin.video.embycon?mode=SHOW_CONTENT&amp;item_type=Movie,return)</onclick>

It shows the movies section fine of the EmbyCon addon I have but when I hit back it goes to

plugin://plugin.video.embycon?mode=SHOW_CONTENT&amp;item_type=Movie
then
plugin://plugin.video.embycon
then
plugin://

which finally takes you back to the home screen.
I have no idea why this is working for you guys and not me.

notspiff:
with return set, the given path shall indeed be the root of the dir history. are there two regression at play?

xhaggi:
The real magic is done in `CGUIMediaWindow::OnBack` where we check if the current item path is not equal `m_startDirectory` and `GetRootPath()`. If that is the case it will call `GoParentFolder()` otherwise it will call `CGUIWindow::OnBack` which call `g_windowManager.PreviousWindow();`.

The changes in https://github.com/xhaggi/xbmc/commit/c00669a607bec7389841a411f31d2607bc65335f will fix that `m_startDirectory` is set to another value after the second thread message was processed.

@faush01 I don't have EmbyCon, so please debug the `CGUIMediaWindow::OnBack` method in your case and let me know what's the reason the condition in `OnBack` is true and therefor it calls `GoParentFolder()`

xhaggi:
@jjd-uk test build is up on our mirror

faush01:
Thanks for the help.

The fix from https://github.com/xhaggi/xbmc/commit/c00669a607bec7389841a411f31d2607bc65335f works for me when I specify the CORRECT addon path in my ActivateWindow

My original Activate Window had the following
plugin://plugin.video.embycon?mode=SHOW_CONTENT&amp;item_type=Movie

and for some reason this gets changed to
plugin://plugin.video.embycon/?mode=SHOW_CONTENT&amp;item_type=Movie

notice the / after the plugin name. So what was happening was the test to check if the current path was equal to the start directory was failing due to the / being added into to the path but not into the start directory.

if I use:

plugin://plugin.video.embycon/?mode=SHOW_CONTENT&amp;item_type=Movie

as the plugin path in my ActivateWindow now the path and the start directory are the same and the OnBack correctly calls CGUIWindow::OnBack(actionID) with calls g_windowManager.PreviousWindow();

So just to confirm it looks like the suggested fix works as long as the plugin URL is specified correctly with the / after the plugin name.

Did you want me to alter this pull request to contain just the changes you recommended or will you follow through with that change in a different channel?

Thanks again for the help.

xhaggi:
Good to here you found it and yes it would be nice if you could alter your​ changes. And please also create a backport PR for Krypton.

faush01:
I have reopened this with the new changes to fix the bug.

jjd-uk:
I've tested the build @xhaggi generated at http://mirrors.kodi.tv/test-builds/win32/KodiSetup-20170504-c00669a607-fix-activatewindow-with-return.exe and looks to be working ok with Youtube.

xhaggi:
Thank you both.

Jenkins build this please

MartijnKaijser:
Jenkins build this please

Flavien:
Thanks for fixing this. FYI, [I reported this in November 2016 for Krypton beta 5](https://trac.kodi.tv/ticket/17075), and it never got looked at.

Paul-Reed:
Has this backport found it's way into the current release please. I'm running Kodi 17.3 Git:147cec4 (installed via Openelec) and the issue still exits - using back arrow doesn't take me back to main menu, instead it takes me to (screenshot)

tv

faush01:
yes this is in 17.3

Paul-Reed:
Thanks for the reply, I now think that this is more to do with the Aeon Nox skin, than Kodi core.
I've posted an issue in the Kodi - Aeon Nox forum section.

PR #12057, opened 2017-05-05
reenable mark as watched/unwatched for folders

notspiff:
this is just a quick hack to reenable mark as watched/unwatched for folders. it works just fine, but i wonder why you went to length to nuiter the functionality so i open this to get the attention.

ref http://forum.kodi.tv/showthread.php?tid=311447&pid=2581701#pid2581701

### DIFF
--- a/xbmc/video/ContextMenus.cpp
+++ b/xbmc/video/ContextMenus.cpp
@@ -56,6 +56,7 @@ bool CMarkWatched::IsVisible(const CFileItem& item) const
 
   if (item.m_bIsFolder) //Only allow db content and recording folders to be updated recursively
   {
+    return true;
     if (item.HasVideoInfoTag())
       return item.IsVideoDb();
     else
@@ -80,6 +81,7 @@ bool CMarkUnWatched::IsVisible(const CFileItem& item) const
 
   if (item.m_bIsFolder) //Only allow db content and recording folders to be updated recursively
   {
+    return true;
     if (item.HasVideoInfoTag())
       return item.IsVideoDb();
     else


Jalle19 (commented):
Could you remove the now dead code instead of just adding return statements?

ksooo:
> Could you remove the now dead code instead of just adding return statements?

 No way. This is not a proper fix (and I think @notspiff knows this; he just wanted to give a heads-up). Returning true for every file item that is a folder results in every folder item having the "mark watched/unwatched" context menu item. We need to identify the "video folder" items properly and *add* this check in addition to the existig ones.

ksooo (changes requested):
This is not a proper fix. Please refer to my other comment for details.

akva2:
Cba to relogin. Yeah what ksooo said. I will do that when and if the explanation for *explicitly* coding the regression makes that the appropriate course of action.

Jalle19:
@ksooo you're totally right, I looked at this a bit too sloppily it seems.

gitAnne:
So now what needs to happen?
Using item.IsVideoDb() instead of the item.m_bIsFolder won't work I presume.
Can you check the content type of a folder item?

krabsheva:
I dunno the politics involved here, but is someone going to fix this problem? Seems like it would take someone with decent knowledge of the codebase 5 minutes to fix this issue properly.

I have no idea in which cases these menu options need to be removed from a folder context menu, or I would change it myself. Personally, I would be happy with the kludge fix, but I don't use the PVR stuff or library mode for video.

Since it looks like this was broken back in December, perhaps that dev should look into this?

ksooo:
No, it takes definitely more than 5 minutes. Please refrain from such statements if you have no idea what you are talking about. Thanks

ksooo:
With old context menu system, the context was given by the window information that wanted to support certain context menu items. 

Now, there us no such context, only a file item and this to my best knowledge currently lacks the information we need - whether the item is a "video folder".

gitAnne:
When I request the [C]ontext menu of a video folder in the user interface, I can check the Content type and change it if needed. I can only change the content to 1 of the 3 video content types (Movies, TV Series and Music videos). That would mean that the informatie what content type a folder has is available somehow?

With the content type of a folder available it should be possible to add the 'mark as watched' option?

Or am I missing something here?

ksooo:
Yes, "somehow" the info is available, but nobody showed up here so far with the knowledge, "how" exactly. 

I know the new context menu system and pvr, but unfortunately not the video and music code base of Kodi, sorry.

krabsheva:
If no one who knows the video side is willing to step in then the kludge is the best solution available. At least you don't continue to err on the side that takes away functionality that has existed for years.

ksooo:
> If no one who knows the video side is willing to step in then the kludge is the best solution available. 

No way, as it breaks other components, like PVR. -1 from my end.

gitAnne:
'-1 from my end'
I assume that is a -1 for using a kludge/work around, 
not towards reintroducing the 'mark as watched' functionality for video folders, I hope ;P

This removed functionality really added quite a lot of work and frustration for me. 
I also still don't get why this was broken as it does not seem like it was a conscious decision to remove it.

ksooo:
>> '-1 from my end'

> I assume that is a -1 for using a kludge/work around,
> not towards reintroducing the 'mark as watched' functionality for video folders, I hope ;P

 Yes, no way to merge the work around as it breaks other stuff.

> This removed functionality really added quite a lot of work and frustration for me.

 Seems that this is a killer feature for you. Why don't you stick with the last version of Kodi that works? Or why don't you switch to library mode? The bug only appears with file mode, right?

> I also still don't get why this was broken 
 
 Simply by accident? Because the dev who changed the implementation to the new context menu system had not enough knowledge (it was not me, btw - so start hating somebody else, please)? OMG.

> as it does not seem like it was a conscious decision to remove it.

 Pffft. Hold yer horses.

> So where do I have to go to to start a voting campaign to get this back?

 As long as no dev gets into the mood to fix it, you can collect millions of votes without any affect. ;-)

ksooo:
@notspiff is it okay for you if I close this (annoying, tbh) PR? A respective trac ticket should be opened, instead.

gitAnne:
Yes it's a feature I use quite extensively.
And I'm kind of surprised that not everyone does ;P

I'll investigate how I can find out which version did not have this update and how to roll back to it, because that does seem like a good plan.

So no amount of icecream will be able to help get this feature back I gather? ;P

Thanks for the comments.

ksooo:
Team Kodi currency is COB (crates of beer).

notspiff:
@ksooo the only reason this is annoying is that the team is unable to answer my question - then i would code whatever fix you would want. but sure, if its is annoying..

PR #12059, opened 2017-05-05
[OSX] - fix compilation for android on OSX hosts.

Memphiz:
This fixes cross compilation for android on MacOSX

## Description
Basically there were still some tools used outside of the toolchain which are not available on osx (on linux you wouldn't notice because they are there on the host).
Basically this ensures those tools are used from the toolchain instead. I also updated the android readme with some stuff. @koying for a review and maybe @hudokkow for the cmake stuff?

## Motivation and Context
Compiling for android on OSX was broken after the cmake transition.

## How Has This Been Tested?
I have compiled for android on Mac OS Sierra 10.12.4. - also @kraka40 who brought up the problems has tested it and also runtime tested the resulting apk. Thx alot for sticking to it @kraka40

## Types of change

## Checklist:

- [X] I have updated the documentation accordingly


wsnipex:
the zlib fix from @kraka40 looks saner then adding a patch.
The rest looks good to me.

jenkins build this please

Memphiz:
@wsnipex - are you sure this even could work when not cross compiling? For me it looks wrong when we don't even have a cross compile toolchain (like on linux native builds) no?

Memphiz:
Also looking at my patch - i don't think you can override that libtool path because it is hardcoded no? (it only workd for kraka40 because he had libtool installed in that path which isn't the normal case on osx imo).

kraka40:
I have a standard osx build. No special libtool install. I am ok with whatever team determines is best but I am a little partial to my zlib fix, seems more inline with intention of zlib maintainers.

Rechi:
I think `remove_fixe_ar_osx.patch` should always applied (not only if target is android) as otherwise it doesn't use the definition of `AR` from our configure / Makefile.include.

MrMC:
looks sane and agree with @Rechi

Memphiz:
@rechi but that hardcoded AR is in a "if darwin" branch inside that configure script ... but of course we can apply it always to make our Makefile a bit cleaner ... will change.

tools/depends/target/zlib/Makefile
```diff
@@ -21,7 +21,8 @@ $(TARBALLS_LOCATION)/$(ARCHIVE):
 $(PLATFORM): $(TARBALLS_LOCATION)/$(ARCHIVE) $(DEPS)
 	rm -rf $(PLATFORM); mkdir -p $(PLATFORM)
 	cd $(PLATFORM); $(ARCHIVE_TOOL) $(ARCHIVE_TOOL_FLAGS) $(TARBALLS_LOCATION)/$(ARCHIVE)
-	cd $(PLATFORM); patch -p0 < ../visibility.patch
+	cd $(PLATFORM); patch -p0 < ../visibility.patch
```
Rechi: unrelated white space change

Memphiz:
@kraka40 i have check my other sierra rig when i am back home from devcon - you are right that there is a libtool on my sierra partition on my hackbook.

kraka40:
Thanks for checking.., was starting to wonder what I might have installed

Memphiz:
@kraka40 i just had another look to it now. Well there is a /usr/bin/libtool indeed. But it can't be used for linking arm objects for android. This is the error it gives when trying to do so:
https://pastebin.com/0adSSeHF

kraka40:
Will check later... not sure I was using it.

Where are you picking it up?

Memphiz:
Sorry - where do I pick up what?

MrMC:
Try doing a 'cd $(PLATFORM); $(AUTORECONF) -vif'  before the configure. sometimes the pre-processed configure is not correct for cross.

Rechi:
zlib doesn't provide a `configure.ac` or `configure.in`

MrMC:
Ahh, one of 'those'.
Is depends configure not picking up zlib from the toolchain ?

kraka40:
@Memphiz 

Makefile.in overrides local libtool

export LIBTOOL=/Volumes/android-dev/kodi/kodi-depends/x86_64-darwin16.6.0-native/bin/libtool
export LIBTOOLIZE=/Volumes/android-dev/kodi/kodi-depends/x86_64-darwin16.6.0-native/bin/libtoolize

there is a conditional in the zlib configuration file to set AR=/usr/bin/libtool, but only if uname is Darwin

my patch #12040  this, so /usr/bin/libtool is not picked up.  Though that looks to have failed some test [ still not sure why or which one ]

Memphiz:
@kraka40 are we in line that your patch enforces cross compile even on native compile targets (f.e. Linux) which will not work? Or are you trying to say your patch works as is and is the better solution?

kraka40:
@Memphiz not sure on the native compile .. I'll need to deploy to my linux platform which I try not to compile on as is my pvr.  I assume the $HOST variable would be the native host so should work with a native compile, and therefore would work in both cases.  Has anyone tried it?

Rechi:
@Memphiz is there anything missing?

Memphiz:
Nothing beside a Jenkins build this please i guess.

PR #12063, opened 2017-05-05
[settings] fix custom localization of setting labels/options

Montellese:
Right now any settings dialog deriving from `CGUIDialogSettingsBase` can provide its own localization implementation for setting labels etc. This works great but I did not consider the labels used as part of the setting controls which currently all use `g_localizeStrings` instead of the settings dialog's specific localization implementation. This PR fixes it by passing an `ILocalizer` interface implementation to all the relevant parts of the settings GUI code.

Right now probably nobody will notice this because all strings are available through `g_localizeStrings` but it's inconsistent with the initial idea and it will fail when migrating add-on settings to core's settings system. Therefore there's no need for a backport to Krypton.

## Types of change

## Checklist:


Montellese:
jenkins build this please

Montellese:
Any objections? jenkins was happy before he was struck down ;-)

PR #12064, opened 2017-05-06
[OSX/Windowing] - replace deprecated functions for display mode switching

Memphiz:
As the topic says. This replaces some deprecated calls for switching display modes on osx (refreshrate switching for example).

## Motivation and Context
Just stumbled over it while investigating on how to switch displays into 3d mode via hdmi modes. (which doesn't work at all so there is no PR for this).

## How Has This Been Tested?
Tested on MacMini OSX Sierra (10.12.4) and OSX El Capitan 10.11.6.

## Types of change

- [X] Cleanup?

## Checklist:

@FernetMenta for a readover.

### DIFF
--- a/xbmc/windowing/osx/WinSystemOSX.mm
+++ b/xbmc/windowing/osx/WinSystemOSX.mm
@@ -227,6 +227,79 @@ CGDirectDisplayID GetDisplayID(int screen_index)
   return(displayArray[screen_index]);
 }
 
+int DisplayBitsPerPixelForMode(CGDisplayModeRef mode)
+{
+  int bitsPerPixel = 0;
+  
+  CFStringRef pixEnc = CGDisplayModeCopyPixelEncoding(mode);
+  if(CFStringCompare(pixEnc, CFSTR(IO32BitDirectPixels), kCFCompareCaseInsensitive) == kCFCompareEqualTo)
+  {
+    bitsPerPixel = 32;
+  }
+  else if(CFStringCompare(pixEnc, CFSTR(IO16BitDirectPixels), kCFCompareCaseInsensitive) == kCFCompareEqualTo)
+  {
+    bitsPerPixel = 16;
+  }
+  else if(CFStringCompare(pixEnc, CFSTR(IO8BitIndexedPixels), kCFCompareCaseInsensitive) == kCFCompareEqualTo)
+  {
+    bitsPerPixel = 8;
+  }
+  
+  CFRelease(pixEnc);
+  
+  return bitsPerPixel;
+}
+
+// mimic former behavior of deprecated CGDisplayBestModeForParameters
+CGDisplayModeRef BestMatchForMode(CGDirectDisplayID display, size_t bitsPerPixel, size_t width, size_t height, boolean_t &match)
+{
+  // Get a copy of the current display mode
+  CGDisplayModeRef displayMode = CGDisplayCopyDisplayMode(kCGDirectMainDisplay);
+  
+  // Loop through all display modes to determine the closest match.
+  // CGDisplayBestModeForParameters is deprecated on 10.6 so we will emulate it's behavior
+  // Try to find a mode with the requested depth and equal or greater dimensions first.
+  // If no match is found, try to find a mode with greater depth and same or greater dimensions.
+  // If still no match is found, just use the current mode.
+  CFArrayRef allModes = CGDisplayCopyAllDisplayModes(kCGDirectMainDisplay, NULL);
+  for(int i = 0; i < CFArrayGetCount(allModes); i++)	{
+    CGDisplayModeRef mode = (CGDisplayModeRef)CFArrayGetValueAtIndex(allModes, i);
+    
+    if(DisplayBitsPerPixelForMode(mode) != bitsPerPixel)
+      continue;
+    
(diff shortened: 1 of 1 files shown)


Memphiz:
JeNkIns build this please

xbmc/windowing/osx/WinSystemOSX.mm
```diff
@@ -227,6 +227,79 @@ CGDirectDisplayID GetDisplayID(int screen_index)
...
+  // Loop through all display modes to determine the closest match.
+  // CGDisplayBestModeForParameters is deprecated on 10.6 so we will emulate it's behavior
+  // Try to find a mode with the requested depth and equal or greater dimensions first.
+  // If no match is found, try to find a mode with greater depth and same or greater dimensions.
+  // If still no match is found, just use the current mode.
+  CFArrayRef allModes = CGDisplayCopyAllDisplayModes(kCGDirectMainDisplay, NULL);
+  for(int i = 0; i < CFArrayGetCount(allModes); i++)	{
+    CGDisplayModeRef mode = (CGDisplayModeRef)CFArrayGetValueAtIndex(allModes, i);
+    
+    if(DisplayBitsPerPixelForMode(mode) != bitsPerPixel)
```
Rechi: sign miss match, either change `DisplayBitsPerPixelForMode(...)` to return `size_t` or `bitsPerPixel` to `int`

Memphiz:
warning fixed @Rechi

MartijnKaijser:
ping @FernetMenta 
jenkins build this please

MartijnKaijser:
jenkins build this please

Memphiz:
I think this can go in then...

xbmc/windowing/osx/WinSystemOSX.mm
MartijnKaijser: @Memphiz you missed this comment?

Memphiz: Nope - i fixed that compile warning in the last commit of this very PR 10 days ago

Rechi: @MartijnKaijser was fixed, this is only the place where the warning occurred, but not the one which needed to be changed

ksooo:
Meh. This causes all kind of trouble on my late 2016 MBP (on macOS Sierra 10.12.5) - from only partially visible kodi UI to crashes.

Memphiz:
@ksooo how sure are you about that. Because it is not meant to change any behavior at all. Can you post a debug log for me? I tried it too of course with macmini and hackbook.

ksooo:
Pretty sure. ;-) maybe related to retina display?

debug log for a startup which ends with a crash: https://pastebin.com/RA4Zeaua
stack trace for the crash: https://pastebin.com/vJu1FqC4

Memphiz:
could you post a debug log with an older version before this change went in? It might be that the resolutions that are returned differ from the old api - thats what i need to check. Sorry for the inconvenience ...

Memphiz:
It tries to find a match for 1680 x 1050 @ 60Hz ... but in the iterated resolutions with the new api there is only a 1650 x 1050 @ 60 Hz (that 1650 looks wrong to me) ... thats why it crashes ... also please tell me which resolution is really selected in osx ...

ksooo:
debug log prior this PR: https://pastebin.com/GeY51EMY
screen resolution selected in Kodi: 1680x1050p
screen resolution selected in macOS: "Looks like 1680x1050" => 
screen shot 2017-05-22 at 22 32 20

Memphiz:
yap - 1680x1050 is not returned anymore ... can you try to remove your guisettings.xml and tell me if it startsup again then?

Memphiz:
@ksoo also please make a backup of your guisettings.xml before (hope its not to late). If deleting the guisettings.xml fixes it - please try again with the broken guisettings.xml and this code change:

https://pastebin.com/gu3LpcP6

It basically adds a last resort when we can't find a matching resolution from settings - it falls back to the resolution that was determined from the current screen (that should always be available of course). This is a transition migration problem i guess because your guisettings.xml doesn't "match" the hardware anymore due to the changed behavior of the resolution enumeration API.

ksooo:
yes, removing guisettings.xml solves the problem. no more crashes, correct kodi display dimensions.

ksooo:
> yes, removing guisettings.xml solves the problem. no more crashes, correct kodi display dimensions.

oops. tested without your PR. So, no good. Will retest this evening.

ksooo:
Still crashes after removing guisettings.xml
Will test your code change tonight.

ksooo:
Well, if a fresh guisettings.xml does not cure the problem, your hotfix will not help, right? I can save the time for testing this I guess.

Memphiz:
That depends if it still can't find a valid resolution.

ksooo:
My guisettings.xml (the"broken" one) does not contain any display dimension values. So, removing the file cannot fix anything, imo.

Memphiz:
Mhh - well if the crash is still because of not finding a match for the resolution - my patch should ensure that a match is found in theory.

ksooo:
Okay, will try out tonight. Not at home currently.

ksooo:
Does not work. See Slack for details.

PR #12066, opened 2017-05-06
Make passed options to ffmpeg works in all case.

Woddy321:
## Description
Problem here http://forum.kodi.tv/showthread.php?tid=312616

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
@@ -386,7 +386,7 @@ bool CDVDDemuxFFmpeg::Open(CDVDInputStream* pInput, bool streaminfo, bool filein
 
     m_pFormatContext->pb = m_ioContext;
 
-    AVDictionary *options = NULL;
+    AVDictionary *options = GetFFMpegOptionsFromInput();
     if (iformat->name && (strcmp(iformat->name, "mp3") == 0 || strcmp(iformat->name, "mp2") == 0))
     {
       CLog::Log(LOGDEBUG, "%s - setting usetoc to 0 for accurate VBR MP3 seek", __FUNCTION__);


xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
```diff
@@ -386,7 +386,7 @@ bool CDVDDemuxFFmpeg::Open(CDVDInputStream* pInput, bool streaminfo, bool filein
 
     m_pFormatContext->pb = m_ioContext;
 
-    AVDictionary *options = NULL;
+    AVDictionary *options = GetFFMpegOptionsFromInput();
```
FernetMenta: ffmpeg options are only supposed to be set for inputStreamFFmpeg

Razzeee: @FernetMenta could you be a bit more specific what that means for this PR?

FernetMenta: that means this change is wrong. ffmpeg options are not for custom io handling.

Woddy321: @FernetMenta , have you read the log in the post http://forum.kodi.tv/showthread.php?tid=312616.
I have same problem, if ffmpeg options are not for custom io handling can you tell me how to pass an User-Agent or a cookie ?
Options are perhaps not for them, but needed.

FernetMenta: fix handling of m3u that it gets picked up by imputstreamffmpeg

Woddy321: Hu, I need more explanations.
You are speaking about DVDInputStreamFFmpeg ?
I have made another try, with a working m3u, and again it's managed by CDVDDemuxFFmpeg (from debug output).

FernetMenta: CDVDDemuxFFmpeg reads data from an inputstream, CDVDInputStreamXYZ. ffmpeg options are set if input stream is DVDInputStreamFFmpeg. 
In the case you linked, the m3u is not opened by DVDInputStreamFFmpeg, otherwise the options would have been set in the block above.

btw: forum posts with truncated logs are in gerneral ignored

Woddy321: I can give a log too if you want but PR is not the good place for that.
Nevermind, you are right, In fact kodi use the extension to "choose its method", if I rename the playlist file with the .m3u8 extension all will works fine.

I m not skilled enough to check why mimetype doesn't work, but it's another problem, so you can set this PR to invalid.

But I have a last question, I haven't finished to read all the code tracking, but where is the problem to use the ffmpeg option with custom io handling ? If I put a for exemple a cookie in command option (with kodi api), it's to use it in all cases, whatever the engine the core will use.

Razzeee:
I'm going to close this as you asked for @Woddy321 
Either way, thanks for the contribution and resulting discussions :)

PR #11739, opened 2017-02-24
changes for Windows x64

Rechi:
## Description
- Registers are different on win32 and x64
- `_USE_32BIT_TIME_T` can't be used on x64
- calling ` _controlfp(...)` on x64 lets the program crash, also most likely has no benefit on win32
- linking with `yajl.lib` is already handled at a  different place in cmake for all platforms

## Motivation and Context
Some changes which are need to get windows x64 compiled.

## How Has This Been Tested?
Changes have been in my local branch for longer time.

## Types of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [x] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #12069, opened 2017-05-07
[PVR] Fix time_t db insertion (typecast time_t -> uint)

peak3d:
[PVR] Fix time_t db insertion (typecast time_t -> uint)

## Description
Win32's time_t is 64 bit, %u for time_t fails in epg and channel insertion with my win32 build.
This PR typecasts the time_t values used into the type expected in formatting string.

## Motivation and Context
Fix crash on EPG update / pvr stall at kodi startup

## Types of change

ping @ksooo

### DIFF
--- a/xbmc/pvr/PVRDatabase.cpp
+++ b/xbmc/pvr/PVRDatabase.cpp
@@ -578,10 +578,10 @@ bool CPVRDatabase::Persist(CPVRChannelGroup &group)
     /* insert a new entry when this is a new group, or replace the existing one otherwise */
     if (group.GroupID() <= 0)
       strQuery = PrepareSQL("INSERT INTO channelgroups (bIsRadio, iGroupType, sName, iLastWatched, bIsHidden, iPosition) VALUES (%i, %i, '%s', %u, %i, %i)",
-          (group.IsRadio() ? 1 :0), group.GroupType(), group.GroupName().c_str(), group.LastWatched(), group.IsHidden(), group.GetPosition());
+          (group.IsRadio() ? 1 :0), group.GroupType(), group.GroupName().c_str(), static_cast<unsigned int>(group.LastWatched()), group.IsHidden(), group.GetPosition());
     else
       strQuery = PrepareSQL("REPLACE INTO channelgroups (idGroup, bIsRadio, iGroupType, sName, iLastWatched, bIsHidden, iPosition) VALUES (%i, %i, %i, '%s', %u, %i, %i)",
-          group.GroupID(), (group.IsRadio() ? 1 :0), group.GroupType(), group.GroupName().c_str(), group.LastWatched(), group.IsHidden(), group.GetPosition());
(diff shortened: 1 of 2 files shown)


Jalle19 (commented):
Isn't there a better modifier we could use to avoid casting everywhere?

ksooo:
Wondering why not every win32 build user experiences those crashes/stalls? Some if affected functions should be called unconditionally as soon as pvr is enabled.

peak3d:
AFAIK there is no format specifier for time_t.
If you push time_t on the list of va_args it has to meet the sizes of the format string.

But yes, @ksooo I was also surprised about the heavy crashes I never realized before.
But, IIRC I never tried PVR on my win machine before.

Behavier is currently (empty dp):
- kodi starts well and db is filled.
- after one minute epg want to update -> crash (access on invalid string when formattig)
- after kodi restart: INSERT CHANNEL segfaults every time with the same issue.

Could be that I use a newer WinSDK than on jenkins build machine and time_t has changed.

peak3d:
Just to avoid a long discussion:

- We all agree that time_t can be 32 or 64 bit, do we?
- We all agree that %u is wrong for 64 bit data types, doesn't we?

So it has to be changed in any way, solution one the typecast as done in the commit, solution 2: change the datatypes / return values of the function to unsigned int instead time_t., solution 3: use std::stringstream for formatting.

We only have to decide to one of those 3.

Rechi:
`time_t` was 32 bit before #11739 was merged.

peak3d:
thanx for sorting this out @Rechi

Paxxi:
I think we should avoid casting to 32bit,won't be an issue until 2038 but it would feel cleaner to cast to uint64_t everywhere

Rechi:
I also think there is a better solution then casting.
@Paxxi it gets casted to **unsigned** int so the problematic year will be 2106.

xhaggi:
We have a platform specific define (PRIu64) for unsigned long long (64bit) formatting.
Cast time_t to unsigned long long and use `%" PRIu64"`.

xhaggi:
BTW we should think about to drop the time_t here and use our datetime class and datetime sql time as we do it in our video db.

ksooo:
> BTW we should think about to drop the time_t here and use our datetime class and datetime sql type as we do it in our video db.

 +1

Paxxi:
@Rechi I thought a time_t was unsigned but yeah in that case I don't have any objections really.

peak3d:
Can we pls. agree to one of both:

1.) leave the PR as it is (90 years safe)
2.) 
> Cast time_t to unsigned long long and use %" PRIu64"
(Safe until end of everything)

Does anyone of us believe that we'll have live-tv in 90 years?

Rechi:
I'm fine with both although version 2 (no casting) is my preferred one.

peak3d:
version 2 needs casting in the other direction.
because on linux 32 we still have 32-bit time_t

Rechi:
Just seen it, pick the version you prefer.

ksooo:
I'm fine with option 1, merge the PR as it is now.

fritsch:
For history:

Dear people that come to this world after us. If you find that commit and it bites you in the ass, sit down, drink a beer and be happy that this part of the code survived that long.

ksooo:
jenkins build this please

PR #12074, opened 2017-05-10
[GUI] avoid unneccesary work

peak3d:
Reduce cpu usage when checking for renderchanges

## Description
Kodi on idle eats a lot of cpu when checking for GUI changes.
Application::FrameMove() calls Process() of the active window and this function iterates through all the skin controls. Profiling this task leads to 2 places where we spend a lot of work for nothing.

This PR changes the idle load of an empty kodi home view from ~12-18% CPU to ~6% CPU usage.
(Android / WETEK HUB)

1.) avoid SendMessage if scrollbar of Grouplist has not changed. SendMessage is one of the most time consuming things, as it iterates again through all the windows to find other receivers.

2.) GUIControl UpdateVisibility; added missing sign that Item is already pushed into the contol info list.

## Motivation and Context
Reduce heat / power consumtion if kodi is in idle

### DIFF
--- a/xbmc/guilib/GUIControlGroupList.cpp
+++ b/xbmc/guilib/GUIControlGroupList.cpp
@@ -35,6 +35,7 @@ CGUIControlGroupList::CGUIControlGroupList(int parentID, int controlID, float po
   m_totalSize = 0;
   m_orientation = orientation;
   m_alignment = alignment;
+  m_lastScrollerValue = 0;
   m_useControlPositions = useControlPositions;
   ControlType = GUICONTROL_GROUPLIST;
   m_minSize = 0;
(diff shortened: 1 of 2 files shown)


MartijnKaijser:
i guess ping @FernetMenta and @popcornmix

popcornmix:
+1 in principle. 
But on testing the movies/poster view is broken with this - the selected poster doesn't update when scrolling through movies.

popcornmix:
Okay without the ``m_pushedUpdates = true`` change.

peak3d:
@popcornmix ok, will investigate, this scroller change was very experimental.
I'll install the ned_scot fake library and look for a safer way, Thx,  for testing.

peak3d:
Will be included in next bigger GUI PR -> closed

PR #12080, opened 2017-05-12
[vpupdates] don't enable vdpau on raspberry-pi

Rechi:
@FernetMenta @peak3d This adds the required switch for enabling vdpau and should fix the RPi compile issue.

@wsnipex Is this okay for you?

### DIFF
--- a/tools/depends/target/ffmpeg/Makefile
+++ b/tools/depends/target/ffmpeg/Makefile
@@ -26,7 +26,10 @@ ifeq ($(CROSS_COMPILING), yes)
 endif
 ifeq ($(OS), linux)
   ffmpg_config += --target-os=$(OS) --cpu=$(CPU)
-  ffmpg_config += --enable-vdpau --enable-vaapi --enable-pic
+  ifneq ($(TARGET_PLATFORM),raspberry-pi)
+    ffmpg_config += --enable-vdpau --enable-vaapi
+  endif
+  ffmpg_config += --enable-pic
 endif
 ifeq ($(OS), android)
   ifeq ($(findstring arm64, $(CPU)), arm64)


peak3d (approved):
Looks as expected for me 👍

wsnipex:
nope sorry, depends configure must never care about single dependencies feature switches.

wsnipex (changes requested):
this has to be fixed in the ffmpeg makefile only

Rechi:
changed to only enable vdpau if TARGET_PLATFORM doesn't equal raspberry-pi
@FernetMenta your decision if you accept it

tools/depends/target/ffmpeg/Makefile
```diff
@@ -26,7 +26,10 @@ ifeq ($(CROSS_COMPILING), yes)
 endif
 ifeq ($(OS), linux)
   ffmpg_config += --target-os=$(OS) --cpu=$(CPU)
-  ffmpg_config += --enable-vdpau --enable-vaapi --enable-pic
+  ifneq ($(TARGET_PLATFORM),raspberry-pi)
+    ffmpg_config += --enable-vdpau
+  endif
+  ffmpg_config += --enable-vaapi --enable-pic
```
wsnipex: Vaapi doesn't cause issues for rbpi?

Rechi: can't test it on my machine, have no crosscompiler for rpi

Rechi:
jenkins build this please

FernetMenta:
@Rechi it is not my decision in this case. merge if you feel happy with it.
if I were in charge of this, I would solve it differently. why do you enable a feature in the first step for all targets and than disable it for those you have issues with? I would enable features only for those targets the feature is supposed to work. if you want to build for imx, you most likely will run into the same issue and have to touch this file again.
you'll end up with something like this
https://github.com/xbmc/xbmc/blob/master/xbmc/windowing/egl/CMakeLists.txt#L1

tools/depends/target/ffmpeg/Makefile
stefansaraev: this one probably not used at all on rbpi. hint tools/depends/target/ffmpeg/CMakeLists.txt

Rechi:
According to @wsnipex we won't move all those checks into the configure which will be only needed to ffmpeg configure.
I have no strong opinion on how this should be fix properly.

tools/depends/target/ffmpeg/Makefile
stefansaraev: in my opinion that "let's rebuild ffmpeg if we find a compatible one" thing is not compatible with depends. FindFFMPEG.cmake should fail and let distribution maintainers fix / update their ffmpeg instead of silently building it. but hey, it's just an opinion.

Rechi: rbpi gets build with depends and doesn't use https://github.com/xbmc/xbmc/blob/master/cmake/modules/FindFFMPEG.cmake#L244-L264.
If building depends fails you will notice it before building kodi.

stefansaraev: then --enable-vaapi is definitely going to fail on that special so-called "platform". how didn't it fail before ?

Rechi: ffmpeg have changed the configure script since version 3.1
they just ignored the options if the needed libraries couldn't be found

stefansaraev: --enable-vaapi definitely fails while building for amlogic here. but if you are right and ./configure silently passes --enable-vaapi on rbpi for some weird reason, we have bigger problem. and fernet's "we dont want auto detection" argument is invalid ;)

Rechi: I don't know if it silently passes --enable-vdpau or this also needs to be ecluded on rbpi, but what I know for sure is, it silently passed --enable-vdpau in version 3.1 and now it fails while searching for vdpau headers.

Rechi:
jenkins build this please

tools/depends/target/ffmpeg/Makefile
stefansaraev: sorry, it seems I didn't express myself right (you can blame the language barrier), I didn't mean that ffmpeg configure defaults changd. I meant whether building with --enable-vdpau / vaapi silently succeeds without va/vdpau headers present. that's definitely not the case for both va/vdpau in 3.3. so --enable-vaapi should fail on rbpi as well, you should move it to the ifneq block

Rechi: just noticed it and started a new rbpi test build

Rechi:
rbpi compiles now

wsnipex:
jenkins build this please.
Lets get this in to finally pull vpudates. But yes, this needs refactoring later.

MartijnKaijser:
@wsnipex it was already compiled
http://jenkins.kodi.tv/job/LINUX-RBPI/3573/
http://jenkins.kodi.tv/view/Automation/job/BuildMulti-PR/2686/

PR #12082, opened 2017-05-12
[Music]Fix Cue sheet regression and improve handling

DaveTBlake:
There are various issues with the way that cuesheets are handled, some are regressions since v16 related to changes to the way the current playlist works or broken the music library and files view were unified by #8011. All came up during testing to make the songs node as efficient as possible (within the limits of the current CFileItem list overall design), see #11785

- Tracks were being wrongly looked up e.g. playback of the last track on a file with an embedded cuesheet and both the first and last songs get highlighted (either from library or fileview), and next not work from first song. The incorrect array access could be a crash source too.
- Odd behaviour when playing music in a file with an embedded cuesheet from fileview. Number of items on the current playlist would increase exponentially (!) every time the screen was switched to that view. The cue document was being expanded recurrsively.
 -   On play from fileview of the folder containing a .cue file that has been scanned into the library, all the tracks on the related music file are played, but the current playlist displays the details (title, track number, artist, duration etc.) of just the first song.
  -  On playback of both kinds of cuesheet songs from the library the replay gain values are not applied (so the song node delaying reading of embedded cuesheet data from DB was disabled in v17 as it was not doing anything useful).

The way embedded cuesheet handling was implemented for v15, using the music db, was taking 7% of the total time taken to process the songs node, even if there were no cuesheets in the music collection. This was all in order get replay gain for embedded cuesheets, that were then only applied in certain playback paths, and as a minority use clearly not worth the overhead. Disabled as a quick work around in v17, this PR redesigns how replay gain for both kinds of cuesheet songs is implemented.

There appeared to be two possible approaches to applying the replay gain values for cuesheet songs in the library: 

1.  Get the player to process any cue document fetching rgain values, at the same point that it scans ordininary music files for metatdata.
2.  Store the rgain values in the db.

The latter seems the best way to avoid additional repeated processing of cue documents during playback, so the schema change stores any replaygain values in the song table. This avoids LEFT joins (which are slow on SQLite and worth avoiding), but most importantly it avoids the original separate query per song approach that is so inefficient on larger music collections. As 4 float values (8 byte each), never used in music slection or filtering, it is most space efficient to store the album and track gain and peak values as a single string.


DaveTBlake:
This has been tested by users and BigNoid , no regerssions, so looking to merge by end of week if no one complains.

@Paxxi perhaps you could check my code if you have a moment

DaveTBlake:
jenkins build this please

xbmc/CueDocument.cpp
```diff
@@ -194,9 +194,41 @@ void CCueDocument::GetSongs(VECSONGS &songs)
...
   for (size_t i = 0; i < m_tracks.size(); ++i)
   {
-    CSong song;
-    GetSong(i, song);
-    songs.push_back(song);
+    CSong aSong;
+    const CCueTrack& track = m_tracks[i];
+    //Pass artist to MusicInfoTag object by setting artist description string only.
+    //Artist credits not used during loading from cue sheet.
+    if ((track.strArtist.length() == 0) && (m_strArtist.length() > 0))
```
Paxxi: I would use `track.strArtist.empty() && !m_strArtist.empty()`

xbmc/CueDocument.cpp
```diff
@@ -194,9 +194,41 @@ void CCueDocument::GetSongs(VECSONGS &songs)
...
+    //Pass album artist to MusicInfoTag object by setting album artist vector. 
+    aSong.SetAlbumArtist(StringUtils::Split(m_strArtist, g_advancedSettings.m_musicItemSeparator));
+    aSong.strAlbum = m_strAlbum;
+    aSong.genre = StringUtils::Split(m_strGenre, g_advancedSettings.m_musicItemSeparator);
+    aSong.iYear = m_iYear;
+    aSong.iTrack = track.iTrackNumber;
+    if (m_iDiscNumber > 0)
+      aSong.iTrack |= (m_iDiscNumber << 16); // see CMusicInfoTag::GetDiscNumber()
+    if (track.strTitle.length() == 0) // No track information for this track!
+      aSong.strTitle = StringUtils::Format("Track %2d", track.iTrackNumber);
```
Paxxi: use fmt syntax, either `{}` if 2 digits isn't important or `{:2d}` if it is

xbmc/CueDocument.cpp
```diff
@@ -194,9 +194,41 @@ void CCueDocument::GetSongs(VECSONGS &songs)
...
+    else
+      aSong.strTitle = track.strTitle;
+    aSong.strFileName = track.strFile;
+    aSong.iStartOffset = track.iStartTime;
+    aSong.iEndOffset = track.iEndTime;
+    if (aSong.iEndOffset)
+      aSong.iDuration = (aSong.iEndOffset - aSong.iStartOffset + 37) / 75;
```
Paxxi: maybe add some comment for these magic values

xbmc/music/tags/ReplayGain.cpp
```diff
@@ -74,6 +75,35 @@ void ReplayGain::SetPeak(Type aType, float aPeak)
...
 }
 
+std::string ReplayGain::Get() const
+{
+  if (!Get(ALBUM).Valid() && !Get(TRACK).Valid())
+    return StringUtils::Empty;
+
+  std::string rg;
+  if (Get(ALBUM).Valid())
+    rg = StringUtils::Format("%.3f,%.3f,", Get(ALBUM).Gain(), Get(ALBUM).Peak());
```
Paxxi: use fmt formatting here as well `{:.3f}`

see more at http://fmtlib.net/latest/syntax.html

xbmc/music/tags/ReplayGain.cpp
```diff
@@ -74,6 +75,35 @@ void ReplayGain::SetPeak(Type aType, float aPeak)
     m_data[index(aType)].SetPeak(aPeak);
 }
 
+std::string ReplayGain::Get() const
+{
+  if (!Get(ALBUM).Valid() && !Get(TRACK).Valid())
+    return StringUtils::Empty;
```
Paxxi: use `return std::string()` the Empty is some legacy thing that should be removed

Paxxi (commented):
Overall looks good to me.

xbmc/CueDocument.cpp
DaveTBlake: I didn't design this just rearranged the code, so I don't know why it is like this!

MartijnKaijser: Perhaps just add a todo or whatever comment so others could fill it in later

DaveTBlake:
Thanks @Paxxi for C++ improvements, code was historic but good to get it fixed

jenkins build this please

MartijnKaijser:
jenkins build this please

PR #12083, opened 2017-05-12
[droid] Define that we possibly use the microphone in Kodi

MartijnKaijser:
## Description
Add the fact that we can use the microphone to the manifest file

## Motivation and Context
We should inform users
https://developer.android.com/guide/topics/manifest/uses-feature-element.html

## How Has This Been Tested?
testing?

## Types of change

## Checklist:

### DIFF
--- a/tools/android/packaging/xbmc/AndroidManifest.xml.in
+++ b/tools/android/packaging/xbmc/AndroidManifest.xml.in
@@ -13,8 +13,10 @@
     <uses-permission android:name="android.permission.GET_TASKS" />
     <uses-permission android:name="android.permission.BLUETOOTH" />
     <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />
+    <uses-permission android:name="android.permission.RECORD_AUDIO" />
 
     <uses-feature android:name="android.hardware.bluetooth" android:required="false" />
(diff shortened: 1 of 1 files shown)


koying:
Assuming it is in the context of speech recog, it's not needed because *we* don't record audio: we call a system service for speech-to-text. We don't see the first bit of it being recorded.

RECORD_AUDIO is needed if we actually do recording, with https://developer.android.com/guide/topics/media/mediarecorder.html for instance.

The perm doesn't do any harm per se, ofc, but can be misleading and lead to unneeded anguish (and potentially stupid clickbait articles ;) )

MartijnKaijser:
That was also the thing that crossed my mind :)

Hedda:
@koying @MartijnKaijser what about Kodi (python) addons that want to use the microphone for new features such as Google Assistant and Amazon Alexa, or an karaoke addon? 

Example: https://developers.google.com/assistant/sdk/develop/python/

Would Kodi not need to declare possibly use the microphone in the manifest file for Kodi's (python) addons to be able to access and use the microphone on Android?

koying:
An app *never* uses the microphone in the context of Assistant or Alexa.
The relevant system services do.

I don't see how a python addon could use a microphone under Android unless we create an interface for it, so it's out-of-scope for now.

PR #12090, opened 2017-05-14
DVDVideoCodecFFmpeg: Consistency fixes (buildsys rework in progress)

fritsch:
DVDVideoCodecFFmpeg currently needs both VAAPI and VDPAU dependencies to succesfully compile, therefore it conflicts with given CMAKE parameters to disable either or on X11.

That's not a fix to make it buildable for each alone, but a consistency workaround in order to reduce the massive ifdeffery.

For people with MIR / DRM/KMS this needs changes.

To my knowledge the buildsystem change that revisits "kodi platforms" is on its way.

@lrusak for information

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp
@@ -1162,7 +1162,11 @@ IHardwareDecoder* CDVDVideoCodecFFmpeg::GetHWAccel()
 }
 
 //------------------------------------------------------------------------------
-// temporary
+// The current way of handling build targets is way too complicated.
+// It's basically a long mess of combining options, which needs revisiting
+// This factory is temporary in its current state. That means certain
+// cmake options to remove certain decoders in certain combinations won't work
+// until the build system is revisited.
 //------------------------------------------------------------------------------
 
 #ifdef HAS_DX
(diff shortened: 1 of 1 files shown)


xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp
```diff
@@ -1176,12 +1180,10 @@ IHardwareDecoder* CDVDVideoCodecFFmpeg::CreateVideoDecoderHW(AVPixelFormat pixfm
...
-#if defined(HAVE_LIBVA) || defined(HAVE_LIBVDPAU)
-#if defined(HAVE_LIBVA)
+// Linux X11 - Currently works only if both
+// VAAPI and VDPAU dependencies are available
+#if defined(HAVE_LIBVA) && defined(HAVE_LIBVDPAU)
 #include "VAAPI.h"
-#endif
-#if defined(HAVE_LIBVDPAU)
 #include "VDPAU.h"
 #endif
```
stefansaraev: the ```#endif``` here should be dropped.

fritsch: Nope?

stefansaraev: I think you do, or I am blind today :) see L1201

fritsch: You are right, thx - the pita of ifdeffery.

FernetMenta:
+1

lrusak:
Thanks!

wsnipex:
I would prefer putting this workaround in cmake instead, so users actually have a chance to get notified they need to activate both.

FernetMenta:
I would prefer that the buld system is adjusted to meet requirements so that this workaround is not needed anymore.

wsnipex:
how is it the build systems fault that DVDVideoCodecFFmpeg doesn't compile with only vaapi or vdpau?

wsnipex (changes requested):
this needs to be done in cmake with proper error messages. Doing it here is totally opaque to the user and will lead to support burden

lrusak:
We cannot build Rome in a day. Incremental changes are just fine.

We are working on a plan via a feature matrix about the needed cmake
changes.

wsnipex:
we are talking about 3 LOC in cmake to implement what this PR intends in a way that does not replace one bug with another.

wsnipex:
LOL

fritsch:
No - I don't find that stuff funny anymore. This has sandbox character ...

PR #12089, opened 2017-05-14
fix build with -DENABLE_VAAPI=no

herrnst:
## Description
fix:
```
cmake -C ... -DENABLE_VAAPI=no -DENABLE_VDPAU=yes
/var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp: In member function ‘IHardwareDecoder* CDVDVideoCodecFFmpeg::CreateVideoDecoderHW(AVPixelFormat, CProcessInfo&)’:
/var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp:1192:16: error: ‘VAAPI’ does not name a type
     return new VAAPI::CDecoder(m_processInfo);
                ^
/var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp:1192:21: error: expected ‘;’ before ‘::’ token
     return new VAAPI::CDecoder(m_processInfo);
                     ^
/var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp:1192:21: error: ‘::CDecoder’ has not been declared
/var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp:1192:21: note: suggested alternative:
In file included from /var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp:1185:0:
/var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/VDPAU.h:546:7: note:   ‘VDPAU::CDecoder’
 class CDecoder
       ^
[...]

PR #12091, opened 2017-05-14
Revert "DVDVideoCodecFFmpeg: Consistency fixes (buildsys rework in progress)"

fritsch:
Reverts xbmc/xbmc#12090

This is not how the game works. PRs are there for review.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp
@@ -1162,11 +1162,7 @@ IHardwareDecoder* CDVDVideoCodecFFmpeg::GetHWAccel()
 }
 
 //------------------------------------------------------------------------------
-// The current way of handling build targets is way too complicated.
-// It's basically a long mess of combining options, which needs revisiting
-// This factory is temporary in its current state. That means certain
-// cmake options to remove certain decoders in certain combinations won't work
-// until the build system is revisited.
+// temporary
 //------------------------------------------------------------------------------
 
 #ifdef HAS_DX
@@ -1180,11 +1176,14 @@ IHardwareDecoder* CDVDVideoCodecFFmpeg::CreateVideoDecoderHW(AVPixelFormat pixfm
 }
 #endif
 
-// Linux X11 - Currently works only if both
-// VAAPI and VDPAU dependencies are available
-#if defined(HAVE_LIBVA) && defined(HAVE_LIBVDPAU)
+// Linux X11
+#if defined(HAVE_LIBVA) || defined(HAVE_LIBVDPAU)
+#if defined(HAVE_LIBVA)
 #include "VAAPI.h"
+#endif
+#if defined(HAVE_LIBVDPAU)
 #include "VDPAU.h"
+#endif
 
 #define VP_VIDEOCODEC_HW
 IHardwareDecoder* CDVDVideoCodecFFmpeg::CreateVideoDecoderHW(AVPixelFormat pixfmt, CProcessInfo &processInfo)


FernetMenta:
building ffmpeg on linux (not pi) already requires wrapper libs for vdpau + vaapi to be present. in this state it does not harm if this is also required for building kodi.
note that this has nothing to do with behavior, just building. in regard to behavior it does not matter if you build with those libs available at build time.

it would be nice if build system supports options bound to windowing target. once this is in place, the factore can be cleaned up. for now it does not really matter if this pr gets reverted or not. it won't change anything but some error logs during building, if a wrapper is not present.

wsnipex:
it DOES matter as in the current state vaapi or vdpau silents get ignored if the other option is disabled.
This is a bug and a bigger one then before.

candrews:
Since this PR was merged, I'm unable to build on Linux. The error is:
```
/var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp: In member function ‘IHardwareDecoder* CDVDVideoCodecFFmpeg::CreateVideoDecoderHW(AVPixelFormat, 
CProcessInfo&)’:
/var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp:1194:6: error: ‘VDPAU’ has not been declared
   if(VDPAU::CDecoder::IsVDPAUFormat(pixfmt) && CServiceBroker::GetSettings().GetBool(CSettings::SETTING_VIDEOPLAYER_USEVDPAU))
      ^~~~~
/var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp:1195:16: error: ‘VDPAU’ does not name a type
     return new VDPAU::CDecoder(m_processInfo);
                ^~~~~
/var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp:1195:21: error: expected ‘;’ before ‘::’ token
     return new VDPAU::CDecoder(m_processInfo);
                     ^~
/var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp:1195:23: error: ‘::CDecoder’ has not been declared
     return new VDPAU::CDecoder(m_processInfo);
                       ^~~~~~~~
/var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp:1195:23: note: suggested alternative:
In file included from /var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp:1182:0:
/var/tmp/portage/media-tv/kodi-9999/work/kodi-9999/xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.h:410:7: note:   ‘VAAPI::CDecoder’
 class CDecoder
       ^~~~~~~~
```

I'm happy to provide additional details.

FernetMenta:
install libvdpau-dev

wsnipex:
this PR is not the cause.
The issue is that you need to enable both vaapi and vdpau if you want one of those.

herrnst:
@candrews fix was right around the corner at https://github.com/xbmc/xbmc/pull/12089

candrews:
> install libvdpau-dev

But I don't want vdpau; I want vaapi.

> this PR is not the cause.
> The issue is that you need to enable both vaapi and vdpau if you want one of those.

So if I want vaapi, I also need vdpau? (If that's the intended change, fine - I have no judgment, I just want to understand so I can update the Gentoo ebuild appropriately)

Also, If that is true, I feel like https://github.com/xbmc/xbmc/blob/master/CMakeLists.txt should check and error out if vaapi^vdpau, instead of waiting for a compilation error later?

hauihau:
@FernetMenta why does one have to enable VDPAU and VAAPI in order to use one of them? Just ifdef them as before, or whats the problem with two ifdefs?

FernetMenta:
the switch does not enable anything, it never did. if a system has no vdpau it won't get used at runtime, no matter if the switch was set during build.

FernetMenta:
more verbose. I already have a working solution that det cts vaapi at startup and chacks capabilities. that can be used to hide settings if not available. you can always build with vdpau and vaapi. ffmeg gets build the same way.

hauihau:
> the switch does not enable anything, it never did. if a system has no vdpau it won't get used at runtime, no matter if the switch was set during build.

Would you then please mind explaining me, what the purpose of https://github.com/xbmc/xbmc/blob/master/CMakeLists.txt#L142-L143 is? https://github.com/xbmc/xbmc/blob/master/cmake/modules/FindVAAPI.cmake#L50 enables HAVE_LIBVA which leads to the build error (analog to FindVDPAU setting HAVE_LIBVDPAU) as there is no more differentiation because you've changed the file in #11946 in a way that now, both VAAPI and VDPAU devel headers must be available in order to successfully compile kodi even when only one of them are requested. So the system sure now must to have VDPAU. I don't get the point what is wrong with #12089. The two ifdefs doesn't make the code hard to read or hard to maintain.

> more verbose. I already have a working solution that detects vaapi at startup and chacks capabilities. that can be used to hide settings if not available. you can always build with vdpau and vaapi. ffmeg gets build the same way.

"You can" is obviously wrong, as already said, VDPAU must be installed, otherwise you get the compiler errors from #12089.

Detecting what is available at startup should have nothing to do with this issue. We're talking about a build time issue and not runtime. No VDPAU installed => no successful compiler run. I consider this a breakage.

herrnst:
Additional note: FFmpeg always builds perfectly with or without vdpau or vaapi libs/headers installed, be it the system package or the embedded one. FFmpeg never complained about missing libva stuff on me in both cases.

PR #12096, opened 2017-05-15
[gui][refactor] rework custom window's visible condition and modality handling

xhaggi:
This PR serves as preparation for #12035. It refactors the custom window handling so expressions used in visible conditions for custom dialogs are resolved in `CGUIWindow::Load` -> `CGUIInclude::ResolveInclude` too.

The following is done to achieve it.
* add a new boolean in GUIWindow to indicate if the window is a custom
window
* move the visible condition handling into CGUIWindow::Load, because it
support skin expressions there.
* move identification of modeless custom dialogs to CGUIDialog::Load
* rename method LoadUserWindows to LoadCustomWindows to unify wording

@ronie @BigNoid not sure who needs a ping

@rmrector fyi

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -1665,8 +1665,8 @@ bool CApplication::LoadSkin(const std::string& skinID)
 
   CLog::Log(LOGINFO, "  load new skin...");
 
-  // Load the user windows
-  LoadUserWindows();
+  // Load custom windows
+  LoadCustomWindows();
 
   int64_t end, freq;
   end = CurrentHostCounter();
@@ -1755,28 +1755,31 @@ void CApplication::UnloadSkin(bool forReload /* = false */)
 // and as a result a race condition on exit can cause a crash.
 }
 
-bool CApplication::LoadUserWindows()
+bool CApplication::LoadCustomWindows()
 {
   // Start from wherever home.xml is
   std::vector<std::string> vecSkinPath;
   g_SkinInfo->GetSkinPaths(vecSkinPath);
-  for (unsigned int i = 0;i < vecSkinPath.size();++i)
+
+  for (const auto &skinPath : vecSkinPath)
   {
-    CLog::Log(LOGINFO, "Loading user windows, path %s", vecSkinPath[i].c_str());
+    CLog::Log(LOGINFO, "Loading custom window XMLs from skin path %s", skinPath.c_str());
+
     CFileItemList items;
-    if (CDirectory::GetDirectory(vecSkinPath[i], items, ".xml", DIR_FLAG_NO_FILE_DIRS))
+    if (CDirectory::GetDirectory(skinPath, items, ".xml", DIR_FLAG_NO_FILE_DIRS))
     {
-      for (int i = 0; i < items.Size(); ++i)
+      for (const auto &item : items.GetList())
       {
-        if (items[i]->m_bIsFolder)
+        if (item->m_bIsFolder)
           continue;
-        std::string skinFile = URIUtils::GetFileName(items[i]->GetPath());
+
+        std::string skinFile = URIUtils::GetFileName(item->GetPath());
         if (StringUtils::StartsWithNoCase(skinFile, "custom"))
         {
           CXBMCTinyXML xmlDoc;
-          if (!xmlDoc.LoadFile(items[i]->GetPath()))
+          if (!xmlDoc.LoadFile(item->GetPath()))
           {
-            CLog::Log(LOGERROR, "unable to load: %s, Line %d\n%s", items[i]->GetPath().c_str(), xmlDoc.ErrorRow(), xmlDoc.ErrorDesc());
+            CLog::Log(LOGERROR, "Unable to load custom window XML %s. Line %d\n%s", item->GetPath().c_str(), xmlDoc.ErrorRow(), xmlDoc.ErrorDesc());
             continue;
           }
 
@@ -1785,13 +1788,14 @@ bool CApplication::LoadUserWindows()
           std::string strValue = pRootElement->Value();
           if (!StringUtils::EqualsNoCase(strValue, "window"))
           {
-            CLog::Log(LOGERROR, "file: %s doesnt contain <window>", skinFile.c_str());
+            CLog::Log(LOGERROR, "No <window> root element found for custom window in %s", skinFile.c_str());
             continue;
           }
 
-          // Read the <type> element to get the window type to create
+          int id = WINDOW_INVALID;
+
+          // Read the type attribute or element to get the window type to create
           // If no type is specified, create a CGUIWindow as default
-          CGUIWindow* pWindow = NULL;
           std::string strType;
           if (pRootElement->Attribute("type"))
             strType = pRootElement->Attribute("type");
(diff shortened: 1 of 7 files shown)


rmrector:
This looks better than my goofy attempt, but modeless custom dialogs aren't being initialized automatically. Once I run an explicit `ActivateWindow(1199)` for the dialog it displays and behaves as it should, though.

xhaggi:
Will take a look at it, thanks for testing.

xhaggi:
Windows or dialogs are loaded on first activation or their load type is `LOAD_ON_GUI_INIT`, which is not the case for custom dialogs (`KEEP_IN_MEMORY`), or (before this changes) their visible condition evaluates to true. As we now set the visible condition while a custom dialog is loaded, custom modeless dialogs are ignored which normally would be open in `CGUIDialog::UpdateVisibility` by its visible condition. So we need to call `CGUIWindow::Initialize` right after we added the custom window/dialog to our window manager.

xhaggi:
@rmrector I've updated the PR. Could you please give it another try?

xhaggi:
I did some runtime tests and updated the PR again. We can't initialize the custom dialogs in `CApplication::LoadCustomWindows` because our window manager is not ready there. Now I change the load type of custom dialogs holding a visible condition to `LOAD_ON_GUI_INIT` and they will be loaded by our window manager. This makes more sense as it follows the window load policy.

BigNoid:
With the latest changes custom dialogs are loaded now when they have a visible tag.

xhaggi:
@BigNoid thanks for testing.

xhaggi:
Again a small update. I noticed that setting the visible condition already exists in `CGUIWindow::Load` at https://github.com/xbmc/xbmc/pull/12096/files#diff-f81df6255b1c6016e7b26a295be24bf3R207. I removed my dupe code now.

xhaggi:
@Paxxi as discussed I've added another commit to simplify the `CGUIWindow::Load` method.

xbmc/Application.cpp
```diff
@@ -1801,38 +1805,56 @@ bool CApplication::LoadUserWindows()
...
+
+          int windowId = id + WINDOW_HOME;
+          if (id == WINDOW_INVALID || g_windowManager.GetWindow(windowId))
+          {
+            // No id specified or id already in use
+            CLog::Log(LOGERROR, "No id specified or id already in use for custom window in %s", skinFile.c_str());
+            continue;
+          }
+
+          CGUIWindow* pWindow = NULL;
```
Paxxi: use `nullptr`

xbmc/guilib/GUIWindow.cpp
```diff
@@ -141,16 +142,28 @@ bool CGUIWindow::LoadXML(const std::string &strPath, const std::string &strLower
...
-    m_windowXMLRootElement = (TiXmlElement*)xmlDoc.RootElement()->Clone();
+    m_windowXMLRootElement = static_cast<TiXmlElement*>(xmlDoc.RootElement()->Clone());
   }
   else
     CLog::Log(LOGDEBUG, "Using already stored xml root node for %s", strPath.c_str());
 
-  return Load(m_windowXMLRootElement);
+  // we must create a copy of the root element as we will manipulate it
+  // when loading and we don't want the original root element to change
+  TiXmlElement* pRootElementCopy = static_cast<TiXmlElement*>(m_windowXMLRootElement->Clone());
```
Paxxi: this Clone seems just wrong. We don't actually modify the xml that I can see and neither should we.

We do touch the pointer but that shouldn't require a clone just to keep our place in the DOM, a regular pointer copy should suffice I think.

xbmc/guilib/GUIWindow.cpp
```diff
@@ -141,16 +142,28 @@ bool CGUIWindow::LoadXML(const std::string &strPath, const std::string &strLower
...
+  bool retVal = Load(pRootElementCopy);
+  delete pRootElementCopy;
```
Paxxi: if we really do need a clone, use a unique_ptr to avoid this setup, then a simple `if (!Load...) return false` avoid the extra nesting below and the extra variable.

xbmc/windows/GUIMediaWindow.cpp
```diff
@@ -103,26 +103,32 @@ CGUIMediaWindow::~CGUIMediaWindow()
...
-    std::vector<std::string> views = StringUtils::Split(allViews, ",");
-    for (std::vector<std::string>::const_iterator i = views.begin(); i != views.end(); ++i)
-    {
-      int controlID = atol(i->c_str());
-      CGUIControl *control = GetControl(controlID);
-      if (control && control->IsContainer())
-        m_viewControl.AddView(control);
+  bool retVal = CGUIWindow::Load(root);
+
+  if (retVal)
```
Paxxi: reverse the check to avoid a layer of nesting

xbmc/windows/GUIMediaWindow.cpp
```diff
@@ -103,26 +103,32 @@ CGUIMediaWindow::~CGUIMediaWindow()
...
+  {
+    // configure our view control
+    m_viewControl.Reset();
+    m_viewControl.SetParentWindow(GetID());
+    TiXmlElement *element = root->FirstChildElement("views");
```
Paxxi: use `auto`

xbmc/windows/GUIMediaWindow.cpp
```diff
@@ -103,26 +103,32 @@ CGUIMediaWindow::~CGUIMediaWindow()
...
+    if (element && element->FirstChild())
+    { // format is <views>50,29,51,95</views>
+      const std::string &allViews = element->FirstChild()->ValueStr();
+      std::vector<std::string> views = StringUtils::Split(allViews, ",");
```
Paxxi: use `auto`

xbmc/windows/GUIMediaWindow.cpp
```diff
@@ -103,26 +103,32 @@ CGUIMediaWindow::~CGUIMediaWindow()
...
+      for (std::vector<std::string>::const_iterator i = views.begin(); i != views.end(); ++i)
```
Paxxi: use for auto loop

Paxxi (commented):
I think the changes look sane.

xbmc/guilib/GUIWindow.cpp
xhaggi: We need to clone here because we alter the XML nodes in `Load`.

xbmc/guilib/GUIWindow.cpp
xhaggi: Thought the same but only moved that code from `Load`.

xhaggi:
Most of the code was just moved, but will address all your comments tommorow ;) Thank you for the review.

xbmc/guilib/GUIWindow.cpp
Paxxi: That seems like a very weird design by someone long ago. I would think that a method called Load would only read from the xml and not modify it in memory. I didn't find any place where this happens during my quick check before commenting though, do you have an example?

xhaggi: We alter the xml while processing includes, constants and expression. Take a look and `ResolveIncludes` in `Load`.

xhaggi: I think it would be better to move the preparation of XML into a separate method `PrepareXML`. The result will be passed into `Load` which do what you expect, only read from XML. That way it should be clear what happens.

xhaggi:
@Paxxi moved the load refactoring into its own PR #12121.

jenkins build this please

xhaggi:
jenkins build this please

PR #12097, opened 2017-05-15
[Android MediaCodecSurface] Change Rendertype from GUILayer to VideoLayer

peak3d:
see title

## Description
Currently MediaCodecSurface is rendered as GUILayer.
This is not correct because rendering takes place on an own surface layer,
This PR removes the binding of RendererMediaCodecSurface to LinuxRendererGLES and tells kodi that Rendering is NOT GUI but VideoPlane.

## Motivation and Context
Video is not visible if home GUI is opened above a running video.

## How Has This Been Tested?
Android / play video / press [ESC] / Esturay Skin

## Types of change

## Notes
- Not tested on mobile phone (screen rotation)
- I have removed some 3D stuff, maybe it breaks already working 3D functionality.

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
@@ -1,5 +1,5 @@
 /*
- *      Copyright (C) 2007-2015 Team Kodi
+ *      Copyright (C) 2007-2017 Team Kodi
  *      http://kodi.tv
  *
  *  This Program is free software; you can redistribute it and/or modify
@@ -18,23 +18,61 @@
  *
  */
 
+#if defined(TARGET_ANDROID)
+
 #include "RendererMediaCodecSurface.h"
 
-#if defined(TARGET_ANDROID)
 #include "../RenderCapture.h"
-
+#include "settings/MediaSettings.h"
 #include "platform/android/activity/XBMCApp.h"
 #include "DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.h"
 #include "utils/log.h"
+#include <thread>
 
 CRendererMediaCodecSurface::CRendererMediaCodecSurface()
+  : m_bConfigured(false)
+  , m_iRenderBuffer(0)
 {
 }
 
 CRendererMediaCodecSurface::~CRendererMediaCodecSurface()
 {
-  for (int i(0); i < m_NumYV12Buffers; ++i)
+  for (int i(0); i < m_numRenderBuffers; ++i)
     ReleaseBuffer(i);
+}
+
+bool CRendererMediaCodecSurface::Configure(unsigned int width, unsigned int height, unsigned int d_width, unsigned int d_height, float fps, unsigned flags, ERenderFormat format, void *hwPic, unsigned int orientation)
+{
+  CLog::Log(LOGNOTICE, "CRendererMediaCodecSurface::Configure");
+
+  m_sourceWidth = width;
+  m_sourceHeight = height;
+  m_renderOrientation = orientation;
+
+  // Save the flags.
+  m_iFlags = flags;
+  m_format = format;
+
+  // Calculate the input frame aspect ratio.
+  CalculateFrameAspectRatio(d_width, d_height);
+  SetViewMode(CMediaSettings::GetInstance().GetCurrentVideoSettings().m_ViewMode);
+  ManageRenderArea();
+
+  m_bConfigured = true;
+
+  for (int i = 0; i < m_numRenderBuffers; ++i)
+    m_buffers[i].hwPic = 0;
+
+  return true;
+}
+
+CRenderInfo CRendererMediaCodecSurface::GetRenderInfo()
+{
+  CRenderInfo info;
+  info.formats.push_back(RENDER_FMT_BYPASS);
+  info.max_buffer_size = m_numRenderBuffers;
+  info.optimal_buffer_size = m_numRenderBuffers;
+  return info;
 }
 
 bool CRendererMediaCodecSurface::RenderCapture(CRenderCapture* capture)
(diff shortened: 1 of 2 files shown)


peak3d:
@MartijnKaijser can you pls. test?

FernetMenta:
this most likeley breaks playing video in video control or background

FernetMenta:
@peak3d guiWindowFullScreen has overwritten this method. I think you should suppress clear there

peak3d:
Correct, it will do if the skin is displayed on the "background color" wich is most likely never the case

What we need is that VideoControl can cut holes in the GUI background, but afaik this is not possible right now

peak3d:
> @peak3d guiWindowFullScreen has overwritten this method. I think you should suppress clear there

thanks, will move it to this place

MartijnKaijser:
Made no difference except that i now see ghosting on GUI controls.

peak3d:
@MartijnKaijser this is a first very basic renderer approach.
Pls test and look for micro stutters while rendering, probably there has to be made some adjustments

koying:
A random 10ms sleep in the render loop is bound to fail. AMCS is not aml bypass, it actually follows VP clock.
If you want to follow that route, revert https://github.com/xbmc/xbmc/commit/8fef65d576dc87957326fd1064f28f339fdbcf4e and use CXBMCApp::WaitVSync, much like mmal does.

peak3d:
Thanx for your suggestion, you may have noticed that I don't wait 10ms
I wait max. 10ms but only in the case of two subsequent calls.

If between 2 calls are 10ms. there is no sleep at all.

CXBMCApp::WaitVSync seems to be a good decision, but we have to keep GUI in focus.
During the time gui renderes, RenderUpdate is called in vsync already.

If we make it wrong, we lead to half update frequency. The "max 10ms" respects this

koying:
> If between 2 calls are 10ms. there is no sleep at all.

Anyway. If you happen to sleep across a vsync, you'll get judder.

> During the time gui renderes, RenderUpdate is called in vsync already

Yep. That's why I always found this VP need for the codec to sleep totally fugly, and switched to follow the gui loop, like it worked (fine) before ;)

koying:
> (But nevertheless, I'll implement WaitForFrame() and log the timings, thanx for the tipp!)

Nah. Wait for frame is unrelated to Surface. Its for Mediacodec EGL only.

peak3d:
@koying 
> Anyway. If you happen to sleep across a vsync, you'll get judder.
This should never happen from concept of the implementation (max. 10ms)

If we need vsync (GUI open), RenderUpdate is called in vsync timing and  no sleep appears (ok if you have > 100hz monitor it will fail).
For Rendering Video there is a long pipeline before the frame we release is displayed (SurfaceFlinger) so my assumption is that we don't have to be exactly here for the time only video is displayed.

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
```diff
@@ -18,202 +18,141 @@
...
 }
 
-void CRendererMediaCodecSurface::AddVideoPictureHW(VideoPicture &picture, int index)
+CRenderInfo CRendererMediaCodecSurface::GetRenderInfo()
 {
-  ReleaseBuffer(index);
-  YUVBUFFER &buf = m_buffers[index];
-  buf.hwDec = picture.hwPic ? static_cast<CDVDMediaCodecInfo*>(picture.hwPic)->Retain() : nullptr;
+  CRenderInfo info;
+  info.formats.push_back(RENDER_FMT_BYPASS);
```
koying: Nope ;)

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
```diff
@@ -18,202 +18,141 @@
...
-        dstRect = CRect(centerX - diffX, centerY - diffY, centerX + diffX, centerY + diffY);
-
-        break;
-      }
-
-      default:
-        break;
-    }
-
-    mci->RenderUpdate(srcRect, dstRect);
```
koying: Is the whole orientation stuff handled somewhere else?

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
```diff
@@ -18,202 +18,141 @@
...
-  return true;
 }
 
-bool CRendererMediaCodecSurface::CreateTexture(int index)
+bool CRendererMediaCodecSurface::Supports(ERENDERFEATURE feature)
 {
-  return true; // nothing todo
+  if (feature == RENDERFEATURE_ZOOM ||
+    feature == RENDERFEATURE_CONTRAST ||
+    feature == RENDERFEATURE_BRIGHTNESS ||
```
koying: Surface does not support contrast/brightness

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
peak3d: see commit message, I still believe that this is handled in Rendermanager::configure change.
If it's not working, let me know pls.

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
peak3d: ok

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
peak3d: Whats the disadvantage?

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
koying: Err... Are you actually coding blind?
You still do not have a Shield?

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
koying: What do you mean?
I return the question: What that's the advantage of returning a wrong RENDER_FMT_BYPASS vs. the correct RENDER_FMT_MEDIACODECSURFACE ?

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
peak3d: wetek hub / minix board / no 3D TV, not rotatable.

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
peak3d: No idea, just a copy from exiting code, no idea, honest :-)

peak3d:
Tested yesterday on my minix board and Live-Tv is not smooth, need some changes to synchronize things better.

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
koying: It's not about rotating the TV ;)
This code handles the rotation tag inside a video, ie it allows to properly playback a video taken with your phone in portrait. You can test with the wetek, nothing fancy.

@da-anda has samples

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
koying: As a matter of fact, I have no clue what this is used for, either.
Probably not much, as it seems to be wrong everywhere, but better use the proper value, I guess ;)

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
peak3d: Isn't this already inside DVDCodecmediaCodecSurface? let me try

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
peak3d: agree

peak3d:
Tested on several devices without issues. Image quality (especially 50fps) is poor compared to LE (micro- unsyncs / stutters) also on shield device  but this is the same / worse without this PR.

peak3d:
Jenkins, build this please

Rechi:
jenkins build this please

koying:
Would you mind explaining "Postpone Render Start to first RenderUpdate call" and "ReleaseBuffer properly" (and maybe include comments)?

Also, did you confirm there is no increase in CPU usage?

peak3d:
CPU usage should not have changed because RenderUpdate is still called in GPU vsync frequency.
The other things have not changed from concept (except that we save some cycles not comparing 2 rects on each frame move (ReorderDrawPoints usage). Lot of optimizations are currently under work (popcornmix / me) to reduce the GPU load (wich could be a reason for the poor quality we currently have).

Sure, I'll add comment for PostPone, the ReleaseBuffer is same as before, render and decrement RefCount.

peak3d:
top with PR
```
11767  7   4% S    51 1101568K 130360K  fg u0_a73   org.xbmc.kodi
11767  7   3% S    51 1103008K 128772K  fg u0_a73   org.xbmc.kodi
11767  5   4% S    47 1097736K 127164K  fg u0_a73   org.xbmc.kodi
11767  7   4% S    47 1096616K 123952K  fg u0_a73   org.xbmc.kodi
11767  5   4% S    47 1098056K 124640K  fg u0_a73   org.xbmc.kodi
11767  4   3% S    47 1096936K 122560K  fg u0_a73   org.xbmc.kodi
11767  4   4% S    47 1098356K 124468K  fg u0_a73   org.xbmc.kodi
11767  5   4% S    47 1097256K 124464K  fg u0_a73   org.xbmc.kodi
11767  5   4% S    47 1098696K 127168K  fg u0_a73   org.xbmc.kodi
11767  7   3% S    47 1097596K 127304K  fg u0_a73   org.xbmc.kodi
```
top without PR (master)
```
11944  5   5% S    51 1101084K 146024K  fg u0_a73   org.xbmc.kodi
11944  6   5% S    51 1099984K 141428K  fg u0_a73   org.xbmc.kodi
11944  4   6% S    47 1097272K 142220K  fg u0_a73   org.xbmc.kodi
11944  4   5% S    47 1096192K 139572K  fg u0_a73   org.xbmc.kodi
11944  5   5% S    47 1097632K 140428K  fg u0_a73   org.xbmc.kodi
11944  6   5% S    47 1096532K 137964K  fg u0_a73   org.xbmc.kodi
11944  5   5% S    47 1097972K 140252K  fg u0_a73   org.xbmc.kodi
11944  7   5% S    47 1096872K 139872K  fg u0_a73   org.xbmc.kodi
11944  4   5% S    47 1095772K 140572K  fg u0_a73   org.xbmc.kodi
11944  5   5% S    47 1097212K 142744K  fg u0_a73   org.xbmc.kodi
11944  4   5% S    47 1096132K 142384K  fg u0_a73   org.xbmc.kodi
```
Both playing the same 720p 50fps h.264 stream from local memory card.

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodecSurface.cpp
```diff
@@ -212,13 +217,14 @@ void CRendererMediaCodecSurface::ReorderDrawPoints()
     dstRect.x1, dstRect.y1, dstRect.Width(), dstRect.Height(),
     adjRect.x1, adjRect.y1, adjRect.Width(), adjRect.Height());
 
+  // Assumption: SurfaceFlinger starts PTS processing on arrival of first frame.
+  // To have the best match between kodi / android clock we start on first RenderUpdate call.
+  // Beside this the first frame should not be rendered before the SurfaceView is resized.
```
koying: Why not putting the "ReleaseOutputBuffer(true)" in RenderUpdate, rather than FlipPage, as it was before?
FlipPage is a GUI layer thing, so you are mixing the 2 loops, now.
And, obviously, putting the ReleaseOutBuffer in RenderUpdate assures you the first frame is only showed on RenderUpdate, without flags.

peak3d: Explained in the comment in FlipPage.

peak3d: Just to clarify: FlipPage is RenderManger, RenderUpdate is GUI loop.
RenderManager / FlipPage is for MediaCodec Surface the "correct" place because video is independent rendered from GUI.

I'll remove the ReleaseBuffer in RenderUpdate as its not really visible what I tried to archieve.

koying: No. 
FlipPage is called from CRenderManager::FrameMove which is GUI
RenderUpdate is called from CRenderManager::PresentSingle which is Video.

The exact same logic is used in RenderAML and MMALRenderer.
I don't know why you want it different here.

User-ex is not the point here, or you'd just have left the glClear

peak3d:
no more objections?

PR #12102, opened 2017-05-16
VideoPlayer: vaapi, vdpau currently only work for X11

FernetMenta:
in current state vdpau, vaapi only work with x11
on X11 those codecs get runtime detected so a build should aleays include them

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp
@@ -411,7 +411,7 @@ void CDVDVideoCodecFFmpeg::SetFilters()
       case 180:
         m_filters_next += "vflip,hflip";
         break;
-      case 270:  
+      case 270:
         m_filters_next += "transpose=2";
         break;
       default:
@@ -1177,13 +1177,9 @@ IHardwareDecoder* CDVDVideoCodecFFmpeg::CreateVideoDecoderHW(AVPixelFormat pixfm
 #endif
 
 // Linux X11
-#if defined(HAVE_LIBVA) || defined(HAVE_LIBVDPAU)
-#if defined(HAVE_LIBVA)
+#ifdef HAVE_X11
 #include "VAAPI.h"
-#endif
-#if defined(HAVE_LIBVDPAU)
 #include "VDPAU.h"
-#endif
 
 #define VP_VIDEOCODEC_HW
 IHardwareDecoder* CDVDVideoCodecFFmpeg::CreateVideoDecoderHW(AVPixelFormat pixfmt, CProcessInfo &processInfo)


FernetMenta:
@stefansaraev this should fix you build issues that I guess are not related to X11

fritsch:
The keeps MIR broken: https://github.com/xbmc/xbmc/blob/master/cmake/scripts/linux/ArchSetup.cmake#L45

FernetMenta:
vaapi for mir is broken anyway but this is a different story

fritsch:
The only real fun thing with all this build discussion (totally unrelated though) is, that VAAPI currently does not work at all - and nobody realized that ...

As long as the auto detection is not properly merged from agile ...

wsnipex (changes requested):
this is totally wrong once again. Now you require vaapi and vdpau when building for X11.
Why do you insist on breaking stuff?

FernetMenta:
vaapi is runtime detected. there is zero advantage when building X11 without vaapi

herrnst:
According to your logic, the next thing people have to do is install AMD and Intel GPU drivers on their NVIDIA windows systems since things "are runtime detected". Nice.

FernetMenta:
vaapi / vdpau libs needed for build are just wrappers. there is absolutely no disadvantage having them installed.
as I mentioned already, there won't be more ifdefs in this file. I'll will agree to ifdefs in separate include files specific to windowing.

hauihau:
@FernetMenta are you kidding me? I  showed you that your code is wrong, independently of if any shit is runtime detected or not or what ffmpeg does. You're guarding the inclusion of VDPAU.h (analog with VAAPI) by checking HAVE_LIBVDPAU.
```
#if defined(HAVE_LIBVDPAU)
#include "VDPAU.h"
#endif
```
But you're not guarding calls to methods in VDPAU namespace some lines later. As I'm compiling with "-DENABLE_VDPAU=OFF" HAVE_LIBVDPAU can never be defined and the compiler will never know what VDPAU is (the same applies to "-DENABLE_VAAPI=OFF" and HAVE_LIBVA). Are you to blind to see that? Your attitude and querulousness sucks! Either guard the calls to the VAAPI/VDPAU methods or remove the ifdef guards around the header inclusions. There is no need for additional ifdefs in header files. The error is in xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp. Your curent logic absolutely makes no sense as well as your comments. I proved you that installing libvdpau does not solve the issue. The only solution for you wrong code to work currently is to compile kodi with "-DENABLE_VAAPI=ON -DENABLE_VDPAU=ON". But there is absolutely no need to force users to enable both VAAPI and VDPAU as kodi clearly works or worked with only one beeing enabled prior to your VP updates PR.

FernetMenta:
disabling vdpau or vaapi at build time for x11 does not make any sense. availability is runtime detected at startup.
distros have to build with vdpau and vaapi anyway because they don't know the users environment. a user won't notice how kodi was built in this regard

stefansaraev:
va/vdpau is not hard requirment for x11. there are platform that use x11 but neither va or vdpau work (or are not possible at all). however those may do sw decoding just fine. such requirment does not make any sense

herrnst:
While you're repeating yourself over and over with the same nonsense, you really should learn, recognise and most importantly accept (!) that there are other distros than debian that don't require installing "everything" but only what the user wants (and even on debian people still can create their wn binpackage without things they don't need).

hauihau:
It's really hard to discuss with you and your stubbornness :| Since cmake does **not** require both VAAPI and VDPAU beeing enabled it **does** make sense! Why should disabling one of these do not make sense? You still haven't provided coherent arguments. I can't find any hard requirement for building for X11 with both enabled. Kodi works just fine with only one beeing enabled. Distros, when needed, will of course still be able to enable both even when you ifdef the already many times mentioned calls to VAAPI/VDPAU methods. Whats wrong with two ifdefs? You should have a look at https://distrowatch.com/ because the world doesn't consist solely of Ubuntu/Debian.

TeamKodi: please consider overruling this stupid and still not technically founded decision by @FernetMenta which restricts the freedom of self builders and distros where people decide on their own what they want and what not.

koenkooi:
FWIW, I use plain X11 on a snapdragon 410 based board, so Aarch64, GL and sw decode. Vaapi (or vdpau) isn't needed or wanted in this specific case.

da-anda:
guys, I'm not a dev and only try to cool down this discussion that seems to get a bit heated. I just had a small chat with Fernet on Slack. The main goal is to move out platform specific adjustments from core files to cmake, and have cmake import the required adjusted files/flags/whatever (I'm neither a linux guy). The cmake adjustments are currently a work in progress (started here https://github.com/xbmc/xbmc/pull/10918) and will ofc be finished until release. So the issues these PRs will currently cause are only temporarily.
Since Kodi is not a one man's show and @fetzerch is currently on vacation (or alike) it would be nice if one of you guys could step in and continue the work on the required cmake changes. Thanks.

hauihau:
@da-anda I don't see how the mentioned PR should solve this specific issue.

@FernetMenta insist on both VAAPI and VDPAU must be installed for X11 (which is not forced by cmake, hmmm) without having shown anything which proves his statement. It may be correct, that VAAPI and VDPAU currently need X11 to work, but @koenkooi proved that it is not true for the opposite direction. The Feedback in the many recent PRs show, that there are users out there who want to build kodi to their needs and not how FernetMenta thinks they have to. Someone should stop him from causing advanced users problems which did not exist before the vpupates PR, it doesn't look like he has any interest in what the people here are asking for.

I kindly would like to ask again: @FernetMenta why do you force users to install and enable both VAAPI and VDPAU? Whats the benefit compared to two simple ifdefs?

hudokkow:
Folks, please calm down. This discussion already pissed off quite a few of us and there is no need to persist with it. It leads to nowhere except exasperation.
Obviously, something has to be done.  To that effect, the best course of action is being discussed on internally.

Right now is obvious that we're hitting some CMake limitations in its current implementation to handle core needs.

hauihau:
@hudokkow I can't see any cmake limitations in this case.

Take a look at https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp#L531-L533. Can't see why https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp should be handled in another way.

hudokkow:
@hauihau, you (probably) can't see the limitations because you're not seeing the hole picture. Heck, I know Kodi's CMake build system better than most and I (still) haven't fully grasped/have a solution about what's needed to make it work with less ifdeffing crap in CMake and core code. It is very much WIP at this stage.

Yes, current solution is ifdeffing code all over the place. Yes, it sucks. Yes, we're working on it.

Point is we build for a lot of platforms/targets and right now that's starting to be a PITA. Our current implementation is starting to show us that we need to fine tune CMake and core instead of ifdeffing. Please see https://github.com/xbmc/xbmc/pull/12021 for a real case scenario where this matters.

Please, please don't turn this PR into a QA about CMake/core limitations. I will close it if it turns into _that_ sort of PR.

@bkuhls, stop giving +1 to everything that moves. Thx!

hauihau:
@hudokkow thanks for clarification. My point is not making it work with less ifdefing, it's about letting the users decide to build kodi with VAAPI or VDPAU, or both or nothing like it has been possible before the big vpupdates PR. The current state forces us to enable both even if we want to use only one of them (which also isn't covered in the cmake files). I don't see any benefit or technical reason in forcing us to have both VAAPI and VDPAU enabled. I kindly want to ask wether you could readd these two ifdefs as long as cmake has no option do deal with that.

MartijnKaijser:
You also seem to forget that this is very early alpha and not anywhere near something production ready. So cool down.

FernetMenta:
obsolete with vpupdates

PR #12103, opened 2017-05-16
X11: check vaapi caps on startup

FernetMenta:
X11:. enable vaapi at runtime if present

### DIFF
--- a/xbmc/settings/AdvancedSettings.cpp
+++ b/xbmc/settings/AdvancedSettings.cpp
@@ -146,7 +146,6 @@ void CAdvancedSettings::Initialize()
   m_videoIgnorePercentAtEnd   = 8.0f;
   m_videoPlayCountMinimumPercent = 90.0f;
   m_videoVDPAUScaling = -1;
-  m_videoVAAPIforced = false;
   m_videoNonLinStretchRatio = 0.5f;
   m_videoEnableHighQualityHwScalers = false;
   m_videoAutoScaleMaxFps = 30.0f;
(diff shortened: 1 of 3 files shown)


FernetMenta:
@BrandonSchaefer if you want vaapi to work with mir, you need this too

wsnipex:
this will fail on all X11 platforms that either disable vaapi - or worse - do not use/provide it.

hauihau:
I guess he lives in his own little castle in the air where and only his opinion is the real deal. with high probability he will answer your comment with the following (still unproven) statements:

> disabling vaapi/vdpau is of no value at runtime

> further building only with one is wrong. x11 is the platform and does require vaapi AND vdpau

> disabling vdpau or vaapi at build time for x11 does not make any sense. availability is runtime detected at startup.

herrnst:
Note (since these changes look familar) that this will require more "ifdeffery" to not cause build failures. Discovered this when trying with the agile fork, I have a kodi ebuild which builds from that agile thing in my overlay, full patch is at https://github.com/herrnst/gentoo-kodi-overlay/blob/kodi-agile/media-tv/kodi/files/kodi-9999-fix-agile-build-vaapi.patch which is the least thats required. No idea if I'll PR this if this hits master, I'm so f**king sick of all this...

wsnipex:
this needs fixing...

wsnipex:
if you are not willing to fix this without introducing another regression, I'll open another PR.
It is annoying that all other PRs fail jenkins tests because of it.

FernetMenta:
fix what? I'll close this for now until I found a better solution.

wsnipex:
fix VAAPI...

FernetMenta:
vpupdates PR with broken vaapi was open for several weeks. since you did not notice you are not in a position to push for a quick hack.

stefansaraev:
it seems vaapi is a second citizen now. just like like aml/imx. I suggest removing x86_64 builds from jenkins. or adjusting it to build with va/vdpau disabled :)

hudokkow:
I've been deliberately trying very hard to ignore this whole Mexican soap-opera but I already drank a bottle of a very fine wine and moved to an even finer scotch, so yeah...

> vpupdates PR with broken vaapi was open for several weeks. since you did not notice you are not in a position to push for a quick hack.

Pardon my French but that is bull crap. You knew it was broken since early February. Please go read https://github.com/FernetMenta/kodi-agile/pull/131

Instead of aiming for a fix, ask for help, coaching, clearly stating your vision, whatever, you shot it down.

Three questions:
1) Why aren't you participating in the internal discussing about this?
2) How on earth do you want us to deal with this on CMake side if code is not ifdefed?
3) What am I missing? (already asked internally but you didn't bothered to answer)

FernetMenta:
Try to ask you question in an non offensive way and you'll get an answer.

MartijnKaijser:
How about you try that yourself as well. Might help in general.

hauihau:
Are there any plans to forcibly add the removed ifdefs back to the code? All PR's trying to fix this have been closed and the originator doesn't seem to care about the opinions of others.

FernetMenta:
@hauihau why don't you just disable all code I wrote? alternatively get lost here, this is a dev space.

hudokkow:
@hauihau, I am deeply sorry for how this comes across and how it may (adversely) reflect on Team Kodi. For better and for worse, no one here does anything forcibly.

PR #12105, opened 2017-05-16
[gui] no need to store/restore modeless dialogs while unload/load skin

xhaggi:
There is no need to store/restore modeless dialogs while unload/load skin. A modeless dialog will automatically be opened/closed depending on its visible condition.

This also leads to problems if different skins use custom modeless dialogs with the same id (skin A and B use window id 1000) and you switch between this skins (see http://forum.kodi.tv/showthread.php?tid=314185)

@BigNoid mind taking a look.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -1600,8 +1600,9 @@ bool CApplication::LoadSkin(const std::string& skinID)
     skin = std::static_pointer_cast<ADDON::CSkinInfo>(addon);
   }
 
-  bool bPreviousPlayingState=false;
-  bool bPreviousRenderingState=false;
+  // store player and rendering state
+  bool bPreviousPlayingState = false;
+  bool bPreviousRenderingState = false;
   if (m_pPlayer->IsPlayingVideo())
   {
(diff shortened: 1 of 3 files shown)


BigNoid:
I see no regressions when testing, e.g. custom windows/dialogs still load when its visible after switching skins.

Not sure if it should be part of this PR or not, but I do think we need an extra function like ```CGUIWindowManager::RemoveFromWindowHistory()``` that can remove custom window id's from history when ```CGUIWindowManager::DeInitialize()``` is called. Right now if you switch to a skin from Estuary's addoninfo (navigated to through the main menu addon menu item) and you go back, you can go back to the new skin's custom window with ```id="1100"``` which can lead to unexpected results like a black screen if that window/dialog should not be visible.

xhaggi:
Added another commit to address your comment. It makes totally sense to remove the custom window from our window history while destroy them.

BigNoid:
Works great, thx for adding that 👍

xhaggi:
Thanks for testing it.

jenkins build this please

MartijnKaijser:
jenkins build this please

xhaggi:
jenkins build this please

PR #12110, opened 2017-05-16
smbclient: cleanup smbclient configuration

chewitt:
This PR does some overdue cleanup on the smbclient configuration used by Kodi.

## Description

* Removed by fritch: ~~`socket options` tuning that is not recommended: https://lists.samba.org/archive/samba/2013-February/171836.html~~

* Removes `preferred master`, `local master`, `domain master` smbd configuration that is irrelevant to smbclient.

* Removes `client lanman auth` so smbclient can use more secure hash algorithms. This breaks compatibility with Windows 95/98 and pre Samba 2.2 shares (c.2000) that do not support NTLM or NTLMv2.

* Removes `lanman auth` smbd configuration that is irrelevant to smbclient.

* Sets `client max protocol` to SMB3. Without this smbclient (and thus Kodi) defaults to NT1 (SMB1) and will not attempt connecting at higher protocol versions. Setting to SMB3 allows smbclient to auto-negotiate a higher protocol with most devices. Some older Windows and Samba versions can require `client max protocol` to be reduced to SMB2 or SMB1 to solve connectivity issues (server side bugs or misconfiguration) so we expose this as a configuration option in Kodi settings. Using SMB3 with smbclient versions below 4.1 that do not support SMB2+ are not affected; older smbclient versions simply negotiate up to what they understand, i.e. NT1. This is part based on https://github.com/koying/SPMC/commit/833aacd4698a7c5af09a94000784c492210959ef that forces NT1 connectivity.

* Fixes a bug where `smb.conf` was not written if the file was not already present. The smb.conf file is also created under the Kodi home folder to prevent clashes with other smb.conf files that may exist on the users' system. The smb.con file is rewritten if the user changes the client max protocol setting (and the user is prompted to restart Kodi to effect the change). This borrows from https://github.com/koying/SPMC/commit/2771cb222d4d09082dac06ce970dba0f6293cd56 that made rewrite on start user configurable and other ideas from Slack.

## Motivation and Context
Wannacry ransomware attacks have woken people up to the security issues with SMB1 and Kodi refusing to use anything newer forces people to continue using insecure shares. LE plans to use a Krypton version of this with samba 4.6 in a future release to bring SMB2/3 support. NB: This PR is intended as a nip/tuck improvement to the existing regime until someone comes up with a more long-term replacement for smbclient.

## How Has This Been Tested?
Lots of experimentation and detective work from @MilhouseVH. An earlier variant of the max client protocol change and samba 4 have been in his nightly builds for several months and this specific PR will be included shortly. It has not been tested with other Kodi build targets that use smbclient (macOS, Android, etc.) but testing with older smbclient 3.6 on Linux hasn't thrown up any obvious issues.

## Types of change

## Checklist:

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -19740,7 +19740,49 @@ msgctxt "#36620"
 msgid "Enable high quality downscaling of pictures (uses more memory and has moderate performance impact)."
 msgstr ""
 
-#empty strings from id 36621 to 36899
+#. Label of a setting, allow the maximum smbclient protocol to be configured
+#: system/settings/settings.xml
+msgctxt "#36621"
+msgid "Maximum protocol version"
+msgstr ""
+
+#. Description of setting with label #36621 "Maximum protocol version"
+#: system/settings/settings.xml
+msgctxt "#36622"
+msgid "Set the maximum SMB protocol version to negotiate when making connections. Forcing SMBv2 or SMBv1 compatibility may be required with older NAS and Windows shares."
+msgstr ""
+
+#. Values for setting with label #36621 "Maximum protocol version" - none means "no protocol version is forced"
+#: system/settings/settings.xml
+msgctxt "#36623"
+msgid "None"
+msgstr ""
+
+#. Values for setting with label #36621 "Maximum protocol version"
+#: system/settings/settings.xml
+msgctxt "#36624"
+msgid "SMBv1"
+msgstr ""
+
+#. Values for setting with label #36621 "Maximum protocol version"
+#: system/settings/settings.xml
+msgctxt "#36625"
+msgid "SMBv2"
+msgstr ""
+
+#. Values for setting with label #36621 "Maximum protocol version"
+#: system/settings/settings.xml
+msgctxt "#36626"
+msgid "SMBv3"
+msgstr ""
+
+#. Label of a group, that allows configuration of the system SMB client
+#: system/settings/settings.xml
+msgctxt "#36627"
+msgid "Client"
+msgstr ""
+
+#empty strings from id 36628 to 36899
 
 #: xbmc/media/MediaType.cpp
 msgctxt "#36900"
--- a/xbmc/filesystem/SMBFile.cpp
+++ b/xbmc/filesystem/SMBFile.cpp
@@ -28,6 +28,7 @@
 #include "ServiceBroker.h"
 #include "SMBDirectory.h"
 #include <libsmbclient.h>
+#include "filesystem/SpecialProtocol.h"
 #include "settings/AdvancedSettings.h"
 #include "settings/Settings.h"
 #include "threads/SingleLock.h"
@@ -92,33 +93,39 @@ void CSMB::Deinit()
 void CSMB::Init()
 {
   CSingleLock lock(*this);
+
   if (!m_context)
   {
-    // Create ~/.smb/smb.conf. This file is used by libsmbclient.
+    // force libsmbclient to use our own smb.conf by overriding HOME
+    std::string truehome(getenv("HOME"));
+    setenv("HOME", CSpecialProtocol::TranslatePath("special://home").c_str(), 1);
+
+    // Create ~/.kodi/.smb/smb.conf. This file is used by libsmbclient.
     // http://us1.samba.org/samba/docs/man/manpages-3/libsmbclient.7.html
     // http://us1.samba.org/samba/docs/man/manpages-3/smb.conf.5.html
-    char smb_conf[MAX_PATH];
+    std::string smb_conf;
     std::string home(getenv("HOME"));
     URIUtils::RemoveSlashAtEnd(home);
-    snprintf(smb_conf, sizeof(smb_conf), "%s/.smb", home.c_str());
-    if (mkdir(smb_conf, 0755) == 0)
+    smb_conf = home + "/.smb";
+    int result = mkdir(smb_conf.c_str(), 0755);
+    if (result == 0 || (errno == EEXIST && IsFirstInit))
     {
-      snprintf(smb_conf, sizeof(smb_conf), "%s/.smb/smb.conf", getenv("HOME"));
-      FILE* f = fopen(smb_conf, "w");
+      smb_conf += "/smb.conf";
+      FILE* f = fopen(smb_conf.c_str(), "w");
       if (f != NULL)
       {
         fprintf(f, "[global]\n");
 
-        // make sure we're not acting like a server
-        fprintf(f, "\tpreferred master = no\n");
-        fprintf(f, "\tlocal master = no\n");
-        fprintf(f, "\tdomain master = no\n");
+        fprintf(f, "\tlock directory = %s/.smb/\n", home.c_str());
 
-        // use the weaker LANMAN password hash in order to be compatible with older servers
-        fprintf(f, "\tclient lanman auth = yes\n");
-        fprintf(f, "\tlanman auth = yes\n");
-
-        fprintf(f, "\tlock directory = %s/.smb/\n", getenv("HOME"));
+        // set maximum smbclient protocol version
+        if (CServiceBroker::GetSettings().GetInt(CSettings::SETTING_SMB_MAXPROTOCOL) > 0)
+        {
+          if (CServiceBroker::GetSettings().GetInt(CSettings::SETTING_SMB_MAXPROTOCOL) == 1)
+            fprintf(f, "\tclient max protocol = NT1\n");
+          else
+            fprintf(f, "\tclient max protocol = SMB%d\n", CServiceBroker::GetSettings().GetInt(CSettings::SETTING_SMB_MAXPROTOCOL));
+        }
 
         // set wins server if there's one. name resolve order defaults to 'lmhosts host wins bcast'.
         // if no WINS server has been specified the wins method will be ignored.
@@ -134,6 +141,9 @@ void CSMB::Init()
         // samba tries to use charset 850 but falls back to ASCII in case it is not available
         if (g_advancedSettings.m_sambadoscodepage.length() > 0)
           fprintf(f, "\tdos charset = %s\n", g_advancedSettings.m_sambadoscodepage.c_str());
+
+        // include users configuration if available
+        fprintf(f, "\tinclude = %s/.smb/user.conf\n", home.c_str());
 
         fclose(f);
       }
@@ -151,6 +161,10 @@ void CSMB::Init()
 
     // setup our context
     m_context = smbc_new_context();
+
+    // restore HOME
+    setenv("HOME", truehome.c_str(), 1);
+
 #ifdef DEPRECATED_SMBC_INTERFACE
     smbc_setDebug(m_context, g_advancedSettings.CanLogComponent(LOGSAMBA) ? 10 : 0);
     smbc_setFunctionAuthData(m_context, xb_smbc_auth);
@@ -647,3 +661,4 @@ int CSMBFile::IoControl(EIoControl request, void* param)
 
   return -1;
 }
+
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -1907,15 +1907,30 @@
     <category id="smb" label="1200" help="36346">
       <requirement>HAS_FILESYSTEM_SMB</requirement>
       <group id="1" label="16000">
+        <setting id="smb.workgroup" type="string" label="1202" help="36348">
+          <level>2</level>
+          <default>WORKGROUP</default>
+          <control type="edit" format="string" />
+        </setting>
+      </group>
+      <group id="2" label="36627">
         <setting id="smb.winsserver" type="string" label="1207" help="36347">
           <level>2</level>
           <default>0.0.0.0</default>
           <control type="edit" format="ip" />
         </setting>
-        <setting id="smb.workgroup" type="string" label="1202" help="36348">
+        <setting id="smb.maxprotocol" type="integer" label="36621" help="36622">
           <level>2</level>
-          <default>WORKGROUP</default>
-          <control type="edit" format="string" />
+          <default>3</default>
+          <constraints>
+            <options>
+              <option label="36623">0</option>
+              <option label="36624">1</option>
+              <option label="36625">2</option>
+              <option label="36626">3</option>
+            </options>
+          </constraints>
+          <control type="list" format="integer" />
         </setting>
       </group>
     </category>
--- a/xbmc/network/NetworkServices.cpp
+++ b/xbmc/network/NetworkServices.cpp
@@ -118,6 +118,7 @@ CNetworkServices::CNetworkServices()
   , m_httpWebinterfaceAddonsHandler(*new CHTTPWebinterfaceAddonsHandler)
 #endif // HAS_WEB_INTERFACE
 #endif // HAS_WEB_SERVER
+
 {
 #ifdef HAS_WEB_SERVER
   m_webserver.RegisterRequestHandler(&m_httpImageHandler);
@@ -439,7 +440,8 @@ void CNetworkServices::OnSettingChanged(std::shared_ptr<const CSetting> setting)
   else
 #endif // HAS_WEB_SERVER
   if (settingId == CSettings::SETTING_SMB_WINSSERVER ||
-      settingId == CSettings::SETTING_SMB_WORKGROUP)
+      settingId == CSettings::SETTING_SMB_WORKGROUP ||
+      settingId == CSettings::SETTING_SMB_MAXPROTOCOL)
   {
     // okey we really don't need to restart, only deinit samba, but that could be damn hard if something is playing
     //! @todo - General way of handling setting changes that require restart
--- a/xbmc/settings/Settings.cpp
+++ b/xbmc/settings/Settings.cpp
@@ -333,6 +333,7 @@ const std::string CSettings::SETTING_SERVICES_AIRPLAYPASSWORD = "services.airpla
 const std::string CSettings::SETTING_SERVICES_AIRPLAYVIDEOSUPPORT = "services.airplayvideosupport";
 const std::string CSettings::SETTING_SMB_WINSSERVER = "smb.winsserver";
 const std::string CSettings::SETTING_SMB_WORKGROUP = "smb.workgroup";
+const std::string CSettings::SETTING_SMB_MAXPROTOCOL = "smb.maxprotocol";
 const std::string CSettings::SETTING_VIDEOSCREEN_MONITOR = "videoscreen.monitor";
 const std::string CSettings::SETTING_VIDEOSCREEN_SCREEN = "videoscreen.screen";
 const std::string CSettings::SETTING_VIDEOSCREEN_RESOLUTION = "videoscreen.resolution";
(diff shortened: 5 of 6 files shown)


MilhouseVH:
https://lists.samba.org/archive/samba-technical/2016-November/117005.html

Samba defaults to NT1 client connections unless `client max protocol` is configured, which explains why Kodi is fixated on establishing NT1 connections regardless of what is (or is not) supported by the server.

I've just checked the latest Samba master[1] and the default client connection is still(!) NT1... utter madness.

1. https://github.com/samba-team/samba/blob/7556c20d4bf90bfcc288ba1c82008105eaf8f261/source3/param/loadparm.c#L4529-L4532 <--- current HEAD on master

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -4065,7 +4065,9 @@ msgctxt "#1208"
 msgid "Mount SMB shares"
 msgstr ""
 
-#empty string with id 1209
+msgctxt "#1209"
```
ksooo: Could you please provide contextual information for the new strings (here and elsewhere). This makes live of translators a lot easier. Thanks.
#. ...
#: ...

ksooo (changes requested):

addons/resource.language.en_gb/resources/strings.po
MartijnKaijser: Also do not use strings below 10000

koying:
You probably do not want to force rewrite smb.conf on a plain linux distro. Or do I miss something?

MilhouseVH:
> Or do I miss something?

The problem is that currently the config isn't updated at all, even if a setting is changed. This currently requires the user to delete the .smb folder before new settings can be applied.

What is the problem with updating `$HOME/.smb/smb.conf` every time Kodi is started - are you concerned about other non-Kodi applications that might be sharing smb.conf?

koying:
Yes. smb.conf is the standard user libsmbclient config file, so other apps or even the user might have tweaked it on a Linux distro, and we probably should not unconditionally overwrite it. 

But I'll completely trust @wsnipex 's opinion on this.

koying:
Note that an automagic way to overwrite config could be to calculate the hash of the existing file and compare with a precalculated one of our "old" default. 
If equal, we can assume it's ours and overwrite.

wsnipex:
We should never overwrite anything outside of our userdata dir.

why don't we use our own smb.conf instead of the standard one? That way there is no problem with overwriting.

And the client max protocol won't work anywhere but linux until we somehow manage to upgrade samba to a more recent version. So this setting should only be visible if supported.

MilhouseVH:
@wsnipex then this would need run-time detection of the smb library version. I can build Kodi on Ubuntu 16.04 which has support for SMB2 and SMB3 client connections (smb library 4.3.11).

@koying an option to overwrite has been added to this PR overnight - once enabled it allows the WINS and max protocol settings to be changed (and written at next restart or when next new context is established). The overwrite setting will be automatically disabled once the new conf is written. 

Not sure if a magic hash would work as once you change any of the defaults your precalculated "default" hash would no longer match and you'd never update the file again, not even to set it back to current defaults (plus this would make life harder for anyone using appliance.xml etc.). :)

chewitt:
@wsnipex If you run smbclient from the console you can append `--config /path/to/smb.conf` but there is no equivalent for API calls so we have to work with a potentially shared config.

wsnipex:
@chewitt https://github.com/samba-team/samba/search?utf8=%E2%9C%93&q=SMB_CONF_PATH&type=

xbmc/filesystem/SMBFile.cpp
```diff
@@ -101,26 +101,23 @@ void CSMB::Init()
...
-        fprintf(f, "\tpreferred master = no\n");
-        fprintf(f, "\tlocal master = no\n");
-        fprintf(f, "\tdomain master = no\n");
-
-        // use the weaker LANMAN password hash in order to be compatible with older servers
-        fprintf(f, "\tclient lanman auth = yes\n");
-        fprintf(f, "\tlanman auth = yes\n");
-
-        fprintf(f, "\tsocket options = TCP_NODELAY IPTOS_LOWDELAY SO_RCVBUF=65536 SO_SNDBUF=65536\n");      
         fprintf(f, "\tlock directory = %s/.smb/\n", getenv("HOME"));
```
Rechi: could you also replace `getenv("HOME")` with `home.c_str()` like you have done above?

Rechi:
@chewitt this needs to be rebased

chewitt:
rebased .. do we want to add the home/env hack @koying suggested in Slack?

Rechi:
With this change `smb.conf` now gets overwrite although it exists already.
To avoid removing / changing some settings a user has set explicitly for the whole os, we should use our own `smb.conf`.

koying:
The only way I found to have our own smb.conf is:
```
    // HACK!! Force libsmbclient to use our own smb.conf by overriding HOME
    char* truehome = getenv("HOME");
    setenv("HOME", CSpecialProtocol::TranslatePath("special://home").c_str(), 1);

    // setup our context
    m_context = smbc_new_context();

    // HACK!! Restore HOME
    setenv("HOME", truehome, 1);
```

SMB_CONF_PATH didn't work for me

koying:
Note that, if we go the "own smb.conf" route, CSettings::SETTING_SMB_OVERWRITECONF doesn't make sense anymore

MilhouseVH:
> Note that, if we go the "own smb.conf" route, CSettings::SETTING_SMB_OVERWRITECONF doesn't make sense anymore

Yes, and no... being able to selectively overwrite would at least allow users to `tune` the smb.conf if they wished. And constant rewriting seems a little unnecessary.

koying:
1) Fair enough regarding "constant rewriting", but then existing smb.conf might never be updated if we change our own again in the future (this PR will create a new one anyway)
2) Fair enough regarding tuning, but then we might want "overwrite" defaulted to "true" because of 1)

MilhouseVH:
It is possible to include extra config files, eg. `include = ${HOME}/.smb/extra.conf`, and then automatically overwriting `smb.conf` once per session (at startup) would be totally fine, as any "tweaking" could occur in `.smb/extra.conf` (it doesn't have to exist) which would be under the users control.

MilhouseVH:
http://sprunge.us/SLBF

@koying the above patch ensures the `smb.conf` settings are written every time before the first (and only the first) connection is initialised. It also adds the `include` setting which allows the user to add their own settings without fear of being overwritten. Tested on Linux (LibreELEC) and it works fine for me.

If you're OK with this as a complete solution then I'll work with @chewitt to remove the overwrite toggle setting and hopefully we can put this to bed. :)

koying:
@MilhouseVH Not 100% sure "IsFirstInit" does what you think it does.

As CSMB is a global (https://github.com/xbmc/xbmc/blob/8d4a5bba55638dfd0bdc5e7de34f3e5293f99933/xbmc/filesystem/SMBFile.h#L67), it should be inited only once, and m_context is only reset on Deinit, which comes when we exit.

Bottom-line: I'd remove the "IsFirstInit" check.
Besides that, I trust you on the "user.conf" include, but it's not me to approve, really.

Did you confirm the hack is working? I did tests myself, but a second confirmation is welcome.

MilhouseVH:
> @MilhouseVH Not 100% sure "IsFirstInit" does what you think it does.

I'm sure you're right, although it would be nice if we could limit the rewriting to only when required (directory doesn't exist) and kodi startup.  Is there a better solution for this, maybe an extra/different static boolean?

I do however see Init() being called each time an SMB connection is established - for instance establish a connection and Init() will be executed, wait for the connection to timeout (3 minutes) and then connect again and Init() will be called again (potentially writing out `smb.conf` etc.).

> Did you confirm the hack is working? I did tests myself, but a second confirmation is welcome.

Yes the hack https://github.com/xbmc/xbmc/pull/12110/commits/2f75f1d403a4cebb7c4255822f0d927823123d9c is working fine for me in Linux.

koying:
> I do however see Init() being called each time an SMB connection is established

It's sure called multiple times, but the call is noop once we have a m_context.
The only reason to not have one is if the "smbc_init_context" fails, but then "IsFirstInit" stays true as well ;)

MilhouseVH:
Ouch.

I'm just trying a version I built with the "home" hack and I'm trying to install an addon from a zip file that I have in `$HOME/downloads`, and when I drill into "Home folder" in the "Install from zip" dialog:

s1

I see this, which looks like the contents of the `.kodi` folder and not `$HOME`:

s2

MilhouseVH:
> It's sure called multiple times, but the call is noop once we have a m_context.
> The only reason to not have one is if the "smbc_init_context" fails, but then "IsFirstInit" stays true as well ;)

Then something is wrong, or you're not taking multiple shares into account or that connections are removed once they become idle (and maybe the `m_context` along with it?)

I added the following patch to log some extra details in `Init()`: http://sprunge.us/AFUY

I then connected to all 3 of my SMB shares (one each on Windows 7 Ultimate, Ubuntu and FreeNAS) and visited them multiple times - sure enough, just a single `m_context` is created/used as you predicted:
```
18:37:48.283 T:1546122144    INFO: NJM1: Init() called
18:37:48.283 T:1546122144    INFO: NJM2: Init() called with no m_context
18:37:48.297 T:1546122144    INFO: NJM1: Init() called
18:37:48.427 T:1546122144    INFO: NJM1: Init() called
18:37:49.306 T:1855976352    INFO: NJM1: Init() called
18:37:49.315 T:1855976352    INFO: NJM1: Init() called
18:37:49.322 T:1855976352    INFO: NJM1: Init() called
18:38:04.804 T:1783456672    INFO: NJM1: Init() called
18:38:04.807 T:1546122144    INFO: NJM1: Init() called
18:38:09.970 T:1943030224    INFO: NJM1: Init() called
18:38:10.371 T:1472197536    INFO: NJM1: Init() called
18:38:13.119 T:1791845280    INFO: NJM1: Init() called
18:38:15.294 T:1943030224    INFO: NJM1: Init() called
18:38:15.349 T:1472197536    INFO: NJM1: Init() called
18:38:17.153 T:1546122144    INFO: NJM1: Init() called
18:38:18.467 T:1943030224    INFO: NJM1: Init() called
18:38:18.680 T:1472197536    INFO: NJM1: Init() called
18:38:24.543 T:1943030224    INFO: NJM1: Init() called
```
However I then waited a few minutes and then connected again to the Windows share:
```
18:41:02.914 T:1943030224    INFO: NJM1: Init() called
18:41:02.914 T:1943030224    INFO: NJM2: Init() called with no m_context
```
So `m_context` is being nulled at some point and when the next SMB connection is established the whole process of initialising `m_context` is repeated (including, potentially, writing `smb.conf`).

koying:
Ah, right you are. "CheckIfIdle" does a Deinit.
So yes, your solution is proper, sorry for the noise.

xbmc/filesystem/SMBFile.cpp
```diff
@@ -151,6 +158,10 @@ void CSMB::Init()
 
     // setup our context
     m_context = smbc_new_context();
+
+    // restore HOME
+    setenv("HOME", truehome, 1);
```
MilhouseVH: Apparently this call *is* succeeding (return value is 0) however `errno` is being set to 95 which (according to `strerror(errno)`) means ` Operation not supported`, so it doesn't look like the $HOME variable is being reset to the original value.

Any ideas?

koying: Maybe save "truehome" in a std::string.
From what I read, the c-string returned from getenv is not immutable (quite obviously).

MilhouseVH: Yes, I've tried that as well but same result.

I added this patch for diagnostics: http://sprunge.us/HbHG
In LibreELEC I get:
```
21:14:10.706 T:1784505248    INFO: NJM3: $HOME is now [/storage/.kodi/]
21:14:10.708 T:1784505248    INFO: NJM4: $HOME needs to be [/storage]
21:14:10.709 T:1784505248    INFO: NJM5: $HOME has been set to [/storage], retcode = 0, errno = 95, error = Operation not supported
```
while in Ubuntu 16.04 it does appear to work:
```
21:57:31.969 T:140159918352128    INFO: NJM3: $HOME is now [/home/neil/.kodi/]
21:57:31.969 T:140159918352128    INFO: NJM4: $HOME needs to be [/home/neil]
21:57:31.969 T:140159918352128    INFO: NJM5: $HOME has been set to [/home/neil], retcode = 0, errno = 2, error = No such file or directory
```
so for now at least it's just an LE issue. Hmmm.

MilhouseVH:
I've determined the issue with LE (although it probably wasn't LE specific in the end). Not sure why `errno` is returning error 95 but that seems to be a red herring.

The issue is this:
```
void CSMB::Init()
{
  CSingleLock lock(*this);

  // force libsmbclient to use our own smb.conf by overriding HOME
  char* truehome = getenv("HOME");
  setenv("HOME", CSpecialProtocol::TranslatePath("special://home").c_str(), 1);

  if (!m_context)
  {
    ...
    smbc_init(xb_smbc_auth, 0);

    // setup our context
    m_context = smbc_new_context();

    // restore HOME
    setenv("HOME", truehome, 1);
    ...
  }
  m_IdleTimeout = 180;
}
```

Each time a share is initialised we would set `$HOME` to the translated path, but only reset it back to the original value if we have to create `m_context` (which, as we've established, would only happen for the first connection or after an idle connection is `Deinit()`). So if you only had one SMB share, then no problem, but with multiple SMB shares (I have 3) your `$HOME` variable would be left set incorrectly. D'oh!

This [patch](http://sprunge.us/gFEf) will fix it, and switches to `std::string`. Tested in LE9 with multiple SMB shares, working fine now. I've asked @chewitt to update the PR with:

http://sprunge.us/gFEf (`$HOME` bug fix)
http://sprunge.us/SLBF (only write file on startup, adds `include=` support)
http://sprunge.us/IJVB (removes the SMB_OVERWRITE GUI setting)

It would be good to get testing feedback from non-Linux platforms to make sure the `$HOME` hack works for everyone.

MilhouseVH:
Please rebase.

chewitt:
@MilhouseVH rebased for changes in xbmc/settings/Settings.cpp - no real-world issues from my side on testing - only the standard BS that you see with browsing when browse master changes in the network. Good to merge?

MartijnKaijser:
What's gonna happen with this and backport?

MilhouseVH:
It's been in my LE test builds for several weeks and is working as intended. It will produce an "invalid parameter" warning to stdout for those systems with libsmbclient prior to 4.1. Users with broken/ancient Samba servers that claim to support SMB3 (but don't) will need to configure SMB2 or SMB1 to restore Samba functionality.

Disabling NT1/SMB1 on the server (as recommended by everyone, which is the reason why this PR is becoming necessary) seems to break network browsing but that's an unrelated issue to this PR (and is a core Samba issue rather than Kodi).

koying:
Is it just me or is `cache directory` somewhat mandatory, now.
Without it, on my droid 4.5.1 build, I get `mkdir failed on directory /home/cbro/spmc17-build/arm-linux-androideabi-21/var/cache/samba: No such file or directory`, `/home/cbro/spmc17-build/arm-linux-androideabi-21` being the default dir from build.

On 3.6, and in the doc, it's clearly stated that `lock directory` should be used, but it seems it's not...

koying:
I was looking at samba configuration, and stumble across posts suggesting auto-negotiation (no "client max protocol" at all) is the preferred way.
So we might need a default "Default" option, which doesn't set the option at all...

chewitt:
@koying PR is updated with a small tweak from @MilhouseVH so we default to no config and smbclient autonegotiates.

MilhouseVH:
@chewitt the default is still [SMBv3](https://github.com/xbmc/xbmc/pull/12110/files#diff-92b91a5c902936f6e9a03f26b054d7d6R1924) which personally I'd keep, but to match the help message (which now says the default is auto-negotiate) this value needs to be changed to 0. The purpose of this PR however is to support better than NT1 and with more servers dropping NT1 (which will continue to be negotiated when the setting is `auto`, and thus not working) I think we should continue to use a default setting of SMBv3 - negotiating a lower protocol should become the exception and only when the user needs it.

koying:
Is "negotiation" not implying client and server agree on a mutually acceptable level?
I'd assume that if the server does not support a level, that one wouldn't be selected, or I don't see the point ;)
Anyway, I'll try to do tests with various combination this WE and report

MilhouseVH:
Negotiation will only happen if `client max protocol` is greater than `NT1` (aka `SMBv1`), ie. if `SMBv2` or `SMBv3` is selected.

Without `client max protocol`, libsmbclient will only ever use `NT1` regardless of what the server may support (which may not include `NT1` meaning the client is unable to establish a connection with the server).

Forcing `NT1` should produce the same behaviour as not setting `client max protocol`, but apparently doesn't always, hence the request for a `default/none/auto` option (which I added in a patch to @chewitt and which didn't change the SMBv3 default - it just added a "do nothing" option).

In order for Samba to automatically negotiate a better than `NT1` connection the `client max protocol` should be set to `SMBv3` as the default, and then knocked back (or disabled) by the minority of users with duff servers.

koying:
Ok, thanks for the clarification.
Thinking out loud, but then shouldn't "Default" (=SMB3) + "SMB1" options be sufficient?

MilhouseVH:
We thought so, but then you asked for the "do nothing" option!

I guess it's possible that samba.org may flip the default connecting behaviour from "forced `NT1` to "auto-negotiate highest version supported by this client" but until that happens then we have to assume `NT1` will always be negotiated in the absence of a `client max protocol`.

Perhaps the posts you found that suggested not setting anything were either misinformed (you'd expect Samba to auto-negotiate, but that's not what it does by default) or were written from the perspective of ensuring maximum compatibility in an age when every server supported `NT1` (not the case today).

MilhouseVH:
And apparently it will be auto-negotiate by default with Samba 4.7

https://git.samba.org/?p=samba.git;a=commitdiff;h=1199907cbe2f003a7df6f56e6cf3878d0732344d

koying:
> We thought so, but then you asked for the "do nothing" option!

Sorry about that ;)
Doing an actual google returns conflicting information on whether not specifying the option default to "NT1" or do auto-negotiation, probably dependent on the samba version.

it was just a suggestion btw, although reading me back might seem more imperative than I intended.

xbmc/filesystem/SMBFile.cpp
```diff
@@ -92,33 +93,39 @@ void CSMB::Deinit()
...
 {
   CSingleLock lock(*this);
+
   if (!m_context)
   {
+    // force libsmbclient to use our own smb.conf by overriding HOME
+    std::string truehome(getenv("HOME"));
+    setenv("HOME", CSpecialProtocol::TranslatePath("special://home").c_str(), 1);
+
     // Create ~/.smb/smb.conf. This file is used by libsmbclient.
```
wsnipex: path is now ~/.kodi/.smb/smb.conf, correct?

koying:
I confirm "auto-negotiation" is kind of BS ;)
From droid 4.5.1 to debian 4.2.14, if I disable smb1 on the NAS (with `min protocol = SMB2`) and leave out `max protocol` on the client, I just can't connect.

chewitt:
The default option and wording is now set to "SMB3" to provide compatibility. Samba <4.1 will still connect using NT1, while Samba 4.1+ will connect using SMB3_11 and should step down to SMB2 if required. The none option (remove config and use Samba defaults) is retained as it does no harm and may be useful with support. The code comment on ~/.kodi/.smb/smb.conf path has been amended.

koying (approved):
Keeping "None" as an option for the future makes sense.
+1, nice work :)

wsnipex (approved):
looks good now, thanks :+1:

PR #12111, opened 2017-05-16
smbclient: cleanup smbclient configuration

chewitt:
backport of #12110 in case it's accepted

### DIFF
--- a/xbmc/filesystem/SMBFile.cpp
+++ b/xbmc/filesystem/SMBFile.cpp
@@ -27,6 +27,7 @@
 #include "PasswordManager.h"
 #include "SMBDirectory.h"
 #include <libsmbclient.h>
+#include "filesystem/SpecialProtocol.h"
 #include "settings/AdvancedSettings.h"
 #include "settings/Settings.h"
 #include "threads/SingleLock.h"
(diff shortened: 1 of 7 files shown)


xbmc/filesystem/SMBFile.cpp
```diff
@@ -100,26 +100,23 @@ void CSMB::Init()
     std::string home(getenv("HOME"));
     URIUtils::RemoveSlashAtEnd(home);
     snprintf(smb_conf, sizeof(smb_conf), "%s/.smb", home.c_str());
```
Paxxi: not specific to this PR really but it would be nice to get rid of the char buffer on the stack and use a string, maybe with `StringUtils::Format` or just concats. snprintf on the stack should probably be avoided.

Not a blocking issue, just a note.

Rechi (dismissed):
`smb.maxprotocol` setting should be only available if samba version is higher or equal 4.1
all builds using depends should hide it as Krypton builds samba 3.6.12

koying:
Looks good.

As discussed on slack, my only concerns are about the below changes, which are totally ok for a major release, but for which I'm more reserved for a point release.

@chewitt / @MilhouseVH confirmed it's no issue, though, and I'll trust them.

```
-        // use the weaker LANMAN password hash in order to be compatible with older servers
-        fprintf(f, "\tclient lanman auth = yes\n");
-        fprintf(f, "\tlanman auth = yes\n");
-
-        fprintf(f, "\tsocket options = TCP_NODELAY IPTOS_LOWDELAY SO_RCVBUF=65536 SO_SNDBUF=65536\n");      
```

Rechi (commented):
smb.maxprotocol changes look now correct
can't say anything about the other changed/removed options

MartijnKaijser:
Are we good for merge?

chewitt:
Yes, good to merge.

PR #12119, opened 2017-05-17
platinum: Fix upnp:lastPlaybackPosition parsing

jmaggard10:
According to the UPnP spec, upnp:lastPlaybackPosition should use the
same format as duration (HH:MM:SS). So we should at least accept the
proper format if it exists, and fall back to the raw integer value if
it's not in the proper format.

Signed-off-by: Justin Maggard <jmaggard@netgear.com>

## How Has This Been Tested?
I verified that both the existing format (number of seconds integer value) and the proper format (HH:MM:SS) both work.

## Types of change

## Checklist:

### DIFF
--- a/lib/libUPnP/Platinum/Source/Devices/MediaServer/PltMediaItem.cpp
+++ b/lib/libUPnP/Platinum/Source/Devices/MediaServer/PltMediaItem.cpp
@@ -762,7 +762,10 @@ PLT_MediaObject::FromDidl(NPT_XmlElementNode* entry)
     m_MiscInfo.original_track_number = value;
 
     PLT_XmlHelper::GetChildText(entry, "lastPlaybackPosition", str, didl_namespace_upnp);
-    if (NPT_FAILED(str.ToInteger(value))) value = 0;
+    if (NPT_FAILED(PLT_Didl::ParseTimeStamp(str, value))) {
+        // fall back to raw integer parsing
+        if (NPT_FAILED(str.ToInteger(value))) value = 0;
+    }
     m_MiscInfo.last_position = value;
 
     PLT_XmlHelper::GetChildText(entry, "lastPlaybackTime", m_MiscInfo.last_time, didl_namespace_upnp, 256);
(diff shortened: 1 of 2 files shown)


lib/libUPnP/Platinum/Source/Devices/MediaServer/PltMediaItem.cpp
```diff
@@ -762,7 +762,10 @@ PLT_MediaObject::FromDidl(NPT_XmlElementNode* entry)
     m_MiscInfo.original_track_number = value;
 
     PLT_XmlHelper::GetChildText(entry, "lastPlaybackPosition", str, didl_namespace_upnp);
-    if (NPT_FAILED(str.ToInteger(value))) value = 0;
+    if (NPT_FAILED(PLT_Didl::ParseTimeStamp(str, value))) {
+        // fall back to raw integer parsing
+        if (NPT_FAILED(str.ToInteger(value))) value = 0;
```
a1rwulf: I know also the old code had the assignment in the same line as the if statement, but would you mind to move it into the next line, so that we can keep style consistence?

a1rwulf:
Thanks for taking the time to look into this.
@Montellese I'm not much into Upnp, can you have a look?

lib/libUPnP/Platinum/Source/Devices/MediaServer/PltMediaItem.cpp
Montellese: @a1rwulf this is not our code but 3rd party code (Platinum SDK) so we should follow their coding style so please leave it as is.

Montellese (changes requested):
Code changes look good, thanks.

Please add a commit with the patch of your current / first commit added to `https://github.com/xbmc/xbmc/tree/master/lib/libUPnP/patches` where all the other custom patches are located as well.

jmaggard10:
Done.

MartijnKaijser:
Jenkins build this please

MartijnKaijser:
jenkins build this please

lib/libUPnP/Platinum/Source/Devices/MediaServer/PltMediaItem.cpp
a1rwulf: yep, got it - they mix newline vs sameline in some places, so nvm.

Montellese (approved):
Thanks, all good for me now!

MartijnKaijser:
jenkins build this please

PR #12120, opened 2017-05-18
[Music]Better artist and album scraping during library update, store scraped mbids

DaveTBlake:
Make online scraping of additional artist and album information a bit more efficient - avoid futile repeat requests and make better use of Musicbrainz IDs.

When fetching additional info on library update:
- Scan all metadata tags from the music files first before scraping (rather than do it album at a time).  
*This populates the library with the basic music without delay so the user can start using it.
*Tagging can be mixed quality - some with mbids, some without, so artist mbid may be obtained from music files later in the scan.
*Avoids unwanted consequnces that can arise from the order in which folders are processed. The common example of this is when the user has their Various Artist compilations in a folder called "Compilations" such that it gets processed _before_ the albums of artists in folders beginning with a letter later in the alphabet. The additional info for the song artists from these compilation albums get added to the library and scraped before the other albums by that artist had been scanned, and so the artist.nfo file is never found and applied. 
*Having passed through all the music during scanning tags we know how many albums we are processing, so during scraping phases the progress bar can show meaningful position though task.
*Repeat of failed scraping due to server timeouts could be added at a later time.
- Scrape album artists even when scraping the album has failed (as proposed in #11673 )
Otherwise known artists can go unscraped just because the albums in the library are not found.
- Avoid repeatedly requesting artists that are unknown (could happen every song, and for every album otherwise), just try once.

Currently only Musicbrainz album ids (release id) derived from music file tags are stored in the db, despite the fact that every time scraping is attempted Musicbrainz ids are fetched for those items that don't have them. This can mean repeated lookups of the same item from Musicbrainz over and over again, leading to their server being unnecessarily overloaded and subsequent requests timing-out. 

Also lookups of artwork at Fanart.tv and TADB use Musicbrainz release group id, which has to be fetched each time despite that it could be scanned from music files tags and stored.

This change scans and stores Musicbrainz release group id (in addition to release id)  from tags, and stores both scraped mbids in the db for later use by scrapers and other addons that do a mbid based lookup such as cdArt etc., but flags them as scraped so that any mis-identified artists or albums can be corrected manually by refresh from the info dialog (it doesn't use any scraped mbids, only those from tags, but does a name lookup again instead). 

When scraping an album, even if "Prefer online information" is disabled, any missing artist mbids are taken from the scraped values when names match.

Also fix the way that scraped albums are merged when "Prefer online information" is enabled. Restore the original intent which was to be able to keep the library upto date with Musicbrainz database changes, including changes to album and song artist credits  and song titles,  without having to re-tag. This only applies when the original music files are tagged with mbids. For music not tagged with mbids, and thus lookup is by name only, with "Prefer online information" enabled Kodi does not attempt to change the artist credits, only the album informations only a few fields of which could have been populated by tag data.

### DIFF
--- a/xbmc/music/Album.cpp
+++ b/xbmc/music/Album.cpp
@@ -51,6 +51,7 @@ CAlbum::CAlbum(const CFileItem& item)
   tag.GetReleaseDate(stTime);
   strAlbum = tag.GetAlbum();
   strMusicBrainzAlbumID = tag.GetMusicBrainzAlbumID();
+  strReleaseGroupMBID = tag.GetMusicBrainzReleaseGroupID();
   genre = tag.GetGenre();
   std::vector<std::string> musicBrainzAlbumArtistHints = tag.GetMusicBrainzAlbumArtistHints();
   strArtistDesc = tag.GetAlbumArtistString();
@@ -206,12 +207,80 @@ CAlbum::CAlbum(const CFileItem& item)
 void CAlbum::MergeScrapedAlbum(const CAlbum& source, bool override /* = true */)
 {
   /*
-   We don't merge musicbrainz album ID so that a refresh of album information
-   allows a lookup based on name rather than directly (re)using musicbrainz.
-   In future, we may wish to be able to override lookup by musicbrainz so
-   this might be dropped.
-   */
-//  strMusicBrainzAlbumID = source.strMusicBrainzAlbumID;
+   Initial scraping of album information when there is a Musicbrainz album ID derived from 
+   tags is done directly using that ID, otherwise the lookup is based on album and artist names
+   but this can sometimes mis-identify the album (i.e. classical music has many "Symphony No. 5").
+   It is useful to store the scraped mbid, but we need to be able to correct any mistakes. Hence 
+   a manual refresh of album information uses either the mbid as derived from tags or the album 
+   and artist names, not any previously scraped mbid.
+
+   When overwritting the data derived from tags, AND the original and scraped album have the same
+   Musicbrainz album ID, then merging is used to keep Kodi up to date with changes in the Musicbrainz 
+   database including album artist credits, song artist credits and song titles. However it is ony 
+   appropriate when the music files are tagged with mbids, these are taken as definative, scraped
+   mbids can not be depended on in this way.
+
+   When the album is megerd in this deep way it is flagged so that the database album update is aware
+   artist credits and songs need to be updated too.
+  */
+
+  bArtistSongMerge = override && !bScrapedMBID
+    && !source.strMusicBrainzAlbumID.empty() && !strMusicBrainzAlbumID.empty() 
+    && (strMusicBrainzAlbumID.compare(source.strMusicBrainzAlbumID) == 0);
+
+  /*
+   Musicbrainz album (release) ID and release group ID values derived from music file tags are
+   always taken as accurate and so can not be overwritten by a scraped value. When the album does 
+   not already have an mbid or has a previously scraped mbid, merge the new scraped value, 
+   flagging it as being from the scraper rather than derived from music file tags.
+  */
+  if (!source.strMusicBrainzAlbumID.empty() && (strMusicBrainzAlbumID.empty() || bScrapedMBID))
+  {
+    strMusicBrainzAlbumID = source.strMusicBrainzAlbumID;
+    bScrapedMBID = true;
+  }
+  if (!source.strReleaseGroupMBID.empty() && (strReleaseGroupMBID.empty() || bScrapedMBID))
+  {
+    strReleaseGroupMBID = source.strReleaseGroupMBID;
+  }
+
+  /*
+   Scraping can return different album artists from the originals derived from tags, even when doing
+   a lookup on name. 
+
+   When overwritting the data derived from tags, AND the original and scraped album have the same
+   Musicbrainz album ID, then merging an album replaces both the album artsts and the song artists
+   with those scraped. 
+   
+   When not doing that kind of merge, for any matching artist names the Musicbrainz artist id
+   returned by the scraper can be used to populate any previously missing Musicbrainz artist id values.
+  */
+  if (bArtistSongMerge)
+  {
+    artistCredits = source.artistCredits; // Replace artists and store mbid returned by scraper    
+    strArtistDesc.clear();  // @todo: set artist display string e.g. "artist1 & artist2" when scraped 
+  }
+  else
+  {
+    // Compare original album artists with those scraped (ignoring order), and set any missing mbid
+    for (auto &artistCredit : artistCredits)
+    {
+      if (artistCredit.GetMusicBrainzArtistID().empty())
+      {
+        for (auto sourceartistCredit : source.artistCredits)
+        {
+          if (StringUtils::EqualsNoCase(artistCredit.GetArtist(), sourceartistCredit.GetArtist()))
+          {
+            artistCredit.SetMusicBrainzArtistID(sourceartistCredit.GetMusicBrainzArtistID());
+            artistCredit.SetScrapedMBID(true);
+            break;
+          }
+        }
+      }
+    }
+  }
+
+  //@todo: scraped album genre needs adding to genre and album_genre tables, this just changes the string
   if ((override && !source.genre.empty()) || genre.empty())
     genre = source.genre;
   if ((override && !source.strAlbum.empty()) || strAlbum.empty())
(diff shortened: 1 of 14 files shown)


xbmc/music/MusicDatabase.cpp
```diff
@@ -482,18 +484,27 @@ bool CMusicDatabase::UpdateAlbum(CAlbum& album, bool OverrideTagData /* = true*/
...
+              album.bScrapedMBID);
 
-  if (OverrideTagData)
+  if (!OverrideTagData)
+  {
+    // Album artist(s) already exist and names are not changing, but may have scraped Musicbrainz ids to add
+    for (const auto &artistCredit : album.artistCredits)
+    {
+      UpdateArtistScrapedMBID(artistCredit.GetArtistId(), artistCredit.GetMusicBrainzArtistID());
+    }
```
Rechi: curly brackets not necessary

xbmc/music/MusicDatabase.cpp
```diff
@@ -1243,12 +1263,12 @@ int CMusicDatabase::AddArtist(const std::string& strArtist, const std::string& s
...
-    // 3) No artist exists at all - add it
+    // 3) No artist exists at all - add it, flagging when has scarped mbid
     if (strMusicBrainzArtistID.empty())
       strSQL = PrepareSQL("INSERT INTO artist (idArtist, strArtist, strMusicBrainzArtistID) VALUES( NULL, '%s', NULL )",
         strArtist.c_str());
     else
-      strSQL = PrepareSQL("INSERT INTO artist (idArtist, strArtist, strMusicBrainzArtistID) VALUES( NULL, '%s', '%s' )",
+      strSQL = PrepareSQL("INSERT INTO artist (idArtist, strArtist, strMusicBrainzArtistID, bScrapedMBID) VALUES( NULL, '%s', '%s', %i )",
         strArtist.c_str(),
         strMusicBrainzArtistID.c_str());
```
Rechi: value of `bScrapedMBID` missing

xbmc/music/infoscanner/MusicInfoScanner.cpp
```diff
@@ -830,75 +848,122 @@ int CMusicInfoScanner::RetrieveMusicInfo(const std::string& strDirectory, CFileI
...
-    m_handle->SetTitle(g_localizeStrings.Get(505));
-
-  return numAdded;
 }
 
 void CMusicInfoScanner::FindArtForAlbums(VECALBUMS &albums, const std::string &path)
 {
   /*
-   If there's a single album in the folder, then art can be taken from
+f   If there's a single album in the folder, then art can be taken from
```
Rechi: changed by accident

xbmc/music/infoscanner/MusicInfoScanner.cpp
```diff
@@ -1047,11 +1112,11 @@ INFO_RET CMusicInfoScanner::UpdateDatabaseArtistInfo(CArtist& artist, const ADDO
   if (!scraper)
     return INFO_ERROR;
 
-  CMusicArtistInfo artistInfo;
-
+  CMusicArtistInfo artistInfo;  
+
```
Rechi: unrelated white space changes

xbmc/music/infoscanner/MusicInfoScanner.h
```diff
@@ -216,8 +221,9 @@ class CMusicInfoScanner : CThread, public IRunnable, public CInfoScanner
   int m_scanType; // 0 - load from files, 1 - albums, 2 - artists
   CMusicDatabase m_musicDatabase;
 
-  std::map<CAlbum, CAlbum> m_albumCache;
-  std::map<CArtistCredit, CArtist> m_artistCache;
+  //std::map<CAlbum, CAlbum> m_albumCache;
+  //std::map<CArtistCredit, CArtist> m_artistCache;
```
Rechi: if not needed, remove it

DaveTBlake:
Many thanks for code check @Rechi 

This could do with some more testing, scraping mixed music sources (some tagged with mbids, some not) from scratch with "Fetch additional information on library update" enabled. Then checking that items mis-identified by auto scraping (it makes mistakes) can be corrected by refreshing from album/artist info dialog.

Also worth trying limited album and artist scraping (from a filtered node), with "Prefer online information" enabled or disabled.

xbmc/music/Album.cpp
```diff
@@ -206,12 +206,58 @@ CAlbum::CAlbum(const CFileItem& item)
...
+  {
+    // Compare original album artists with those scraped (ignoring order), and set any missing mbid
+    int matchcount = 0;
+    for (auto &artistCredit : artistCredits)
+    {
+      if (artistCredit.GetMusicBrainzArtistID().empty())
+      {
+        bool found = false;
+        auto sourceartistCredit = source.artistCredits.begin();
+        while (sourceartistCredit != source.artistCredits.end() && !found)
```
DaveTBlake: So break from within a for loop is prefered?

Rechi: for me a range-based for loop with a break is easier to read

xhaggi: Yep it's preferred and as @Rechi said more readable. And drop found just if string is equal set and break.

xbmc/music/Album.cpp
```diff
@@ -206,12 +206,56 @@ CAlbum::CAlbum(const CFileItem& item)
...
-//  strMusicBrainzAlbumID = source.strMusicBrainzAlbumID;
+   Initial scraping of album information when there is a Musicbrainz album ID derived from 
+   tags is done directly using that ID, otherwise the lookup is based on album and artist names
+   but this can sometimes mis-identify the album (i.e. classical music has many "Symphony No. 5").
+   It is useful to store the scraped mbid, but we need to be able to correct any mistakes. Hence 
+   a manual refresh of album information uses either the mbid as derived from tags or the album 
+   and artist names, not any previously scraped mbid.
+  
+  A Musicbrainz album ID derived from music file tags is always taken as accurate and so can
+  not be overwritten by a scraped value. When the album does not already have an mbid or has
```
xhaggi: Please add another space for the whole block to fit the one above.

xbmc/music/Album.cpp
```diff
@@ -206,12 +206,56 @@ CAlbum::CAlbum(const CFileItem& item)
...
+  {
+    strMusicBrainzAlbumID = source.strMusicBrainzAlbumID;
+    bScrapedMBID = true;
+  }
+
+  /*
+  Scraping can return different album artists from the originals derived from tags. When overwritting
+  the data derived from tags the original album artists are replaced with those scraped. However even
+  when not overwritting derived data, for any matching artist names the Musicbrainz artist id returned 
+  by the scraper can be used to populate any previously missing Musicbrainz artist id values.
```
xhaggi: Please add a space to all comments.

```
/*
 My comment
 
 My comment2
*/
```

xbmc/music/infoscanner/MusicInfoScanner.cpp
```diff
@@ -791,17 +804,21 @@ int CMusicInfoScanner::RetrieveMusicInfo(const std::string& strDirectory, CFileI
...
 
-  int numAdded = 0;
-  ADDON::AddonPtr addon;
-  ADDON::ScraperPtr albumScraper;
-  ADDON::ScraperPtr artistScraper;
-  if(ADDON::CAddonSystemSettings::GetInstance().GetActive(ADDON::ADDON_SCRAPER_ALBUMS, addon))
-    albumScraper = std::dynamic_pointer_cast<ADDON::CScraper>(addon);
+  /* Strategy: Having scanned tags and made a list of albums, add them to the library. Only then try
+  to scrape additional album and artist information. Music is often tagged to a mixed standard
+  - some albums have mbid tags, some don't. Once all the musc files have been added to the library,
```
scott967: +  - some albums have mbid tags, some don't. Once all the music files have been added to the library,

xbmc/music/infoscanner/MusicInfoScanner.cpp
```diff
@@ -830,69 +848,116 @@ int CMusicInfoScanner::RetrieveMusicInfo(const std::string& strDirectory, CFileI
...
+    m_handle->SetTitle(g_localizeStrings.Get(505));
+
+  return numAdded;
+}
+
+void MUSIC_INFO::CMusicInfoScanner::ScrapeInfoAddedAlbums()
+{
+  /* Strategy: Having scanned tags, make a list of albums and add them to the library, only then try
+  to scrape additional album and artist information. Music is often tagged to a mixed standard
+  - some albums have mbid tags, some don't. Once all the musc files have been added to the library,
```
scott967: +  - some albums have mbid tags, some don't. Once all the music files have been added to the library,

xbmc/music/infoscanner/MusicInfoScanner.cpp
```diff
@@ -830,69 +848,116 @@ int CMusicInfoScanner::RetrieveMusicInfo(const std::string& strDirectory, CFileI
...
+      // Fetch any artist mbids for album artist(s) and song artists when scraping those too.
+      m_musicDatabase.GetAlbum(albumId, album, !albumartistsonly);
+      album.bScrapedMBID = m_musicDatabase.HasScrapedAlbumMBID(albumId);
+      UpdateDatabaseAlbumInfo(album, albumScraper, false);
 
-      INFO_RET albumScrapeStatus = INFO_NOT_FOUND;
-      if (!m_musicDatabase.HasAlbumBeenScraped(album->idAlbum))
-        albumScrapeStatus = UpdateDatabaseAlbumInfo(*album, albumScraper, false);
+      // Scrape information for artists that have not been scraped before, avoiding repeating
+      // unsucessful attempts for every alnum and song.
```
scott967: +      // unsuccessful attempts for every album and song.

DaveTBlake:
Thanks for the code syntax and spelling corrections guys, it is appreciated.

Some further design discussion here http://forum.kodi.tv/showthread.php?tid=306218&pid=2588005#pid2588005

DaveTBlake:
After chatting with @night199uk about the original intent of album merge and the "Prefer online information" setting, I have added a post to correct this behaviour.

The idea is to able to keep the library upto date with Musicbrainz database changes, including changes to album and song artist credits  and song titles,  without having to re-tag. This only applies when the original music files are tagged with mbids. 

For music not tagged with mbids, and thus lookup is by name only, with "Prefer online information" enabled Kodi does not attempt to change the artist credits and song titles, only the album information (only a few fields of which could have been populated by tag data).

DaveTBlake:
Tyring to create a new test build with this and #12402 (Fix Musicbrainz Throttling) combined, but Jenkins won't play.

Adjusted to scan from tags and store scraped values of Musicbrainz release group id for use in subsequent art or info lookups by addons etc.

zag2me:
Since TADB only stores the MB release group ID, this change is most welcome!

DaveTBlake:
Test build has received some user testing, so  have squashed the commits and intend to merge at the end of today unless anyone has any objections.

PR #12125, opened 2017-05-18
[add-ons/settings] migrate add-on settings to settings library

Montellese:
## Description
So this is it. I've been "dreaming" of this final step ever since I worked on the original settings rework in core but it turned out to be much more complicated than expected which is also why it took so long. This PR introduces (almost) all the necessary bits to be able to use the same settings system for add-ons that we already in core. Furthermore it contains a backwards compatibility layer which reads in old settings definitions (and values) and translates those into settings from the new system. After this PR settings will always be saved in the same format as `guisettings.xml` but it can read the settings definition either from the old add-on settings format or from the "new" core settings system format.

Since I'm not that much of an add-on user myself and since I've never written an add-on myself I'm pretty sure I missed a few things in the backwards compatibility layer. So it would be greatly appreciated if people with more experience in this area could give this a try. But beware once Kodi has written the setting values in the new format you can't go back to the old format without losing all your add-on settings. So best make a backup of your add-on data before giving this a try.

There are a few parts that are known not to be backwards compatible but I've tried to add log messages in those places so that it becomes obvious. I hope that those cases are edge cases that can live without backwards compatibility but who knows.

There are also parts that I wasn't able to test like binary add-ons defining settings through the binary add-on API since I don't know if and which add-on is using this at all.

In the end the major benefit is that `GUIDialogAddonSettings` has become very simple because it can derive from the existing settings related base classes. The major part of the work is in `CAddonSettings` which also contains the backwards compatibility layer which we can hopefully drop someday (a few releases into the future).

I'm not sure if we want to "mark" the old add-on settings approach as deprecated to try and get add-ons to adopt to the new approach or not.

Some (basically the first 25) of the commits could go into a separate PR (as I've already done with a few fixes before) but since they didn't bring any real benefit to the existing code I didn't do that yet. If someone would like me to do that to make reviewing easier I'm fine with that. Just let me know.

Part of this work also overlaps with my media importing work (`CSettingsBase` et al.) and should make things a bit easier there as well code-wise.

## How Has This Been Tested?
Locally by installing some add-ons, opening their add-on settings and comparing the result to before.

## Types of change

## Checklist:


MartijnKaijser:
Just awesome :) 
I'd say mark old settings as deprecated and get v18 Addons on the new system.

xbmc/addons/settings/AddonSettings.cpp
```diff
@@ -0,0 +1,1278 @@
...
+  else
+    return false;
+
+  expression.m_relativeSettingIndex = static_cast<int32_t>(strtol(str.substr(posOpen + 1, posSep - posOpen - 1).c_str(), nullptr, 10));
+  expression.m_value = str.substr(posSep + 1, posClose - posSep - 1);
+
+  return true;
+}
+
+void CAddonSettings::FileEnumSetingOptionsFiller(std::shared_ptr<const CSetting> setting, std::vector< std::pair<std::string, std::string> > &list, std::string &current, void *data)
```
AlwinEsch: Maybe a small type :) "tt"

AlwinEsch:
Has looked over them, is really big :) From my side looks it good.

Just a question of understanding. Are the shared pointers present all the time or are them released after usage?

xbmc/addons/settings/AddonSettings.cpp
Montellese: Thanks and fixed.

Montellese:
Well as long as the `CSettingsManager` instance which created the settings is alive the (shared) pointers will be alive as well. Nothing changes in that matter.

But to be honest I've noticed that add-on settings are created and destroyed waaaaaay too often because add-on instances are created and destroyed waaaaaay too often. But I didn't wanna go down that rabbit hole :-/

AlwinEsch:
I know :-D that's part of my rework to reduce them.

Addon becomes loaded and stays loaded as long it is used if now a new instance becomes used is still the loaded present and not need to do all the work.

Montellese:
I forgot to mention that there are two known backwards incompatibilities (both marked with a TODO in the code):
* it's not possible to specify that the add-on settings dialog should close after having edited a setting. I don't really know what the intention / purpose behind this is / was.
* it's not possible to specify multiple add-on types for an add-on setting. but then again I couldn't find any add-on using this.

MartijnKaijser:
Doesn't sound like a problem

xbmc/interfaces/json-rpc/SettingsOperations.cpp
```diff
@@ -684,6 +695,26 @@ bool CSettingsOperations::SerializeSettingAddon(const CSettingAddon* setting, CV
   return true;
 }
 
+bool CSettingsOperations::SerializeSettingDate(const CSettingDate* setting, CVariant &obj)
```
notspiff: just because it piqued my curiosity; why isn't this done as CSettingDate::Serialize ?

Montellese: You mean add the JSON serialization to the `CSettingFoo` classes? Right now those classes only have deserialization logic to read the setting definition from XML. This would be serializing the setting definition to JSON. So it might be a bit confusing.

Furthermore any changes to the JSON serialization implementation could require changes to the JSON-RPC API schema. That is more obvious if the serialization code is located with the rest of the JSON-RPC API code.

But I agree that it's not the cleanest solution.

notspiff: makes perfect sense, thanks for explaining.

notspiff: follow up question though: is the xml really of any value here or could you use json on disk as well? i doubt many ever edit these by hand.

Montellese: It shouldn't be too hard to write the setting definitions in JSON instead of XML and write a parser for JSON. Now (after all the GUI integration rewriting) it would probably even be possible to write such a parser outside of the whole settings library. That's basically what I did for the backwards compatibility layer. Maybe it would be worth extracting all the deserialization logic from the settings library and leave that up to the "integrator". There might be some tricky business around being able to overwrite setting definitions e.g. per platform but it should all be doable.

MartijnKaijser:
Needs a rebase.
* I never knew we could close the dialog after editing a setting so i doubt any one used that.
* Same for the multiple addon types for settings. Maybe it never got finished?

MartijnKaijser:
jenkins build this please

Montellese:
Should now also build on Linux and on win32 with tests. jenkins build this please

Montellese:
jenkins build this please

Montellese:
jenkins build this please

MartijnKaijser:
All green :) 
Go for it

MartijnKaijser:
tested and so far all fine

Montellese:
@MartijnKaijser thanks for testing. Did you check the log file for any warnings / errors from `CAddonSettings`?

I'd appreciate it if someone else would be able to test this as well before we merge this and break everyone's add-on settings ;-) Maybe @ronie has some time as he probably knows this area best?

MartijnKaijser:
Nope all looks fine so far.

AlwinEsch:
You need to do a rebase, there is a new function added (https://github.com/xbmc/xbmc/blob/master/xbmc/addons/AddonDll.cpp#L629-L704) who bring faults to them.

Montellese:
@AlwinEsch: thanks for the heads-up. Github doesn't realize it but obviously the build will break. Will fix it tomorrow.

Montellese:
jenkins build this please

Montellese:
jenkins build this please

Montellese:
Something's wrong with the win32 build slave. It always fails building `libavfilter/avfilter-6.dll`.

MartijnKaijser:
jenkins build this please

MartijnKaijser:
Perhaps @MilhouseVH can include it in his LE builds?

ronie:
this PR doesn't compile anymore, due to https://github.com/xbmc/xbmc/pull/12187

Montellese:
*grml* will fix later

Montellese:
done (until the next merge)

ronie:
first issue i came across after a brief test, we don't seem to handle settings which aren't part of the settings.xml file of an addon.
i'm referring to settings which are set directly by the python addon, 
using _xbmcaddon.setSetting(id, value)_

for instance, this is (part of) the settings.xml file of the last.fm scrobbler:
```
<settings>
	<category label="30000">
		<setting id="lastfmuser" type="text" label="30003" default=""/>
		<setting id="lastfmpass" type="text" label="30004" default="" option="hidden"/>
	</category>
</settings>
```

after the user has provided his username and password, the addon will retrieve a session key and save it to the settings.xml in userdata:
```
<settings>
    <setting id="lastfmuser" value="ronie" />
    <setting id="lastfmpass" value="iloveponies" />
    <setting id="lastfmkey" value="d6bd7378515ec88b685070f4e9780d73" />
</settings>
```

the next time i start kodi, this is logged:
```
21:21:47.112 T:140586887669504   DEBUG: CAddonSettings[service.scrobbler.lastfm]: loading setting definitions
21:21:47.112 T:140586887669504 WARNING: CAddonSettings[service.scrobbler.lastfm]: trying to load setting definitions from old format...
21:21:47.112 T:140586887669504   DEBUG: CAddonSettings[service.scrobbler.lastfm]: loading setting values
21:21:47.112 T:140586887669504   DEBUG: CSettingsManager: requested setting (lastfmkey) was not found.
21:21:47.112 T:140586887669504   DEBUG: CAddonSettings[service.scrobbler.lastfm]: failed to find definition for old setting lastfmkey
```

and the value of the lastfmkey setting is **not** passed to the addon.


i see the same issue with quite a few other addons as well, versioncheck for instance:
```
21:21:52.925 T:140587052844800   DEBUG: CAddonSettings[service.xbmc.versioncheck]: loading setting definitions
21:21:52.925 T:140587052844800 WARNING: CAddonSettings[service.xbmc.versioncheck]: trying to load setting definitions from old format...
21:21:52.925 T:140587052844800   DEBUG: CAddonSettings[service.xbmc.versioncheck]: loading setting values
21:21:52.925 T:140587052844800   DEBUG: CSettingsManager: requested setting (lastnotified_version) was not found.
21:21:52.925 T:140587052844800   DEBUG: CAddonSettings[service.xbmc.versioncheck]: failed to find definition for old setting lastnotified_version
```

ronie:
second 'issue' i noticed is a failure to parse a setting which is defined in settings.xml, but without a type definition.

for instance:
```
<?xml version="1.0" encoding="utf-8" standalone="yes"?>
<settings>
	<category label="30000">
		<setting id="lastfmstatus" default="xxxxxxxxxx"/>
	</category>
</settings>
```

result in log:
```
21:21:47.112 T:140586887669504   DEBUG: CAddonSettings[service.scrobbler.lastfm]: loading setting definitions
21:21:47.112 T:140586887669504 WARNING: CAddonSettings[service.scrobbler.lastfm]: trying to load setting definitions from old format...
21:21:47.112 T:140586887669504 WARNING: CAddonSettings[service.scrobbler.lastfm]: failed to parse old setting definition for "lastfmstatus" of type ""
21:21:47.112 T:140586887669504   DEBUG: CAddonSettings[service.scrobbler.lastfm]: loading setting values
221:21:47.112 T:140586887669504   DEBUG: CSettingsManager: requested setting (lastfmstatus) was not found.
21:21:47.112 T:140586887669504   DEBUG: CAddonSettings[service.scrobbler.lastfm]: failed to find definition for old setting lastfmstatus
```

perhaps not a big issue and the addon dev is to blame here for not specifying a type,
but this wasn't a problem in the old situation and it might break some addons.

ronie:
for settings of type="text" and option="hidden", we don't display asterisks symbols anymore...

before:
before


after:
after


so there's no indication anymore you've already provided your password or not.

Montellese:
@ronie: Thanks a lot for your feedback. This is much appreciated.
* I didn't know that add-ons could simply set any additional settings through the python interface. I'll have to look into that. Might be tricky.
* Parsing setting definitions without a type should be ok if we can simply assume that it's a string / text. Is that ok?
* I'll take a look at the missing asterisks. That should be easy to fix.

Let me know if you come across any other issues. It's much easier for you to spot them since you know how things work or rather should work.

Montellese:
@ronie: could you check if the "lastfmstatus" setting without a type definition appears in the GUI or not? If I read the old code correctly it shouldn't.

ronie:
> @ronie: could you check if the "lastfmstatus" setting without a type definition appears in the GUI or not? If I read the old code correctly it shouldn't.

nope, it doesn't (and shouldn't) appear in the settings dialog.
as mentioned, this one is perhaps not a big issue and maybe something we don't have to support...
i don't even recall why i made 'lastfmstatus' a setting, it's basically an api key (which never changes) so i might as well hardcode it in the python code instead.

ronie:
> Parsing setting definitions without a type should be ok if we can simply assume that it's a string / text. Is that ok?

yup, that's certainly ok.

xbmc/addons/settings/AddonSettings.cpp
```diff
@@ -316,7 +316,7 @@ bool CAddonSettings::InitializeFromOldSettingDefinitions(const CXBMCTinyXML& doc
     std::vector<SettingWithConditions> settingsWithConditions;
 
     // prepare a setting group
-    std::shared_ptr<CSettingGroup> group(new CSettingGroup("0", GetSettingsManager()));
+    auto group = std::make_shared<CSettingGroup>("0", GetSettingsManager());
```
akva2: Afaik kodi is still c++11 not 14?

Montellese:
@ronie 
* setting definitions without a type should be handled as string settings not shown in the GUI now
* for settings with a hidden value asterisks should be shown now
* fixing the problem with settings being added on-the-fly will take some time because it's non-trivial

xbmc/addons/settings/AddonSettings.cpp
Montellese: Hm? `std::make_shared<>()` should be available in C++11. It's only `std::make_unique<>()` which is missing in C++11 and has been added in C++14. Or did you mean something else?

notspiff: ah, i got them mixed up, sorry.

Montellese:
@ronie: Please give it another try. I have extended the settings library to support adding settings after it has been initialized and integrated it into the add-on settings stuff.

Concerning the python API should I also add methods to get/set settings of a specific type (bool, int, float, ...) instead of just string? That should be pretty easy now.

MartijnKaijser:
@Montellese yes that would be pretty nice to have extra :)

MartijnKaijser:
@Montellese i've tested that scenario and it seems to work now as settings are saved and retrieved. thx

Montellese:
@MartijnKaijser done. Let me know if the new methods in the python `Addon` class make sense like this. Do I need to bump some version number somewhere because the API has changed?

xbmc/interfaces/legacy/Addon.h
```diff
@@ -140,6 +142,122 @@ namespace XBMCAddon
...
+      /// Returns the value of a setting as a floating point number.
+      ///
+      /// @param id                      string - id of the setting that the module
+      ///                                needs to access.
+      /// @return                        Setting as a floating point number
+      ///
+      ///
+      ///-----------------------------------------------------------------------
+      /// @python_v13
+      /// **id** is optional as it will be auto detected for this add-on instance.
```
MartijnKaijser: Should be python_v18
and with the text "New function added."
The current text about id being option should be added few lines above.

xbmc/interfaces/legacy/Addon.h
```diff
@@ -168,6 +286,122 @@ namespace XBMCAddon
...
+      ///
+      /// @note You can use the above as keywords for arguments.
+      ///
+      ///
+      ///-----------------------------------------------------------------------
+      ///
+      /// **Example:**
+      /// ~~~~~~~~~~~~~{.py}
+      /// ..
+      /// self.Addon.setSettingNumber(id='username', value='teamkodi')
```
MartijnKaijser: example doesn't fit the function

MartijnKaijser (commented):
I'll let @ronie also comment on this is if this is the best way to add these functions.
Edit:
Perhaps we can do this addition as separate PR so we can get the rest merged?
Yes a python version bump is needed i guess and we can keep it at that version for v18.

xbmc/interfaces/legacy/Addon.h
Montellese: Fixed `python_v18` and added the text `New function added.`.
Where should I move the text about id being optional? Also we should probably change one of the referenced `id` parameters to something else to avoid confusion. The `id` parameter of the `xbmcaddon.Addon([id])` call is optional but the one of the `getSettingNumber(id)` isn't.

xbmc/interfaces/legacy/Addon.h
Montellese: Nice spotting. Fixed it to `setSettingString()`.

xbmc/interfaces/legacy/Addon.h
MartijnKaijser: Well that certainly makes it confusing having the same `id`.
Perhaps make it `getSettingBool(setting_id)` or something because `id` is generally use for the plugin id already and we shouldn't change that one.
So we'd get:
`@param id  optional as it will be auto detected for this add-on instance`
`@param setting_id   string - id of the setting that the module needs to access`

xbmc/interfaces/legacy/Addon.h
```diff
@@ -168,6 +289,134 @@ namespace XBMCAddon
...
+      /// @param value               boolean - value of the setting.
+      /// @return                    True if the value of the setting was set, false otherwise
+      ///
+      ///
+      /// @note You can use the above as keywords for arguments.
+      ///
+      ///
+      ///-----------------------------------------------------------------------
+      /// @python_v18 New function added.
+      /// **id** is optional as it will be auto detected for this add-on instance.
```
MartijnKaijser: you can remove the **id** is `optional` lines as that's added to the changelog. Since this is new the param is already optional from start.

xbmc/interfaces/legacy/Addon.h
```diff
@@ -168,6 +289,134 @@ namespace XBMCAddon
...
+      ///
+      ///
+      /// **Example:**
+      /// ~~~~~~~~~~~~~{.py}
+      /// ..
+      /// self.Addon.setSettingBool(id='enabled', value=True)
```
MartijnKaijser: will `self.Addon.setSettingBool('enabled', value=True)` (without the id= )also work as iirc we don't really mention this in other places. Maybe @ronie or @phil65 has an opinion on this?

xbmc/interfaces/legacy/Addon.h
Montellese: IMO the `id` parameter of the `xbmcaddon.Addon()` method shouldn't be mentioned here at all because there should be doxygen for that method somewhere else.

xbmc/interfaces/legacy/Addon.h
Montellese: Sure as always the keywords are optional. You can also drop the `value=` part and it will still work. I just copied the doxygen description of the existing `getSetting()` and `setSetting()` methods.

xbmc/interfaces/legacy/Addon.h
Montellese: Ah I didn't realize this was the changelog. I adjusted it again by
* removing the "**id** is optional" part
* moving the "New function added." part to a separate line

xbmc/interfaces/legacy/Addon.h
MartijnKaijser: Also fine 👍

Montellese:
@MartijnKaijser just saw your last edit. I removed the two commits for the new python methods. They are now in a separate branch which can be rebased and PRed once this has been merged.

MartijnKaijser:
Merged :)

Montellese:
Now let's watch the world burn :-D

MartijnKaijser:
@Montellese http://paste.ubuntu.com/24770395/  :)

MartijnKaijser:
@Montellese should we really log the loading old settingsformat as warning?

> 16:39:35.447 T:367774405600 WARNING: CAddonSettings[pvr.hts]: trying to load setting definitions from old format...
16:39:35.984 T:367790018256 WARNING: Previous line repeats 16 times.
16:39:35.985 T:367790018256 WARNING: CAddonSettings[plugin.video.youtube]: trying to load setting definitions from old format...
16:39:36.061 T:367790018256 WARNING: Previous line repeats 2 times.
16:39:36.062 T:367790018256  NOTICE: [plugin.video.youtube] Startup: detected Krypton (Kodi-18.0), setting DASH_SUPPORT_BUILTIN = False, DASH_SUPPORT_ADDON = True
16:39:36.337 T:367774405600  NOTICE: Start - EPG thread started
16:39:36.354 T:367797687504 WARNING: CAddonSettings[weather.openweathermap.extended]: trying to load setting definitions from old format...
16:39:40.076 T:367781754448 WARNING: Previous line repeats 10 times.
16:39:40.077 T:367781754448 WARNING: CAddonSettings[service.xbmc.versioncheck]: trying to load setting definitions from old format...
16:39:56.575 T:367809422720 WARNING: CAddonSettings[script.kodi.loguploader]: trying to load setting definitions from old format...

stefansaraev:
uhm. maybe stupid question but.. is

```
<setting id="xxx" ... enable="eq(-1,1)" />
```

or
```
<setting id="xxx" ... visible="eq(-1,1)" />
```

supposed to work after this change?

Montellese:
@MartijnKaijser: will look at the reports later this week. Concerning the logging I did that on purpose to try to force add-on developers to switch to the new approach but I'm fine with changing log level to debug because (as already mentioned before) add-on settings are loaded waaaay too often.

@stefansaraev: It should still work. I tested it with one of the weather add-ons which enables/disables certain settings based on other settings. If something doesn't work you should probably see an error in the log file.

stefansaraev:
it works if the setting checked in condition is bool:

https://github.com/stefansaraev/tb.settings/blob/master/src/resources/settings.xml#L19-L20

not for labelenums: 

https://github.com/stefansaraev/tb.settings/blob/master/src/resources/settings.xml#L24-L29

Montellese:
@setfansaraev this might be limited to labelenums. It took me ages to figure out how those work and I probably still got something wrong. I'll look into it, thanks.

Montellese:
@MartijnKaijser any chance we can silence CID 170206, 170207, 170209 and 170210? I want those checks to stay there because they will become important when the value of `MinimumSupportedVersion` changes.

Rechi:
set classification to intentional at coverity.com

Montellese:
@Rechi: done, thanks.

Montellese:
@stefansaraev are you sure you're using the `labelenum` setting type correctly? According to the documentation at http://kodi.wiki/view/Add-on_settings#type.3D.22enum.22_or_.22labelenum.22 when using `labelenum` the value of the setting is the actual (string) value (i.e. in your case either `dhcp` or `manual`. If you want the value to be the index you need to use `enum`.

But as I said the code around `enum` and `labelenum` was extremely convoluted and therefore very hard to read, follow and understand. Maybe the old code contained some additional magic not mentioned in the wiki. But I just checked again and couldn't find anything. This is the code where it uses the actual string value for `labelenum` and the index for `enum`: https://github.com/xbmc/xbmc/blob/f22d62dc3f6e811a538dda9c434e1804abb8b95f/xbmc/addons/GUIDialogAddonSettings.cpp#L801

stefansaraev:
yep. the actual value is string I know, but comparing against int in eq() worked just like enum.

I dont mind comparing string values (those can be translated, btw, so I am not sure it's the right thing to do). however, it's not just me: ref:
https://github.com/xbmc/xbmc/blob/master/addons/metadata.tvdb.com/resources/settings.xml#L9-L10
https://github.com/xbmc/xbmc/blob/master/addons/metadata.artists.universal/resources/settings.xml#L18-L19

Montellese:
@stefansaraev ah right this is probably as ugly as it can get. It stores the setting value as a string but the comparison is done against the index. *grml* I'll try to add an ugly workaround :-/

Montellese:
@stefansaraev could you try with 640b46db0572a93d0279c8d4940ef128613e4c09? That fixes your issue for me.

stefansaraev:
@Montellese works for me.

MartijnKaijser:
@Montellese perhaps add some comment that this is for the old settings or is that already covered somewhere?

Montellese:
@MartijnKaijser it's in the method `ParseOldCondition()` and all the parsing code in `CAddonSettings` is for the old settings.

basrieter:
My add-on (Retrospect, see www.rieter.net) uses a lot (really a lot) of settings. I will give it go tomorrow and see what happens. 

One thing I created myself is the option to use absolute settingid's for visibility conditions instead of relative pointers to settings (I actually generate a setting.xml with relative values with every release). Would this now be possible to add  within the standard Kodi functionality?

Montellese:
@basrieter thanks for giving this a try and for reporting your feedback. That's much appreciated.

If you move your add-on's settings definition to the "new" settings syntax (i.e. the same Kodi uses in https://github.com/xbmc/xbmc/blob/master/system/settings/settings.xml) you can define dependencies using the setting identifiers instead of relative indices (see e.g. https://github.com/xbmc/xbmc/blob/master/system/settings/settings.xml#L1454).

MartijnKaijser:
@basrieter if you have any questions or issues feel free to make a forum thread. Might makes it easier to keep track of.

basrieter:
Is there a nightly installer I can use to test? I can only find one for older commits?

MartijnKaijser:
@basrieter it should be in nightlies. Only win32 failed for some reason so you can check again tomorrow

basrieter:
Ok, I will wait with patience on a successful Win32 build tonight.

Montellese:
@basrieter looks like @MartijnKaijser has started a win32 build for you. Once the mirrors have synced you should be able to download a test build from http://mirrors.kodi.tv/test-builds/win32/KodiSetup-20170606-34322d847d-master-x86.exe.

basrieter:
Thanks @MartijnKaijser. I just tested the build and it crashes Kodi. Here is a logfile: https://pastebin.com/iRqtpPcC

There seem to be some parsing issues but the reason for the crash is not really clear to me. What I do see is that it does not seem to like settings that show up more than once. Anything I can do to help you guys with this? 

Just for clarity: my add-on (Retrospect) is used for playing streams from public and free re-play sites that many public TV companies provide. At the initial start (and after each update) the add-on generates an updated settings.xml. It does that based on the available Retrospect related channel add-ons. The generation creates a valid settings.xml (for Kodi 17.x and below). Although I admit it is not the best of solutions, it does allow me to show all settings for all channels from the main add-on.

Montellese:
@basrieter: Thanks for your feedback. Could you provide a generated `settings.xml` from your add-on so that I can easily and reproducably test this?

Montellese:
@basrieter: Took a quick look at your log file and there are a lot of warnings of the type
```
failed to parse old setting definition for "FOO" of type "hidden"
```
There is no setting type `"hidden"`. There only is a setting type `"text"` with `option="hidden"`. What then happens is that the settings parser ignores those invalid setting definitions and then all your enable/visible conditions which use relative indices don't fit anymore because the number of settings has changed.

basrieter:
@Montellese Right. Removed those and the warnings disappeared, but the same crash: https://pastebin.com/3shvFf75

MartijnKaijser:
Crash might be related to https://github.com/xbmc/xbmc/pull/12244

basrieter:
@MartijnKaijser I will verify with a more recent build then. 
@Montellese if you want to test with my settings.xml: https://pastebin.com/Qu6guKvv

basrieter:
I checked with KodiSetup-20170606-b196817-master-x86 still the same issue. I hope my settings.xml will help you find the source.

MartijnKaijser:
@basrieter please try http://mirrors.kodi.tv/test-builds/win32/KodiSetup-20170608-78c0d2d-master-x86.exe  (might take a while)

basrieter:
Right, some of the settings actually fire my add-on again using RunScript to do some work. Sometimes it requires some input from the user and it could result in settings being updated. E.g.: I store some encrypted strings in the settings and encryption is done by a call to RunScript. 

The `option="close"` saves the current settings and closes the settings dialog before executing the RunScript. If I would not close the settings, any settings that I updated from the RunScript, while the settings dialog remains open, will be reverted as soon as the user selects "OK" in the settings dialog. 

Out of curiousity: I haven't seen a working conversion yet, but will it actually overwrite the settings.xml or will it create a guisettings.xml in the add-on's `resources` folder? I ask this because I would like to keep backwards compatibility with older Kodi versions and it would be nice to include a settings.xml for older Kodi's and a guisetttings.xml for 18 and up?

primaeval:
This has totally borked up most of my addons. I hope it is just a teething problem.

Here is the worst affected settings.xml file from TV Guide Fullscreen.
https://github.com/primaeval/script.tvguide.fullscreen/blob/c797440aa8bdc2511cfd7c3a90dcbb214f85c881/resources/settings.xml

It looks like you have only implemented using numeric ids for labels so most of my categories say General and the other settings have no label or just the id.

You have interpreted a period as a . for levels in the recreated settings.xml file. My addons use that as a space character so most of the setttings are now nonsense and not interpreted correctly in python.

The settings that use a Visible condition using System.HasAddon() don't work any more.

I expect there is a problem with RunScript() too as basreiter noted.

As ronie noticed setting settings from the python on the fly is also borked. 
I have many of them to keep track of state.

I hope these issues can be fixed. 
There are many addons out there that use the same methods as I described here.

I really don't see the need for such a drastic change as this but I don't poke my nose in the core very often unless I can't find a solution at a higher level.

notspiff:
no cheese for you!

MartijnKaijser:
@primaeval if you have no constructive criticism you might as well stay away

primaeval:
That was supposed to be constructive. 
There are some problems that some of my users have pointed out to me and I'm pointing them out here.
Of course I am a little upset as it is going to take a lot of time to fix them in my code.
As Montellese stated he doesn't make addons or delve into them very often.
As I have worked on one of the most complicated ones I would like to add my observations.
I'll try to keep the discussion more neutral if that helps from now on.

notspiff:
just in case: everything was fine up till your last line.

Montellese:
@primaeval the most important thing for the feedback to be constructive is to provide a log file (and in this case the respective settings.xml). Without the log file I have no idea which part is working, might be working or is not working.

I have already fixed the enable/visible conditions which use builtins in https://github.com/Montellese/xbmc/commits/addon_settings_fallout_2 but obviously that hasn't been merged yet.
Setting undefined settings from python should already be working as I should have fixed that before this PR was merged.

There's no need for you to change anything right now. I'll try to sort out all of the issues that are reported.

primaeval:
Thanks @Montellese.

From the settings.xml file from TV Guide Fullscreen:
https://github.com/primaeval/script.tvguide.fullscreen/blob/c797440aa8bdc2511cfd7c3a90dcbb214f85c881/resources/settings.xml

I am getting some log messages like this so far:
```
CAddonSettings[script.tvguide.fullscreen]: ignoring old setting definition of type "text" without id
CAddonSettings[script.tvguide.fullscreen]: failed to parse old setting condition "System.HasAddon(service.vpn.manager)" for "vpnmgr.connect"
```

The settings now look like this:
settings

You can see the General and empty labels.

The new settings.xml file comes out like this with unwanted nesting:
```
<settings version="1">
    <action>
        <bar>true</bar>
    </action>
    <addon>
        <logos default="true">false</logos>
        <match>2</match>
    </addon>
    <addons>
        <ini.enabled default="true">false</ini.enabled>
        <ini.file default="true"></ini.file>
        <ini.overwrite>1</ini.overwrite>
        <ini.subscriptions default="true">false</ini.subscriptions>
        <ini.type default="true">0</ini.type>
        <ini.url default="true"></ini.url>
    </addons>
    <alternative>
        <playback default="true">false</playback>
    </alternative>
    <append>
        <folder default="true">false</folder>
    </append>
    <authentication default="true">false</authentication>
    <autoplays>
        <after>1</after>
        <before>1</before>
        <enabled>true</enabled>
        <stop>true</stop>
    </autoplays>
    <autoplaywiths>
        <after>1</after>
        <before>1</before>
        <enabled>true</enabled>
        <ffmpeg default="true"></ffmpeg>
        <folder default="true"></folder>
        <player default="true"></player>
        <stop>true</stop>
    </autoplaywiths>
    <autostart default="true">false</autostart>
    <background>
        <notify default="true">false</notify>
        <service default="true">false</service>
        <startup default="true">false</startup>
    </background>
    <cat>
        <order default="true"></order>
    </cat>
    <catchup>
        <channel default="true">false</channel>
        <text default="true">Meta</text>
    </catchup>
    <categories>
        <background.color default="true">[COLOR ff4682b4]steelblue[/COLOR]</background.color>
        <ini.enabled default="true">false</ini.enabled>
        <ini.file default="true"></ini.file>
        <ini.type default="true">0</ini.type>
        <ini.url default="true"></ini.url>
        <remember>true</remember>
    </categories>
    <channel>
        <filter.sort default="true">0</filter.sort>
        <index.digits>3</index.digits>
        <logo default="true">false</logo>
        <shortcut default="true">0</shortcut>
        <shortcut.behaviour>1</shortcut.behaviour>
        <shortcut.direct default="true">false</shortcut.direct>
    </channel>
    <channels>
        <per.page default="true">10</per.page>
    </channels>
    <date>
        <long>true</long>
    </date>
    <down>
        <action default="true">false</action>
    </down>
    <dummy>
        <channels>true</channels>
    </dummy>
    <enable>
        <nextup>true</nextup>
        <nextup.showTime default="true">false</nextup.showTime>
        <osd>true</osd>
    </enable>
    <epg>
        <box.spacing default="true">2</box.spacing>
        <focus.color default="true">[COLOR ff606060]black[/COLOR]</focus.color>
        <font default="true">font13</font>
        <nofocus.color default="true">[COLOR ffffffff]white[/COLOR]</nofocus.color>
        <video.pip default="true">false</video.pip>
    </epg>
    <exit>
        <on.back default="true">false</on.back>
    </exit>
    <favourites default="true">Favourites</favourites>
    <find>
        <program.images default="true">false</program.images>
    </find>
    <fixtures default="true">
        false
        <country default="true">all</country>
        <timezone default="true">None</timezone>
    </fixtures>
    <gz default="true">false</gz>
    <help>
        <invisiblebuttons default="true">false</invisiblebuttons>
    </help>
    <imdb>
        <big>true</big>
        <match>1</match>
    </imdb>
    <last>
        <channel.popup default="true">0</channel.popup>
    </last>
    <listing>
        <days>1</days>
        <hours>6</hours>
        <sort.time>true</sort.time>
    </listing>
    <logos>
        <enabled>true</enabled>
        <folder default="true"></folder>
        <keep default="true">false</keep>
        <source default="true">0</source>
        <url default="true"></url>
    </logos>
    <m3u>
        <read default="true">false</read>
    </m3u>
    <mapping>
        <ini.enabled default="true">false</ini.enabled>
        <ini.file default="true"></ini.file>
        <ini.type default="true">0</ini.type>
        <ini.url default="true"></ini.url>
        <m3u.enabled default="true">false</m3u.enabled>
        <m3u.file default="true"></m3u.file>
        <m3u.type default="true">0</m3u.type>
        <m3u.url default="true"></m3u.url>
    </mapping>
    <md5 default="true">false</md5>
    <menu>
        <addon default="true">false</addon>
    </menu>
    <mine1 default="true">ActivateWindow(Videos,library://video,return)</mine1>
    <mouse>
        <controls>true</controls>
    </mouse>
    <nextup>
        <showTime>10</showTime>
        <time>20</time>
    </nextup>
    <no>
        <programs.message default="true">-</programs.message>
    </no>
    <notifications>
        <enabled>true</enabled>
    </notifications>
    <omdb default="true">false</omdb>
    <password default="true"></password>
    <play>
        <alt.choose default="true">false</alt.choose>
        <alt.continue default="true">false</alt.continue>
        <menu default="true">false</menu>
        <minimized default="true">false</minimized>
    </play>
    <playback>
        <osd.timeout default="true">2</osd.timeout>
        <timeout default="true">5</timeout>
    </playback>
    <program>
        <background.color default="true">[COLOR ff4682b4]steelblue[/COLOR]</background.color>
        <background.enabled default="true">false</background.enabled>
        <background.flat default="true">false</background.flat>
        <background.image default="true"></background.image>
        <background.image.source>2</background.image.source>
        <background.image.url default="true">http://i.imgur.com/yQTvmIA.jpg</background.image.url>
        <background.texture.url default="true">special://profile/addon_data/script.tvguide.fullscreen/backgrounds/sunburst.png</background.texture.url>
        <channel.logo default="true">false</channel.logo>
        <image>true</image>
        <image.scale>true</image.scale>
        <search.plot default="true">false</search.plot>
    </program>
    <redraw>
        <epg>true</epg>
    </redraw>
    <sd>
        <changed default="true">0</changed>
        <interval>2</interval>
        <password default="true"></password>
        <range>3</range>
        <username default="true"></username>
    </sd>
    <search>
        <type default="true">MySearch</type>
    </search>
    <service>
        <addon.folders default="true">false</addon.folders>
        <interval>1</interval>
        <time default="true"></time>
        <type default="true">0</type>
    </service>
    <skin>
        &#x0A;        &#x0A;        Default&#x0A;        &#x0A;        
        <folder default="true"></folder>
        <source default="true">0</source>
        <user default="true"></user>
    </skin>
    <source>
        <source>tvguide.co.uk now</source>
    </source>
    <stop>
        <on.exit>true</on.exit>
    </stop>
    <stream>
        <addon.list default="true">false</addon.list>
    </stream>
    <thelogodb default="true">0</thelogodb>
    <timebar>
        <color default="true">[COLOR ff87ceeb]skyblue[/COLOR]</color>
    </timebar>
    <tvdb>
        <imdb default="true">false</imdb>
        <match>1</match>
    </tvdb>
    <tvguide>
        <co.uk.days>2</co.uk.days>
        <co.uk.email default="true"></co.uk.email>
        <co.uk.systemid default="true">Sky</co.uk.systemid>
    </tvguide>
    <up>
        <cat.mode default="true">First Channel</cat.mode>
    </up>
    <update>
        <progress>true</progress>
    </update>
    <user default="true"></user>
    <version default="true">0.17.349</version>
    <vpnmgr>
        <connect>true</connect>
        <default>true</default>
    </vpnmgr>
    <xmltv>
        <date>true</date>
        <file default="true"></file>
        <interval>2</interval>
        <logos default="true">false</logos>
        <refresh default="true">false</refresh>
        <type default="true">0</type>
        <url default="true"></url>
    </xmltv>
    <xmltv2>
        <enabled default="true">false</enabled>
        <file default="true"></file>
        <type default="true">0</type>
        <url default="true"></url>
    </xmltv2>
</settings>

```

Montellese:
@basrieter 
Kodi doesn't touch the `settings.xml` file your add-on provides. It currently tries to support reading the old format and supports reading the "new" format as also used by Kodi itself. It however overwrites the `settings.xml` in the `addon_data` directory containing all the setting values. But that shouldn't be a problem for you as Kodi v17 will simply write in the old format and Kodi v18+ in the new one.

@primaeval 
* Why is the nesting unwanted? Your add-on shouldn't care one bit about how the `settings.xml` containing the actual setting values looks like. You should still be able to use the setting's identifiers as defined in the your add-on's `settings.xml`. I agree that the format isn't great but that's a backwards compatibility burden we are carrying from the old settings system used by Kodi core.
* As mentioned the issue with InfoBool expressions not working is already confirmed and fixed in my fix branch.
* Did you verify that creating settings on the fly from python doesn't work? Because it worked fine for me when I tested with @ronie's `service.scrobbler.lastfm`.

primaeval:
@Montellese
Now that I've calmed down a bit maybe it's not the end of the world. ;)

Some of the nesting does make sense, such as the xmltv section grouping near the end of the file.
Others such as the tvguide.co.uk settings don't make sense but as you say should be ok from the code.

Now that I have found the right settings by fumbling in the dark to get past the start screen, it looks like setting the values in python is still working. eg
```<version default="true">0.17.349</version>```

Are the System.HasAddon() and RunScript() calls still going to work?

Are you planning on bringing back the labels using direct text values rather than numeric ids?
That is going to cause the most amount of work. 
As with most devs and probably yourself this is a hobby that has to be low priority.
Core changes can cause a lot of extra work that we don't have time for. 
I would probably have given up by now if it weren't for the kind words of some of my users.
I don't even watch much TV. I prefer reading books. :)

Thanks for looking into this.

**Comment by Montellese** [comment]
* `System.HasAddon()` and `RunScript()` will be working again.
* I'm currently working on the label stuff.

IMO early adopters i.e. people using these nightly builds should be used to things not working (perfectly) and you shouldn't take their "screaming" too much to heart. It's great to get the feedback so early but there's still a few months until Kodi v18 will be released so no need to worry already ;-)

**Comment by primaeval** [comment]
Nice. Thanks.

**Comment by Montellese** [comment]
@primaeval already looking a bit better:
screenshot 2017-06-08 21 33 02

**Comment by primaeval** [comment]
:)

**Comment by basrieter** [comment]
@Montellese Thanks for explaining that! So I can choose to either create a new format settings.xml or an old one (for backward compatibility) and let Kodi do the in-memory translation itself?

**Comment by Montellese** [comment]
@basrieter that's correct.

Unfortunately I haven't gotten around to documenting the "new" format yet. The best thing available is Kodi's settings definitions at https://github.com/xbmc/xbmc/blob/master/system/settings/settings.xml which should probably cover almost all available use cases.

**Comment by basrieter** [comment]
Great! I will keep an eye out on your addon_settings_fallout2 branch and do a quick build to see the status. Let me know if you need me to verify/test something specific.

**Comment by AlwinEsch** [comment]
Have a question about things like `<setting id="smooth_duration" type="rangeofnum" label="30006" rangestart="5" rangeend="50" elements="10" valueformat="30050" default="0"/>`.

How is the Kodi alternative?

Parts like them becomes always returned as string what bring big conflicts if the binary addon expect a integer.

**Comment by hudokkow** [comment]
@Montellese, there's a crash caused by this. I bisected but can't pin point the offending commit because they aren't self contained and build fails.
Crashlog -> https://pastebin.com/AwAtQ4C9

Steps to reproduce: go to Settings -> Services -> Control and enable "Allow remote control from applications on other systems"

Thanks @candrews for pointing it out on kodi-dev

**Comment by Montellese** [comment]
@basrieter: Thanks for testing. The following issues should be fixed:
* string labels for categories, groups and settings
* InfoBool conditions for `enable` and `visible`
* settings with a type != `action` and without an identifier are treated as text / labels
* action settings are properly triggered

I'm still looking into the rest. Support for `option="close"` is almost done and then I'll look into the multiple settings with the same identifier problem.

@AlwinEsch: what API are you using? I thought I adjusted the binary add-on interface implementation but I wasn't able to test it. `rangeofnum` settings are handled as `CSettingNumber` so it should return a double or float (whatever the binary add-on interface expects.

@hudokkow Thanks for the hint. I'll look into it. I keep the changes to the settings library and the settings integration separated which can cause builds to fail.

**Comment by Montellese** [comment]
See #12276 for the latest set of fixes (the ones listed above plus the fix for @hudokkow's issue and `option="close"`).

**Comment by Montellese** [comment]
@AlwinEsch I can't reproduce your problem because I can't find an add-on that actually does the `get_setting()` call. They seem to be using the `ADDON_SetSetting()` to ask Kodi for setting values.

**Comment by basrieter** [comment]
Still crashes for me with the latest code and this settings.xml: https://pastebin.com/xWtYSPBH

**Comment by Montellese** [comment]
@basrieter yes because I haven't looked into the multiple settings with the same identifier issue yet.

**Comment by basrieter** [comment]
Ok, NP, will keep tracking your commits

**Comment by primaeval** [comment]
Most of the label problems seem to be ok. Thanks.

There is a new problem with the "fixtures" setting.
It tries to make a top level setting which doesn't get converted properly because there are also subsettings.

kodi.log
```
Starting Kodi (18.0-ALPHA1 Git:20170610-a0a04d6011-di). Platform: Windows NT x86 32-bit
20:00:03.472 T:6632 WARNING: CSettingsManager: unable to read value of setting "fixtures"
20:00:03.472 T:6632 WARNING: CAddonSettings[script.tvguide.fullscreen]: failed to load value "
                                                    false
                                                    " for setting fixtures
```

addon settings.xml
```
        <setting id="fixtures" label="Include Sports Fixtures in xmltv Source" default="false" type="bool"/>
        <setting id="fixtures.country" label="Sports Fixtures Country Code" type="select" default="all" values="all|ai|al|ao|ar|au|az|ba|be|bg|br|ca|ch|cn|cy|cz|de|dk|en|es|fi|fr|gr|hk|hr|hu|id|ie|il|in|ir|is|it|jp|kr|lv|mt|mx|my|nl|no|nz|pl|pt|qa|ro|rs|ru|s2|se|sg|si|sk|th|tr|ua|uk|us|vn|wa|za" />
        <setting id="fixtures.timezone" label="Sports Fixtures Time Zone" type="select" default="None" values="None|-12:00|-11:00|-10:00|-09:00|-08:00|-07:00|-06:00|-05:00|-04:00|-03:00|-02:00|-01:00|+00:00|+01:00|+02:00|+03:00|+04:00|+05:00|+06:00|+07:00|+08:00|+09:00|+10:00|+11:00|+12:00" />

```
addon_data settings.xml
```
    <fixtures default="true">
        false
        <country default="true">all</country>
        <timezone default="true">None</timezone>
    </fixtures>
```

**Comment by primaeval** [comment]
The "skin" settings group has the same problem.
```
    <skin default="true">
        Default
        <folder default="true"></folder>
        <source default="true">0</source>
        <user default="true"></user>
    </skin>
```

**Comment by Stevehans** [comment]
vpn manager from @zomboided crashes on install with the new changes.

**Comment by Montellese** [comment]
@primaeval I've already noticed that as well which is why I have proposed #12277 a few days ago.
@Stevehans the problem is known but I haven't gotten around to fixing it yet.

**Comment by manuelm** [comment]
@Montellese It seems support for file selection is broken. No dialog appears. Path selection works fine. I tried to convert the XML to the core settings format but all I get is a crash in `StringUtils::EqualsNoCase` called from `CGUIControlButtonSetting::GetPath`

XML from my PVR:
```
<setting label="30012" id="favouritesfile" type="file" default="" />
```

**Comment by Montellese** [comment]
@manuelm thanks for the report. I'll look into it.

**Comment by Montellese** [comment]
@manuelm: See #12291 for a fix for the "old" add-on settings. Could you post the XML that you wrote for the new settings format? There is a `StringUtils::EqualsNoCase()` call in `CGUIControlButtonSetting::GetPath()` but I don't see how it could crash.

**Comment by manuelm** [comment]
> Could you post the XML that you wrote for the new settings format?

Sure.
```
<setting id="favouritesfile" type="path" label="30012">
  <level>0</level>
  <default></default>
  <constraints>
    <allowempty>true</allowempty>
  </constraints>
  <control type="button" format="file">
    <heading>657</heading>
  </control>
</setting>
```

results in: `CSettingPath: invalid <control> of "favouritesfile"`, probably because `file` is missing in https://github.com/xbmc/xbmc/blob/master/xbmc/settings/SettingPath.cpp#L58

If i change the baseclass from `type="path"` to `type="string"` I get the above crash. But that might be because `type="string"` is just wrong :-)

**Comment by HitcherUK** [comment]
@Montellese Some oddness with DialogAddonSettings.xml

https://forum.kodi.tv/showthread.php?tid=316560

**Comment by Montellese** [comment]
@HitcherUK: I don't really have any clue about skinning so I simply copied all the controls from another settings dialog XML.
I can look into the Window property though as that probably just got lost when I deleted everything from `CGUIDialogAddonSettings`.

**Comment by DaveTBlake** [comment]
@Montellese it seems that this PR has broken the fetch of settings for the Universal artist and album scrapers, (maybe any content specific video scraper settings too). Scraper settings can be specific to artist and album, the local values are read from the db as xml and then loaded by `CSetingsManager`, however the loading is no longer working. 

To see this issue change the universal scraper settings for an album from "Change Infomation Provider" on the context menu, save them and then revisit the settings dialog, The values are reset to the default.

I have tested by bisection of nightlies and it works prior to 3rd June when this PR was merged, but I don't know enough about settings to see what the issue is. Initialisation of m_loaded perhaps?

**Comment by HitcherUK** [comment]
@Montellese One more thing, the Default Category Button (id=10) isn't fading when the Categories Area grouplist loses focus.

**Comment by HitcherUK** [comment]
@Montellese The above problem and scrollbar problem started with this commit [12245](https://github.com/xbmc/xbmc/pull/12245/commits) - previous to this they worked correctly.

**Comment by Montellese** [comment]
@HitcherUK: that's odd because that PR doesn't touch the GUI presentation at all.

**Comment by HitcherUK** [comment]
I'll try to find the exact day it changed but it's easy enough to test using Estuary by going to any system settings window and noting the category button and scrollbars behaviour before and after.

**Comment by HitcherUK** [comment]
Working fine - `KodiSetup-20170604-49ad3bb2a6-master-x86.exe
79M	2017-Jun-05 04:35`
Broken category button and scrollbars - `KodiSetup-20170606-b196817-master-x86.exe
79M	2017-Jun-07 04:50`

**Comment by Montellese** [comment]
@HitcherUK: it could also be due to #12213 which has been merged between those two builds.

**Comment by DaveTBlake** [comment]
@Montellese any thoughts on how the music scraper settings functionality can be restored? Have you been able to reproduce the issue, or do you need more from me?

**Comment by Montellese** [comment]
@DaveTBlake I'm looking into it now (though with video scrapers because I don't use the music library).

**Comment by DaveTBlake** [comment]
Great. Not using the video lib I wasn't sure if it had the same issue or not, but suspected that it might (if scraper settings are content dependant and settings get fetched from the db). If I can help let me know.

**Comment by Montellese** [comment]
@DaveTBlake can you check whether the two commits in https://github.com/Montellese/xbmc/tree/addon_settings_fix_scrapers solve the problem for you as well?

**Comment by DaveTBlake** [comment]
@Montellese yes, that fixes the music scraper settings issue I was seeing, thanks.

**Comment by peak3d** [comment]
@Montellese if I understand the introduction correctly "old" settings are transformed to "new" settings and after that the new settings should be used for further kodi read ops.

```
11:17:10.727 T:2452   DEBUG: CAddonSettings[plugin.video.netflix]: failed to find definition for setting setting. Creating a setting on-the-fly...
11:17:10.729 T:2452   DEBUG: CAddonSettings[plugin.video.netflix]: loading setting definitions
11:17:10.729 T:2452   DEBUG: CAddonSettings[plugin.video.netflix]: trying to load setting definitions from old format...
11:17:10.732 T:2452   DEBUG: CAddonSettings[plugin.video.netflix]: loading setting values
11:17:10.732 T:2452   DEBUG: CSettingsManager: requested setting (setting) was not found.
```
Tose lines are flooding my log file and it does not seem that the new settings file is written.
Just a misunderstanding? Do I have to tell the addon-dev to change the settings.xml file?

Montellese:
@peak3d the add-on settings system has a legacy parser which reads the old settings XML and turns it into new settings object in the code. But it does not serialize that into a new settings XML file (because we don't have a serializer for that). It also reads the old settings XML values but stores those in the new settings XML value format. So as long as add-on authors don't update their add-on with a new `settings.xml` you will see those debug messages.

DaveTBlake:
@Montellese I noticed that there are still many settings in guisettings.xml in the old format. Is there any plan for these to be updated to v2? Does `CMediaSettings` also need some kind of settings migration?

basrieter:
I still get a lot of warnings on the latest nightly for my add-on and as a result a lot of missing visibility settings: https://pastebin.com/H6cDWW7q

The settings.xml can be found here: https://pastebin.com/dymVsvUt

Any thing I can do to help out with that?

Montellese:
Hey guys, sorry I'm pretty slammed with real life right now and probably for a while still so I can't make any promises when I'll get around to look into this.

@DaveTBlake: Those are settings that are not controlled by the settings system but are manually written by different classes derived from `ISubSettings`. We simply provide them with the XML document / tree and they are free to fill it with whatever they want.

reddit-reaper:
> @Montellese The above problem and scrollbar problem started with this commit [12245](https://github.com/xbmc/xbmc/pull/12245/commits) - previous to this they worked correctly.

just letting you know, looks like this is back. i heard somewhere that if the category has more than 100 elements it will cause this issue. is this true? because after a certain point in this addons settings it just stops and you have to go in reverse to get past it or use the mouse wheel to scroll painfully slow through it

basrieter:
This is true. 100 settings per category is the current limit.

There is a PR that pushes the limit to 512: https://github.com/xbmc/xbmc/pull/14154

PR #12127, opened 2017-05-19
detect and delete zero-byte database files

chewitt:
## Description
This change detects and deletes zero-byte sqlite database files so they don't block the migration process when users upgrade from earlier Kodi versions.

## Motivation and Context
There is growing evidence that issues with Addons27.db are primarily caused by pirate add-ons opening the DB to perform self-enablement at startup, but allowing users to fester in their own self-inflicted pain results in onerous support work; we gain more from the hack.

## How Has This Been Tested?
This code change was tested in Milhouse builds for several weeks before LE elected to detect the problem and delete zero-byte files in a startup script (to avoid a code patch). Most of the Kodi-specific Linux distro's are now doing similar things but the issue is still seen in other Kodi build targets.

## Types of change

## Checklist:


MartijnKaijser:
jenkins build this please

PR #12128, opened 2017-05-19
detect and delete zero-byte database files

chewitt:
forward port of #12127 so we're consistently hacky

### DIFF
--- a/xbmc/dbwrappers/sqlitedataset.cpp
+++ b/xbmc/dbwrappers/sqlitedataset.cpp
@@ -33,10 +33,13 @@
 #include "utils/log.h"
 #include "system.h" // for Sleep(), OutputDebugString() and GetLastError()
 #include "utils/URIUtils.h"
+#include "filesystem/File.h"
 
 #ifdef TARGET_POSIX
 #include "linux/XTimeUtils.h"
 #endif
+
+using namespace XFILE;
 
 namespace dbiplus {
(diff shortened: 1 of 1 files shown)


Paxxi:
Thank you for taking the time to fix and submit this, however I don't really agree with this fix as there's a few issues.
1. It doesn't handle corrupt databases, only empty ones which is only a subset of the scenarios we might encounter.
2. It introduces a dependency from sqlite on our VFS layer which "violates" architecture layering.

I've been thinking about this for a while and my idea so far is
1. actually catch the exception thrown today so we don't crash
2. handle the exception in a higher layer, possibly in dbmanager or in CApplication as this happens during init. Would be possible to do it in two steps, catch the exception in dbmanager, move/delete the broken db and rethrow to signal to CApplication that things went horribly wrong and we should close down and ask user to restart kodi so we're in a known good state.

There are certainly more ways to handle it that I haven't thought of.

Because of the lack of time(from devs) and the wish to get 17.2 shipped this could be acceptable for v17 but I'd rather avoid having it in master.

MartijnKaijser:
I agree with @Paxxi that we should add some catch all for v18 as there are multiple ways to crash. Like with mysql not being able to open a database also causes a crash.

Going to merge it for v17 so we can shove that out.

chewitt:
okay, closing here

PR #12129, opened 2017-05-19
[gui] refactor GUIIncludes class

xhaggi:
Refactor `CGUIIncludes` to be more structured and understandable. There should be no functional changes.

Changes:
* change method signatures of various methods
* move loading/resolving of constants, expressions, defaults etc. to its own methods
* better logging messages in case something goes wrong
* add doxygen comments to various methods

Runtime tested on macOS.

@Paxxi mind taking a look.

### DIFF
--- a/xbmc/guilib/GUIIncludes.cpp
+++ b/xbmc/guilib/GUIIncludes.cpp
@@ -96,7 +96,7 @@ CGUIIncludes::~CGUIIncludes()
 {
 }
 
-void CGUIIncludes::ClearIncludes()
+void CGUIIncludes::Clear()
 {
   m_includes.clear();
   m_defaults.clear();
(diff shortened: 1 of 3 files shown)


xbmc/guilib/GUIIncludes.cpp
```diff
@@ -106,193 +106,280 @@ void CGUIIncludes::ClearIncludes()
...
+bool CGUIIncludes::Load(const TiXmlElement *root)
 {
-  if (!root || strcmpi(root->Value(), "includes"))
+  LoadDefaults(root);
+  LoadConstants(root);
+  LoadExpressions(root);
+  LoadVariables(root);
+  LoadIncludes(root);
+
+  return true;
```
Rechi: function never returns false

xhaggi: Thanks dropped that method completely.

xbmc/guilib/GUIIncludes.cpp
```diff
@@ -106,193 +106,272 @@ void CGUIIncludes::ClearIncludes()
...
+    return;
+  
+  const TiXmlElement* child = node->FirstChildElement("default");
+  while (child)
   {
-    CLog::Log(LOGERROR, "Skin includes must start with the <includes> tag");
-    return false;
+    const char *type = child->Attribute("type");
+    if (type && child->FirstChild())
+      m_defaults.insert(std::pair<std::string, TiXmlElement>(type, *child));
```
Paxxi: use `std::make_pair(type, *child)` as you correctly do below :)

xbmc/guilib/GUIIncludes.cpp
```diff
@@ -106,193 +106,272 @@ void CGUIIncludes::ClearIncludes()
...
+
+void CGUIIncludes::SetDefaults(TiXmlElement *node)
+{
+  if (node->ValueStr() != "control")
+    return;
 
-  node = root->FirstChildElement("expression");
-  while (node)
+  std::string type = XMLUtils::GetAttribute(node, "type");
+  std::map<std::string, TiXmlElement>::const_iterator it = m_defaults.find(type);
```
Paxxi: maybe use `auto`?

Paxxi (commented):
Looks sane to me. Quite a lot of duplication of code for the Load methods but trying to unify them might just make it messy so it's better this way.

Some nitpicks but no blockers.

xhaggi:
@Paxxi addressed your comments

jenkins build this please

xhaggi:
jenkins build this please

PR #12130, opened 2017-05-19
Font addons

notspiff:
Ref http://forum.kodi.tv/showthread.php?tid=314595

Looks like you want this, and I had some spare time so I threw the basics together.


da-anda:
thanks @notspiff Would you also be interested in taking the font resource add-ons further? Like some stuff I mentioned on the forum thread (additional metadata in addon.xml for things like supported charset, and maybe other stuff once we have a worked out a concept)?

akva2:
sure. but others are free to work on it as well, time i have to offer is erradic these days.

da-anda:
tested on Windows and seemed to work fine. Any objections to merge it and fix issues as they show up?
@MartijnKaijser do you want to do a test with font add-on updates/dependencies via repository upfront?

MartijnKaijser:
What needs to be tested is what happens if you receive a font update while running. You can simply test this by creating a zipped font Addon with version bumped and use install from zip.

MartijnKaijser:
Second question is do we want to have Estuary/Estouchy using this addon stright away or only later?

da-anda:
just as a heads up: I haven't forgotten about doing further tests for this PR - just didn't find the time yet due to stress in real life.

jjd-uk:
For Martijn's second question on Estuary/Estouchy I guess we might need feedback from @ronie @phil65

phil65:
I'm fine with using it straight away.

Rechi:
@notspiff this needs a rebase

notspiff:
rebased. not really runtime tested, it builds and stuff shows in the browser is all i have time for.

notspiff:
properly runtime tested. works just fine. both skins updated, and the new roboto font split to an add-on as well. probably want to do those changes outside mainline though.

notspiff:
@MartijnKaijser removed the commit as requested. if somebody could quickly look at code @xhaggi @Rechi @Paxxi it is ready as a starting point afaik.

xbmc/addons/FontResource.cpp
```diff
@@ -0,0 +1,87 @@
...
+    CApplicationMessenger::GetInstance().PostMsg(TMSG_EXECUTE_BUILT_IN, -1, -1, nullptr, "ReloadSkin");
+  }
+}
+
+bool CFontResource::HasFont(std::string& path, const std::string& file) const
+{
+#ifdef TARGET_POSIX
+  std::string result = CSpecialProtocol::TranslatePathConvertCase(Path()+"/resources/"+file);
+#else
+  std::string result = Path()+"/resources/"+file;
```
Rechi: ifdef not needed here, it is already handled in `CSpecialProtocol::TranslatePathConvertCase` function

notspiff: apparently it wasn't back in may as this code was copy-pasted ;) removed.

xbmc/guilib/GUIFontManager.cpp
```diff
@@ -121,7 +126,17 @@ CGUIFont* GUIFontManager::LoadTTF(const std::string& strFontName, const std::str
...
   if (!CheckFont(strPath,"special://home/media/Fonts",file))
-    CheckFont(strPath,"special://xbmc/media/Fonts",file);
+    if (!CheckFont(strPath,"special://xbmc/media/Fonts",file))
+    {
+      VECADDONS addons;
+      CServiceBroker::GetAddonMgr().GetAddons(addons, ADDON_RESOURCE_FONT);
+      for (auto& it : addons)
+      {
+        std::shared_ptr<CFontResource> font(std::static_pointer_cast<CFontResource>(it));
+        if (font->HasFont(strPath,file))
```
xhaggi: Spaces between parameters, here and everywhere pls

xbmc/addons/FontResource.h
```diff
@@ -0,0 +1,51 @@
...
+  virtual bool IsInUse() const;
+
+  //! \brief Check whether file is allowed or not (no filters here).
+  bool IsAllowed(const std::string& file) const override { return true; }
+
+  //! \brief Check whether add-on has font
+  //! \param[out] path Full path to font if found.
+  //! \param[in] file File name of font.
+  //! \return True if font was found, false otherwise.
+  bool HasFont(std::string& path, const std::string& file) const;
```
xhaggi: `GetFont` and switch out and in parameter? IMO it gets the font path if font file is served by the addon.

notspiff: you want
```
GetFont(const std::string& file, std::string& path)
```
?

notspiff: i guess that's what you wanted. updated like this, i can change it to whatever you want if i misunderstood.

xhaggi: Yep that's what I meant and pls adjust the method doc to `Get the font path if give font file is served by the addon.` or something like that

notspiff: done.

xhaggi: Thx ;)

xhaggi (approved):

Rechi:
wrong revision was build
jenkins build this please

xbmc/addons/FontResource.cpp
```diff
@@ -0,0 +1,83 @@
...
+    if (it.first == ID())
+      return true;
+
+  return false;
+}
+
+void CFontResource::OnPostInstall(bool update, bool modal)
+{
+  if (IsInUse() || (!update && !modal &&
+    HELPERS::ShowYesNoDialogText(CVariant{Name()}, CVariant{24099}) == DialogResponse::YES))
```
tamland: this makes zero sense..

tamland:
Correct me if I'm wrong but disabling, updating and uninstalling of font addons does not seem to be handled in any way in FontManager here

notspiff:
FontManager is updated through a skin reload, ref https://github.com/xbmc/xbmc/blob/master/xbmc/Application.cpp#L1736 and https://github.com/xbmc/xbmc/blob/master/xbmc/Application.cpp#L1637

as these are dependencies of skins, i thought the generic handling would cover uninstall? i.e. not allowed to uninstall unless skin is gone etc, and in that case it wouldn't be loaded. the dialog should perhaps have a better message, suggestions welcome.

tamland:
As long as there's no stale cache of uninstalled/disabled addons in FontManager that somehow can be used, it's fine.

> the dialog should perhaps have a better message, suggestions welcome.

If they are strictly dependencies of skins tho, then there is no point having any dialogs or skin reloading, as they will always be present before the skin is

xbmc/addons/FontResource.cpp
```diff
@@ -0,0 +1,83 @@
...
+bool CFontResource::IsInUse() const
+{
+  std::string skin = CServiceBroker::GetSettings().GetString(CSettings::SETTING_LOOKANDFEEL_SKIN);
+  ADDONDEPS deps = CServiceBroker::GetAddonMgr().GetDepsRecursive(skin);
+  for (const auto& it : deps)
+    if (it.first == ID())
+      return true;
+
+  return false;
+}
```
tamland: Also, this method is redundant. Dependency check is already done elsewhere for all addons

notspiff: i do not understand how this can be redundant, i need to check if the font is in use on update, and reload skin if it is. there are glyph caches and such in memory that has to be redone, and the only way to do that right now is to reload the skin.

notspiff: right, i see where the confusion comes from now. i have been a little bit too eager when i c&p'd the code from the skin, and that dialog shouldn't be there at all. i thought it was something different, not the want to switch dialog.

tamland: Nah, my confusion was from IsInUse being implemented.. But I get it now, so can you please move this into OnPostInstall instead then? Because `IsInUse` is used by external stuff which means the check will done multiple times. (and IsInUse should really be removed in general)

notspiff: done.

xbmc/guilib/GUIFontManager.cpp
```diff
@@ -120,8 +125,18 @@ CGUIFont* GUIFontManager::LoadTTF(const std::string& strFontName, const std::str
 
   // Check if the file exists, otherwise try loading it from the global media dir
   std::string file = URIUtils::GetFileName(strFilename);
-  if (!CheckFont(strPath,"special://home/media/Fonts",file))
-    CheckFont(strPath,"special://xbmc/media/Fonts",file);
+  if (!CheckFont(strPath, "special://home/media/Fonts", file))
+    if (!CheckFont(strPath, "special://xbmc/media/Fonts", file))
```
xhaggi: Missing bracket?

xhaggi: Btw combine both of them into one if statement

notspiff: if you say so. it's hard to remember which project has which brace rules, at work i am not allowed to use those braces in this case...

notspiff: braced, combined.

xhaggi: non-single lines of code should be surrounded by braces if you asked me, it's better for readability.

xbmc/addons/FontResource.h
```diff
@@ -0,0 +1,48 @@
...
+  bool IsAllowed(const std::string& file) const override { return true; }
+
+  //! \brief Get the font path if given font file is served by the add-on.
+  //! \param[in] file File name of font.
+  //! \param[out] path Full path to font if found.
+  //! \return True if font was found, false otherwise.
+  bool GetFont(const std::string& file, std::string& path) const;
+
+  //! \brief Callback executed after installation
+  virtual void OnPostInstall(bool update, bool modal);
```
Rechi: `override` missing

xbmc/addons/AddonInfo.cpp
```diff
@@ -71,6 +71,7 @@ static const TypeMapping types[] =
    {"kodi.resource.language",            ADDON_RESOURCE_LANGUAGE,   24026, "DefaultAddonLanguage.png" },
    {"kodi.resource.uisounds",            ADDON_RESOURCE_UISOUNDS,   24006, "DefaultAddonUISounds.png" },
    {"kodi.resource.games",               ADDON_RESOURCE_GAMES,      35209, "DefaultAddonGame.png" },
+   {"kodi.resource.font",                ADDON_RESOURCE_FONT,       13303, "DefaultAddonFont.png" },
```
Rechi: file has to be added to skins

MartijnKaijser: @ronie @phil65 could you come up with some icon?

MartijnKaijser:
jenkins build this please

tamland:
The install hook still not fixed here fyi

xbmc/addons/FontResource.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+std::unique_ptr<CFontResource> CFontResource::FromExtension(CAddonInfo addonInfo, const cp_extension_t* ext)
+{
+  return std::unique_ptr<CFontResource>(new CFontResource(std::move(addonInfo)));
+}
+
+bool CFontResource::IsInUse() const
+{
+  std::string skin = CServiceBroker::GetSettings().GetString(CSettings::SETTING_LOOKANDFEEL_SKIN);
+  const auto& deps = CServiceBroker::GetAddonMgr().GetDepsRecursive(skin);
```
tamland: Unless there's some specific need, there is no need to implement this. Dependencies are enforced elsewhere and common to everything.

notspiff:
thanks for heads-up i thought i had taken care of it.

notspiff:
i have no idea what github was up to here. i had removed the dialog in the install hook, i had removed isinuse. the branch from my remote held the changes, but somehow the pr did not reflect the updated branch.

notspiff:
jenkins build this please

MartijnKaijser:
@tamland any last remarks so we can merge this?

tamland:
Is there something screwed up with github? Because I'm still seeing the old version here..

notspiff:
it's been a complete cunt in this pr. it does however show the new version for me now.

tamland:
It's reflecting what's here: https://github.com/notspiff/xbmc/commits/font_addons so I don't thinks there's anything wrong with the PR. Maybe just push to new branch?

notspiff:
nope. it's back again! the old version. even in my branch. ffs.

notspiff:
i have pushed up a new version at least 3 times. it's NOT reflecting what i push in there. i'm trying from a new branch.

MartijnKaijser:
Replaced by https://github.com/xbmc/xbmc/pull/13515

pannal:
Sorry for digging up this ancient PR.

This is working well. Is there a reason font addons don't supply fonts to the subtitle font selection in Kodi?

PR #12132, opened 2017-05-19
[cmake] fix addon depends check if headers are in separate folders

AlwinEsch:
## Description
Is to handle the cases where header is in different folders, e.g. the coming "#include <kodi/gui/DialogOK.h>"

## Types of change

## Checklist:

### DIFF
--- a/cmake/scripts/common/AddonHelpers.cmake
+++ b/cmake/scripts/common/AddonHelpers.cmake
@@ -93,7 +93,7 @@ macro (build_addon target prefix libs)
       if(loop_var MATCHES "#define ADDON_")
         string(REGEX REPLACE "\\\n" " " loop_var ${loop_var}) # remove header line breaks 
         string(REGEX REPLACE "#define " "" loop_var ${loop_var}) # remove the #define name from string
-        string(REGEX MATCHALL "[a-zA-Z0-9._]+" loop_var "${loop_var}") # separate the define values to a list
+        string(REGEX MATCHALL "[//a-zA-Z0-9._-]+" loop_var "${loop_var}") # separate the define values to a list
 
         # Get the definition name
         list(GET loop_var 0 include_name)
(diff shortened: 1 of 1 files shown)


AlwinEsch:
jenkins build this please

cmake/scripts/common/AddonHelpers.cmake
```diff
@@ -117,7 +117,7 @@ macro (build_addon target prefix libs)
                 if("${matchres}" EQUAL 0)
                   string(REPLACE " " ";" loop_var "${loop_var}")
                   list(GET loop_var 1 include_name)
-                  string(REGEX REPLACE "[<>\"]" "" include_name "${include_name}")
+                  string(REGEX REPLACE "[<>\"]|kodi/" "" include_name "${include_name}")
```
da-anda: do we have some constant/variable that contains the app name? I'm just thinking about making it easier for people to rebrand if needed.

AlwinEsch: :-D
I'm mostly want to have it as '#include <kodi/...>' but some use it as local headers with `#include "..."` and to check depends also on them ;-)

AchimTuran: @da-anda
So far I know the code that you commented isn't related to UI stuff. Right @AlwinEsch?

da-anda: rebranding is not limited to the UI. Also things like useragent, installation folder, app name etc need to change in code as well. So if a string is related f.e. to the installation folder, ideally some cmake wide variable should be used for that. If the string is however related to extension points that likely won't change on rebrands, hardcoding should be fine.
I was only bringing this up because we had a related discussion at DevCon if you guys remember.

da-anda: to clarify - if the "kodi" folder is part of the file tree from add-ons and is not dynamically generated/referenced from the app name in some other place, hardcoding it is fine. But as I have no idea of cmake I thought I just mention it because of the DevCon discussion we had. I didn't mean to block merging this PR.

AchimTuran (commented):
@da-anda
Thanks for sharing points. So far I understand @AlwinEsch's PR this is only for the include path during development time.
It might be better to create a new forums thread about that topic. I guess you will reach more developers then in this PR. ;)

da-anda:
sure, I only mentioned it because this PR added a hardcoded reference to "kodi" and I wasn't sure if it's related to the app name or not.

AchimTuran:
So then all is fine.

jenkins build addons please

AchimTuran:
jenkins build addons please

AlwinEsch:
jenkins build addons please

AlwinEsch:
jenkins build this please

AlwinEsch:
jenkins is currently not configured for the last changes to build addons!

PR #12133, opened 2017-05-19
[game] change to new addon to kodi callback way

AlwinEsch:
This is the first major change to become multiple instances to work.

There are the callbacks moved to the GAME::CGameClient class. Reason is
to have this parts together and later with multiple instances to call
always right related GAME::CGameClient class.

On addon is currently nothing needed to change, only a recompile to
match them.

The last request to move addon instance callbacks to his related Kodi class.

The other three for global callbacks comes later.

## Types of change

## Checklist:


xbmc/games/addons/GameClientProperties.h
```diff
@@ -77,7 +77,7 @@ class CGameClientProperties
   bool HasProxyDll(const std::string& strLibPath) const;
 
   const CGameClient* const  m_parent;
-  game_client_properties    m_properties;
+  AddonProps_Game&          m_properties;
```
AlwinEsch: This is one of the biggest changes where the pointer becomes moved there and the static is now in `CGameClient`, during test here has it worked with the addon properties on the classes.

@garbear are the changes here and on this request OK for you? During test here at home has it work. Due to a renamed structure (to match style of other types) need also the game addon a update

ksooo (commented):
despite from the open question to @garbear, which I cannot comment on, the changes are looking good to me.

xbmc/games/addons/GameClientProperties.h
garbear: This var is a hack from early in game development, entirely to avoid modifying CAddonDll to keep the size of the patch down. I'd be glad to see it go, so +1 from me.

ksooo:
jenkins build this please

MartijnKaijser:
jenkins build this please

PR #12134, opened 2017-05-19
[cmake] platform changes

lrusak:
Would like to get the ball rolling here. Thanks to @fetzerch for the initial work.

This only affects linux, this _shouldn't_ change anything for the other platforms.

Basically this allows for us to set required and optional dependencies **per** platform (x11, mir, wayland, aml, rpi, imx, gbm, etc).

You will now need to specify `CORE_PLATFORM_NAME=Mir` if you want to build for mir or similar for other platforms.
If `CORE_PLATFORM_NAME` isn't specified it will default to X11.

I've added temp commits for RPI and GBM, I will have to move these to the rpi cleanup PR and the drm/kms PR's.

I've built this successfully using GBM, I will try with the other platforms.


lrusak:
jenkins built this please

lrusak:
jenkins build this please

BrandonSchaefer:
A couple issues, one PkgConfig isnt checked until after the platform probing seems to happen. So for Mir it fails to have pkg config found so it fails to find Mir.

The other issue is, if set CORE=Mir and Mir is not found its more then happy to produce a Makefile without any core platform really found. For probing a platform it should be required to find it.

Otherwise this looks good. Easier to read (for new platforms) and cleans up some logic :)

cmake/scripts/common/Macros.cmake
```diff
@@ -499,7 +499,7 @@ macro(core_add_optional_subdirs_from_filelist pattern)
       list(GET line 1 subdir_dest)
       list(GET line 3 opts)
       foreach(opt ${opts})
-        if(ENABLE_${opt})
+        if(ENABLE_${opt} OR ${opt}_FOUND)
```
wsnipex: this looks wrong. It will cause existing  but disabled stuff to be used. Maybe AND was intended here?

cmake/scripts/linux/platform/X11.cmake
```diff
@@ -0,0 +1,8 @@
+core_require_dep(OpenGl)
+core_require_dep(EGL)
+core_require_dep(X)
+core_require_dep(XRandR)
+core_require_dep(LibDRM)
+core_optional_dep(VAAPI)
+core_optional_dep(VDPAU)
+core_optional_dep(OpenGLES)
```
wsnipex: can we replace those with 2 vars, like in main CMakeLists.txt?
e.g. PLATFORM_REQUIRED_DEPS and PLATFORM_OPTIONAL_DEPS?

wsnipex (commented):
looks good part from some minors

FernetMenta (approved):
The concept looks great to me. I am no expert in that domain and can't judge the details.

cmake/scripts/common/Macros.cmake
lrusak: maybe @fetzerch can comment, but I agree `AND` is better here

cmake/scripts/linux/platform/X11.cmake
lrusak: I'm not sure what you mean exactly. Can you clarify please?

lrusak:
@BrandonSchaefer I've moved the relevant code to after the pkgconfig check. Please test again.

I've also only made this for linux for now as it is the only thing relevant

lrusak:
jenkins build this please

BrandonSchaefer:
Nice found a bug in GLES for Mir. Forgot to rename m_surface -> m_window. Opps.

Yup works for me now, reason is Mir puts its headers in /usr/include/mirclient vs /usr/include soo it needs the *.pc files to actually find them.

Though Mir wont build (due to doing vaapi=1 and vdpau=0). Ill have to check out whats changed in the cdecorder. Not this branches issue.

Yeah I was mainly thinking when wayland support comes this will be things simpler (cmake wise).  So +1 on Mir side of things here.

cmake/scripts/linux/platform/X11.cmake
hudokkow: @wsnipex means something like https://github.com/xbmc/xbmc/blob/master/CMakeLists.txt#L103 and https://github.com/xbmc/xbmc/blob/master/CMakeLists.txt#L127

Not a blocker imo.

I've been playing with @fetzerch's work for the last 2 days. Not enough time to finish it, though. I'm very glad you managed to do it!

A big :+1: from me!

CMakeLists.txt
```diff
@@ -213,42 +214,6 @@ if(CORE_SYSTEM_NAME STREQUAL osx)
   core_require_dep(Sdl)
 endif()
 
-if(NOT WIN32)
```
fetzerch: This whole block is not linux only, so for the other platforms we either need to put these dependencies in the ArchSetup files, or also create platform files for them. If there's a change that we will have more platforms also for non-linux then the later is probably preferable. (Maybe @FernetMenta / @Memphiz / @Paxxi can give input about how likely that is).

cmake/scripts/common/Macros.cmake
fetzerch: Hm, can it be that optional dependencies didn't have ENABLE_ set? I'd just remove it and see if it builds.

cmake/scripts/linux/platform/X11.cmake
fetzerch: core_optional/require_dep would just have to loop over the parameters. Maybe that would allow us in the main cmake file to not use extra vars (optional/required_deps), but to directly use the functions. Could be done independently though.

CMakeLists.txt
FernetMenta: as far as I can tell this block was only for X11 and Mir. this change looks ok to me

Memphiz:
That looks nice :)

cmake/scripts/linux/platform/X11.cmake
wsnipex: see https://github.com/lrusak/xbmc/pull/1

CMakeLists.txt
fetzerch: It did also optionally enable GL/GLES for osx/ios/droid.

FernetMenta: right but not obvious. good that it is cleaned up :)

Paxxi: dropping this looks good to me as WindowsStore isn't WIN32 afaik.

lrusak:
This builds again, though the last commit is kind of a hack. Basically the problem it solves is the when enabling an optional dep (say mir, or gbm) but it is set in the `CORE_PLATFORM_NAME=GBM` then the option `ENABLE_GBM` won't be set and the optional cmake directory won't be included. So I have added this to automatically add `ENABLE_GBM` to the build, I'm not sure if there is a better solution or not.

lrusak:
jenkins build this please

cmake/scripts/linux/platform/X11.cmake
```diff
@@ -0,0 +1,2 @@
+set(PLATFORM_REQUIRED_DEPS OpenGl EGL X XRandR LibDRM)
```
FernetMenta: glx is required too

cmake/scripts/common/Macros.cmake
```diff
@@ -333,6 +333,7 @@ endmacro()
 #   dependency added to ${SYSTEM_INCLUDES}, ${DEPLIBS} and ${DEP_DEFINES}
 function(core_require_dep)
   foreach(dep ${ARGN})
+    setup_enable_switch()
```
wsnipex: required deps do not have an enable_switch.
If you cannot change relevant code to use FOO_FOUND instead, set ENABLE_FOO in the platform file.

cmake/scripts/linux/platform/X11.cmake
lrusak: hmm we have never explicitly set glx as a dependency anywhere I can find. Maybe we need to add a FindGLX.cmake module?

lrusak:
jenkins build this please

cmake/scripts/common/Macros.cmake
```diff
@@ -327,37 +327,43 @@ endmacro()
...
+    export_dep()
+    set(${depup}_SONAME ${${depup}_SONAME} PARENT_SCOPE)
+  endforeach()
 endfunction()
 
 # helper macro for optional deps
 macro(setup_enable_switch)
   string(TOUPPER ${dep} depup)
-  if(ARGV1)
+  if(${ARGV1})
```
wsnipex: why?

lrusak: see, https://cmake.org/cmake/help/v3.0/command/macro.html#macro-argument-caveats

lrusak:
jenkins build this please

cmake/scripts/linux/platform/X11.cmake
FernetMenta: most people install libgl-mesa-dev which includes libgl1-mesa-glx. that may be the reason the issue has not come up yet.
not sure if a proper solution is worth the effort. maybe just put a comment in the code that states that glx is a package by itself

lrusak:
jenkins build this please

lrusak:
jenkins build this please

lrusak:
This now builds on all platforms

basically the ENABLE_* is set for all required deps now because of how our tree data paths work. Otherwise we would need something like, https://github.com/lrusak/xbmc/commit/7e3a4d5e2d1b03c0103ee7af1cd6a87270e7a17b which I'm not sure if that's really any better.

lrusak:
Would like to get some more comments on this please. This is holding up some other work so I'd like to get it in ASAP.

@BrandonSchaefer does this work for mir in it's current state?

Any oobjections?

Memphiz:
looks ok for darwin at a glance

BrandonSchaefer:
Yup looks good for mir

lrusak:
I consider this good to go. Please have one last look before squashing

@Paxxi @Rechi currently the windows.cmake platform file is empty. Would you rather me remove it or just have it there as a placeholder?

lrusak:
jenkins build this please

lrusak:
Jenkins is happy, Just waiting to hear from the windows guys then I'll squash and merge

MartijnKaijser:
I'd say keep the empty file

lrusak:
pushed a commit to leave a comment in the blank file

Rechi:
@lrusak windows currently doesn't need any platform specific dependencies.
If it is easier just keep the empty `windows.cmake` file.

lrusak:
jenkins build and merge

asavah:
This has broken disabling optional components (like libsnfs,libplist)
it doesn't matter what I do
```
 -DENABLE_PLIST=OFF 
``
Which worked before now fails miserably.
```
-- Could NOT find Plist (missing:  PLIST_LIBRARY PLIST_INCLUDE_DIR)
CMake Error at cmake/scripts/common/Macros.cmake:431 (message):
  PLIST enabled but not found
  CMakeLists.txt:188 (core_optional_dyload_dep)


```

Let's try with libnfs (which is present)
 -DENABLE_NFS=OFF 
```
-- Found Bluray: /home/asavah/kross/build/rpi3/rootfs/usr/lib/libbluray.so (found version "0.9.3")
-- Found CEC: /home/asavah/kross/build/rpi3/rootfs/usr/lib/libcec.so (found suitable version "4.0.2", minimum required is "4.0.0")
CMake Error at cmake/scripts/common/Macros.cmake:431 (message):
  NFS enabled but not found
  CMakeLists.txt:188 (core_optional_dyload_dep)


```

Nonsense ...

asavah:
Found the cause (it's trivial) will submit a PR.

PR #12137, opened 2017-05-20
[python] fix ListItem being forced to be a video item after 27dc9a69

Montellese:
## Description
This should fix the issues reported by @notspiff in #11998 by introducing a `ListItem::GetVideoInfoTag()` wrapper which resets the video playcount if and only if the plugin accesses the video properties of the `ListItem`.

## How Has This Been Tested?
I haven't tested this myself. @faush01 could you give this a try please?
## Types of change

## Checklist:

### DIFF
--- a/xbmc/interfaces/legacy/ListItem.cpp
+++ b/xbmc/interfaces/legacy/ListItem.cpp
@@ -63,8 +63,6 @@ namespace XBMCAddon
         item->SetArt("thumb",  thumbnailImage );
       if (!path.empty())
         item->SetPath(path);
-
-      item->GetVideoInfoTag()->ResetPlayCount();
     }
 
     ListItem::~ListItem()
@@ -158,7 +156,7 @@ namespace XBMCAddon
       if (!item) return;
 
       LOCKGUIIF(m_offscreen);
-      CVideoInfoTag& vtag = *item->GetVideoInfoTag();
+      CVideoInfoTag& vtag = *GetVideoInfoTag();
       for (const auto& it : dictionary)
         vtag.SetUniqueID(it.second, it.first, it.first == defaultrating);
     }
@@ -168,7 +166,7 @@ namespace XBMCAddon
       if (!item) return;
 
       LOCKGUIIF(m_offscreen);
-      item->GetVideoInfoTag()->SetRating(rating, votes, type, defaultt);
+      GetVideoInfoTag()->SetRating(rating, votes, type, defaultt);
     }
 
     void ListItem::addSeason(int number, std::string name /* = "" */)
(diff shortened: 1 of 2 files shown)


xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -862,5 +860,19 @@ namespace XBMCAddon
       }
       return els;
     }
+
+    CVideoInfoTag* ListItem::GetVideoInfoTag()
+    {
+      // make sure the playcount is reset to -1
+      if (!item->HasVideoInfoTag())
+        GetVideoInfoTag()->ResetPlayCount();
```
xhaggi: A circular call?

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -862,5 +860,19 @@ namespace XBMCAddon
...
+
+      return GetVideoInfoTag();
```
xhaggi: same here

xbmc/interfaces/legacy/ListItem.cpp
Montellese: Ha, that's what you get when you do search & replace :-/

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -862,5 +860,19 @@ namespace XBMCAddon
...
+      // make sure the playcount is reset to -1
+      if (!item->HasVideoInfoTag())
+        item->GetVideoInfoTag()->ResetPlayCount();
+
+      return item->GetVideoInfoTag();
+    }
+
+    const CVideoInfoTag* ListItem::GetVideoInfoTag() const
+    {
+      return item->GetVideoInfoTag();
```
xhaggi: Should we call `ListItem::GetVideoInfoTag` or shouldn't we?

Montellese: No this is the `const` version so we can't call into the non-`const` version. `CFileItem::GetVideoInfoTag() const` simply returns `m_videoInfoTag` independent of whether it's `nullptr` or not.

xhaggi: Okay, thanks for explanation.

faush01:
Just got home, I will test this now.

faush01:
I have tested this and it works as intended for me.
When the addon sets the playcount that is the value that is used instead of the one from the DB. 
This looks good.

faush01:
Let me know if you want anything else tested or done.

MartijnKaijser:
jenkins build this please

MartijnKaijser:
jenkins build this please

MartijnKaijser:
jenkins build this please

Rechi:
random failure
jenkins build this please

Rechi:
jenkins build this please

MartijnKaijser:
jenkins build this please

PR #12138, opened 2017-05-20
[xbmc][cmake][win32] Use PCH as intended to speed up build

Paxxi:
PCH was being built in a separate library to work around the lack of includes in certain parts of other libraries.
This only complicated the build and wasn't really the intended use of a pch which is to speed up the build process.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/interfaces/swig/AddonModuleXbmc.i
+++ b/xbmc/interfaces/swig/AddonModuleXbmc.i
@@ -21,6 +21,13 @@
 %module(directors="1") xbmc
 
 %{
+#if defined(TARGET_WINDOWS) || defined(TARGET_WIN10)
+#  if !defined(WIN32_LEAN_AND_MEAN)
+#    define WIN32_LEAN_AND_MEAN
+#  endif
+#  include <windows.h>
+#endif
+
 #include "interfaces/legacy/Player.h"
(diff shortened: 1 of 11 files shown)


xbmc/interfaces/swig/AddonModuleXbmc.i
```diff
@@ -21,6 +21,13 @@
 %module(directors="1") xbmc
 
 %{
+#if defined(TARGET_WINDOWS) || defined(TARGET_WIN10)
+#  if !defined(WIN32_LEAN_AND_MEAN)
+#    define WIN32_LEAN_AND_MEAN
+#  endif
+#  include <windows.h>
+#endif
```
Rechi: why do we need to include `windows.h` here and in the other files?

Rechi (approved):

xbmc/interfaces/swig/AddonModuleXbmc.i
Paxxi: for filesystem, Sleep, sockets. Without the include it breaks with missing bit or redefinition because a struct gets defined here before the proper one in windows.h

Paxxi:

Rechi:
jenkins build and merge

PR #12145, opened 2017-05-21
SMBFile: Clean Up

fritsch:
See: https://github.com/kevmitch/mpv/blob/read_chunk/TOOLS/wlan_run_1.eth_run_2.eth_run_1.wlan_run_2.png

### DIFF
--- a/xbmc/filesystem/SMBFile.cpp
+++ b/xbmc/filesystem/SMBFile.cpp
@@ -503,15 +503,6 @@ ssize_t CSMBFile::Read(void *lpBuf, size_t uiBufSize)
 
   CSingleLock lock(smb); // Init not called since it has to be "inited" by now
   smb.SetActivityTime();
-  /* work around stupid bug in samba */
-  /* some samba servers has a bug in it where the */
-  /* 17th bit will be ignored in a request of data */
-  /* this can lead to a very small return of data */
-  /* also worse, a request of exactly 64k will return */
-  /* as if eof, client has a workaround for windows */
-  /* thou it seems other servers are affected too */
-  if( uiBufSize >= 64*1024-2 )
-    uiBufSize = 64*1024-2;
 
   ssize_t bytesRead = smbc_read(m_fd, lpBuf, (int)uiBufSize);
 
--- a/xbmc/filesystem/SMBFile.cpp
+++ b/xbmc/filesystem/SMBFile.cpp
@@ -118,7 +118,6 @@ void CSMB::Init()
         fprintf(f, "\tclient lanman auth = yes\n");
         fprintf(f, "\tlanman auth = yes\n");
 
-        fprintf(f, "\tsocket options = TCP_NODELAY IPTOS_LOWDELAY SO_RCVBUF=65536 SO_SNDBUF=65536\n");      
         fprintf(f, "\tlock directory = %s/.smb/\n", getenv("HOME"));
 
         // set wins server if there's one. name resolve order defaults to 'lmhosts host wins bcast'.


fritsch:
Removed a year old workaround: https://bugzilla.samba.org/show_bug.cgi?id=3668 which "should be fixed" in recent version. At least for my testing everything is fine.

Removing the workaround:
Before: 8 MB/s After: 15 MB/s

Increasing the rcbuf and sendbuf to 1 MB
Before: 15 MB/s After: 33 MB/s (my wireless bridge max speed)

fritsch:
@chewitt @MilhouseVH we should sync the other patches somehow. My testing was done with kodi's filemanager by copying 2 GB from the smb share to my local ssd.

popcornmix:
@fritsch the graph seems to show that 256K has almost all the performance benefit and may be a safer default.

fritsch:
@popcornmix did you test with the second patch applied? There is an old workaround that limits to 64 K - can you remove that?

popcornmix:
@fritsch no I'm just going by the graph you posted.
But I have cherry picked this into my tree to get some testing in nightly builds.
I agree the hack should be dropped if no one is sure whether it is still needed as it will cause some performance loss.

fritsch:
Btw. removing the socket option completely:
Before: 33 MB/s After: 47 MB/s

Will update the PR and wait on Milhouse's figures.

fritsch:
``` socket options = TCP_NODELAY ``` is default since samba 2
``` IPTOS_LOWDELAY ``` is enabled by default whenever TCP_NODELAY is set

Let's keep samba's defaults.

**Comment by MilhouseVH** [comment]
Slightly different results on LE: http://sprunge.us/QgNG

Removing socket options: no difference.
Switching between NT1/SMB2/SMB3: no difference
Increasing chunk size to 1MB: no difference
Removing 10 year workaround: 50-60% improvement

**Comment by fritsch** [comment]
Some results from my laptop to NAS (smb.conf removed):
```
6.8 MB /s defaults
1M chunks 9.4 MB/s
years old workaround removed + 1 MB chunks: 16 MB /s
years old workaround removed + 1 Byte chunks: 12.6 MB /s
```

So from my testing I'd like to keep the workaround removed and the 1M chunks.

**Comment by MrMC** [comment]
FYI: https://github.com/MrMC/mrmc/commit/becb14d3b28591a9bc9227120e775c83e81fe2d9

**Comment by fritsch** [comment]
@MrMC thx for heads up - do you have a debuglog that shows the output? It's printing on every read right?

**Comment by MrMC** [comment]
Not every read, only when the read size chunk increases. Keeps from spamming log.

GetChunkSize = 1
Read - max read bytes = 131072

GetChunkSize = 1024 * 1024
Read - max read bytes = 1048576

But with GetChunkSize = 1M, bluray ISO's open much slower.

20-25MB/s using either, over 802.11ac under OSX.

For comparison, libdsm performance a little slower (18-20MB/s) using a 62k payload size. SMB packet size has limits so what you are seeing in libsmbclient is internal async threading tricks to speed up transfers.

**Comment by fritsch** [comment]
@MrMC: Which chunksize do you use? I see direct linear correlation to my raw throughput tests with the chunksize (after workaround removed).

Can you reproduce the ISO issue locally?

**Comment by MrMC** [comment]
For SMB, we use GetChunkSize=1. 

Not sure what you mean by 'Can you reproduce the ISO issue locally'. Local disk reads are very fast and this tends to mask issues with transfer and/or caching.

**Comment by fritsch** [comment]
Nope, I wanted to know: Is this iso issue some mistery someone reported on the forums or did you benchmark that yourself, too? If yes, we should find a chunksize fine for that isos and fine for consecutive reads.

**Comment by MrMC** [comment]
Under my testing. Two SMB servers, QNAP (quad atom) and my mighty linux/i7. We were going to bump up smb chunksize to 512k around the time of that commit but testing showed the issue with bluray ISOs so we left it at GetChunkSize=1. 256k might be ok, as 1 resulted in 128k (from my SMB servers)

**Comment by fritsch** [comment]
I'll wait for @MilhouseVH nightly tests and if we find issues I'll change it to 256 KByte - I currently have not bluray isos arround and therefore cannot contribute to testing.

**Comment by da-anda** [comment]
I could test on a native bluray via network share (SMB) if that is supposed to show the same issue as an ISO

**Comment by MrMC** [comment]
Has to be bluray ISO. Raw blurays are fine.

**Comment by fritsch** [comment]
This one is fine? http://download.blender.org/durian/movies/Sintel-Bluray.iso

**Comment by fritsch** [comment]
I see no difference running on my laptop.

**Comment by fritsch** [comment]
Testing with @MilhouseVH revealed problems with Bluray Menus when chunksize is too large. We decided between "throughput" and reading too much data. There is a lot increase of speed between 128K and 256K chunks on my testing setup (24.8 MB/s to 41.2 MB/s). Also his blurays work okayish again - people on gigabit should not realize anyways for the menus. For 120+ Mbit/s HEVC everyone will realize cause with those chunks one can read much faster in readahead direction.

**Comment by fritsch** [comment]
jenkins build this please

**Comment by fritsch** [comment]
And finally I removed the ChunkSize commit completely. We need a better solution - something similar like with NFS to match what the server does.

**Comment by fritsch** [comment]
jenkins build and merge

**Comment by MilhouseVH** [comment]
@fritsch, apologies for necro'ing this PR.

I've just found an issue with https://github.com/xbmc/xbmc/pull/12145/commits/58d4b3c68144b377ef57bcb08446de0d0e0c5c97

I've got 4 Samba servers: Win7, Ubuntu 16.04 (Samba 4.3.11), FreeNAS 8.3.x (Samba 3.6.12) and LibreELEC (Samba 4.6.7).

This PR originally started out as http://sprunge.us/SbPa, which I included in build [#0521](https://forum.kodi.tv/showthread.php?tid=298461&pid=2589560#pid2589560) and I didn't notice any problem with any of these Samba servers.

However, on 22 May the PR changed, dropping http://sprunge.us/SbPa and adding "removal of 10 year workaround" (https://github.com/xbmc/xbmc/pull/12145/commits/58d4b3c68144b377ef57bcb08446de0d0e0c5c97).

Since then, ie. [#0522](https://forum.kodi.tv/showthread.php?tid=298461&pid=2590130#pid2590130), on RPi (but strangely not on x86_64), when streaming `smb://` from the FreeNAS server (and only the FreeNAS server - the other 3 Samba servers are all OK), Kodi is unable to play files reliably - playback ends prematurely after about 4-5 seconds (mp4), or there is no playback at all (m2ts), or will end when attempting to seek (avi).

Premature end of playback, mp4: http://sprunge.us/EKCP
No playback at all, mt2s: http://sprunge.us/IVBh (crashlog)

Often Kodi will crash after trying to play a file from the FreeNAS server (possibly a thread is left in an inconsistent state?)

The FreeNAS server is configured with:
```
min protocol = SMB2
max protocol = SMB2
```
and Kodi libsmbclient is 4.6.7 and will negotiate a better than NT1 protocol (preferably SMB3 if available, then SMB2 and finally SMB1/NT1).

However, if I remove the above FreeNAS Samba server settings so that the server supports:
```
min protocol = CORE
max protocol = NT1
```
then this PR is no longer an issue, as Kodi then uses an SMB1/NT1 connection.

Which is to say, this PR is unreliable with SMB2 connections and a Samba 3.6.12 server.

Reverting https://github.com/xbmc/xbmc/pull/12145/commits/58d4b3c68144b377ef57bcb08446de0d0e0c5c97 fixes the issue for SMB2 on FreeNAS, allowing Kodi to reliably play smb:// streams from all 4 servers (none of which support SMB1).

mp4 and then mt2s playback with https://github.com/xbmc/xbmc/pull/12145/commits/58d4b3c68144b377ef57bcb08446de0d0e0c5c97 reverted: http://sprunge.us/fYXN

What is wierd is that https://github.com/xbmc/xbmc/pull/12145/commits/58d4b3c68144b377ef57bcb08446de0d0e0c5c97 isn't an issue on x86_64, it's only a problem on RPi (tested on RPi3).

I don't know what is the best thing to do here - "nothing" is an option.

Obviously this FreeNAS Samba server is quite old, and its age might be contributing to the issue (but why is there no issue with an x86_64 client and SMB2, and only with RPi?), and it's something I only use for testing so personally I don't really care about it (and can personally live without a fix), but obviously many users might be stuck on a Samba server like this and might like to disable SMB1.

Leaving SMB1 enabled on servers of this vintage might be one (the best?) solution, but not a great one for security-type reasons (although given the numerous bugs in old versions of Samba, there's probably many other good reasons to upgrade this server).

Reverting this fix would resolve the issue for old and new servers, but might have a performance impact. Decisions, decisions.

fritsch:
Choose what you like best - I don't use samba so no hard feelings involved.

Any chance to update SW on the NAS?

Am 20.08.2017 10:26 nachm. schrieb "MilhouseVH" <notifications@github.com>:

MilhouseVH:
> Choose what you like best - I don't use samba so no hard feelings involved.

Me neither - I only use it for testing otherwise I really do my best to avoid it (unfortunately that hasn't been working out so well over the last few months!)

At the very least I thought it best to post a comment that this situation now exists after I noticed the  regression. Since there hasn't been any noise about it, maybe it will be OK to leave it as it is.

> Any chance to update SW on the NAS?

For me, yes - upgrading FreeNAS has been on my list for only 2 years (!) and is something I still plan to do eventually, once I find the time (another reason why this is no big deal for me). But for other people with vendor maintained NAS devices, that isn't always an option.

I'm happy to "do nothing" for now, and review that decision based on feedback in the future - hopefully the number of users with this issue is enough to ignore (worst case, they can leave SMB1 enabled).

PR #12148, opened 2017-05-22
[addons] Bring in new binary style (First part)

AlwinEsch:
## Description
This two commits bring in the first parts to allow multiple instances on Kodi.

The first add the new functions in Kodi to create the interface. This are currently Independent from old calls to have the old still compatible.

There becomes now additionally to ADDON_Create the ADDON_CreateInstance used. Due to the "C++" Headers are most parts be done from the classes.

The Second commit change the Screensaver Interface to the new way, who is already used on Kodi agile.

## Types of change

## Checklist:


AlwinEsch:
As note, except screensaver, are on no addons changes on code or version needed.

xbmc/addons/kodi-addon-dev-kit/include/kodi/AddonBase.h
```diff
@@ -0,0 +1,439 @@
...
+  CAddonBase()
+  {
+    CAddonBase::m_interface->toAddon.create_instance = ADDONBASE_CreateInstance;
+    CAddonBase::m_interface->toAddon.destroy_instance = ADDONBASE_DestroyInstance;
+    CAddonBase::m_interface->toAddon.set_setting = ADDONBASE_SetSetting;
+  }
+
+  virtual ~CAddonBase()
+  {
+  }
```
ksooo: <code>virtual ~CAddonBase() = default;</code> maybe?

xbmc/addons/kodi-addon-dev-kit/include/kodi/AddonBase.h
```diff
@@ -0,0 +1,439 @@
...
+ * Mainly is this currently used to identify requested instance types.
+ *
+ * @note This class is not need to know during add-on development thats why
+ * commented with "*".
+ */
+class IAddonInstance
+{
+public:
+  IAddonInstance(ADDON_TYPE type) : m_type(type) { }
+  virtual ~IAddonInstance() { }
```
ksooo: <code>virtual ~IAddonInstance() = default;</code> maybe?

ksooo (approved):
Did a very careful review, found nothing suspicious. The only thing I want to suggest is to do a cleanup of some log strings and code comments wrt. correct English, but this can be done in a followup PR.

ksooo:
jenkins build this please

AlwinEsch:
Thanks @ksooo, the english correction can be easily done during add of addon documentation.

xbmc/addons/kodi-addon-dev-kit/include/kodi/AddonBase.h
AlwinEsch: Here and above is changed to the default, it works too.

The destructors are added there to destruct also addons own class. During tests I has seen that without them a destructor on addon itself was not called (don't know why).

AlwinEsch:
Also some screensaver addons need to have settings access, due to request https://github.com/xbmc/xbmc/pull/12125 I'm not sure who comes first.

For me the best is to insert https://github.com/xbmc/xbmc/pull/12125 before, that I can change settings functions to new style.

After them is in becomes a commit added here to have on new way.

ksooo:
To not delay your work I would opt for merging your stuff first. We can adapt screensaver later.

ksooo:
Alternatively, you leave out the screensaver commit and merge only the core change with this PR.

xbmc/addons/AddonDll.h
```diff
@@ -54,19 +76,51 @@ namespace ADDON
...
     virtual ADDON_STATUS TransferSettings();
 
     static void AddOnStatusCallback(void *userData, const ADDON_STATUS status, const char* msg);
     static bool AddOnGetSetting(void *userData, const char *settingName, void *settingValue);
     static void AddOnOpenSettings(const char *url, bool bReload);
     static void AddOnOpenOwnSettings(void *userData, bool bReload);
+
+    /// addon to kodi basic callbacks below
+    //@{
+    AddonGlobalInterface m_interface;
```
Paxxi: Where does this come from? Tried to find any reference to it and failed.

xbmc/addons/AddonDll.h
```diff
@@ -46,6 +46,28 @@ namespace ADDON
...
+    /*!
+     * @brief Function to destroy a on addon created instance class
+     *
+     * @param[in] instanceID The from Kodi used ID string of active instance
+     */
+    void DestroyInstance(const std::string& instanceID);
+
   protected:
     bool Initialized() { return m_initialized; }
     static uint32_t GetChildCount() { static uint32_t childCounter = 0; return childCounter++; }
```
Paxxi: this isn't threadsafe, somewhat weird design to initialize on a get call and doesn't follow coding style.

Not part of this PR but yeah,

xbmc/addons/AddonDll.cpp
```diff
@@ -46,6 +48,7 @@ CAddonDll::CAddonDll(AddonProps props)
   m_pHelpers    = NULL;
   m_needsavedsettings = false;
   m_parentLib.clear();
+  memset(&m_interface, 0, sizeof(m_interface));
```
Paxxi: this is generally not a good way to initialize a struct/class, it'll destroy the vtable if it happens to have virtual methods.

xbmc/addons/AddonDll.cpp
```diff
@@ -236,6 +240,62 @@ ADDON_STATUS CAddonDll::Create(ADDON_TYPE type, void* funcTable, void* info)
...
+     needed to become the AddOn running */
+  ADDON_STATUS status = m_pDll->Create(&m_interface, nullptr);
+  if (status == ADDON_STATUS_OK)
+  {
+    m_initialized = true;
+  }
+  else if ((status == ADDON_STATUS_NEED_SETTINGS) || (status == ADDON_STATUS_NEED_SAVEDSETTINGS))
+  {
+    m_needsavedsettings = (status == ADDON_STATUS_NEED_SAVEDSETTINGS);
+    if ((status = TransferSettings()) == ADDON_STATUS_OK)
```
Paxxi: don't assign in conditional statements, it saves a line but so easy to mess up and break things.

xbmc/addons/AddonDll.h
AlwinEsch: Is here https://github.com/AlwinEsch/kodi/blob/12213e445eb79ab69e8f291dc0b4d2e08bd234f9/xbmc/addons/kodi-addon-dev-kit/include/kodi/AddonBase.h#L191-L216 because of the size
 is it not direct visible.

I add a note here to know direct the source

xbmc/addons/AddonDll.cpp
```diff
@@ -236,6 +240,62 @@ ADDON_STATUS CAddonDll::Create(ADDON_TYPE type, void* funcTable, void* info)
...
+      m_initialized = true;
+    else
+      new CAddonStatusHandler(ID(), status, "", false);
```
Paxxi: this leaks the status handler

xbmc/addons/AddonDll.cpp
```diff
@@ -236,6 +240,62 @@ ADDON_STATUS CAddonDll::Create(ADDON_TYPE type, void* funcTable, void* info)
...
+  }
+  else
+  { // Addon failed initialization
+    CLog::Log(LOGERROR, "ADDON: Dll %s - Client returned bad status (%i) from Create and is not usable", Name().c_str(), status);
+
+    // @todo currently a copy and paste from other function and becomes improved.
+    CGUIDialogOK* pDialog = g_windowManager.GetWindow<CGUIDialogOK>(WINDOW_DIALOG_OK);
```
Paxxi: This should be removed, but currently there's no support for the dialogok through appmessenger so it's ok for now.

xbmc/addons/AddonDll.cpp
```diff
@@ -278,6 +340,42 @@ void CAddonDll::Destroy()
...
+  {
+    m_usedInstances[instanceID] = std::make_pair(instanceType, addonInstance);
+  }
+
+  return status;
+}
+
+void CAddonDll::DestroyInstance(const std::string& instanceID)
+{
+  std::map<std::string, std::pair<ADDON_TYPE, KODI_HANDLE>>::iterator it = m_usedInstances.find(instanceID);
```
Paxxi: use `auto`

xbmc/addons/AddonDll.cpp
```diff
@@ -425,5 +523,233 @@ bool CAddonDll::CheckAPIVersion(int type)
...
+ * where needed on nearly every add-on (e.g. addon_log_msg) are to add here.
+ *
+ * More specific parts like e.g. to open files should be added to a separate
+ * part.
+ */
+//@{
+ 
+bool CAddonDll::InitInterface(KODI_HANDLE firstKodiInstance)
+{
+  memset(&m_interface, 0, sizeof(m_interface));
```
Paxxi: same as the memset comment above

xbmc/addons/AddonDll.cpp
```diff
@@ -425,5 +523,233 @@ bool CAddonDll::CheckAPIVersion(int type)
...
+
+    return false;
+  }
+
+  CLog::Log(LOGDEBUG, "kodi::General::%s - add-on '%s' requests setting '%s'",
+                        __FUNCTION__, addon->Name().c_str(), settingName);
+
+  if (strcasecmp(settingName, "__addonpath__") == 0)
+  {
+    strcpy((char*) settingValue, addon->Path().c_str());
```
Paxxi: don't use strcpy, so easy to mess up and write random memory.
get_Setting need to allocate the string here and return it or know about the size but preferably allocate it.

xbmc/addons/AddonDll.cpp
```diff
@@ -425,5 +523,233 @@ bool CAddonDll::CheckAPIVersion(int type)
...
+
+      if (id == settingName && !type.empty())
+      {
+        if (type == "text"     || type == "ipaddress" ||
+            type == "folder"   || type == "action"    ||
+            type == "music"    || type == "pictures"  ||
+            type == "programs" || type == "fileenum"  ||
+            type == "file"     || type == "labelenum")
+        {
+          strcpy((char*) settingValue, addon->GetSetting(id).c_str());
```
Paxxi: same as above

xbmc/addons/AddonDll.cpp
```diff
@@ -425,5 +523,233 @@ bool CAddonDll::CheckAPIVersion(int type)
...
+          return true;
+        }
+        else if (type == "number" || type == "enum")
+        {
+          *(int*) settingValue = (int) atoi(addon->GetSetting(id).c_str());
```
Paxxi: use c++ style casts

xbmc/addons/AddonDll.cpp
```diff
@@ -425,5 +523,233 @@ bool CAddonDll::CheckAPIVersion(int type)
...
+          return true;
+        }
+        else if (type == "bool")
+        {
+          *(bool*) settingValue = (bool) (addon->GetSetting(id) == "true" ? true : false);
```
Paxxi: use c++ style casts

xbmc/addons/AddonDll.cpp
```diff
@@ -425,5 +523,233 @@ bool CAddonDll::CheckAPIVersion(int type)
...
+          return true;
+        }
+        else if (type == "slider")
+        {
+          const char *option = setting->Attribute("option");
+          if (option && strcmpi(option, "int") == 0)
+          {
+            *(int*) settingValue = (int) atoi(addon->GetSetting(id).c_str());
```
Paxxi: use c++ style casts

xbmc/addons/AddonDll.cpp
```diff
@@ -425,5 +523,233 @@ bool CAddonDll::CheckAPIVersion(int type)
...
+            return true;
+          }
+          else
+          {
+            *(float*) settingValue = (float) atof(addon->GetSetting(id).c_str());
```
Paxxi: use c++ style casts

Paxxi (changes requested):
sorry to be so harsh but these things need to be addressed.

Keep up the good work though, the overall design looks good.

xbmc/addons/AddonDll.h
AlwinEsch: Does go away complete soon :D

xbmc/addons/AddonDll.cpp
AlwinEsch: Is only a poor "C" part. Does a `m_interface = 0` also work?

xbmc/addons/AddonDll.cpp
AlwinEsch: This is currently a c&p from old, this way with "ADDON_STATUS_NEED_SAVEDSETTINGS" becomes complete removed in future.

Wanted to keep the changes as small as possible.

No problem I add a new line.

xbmc/addons/AddonDll.cpp
ksooo: ALthough you are right, please don't forget that larger parts of this code are just moved from elsewhere during refactoring. Question is how far to go - means is fixing all sh*t contained in the original code in scope?

xbmc/addons/AddonDll.cpp
AlwinEsch: The next part who must be removed asap, also a remnant from the old system who comes with c&p to there.

If acceptable for you becomes this done on soon coming requests. Is related to my comment before (only be used for this "ADDON_STATUS_NEED_SAVEDSETTINGS" and "ADDON_STATUS_NEED_SETTINGS")

xbmc/addons/AddonDll.cpp
AlwinEsch: Want to leave them currently like it is, also thought to change them but with the coming https://github.com/xbmc/xbmc/pull/12125 I want to hold differences to them on "xbmc/addons/interfaces/Addon/AddonCallbacksAddon.cpp" so small as possible.

OK for you?

AlwinEsch:
Currently is much only a c&p from the old way. Because of the size I want todo them step by step, otherwise it breaks the current addons who not attached.

AlwinEsch:
The good think with the c++ addon headers is that it can be fixed without bigger changes on addons, the "C" part is completely done from the headers.

xbmc/addons/AddonDll.cpp
AlwinEsch: Tried it, a memset with a `= 0` not possible. Have no alternative to them, the structure must complete in "C" style and also not want to add a "C++" constructor there.

ksooo:
@Paxxi all very valid points, but may I suggest to do the cleanup once all pieces are in place? Currently, larger parts are just temporary and will not be present once everything for the addon rewrite got merged, thus tweaking right now might be work for the garbage bin.

AlwinEsch:
jenkins build this please

AlwinEsch:
jenkins build this please

Paxxi:
@ksooo I would agree to that in theory. In reality there's almost never anyone who goes back and cleans up newly merged code. Style issues I can let pass but the bits like leaking memory shouldn't be merged and forgotten about.

xbmc/addons/AddonDll.cpp
Paxxi: `m_interface = {0}` should work on all platforms.

xbmc/addons/AddonDll.h
Paxxi: if it's temporary then it's ok.

xbmc/addons/AddonDll.cpp
Paxxi: if it's planned for a future PR then it's OK for me.

ksooo:
> the bits like leaking memory shouldn't be merged and forgotten about.

 @Paxxi agreed, but there is no memleak with <code>CAddonStatusHandler</code>. This class is derived from <code>CThread</code> and if constructed with parameter bSameProcess set to false, <code>CThread::Create(true)</code> will be called from inside <code>CAddonStatusHandler</code> ctor. "true" means "thread auto deletes itself" here. I agree that this is no good class design, but no memleak.

 What else do you see as blocker for this PR?

xbmc/addons/AddonDll.h
AlwinEsch: Is a remnant of the old path of Hack was introduced to make multiple instances of an addon.

There was the addon dll copied to another name to open them independent again.

xbmc/addons/AddonDll.cpp
AlwinEsch: Is removed on this file and also removed them on Screensaver class who was brought before.

AlwinEsch:
@Paxxi think to have all the parts changed and fixed, thanks a lot! Looks now better and learned a new part (`= { 0 }` :D ).

AlwinEsch:
jenkins build this please

AlwinEsch:
OK to merge?

ksooo:
Go for it!

AlwinEsch:
jenkins build this please

AlwinEsch:
Thanks a lot @MartijnKaijser

PR #12149, opened 2017-05-22
Fix libdvdcss-2 hooking which was broken after a8a3a0db60171328fb962b…

Voyager1:
…700ba9b85ac25373a3

## Description
removal of kernel32 emulation also partially broke the DLL loading for libdvdcss-2 (because that's not directly called, rather indirectly by libdvdnav.dll)

## Motivation and Context
DVD files on our virtual file system (such as SMB shares etc.) now play correctly again, as before only local drives worked correctly.

## How Has This Been Tested?
using ISO files on an SMB share

## Types of change

## Checklist:


Paxxi (approved):
Looks good to me.

Need to fix this properly but this is good in the mean time.

PR #12154, opened 2017-05-24
[fix] gcc 4.8 doesn’t correctly implement std::regex

Rechi:
## Description
Use our own `CRegExp` implementation instead of `std::regex`.

## Motivation and Context
Kodi compiled with GCC 4.8 (Ubuntu 14) crashes on startup, because `std::regex` is not implemented for that gcc version.

## How Has This Been Tested?
tested with a malicious file and run the test suite

## Types of change

## Checklist:


MartijnKaijser:
jenkins build this please

PR #12156, opened 2017-05-24
sqlite: Close file before attempting deletion

MilhouseVH:
Thanks for the [suggestion](http://forum.kodi.tv/showthread.php?tid=314965&pid=2591323#pid2591323), @stefansaraev 

This is just a hack on a hack, Windows might still fail to delete the file for many other reasons but it shouldn't fail because we still have the file open.


Rechi (approved):
works with this change on windows

PR #12164, opened 2017-05-26
[addons] add new IAddonInstanceHandler

AlwinEsch:
## Description
This class becomes in future used as parent on the various binary
addon instances as alternative to the CAddonDll who can no more
used as parent.

Currently this class include only the minimum but predefine the
coming multiple instance support.

Also is screensaver changed to new way and nearly in the style on end.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/AddonInstanceHandler.cpp
+++ b/xbmc/addons/AddonInstanceHandler.cpp
@@ -1,0 +1,57 @@
+/*
+ *      Copyright (C) 2005-2017 Team Kodi
+ *      http://kodi.tv
+ *
+ *  This Program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2, or (at your option)
+ *  any later version.
+ *
+ *  This Program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with Kodi; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "AddonInstanceHandler.h"
+
+#include "utils/log.h"
+#include "utils/StringUtils.h"
+
+namespace ADDON
+{
+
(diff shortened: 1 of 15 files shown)


AlwinEsch:
Needed changes for this request are done. @Paxxi @Montellese @ksooo is this style OK?

xbmc/addons/AddonInstanceHandler.cpp
```diff
@@ -0,0 +1,57 @@
...
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with Kodi; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "AddonInstanceHandler.h"
+
+#include "utils/log.h"
```
ksooo: No functionality from utils/logs.h used in this ccp file, as far as I can see. Could be removed, imo.

xbmc/addons/AddonInstanceHandler.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+IAddonInstanceHandler::~IAddonInstanceHandler()
+{
+}
+
+bool IAddonInstanceHandler::CreateInstance(KODI_HANDLE instance)
+{
+  if (m_addon && m_addon->CreateInstance(m_type, m_instanceId, instance, m_parentInstance) == ADDON_STATUS_OK)
+    return true;
+  return false;
```
ksooo: Why not just <code>return (m_addon && m_addon->CreateInstance(m_type, m_instanceId, instance, m_parentInstance) == ADDON_STATUS_OK)</code>?

xbmc/addons/AddonInstanceHandler.h
```diff
@@ -0,0 +1,57 @@
...
+    const ADDON_TYPE UsedType() const { return m_type; }
+    const std::string& InstanceID() { return m_instanceId; }
+
+    std::string ID() const { return m_addon->ID(); }
+    std::string Name() const { return m_addon->Name(); }
+    std::string Author() const { return m_addon->Author(); }
+    std::string Icon() const { return m_addon->Icon(); }
+    std::string Path() const { return m_addon->Path(); }
+    std::string Profile() const { return m_addon->Profile(); }
+    AddonVersion Version() const { return m_addon->Version(); }
```
ksooo: In the cpp file you are checking m_addon for nullptr before accessing it. Why not in the header file?

AlwinEsch:
@ksooo have added a check to the functions on header, further is a new commit there who translate the addon status to a string and becomes used on IAddonInstanceHandler to bring a error.

The `m_addon` as nullptr brings now a LOGFATAL, normally should this never happen.

ksooo:
@AlwinEsch From my point of view okay to merge now. cannot really judge on the screensaver changes but I guess this is good to go as well.

ksooo:
jenkins build this please

AlwinEsch:
jenkins build this please

PR #12166, opened 2017-05-26
Fix crash when closing standalone games

garbear:
Currently, if a game add-on is loaded without a file (standalone mode), when closed it will segfault.

The cause is from #11380, where the in-game saves variable was only initialized for file-based games.

Also included are three code improvements for the games code.


AlwinEsch:
Looks good for me, has also seen also sometimes crashes after the close of a dosbox, thought only that it comes from a bad dosbox dll here at home :)

garbear:
I think this is the dosbox error you're talking about: https://github.com/libretro/dosbox-libretro/pull/47

garbear:
jenkins build this please

PR #12169, opened 2017-05-27
ADD: [droid] allow autostarting at boot (via xbmc_env.proprties)

koying:
As the title says, add much-requested possibility to start kodi on boot.
Enable via xbmc_env.properties, `xbmc.autostart=yes`


koying:
jenkins build this please

PR #12170, opened 2017-05-27
CHG: [droid] stop video if screen goes off

koying:
as the title says

### DIFF
--- a/xbmc/platform/android/activity/XBMCApp.cpp
+++ b/xbmc/platform/android/activity/XBMCApp.cpp
@@ -179,6 +179,7 @@ void CXBMCApp::onResume()
   intentFilter.addAction("android.intent.action.SCREEN_ON");
   intentFilter.addAction("android.intent.action.HEADSET_PLUG");
   intentFilter.addAction("android.intent.action.HDMI_AUDIO_PLUG");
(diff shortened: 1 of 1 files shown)


peak3d (approved):

MartijnKaijser:
jenkins build this please

CiNcH83:
Are you sure about that? On Sony you can switch off the screen via ACTION MENU with audio still playing. I don't know whether this triggers the "android.intent.action.SCREEN_OFF" though. If I interpret the PR correctly, it would stop playback alltogether?

Currently this does not matter though as bringing up the ACTION MENU still stops playback anyway.

koying:
Is it so? 
Anyway, this one only stops video, not audio only. 
It's in spmc since ages, so you can test

CiNcH83:
Tried it out now with SPMC 16.6.0. It behaves well. When I do ACTION MENU > Picture Off on my Sony, the screen goes off, but video keeps playing, so you still hear sound. So it seems that Picture Off does not trigger the "android.intent.action.SCREEN_OFF".

koying:
Yeah. Pretty sure anything in the Action menu relates to the "TV" subsystem, and is completely independedant from the "Android" subsystem.

PR #12171, opened 2017-05-27
FIX: Handle removable drives for media window title

koying:
If we don't have a label for the media window, check if it's a removable drive, and use its label if it is

### DIFF
--- a/xbmc/windows/GUIMediaWindow.cpp
+++ b/xbmc/windows/GUIMediaWindow.cpp
@@ -770,6 +770,21 @@ bool CGUIMediaWindow::Update(const std::string &strDirectory, bool updateFilterP
   }
 
   if (m_vecItems->GetLabel().empty())
+  {
+    // Removable sources
+    VECSOURCES removables;
+    g_mediaManager.GetRemovableDrives(removables);
+    for (auto s : removables)
+    {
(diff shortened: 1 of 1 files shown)


xbmc/windows/GUIMediaWindow.cpp
```diff
@@ -770,8 +770,22 @@ bool CGUIMediaWindow::Update(const std::string &strDirectory, bool updateFilterP
   }
 
   if (m_vecItems->GetLabel().empty())
+  {
     m_vecItems->SetLabel(CUtil::GetTitleFromPath(m_vecItems->GetPath(), true));
 
+    // Removable sources
```
xhaggi: Should we need another empty check here?

koying: No, GetTitleFromPath will always returns a cleaned up path suitable as a label.
But maybe you mean the removables should be checked first, and GetTitleFromPath only called if empty, and then I agree ;)

xhaggi: Yeah something like that if we always want the label be set if a removable path match.

xhaggi (approved):

MartijnKaijser:
jenkins build this please

PR #12176, opened 2017-05-27
Remove fallback definition of FILE

coypoop:
the typedef provided is wrong on most platforms, so they have chosen to use -D_FILE_DEFINED.
rather than do this, don't provide a fallback definition. standard C promises that FILE is defined in stdio.h

build tested only on a currently unsupported platform (NetBSD) - I haven't succeeded in running it yet.


coypoop:
that failure is due to hitting a time limit on the tests, not a build issue

notspiff:
the xbox pretty much ignored all such standards. this should be all good to kill.

Rechi:
maybe @Paxxi knows something about it

notspiff:
this codes originates from xbplayer i believe, back on the original xbox in 2002. nobody around except me should anything about this ;) we basically had to hook a homegrown posix io right into the kernel, and this definition was part of that effort. it was then put under an ifdef when the initial linux/windows ports happened instead of simply removed. i believe that struct is compatible with what mingw had back then (so we could load mplayer.dll).

MartijnKaijser:
jenkins build this please

Rechi:
jenkins build this please

MartijnKaijser:
@coypoop thanks for your contribution

PR #12177, opened 2017-05-28
[addons] change addon audio engine callbacks to new way

AlwinEsch:
## Description
This add the currently with libKODI_audioengine.h done part to the
new style.

Due to missing addons who use them was a complete test currently not
possible and must be check in future, from the base is the work
equal to the old.

The old parts becomes in next commit removed.

Also includes this the complete documentation for them who need be maybe improved in future (in a well-English speaking).

## Types of change

## Checklist:


AlwinEsch:
Found during add and test of this a fault who I want to prevent, have added this https://github.com/xbmc/xbmc/pull/12178 as fix. Without them need every addon a version change for changes like here.

AlwinEsch:
jenkins build this please

PR #12181, opened 2017-05-28
When playing back an item that is an addon pass a new script param to…

faush01:
… addon script that lets the addon know if the user selected resume of not in the built in resume dialog question.

If a user selected resume then the string resume will be passed to the addon script.

## Description
When the user plays an item that is an addon item and there is a progress set they will be given a resume dialog. All this change does is pass the result of the resume question on to the addon via a new script param string "resume".
This allows addons to use the built in resume question response instead of having to handle this in the addon, the addon will have the answer to what the user wanted to do.

## Motivation and Context

I have written a number of addons over the years and this allows you to build addons that are more in line with the global look and feel of the app.
Having this data passed to the addon script gives the user a smoother experience.

## How Has This Been Tested?

Tested with addons and playback process to test regressions.

Tested on Windows, Run all the built in tests.

This change is low risk, none of the internal logic is being changed, if a user asks for resume a new script param is added when the addon is called, that is all.
## Types of change

## Checklist:

### DIFF
--- a/xbmc/filesystem/PluginDirectory.cpp
+++ b/xbmc/filesystem/PluginDirectory.cpp
@@ -112,7 +112,7 @@ CPluginDirectory *CPluginDirectory::dirFromHandle(int handle)
   return NULL;
 }
 
-bool CPluginDirectory::StartScript(const std::string& strPath, bool retrievingDir)
+bool CPluginDirectory::StartScript(const std::string& strPath, bool retrievingDir, bool resume)
 {
   CURL url(strPath);
 
@@ -151,8 +151,13 @@ bool CPluginDirectory::StartScript(const std::string& strPath, bool retrievingDi
   argv.push_back(strHandle);
   argv.push_back(options);
 
+  std::string strResume = "resume:false";
+  if (resume)
+    strResume = "resume:true";
+  argv.push_back(strResume);
+
   // run the script
-  CLog::Log(LOGDEBUG, "%s - calling plugin %s('%s','%s','%s')", __FUNCTION__, m_addon->Name().c_str(), argv[0].c_str(), argv[1].c_str(), argv[2].c_str());
+  CLog::Log(LOGDEBUG, "%s - calling plugin %s('%s','%s','%s','%s')", __FUNCTION__, m_addon->Name().c_str(), argv[0].c_str(), argv[1].c_str(), argv[2].c_str(), argv[3].c_str());
   bool success = false;
   std::string file = m_addon->LibPath();
   int id = CScriptInvocationManager::GetInstance().ExecuteAsync(file, m_addon, argv);
@@ -170,12 +175,12 @@ bool CPluginDirectory::StartScript(const std::string& strPath, bool retrievingDi
   return success;
 }
 
-bool CPluginDirectory::GetPluginResult(const std::string& strPath, CFileItem &resultItem)
+bool CPluginDirectory::GetPluginResult(const std::string& strPath, CFileItem &resultItem, bool resume)
 {
   CURL url(strPath);
   CPluginDirectory newDir;
 
-  bool success = newDir.StartScript(strPath, false);
+  bool success = newDir.StartScript(strPath, false, resume);
 
   if (success)
   { // update the play path and metadata, saving the old one as needed
@@ -429,7 +434,7 @@ void CPluginDirectory::AddSortMethod(int handle, SORT_METHOD sortMethod, const s
 bool CPluginDirectory::GetDirectory(const CURL& url, CFileItemList& items)
 {
   const std::string pathToUrl(url.Get());
-  bool success = StartScript(pathToUrl, true);
+  bool success = StartScript(pathToUrl, true, false);
 
   // append the items to the list
   items.Assign(*m_listItems, true); // true to keep the current items
@@ -437,7 +442,7 @@ bool CPluginDirectory::GetDirectory(const CURL& url, CFileItemList& items)
   return success;
 }
 
-bool CPluginDirectory::RunScriptWithParams(const std::string& strPath)
+bool CPluginDirectory::RunScriptWithParams(const std::string& strPath, bool resume)
 {
   CURL url(strPath);
   if (url.GetHostName().empty()) // called with no script - should never happen
@@ -464,8 +469,13 @@ bool CPluginDirectory::RunScriptWithParams(const std::string& strPath)
   argv.push_back(strHandle);
   argv.push_back(options);
 
+  std::string strResume = "resume:false";
+  if (resume)
+    strResume = "resume:true";
+  argv.push_back(strResume);
+
   // run the script
-  CLog::Log(LOGDEBUG, "%s - calling plugin %s('%s','%s','%s')", __FUNCTION__, addon->Name().c_str(), argv[0].c_str(), argv[1].c_str(), argv[2].c_str());
+  CLog::Log(LOGDEBUG, "%s - calling plugin %s('%s','%s','%s','%s')", __FUNCTION__, addon->Name().c_str(), argv[0].c_str(), argv[1].c_str(), argv[2].c_str(), argv[3].c_str());
   if (CScriptInvocationManager::GetInstance().ExecuteAsync(addon->LibPath(), addon, argv) >= 0)
     return true;
   else
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -2918,8 +2918,9 @@ bool CApplication::PlayMedia(const CFileItem& item, const std::string &player, i
   //If item is a plugin, expand out now and run ourselves again
   if (item.IsPlugin())
   {
+    bool resume = item.m_lStartOffset == STARTOFFSET_RESUME;
     CFileItem item_new(item);
-    if (XFILE::CPluginDirectory::GetPluginResult(item.GetPath(), item_new))
+    if (XFILE::CPluginDirectory::GetPluginResult(item.GetPath(), item_new, resume))
       return PlayMedia(item_new, player, iPlaylist);
     return false;
   }
@@ -3197,8 +3198,9 @@ PlayBackRet CApplication::PlayFile(CFileItem item, const std::string& player, bo
 
   if (item.IsPlugin())
   { // we modify the item so that it becomes a real URL
+    bool resume = item.m_lStartOffset == STARTOFFSET_RESUME;
     CFileItem item_new(item);
-    if (XFILE::CPluginDirectory::GetPluginResult(item.GetPath(), item_new))
+    if (XFILE::CPluginDirectory::GetPluginResult(item.GetPath(), item_new, resume))
       return PlayFile(std::move(item_new), player, false);
     return PLAYBACK_FAIL;
   }
@@ -4227,7 +4229,7 @@ bool CApplication::OnMessage(CGUIMessage& message)
       // handle plugin://
       CURL url(file.GetPath());
       if (url.IsProtocol("plugin"))
-        XFILE::CPluginDirectory::GetPluginResult(url.Get(), file);
+        XFILE::CPluginDirectory::GetPluginResult(url.Get(), file, false);
 
       // Don't queue if next media type is different from current one
       if ((!file.IsVideo() && m_pPlayer->IsPlayingVideo())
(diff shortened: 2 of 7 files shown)


notspiff:
uhm, you can read the startoffset in the add-on can't you ? that was at least the original intent.

faush01:
That makes sense, how is it surfaced in the listitem? As a property?

notspiff:
it is available through getProperty('startoffset') indeed. but yes, i now recall how it works;

there are two ways to do this: one is using setResolvedUrl type operation. in that case, if the *original* url (i.e. not the resolved media url!) is a constant, kodi will save info based on the original url and handle the resume for that item as normal.

alternatively you can set the startoffset on the resolved item returned in the add-on. in that case kodi will use whatever value you put in there (using setProperty('startoffset', secs-as-float)). the second here was intended to be the hook used if you want to handle this on the add-on side.

faush01:
setResolvedUrl  might work but the addon I am writing does not use this approach.

The addon uses the following approach:
 - Get a list of media items from the Emby Media server https://emby.media
 - Build a list of ListItems and return that to Kodi to be displayed
 - In the ListItems the progress is set so Kodi and the Kodi skins show the correct progress icons and indicators
 - The URL of the ListItems are set to a plugin:// play action in the addon so when a user selects an item the addon is called with the ID of the item to play
 - Because the progress is set in the ListItem when a user selects an in progress item they will be promoted for resume or start from beginning by the default built in Kodi resume question
 - The play action of the addon is called and this is responsible for getting the actual URL and using the kodi player to play the file
 - If the item is part played (details from the Emby Server) then seek is called on the playback item to take the player to the users progress point

I the above scenario the only way I can get access to the answer to the built in resume or from start question is using the code change I have submitted.

This should be all low risk as the only logic change is passing along to the addon script the extra param of resume:true|false

faush01:
It's coming up to the weekend here so if anyone has any feedback or changes let me know and I will spend some time over the next few days sorting them out.

Otherwise if no feedback are we happy to merge this?

faush01:
Any more feedback?
Can we merge this?

faush01:
Hi guys, not sure what you want to do with this one now, its been a while, just wondering if you want to merge or just close it.

xbmc/filesystem/PluginDirectory.cpp
```diff
@@ -151,6 +151,11 @@ bool CPluginDirectory::StartScript(const std::string& strPath, bool retrievingDi
   argv.push_back(strHandle);
   argv.push_back(options);
 
+  std::string strResume = "resume:false";
+  if (resume)
+    strResume = "resume:true";
+  argv.push_back(strResume);
+
   // run the script
   CLog::Log(LOGDEBUG, "%s - calling plugin %s('%s','%s','%s')", __FUNCTION__, m_addon->Name().c_str(), argv[0].c_str(), argv[1].c_str(), argv[2].c_str());
```
Razzeee: You might want to also log it here

xbmc/filesystem/PluginDirectory.cpp
```diff
@@ -464,6 +469,11 @@ bool CPluginDirectory::RunScriptWithParams(const std::string& strPath)
   argv.push_back(strHandle);
   argv.push_back(options);
 
+  std::string strResume = "resume:false";
+  if (resume)
+    strResume = "resume:true";
+  argv.push_back(strResume);
+
   // run the script
   CLog::Log(LOGDEBUG, "%s - calling plugin %s('%s','%s','%s')", __FUNCTION__, addon->Name().c_str(), argv[0].c_str(), argv[1].c_str(), argv[2].c_str());
```
Razzeee: Missing log again

xbmc/video/windows/GUIWindowVideoBase.cpp
```diff
@@ -763,7 +763,7 @@ std::string CGUIWindowVideoBase::GetResumeString(const CFileItem &item)
 
 bool CGUIWindowVideoBase::ShowResumeMenu(CFileItem &item)
 {
-  if (!item.m_bIsFolder && !item.IsPVR())
+  if (!item.m_bIsFolder && !item.IsPVR() || item.IsPlugin())
```
Razzeee: This seems weird and/or dangerous. A plugin should already confirm to the first two rules. It's not a folder and it's not PVR.

faush01: yes, just confirmed, item.m_bIsFolder is True when an addon is called from the Info screen.

Rechi: m_bIsFolder depends on the value specified for isFolder at  xbmcplugin.addDirectoryItem (https://codedocs.xyz/xbmc/xbmc/group__python__xbmcplugin.html#ga4c357cf7d3dbd4e2823755ecfd7dfa01)

faush01: I am using:

dirItems is an array of (u, list_item, folder)
xbmcplugin.addDirectoryItems(pluginhandle, dirItems)

and folder is being set correctly, it is false for the items I am adding.

This is occurring when I hit the play button on the media info screen. It does not happen when I select play for the item in the dir list, but if I hit Info to show info screen and then hit the play this is when I see item.m_bIsFolder being True

Razzeee: Then that's a bug. Nothing you should try to tackle in this part of the code :)

faush01: Ok, I will remove this change.
So do you want me to try to find the root cause for this?

Razzeee: If you can put time into that, that would be appreciated.

faush01: The root cause of this was the new movie item not being created from the original CFileItemPtr but it used the video info tag to populate it. This used some weird logic to work out if it was a folder or not. If there was no FileAndPath set in the video info tag then it was assumed it was a dir and set the path to / and marked it as a Folder.

I made a small change to populate the movie item from the base CFileItemPtr, tjis uses the original items IsFolder value.

This is now working.

Razzeee:
@notspiff I would like to hear your thought on this, with the added information :)

ios and linux failed, let's try again

jenkins build this please

Rechi:
please rebase first, before trying to build it again

Razzeee:
whoops, got ahead of myself

akva2:
sure but i'm busy busy, monday at the earliest.

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -597,7 +597,7 @@ void CGUIDialogVideoInfo::Play(bool resume)
     return;
   }
 
-  CFileItem movie(*m_movieItem->GetVideoInfoTag());
+  CFileItem movie(*m_movieItem);
```
notspiff: this looks potentially icky. it will change from an item pointing to the actual underlying file to an item pointing to a library item. not sure if it has consequences but why is this change needed?

notspiff:
it seems reasonable. only thing i question is whether the bool could be parsed more elegantly somehow. a lot of signature impact. it for sure could be given a default of false in any case.

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
faush01: There are overloaded constructors, the one that takes a VideoInfoTag does not populate all the member level isFolder type data it uses some internal logic to set them from data in the VideoInfoTag which is not consistent with what the original item has.
Using the base Item it allows the new movie item to be populated using the full item info.
I feel it is actually more correct to use the constructor that passes in the full item if it is available, in this instance the full item is available and should be used.

notspiff: i know for a fact that 5 years ago i tracked a bug down to the fact that we did what you suggest, and changed it to what it is now. i cannot guarantee whether the problem still exists but this potentially causes regressions and i this needs to be investigated! i do not recall the details..

notspiff: actually it's more like 10 years ago, sometimes prior to 2009 so it's all smashed into the svn to git conversion ;) but i looked into it. the problem is that we hop into the ShowResumeMenu, which calls into the Database to get resume points. these are stored by path, and not by the virtual db path, so i think that was the issue. but then again maybe the surrounding code has changed so the m_movieItem points to proper paths in the first place now?

faush01: I had a look back through the git log for this file but I could only go back to 25/1/2011 and could not see any changes to this function.
I am not sure what or how to regression test this with no idea where to start with what to look for.

faush01: ok so its a DB path resume point issue, I can test this.

faush01: the other option is just to use the first approach and check if the item is a plugin when displaying the resume dialog.

notspiff: sounds safer. but current code is non-intuitive so if it can be changed that is preferable.

faush01: So the original idea was to add
if (!item.m_bIsFolder && !item.IsPVR() **|| item.IsPlugin()**)
to
xbmc/video/windows/GUIWindowVideoBase.cpp 
See above in the hidden change request, it was requested that I dont do that there and find the root cause. 
I found the root cause as the IsFolder not being set so then is when I realised the IsFolder was not being set when using the VideoInfoTag constructor.
If you think it is better to go back to the original idea of adding || item.IsPlugin() then I can do that.

faush01:
Ok I changed it back to use the original approach of using item.IsPlugin() in the if statement.
I am really not sure where to go from here, if we have people unhappy with both approaches perhaps someone can suggest a new approach.
Or we can just go with this approach as it is quite descriptive, if it is a plugin then show the resume, cant get more descriptive than that :-)

notspiff:
rest assured we will find a solution. nobody is objecting your change, it's just takes some stamina to work in huge code bases :)

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
Razzeee: @faush01 did you find some time to go back and look at "ok so its a DB path resume point issue, I can test this."?
might be the best way to proceed, otherwise we'll have to live with the "IsPlugin-Hack" But it will need to be documented/commented.

faush01:
So ultimately the root cause is that the VideoInfoTag did not have a m_strFileNameAndPath set.
To allow an addon to set this I added a info tag to the video type on the ListItem.cpp to allow an addon to set FileNameAndPath, this fixes the root cause, is backwards compatible and should have the lowest code impact of all.
What do people think?

Razzeee:
should work fine, as long as addons make sure to always set it consistently, but that's out of our hands obviously :)
sounds fine to me (without seeing any code)

faush01:
I have amended the changes in this pull request to include the video info tag FileNameAndPath already, you should be able to see the code in the change set in this pull request.

faush01:
I have re-based on master to fix a conflict with the changes on master since this pull request has been open for so long.
It would we good if we could nail this down soon.

faush01:
Hi Guys, its getting close to 2 months this has been open.
I think we have covered all the requests for change and found the lowest risk and most compatible solution.
I was just wondering if this can be merged now.
Thanks.

Razzeee:
jenkins build this please

faush01:
Not sure why the failed build, the failed build logs look like they been deleted now.

Razzeee:
jenkins build this please

faush01:
Anyone know what this error means:

```
Checking out Revision 45d0cd49da1028f714e6b2bde920a3c0695b8f0c (refs/remotes/origin/pr/12181/merge)
org.eclipse.jgit.errors.MissingObjectException: Missing unknown 45d0cd49da1028f714e6b2bde920a3c0695b8f0c
```

MartijnKaijser:
Jenkins build this please

MartijnKaijser:
@faush01 could you please do a rebase on latest master and force push?

faush01:
@MartijnKaijser I have rebased on latest master

faush01:
Jenkins build this please

Razzeee:
jenkins build this please

PR #11009, opened 2016-11-29
[GUIInfo] calculate correct remaining play time for speed up playback

Rechi:
## Description

Calculate the correct remaining play time for speed up or slowed playback.

## Motivation and Context

Always get the correct time during playback for `Ends at: ??:??` in the GUI.

## How Has This Been Tested?

Runtime tested

## Types of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [x] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #12186, opened 2017-05-28
Expand tempo range to 0.5x-5.0x

aschmitz:
## Description
This expands the range of tempo settings available via tempoup /
tempodown commands from 0.75x-1.5x to 0.5x-5.0x.

There is a minor chance that this will interact with fast forward
behavior, but that seems unlikely to be problematic: if the user
modifies the tempo at 2x or 4x fast forward, the new tempo setting
will take effect, and will be used as the base when multiplied
or divided by 2 if the user then presses "rewind" or "forward".

When the value exceeds 32x or gets down below 1.55x, the "rewind"
and "forward" commands will treat it as 1, escaping any odd
behavior. If necessary, we can clamp values in "rewind" and
"forward" to the nearest power of 2, but that seems like overkill
in this situation.

## Motivation and Context
I frequently watch content at speeds in excess of the existing 1.5x limit (2.0x-2.5x typically, occasionally higher). Merging this change allows me to do so in Kodi. I chose 5x as an arbitrary upper bound that seems excessive. I decreased the tempo minimum as well, in case other users would prefer to view media even slower.

## How Has This Been Tested?
It hasn't. I was unable to find any relevant tests, although I am extremely new to this codebase.

N/A

## Types of change

## Checklist:

### DIFF
--- a/xbmc/interfaces/builtins/PlayerBuiltins.cpp
+++ b/xbmc/interfaces/builtins/PlayerBuiltins.cpp
@@ -142,7 +142,7 @@ static int PlayerControl(const std::vector<std::string>& params)
     if (g_application.m_pPlayer->IsPlaying() && !g_application.m_pPlayer->IsPaused())
     {
       float playSpeed = g_application.m_pPlayer->GetPlaySpeed();
-      if (playSpeed >= 0.75 && playSpeed <= 1.55)
+      if (playSpeed >= 0.45 && playSpeed <= 1.55)
         playSpeed = 1;
 
       if (paramlow == "rewind" && playSpeed == 1) // Enables Rewinding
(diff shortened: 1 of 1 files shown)


Rechi:
You have missed some changes.
- It doesn't show `Play Speed: x` if speed is lower then 0.8 or higher then 1.5
- Finish Time is only correct for play speed between 0.8 and 1.5

This change currently also conflicts with fast forward. If playing with 1.6 and then pressing fast forward it doesn't switch to fast forward 2x, with play speed 1.5 it does.

@FernetMenta any comments?

FernetMenta:
ffmpeg's atempo filter works in a range from 0.5 to 2.0. we have to stay within this range as well. I won't push min/max to the very limits, maybe 0.6 - 1.9
changing limits needs proper testing because AE will trigger resync if a/v error gets too big.

FernetMenta:
with regard to player controls:
currently ff/rw and tempo uses playes's SetPlaySpeed. that is not ideal and has limitations, i.e. tempo gets lost when pausing tempo mode. if we indroduced a method SetTempo for player, I would get more control. interference with ff/fw would also get resolved.

aschmitz:
I'll admit I didn't follow the `SetPlaySpeed` call all the way through, and I didn't realize this was going through `atempo`. I know much higher tempos work in VLC, which I believe is also using ffmpeg's libavcodec, so I would expect that there's *some* way to make it work, but I'm not sure what the best way to do that is. Is that something that's likely to be practical in the near term? If so, can you point me in the general direction of that code?

An increase to 2.0x (or 1.9x, I suppose) would certainly be nice, but a bit faster would be handy as well.

@Rechi Have you found that the finish time is accurate even before this change? I've found that it isn't for me, but I was going to address that in a separate change when I got sufficiently annoyed by it.

I agree that if we do get a `SetTempo`, a few of these issues would probably be easy to resolve: if it's not [within some epsilon of] 1.0, display "Play Speed: x", and the extra checks around FF/RW could be removed.

Rechi:
@asavah Yes finish time is correct if playing at a speed which is within the atempo range (#11009).

The issue with a higher speed then 1.9 is that it will conflict with `PLAYER_FORWARDING_2x` and maybe also higher ones.

FernetMenta:
In order to meke atempo work for > 2.0 you would need to cascade multiple filters. IMO this effort and additional code path does not make too much sense because most systems will struggle with video decoding at this speed.

MartijnKaijser:
See https://github.com/xbmc/xbmc/pull/12470

PR #12196, opened 2017-05-30
X11 Cleanup

lrusak:
This does the following
 - adds cmake module `FindGLX.cmake` to add extra definitions
 - adjust `FindEGL.cmake` to use the proper definition
 - splits up the source and header files in `windowing/X11/CMakeLists.txt`
 - removes unused and unmaintained `WinSystemX11GLESContext`
 - removes uneeded ifdefs because this is controlled by cmake now.


lrusak:
jenkins build this please

xbmc/windowing/X11/CMakeLists.txt
```diff
@@ -1,23 +1,33 @@
...
-            WinSystemX11.h
             WinSystemX11GLContext.h
-            WinSystemX11GLESContext.h
-            XRandR.h
-            VideoSyncGLX.h
-            VideoSyncDRM.h)
+            WinSystemX11.h
+            WinEventsX11.h)
+
+if(EGL_FOUND)
```
FernetMenta: EGL and friends are required (X11.cmake)
why don't you add it unconditionally?

lrusak: I guess it doesn't really make much sense to break it up as all of the sources are required.

Would it make sense to have the ability to disable glx? This would require minimal ifdefs

FernetMenta: yes, glx is about to die and we want to get rid of it as soon as possible.

lrusak: I'll do that in a separate PR

FernetMenta (approved):

cmake/modules/FindGLX.cmake
```diff
@@ -0,0 +1,44 @@
...
+                          PATHS ${PC_GLX_INCLUDEDIR})
+find_library(GLX_LIBRARY NAMES GL
+                         PATHS ${PC_GLX_LIBDIR})
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(GLX
+                                  REQUIRED_VARS GLX_LIBRARY GLX_INCLUDE_DIR)
+
+if(GLX_FOUND)
+  set(GLX_LIBRARIES ${X_LIBRARY} ${X_EXT_LIBRARY})
```
wsnipex: GLX_LIBRARY

cmake/modules/FindGLX.cmake
```diff
@@ -0,0 +1,44 @@
...
+  set(GLX_INCLUDE_DIRS ${X_INCLUDE_DIR})
```
wsnipex: GLX_INCLUDE_DIR

mac-l1:
hi all,

kodi for x11 + gles and without gl runs fine with me with recent master on my debian 9 linux/X11 rockchip rk3288 device.
this is actually a very decent way to run kodi on linux, x11 on architectures that have a gpu that only support gles but dont support gl (like arm mali gpus) without patching kodi sources (probably to current situation) to make it happen.
so why remove it? can you please reconsider?

regards-
mac-l1

FernetMenta:
use mir or gbm. those platforms work with gles

lrusak:
@mac-l1 unfortunately gles is unmaintained and has fallen behind in terms of x11. I'm surprised if it would even work in current master anyways. On that platform you probably don't have hw decode support anyway so it's probably not worth it.

as fernet said, you should look into using gbm or mir. Both can use vaapi decoding.

Rechi:
`HAS_EGL` definition is now missing on ios and hence the build fails

lrusak:
thanks, I fixed the build by removing the ifdefs

lrusak:
jenkins build this please

lrusak:
jenkins is happy. Would be nice to hear some other opinions before merging.

MartijnKaijser:
@wsnipex @FernetMenta all ok?

PR #12204, opened 2017-05-30
[addons] add callback function to get the MD5 digest of the given text

AlwinEsch:
## Description
Is needed on my test on GUI addon callbacks.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/General.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/General.h
@@ -39,6 +39,7 @@ typedef struct AddonToKodiFuncTable_kodi
   char* (*get_localized_string)(void* kodiBase, long dwCode);
   char* (*get_language)(void* kodiBase, int format, bool region);
   bool (*queue_notification)(void* kodiBase, int type, const char* header, const char* message, const char* imageFile, unsigned int displayTime, bool withSound, unsigned int messageTime);
+  void (*get_md5)(void* kodiBase, const char* text, char* md5);
 } AddonToKodiFuncTable_kodi;
 
 //==============================================================================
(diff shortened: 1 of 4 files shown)


AlwinEsch:
jenkins build this please

ksooo (approved):

akva2:
Kodi is not a complete toolbox for everything needed by addons. kodiplatform was invented for stuff like this as there is no interaction with kodi needed.

AlwinEsch:
@akva2 I understand you but in this case it is acceptable for me to have from Kodi.

Reasons are:
- It is only one function used, in case the whole class for them or something other becomes used makes it also no sense for me and is really better to have in kodiplatform.
- One another reason for me is to have it so simple as possible for addon developer. On Python addons can it be easily used with https://docs.python.org/2/library/md5.html and only add the line `import md5` for C++ is it more complicated with add of external libraries.

AlwinEsch:
The current request are primary to bring in a new way for inputstream addons to allow encrypted streams. But want to have it also predefined as a basic framework for other languages in future.

akva2:
it's a slippery slope and i would hold the principles high in this case. but it's not my call to make.

btw, the idea was to make kodi a consumer of kodiplatform. most generic utils in xbmc/utils would go there. that way all this code is available to kodi and its add-ons, and there is only a single version to maintain.

xbmc/addons/kodi-addon-dev-kit/include/kodi/General.h
```diff
@@ -374,3 +375,35 @@ inline void QueueNotification(QueueMsg type, const std::string& header,
...
+///
+///
+/// ------------------------------------------------------------------------
+///
+/// **Example:**
+/// ~~~~~~~~~~~~~{.cpp}
+/// #include <kodi/General.h>
+/// ...
+/// std::string md5 = kodi::GetMD5("Make me as md5");
+/// fprintf(stderr, "My md5 digest is: '%s'\n", md5.c_str());
```
AchimTuran: It's great that you add examples to the Doxygen documenation.

xbmc/addons/kodi-addon-dev-kit/include/kodi/General.h
```diff
@@ -374,3 +375,35 @@ inline void QueueNotification(QueueMsg type, const std::string& header,
...
+/// ...
+/// ~~~~~~~~~~~~~
+///
+inline std::string GetMD5(const std::string& text)
+{
+  char* md5ret = static_cast<char*>(malloc(40*sizeof(char))); // md5 size normally 32 bytes
```
AchimTuran: You could add a fixed size array here to avoid `malloc` and `free`. This could be done with char `md5ret[40]`.

AchimTuran:
@akva2 
> btw, the idea was to make kodi a consumer of kodiplatform. most generic utils in xbmc/utils would go there. that way all this code is available to kodi and its add-ons, and there is only a single version to maintain.

This sounds like a good way to go. So there would be no effort to maintain the callbacks. But this is not my call too as I don't have the huge overview like you both.
I also think moving this code to kodiplatform could be done later.

ksooo:
What @AchimTuran said.

AlwinEsch:
jenkins build this please

AlwinEsch:
jenkins build this please

PR #12212, opened 2017-06-01
VideoPlayer updates

FernetMenta:
This adds video buffers to be used by decoders and renderers. platforms can register custom buffers like dma.
render formats are history, renderers need to check type of video buffer

platforms register 
- decoders 
- renderers
- processInfo


xbmc/cores/VideoPlayer/Process/ProcessInfo.h
```diff
@@ -110,4 +114,5 @@ class CProcessInfo
   // player states
   CCriticalSection m_stateSection;
   bool m_stateSeeking;
+  std::atomic_int m_levelVQ;
```
garbear: The default constructor of `std::atomic_int` leaves the object in an uninitialized state. Calling GetLevelVQ() before SetLevelVQ() will result in UB. `m_levelVQ` should be initialized by `CProcessInfo` or the UB should be clearly documented.

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/MMALRenderer.h
```diff
@@ -94,8 +94,14 @@ class CMMALRenderer : public CBaseRenderer, public CThread, public IRunnable
...
   bool RenderCapture(CRenderCapture* capture);
 
+<<<<<<< HEAD
   virtual bool         Configure(unsigned int width, unsigned int height, unsigned int d_width, unsigned int d_height, float fps, unsigned flags, ERenderFormat format, void* hwPic, unsigned int orientation);
   virtual int          GetImage(YV12Image *image, int source = AUTOSOURCE, bool readonly = false);
+=======
+  // Player functions
+  virtual bool         Configure(unsigned int width, unsigned int height, unsigned int d_width, unsigned int d_height, float fps, unsigned flags, ERenderFormat format, unsigned extended_format, unsigned int orientation);
+  virtual int          GetImage(YuvImage *image, int source = AUTOSOURCE, bool readonly = false);
+>>>>>>> 50fb06e... VideoPlayer: rename and move YuvImage
```
garbear: Oops, rebase error

xbmc/cores/VideoPlayer/Process/ProcessInfo.h
FernetMenta: > Calling GetLevelVQ() before SetLevelVQ()

that can't happen, even without docu

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/MMALRenderer.h
FernetMenta: intentional, has to be touched anyway

xbmc/cores/VideoPlayer/Process/ProcessInfo.h
garbear: What about if RP doesn't use levels and doesn't call SetLevelVQ()? does the GUI know not to call GetLevelVQ()?

FernetMenta: RP must not use VP's private APIs. I told you many times already. VP takes care that VQ represents actual state of MQ

garbear: is CProcessInfo a private interface? I assumed otherwise because PAPlayer uses it

xbmc/cores/VideoPlayer/VideoPlayer.h
```diff
@@ -402,14 +400,16 @@ class CVideoPlayer : public IPlayer, public CThread, public IVideoPlayer, public
...
+  virtual void OnStartup() override;
+  virtual void OnExit() override;
+  virtual void Process() override;
   virtual void VideoParamsChange() override;
   virtual void GetDebugInfo(std::string &audio, std::string &video, std::string &general) override;
   virtual void UpdateClockSync(bool enabled) override;
   virtual void UpdateRenderInfo(CRenderInfo &info) override;
   virtual void UpdateRenderBuffers(int queued, int discard, int free) override;
+  virtual void UpdareGuiRender(bool gui) override;
+  virtual void UpdareVideoRender(bool video) override;
```
garbear: Small typo

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererVTBGL.cpp
```diff
@@ -56,38 +51,31 @@ CRenderInfo CRendererVTB::GetRenderInfo()
...
+    VTB::CVideoBufferVTB *vb = dynamic_cast<VTB::CVideoBufferVTB*>(buf.videoBuffer);
+    if (vb)
+    {
+      if (vb->m_fence && glIsFenceAPPLE(vb->m_fence))
+      {
+        glDeleteFencesAPPLE(1, &vb->m_fence);
+        vb->m_fence = 0;
+      }
+    }
+    vb->Release();
```
garbear: this should be moved 1 line above, right?

xbmc/cores/VideoPlayer/Process/ProcessInfo.h
FernetMenta: everything not in the top level folder of VideoPlayer is not supposed to be used by other components. CProcessInfo represents the state and behavior of VP that is totally different to the one of RP

xbmc/cores/VideoPlayer/VideoPlayer.h
FernetMenta: yes, already corrected in one of my branches and will update tomorrow

xbmc/cores/VideoPlayer/Process/ProcessInfo.h
garbear: Understood. One of the outcomes of the GSoC project will be to remove all uses of VP code from retroplayer.

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererVTBGL.cpp
FernetMenta: that entire block needs rework. testing that dynamic cast makes no sense in ReleaseBuffer. if it were no vtb type, this would not have been added.

xbmc/cores/VideoPlayer/Process/ProcessInfo.h
FernetMenta: that would be best for now but does not mean that we can't factor out common code at a later time. as long VP's code base has not stabilised I think that is best strategy.

garbear:
I submitted some fixes in #12220 and updated RP in #12221.

xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
```diff
@@ -1645,17 +1611,54 @@ void CVaapiRenderPicture::GLUnMapSurface()
...
 
-VaapiBufferPool::VaapiBufferPool()
+/**
+ * Buffer pool holds allocated vaapi and gl resources
+ * Embedded in COutput
+ */
+class VAAPI::CVaapiBufferPool : public IVideoBufferPool
+{
+public:
+  CVaapiBufferPool(CDecoder &decoder);
```
peak3d: What is CDecoder here? MemoryBuffer is MemoryBuffer and should not rely on any thing like Decoder.
From theory anything could write into a buffer, not only decoders,

peak3d:
From my understanding BufferPools must be part of the platform, not decoder (codec).
RetroPlayer / AddonVideoCodec must not know anything about VAAPI or DMA or whatever.

They should request an buffer from any global place and write to it without knowing whats the underlying buffer-type. Why shouldn't RetoPlayer be able to write into an mmapped VABuffer?

Hopefully I missed something when looking over the code, from (short!!) reading I can only say that the concept does not match the needs I have with AddonVideoCodec (and what I thought is optimal for RetroPlayer).

FernetMenta:
every component can own the bufferpool. it depends on the requiremetns. in case of vaapi, vaapi owns the pool. you are free to register any kind of pool with the bufferPoolManager for use in AddonVideoCodec.

> Why shouldn't RetoPlayer be able to write into an mmapped VABuffer?

you cannot map vaapi surface to system mem. hence RP won't ever be able to write into a vaapi surface.

xbmc/cores/DataCacheCore.h
```diff
@@ -67,6 +67,9 @@ class CDataCacheCore
   void SetStateSeeking(bool active);
   bool IsSeeking();
   bool IsPlayerStateChanged();
+  void SetGuiRender(bool gui);
+  bool GetGuiRender();
+  void SetVideoRender(bool video);
```
Rechi: `bool GetVideoRender();` is missing

FernetMenta: somehow got lost on rebase, fixed now

xbmc/cores/VideoPlayer/Process/VideoBuffer.cpp
```diff
@@ -0,0 +1,386 @@
...
+  return m_pixFormat;
+}
+
+bool CVideoBuffer::CopyPicture(YuvImage* pDst, YuvImage *pSrc)
+{
+  uint8_t *s = pSrc->plane[0];
+  uint8_t *d = pDst->plane[0];
+  int w = pDst->width * pDst->bpp;
+  int h = pDst->height;
+  if ((w == pSrc->stride[0]) && ((unsigned int) pSrc->stride[0] == pDst->stride[0]))
```
Rechi: casting not needed

Rechi: why casting to `unsigned int` here?
`pSrc->stride[1]` and `pDst->stride[1]` have the same type
same is done multiple times below

FernetMenta: I just copied the old code: https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDCodecs/DVDCodecUtils.cpp#L118

right, cast is not needed anymore

FernetMenta: updated, @Rechi  thanks for bringing this up

xbmc/cores/VideoPlayer/Process/VideoBuffer.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+bool CVideoBuffer::CopyYUV422PackedPicture(YuvImage* pDst, YuvImage *pSrc)
+{
+  uint8_t *s = pSrc->plane[0];
+  uint8_t *d = pDst->plane[0];
+  int w = pDst->width;
+  int h = pDst->height;
+
+  // Copy YUYV
+  if ((w * 2 == pSrc->stride[0]) && ((unsigned int) pSrc->stride[0] == pDst->stride[0]))
```
Rechi: @FernetMenta you missed one superfluous cast

afedchin:
windows changes are done (not tested)

FernetMenta:
great, thanks!

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/DXVAHD.h
```diff
@@ -104,6 +105,7 @@ class CProcessorHD : public ID3DResource
   D3D11_VIDEO_PROCESSOR_RATE_CONVERSION_CAPS m_rateCaps;
   D3D11_TEXTURE2D_DESC m_texDesc;
   PROCESSOR_VIEW_TYPE m_eViewType;
+  std::vector<AVPixelFormat> m_formats;
```
FernetMenta: this is obsolete, formates have to be supplied by CProcessInfo::GetRenderFormats

afedchin: how I can decide in CProcessInfo::GetRenderFormats wich format is supported by dxva renderer if I don't know will it be used or not?

afedchin: there is CWinRenderer::GetRenderInfo() where I provide supported formats.

FernetMenta: >  if I don't know will it be used or not

this problem exists in current master.

in this branch a decoder delivers a videobuffer and renderer decides if it can handle it. capabilites don't change at runtime, means this is very static behaviour.  in case it depends on runtime environemnt, you can ask windowing in CProcessInfo::GetRenderFormats

afedchin: what I should ask windowing? is setting enabled && can dxvahd be inited && will HW decoder used?

FernetMenta: m_formats only ever had a single purpose: to tell sw decoder if it needs to convert the output of ffmpeg decoder. a supported format should not depend on any settings. 
nevertheless you can query settings in windows' implementation of CProcessInfo::GetRenderFormats. but if you need that, something goes really wrong.

afedchin: In case of sw decoding and DXVA render method enabled with yuv420p10 video. DXVA may not support yuv420p10 format, actually it is possible to detect supported fornat after initing DXVA renderer
it's not a problem to drop it, but I think it will be a pain

FernetMenta: this is static information you can collect at system start. it won't change during runtime, hence no need to overcomplicate startup of playback. take this as an example: https://github.com/FernetMenta/kodi-agile/blob/videobuffer/xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp#L1195
this is called at startup by windowing. you could do the same with your renderer

afedchin: > it won't change during runtime

different adapters has different caps. I think restart app after changing adapter it more worst idea than do not check caps at all.

FernetMenta: simply call the static method again after adapters have changed. I hooked vaapi/vdpau into RefreshContext: https://github.com/FernetMenta/kodi-agile/blob/videobuffer/xbmc/windowing/X11/WinSystemX11GLContext.cpp#L213

afedchin:
now windows code works as expected (I think)

FernetMenta:
thanks much. I'll rebase and squash your changes end of this week when I'll have vdpau changes ready.

Rechi:
@FernetMenta this needs a rebase

FernetMenta:
@peak3d AddonVideoCodec needs update

FernetMenta:
@popcornmix I introduced a factory for renderers. renderers need to be registered like decoders. see x11, windows, osx for examples

xbmc/cores/VideoPlayer/DVDCodecs/Video/VTB.h
```diff
@@ -19,24 +19,44 @@
  */
 #pragma once
 
-#include "system_gl.h"
+#include <OpenGL/gl.h>
```
Rechi: iOS need `OpenGLES/ES2/gl.h` instead of `OpenGL/gl.h`

FernetMenta: updated

xbmc/cores/VideoPlayer/DVDCodecs/Video/VTB.h
```diff
@@ -19,24 +19,48 @@
  */
 #pragma once
 
-#include "system_gl.h"
+#ifdef HAS_GL
+#include <OpenGL/gl.h>
+#else
+#include <OpenGLES/ES2/gl.h>
+#endif
```
Rechi: this change isn't enough, have a look at https://github.com/xbmc/xbmc/blob/f743b10/xbmc/system.h#L189-L193
system.h undefines `HAS_GL` for iOS

FernetMenta:
@afedchin @popcornmix @peak3d the concept of having platforms register their components is now also applied to ProcessInfo 5a2d60e

FernetMenta:
@Memphiz ios added. I did a quick test with vtb and sw decoding

Memphiz:
Thx for doing all the Darwin work!

FernetMenta:
jenkins build this please

MartijnKaijser:
@popcornmix rpi fails

popcornmix:
Yep, I'll PR a fix

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

MartijnKaijser:
fyi playback seems fine on android

FernetMenta:
@MartijnKaijser thanks. Do we have volunteers for runtime tests on Windows?

afedchin:
I've tested very quickly while adapting windows changes.

FernetMenta:
@afedchin could you please review 1b5e4966ebb66215be2228e172026835ee42adc3

xbmc/cores/VideoPlayer/Process/ProcessInfo.cpp
```diff
@@ -19,29 +19,46 @@
...
+  CSingleLock lock(createSection);
+  
+  CProcessInfo *ret = nullptr;
+  for (auto &info : m_processControls)
+  {
+    ret = info.second();
+    if (ret)
+      return ret;
+  }
+  if (!ret)
```
Rechi: why doing this check here?
This produces a warning that the function end may be reached without a `CProcessInfo*` return.

FernetMenta: right, I'll drop it

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

MartijnKaijser:
Last I build rpi didn't finish. http://jenkins.kodi.tv/job/LINUX-RBPI/4883/

FernetMenta:
how can we make jenkins build this again?

MartijnKaijser:
After removing the label it needs a while to adapt. 
jenkins build this please

FernetMenta:
jenkins build this please

FernetMenta:
as far as I am concered this is ready to go now. is anybody aware of remaining issues that should be addressed prior to merge?

xbmc/cores/VideoPlayer/DVDDemuxers/DemuxMultiSource.h
```diff
@@ -45,22 +45,21 @@ class CDemuxMultiSource : public CDVDDemux
...
+
+  bool Open(CDVDInputStream* pInput);
+
+  // implementation of CDVDDemux
   void Abort() override;
   void EnableStream(int64_t demuxerId, int id, bool enable) override;
   void Flush() override;
-  std::string GetFileName() override { return ""; };
   int GetNrOfStreams() const override;
-  CDemuxStream* GetStream(int iStreamId)const override { return nullptr; } ;
```
garbear: Uh oh, this doesn't look good. In the superclass this is pure-virtual. I think I moved this to `protected` access but it got lost in the rebase. I'll try running a test build and see if I'm correct.

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

MartijnKaijser:
let's get it in

FernetMenta:
@lrusak wanted to submit gbm updates in the morning (his timezone)

lrusak:
gbm works fine with vaapi for me :+1:

FernetMenta:
jenkins build this please

FernetMenta:
win64 error unrelated.

dapharsyde:
Sorry I know I'm late to the party, but I just discovered that this merge seems to be the cause of stuttering 1080p live TV streams.  720p is fine.  I have confirmed it is not network related, this issue is only present when I upgrade my build from version #0709 to #0710.  I am using VDPAU acceleration, in case that is relevant.   Can anyone help?

popcornmix:
I imagine creating a [trac](https://trac.kodi.tv/) ticket would have a better chance of a response than posting on an old github PR.

PR #12213, opened 2017-06-01
Improve GUI message handling / gui::smartredraw

peak3d:
GUI optimizations (Message handling through lookup) and SmartRedraw (only Processes GUI if anything has changed).

## Description
1.) Use alreaedy existing gui-control-lookup for sending control messages.
2.) [Experimental]: If advancedsettings::gui:guismartredraw is set to true, the GUI Process call (and therefore the render call), wich is usually called on every vsync, is only done if anything is changed.

## Motivation and Context
Speed up GUI / Remove load if kodi idle / plays videos (factor ~10)

## How Has This Been Tested?
Win32 / Android / Esturay

## Types of change

## Checklist:

##Note
I'm pushing this PR quite early (regarding smartredraw) because @mkortstiege seems to play around with GUI.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -4594,6 +4594,8 @@ void CApplication::ProcessSlow()
 
   m_ServiceManager->GetActiveAE().GarbageCollect();
 
+  RefreshControls();
+
   // if we don't render the gui there's no reason to start the screensaver.
   // that way the screensaver won't kick in if we maximize the XBMC window
   // after the screensaver start time.
@@ -5230,6 +5232,12 @@ void CApplication::CloseNetworkShares()
     VFSEntryPtr vfs = std::static_pointer_cast<CVFSEntry>(addon);
     vfs->DisconnectAll();
   }
+}
+
+void CApplication::RefreshControls()
+{
+  CGUIMessage msg(GUI_MSG_REFRESH_TIMER, 0, 0, 0);
+  g_windowManager.SendThreadMessage(msg);
 }
 
 void CApplication::RegisterActionListener(IActionListener *listener)
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -33,7 +33,6 @@
 #include "Util.h"
 #include "URL.h"
 #include "guilib/TextureManager.h"
-#include "cores/DataCacheCore.h"
 #include "cores/IPlayer.h"
 #include "cores/VideoPlayer/DVDFileInfo.h"
 #include "cores/AudioEngine/Engines/ActiveAE/AudioDSPAddons/ActiveAEDSP.h"
@@ -4595,7 +4594,7 @@ void CApplication::ProcessSlow()
 
   m_ServiceManager->GetActiveAE().GarbageCollect();
 
-  SendGUIMessage(GUI_MSG_REFRESH_TIMER);
+  g_windowManager.SendMessage(GUI_MSG_REFRESH_TIMER,0,0);
 
   // if we don't render the gui there's no reason to start the screensaver.
   // that way the screensaver won't kick in if we maximize the XBMC window
@@ -5233,12 +5232,6 @@ void CApplication::CloseNetworkShares()
     VFSEntryPtr vfs = std::static_pointer_cast<CVFSEntry>(addon);
     vfs->DisconnectAll();
   }
-}
-
-void CApplication::SendGUIMessage(int dwMsg, int senderID/* = 0*/, int controlID/* = 0*/, int param1/* = 0*/, int param2/* = 0*/)
-{
-  CGUIMessage msg(dwMsg, senderID, controlID, param1, param2);
-  g_windowManager.SendThreadMessage(msg);
 }
 
 void CApplication::RegisterActionListener(IActionListener *listener)
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -1909,13 +1909,13 @@ void CApplication::Render()
 
   g_Windowing.EndRender();
 
-  // reset our info cache - we do this at the end of Render so that it is
-  // fresh for the next process(), or after a windowclose animation (where process()
-  // isn't called)
-  g_infoManager.ResetCache();
-
   if (hasRendered)
   {
+    // reset our info cache - we do this at the end of Render so that it is
+    // fresh for the next process(), or after a windowclose animation (where process()
+    // isn't called)
+    g_infoManager.ResetCache();
+
     g_infoManager.UpdateFPS();
   }
 
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -1909,13 +1909,13 @@ void CApplication::Render()
 
   g_Windowing.EndRender();
 
+  // reset our info cache - we do this at the end of Render so that it is
+  // fresh for the next process(), or after a windowclose animation (where process()
+  // isn't called)
+  g_infoManager.ResetCache();
+
   if (hasRendered)
   {
-    // reset our info cache - we do this at the end of Render so that it is
-    // fresh for the next process(), or after a windowclose animation (where process()
-    // isn't called)
-    g_infoManager.ResetCache();
-
     g_infoManager.UpdateFPS();
   }
 
@@ -4594,7 +4594,7 @@ void CApplication::ProcessSlow()
 
   m_ServiceManager->GetActiveAE().GarbageCollect();
 
-  RefreshControls();
+  SendGUIMessage(GUI_MSG_REFRESH_TIMER);
 
   // if we don't render the gui there's no reason to start the screensaver.
   // that way the screensaver won't kick in if we maximize the XBMC window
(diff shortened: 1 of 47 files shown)


xbmc/guilib/GUIWindowManager.cpp
```diff
@@ -1148,7 +1148,12 @@ void CGUIWindowManager::AfterRender()
   for (const auto& window : activeDialogs)
   {
     if (window->IsDialogRunning())
+    {
       window->AfterRender();
+      // Dialog state can affect visibility states
+      if (pWindow && window->IsControlDirty())
```
Rechi: `class CGUIWindow` has no member named `IsControlDirty`

xbmc/cores/DataCacheCore.cpp
```diff
@@ -235,3 +237,13 @@ bool CDataCacheCore::CDataCacheCore::IsSeeking()
 
   return m_stateInfo.m_stateSeeking;
 }
+
+bool CDataCacheCore::PlayStateChanged()
```
Rechi: definition in `DataCacheCore.h` is missing

xbmc/cores/DataCacheCore.cpp
```diff
@@ -23,11 +23,12 @@
 #include "ServiceBroker.h"
 
 CDataCacheCore::CDataCacheCore()
+  :m_playStateChanged(false)
```
FernetMenta: init by assignment in header

xbmc/guilib/GUIWindowManager.cpp
```diff
@@ -1174,6 +1175,9 @@ void CGUIWindowManager::FrameMove()
     m_deleteWindows.clear();
   }
 
+  if (CDataCacheCore::GetInstance().PlayStateChanged())
```
FernetMenta: CApplicationPlayer::FrameMove is a much better place to check this. That method also knows if a player is active

xbmc/cores/DataCacheCore.cpp
FernetMenta: please rename to PlayerStateChenged

xbmc/cores/DataCacheCore.cpp
```diff
@@ -227,6 +227,7 @@ void CDataCacheCore::SetStateSeeking(bool active)
   CSingleLock lock(m_stateSection);
 
   m_stateInfo.m_stateSeeking = active;
+  m_playStateChanged = true;
```
Rechi: `m_playStateChanged` was not declared in this scope

Rechi: `m_playerStateChanged`

peak3d:
jenkins build this please

peak3d:
Any objections @FernetMenta ? Its approved from popcornmix (the last commit) and the others are running quite some days on my home equipment.

xbmc/Application.cpp
```diff
@@ -33,6 +33,7 @@
 #include "Util.h"
 #include "URL.h"
 #include "guilib/TextureManager.h"
+#include "cores/DataCacheCore.h"
```
FernetMenta: left over from previous attempt

xbmc/Application.cpp
```diff
@@ -5232,6 +5235,12 @@ void CApplication::CloseNetworkShares()
   }
 }
 
+void CApplication::SendGUIMessage(int dwMsg, int senderID/* = 0*/, int controlID/* = 0*/, int param1/* = 0*/, int param2/* = 0*/)
```
FernetMenta: https://github.com/xbmc/xbmc/blob/master/xbmc/messaging/ApplicationMessenger.cpp#L272

please elaborate why you create an additionsl SendGUIMessage

xbmc/dialogs/GUIDialogSeekBar.cpp
```diff
@@ -30,6 +30,7 @@
 
 CGUIDialogSeekBar::CGUIDialogSeekBar(void)
   : CGUIDialog(WINDOW_DIALOG_SEEK_BAR, "DialogSeekBar.xml", DialogModalityType::MODELESS)
+  , m_lastPercent(~0)
```
FernetMenta: please init by assignment in header

peak3d:
jenkins build this please

peak3d:
Because you build the wrong ref: jenkins build this please

MartijnKaijser:
jenkins build this please

HitcherUK:
This could be the cause of problems with scrollbars and dialog grouplist buttons not fading when not focused. https://forum.kodi.tv/showthread.php?tid=316560

HitcherUK:
@peak3d I can confirm that this is the cause of the grouplist problems mentioned [here](https://forum.kodi.tv/showthread.php?tid=316560&pid=2604968#pid2604968) and reported [here](https://trac.kodi.tv/ticket/17504). This was tested using a [build](http://mirrors.kodi.tv/test-builds/windows/win32/KodiSetup-20170605-a16e0da-PR12213exc-x86.exe) before this commit and a [build ](http://mirrors.kodi.tv/test-builds/windows/win32/KodiSetup-20170606-40a1a0c513-PR12213inc-x86.exe)after.

Thanks.

pkerling:
@peak3d I was seeing another (i.e. not yet reported as far as I know) strange behavior in Kodi for some time and finally got around to bisecting it. Commit 40a1a0c5 in this PR introduces it:
You can see it by opening e.g. the system settings dialog and then move the mouse around at the upper left edge where it says "Settings / System". The currently selected category item will randomly flicker (it did not before). It also flickers sometimes when you move the mouse over the individual settings on the right side.

peak3d:
thanks @pkerling for the report, yes, I have noticed this one as well, but hoped that nobody else will see it (joking). The selection / focus things are a little bit messed up in GUI, I have one more issue and will focus on it next week (unless you already have the reason for it (???) :-)

pkerling:
@peak3d No I didn't look further into it, only bisected

xbmc/Application.cpp
```diff
@@ -4594,6 +4594,8 @@ void CApplication::ProcessSlow()
 
   m_ServiceManager->GetActiveAE().GarbageCollect();
 
+  g_windowManager.SendMessage(GUI_MSG_REFRESH_TIMER,0,0);
```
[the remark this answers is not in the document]
ksooo: no idea what this is for, pvr does not use  this message (afaik)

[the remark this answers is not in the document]
FernetMenta: ok, I'll move it to FrameMove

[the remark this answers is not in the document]
peak3d: This message triggers a "change detection" in GUI in smartredraw mode.
It should NOT be done on every FrameMove, it should be done on reasonable time.

Example is the GUI clock wich forces GUI actions on each frame, 500ms between the refresh (ProcessSlow) is quite fine, so FrameMove is not the best option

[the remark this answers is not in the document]
FernetMenta: already done

JimmyS83:
_If advancedsettings::gui:guismartredraw is set to true, the GUI Process call (and therefore the render call), wich is usually called on every vsync, is only done if anything is changed._

Does it means, when black screensaver is active, then there is no render at all and CPU load is reduced?

peak3d:
yes, schould be. There is still load because of other threads (network e.g.), but CPU load is in general very low.

JimmyS83:
> 
> 
> yes, schould be. There is still load because of other threads (network e.g.), but CPU load is in general very low.

Just start testing this feature.. I think CPU is still [higher than its needed](https://forum.kodi.tv/showthread.php?tid=338742&pid=2803158#pid2803158) - compare with Jarvis release, and I have the same network / configuration / addons...  But at least I see some CPU improvement with smart redraw (I didnt do any actual user testing, so cant comment more now).

peak3d:
There are still some GUI elements which don't draw / update correctly.
If benefit is not high for you (for me on low powered Android devices it is), than its better to disable.

All this depends on what addons - especially service addons - do in background. I have no influence on this.

JimmyS83:
Any lowering CPU is benefit for me. The Rpi / C2 boxes running 24/7, and actually get used 1-4h per day.

I am almost sure that no addons causes this difference, because I use just a few of them, and all of them (and also configuration of services like Samba, CEC, etc. remains exactly same) already used before on Rpi2+Jarvis with 1-3% CPU..

JimmyS83:
Is this good platform to report bugs? I didnt find forum thread, so if its not bother you here:
- Confluence skin, little window showing library update progress doesnt dissappears, after library update.

MartijnKaijser:
@JimmyS83 no it's not. Use the correct way and create a github ticket with enough info. Do not start asking questions in a very old PR

zaphod24:
What is the recommended setting for dirty regions when smartredraw is set to true? There seem to be multiple opinions so I figured I’d ask here. Unless something has changed in Leia, the default for dirty regions is still 3. Should it be set to zero to disable dirty regions? Should only smartredraw=true be specified leaving dirty regions defaulted to 3?

For reference I’m running Leia on a c2

PR #11080, opened 2016-12-06
log: Don't force logging with a debug build

popcornmix:
## Description
Don't force logging with a debug build

## Motivation and Context
A debug build is to allow use of a debugger and shouldn't be tied to logging level.
There are perfectly good ways of enabling debug logging anyway.

## Types of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [X] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #12215, opened 2017-06-02
Posix/Win32 main refactor

yol:
Follow-up from #12207 

## Description
Remove TARGET_POSIX ifdefs from posix/main.cpp and make win32 platform independent of posix/main.cpp

## Motivation and Context
WinMain() currently uses posix/main.cpp, which is unexpected and unnecessary.

## How Has This Been Tested?
Linux x86_64 works, other builds could fail - will update according to Jenkins results

## Types of change

## Checklist:

### DIFF
--- a/cmake/scripts/windows/ArchSetup.cmake
+++ b/cmake/scripts/windows/ArchSetup.cmake
@@ -13,6 +13,8 @@ endif()
 
 set(PLATFORM_DIR platform/win32)
 
+set(CORE_MAIN_SOURCE ${CMAKE_SOURCE_DIR}/xbmc/platform/win32/WinMain.cpp)
+
 # Precompiled headers fail with per target output directory. (needs CMake 3.1)
 set(PRECOMPILEDHEADER_DIR ${PROJECT_BINARY_DIR}/${CORE_BUILD_CONFIG}/objs)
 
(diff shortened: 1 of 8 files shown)


pkerling:
jenkins build this please

MartijnKaijser:
Only authorized people can trigger jenkins manually ;) 
You can do this by simply repushing to a branch.

pkerling:
It did trigger the build though? Or was that coincidence?
And I can't repush since nothing has changed. Jenkins just failed the build due to internal issues previously.

fritsch:
jenkins is happy - so all fine here.

FernetMenta (approved):
looks ok to me

Rechi:
In #11080 set log level to debug logging for debug builds was disabled, so I'm not sure if we should change it back again.

MartijnKaijser:
Indeed debuglogging shouldn't be auto enabled.

pkerling:
Ah didn't know about #11080 - I just assumed that forcing DEBUG was the original intention of the broken code. But seems it was just leftovers. Removing the #ifdef _DEBUG in xbmc.cpp should fix it then.

Rechi:
This was just forgotten in #11080 because it didn't wok as expected.

pkerling:
Related code removed from xbmc.cpp

MartijnKaijser:
No objections?

PR #12218, opened 2017-06-02
fix in-tree add-on build environment

notspiff:
The only convenient way to develop add-ons was broken. Since all my add-ons crash atm, I need this to be able to debug.


hudokkow:
Thx!

PR #12219, opened 2017-06-02
[X11] factor out glx support

lrusak:
GLX is almost dead. Let's allow building without it.

If someone has a better solution to the overrides I'd like to hear it as I don't really like adding the two header files to `GLContext.h`.

Build tested only. Could use a runtime test on intel and nvidia hardware with and with `-DENABLE_GLX=OFF`


FernetMenta (approved):
vdpau requires glx but that dies as well so should not matter much.

hudokkow:
`-DENABLE_GLX=OFF`: Crash on start and no cigar using i7-7700HQ with 630 GPU -> https://pastebin.com/eMJg2xPe

`-DENABLE_GLX=ON`: Doesn't crash. Testing now with a few videos.
Will test with 1050Ti tomorrow.

lrusak:
@hudokkow without glx (egl only) will only work on intel graphics.

xbmc/windowing/X11/WinSystemX11GLContext.cpp
```diff
@@ -175,18 +180,25 @@ bool CWinSystemX11GLContext::DestroyWindow()
...
+  }
+
+#ifdef HAS_GLX
+  std::string gpuvendor;
+  const char* vend = (const char*) glGetString(GL_VENDOR);
+  if (vend)
+    gpuvendor = vend;
+
+  std::transform(gpuvendor.begin(), gpuvendor.end(), gpuvendor.begin(), ::tolower);
+  if (gpuvendor.compare(0, 5, "intel") != 0)
```
wsnipex: what about AMD?

xbmc/windowing/X11/WinSystemX11GLContext.cpp
```diff
@@ -175,18 +180,25 @@ bool CWinSystemX11GLContext::DestroyWindow()
...
-    GLX_ALPHA_SIZE, 8,
-    GLX_DEPTH_SIZE, 24,
-    GLX_DOUBLEBUFFER,
-    None
-  };
-  return glXChooseVisual(m_dpy, m_nScreen, att);
+    m_pGLContext = new CGLContextEGL(m_dpy);
+  }
+
+#ifdef HAS_GLX
```
FernetMenta: @lrusak this entire block should be enclosed by if(!m_pGLContext)
currently it kills the context every time we ask for a visual

xbmc/windowing/X11/WinSystemX11GLContext.cpp
lrusak: amd can only use vdpau, so glx is the only option for them.

xbmc/windowing/X11/WinSystemX11GLContext.cpp
lrusak: yes, that sounds right

fritsch:
Would be time to tell Christian and Alex of AMD. I think it's time to go with VAAPI and use the upcoming Videobuffers work to get AMD specifics into the rendering part.

lrusak:
@fritsch sounds good. Do you want to lead the conversation as I do not know the specifics of what is required.

fritsch:
@lrusak it's okay. Fernet and I know these guys since day one of VDPAU on AMD - I will write an email to christian tomorrow and put you on cc.

FernetMenta:
Both of you were on the distribution list of the mail thread with Christian. I don't think anything has changed since last info of March, 18th.

FernetMenta:
you should add some comment that currently GLX can't be disabled when building for X11. libGL.so does include GLX symbols, one of the reasons, Weston (wayland) uses GLES.
nevertheless this PR makes it clear where kodi depends on GLX.

MartijnKaijser:
@lrusak needs a rebase

lrusak:
Updated and tested with and without vdpau. This is worth a review

FernetMenta:
sometimes GLX was used instead of X11. https://github.com/xbmc/xbmc/search?utf8=%E2%9C%93&q=HAS_GLX&type=
this needs to be corrected or some things will break

lrusak:
pushed a commit to change HAS_GLX to HAVE_X11

FernetMenta:
+1

xbmc/windowing/X11/WinSystemX11GLContext.cpp
```diff
@@ -177,18 +183,14 @@ bool CWinSystemX11GLContext::DestroyWindow()
...
-    GLX_RED_SIZE, 8,
-    GLX_GREEN_SIZE, 8,
-    GLX_BLUE_SIZE, 8,
-    GLX_ALPHA_SIZE, 8,
-    GLX_DEPTH_SIZE, 24,
-    GLX_DOUBLEBUFFER,
-    None
-  };
-  return glXChooseVisual(m_dpy, m_nScreen, att);
+    RefreshGLContext(true);
```
FernetMenta: this can't work. visual is required when the window is created. creating the context requires the window

FernetMenta (changes requested):
see my comment

xbmc/windowing/X11/WinSystemX11GLContext.cpp
FernetMenta: you need to use XGetVisualInfo instead of any EGL method

lrusak:
fixed up, it's a little ugly but it works on GL with and without GLX enabled (mirroring if VDPAU is enabled or not).

FernetMenta:
It is still wrong and I won't accept it this way. A context must not be created before the window. You have to use XLib for getting the visual.

lrusak:
@FernetMenta apologies, I did not see your comment about XGetVisualInfo. I have adapted the code to use this and it works as expected with and without GLX enabled.

xbmc/windowing/X11/WinSystemX11GLContext.cpp
```diff
@@ -177,18 +183,12 @@ bool CWinSystemX11GLContext::DestroyWindow()
...
-    GLX_ALPHA_SIZE, 8,
-    GLX_DEPTH_SIZE, 24,
-    GLX_DOUBLEBUFFER,
-    None
-  };
-  return glXChooseVisual(m_dpy, m_nScreen, att);
+  int count;
+  XVisualInfo* visuals = nullptr;
+
+  visuals = XGetVisualInfo(m_dpy, 0, NULL, &count);
```
FernetMenta: you have to select a suitable visual here. create a template, then iterate through the number of returned and select the best matching. use XFree when finished
ref: https://www.x.org/archive/X11R7.5/doc/man/man3/XVisualInfo.3.html

lrusak: Is that really needed? after the context is created a new suitable visual is selected either for glx or egl. 

https://github.com/xbmc/xbmc/blob/master/xbmc/windowing/X11/GLContextEGL.cpp#L266
https://github.com/xbmc/xbmc/blob/master/xbmc/windowing/X11/GLContextGLX.cpp#L142

FernetMenta: I don't think you understand the situation correctly. visual is requred for creating the window. this is before any gl context is created. 
When we create the GL/EGL context, we just verify if the visual is suitable.

lrusak: I don't think that's correct. It looks to my like it will try to find and use a suitable visual for each context.
https://github.com/xbmc/xbmc/blob/master/xbmc/windowing/X11/GLContextEGL.cpp#L89-L136
https://github.com/xbmc/xbmc/blob/master/xbmc/windowing/X11/GLContextGLX.cpp#L52-L128

FernetMenta: Right, visual is used to get i.e. eglConfig but the visual used here should match the visual used for creating the window. 
```
For EGL window and pixmap surfaces, a suitable native window or pixmap with a matching native visual must be created first.
```
https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglIntro.xhtml

We don't want to make this working by chance if the first visual returned matches. Further I am not sure if there are any internal convertions required if both visuals do not match.

lrusak: what about using XMatchVisualInfo instead (https://tronche.com/gui/x/xlib/utilities/XMatchVisualInfo.html)

something like
```
  XVisualInfo *visual = new XVisualInfo();
  XMatchVisualInfo(m_dpy, m_nScreen, 24, TrueColor, visual);
  return visual;
```
This will match a visual with the correct depth and color that we want.

FernetMenta: this is not what we want. we want exaxtly the same parameters we use for glx/egl. the solution is rather simple: submit a template with required parameters to XGetVisualInfo

lrusak:
@FernetMenta is that better? Is there other values we would like to use in the template?

FernetMenta:
@lrusak yes, but don't forget to free allocated resources with XFree

candrews:
Can you please also update the documentation in docs/README.* to reflect the fact this dependency no longer exists?

lrusak:
@FernetMenta I'm honestly not sure what needs freeing using xfree?

FernetMenta:
sorry, it is done in SetWindow already.

lrusak:
@FernetMenta I've rebased and squashed. Please test this is you can so we can get it merged.

@candrews no need to adjust the docs. glx is still required when using VDPAU.

PR #12220, opened 2017-06-03
VideoPlayer: Fix missing include and virtual destructors

garbear:
Just some improvements I came across while adapting RetroPlayer to the new video buffers.


FernetMenta:
thanks

PR #12221, opened 2017-06-03
RetroPlayer: Update for video buffers

garbear:
This adapts RetroPlayer to use the new video buffers. This code is temporary and will be removed when RetroPlayer's rendering is rewritten to not use private VP interfaces.


garbear:
RPi is still broken, I've started working on a fix for this

PR #12222, opened 2017-06-03
Fix reversed condition

notspiff:
missing a !


AlwinEsch:
Thought that I fixed it, thanks a lot and Sorry!

AlwinEsch:
jenkins build this please

garbear:
Alwin, you're fired!

But seriously, thanks a lot for undertaking this insanely massive project.

akva2:
No worries alwin shit will break and then we fix it. It might be my previous pr description came off wrong. I didnt sulk that things broke, i just meant that if it was a temporary measure, it was a bad thing to have broken even for a while as i rely on that during dev, was all.

PR #12223, opened 2017-06-03
Games: Fix crash due to circular dependency

garbear:
## Description
If Kodi is aborted during startup, we can segfault while cleaning up. I believe this is only an issue if Kodi fails to start fully, but the circular dependency should be fixed nonetheless.

## Types of change


garbear:
jenkins build this please

PR #12225, opened 2017-06-03
Add CORS headers to HTTP web server

jaylinski:
## Description
This PR adds the CORS headers

* `Access-Control-Allow-Credentials: true`
* `Access-Control-Allow-Origin: *`

to every `MHD_OK` and `AskForAuthentication` response from the HTTP web server.

## Motivation and Context
This change enables usage of the new [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) which requires CORS headers to be present.

Code example:
```js
var headers = new Headers({
  "Authorization": "Basic " + window.btoa("kodi:kodi"),
  "Accept": "application/json",
  "Content-Type": "application/json;charset=UTF-8"
});

var params = {
  method: "POST",
  credentials: "include",
  headers: headers,
  mode: "cors",
  body: '{"foo": "bar"}'
};

fetch("http://kodi:8080/jsonrpc", params).then(function(response) {
  console.log(response);
});
```

### Long description

I'm developing a web-extension which talks to Kodi via HTTP. (Just like https://github.com/khloke/play-to-xbmc-chrome).

I could prevent cross-origin errors by requesting permissions to talk to all websites (see https://github.com/khloke/play-to-xbmc-chrome/blob/master/manifest.json#L12), but due to security concers I want to avoid this.

By adding CORS headers to Kodi, I can make cross-origin requests to Kodi and use the HTTP API without having to use a cross-origin workaround.

### Security concerns

Assuming a user enables remote control via HTTP and **does not** set a password, a malicious website could send cross-site requests to Kodi without the user noticing.    
But: in order for the attacker to succeed, he must know the IP and port of the Kodi HTTP server.

In order to mitigate this issue, a `Enable CORS` setting could be introduced.

## How Has This Been Tested?
I tested this change by sending fetch requests to Kodi (see code example above) and making `curl` requests to Kodi.

### Unauthenticated response
```
$ curl -v http://192.168.99.100:8080/jsonrpc
*   Trying 192.168.99.100...
* Connected to 192.168.99.100 (192.168.99.100) port 8080 (#0)
> GET /jsonrpc HTTP/1.1
> Host: 192.168.99.100:8080
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 401 Unauthorized
< Content-Length: 0
< WWW-Authenticate: Basic realm="XBMC"
< Access-Control-Allow-Credentials: true
< Access-Control-Allow-Origin: *
< Connection: close
< Date: Sat, 03 Jun 2017 11:51:05 GMT
< 
* Closing connection 0
```

### Authenticated response
``` 
curl -v -o /dev/null http://kodi:kodi@192.168.99.100:8080/jsonrpc
*   Trying 192.168.99.100...
* Connected to 192.168.99.100 (192.168.99.100) port 8080 (#0)
* Server auth using Basic with user 'kodi'
> GET /jsonrpc HTTP/1.1
> Host: 192.168.99.100:8080
> Authorization: Basic a29kaTprb2Rp
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Connection: Keep-Alive
< Content-Type: application/json
< Content-Length: 232629
< Cache-Control: private, max-age=0, no-cache
< Accept-Ranges: none
< Access-Control-Allow-Credentials: true
< Access-Control-Allow-Origin: *
< Date: Sat, 03 Jun 2017 12:23:21 GMT
```

I also successfully ran the test suite.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/network/WebServer.cpp
+++ b/xbmc/network/WebServer.cpp
@@ -134,6 +134,9 @@ int CWebServer::AskForAuthentication(HTTPRequest request) const
 
   LogResponse(request, MHD_HTTP_UNAUTHORIZED);
 
+  if (HEADER_CORS_ENABLED)
+    EnableCors(response);
+
   ret = MHD_queue_basic_auth_fail_response(request.connection, "XBMC", response);
   MHD_destroy_response(response);
 
@@ -452,14 +455,13 @@ int CWebServer::FinalizeRequest(const std::shared_ptr<IHTTPRequestHandler>& hand
   if (responseDetails.totalLength > 0)
     handler->AddResponseHeader(MHD_HTTP_HEADER_CONTENT_LENGTH, StringUtils::Format("%" PRIu64, responseDetails.totalLength));
 
+  // add CORS headers
+  if (HEADER_CORS_ENABLED)
+    EnableCors(response);
+
   // add all headers set by the request handler
   for (std::multimap<std::string, std::string>::const_iterator it = responseDetails.headers.begin(); it != responseDetails.headers.end(); ++it)
     AddHeader(response, it->first, it->second);
-
-  // add CORS headers
-  if (HEADER_CORS_ENABLED)
-    AddHeader(response, MHD_HTTP_HEADER_ACCESS_CONTROL_ALLOW_ORIGIN, "*");
-    AddHeader(response, "Access-Control-Allow-Credentials", "true");
 
   return SendResponse(request, responseStatus, response);
 }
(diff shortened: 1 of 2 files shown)


tamland:
>Assuming a user enables remote control via HTTP and does not set a password, a malicious website could send cross-site requests to Kodi without the user noticing.
But: in order for the attacker to succeed, he must know the IP and port of the Kodi HTTP server.
>
>In order to mitigate this issue, a Enable CORS setting could be introduced.

That's trivial to guess/brute force.. This should at the very most be limited to localhost/whatever the origin for webextensions are or an explicit whitelist as uses are definitely not going to understand the implication of an 'Enable CORS' setting.

Montellese:
Thanks for looking into this @jaylinski. I have actually been looking into CORS support myself a few weeks ago, see 78efe9739832bdf84dc8193207cf9fcecdab919e. It is more complete than your approach due to the following:
* it supports the HTTP `OPTIONS` request method which can be used by clients to figure out if the server supports CORS at all for specific URLs and specific request methods.
* it has a setting to enable/disable CORS which IMO is a must have because there's a reason CORS is disabled by default in every webserver (as @tamland pointed out as well)
* it has a setting to specify the server/host name to be able to differentiate between non-CORS (anything from the same server/host) and CORS requests
* it supports additional CORS related HTTP headers
* it never returns `Access-Control-Allow-Origin: *` but always a value specific to the request's `Origin`

Unfortunately it doesn't support user defined lists of origins yet which means that it basically accepts CORS requests from any origin. That's why I haven't PRed it yet. I'm not sure how best to implement the origin whitelist.

jaylinski:
@Montellese Perfect, your commit looks much better. :+1: 

Implementing an origin whitelist could solve the security issue (although I'm not a big fan of complex configurations).   
Why not just force the user to set a (secure) password in order to enable CORS?

Would be nice if you could open a WIP PR with your current implementation, so we can discuss this further.

I'm closing my PR in favor of your commit 78efe97.

PR #12228, opened 2017-06-04
[add-ons/settings] add python setting getters/setters for bool/int/number/string

Montellese:
This is a follow-up of #12125 which adds typed getters and setters for bool/int/number/string settings. The difference to the existing getter/setter is that when a plugin e.g. tries to retrieve an integer setting using `getSettingString()` it will fail because the types don't match.

## How Has This Been Tested?
I've adjusted @ronie's `service.scrobbler.lastfm` to use the new methods and it still worked.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/Addon.cpp
+++ b/xbmc/addons/Addon.cpp
@@ -319,6 +319,40 @@ std::string CAddon::GetSetting(const std::string& key)
   return "";
 }
 
+template<class TSetting>
+bool GetSettingValue(CAddon& addon, const std::string& key, typename TSetting::Value& value)
+{
+  if (key.empty() || !addon.HasSettings())
+    return false;
+
+  auto setting = addon.GetSettings()->GetSetting(key);
+  if (setting == nullptr || setting->GetType() != TSetting::Type())
+    return false;
+
+  value = std::static_pointer_cast<TSetting>(setting)->GetValue();
+  return true;
+}
+
+bool CAddon::GetSettingBool(const std::string& key, bool& value)
+{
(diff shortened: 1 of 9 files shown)


phil65 (approved):

romanvm:
Looks nice. Just one question though: how is attempt to get/set an invalid type handled? The Pythonic way is to throw `TypeError` exception on an invalid type or call `PyErr_SetString(PyExc_TypeError, "<some exception message here>")` on C level.

MartijnKaijser:
Thinking about this why would we need to specify the type when doing get? Understandable you need to set the correct type when writing a setting but for retrieving we already know what was store correct?

romanvm:
@MartijnKaijser Foolpoofing against accidental getter/setter mismatch?

Montellese:
@romanvm: Right now it returns a default value but I could change it to throw an error/exception (just need to check how). Is this preferred?

@MartijnKaijser AFAIK our C++ Python API doesn't allow us to return any type so we'd have to go back to returning a string and then the add-on script has to parse the actual type and value from a string which IMO doesn't make any sense.

romanvm:
@Montellese Yes, an exception would be better.  "Explicit is better than implicit". Of course, if it's not too difficult. Kodi's Python <-> C interoperability layer is a complete mystery to me.

Montellese:
@romanvm It actually wasn't that hard. The necessary exception and translation to `PyExc_TypeError` was already in place. Now all the new getters and setters throw a `TypeError` if a script tries to access a setting of a wrong type.

PR #12230, opened 2017-06-04
[addons] add new CAddonInfo class

AlwinEsch:
## Description
This add the class who becomes extended step by step to replace cpluff.

Further replace this in future the complete AddonProps structure.

Currently include it only the static functions and table who was before
on other place.

This is done on this commit and request to hold coming request smaller
and prevent parts who not direct related to changes.

As next requests comes the basic read of addon.xml and to go over there.

## Types of change

## Checklist:


AlwinEsch:
Here a example how it can look on end https://github.com/FernetMenta/kodi-agile/blob/master/xbmc/addons/AddonInfo.cpp (need some cleanup ;) )

tamland:
Do we really need a *third* addon info thing in addition to CAddon and AddonProps? 
I think it would be better if you PRed something a bit more complete here. Just converting a function to a static method and adding an empty class doesn't make any sense..

ksooo:
I like @AlwinEsch's approach to bring in the changes step by step. For me, this makes reviewing stuff a lot easier (if not possible at all). So, for me, small incremental steps are fine.

But I'm with @tamland that without a big picture (like: "at the end new CAddonInfo is going to elminate the redundancies we have today with CAddon and AddonProps") would be very helpful.

ksooo (commented):
Code-wise this looks okay to me, but I have not enough knowledge to jugde on the conceptual concerns @tamland raised.

AlwinEsch:
Hi @tamland the AddonProps parts becomes moved to CAddonInfo and the old structure removed.

Because CAddonInfo makes much more, the name AddonProps no longer fits.

AchimTuran:
@ksooo @AlwinEsch 
> I like @AlwinEsch's approach to bring in the changes step by step. For me, this makes reviewing stuff a lot easier (if not possible at all). So, for me, small incremental steps are fine.

That's fine for me too. No one likes to review PRs greater than 1000 additions.

AlwinEsch:
jenkins build this please

PR #12235, opened 2017-06-04
[cmake] clang: disable annoying and pointless warnings

Rechi:
## Description
Disable the following warnings for clang
- overloaded-virtual
- inconsistent-missing-override

## Motivation and Context
These warnings spam the output and make it difficult to spot other warnings.

## How Has This Been Tested?
Compiled on OS X. Also compiled on windows to check it doesn't break anything.

## Types of change

## Checklist:


cmake/scripts/common/ArchSetup.cmake
```diff
@@ -157,3 +157,9 @@ if(CMAKE_BUILD_TYPE STREQUAL "Debug")
   add_options (ALL_LANGUAGES DEBUG "-g" "-D_DEBUG" "-Wall")
 endif()
 
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_options(CXX "DEBUG;RELEASE" "-Wno-overloaded-virtual")
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND (CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 7 OR CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 7))
```
wsnipex: clang >= 3.5 is supposed to support this. I guess we should do this on all platforms, not just OSX?
Also those options should only be enabled due to -Wall in Debug builds, so disabling in Release shouldn't be necessary?

Rechi: I couldn't find the exact version for clang where it got first introduced, will add it and also check if it's needed for release builds.

Memphiz:
I had a pr around which added the missing overrides and removed about 1000 warnings - iirc it got turned down by fernetmenta ... I'd rather see those overrides added (which is c++ syntax) instead of ignoring the warning. But the fact that I don't have the time to do that PR again - I am sort of fine with this ... those warnings are overdue since years anyway ;)

Rechi:
@wsnipex you are right, `overloaded-virtual` gets enabled by `-Wall` (debug builds), so I've changed to only disable it for debug builds.

The `inconsistent-missing-override` warning with AppleClang 7 and newer is always enabled (also release build) so I kept the disabling of this warning for debug and release.
Can you check if this behaviour is also true for the real clang 3.5 or newer?

wsnipex:
I've only tested with clang 4.0 on linux. inconsistent-missing-override is always active there as well.

Memphiz:
anything missing here?

Memphiz (approved):
As i have no time to do it any better - i am all for getting rid of those warnings the one or the other way.

Rechi:
@wsnipex should I disable `inconsistent-missing-override` warning only for clang 4 and newer as you haven't tested with a lower version?

garbear:
How many warnings are left? I'm going to be completely isolated with a macbook pro for the next week. I can address all the remaining warnings and remove the need for this change.

Rechi:
Latest OSX and IOS builds gave 15 `overloaded-virtual` warnings so if you can resolve them, that would be great.
I have to check how many `inconsistent-missing-override` warnings I get. Jenkins doesn't warn about them because that warning was first added in apple clang 7 and we use some apple clang 6 version on jenkins.

Paxxi:
I don't really like disabling these warnings, I'd rather fix them as @Memphiz tried to do.

it can be done automatically with clang-tidy and I think I have a commit for it here https://github.com/Paxxi/xbmc/commit/017747072de0d1f8af1dd949824e7f54dbe45844

Rechi:
I'll drop `-Wno-overloaded-virtual` as it is now only produces a few warnings which is okay.
A clean build with apple clang 7 produces 3907 `inconsistent-missing-override` warnings for current master. Can clang-tidy fix those as well? Otherwise I would like to disable this warning.

Paxxi:
inconsistent-missing-override is solved by the commit I linked. Name of commit is the clang-tidy rule. 
The other one is a bug and should be fixed. Someone got parameters wrong and failed to override the method and instead hid it.

MartijnKaijser:
Still needed?

Rechi:
No video player will be adapted soon and the rest is already fixed.

PR #12236, opened 2017-06-04
[cmake][addons] Adapted addons subdirs to latest changes.

ksooo:
Allows me to see headers-only directories xbmc/addons/kodi-addon-dev-kit/include/* in my Kodi Xcode project:

screen shot 2017-06-04 at 23 01 17

@hudokkow might taking a look.


Rechi:
@ksooo slashes in `core_add_library` aren't allowed

AlwinEsch:
To add a CMakeLists.txt in addon include folder make problems with https://github.com/xbmc/xbmc/blob/master/cmake/scripts/linux/Install.cmake#L149 and https://github.com/xbmc/xbmc/blob/master/cmake/scripts/common/PrepareEnv.cmake#L44
this install the whole directory.

It can work if the cmake parts becomes changed to:
```
# copy standard add-on include files
file(COPY ${CORE_SOURCE_DIR}/xbmc/addons/kodi-addon-dev-kit/include/kodi/
     DESTINATION ${APP_INCLUDE_DIR} REGEX ".txt" EXCLUDE)
```
and
```
install(DIRECTORY ${CMAKE_SOURCE_DIR}/xbmc/addons/kodi-addon-dev-kit/include/kodi/
        DESTINATION ${includedir}/${APP_NAME_LC}
        COMPONENT kodi-addon-dev
        REGEX ".txt" EXCLUDE)
```

ksooo:
> slashes in core_add_library aren't allowed

@Rechi: Okay, copied this from here: https://github.com/xbmc/xbmc/blob/master/cmake/treedata/common/subdirs.txt#L13 Interestingly it was done right in the repective CMakeLists.txt (https://github.com/xbmc/xbmc/blob/master/xbmc/input/joysticks/CMakeLists.txt#L31). I will fix my changes and the existing wrong entries. Thanks for pointing this out.

> To add a CMakeLists.txt in addon include folder make problems

@AlwinEsch as I'm not in cmake at all, could you please send me a diff for this?

ksooo:
@Rechi updated. Better now?

Rechi:
Slashes are only not allowed for `core_add_library(...)`, because it calls `add_library(...)` and this native cmake method can't handle slashes  in the library name.

Slashes at the second column at `cmake/treedata/common/subdirs.txt` are allowed so I don't see a benefit to change it.

ksooo:
> Slashes at the second column at cmake/treedata/common/subdirs.txt are allowed so I don't see a benefit to change it.

 I thought this second column entries must match what is used in core_add_library?

Rechi:
The argument supplied at `core_add_library` and the second column entry do not have to match.
The first one defines the name of the target if you use make files (e.g you can call `make input_joystick`) the other one defines where the build files are saved.

Now we have another issue, because `addons_kodi-addon-dev-kit_include_kodi` and others only have headers which doesn't work for Makefile based builds.

AlwinEsch:
Hi @ksooo,

here a working diff to prevent *.txt files during install

[0003-cmake-addons-prevent-the-use-of-.txt-files.patch.txt](https://github.com/xbmc/xbmc/files/1051998/0003-cmake-addons-prevent-the-use-of-.txt-files.patch.txt)

ksooo:
Okay, cmake is really not my playground and honestly, I cannot motivate myself to improve my cmake skills. Thus, closing this PR. Maybe somebody else can solve the problem with header-only directories not appearing in Xcode project one day.

xbmc/addons/kodi-addon-dev-kit/include/kodi/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+            xbmc_addon_types.h
+            xbmc_audioenc_dll.h
+            xbmc_audioenc_types.h
+            xbmc_epg_types.h
+            xbmc_pvr_dll.h
+            xbmc_pvr_types.h
+            xbmc_vis_dll.h
+            xbmc_vis_types.h)
+
+core_add_library(addons_kodi-addon-dev-kit_include_kodi)
```
Rechi:
```
if(NOT ENABLE_STATIC_LIBS)
  core_add_library(addons_kodi-addon-dev-kit_include_kodi)
endif()
```

@ksooo This solves the header only issue. The same needs to done for the other header only libs.

PR #12241, opened 2017-06-05
FFmpeg: Bump to 3.3.1-Leia-Alpha

fritsch:
This bumps ffmpeg.

I redid the 3.3.1-Leia-Alpha tag as it was not used before. This time it includes HLS fixes by @micahg


fritsch:
jenkins build this please

FernetMenta (approved):

micahg:
Hi,

Just tested and confirmed my fixes are in and work.

But in CDVDDemuxFFmpeg::GetFFMpegOptionsFromInput the cookie option kind of screws things up as it sets a header value instead of the http protocol cookies AVOption.  I might pursue a fix or alternative option so addons (which generally use that cookie option) can send in a proper cookie string.

PR #12242, opened 2017-06-05
added: ability to inspect add-on dependencies prior to installation

notspiff:
This is a simple way to achieve the request in https://forum.kodi.tv/showthread.php?tid=315528

Needs some skinner love and in general pointers on how to paint the shed.

### DIFF
--- a/xbmc/addons/AddonManager.cpp
+++ b/xbmc/addons/AddonManager.cpp
@@ -1234,6 +1234,48 @@ bool CAddonMgr::IsCompatible(const IAddon& addon)
   return true;
 }
 
+ADDONDEPS CAddonMgr::GetDepsRecursive(const std::string& id)
+{
+  ADDONDEPS added;
+  AddonPtr root_addon;
+  if (!FindInstallableById(id, root_addon) && !GetAddon(id, root_addon))
+    return added;
+
+  ADDONDEPS toProcess = root_addon->GetDeps();
+  while (!toProcess.empty())
+  {
+    auto current_dep = *toProcess.begin();
+    toProcess.erase(toProcess.begin());
+    if (StringUtils::StartsWith(current_dep.first, "xbmc.") ||
+        StringUtils::StartsWith(current_dep.first, "kodi."))
+      continue;
+
+    auto added_it = added.find(current_dep.first);
+    if (added_it != added.end())
+    {
+      if (current_dep.second.first < added_it->second.first)
+        continue;
+
+      bool aopt = added_it->second.second;
+      added.erase(added_it);
+      added.insert(current_dep);
+      if (!current_dep.second.second && aopt)
+        continue;
+    }
+    else
+      added.insert(current_dep);
+
+    AddonPtr current_addon;
+    if (FindInstallableById(current_dep.first, current_addon))
+    {
+      toProcess.insert(current_addon->GetDeps().begin(),
+                       current_addon->GetDeps().end());
+    }
+  }
+
+  return added;
+}
+
 int cp_to_clog(cp_log_severity_t lvl)
 {
   if (lvl >= CP_LOG_ERROR)
--- a/xbmc/addons/GUIDialogAddonInfo.cpp
+++ b/xbmc/addons/GUIDialogAddonInfo.cpp
@@ -39,6 +39,7 @@
 #include "GUIUserMessages.h"
 #include "guilib/GUIWindowManager.h"
 #include "input/Key.h"
+#include "messaging/helpers/DialogHelper.h"
 #include "pictures/GUIWindowSlideShow.h"
 #include "settings/Settings.h"
 #include "utils/JobManager.h"
@@ -51,12 +52,14 @@
 #include "interfaces/builtins/Builtins.h"
 
 #include <functional>
+#include <sstream>
 #include <utility>
 
 #define CONTROL_BTN_INSTALL          6
 #define CONTROL_BTN_ENABLE           7
 #define CONTROL_BTN_UPDATE           8
 #define CONTROL_BTN_SETTINGS         9
+#define CONTROL_BTN_DEPENDENCIES    10
 #define CONTROL_BTN_SELECT          12
 #define CONTROL_BTN_AUTOUPDATE      13
 #define CONTROL_LIST_SCREENSHOTS    50
@@ -64,6 +67,8 @@
 using namespace KODI;
 using namespace ADDON;
 using namespace XFILE;
+using namespace KODI::MESSAGING;
+using KODI::MESSAGING::HELPERS::DialogResponse;
 
 CGUIDialogAddonInfo::CGUIDialogAddonInfo(void)
   : CGUIDialog(WINDOW_DIALOG_ADDON_INFO, "DialogAddonInfo.xml"),
@@ -113,6 +118,12 @@ bool CGUIDialogAddonInfo::OnMessage(CGUIMessage& message)
       else if (iControl == CONTROL_BTN_SETTINGS)
       {
         OnSettings();
+        return true;
+      }
+      else if (iControl == CONTROL_BTN_DEPENDENCIES)
+      {
+        ADDONDEPS deps = CServiceBroker::GetAddonMgr().GetDepsRecursive(m_item->GetAddonInfo()->ID());
+        ShowDependencyList(deps, true);
         return true;
       }
       else if (iControl == CONTROL_BTN_AUTOUPDATE)
@@ -194,6 +205,9 @@ void CGUIDialogAddonInfo::UpdateControls()
   SET_CONTROL_LABEL(CONTROL_BTN_SELECT, CanUse() ? 21480 : (CanOpen() ? 21478 : 21479));
 
   CONTROL_ENABLE_ON_CONDITION(CONTROL_BTN_SETTINGS, isInstalled && m_localAddon->HasSettings());
+
+  ADDONDEPS deps = CServiceBroker::GetAddonMgr().GetDepsRecursive(m_item->GetAddonInfo()->ID());
+  CONTROL_ENABLE_ON_CONDITION(CONTROL_BTN_DEPENDENCIES, !deps.empty());
 
   CFileItemList items;
   for (const auto& screenshot : m_item->GetAddonInfo()->Screenshots())
@@ -336,6 +350,10 @@ void CGUIDialogAddonInfo::OnInstall()
   if (i != -1)
   {
     Close();
+    ADDONDEPS deps = CServiceBroker::GetAddonMgr().GetDepsRecursive(m_item->GetAddonInfo()->ID());
+    if (!deps.empty() && !ShowDependencyList(deps, false))
+      return;
+
     CAddonInstaller::GetInstance().Install(addonId, versions[i].first, versions[i].second);
   }
 }
@@ -458,6 +476,78 @@ void CGUIDialogAddonInfo::OnEnableDisable()
 void CGUIDialogAddonInfo::OnSettings()
 {
   CGUIDialogAddonSettings::ShowForAddon(m_localAddon);
+}
+
+bool CGUIDialogAddonInfo::ShowDependencyList(const ADDONDEPS& deps, bool reactivate)
+{
+  auto pDialog = g_windowManager.GetWindow<CGUIDialogSelect>(WINDOW_DIALOG_SELECT);
+  CFileItemList items;
+  for (auto& it : deps)
+  {
+    AddonPtr dep_addon, local_addon;
+    CServiceBroker::GetAddonMgr().FindInstallableById(it.first, dep_addon);
+    CServiceBroker::GetAddonMgr().GetAddon(it.first, local_addon);
+    if (dep_addon)
+    {
+      CFileItemPtr item(new CFileItem(dep_addon->Name()));
+      std::stringstream str;
+      str << it.first << " " << it.second.first.asString();
+      if ((it.second.second && !local_addon) || (!it.second.second && local_addon))
+        str << " " << StringUtils::Format(g_localizeStrings.Get(39022).c_str(),
+                                          local_addon ? g_localizeStrings.Get(39019).c_str()
+                                                      : g_localizeStrings.Get(39018).c_str());
+      else if (it.second.second && local_addon)
+        str << " " << StringUtils::Format(g_localizeStrings.Get(39023).c_str(),
+                                          g_localizeStrings.Get(39019).c_str(),
+                                          g_localizeStrings.Get(39018).c_str());
+
+      item->SetLabel2(str.str());
+      item->SetIconImage(dep_addon->Icon());
+      item->SetProperty("addon_id", it.first);
+      items.Add(item);
+    }
+    else
+    {
+      CFileItemPtr item(new CFileItem(it.first));
+      item->SetLabel2(g_localizeStrings.Get(161));
+      items.Add(item);
+    }
+  }
+
+  CFileItemPtr backup_item = GetCurrentListItem();
+  while (true)
+  {
+    pDialog->Reset();
+    pDialog->SetHeading(reactivate ? 39024 : 39020);
+    pDialog->SetUseDetails(true);
+    for (auto& it : items)
+      pDialog->Add(*it);
+    pDialog->EnableButton(!reactivate, 186);
+    pDialog->Open();
+
+    if (pDialog->IsButtonPressed())
+      return true;
+
+    if (pDialog->IsConfirmed())
+    {
+
+      const CFileItemPtr& item = pDialog->GetSelectedFileItem();
+      std::string addon_id = item->GetProperty("addon_id").asString();
+      AddonPtr dep_addon;
+      if (CServiceBroker::GetAddonMgr().FindInstallableById(addon_id, dep_addon))
+      {
+        Close();
+        ShowForItem(CFileItemPtr(new CFileItem(dep_addon)));
+      }
+    }
+    else
+      break;
+  }
+  SetItem(backup_item);
+  if (reactivate)
+    Open();
+
+  return false;
 }
 
 bool CGUIDialogAddonInfo::ShowForItem(const CFileItemPtr& item)
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -816,6 +816,7 @@ msgid "Searching album"
 msgstr ""
 
 #: xbmc/addons/GUIViewStateAddonBrowser.cpp
+#: xbmc/addons/GUIDialogAddonInfo.cpp
 #: xbmc/dialogs/GUIDialogSelect.cpp
 #: addons/skin.estuary/xml/DialogPVRGroupManager.xml
 msgctxt "#186"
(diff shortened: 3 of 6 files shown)


Montellese:
I prefer blue i.e. we could show the dependencies in a popup list.

notspiff:
don't follow. which dialog do you have in mind?

Montellese:
Hm maybe CGUIDialogSelect? We could either not allow selecting any item (is that already possible?) or jump to the info dialog of the dependency if one is selected.

notspiff:
it makes the hierarchy more messy but in general it's probably a better idea (mine shows deps of deps).

AlwinEsch:
Think the the same, with them can be the defined in list and maybe  allow to open for addon a text dialog with information of them.

notspiff:
i'll draft a flattened info-dialog enabling select dialog.

notspiff:
there.

notspiff:
i believe we need to have an enable/disable condition for the button, ref the return.

MartijnKaijser:
Some remarks:
Already looks good though it's a long list when installing a skin as I had 69 items on it. Currently it also shows Kodi's internal xbmc.foo and kodi.foo addons which show as "unavailable".
image

UX flow wise when you open the info of a dependency and you want to go back you don't go back to the list but to the main addon info page. Is it possible to end back in the list again?
Is it possible to show "installed"  for the ones already present?

Not sure how much work would it be but actually asking the user on install of the main addon if he/she wants to accept the list of needed dependencies as well before continuing would be great.
"This add-on requires additional dependencies to function. Would you want to continue or reviews them?"

notspiff:
hid the interface addons, added the enable condition for the button.

the navigation you request is very tricky to pull off due to the single instance dialogs. it would take a rather evil hack and i am not sure i want to be responsible for such code.

notspiff:
added installation status in the list.

notspiff:
added the prompt on installation as requested. here it's no problem to have the expected navigation, it's only when we launch from inside an active dialog that is an issue.

notspiff:
please give it a run and report back.

MartijnKaijser:
Will do tonight. Thank you :)

notspiff:
@tamland please review and see if i have violated some of your design constrains for addons with the new method(s).

notspiff:
and do not forget this still needs skin work, my changes is just a hackjob.

notspiff:
CI is annoyingly unstable :(

hudokkow:
yell at Memphiz and Rechi ;)

jenkins build this please

xbmc/addons/GUIDialogAddonInfo.cpp
```diff
@@ -336,6 +359,31 @@ void CGUIDialogAddonInfo::OnInstall()
   if (i != -1)
   {
     Close();
+    if (HELPERS::ShowYesNoDialogLines(CVariant{m_item->GetAddonInfo()->Name()},
```
xhaggi: we should use `ShowYesNoDialogText` instead, because `CGUIDialogBoxBase::SetLine` do some superfluous split magic and later call `CGUIDialogBoxBase::SetText`. IMO we should think about to mark `ShowYesNoDialogLines` as deprecated.

xbmc/addons/GUIDialogAddonInfo.cpp
```diff
@@ -460,6 +508,85 @@ void CGUIDialogAddonInfo::OnSettings()
   CGUIDialogAddonSettings::ShowForAddon(m_localAddon);
 }
 
+//! \brief Recursively add add-on dependencies to select dialog
+static void HandleAddon2(std::map<std::string, std::string>& added,
```
xhaggi: could we use a more applicable method name e.g. `FillSelectDialogWithAddOns`

xbmc/addons/GUIDialogAddonInfo.cpp
notspiff: done.

xbmc/addons/GUIDialogAddonInfo.cpp
notspiff: javafication applied ;)

xbmc/addons/IAddon.h
```diff
@@ -92,6 +92,7 @@ namespace ADDON
     virtual bool GetSettingString(const std::string& key, std::string& value) = 0;
     virtual CAddonSettings* GetSettings() const =0;
     virtual const ADDONDEPS &GetDeps() const =0;
+    virtual int NoDeps(std::map<std::string,std::string>& added) const = 0;
```
xhaggi: same here e.g. `GetMissingDependencies` and if we only need the map for the recursive call, we should think about to add a convenient method without the parameter and init the map there.

```
virtual int GetMissingDependencies() const {
  std::map<std::string,std::string>& addons;
  return GetMissingDependencies(addons);
}

Comment (notspiff) [inline xbmc/addons/IAddon.h]: I can consider a wrapper but your name suggestion says read code again;) No is englisj abbrevation of number, ie this counts deps recursively.

File: xbmc/addons/GUIDialogAddonInfo.cpp [hunk 4]
```diff
@@ -460,6 +508,85 @@ void CGUIDialogAddonInfo::OnSettings()
...
+          str << " (";
+          if (ad)
+            str << g_localizeStrings.Get(39019);
+          if (it.second.second)
+          {
+            if (ad)
+              str << ", ";
+            str << g_localizeStrings.Get(39018);
+          }
+          str << ")";
```
xhaggi: we could use a localized string to drop the hard-coded brackets

```
msgid "({0:s})"
```

notspiff: It is combining multiple depending on bools. I dont think that will be cleaner

xbmc/addons/IAddon.h
xhaggi: Should we  use `Count` instead of `No` then ;)

xbmc/addons/GUIDialogAddonInfo.cpp
xhaggi: Okay np, don't like these hard-coded strings.

garbear: and maybe also add a msg to adapt for RTL languages

MartijnKaijser:
@notspiff functionality looks good to me :)

xbmc/addons/GUIDialogAddonInfo.cpp
```diff
@@ -336,6 +359,31 @@ void CGUIDialogAddonInfo::OnInstall()
...
+          {
+            AddonPtr addon;
+            if (CServiceBroker::GetAddonMgr().FindInstallableById(show,addon))
+              ShowForItem(CFileItemPtr(new CFileItem(addon)));
+          }
+        }
+        if (HELPERS::ShowYesNoDialogText(CVariant{name},
+                                         CVariant{39021}) != DialogResponse::YES)
+          return;
+      }
```
tamland: This is way too many dialogs to click though. Just make the select dialog say "The following add-ons will be installed." Continue/Cancel, or something. No need for either of the yesno dialogs

xbmc/addons/GUIDialogAddonInfo.cpp
```diff
@@ -460,6 +508,85 @@ void CGUIDialogAddonInfo::OnSettings()
...
+      else
+      {
+        CFileItem item(it.first);
+        item.SetLabel2(g_localizeStrings.Get(161));
+        pDialog->Add(item);
+      }
+      FillSelectDialogWithAddons(added, pDialog, it.first);
+    }
+  }
```
tamland: This whole method seems to be pretty much a duplicate of NoDeps. Cant you just use the result of that to create the file items?

xbmc/addons/GUIDialogAddonInfo.cpp
notspiff: done (though i doubt seeing (optional, installed) instead of (installed, optional) would kill any kittens).

xbmc/addons/IAddon.h
notspiff: renamed, doxified and wrapped. due to the IAddon mess i need the recursive function as a virtual as well, public at that. sigh. oh, and yes i know i could have used a cast, but this way it's transparent when IAddon dies and diff would simply be moving the CAddon method to protected and removing override (right @AlwinEsch ?)

xbmc/addons/Addon.cpp
```diff
@@ -391,6 +391,30 @@ AddonVersion CAddon::GetDependencyVersion(const std::string &dependencyID) const
...
+    if (added.find(it.first) != added.end())
+    {
+      if (!added[it.first].empty() &&
+          AddonVersion(it.second.first.asString()) < AddonVersion(added[it.first]))
+        continue;
+    }
+    added[it.first] = it.second.first.asString();
+    AddonPtr addon;
+    result++;
+    if (CServiceBroker::GetAddonMgr().FindInstallableById(it.first, addon))
```
notspiff: there is no cycles for well behaved descriptions but yes, it can possibly do a cycle of course.

xbmc/addons/GUIDialogAddonInfo.cpp
notspiff: @MartijnKaijser

xbmc/addons/GUIDialogAddonInfo.cpp
```diff
@@ -460,6 +508,85 @@ void CGUIDialogAddonInfo::OnSettings()
...
+      }
+      added[it.first] = it.second.first.asString();
+      AddonPtr addon2;
+      CServiceBroker::GetAddonMgr().FindInstallableById(it.first, addon2);
+      if (addon2)
+      {
+        CFileItem item(addon2->Name());
+        std::stringstream str;
+        str << it.first << " " << it.second.first.asString();
+        AddonPtr ad;
```
notspiff: whatever you want, they are just temporaries needed in the recursion.

xbmc/addons/GUIDialogAddonInfo.cpp
notspiff: I wanted to avoid creating a list just to count the number of entries for the enable/disable condition.

xbmc/addons/GUIDialogAddonInfo.cpp
MartijnKaijser: is that possible (i have no idea)?
Maybe @ronie and @phil65 can share their view on this?

xbmc/addons/IAddon.h
tamland: So move it to AddonMgr as I said, where it belogs; mess gone

xbmc/addons/GUIDialogAddonInfo.cpp
tamland: Not sure what you mean. The enable/disable condition calls NoDeps/CountDeps..

notspiff: Yes. But you ask me to use the same method as the one which builds the list of items for the select dialog. Which means I will just build a list to check its size for the enable/disable.That's the reason why i slightly duped the code in a count method.

xbmc/addons/IAddon.h
notspiff: sure. but i cannot go back in time...

xbmc/addons/Addon.cpp
notspiff: ehm, disregard. i was working on removing the recursion and my head fooled me into thinking you were on about the same.

notspiff: completely rewritten method added to the manager.

xbmc/addons/GUIDialogAddonInfo.cpp
notspiff: gone.

xbmc/addons/GUIDialogAddonInfo.cpp
notspiff: new code should have all neat names for all temporaries etc.

xbmc/addons/GUIDialogAddonInfo.cpp
notspiff: i reconsidered after i rewrote the method. i rewrote to a method that returns the deps in an ADDONDEPS structure, then i use this to either check size or to create the list items. i have to build that map anyways..

xbmc/addons/GUIDialogAddonInfo.cpp
```diff
@@ -460,6 +508,85 @@ void CGUIDialogAddonInfo::OnSettings()
...
+      else
+      {
+        CFileItem item(it.first);
+        item.SetLabel2(g_localizeStrings.Get(161));
+        pDialog->Add(item);
+      }
+      HandleAddon2(added, pDialog, it.first);
+    }
+  }
+}
```
notspiff: see above.

notspiff:
rewritten to have almost proper navigation. the only quirk you can notice is that if you click a dependency to show the info dialog, then return, the original dialog is not visible in the background.

xbmc/addons/AddonManager.cpp
```diff
@@ -1233,6 +1233,49 @@ bool CAddonMgr::IsCompatible(const IAddon& addon)
...
 
+ADDONDEPS CAddonMgr::GetDepsRecursive(const std::string& id)
+{
+  ADDONDEPS added;
+  AddonPtr root_addon;
+  if (!FindInstallableById(id, root_addon) && !GetAddon(id, root_addon))
+    return added;
+
+  ADDONDEPS toProcess = root_addon->GetDeps();
+  int result = 0;
```
xhaggi: superfluous?

notspiff: aye

xbmc/addons/GUIDialogAddonInfo.cpp
```diff
@@ -460,6 +488,69 @@ void CGUIDialogAddonInfo::OnSettings()
...
+    else
+    {
+      CFileItemPtr item(new CFileItem(it.first));
+      item->SetLabel2(g_localizeStrings.Get(161));
+      items.Add(item);
+    }
+  }
+
+  CFileItemPtr backup_item = GetCurrentListItem();
+  std::string result="1";
```
xhaggi: Missing whitespaces

xbmc/addons/GUIDialogAddonInfo.cpp
```diff
@@ -460,6 +488,69 @@ void CGUIDialogAddonInfo::OnSettings()
...
+    pDialog->SetHeading(39017);
+    pDialog->SetUseDetails(true);
+    for (auto& it : items)
+      pDialog->Add(*it);
+    pDialog->Open();
+    if (pDialog->IsConfirmed())
+    {
+      const CFileItemPtr& item = pDialog->GetSelectedFileItem();
+      result = item->GetProperty("addon_id").asString();
+      AddonPtr dep_addon;
```
xhaggi: Should we do the following to reduce the scope for `result` and save some calls.

```
while(true)
{
  if (result.empty())
    break;
}
```

xhaggi:
BTW looks way better now.

xbmc/addons/GUIDialogAddonInfo.cpp
notspiff: while(true) construct utilized. *shudders*.

xbmc/addons/GUIDialogAddonInfo.cpp
notspiff: this code has been removed.

MartijnKaijser:
Tested it again and now all addon are marked optional while they are still pulled in and there's no dialog on installing a skin having dependencies.

notspiff:
fixed.

notspiff:
now, for the third time, can i get some skin help for the dialog? and what to do with the ui flow (too many dialogs etc)?

HitcherUK:
What sort of skinning help do you need exactly?

notspiff:
i have hijacked an existing button in the dialog for the dependency list.

HitcherUK:
Would it not be best to just add 'Dependencies' as a new button?

HitcherUK:
PS Feel free to ping me in future for skin help.

notspiff:
that's what i am asking for. but i'm not the one to decide where to stick the new button in the layout et al, that would not be a good idea ;)

HitcherUK:
I'd personally add it towards the end of the grouplist as it's probably going to be less used but I'm sure Phil wouldn't mind where it goes for now as it's easy enough to change at a later date.

notspiff:
right now the install button is at the far right, so i guess applying your logic means add it first.

HitcherUK:
Downloading a test build now so I can check it out.

HitcherUK:
How's this?

screenshot001

Code - https://pastebin.com/Zh6cLEfY

HitcherUK:
Flow wise it works exactly I'd expect it to.

notspiff:
flow question is due to this statement from tamland

```
This is way too many dialogs to click though. Just make the select dialog say "The following add-ons will be installed." Continue/Cancel, or something. No need for either of the yesno dialogs
```

notspiff:
and dialog looks good to me. took your code and squashed the pr up.

HitcherUK:
How do you get the yes/no dialog to appear? Whenever I install an addon it just installs all the dependencies anyway.

notspiff:
then there must still be a bug as it should pop up on install. are you sure you have the latest build? i fixed that earlier today. i cannot reproduce the issue with the new code so i'm a bit lost right now.

MartijnKaijser:
just kicked new build to test tonight

notspiff:
as for the other ui flow i would have to extend the select dialog to allow two buttons.

tamland:
@notspiff It already allow two for buttons. Call `CGUIDialogSelect::EnableButton`

notspiff:
ah i thought that was for the single cancel button. let's see if i can grok it.

notspiff:
there. bit of a mess left in strings and confusing parameter reuse. if everybody prefer this ui flow i will update.

notspiff:
.. how impolite of me. Thanks a bunch @HitcherUK

HitcherUK:
> .. how impolite of me. Thanks a bunch @HitcherUK

No problem, and yes I must have been using an older build.

MartijnKaijser:
Flow looks fine to me :)
One more thing i noticed is that it says `script.module.pil unavailable`. Not sure if we need to start hiding it as well next to core addons. With python scrapers coming up we might need to add another module and hard coding that might seem ugly. Thoughts?

Additional flow tweak might be to skip the dependency list dialog on install if all needed ones are already installed. So only show the dialog if additional ones need to be installed. Opinions?

HitcherUK:
> Additional flow tweak might be to skip the dependency list dialog on install if all needed ones are already installed. So only show the dialog if additional ones need to be installed.

I was going to suggest the same thing.

Also, although not necessary, it would be nice to have a setting to 'Always install dependencies' in addon settings.

MartijnKaijser:
I wouldn't go that far to add that setting to always auto install the dependencies.

Another possible scenario is when an already installed add-on gets additional new dependencies. What should the action be? Just blind install or ask user (of course when no video is playing) through same dialog if he wants to upgrade? Currently it will blind install the dependencies i suspect.

Already many thx spiff for current work.

MartijnKaijser:
@notspiff I think we can merge current work and go from there if there are no code objections?

notspiff:
agreed, rebased.

PR #12243, opened 2017-06-05
ADD: [droid] network stuff via api

koying:
Use Android API's for the network stuff, rather than POSIX

### DIFF
--- a/cmake/treedata/android/subdirs.txt
+++ b/cmake/treedata/android/subdirs.txt
@@ -2,6 +2,7 @@ xbmc/linux                              linuxsupport
 xbmc/input/touch                        input/touch
 xbmc/input/touch/generic                input/touch/generic
 xbmc/network/linux                      network/linux
+xbmc/network/android                    network/android
 xbmc/peripherals/bus/linux              peripherals/bus/linux
 xbmc/peripherals/bus/android            peripherals/bus/android
 xbmc/powermanagement/android            powermanagement/android
--- a/tools/android/packaging/xbmc/AndroidManifest.xml.in
+++ b/tools/android/packaging/xbmc/AndroidManifest.xml.in
@@ -15,6 +15,7 @@
     <uses-permission android:name="android.permission.BLUETOOTH" />
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
     <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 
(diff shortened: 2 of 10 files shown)


MartijnKaijser:
jenkins build this please

MartijnKaijser:
runtime tested on shield and aft2

MartijnKaijser:
@peak3d any objections?

xbmc/network/Network.cpp
```diff
@@ -466,6 +470,40 @@ bool CNetwork::PingHost(unsigned long ipaddr, unsigned short port, unsigned int
   return err_msg == 0;
 }
 
+bool CNetwork::PingHost(unsigned long remote_ip, unsigned int timeout_ms)
+{
+#ifdef TARGET_WINDOWS
+  return false;
```
FernetMenta: promoting PingHost to superclass and do this really bad

peak3d: I like the idea of using Androids Network implementation, but we all try in the past to avoid ifderrery.
Moving PingHost from LinuxNetwork to Network makes it worse,

Pls. let us go and make it better, means something like a new file xbmc/network/android/Network.cpp/h.
The Darwin code is not fine as well, and will hopefully be moved to its own platform place next time someone touches it.

koying:
> but we all try in the past to avoid ifderrery

I don't. Anyone thinking a "thing" like https://github.com/xbmc/xbmc/pull/10918/files (i.e. build time overrides) is better than an IFDEF is insane, imho ;)

But yeah, in this case, I could just duplicate the code, which is not insane, just stupid as a fix on one will surely be forgotten on the other.

peak3d:
With copying the code I don't absolutely mean making a 1:1 copy, but xbmc/network(Network.cpp could be something abstract and only the part wich differes from platform to platform should be overridden.

I haven't looked so close into the code, but if someone is in this area he may think about solutions wich remove ifdefs in sane manner.

koying:
Network.cpp is already full of TARGET_WINDOWS, and this specific one is just to make it build on win32.
The code is never reached from Win32 as it's overriden in https://github.com/xbmc/xbmc/blob/master/xbmc/network/windows/NetworkWin32.cpp#L258

So duplicating would actually be insane in the current situation as well ;)
Obviously, this is a good candidate for refactor, but it can happen anytime after this PR.

koying:
Put back PingHost into NetworkAndroid to please the code aesthetes

koying:
jenkins build this please

MartijnKaijser:
No comments anymore?

MartijnKaijser:
jenkins build this please

PR #12246, opened 2017-06-05
fix favourites reloading on profile change

tamland:
Just another profiles hack. Should fix https://trac.kodi.tv/ticket/17478

### DIFF
--- a/xbmc/favourites/FavouritesService.cpp
+++ b/xbmc/favourites/FavouritesService.cpp
@@ -77,8 +77,16 @@ static bool LoadFromFile(const std::string& strPath, CFileItemList& items)
   return true;
 }
 
-CFavouritesService::CFavouritesService(std::string userDataFolder) : m_userDataFolder(std::move(userDataFolder))
+CFavouritesService::CFavouritesService(std::string userDataFolder)
 {
+  ReInit(std::move(userDataFolder));
+}
+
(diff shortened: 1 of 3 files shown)


MartijnKaijser:
No objections?

xbmc/favourites/FavouritesService.cpp
```diff
@@ -77,8 +77,16 @@ static bool LoadFromFile(const std::string& strPath, CFileItemList& items)
   return true;
 }
 
-CFavouritesService::CFavouritesService(std::string userDataFolder) : m_userDataFolder(std::move(userDataFolder))
+CFavouritesService::CFavouritesService(std::string userDataFolder)
```
ksooo: would it make sense to change parameter type to <code>const std::string &</code> as you're no longer using the path-by-value-then-move idiom here? Alternatively, you could use <code>std::move</code> to assign <code>m_userDataFolder</code> in <code>CFavouritesService::ReInit</code>.

ksooo (commented):
Just one question/suggestion.

xbmc/favourites/FavouritesService.cpp
tamland: added the move!

ksooo: yep, imo the better approach of the two alternatives. thx.

ksooo (approved):
👍

PR #12248, opened 2017-06-05
RetroPlayer: Fixes for RPi after videobuffers update

garbear:
RPi is still broken because it needs to be updated for the videobuffers update, but these changes should fix RetroPlayer for when RPi is working again


Rechi:
@garbear this needs a rebase

PR #12238, opened 2017-06-04
Fix hidden overloaded virtual functions and override warnings

garbear:
Alternative to #12235. The fix for hidden functions comes from @notspiff .

These fixes have been sitting in my repo for about a month, and by this point I'm sure several other people have fixes for these sitting around as well.

PR #12250, opened 2017-06-05
VideoPlayer: Add missing override keywords

garbear:
Split out from #12238

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemux.h
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemux.h
@@ -243,29 +243,29 @@ public:
   /*
    * Reset the entire demuxer (same result as closing and opening it)
    */
-  virtual void Reset() = 0;
+  virtual void Reset() { }
 
   /*
    * Aborts any internal reading that might be stalling main thread
    * NOTICE - this can be called from another thread
    */
-  virtual void Abort() = 0;
+  virtual void Abort() { }
 
   /*
    * Flush the demuxer, if any data is kept in buffers, this should be freed now
    */
-  virtual void Flush() = 0;
+  virtual void Flush() { }
 
   /*
    * Read a packet, returns NULL on error
    *
    */
-  virtual DemuxPacket* Read() = 0;
+  virtual DemuxPacket* Read() { return nullptr; }
 
   /*
    * Seek, time in msec calculated from stream start
    */
-  virtual bool SeekTime(double time, bool backwards = false, double* startpts = NULL) = 0;
+  virtual bool SeekTime(double time, bool backwards = false, double* startpts = NULL) { return false; }
 
   /*
    * Seek to a specified chapter.
@@ -300,29 +300,29 @@ public:
    * Set the playspeed, if demuxer can handle different
    * speeds of playback
    */
-  virtual void SetSpeed(int iSpeed) = 0;
+  virtual void SetSpeed(int iSpeed) { }
 
   /*
    * returns the total time in msec
    */
-  virtual int GetStreamLength() = 0;
+  virtual int GetStreamLength() { return 0; }
 
   /*
    * returns the stream or NULL on error
    */
   virtual CDemuxStream* GetStream(int64_t demuxerId, int iStreamId) const { return GetStream(iStreamId); };
 
-  virtual std::vector<CDemuxStream*> GetStreams() const = 0;
+  virtual std::vector<CDemuxStream*> GetStreams() const { return std::vector<CDemuxStream*>(); }
 
   /*
    * return nr of streams, 0 if none
    */
-  virtual int GetNrOfStreams() const = 0;
+  virtual int GetNrOfStreams() const { return 0; }
 
   /*
    * returns opened filename
    */
-  virtual std::string GetFileName() = 0;
+  virtual std::string GetFileName() { return ""; }
 
   /*
    * return nr of subtitle streams, 0 if none
@@ -352,7 +352,7 @@ public:
 
 protected:
   virtual void EnableStream(int id, bool enable) {};
-  virtual CDemuxStream* GetStream(int iStreamId) const = 0;
+  virtual CDemuxStream* GetStream(int iStreamId) const { return nullptr; }
   virtual std::string GetStreamCodecName(int iStreamId) { return ""; };
 
   int GetNrOfStreams(StreamType streamType);
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxBXA.h
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxBXA.h
@@ -62,8 +62,6 @@ public:
   virtual void Abort() override;
   virtual void Flush() override;
   virtual DemuxPacket* Read() override;
-  bool SeekTime(double time, bool backwards = false, double* startpts = NULL) override { return false; }
-  virtual void SetSpeed(int iSpeed) override {}
   virtual int GetStreamLength() override { return (int)m_header.durationMs; }
   CDemuxStream* GetStream(int iStreamId) const override;
   std::vector<CDemuxStream*> GetStreams() const override;
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxBXA.h
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxBXA.h
@@ -56,17 +56,19 @@ public:
 
   bool Open(CDVDInputStream* pInput);
   void Dispose();
-  void Reset();
-  void Abort();
-  void Flush();
-  DemuxPacket* Read();
+
+  // implementation of CDVDDemux
+  virtual void Reset() override;
+  virtual void Abort() override;
+  virtual void Flush() override;
+  virtual DemuxPacket* Read() override;
   bool SeekTime(double time, bool backwards = false, double* startpts = NULL) override { return false; }
-  void SetSpeed(int iSpeed) {};
-  int GetStreamLength() { return (int)m_header.durationMs; }
+  virtual void SetSpeed(int iSpeed) override {}
+  virtual int GetStreamLength() override { return (int)m_header.durationMs; }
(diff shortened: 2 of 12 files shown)


xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxBXA.h
```diff
@@ -56,17 +56,19 @@ class CDVDDemuxBXA : public CDVDDemux
...
+
+  // implementation of CDVDDemux
+  virtual void Reset() override;
+  virtual void Abort() override;
+  virtual void Flush() override;
+  virtual DemuxPacket* Read() override;
   bool SeekTime(double time, bool backwards = false, double* startpts = NULL) override { return false; }
-  void SetSpeed(int iSpeed) {};
-  int GetStreamLength() { return (int)m_header.durationMs; }
+  virtual void SetSpeed(int iSpeed) override {}
```
FernetMenta: having SetSpeed as pure virtual does not make much sense if we need those kind of overrides in subclasses, right?

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxVobsub.h
```diff
@@ -35,19 +35,20 @@ class CDVDDemuxVobsub : public CDVDDemux
...
 
-  virtual void          Reset();
-  virtual void          Abort() {};
-  virtual void          Flush();
-  virtual DemuxPacket*  Read();
+  bool Open(const std::string& filename, int source, const std::string& subfilename);
+
+  // implementation of CDVDDemux
+  virtual void          Reset() override;
+  virtual void          Abort() override {}
```
FernetMenta: should be promoted to superclass

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxVobsub.h
```diff
@@ -35,19 +35,20 @@ class CDVDDemuxVobsub : public CDVDDemux
...
-  virtual void          SetSpeed(int speed) {}
+  virtual void          SetSpeed(int speed) override {}
   virtual CDemuxStream* GetStream(int index) const override { return m_Streams[index]; }
   virtual std::vector<CDemuxStream*> GetStreams() const override;
   virtual int           GetNrOfStreams() const override { return m_Streams.size(); }
-  virtual int           GetStreamLength()    { return 0; }
-  virtual std::string   GetFileName()        { return m_Filename; }
-
-  bool                  Open(const std::string& filename, int source, const std::string& subfilename);
+  virtual int           GetStreamLength() override { return 0; }
```
FernetMenta: also promote to superclass

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxVobsub.h
```diff
@@ -35,19 +35,20 @@ class CDVDDemuxVobsub : public CDVDDemux
...
+  virtual std::string   GetFileName() override { return m_Filename; }
```
FernetMenta: same here

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxBXA.h
garbear: You tell me :) You could make the argument that nearly any virtual method could have a trivial base class implementation. On the other hand, there might be value in being explicit by declaring the lack of any effect in its subclass.

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxVobsub.h
garbear: but the superclass doesn't have access to the private `m_Filename`?

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxBXA.h
FernetMenta: I don't kill warnings for the sake of killing warnings. We can take the opportunity to improve things that pop up here or just do nothing.
SetSpeed obvously was hacked into the interface at some time for a special case. Now we can take corrective action.

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxVobsub.h
FernetMenta: then it needs a closer look

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxBXA.h
garbear: ok, if you don't see anything obvious I'll look into these issues later tonight

garbear: SetSpeed() doesn't look like a hack. It's called upon a `CDVDMsg::PLAYER_SETSPEED` msg. `DVDDemuxFFmpeg` uses this function to decide what kinds of frames should be discarded. `DVDDemuxFFmpeg` is the only subclass that implements this, so the superclass should provide the trivial implementation.

garbear:
@FernetMenta added a commit moving all the functions you commented on (and more) to the superclass. Also added a commit for an unused function I just found.

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxBXA.h
FernetMenta: SetSpeed for the ffmpeg demuxer has a single purpose: it pauses network streams that are handled directly by ffmpeg. this internal information is exposed to the upper layers and all other demuxers and devs how write demuxers have to deal with this. This is what I call a hack. disacrding frames on the demuxers is max nonsense, won't work this way. Onother hack.

garbear: What's the next step? Drop this function?

garbear:
Comment got hidden after the rebase, so I'll reply here

> SetSpeed for the ffmpeg demuxer has a single purpose: it pauses network streams that are handled directly by ffmpeg. this internal information is exposed to the upper layers and all other demuxers and devs how write demuxers have to deal with this. This is what I call a hack. disacrding frames on the demuxers is max nonsense, won't work this way. Onother hack.

What's the next step? Drop this function?

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemux.h
```diff
@@ -243,29 +243,29 @@ class CDVDDemux
...
    * Flush the demuxer, if any data is kept in buffers, this should be freed now
    */
-  virtual void Flush() = 0;
+  virtual void Flush() { }
 
   /*
    * Read a packet, returns NULL on error
    *
    */
-  virtual DemuxPacket* Read() = 0;
```
FernetMenta: why did you change even the required methods? now the class is not pure virtual anymore.

garbear: because function is not universally required: https://github.com/xbmc/xbmc/pull/12250/files#diff-3b3e940dfa4c42a41d871daaaabe3d70L37

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxBXA.h
FernetMenta: demuxer could handle this internally. it knows about the nature of the stream and if it should be paused. if read is not called for a given time, it can pause ffmpeg. when player continues reading, the stream can continue as well. why should upper layers bother with those internals.

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemux.h
FernetMenta: CDVDDemuxCC is not a regular demuxer. It was done before we got DemuxMultiSource and will be migrated.

Rechi:
@garbear this needs a rebase

garbear:
Just to make sure, the commit "VideoPlayer: fixup 2baba89" is a rebase error and should be dropped, right?

FernetMenta:
@garbear yes, 2baba892693a68d32ad4fb54df7018a456e1d367 should not be in this PR here

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxBXA.h
garbear: The ownership of the input stream is confusing. It's held by both VideoPlayer and the demuxer. What's the reason for this? Would moving ownership to the demuxer exclusively solve this issue?

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemux.h
garbear: Which methods of the ones I changed are required?

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxBXA.h
FernetMenta: lifetime of input stream is not bound to lifetime of demuxer. SetSpeed was hacked into the code violating a basic OO principle: information hiding

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemux.h
FernetMenta: all but SetSpeed

garbear: but you commented that I should promote `Abort()`, `GetStreamLength()` and `GetFileName()` to the superclass?

FernetMenta: right, I mean those are required by VP. if overridden they can be used to change behaviour.  VP uses those methods to control demuxers and get info from demuxers. SetSpeed does not belong to this category. it was implemented because the demuxer wanted info from VP.

PR #12113, opened 2017-05-16
allow using VAAPI with OpenGLES

lrusak:
Now with the vpupdates merged I wanted to get this rolling.

The changes are pretty trivial.

This is needed to be able to use VAAPI with OpenGLES (drm/kms, mir, wayland, etc)

PR #12251, opened 2017-06-06
[settings] move to C++11

Montellese:
After the huge add-on settings work I figured it's finally time to move the settings related code to C++11 by
* using `nullptr` instead of `NULL`
* using the `default` keyword for ctors and dtors
* initializing member variables in the class definition
* using `enum class` instead of just `enum` where it makes sense
* using `using` instead of `typedef`

## Motivation and Context
Easier to read and maintain code

## Checklist:

### DIFF
--- a/xbmc/addons/AddonDll.cpp
+++ b/xbmc/addons/AddonDll.cpp
@@ -432,28 +432,28 @@ ADDON_STATUS CAddonDll::TransferSettings()
             const char* id = setting->GetId().c_str();
             switch (setting->GetType())
             {
-              case SettingTypeBool:
+              case SettingType::Boolean:
               {
                 bool tmp = std::static_pointer_cast<CSettingBool>(setting)->GetValue();
                 status = m_pDll->SetSetting(id, &tmp);
                 break;
               }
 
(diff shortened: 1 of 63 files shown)


MartijnKaijser:
Only win32 was happy :)

Montellese:
Yeah already looking into it. This is the problem with doing copy & paste in VS as it doesn't cover all the files. And the VS compiler is far less strict about C++11 features than GCC ;-)

Montellese:
Finally! Linux didn't like me using `None` as an enum value because it already seems to be defined globally as an integer :-/

MartijnKaijser:
You mean somewhere define that ourselves?

Montellese:
TBH I don't know and I didn't check. Just saw it in the Jenkins build logs.

MartijnKaijser:
I guess this is quite forward and can just be merged?

hauihau:
@Montellese, @MartijnKaijser: there was a commit in #12113  which stated that None is a libX11 macro.

Montellese:
@hauihau thanks for the hint. That's really stupid :-/

PR #12252, opened 2017-06-06
[network] CWebServer: add support for Cross-Origin Resource Sharing (CORS)

Montellese:
This is a WIP PR for adding Cross-Origin Resource Sharing (CORS) to Kodi's webserver. I open the PR because of #12225 and because I need input on how to handle configuring the list of allowed domains. I also haven't done much testing yet because I don't have a client ready to hand.

## Motivation and Context
In general browsers don't allow performing POST requests from one domain to another due to security reasons. This however can stop clients from interacting with Kodi. The solution for this is Cross-Origin Resource Sharing (CORS) which adds additional HTTP headers to the request and responses indicating whether CORS is supported or not.
Due to this being a potential security issue (if not properly configured) CORS is disabled by default and needs to be manually enabled by the user in the settings.

Furthermore the proper thing to do would be to ask the user to configure the list of domains which should be allowed to use CORS. This is however very technical and I'm not sure how best to integrate this into the user interface which is why I haven't done it yet. Right now when CORS is enabled the webserver accepts cross domain requests from any domain.

## Types of change

## Checklist:


da-anda:
would it make sense to only apply CORS to anything WAN while having it disabled for local clients (same subnet f.e.)? Just thinking about remote apps. I don't mind eitherway.

Montellese:
@da-anda: you already can set a server/host name in the settings and the webserver will treat any request with that `Origin` as not a CORS request.

jaylinski:
@Montellese How about only allowing to enable CORS if the user has set a (secure) password?

xbmc/network/WebServer.cpp
```diff
@@ -599,6 +695,68 @@ void CWebServer::FinalizePostDataProcessing(ConnectionHandler *connectionHandler
...
+  {
+    CLog::Log(LOGERROR, "CWebServer[%hu]: couldn't find a potential request handler for %s", m_port, request.pathUrl.c_str());
+    return SendErrorResponse(request, MHD_HTTP_NOT_FOUND, request.method);
+  }
+
+  struct MHD_Response *response = create_response(0, nullptr, MHD_NO, MHD_NO);
+  if (response == nullptr)
+  {
+    CLog::Log(LOGERROR, "CWebServer[%hu]: failed to create a HTTP 200 OPTIONS response", m_port);
+    return MHD_NO;
```
tamland: CORS headers needs to be added to all non-ok responses as well, otherwise the client will be unable to read the error response.

xbmc/network/WebServer.cpp
```diff
@@ -209,12 +233,32 @@ int CWebServer::HandlePartialRequest(struct MHD_Connection *connection, Connecti
...
+
+    // abort if CORS support is disabled
+    if (request.isCORSRequest && !m_corsEnabled)
+    {
+      CLog::Log(LOGWARNING, "CWebServer[%hu]: CORS support is disabled", m_port);
+      return SendErrorResponse(request, MHD_HTTP_BAD_REQUEST, request.method);
+    }
+
+    // TODO: abort if the origin isn't allowed access
+  }
```
tamland: If I'm reading this correct, I don't think this is correct way to implement it. It should not be necessary to abort any requests. One simply A: allow CORS by adding the CORS headers to the response, or B: disallow CORS by *not* adding the headers. OPTIONS can be used for other things

tamland:
> I need input on how to handle configuring the list of allowed domains. 

Open a popup when receiving a request and ask user to allow the client access, if yes add it to a whitelist? That would be the most user friendly I think.

Alternatively: a manually entered setting. I think a wildcard is a really bad idea at least, because I *think* what most user want is to just have clients on the local network access (like @da-anda) not websites, or having to set a strong password.

jaylinski:
@Montellese What do you think about the suggestion from tamland?

I guess the webserver would have to listen to all `OPTIONS` requests and look up an origin whitelist.   
If no match is found in the whitelist, present a pop-up to the user with an info about the request-origin and request permission.

Rechi:
@Montellese this needs a rebase

jaylinski:
@Montellese Any news / plans on finishing CORS support?

Zw3rv3r:
When can we enjoy this nice feature? :-)

I'm stuck ATM trying to create a JavaScript client interface for an external Kodi in my LAN. This code should solve it.

Zw3rv3r:
BTW I'm not a security specialist, but would enabling CORS and enabling "Allow remote control from applications on other systems" do the trick? Implementing the use of an API key could enhance security further. My 2 ct

Zw3rv3r:
> > When can we enjoy this nice feature? :-)
> > I'm stuck ATM trying to create a JavaScript client interface accessing an external Kodi in my LAN. This code should solve it.
> 
> @Zw3rv3r I circumvented this issue by using websockets. (https://github.com/jaylinski/kodi-echo/blob/master/src/modules/WebSocketApi.js)

I was assuming that Websockets had to follow the CORS rules too like the HTTP protocol. Thanks for the insight! I'm going to explorer that path.

JumBay:
Is this gonna be merged?
With PWA becoming more and more popular this is a must have. 

Using the websocket as a work around is not viable since not all methods seem to work with this. Moreover the default websocket port cannot be changed from the kodi interface and this port maybe blocked on some android tv boxes. 

Secondly I disagree the fact that CORS should be enabled only if a password is set. That’s not CORS role and you still can access kodi web client and do what you want.

tamland:
> Secondly I disagree the fact that CORS should be enabled only if a password is set. That’s not CORS role and you still can access kodi web client and do what you want.

The password is just a workaround in case allowed origin is `*` and not a whitelist. I doubt most people would set no password + enable wildcard cors if they understood the implications it. Also, most people probably run the web server on lan only which makes the web client protected, but not the server if you enable cors.

da-anda:
my recent idea would be to use OAuth2 instead of a username:password (HTTP-basic auth) approach if we are going that route. When a new client tries to get an access token we show a dialog in KODI and the user can either grant access, deny it or add the client to an ignore list (just in case an app would be spamming). That would IMO be the most userfriendly approach.
For headless setups we could still allow HTTP-basic auth via username:password (both can easily run in parallel).
Having that in place, we could probably allow `*` in the CORS headers and we would even be able (in theory) to grant different permissions per access token.

phunkyfish:
@Montellese are you still interested in merging this, close it or put it on the backburner?

Montellese:
I'm not really interested in this right now because from experience it will first take 2 years of discussion before we can agree on how to integrate this into Kodi and its settings.

phunkyfish:
Ok, will I set the back burner label and close it so?

mattmattmatt:
So what's the recommended best practice to interact with the Kodi API from a web client? @phunkyfish @Montellese

phunkyfish:
Not a question I have the answer to. I’m just doing a PR cleanup 😉

PR #12253, opened 2017-06-06
[addons] move AddonProps to new CAddonInfo class

AlwinEsch:
## Description

This add all previous on AddonProps defined values to
CAddonInfo who was added by commits before.

On this change is mostly only an all parts a rename done.

The commit comes in alone to hold next changes smaller
and better for review.

The CAddonInfo becomes on coming changes more and more
changed to allow multiple addon instances.

For this is it later not a static definition in CAddon
and also replace the old cpluff system.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/AddonInfo.h
+++ b/xbmc/addons/AddonInfo.h
@@ -19,7 +19,12 @@
  *
  */
 
+#include "XBDateTime.h"
+#include "addons/AddonVersion.h"
+
+#include <map>
 #include <string>
+#include <vector>
 
 namespace ADDON
 {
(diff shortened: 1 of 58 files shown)


xbmc/addons/AddonInfo.h
```diff
@@ -76,9 +81,15 @@ namespace ADDON
     ADDON_MAX
   } TYPE;
 
+  typedef std::map<std::string, std::string> InfoMap;
+  typedef std::map<std::string, std::pair<const AddonVersion, bool> > ADDONDEPS;
```
garbear: FYI, having two >> next to each other isn't a problem anymore since switching to c++11

xbmc/addons/AddonVersion.h
```diff
@@ -1,3 +1,4 @@
+#pragma once
```
garbear: FYI, unlike #define guards, `#pragma once` can appear anywhere in the file with the same effect, so I usually like placing the copyright header first

xbmc/addons/AddonInfo.h
garbear: Oh I see, this was moved from somewhere else :)

garbear (approved):
LGTM, you can read my comments but no changes required

ksooo (approved):
Nothing to complain about. Go ahead, merge it! :-)

AlwinEsch:
jenkins build this please

PR #12255, opened 2017-06-06
Dont force dirtyRegions on parent controls

peak3d:
Dont force dirtyRegions on parent controls

## Description
Currently we draw visible regions of recursivw parents if a control gets dirty
This  PR changes this behaviour (only Control dirty Rect is displayed)

## Types of change

@mkortstiege can you pls verify that its working as expected? Thx

### DIFF
--- a/xbmc/guilib/GUIWindow.cpp
+++ b/xbmc/guilib/GUIWindow.cpp
@@ -72,7 +72,6 @@ CGUIWindow::CGUIWindow(int id, const std::string &xmlFile)
   m_menuControlID = 0;
   m_menuLastFocusedControlID = 0;
   m_custom = false;
-  m_forceProcess = true;
 }
 
 CGUIWindow::~CGUIWindow()
@@ -343,10 +342,8 @@ void CGUIWindow::CenterWindow()
 
 void CGUIWindow::DoProcess(unsigned int currentTime, CDirtyRegionList &dirtyregions)
 {
-  if (!m_controlIsDirty && !m_forceProcess && g_advancedSettings.m_guiSmartRedraw)
+  if (!IsControlDirty() && g_advancedSettings.m_guiSmartRedraw)
     return;
-
-  m_forceProcess = false;
 
   g_graphicsContext.SetRenderingResolution(m_coordsRes, m_needsScaling);
   g_graphicsContext.AddGUITransform();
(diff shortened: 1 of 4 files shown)


FernetMenta:
what exactly does this do. please elaborate.

peak3d:
Beginning with PR 12213 we transport m_IsControlDirty down to the parents to get an early hint if we have to visit tree nodes in the GUI tree.

This change lead to the fact, that DirtyRegions are drawn / recorded for parents of a control wich becomes dirty ( a ListItemLabel e.g.). This was not the fact before the PR 12213.

We now have 2 states, 1 saying that the Control is dirty, and one wich say that some of its childs are dirty.

The first state is used for displaying DirtyRegions in default DirtyRegion mode, but also used for collecting only DirtyRects of dirty Controls, and not the parents of it.

The second state is base of the optimizations in PR 12213.

FernetMenta:
we still draw the entire screen if only a single control is dirty?

popcornmix:
@FernetMenta http://kodi.wiki/view/HOW-TO:Modify_dirty_regions
"Whole screen" is the default setting.
Other settings produce artefacts due to alpha overwriting without clearing the screen (at least on GLES).

FernetMenta:
I know, my question was if this PR would change this. the title "Dont force dirtyRegions on parent controls" kind of suggests this.

peak3d:
PR 12213 introduced following issues:
- dirty-mode==default (draw full screen always): if you enable "show dirty rects" you will see the rects of the parents of a dirty control.  From technical POV its meaningless, because everything is redrawn because of the mode.

- dirty-mode!= default(try to refresh only dirty regions): PR 12213 broke this, because the rects of all parents are set dirty (wich lead in effect to full redraw).

Especially the 1. part took already 2 hours discussion. This PR here fixes both issues (You'll see only the dirty rects of the dirty contrl / only the dirty controls dirty-rect is updated if anyone uses dirty-mode != default.

FernetMenta:
thanks for clarification

xbmc/guilib/GUIWindow.cpp
```diff
@@ -755,7 +752,7 @@ bool CGUIWindow::OnMessage(CGUIMessage& message)
           }
         }
         else if (message.GetParam1() == GUI_MSG_STATE_CHANGED)
-          m_forceProcess = true;
+          MarkDirtyRegion(DIRTY_STATE_CHILD); //Dont force an dirtyRect, we don't knw if / what has changed.
```
mkortstiege: typo in comment

mkortstiege:
Works as expected now. Thanks!

mkortstiege (approved):

peak3d:
thx for the time and looking at it @mkortstiege

PR #12256, opened 2017-06-06
[jenkins/docs] - added README for setting up a jenkins build slave

Memphiz:
… mac os (for building ios and osx with jenkins)

This is the HowTo for setting up a jenkins slave on macos. I did it exactly like that for the i7-1 slave.

@MartijnKaijser fyi

### DIFF
--- a/tools/buildsteps/jenkins_docs/README.mac
+++ b/tools/buildsteps/jenkins_docs/README.mac
@@ -1,0 +1,98 @@
+This are the steps to be done for configuring a mac for being a build slave to the Kodi jenkins CI system
+--------------------------------------------------------------------------------------------------------
+
+1. setup user jenkins as follows:
+
+# create jenkins group
+NEXT_GID=$((`dscl /Local/Default list /Groups gid | awk '{ print $2 }' | sort -n | grep -v ^[5-9] | tail -n1` + 1))
+sudo dscl /Local/Default create /Groups/jenkins
+sudo dscl /Local/Default create /Groups/jenkins PrimaryGroupID $NEXT_GID
+sudo dscl /Local/Default create /Groups/jenkins Password \*
+sudo dscl /Local/Default create /Groups/jenkins RealName 'Jenkins Node Service'
+# create jenkins user
+NEXT_UID=$((`dscl /Local/Default list /Users uid | awk '{ print $2 }' | sort -n | grep -v ^[5-9] | tail -n1` + 1))
+sudo dscl /Local/Default create /Users/jenkins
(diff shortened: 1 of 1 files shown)


Rechi:
Can you add the following after install ccache?
`Increase ccache max_size if enough disk space available`

If you want the exact location of the file that needs to be changed just tell me.

Memphiz:
Ah right - totally forgot about that one ... can you tell me the file and the size? so i can add it to the readme (and to the 2 new slaves too)

Rechi:
The file location is `/var/lib/jenkins/.ccache/ccache.conf` (`$HOME/.ccache/ccache.conf`) and I had set the content to `max_size = 15.0G` on the i5 nodes.

tools/buildsteps/jenkins_docs/README.mac
```diff
@@ -0,0 +1,96 @@
...
+sudo chown -R jenkins:wheel /var/lib/jenkins
+# create a logging space
+sudo mkdir /var/log/jenkins
+sudo chown jenkins:wheel /var/log/jenkins
+
+3. mkdir /Users/Shared/jenkins
+
+4. sudo chown jenkins:wheel /Users/Shared/jenkins
+
+5. mkdir -p /Users/Shared/xbmc-depends/dSyms
```
Rechi: seems like `/dSyms` went in by accident
same in the next line

Kwiboo:
You can use a script such as https://github.com/RasPlex/jenkins-slave-osx to simplify part of the jenkins slave setup. I used that script to configure my osx jenkins slaves for openpht in a matter of seconds.

Memphiz:
@Kwiboo i know that script and i used part of it. I am not comfortable to run an "arbitrary" script on our jenkins nodes thats why i pulled the steps we need out of it and put those in the readme. So i know what is done to the slave.

tools/buildsteps/jenkins_docs/README.mac
Memphiz: nope its not. Jenkins puts the generated debug info for each build into that directory when it exists. We use it mostly during releases when i need to symbolicate osx or ios crashlogs which don't have symbols when coming from a user.

Rechi: but jenkins must have permission for /Users/Shared/xbmc-depends
so at least it is wrong in the next line

Memphiz:
added ccache cache size adaption to the readme and also set it on the new nodes... - also fixed the dSyms dir permissions in the Readme (thx for pointing it out @Rechi

Memphiz:
squashed - i think this is good to go?

Rechi:
looks correct

PR #12257, opened 2017-06-07
Use http cookies AVOption (instead of sticking cookies in the header AVOption)

micahg:
## Description
Add an option to pass the ffmpeg cookies http protocol avoption instead of sticking cookie values inside the ffmpeg headers http protocol avoption.

This is preferable (and necessary) for streams that periodically update authentication values (such as a session token) that are stored as cookies. If the cookie values are specified in the headers avoption, it conflicts with the http protocols cookies option, which does get updated properly.

The cookies value format is different than the headers value format. The value used with the headers avoption is just the cookie value, whereas, the cookies avoption expects a newline delimited list of Set-Cookie values (eg: everything after ‘Set-Cookie: ‘).

## Motivation and Context

FFMPEG http protocol has an AVOption _just for cookies_ -- we shouldn't be forcing people to cram cookies into the headers AVOption. Also, the cookies AVOption is self updating if a server sends back a new Set-Cookie header in a response. As stated above, this is critical for streams that authenticate using cookies and periodically update those authentication values (eg: neulion).

## How Has This Been Tested?

I've tested this in my own sportsnet addon (https://github.com/micahg/plugin.video.snnow/tree/nologon). Basically instead of calling http://whatever.com/whatever.m3u8|Cookie=just_the_cookie_value I would call http://whatever.com/whatever.m3u8|Cookies=whole_set_cookie_value_including_domain_and_expiry. The stream works, and authentication values are properly updated. Without this change the stream plays until a new authentication value is sent down (which doesn't get updated) and then the stream ends.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
@@ -681,9 +681,16 @@ AVDictionary *CDVDDemuxFFmpeg::GetFFMpegOptionsFromInput()
         CLog::Log(LOGDEBUG, "CDVDDemuxFFmpeg::GetFFMpegOptionsFromInput() adding ffmpeg option 'user-agent: %s'", value.c_str());
         hasUserAgent = true;
       }
+      else if (name == "cookies")
+      {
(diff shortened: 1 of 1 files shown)


FernetMenta (approved):

FernetMenta:
thanks, build error were unrelaed

micahg:
Tyvm

PR #12258, opened 2017-06-07
[lang] Add ISO 639-2/T codes.

ace20022:
## Motivation and Context
In 5acbcc322ffde6f829a22539ea26bac9bb8d8e5f some methods where renamed to *ISO6392T which is wrong because they all convert/retrieve the ISO 639-2/B code of a given input.

Since libbluray requires ISO 639-2/T codes for the language selection (menu, audio, subtitle, metadata) I've added  ISO 639-2/T code retrieval.

## How Has This Been Tested?
No runtime tests done.

## Types of change


xbmc/utils/LangCodeExpander.cpp
```diff
@@ -1249,195 +1284,196 @@ extern const LCENTRY g_iso639_2[538] =
...
-  { "yi", "yid", NULL },
-  { "yo", "yor", NULL },
-  { "za", "zha", NULL },
-  { "zh", "chi", "zho" },
-  { "zu", "zul", NULL },
-  { "zv", "und", NULL }, // Kodi intern mapping for missing "Undetermined" iso639-1 code
-  { "zx", "zxx", NULL }, // Kodi intern mapping for missing "No linguistic content" iso639-1 code
-  { "zy", "mis", NULL }, // Kodi intern mapping for missing "Miscellaneous languages" iso639-1 code
-  { "zz", "mul", NULL }  // Kodi intern mapping for missing "Multiple languages" iso639-1 code
+  { "aa","aar",NULL,NULL },
```
Montellese: Could we please keep the spaces after the commas for easier readability?

xbmc/utils/LangCodeExpander.cpp
```diff
@@ -208,14 +209,33 @@ bool CLangCodeExpander::ConvertToISO6392T(const std::string& strCharCode, std::s
...
   return false;
 }
 
+bool CLangCodeExpander::ConvertToISO6392T(const std::string& strCharCode, std::string& strISO6392T, bool checkWin32Locales /* = false */)
+{
+
+  if (!ConvertToISO6392B(strCharCode, strISO6392T, checkWin32Locales))
+    return false;
+
+    for (unsigned int index = 0; index < ARRAY_SIZE(LanguageCodes); ++index)
```
Montellese: You could use `std::begin()` and `std::end()` with iterators instead of the index.

xbmc/utils/LangCodeExpander.cpp
ace20022: Sure, happend by accident.

xbmc/utils/LangCodeExpander.cpp
pkerling: You could also make LanguageCodes an `std::vector` (or `std::array`) and use a range-based for even more convenience.

ace20022: Not sure it's worth the effort. IIRC @Paxxi had the plan to replace it with icu.

ace20022:
@Montellese @pkerling I've changed my opinion ;) Could you please check again?

xbmc/utils/LangCodeExpander.cpp
```diff
@@ -480,25 +482,26 @@ bool CLangCodeExpander::CompareFullLanguageNames(const std::string& lang1, const
...
-  const LCENTRY *lang = g_iso639_1;
-  size_t length = sizeof(g_iso639_1);
+
   if (format == CLangCodeExpander::ISO_639_2)
   {
-    lang = g_iso639_2;
-    length = sizeof(g_iso639_2);
+    for (const auto& codes : g_iso639_2)
+    {
+      languages.emplace_back(codes.name);
```
pkerling: Minor nitpick: You could use std::copy here and further down, would make the function even shorter

pkerling:
Looks way better now :)

Disclaimer: I'm only voicing my personal opinion, I'm not officially reviewing here - actual maintainers might disagree.

xbmc/utils/LangCodeExpander.cpp
```diff
@@ -1245,204 +1283,206 @@ extern const LCENTRY g_iso639_2[538] =
...
+  { "yi", "yid", NULL, NULL },
+  { "yo", "yor", NULL, NULL },
+  { "za", "zha", NULL, NULL },
+  { "zh", "chi", "zho", "zho" },
+  { "zu", "zul", NULL, NULL },
+  { "zv", "und", NULL, NULL }, // Kodi intern mapping for missing "Undetermined" iso639-1 code  
+  { "zx", "zxx", NULL, NULL }, // Kodi intern mapping for missing "No linguistic content" iso639-1 code
+  { "zy", "mis", NULL, NULL }, // Kodi intern mapping for missing "Miscellaneous languages" iso639-1 code 
+  { "zz", "mul", NULL, NULL }  // Kodi intern mapping for missing "Multiple languages" iso639-1 code 
+
```
Montellese: empty line on purpose?

Montellese:
Looks better this way :+1:

ace20022:
@Montellese std::copy does not work, or I can't make it work... Not sure if the modern one is nicer/faster/better to read ;)

pkerling:
Sorry, I overlooked that you're using subfields of the entries (that's why I suggested `copy`). This makes it very debatable if `std::transform` is better than a range-based for, as you rightfully say.

You could do
```
auto const& codes = (format == CLangCodeExpander::ISO_639_2) ? g_iso639_2 : g_iso639_1;
std::transform(codes.begin(), codes.end(), std::back_inserter(languages), [](const LCENTRY& e) {return e.name; });
```
though, no need to repeat the code.

ace20022:
Does not work. ```no conversion from 'const std::array<LCENTRY,185>' to 'const std::array<LCENTRY,538>'```

Montellese:
Yeah these kind of games usually don't work with `std::array` because unline a builtin array `std::array` isn't implicitly cast to a pointer.

pkerling:
Yes, you'd have to make it a `std::vector` for the types to be compatible.

ace20022:
There's no ISO-3166 ;) I'll change that too.

PR #12260, opened 2017-06-07
[addons] add next step on CAddonInfo to become independent

AlwinEsch:
## Description

This change all values on CAddonInfo to private and access
them over functions.

Main reason for them is to make the class more safe for next changes
where CAddonInfo is no longer in CAddon, but globally as replacement
for cpluff.

Set of this value becomes in future mainly done from the class constructor
and no more by external parts.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/Addon.h
+++ b/xbmc/addons/Addon.h
@@ -56,38 +56,38 @@ public:
   explicit CAddon(CAddonInfo addonInfo);
   virtual ~CAddon() {}
 
-  TYPE Type() const override { return m_addonInfo.type; }
+  TYPE Type() const override { return m_addonInfo.MainType(); }
   TYPE FullType() const override { return Type(); }
-  bool IsType(TYPE type) const override { return type == m_addonInfo.type; }
-  std::string ID() const override{ return m_addonInfo.id; }
-  std::string Name() const override { return m_addonInfo.name; }
+  bool IsType(TYPE type) const override { return type == m_addonInfo.MainType(); }
+  std::string ID() const override{ return m_addonInfo.ID(); }
+  std::string Name() const override { return m_addonInfo.Name(); }
   bool IsInUse() const override{ return false; };
-  AddonVersion Version() const override { return m_addonInfo.version; }
-  AddonVersion MinVersion() const override { return m_addonInfo.minversion; }
-  std::string Summary() const override { return m_addonInfo.summary; }
-  std::string Description() const override { return m_addonInfo.description; }
-  std::string Path() const override { return m_addonInfo.path; }
+  AddonVersion Version() const override { return m_addonInfo.Version(); }
+  AddonVersion MinVersion() const override { return m_addonInfo.MinVersion(); }
+  std::string Summary() const override { return m_addonInfo.Summary(); }
+  std::string Description() const override { return m_addonInfo.Description(); }
+  std::string Path() const override { return m_addonInfo.Path(); }
   std::string Profile() const override { return m_profilePath; }
   std::string LibPath() const override;
-  std::string Author() const override { return m_addonInfo.author; }
-  std::string ChangeLog() const override { return m_addonInfo.changelog; }
-  std::string Icon() const override { return m_addonInfo.icon; };
-  ArtMap Art() const override { return m_addonInfo.art; }
-  std::vector<std::string> Screenshots() const override { return m_addonInfo.screenshots; };
-  std::string Disclaimer() const override { return m_addonInfo.disclaimer; }
-  std::string Broken() const override { return m_addonInfo.broken; }
-  CDateTime InstallDate() const override { return m_addonInfo.installDate; }
-  CDateTime LastUpdated() const override { return m_addonInfo.lastUpdated; }
-  CDateTime LastUsed() const override { return m_addonInfo.lastUsed; }
-  std::string Origin() const override { return m_addonInfo.origin; }
-  uint64_t PackageSize() const override { return m_addonInfo.packageSize; }
+  std::string Author() const override { return m_addonInfo.Author(); }
+  std::string ChangeLog() const override { return m_addonInfo.ChangeLog(); }
+  std::string Icon() const override { return m_addonInfo.Icon(); };
+  ArtMap Art() const override { return m_addonInfo.Art(); }
+  std::vector<std::string> Screenshots() const override { return m_addonInfo.Screenshots(); };
+  std::string Disclaimer() const override { return m_addonInfo.Disclaimer(); }
+  std::string Broken() const override { return m_addonInfo.Broken(); }
+  CDateTime InstallDate() const override { return m_addonInfo.InstallDate(); }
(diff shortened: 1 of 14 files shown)


ksooo (commented):
Some suggestions and questions...

xbmc/addons/Addon.h
```diff
@@ -56,38 +56,38 @@ class CAddon : public IAddon
   explicit CAddon(CAddonInfo addonInfo);
   virtual ~CAddon() {}
 
-  TYPE Type() const override { return m_addonInfo.type; }
+  TYPE Type() const override { return m_addonInfo.MainType(); }
```
tamland: In the end though there really should be no such thing as an 'addon type', but only addons with N possible extension points (which can have a type). So assuming this is only temporary for legacy code I guess 'MainType' works as to making it clearer..

xbmc/addons/AddonInfo.h
```diff
@@ -85,13 +85,240 @@ namespace ADDON
...
 
     /*!
+     * @brief To ask generated class is usable and all needed parts are set.
+     *
+     * @return True is OK and usable, otherwise false.
+     *
+     * @note This function is only be related to constructor's who read the
+     * addon xml part.
+     */
+    bool IsUsable() const { return m_usable; }
```
tamland: This doesn't belong to this commit

xbmc/addons/AddonInfo.h
```diff
@@ -85,13 +85,240 @@ namespace ADDON
...
+
+    /*!
+     * @brief Parts used from ADDON::CAddonDatabase
+     */
+    //@{
+    void SetInstallDate(CDateTime installDate) { m_installDate = installDate; }
+    void SetLastUpdated(CDateTime lastUpdated) { m_lastUpdated = lastUpdated; }
+    void SetLastUsed(CDateTime lastUsed) { m_lastUsed = lastUsed; }
+    void SetOrigin(std::string origin) { m_origin = std::move(origin); }
+    //@}
```
tamland: Same here: not used anywhere. And definitely not used by CAddonDatabase.

xbmc/addons/AddonInfo.h
```diff
@@ -85,13 +85,240 @@ namespace ADDON
...
+
+    /*!
+     * @brief Get the ID from add-on
+     *
+     * The ID is defined in addon.xml but should also equal with add-on path!
```
tamland: This isn't correct. There is no such requirement.

xbmc/addons/AddonInfo.h
```diff
@@ -85,13 +85,240 @@ namespace ADDON
...
+     *
+     * @return Add-on name
+     */
+    const std::string& Name() const { return m_name; }
+
+    /*!
+     * @brief Licence text of add-on
+     *
+     * @return the complete licence text defined on addon.xml
+     */
```
tamland: I find all these getter 'docstrings' very superficial.. It's completely obvious to everyone what all these methods does.

xbmc/addons/AddonInfo.h
```diff
@@ -85,13 +85,240 @@ namespace ADDON
...
+     * @brief To get broken parts
+     *
+     * @todo make description more detailed
+     */
+    const std::string& Broken() const { return m_broken; }
+
+    /*!
+     * @brief To get install date of add-on
+     *
+     * @note this is only be set from add-on browser and not always present!
```
tamland: That's not true. It should be set everywhere except when getting a single addon from addon manager. (And that is just a workaround due to some parts calling GetAddon way too much)

xbmc/addons/AddonBuilder.h
```diff
@@ -30,36 +30,36 @@ class CAddonBuilder
...
+  void SetInstallDate(CDateTime installDate) { m_addonInfo.m_installDate = installDate; }
+  void SetLastUpdated(CDateTime lastUpdated) { m_addonInfo.m_lastUpdated = lastUpdated; }
+  void SetLastUsed(CDateTime lastUsed) { m_addonInfo.m_lastUsed = lastUsed; }
+  void SetOrigin(std::string origin) { m_addonInfo.m_origin = std::move(origin); }
+  void SetPackageSize(uint64_t size) { m_addonInfo.m_packageSize = size; }
 
-  const std::string& GetId() const { return m_addonInfo.id; }
-  const AddonVersion& GetVersion() const { return m_addonInfo.version; }
+  const std::string& GetId() const { return m_addonInfo.m_id; }
+  const AddonVersion& GetVersion() const { return m_addonInfo.m_version; }
```
AlwinEsch: The add of `friend class CAddonBuilder` is only temporarily to allow them to set the values in CAddonInfo, on end create CAddonInfo the values itself by his constructor.

tamland (changes requested):

xbmc/addons/AddonInfo.h
```diff
@@ -101,31 +328,40 @@ namespace ADDON
...
-    InfoMap extrainfo;
-    CDateTime installDate;
-    CDateTime lastUpdated;
-    CDateTime lastUsed;
-    std::string origin;
-    uint64_t packageSize;
+    InfoMap extrainfo; // defined here, but removed in future with cpluff
+
+  private:
+    friend class ADDON::CAddonBuilder;
```
AlwinEsch: CAddonBuilder is the only class who currently need to set the values on CAddonInfo and only temporarily, the builder becomes on next steps complete removed. See also my other comment,

xbmc/addons/AddonInfo.h
AlwinEsch: Can remove it now, but becomes on next commits used the know class construction with read of addon.xml was OK.

xbmc/addons/AddonInfo.h
ksooo: Okay, then.

xbmc/addons/AddonBuilder.h
ksooo: Okay.

xbmc/addons/AddonInfo.h
AlwinEsch: Also a part who becomes in future used from CAddonDatabase. Can remove it and add later.

xbmc/addons/AddonInfo.h
AlwinEsch: Better remove the docs, for me also not really needed?

Thought only that maybe wanted for there.

ksooo: What @tamland said. Remove it. Does  not really add any value.

garbear (approved):

ksooo (approved):

ksooo:
Let's handle documentation issues in a separate PR, once all code changes are in, like agreed before.

AlwinEsch:
Have removed the comments, but want to leave the "Parts used from ADDON::CAddonDatabase" in. Becomes used on next commits and want them to hold small to see the critical parts.

ksooo:
Fine by me. Merge it!

AlwinEsch:
jenkins build this please

PR #12244, opened 2017-06-05
[addons] add support for banner and clearlogo

tamland:
Just some low-hanging fruit someone asked for long ago.. If anyone wants to add support for more later do this: https://github.com/xbmc/xbmc/commit/6ba38a1a2640193c5fffdd320aa6c6f1f7b2c58a

PR #12261, opened 2017-06-07
[addons] fix usage of CVariant map after 8851244

Rechi:
## Description
Why this is needed: https://github.com/xbmc/xbmc/pull/12244#pullrequestreview-42553186

## Motivation and Context
Fix crash caused by 8851244 (#12244)
@ksooo please check if this solves your crash

## How Has This Been Tested?
runtime tested, now crash is gone

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/AddonDatabase.cpp
+++ b/xbmc/addons/AddonDatabase.cpp
@@ -99,8 +99,11 @@ static void DeserializeMetadata(const std::string& document, CAddonBuilder& buil
   builder.SetIcon(variant["icon"].asString());
 
   std::map<std::string, std::string> art;
-  for (auto it = variant["art"].begin_map(); it != variant["art"].end_map(); ++it)
-    art.emplace(it->first, it->second.asString());
+  if (!variant["art"].empty())
+  {
+    for (auto it = variant["art"].begin_map(); it != variant["art"].end_map(); ++it)
+      art.emplace(it->first, it->second.asString());
+  }
   builder.SetArt(std::move(art));
 
   std::vector<std::string> screenshots;


tamland:
I don't think this is correct.. At least this is not how std containers work (you can iterate an empty map just fine)

Rechi:
You are partly correct, normally the empty check for pure std::map isn't necessary, but here we interact with `CVariant`.
Have a look at https://github.com/xbmc/xbmc/blob/d8fbb5a/xbmc/utils/Variant.cpp#L784-L814, especially the else branches. If art wasn't parsed into the CVariant you get different iterators for begin_map and end_map and you will iterate over unexpected memory locations.
The real issue is in the CVariant class. I'll have a look if it can be fixed there easily, which is of course the preferred way.

xbmc/addons/AddonDatabase.cpp
```diff
@@ -99,8 +99,11 @@ static void DeserializeMetadata(const std::string& document, CAddonBuilder& buil
   builder.SetIcon(variant["icon"].asString());
 
   std::map<std::string, std::string> art;
-  for (auto it = variant["art"].begin_map(); it != variant["art"].end_map(); ++it)
-    art.emplace(it->first, it->second.asString());
+  if (!variant["art"].empty())
```
Montellese: The proper check is `if (variant.isMember("art") && variant["art"].isObject())`. Then you  can just iterate over it even if the "art" member is an empty map.

Montellese:
The correct usage of CVariant is to first check if a member exists using `isMember()` and if the type of the member is what you expect it to be. You can't just assume that it's an object/map and start iterating over it with `begin_map()`. CVariant wasn't designed to just magically work like a python dict.

tamland:
> You can't just assume that it's an object/map and start iterating over it with begin_map(). CVariant wasn't designed to just magically work like a python dict.

It definitely *reads* like it should as every other method in CVariant returns the empty variant when the type is 'wrong'. So I expected the empty variant to also return an empty iterator, not undefined behavior. I mean what's event the point of returning an iterator for unexpected types if it cannot be used?

Rechi:
`begin_map()` and `end_map()` both return an empty iterator, but not the same.
Problem now is, begin != end, and therefore it iterates over unexpected memory locations.

#12264 would correct the behaviour to return the same iterator for begin and end, but I'm not sure what @Montellese thinks about that.

Rechi:
fixed with #12264

PR #12264, opened 2017-06-07
[utils] Variant: return matching iterators for (begin|end)_(array|map)

Rechi:
## Description
[utils] Variant: return matching iterators for (begin|end)_(array|map), if type is not array or map

With this change std::distance(...) is also correct if the type isn't an array or map.
Without this change there needs to be a check for isArray respectively isObject before iterating over the elements.

## Motivation and Context
Incorrect usage of CVariant at #12244.
@Montellese I don't know if you like this idea of the static array and map.

## How Has This Been Tested?
compiled and runtime tested on Windows and OSX

## Types of change

## Checklist:

### DIFF
--- a/xbmc/utils/Variant.cpp
+++ b/xbmc/utils/Variant.cpp
@@ -135,6 +135,8 @@ CVariant::CVariant()
 }
 
 CVariant CVariant::ConstNullVariant = CVariant::VariantTypeConstNull;
+CVariant::VariantArray CVariant::EMPTY_ARRAY;
+CVariant::VariantMap CVariant::EMPTY_MAP;
 
 CVariant::CVariant(VariantType type)
 {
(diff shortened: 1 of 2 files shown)


xbmc/utils/Variant.cpp
```diff
@@ -135,6 +135,8 @@ CVariant::CVariant()
 }
 
 CVariant CVariant::ConstNullVariant = CVariant::VariantTypeConstNull;
+CVariant::VariantArray CVariant::EMPTY_ARRAY = VariantArray();
+CVariant::VariantMap CVariant::EMPTY_MAP = VariantMap();
```
Montellese: You shouldn't need the `= VariantArray()` and `= VariantMap()` parts.

Rechi:
@Montellese addressed your comment and runtime tested again.

Montellese (approved):
Good to go IMO.

MartijnKaijser:
@Rechi could you make a backport PR?

Rechi:
I don't think its necessary, but I'll backport it.

PR #12265, opened 2017-06-07
[webserver] dont allow jsonrpc to use http get

tamland:
As discussed in the forums, allowing jsonrpc to be sent over HTTP GET:

1. violates the idempotence requirement of GET.
2. makes the jsonrpc endpoint vulnerable to CSRF.

This is fixed by requiring all HTTP request to jsonrpc endpoint use POST. This is a security patch so all clients still using the endpoint must switch to POST immediately!

### DIFF
--- a/xbmc/network/httprequesthandler/HTTPJsonRpcHandler.cpp
+++ b/xbmc/network/httprequesthandler/HTTPJsonRpcHandler.cpp
@@ -34,7 +34,7 @@
 
 bool CHTTPJsonRpcHandler::CanHandleRequest(const HTTPRequest &request) const
 {
-  return (request.pathUrl.compare("/jsonrpc") == 0);
+  return (request.method == POST && request.pathUrl.compare("/jsonrpc") == 0);
 }
 
 int CHTTPJsonRpcHandler::HandleRequest()
@@ -61,15 +61,6 @@ int CHTTPJsonRpcHandler::HandleRequest()
     }
 
     isRequest = true;
-  }
-  else if (m_request.method == GET)
-  {
-    std::map<std::string, std::string>::const_iterator argument = arguments.find("request");
-    if (argument != arguments.end() && !argument->second.empty())
-    {
-      m_requestData = argument->second;
-      isRequest = true;
-    }
   }
 
   std::map<std::string, std::string>::const_iterator argument = arguments.find("jsonp");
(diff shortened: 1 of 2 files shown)


Montellese:
Could you check if doing a HTTP GET request to just `/jsonrpc` without any options still works? It should return the full JSON-RPC API definition and IMO it makes sense to support that through HTTP GET.

Montellese:
Theoretically we could change the logic so that HTTP GET requests are possible but only to JSON-RPC methods with `"permission": "ReadData"`. That way we could still support simple HTTP GET clients which only read data.

MilhouseVH:
I'll include this in tonight's RPi/x86 LibreELEC #0607 test builds which should get it some early exposure.

Jalle19:
Good idea IMO, even though it may break a bunch of custom scripts that people use to e.g. trigger a library update.

uvjim:
@Montellese @tamland @MilhouseVH - just updated to Milhouse LibreELEC build #0608 and can no longer access the API definition at `/jsonrpc` by doing a HTTP GET request with no parameters.  An empty HTTP POST request produces a parse error.

I have another RPi in the house which I have left at Milhouse LibreELEC build #0606 and that is behaving as expected, in regards to retrieving the API definition.  Looking at the changes between those versions this seemed the most likely candidate.

Is there any other way to get to the API definition?  Was this by design?

Montellese:
@uvjim in the current state of this PR that behaviour is to be expected but as you can see in my previous comment I already requested that this is changed so that at least the full API definition can still be accessed using a HTTP GET to `/jsonrpc`. But @tamland hasn't reacted to it yet.

uvjim:
Thanks @Montellese.  I did see your comment but wasn't sure if it had been acted upon.

tamland:
> Could you check if doing a HTTP GET request to just /jsonrpc without any options still works? It should return the full JSON-RPC API definition and IMO it makes sense to support that through HTTP GET.

Sorry for slow reply. I tried to keep this at first but could make a clean solution due to all the other stuff floating around in this handler.

> Theoretically we could change the logic so that HTTP GET requests are possible but only to JSON-RPC methods with "permission": "ReadData". That way we could still support simple HTTP GET clients which only read data.

That also works. The important thing is that GET request can be made by malicious websites and therefore must not do any modifications. This PR is just a a quick fix as I don't really have the insight or time to look into this further, so feel free to supersede!

@uvjim You can call the Introspect method as shown here: https://github.com/xbmc/xbmc/pull/12265/files#diff-58a2b2a422ef05cf3007dbdaeaf3ce39R369

uvjim:
Thanks @tamland.

Montellese:
@tamland see #12281 for my alternative approach.

PR #12268, opened 2017-06-08
RFC: [droid] translate to ascii for sorting

koying:
As droid NDK does not support locale, use iconv to translate utf8 accentuated characters to ascii, e.g "â" to "a"

Obviously has a performance cost (limited, afaict), but really helps when your media titles are in French (or German, I guess).
Not idea how "ASCII//TRANSLIT" fares for non-latin charsets.

Comments welcome

PS: fixup because TRANSLIT adds accents as a separate char, so `È` becomes `'E` (stupid, I know)

### DIFF
--- a/xbmc/utils/CharsetConverter.cpp
+++ b/xbmc/utils/CharsetConverter.cpp
@@ -87,6 +87,7 @@ enum SpecialCharset
   SystemCharset,
   UserCharset /* locale.charset */, 
   SubtitleCharset /* subtitles.charset */,
+  AsciiCharset
 };
 
 class CConverterType : public CCriticalSection
@@ -244,6 +245,8 @@ std::string CConverterType::ResolveSpecialCharset(enum SpecialCharset charset)
     return g_langInfo.GetGuiCharSet();
   case SubtitleCharset:
     return g_langInfo.GetSubtitleCharSet();
+  case AsciiCharset:
+    return "ASCII//TRANSLIT";
   case NotSpecialCharset:
   default:
     return "UTF-8"; /* dummy value */
@@ -269,6 +272,8 @@ enum StdConversionType /* Keep it in sync with CCharsetConverter::CInnerConverte
   Utf8ToSystem,
   SystemToUtf8,
   Ucs2CharsetToUtf8,
+  WtoAscii,
+  Utf8toAscii,
   NumberOfStdConversionTypes /* Dummy sentinel entry */
 };
 
@@ -312,7 +317,9 @@ CConverterType CCharsetConverter::CInnerConverter::m_stdConversion[NumberOfStdCo
   /* Utf8toW */             CConverterType(UTF8_SOURCE,     WCHAR_CHARSET),
   /* Utf8ToSystem */        CConverterType(UTF8_SOURCE,     SystemCharset),
   /* SystemToUtf8 */        CConverterType(SystemCharset,   UTF8_SOURCE),
-  /* Ucs2CharsetToUtf8 */   CConverterType("UCS-2LE",       "UTF-8", CCharsetConverter::m_Utf8CharMaxSize)
+  /* Ucs2CharsetToUtf8 */   CConverterType("UCS-2LE",       "UTF-8", CCharsetConverter::m_Utf8CharMaxSize),
+  /* WtoAscii */            CConverterType(WCHAR_CHARSET,   AsciiCharset),
+  /* Utf8toAscii */         CConverterType(UTF8_SOURCE,     AsciiCharset),
 };
 
 CCriticalSection CCharsetConverter::CInnerConverter::m_critSectionFriBiDi;
@@ -723,6 +730,11 @@ bool CCharsetConverter::utf8ToW(const std::string& utf8StringSrc, std::wstring& 
   return CInnerConverter::stdConvert(Utf8toW, utf8StringSrc, wStringDst, failOnBadChar);
 }
 
+bool CCharsetConverter::utf8ToASCII(const std::string& utf8StringSrc, std::string& asciiStringDst, bool failOnBadChar)
+{
+  return CInnerConverter::stdConvert(Utf8toAscii, utf8StringSrc, asciiStringDst, failOnBadChar);
+}
+
 bool CCharsetConverter::subtitleCharsetToUtf8(const std::string& stringSrc, std::string& utf8StringDst)
 {
   return CInnerConverter::stdConvert(SubtitleCharsetToUtf8, stringSrc, utf8StringDst, false);
@@ -803,6 +815,11 @@ bool CCharsetConverter::unknownToUTF8(const std::string& stringSrc, std::string&
 bool CCharsetConverter::wToUTF8(const std::wstring& wStringSrc, std::string& utf8StringDst, bool failOnBadChar /*= false*/)
 {
   return CInnerConverter::stdConvert(WtoUtf8, wStringSrc, utf8StringDst, failOnBadChar);
+}
+
+bool CCharsetConverter::wToASCII(const std::wstring& wStringSrc, std::string& asciiStringDst, bool failOnBadChar)
+{
+  return CInnerConverter::stdConvert(WtoAscii, wStringSrc, asciiStringDst, failOnBadChar);
 }
 
 bool CCharsetConverter::utf16BEtoUTF8(const std::u16string& utf16StringSrc, std::string& utf8StringDst)
--- a/xbmc/utils/CharsetConverter.h
+++ b/xbmc/utils/CharsetConverter.h
@@ -126,6 +126,7 @@ public:
   static bool utf8ToW(const std::string& utf8StringSrc, std::wstring& wStringDst,
                 bool bVisualBiDiFlip = true, bool forceLTRReadingOrder = false,
                 bool failOnBadChar = false);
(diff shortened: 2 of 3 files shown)


Razzeee:
Whats currently happening when you sort and hit a fancy character?

peak3d (commented):
1.) The ifdeffery here is ugly
2.) SortItem has IMO a FieldSort Map element, wich can be translated once on initialization time. This would remove the need of changes at least in SortUtils.

koying:
> Whats currently happening when you sort and hit a fancy character?

It gets at the bottom of the list, ie you get ..., "Y", "Z", "À", "É" for french

koying:
> SortItem has IMO a FieldSort Map element

If not wrong, this is a user field, stored in the DB. We cannot override it with something else, or we miss the point of the field.

peak3d:
I don't know whythis should not be done, If you fake the value on ever sort, then you can fake it once completely. This is more truthful then hiding what you do in this really sensitive loop (and jes, it will cost a lot of CPU.

koying:
If the user goes with its own sorting like "1 - yyzz", "2 - brol", "3 - toto", we are not speaking about the same sorting override.

peak3d:
Not optimal in your approach is that it takes place in comparator. left and / or right will accessed multiple times until the sorting is done.

After a short overview it seems that all elements are copied into a SortItem structure. This is at least true for FileItem and Database. This would be a nice place to transform the char (if necessary) to ascii.

Maybe you want to change first one of both (FileItem or Database), TBH I never came int this issue, but (as most already know) I'm not a power user.

Power users using this feature will have lots of entries, wich will lead to noticable performance decrease.

koying:
Well, this is a RFC. 
I agree this is very low, so conversion is done multiple times. If someone has ideas (and preferably code) on better place to do the conversion (ie higher), they are most welcome. 

That being said, the same is probably true for StringUtils::AlphaNumericCompare

koying:
I got what @peak3d said ;)

Interestingly, I noticed the sort isn't proper even on locale-enabled platforms due to https://github.com/xbmc/xbmc/blob/master/xbmc/utils/StringUtils.cpp#L857  

It produce something like this on windows. "B" comes after, so I guess the french collate ignores caps anyway. /cc @Paxxi
kodi_windows_bad_sort

I start to wonder if we shouldn't enable it for all, as a further side effect is that non-alphanumeric chars are just ignored, probably as it should be...
It all depends on what is produced in non-latin languages, ofc.

MartijnKaijser:
ready to merge? Any comments left?

Rechi:
osx and ios have the same (or a similar issue)
have done a quick test with this PR on osx and it fixes the issue (after enabling osx in the ifdef)

koying:
@MartijnKaijser If this is green-lit, please allow me to squash before merging

MartijnKaijser:
Green? :)

koying:
Squashed and rebased

Razzeee:
We can merge this but might want to revisit the `ifdef` as  @Rechi seems to say it does also fix a osx/ios bug

Rechi:
From my side it can be merged, I'll look at the apple issue later.

pkerling:
This breaks sorting completely for all languages not based on the Latin alphabet. It just removes everything (#15026), even ASCII characters such as `@` or `!` (#14952).

I'm leaning towards a revert of this if we do not find another solution.

@koying @ksooo Comments? Can we somehow keep the benefits without breaking lots of other stuff like it is now?

pkerling:
Would it help to decompose to Unicode NKFD (Normalization Form Compatibility Decomposition), so for example `È` becomes two characters (`E` + mark)? We'd need libunistring or libicu for that, though.

pkerling:
Ah, libicu has locale-aware sorting. Nice. But it's a quite heavy library and I don't think we can properly integrate it this late in the release cycle. We should consider using it on every platform instead of libc locales though to get consistent results in the future.

I'm looking for a solution/workaround we can implement on Android in the meantime.

koying:
Proper (but painful) way is to go jni, I guess.
Android happens to use ICU as of 24 btw: https://developer.android.com/guide/topics/resources/internationalization.

Workaround is to be smarter about what goes thru the filter ;)

Paxxi:
Yeah it's not a quick fix to use and I don't have any useful input for that.
I did investigate icu and there's a few branches in my fork that should hopefully still work. If someone's willing to push it through then feel free to rebase it and pr

pkerling:
> Workaround is to be smarter about what goes thru the filter ;)

I'm afraid that won't help. `iconv` transliterates every non-Latin character to `?`. So even if we removed the filter the situation would not improve for CJK and Russian users. An option would be to apply the transliteration only to completely Latin titles but I don't really have a reliable idea how to detect that. We could just disregard the transliteration if it contains a question mark maybe? Doesn't sound too bad as a workaround IMO if nobody has a better idea.

> Composition or decomposition doesn't help without knowing the sorting rules for the locale

Well, it does help in the examples mentioned in this PR. I think it's comparable to the `iconv` transliteration here except that it doesn't bork all non-Latin chars. The situation on Android before this PR also was that the locale-specific rules were completely ignored.

> Proper (but painful) way is to go jni, I guess.

+1, but someone with Android experience has to do that. Roadmap for v19 I guess.

pkerling:
Oh OK. It seems Android `iconv` works differently from glibc. E.g. I don't get `'E`:

```
$ echo 'äöüÄÖÜÀÉあいうえお' | LC_ALL=de_DE.UTF-8 iconv --from-code=UTF-8 --to-code=ASCII//TRANSLIT
aeoeueAEOEUEAE?????
```

@koying Do you remember or can you test what non-Latin characters output on Android `iconv`? Here I get question marks.

koying:
Looks like there is a semi-official backdoor to use the shipped android icu from C/C++:
https://github.com/android-ndk/ndk/issues/548

pkerling:
Doesn't seem to be available yet, but for the future we should keep it in mind

pkerling:
Yes, you're right. Only allows access to the C API though (not C++), which is suboptimal. Anyway if possible we should probably avoid bringing libicu into the mix in RC4.

koying:
Some code:
The ICU wrapper: https://gitlab.com/koying/libicundk

You can replace the whole of `StringUtils::AlphaNumericCompare` (`g_langInfo.GetSystemLocaleString()` being, eg, `"fr_BE"`) by

```
  // Use ICU
  UErrorCode ustatus = U_ZERO_ERROR;
  UCollator* ucoll = ucol_open(g_langInfo.GetSystemLocaleString().c_str(), &ustatus);
  if (U_FAILURE(ustatus))
  {
    CLog::Log(LOGWARNING, "Cannot open icu collation '%s': %s", g_langInfo.GetSystemLocaleString().c_str(), u_errorName(ustatus));
    ucol_close(ucoll);
    ucoll = nullptr;
  }
  ucol_setStrength(ucoll, UCOL_PRIMARY);
  ucol_setAttribute(ucoll, UCOL_NUMERIC_COLLATION, UCOL_ON, &ustatus);
  if (U_FAILURE(ustatus))
  {
    CLog::Log(LOGWARNING, "Cannot set icu numeric attribute: %s", u_errorName(ustatus));
    ucol_close(ucoll);
    ucoll = nullptr;
  }

  if (ucoll)
  {
    UChar *ul, *ur;
    int32_t len;

    size_t ll = wcslen(left);
    size_t lr = wcslen(right);
    ul = static_cast<UChar*>(calloc(ll + 1, sizeof(UChar)));
    u_strFromWCS(ul, ll + 1, &len, left, -1, &ustatus);
    assert(U_SUCCESS(ustatus));
    ur = static_cast<UChar*>(calloc(lr + 1, sizeof(UChar)));
    u_strFromWCS(ur, lr + 1, &len, right, -1, &ustatus);
    assert(U_SUCCESS(ustatus));

    UCollationResult ures = ucol_strcoll(ucoll, ul, -1, ur, -1);

    ucol_close(ucoll);
    free(ul);
    free(ur);

    switch (ures)
    {
      case UCOL_LESS:
        return -1;
      case UCOL_GREATER:
        return 1;
      default:
        return 0;
    }
  }
  else
  {
    // fallback to previous implementation
  }

```

pkerling:
Very nice, thanks! :+1: I agree that this is the solution we want to have (at least) for Android going forward. Question is: Do we want to bring it into v18 (and figure out how to build the wrapper etc.) or workaround for now and implement it properly in v19? Workaround would be to use the original string when `iconv` fails. I've verified that this is the case on Android when the string contains non-Latin characters, so at least we would not regress compared to v17. I'm not usually involved with Android at all and thus don't really feel qualified to make that decision. @MartijnKaijser @peak3d ?

MartijnKaijser:
Of course we should at least bring the best fix to the table for v19. Depending how evasive it is the "quick"  fix can be considered for v18. We can decide once we see the PRs for them

pkerling:
@koying Are you interested in PRing the ICU solution?

koying:
Nope. Not like I have a Kodi build tree, anyway.
Building the wrapper is just copy/paste of the libandroidjni from depends.

Reverting is way to go for 18, I guess.

pkerling:
@koying Understood. Happy christmas and thanks a lot for your input! :christmas_tree:

@MartijnKaijser An Android dev has to step up and integrate the ICU thingy for v19 (mostly infra). I'll make a feature ticket. As I'm not sure what other problems we'd encounter with the workaround, my suggestion is also to revert this for v18 so we can be sure there are no regressions at least.

PR #12273, opened 2017-06-08
Quell warnings

notspiff:
Is that cast needed for some reason I do not see?

### DIFF
--- a/xbmc/utils/StringUtils.cpp
+++ b/xbmc/utils/StringUtils.cpp
@@ -1293,9 +1293,9 @@ std::string StringUtils::FormatFileSize(uint64_t bytes)
   if (bytes < 1000)
     return Format("%" PRIu64 "B", bytes);
 
-  unsigned int i = 0;
+  size_t i = 0;
   double value = static_cast<double>(bytes);
-  while (i < static_cast<int>(units.size()) - 1 && value >= 999.5)
+  while (i + 1 < units.size() && value >= 999.5)
   {
     ++i;
     value /= 1024.0;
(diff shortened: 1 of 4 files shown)


Montellese:
Can't think of any reason. First I thought it might be that the expressing (including the `- 1` can become negative but `units` has a fixed positive size so that can't be it either.

xbmc/utils/StringUtils.cpp
```diff
@@ -1295,7 +1295,7 @@ std::string StringUtils::FormatFileSize(uint64_t bytes)
 
   unsigned int i = 0;
   double value = static_cast<double>(bytes);
-  while (i < static_cast<int>(units.size()) - 1 && value >= 999.5)
+  while (i < units.size() - 1 && value >= 999.5)
```
garbear: Should be `while (i + 1 < units.size() && ...`

xbmc/addons/settings/AddonSettings.cpp
```diff
@@ -1350,15 +1350,15 @@ bool CAddonSettings::ParseOldCondition(std::shared_ptr<const CSetting> setting,
         case SettingOptionsType::Static:
         {
           const auto& options = referencedSettingString->GetOptions();
-          if (options.size() > valueIndex)
+          if (options.size() > static_cast<size_t>(valueIndex))
```
garbear: valueIndex should be made unsigned

Montellese: Hm I already have this fixed somewhere...

xbmc/utils/StringUtils.cpp
Paxxi: i should be std::size_t and not unsigned int otherwise it'll generate warnings as size_t can be 32/64 bit

notspiff: done.

xbmc/addons/settings/AddonSettings.cpp
notspiff: done.

notspiff:
whadya know. found another one hiding within the ffmpeg deprecation warnings.

notspiff:
rebased. last time. i will just close on next conflict.

PR #12084, opened 2017-05-13
VideoPlayer: Fix uninitialized variables

garbear:
First of all, sorry to send a PR so close to the VP update merge. Feel free to merge the PR without this patch and I'll resubmit to master.

I came across an uninitialized var that led to UB when used without initialization, so I went through and made sure all vars were initialized in their constructors.

PR #12274, opened 2017-06-09
VideoPlayerRadioRDS: Revert ResetRDSCache on init

popcornmix:
## Description

Since this change we have had random crashes on playback of videos.
I believe it is due to calling g_infoManager.SetCurrentItem with an uninitialised CurrentFileItem

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayerRadioRDS.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayerRadioRDS.cpp
@@ -518,8 +518,6 @@ CDVDRadioRDSData::CDVDRadioRDSData(CProcessInfo &processInfo)
   , m_speed(DVD_PLAYSPEED_NORMAL)
   , m_messageQueue("rds")
 {
-  ResetRDSCache();
-
   CLog::Log(LOGDEBUG, "Radio UECP (RDS) Processor - new %s", __FUNCTION__);
 
   m_messageQueue.SetMaxDataSize(40 * 256 * 1024);


popcornmix:
The crash is avoided with this line commented out:
https://github.com/popcornmix/xbmc/blob/f4b3dc56068b27922ca0bd6b55c355f11afc4917/xbmc/cores/VideoPlayer/VideoPlayerRadioRDS.cpp#L633

Seg faults like:
Thread 1 (Thread 0x6d0ff3a0 (LWP 1091)):
#0  0x751bc3e4 in std::string::swap(std::string&) () from /usr/lib/libstdc++.so.6
#1  0x00565908 in CVideoInfoTag::operator=(CVideoInfoTag&&) ()
#2  0x005a3180 in CVideoDatabase::GetEpisodeInfo(std::string const&, CVideoInfoTag&, int, int) ()
#3  0x005a516c in CVideoDatabase::LoadVideoInfo(std::string const&, CVideoInfoTag&, int) ()
#4  0x007de254 in CGUIInfoManager::SetCurrentMovie(CFileItem&) ()
#5  0x007de5b8 in CGUIInfoManager::SetCurrentItemJob(std::shared_ptr<CFileItem>) ()
#6  0x007df85c in CSetCurrentItemJob::DoWork() ()
#7  0x0061531c in CJobWorker::Process() ()
#8  0x0065c574 in CThread::Action() ()

and http://sprunge.us/WTiK have been observed.

Ping @FernetMenta @garbear

FernetMenta:
@AlwinEsch please review

AlwinEsch (changes requested):
Is not a good way to prevent this crash, without the call is the RDS info not correct and can produce other problems.

Better search the reason inside this function and fix it there.

AlwinEsch:
Becomes this class now also created if a stream without RDS is started?

popcornmix:
Yes, this code is called on all files (even without RDS stream).

AlwinEsch:
Not sure think its wrong then, why need someone to open a RDS stream if it becomes never used.

popcornmix:
Master is currently broken with random crashes on playback start.
The addition of the ResetRDSCache call (https://github.com/xbmc/xbmc/pull/12084) didn't fix any actual problem (i.e. didn't solve any visible bug).
We should revert this now and any improved solution can come in later.

FernetMenta:
I wasn't awre that the problem came with #12084 
+1 for this PR

AlwinEsch:
OK
then also for me +1, Maybe this was also the reason why I not added there on begin? :)

PR #12276, opened 2017-06-09
[addons/settings] Fix fallout #2

Montellese:
These commits should fix the following problems caused by #12125:
* support string-based labels (not just id-based labels) for categories, groups and settings
* fix handling of InfoBool conditions for `enable` and `visible`
* treat settings (whose type is not `action`) as labels (new `label` control introduced)
* fix action settings not properly being triggered
* fix crash when changing settings with a complex setting callback handler
* support `option="close"` for action settings

There are still other issues that need fixing which I will cover in separate PRs because they are more complex.

## How Has This Been Tested?
Locally with add-ons provided by the people who reported above issues (thanks @basrieter and @primaeval).

## Types of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [x] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [x] All new and existing tests passed

PR #12277, opened 2017-06-09
[settings] change XML format of setting values

Montellese:
(This PR is on top of #12276.)

I've always wondered why we stored our settings in `guisettings.xml` split up into categories and settings where the categories were the first part of the settings' identifiers. This can actually cause problems if we e.g. have a setting `foo` and then a setting `foo.bar` the resulting XML will be
```xml
<settings>
  <foo>
    value
    <bar>othervalue</bar>
  </foo>
</settings>
```
which the settings library is unable to parse. With core settings this hasn't happened (yet) but I've seen a few add-ons who suffer from this since #12125. IMO it is much cleaner to just store the settings as
```xml
<settings>
  <foo>value</foo>
  <foo.bar>othervalue</foo.bar>
</settings>
```
because it makes it more obvious what the actual setting identifiers are.

The implementation still supports reading/parsing the old format (which is also used in `advancedsettings.xml`) but always writes settings to XML in the new format.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/settings/lib/SettingsManager.h
+++ b/xbmc/settings/lib/SettingsManager.h
@@ -58,7 +58,7 @@ public:
   CSettingsManager() = default;
   virtual ~CSettingsManager();
 
-  static const uint32_t Version = 1;
+  static const uint32_t Version = 2;
   static const uint32_t MinimumSupportedVersion = 0;
 
   // implementation of ISettingCreator
--- a/xbmc/settings/lib/SettingsManager.cpp
+++ b/xbmc/settings/lib/SettingsManager.cpp
@@ -759,46 +759,22 @@ bool CSettingsManager::Serialize(TiXmlNode *parent) const
     if (setting.second.setting->GetType() == SettingType::Action)
       continue;
 
-    std::string categoryTag, settingTag;
-    if (!ParseSettingIdentifier(setting.second.setting->GetId(), categoryTag, settingTag))
-    {
-      CLog::Log(LOGWARNING, "CSettingsManager: unable to save setting \"%s\"", setting.first.c_str());
-      continue;
-    }
+    TiXmlElement settingElement(SETTING_XML_ELM_SETTING);
+    settingElement.SetAttribute(SETTING_XML_ATTR_ID, setting.second.setting->GetId());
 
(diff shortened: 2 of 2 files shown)


MartijnKaijser:
Any takers on doing a review or have objections throwing this in?

xbmc/settings/lib/SettingsManager.h
```diff
@@ -58,7 +58,7 @@ class CSettingsManager : public ISettingCreator, public ISettingControlCreator,
   CSettingsManager() = default;
   virtual ~CSettingsManager();
 
-  static const uint32_t Version = 1;
+  static const uint32_t Version = 2;
```
AlwinEsch: I this OK with in settings.xml defined `<settings version="1">`?

AlwinEsch:
Have don't so many knowledge about the TinyXML but make sense.

xbmc/settings/lib/SettingsManager.h
Montellese: This is just the current version (which will be written to `guisettings.xml`). One line further down is the minimum supported version which is still at `0` which is identical to not defining a version at all (which we didn't until a few weeks ago).

Montellese:
I realized that this is not sufficient. There are add-on's that use setting identifiers that
* start with a number
* contain spaces

which both lead to invalid XMLs when we use the setting identifiers as XML tag names. That means we would have to move to a format like `<setting id="123abc">value</setting>`. Would that be acceptable for `guisettings.xml` as well? Or should I use different serializations for `guisettings.xml` and add-on settings?

MartijnKaijser:
Isn't that bad practice to use number and spaces?

Montellese:
Well there's nothing in the docs that says you can't / shouldn't do it. The setting identifier is defined as an XML attribute in `<setting id="123abc" ...>` so there it's perfectly valid.

MartijnKaijser:
Well i think you know the settings system the best of all of us so  i'll trust your judgement what is best :)

Montellese:
I've updated the code to serialize setting values as `<setting id="foo">bar</setting>` which should be the most universal format.

Montellese:
jenkins build and merge

PR #12279, opened 2017-06-09
[Python] Fix xbmcgui.Window methods to get the actual size of a Window

romanvm:
## Description
This PR fixes `xbmcgui.Window.getWidth()` and `xbmcgui.Window.getHeight()` methods so that they return the actual pixel size of a Window instance instead of the current display resolution.
Display resolution can now be obtained using `xbmcgui.getScreenWidth()` and `xbmcgui.getScreenHeight()` free functions.
Also this PR adds deprecation information to `Window.getResolution()` and `Window.setCoordinateResolution()` methods because currently they are practically useless.

## Motivation and Context
xbmcgui Controls are placed using pixel coordinates from the top left corner. However, currently there is no way to determine the the actual pixel size of the visible area (pixel coordinate grid) of a Window instance. Existing `getWidth` and `getHeight` methods return the current display resolution which is not the same as Window's coordinate grid resolution. The fixed methods will help to correctly place Controls on the screen, especially if the Controls are placed in a Window with existing ID, e.g. over a fullscreen video, and display resolution can still be obtained using the new free functions.

## How Has This Been Tested?
I compiled the code in Vusial Studio 2015 on Windows 10. Then I built a Windows installer and installed Kodi from it. After that I ran a simple testing addon and checked if those fixed methods and new functions actually work using an interactive Python debugger.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/interfaces/legacy/ModuleXbmcgui.h
+++ b/xbmc/interfaces/legacy/ModuleXbmcgui.h
@@ -83,6 +83,40 @@ namespace XBMCAddon
 #else
     long getCurrentWindowDialogId();
 #endif
+
+#ifdef DOXYGEN_SHOULD_USE_THIS
+    ///
+    /// \ingroup python_xbmcgui_window
+    /// @brief \python_func{ getScreenHeight() }
+    ///-------------------------------------------------------------------------
+    /// Returns the height of this screen.
+    ///
+    /// @return                       Screen height
+    ///
+    ///-------------------------------------------------------------------------
+    /// @python_v18 New function added.
+    ///
+    getScreenHeight();
+#else
+    long getScreenHeight();
+#endif
+
+#ifdef DOXYGEN_SHOULD_USE_THIS
+    ///
+    /// \ingroup python_xbmcgui_window
+    /// @brief \python_func{ getScreenWidth() }
+    ///-------------------------------------------------------------------------
+    /// Returns the width of this screen.
+    ///
+    /// @return                       Screen width
+    ///
+    ///-------------------------------------------------------------------------
+    /// @python_v18 New function added.
+    ///
+    getScreenWidth();
+#else
+    long getScreenWidth();
+#endif
     ///@}
 
 #ifndef DOXYGEN_SHOULD_SKIP_THIS
--- a/xbmc/interfaces/legacy/Window.cpp
+++ b/xbmc/interfaces/legacy/Window.cpp
@@ -319,7 +319,7 @@ namespace XBMCAddon
         break;
       case CGUIControl::GUICONTROL_SLIDER:
         pControl = new ControlSlider();
-        break;			
+        break;
       case CGUIControl::GUICONTAINER_LIST:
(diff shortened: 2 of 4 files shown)


xhaggi:
A window should normally return it's actually width and height not the display resolution. If that's the case we should change the implementation of `GetWidth` and `GetHeight` and do not introduce new methods. Not sure if there is another way to get the current display resolution, if not that should be addressed too.

romanvm:
I agree with you but since this my very first PR to the Kodi core code, I didn't feel right to change the existing code. But yes, logically, a `Window` instance should return its coordinate grid resolution and the display resolution, if needed, should be returned by free functions, because it is a property of Kodi itself.

Also methods `xbmcgui::Window::getResolution ()` and `xbmcgui::Window::setCoordinateResolution (...)` are basically useless because they operate with some arcane numeric constants that are meaningless from addon developer's perspective. For example, on my Win 10 machine with 1680x1050 display `getResolution()` returns `16` or `RES_DESKTOP`: https://github.com/xbmc/xbmc/blob/master/xbmc/guilib/Resolution.h#L33.

xbmc/interfaces/legacy/Window.cpp
```diff
@@ -319,7 +319,7 @@ namespace XBMCAddon
         break;
       case CGUIControl::GUICONTROL_SLIDER:
         pControl = new ControlSlider();
-        break;			
+        break;
```
DaveTBlake: Remove trailing blanks please

romanvm:
I have updated this PR. Please see the new title and description.

xhaggi:
@romanvm thank you. Some minor things to do before we are ready to merge.

Please fixup/squash commit 9206e878295f8832944c1dd801483feb7662ae46 into 4fdd5bec4590efdbb522eecb80e693baafe6ce49 and c46e7ffca5f65d8f1581c659da465233c3ea3716 into 266f75c7540757338ec05250cb005ee7b01bba51.
We don't need the history for that.

romanvm:
Done

xhaggi:
Thanks looks good.

xhaggi:
@phil65 or @ronie any objections?

xbmc/interfaces/legacy/ModuleXbmcgui.h
```diff
@@ -83,6 +83,34 @@ namespace XBMCAddon
...
+
+#ifdef DOXYGEN_SHOULD_USE_THIS
+    ///
+    /// \ingroup python_xbmcgui_window
+    /// @brief \python_func{ getScreenHeight() }
+    ///-----------------------------------------------------------------------
+    /// Returns the height of this screen.
+    ///
+    /// @return                       Screen height
+    ///
```
MartijnKaijser: Shouldn't the `@python_v18` be added here as wel?

romanvm: Yes, I forgot about that

romanvm:
@MartijnKaijser done

phil65 (approved):

PR #12281, opened 2017-06-10
[network] CHTTPJsonRpcHandler: limit any other request method than POST to ReadData permissions on the JSON-RPC API

Montellese:
This is an alternative approach to #12265 which still accepts HTTP GET requests to `/jsonrpc` but limits all non-POST requests to `ReadData` permissions. That means that it is still possible to use HTTP GET to `/jsonrpc` to get the whole API definition and it is still possible to call methods like `VideoLibrary.GetMovies`. But when trying to call a modifying JSON-RPC method like `Player.PlayPause` the following error will be returned:
```json
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32099,
    "message": "Bad client permission."
  },
  "id": 1
}
```
That way it's not possible anymore to perform any modifications using anything but HTTP POST.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/network/httprequesthandler/HTTPJsonRpcHandler.h
+++ b/xbmc/network/httprequesthandler/HTTPJsonRpcHandler.h
@@ -61,7 +61,7 @@ private:
   {
   public:
     CHTTPTransportLayer() = default;
-    ~CHTTPTransportLayer() = default;
+    virtual ~CHTTPTransportLayer() = default;
 
     // implementations of JSONRPC::ITransportLayer
     bool PrepareDownload(const char *path, CVariant &details, std::string &protocol) override;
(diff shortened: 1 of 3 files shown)


xbmc/network/httprequesthandler/HTTPJsonRpcHandler.h
```diff
@@ -73,8 +73,14 @@ class CHTTPJsonRpcHandler : public IHTTPRequestHandler
   class CHTTPClient : public JSONRPC::IClient
   {
   public:
-    virtual int  GetPermissionFlags();
-    virtual int  GetAnnouncementFlags();
-    virtual bool SetAnnouncementFlags(int flags);
+    CHTTPClient(HTTPMethod method);
+    ~CHTTPClient() = default;
```
garbear: destructor should be virtual

Montellese: will fix, thanks

tamland:
Looks good to me 👍

MartijnKaijser:
@MilhouseVH maybe you can include this PR instead of the other for testing?

MilhouseVH:
Yes

MartijnKaijser:
@MilhouseVH any negative responses?

So should we go with this version?

MilhouseVH:
No negative responses in relation to this PR.

sej7278:
this breaks calling a library update from a web browser e.g.

http://kodibox:8080/jsonrpc?request={"jsonrpc":"2.0","method":"VideoLibrary.Scan","id":"mybash"}

whilst curl works, firefox doesn't. if this isn't a bug, the wiki needs updating:

http://kodi.wiki/view/HOW-TO:Remotely_update_library#HTTP

PR #12283, opened 2017-06-10
[addons] add parts to CAddonInfo to read addon.xml

AlwinEsch:
## Description

This change is the first big major part to remove cpluff on next
steps.

On them here becomes the parts for addon info (old props) readed by
CAddonInfo class contructor, by give of path to addon or with a tinyXML part
(used later for repo parts). With them is cpluff no more needed to read
addon.xml, further is it direct designed and required to support multiple
types (e.g. PVR with inputstream or inputstream with videocodec).

Screensaver addon is with them changed to create with new way.

Temporary is for this on CAddonBuilder the from cpluff given path used for
creation of CAddonInfo.

On next steps becomes all addon types changed to this way and shortly only
the from cpluff given parts used.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/Addon.cpp
+++ b/xbmc/addons/Addon.cpp
@@ -61,13 +61,12 @@ namespace ADDON
 
 CAddon::CAddon(CAddonInfo addonInfo)
   : m_addonInfo(std::move(addonInfo))
-  , m_userSettingsPath()
   , m_loadSettingsFailed(false)
   , m_hasUserSettings(false)
   , m_profilePath(StringUtils::Format("special://profile/addon_data/%s/", m_addonInfo.ID().c_str()))
+  , m_userSettingsPath(URIUtils::AddFileToFolder(m_profilePath, "settings.xml"))
   , m_settings(nullptr)
 {
-  m_userSettingsPath = URIUtils::AddFileToFolder(m_profilePath, "settings.xml");
 }
 
 /**
@@ -377,9 +376,9 @@ CAddonSettings* CAddon::GetSettings() const
 
 std::string CAddon::LibPath() const
 {
-  if (m_addonInfo.LibName().empty())
+  if (m_addonInfo.MainLibName().empty())
     return "";
-  return URIUtils::AddFileToFolder(m_addonInfo.Path(), m_addonInfo.LibName());
+  return URIUtils::AddFileToFolder(m_addonInfo.Path(), m_addonInfo.MainLibName());
 }
 
 AddonVersion CAddon::GetDependencyVersion(const std::string &dependencyID) const
--- a/xbmc/addons/AddonInfo.cpp
+++ b/xbmc/addons/AddonInfo.cpp
@@ -21,7 +21,13 @@
 #include "AddonInfo.h"
 
 #include "Util.h"
+#include "filesystem/SpecialProtocol.h"
 #include "guilib/LocalizeStrings.h"
+#include "utils/log.h"
+#include "utils/JSONVariantParser.h"
+#include "utils/JSONVariantWriter.h"
+#include "utils/URIUtils.h"
+#include "utils/XBMCTinyXML.h"
 
 namespace ADDON
 {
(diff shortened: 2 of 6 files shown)


xbmc/addons/AddonInfo.h
```diff
@@ -81,16 +84,128 @@ namespace ADDON
...
+    const bool empty() const { return str.empty(); }
+    const std::string str;
+  };
+
+  class CAddonExtensions;
+  typedef std::vector<std::pair<std::string, CAddonExtensions> > EXT_ELEMENTS;
+  typedef std::vector<std::pair<std::string, extValue> > EXT_VALUE;
+  typedef std::vector<std::pair<std::string, EXT_VALUE> > EXT_VALUES;
+
+  class CAddonBuilder;
```
ksooo: Is this forward decl actually needed?

xbmc/addons/AddonInfo.h
```diff
@@ -81,16 +84,128 @@ namespace ADDON
...
+
+    void Insert(const std::string& id, const std::string& value);
+
+  private:
+    std::string m_point;
+    EXT_VALUES m_values;
+    EXT_ELEMENTS m_childs;
+  };
+
+  class CAddonInfo;
```
ksooo: Is this forward decl actually needed?

xbmc/addons/AddonInfo.h
```diff
@@ -81,16 +84,128 @@ namespace ADDON
...
+    }
+
+    int ProvidedSubContents() const
+    {
+      return m_providedSubContent.size();
+    }
+
+  private:
+    friend class ADDON::CAddonBuilder; // temporary until cpluff is removed
+    friend class CAddonInfo;
```
ksooo: Can't this be avoided?

xbmc/addons/AddonInfo.h
```diff
@@ -107,13 +222,18 @@ namespace ADDON
...
+    std::string MainLibPath() const;
+
     const AddonVersion& Version() const { return m_version; }
     const AddonVersion& MinVersion() const { return m_minversion; }
     const std::string& Name() const { return m_name; }
     const std::string& License() const { return m_license; }
     const std::string& Summary() const { return m_summary; }
     const std::string& Description() const { return m_description; }
-    const std::string& LibName() const { return m_libname; }
+    const std::string& LibName() const { return m_types[0].m_libname; }
```
ksooo: Again, this looks strange. Why is it always the libname for the first type?

ksooo (changes requested):
A couple of questions/suggestions...

xbmc/addons/Addon.cpp
```diff
@@ -59,6 +59,17 @@ using XFILE::CFile;
 namespace ADDON
 {
 
+  CAddon::CAddon(const std::string& addonPath)
+  : m_addonInfo(addonPath)
+  , m_userSettingsPath()
```
AlwinEsch: c&p becomes removed.

xbmc/addons/AddonBuilder.h
```diff
@@ -45,7 +45,7 @@ class CAddonBuilder
   void SetChangelog(std::string changelog) { m_addonInfo.m_changelog = std::move(changelog); }
   void SetBroken(std::string broken) { m_addonInfo.m_broken = std::move(broken); }
   void SetPath(std::string path) { m_addonInfo.m_path = std::move(path); }
-  void SetLibName(std::string libname) { m_addonInfo.m_libname = std::move(libname); }
+  void SetLibName(std::string libname) { m_addonInfo.m_types[0].m_libname = std::move(libname); }
```
AlwinEsch: This is a temporary for old cpluff where always only one lib name is. On new way can be this no more direct accessed.

For a binary can it be only one, but on Python is it mostly on different scripts.

xbmc/addons/AddonInfo.cpp
```diff
@@ -77,6 +83,207 @@ static const TypeMapping types[] =
    {"kodi.imagedecoder",                 ADDON_IMAGEDECODER,        39015, "DefaultAddonImageDecoder.png" },
   };
 
+bool CAddonExtensions::ParseExtension(const TiXmlElement* element)
+{
+  const char* cstring;
+
+  cstring = element->Attribute("point");
```
AlwinEsch: Can change, it becomes set also from another places thats why I leaved alone.

xbmc/addons/AddonInfo.cpp
```diff
@@ -77,6 +83,207 @@ static const TypeMapping types[] =
...
+}
+
+const EXT_VALUES& CAddonExtensions::GetValues() const
+{
+  return m_values;
+}
+
+const CAddonExtensions* CAddonExtensions::GetElement(const std::string& id) const
+{
+  for (EXT_ELEMENTS::const_iterator it = m_childs.begin(); it != m_childs.end(); ++it)
```
AlwinEsch: Is better, still have mostly old style in mind ;-)

xbmc/addons/AddonInfo.h
AlwinEsch: Here it is needed in future for binary addons. I think to change the name to `MainLibName()`.

tamland:
Getting rid of the cpluff parser is great!, so sorry to be blunt here, but design wise this turning into a huge mess. Everything you are trying to do here have already been solved a long time ago!

What you need to do instead is:

1. Add a new factory method similar to [this](https://github.com/xbmc/xbmc/blob/a0a04d6011eedfd149709669e3526cd5cd1abce1/xbmc/addons/AddonManager.h#L254) that uses tinyXML instead of cpluff. (Or even better: make a new factory class)

2. Move all xml parsing logic from CAddonInfo and CAddonType constuctors to here. 

3. You then pass the parsed info, including a fully constructed vector<CAddonType> to the CAddonInfo constructor

xbmc/addons/AddonInfo.cpp
```diff
@@ -151,7 +414,448 @@ CAddonInfo::CAddonInfo(std::string id, TYPE type)
...
+
+bool CAddonInfo::ProvidesSeveralSubContents() const
+{
+  int contents = 0;
+  for (auto addonType : m_types)
+    contents += addonType.ProvidedSubContents();
+  return contents > 0 ? true : false;
+}
+
+std::string CAddonInfo::SerializeMetadata()
```
tamland: This is a database implementation detail and is tied to the schema and migration. This should *not* be moved into to a generic class, but kept as a private function in CAddonDatabse.

xbmc/addons/AddonInfo.cpp
```diff
@@ -151,7 +414,448 @@ CAddonInfo::CAddonInfo(std::string id, TYPE type)
...
+      TYPE type = TranslateType(point);
+      if (type == ADDON_UNKNOWN || type >= ADDON_MAX)
+      {
+        CLog::Log(LOGERROR, "CAddonInfo: file '%s' doesn't contain a valid add-on type name (%s)", addonPath.c_str(), point.c_str());
+        return false;
+      }
+
+      m_types.push_back(CAddonType(type, this, child));
+    }
+  }
```
tamland: There are multiple errors in this method compared to original cpluff implementation. They are easily found by running it again the existing tests: https://github.com/xbmc/xbmc/blob/a0a04d6011eedfd149709669e3526cd5cd1abce1/xbmc/addons/test/TestAddonFactory.cpp

xbmc/addons/AddonInfo.cpp
AlwinEsch: This is a Chicken and egg problem,

CAddonInfo should hold everything private to allow only a read of the values but no changes.
Reason behind is that the class is on end only created one time and hold in a list from AddonManager. If some part need then something, becomes the already stored CAddonInfo from the list used.

Only way but also not the best is to add the database as "friend" to CAddonInfo.

See also the coming parts on other comment.

AlwinEsch:
Hi @tamland, to remove the cpluff is not in a easy small step possible.

To change the complete addon system to new way is a big part.

Currently is this CAddonInfo still as static in CAddon class. This is must be removed complete (on next requests) and only given as shared_ptr to the addon class.

Goal is (and used on kodi-agile) that Kodi e.g. go over the addon directory, if there is a addon with addon.xml becomes this by the constructor of CAddonInfo read and the class stored in a list on addon manager. What Kodi does with it does not interest. This mean also a related addon class is stll not needed and created.

After all steps are done contains the addon class only a `CAddon(std::shared_ptr<CAddonInfo> addonInfo)` contructor where the stored parts becomes given.

XML parsing must only be done inside this class and should not done somewhere outside. To do outside bring a lot bad parts and nothing good.

Big problem is to bring in without destroy something and allow a better review. This need for short times bad parts. Thats why I mostly use screensaver where is not so critical.

tamland:
> XML parsing must only be done inside this class and should not done somewhere outside. To do outside bring a lot bad parts and nothing good.

Sorry but this is a horrible idea. There nothing 'bad' about keeping it outside. It's called the factory pattern and it's great for this sort of thing. Not too long ago this exact 'inside this class' spaghetti was removed. And the reasons for that was:

- Having complex logic like parsing and filesystem code in constructor is an anti-pattern. 

- It breaks testability.

- It makes error handing complicated as your only option is to leave objects in an invalid state and work around it by setting flags (this is exactly what you have done with m_usable)

- It's a pain to deal with configuration as every new config must be added to every constructors. 

- Makes it a lot harder to construct something partially when needed. E.g. local addons have metadata from the filesystem and from the database, and potentially also from the repository and they must all be combined.

Additionally, the way this is implemented violates the single responsibility principle. `CAddonInfo` now holds the info, it holds state, it's an xml parser, it's an json parser, it's responsible for the extensions. 
The same for `CAddonType`: it's hold extension info, it's a parser, it's a mutable collection. This is all effects of having the classes themselves do the parsing.

The solution I explained in my previous post does not have these issues.

xbmc/addons/AddonInfo.cpp
tamland: Yet, in the current code all values are read only, this function is in CAddonDatabse, and there are no friend classes needed. 

If you need to break that to make it "fit" your solution, then that solution is the wrong one.

AlwinEsch:
Maybe I improve again

AlwinEsch:
Excuse me, but after over two years I want and I have to come to an end.

Always comes a no and see how the current addon interface is getting more and more mixed up or hacking paths to achieve a goal.

Since I am surprised again and again something larger is rejected by me but otherwise which is unreliable somewhere else comes in.

Excuse me that I bring these words, we are a good team.

But I do not want again this cross and cross between the different parts, this brings again and again problems since the one then no longer fits with the other.

So far, where the AddonProps was recreated with every little thing, it was easy, but since in the future this is done only once must be ensured that does not change from now on equal in what.

**I look at what I can do, but I want to keep a few things from it.**

Sorry for my words

AlwinEsch:
I create now a another request to remove the static declaration of CAddonInfo (AddonProps) in CAddon and pass as shared pointer to them.

Is a part also required for them here.

tamland:
@AlwinEsch You can't just throw out all the existing code and design choices, make a complete re-implementation with tons of issues, and then expect everything to be just fine. (That's not collaboration).

I have tried to tell you this and propose better solutions multiple times, in this pr and others, but you seem too fixated on continuing down path you have already decided on, ignoring any suggestions.

Why don't you just *try* and do what [I proposed earlier](https://github.com/xbmc/xbmc/pull/12283#issuecomment-307575101)? Everything you want to achieve with removing cpluff and supporting multiple types fits well within the current design with minor changes. I know it does because everything here has already been made with this in mind long before you started with it.

ksooo:
For me, what @tamland suggested makes sense. I especially would be happy if he's right on that not everything needs to be reimplemented because the existing design already fits well an can be extended with not too much effort.

ksooo:
Maybe you two guys should discuss this 1:1 on Slack?

xbmc/addons/AddonInfo.cpp
AlwinEsch: It is a from database used part complete right!

But I must say again, to have in Database is complete bad!

It use values from CAddonInfo and use only a string as "in" or "out" for database. All used values from the string are from CAddonInfo.

**In case now new values becomes added (removed bring God thanks to a compile error), is the database be easily forgotten!**

Further is CAddonInfo not a generic class, it is the base addon class (and on end only one time for one addon present) to declare the whole addon in Kodi.

xbmc/addons/AddonInfo.cpp
AlwinEsch: The existing tests must be complete reworked after cpluff is removed. Everything related on current test parts must be ignored for this.

If you see a error please describe me detailed what it bring.

xbmc/addons/AddonInfo.h
AlwinEsch: Is for the temporary declared as child class below

xbmc/addons/AddonInfo.h
AlwinEsch: Is for the contructor on CAddonType and also is CAddonType used from CAddonInfo, thats why the decl was needed.

xbmc/addons/AddonInfo.h
AlwinEsch: Yes, was only a test mistake from.

AlwinEsch:
I write you something on slack.

But I say, **not** expect everything to be just fine, never possible on world. But it must come a day where the old way becomes changed to a new.

AlwinEsch:
Is updated once, thanks @ksooo for the comments.

But not yet final.

xbmc/addons/AddonInfo.cpp
```diff
@@ -142,7 +352,63 @@ CAddonInfo::CAddonInfo()
...
+
+  m_usable = LoadAddonXML(xmlDoc.RootElement(), m_path);
+}
+
+CAddonInfo::CAddonInfo(const TiXmlElement* baseElement, const std::string& addonRepoXmlPath)
+  : m_usable(false),
+    m_mainType(ADDON_UNKNOWN),
+    m_packageSize(0)
+{
+  m_usable = LoadAddonXML(baseElement, URIUtils::AddFileToFolder(addonRepoXmlPath, m_id));
```
ksooo: This could be moved to init list.

xbmc/addons/AddonInfo.cpp
```diff
@@ -142,7 +352,63 @@ CAddonInfo::CAddonInfo()
...
+    m_mainType(ADDON_UNKNOWN),
+    m_version(version),
+    m_name(name),
+    m_summary(summary),
+    m_description(description),
+    m_changelog(changelog),
+    m_origin(origin),
+    m_packageSize(0)
+{
+  m_usable = DeserializeMetadata(metadata);
```
ksooo: This could be moved to init list.

xbmc/addons/AddonInfo.cpp
tamland: The point here is having some basic separation of concern. CAddonDatabse is responsible for mapping an unmapping. By exposing the raw unmapping `metadata` value, CAddonInfo now has to details about how things are stored in the database.

I don't see you put reading/writing from the db `Dataset` in this class so why do it with `metadata`? (hint: you don't because mapping of `Dataset` is the entire job of the database). The `metadata` value is the same thing as `Dataset`, it just happens to be json right now.

tamland: And here's a lengthy explanation for why it's done this way: https://msdn.microsoft.com/en-us/library/ff649690.aspx

FernetMenta: even Microsoft architects call this an anti pattern: http://www.infoworld.com/article/3117713/application-development/design-patterns-that-i-often-avoid-repository-pattern.html

tamland: @FernetMenta you should read the link you post.. He call the *generic* repository pattern (over entity frameworks) an anti-pattern. And I agree.

PR #12284, opened 2017-06-10
[droid] Packaging cleanup

koying:
Sync with SPMC.

The java files move prevents Android Studio from complaining if package name != directory structure for rebranded projects.

I also have some basic gradle files for Android Studio (just sufficient to be able to open Kodi in it), but not sure if that would be accepted in main tree as it adds a "build.gradle" in the root.

### DIFF
--- a/tools/android/packaging/Makefile.in
+++ b/tools/android/packaging/Makefile.in
@@ -129,8 +129,8 @@ res:
 	@rm -rf tmp/
 
 libs: $(PREFIX)/lib/@APP_NAME_LC@/lib@APP_NAME_LC@.so
-	rm -rf xbmc/lib/$(CPU) xbmc/libs/$(CPU) xbmc/obj/local/$(CPU)
-	mkdir -p xbmc/lib/$(CPU) xbmc/assets/python2.7/lib/ xbmc/libs/$(CPU) xbmc/obj/local/$(CPU)
+	rm -rf xbmc/lib/$(CPU) xbmc/obj/local/$(CPU)
+	mkdir -p xbmc/lib/$(CPU) xbmc/assets/python2.7/lib/ xbmc/obj/local/$(CPU)
 	cp -fpL $(SRCLIBS) xbmc/obj/local/$(CPU)/
 	cp -fp $(PREFIX)/lib/@APP_NAME_LC@/lib@APP_NAME_LC@.so xbmc/obj/local/$(CPU)/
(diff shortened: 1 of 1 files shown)


hudokkow:
Sorry @koying, can't do much reviewing here. Droids + makefiles is not even close to my territory.

wsnipex (commented):
to be complete, at least https://github.com/xbmc/xbmc/blob/master/cmake/scripts/android/Install.cmake#L32-L51 should also be adjusted to not hardcode the path anymore.

Otherwise, looks good to me.

wsnipex:
untested: http://paste.ubuntu.com/24823798/

koying:
@wsnipex If not mistaken, those are the "source" java files, generated from the "*.in" in https://github.com/xbmc/xbmc/blob/master/cmake/scripts/android/Install.cmake#L52-L55, mimicking an autoconf setup.

The .in are still in src/org/xbmc/kodi, and the Makefile still expects the . java there as well , so I don't think your patch will work.

koying:
jenkins build this please

wsnipex:
the patch should work at least as long as the package_name is org.xbmc.kodi as it only replaces the hardcoded path with a generated one. But yes, it will fail for a different name unless the java.in files are moved. 
This poses the following questions:
- do we want rebranders to move those java files
- or do we want to automate this
- or just skip this patch and leave as is

koying:
No source files are moved by this one, it just changes packaging so that it's usable by android studio *including* for rebranded app. 

Basically, it makes *my* life easier as it eases backporing to Kodi ;)

koying:
@wsnipex Still comments?

I could take your patch and do something like (removing the paths from package_files)
```
foreach(file IN LISTS package_files)
  configure_file(${CMAKE_SOURCE_DIR}/tools/android/packaging/xbmc/src/org/xbmc/kodi/${file}.in
                 ${CMAKE_BINARY_DIR}/tools/android/packaging/xbmc/java/${APP_PACKAGE_DIR}/${file} @ONLY)
endforeach()
```

But that would be splitting the android packaging process between cmake and the actual packaging Makefile, which makes it hardly readable, for no practical advantage.

Call me old-fashioned, but I like being able to understand code by looking at just one place (reason why I prefer IFDEF's rather than obscure cmake magic ;) ) and I'm sure it makes it easier for potential contributors as well.

Actually, we could remove the whole "org/xbmc/kodi" tree and put everything in src.
Package name = directory name is a java convention, it doesn't hinder compilation at all if not followed, but it makes IDEs choking, and editing java files with vim is no fun ;)

wsnipex:
I'm fine with merging as is.

koying:
Thanks

PR #12288, opened 2017-06-11
[jenkins] - ensure that depends are rebuilt when darwim arm arch is changing

Memphiz:
When changing between armv7 and arm64 on the ios job - depends were not cleaned.

@MartijnKaijser i think that might be the reason why you had success with master and ios-arm64 - i cleared the workspace and can see that it doesn't build for arm64 actually (because of the missing fix in ffmpeg).

@Rechi fyi ping as you had arm64 fixes in before too:

this is the printout https://pastebin.com/dBBG1C01

where i think only the first one is the cause and the others are follow up errors. We are missing this commit from ffmpeg upstream:

https://github.com/FFmpeg/FFmpeg/commit/2425d7329fdccfa9954faba748f3865151354f0c

@FernetMenta fyi

### DIFF
--- a/tools/buildsteps/defaultenv
+++ b/tools/buildsteps/defaultenv
@@ -88,7 +88,7 @@ function getBuildHash ()
   checkPath="$1"
   local hashStr
   hashStr="$(git rev-list HEAD --max-count=1  -- $checkPath)"
-  hashStr="$hashStr $SDK_PATH $NDK_PATH $NDK_VERSION $SDK_VERSION $TOOLCHAIN $XBMC_DEPENDS_ROOT"
+  hashStr="$hashStr $SDK_PATH $NDK_PATH $NDK_VERSION $SDK_VERSION $TOOLCHAIN $XBMC_DEPENDS_ROOT $DARWIN_ARM_CPU"
   echo $hashStr
 }
 


MartijnKaijser:
@Memphiz is fails for iOS

wsnipex:
is this really needed? Depends should have their own build dir for every platform and architecture.

Rechi:
The `DARWIN_ARM_CPU` parameter at the jenkins job for IOS and IOS-ARM64 should be removed then this change isn't necessary.
FFmpeg may need fixes for Xcode 6.1.1, I only tested it with 7.3.1 and there it works.

Memphiz:
You are right - this is superfluous b cause of splitting it up in 2 jobs. Well that only points out that the arm64 job can't work yet due to the mssing ffmpeg commit i pointed out in the PR description. @FernetMenta can we backport that one commit?

FernetMenta:
yes, will do this evening

FernetMenta:
@Memphiz https://github.com/xbmc/FFmpeg/releases/tag/3.3.2-Leia-Alpha

Memphiz:
@FernetMenta thx - testbuild running ... closing this PR as it will be unneeded

PR #12289, opened 2017-06-12
[adsp] change addon interface to match new style

AlwinEsch:
## Description
This commit change the audio DSP addon system to the new interface
style.

This becomes be improved much in future.

Now only to match with the coming style!

## Types of change

## Checklist:


AchimTuran:
Jenkins build this please

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
```diff
@@ -0,0 +1,1288 @@
...
+    /// Only one DSP add-on is allowed to-do this!
+    /// @param handle identification data for stream
+    /// @param array_in Pointer to input data memory
+    /// @param array_out Pointer to output data memory
+    /// @param samples Amount of samples inside array_in
+    /// @return Amount of samples processed
+    /// @remarks Optional. Is set by AE_DSP_ADDON_CAPABILITIES and asked with
+    /// GetCapabilities
+    ///
+    virtual unsigned int MasterProcess(const ADDON_HANDLE handle, float **array_in, float **array_out, unsigned int samples) { return 0; }
```
ksooo: The pointer to pointer stuff looks odd. Can't you use (const) references here?

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
```diff
@@ -0,0 +1,1288 @@
...
+    /// RegisterMode callback, and can be defined from add-on as a structure
+    /// pointer or anything else what is needed to find it.
+    /// @param array_in Pointer to input data memory
+    /// @param array_out Pointer to output data memory
+    /// @param samples Amount of samples inside array_in
+    /// @return Amount of samples processed
+    /// @remarks Optional. Is set by AE_DSP_ADDON_CAPABILITIES and asked with
+    /// GetCapabilities
+    ///
+    virtual unsigned int PostProcess(const ADDON_HANDLE handle, unsigned int mode_id, float **array_in, float **array_out, unsigned int samples) { return 0; }
```
ksooo: Same here. Use refs instead of pointer to pointer?

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
```diff
@@ -0,0 +1,1288 @@
...
+    /// one add-on can perform it.
+    /// @param handle identification data for stream
+    /// @param array_in Pointer to input data memory
+    /// @param array_out Pointer to output data memory
+    /// @param samples Amount of samples inside array_in
+    /// @return Amount of samples processed
+    /// @remarks Optional. Is set by AE_DSP_ADDON_CAPABILITIES and asked with
+    /// GetCapabilities
+    ///
+    virtual unsigned int OutputResampleProcess(const ADDON_HANDLE handle, float **array_in, float **array_out, unsigned int samples) { return 0; }
```
ksooo: Again, point to pointer vs. refs

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AchimTuran: const for what?

I think only handle and array_in makes sense for const.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AchimTuran: No. Pointers are the right way for 2 dimensional  arrays like the audio channels. But array_in could be made const.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AchimTuran: See my privious comments.

xbmc/addons/Addon.cpp
```diff
@@ -413,8 +412,7 @@ void OnDisabled(const std::string& id)
 {
 
   AddonPtr addon;
-  if (CAddonMgr::GetInstance().GetAddon(id, addon, ADDON_PVRDLL, false) ||
-      CAddonMgr::GetInstance().GetAddon(id, addon, ADDON_ADSPDLL, false))
+  if (CAddonMgr::GetInstance().GetAddon(id, addon, ADDON_PVRDLL, false))
```
AchimTuran: @AlwinEsch 
Thanks for cleaning up the old mess. Looks good.

xbmc/addons/AddonBuilder.cpp
```diff
@@ -199,13 +197,12 @@ AddonPtr CAddonBuilder::FromProps(CAddonInfo addonInfo)
       return AddonPtr(new CScraper(std::move(addonInfo)));
     case ADDON_SKIN:
       return AddonPtr(new CSkinInfo(std::move(addonInfo)));
+    case ADDON_ADSPDLL:
     case ADDON_VIZ:
     case ADDON_SCREENSAVER:
       return AddonPtr(new CAddonDll(std::move(addonInfo)));
```
AchimTuran: Sorry for the dump question, do all add-on extenstion points have to use CAddonDll as a base class? If so this makes totally sense to me. First time that I really look into your work.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
```diff
@@ -0,0 +1,1288 @@
...
+    AE_DSP_CH_MAX
+  } AE_DSP_CHANNEL;
+
+  /*!
+   * @brief Present channel flags
+   */
+  typedef enum
+  {
+    AE_DSP_PRSNT_CH_UNDEFINED = 0,
+    AE_DSP_PRSNT_CH_FL        = 1<<0,
```
AchimTuran: I don't know if it also makes sense for you @AlwinEsch, but we could use `AE_DSP_PRSNT_CH_FR        = 1<<AE_DSP_CH_FL,` ... here.
So we only have to define the number of the channels and then define on top of them the channel mask.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
```diff
@@ -0,0 +1,1288 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with Kodi; see the file COPYING.  If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "../AddonBase.h"
```
AchimTuran: @AlwinEsch 
Please try to avoid relative paths. I had so many with them in the past.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
```diff
@@ -0,0 +1,1288 @@
...
+    /// Can be used to have unchanged stream..
+    /// All DSP add-ons allowed to-do this.
+    /// @param handle identification data for stream
+    /// @param array_in Pointer to data memory
+    /// @param samples Amount of samples inside array_in
+    /// @return true if work was OK
+    /// @remarks Optional. Is set by AE_DSP_ADDON_CAPABILITIES and asked with
+    /// GetCapabilities
+    ///
+    virtual bool InputProcess(const ADDON_HANDLE handle, const float **array_in, unsigned int samples) { return true; }
```
AchimTuran: @AlwinEsch 
Look here we already have a const pointer for input data. I think this is the way to go.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
```diff
@@ -0,0 +1,1288 @@
...
+      m_instanceData->toAddon.post_process_get_delay = ADDON_PostProcessGetDelay;
+      m_instanceData->toAddon.post_process = ADDON_PostProcess;
+
+      m_instanceData->toAddon.output_resample_process_needed_samplesize = ADDON_OutputResampleProcessNeededSamplesize;
+      m_instanceData->toAddon.output_resample_process = ADDON_OutputResampleProcess;
+      m_instanceData->toAddon.output_resample_get_delay = ADDON_OutputResampleGetDelay;
+      m_instanceData->toAddon.output_resample_samplerate = ADDON_OutputResampleSampleRate;
+    }
+
+    static inline void ADDON_GetCapabilities(AddonInstance_AudioDSP* instance, AE_DSP_ADDON_CAPABILITIES *capabilities)
```
AchimTuran: If `AE_DSP_ADDON_CAPABILITIES` is a struct please use `const AE_DSP_ADDON_CAPABILITIES&` as it avoids crashes in the case we have a `nullptr`.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
```diff
@@ -0,0 +1,1288 @@
...
+    {
+      instance->toAddon.addonInstance->GetCapabilities(*capabilities);
```
AchimTuran: Here it would crash Kodi if `capabilities` is a `nullptr` because it was not checked anywhere.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
```diff
@@ -0,0 +1,1288 @@
...
+
+    static inline const char* ADDON_GetDSPName(AddonInstance_AudioDSP* instance)
+    {
+      instance->toAddon.addonInstance->m_dspName = instance->toAddon.addonInstance->GetDSPName();
+      return instance->toAddon.addonInstance->m_dspName.c_str();
+    }
+
+    static inline const char* ADDON_GetDSPVersion(AddonInstance_AudioDSP* instance)
+    {
+      instance->toAddon.addonInstance->m_dspVersion = instance->toAddon.addonInstance->GetDSPVersion();
```
AchimTuran: Why the assignment? Does it make sense to return `instance->toAddon.addonInstance->GetDSPVersion()` imidiatly?

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
```diff
@@ -0,0 +1,1288 @@
...
+    }
+
+    static inline void ADDON_GetCapabilities(AddonInstance_AudioDSP* instance, AE_DSP_ADDON_CAPABILITIES *capabilities)
+    {
+      instance->toAddon.addonInstance->GetCapabilities(*capabilities);
+    }
+
+    static inline const char* ADDON_GetDSPName(AddonInstance_AudioDSP* instance)
+    {
+      instance->toAddon.addonInstance->m_dspName = instance->toAddon.addonInstance->GetDSPName();
```
AchimTuran: Why the assignment? Does it make sense to return instance->toAddon.addonInstance->GetDSPVersion() imidiatly?

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
```diff
@@ -0,0 +1,1288 @@
...
+    }
+
+    static inline unsigned int ADDON_MasterProcess(AddonInstance_AudioDSP* instance, const ADDON_HANDLE handle, float **array_in, float **array_out, unsigned int samples)
+    {
+      return instance->toAddon.addonInstance->MasterProcess(handle, array_in, array_out, samples);
+    }
+
+    static inline const char* ADDON_MasterProcessGetStreamInfoString(AddonInstance_AudioDSP* instance, const ADDON_HANDLE handle)
+    {
+      instance->toAddon.addonInstance->m_streamInfoString = instance->toAddon.addonInstance->MasterProcessGetStreamInfoString(handle);
```
AchimTuran: Why the assignment? Does it make sense to return instance->toAddon.addonInstance->GetDSPVersion() imidiatly?

AchimTuran:
From the first review all is looking fine. I will try a next round during the next days. If I see it correctly it does not really damage anything because AudioDSP is disabled anyway. So a merge should be save.

But I guess that I have to adapt some bits of the new API, because AudioDSP V2 is a bit different compared to the current implementation.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AlwinEsch: Here on addon headers not so easy to include. On addon can we not go with the on Kodi's cmake defined include points.
Only alternative can be <kodi/AddonBase.h> but to lead with them can bring another problems.

AchimTuran: > Only alternative can be <kodi/AddonBase.h> but to lead with them can bring another problems.

From my point of view, a much cleaner way. If that is easily possible please use this way. You can not estimate where this file will be included. But with <kodi/AddonBase.h> it is very clear and you have to define a proper include path.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AlwinEsch: I check, not sure if it works to use a enum in enum (maybe the reason why I done with number), if works is it really better.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AlwinEsch: @ksooo how you see includes on headers from external positions?

Normally use the addon `#include <kodi/..>`, also OK then if inside related headers? And the other reason if someone go again over the old way with `#include "AddonBase.h"` on his addon?

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AlwinEsch: With `%` does not match "C", must be leaved as `*`, sorry.

AchimTuran: You are right @AlwinEsch, I forgot that these static methods are used as the callbacks. So they have to be C compatible right?

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
```diff
@@ -0,0 +1,1288 @@
...
+  {
+  public:
+    //==========================================================================
+    /// @brief Class constructor
+    ///
+    CInstanceAudioDSP()
+      : IAddonInstance(ADDON_INSTANCE_ADSP)
+    {
+      if (CAddonBase::m_interface->globalSingleInstance != nullptr)
+        throw std::logic_error("kodi::addon::CInstanceAudioDSP: Creation of more as one in single instance way is not allowed!");
```
AlwinEsch: With `throw` is it not direct possible. Only can it then done with a:
```cpp
std::logic_error(StringUtils::Format("kodi::addon::%s: Creation of more as one in single instance way is not allowed!", __FUNCTION__).c_str())
```
But this is for a part who normally never come a bit much :)

AlwinEsch: Ah forgot, about the StringUtils.  Want to prevent to have the headers with depends of external parts, in this case p8platform.

AlwinEsch: Only externel lib who I have a bit in mind is to increase the https://github.com/xbmc/kodi-platform and have as default included in the dev-kit.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AlwinEsch: There are you fully right, to allow a change of input stream was a bit bad thing in my mind.

Thought on begin with them to allow the input also as temporary array to use for math parts. But was a bit stupid from me :)

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AlwinEsch: @ksooo how exact you mean with "refs"?

You mean to use as `const float *array_in[], float *array_out[]`, with the `[]`?

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AlwinEsch: It works, thanks a lot!
Is then more safe on changes.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AlwinEsch: Right, but only on this place. The function direct on addon c++ class `virtual void GetCapabilities(AE_DSP_ADDON_CAPABILITIES& capabilities) = 0;` is with `&`

AlwinEsch: With the `const` is not possible, this must be edited by addon :)

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AlwinEsch: I add a check there, but normally should this never happen, see https://github.com/AlwinEsch/kodi/blob/change-adsp/xbmc/cores/AudioEngine/Engines/ActiveAE/AudioDSPAddons/ActiveAEDSPAddon.cpp#L140-L144

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AlwinEsch: Not complete sure in cases like them, normally is the returned `std::string` destructed from c++. If now this becomes given as "C" string to Kodi, is the class no longer available from my understanding.

Thats also a reason why mostly on "C" interface wrongly it was used like them:
```cpp
const char *GetBackendVersion(void)
{
  static std::string strBackendVersion = "0.0.0";
  return strBackendVersion.c_str();
}
```
there can a string direct returned;

AchimTuran: Ah right @AlwinEsch, I forgot about that use case. Your way now makes a lot of sense. Forget what I had written.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
ksooo: I'm a fan of absolute paths here, e.g. #include "kodi/foo.h".

But, as we're currently using relative paths all over the place in the new add-on framework, may I suggest to change this everywhere, in a follow-up PR?

AlwinEsch: Think also so with the fix about them: https://github.com/notspiff/visualization.spectrum/blob/master/CMakeLists.txt#L64 (My comment with Hack Way)

Then is forced to use <kodi/...>

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
ksooo: Forget about my comment re __FUNCTION__. It's not worth the hassle. I overlooked that place holders  can't be used easily here. Keep it as it is, please.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
```diff
@@ -0,0 +1,1288 @@
...
+    /// Can be used to have unchanged stream..
+    /// All DSP add-ons allowed to-do this.
+    /// @param handle identification data for stream
+    /// @param array_in Pointer to data memory
+    /// @param samples Amount of samples inside array_in
+    /// @return true if work was OK
+    /// @remarks Optional. Is set by AE_DSP_ADDON_CAPABILITIES and asked with
+    /// GetCapabilities
+    ///
+    virtual bool InputProcess(const ADDON_HANDLE handle, const float** array_in, unsigned int samples) { return true; }
```
ksooo: I still would prefer std containers over raw arrays in c++, but ...

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
ksooo: No check if is a precondition. If the caller  violates the contract, crash please. You could add an assert(), if you want to be nice. ;-)

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AchimTuran: @ksooo
I know that this would be better, but our general approach is to use raw C types for communication between binary add-ons and Kodi core. STD containers will always create problems if the compiler for the binary addon was not the same as for Kodi.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AchimTuran: Sure that's also a valid way.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
ksooo: That's not the point. I do not suggest to use std containers for the addon C interface, which would be a nogo. This comment relays to a c++ member function and for c++ I would almost in all cases prefer std containers over raw arrays. Exception could be for performance reasons.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AchimTuran: Great! No problem.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AchimTuran: I know it's a hacky way, but I think that's the way to go until we have fixed that in the CMake build system.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
AchimTuran: It has to be a double pointer to a an array because of the performance and how AE manages the audio stream packets. All audio packets in AE are based on [CSoundPacket](https://github.com/AchimTuran/xbmc/blob/AudioDSP/v2/core-reimplementation_V18/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAEBuffer.cpp#L34) and the [CActiveAE::AllocSoundSample](https://github.com/AchimTuran/xbmc/blob/AudioDSP/v2/core-reimplementation_V18/xbmc/cores/AudioEngine/Engines/ActiveAE/ActiveAE.cpp#L3087-L3098) method. Also in future AudioDSP V2 will use this class.

AlwinEsch: For me the same reason

AchimTuran: @AlwinEsch
I guess it is better define the return value as false per default.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/AudioDSP.h
```diff
@@ -0,0 +1,1298 @@
...
+    //--------------------------------------------------------------------------
+
+    //==========================================================================
+    ///
+    /// Used to get a information string about the processed work to show on skin
+    /// @return A string to show
+    /// @remarks Optional. Is set by AE_DSP_ADDON_CAPABILITIES and asked with
+    /// GetCapabilities
+    ///
+    virtual std::string MasterProcessGetStreamInfoString(const ADDON_HANDLE handle) { return ""; }
```
AchimTuran: @AlwinEsch
Will the std::string return value passed through our C interface boundaries?

AlwinEsch:
Is updated to match new multiple instance way after last changes.

Also created a request for adsp.freesurround to have one to check.

**WARNING** Audio DSP currently not working and addon can not be used!

PR #12299, opened 2017-06-14
[addons] make binary addon management independent

AlwinEsch:
## Description
This changes add a new binary addon manager who handle the
addons on request.

This way allow the use of several instances on the same addon.

Currently only used on Visualization and Screensaver, but other binary
addon types becomes on next steps changed to use them here.

Further becomes on next changes everything related to binary addons
moved to the new folder "./xbmc/addons/binary-addons".

The old system is not changed!

## Types of change

## Checklist:


tamland:
Yes, architecturally I think this is fine.

Ideally GetInstalledBinaryAddonPaths would return a CAddon with all the info CBinaryAddonManger needs instead of the paths directly. But if you can't make it work, that's ok. I will be easy to fix later once CAddonDll and CAddon are separated.

xbmc/addons/AddonManager.h
```diff
@@ -117,6 +117,8 @@ namespace ADDON
 
     bool GetInstallableAddons(VECADDONS& addons, const TYPE &type);
 
+    bool GetInstalledBinaryAddonPaths(std::vector<std::pair<std::string, bool>>& binaryAddonList);
```
ksooo: Could you please add documentation for the new member function? Especially the content of the vector is not self explaining.

xbmc/addons/binary-addons/BinaryAddonBase.cpp
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+#include "BinaryAddonBase.h"
+
+using namespace ADDON;
+
+CBinaryAddonBase::CBinaryAddonBase(const std::string& addonPath)
+{
+
+  m_usable = false;
```
ksooo: blank line should be removed. can "m_usable" initialized in ctor init list?

xbmc/addons/binary-addons/BinaryAddonBase.cpp
```diff
@@ -0,0 +1,34 @@
...
+}
+
+CBinaryAddonBase::~CBinaryAddonBase()
+{
+
+}
```
ksooo: No need for this. "~CBinaryAddonBase() = default;" in the header file is sufficient.

xbmc/addons/binary-addons/BinaryAddonBase.h
```diff
@@ -0,0 +1,48 @@
...
+  typedef std::shared_ptr<CBinaryAddonBase> BinaryAddonBase;
+
+  class CBinaryAddonBase
+  {
+  public:
+    CBinaryAddonBase(const std::string& addonPath);
+    ~CBinaryAddonBase();
+
+    const std::string& ID() { return m_id; }
+    const std::string& Path() { return m_path; }
```
ksooo: these two member functions should be "const".

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include "addons/AddonManager.h"
+#include "utils/log.h"
+
+using namespace ADDON;
+
+CBinaryAddonManager::CBinaryAddonManager()
+{
+
+}
```
ksooo: no need for an empty definition. "= default;" in the header does the trick.

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+CBinaryAddonManager::~CBinaryAddonManager()
+{
+
+}
```
ksooo: no need for an empty definition. "= default;" in the header does the trick.

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+bool CBinaryAddonManager::Init()
+{
+  std::vector<std::pair<std::string, bool>> binaryAddonList;
+  if (!CAddonMgr::GetInstance().GetInstalledBinaryAddonPaths(binaryAddonList))
+  {
+    CLog::Log(LOGERROR, "Binary Addons: failed to get list");
```
ksooo: could you adjust the wording to follow the log "syntax" you're using everywhere else (\_\_FUNCTION\_\_ and the like)?

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,58 @@
...
+    return false;
+  }
+
+  for (auto addon : binaryAddonList)
+  {
+    BinaryAddonBase base = std::make_shared<CBinaryAddonBase>(addon.first);
```
ksooo: just a question: why don't you fill the vector with the addon instances instaed of strings in<code>CAddonMgr::GetInstalledBinaryAddonPaths</code> as you're here instanciate them unconditionally? (member function should be renamed to GetInstalledBinaryAddons, then).

Something like comes to mind:
<code>
typedef std::vector<std::pair<std::shared_ptr<CBinaryAddonBase>, bool>> BinaryAddonList;
</code>
<code>
bool GetInstalledBinaryAddons(BinaryAddonList& binaryAddonList);
</code>

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,58 @@
...
+    if (!base->IsUsable())
+      continue;
+    m_installedAddons[base->ID()] = base;
+    if (addon.second)
+      m_enabledAddons[base->ID()] = base;
```
ksooo: what's the advantage of having two vectors here, one for all usable installed addons, another for the usable installed enabled addons? The addons contained in the second vector are 100% subset of the addons in the first vector. would it be more efficient to have just one vector containing the addon along with its 'enabled' flag? => header change: <code>std::vector<std::pair<std::shared_ptr<CBinaryAddonBase>, bool>> m_usableAddons;</code>?

AlwinEsch:
The request is now complete updated and use a working way.

ksooo (changes requested):
some questions and suggestions...

xbmc/ServiceManager.cpp
```diff
@@ -78,6 +79,13 @@ bool CServiceManager::Init2()
     return false;
   }
 
+  m_binaryAddonManager.reset(new ADDON::CBinaryAddonManager());
+  if (!m_binaryAddonManager->Init())
+  {
+    CLog::Log(LOGFATAL, "CServiceManager::Init: Unable to start CBinaryAddonManager");
```
AlwinEsch: Always this c&p ;)

xbmc/addons/AddonManager.h
```diff
@@ -117,6 +117,8 @@ namespace ADDON
 
     bool GetInstallableAddons(VECADDONS& addons, const TYPE &type);
 
+    bool GetInstalledBinaryAddonPaths(std::vector<std::tuple<std::string, bool, std::string>>& binaryAddonList);
```
ksooo: + please add documentation, esp. the content of the vector is far from being self-explaining. ;-)
+ you should introduce a typedef for the complex type used here:
<code>typedef std::vector<std::tuple<std::string, bool, std::string>> BinaryAddonList;</code>

xbmc/addons/binary-addons/BinaryAddonManager.cpp
AlwinEsch: Was from me separated to two places to have more efficent, to get wanted type.
If enabled becomes requested only the list from there must be checked, otherwise the whole complete with all, was come on my old way, where every addon was stored like them and mostly only few are enabled.

xbmc/addons/binary-addons/AddonDll.cpp
```diff
@@ -47,18 +48,31 @@
...
 {
-  m_initialized = false;
-  m_pDll        = NULL;
-  m_pHelpers    = NULL;
-  m_needsavedsettings = false;
-  m_parentLib.clear();
-  m_interface = {0};
 }
 
+
```
ksooo: no need for a second blank line.

xbmc/addons/binary-addons/AddonDll.cpp
```diff
@@ -385,6 +399,14 @@ void CAddonDll::DestroyInstance(const std::string& instanceID)
     Destroy();
 }
 
+AddonPtr CAddonDll::GetRunningInstance() const
+{
+  AddonPtr addon = CServiceBroker::GetBinaryAddonManager().GetRunningAddon(ID());
+  if (addon)
+    return addon;
+  return AddonPtr();
```
ksooo: the whole implementation of this member function can be reduced to one line: <code>return CServiceBroker::GetBinaryAddonManager().GetRunningAddon(ID());</code> ;-)

xbmc/addons/binary-addons/AddonInstanceHandler.cpp
```diff
@@ -19,35 +19,70 @@
...
-    m_addon(addonInfo)
+    m_addonBase(addonBase)
 {
   // if no special instance ID is given generate one from class pointer (is
   // faster as unique id and also safe enough for them).
   m_instanceId = !instanceID.empty() ? instanceID : StringUtils::Format("%p", static_cast<void*>(this));
+
+  m_addon = m_addonBase->GetAddon(this);
+  if (!m_addon)
+    CLog::Log(LOGFATAL, "ADDON::IAddonInstanceHandler: Tried to get add-on '%s' who not available!", m_addonBase->ID().c_str());
```
ksooo: use \_\_FUNCTION\_\_ instead of hardcoded function name

xbmc/addons/binary-addons/AddonInstanceHandler.cpp
```diff
@@ -59,8 +94,7 @@ bool IAddonInstanceHandler::CreateInstance(KODI_HANDLE instance)
 
 void IAddonInstanceHandler::DestroyInstance()
 {
-  if (m_addon)
-    m_addon->DestroyInstance(m_instanceId);
+  m_addon->DestroyInstance(m_instanceId);
```
ksooo: is it guaranteed that m_addon is always set here? In all (many?) other functions you take care to check before access. why not here?

xbmc/addons/binary-addons/BinaryAddonBase.cpp
```diff
@@ -0,0 +1,383 @@
...
+  m_path = CSpecialProtocol::TranslatePath(addonPath);
+
+  StringUtils::TrimRight(m_path, "/\\");
+
+  auto addonXmlPath = URIUtils::AddFileToFolder(m_path, "addon.xml");
+
+  CXBMCTinyXML xmlDoc;
+  if (!xmlDoc.LoadFile(addonXmlPath))
+  {
+    CLog::Log(LOGERROR, "CBinaryAddonBase: Unable to load '%s', Line %d\n%s",
```
ksooo: la la la, singing... \_\_FUNCTION\_\_ ...

xbmc/addons/binary-addons/BinaryAddonBase.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+  return false;
+}
+
+bool CBinaryAddonBase::ProvidesSeveralSubContents() const
+{
+  int contents = 0;
+  for (auto addonType : m_types)
+    contents += addonType.ProvidedSubContents();
+  return contents > 0 ? true : false;
```
ksooo: could be just <code>return contents > 0;</code>

xbmc/addons/binary-addons/BinaryAddonBase.cpp
```diff
@@ -0,0 +1,383 @@
...
+bool CBinaryAddonBase::MeetsVersion(const AddonVersion &version) const
+{
+  return m_minversion <= version && version <= m_version;
+}
+
+AddonDllPtr CBinaryAddonBase::GetAddon(const IAddonInstanceHandler* handler)
+{
+  if (handler == nullptr)
+  {
+    CLog::Log(LOGERROR, "ADDONS: GetAddon for Id '%s' called with empty instance handler", ID().c_str());
```
ksooo: la la la, singing... \_\_FUNCTION\_\_ ... here and everywhere else in this PR.

xbmc/addons/binary-addons/BinaryAddonBase.cpp
```diff
@@ -0,0 +1,383 @@
...
+{
+  if (handler == nullptr)
+  {
+    CLog::Log(LOGERROR, "ADDONS: ReleaseAddon for Id '%s' called with empty instance handler", ID().c_str());
+    return;
+  }
+
+  CSingleLock lock(m_critSection);
+
+  const auto& presentHandler = m_activeAddonHandlers.find(handler);
```
ksooo: no need for const and & around auto

xbmc/addons/binary-addons/BinaryAddonBase.cpp
```diff
@@ -0,0 +1,383 @@
...
+    return;
+
+  m_activeAddonHandlers.erase(presentHandler);
+
+  // if no handler is present anymore reset and delete the add-on class on informations
+  if (m_activeAddonHandlers.empty())
+  {
+    m_activeAddon.reset();
+  }
+
```
ksooo: this empty line could be removed.

xbmc/addons/binary-addons/BinaryAddonBase.cpp
```diff
@@ -0,0 +1,383 @@
...
+  /*
+   * If nothing is defined in addon.xml set this as unknown to have minimum one
+   * instance type present.
+   */
+  if (m_types.empty())
+  {
+    CBinaryAddonType addonType(ADDON_UNKNOWN, this, nullptr);
+    m_types.push_back(addonType);
```
ksooo: no need for the variable here. could just read <code>m_types.emplace_back(CBinaryAddonType(ADDON_UNKNOWN, this, nullptr));</code>

xbmc/addons/binary-addons/BinaryAddonExtensions.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+  return nullptr;
+}
+
+const EXT_ELEMENTS CBinaryAddonExtensions::GetElements(const std::string& id) const
+{
+  if (id.empty())
+    return m_children;
+
+  EXT_ELEMENTS childs;
```
ksooo: the plural of "child" is "children", not "childs".

xbmc/addons/binary-addons/BinaryAddonExtensions.h
```diff
@@ -0,0 +1,69 @@
...
+
+class TiXmlElement;
+
+namespace ADDON
+{
+
+  struct SExtValue
+  {
+    SExtValue(const std::string& strValue) : str(strValue) { }
+    const std::string& asString()  const { return str; }
```
ksooo: one of the two blanks between "ring()" and "const {" could be removed here.

xbmc/addons/binary-addons/BinaryAddonManager.cpp
AlwinEsch: A from CAddonManager created CAddon is not complete usable, it can then only read the Path from him (and maybe things like names ...) and then destroy it again.

But this becomes also done from the xml parser on CBinaryAddonBase.

xbmc/addons/binary-addons/BinaryAddonExtensions.h
```diff
@@ -0,0 +1,69 @@
...
+    int asInteger() const { return atoi(str.c_str()); }
+    float asFloat() const { return static_cast<float>(atof(str.c_str())); }
+    bool empty() const { return str.empty(); }
+    const std::string str;
+  };
+
+  class CBinaryAddonExtensions;
+  typedef std::vector<std::pair<std::string, CBinaryAddonExtensions> > EXT_ELEMENTS;
+  typedef std::vector<std::pair<std::string, SExtValue> > EXT_VALUE;
+  typedef std::vector<std::pair<std::string, EXT_VALUE> > EXT_VALUES;
```
ksooo: with c++11, no need for a blank between two <code>></code> anymore.

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,240 @@
...
+CBinaryAddonManager::~CBinaryAddonManager()
+{
+  CAddonMgr::GetInstance().Events().Unsubscribe(this);
+}
+
+bool CBinaryAddonManager::Init()
+{
+  CAddonMgr::GetInstance().Events().Subscribe(this, &CBinaryAddonManager::OnEvent);
+
+  std::vector<std::tuple<std::string, bool, std::string>> binaryAddonList;
```
ksooo: "auto" instead of the spaghetti type name.

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,240 @@
...
+  if (!CAddonMgr::GetInstance().GetInstalledBinaryAddonPaths(binaryAddonList))
+  {
+    CLog::Log(LOGNOTICE, "Binary Addons: No binary addons present and related manager, init not necessary");
+    return true;
+  }
+
+  for (auto addon : binaryAddonList)
+  {
+    BinaryAddonBasePtr base = std::make_shared<CBinaryAddonBase>();
```
ksooo: just a question: why don't you fill the vector with the addon instances instead of strings in <code>CAddonMgr::GetInstalledBinaryAddonPaths</code> as you're instanciate them unconditionally here? (member function should be renamed to GetInstalledBinaryAddons, then).

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,240 @@
...
+    if (!base->Create(std::get<0>(addon)))
+    {
+      CLog::Log(LOGERROR, "Binary Addons: Failed to create base for '%s' and addon not usable", std::get<0>(addon).c_str());
+      continue;
+    }
+    m_installedAddons[base->ID()] = base;
+    if (std::get<1>(addon))
+      m_enabledAddons[base->ID()] = base;
```
ksooo: what's the advantage of having two vectors here, one for all usable installed addons, another for the usable installed enabled addons? The addons contained in the second vector are 100% subset of the addons in the first vector. would it be more efficient to have just one vector containing the addon along with its 'enabled' flag? => header change: <code>BinaryAddonList m_usableAddons;</code>?

Or is there a real performance gain, because the functions using the two vectors are called really very often?

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,240 @@
...
+{
+  CSingleLock lock(m_critSection);
+
+  BinaryAddonMgrBaseList* addons;
+  if (enabledOnly)
+    addons = &m_enabledAddons;
+  else
+    addons = &m_installedAddons;
+
+  for (auto& info : *addons)
```
ksooo: no need for "&" after "auto"

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,240 @@
...
+      addonInfos.push_back(info.second);
+    }
+  }
+}
+
+const BinaryAddonBasePtr CBinaryAddonManager::GetInstalledAddonInfo(const std::string& addonId, const TYPE &type/* = ADDON_UNKNOWN*/)
+{
+  CSingleLock lock(m_critSection);
+
+  const auto& addon = m_installedAddons.find(addonId);
```
ksooo: no need for "const" and "&" before/after "auto" here and everywhere else in this PR

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,240 @@
...
+   *
+   * Currently only Screensaver and Visualization use the new way and not need informed.
+   */
+}
+
+void CBinaryAddonManager::InstalledChangeEvent()
+{
+  CSingleLock lock(m_critSection);
+
+  std::vector<std::tuple<std::string, bool, std::string>> binaryAddonList;
```
ksooo: auto instead of full type name?

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+      BinaryAddonBasePtr base = std::make_shared<CBinaryAddonBase>();
+      if (!base->Create(std::get<0>(addon)))
+      {
+        CLog::Log(LOGERROR, "Binary Addons: Failed to create base for '%s' and addon not usable", std::get<0>(addon).c_str());
+        continue;
+      }
+      m_installedAddons[base->ID()] = base;
+      if (std::get<1>(addon))
+        m_enabledAddons[base->ID()] = base;
```
ksooo: at the first glance, the above code lines look pretty much the same as found in <code>bool CBinaryAddonManager::Init()</code>. maybe you should factor out the identical code?

xbmc/addons/binary-addons/BinaryAddonManager.h
```diff
@@ -0,0 +1,113 @@
...
+    bool Init();
+
+    /*!
+     * @brief Checks system about given type to know related add-on's are
+     * installed.
+     *
+     * @param[in] type Add-on type to check installed
+     * @return true if given type is installed
+     */
+    bool HasInstalledAddons(const TYPE &type);
```
ksooo: this member function could be const.

xbmc/addons/binary-addons/BinaryAddonManager.h
```diff
@@ -0,0 +1,113 @@
...
+
+    /*!
+     * @brief Checks system about given type to know related add-on's are
+     * installed and also minimum one enabled.
+     *
+     * @param[in] type Add-on type to check enabled
+     * @return true if given type is enabled
+     */
+    bool HasEnabledAddons(const TYPE &type);
```
ksooo: this member function could be const.

xbmc/addons/binary-addons/BinaryAddonManager.h
```diff
@@ -0,0 +1,113 @@
...
+     * @param[out] addonInfos list where finded addon information becomes stored
+     * @param[in] enabledOnly If true are only enabled ones given back,
+     *                        if false all on system available. Default is true.
+     * @param[in] type        The requested type, with "ADDON_UNKNOWN"
+     *                        are all add-on types given back who match the case
+     *                        with value before.
+     *                        If a type id becomes added are only add-ons
+     *                        returned who match them. Default is for all types.
+     */
+    void GetAddonInfos(BinaryAddonBaseList& addonInfos, bool enabledOnly, const TYPE &type);
```
ksooo: this member function could be const.

xbmc/addons/binary-addons/BinaryAddonManager.h
```diff
@@ -0,0 +1,113 @@
...
+
+    /*!
+     * @brief To get information from a installed add-on
+     *
+     * @param[in] addonId the add-on id to get the info for
+     * @param[in] type if used becomes used type confirmed and is supported, if
+     *                 not a nullptr is returned
+     * @return add-on information pointer of installed add-on
+     */
+    const BinaryAddonBasePtr GetInstalledAddonInfo(const std::string& addonId, const TYPE &type = ADDON_UNKNOWN);
```
ksooo: this member function could be const.

ksooo (changes requested):
Here we go. Lots of stuff...

xbmc/addons/binary-addons/AddonDll.cpp
AlwinEsch: Ah yes, a remnant from my test of them :)

xbmc/addons/binary-addons/BinaryAddonBase.cpp
AlwinEsch: :-D already done

xbmc/addons/binary-addons/BinaryAddonManager.cpp
AlwinEsch: A from CAddonManager created CAddon is not complete usable, it can then only read the Path from him (and maybe things like names ...) and then must it be destroyed (The CAddonDll need them also as parent).

But this becomes also done from the xml parser on CBinaryAddonBase.

xbmc/addons/binary-addons/BinaryAddonManager.cpp
AlwinEsch: Was from me separated to two places to have more efficent, to get wanted type.
If enabled becomes requested only the list from there must be checked, otherwise the whole complete with all, was come on my old way, where every addon was stored like them and mostly only few are enabled.

ksooo: okay, if it was a well profen decision for performance reasons, I'm fine with this.

xbmc/addons/binary-addons/BinaryAddonManager.cpp
ksooo: okay, you are the boss, was just a question anyway

xbmc/addons/binary-addons/BinaryAddonManager.cpp
AlwinEsch: Has testet now, there does not work a auto, I replace it with a typedef
```
  typedef std::tuple<std::string, bool, std::string> BINARY_ADDON_LIST_ENTRY;
  typedef std::vector<BINARY_ADDON_LIST_ENTRY> BINARY_ADDON_LIST;
```

xbmc/addons/binary-addons/BinaryAddonManager.h
AlwinEsch: Has tried with several ways, has produced always errors, with the try to write on addonInfos.

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,237 @@
...
+   * Currently only Screensaver and Visualization use the new way and not need informed.
+   */
+}
+
+void CBinaryAddonManager::InstalledChangeEvent()
+{
+  CSingleLock lock(m_critSection);
+
+  BINARY_ADDON_LIST binaryAddonList;
+  CAddonMgr::GetInstance().GetInstalledBinaryAddonPaths(binaryAddonList);
```
tamland: @AlwinEsch Can you (as we discussed) change `GetInstalledBinaryAddonPaths` to in addition to the paths return `CAddon`s directly, then pass them on to`CBinaryAddonBase`? This way, parsing of all the common info doesn't have to be duplicated.

ksooo: This is what I suggested as well.

AlwinEsch: I change it

AlwinEsch:
Is complete updated, instead of the CAddon use it CAddonInfo who was created by CAddonBuilder.

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,231 @@
...
+  CLog::Log(LOGERROR, "CBinaryAddonManager::%s: Requested addon '%s' unknown as binary", __FUNCTION__, addonId.c_str());
+  return nullptr;
+}
+
+AddonPtr CBinaryAddonManager::GetRunningAddon(const std::string& addonId) const
+{
+  CSingleLock lock(m_critSection);
+
+  auto addon = m_installedAddons.find(addonId);
+  if (addon != m_installedAddons.end() && addon->second)
```
ksooo: Is it really possible that an empty addoninfo is stored in the map? I think the addon->second check can be removed.

xbmc/addons/binary-addons/BinaryAddonType.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+std::string CBinaryAddonType::LibPath() const
+{
+  if (m_libname.empty())
+    return "";
+  return URIUtils::AddFileToFolder(m_path, m_libname);
+}
+
+const char* CBinaryAddonType::GetPlatformLibraryName(const TiXmlElement* element)
```
ksooo: This member function should be static or at least const.

ksooo (changes requested):
Very  nice. Just some nitpicking...

xbmc/ServiceManager.cpp
```diff
@@ -71,13 +72,20 @@ bool CServiceManager::Init2()
...
   m_addonMgr.reset(new ADDON::CAddonMgr());
   if (!m_addonMgr->Init())
   {
     CLog::Log(LOGFATAL, "CServiceManager::Init: Unable to start CAddonMgr");
     return false;
   }
 
+  if (!m_binaryAddonManager->Init())
+  {
+    CLog::Log(LOGFATAL, "CServiceManager::Init: Unable to initialize CBinaryAddonManager");
```
AlwinEsch: I change it also on the other parts on CServiceManager :)

AlwinEsch: Was the last one :D, them on moved CAddonDll becomes updated on other request.

tamland:
> Is complete updated, instead of the CAddon use it CAddonInfo who was created by CAddonBuilder.

Why didn't using CAddon work?

xbmc/addons/AddonManager.cpp
```diff
@@ -564,6 +567,40 @@ bool CAddonMgr::FindInstallableById(const std::string& addonId, AddonPtr& result
   return true;
 }
 
+bool CAddonMgr::GetInstalledBinaryAddons(BINARY_ADDON_LIST& binaryAddonList)
```
AlwinEsch: Here it can not "const", becomes in big conflict on much places then

AlwinEsch:
Make no sense to use, it need only to have the CAddonInfo (who becomes also stored) to have for binary addon creation.

CAddon becomes created with CAddonInfo as value, the binary part need then only to read the CAddonInfo and destroy the CAddon.

With that he gets it directly:
```cpp
binaryAddonList.push_back(BINARY_ADDON_LIST_ENTRY(!IsAddonDisabled(cp_addon->identifier), std::move(builder.GetAddonInfo())));
```

Otherwise must be the
```cpp
AddonPtr addon;
...
builder.Build();` 
binaryAddonList.push_back(BINARY_ADDON_LIST_ENTRY(!IsAddonDisabled(cp_addon->identifier), addon));
...
```
added who then make never needed work.

AlwinEsch:
Another part is the performance, here the measured from me:

The first way where addon.xml was readed complete on CBinaryAddonBase
0.002619 seconds
0.002962 seconds
0.002603 seconds
0.002628 seconds
0.001032 seconds

Here with CAddonBuiler and CAddon
0.010468 seconds
0.010382 seconds
0.010420 seconds
0.010542 seconds
0.010504 seconds

Here the current on request
0.007633 seconds
0.007950 seconds
0.008571 seconds
0.007605 seconds
0.008558 seconds

Measured on a 8x "Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz" CPU with 38 binary addons.

On the first can be seen how big the work is where cpluff is only used to get path.

These are small times, but times on a larger addon quantity and a weaker system can give it greater differences.

xbmc/addons/AddonManager.cpp
ksooo: I'm sure it can. Try harder. Mostly it does not work, because functions caked from this method are not convert, but they could and should be. See for example my other comment re GetPlatformLibraryName which currently is not const, but could even be static.

ksooo: but yeah if it is too much effort, leave it as it is for now. can be optimized later.

AlwinEsch:
jenkins build this please

tamland:
> Make no sense to use, it need only to have the CAddonInfo (who becomes also stored) for binary addon creation.
>
> CAddon becomes created with CAddonInfo as value, the binary part need then only to read the CAddonInfo and destroy the CAddon.

Everything else that interface with addon manager use IAddon as the 'info' so it makes sense to do the same for binary addons. Even if you don't to directly *need* it. When you don't set an extension point there is 0 overhead of `Build()` (it just calls the CAddon constructor). If it's impossible to do now, at the very least be prepared to fix it later

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+  BINARY_ADDON_LIST binaryAddonList;
+  if (!CAddonMgr::GetInstance().GetInstalledBinaryAddons(binaryAddonList))
+  {
+    CLog::Log(LOGNOTICE, "CBinaryAddonManager::%s: No binary addons present and related manager, init not necessary", __FUNCTION__);
+    return true;
+  }
+
+  for (auto addon : binaryAddonList)
+    AddAddonBaseEntry(addon);
```
tamland: Missing lock

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -0,0 +1,231 @@
...
+  if (addon != m_installedAddons.end())
+    return addon->second->GetActiveAddon();
+
+  return nullptr;
+}
+
+bool CBinaryAddonManager::AddAddonBaseEntry(BINARY_ADDON_LIST_ENTRY& entry)
+{
+  BinaryAddonBasePtr base = std::make_shared<CBinaryAddonBase>(entry.second);
+  if (!base->Create())
```
tamland: Holding the internal lock while doing io and external calls like this is a recipe for deadlocks. I don't have an easy fix for this, but it's better to write initialization in such a way that it doesn't require calls to be made in-between reading and writing to m_installedAddons/m_enabledAddons

xbmc/addons/binary-addons/BinaryAddonManager.cpp
AlwinEsch: Becomes added :-) , had it omitted since this is called only at the Kodi start

AlwinEsch:
jenkins build this please

PR #11954, opened 2017-04-11
[bluray][win32] Update libbluray

ace20022:

PR #12310, opened 2017-06-16
[bluray] jvm check is available since libbluray 1.0.1

Rechi:
## Description
The jvm check is available since libbluray version 1.0.1 and not windows only.

## Motivation and Context
#11954 added the check for windows only.

## How Has This Been Tested?
Tested on Windows (with and without libbluray), OS X and Linux (with 0.9.3 and 1.0.1)

## Types of change

## Checklist:

### DIFF
--- a/xbmc/settings/DiscSettings.cpp
+++ b/xbmc/settings/DiscSettings.cpp
@@ -31,6 +31,7 @@
 #include "dialogs/GUIDialogOK.h"
 #include "utils/Variant.h"
 
+#include <libbluray/bluray-version.h>
 
 CDiscSettings::CDiscSettings(void)
 {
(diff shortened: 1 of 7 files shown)


akva2:
system.h cleanup?

Rechi:
@ace20022 any objections?

xbmc/settings/DiscSettings.cpp
```diff
@@ -58,12 +59,12 @@ CDiscSettings& CDiscSettings::GetInstance()
 
 void CDiscSettings::OnSettingChanged(std::shared_ptr<const CSetting> setting)
 {
-#ifdef TARGET_WINDOWS
   if (setting == NULL)
     return;
 
   const std::string &settingId = setting->GetId();
 
+#if (BLURAY_VERSION >= BLURAY_VERSION_CODE(1,0,1))
```
ace20022: please move that to the beginning of the method.

Rechi: @ace20022 I can do, but shouldn't the ifdef only enclose as little as actually necessary?
If one adds another setting to DiscSettings later the setting null check would also apply to the new one.

ace20022: If you really want to discuss such things, I don't care.

Rechi:
@ace20022 moved the ifdef, your button

PR #12311, opened 2017-06-16
FIX: ensure real filename from videodb in playlists

koying:
fixes issue when downloading subtitles side-by-side

### DIFF
--- a/xbmc/PlayListPlayer.cpp
+++ b/xbmc/PlayListPlayer.cpp
@@ -25,7 +25,6 @@
 #include "input/Key.h"
 #include "URL.h"
 #include "messaging/ApplicationMessenger.h"
-#include "filesystem/VideoDatabaseFile.h"
 #include "messaging/helpers/DialogOKHelper.h"
 #include "ServiceBroker.h"
 
@@ -296,9 +295,6 @@ bool CPlayListPlayer::Play(int iSong, std::string player, bool bAutoPlay /* = fa
 
   m_iCurrentSong = iSong;
   CFileItemPtr item = playlist[m_iCurrentSong];
-  if (item->IsVideoDb() && !item->HasVideoInfoTag())
-    *(item->GetVideoInfoTag()) = XFILE::CVideoDatabaseFile::GetVideoTag(CURL(item->GetPath()));
-
   playlist.SetPlayed(true);
 
   m_bPlaybackStarted = false;
--- a/xbmc/playlists/PlayList.cpp
+++ b/xbmc/playlists/PlayList.cpp
@@ -83,6 +83,9 @@ void CPlayList::Add(const CFileItemPtr &item, int iPosition, int iOrder)
   else
     item->m_iprogramCount = iOrder;
 
+  if (item->IsVideoDb())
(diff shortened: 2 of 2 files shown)


MilhouseVH:
Thanks @koying, this has fixed the subtitle download issue: https://forum.kodi.tv/showthread.php?tid=298462&pid=2606809#pid2606809

Unfortunately it doesn't appear to have had any effect on the watched statuses: https://forum.kodi.tv/showthread.php?tid=298461&pid=2605996#pid2605996

koying:
Ok, thanks for the feedback

mlburgett:
@koying, is there any info I can gather or tests I can run, to help with the 'watched statuses' issue @MilhouseVH refers to above?

koying:
@MilhouseVH Would you mind trying the current version, please

MilhouseVH:
@mlburgett can you try build [#0629b](http://milhouse.libreelec.tv/builds/testing/RPi2/LibreELEC-RPi2.arm-9.0-Milhouse-20170630143747-%230629b-g1196fc5.tar) which includes the latest commit from @koying.

Note this PR continues to be be included in regular test builds, so this updated PR with the new commit will be in tonight's `#0630` build.

mlburgett:
@MilhouseVH, I don't see any change using #0629b, all my tv->recordings continue to show 'watched' status.  Partially watched recordings do still show correct status (as they have since this first showed up)

da-anda:
I suppose the new "DynPath" thingy could/should be use here instead of altering the actual "path" that basically acts like the file ID? @FernetMenta ?

FernetMenta:
@da-anda nope, DynPath only get evaluated by inputstream at a time when the file gets opened for playback. 
but yes, id (path) should never be changed on a fileitem. better construct a new item insteead of chaning an existing one.

da-anda:
@FernetMenta but if you create a new file item you will loose the identity
of the item. Shouldn't it be a goal to always keep the identity regardless
where the file is processed?

Am 20.08.2017 11:54 vorm. schrieb "Rainer Hochecker" <
notifications@github.com>:

FernetMenta:
@da-anda what do you call the "identity of the item"?

da-anda:
@FernetMenta well, the VFS path is basically the identiy of the fileItem, isn't it? As soon as you change the "path" property to something else you a) change the file type (your own words) and b) some things that rely on the VFS path won't work anymore (like some context menu stuff IIRC). Therefor IMO we neither should pass around stub fileItems (like we f.e. do for VideoInfoDialog which makes it impossible to implement a context menu there atm) nor alter the path.

FernetMenta:
@da-anda right. IMO we should make those methods private like the comment in the code suggests:
https://github.com/xbmc/xbmc/blob/master/xbmc/FileItem.cpp#L1698
 Making use of SetPath is a step in the wrong direction an makes a fix harder.

pogo1975:
i dont know if it was reported but whitout this patch if you disable play next video automaticly subtitles works well and downloading subtitles works, with this patch applied if you enable play next video automaticly subtitles works but if next video is played subs remains from the first one...

MilhouseVH:
@koying: I don't know if this is ever going to be merged, but would you mind rebasing?

PR #12318, opened 2017-06-18
[imagedecoder] change image decoder addon system to new way

AlwinEsch:
## Description

This change the addon system headers from "C" to "C++", further is this the
first who use the new way to get extensions from addon.xml. In this the
mimetype and supported file extensions.

This extension way allow to check several parts on different types without
load of addon.

## Types of change

## Checklist:


xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/ImageDecoder.h
```diff
@@ -0,0 +1,163 @@
...
+    KODI_HANDLE kodi_instance;
+  } AddonToKodiFuncTable_ImageDecoder;
+
+  struct AddonInstance_ImageDecoder;
+  typedef struct KodiToAddonFuncTable_ImageDecoder
+  {
+    kodi::addon::CInstanceImageDecoder* addonInstance;
+    bool (__cdecl* load_image_from_memory) (AddonInstance_ImageDecoder const* instance,
+                                         unsigned char* buffer, unsigned int buf_size,
+                                         unsigned int* width, unsigned int* height);
```
ksooo: indentation?

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/ImageDecoder.h
```diff
@@ -0,0 +1,163 @@
...
+      if (instance == nullptr)
+        throw std::logic_error("kodi::addon::CInstanceImageDecoder: Creation with empty addon structure not allowed, table must be given from Kodi!");
+
+      m_instanceData = static_cast<AddonInstance_ImageDecoder*>(instance);
+      m_instanceData->toAddon.addonInstance = this;
+      m_instanceData->toAddon.load_image_from_memory = ADDON_LoadImageFromMemory;
+      m_instanceData->toAddon.decode = ADDON_Decode;
+    }
+
+    inline static bool ADDON_LoadImageFromMemory(AddonInstance_ImageDecoder const* instance,
```
ksooo: "type const *" is the same as "const type *". Why do you use other syntax here than you do usually? Is your intention to denote that the pointer value itself is const and the data behind the pointer can be written? Then you must use "type * const". Applies also to other methods in the PR.

xbmc/guilib/imagefactory.cpp
```diff
@@ -46,16 +46,15 @@ IImage* ImageFactory::CreateLoader(const CURL& url)
 
 IImage* ImageFactory::CreateLoaderFromMimeType(const std::string& strMimeType)
 {
-  VECADDONS codecs;
-  ADDON::CBinaryAddonCache &addonCache = CServiceBroker::GetBinaryAddonCache();
-  addonCache.GetAddons(codecs, ADDON::ADDON_IMAGEDECODER);
-  for (auto& codec : codecs)
```
ksooo: Does this mean that we will no longer need binary addon cache once all your changes are in? Is <code>CServiceBroker::GetBinaryAddonManager().GetAddonInfos()</code> performance-wise comparable with  <code>CServiceBroker::GetBinaryAddonCache().GetAddons()</code>?

ksooo (commented):
some notes/questions.

xbmc/guilib/imagefactory.cpp
AlwinEsch: Yes, this replace complete the binary addon cache, mostly is only the info like here for supported types needed.

This parts are continuous stored in binary addon manager.

In case now a running addon need to be asked becomes them during active time also stored and can be used by new way.

From performance it is equal with the other.

AlwinEsch: Have measured performance (with one addon present).

|  |    |
|--|---|
| With binary addon cache | ~0.000040 sec
| Complete without cache | ~0.005000 sec
| With new on this | ~0.000065 sec

ksooo: Okay, thanks for clarifying.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/ImageDecoder.h
```diff
@@ -0,0 +1,163 @@
...
+    KODI_HANDLE kodi_instance;
+  } AddonToKodiFuncTable_ImageDecoder;
+
+  struct AddonInstance_ImageDecoder;
+  typedef struct KodiToAddonFuncTable_ImageDecoder
+  {
+    kodi::addon::CInstanceImageDecoder* addonInstance;
+    bool (__cdecl* load_image_from_memory) (const AddonInstance_ImageDecoder* instance,
+                                         unsigned char* buffer, unsigned int buf_size,
+                                         unsigned int* width, unsigned int* height);
```
ksooo: indentation?

ksooo (approved):
Only a small indentation glitch, otherwise good to go.

AlwinEsch:
jenkins build this please

AlwinEsch:
@notspiff OK from your side to bring in?

notspiff:
sure

PR #12320, opened 2017-06-19
changed: Default libcurl timeout value to 30

arnova:
Ages ago when libcurl still had several issues with timeout handling and we also didn't have the ability to abort the opening of a (new) connection, I lowered this value to 10 to make it more user friendly. Fortunately much has improved ever since, and now the value of 10 is actually too low for some scenarios. Mainly for setups with poor (wifi) connections or NAS systems which have HDD spindown enabled, the value should be much higher. According to docs one can assume a harddisk will at max take 30 seconds to spinup, so let's use this as our timeout default as well.

### DIFF
--- a/xbmc/settings/AdvancedSettings.cpp
+++ b/xbmc/settings/AdvancedSettings.cpp
@@ -320,7 +320,7 @@ void CAdvancedSettings::Initialize()
   m_screenAlign_xStretchFactor = 1.0;
   m_screenAlign_yStretchFactor = 1.0;
 
-  m_curlconnecttimeout = 10;
+  m_curlconnecttimeout = 30;
   m_curllowspeedtime = 20;
   m_curlretries = 2;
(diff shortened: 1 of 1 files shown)


arnova:
Also increased the timeout for Samba as obviously the same thing applies for it.

Rechi:
Just to be sure curl 7.21.6 doesn't have those timeout handling issues?
We have to use that old version on [OSX](https://github.com/xbmc/xbmc/blob/320fda1/tools/depends/target/curl/Makefile#L7), because otherwise Kodi would crash.

arnova:
@rechi: The timout issues in libcurl probably existed back in the Xbox age, or at least when it was still XBMC. We've also worked around many problems in our CCurlFile and CFileCache classes, so no worries, 7.21.6 should be fine.

MartijnKaijser:
Let's try it in nightlies before adding to v17

PR #12321, opened 2017-06-19
changed: Default libcurl timeout value to 30

arnova:
Backport of https://github.com/xbmc/xbmc/pull/12320


arnova:
Also increased the timeout for Samba as obviously the same thing applies for it.

PR #12325, opened 2017-06-21
[inputstream] change addon interface to new C++ style

AlwinEsch:
## Description
This commit change the inputstream addon system to the new C++ style
and remove the old parts.

The parts from here are taken from kodi-agile and becomes mainly improved
and optimized from @peak3d on next changes.

Currently it supports only the same as the old, only a new interface way.

## Types of change

## Checklist:


xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/Inputstream.h
```diff
@@ -0,0 +1,585 @@
...
+      SUPPORTSIPOSTIME = (1 << 1),
+
+      /// supports interface IDisplayTime
+      SUPPORTSIDISPLAYTIME = (1 << 2),
+
+      /// supports seek
+      SUPPORTSSEEK = (1 << 3),
+
+      /// supports pause
+      SUPPORTSPAUSE = (1 << 4)
```
ksooo: Maybe it's just personal style, but I would insert an underline between the verb and the action (e.g. SUPPORTS_PAUSE). Increases readability and this is how we usuallly name enum members, iirc.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/Inputstream.h
```diff
@@ -0,0 +1,585 @@
...
+    DemuxPacket* (*allocate_demux_packet)(void* kodiInstance, int data_size);
+    void (*free_demux_packet)(void* kodiInstance, DemuxPacket* packet);
+  } AddonToKodiFuncTable_InputStream;
+
+  struct AddonInstance_InputStream;
+  typedef struct KodiToAddonFuncTable_InputStream /* internal */
+  {
+    kodi::addon::CInstanceInputStream* addonInstance;
+
+    bool (__cdecl* open)(AddonInstance_InputStream* instance, INPUTSTREAM&);
```
ksooo: For other addon types, like ADSP, you are using "const pointer to the addon instance", which is imo correct, and you name the parameter <code>addonInstance</code> - would read <code>AddonInstance_InputStream const* addonInstance</code> here. I suggest to be consistent across addon types re naming and data types.

I also suggest to check older commits and to correct this. IIRC, the "const" came in with the ADSP PR, before we did not spot this and always used non-const pointers.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/Inputstream.h
```diff
@@ -0,0 +1,585 @@
...
+    bool (__cdecl* demux_seek_time)(AddonInstance_InputStream* instance, double, bool, double*);
+    void (__cdecl* demux_set_speed)(AddonInstance_InputStream* instance, int);
+    void (__cdecl* set_video_resolution)(AddonInstance_InputStream* instance, int, int);
+
+    // IDisplayTime
+    int (__cdecl* get_total_time)(AddonInstance_InputStream* instance);
+    int (__cdecl* get_time)(AddonInstance_InputStream* instance);
+
+    // IPosTime
+    bool (__cdecl* pos_time)(AddonInstance_InputStream* instance, int);
```
ksooo: May I suggest to find a more speaking name for this function? "pos_time" means nothing to me.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/Inputstream.h
```diff
@@ -0,0 +1,585 @@
...
+    // Seekable (mandatory)
+    bool (__cdecl* can_pause_stream)(AddonInstance_InputStream* instance);
+    bool (__cdecl* can_seek_stream)(AddonInstance_InputStream* instance);
+
+    int (__cdecl* read_stream)(AddonInstance_InputStream* instance, uint8_t*, unsigned int);
+    int64_t(__cdecl* seek_stream)(AddonInstance_InputStream* instance, int64_t, int);
+    int64_t (__cdecl* position_stream)(AddonInstance_InputStream* instance);
+    int64_t (__cdecl* length_stream)(AddonInstance_InputStream* instance);
+    void (__cdecl* pause_stream)(AddonInstance_InputStream* instance, double);
+    bool (__cdecl* is_real_time_stream)(AddonInstance_InputStream* instance);
```
ksooo: I think you should either supply variable names for all parameters or for none. Currently, you supply a name for AddonInstance_InputStream parameter, but none for all other paramters...

Maybe we should also change this for addon types that came in with older PRs, where we missind this inconsistency during review.

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
```diff
@@ -20,40 +20,143 @@
...
+CInputStreamAddon::CInputStreamAddon(BinaryAddonBasePtr& addonBase, IVideoPlayer* player, const CFileItem& fileitem)
+  : IAddonInstanceHandler(ADDON_INSTANCE_INPUTSTREAM, addonBase),
+    CDVDInputStream(DVDSTREAM_TYPE_ADDON, fileitem),
+    m_player(player)
 {
+  std::string listitemprops = addonBase->Type(ADDON_INPUTSTREAM)->GetValue("@listitemprops").asString();
+  std::string name(addonBase->ID());
+
+  m_fileItemProps = StringUtils::Tokenize(listitemprops, "|");
+  for (auto &key : m_fileItemProps)
```
ksooo: No need for & after auto. Trust the compiler. ;-)

Here and at other places in this source file.

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
```diff
@@ -20,40 +20,143 @@
...
+bool CInputStreamAddon::Supports(BinaryAddonBasePtr& addonBase, const CFileItem &fileitem)
+{
+  // check if a specific inputstream addon is requested
+  CVariant addon = fileitem.GetProperty("inputstreamaddon");
+  if (!addon.isNull())
+  {
+    if (addon.asString() != addonBase->ID())
+      return false;
+    else
+      return true;
```
ksooo: I would write this as <code>return addon.asString() == addonBase->ID();</code>

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
```diff
@@ -20,40 +20,143 @@
...
+
+  bool ret = m_struct.toAddon.open(&m_struct, props);
   if (ret)
   {
-    m_hasDemux = m_addon->HasDemux();
-    m_hasDisplayTime = m_addon->HasDisplayTime();
-    m_hasPosTime = m_addon->HasPosTime();
-    m_canPause = m_addon->CanPause();
-    m_canSeek = m_addon->CanSeek();
+    memset(&m_caps, 0, sizeof(m_caps));
```
ksooo: replace `memset` by `m_struct = {{ 0 }};`?

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
```diff
@@ -237,29 +357,129 @@ bool CInputStreamAddon::SeekTime(double time, bool backward, double* startpts)
...
+    }
+
+    m_streams[demuxStream->uniqueId] = demuxStream;
+  }
+}
+
+void CInputStreamAddon::DisposeStreams()
+{
+  for (auto &stream : m_streams)
+    delete stream.second;
```
ksooo: does the dtor of stream.second take care to delete `ExtraData`, which is allocated here https://github.com/xbmc/xbmc/pull/12325/files#diff-3590db0d5bee056ad4a07281388f0531R455?

ksooo (changes requested):
some questions and suggstions

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
AlwinEsch: No, in this case is needed :D

Has changed them, but there is a write done and was wrote to somewhere else ;)

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
AlwinEsch: In this case not work, there is only one part defined in structure and brought errors.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/Inputstream.h
FernetMenta: It positions the stream to the given time. how much more meanful could it be? :)
this is a demuxer/inputstream interface. addons just implement it and breaking naming in the addon interface does not make sense. you would need to change the entire interace which has a much broader scope than this pr.

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
Montellese: @ksooo you can't trust the compiler because the compiler can't know what you want to do. There are basically three options:
* `for (auto value : values)` which will create a full copy of the object (most often not what you want)
* `for (auto& value : values)` which will create a modifyable reference to the object (similar to `iterator`)
* `for (const auto& value : values` which will create a const reference to the object (similar to `const_iterator`)

This is the same as with regular C++ iterators. You have to tell the compiler whether to use `iterator` or `const_iterator`.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/Inputstream.h
ksooo: > It positions the stream to the given time. how much more meanful could it be? :)

 set_stream_position would be MUCH more meaningful, imo.

 Alwin is actually changing the whole interface, all addons must be adapted anyway. So we could and should take the opportunity to improve even small things like this.

 But if anybody else is fine with pos_time feel free to ignore me, I was just suggesting, not demanding.

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
ksooo: k

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
ksooo: @Montellese thanks for explaining. Makes perfect sense.

xbmc/addons/kodi-addon-dev-kit/include/kodi/addon-instance/Inputstream.h
FernetMenta: not really what it does. you can set position to a position or a time.

AlwinEsch: I leave it, there is only the "C" structure who handle the interface between the both.

There comes the next time a complete reworked addon documentation and there can be the `CInstanceInputStream::PosTime(...)` detailed descripted.

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
AlwinEsch: What mean "k"?

ksooo: it's like "if you say so, although I'm not fully convinced, but I don't really care to investigate"

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
AlwinEsch: Yes is in his desctructor

ksooo: Okay.

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
hudokkow: Means OK.

ksooo: Not exactly, it's more like okay-ish.

AlwinEsch:
Is updated.

ksooo (approved):

AlwinEsch:
jenkins build this please

AlwinEsch:
jenkins build this please

AlwinEsch:
Can someone help me to explain what on Windows now the second time fails? For me not related to this request but I'm not sure.

xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
peak3d: auto & makes sense, if not the loop is working with a copy. Already fall ito this trap

peak3d:
> fatal: repository 'https://github.com/xbmc/FFmpeg/archive/' not found

peak3d:
thanx @AlwinEsch for this PR, we already approoved the changes in agile branch and it seems to be mostly identical. 👍 from my side.

Rechi:
strange build machine error, disabled it
jenkins build this please

AlwinEsch:
jenkins build this please

PR #12326, opened 2017-06-21
Fix 3d mode switching for Wetek HUB on Libreelec

Memphiz:
Backort of #12316

@kszaq in case this makes it in and gets pulled into libreelec 8.2 - this will conflict with this patch:

https://github.com/LibreELEC/LibreELEC.tv/blob/libreelec-8.2/packages/mediacenter/kodi/patches/aarch64/kodi-0003-EGLNativeTypeAmlogic-Enable-GUI-free_scale-when-framebuffer-size-is-less-than-output.patch

Just ensure that the stereo_mode handling i added here is done after the call to "DealWithScale" ...

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
@@ -2066,10 +2066,20 @@ void CAMLCodec::SetVideoSaturation(const int saturation)
   SysfsUtils::SetInt("/sys/class/video/saturation", saturation);
 }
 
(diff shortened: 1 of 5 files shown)


kszaq:
Thank you Memphiz! I have included this patch in my latest community LE build for people to test.

Memphiz:
Thx - feedback would be cool. From time to time my projector doesn't switch back from 3d mode to normal mode. This is also commented and worked around in SPMC by @koying - sadly that workaround leads to loosing audio for me so i didn't include it. Just curious how many users this hits. (I doubt there are that many using 3D at all but we will see ;) ).

MartijnKaijser:
Jenkins build this please

PR #12327, opened 2017-06-21
CHG: Allow to (really) disable upnp altogether

koying:
aka the "@da-anda shitty radio" PR ;)

Currently, there is no way to disable the upnp "listener". This PR implements it.
There are a number of reason why you wouldn't want a daemon like this to run (if you don't need it), but the 1st one would probably be wasted resources.

The side-effect is that it indirectly solves an issue where an ill-behaving upnp device on the network flooding with upnp messages can affect Kodi (ie da-anda shitty radio ;) )

### DIFF
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -1821,16 +1821,29 @@
     <category id="upnp" label="20187" help="36322">
       <requirement>HAS_UPNP</requirement>
       <group id="1" label="16000">
-        <setting id="services.upnpserver" type="boolean" label="21360" help="36323">
+        <setting id="services.upnp" type="boolean" label="21358" help="39017">
           <level>0</level>
           <default>false</default>
+          <control type="toggle" />
+        </setting>
+        <setting id="services.upnpserver" type="boolean" parent="services.upnp" label="21360" help="36323">
+          <level>0</level>
+          <default>false</default>
+          <dependencies>
+            <dependency type="enable" setting="services.upnp">true</dependency>
+          </dependencies>
           <control type="toggle" />
         </setting>
         <setting id="services.upnpannounce" type="boolean" parent="services.upnpserver" label="20188" help="36324">
           <level>2</level>
           <default>true</default>
           <dependencies>
-            <dependency type="enable" setting="services.upnpserver">true</dependency>
+            <dependency type="enable">
+              <and>
+                <condition setting="services.upnp" operator="is">true</condition>
+                <condition setting="services.upnpserver" operator="is">true</condition>
+              </and>
+            </dependency>
           </dependencies>
           <control type="toggle" />
         </setting>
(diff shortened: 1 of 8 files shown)


system/settings/settings.xml
```diff
@@ -1821,38 +1821,64 @@
     <category id="upnp" label="20187" help="36322">
       <requirement>HAS_UPNP</requirement>
       <group id="1" label="16000">
-        <setting id="services.upnpserver" type="boolean" label="21360" help="36323">
+        <setting id="services.upnp" type="boolean" label="1270">
```
da-anda: like mentioned on Slack, it's the wrong label ID (it's the one from AirPlay). Correct ID is `21358`.

system/settings/settings.xml
```diff
@@ -1821,38 +1821,64 @@
...
           <level>0</level>
           <default>false</default>
           <control type="toggle" />
         </setting>
+        <setting id="services.upnpserver" type="boolean" label="21358" help="36323">
```
da-anda: and this setting should probably keep label id `21360`?

da-anda:
thanks. The correct unofficial PR title would be "fix da-anda shitty radio issue that only occurs on Android", because no other platform running Kodi is having this issue (video/UI freeze while audio continues and Kodi is still responding to keypresses). The root cause might be some libplatinum core issue on Android leading to a scenario that prevents our rendering process to update frames, but no idea on how to hunt this one down.

@Montellese if you'd like to see the network traffic/spam from my UPNP enabled radio I'm happy to provide a Wireshark trace (already have prepared one). My radio basically is sending at least 2 (up to 6) UPNP related messages every second and the screen freezes whenever I play something in Kodi on Android (Kodi17 on SHIELD and Kodi 18 nightly on Nexus10).

system/settings/settings.xml
Montellese: And it would be great to also have a description.

system/settings/settings.xml
```diff
@@ -1821,38 +1821,64 @@
...
+          <level>1</level>
```
Montellese: Why did you change the level from 0 to 1?

Montellese (changes requested):
Looks ok apart from the minors. It might be worth to check the initialization logic of Platinum SDK in more detail to see if we are already performing certain setup steps before `StartUPnPClient()` or `StartUPnPServer()` is called.

system/settings/settings.xml
koying: The idea is to enable/disable UPnP altogether. As such, it makes sense to me that anything related to UPnP be under the new setting at level 0.

Montellese: Hm maybe you are misinterpreting the term `level`. In this case it means the setting level visibility i.e.
* 0 = Basic
* 1 = Standard
* 2 = Advanced
* 3 = Expert
* 4 = Internal

If you want to have the setting indented under the general "Enable UPnP" setting you need to add the `parent="services.upnp"` attribute on the `<setting>` tag.

koying: Oops, indeed

MartijnKaijser:
@koying build failed

koying:
jenkins build this please

MartijnKaijser:
@Montellese all good?

Montellese (approved):

PR #12331, opened 2017-06-21
FIXUP: [droid] addon packaging

koying:
Stuff is scattered everywhere :(
The `lib(XBMC|KODI)_*.so` are there

Fixup of https://github.com/xbmc/xbmc/pull/12304

### DIFF
--- a/tools/android/packaging/Makefile.in
+++ b/tools/android/packaging/Makefile.in
@@ -135,6 +135,8 @@ libs: $(PREFIX)/lib/@APP_NAME_LC@/lib@APP_NAME_LC@.so
 	mkdir -p xbmc/lib/$(CPU) xbmc/assets/python2.7/lib/ xbmc/obj/local/$(CPU)
 	cp -fpL $(SRCLIBS) xbmc/obj/local/$(CPU)/
 	cp -fp $(PREFIX)/lib/@APP_NAME_LC@/lib@APP_NAME_LC@.so xbmc/obj/local/$(CPU)/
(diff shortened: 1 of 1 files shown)


wsnipex (approved):
heh, depends path is not really "everywhere" ;)

koying:
Nah, I mean the libs are kinda scattered at too many different places.
Between the main build, the depend target, the binary addons build, the packaging build and whatnot, I fear for my sanity ;)

MrMC:
all roads lead to prefix :)

koying:
@MrMC Yeah, but what does prefix point to?
Nope, not the depends anymore, it points to build/install ;)

koying:
jenkins build and merge

MrMC:
In the long past, the rule for non-desktop (using depends) was build/install to prefix. packaging pull from there. Then too many cooks entered the kitchen and now no one knows where the knives are anymore :)

wsnipex:
everything still installs to prefix on depends builds. Only binary addons have their own, cause they handle their own independent dependencies.

koying:
No bikeshedding, but was clearer before when *everything* installed to the "PREFIX" specified on the configure line, whether main program & addons, depends and binary addons.

Nowadays, main "installs" into the cmake "build" (which ends up being PREFIX), depends still goes to "prefix", (which ends up being "DEPENDS_PATH") and binary addons go... somewhere ;)

But hey, adapt or die, isn't it ;)

MrMC:
or fork :)

PR #12332, opened 2017-06-22
VideoBuffer: Avoid unsafe race when resetting pool

popcornmix:
Currently we are resetting the m_pool variable after Returning the buffer.
As the pool will add the buffer back to the free list there is a danger it
will be reallocated and m_pool assigned before the reset to nullptr occurs.

That leaves you with a live buffer with an invalid pool pointer.
So, make sure we don't touch m_pool after returning the buffer.

## Types of change

## Checklist:


FernetMenta:
thanks!

PR #12333, opened 2017-06-22
[peripheral] change addon interface to new C++ style

AlwinEsch:
## Description
This change also the peripheral addon system to use C++ class as
interface to Kodi.

There is only the interface itself changed, the functions style is still
the same as before with the use of PeripheralUtils.h (kodi_peripheral_utils.hpp).
My experience for the system is not good enough to change also the list functions
to use direct a vector on addon (not sure how I can test every function).

Maybe can be this improved from @garbear who have more experience.

The test of them here was good and has played some games :-) .

## Types of change

## Checklist:


xbmc/peripherals/bus/virtual/PeripheralBusAddon.cpp
```diff
@@ -457,23 +402,26 @@ void CPeripheralBusAddon::UpdateAddons(void)
...
 
-      if (bCreated)
-        m_addons.emplace_back(std::move(newAddon));
-      else
-        m_failedAddons.emplace_back(std::move(newAddon));
+        {
+          CSingleExit exit(m_critSection);
+          bCreated = newAddon->CreateAddon();
+        }
+fprintf(stderr, "-----3- %s - %i\n", __PRETTY_FUNCTION__, bCreated);
```
notspiff: Nah

xbmc/peripherals/bus/virtual/PeripheralBusAddon.cpp
```diff
@@ -514,9 +462,9 @@ void CPeripheralBusAddon::PromptEnableAddons(const ADDON::VECADDONS& disabledAdd
   bool bAccepted = false;
 
   auto itAddon = std::find_if(disabledAddons.begin(), disabledAddons.end(),
-    [](const AddonPtr& addon)
+    [](const BinaryAddonBasePtr& addonInfo)
   {
-    return std::static_pointer_cast<CPeripheralAddon>(addon)->HasButtonMaps();
+    return addonInfo->Type(ADDON::ADDON_PERIPHERALDLL)->GetValue("@provides_buttonmaps").asBoolean();
```
garbear: Any way to not scatter XML tags in this function?

xbmc/peripherals/bus/virtual/PeripheralBusAddon.cpp
AlwinEsch: Uppsss, readed several times the commit and stil this test print inside, SORRY

notspiff: No worries. last week i sent our customers code which spewed 'hore' five times on every run of the simulator :) (add a w in front to translate the norwegian..) proud moment ;)

xbmc/peripherals/bus/virtual/PeripheralBusAddon.cpp
AlwinEsch: Not direct, it can be maybe a static function added on ./xbmc/peripherals/addons/PeripheralAddon.h who do this with pass of addonInfo as value.

This way was thought (on agile) to replace cpluff and allow a read of addon.xml value without creation of related addon class, in this case CPeripheralAddon. Also used here to allow read of values of several different instance types without creation of them.

garbear:
@AlwinEsch https://github.com/AlwinEsch/kodi/pull/30

xbmc/peripherals/bus/virtual/PeripheralBusAddon.cpp
garbear: Makes sense, maybe a static function HasButtonMaps() that takes the xml parameter?

AlwinEsch: I add them

ksooo (commented):
Looking good. Nothing to moan this time.

AlwinEsch:
Is updated and use now `static inline ...` inside CPeripheralAddon to get extension values.

xbmc/addons/binary-addons/BinaryAddonManager.h
```diff
@@ -63,6 +63,24 @@ namespace ADDON
...
+
+    /*!
+     * @brief Check whether an addon has been enabled.
+     *
+     * @param[in] addonId id of the addon
+     * @param[in] type Add-on type to check installed and enabled
+     * @return true if enabled
+     */
+    bool IsAddonEnabled(const std::string& addonId, const TYPE &type = ADDON_UNKNOWN);
+
```
tamland: This is not something that should be added as public methods on BinaryAddonManager (same with GetDisabledAddonInfos). AddonMgr still have the authority over which addons are installed an enabled so all of these might give wrong/outdated answers.

xbmc/peripherals/bus/virtual/PeripheralBusAddon.cpp
```diff
@@ -439,8 +383,9 @@ void CPeripheralBusAddon::UpdateAddons(void)
   std::set<std::string> removed;
 
   // Get new add-ons
-  PeripheralAddonVector newAddons = GetEnabledAddons();
-  std::transform(newAddons.begin(), newAddons.end(), std::inserter(newIds, newIds.end()), GetPeripheralAddonID);
+  BinaryAddonBaseList newAddons;
+  CServiceBroker::GetBinaryAddonManager().GetAddonInfos(newAddons, true, ADDON_PERIPHERALDLL);
```
tamland: Like here. This doesn't work. CPeripheralBusAddon listens to events from addon manager, so there is no guarantee BinaryAddonManager will have the correct list at this point. And then we are out of sync.

AlwinEsch: This and the other is thought as replacement for the CBinaryAddonCache and to remove them in future, both does mostly the same.

AlwinEsch: With use of GetAddon direct is everything lost why them both are present.

tamland: > This and the other is thought as replacement for the CBinaryAddonCache and to remove them in future, both does mostly the same.

Ok. No need to fix right now then. I see that the set_difference etc. below isn't correct either (doesn't not account for updates) so this whole method will need to be changed at some point.

AlwinEsch:
@garbear OK from your side to bring in?

jenkins build this please

AlwinEsch:
jenkins build this please

garbear:
OK from my side. Concerning `PeripheralUtils.h`, it would be good to convert this to the class system that you've implemented for the rest of the add-ons. This header was just my filler until we had a C++ interface. But this can happen anytime... I'll be in the mountains for the next week or two so I can look into this when I return.

AlwinEsch:
jenkins build this please

AlwinEsch:
jenkins build this please

PR #12335, opened 2017-06-22
[peripheral] Cosmetic: Fix doxy indentation

garbear:
Just a detail I noticed while reviewing #12333

I also discovered Input Stream add-ons had the same doxy problem, so I included that fix too


garbear:
Sorry, wrong repo. Correct PR is https://github.com/AlwinEsch/kodi/pull/30

PR #12337, opened 2017-06-23
[network] fix CID 1364572 - ensure we're not leaking the ipv6 test socket

mkortstiege:


ace20022 (approved):

wsnipex:
Since the class is called CPosixUDPSocket, shouldn't we be using close instead of the windows function closesocket?

ace20022:
> Since the class is called CPosixUDPSocket, shouldn't we be using close instead of the windows function closesocket?

Might be better. This is quite messy. At line 34 there's
```
#ifdef WINSOCK_VERSION
#define close closesocket
#endif
```

and in \xbmc\linux\PlatformDefs.h there's

```
#ifdef TARGET_POSIX
....

#define closesocket(s)  close(s)
...
```

mkortstiege:
IMO we should merge this one and do a cleanup and rename of this class in another PR. The name is misleading as it's not Posix exclusive.

Paxxi:
Agree with @mkortstiege, whole networking implementation needs a cleanup and proper abstraction layer to avoid all of this

PR #12339, opened 2017-06-23
[bluray] Blu-ray updates

ace20022:
This PR

- sets the player profile to 5.0
- adds bd-j menu support for isos
- adds bd-j menu support for disc on samba shares and the like

### DIFF
--- a/lib/DllLibbluray.h
+++ b/lib/DllLibbluray.h
@@ -31,6 +31,8 @@ extern "C"
 #include <libbluray/overlay.h>
 }
 
+typedef int(*read_blocks_f)(void *handle, void *buf, int lba, int num_blocks);
+
 class DllLibblurayInterface
 {
 public:
@@ -41,6 +43,7 @@ public:
   virtual void bd_free_title_info(BLURAY_TITLE_INFO *title_info)=0;
   virtual BLURAY *bd_open(const char* device_path, const char* keyfile_path)=0;
   virtual int bd_open_disc(BLURAY *bd, const char *device_path, const char *keyfile_path)=0;
+  virtual int bd_open_stream(BLURAY *bd, void *read_blocks_handle, read_blocks_f func) = 0;
   virtual BLURAY *bd_init(void)= 0;
   virtual void bd_close(BLURAY *bd)=0;
   virtual int64_t bd_seek(BLURAY *bd, uint64_t pos)=0;
@@ -92,6 +95,7 @@ class DllLibbluray : public DllDynamic, DllLibblurayInterface
   DEFINE_METHOD1(void,                bd_free_title_info,     (BLURAY_TITLE_INFO *p1))
   DEFINE_METHOD2(BLURAY*,             bd_open,                (const char* p1, const char* p2))
   DEFINE_METHOD3(int,                 bd_open_disc,           (BLURAY *p1, const char *p2, const char *p3))
+  DEFINE_METHOD3(int,                 bd_open_stream,         (BLURAY *p1, void *p2, read_blocks_f p3))
   DEFINE_METHOD0(BLURAY*,             bd_init)
   DEFINE_METHOD1(void,                bd_close,               (BLURAY *p1))
   DEFINE_METHOD2(int64_t,             bd_seek,                (BLURAY *p1, uint64_t p2))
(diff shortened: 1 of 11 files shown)


ace20022:
@koying @Memphiz @wsnipex what is the min. version of libburay on your platforms? Would be nice to drop the preprocessor version checks.

wsnipex:
per default ubuntu, both current LTS releases have a lower version the 0.9.3. I'm currently looking into getting 1.0.1 into our PPAs though.
I'll let you know how it goes over the weekend.

Rechi:
@ace20022 libbluray version for depends based platforms (osx, ios, android) is 1.0.1, but it is compiled without bd-j support.

wsnipex:
1.0.1 with bdj support enabled is now in our nightly PPA - untested though

xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamBluray.cpp
```diff
@@ -50,6 +50,22 @@
 
 using namespace XFILE;
 
+
+static int read_blocks(void* handle, void* buf, int lba, int num_blocks)
```
Paxxi: don't methods like this usually return a `size_t` or `ssize_t`?

xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamBluray.cpp
```diff
@@ -153,26 +137,72 @@ int DllLibbluray::dir_read(BD_DIR_H *dir, BD_DIRENT *entry)
...
-    if(!CDirectory::GetDirectory(strDirname, st->list))
-    {
-      CLog::Log(LOGDEBUG, "CDVDInputStreamBluray - Error opening dir! (%s)\n", dirname);
-      delete st;
-      return NULL;
-    }
+  std::string strDirname = URIUtils::AddFileToFolder(*strBasePath, strRelPath);
+  URIUtils::RemoveSlashAtEnd(strDirname);
+
+  SDirState *st = new SDirState();
```
Paxxi: might just use a local here as it doesn't outlive the method scope

xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamBluray.cpp
```diff
@@ -467,6 +529,15 @@ void CDVDInputStreamBluray::Close()
   }
   m_bd = NULL;
   m_title = NULL;
+  if (m_pstream != nullptr)
```
Paxxi: `m_pstream` and `m_rootPath` will never be null here, their content might be that's not relevant so this can be simplified as
```
m_pstream.reset();
m_rootPath.reset();
```
as `Close()` will be called by the destructor when calling `reset()`

xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamBluray.h
```diff
@@ -172,5 +173,8 @@ class CDVDInputStreamBluray
 #endif
 
   private:
+    bool OpenStream(CFileItem &item);
     void SetupPlayerSettings();
+    std::unique_ptr<CDVDInputStreamFile> m_pstream;
+    std::unique_ptr<std::string> m_rootPath;
```
Paxxi: this could just be a regular `std::string` and pass it around using & as I understand the code. Simplifies things a bit.

Paxxi (approved):
Looks good to me overall. Some nitpicks or tips depending on how you look at it 😄 . Nothing blocking a merge though.

xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamBluray.cpp
ace20022: Not in this case ;)

xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamBluray.cpp
ace20022: it does, see line 167

xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamBluray.cpp
ace20022: right

lib/DllLibbluray.h
```diff
@@ -29,8 +29,13 @@ extern "C"
 #include <libbluray/log_control.h>
 #include <libbluray/keys.h>
 #include <libbluray/overlay.h>
+#include <libbluray/bluray-version.h>
```
Rechi: as you have removed the ifdef version checks again this isn't needed

Rechi:
Your changes require now libbluray 0.9.3 or higher, can you please update `cmake/modules/FindBluray.cmake`, `docs/README.linux` and `docs/README.opensuse`

ace20022:
jenkins build this please

PR #12343, opened 2017-06-23
Fix massive faults

AlwinEsch:
## Description
To see fault must be somewhere from settings a folder selected where
it start with "/../" to reduce length.

This freeze complete Kodi and can only be closed by a force,
is also confirmed fault on libreelec (tested on raspberry).

There becomes std::string::npos required to exit loop who can never come!
This is replaced to a "0" who works.

and fix crash if <control type="button" format="file"/> is used, currently not used from Kodi itself therefore not yet noticed.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/Util.cpp
+++ b/xbmc/Util.cpp
@@ -1465,9 +1465,14 @@ bool CUtil::MakeShortenPath(std::string StrInput, std::string& StrOutput, size_t
   {
     nPos = StrInput.find_last_of( cDelim, nPos );
     nGreaterDelim = nPos;
-    if ( nPos != std::string::npos )
-      nPos = StrInput.find_last_of( cDelim, nPos - 1 );
-    if ( nPos == std::string::npos ) break;
+
+    if (nPos == std::string::npos || nPos == 0)
+      break;
+
+    nPos = StrInput.find_last_of( cDelim, nPos - 1 );
+
+    if ( nPos == std::string::npos)
+      break;
     if ( nGreaterDelim > nPos ) StrInput.replace( nPos + 1, nGreaterDelim - nPos - 1, ".." );
     iStrInputSize = StrInput.size();
   }
--- a/xbmc/settings/SettingPath.cpp
+++ b/xbmc/settings/SettingPath.cpp
@@ -56,8 +56,8 @@ bool CSettingPath::Deserialize(const TiXmlNode *node, bool update /* = false */)
   if (!CSettingString::Deserialize(node, update))
     return false;
     
-  if (m_control != NULL &&
-     (m_control->GetType() != "button" || m_control->GetFormat() != "path"))
+  if (m_control != nullptr &&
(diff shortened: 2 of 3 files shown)


xbmc/Util.cpp
```diff
@@ -1467,8 +1467,10 @@ bool CUtil::MakeShortenPath(std::string StrInput, std::string& StrOutput, size_t
     nGreaterDelim = nPos;
     if ( nPos != std::string::npos )
       nPos = StrInput.find_last_of( cDelim, nPos - 1 );
-    if ( nPos == std::string::npos ) break;
-    if ( nGreaterDelim > nPos ) StrInput.replace( nPos + 1, nGreaterDelim - nPos - 1, ".." );
+    if ( nPos == 0 )
```
Paxxi: Why is this compared against 0? Not found returns `std::string::npos`

AlwinEsch: No direct idea why not found. But it freeze the whole Kodi by Dead Loop and bring the CPU in a hot round.

Paxxi: I don't argue that there's a bug but this solution isn't a proper one.
Can you provide an example input that causes the issue?

AlwinEsch:
On libreelec the following steps (you need to have "Expert" settings enabled):

Go to "Settings" -> "Player" -> "Discs" -> "Saved music folder" ->
Select a long folder, where you see the "/../../" -> Press OK.

Not all lengths bring the problem, sometimes on much bigger it works.

But e.g. this bring the fault:
"/storage/videos/Star_Trek_Das_nächste_Jahrhundert"

xbmc/Util.cpp
AlwinEsch: On libreelec the following steps (you need to have "Expert" settings enabled):

Go to "Settings" -> "Player" -> "Discs" -> "Saved music folder" ->
Select a long folder, where you see the "/../../" -> Press OK.

Not all lengths bring the problem, sometimes on much bigger it works.

But e.g. this bring the fault:
"/storage/videos/Star_Trek_Das_nächste_Jahrhundert"

Paxxi: like so https://github.com/Paxxi/xbmc/commit/f7bc7dd60cf12043f3aef00d65f76e1fa9e71b3c

AlwinEsch: Looks better :-) , if works can it be also like them in master (where I added already the way from here, Sorry!).

Paxxi: the test cases pass, the last case failed before the changes so I'm fairly sure it's a valid solution

AlwinEsch: How do you think we'll take your better fix. I can cherry-pick your one or do you create a separate request?

AlwinEsch:
jenkins build this please

MartijnKaijser:
@Paxxi this backport ok?

Paxxi (approved):

PR #12350, opened 2017-06-23
[cleanup] Remove HAS_FILESYSTEM

hudokkow:
Another define that doesn't make sense.
Build tested, brief runtime test.


notspiff:
It made lots of sense when porting to new platforms.

MartijnKaijser:
jenkins build this please

PR #11775, opened 2017-02-27
[PVR] Krypton: fix timer type display for 'invalid' types

metaron-uk:
Backport of https://github.com/xbmc/xbmc/pull/11753

Tested on x86 with pvr.mythtv

PR #12352, opened 2017-06-24
Ignore FORBIDS_EPG_TAG_ON_CREATE timer types when creating via CreateFromEpg()

djp952:
## Description
After pull request 11775 [https://github.com/xbmc/xbmc/pull/11775](https://github.com/xbmc/xbmc/pull/11775), selecting "Add Timer" from an EPG context menu will erroneously add a timer(s) that have the PVR_TIMER_TYPE_FORBIDS_EPG_TAG_ON_CREATE attribute set.  

Assuming that pull request 11775 must remain in place to resolve the original problem, my proposed solution is to add filtering for PVR_TIMER_TYPE_FORBIDS_EPG_TAG_ON_CREATE into CPVRTimerInfoTag::CreateFromEpg() such that a timer with this attribute set will never be selected in this context and subsequently passed onto CGUIDialogPVRTimerSettings.

## Motivation and Context
The expectation is that timer(s) created with the PVR_TIMER_TYPE_FORBIDS_EPG_TAG_ON_CREATE attribute set would never appear in the selection dialog when accessed via an EPG element's "Add Timer" context menu.

## How Has This Been Tested?
Testing was performed on Windows, using a debug build of 17.3-Krypton (x86).  The tested PVR module is an unofficial custom PVR [https://github.com/djp952/pvr.hdhomerundvr](https://github.com/djp952/pvr.hdhomerundvr) that defines timers with the PVR_TIMER_TYPE_FORBIDS_EPG_TAG_ON_CREATE.  Prior to the change a timer with that attribute set would appear in the EPG context menu.  After the change the erroneous timer no longer appears.  The PVR has a limited number of timer types implemented (specifically there are no MANUAL timer types), but tests of various other timer integration points appear to be normal afterwards.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/pvr/timers/PVRTimerInfoTag.cpp
+++ b/xbmc/pvr/timers/PVRTimerInfoTag.cpp
@@ -816,7 +816,7 @@ CPVRTimerInfoTagPtr CPVRTimerInfoTag::CreateFromEpg(const CEpgInfoTagPtr &tag, b
     // create epg-based timer rule
     timerType = CPVRTimerType::CreateFromAttributes(
       PVR_TIMER_TYPE_IS_REPEATING,
-      PVR_TIMER_TYPE_IS_MANUAL | PVR_TIMER_TYPE_FORBIDS_NEW_INSTANCES, channel->ClientID());
+      PVR_TIMER_TYPE_IS_MANUAL | PVR_TIMER_TYPE_FORBIDS_NEW_INSTANCES | PVR_TIMER_TYPE_FORBIDS_EPG_TAG_ON_CREATE, channel->ClientID());
 
     if (timerType)
     {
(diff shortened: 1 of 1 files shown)


Rechi:
@djp952 is this also needed for master? If yes please create a PR for master too.
Normally we create a PR for master first and then decide if we are going to backport it.

djp952:
It shouldn't hurt to be in master, but I think the original pull request that caused the concern may need to be reviewed since it changed the code that would have prevented the problem.  My PR feels more like a Band-Aid than a final solution. I'll submit a PR for master and indicate that in my commentary - thank you very much!

MartijnKaijser:
ping @ksooo @metaron-uk 
We rather fix it correctly than band-aid (even for release branch)

metaron-uk (changes requested):
Makes sense to me and aligns with the definition in the include file:
xbmc_pvr_types.h:143:  const unsigned int PVR_TIMER_TYPE_FORBIDS_EPG_TAG_ON_CREATE         = 0x00400000; /*!< @brief this type should not appear on any create menus which provide an associated EPG tag */

In my PR, I was expecting this line:
https://github.com/xbmc/xbmc/pull/11775/files#diff-81ee6b50ffdcedb2e947a807469bf195R760
To prevent excluded timer types from appearing as an option in this case.

The problem is that having experimented (making the 'manual type in mythtv forbid an associated EPG entry, which it normally doesn't and cherry picking into a not-so-recent leia build to test) it doesn't seem to limit the options available from an EPG tag's 'Add timer' context menu (which would be the proper fix i.m.o.)

Maybe the route cause is that the context menu selection doesn't create and associate the draft timer record with the parent EPG entry before asking CGUIDialogPVRTimerSettings::InitializeTypesList() to build the selection list.

Are you able to investigate this?

ksooo (approved):
Makes sense and is the proper fix, imo.

metaron-uk:
Oops. Seems my previous run-time test was flawed. This patch seems to work fine on its own now I've properly installed my modified pvr.mythtv version - sorry. Good to go now from my perspective.

ksooo:
@metaron-uk this PR will be superceeded by backport of #12380

ksooo:
jenkins build this please

ksooo:
jenkins errors are unrelated

PR #12354, opened 2017-06-24
Ignore FORBIDS_EPG_TAG_ON_CREATE timer types when creating a timer info tag via CreateFromEpg()

djp952:
## Description
After pull request 11775 [https://github.com/xbmc/xbmc/pull/11775](https://github.com/xbmc/xbmc/pull/11775) applied in 17.1-Krypton, selecting "Add Timer" from an EPG context menu will erroneously add a timer(s) that have the PVR_TIMER_TYPE_FORBIDS_EPG_TAG_ON_CREATE attribute set.  

Assuming that pull request 11775 must remain in place to resolve the original problem, my proposed solution is to add filtering for PVR_TIMER_TYPE_FORBIDS_EPG_TAG_ON_CREATE into CPVRTimerInfoTag::CreateFromEpg() such that a timer with this attribute set will never be selected in this context and subsequently passed onto CGUIDialogPVRTimerSettings.

I have a concern that the behavior applied for PR 11775 may need to be re-reviewed as no other functions in PVRTimerInfoTag perform this level of filtering, making this PR feel more like a Band-Aid that could be best served as just a temporary patch to restore the functionality without breaking the mythtv PVR addon in the process.

## Motivation and Context
The expectation is that timer(s) created with the PVR_TIMER_TYPE_FORBIDS_EPG_TAG_ON_CREATE attribute set would never appear in the selection dialog when accessed via an EPG element's "Add Timer" context menu.

## How Has This Been Tested?
Testing was performed on Windows, using a debug build of 17.3-Krypton (x86).  The tested PVR module is an unofficial custom PVR [https://github.com/djp952/pvr.hdhomerundvr](https://github.com/djp952/pvr.hdhomerundvr) that defines timers with the PVR_TIMER_TYPE_FORBIDS_EPG_TAG_ON_CREATE.  Prior to the change a timer with that attribute set would appear in the EPG context menu.  After the change the erroneous timer no longer appears.  The PVR has a limited number of timer types implemented (specifically there are no MANUAL timer types), but tests of various other timer integration points appear to be normal afterwards.

I have not tested the PR on the master branch (my PVR is not updated for Leia/master yet) although I have done a static analysis of the affected code and believe the behavior would be the same.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/pvr/timers/PVRTimerInfoTag.cpp
+++ b/xbmc/pvr/timers/PVRTimerInfoTag.cpp
@@ -815,7 +815,7 @@ CPVRTimerInfoTagPtr CPVRTimerInfoTag::CreateFromEpg(const CPVREpgInfoTagPtr &tag
     // create epg-based timer rule
     timerType = CPVRTimerType::CreateFromAttributes(
       PVR_TIMER_TYPE_IS_REPEATING,
-      PVR_TIMER_TYPE_IS_MANUAL | PVR_TIMER_TYPE_FORBIDS_NEW_INSTANCES, channel->ClientID());
+      PVR_TIMER_TYPE_IS_MANUAL | PVR_TIMER_TYPE_FORBIDS_NEW_INSTANCES | PVR_TIMER_TYPE_FORBIDS_EPG_TAG_ON_CREATE, channel->ClientID());
 
     if (timerType)
     {
@@ -837,7 +837,7 @@ CPVRTimerInfoTagPtr CPVRTimerInfoTag::CreateFromEpg(const CPVREpgInfoTagPtr &tag
     // create one-shot epg-based timer
     timerType = CPVRTimerType::CreateFromAttributes(
       PVR_TIMER_TYPE_ATTRIBUTE_NONE,
-      PVR_TIMER_TYPE_IS_REPEATING | PVR_TIMER_TYPE_IS_MANUAL | PVR_TIMER_TYPE_FORBIDS_NEW_INSTANCES, channel->ClientID());
+      PVR_TIMER_TYPE_IS_REPEATING | PVR_TIMER_TYPE_IS_MANUAL | PVR_TIMER_TYPE_FORBIDS_NEW_INSTANCES | PVR_TIMER_TYPE_FORBIDS_EPG_TAG_ON_CREATE, channel->ClientID());
   }
 
   if (!timerType)


ksooo:
@metaron-uk did I understand your comment in the coresponding Krypton-PR right, that this fix does not work for Leia, but for Kypton?

metaron-uk:
Not quite.
I haven't tested krypton, but the behavior in leia isn't what I was 
expecting. In my opinion the propper fix is to stop FORBIDS_EPG timers 
appearing in the list  but i'm still seeing them with this patch applied.
My pr didn't change the behaviour as far as I can tell, so probably it was 
already there.
More investigation required imo.

Sent with AquaMail for Android
http://www.aqua-mail.com

ksooo:
@metaron-uk so, what do you suggest? Should we merge this PR?

ksooo (approved):
@metaron-uk we need this PR and another fix (in another PR) somewhere else to get it fully working.

ksooo:
@metaron-uk sorry, cannot reproduce. I applied this PR to latest kodi master and patched pvr.hts manual oneshot tomertype to inlcude PVR_TIMER_TYPE_FORBIDS_EPG_TAG_ON_CREATE. When selecting opening the context menu of an epg item in the guide window, then "Add timer", the type in question does NOT appear in the list of available timer types. All good. Or do I have to do something dofferent to reproduce the problem?

ksooo:
@metaron-uk does this solve your problem:

<pre>
diff --git a/xbmc/pvr/timers/PVRTimerInfoTag.cpp b/xbmc/pvr/timers/PVRTimerInfoTag.cpp
index 42144cfc7d..53c09df076 100644
--- a/xbmc/pvr/timers/PVRTimerInfoTag.cpp
+++ b/xbmc/pvr/timers/PVRTimerInfoTag.cpp
@@ -848,7 +848,7 @@ CPVRTimerInfoTagPtr CPVRTimerInfoTag::CreateFromEpg(const CPVREpgInfoTagPtr &tag
 
   newTag->SetTimerType(timerType);
   newTag->UpdateSummary();
-  newTag->UpdateEpgInfoTag();
+  newTag->SetEpgTag(tag);
 
   /* unused only for reference */
   newTag->m_strFileNameAndPath = CPVRTimersPath::PATH_NEW;
</pre>

metaron-uk:
@ksooo - I have a confession to make. 
It appears the test build I made of pvr.mythtv didn't install correctly. I tried your patch above and it didn't work.
I then re-installed my hacked pvr.mythtv version after changing the version reference in addon.xml.in (just to check it had actually installed properly) and it magically started working.
I then compiled and installed kodi without your patch and guess what - it was still working.
My mistake - sorry.

ksooo:
Yeah, #12380 makes the original fix "bullet-proof", but the core of the fix is still @djp952's commit. I just cherry picked it and added some more fail safeness for cases that actually could occur.

ksooo:
This PR can be closed as we go with #12380 for master.

PR #12355, opened 2017-06-24
[depends] - fixes for tvos

Memphiz:
Currently tvos compile of depends is broken in master due to new depends. This PR fixes compilation for tvos again.

### DIFF
--- a/tools/depends/target/libbluray/Makefile
+++ b/tools/depends/target/libbluray/Makefile
@@ -26,6 +26,9 @@ $(PLATFORM): $(TARBALLS_LOCATION)/$(ARCHIVE) $(DEPS)
 	sed -i -e "s|CFLAGS=|#CFLAGS=|" $(PLATFORM)/Makefile.am
 	cd $(PLATFORM); ./bootstrap
 	cd $(PLATFORM); $(CONFIGURE)
+ifeq ($(TARGET_PLATFORM),appletvos)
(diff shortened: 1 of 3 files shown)


tools/depends/target/libbluray/Makefile
```diff
@@ -26,6 +26,11 @@ $(PLATFORM): $(TARBALLS_LOCATION)/$(ARCHIVE) $(DEPS)
 	sed -i -e "s|CFLAGS=|#CFLAGS=|" $(PLATFORM)/Makefile.am
 	cd $(PLATFORM); ./bootstrap
 	cd $(PLATFORM); $(CONFIGURE)
+ifeq ($(OS),ios)
+ifeq ($(CPU), arm64)
```
Rechi: Shouldn't this be `ifeq ($(TARGET_PLATFORM),appletvos)`?
If it is also valid for iOS 64 bit please change the patch filename.

Memphiz:
thx @Rechi you are right - i didn't remember that i can distinguish tvos and ios in depends ;) - i also did a better patch to the pythonmodule-pycryptodome (the same approach as for pillow)

Memphiz:
jenkins is basically fine - http://jenkins.kodi.tv/job/BuildMulti-PR/3420/ win32 still builds, win64 failed because it couldn't remove the cmake directory - but i guess those platforms are uninteresting anyways. If this is ok with you @Rechi please just push the button - have to go afk now.

PR #12361, opened 2017-06-25
VAAPI: Return false on FF_PROFILE_HEVC_MAIN_10

fritsch:
Background: https://forum.kodi.tv/showthread.php?tid=317137


fritsch:
@wsnipex @FernetMenta for information.

Citation from the forum:
```
Also, Kodi explicitly allows and supports the use of system ffmpeg - the CMake build system has an option for it. So it's not like Gentoo is patching or otherwise messing with Kodi in some crazy way.
```

wsnipex:
good to go?

fritsch:
Better than black screen, yes.

Am 07.07.2017 9:15 vorm. schrieb "Wolfgang Schupp" <notifications@github.com

PR #12362, opened 2017-06-25
FFmpeg: Bump to 3.1.9-Krypton-17.4 for 17.4 release

fritsch:
This bumps our internal ffmpeg from 3.1.6 to 3.1.9 to receive the multiple security bugs fixea in ffmpeg branch. It seems an automated code checker was involved upstream.

### DIFF
--- a/tools/depends/target/ffmpeg/FFMPEG-VERSION
+++ b/tools/depends/target/ffmpeg/FFMPEG-VERSION
@@ -1,5 +1,5 @@
 LIBNAME=ffmpeg
 BASE_URL=https://github.com/xbmc/FFmpeg/archive
-VERSION=3.1.6-Krypton
+VERSION=3.1.9-Krypton-17.4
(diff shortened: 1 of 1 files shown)


fritsch:
@FernetMenta What do you think? I know we neither have the capability to properly test that nor is development still Krypton centric. Never the less I think if we release a new version based on internal ffmpeg, we should use the latest and greatest fixes upstream does for our 3.1 branch.

If you think differently, just close the PR.

FernetMenta:
Sorry, I don't now. I have not even looked at the changelog of ffmpeg. What does it fix? Are the changes important. If there are plans for a 17.5 it still can be reverted.
I'd say yes, but I leave it to @MartijnKaijser to take the final decision.

MartijnKaijser:
We still have nightlies for 17.4 and we can do a RC before final. Revert is always an option.
Did some quick playback tests on windows and no obvious issues.

mazeyx:
seems to work fine on android arm also.

PR #12365, opened 2017-06-26
Joystick improvements for Game OSD

garbear:
This PR is the first part of #12367. It contains a set of improvements to Kodi's joystick code. The functionality should remain unchanged.

## Motivation and Context
Cleanup so that bad code doesn't make its way into the new keymap handling.

## How Has This Been Tested?
Input works in and out of games.


garbear:
jenkins build this please

PR #12369, opened 2017-06-26
[fix][pictures] CID 77607: prevent buffer overflow

Rechi:
## Description
Prevent a buffer overflow by copying maximal `sizeof(dest) - 1` chars.

## Motivation and Context
calling `ListItem.setInfo('pictures', {'exif:exiftime': '<VALUE>'})` from a python addon with a too long value sets other exif info with invalid data.

## How Has This Been Tested?
above call with a long value

## Types of change

## Checklist:

### DIFF
--- a/xbmc/pictures/PictureInfoTag.cpp
+++ b/xbmc/pictures/PictureInfoTag.cpp
@@ -1033,7 +1033,8 @@ void CPictureInfoTag::SetInfo(int info, const std::string& value)
     }
   case SLIDE_EXIF_DATE_TIME:
     {
-      strcpy(m_exifInfo.DateTime, value.c_str());
+      strncpy(m_exifInfo.DateTime, value.c_str(), sizeof(m_exifInfo.DateTime) - 1);
+      m_exifInfo.DateTime[sizeof(m_exifInfo.DateTime) - 1] = '\0';
       m_isInfoSetExternally = true; // Set the internal state to show metadata has been set by call to SetInfo
       ConvertDateTime();
       break;


Rechi:
@MartijnKaijser should this be backported?

xbmc/pictures/PictureInfoTag.cpp
```diff
@@ -1033,7 +1033,7 @@ void CPictureInfoTag::SetInfo(int info, const std::string& value)
     }
   case SLIDE_EXIF_DATE_TIME:
     {
-      strcpy(m_exifInfo.DateTime, value.c_str());
+      strncpy(m_exifInfo.DateTime, value.c_str(), sizeof(m_exifInfo.DateTime) - 1);
```
popcornmix: This string will be unterminated if length is exceeded. Adding;
m_exifInfo.DateTime[sizeof(m_exifInfo.DateTime) - 1)] = '\0';
should avoid that.
See notes here: https://linux.die.net/man/3/strncpy

Rechi: not really needed, because of `memset(&m_exifInfo, 0, sizeof(m_exifInfo))` (https://github.com/xbmc/xbmc/blob/4a913d0/xbmc/pictures/PictureInfoTag.cpp#L33)
If I still should add it just tell me.

MartijnKaijser:
If it fixes something why not

xbmc/pictures/PictureInfoTag.cpp
Rechi: added the explicit null termination, to not get in trouble if something gets changed in the future.

PR #12370, opened 2017-06-26
service addons: replace synchronous add-on hooks with event handlers 

tamland:
This should fix the issues with services getting started multiple times.

### DIFF
--- a/xbmc/ServiceManager.cpp
+++ b/xbmc/ServiceManager.cpp
@@ -98,6 +98,7 @@ bool CServiceManager::Init2()
 
   m_favouritesService.reset(new CFavouritesService(CProfilesManager::GetInstance().GetProfileUserDataFolder()));
   m_contextMenuManager.reset(new CContextMenuManager(*m_addonMgr.get()));
+  m_serviceAddons.reset(new ADDON::CServiceAddonManager(*m_addonMgr));
 
   init_level = 2;
   return true;
@@ -145,6 +146,7 @@ bool CServiceManager::Init3()
 
 void CServiceManager::Deinit()
 {
+  m_serviceAddons.reset();
   m_gameServices->Deinit();
   m_peripherals.reset();
   m_contextMenuManager.reset();
@@ -182,6 +184,11 @@ ADDON::CBinaryAddonManager &CServiceManager::GetBinaryAddonManager()
 ADDON::CVFSAddonCache &CServiceManager::GetVFSAddonCache()
 {
   return *m_vfsAddonCache.get();
+}
+
+ADDON::CServiceAddonManager &CServiceManager::GetServiceAddons()
+{
+  return *m_serviceAddons;
 }
 
 ANNOUNCEMENT::CAnnouncementManager& CServiceManager::GetAnnouncementManager()
--- a/xbmc/addons/AddonEvents.h
+++ b/xbmc/addons/AddonEvents.h
@@ -48,6 +48,28 @@ namespace ADDON
       MetadataChanged(std::string id) : id(std::move(id)) {}
     };
 
+    /**
+     * Emitted when a different version of the add-on has been installed
+     * to the file system and should be reloaded.
+     */
+    struct ReInstalled: AddonEvent
+    {
+      std::string id;
+      ReInstalled(std::string id) : id(std::move(id)) {}
+    };
+
+    /**
+     * Emitted after the add-on has been uninstalled.
+     */
+    struct UnInstalled : AddonEvent
+    {
+      std::string id;
+      UnInstalled(std::string id) : id(std::move(id)) {}
+    };
+
+    /**
+     * @deprecated Use Enabled, ReInstalled and UnInstalled instead.
+     */
     struct InstalledChanged : AddonEvent {};
   };
 };
--- a/xbmc/addons/AddonManager.cpp
+++ b/xbmc/addons/AddonManager.cpp
@@ -273,8 +273,7 @@ static bool LoadManifest(std::set<std::string>& system, std::set<std::string>& o
 
 CAddonMgr::CAddonMgr()
   : m_cp_context(nullptr),
-  m_cpluff(nullptr),
-  m_serviceSystemStarted(false)
+  m_cpluff(nullptr)
 { }
 
 CAddonMgr::~CAddonMgr()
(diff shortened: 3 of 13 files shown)


xbmc/addons/Service.cpp
```diff
@@ -20,89 +20,148 @@
...
+void CServiceAddonManager::Start()
 {
-  bool ret = true;
+  m_addonMgr.Events().Subscribe(this, &CServiceAddonManager::OnEvent);
+  VECADDONS addons;
+  if (m_addonMgr.GetAddons(addons, ADDON_SERVICE))
+  {
+    for (const auto& addon : addons)
+    {
+     Start(addon);
```
AlwinEsch: One pointer missing

xbmc/addons/Service.h
```diff
@@ -19,41 +19,74 @@
...
+     * Stop all services.
+     */
+    void Stop();
+
+    /**
+     * Stop service by add-on id.
+     */
+    void Stop(const std::string& addonId);
+
+    void OnEvent(const AddonEvent& event);
```
AlwinEsch: Is not needed to have public.

xbmc/addons/Service.cpp
```diff
@@ -20,89 +20,148 @@
...
+  return std::unique_ptr<CService>(new CService(std::move(addonInfo), startOption));
+}
+
+CService::CService(CAddonInfo addonInfo, START_OPTION startOption)
+  : CAddon(std::move(addonInfo)), m_startOption(startOption)
+{
 }
 
+CServiceAddonManager::CServiceAddonManager(CAddonMgr& addonMgr) :
+    m_addonMgr(addonMgr)
```
AlwinEsch: Think 2 pointers are enough :)

xbmc/addons/AddonEvents.h
```diff
@@ -48,6 +48,28 @@ namespace ADDON
...
+      ReInstalled(std::string id) : id(std::move(id)) {}
+    };
+
+    /**
+     * Emitted after the add-on has been uninstalled.
+     */
+    struct UnInstalled : AddonEvent
+    {
+      std::string id;
+      UnInstalled(std::string id) : id(std::move(id)) {}
```
AlwinEsch: Thats a point what make me unsure. The AddonEvent's becomes changed without blocking of related place who called them.

There can be a addon uninstalled before the related addon becomes stopped by disable from his Manager.

AlwinEsch:
On the whole the whole it looks good for me.

xbmc/addons/AddonEvents.h
AlwinEsch: For the most I find the non blocking very good, but think on one place the related addon manager need to become the time to do his work.

xbmc/addons/AddonManager.cpp
```diff
@@ -734,10 +734,34 @@ bool CAddonMgr::ReloadAddon(AddonPtr& addon)
   if (!addon ||!m_cpluff || !m_cp_context)
     return false;
 
-  m_cpluff->uninstall_plugin(m_cp_context, addon->ID().c_str());
-  return FindAddons()
-      && GetAddon(addon->ID(), addon, ADDON_UNKNOWN, false)
-      && EnableAddon(addon->ID());
+  AddonPtr tmp;
```
mkortstiege: Is there a reason you're using a temporary AddonPtr for the reinstall check?

xbmc/addons/AddonEvents.h
tamland: It doesn't really matter if something is stopped a few moments after uninstall instead of before, that's why.
Services have actually always been stopped ascync on uninstall even though Stop was called in the pre-hook (since CScriptInvocationManager::Stop is non-blocking).

xbmc/addons/AddonManager.cpp
tamland: It's an out parameter of GetAddon

xbmc/addons/AddonEvents.h
AlwinEsch: Yes but if you have a running dll and this becomes deleted or changed during his runtime does Kodi crash.

This can maybe happen if AddonEvents::Disabled is called and the addon type manager need a bit to stop his running addon.

xbmc/addons/AddonManager.cpp
mkortstiege: Sure. You're not using a temp AddonPtr in Line 750 when doing a GetAddon() check there.

xbmc/addons/AddonEvents.h
AlwinEsch: Currently use CPVRManager the `ADDON::OnPreUnInstall` way, where the addon calls then the PVR manager and this stop his addon.

xbmc/addons/AddonManager.cpp
tamland: Because `addon` is the out parameter of this method and it will be the new instance after reload, so if you use it here and it failed (and the caller doesn’t check the return status) it will have the wrong one (maybe it doesn't really matter as you always *should* check the return value)

xbmc/addons/AddonEvents.h
tamland: > Yes but if you have a running dll and this becomes deleted or changed during his runtime does Kodi crash.

Not on *nix, as long as it's kept open. I don't know what windows does (probably something stupid) but if this is an issue there, we can still handle it just fine in a non-blocking way (e.g by 'uninstalling' it from the addon system, but not actually delete the underlying files until later)

xbmc/ServiceManager.cpp
```diff
@@ -174,6 +176,11 @@ ADDON::CBinaryAddonManager &CServiceManager::GetBinaryAddonManager()
   return *m_binaryAddonManager.get();
 }
 
+ADDON::CServiceAddonManager &CServiceManager::GetServiceAddons()
+{
+  return *m_serviceAddons.get();
```
garbear: This can be just `return *m_serviceAddons;`. Same as above (line 97) too.

Aside from that, this PR looks good.

MartijnKaijser:
@tamland this needs a rebase

tamland:
All good?

MartijnKaijser:
Jenkins build this please

xbmc/addons/Service.cpp
Montellese: You missed this comment.

xbmc/addons/Service.cpp
```diff
@@ -20,89 +20,148 @@
...
+  m_addonMgr.Events().Unsubscribe(this);
 }
 
-bool CService::Start()
+void CServiceAddonManager::OnEvent(const ADDON::AddonEvent& event)
 {
-  bool ret = true;
-  switch (m_type)
+  if (auto enableEvent = dynamic_cast<const AddonEvents::Enabled*>(&event))
   {
```
Montellese: All these curly braces for a single line block does not match our coding style (throughout the whole `CServiceAddonManager` implementation.

xbmc/addons/Service.h
```diff
@@ -19,41 +19,74 @@
...
+  private:
+    START_OPTION m_startOption;
+  };
 
-    explicit CService(CAddonInfo addonInfo) : CAddon(std::move(addonInfo)), m_type(UNKNOWN), m_startOption(LOGIN) {}
-    CService(CAddonInfo addonInfo, TYPE type, START_OPTION startOption);
+  class CServiceAddonManager
+  {
+  public:
+    CServiceAddonManager(CAddonMgr& addonMgr);
```
Montellese: This constructor should be marked `explicit`.

Montellese (commented):
Sorry for the late review but I just came across this code while rebasing my media import stuff which also requires starting/stopping add-ons similar to services and noticed a few minors.

PR #12372, opened 2017-06-26
[database] correctly closing video and music databases

lobermann:
While rebasing feature_odb branch to current master and debugging some issues, I added correct close handling to all occurrences of the video and music database.

## Description
Every opened video and music database is now closed correctly, even on early returns.

## Motivation and Context
While testing odb I added some open and close code to the music and video database classes. So I had to make sure that close is always called before the datbase object is beeing destroyed. Although this is not that usefull right now, as it is not really needed to close it correctly (afaik), odb may need it. And it is more clear that Close is called all the time, and not only half of the time or not always. And, as I already wrote the code, I thought to create a PR.

## How Has This Been Tested?
I have imported a large movie, tvshow and music database, played various items, used the webinterface and android app, added some stuff to the library and removed some content.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/Autorun.cpp
+++ b/xbmc/Autorun.cpp
@@ -508,10 +508,12 @@ bool CAutorun::CanResumePlayDVD(const std::string& path)
   if (!strUniqueId.empty())
   {
     CVideoDatabase dbs;
-    dbs.Open();
     CBookmark bookmark;
-    if (dbs.GetResumeBookMark(strUniqueId, bookmark))
+    if (dbs.Open() && dbs.GetResumeBookMark(strUniqueId, bookmark))
+    {
+      dbs.Close();
       return true;
+    }
   }
   return false;
 }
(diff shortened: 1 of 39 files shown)


Montellese:
You do realize that the destructor of `CDatabase` is calling `Close()` right? IMO it would make more sense to remove all the unnecessary calls to `Close()`. There may be some cases where `Close()` is called on purpose before the instance is destroyed but that should be very few.

lobermann:
@Montellese I am aware of that, yes. But, during debugging I added some mutexes to the odb database class, and I got some deadlocks during that. Thats why I did this. 
I am also good if it goes in the direction of removing all the close() calls. But it may be better to do this later on when odb is ready? As you say, there may be some close() calls on purpose in there.

Tolriq:
For the record, Open / Close each times have a huge impact on the fact that MySQL is so slow. If a rewrite is made and keep MySQL support, going to a better handling would have huge performance gains in many cases (including local sqlite)

lobermann:
@Tolriq with odb we are using database pooling, so when you "close" a connection it will return to the pool.

Paxxi:
I don't really like this. We almost never should see an explicit `Close()` call. It means we failed to design CDatabase correctly and that should be fixed instead. We do the same thing with our CFile* classes which I also think is bad style.

lobermann:
@Paxxi then I suggest we close this PR and we will do it right in the feature_odb branch.

MartijnKaijser:
Closing as suggested.

PR #12373, opened 2017-06-26
[imdb] fix crash on empty videoInfoTag

lobermann:
Scraping with an invalid api key caused a crash after trying to scrape a few times.

## Description
When cleaning my database to re-scrape everything, I still had the old tmdb addon running, as there was an issue with the update. This caused kodi to crash after trying to scan for content a few times. As I catched the crash in the debugger I added some checks, as it seems in that case the videoInfoTag is not assigned. This should just prevent the crash when it tries to access the info tag.

## Motivation and Context
This prevents the crash I catched in the debugger.

## How Has This Been Tested?
I re-tested scraping a few times with the invalid key and it was not crashing any more. Noticed no issues.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/video/dialogs/GUIDialogVideoInfo.cpp
+++ b/xbmc/video/dialogs/GUIDialogVideoInfo.cpp
@@ -306,6 +306,11 @@ void CGUIDialogVideoInfo::SetMovie(const CFileItem *item)
 
   // setup cast list
   ClearCastList();
+  
+  // When the scraper throws an error, the video tag can be null here
+  if (!item->HasVideoInfoTag())
+    return;
+  
   MediaType type = item->GetVideoInfoTag()->m_type;
 
   m_startUserrating = m_movieItem->GetVideoInfoTag()->m_iUserRating;
--- a/xbmc/video/windows/GUIWindowVideoBase.cpp
+++ b/xbmc/video/windows/GUIWindowVideoBase.cpp
@@ -398,6 +398,10 @@ bool CGUIWindowVideoBase::ShowIMDB(CFileItemPtr item, const ScraperPtr &info2, b
     CGUIDialogOK::ShowAndGetInput(CVariant{13346}, CVariant{14057});
     return false;
   }
+  
+  // If the scraper failed above and no videoinfotag was created, return
+  if (!item->HasVideoInfoTag())
+    return false;
 
   bool listNeedsUpdating = false;
   // 3. Run a loop so that if we Refresh we re-run this block


MartijnKaijser:
backport perhaps?

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -306,6 +306,11 @@ void CGUIDialogVideoInfo::SetMovie(const CFileItem *item)
 
   // setup cast list
   ClearCastList();
+  
+  // When the scraper throws an error, the video tag can be null here
+  if (item->GetVideoInfoTag() == nullptr)
```
garbear: `if (item->HasVideoInfoTag())`

lobermann: nice, missed that one. Updating it now. Thanks.

mkortstiege:
Looks good to me. Backport won't hurt.

mkortstiege:
jenkins build this please

MartijnKaijser:
jenkins build this please

arnova:
Don't we want the video dialog to appear after failure so the user can refresh like I did here?: https://github.com/xbmc/xbmc/pull/12385

mkortstiege:
> Don't we want the video dialog to appear after failure so the user can refresh like I did here?: #12385

Unsure, but since this most likely only happens when there's a massive failure (eg. due to invalid api key) i don't think raising the info dialog would make sense here.

arnova:
@mkortstiege : I can't think of all the possible scenarious when this occurs. I'm just wondering whether still raising it could help as a temporary workaround for certain failures. Not sure though...

mkortstiege:
@arnova If we're going to re-open the info dialog it might look like we're done refreshing. I am unable to test it myself right now, but from a brief look there's no toast or error dialog involved.

arnova:
@mkortstiege : There are 2 error dialogs raised (see my comment in https://github.com/xbmc/xbmc/pull/12385), they are triggered by the RefreshItemModal-call in the refresh-loop.

lobermann:
I think asking for a reload does not make sense in this case, as it will fail again and then we still will not have a videoinfotag. But we could do a simple OK-Dialog/Notification saying there was an error scraping, just to inform the user.

arnova:
@lobermann : If you add an additional dialog you might as well move it down like I suggested as that will also give you an error dialog (followed by an empty info screen) which does the job just as well IMO. @mkortstiege : Your opinion please.

MartijnKaijser:
So what will it be so we can add it to v17. 4 as well

arnova:
@MartijnKaijser : Including this for 17.4 should be fine, although imo for master this could use some more thinking/refinement.

MartijnKaijser:
jenkins build this please

PR #12379, opened 2017-06-27
[addons] add help for load of shared library and allow use of own addon temp folder

AlwinEsch:
## Description
This is a part to simplify the load of shared library functions on a
addon.

This part is mainly needed for audio decoder addons.

Further add it a new addon callback function to get a temporary folder for addon itself. This becomes used e.g. on audio decoders to copy temporary a used dll to there.
Will make this way to ensure that temporary things are really erased again and there is no flooding on the disk.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/binary-addons/AddonDll.cpp
+++ b/xbmc/addons/binary-addons/AddonDll.cpp
@@ -323,6 +323,11 @@ void CAddonDll::Destroy()
   /* Unload library file */
   if (m_pDll)
   {
+    /* If temporary directory was used from addon delete them */
+    const std::string tempPath = URIUtils::AddFileToFolder("special://temp/addons", ID());
+    if (XFILE::CDirectory::Exists(tempPath))
+      XFILE::CDirectory::RemoveRecursive(CSpecialProtocol::TranslatePath(tempPath));
+
     /* Inform dll to stop all activities */
     if (m_needsavedsettings)  // If the addon supports it we save some settings to settings.xml before stop
     {
--- a/xbmc/addons/interfaces/General.cpp
+++ b/xbmc/addons/interfaces/General.cpp
@@ -26,11 +26,14 @@
 #include "addons/binary-addons/AddonDll.h"
 #include "addons/settings/GUIDialogAddonSettings.h"
 #include "dialogs/GUIDialogKaiToast.h"
+#include "filesystem/Directory.h"
+#include "filesystem/SpecialProtocol.h"
 #include "utils/CharsetConverter.h"
 #include "utils/log.h"
 #include "utils/LangCodeExpander.h"
 #include "utils/md5.h"
 #include "utils/StringUtils.h"
+#include "utils/URIUtils.h"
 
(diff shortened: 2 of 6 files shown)


AlwinEsch:
@notspiff Thats the part what I testet on your addon

xbmc/addons/kodi-addon-dev-kit/include/kodi/tools/DllHelper.h
```diff
@@ -0,0 +1,129 @@
...
+
+  /// @brief Function to register requested library symbol
+  ///
+  /// @note This function should not be used, use instead the macro
+  /// REGISTER_DLL_SYMBOL to register the symbol pointer.
+  ///
+  template <typename T>
+  bool RegisterSymbol(T& functionPtr, const char* strFunctionPtr)
+  {
+    functionPtr = (T)dlsym(m_dll, strFunctionPtr);
```
ksooo: c++ style cast instead of c style cast, please.

ksooo (commented):
Looking good. But where do you want to use this?

AlwinEsch:
See there the second commit https://github.com/notspiff/audiodecoder.timidity/pull/17

ksooo:
Okay. I assume you will remove Dllhelper.h from notspiff/audiodecoder.timidity#17 ?

AlwinEsch:
Yes was only to ask him.

There becomes also another commit added here to become a temporary folder for addon. On test it use the addon user folder, but not like to do there.

I add a function where a addon can get a individual folder where he can use for temporary parts, if now the addon becomes closed, becomes this deleted. Is to prevent in case of errors a flooded harddisk.

AlwinEsch:
Is updated

xbmc/addons/binary-addons/AddonDll.cpp
```diff
@@ -323,6 +323,11 @@ void CAddonDll::Destroy()
   /* Unload library file */
   if (m_pDll)
   {
+    /* If temporary directory was used from addon delete them */
+    std::string tempPath = URIUtils::AddFileToFolder("special://temp/addons", ID());
```
ksooo: this could be a "const std::string".

xbmc/addons/interfaces/General.cpp
```diff
@@ -239,4 +243,20 @@ void Interface_General::get_md5(void* kodiBase, const char* text, char* md5)
...
+char* Interface_General::get_temp_path(void* kodiBase)
+{
+  CAddonDll* addon = static_cast<CAddonDll*>(kodiBase);
+  if (addon == nullptr)
+  {
+    CLog::Log(LOGERROR, "Interface_General::%s - called with empty kodi instance pointer", __FUNCTION__);
+    return nullptr;
+  }
+
+  std::string tempPath = URIUtils::AddFileToFolder("special://temp/addons", addon->ID());
```
ksooo: this could be a "const std::string".

ksooo (approved):
you should rename the PR to reflect that it now also contains the temp path access function.

PR #12371, opened 2017-06-26
DBus refactoring

yol:
Refactor the DBus code on Linux to use RAII and have less repetition

## Description
Introduce RAII wrappers for `DBusConnection`, `DBusMessage`, and `DBusError`. The goal is still not to provide full C++ bindings for DBus, but just to make life a bit easier and make use of C++ comfort features.
I've tried not to change Kodi's behavior regarding PM etc.

## Motivation and Context
The DBus code suffers of a great deal of copy&paste because RAII isn't used and simple helpers for common stuff were not added.
Also, `dbus_connection_close` and  `dbus_connection_unref` are potentially called on a `NULL` connection in some places which is illegal.

## How Has This Been Tested?
Compiled for X11, log output checked

## Types of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [x] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #12383, opened 2017-06-28
DeviceKit cleanup

yol:
DeviceKit-disks was [renamed udisks](https://lists.freedesktop.org/archives/devkit-devel/2009-December/000567.html) in 2009, DeviceKit-power was [renamed upower](https://lists.freedesktop.org/archives/devkit-devel/2010-January/000728.html) in 2010. Kodi has support for UDisks and UPower. RIP DeviceKit.

## Description
Remove PM and storage implementations using DeviceKit.

## Motivation and Context
Follow-up from #12371 

## How Has This Been Tested?
Run-tested on X11

## Types of change

## Checklist:


MartijnKaijser:
jenkins build this please

Rechi:
There are two different ways to fix the raspberry pi build issue
- remove `storage/linux` directory at `cmake/treedata/rbpi/subdirs.txt`
- surround `core_add_library(...)` at `xbmc/storage/linux/CMakeLists.txt` with `if(NOT ENABLE_STATIC_LIBS OR DBUS_FOUND OR UDEV_FOUND)`

pkerling:
Just wondering, raspberry pi does not have udev?

MartijnKaijser:
@popcornmix?

popcornmix:
Pi does use udev (it is just standard linux).
I believe udev isn't enabled for jenkins builds purely due to limitations of the build system when cross compiling. @wsnipex may know more about this.

wsnipex:
udev is provided by systemd these days and we don't have that (nor old standalone udev) in depends.
It only works on x86 because we "cheat" too much there and allow the use of system libs.

I'm a bit puzzled about dbus though. We have it in depends, but it is never built. I've tried to track down if that was removed in the past, but can't find a ref that it was actually ever built...

Rechi:
@pkerling this needs a rebase

pkerling:
Done. Can we merge this then? Pi build stuff is peculiar, but not really related to this PR.

popcornmix:
If nothing has changed (i.e. Jenkins builds for Pi successfully without udev enabled) then no objections here.

PR #12385, opened 2017-06-28
fixed: Crash when we're unable to retrieve info from remote server

arnova:
Fix for the crash mentioned in my team-email. If ok'ed, I'll prepare a backport for Krypton.


arnova:
Note that I'm not sure the fix is in the right location, perhaps it should be somewhere in CVideoLibraryRefreshingJob

Rechi:
maybe adapt it to the way of #12373

arnova:
@Rechi : Thanks for the heads up. See my comment in the other PR

mkortstiege:
Issue handled in https://github.com/xbmc/xbmc/pull/12373

PR #12387, opened 2017-06-28
[settings] fix seq fault if addon use same setting in different groups

AlwinEsch:
## Description
Related to https://pastebin.com/nEa0TQBc and https://forum.kodi.tv/showthread.php?tid=298462&pid=2610026#pid2610026

Crash was from settings.xml here: https://github.com/Zomboided/service.vpn.manager/blob/master/resources/settings.xml

## Types of change

## Checklist:

### DIFF
--- a/xbmc/settings/lib/SettingsManager.cpp
+++ b/xbmc/settings/lib/SettingsManager.cpp
@@ -1248,7 +1248,9 @@ void CSettingsManager::ResolveReferenceSettings(std::shared_ptr<CSettingSection>
         else
         {
           referencedSetting = itReferencedSetting->second.setting;
-          m_settings.erase(FindSetting(referenceSetting->GetId()));
+          itReferencedSetting = FindSetting(referenceSetting->GetId());
+          if (itReferencedSetting != m_settings.end())
+            m_settings.erase(itReferencedSetting);
         }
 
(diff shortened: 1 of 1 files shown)


xbmc/settings/lib/SettingsManager.cpp
```diff
@@ -1248,7 +1248,9 @@ void CSettingsManager::ResolveReferenceSettings(std::shared_ptr<CSettingSection>
         else
         {
           referencedSetting = itReferencedSetting->second.setting;
-          m_settings.erase(FindSetting(referenceSetting->GetId()));
+          auto itReferencedSetting2 = FindSetting(referenceSetting->GetId());
```
Montellese: You could reuse the existing `itReferencedSetting` variable.

AlwinEsch: Oh yes, you are right. Becomes changed.

Is this a fix enough, or maybe another part who need change?

Montellese: This is ok for solving the crash if the same setting is used more than twice (in which case it has already been removed from `m_settings`. There are still issues however with reference settings and enable/visibility conditions which I'm trying to solve.

Montellese (approved):
Change the local variable being used and then this is good to go in. Thanks for looking into it.

PR #12394, opened 2017-06-29
Revert "[droid] packaging: don't delete pil libs from script.module.pil"

koying:
This reverts commit a59e00f619e179cdea5d205e2ec0a3ae5a7bd4d2.

The PIL libraries are alredy globbed in https://github.com/xbmc/xbmc/blob/Krypton/tools/android/packaging/Makefile.in#L135

### DIFF
--- a/tools/android/packaging/Makefile.in
+++ b/tools/android/packaging/Makefile.in
@@ -88,7 +88,7 @@ shared:
 	mkdir -p assets
 	cp -rfp $(PREFIX)/share/@APP_NAME_LC@/* ./assets
 	find `pwd`/assets/ -depth -name ".git" -exec rm -rf {} \;
-	find `pwd`/assets/ -name "*.so" -not -name "*imaging*.so" -exec rm {} \;
+	find `pwd`/assets/ -name "*.so" -exec rm {} \;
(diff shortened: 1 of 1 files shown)


wsnipex:
as long as you tested it properly, I'm fine with it.

koying:
I'll do a test build to confirm

koying:
Confirmed

image

wsnipex (approved):

Rechi:
@koying this breaks script.module.pil, see https://forum.kodi.tv/showthread.php?tid=322184

koying:
Shouldn't, unless those scripts uses non-python or non-standard dlopen.

There is an explicit python patch to prepend "lib" on android

koying:
The "./" is already suspect

Rechi:
So https://github.com/xbmc/xbmc/blob/ae91f5e/tools/depends/target/python27/python-android-binmodule.patch is missing in Krypton?

koying:
Don't know, might be. On phone...

koying:
Yep. Binmodule patch is in Jarvis and master but not in K, apparently.
We should backport https://github.com/xbmc/xbmc/commit/ae91f5e357ef59ee15c465ed3a5d7a08400df808 to K

Rechi:
If done a testbuild with the binmodule part only and will ask in forum if anyone can test it.

koying:
Great. Thanks

PR #12399, opened 2017-06-30
[utils] coverity: ignore uncaught 'fmt::FormatError' exception

Rechi:
## Description
Disable an `fmt::FormatError` uncaught exception issue at coverity.

## Motivation and Context
The windows coverity project currently shows 63 warnings because of the exception thrown in fmt.
If fmt throws an exception it should result in a crash as in that case the format string is wrong.

## How Has This Been Tested?
Next coverity after this has been merged will tell if it does what I expect.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/utils/StringUtils.h
+++ b/xbmc/utils/StringUtils.h
@@ -63,6 +63,7 @@ public:
   template<typename... Args>
   static std::string Format(const std::string& fmt, Args&&... args)
   {
+    // coverity[fun_call_w_exception : FALSE]
     auto result = fmt::format(fmt, std::forward<Args>(args)...);
     if (result == fmt)
       result = fmt::sprintf(fmt, std::forward<Args>(args)...);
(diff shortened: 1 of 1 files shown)


garbear:
Instead of hard-crashing, what about this as a workaround? #11915

When the fmt conversion happened, one of the strings had a wrong format type and it took a long time to track it down. Adding this log message instead of crashing would be helpful in that case.

FernetMenta:
A wrong format is not a runtime error. Having the app crash is much better than hiding the error.

garbear:
> A wrong format is not a runtime error. Having the app crash is much better than hiding the error.

But crash without error logging? And now a translator can maliciously or accidentally crash Kodi when a new language is installed.

AlwinEsch:
This is also a point where I must say that a "try" is helpful, after removing from the addons it makes me almost impossible to find the source caused.

Why not the "try" log the error and "throw" again?

PR #12400, opened 2017-07-01
[modernize] Prefer 'default' for declarations

hudokkow:
## Description
Prefer `= default;`over `{}` for declarations

## Motivation and Context
Modernize

## How Has This Been Tested?
Build and brief run under 17.04


hudokkow:
I'll separate VP, games and input/joysticks after everything is OK and if it's accepted.

FernetMenta:
+1 no need to separate VP. I don't expect any conflics with those changes

hudokkow:
jenkins build this please

Paxxi:
One annoying thing with this clang-tidy check is that it doesn't move the default into the header file.

Paxxi (approved):

AlwinEsch (approved):

MartijnKaijser:
@hudokkow can you rebase and merge?

garbear:
> One annoying thing with this clang-tidy check is that it doesn't move the default into the header file.

The reason for this is that if the destructor is only defined in the header, the `delete` operator doesn't know the size of the object. This is a problem for `std::unique_ptr` because the deletion occurs in the header (`std::shared_ptr` still works).

hudokkow:
ffs, jenkins build this please

Paxxi:
@garbear that doesn't sound right for the general case I'm complaining about. The destructor is still generated by the compiler for the translation unit, where the =default is specified should not be relevant to this. 
I'm more thinking that clang-tidy works on the file it's given, it doesn't go find the header and make changes to it as that wasn't specified

hudokkow:
jenkins build this please

hudokkow:
jenkins build this please

hudokkow:
Finally! Win64 failure unrelated.

PR #12402, opened 2017-07-01
[music]Fix throttling of Musicbrainz requests by music scrapers

DaveTBlake:
Musicbrainz rate-limits queries to 1 per sec, once we hit the rate-limiter the server returns 503 errors for all calls from that IP address. This results in leaving albums and artists unscraped, taking many manually initiaited re-scraping attempts to fill the gaps and much unnecessary server traffic, but also blocks the user from using Picard or any other Musicbrainz lookups. See https://musicbrainz.org/doc/XML_Web_Service/Rate_Limiting

To avoid this a 2s delay was designed into the music scraper when it was first implemented and we all thought that this was sufficiently throttling the Musicbrainz requests. Unfortunately that was not the case, and the scraping of artists and albums that do not have mbids (from tags) takes place ***without any throttling attempts at all***.

In testing at one point (when MB server traffic was heavy globally) I was getting 5 out of 6 scraped items returning a 503 error, it took 15 attempts to scrape 850 albums, there has also been user compliants about missing info and art. This is unacceptable, but it was quite a surprize to find it was mostly Kodi's fault for making unthrottled requests!

The delay was placed at the point where a known mbid was converted into a url, this is flawed as this route is not taken by all music scraping. It also does not space out the call to fetch album or artist details from Musicbrainz, from any prior request to to find that artist or album. 

The fix is to replace this delay with 1s wait after find (when mbid unknown) and another after details download.

Users without mbid tags in their music  files may notice that scraping takes longer - it will be throttling requests for the first time - but it will produce far fewer "gaps" in scraping results, and significantly reduce the need to retry scraping. Everyone should experience fewer 503 errors and more efficient scraping.

Unfortunately **as before** this still slows down (xml based) **music** scrapers that do not use Musicbrainz too e.g.TADB scraper. It will have not have an effect on any other kind of scrapers, nor the Python music scrapers when they are available.  It also does not limit any multiple MB requests from within the xml chain while processing an album e.g. getting release group details for the link to allmusic.com immediately after requesting the release details.

A better, and more involved, solution could be to make the throttling part of the scraper parameters that can be set from the scraper xml. However I am unclear how much effort it is worth putting into xml scraper support when in the future we will be converting to Python based scraper addons. This fix is a worthwhile step.

This could usefully be backported to Krypton too.

### DIFF
--- a/xbmc/music/infoscanner/MusicInfoScanner.cpp
+++ b/xbmc/music/infoscanner/MusicInfoScanner.cpp
@@ -1160,6 +1160,14 @@ INFO_RET CMusicInfoScanner::DownloadAlbumInfo(const CAlbum& album, const ADDON::
       }
       Sleep(1);
     }
+    /*
+    Finding album using xml scraper may request data from Musicbrainz.
+    MusicBrainz rate-limits queries to 1 per sec, once we hit the rate-limiter the server
+    returns 503 errors for all calls from that IP address.
+    To stay below the rate-limit threshold wait 1s before proceeding
+    */
+    if (!info->IsPython())
+      Sleep(1000);
   }
 
   CGUIDialogSelect *pDlg = NULL;
@@ -1277,9 +1285,17 @@ INFO_RET CMusicInfoScanner::DownloadAlbumInfo(const CAlbum& album, const ADDON::
     }
     Sleep(1);
   }
-
   if (!scraper.Succeeded())
     return INFO_ERROR;
+  /*
+  Fetching album details using xml scraper may makes requests for data from Musicbrainz.
+  MusicBrainz rate-limits queries to 1 per sec, once we hit the rate-limiter the server
+  returns 503 errors for all calls from that IP address.
+  To stay below the rate-limit threshold wait 1s before proceeding incase next action is
+  to scrape another album or artist
+  */
+  if (!info->IsPython())
+    Sleep(1000);
 
   albumInfo = scraper.GetAlbum(iSelectedAlbum);
   
@@ -1375,6 +1391,14 @@ INFO_RET CMusicInfoScanner::DownloadArtistInfo(const CArtist& artist, const ADDO
       }
       Sleep(1);
     }
+    /*
+    Finding artist using xml scraper makes a request for data from Musicbrainz.
+    MusicBrainz rate-limits queries to 1 per sec, once we hit the rate-limiter
+    the server returns 503 errors for all calls from that IP address. To stay 
+    below the rate-limit threshold wait 1s before proceeding
+    */
+    if (!info->IsPython())
+      Sleep(1000); 
   }
 
   int iSelectedArtist = 0;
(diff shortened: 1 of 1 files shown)


DaveTBlake:
Of course I have tested extensively with my music library, but some other runs will be useful. For best effect I recommend installing the lastest Universal Album Scraper v2.7.2 from the repo (that also halves the number of mb requests). However there are other music scraper changes comming for v18 (see #12120 a first step), which will need more substantial tests. 

I will create a test build that has both PRs if that will be useful?

xbmc/music/infoscanner/MusicInfoScanner.cpp
```diff
@@ -1440,6 +1464,15 @@ INFO_RET CMusicInfoScanner::DownloadArtistInfo(const CArtist& artist, const ADDO
     else
       return INFO_NOT_FOUND;
   }
+  /*
+  Fetching artist detais using xml scraper makes requests for data from Musicbrainz.
```
AchimTuran: detais=details

AchimTuran:
Well there is not much to say about your PR. If it works you should try to merge it :-) also make sure that it doesn't break other scrapers.

AchimTuran:
> However I am unclear how much effort it is worth putting into xml scraper support when in the future we will be converting to Python based scraper addons. This fix is a worthwhile step.

If that is the case, I think it would not be very useful to put more effort into it. Better put more effort into implementing Python scrapers then XML. :-)

xbmc/music/infoscanner/MusicInfoScanner.cpp
```diff
@@ -1160,6 +1160,14 @@ INFO_RET CMusicInfoScanner::DownloadAlbumInfo(const CAlbum& album, const ADDON::
       }
       Sleep(1);
     }
+    /*
+    Finding album using xml scraper may request data from Musicbrainz.
+    MusicBrainz rate-limits queries to 1 per sec, once we hit the rate-limiter the server
+    returns 503 errors for all calls from that IP address.
+    To stay below the rate-limit threshold wait 1s before proceeding
+    */
+    if (!info->IsPython())
```
Razzeee: Can't we invert this filter? Might make more sense to check if it's an xml scraper? We might potentially get more scrapers then python in the future ;)

DaveTBlake: Sorry no, m_isPython is the only flag CScraper has, and IsPython the only method of this kind. 

If we do add support of other kinds of scraper a lot more would need to change anyway, a refactoring of scraping would be sensible

zag2me:
Its a shame this effects the TADB scraper, as that has no rate limiter and was very fast. But this is a correct fix as I have seen the problems with the universal scraper as well in the past and i'm sure most people just use that as its default.

DaveTBlake:
@zag2me yes I know. Of course for music tagged with Musicbrainz ids _**all**_ the xml music scrapers were throttled to 1 album or artist per **2** seconds by the flawed attempt to avoid hitting the Musicbrainz rate limiter anyway. Although the TADB scraper does not need throttling, this fix does n't make the situation worse for that scraper, it just doen't improve it either.

Since the future lies in Python scrapers that can do their own rate limiting (or not), I think it is best to focus effort there

DaveTBlake:
Received some user testing that went well, and I have tested as best I can, so time to merge

PR #12386, opened 2017-06-28
Fix crash during close of kodi if vfs and pvr addons are present.

AlwinEsch:
## Description

Error was coming with https://github.com/xbmc/xbmc/pull/12379. During my test there was no PVR addon active and seen it now. This make the m_vfsAddonCache reset short before the manager, then are the related addons, normally closed.

Related to:
```
Thread 1 "kodi.bin" received signal SIGSEGV, Segmentation fault.

#0  __GI___pthread_mutex_lock (mutex=0x8) at ../nptl/pthread_mutex_lock.c:67
#1  0x0000555d3ee0e0ae in (anonymous namespace)::(anonymous namespace)::RecursiveMutex::lock (this=0x8)
    at /home/alwin/Development/kodi/xbmc/threads/platform/pthreads/CriticalSection.h:49
#2  0x0000555d3ee0f1da in (anonymous namespace)::CountingLockable<XbmcThreads::pthreads::RecursiveMutex>::lock (this=0x8)
    at /home/alwin/Development/kodi/xbmc/threads/Lockables.h:60
#3  0x0000555d3ee0f154 in (anonymous namespace)::UniqueLock<CCriticalSection>::UniqueLock (this=0x7ffe8296b2f0, lockable=...)
    at /home/alwin/Development/kodi/xbmc/threads/Lockables.h:127
#4  0x0000555d3ee0e0f1 in CSingleLock::CSingleLock (this=0x7ffe8296b2f0, cs=...)
    at /home/alwin/Development/kodi/xbmc/threads/SingleLock.h:38
#5  0x0000555d3f752245 in (anonymous namespace)::CVFSAddonCache::GetAddonInstances (this=0x0)
    at /home/alwin/Development/kodi/xbmc/addons/VFSEntry.cpp:49
#6  0x0000555d3fcdd531 in (anonymous namespace)::CFileDirectoryFactory::Create (url=..., pItem=0x7ffe8296bb90, strMask=...)
    at /home/alwin/Development/kodi/xbmc/filesystem/FileDirectoryFactory.cpp:88
#7  0x0000555d3fccad15 in (anonymous namespace)::CDirectoryFactory::Create (url=...)
    at /home/alwin/Development/kodi/xbmc/filesystem/DirectoryFactory.cpp:110
#8  0x0000555d3fcc90f1 in (anonymous namespace)::CDirectory::Exists (url=..., bUseCache=true)
    at /home/alwin/Development/kodi/xbmc/filesystem/Directory.cpp:308
[...]

PR #12404, opened 2017-07-01
ServiceManager: Fix order until we hit the wall again

fritsch:
Filesystem implementation wants to check VFS, but VFS might be a binaddon now, vfs tries to ask bin addon cache but this is already gone.

### DIFF
--- a/xbmc/ServiceManager.cpp
+++ b/xbmc/ServiceManager.cpp
@@ -151,11 +151,11 @@ void CServiceManager::Deinit()
   m_peripherals.reset();
   m_contextMenuManager.reset();
   m_favouritesService.reset();
-  m_vfsAddonCache.reset();
-  m_binaryAddonCache.reset();
   if (m_PVRManager)
     m_PVRManager->Deinit();
   m_PVRManager.reset();
+  m_vfsAddonCache.reset();
+  m_binaryAddonCache.reset();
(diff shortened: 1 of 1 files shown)


ace20022:
You could manage them with a stack for example.

fritsch:
Yes. That's a nice idea using a LIFO to manage the order of destructing the objects. It saves us topological sorting afterwards.

This code is currently heavily in development (not by me) - I wanted to make sure master shutsdown correctly with given design at hand.

Montellese:
I've written a very basic implementation of topological sorting if we need it: https://github.com/Montellese/xbmc/commit/370bf86f5b10ef45d8ea8f6155471ecf2e74cff0. It could be updated with some C++11 magic.

fritsch:
The good thing about the lifo stack is, that by making sure everything is in order when enqueing, the correct order is already done. Those elements without input are at the end of the list, those that others depend on are more in front. E.g. position 2 is not dependend on position 3, but position 3 can depend on 2 and 1. That means destructing those objects is a linear loop popping of the stack values.

Cycles won't obviously work.

da-anda:
is this sort of obsolete now after #12386 got merged?

AlwinEsch:
Yes there was the same done to prevent in future and also removed the causer of them and done now on CBinaryAddonManger instead of CAddonDll.

PR #12405, opened 2017-07-02
Explicitly initialize VIS_INFO structure; may be bypassed

djp952:
## Description
The initialization of the VIS_INFO structure in CGUIVisualisationControl::CreateBuffers() may be bypassed, this causes a runtime error when debugging through Visual Studio.

## Motivation and Context
Running a DEBUG build of the master branch through Visual Studio stops frequently at this location due to the VIS_INFO structure members being accessed prior to initialization when the member variable m_alreadyStarted is set to false.  This happens every time a PVR Live TV channel is opened, for example.

## How Has This Been Tested?
I tested this change on Windows (Win32) using a local build of the master branch.  The resultant binaries (.dll, .pdb) were copied over a recently installed nightly build.  The only non-default add-on installed was my own PVR, and the only non-default settings are PVR related.  I stepped through the code via a breakpoint set by Visual Studio while executing tests against a custom PVR client and opening channels.  No adverse side-effects have been noted after the change.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/guilib/GUIVisualisationControl.cpp
+++ b/xbmc/guilib/GUIVisualisationControl.cpp
@@ -442,7 +442,7 @@ void CGUIVisualisationControl::CreateBuffers()
   ClearBuffers();
 
   // Get the number of buffers from the current vis
-  VIS_INFO info;
+  VIS_INFO info { false, 0 };
 
   if (m_instance && m_alreadyStarted)
     m_instance->GetInfo(&info);


Paxxi:
It would be a better idea to handle this inside of VIS_INFO.
Simplest would be to do
```
struct VIS_INFO /* internal */
{
  bool bWantsFreq{false};
  int iSyncDelay{0};
};
```

but using a constructor is also an option. Having this handled outside of VIS_INFO means it's easy to miss it like what's happened here.

AlwinEsch:
@Paxxi The parts there with internal should stay complete in "C". 

For the case that it is required in other languages completely as independent "C" header.

AlwinEsch (approved):

AlwinEsch:
Thanks a lot @djp952!

PR #12406, opened 2017-07-02
[Backport] [Music]Fix throttling of Musicbrainz requests

DaveTBlake:
Backport of #12402  (see discussion and testing there)

To avoid hitting the Musicbrainz 1 query per sec rate-limiter (503 errors for all calls from that IP address) a 2s delay was designed into the music scraper when it was first implemented, and we all thought that this was sufficiently throttling the Musicbrainz requests. Unfortunately that was not the case, and the scraping of artists and albums that do not have mbids (from tags) takes place **_without any throttling_** attempts at all.

The fix is to replace this delay with 1s wait after find (when mbid unknown) and another after details download.

Users without mbid tags in their music files may notice that scraping takes longer - it will be throttling requests for the first time - but it will produce far fewer "gaps" in scraping results, and significantly reduce the need to retry scraping. Everyone should experience fewer 503 errors and more efficient scraping.

As before this still slows down music scrapers that do not use Musicbrainz too e.g.TADB scraper. But it will have not have an effect on any other kind of scrapers. Unfortunately it does not limit any multiple MB requests from within the xml chain while processing an album e.g. getting release group details for the link to allmusic.com immediately after requesting the release details.

### DIFF
--- a/xbmc/music/infoscanner/MusicInfoScanner.cpp
+++ b/xbmc/music/infoscanner/MusicInfoScanner.cpp
@@ -1157,6 +1157,13 @@ INFO_RET CMusicInfoScanner::DownloadAlbumInfo(const CAlbum& album, const ADDON::
       }
       Sleep(1);
     }
+    /*
+    Finding album may request data from Musicbrainz.
+    MusicBrainz rate-limits queries to 1 per sec, once we hit the rate-limiter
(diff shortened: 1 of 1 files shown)


MartijnKaijser:
jenkins build this please

PR #12409, opened 2017-07-02
CVideoPlayer: cannot pause if seek is not supported

djp952:
## Description
CVideoPlayer does not allow for a pause operation to occur if seek is not supported.  The call to SetPlaySpeed will be specifically bypassed by a test to CanSeek().

## Motivation and Context
While developing a PVR client I had noted a while back that if the client indicates it supports pausing but it does not support seeking, the player UI control for Pause was present but inoperable.  I ultimately tracked down the problem to CVideoPlayer::SetSpeed disallowing the operation by testing CanSeek().  

The proposed modification checks both CanSeek() and CanPause() up front and returns early if both are false.  Subsequently if CanSeek() is false, implying that CanPause() is true, and the request is for a pause (DVD_PLAYSPEED_PAUSE) or resume (DVD_PLAYSPEED_NORMAL) operation it will be allowed to go through.

## How Has This Been Tested?
I tested this change on Windows (Win32) using a local build of the master branch.  The resultant binaries (.dll, .pdb) were copied over a recently installed nightly build.  I confirmed the change resolves the specific problem I'm trying to resolve with the PVR Client.

Additionally, I temporarily hard-coded CanSeek() to return false and tested the player with a number of different playback mediums - DVD, Blu-Ray, MP4, WMV and M2TS.  Each media type was paused and resumed successfully while disallowing seek and playback speed changes as expected.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -3590,14 +3590,14 @@ void CVideoPlayer::SetSpeed(float speed)
   if (speed < 0 && IsInMenu())
     return;
  
-  if(!CanSeek() && !CanPause())
+  if (!CanSeek() && !CanPause())
     return;
   
   int iSpeed = static_cast<int>(speed * DVD_PLAYSPEED_NORMAL);
 
-  if(!CanSeek())
+  if (!CanSeek())
   {
-    if((iSpeed != DVD_PLAYSPEED_NORMAL) && (iSpeed != DVD_PLAYSPEED_PAUSE))
+    if ((iSpeed != DVD_PLAYSPEED_NORMAL) && (iSpeed != DVD_PLAYSPEED_PAUSE))
       return;
   }
 
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -3589,10 +3589,10 @@ void CVideoPlayer::SetSpeed(float speed)
   // forward is fine
   if (speed < 0 && IsInMenu())
     return;
- 
+
   if (!CanSeek() && !CanPause())
     return;
-  
+
   int iSpeed = static_cast<int>(speed * DVD_PLAYSPEED_NORMAL);
 
   if (!CanSeek())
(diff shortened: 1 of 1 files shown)


xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3589,12 +3589,18 @@ void CVideoPlayer::SetSpeed(float speed)
   // forward is fine
   if (speed < 0 && IsInMenu())
     return;
-
-  if (!CanSeek())
+ 
+  if(!CanSeek() && !CanPause())
```
ksooo: Meh. Please nclude a space between "if" and "(".

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3589,12 +3589,18 @@ void CVideoPlayer::SetSpeed(float speed)
...
     return;
   
   int iSpeed = static_cast<int>(speed * DVD_PLAYSPEED_NORMAL);
 
+  if(!CanSeek())
```
ksooo: Please nclude a space between "if" and "(".

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3589,12 +3589,18 @@ void CVideoPlayer::SetSpeed(float speed)
...
+  {
+    if((iSpeed != DVD_PLAYSPEED_NORMAL) && (iSpeed != DVD_PLAYSPEED_PAUSE))
```
ksooo: Please nclude a space between "if" and "(".

djp952:
Spaces added

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3589,12 +3589,18 @@ void CVideoPlayer::SetSpeed(float speed)
   // forward is fine
   if (speed < 0 && IsInMenu())
     return;
-
-  if (!CanSeek())
+
```
hudokkow: stray space.

Thanks for your contribution. ping @FernetMenta

djp952:
Stray spaces eliminated like the rabid dogs that they are.

FernetMenta:
thanks. build errors unrelated

FernetMenta (commented):
+1

ksooo:
Maybe next time we should at least take our time to squash into a single commit before we merge... This is, how we usually work - eliminate temporary commits, then merge.

FernetMenta:
oops, missed that. sorry

djp952:
Thanks ksooo, I have practiced squashing commits offline so I'll get that right next time.  'preciate it!

PR #12416, opened 2017-07-03
[PVR][guiinfo][Estuary] PVR info dialog, recordings window: cleanup and extend

ksooo:
guiinfo: Add LISTITEM_EXPIRATION_DATE and LISTITEM_EXPIRATION_TIME 
Estuary: PVR info dialog & recordings window: Reorder some items and add expriration date/time for recordings:

PVR info dialog (before):
screenshot000

PVR info dialog (after):
screenshot003

Recordings window (before):
screenshot001

Recordings window (after):
screenshot004

This has been runtime tested on macOS, latest master.

@ronie, phil65 fyi

### DIFF
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -3660,7 +3660,7 @@ const infomap container_str[]  = {{ "property",         CONTAINER_PROPERTY },
 ///   \table_row3{   <b>`ListItem.EpgEventTitle`</b>,
 ///                  \anchor ListItem_EpgEventTitle
 ///                  _string_,
-///     Returns the title of the epg event associated with the item, if any
+///     Returns the title of the epg event associated with the item\, if any
 ///   }
 ///   \table_row3{   <b>`ListItem.InProgress`</b>,
 ///                  \anchor ListItem_InProgress
@@ -3731,6 +3731,16 @@ const infomap container_str[]  = {{ "property",         CONTAINER_PROPERTY },
 ///                  \anchor ListItem_AddonOrigin
 ///                  _string_,
 ///     Name of the repository the add-on originates from.
+///   }
+///   \table_row3{   <b>`ListItem.ExpirationDate`</b>,
+///                  \anchor ListItem_ExpirationDate
+///                  _string_,
+///     Expiration date of the selected item in a container\, empty string if not supported
+///   }
+///   \table_row3{   <b>`ListItem.ExpirationTime`</b>,
+///                  \anchor ListItem_ExpirationTime
+///                  _string_,
+///     Expiration time of the selected item in a container\, empty string if not supported
 ///   }
 /// \table_end
 ///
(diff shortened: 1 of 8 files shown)


xbmc/GUIInfoManager.cpp
```diff
@@ -3732,6 +3732,16 @@ const infomap container_str[]  = {{ "property",         CONTAINER_PROPERTY },
...
+///   \table_row3{   <b>`ListItem.ExpirationDate`</b>,
+///                  \anchor ListItem_ExpirationDate
+///                  _string_,
+///     Expiration date of the selected item in a container, empty string if not supported
+///   }
+///   \table_row3{   <b>`ListItem.ExpirationTime`</b>,
+///                  \anchor ListItem_ExpirationTime
+///                  _string_,
+///     Expiration time of the selected item in a container, empty string if not supported
+///   }
```
AlwinEsch: You need to add a **`\`** before **`,`** inside text lines! Also seen on `ListItem.EpgEventTitle` missing (may be also somewhere else).

```
Expiration time of the selected item in a container\, empty string if not supported
```

Thats a stupid bad thing on doxygen by use of my table style.

AlwinEsch (commented):
How it is shown if addon not support expire date for recordings?

xbmc/GUIInfoManager.cpp
ksooo: okay, will change

da-anda:
could you move the "genre" and "expires" info to a second column in the PVR info dialog? Currently there is so much lost space on the right handside of it. Alternatively, have you tried moving the plot on the right? These are just two quick ideas. I also just thought about moving the "meta data" below the thumbnail. But this dialog might in general need a bit more love from our skinners. @phil65 ?

ksooo:
I agree, but due to lack of knowledge on my end I leave this to our skinners, in a follow-up PR.

xbmc/GUIInfoManager.cpp
ksooo: done

ksooo:
> How it is shown if addon not support expire date for recordings?

 :-) The "Expiration: foo" line is not present, then.

AlwinEsch (approved):

da-anda:
@ksooo fair point. I hope one of our skinners will give this dialog some love. Btw, is the channel info below the recording title really important? I think the episode title ("Steg mit Aussicht") might be better suited there, and the channel name moved down to the meta data. Just an idea and no hard feelings.

ksooo:
jenkins build this please

ksooo:
@da-anda better now?

screenshot001

da-anda:
yep, better IMO. But I would have done it like "Sender: Das Erste HD" to match the other meta data. Up to you if you want to adjust it that way. And thanks much for maintaining and improving PVR.

ksooo:
I had it exactly like you suggested, but channel name imo is special and important; it needs to be highlighted from other less important properties.

da-anda:
> but channel name imo is special and important; it needs to be highlighted from other less important properties

is it? That's probably subject of personal preferences and recording habits. I primarily care for the show I recorded and not on which channel it was on. But again, no hard feelings, fine with whatever you prefer.

Btw, because it just came to my mind. Might be nice to also have a "delete" button in the OSD as well as "find similar" and maybe even "edit timer" (in case it's from a recurring timer rule, if we store that info somewhere). Just some ideas for the future, as the "play" button is kinda lost down there.

ksooo:
@phil65 are the skin changes okay?

phil65:
yup, looks fine.

phil65 (approved):

ksooo:
jenkins build this please

ksooo:
jenkins errors are unrelated

PR #12417, opened 2017-07-03
AddonVideoCodec: Switch to using CVideoBufferManager

popcornmix:
## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/AddonVideoCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/AddonVideoCodec.cpp
@@ -31,70 +31,12 @@ using namespace kodi::addon;
 
 #define ALIGN(value, alignment) (((value)+(alignment-1))&~(alignment-1))
 
-class BufferPool
-{
-public:
-  ~BufferPool()
-  {
-    for (BUFFER &buf : freeBuffer)
-      free(buf.mem);
-    for (BUFFER &buf : usedBuffer)
-      free(buf.mem);
-  }
-
-  bool GetBuffer(VIDEOCODEC_PICTURE &picture)
-  {
-    if (freeBuffer.empty())
-      freeBuffer.resize(1);
-
-    BUFFER &buf(freeBuffer.back());
-    if (buf.memSize < picture.decodedDataSize)
-    {
-      buf.memSize = picture.decodedDataSize;
-      buf.mem = malloc(buf.memSize);
-      if (buf.mem == nullptr)
-      {
-        buf.memSize = 0;
-        picture.decodedData = nullptr;
-        return false;
-      }
-    }
-    picture.decodedData = (uint8_t*)buf.mem;
-
-    usedBuffer.push_back(buf);
-    freeBuffer.pop_back();
-    return true;
-  }
-
-  void ReleaseBuffer(void *bufferPtr)
-  {
-    std::vector<BUFFER>::iterator res(std::find(usedBuffer.begin(), usedBuffer.end(), bufferPtr));
-    if (res == usedBuffer.end())
-      return;
-    freeBuffer.push_back(*res);
-    usedBuffer.erase(res);
-  }
-
-private:
-  struct BUFFER
-  {
-    BUFFER():mem(nullptr), memSize(0) {};
-    BUFFER(void* m, size_t s) :mem(m), memSize(s) {};
-    bool operator == (const void* data) const { return mem == data; };
-
-    void *mem;
-    size_t memSize;
-  };
-  std::vector<BUFFER> freeBuffer, usedBuffer;
-};
-
 CAddonVideoCodec::CAddonVideoCodec(CProcessInfo &processInfo, ADDON::BinaryAddonBasePtr& addonInfo, kodi::addon::IAddonInstance* parentInstance)
   : CDVDVideoCodec(processInfo),
     IAddonInstanceHandler(ADDON_INSTANCE_VIDEOCODEC, addonInfo, parentInstance)
   , m_codecFlags(0)
   , m_displayAspect(0.0f)
   , m_lastPictureBuffer(nullptr)
-  , m_bufferPool(new BufferPool())
 {
   m_struct = { { 0 } };
   m_struct.toKodi.kodiInstance = this;
@@ -111,9 +53,10 @@ CAddonVideoCodec::~CAddonVideoCodec()
 {
   DestroyInstance();
 
-  m_bufferPool->ReleaseBuffer(m_lastPictureBuffer);
-
-  delete m_bufferPool;
+  CVideoBuffer *videoBuffer = m_map[m_lastPictureBuffer];
+  if (videoBuffer)
+    videoBuffer->Release();
+  m_map.erase(m_lastPictureBuffer);
 }
 
 bool CAddonVideoCodec::CopyToInitData(VIDEOCODEC_INITDATA &initData, CDVDStreamInfo &hints)
@@ -242,6 +185,7 @@ bool CAddonVideoCodec::AddData(const DemuxPacket &packet)
 
 CDVDVideoCodec::VCReturn CAddonVideoCodec::GetPicture(VideoPicture* pVideoPicture)
 {
+  CVideoBuffer *videoBuffer;
   if (!m_struct.toAddon.get_picture)
     return CDVDVideoCodec::VC_ERROR;
 
@@ -257,22 +201,23 @@ CDVDVideoCodec::VCReturn CAddonVideoCodec::GetPicture(VideoPicture* pVideoPictur
   case VIDEOCODEC_RETVAL::VC_BUFFER:
     return CDVDVideoCodec::VC_BUFFER;
   case VIDEOCODEC_RETVAL::VC_PICTURE:
-//    pVideoPicture->data[0] = picture.decodedData + picture.planeOffsets[0];
-//    pVideoPicture->data[1] = picture.decodedData + picture.planeOffsets[1];
-//    pVideoPicture->data[2] = picture.decodedData + picture.planeOffsets[2];
-//    pVideoPicture->iLineSize[0] = picture.stride[0];
-//    pVideoPicture->iLineSize[1] = picture.stride[1];
-//    pVideoPicture->iLineSize[2] = picture.stride[2];
-//    pVideoPicture->iWidth = picture.width;
-//    pVideoPicture->iHeight = picture.height;
-//    pVideoPicture->pts = DVD_NOPTS_VALUE;
-//    pVideoPicture->dts = DVD_NOPTS_VALUE;
-//    pVideoPicture->color_range = 0;
-//    pVideoPicture->color_matrix = 4;
-//    pVideoPicture->iFlags = DVP_FLAG_ALLOCATED;
-//    if (m_codecFlags & DVD_CODEC_CTRL_DROP)
-//      pVideoPicture->iFlags |= DVP_FLAG_DROPPED;
-//    pVideoPicture->format = RENDER_FMT_YUV420P;
+    pVideoPicture->iWidth = picture.width;
+    pVideoPicture->iHeight = picture.height;
+    pVideoPicture->pts = picture.pts;
+    pVideoPicture->dts = DVD_NOPTS_VALUE;
+    pVideoPicture->color_range = 0;
+    pVideoPicture->color_matrix = 4;
+    pVideoPicture->iFlags = 0;
+    if (m_codecFlags & DVD_CODEC_CTRL_DROP)
+      pVideoPicture->iFlags |= DVP_FLAG_DROPPED;
+
+    if (pVideoPicture->videoBuffer)
+      pVideoPicture->videoBuffer->Release();
+
+    pVideoPicture->videoBuffer = m_map[picture.decodedData];
+    assert(pVideoPicture->videoBuffer);
+    pVideoPicture->videoBuffer->Acquire();
+    pVideoPicture->videoBuffer->SetDimensions(picture.width, picture.height, picture.stride[0], picture.height);
 
     pVideoPicture->iDisplayWidth = pVideoPicture->iWidth;
     pVideoPicture->iDisplayHeight = pVideoPicture->iHeight;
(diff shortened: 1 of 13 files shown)


xbmc/cores/VideoPlayer/Process/overrides/rbpi/ProcessInfoPi.h
```diff
@@ -22,13 +22,36 @@
 #include "cores/IPlayer.h"
 #include "../../ProcessInfo.h"
 
+namespace MMAL {
+  class CMMALYUVBuffer;
+};
```
AlwinEsch: Ah `;` is not needed on namespaces

popcornmix:
Ping @FernetMenta @peak3d 
This is a RFC rather than something to merge immediately.
This gets AddonVideoCodec working with MMAL buffers in a platforms independent way.
Other platforms will need similar adjustments.

AddonVideoCodec initially provides just the buffer size desired and the dimensions are only available later which required splitting the Configure API.

Annoyingly it uses the pixel data pointer as an identifier in the API, rather than a general user data pointer, so the map between pixel data pointer and videoBuffer is required.

The whole Configure API and CVideoBufferManager needs more thought when video resolution can change on the fly.

FernetMenta:
> The whole Configure API and CVideoBufferManager needs more thought when video resolution can change on the fly.

@popcornmix a change in resolution triggers a streamchange event in VP. VPV will open the new stream and call CVideoBufferManager::ReleasePools. The will call Released() on the pool. Implementations can use this callback and register a new pool if required.

xbmc/cores/VideoPlayer/DVDCodecs/Video/AddonVideoCodec.cpp
```diff
@@ -330,19 +289,38 @@ void CAddonVideoCodec::Reset()
...
+  m_map.erase(m_lastPictureBuffer);
   m_lastPictureBuffer = nullptr;
 
   m_struct.toAddon.reset(&m_struct);
 }
 
 bool CAddonVideoCodec::GetFrameBuffer(VIDEOCODEC_PICTURE &picture)
 {
-  return m_bufferPool->GetBuffer(picture);
+  CVideoBuffer *videoBuffer = m_videoBufferManager.Get(AV_PIX_FMT_YUV420P, picture.decodedDataSize);
```
FernetMenta: what is decodedDataSize and why is this needed?

FernetMenta: I see, why don't you drop Get(AVPixelFormat format, int width, int height)? clieant should call Get(fmt, width * height) instead

xbmc/cores/VideoPlayer/DVDCodecs/Video/AddonVideoCodec.cpp
```diff
@@ -263,16 +208,24 @@ CDVDVideoCodec::VCReturn CAddonVideoCodec::GetPicture(VideoPicture* pVideoPictur
...
+    pVideoPicture->iHeight = picture.height;
+    pVideoPicture->pts = picture.pts;
+    pVideoPicture->dts = DVD_NOPTS_VALUE;
+    pVideoPicture->color_range = 0;
+    pVideoPicture->color_matrix = 4;
+    pVideoPicture->iFlags = 0;
+    if (m_codecFlags & DVD_CODEC_CTRL_DROP)
+      pVideoPicture->iFlags |= DVP_FLAG_DROPPED;
+
+    m_videoBufferManager.Configure(AV_PIX_FMT_YUV420P, picture.width, picture.height, picture.stride[0], picture.height, picture.decodedDataSize);
```
FernetMenta: clients are not supposed to call internals. clients just request a buffer, the rest is handled internally

FernetMenta: see here https://github.com/FernetMenta/kodi-agile/blob/master/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoPPFFmpeg.cpp#L122

xbmc/cores/VideoPlayer/Process/VideoBuffer.cpp
```diff
@@ -318,14 +318,21 @@ void CVideoBufferPoolSysMem::Return(int id)
   m_free.push_back(id);
 }
 
-void CVideoBufferPoolSysMem::Configure(AVPixelFormat format, int width, int height)
+void CVideoBufferPoolSysMem::Configure(AVPixelFormat format, int width, int height, int aligned_width, int aligned_height, int size)
```
FernetMenta: aligned_width looks like dead code. why is this needed.
btw. we use camelCase here, no "_"

xbmc/cores/VideoPlayer/DVDCodecs/Video/AddonVideoCodec.cpp
popcornmix: Will do. 
It's a bit more complicated - YUV has a 3/2 factor,
and you really need to use the aligned width/height.
ffmpeg when decoding H.264 1280x720 will actually require a buffer of 1536x722 due to alignment requirements of various bits of simd code.
But the client that uses the buffer will need to know about that.

xbmc/cores/VideoPlayer/Process/VideoBuffer.cpp
popcornmix: aligned width / height are used in the Pi implementation, as that information gets passed back to renderer.
But lets see if I can just use the size on Configure, and work out the aligned_width/aligned_height from GetStride/GetPlanes in the renderer.

popcornmix: Ah - iLineSize/data has gone from VideoPicture, so can't use that for renderer to work out what the aligned width and height are.
So I need the CVideoBuffer (or mmal derived class) to contain this information and some way for AddonVideoCodec to pass that in. Configuring the pool (which the CVideoBuffer has access to) seems like the obvious way.

peak3d:
There should not be a special CVideoBuffer for AddonVideoCodec
AddonVideoCodec should use the existing Buffer(pools) provided by the platform.

xbmc/cores/VideoPlayer/Process/VideoBuffer.cpp
FernetMenta: ok, then we shoud add stride to Get()

popcornmix: Unfortunately that doesn't work for AddonVideoCodec.
It only provides the buffer size and not the width/height (or stride) when Get is called.
You only see the width/height (and stride) when the decoded picture is returned.

FernetMenta: the client needs a buffer with a set of specific properties. the should be parameters of Get

popcornmix: @peak3d - is it possible to provide width/height/stride at point Get is called (CAddonVideoCodec::GetFrameBuffer)?

FernetMenta: they are parameters of picture (VIDEOCODEC_PICTURE)

FernetMenta: confused ... AddonVideoCodec as a client does not define stride, the buffer pool does

xbmc/cores/VideoPlayer/DVDCodecs/Video/AddonVideoCodec.cpp
FernetMenta: that logic here is wrong btw. at the time the codec outputs a picture is has already requested a buffer and configuration is already defined.

popcornmix: This is the first point in time that picture.stride[0] is filled in.
It is undefined when GetBuffer is called (https://github.com/xbmc/xbmc/pull/12417/commits/ceced4c3f34a4ac04f8392978f0d63e967853c47#diff-7c4d93a31a66a47a05f4bda3208d9edbR310)
@peak3d will have to explain if there is any way round this.
I did initially request getting the width/height/stride when GetBuffer was called but that was not possible with the API widevine provides.

FernetMenta: hmm, in general stride is defined by the buffer and the decoder honors this when writing to the buffer rather than the other way round.

peak3d:
libwidevine calls "allocate" callback with only the size needed.
This is done before any picture is produced and therefore there is no way to know currently what will be done with the block of memory.

xbmc/cores/VideoPlayer/DVDCodecs/Video/AddonVideoCodec.cpp
FernetMenta: check out ffmpeg: https://github.com/FFmpeg/FFmpeg/blob/cdd3048134b8b032b69d3c076d73d24e329dc9ba/libavfilter/framepool.c#L51
linesize is set by the pool an honored by decoders. this is why CVideoBuffer provides a method GetStrides and nothing to set those

FernetMenta:
@peak3d as mentioned, stride is defined by the buffer but if this is not possible with widevine we have to add SetStrides to CVideoBuffer

peak3d:
@FernetMenta the decoder is NOT designed by me so I have no influence on it.

popcornmix:
Updated to use "size" as the primary configuration option and SetDimensions as a way of filling in dimensions after allocation.

xbmc/cores/VideoPlayer/DVDCodecs/Video/AddonVideoCodec.cpp
```diff
@@ -263,16 +208,24 @@ CDVDVideoCodec::VCReturn CAddonVideoCodec::GetPicture(VideoPicture* pVideoPictur
...
+    pVideoPicture->iHeight = picture.height;
+    pVideoPicture->pts = picture.pts;
+    pVideoPicture->dts = DVD_NOPTS_VALUE;
+    pVideoPicture->color_range = 0;
+    pVideoPicture->color_matrix = 4;
+    pVideoPicture->iFlags = 0;
+    if (m_codecFlags & DVD_CODEC_CTRL_DROP)
+      pVideoPicture->iFlags |= DVP_FLAG_DROPPED;
+
+    m_videoBufferManager.SetDimensions(picture.width, picture.height, picture.stride[0], picture.height);
```
FernetMenta: this should be called on the buffer, not the bufferManager

xbmc/cores/VideoPlayer/Process/VideoBuffer.h
```diff
@@ -161,6 +169,8 @@ class CVideoBufferManager
   void RegisterPool(std::shared_ptr<IVideoBufferPool> pool);
   void ReleasePools();
   CVideoBuffer* Get(AVPixelFormat format, int width, int height);
+  CVideoBuffer* Get(AVPixelFormat format, int size);
```
FernetMenta: the other Get method is not needed anymore

xbmc/cores/VideoPlayer/DVDCodecs/Video/AddonVideoCodec.h
```diff
@@ -62,5 +62,6 @@ class CAddonVideoCodec
 
   void * m_lastPictureBuffer;
 
-  BufferPool *m_bufferPool;
+  std::map<void *,CVideoBuffer *> m_map;
+  CVideoBufferManager m_videoBufferManager;
```
FernetMenta: a codec is not supposed to own an instance of a buffermanager

xbmc/cores/VideoPlayer/DVDCodecs/Video/AddonVideoCodec.cpp
```diff
@@ -31,70 +31,13 @@ using namespace kodi::addon;
...
-};
-
 CAddonVideoCodec::CAddonVideoCodec(CProcessInfo &processInfo, ADDON::BinaryAddonBasePtr& addonInfo, kodi::addon::IAddonInstance* parentInstance)
   : CDVDVideoCodec(processInfo),
     IAddonInstanceHandler(ADDON_INSTANCE_VIDEOCODEC, addonInfo, parentInstance)
   , m_codecFlags(0)
   , m_displayAspect(0.0f)
   , m_lastPictureBuffer(nullptr)
-  , m_bufferPool(new BufferPool())
+  , m_videoBufferManager(processInfo.GetVideoBufferManager())
```
FernetMenta: see above, manager is owned by ProcessIndo

xbmc/cores/VideoPlayer/Process/ProcessInfo.h
```diff
@@ -62,7 +62,7 @@ class CProcessInfo
   bool Supports(EINTERLACEMETHOD method);
   void SetDeinterlacingMethodDefault(EINTERLACEMETHOD method);
   EINTERLACEMETHOD GetDeinterlacingMethodDefault();
-  CVideoBufferManager& GetVideoBufferManager();
+  virtual CVideoBufferManager& GetVideoBufferManager();
```
FernetMenta: it should not be necessary to override the manager. the manager provices methods for pltforms to register their pools

xbmc/cores/VideoPlayer/Process/VideoBuffer.cpp
```diff
@@ -372,16 +379,30 @@ void CVideoBufferManager::ReleasePools()
 
 CVideoBuffer* CVideoBufferManager::Get(AVPixelFormat format, int width, int height)
 {
+  // todo: fix up other formats
```
FernetMenta: drop this Get method

peak3d:
I think its a good idea to keep both Get() methods. he one with width/height will be the one wich lets Buffer choose the best matching alignement as imlemented today.

FernetMenta:
This would be dead code and keeping dead code is imo not a good idea. It can be implemented again if a client requires it for some reason

popcornmix:
Updated to include passing videoBufferManager to Released callback.

PR #12424, opened 2017-07-04
Fix Settings::FadeButton / initial scrollbar display

peak3d:
Fix Settings::FadeButton / initial scrollbar display

## Description
- Moving focus in settings from menu into settings::grouplist shoud fade the category button
- Selecting new category in settings should update the scrollbar

Both issues broke with PR 12213, this PR fixes them

## Types of change

### DIFF
--- a/xbmc/guilib/GUIControlGroupList.cpp
+++ b/xbmc/guilib/GUIControlGroupList.cpp
@@ -35,7 +35,7 @@ CGUIControlGroupList::CGUIControlGroupList(int parentID, int controlID, float po
   m_totalSize = 0;
   m_orientation = orientation;
   m_alignment = alignment;
-  m_lastScrollerValue = 0;
+  m_lastScrollerValue = -1;
   m_useControlPositions = useControlPositions;
   ControlType = GUICONTROL_GROUPLIST;
   m_minSize = 0;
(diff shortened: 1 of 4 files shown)


peak3d:
@HitcherUK <- hope it solves the issues you tracked.

da-anda:
wouldn't it actually be better if skins could define how the "active but unfocused" state looks like? It's odd that the core is dictating how this should look like

peak3d:
agree 1000% @da-anda, that would mean a triple state, but maybe something like this is already implemented. What I was searching for is that a control has beside "focus" also a "selected" state. and skinner can design a layout for this.

HitcherUK:
@da-anda we can by using variables for the focus texture but it's a pain to do it for each different grouplist across a whole skin. This [fading] has been the default behaviour since forever.

peak3d:
@HitcherUK but wouldn't be a good idea to have a third state in the future?
The fade effect could be also make sense not only in Settings (where it is implemented right now) but as well in home menu.

HitcherUK:
Yes it would be a nice addition in the future for sure.
Note: A fade could easily be added to the Estuary home menu (it's actually a list control in this case) but I just assumed this was a design decision.

HitcherUK:
> @HitcherUK <- hope it solves the issues you tracked.

All fixed, thanks.

PR #12426, opened 2017-07-05
Fix crash on startup

garbear:
This fixes a crash on startup introduced in commit 26349e1, "Move CInputManager into service manager" from #12366.

Specifically, `CAppParamParser` references `CInputManager` which is owned by `CServiceManager`, which hasn't been created or initialized yet.

To fix this, we create `CServiceManager` before the input manager is accessed. This causes the IR remote commands to be performed on an uninitialized CInputManager, which is perfectly acceptable as these parameters are meant to be set before initialization.


garbear:
jenkins build this please

garbear:
Confirmed working in slack

garbear:
Reverted and fixed properly in #12432

PR #12431, opened 2017-07-06
game: fix order of init/deinit in ServiceManager

FernetMenta:
see title

@garbear /cc

PR #12432, opened 2017-07-06
Service Manager: Fix order of init/deinit

garbear:
This PR takes #12431 and finishes the task of ordering the individual services in a LIFO manner. Several circular dependencies arose, so some code was split up and refactored.

### DIFF
--- a/xbmc/ServiceManager.cpp
+++ b/xbmc/ServiceManager.cpp
@@ -40,10 +40,7 @@
 
 using namespace KODI;
 
-CServiceManager::CServiceManager() :
-  m_gameServices(new GAME::CGameServices),
-  m_peripherals(new PERIPHERALS::CPeripherals),
-  m_inputManager(new CInputManager)
+CServiceManager::CServiceManager()
 {
 }
 
@@ -55,6 +52,10 @@ bool CServiceManager::Init1()
 {
   m_announcementManager.reset(new ANNOUNCEMENT::CAnnouncementManager());
   m_announcementManager->Start();
+
+  m_gameServices.reset(new GAME::CGameServices);
+  m_peripherals.reset(new PERIPHERALS::CPeripherals);
+  m_inputManager.reset(new CInputManager);
 
 #ifdef HAS_PYTHON
   m_XBPython.reset(new XBPython());
@@ -152,9 +153,10 @@ bool CServiceManager::Init3()
 void CServiceManager::Deinit()
 {
   m_serviceAddons.reset();
-  m_gameServices->Deinit();
-  m_peripherals.reset();
+
   //m_inputManager->Deinitialize(); //! @todo
+  m_gameServices->Deinit();
+
   m_contextMenuManager.reset();
   m_favouritesService.reset();
   m_binaryAddonCache.reset();
@@ -169,6 +171,11 @@ void CServiceManager::Deinit()
   CScriptInvocationManager::GetInstance().UnregisterLanguageInvocationHandler(m_XBPython.get());
   m_XBPython.reset();
 #endif
+
+  m_inputManager.reset();
+  m_peripherals.reset();
+  m_gameServices.reset();
+
   m_announcementManager.reset();
   init_level = 0;
 }
--- a/xbmc/ServiceManager.cpp
+++ b/xbmc/ServiceManager.cpp
@@ -67,7 +67,7 @@ bool CServiceManager::InitStageOne()
   return true;
 }
 
-bool CServiceManager::InitStageTwo()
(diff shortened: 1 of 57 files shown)


xbmc/ServiceManager.h
```diff
@@ -153,6 +157,7 @@ class CServiceManager
   std::unique_ptr<CPlatform> m_Platform;
   std::unique_ptr<PLAYLIST::CPlayListPlayer> m_playlistPlayer;
   std::unique_ptr<CSettings> m_settings;
+  std::unique_ptr<KODI::GAME::CControllerManager> m_controllerManager;
```
FernetMenta: how about m_gameControllerManager and GetGameControllerManager? At first glance I was confused about this additional manager because I did not notice that this manager is in the namespace GAME.

FernetMenta:
thanks for cleaning up the mess. feel free to squash my commit into yours.

xbmc/ServiceManager.h
garbear: I actually moved this from `CGameServices` because input (and peripherals) depend on some controller stuff, but game services are brought up last. There still exists `CGameServices::GetController*()`, but this is only called from within GAME.

garbear: Oh, I see what you're saying. Just rename it. I could do that, but it would be redundant - Anything outside of game should reference this as `GAME::CControllerManager`

FernetMenta: > Anything outside of game should reference this as GAME::CControllerManager

right but ServiceManager is outside of namaspace game. CServiceManager::GetControllerManager vs CServiceManager::GetGameControllerManager

assume some other namespace has a ControllerManager too. I don't know id C++ allows having the same method more than once with just different return types. 
even if it is allowed, it does not look nice

garbear: > right but ServiceManager is outside of namaspace game. 

Good point. Renamed.

garbear:
OK, I've tested on win, linux and OSX, and blindly adapted the remaining platforms. Jenkins is building an old version, so when that's failed I'll start another run.

FernetMenta:
> Jenkins is building an old version

I noticed this too. Makes our lives a little bit harder :)

MartijnKaijser:
if you continuously push new revisions you build up a huge queue so you shouldn't blame jenkins ;)

garbear:
I've included a better fix for #12426 that involves wrapping up the remote params in `CAppParamParser` and then unwrapping them inside `CInputManager`, removing the need for calling into `CInputManager` directly.

>> Jenkins is building an old version
> I noticed this too. Makes our lives a little bit harder :)

I'm making liberal use of the `No-Jenkins` flag :)

garbear:
jenkins build this please

garbear:
jenkins build this please

garbear:
jenkins build this please

garbear:
jenkins build this please

garbear:
jenkins build this please

garbear:
jenkins build this please

garbear:
jenkins build this please

garbear:
jenkins build this please

garbear:
Jenkins build this please

garbear:
jenkins build this please

garbear:
jenkins build this please

garbear:
The `No-Jenkins` tag has tainted the pr. Now jenkins won't touch this with a 10-foot pole.

garbear:
jenkins build this please

garbear:
jenkins build this please

garbear:
Windows errors look unrelated.

xbmc/ServiceManager.cpp
```diff
@@ -138,38 +141,57 @@ bool CServiceManager::StartAudioEngine()
   return m_ActiveAE->Initialize();
 }
 
-bool CServiceManager::Init3()
+bool CServiceManager::InitStageThree()
 {
+  m_peripherals.reset(new PERIPHERALS::CPeripherals(*m_announcementManager));
   m_peripherals->Initialise();
-  m_PVRManager->Init();
```
FernetMenta: there was a reason why this was in stage three. stage 3 is init after WindowManager is initialized. PVR can have its own startup window, hence need to be in stage 3. on the other hand WindowManager depends on peripherials so peripherials must not be in stage 3

garbear: peripherals must not be in stage 2, because our test suite inits stage 2 and doesn't provide enough for the peripherals subsystem

FernetMenta: I care more about the app itself and less about the test suite. I created stage 3 for the reason given.

PR #12425, opened 2017-07-04
Compiler warnings and dead code

garbear:
This PR fixes some compiler warnings (first 7 commits) and removes some dead code (final 5 commits).

PR #12433, opened 2017-07-07
Fix missing brackets in dbwrappers code

garbear:
While hunting down compiler warnings for #12425, I encountered an actual error in the code. Due to the use of early returns this is benign, but a good lesson nonetheless.


Montellese:
And that's why I don't use `else` blocks at all after an `if` block with a guaranteed early return.

And since there is a guaranteed early return IMO you can't call it an "actual error" because it behaves exactly as it should. It's probably not what the developer wanted to write but technically it's not wrong.

PR #12434, opened 2017-07-07
Implement basic gameplay persistence (autosave)

garbear:
This implements a rudimentary autosave/autoload feature. Basically, when you close the game it's saved, and when you open the game the savestate is restored.

The savestate and metadata is stored alongside the ROM:

```
Chrono Trigger (U) [!].smc   (the ROM)
Chrono Trigger (U) [!].sav   (the state)
Chrono Trigger (U) [!].xml   (the metadata)
```

To guard against power loss, the game is autosaved every 10 seconds.

If a different game client is used, Kodi will warn the user that it will overwrite their existing save. This is because savestates are incompatible between emulators, and storing more than a single savestate isn't implemented yet:

screen shot 2017-07-10 at 12 57 00 pm

## Types of change

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -16969,7 +16969,19 @@ msgctxt "#35216"
 msgid "Support add-ons"
 msgstr ""
 
-#empty strings from id 35217 to 35219
+#. Dialog text asking user to overwrite their savestate file
+#: xbmc/cores/RetroPlayer/RetroPlayer.cpp"
+msgctxt "#35217"
+msgid "The savestate can only be loaded with \"{0:s}\". Erase savestate and continue?"
+msgstr ""
+
+#. Button to erase the savestate
+#: xbmc/cores/RetroPlayer/RetroPlayer.cpp"
+msgctxt "#35218"
+msgid "Erase savestate"
+msgstr ""
+
+#empty string with id 35219
 
 #. Description of add-on category for game providers
 #: xbmc/filesystem/AddonsDirectory.cpp
--- a/xbmc/cores/RetroPlayer/RetroPlayer.cpp
+++ b/xbmc/cores/RetroPlayer/RetroPlayer.cpp
@@ -20,9 +20,15 @@
 
 #include "RetroPlayer.h"
 #include "RetroPlayerAudio.h"
+#include "RetroPlayerAutoSave.h"
 #include "RetroPlayerVideo.h"
+#include "addons/AddonManager.h"
 #include "cores/VideoPlayer/Process/ProcessInfo.h"
+#include "dialogs/GUIDialogYesNo.h"
+#include "filesystem/File.h"
 #include "games/addons/playback/IGameClientPlayback.h"
+#include "games/addons/savestates/Savestate.h"
+#include "games/addons/savestates/SavestateUtils.h"
 #include "games/addons/GameClient.h"
 #include "games/ports/PortManager.h"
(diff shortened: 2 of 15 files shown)


addons/resource.language.en_gb/resources/strings.po
```diff
@@ -16963,7 +16963,19 @@ msgctxt "#35216"
...
+#. Dialog text asking user to overwrite their savestate file
+#: xbmc/cores/RetroPlayer/RetroPlayer.cpp"
+msgctxt "#35217"
+msgid "The savestate can only be loaded with \"{0:s}\". Erase savestate and continue?"
+msgstr ""
+
+#. Button to erase the savestate
+#: xbmc/cores/RetroPlayer/RetroPlayer.cpp"
+msgctxt "#35218"
+msgid "Erase save"
```
Montellese: Above you use "savestate" twice but here only "save". Is that intended or should we keep it consistent?

xbmc/cores/RetroPlayer/RetroPlayerAutoSave.h
```diff
@@ -0,0 +1,46 @@
...
+namespace KODI
+{
+namespace RETRO
+{
+  class CRetroPlayerAutoSave : protected CThread
+  {
+  public:
+    CRetroPlayerAutoSave(IPlayer *player);
+
+    virtual ~CRetroPlayerAutoSave() override;
```
Montellese: You don't need the `virtual` here and IIR there are compilers who spit out a warning if you combine `virtual` and `override` for a method definition.

Montellese (changes requested):
Looks good apart from minors. Did you add and commit the two `RPOverlayRendererGL` files on purpose or by accident? They haven't been added to any `CMakeLists.txt` so they aren't built and they also aren't included anywhere else.

addons/resource.language.en_gb/resources/strings.po
garbear: I was going for shorter text for the button, but probably best to keep it consistent

xbmc/cores/RetroPlayer/RetroPlayerAutoSave.h
garbear: I've never come across a compiler with this warning, but it can't hurt to remove `virtual`

garbear:
@Montellese addressed your comments

xbmc/cores/RetroPlayer/RetroPlayer.cpp
```diff
@@ -106,18 +113,43 @@ bool CRetroPlayer::OpenFile(const CFileItem& file, const CPlayerOptions& options
...
+      // Check if game client is the same
+      if (save.GameClient() != m_gameClient->ID())
+      {
+        ADDON::AddonPtr addon;
+        if (ADDON::CAddonMgr::GetInstance().GetAddon(save.GameClient(), addon))
+        {
+          // "Opening file"
+          // "The savestate can only be loaded with \"{0:s}\". Erase savestate and continue?"
+          // "Cancel"
+          // "Erase save"
```
Montellese: This isn't correct anymore. I wouldn't write this here anyway because it can get out of sync quickly.

garbear: OK, I'll summarize

garbear:
I've uploaded builds to the [test builds thread](https://forum.kodi.tv/showthread.php?tid=173361) and attached a screen shot of the warning message.

garbear:
jenkins build this please

PR #12438, opened 2017-07-07
[addons] add callback to remove slash at end of path

AlwinEsch:
## Description
This add a addon callback to remove a slash on path end if present. Is small but seen on several addons the need of them and thought to add in headers.

This change does not touch Kodi itself and not need changes on addon.

But that also brings a question from me. Can we add a git label with e.g. "Not down-compatible" for cases where a addon need update or recompile?

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
@@ -906,6 +906,37 @@ namespace vfs
   }
   //----------------------------------------------------------------------------
 
+
+  //============================================================================
+  ///
+  /// @ingroup cpp_kodi_vfs
+  /// @brief Remove the slash on given path name
+  ///
+  /// @param[in,out] path The complete path
+  ///
+  ///
+  /// ------------------------------------------------------------------------
+  ///
+  /// **Example:**
+  /// ~~~~~~~~~~~~~{.cpp}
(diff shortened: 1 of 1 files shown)


Paxxi:
Why isn't it using our existing removeslashatend method?

AlwinEsch:
Want to hold callbacks to kodi itself small as possible and has not seen there a need for "void URIUtils::RemoveSlashAtEnd(std::string& strFolder)", the inline do mostly the same.

Paxxi:
It's the "mostly the same" that worries me, can lead to some annoying issues.

AlwinEsch:
You mean that someone use "kodi::vfs::RemoveSlashAtEnd" in Kodi or someone try to use "URIUtils::RemoveSlashAtEnd" on addon?

Another thing is to hold separate is the hard way to become them called with "C" from addon.

Paxxi:
yes it does require more work to get it done in a nice way unfortunately. Another reason we want to isolate all of these utility methods into their own library really.

notspiff:
I repeat my previous statement: kodi is not a toolbox for addons.

ksooo:
Why do you introduce API versions if you never bump them? I'm getting sick of having to write this again and again.

AlwinEsch:
As long it is a development version and not direct in the interface to Kodi itself is changed is it not so critical for me.

AlwinEsch:
Think it was there but loosed during rebase and has it overlook.

PR #12086, opened 2017-05-13
Fix seeking for pts values larger than max_int

rbuehlma:
Fix seeking for PTS values larger than max-int

## Description
If a stream uses very high PTS/DTS values, the current code truncates this value to integer instead of using double.

## Motivation and Context
Seeking in such streams is possible

## How Has This Been Tested?
Successfully seek forward and backwards in such a stream

## Types of change
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [X] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [X] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #12442, opened 2017-07-07
Fix seeking with PTS values exceeding max int

rbuehlma:
This is a follow up to https://github.com/xbmc/xbmc/pull/12086

I found some additional places where double got pushed into an int.

### DIFF
--- a/xbmc/cores/VideoPlayer/Edl.cpp
+++ b/xbmc/cores/VideoPlayer/Edl.cpp
@@ -770,19 +770,19 @@ int CEdl::RemoveCutTime(int iSeek) const
   return iSeek - iCutTime;
 }
 
-int CEdl::RestoreCutTime(int iClock) const
+double CEdl::RestoreCutTime(double dClock) const
 {
   if (!HasCut())
-    return iClock;
+    return dClock;
 
-  int iSeek = iClock;
+  double dSeek = dClock;
   for (int i = 0; i < (int)m_vecCuts.size(); i++)
   {
-    if (m_vecCuts[i].action == CUT && iSeek >= m_vecCuts[i].start)
-      iSeek += m_vecCuts[i].end - m_vecCuts[i].start;
+    if (m_vecCuts[i].action == CUT && dSeek >= m_vecCuts[i].start)
+      dSeek += static_cast<double>(m_vecCuts[i].end - m_vecCuts[i].start);
   }
 
-  return iSeek;
+  return dSeek;
 }
 
 bool CEdl::HasSceneMarker() const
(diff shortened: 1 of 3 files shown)


xbmc/cores/VideoPlayer/Edl.cpp
```diff
@@ -770,16 +770,16 @@ int CEdl::RemoveCutTime(int iSeek) const
   return iSeek - iCutTime;
 }
 
-int CEdl::RestoreCutTime(int iClock) const
+double CEdl::RestoreCutTime(double iClock) const
```
ksooo: If you change the data type, then please also change the variable's name. Prefix "i" stands for "int". Applies also to other places of your commit.

ksooo (changes requested):

xbmc/cores/VideoPlayer/Edl.cpp
rbuehlma: Sure, here we are :)

xbmc/cores/VideoPlayer/Edl.cpp
```diff
@@ -770,19 +770,19 @@ int CEdl::RemoveCutTime(int iSeek) const
...
+    return dClock;
 
-  int iSeek = iClock;
+  double dSeek = dClock;
   for (int i = 0; i < (int)m_vecCuts.size(); i++)
   {
-    if (m_vecCuts[i].action == CUT && iSeek >= m_vecCuts[i].start)
-      iSeek += m_vecCuts[i].end - m_vecCuts[i].start;
+    if (m_vecCuts[i].action == CUT && dSeek >= m_vecCuts[i].start)
+      dSeek += (double)m_vecCuts[i].end - (double)m_vecCuts[i].start;
```
ksooo: please use c++ style casts

ksooo (changes requested):

xbmc/cores/VideoPlayer/Edl.cpp
rbuehlma: I removed the cast. If I understand C++ correctly, the calculation is done in double any way as it expands to dSeek = dSeek + int.
As we calculate here with a double, the int is promoted to a double.
Can you confirm that?

garbear: Better include the cast. Windows warns about this conversion:

```
ActiveAE.cpp(1911): warning C4244: 'initializing': conversion from 'int64_t' to 'double', possible loss of data
```

FernetMenta (approved):

xbmc/cores/VideoPlayer/Edl.cpp
rbuehlma: Done. Better now?

MartijnKaijser:
jenkins build this please

ksooo (approved):

MartijnKaijser:
jenkins build this please

PR #12444, opened 2017-07-08
[addons] change OnEnabled and OnDisabled to use AddonPtr

AlwinEsch:
## Description

Before was the call done with the ID string only and limited to PVR
addon.

This change them to use the on CAddonManager created IAddon class
to give related addon.

Test here with PVR and Peripheral was successful.

This is required part on next changes where the CBinaryAddonManager
handle the addon changes and to inform related parts thread safe.

As note, PVR and some other binary addons can not work with the
"AddonEvents"! There must be e.g. before uninstall confirmed
that the running addon is stopped and by install thats present.

Further has I seen that currently e.g. on peripheral addons the disable
and enable not work with "AddonEvents" from the above described reasons.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/Addon.h
+++ b/xbmc/addons/Addon.h
@@ -36,8 +36,8 @@ namespace ADDON
 
   const char* const ORIGIN_SYSTEM = "b6a50484-93a0-4afb-a01c-8d17e059feda";
 
-void OnEnabled(const std::string& id);
-void OnDisabled(const std::string& id);
+void OnEnabled(const AddonPtr& addon);
+void OnDisabled(const AddonPtr& addon);
 void OnPreInstall(const AddonPtr& addon);
 void OnPostInstall(const AddonPtr& addon, bool update, bool modal);
 void OnPreUnInstall(const AddonPtr& addon);
--- a/xbmc/addons/Addon.cpp
+++ b/xbmc/addons/Addon.cpp
@@ -393,20 +393,14 @@ AddonVersion CAddon::GetDependencyVersion(const std::string &dependencyID) const
   return AddonVersion("0.0.0");
 }
 
-void OnEnabled(const std::string& id)
+void OnEnabled(const AddonPtr& addon)
 {
-  // If the addon is a special, call enabled handler
-  AddonPtr addon;
-  if (CAddonMgr::GetInstance().GetAddon(id, addon, ADDON_PVRDLL))
-    return addon->OnEnabled();
+  addon->OnEnabled();
 }
 
-void OnDisabled(const std::string& id)
+void OnDisabled(const AddonPtr& addon)
 {
-
-  AddonPtr addon;
-  if (CAddonMgr::GetInstance().GetAddon(id, addon, ADDON_PVRDLL, false))
-    return addon->OnDisabled();
+  addon->OnDisabled();
 }
 
 void OnPreInstall(const AddonPtr& addon)
--- a/xbmc/addons/AddonManager.cpp
+++ b/xbmc/addons/AddonManager.cpp
@@ -827,11 +827,12 @@ bool CAddonMgr::DisableAddon(const std::string& id)
     return false;
 
   //success
-  ADDON::OnDisabled(id);
-
   AddonPtr addon;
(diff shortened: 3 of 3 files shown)


garbear (approved):

xbmc/addons/Addon.h
```diff
@@ -36,8 +36,8 @@ namespace ADDON
 
   const char* const ORIGIN_SYSTEM = "b6a50484-93a0-4afb-a01c-8d17e059feda";
 
-void OnEnabled(const std::string& id);
-void OnDisabled(const std::string& id);
+void OnEnabled(const AddonPtr& addon);
+void OnDisabled(const AddonPtr& addon);
 void OnPreInstall(const AddonPtr& addon);
 void OnPostInstall(const AddonPtr& addon, bool update, bool modal);
 void OnPreUnInstall(const AddonPtr& addon);
```
AlwinEsch: @tamland what do you thing. Leave this functions there with:
```cpp
ADDON::OnPreUnInstall(addon);
```
or do direct on related call places a:
```cpp
addon->OnPreUnInstall();
```
for me does not make sense any more the use of the static functions.

tamland: The reason these static functions were added in the first place is because there are multiple different versions of add-ons (e.g the one in the repository, and ones that were loaded as a different type). And the result is that you if you call it on the "wrong" instance you will end up calling the wrong hook.

Better to just fix the remaining implementations of these On* methods than polish this turd.

tamland:
> This is required part on next changes where the CBinaryAddonManager
handle the addon changes and to inform related parts thread safe.
>
> As note, PVR and some other binary addons can not work with the
"AddonEvents"! There must be e.g. before uninstall confirmed
that the running addon is stopped and by install thats present.

As I've mentioned before adding back external, global, blocking calls into addon manager is a no go. I've already patched this inherently flawed design like 15 times. If AddonEvents really cannot work (I still don't see why not btw), then it needs a different solution doesn't ignore dependency management and blocking.

AlwinEsch:
There comes a complete another way not related to AddonManager.

Becomes also done with non blocking calls, but confirmed that a disable is done before the next comes.
In binary addon manager comes a thread who work over a StateMachine (https://github.com/AlwinEsch/kodi/pull/31/commits/480e3cff17de2b1657ee22d55a8caa891517a2ae)

With PVR, where know since I created 2009 is it very, very critical to handle addon changes and also @ksooo can sing a song of.

tamland:
> AlwinEsch/kodi@480e3cf

Globals inside the add-on instances like that is exactly what I'm talking about.

AlwinEsch:
Yes, but seen currently during test a abolute no go for me. 

Was on peripheral.joystick with the current becomes the AddonEvents::UnInstalled called complete alone **after the dll is complete deleted**. Until then was the addon still running here **without dll** and his path no more present. 

If now any addon (no matter what kind) would try at the end, to access its folder but this would be no longer there!

Then can bring a addon a crash where no one can understand why.

There is also no kind of disable called before to give the time to stop addon.

tamland:
Exactly what isn't working?

I'm not saying we cannot have synchronous hooks in addition to events, *if* it is actually necessary. I'm saying these old instance hooks with globals is a dead end and there is no point in adding new code that use it. (Likely it won't even work anymore and just crash now that we have pre-startup add-on migration).

AlwinEsch:
I see the same and should be gone. 

Problem for me is simply not to touch the CAddonManager and add not another call type there.

There should be only one way to do that. But is very hard to implement.
Now that the CBinaryAddonManager is there makes it even more difficult (and did not want it at the beginning because of it).

This must before know that an addon is changed and only then the corresponding Type Manager.

The "AddonEvents" is really a good way but bring some bad things:
- No synchronous calls possible
- Every asynchronous call is in own thread, if several comes together can one confuse the other
- Currently the event give only the changed ID of addon, there must be every called place compare this with his addons who brings a lot of work for one changed addon.

So I would also gone through this old body of callbacks to IAddon, since Ensured there the only one CAddonDll calls the associated object in CBinaryAddonManager and this then makes his stuff.

The "IAddonMgrCallback" type I want to remove complete, this have only blocking calls who is really bad.

PR #12445, opened 2017-07-08
[PVR] Do not mark in progress recordings as watched.

ksooo:
Do not automatically mark in progress recordings as watched. Makes no sense.

@Jalle19 mind taking a look.

### DIFF
--- a/xbmc/FileItem.cpp
+++ b/xbmc/FileItem.cpp
@@ -847,6 +847,11 @@ bool CFileItem::IsDeletedPVRRecording() const
   return (m_pvrRecordingInfoTag && m_pvrRecordingInfoTag->IsDeleted());
 }
 
+bool CFileItem::IsInProgressPVRRecording() const
+{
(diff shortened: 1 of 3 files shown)


Jalle19:
It's probably fine, but why use separate logic for recordings in general?

ksooo:
> It's probably fine, but why use separate logic for recordings in general?

How would the general logic look like?

Jalle19:
Doesn't it behave just like for library items now, i.e. if you watch something but don't watch it completely til the end it's marked as watched but with a different icon indicating that it is in progress (this possibly depends on the skin).

ksooo:
No, currently, if you watch a in-progress recording, and skip past the current end of the recording, it will be stopped and marked watched.

Jalle19:
You're right, didn't think about that

Jalle19 (approved):

AlwinEsch (approved):

PR #12446, opened 2017-07-08
Fix compiler warnings: 'foo' has virtual functions but non-virtual destructor

ksooo:
... just what the title says.

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.h
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.h
@@ -50,6 +50,8 @@ class CRenderManager;
 class IRenderMsg
 {
   friend CRenderManager;
+public:
+  virtual ~IRenderMsg() = default;
 protected:
   virtual void VideoParamsChange() = 0;
   virtual void GetDebugInfo(std::string &audio, std::string &video, std::string &general) = 0;
--- a/xbmc/ContextMenuItem.h
+++ b/xbmc/ContextMenuItem.h
@@ -37,6 +37,7 @@ namespace ADDON
 class IContextMenuItem
 {
 public:
+  virtual ~IContextMenuItem() = default;
   virtual bool IsVisible(const CFileItem& item) const = 0;
   virtual bool Execute(const CFileItemPtr& item) const = 0;
   virtual std::string GetLabel(const CFileItem& item) const = 0;
--- a/xbmc/addons/AddonProvider.h
+++ b/xbmc/addons/AddonProvider.h
@@ -36,6 +36,7 @@ namespace ADDON
   class IAddonProvider
   {
   public:
+    virtual ~IAddonProvider() = default;
     enum INSTANCE_TYPE
     {
       INSTANCE_VIDEOCODEC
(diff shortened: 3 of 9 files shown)


AlwinEsch:
Thats good to have, during addon tests with my new interface I seen also that the parent-parent destructor was not called without them on parent of child.

Mean like:

| CMyAddon -> | CInstanceScreensaver -> | IAddonInstance |
|:--------------------|:------------------------------------|:---------------------|
| no matter | if destructor missing | present desctructor not called

xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.h
```diff
@@ -50,6 +50,8 @@ class CRenderManager;
 class IRenderMsg
 {
   friend CRenderManager;
+public:
+  virtual ~IRenderMsg() = default;
```
FernetMenta: declaring a dtor on an interface looks weird. there is nothing to destruct.

ksooo: I guess you are right for interfaces (== classes with only pure virtual member functions). Will remove the respective changes from this PR.

ksooo: Well, maybe it is actually a good idea (if not needed) to add a default virtual ctor, even to a c++ interface, as the compiler may generate a non-virtual dtor for that class:

<pre>
class IBatch
{
    public:
       virtual void f() = 0;
};

class A : public IBatch
{
    public:
       void f() {}
       ~A() {}
};

IBatch * a = new A();
a->f();    // calls A::f()
delete a;  // calls IBatch::~IBatch() not A::~A()
</pre> 

Source: https://stackoverflow.com/questions/2691611/destructors-for-c-interface-like-classes

ksooo: @paxxi ^^^?

FernetMenta: good catch. i think we should check all of our interface classes for this.

ksooo: I did, on macOS, using the respective clang warning. So, only non-Mac classes still need to be checked.

pkerling: Do note that this is only relevant if you are (potentially) going to destroy the class via a pointer to the interface.

ksooo: > Do note that this is only relevant if you are (potentially) going to destroy the class via a pointer to the interface.

 ofc, but as you never know before and imo nobody checks the presence of a virtual dtor when first heap allocating a respective object and the price is just one more vtbl entry I tend to add this everywhere and be on the safe side.

ksooo:
jenkins build this please

ksooo:
jenkins error is unrelated

xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.h
Paxxi: sorry about the late response, good catch ksoo, fully approve

PR #12457, opened 2017-07-10
Restore = default constructor for DBusConnectionDeleter

yol:
Proper fix that replaces #12448 

@tamland @candrews please test

## Description
The "= default" was previously removed because it did not work when
in the header file with newer gcc and clang. But it seems to be ok in
the implementation file.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/linux/DBusUtil.cpp
+++ b/xbmc/linux/DBusUtil.cpp
@@ -222,6 +222,8 @@ CDBusConnection::operator DBusConnection*()
   return m_connection.get();
 }
 
+CDBusConnection::DBusConnectionDeleter::DBusConnectionDeleter() = default;
+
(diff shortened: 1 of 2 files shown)


Paxxi:
Unique_ptr can be tricky if all it's arguments can't be fully determined. Moving the operator implementation to the header file should solve it, obviously not wanted because of the dbus method usage. 

We do have utility class that can be used in these situations in utils/scopeguard.h it's designed for these kinds of third-party handles that requires custom deletion

pkerling:
Oh the arguments should be fully determinable here - DBusConnection and DBusConnectionDeleter are fully declared before the unique_ptr comes into play. I'm really not sure what the problem of the compiler is here and why it makes a difference whether you write ` = default` or ` {}` for the constructor. Remember, with `DBusConnectionDeleter() {}` in the header everything was ok.

PR #12458, opened 2017-07-10
Modernize CEventGroup

yol:
## Description
The va_arg stuff is not necessary any more with initializer lists. Other code can be simplified with range-based for and smart pointers.

## Motivation and Context
Modernization/cleanup

## How Has This Been Tested?
Build/run on linux, testsuite

## Types of change

## Checklist:

### DIFF
--- a/xbmc/threads/Event.cpp
+++ b/xbmc/threads/Event.cpp
@@ -20,7 +20,7 @@
  *
  */
 
-#include <stdarg.h>
+#include <algorithm>
 #include <limits>
 
 #include "Event.h"
@@ -28,8 +28,8 @@
 void CEvent::addGroup(XbmcThreads::CEventGroup* group)
 {
   CSingleLock lock(groupListMutex);
-  if (groups == NULL)
-    groups = new std::vector<XbmcThreads::CEventGroup*>();
+  if (!groups)
+    groups.reset(new std::vector<XbmcThreads::CEventGroup*>);
 
   groups->push_back(group);
 }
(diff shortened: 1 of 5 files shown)


xbmc/threads/Event.h
```diff
@@ -111,35 +113,24 @@ namespace XbmcThreads
...
+    XbmcThreads::TightConditionVariable<CEvent*&> condVar{actualCv, signaled};
     CCriticalSection mutex;
 
-    unsigned int numWaits;
+    unsigned int numWaits{0};
 
     // This is ONLY called from CEvent::Set.
     inline void Set(CEvent* child) { CSingleLock l(mutex); signaled = child; condVar.notifyAll(); }
-
+
```
MartijnKaijser: Unneeded white space

pkerling:
Fixed, thanks. Turns out `git rebase --whitespace=fix` does not work with `-i` for some reason.

xbmc/threads/Event.cpp
```diff
@@ -115,11 +105,9 @@ namespace XbmcThreads
     // signaled and sets 'signaled' to the first one it
     // finds.
     // ==================================================
-    signaled = NULL;
-    for (std::vector<CEvent*>::iterator iter = events.begin();
-         signaled == NULL && iter != events.end(); ++iter)
+    signaled = nullptr;
+    for (auto* cur : events)
```
Paxxi: Should not be required to specify * here, auto will deduce it since the vector members are pointers. Doesn't hurt, just a tip.

Paxxi (approved):
Very nice!

xbmc/threads/Event.cpp
pkerling: Yep, `*` is not required here - it's actually on purpose. I wanted it to be immediately visible that you're iterating over plain pointers.

pkerling:
jenkins failure seems to be unrelated

PR #12459, opened 2017-07-10
Fix missing member initialization

notspiff:
Ran into this while trying to use valgrind for something useful. tons of things like

 ==7579== Conditional jump or move depends on uninitialised value(s)
==7579==    at 0x198DA74: INFO::InfoBool::Get(CGUIListItem const*) (InfoBool.h:50)

also fixes a variable name and a rather amusing use of auto.

### DIFF
--- a/xbmc/interfaces/info/InfoBool.cpp
+++ b/xbmc/interfaces/info/InfoBool.cpp
@@ -28,7 +28,7 @@ namespace INFO
       m_context(context),
       m_listItemDependent(false),
       m_expression(expression),
-      m_refeshCounter(0),
+      m_refreshCounter(0),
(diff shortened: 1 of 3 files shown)


Montellese (approved):

BigNoid:
This PR somehow breaks the following in Aeon Nox on startup:
```
<include file="Includes_16x9.xml" condition="!String.IsEqual(Skin.AspectRatio,21:9)" />
<include file="Includes_21x9.xml" condition="String.IsEqual(Skin.AspectRatio,21:9)" />
```
A skin.reload after startup will fix it and switching to the skin also doesn't give the problem.
Not sure where to start to look how to fix this (other than reverting).

PR #12460, opened 2017-07-10
gbm and vaapi gles updates

lrusak:
This brings RendererVAAPIGLES up to par with RendererVAAPIGL

Doing so requires using OpenGLES 3.0

I've had to change how HAS_GLES works a little bit so I had to adjust some ifdefs.

In the future it would be nice to make this GLES_VERSION and GL_VERSION instead.

This is tested working on my kaby lake NUC :+1:


xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererVAAPIGL.cpp
```diff
@@ -229,7 +229,7 @@ void CRendererVAAPI::AfterRenderHook(int idx)
   if (glIsSync(m_fences[idx]))
   {
     glDeleteSync(m_fences[idx]);
-    m_fences[idx] = None;
+    m_fences[idx] = 0;
```
FernetMenta: GL_NONE is better

PR #12462, opened 2017-07-11
[cmake] fix lowercase x11

lrusak:
@hauihau thanks for pointing this out.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,7 +97,7 @@ list(APPEND DEPLIBS ${CMAKE_THREAD_LIBS_INIT})
 if(CORE_SYSTEM_NAME STREQUAL linux)
   # Set default CORE_PLATFORM_NAME to X11
   # This is overridden by user setting -DCORE_PLATFORM_NAME=<platform>
-  set(_DEFAULT_PLATFORM X11)
+  set(_DEFAULT_PLATFORM x11)
 else()
   string(TOLOWER ${CORE_SYSTEM_NAME} _DEFAULT_PLATFORM)
 endif()


hudokkow:
jenkins build this please

pkerling:
What about when the user specifies `X11` (uppercase) by accident? Shouldn't we just always lowercase `CORE_PLATFORM_NAME`?

hudokkow:
Yep, you're right.
Code needs some refactoring. I'll do it later if no one beats me to it.

wsnipex:
we already use lower case: https://github.com/xbmc/xbmc/blob/master/CMakeLists.txt#L107-L111
So this PR shouldn't fix anything

pkerling:
@wsnipex this only applies to `CORE_PLATFORM_NAME_LC` which is only available in the toplevel `CMakeLists.txt` and only used for the `include`. `CORE_PLATFORM_NAME` can still have uppercase characters.

hauihau:
From my understanding, thats not correct: you're saving the lowercase value of **CORE_PLATFORM_NAME** in **CORE_PLATFORM_NAME_LC**.

FindVAAPI.cmake uses **CORE_PLATFORM_NAME** in a STREQUAL check against lowercase x11 while **CORE_PLATFORM_NAME** contains uppercase X11. Also cmake output says uppercase X11:

```
-- #---- CONFIGURATION ----#
-- -- PATH config --
-- Prefix: /usr
-- Libdir: /usr/lib64
-- Bindir: /usr/bin
-- Includedir: /usr/include
-- Datarootdir: /usr/share
-- Datadir: /usr/share
-- Docdir: /usr/share/doc/kodi-9999
-- Platform: X11
[snip]
```

hudokkow:
That ^^

wsnipex:
ok, then everything that currently uses CORE_PLATFORM_NAME as condition should be changed to use CORE_PLATFORM_NAME_LC instead

hudokkow:
I'll refactor the code and fix the logic. Meanwhile this can go in.

hudokkow:
Okido!

PR #12464, opened 2017-07-11
[video] Readd mark watched / unwatched to video folder's context menu

ksooo:
This is my desperate ( ;-) ) attempt to fix the problem reported here: https://forum.kodi.tv/showthread.php?tid=311447&page=2

It works, but I'm not sure whether the approach to use an item property to transport the type of the folder is okay. Feedback from devs is welcome.

This has been tested on latest master on macOS.

@notspiff fyi

### DIFF
--- a/xbmc/video/ContextMenus.cpp
+++ b/xbmc/video/ContextMenus.cpp
@@ -54,10 +54,12 @@ bool CMarkWatched::IsVisible(const CFileItem& item) const
   if (item.IsDeleted()) // e.g. trashed pvr recording
     return false;
 
-  if (item.m_bIsFolder) //Only allow db content and recording folders to be updated recursively
+  if (item.m_bIsFolder) // Only allow video db content, video and recording folders to be updated recursively
   {
     if (item.HasVideoInfoTag())
       return item.IsVideoDb();
+    else if (item.GetProperty("IsVideoFolder").asBoolean())
+      return true;
     else
       return CUtil::IsTVRecording(item.GetPath());
   }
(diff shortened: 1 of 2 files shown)


ksooo:
Too whom it may concern: If nobody speaks  up I will merge this PR the next days. ;-)

ksooo:
jenkins build this please

xbmc/video/ContextMenus.cpp
```diff
@@ -58,6 +58,8 @@ bool CMarkWatched::IsVisible(const CFileItem& item) const
   {
     if (item.HasVideoInfoTag())
       return item.IsVideoDb();
+    else if (item.GetProperty("IsVideoFolder").asBoolean())
```
xhaggi: If this can only be a recording folder or a video item, why not change the if-else to check video info tag, is recording, not parent folder and else return false.

xhaggi: And please adjust the comment in line 57

xbmc/video/ContextMenus.cpp
```diff
@@ -82,6 +84,8 @@ bool CMarkUnWatched::IsVisible(const CFileItem& item) const
...
```
xhaggi: Same here + comment in line 83

xbmc/video/ContextMenus.cpp
ksooo: Not sure I understand what you mean. The problem is that item can be anything (!) and we must only return true in case of certain items - namely videos/recordings and video folders/recording folders.

xhaggi: I thought that this context class is only involved if the item is holds a video item, because it's `video/ContextMenus.cpp`. If that is not the case ignore my comment.

ksooo:
Updated the comments, as requested.

I will merge this PR tomorrow, if nobody abjects.

ksooo:
jenkins error is unrelated

Nostalgist:
Thanks for looking into this! Does this fix work for anyone? I can neither reset the watched status of files inside a folder via context menu nor by pressing "w".

PR #12465, opened 2017-07-11
[cmake] use normalized to lower case CORE_PLATFORM_NAME in conditionals

wsnipex:
replaces https://github.com/xbmc/xbmc/pull/12462

@hudokkow @lrusak @pkerling

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,6 +103,10 @@ else()
 endif()
 
 set(CORE_PLATFORM_NAME ${_DEFAULT_PLATFORM} CACHE STRING "Platform port to build")
+#
+# Note: please do not use CORE_PLATFORM_NAME in any checks,
+# use the normalized to lower case CORE_PLATFORM_NAME_LC (see below) instead
+#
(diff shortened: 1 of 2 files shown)


pkerling:
Is there a particular reason not to modify CORE_PLATFORM_NAME directly? Would be less confusing.

wsnipex:
the idea is not to modify user supplied configs/values.

pkerling:
Mh I see. Is there something we can do to "prevent" use of `CORE_PLATFORM_NAME`? It's just very tempting to accidentally use it and then get into trouble. At the very least, I think there should be a comment somewhere that says not to use it :)

wsnipex:
valid point, added a comment

hudokkow:
I came up with https://github.com/hudokkow/xbmc/commit/3f4e00226d79f2e4bad638f43c75c80493789869 but violates the "... not to modify user supplied configs/value" part...

I'm fine with whatever solution is agreed upon.

lrusak:
I'm fine with this

lrusak (approved):

PR #12468, opened 2017-07-11
[addons] Fix compiler warning: xbmc/xbmc/addons/RepositoryUpdater.h: …

ksooo:
…Private field 'm_addonMgr' is not used.

Actually use the member instead of still using global instance. ;-)

@tamland good to go?


tamland:
Yeah, thanks!

PR #12470, opened 2017-07-11
VideoPlayer: rework speed and tempo

FernetMenta:
Rework on tempo and speed. This allows VideoPlayer to continue with tempo set after pause/resume. Tempo range is now defined by ProcessInfo and can be overridden by platforms and set at runtime.

### DIFF
--- a/xbmc/ApplicationPlayer.cpp
+++ b/xbmc/ApplicationPlayer.cpp
@@ -122,7 +122,6 @@ PlayBackRet CApplicationPlayer::OpenFile(const CFileItem& item, const CPlayerOpt
     m_audioStreamUpdate.SetExpired();
     m_videoStreamUpdate.SetExpired();
     m_subtitleStreamUpdate.SetExpired();
-    m_speedUpdate.SetExpired();
   }
   return iResult;
 }
@@ -245,7 +244,6 @@ void CApplicationPlayer::Pause()
   if (player)
   {
     player->Pause();
-    m_speedUpdate.SetExpired();
   }
 }
 
@@ -476,6 +474,13 @@ void CApplicationPlayer::SetSpeed(float speed)
   std::shared_ptr<IPlayer> player = GetInternal();
   if (player)
     player->SetSpeed(speed);
+}
+
+void CApplicationPlayer::SetTempo(float tempo)
+{
+  std::shared_ptr<IPlayer> player = GetInternal();
+  if (player)
+    player->SetTempo(tempo);
 }
 
 void CApplicationPlayer::DoAudioWork()
@@ -721,20 +726,25 @@ void CApplicationPlayer::SetPlaySpeed(float speed)
     return ;
 
   SetSpeed(speed);
-  m_speedUpdate.SetExpired();
 }
 
 float CApplicationPlayer::GetPlaySpeed()
 {
-  if (!m_speedUpdate.IsTimePast())
-    return m_fPlaySpeed;
-
   std::shared_ptr<IPlayer> player = GetInternal();
   if (player)
   {
-    m_fPlaySpeed = player->GetSpeed();
-    m_speedUpdate.Set(1000);
-    return m_fPlaySpeed;
+    return CDataCacheCore::GetInstance().GetSpeed();
+  }
+  else
+    return 0;
+}
+
+float CApplicationPlayer::GetPlayTempo()
+{
+  std::shared_ptr<IPlayer> player = GetInternal();
+  if (player)
+  {
+    return CDataCacheCore::GetInstance().GetTempo();
   }
   else
     return 0;
--- a/xbmc/GUIInfoManager.cpp
+++ b/xbmc/GUIInfoManager.cpp
@@ -6089,7 +6089,12 @@ std::string CGUIInfoManager::GetLabel(int info, int contextWindow, std::string *
     break;
   case PLAYER_PLAYSPEED:
       if(g_application.m_pPlayer->IsPlaying())
-        strLabel = StringUtils::Format("%.2f", g_application.m_pPlayer->GetPlaySpeed());
+      {
+        float speed = g_application.m_pPlayer->GetPlaySpeed();
+        if (speed == 1.0)
+          speed = g_application.m_pPlayer->GetPlayTempo();
+        strLabel = StringUtils::Format("%.2f", speed);
+      }
       break;
   case MUSICPLAYER_TITLE:
   case MUSICPLAYER_ALBUM:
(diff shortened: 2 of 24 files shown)


xbmc/ApplicationPlayer.cpp
```diff
@@ -478,6 +476,13 @@ void CApplicationPlayer::SetSpeed(float speed)
     player->SetSpeed(speed);
 }
 
+void CApplicationPlayer::SetTempo(float speed)
```
ace20022: float tempo

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -2953,7 +2963,8 @@ void CVideoPlayer::HandleMessages()
             m_ChannelEntryTimeOut.SetInfinite();
             CloseDemuxer();
             m_playSpeed = DVD_PLAYSPEED_NORMAL;
-            m_newPlaySpeed = DVD_PLAYSPEED_NORMAL;
+            m_processInfo->SetSpeed(1.0);
+            m_processInfo->SetTempo(1.0);
```
ace20022: Not sure if it's worth, but these two lines repeat 3 times, maybe add a function ```SetSpeedAndTempo```?

ace20022 (commented):
A general question: Is ```speed``` allowed (or desinged) to have a value outside of \{0.0, 1.0\}?

xbmc/cores/VideoPlayer/VideoPlayer.cpp
FernetMenta: That is the pvr channel switch code that will go away completely (hopefully soon)

FernetMenta:
> A general question: Is speed allowed (or desinged) to have a value outside of {0.0, 1.0}?

yes of course, i.e. -1,-2,-4, ... for rewind or 2,4,8 for ff. It can even be float values.

xbmc/ApplicationPlayer.cpp
FernetMenta: fixed

xbmc/cores/VideoPlayer/Process/ProcessInfo.cpp
```diff
@@ -424,6 +424,70 @@ bool CProcessInfo::IsSeeking()
...
+
+  return m_newTempo;
+}
+
+bool CProcessInfo::IsTempoAllowed(float tempo)
+{
+  if (tempo > 0.75 && tempo < 1.55)
+    return true;
+
+  return false;
```
Rechi: any reason for not using `return tempo > 0.75 && tempo < 1.55`?

FernetMenta: i avoid multiple statements on a single line. those are pest when it comes to debugging.

xbmc/GUIInfoManager.cpp
```diff
@@ -7182,7 +7187,7 @@ bool CGUIInfoManager::GetBool(int condition1, int contextWindow, const CGUIListI
     case PLAYER_PLAYING:
       {
         float speed = g_application.m_pPlayer->GetPlaySpeed();
-        bReturn = (speed >= 0.75 && speed <= 1.55);
+        bReturn = (speed == 1.0);
```
AlwinEsch: Speed is a float, not better to use as "1.0f"?

xbmc/GUIInfoManager.cpp
```diff
@@ -7238,8 +7243,9 @@ bool CGUIInfoManager::GetBool(int condition1, int contextWindow, const CGUIListI
       break;
     case PLAYER_IS_TEMPO:
       {
+        float tempo = g_application.m_pPlayer->GetPlayTempo();
         float speed = g_application.m_pPlayer->GetPlaySpeed();
-        bReturn = (speed >= 0.75 && speed <= 1.55 && speed != 1);
+        bReturn = (speed == 1.0 && tempo != 1);
```
AlwinEsch: Here too.

AlwinEsch: Tempo as float but compared as integer, maybe bring a warning.

xbmc/GUIInfoManager.cpp
FernetMenta: I don't think that makes any difference

xbmc/GUIInfoManager.cpp
FernetMenta: i didn't get any warning and that did not change

xbmc/GUIInfoManager.cpp
garbear: In general, windows warns on conversion of double to float:

`warning C4244: conversion from 'double' to 'float', possible loss of data`

Kodi currently has 200 of such warnings.

I don't think that this line generates a warning on Windows, but I prefer to always match precisions because it makes it easier to reason about the code I maintain.

FernetMenta: if it does not generate a warning, I see no benefit in adding this silly f. I'll leave this to whoever wants to be the maintainer of this part of the app that would need a rewrite from scratch anyway

FernetMenta:
jenkins build this please

xbmc/cores/RetroPlayer/RetroPlayer.cpp
```diff
@@ -357,14 +358,9 @@ void CRetroPlayer::SetSpeed(float speed)
...
       CloseOSD();
-  }
-}
 
-float CRetroPlayer::GetSpeed()
-{
-  if (m_gameClient)
-    return static_cast<float>(m_gameClient->GetPlayback()->GetSpeed());
-  return 0;
+    CDataCacheCore::GetInstance().SetSpeed(1.0, speed);
```
FernetMenta: @garbear btw: I won't use the generic IPlayer interface for this case where you know the type. you can implement a GetSpeed method for RP if you need one. GetSpeed in IPlayer is gone

xbmc/cores/RetroPlayer/RetroPlayerAutoSave.cpp
```diff
@@ -47,7 +48,7 @@ void CRetroPlayerAutoSave::Process()
     if (m_bStop)
       break;
 
-    if (m_player->GetSpeed() > 0.0f)
+    if (CDataCacheCore::GetInstance().GetSpeed() > 0.0f)
```
FernetMenta: @garbear btw: I won't use the generic IPlayer interface for this case where you know the type. you can implement a GetSpeed method for RP if you need one. GetSpeed in IPlayer is gone

garbear: @FernetMenta No problem, RetroPlayer has its own internal playback API. I've moved over to it in #12493

PR #12472, opened 2017-07-11
VideoPlayer: flush renderer on channel switch

FernetMenta:
long demanded by users. kicks out the last frame of last channel when switch to a new one.


FernetMenta:
@ksooo ok?

ksooo:
No chance to runtime test the next days, but I guess it's okay. If not, we can easily fix it later.

PR #12473, opened 2017-07-11
[addons] add blocked addon call event to prevent problems after disable

AlwinEsch:
## Description

This add a major part from me to handle binary addon changes in a secure way!

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/AddonInstaller.cpp
+++ b/xbmc/addons/AddonInstaller.cpp
@@ -604,11 +604,17 @@ bool CAddonInstallJob::DoWork()
   // run any pre-install functions
   ADDON::OnPreInstall(m_addon);
 
+  if (!CAddonMgr::GetInstance().UnloadAddon(m_addon))
+  {
+    CLog::Log(LOGERROR, "CAddonInstallJob[%s]: failed to unload addon.", m_addon->ID().c_str());
+    return false;
+  }
+
   // perform install
   if (!Install(installFrom, m_repo))
     return false;
 
-  if (!CAddonMgr::GetInstance().ReloadAddon(m_addon))
+  if (!CAddonMgr::GetInstance().LoadAddon(m_addon->ID()))
   {
     CLog::Log(LOGERROR, "CAddonInstallJob[%s]: failed to reload addon", m_addon->ID().c_str());
     return false;
--- a/xbmc/addons/AddonManager.cpp
+++ b/xbmc/addons/AddonManager.cpp
@@ -706,6 +706,10 @@ bool CAddonMgr::UnloadAddon(const AddonPtr& addon)
     if (m_cpluff->uninstall_plugin(m_cp_context, addon->ID().c_str()) == CP_OK)
     {
       CLog::Log(LOGDEBUG, "CAddonMgr: %s unloaded", addon->ID().c_str());
+
+      lock.Leave();
+      AddonEvents::Unload event(addon->ID());
+      m_unloadEvents.HandleEvent(event);
       return true;
     }
   }
@@ -736,6 +740,11 @@ bool CAddonMgr::LoadAddon(const std::string& addonId)
     CLog::Log(LOGERROR, "CAddonMgr: could not load add-on %s. No add-on with that ID is installed.", addon->ID().c_str());
     return false;
   }
+
+  lock.Leave();
+
+  AddonEvents::Load event(addon->ID());
+  m_unloadEvents.HandleEvent(event);
 
   if (IsAddonDisabled(addon->ID()))
   {
--- a/xbmc/addons/AddonManager.cpp
+++ b/xbmc/addons/AddonManager.cpp
@@ -706,8 +706,6 @@ bool CAddonMgr::UnloadAddon(const AddonPtr& addon)
     if (m_cpluff->uninstall_plugin(m_cp_context, addon->ID().c_str()) == CP_OK)
     {
       CLog::Log(LOGDEBUG, "CAddonMgr: %s unloaded", addon->ID().c_str());
-      m_events.Publish(AddonEvents::InstalledChanged());
-      m_events.Publish(AddonEvents::UnInstalled(addon->ID()));
       return true;
     }
   }
@@ -715,17 +713,16 @@ bool CAddonMgr::UnloadAddon(const AddonPtr& addon)
   return false;
 }
 
-bool CAddonMgr::ReloadAddon(AddonPtr& addon)
+bool CAddonMgr::LoadAddon(const std::string& addonId)
 {
   CSingleLock lock(m_critSection);
-  if (!addon ||!m_cpluff || !m_cp_context)
+  if (!m_cpluff || !m_cp_context)
     return false;
 
-  AddonPtr tmp;
-  bool isReinstall = GetAddon(addon->ID(), tmp, ADDON_UNKNOWN, false);
-  if (isReinstall)
+  AddonPtr addon;
+  if (GetAddon(addonId, addon, ADDON_UNKNOWN, false))
   {
-    m_cpluff->uninstall_plugin(m_cp_context, addon->ID().c_str());
+    return true;
   }
 
   if (!FindAddons())
@@ -734,22 +731,21 @@ bool CAddonMgr::ReloadAddon(AddonPtr& addon)
     return false;
   }
 
-  if (!GetAddon(addon->ID(), addon, ADDON_UNKNOWN, false))
+  if (!GetAddon(addonId, addon, ADDON_UNKNOWN, false))
   {
-    CLog::Log(LOGERROR, "CAddonMgr: could not reload add-on %s. No add-on with that ID is installed.", addon->ID().c_str());
+    CLog::Log(LOGERROR, "CAddonMgr: could not load add-on %s. No add-on with that ID is installed.", addon->ID().c_str());
     return false;
   }
 
-  if (isReinstall)
-    m_events.Publish(AddonEvents::ReInstalled(addon->ID()));
-
-  if (!EnableAddon(addon->ID()))
+  if (IsAddonDisabled(addon->ID()))
   {
-    CLog::Log(LOGERROR, "CAddonMgr: '%s' was installed but could not be enabled", addon->ID().c_str());
-    return false;
+    EnableAddon(addon->ID());
+    return true;
   }
 
-  CLog::Log(LOGDEBUG, "CAddonMgr: %s reloaded", addon->ID().c_str());
+  m_events.Publish(AddonEvents::ReInstalled(addon->ID()));
+  m_events.Publish(AddonEvents::InstalledChanged());
+  CLog::Log(LOGDEBUG, "CAddonMgr: %s successfully loaded", addon->ID().c_str());
   return true;
 }
 
@@ -758,6 +754,8 @@ void CAddonMgr::OnPostUnInstall(const std::string& id)
   CSingleLock lock(m_critSection);
   m_disabled.erase(id);
   m_updateBlacklist.erase(id);
+  m_events.Publish(AddonEvents::InstalledChanged());
+  m_events.Publish(AddonEvents::UnInstalled(id));
 }
 
 bool CAddonMgr::RemoveFromUpdateBlacklist(const std::string& id)
--- a/xbmc/addons/AddonManager.h
+++ b/xbmc/addons/AddonManager.h
@@ -178,7 +178,7 @@ namespace ADDON
      *
      * Returns true if the addon was successfully loaded and enabled; otherwise false.
      */
-    bool ReloadAddon(AddonPtr& addon);
+    bool LoadAddon(const std::string& addonId);
 
     /*! @note: should only be called by AddonInstaller
      *
--- a/xbmc/addons/AddonManager.h
+++ b/xbmc/addons/AddonManager.h
@@ -82,6 +82,7 @@ namespace ADDON
     virtual ~CAddonMgr();
 
     CEventStream<AddonEvent>& Events() { return m_events; }
+    CEventStream<AddonEvent>& UnloadEvents() { return m_unloadEvents; }
 
     IAddonMgrCallback* GetCallbackForType(TYPE type);
     bool RegisterAddonMgrCallback(TYPE type, IAddonMgrCallback* cb);
@@ -312,6 +313,7 @@ namespace ADDON
     CCriticalSection m_critSection;
     CAddonDatabase m_database;
     CEventSource<AddonEvent> m_events;
+    CBlockingEventSource<AddonEvent> m_unloadEvents;
     std::set<std::string> m_systemAddons;
     std::set<std::string> m_optionalAddons;
   };
--- a/xbmc/addons/AddonEvents.h
+++ b/xbmc/addons/AddonEvents.h
@@ -71,5 +71,17 @@ namespace ADDON
      * @deprecated Use Enabled, ReInstalled and UnInstalled instead.
      */
     struct InstalledChanged : AddonEvent {};
+
+    struct Load : AddonEvent
+    {
+      std::string id;
+      explicit Load(std::string id) : id(std::move(id)) {}
+    };
+
+    struct Unload : AddonEvent
+    {
+      std::string id;
+      explicit Unload(std::string id) : id(std::move(id)) {}
+    };
   };
 };
--- a/xbmc/addons/Service.cpp
+++ b/xbmc/addons/Service.cpp
@@ -50,6 +50,7 @@ CServiceAddonManager::CServiceAddonManager(CAddonMgr& addonMgr) :
 CServiceAddonManager::~CServiceAddonManager()
 {
   m_addonMgr.Events().Unsubscribe(this);
+  m_addonMgr.UnloadEvents().Unsubscribe(this);
 }
 
 void CServiceAddonManager::OnEvent(const ADDON::AddonEvent& event)
@@ -68,8 +69,25 @@ void CServiceAddonManager::OnEvent(const ADDON::AddonEvent& event)
   }
   else if (auto reinstallEvent = dynamic_cast<const AddonEvents::ReInstalled*>(&event))
   {
-    Stop(reinstallEvent->id);
     Start(reinstallEvent->id);
+  }
+
+  else if (auto e = dynamic_cast<const ADDON::AddonEvents::Unload*>(&event))
+  {
+    CLog::Log(LOGINFO, "CServiceAddonManager: Unloading %s", e->id.c_str());
+    CSingleLock lock(m_criticalSection);
+    Stop(e->id);
+    m_blacklistedAddons.push_back(e->id);
+  }
+  else if (auto e = dynamic_cast<const ADDON::AddonEvents::Load*>(&event))
+  {
+    CLog::Log(LOGINFO, "CServiceAddonManager: Removing %s from blacklist", e->id.c_str());
+    CSingleLock lock(m_criticalSection);
+    auto it = std::find(m_blacklistedAddons.begin(), m_blacklistedAddons.end(), e->id);
+    if (it != m_blacklistedAddons.end())
+    {
+      m_blacklistedAddons.erase(it);
+    }
   }
 }
 
@@ -92,6 +110,7 @@ void CServiceAddonManager::StartBeforeLogin()
 void CServiceAddonManager::Start()
 {
   m_addonMgr.Events().Subscribe(this, &CServiceAddonManager::OnEvent);
+  m_addonMgr.UnloadEvents().Subscribe(this, &CServiceAddonManager::OnEvent);
   VECADDONS addons;
   if (m_addonMgr.GetAddons(addons, ADDON_SERVICE))
   {
@@ -117,6 +136,12 @@ void CServiceAddonManager::Start(const AddonPtr& addon)
   if (m_services.find(addon->ID()) != m_services.end())
   {
     CLog::Log(LOGDEBUG, "CServiceAddonManager: %s already started.", addon->ID().c_str());
+    return;
+  }
+
+  if (std::find(m_blacklistedAddons.begin(), m_blacklistedAddons.end(), addon->ID()) != m_blacklistedAddons.end())
+  {
+    CLog::Log(LOGINFO, "CServiceAddonManager: Not executing blacklisted addon %s", addon->ID().c_str());
     return;
   }
 
--- a/xbmc/addons/Service.h
+++ b/xbmc/addons/Service.h
@@ -89,5 +89,7 @@ namespace ADDON
     CCriticalSection m_criticalSection;
     /** add-on id -> script id */
     std::map<std::string, int> m_services;
+
+    std::vector<std::string> m_blacklistedAddons;
   };
 }
--- a/xbmc/utils/EventStream.h
+++ b/xbmc/utils/EventStream.h
@@ -88,3 +88,18 @@ public:
     CJobManager::GetInstance().Submit(std::move(task));
   }
 };
+
+template<typename Event>
+class CBlockingEventSource : public CEventStream<Event>
+{
+public:
+  template<typename A>
+  void HandleEvent(A event)
+  {
+    CSingleLock lock(this->m_criticalSection);
+    for (const auto& subscription : this->m_subscriptions)
+    {
+      subscription->HandleEvent(event);
+    }
+  }
+};


tamland:
> How this parts on binary addon manager are called I do not really like it, but bring currently the best way. There is shortly the call blocked to give CBinaryAddonManager the time to start his unblocked calls and with parts on CAddonDll is guaranteed that only binary addons are called.

You already know this isn't acceptable, but if it's already non-blocking why even bother with this!? It makes no difference compared to events.

And you *still* haven't explained what the problem you are trying to solve is Alwin, so I still fail to see why any of this is needed and why binary addons need special treatment.

AlwinEsch:
As I have already written in the other.

"EventSource <AddonEvent>" allows only non-blocking calls. Even if a call for pre-uninstall or before addon update would be added, e.g. PVR does not have time to carry out his work. Its necessary files are nicely deleted although it still runs.

Furthermore, you must ensure that "CBinaryAddonManager" is the "addon" change before the associated systems.

Also, every call to "CEventSource <AddonEvent>" is in a new thread, even if now each called can set its own lock under circumstances the data of the addon have already changed again.

With most addon types (skins, languages, metadata ...) is not interested whether the related things have changed since they are only once loaded and after changes again.

Also, a Python addon may be more likely to live with it since it usually only background processes might make problems but Python can not direct bring Kodi to a crash.

PVR is the binary addon which brings most of the work, but these things can also affect other binary addon types.

Also, I noticed earlier that even with a larger auto update of script addons Kodi can crashen.

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -172,105 +413,68 @@ bool CBinaryAddonManager::AddAddonBaseEntry(BINARY_ADDON_LIST_ENTRY& entry)
...
+  msg->modal = modal;
+  m_control.SendOutMessage(CBinaryAddonDataProtocol::POST_INSTALL_EVENT, &msg, sizeof(MsgAddonPostInstall*));
+}
 
-  BinaryAddonMgrBaseList deletedAddons = m_installedAddons;
-  for (auto addon : binaryAddonList)
+void CBinaryAddonManager::OnPreUnInstallEvent(const std::string& addonId)
+{
+  Actor::Message *reply;
+  if (m_control.SendOutMessageSync(CBinaryAddonDataProtocol::PRE_UNINSTALL_EVENT, &reply, 5000, const_cast<char*>(addonId.c_str()), addonId.size()+1))
```
AlwinEsch: With 5000 the call is given 5 seconds of time, if it is finished faster it returns before, otherwise the wait is finished.

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -172,105 +413,68 @@ bool CBinaryAddonManager::AddAddonBaseEntry(BINARY_ADDON_LIST_ENTRY& entry)
...
-     *
-     * Currently only Screensaver and Visualization use the new way and not need informed.
-     */
+    CLog::Log(LOGERROR, "CBinaryAddonManager::%s - timed out", __FUNCTION__);
   }
 }
+
+void CBinaryAddonManager::OnPostUnInstallEvent(const std::string& addonId)
+{
+  m_control.SendOutMessage(CBinaryAddonDataProtocol::POST_UNINSTALL_EVENT, const_cast<char*>(addonId.c_str()), addonId.size()+1);
```
AlwinEsch: Like them is only non blocked for CAddonManager. But with the message can the CBinaryAddonManger do his work.

xbmc/peripherals/bus/virtual/PeripheralBusAddon.cpp
```diff
@@ -330,14 +355,6 @@ void CPeripheralBusAddon::GetDirectory(const std::string &strPath, CFileItemList
     addon->GetDirectory(strPath, items);
 }
 
-void CPeripheralBusAddon::OnEvent(const ADDON::AddonEvent& event)
-{
-  if (typeid(event) == typeid(ADDON::AddonEvents::Enabled) ||
-      typeid(event) == typeid(ADDON::AddonEvents::Disabled) ||
-      typeid(event) == typeid(ADDON::AddonEvents::InstalledChanged))
-    UpdateAddons();
```
AlwinEsch: This with "ADDON::AddonEvents::InstalledChanged" has not worked for updated (only install and deinstall).

A little bit of the problem of himself, but he compares installed addon id's with those already known to him.

Although the dll and the old addon are no longer present, he still thinks it would be there.

This way, the deleted dll continues nicely and the update is not used at all.

AlwinEsch: I know should be done now with new "ADDON::AddonEvents::ReInstalled".

But again, informated about reinstall after the old becomes deleted and still on memory in use.

tamland:
> "EventSource " allows only non-blocking calls. Even if a call for pre-uninstall or before addon update would be added, e.g. PVR does not have time to carry out his work. Its necessary files are nicely deleted although it still runs.

I hear you, but exactly what 'work' does PVR do that *must* be done prior to uninstall?


> Also, every call to "CEventSource " is in a new thread, even if now each called can set its own lock under circumstances the data of the addon have already changed again.

It does not use a new thread, it uses tasks. And this is an implementation detail of CEventSource.

> Also, I noticed earlier that even with a larger auto update of script addons Kodi can crash.

Which part is crashing? I would like to know this, and if so this is something we should fix

xbmc/peripherals/bus/virtual/PeripheralBusAddon.cpp
tamland: > This with "ADDON::AddonEvents::InstalledChanged" has not worked for updated (only install and deinstall).
>
> A little bit of the problem of himself, but he compares installed addon id's with those already known to him.

I know, but this is an issue with CPeripheralBusAddon::UpdateAddons only (and it's still broken) as it just looks at added/removed addon ids and does not account for updates. The fix is to remove this stuff from UpdateAddons and instead first unload the addon with that id completely, then try to load addon again, when an addon reinstall occur.

AlwinEsch:
>> "EventSource " allows only non-blocking calls. Even if a call for pre-uninstall or before addon update would be added, e.g. PVR does not have time to carry out his work. Its necessary files are nicely deleted although it still runs.

> I hear you, but exactly what 'work' does PVR do that *must* be done prior to uninstall?

Not only PVR, for every binary (actually even Python) must be ensured that it is not changed when it is running.

For most, it just does not matter. Since any addons mostly work at this moment, but this is never safe.

Most of the time it will also work because between two versions are often only small differences.

>> Also, every call to "CEventSource " is in a new thread, even if now each called can set its own lock under circumstances the data of the addon have already changed again.

> It does not use a new thread, it uses tasks. And this is an implementation detail of CEventSource.

CEventSource goes through CJobManager which in turn goes into a list of CJobWorker's as thread's, each in principle independent of the other.

>> Also, I noticed earlier that even with a larger auto update of script addons Kodi can crash.

> Which part is crashing? I would like to know this, and if so this is something we should fix

The update itself was done and then crashed. Was after the addon updates no more reproducable.

tamland:
>>> "EventSource " allows only non-blocking calls. Even if a call for pre-uninstall or before addon update would be added, e.g. PVR does not have time to carry out his work. Its necessary files are nicely deleted although it still runs.
>>>
>> I hear you, but exactly what 'work' does PVR do that *must* be done prior to uninstall?
>>
>Not only PVR, for every binary (actually even Python) must be ensured that it is not changed when it is running.
>
>For most, it just does not matter. Since any addons mostly not work at this moment, but this is never safe.
>
>Most of the time it will also work because between two versions are often only small differences.

Changing the files of a program that is linked or loaded into memory will not interfere with the running of said program.


>>> Also, every call to "CEventSource " is in a new thread, even if now each called can set its own lock under circumstances the data of the addon have already changed again.
>>
>> It does not use a new thread, it uses tasks. And this is an implementation detail of CEventSource.
>>
>CEventSource goes through CJobManager which in turn goes into a list of CJobWorker's as thread's, each in principle independent of the other.

I don't see what the issue is. From the subscribers perspective they come in one-by-one in order; why do you care if the underlying implementation use threads or tasks?

>>> Also, I noticed earlier that even with a larger auto update of script addons Kodi can crash.
>>>
>> Which part is crashing? I would like to know this, and if so this is something we should fix
>>
>The update itself was done and then crashed. But was after the addon updates no more reproducable.
EDIT: Thats a part I mean, it can even work, but not always.

I mean: which line?

AchimTuran:
@tamland
> Changing the files of a program that is linked or loaded into memory will not interfere with the running of said program.

This is not a clean implementation. What is when e.g. I uninstall an adsp add-on during a video file is playing? I can not estimate what an add-on does during shutdown because that is up to the developer. In my view we only should delete files if the add-on is safely released from Kodi and not before. This will break somehow.

FernetMenta:
I agree with @AlwinEsch and @AchimTuran. At the time an addon is updated, loaded, or unloaded there must absolutely no activity perfomed by this addon nor must the addon have any registrations into the application.

tamland:
If so, the way to do that is to use an unloading mechanism (which is already done in addon manager actually) where other services can register their blocking callbacks with the addon manager. Then you can choose to between using hot-reloading or not. All the other stuff isn't necessary.

AlwinEsch:
You mean the "virtual bool RequestRemoval(AddonPtr addon)=0;"? If yes is no more called since years.

I add a temporary commit to go with changed IAddonMgrCallback.

xbmc/addons/AddonManager.h
```diff
@@ -59,8 +59,18 @@ namespace ADDON
...
+      virtual bool RequestRestart(AddonPtr addon, bool datachanged) { return false; }
+      virtual bool RequestRemoval(AddonPtr addon) { return false; }
+
+      /* callbacks needed for CBinaryAddonManager */
+      virtual void EnableEvent(const std::string& addonId) {}
+      virtual void DisableEvent(const std::string& addonId) {}
+      virtual void PreInstallEvent(const std::string& addonId) {}
+      virtual void PostInstallEvent(const std::string& addonId, bool update, bool modal) {}
+      virtual void PreUnInstallEvent(const std::string& addonId) {}
+      virtual void PostUnInstallEvent(const std::string& addonId) {}
```
AlwinEsch: Is changed to use the IAddonMgrCallback instead of CAddonDll

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -55,6 +57,20 @@ bool CBinaryAddonManager::Init()
...
+  CServiceBroker::GetAddonMgr().RegisterAddonMgrCallback(ADDON_AUDIOENCODER, this);
+  //CServiceBroker::GetAddonMgr().RegisterAddonMgrCallback(ADDON_GAMEDLL, this);
+  CServiceBroker::GetAddonMgr().RegisterAddonMgrCallback(ADDON_IMAGEDECODER, this);
+  CServiceBroker::GetAddonMgr().RegisterAddonMgrCallback(ADDON_INPUTSTREAM, this);
+  CServiceBroker::GetAddonMgr().RegisterAddonMgrCallback(ADDON_PERIPHERALDLL, this);
+  //CServiceBroker::GetAddonMgr().RegisterAddonMgrCallback(ADDON_PVRDLL, this);
+  CServiceBroker::GetAddonMgr().RegisterAddonMgrCallback(ADDON_SCREENSAVER, this);
+  CServiceBroker::GetAddonMgr().RegisterAddonMgrCallback(ADDON_VFS, this);
+  CServiceBroker::GetAddonMgr().RegisterAddonMgrCallback(ADDON_VIDEOCODEC, this);
+  CServiceBroker::GetAddonMgr().RegisterAddonMgrCallback(ADDON_VIZ, this);
```
AlwinEsch: Register added here, the ADDON_GAMEDLL and ADDON_PVRDLL are commented out until this interfaces are changed.

tamland:
> You mean the "virtual bool RequestRemoval(AddonPtr addon)=0;"? If yes is no more called since years.
>
>I add a temporary commit to go with changed IAddonMgrCallback.

That wont work. The whole IAddonMgrCallback thing is broken (not thread safe), and really makes no sense as it allows anyone to invoke whatever callbacks they want at any time. For unloading to be safe it must 1. be done CAddonMgr::UnloadAddon such that there are no race conditions. 2. have thread safe unregistration and 3. not hold the addon manager lock while executing external callbacks. This is essentially what eventstream does except it's async. I'm sure it's possible to reuse some of that.

AlwinEsch:
It works, but that's exactly why I was previously gone over the unsightly way to CAddonDll.

"CBinaryAddonManager" and "CAddonManager" are siblings, so they should also be able to be so far.

For me it makes no sense are two different things (CEventSource and what else) to be used for one thing and see something as confused.

tamland:
> For me it makes no sense are two different things (CEventSource and what else) to be used for one thing and see something as confused.

Nobody's forcing you to use two different things. You are free to make something that is *better* and unifies them, or hide the details behind an api. But there *are* different things here and there *are* different needs, so see no point of pretending it isn't and forcing everything to adhere to the lowest common denominator.

Paxxi:
Why do you re-invent CApplicationMessenger using the actor implementation? Actor has it's uses, e.g a state machine but here it's used as a simple event/message transport and CApplicationMessenger is already designed for that.

I strongly disagree with this design.

AchimTuran:
@paxxi
Can you highlight the lines you mean? So far I only see that the ActorProtocol is used in the state machine. No message goes outside or did I miss something?

Paxxi:
@AchimTuran yes, the ActorProtocol is used to implement a simple messaging solution which we already have and adds a new thread to run it's own message loop.

There's no real state machine required here, it's just message passing and handling so I don't see why it should be invented again. We have enough with AppMessenger, observer, announcement manager, Actor. We don't need a new message implementation.

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -55,6 +57,20 @@ bool CBinaryAddonManager::Init()
   for (auto addon : binaryAddonList)
     AddAddonBaseEntry(addon);
 
+  CServiceBroker::GetAddonMgr().RegisterAddonMgrCallback(ADDON_ADSPDLL, this);
```
Paxxi: What's the point here registering for every type? The callback is stored in a map so there's only one possible listener for an addon type so why not just register all types in one method call?

Paxxi (changes requested):
Don't reinvent messaging systems, we have enough of them

if a "non-blocking" call has a 5 second timeout it's blocking and an ugly hack. This is not a good thing to have in a new implementation and needs to be designed properly.

AchimTuran:
> We have enough with AppMessenger, observer, announcement manager, Actor. We don't need a new message implementation.

That's exactly what @FernetMenta wrote when you introduced AppMessenger. :smile: 

But the ActorProtocol is used internally not outside of an active object.

> This is not a good thing to have in a new implementation and needs to be designed properly.

Sure new implementations should always designed properly.

xbmc/addons/binary-addons/BinaryAddonManager.cpp
AchimTuran: Well this is because @AlwinEsch tries to implement a new manager for binary add-ons as it was not wanted to merge them.

Paxxi:
I didn't introduce appmessenger, it was already there and in use. I designed it properly to not have a billion dependencies and made a plan to move it forward.

AlwinEsch:
Hi @Paxxi and @tamland,

Have the request updated and does not go beyond the IAddonMgrCallback or CAddonDll, CAddonManager now calls CBinaryAddonManager directly.

CBinaryAddonManager is just as critical as CAddonManager, so both must work directly with each other.

The other why about the "Actor :: Protocol", this brings the greatest security. Certainly, "CApplicationMessenger" would go better as "CEventSource" but in the case that an unblocked call might come several times in a very short time, which can only be managed with a "MutexLock". This is not directly required by the thread in the CBinaryAddonManager.

In addition, the thread in the future will also other things such as e.g. Monitor a running binary addon.

But I add a temporary commit, where I go through CApplicationMessenger.

AlwinEsch:
@Paxxi have added on end a temporary commit, if this is OK for you then I include it.

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -63,8 +66,180 @@ void CBinaryAddonManager::DeInit()
...
+          break;
+
+        base = addon->second;
+        m_installedAddons.erase(addonId);
+        m_enabledAddons.erase(addonId);
+      }
+
+      CLog::Log(LOGDEBUG, "CBinaryAddonManager::%s: Removing binary addon '%s'", __FUNCTION__, addonId);
+
+      for (auto &type : base->Types())
```
Paxxi: make this a method or lambda to avoid duplicating it

xbmc/addons/binary-addons/BinaryAddonManager.cpp
```diff
@@ -63,8 +66,180 @@ void CBinaryAddonManager::DeInit()
...
+      {
+        auto manager = m_managers.find(type.Type());
+        if (manager != m_managers.end())
+          manager->second->PreUnInstallEvent(base);
+      }
+      break;
+    }
+    case TMSG_BINADDON_POST_UNINSTALL:
+    {
+      BinaryAddonBasePtr base;
```
Paxxi: make this down to CLog a method or lambda to avoid duplicating it

xbmc/addons/binary-addons/BinaryAddonManager.h
```diff
@@ -36,15 +40,31 @@ namespace ADDON
   typedef std::shared_ptr<CBinaryAddonBase> BinaryAddonBasePtr;
   typedef std::vector<BinaryAddonBasePtr> BinaryAddonBaseList;
 
-  class CBinaryAddonManager
+  class IBinaryAddonManagerCallback
```
Paxxi: methods should be prefixed with On to make it clear they're events, also matches regular addons.

xbmc/addons/AddonManager.cpp
```diff
@@ -830,6 +831,7 @@ bool CAddonMgr::DisableAddon(const std::string& id)
   AddonPtr addon;
   if (GetAddon(id, addon, ADDON_UNKNOWN, false) && addon != NULL)
   {
+    CServiceBroker::GetBinaryAddonManager().DisableEvent(addon->Type(), addon->ID());
```
Paxxi: Not really sure why GetBinaryAddonManager()->*Event is called from here. Would make more sense to just Post/Send the events instead of going through an extra method to do it.

xbmc/peripherals/bus/virtual/PeripheralBusAddon.h
```diff
@@ -34,12 +34,18 @@
 
 namespace PERIPHERALS
 {
-  class CPeripheralBusAddon : public CPeripheralBus
+  class CPeripheralBusAddon : public CPeripheralBus, public ADDON::IBinaryAddonManagerCallback
```
Paxxi: I don't really like this. IMO nothing outside of CAddonManager should know that CBinaryAddonManager exists and there shouldn't be separate events or notifications depending on which type of addon is used. It completely breaks encapsulation and makes things very messy.

I suggest fixing the design so the events go through the existing CAddonManager events mechanism that we already have instead of trying to replace it.

Paxxi (changes requested):
I thinks it looks much better now but still plenty of design choices I'm wondering about and I don't think IBinaryAddonManagerCallback should exist, or possibly it could be used to notify CAddonManager about events but nothing else in Kodi should be exposed to it.

xbmc/peripherals/bus/virtual/PeripheralBusAddon.h
AlwinEsch: The CAddonManager events mechanism is not usable, it allow only non blocking calls. 

Further is wanted from @tamland to remove the old callback system with IAddonMgrCallback and the way over the addons (what is good) and go only over "CEventSource".

xbmc/addons/AddonManager.cpp
AlwinEsch: Will the CBinaryAddonManager define itself what it does. In case of way changes need only the related manager updated.

AlwinEsch:
Is updated

AlwinEsch:
@Paxxi @tamland is rebased and updated, whats your opinion? Is not perfect, but it works.

To become really good need the complete addon system a improvment.
.

tamland:
Ok Im not sure this is going anywhere and I don't see that anything has really changed so I'll try to explain this again. It is still duplicating the whole state system that exist in AddonMgr. BinaryAddonManager have no business, not does it need to know about *any* of these install/enable/disable event. All it needs to know to achieve this in a safe way is 1. when the addon should be destroyed and 2. when it's safe to create. The other events doesn't make any sense and aren't needed by anything.

Additionally, calling `PreInstallEvent` from `AddonInstaller` makes the whole thing even *less* safe (it must be done in the AddonMgr lock to be safe). And having AddonMgr, AddonInstaller and other 'core' components call BinaryAddonManager goes against the architecture. It should be the other way around.

As state previously the way to fix this is to:
1 Keep all states and authority when an addon is considered 'installed' and not installed etc. in AddonMgr like it is currently.
2 Add a registration mechanism for BinaryAddonManager to register with AddonMgr.
3 Have AddonMgr make a blocking call to the `Unload` method on the callback. The opposite (`Load`) can still be async.

ksooo:
omg. I feel like for Kodi 18 ++ I will have to get used with pvr (and other binary) addons crashing while being updated... :-(

AlwinEsch:
As it is now, it is far from good, but as long as the old AddonManager is so because there are not many alternatives for a safe way.

In order to have it really good, the current CAddonManager would have to be completely rebuilt.

FernetMenta:
@tamland why don't you come up with an alternative suggestion implemented as executable code. Then we vote the better one. I agree with @ksooo and would rather take this than waiting for the perfect solition.

peak3d:
@tamland if I understand right you prefer an AddonManager wich handles both binary and other addons in the same way (I think we all agree that this is the best way).

This would mean a removal of the current BinaryAddonManager, but needs some abstraction for the special cases binary addons have (e.g. secure finalization on remove).
 
Reading your comments above it seems that you have now a complete understanding about the things @AlwinEsch  wants to archieve.  Would it be an idea if you write some pseudo-code in AddonManager just to make clear where what to add and @AlwinEsch continues with making real code from pseudo code?
IMO most time is spend here because of misunderstandings caused of the complexity of this topic here.

tamland:
Here's a quick but working implementation of the necessary changes for addon manager, and an implementation for service addons to demonstrate (since those are already in a fully working state): https://github.com/xbmc/xbmc/compare/master...tamland:addon_unload

How you actually prevent the addons from being interacted with while being updated will depend though. For services, everything goes via CServiceAddonManager so preventing it should be sufficient to (globally) ensure it can't be executed. This can probably be done in numerous different ways though.

AlwinEsch:
Thanks @tamland on the first view it looks good. 

Is guaranteed if CBinaryAddonManager register as first that it becomes called before addons?

AlwinEsch:
@tamland do you bring your changes in own request or that I take to them?

tamland:
@AlwinEsch Feel free to do it as I don't have a dev computer available right now due to hard drive failure (and I'm out for at least another week)

AlwinEsch:
@tamland your commits are added here

AlwinEsch:
jenkins build this please

AlwinEsch:
jenkins build this please

xbmc/addons/AddonInstaller.cpp
```diff
@@ -604,11 +604,17 @@ bool CAddonInstallJob::DoWork()
   // run any pre-install functions
   ADDON::OnPreInstall(m_addon);
 
+  if (!CAddonMgr::GetInstance().UnloadAddon(m_addon))
+  {
+    CLog::Log(LOGERROR, "CAddonInstallJob[%s]: failed to unload addon.", m_addon->ID().c_str());
+    return false;
```
stefansaraev: it is impossible to install addons now:

```
21:00:45.707 T:546777977296   ERROR: CAddonMgr: failed to unload plugin.video.youtube
21:00:45.708 T:546777977296   ERROR: CAddonInstallJob[plugin.video.youtube]: failed to unload addon.
```

AlwinEsch: I look to fix later.

tamland:
@AlwinEsch This last commit (8bd7213) shouldn't have been included. This was just an example of how to use it..

AlwinEsch:
Ah sorry, you marked here https://github.com/xbmc/xbmc/compare/master...tamland:addon_unload the last with `tmp: test that it's working` and not seen https://github.com/xbmc/xbmc/commit/8bd7213d2ec081717f3c3f4fc0ed43eea843e019 as example too

PR #12357, opened 2017-06-25
[WARNING Do not merge] ADD: GLES HQ scalers

koying:
[WARNING Jarvis code]

On request of @fritsch , the HQ scaler code from SPMC jarvis for reference / discussion
Not yet backported to Keypton or Leia.

/cc @peak3d @lrusak

PR #12474, opened 2017-07-12
LinuxRendererGLES: implement hq scalers

lrusak:
taken from @koying PR #12357 

This is not working properly yet. It can run with the fbo but the video does not get scaled properly, AKA. a 1280x720 video will only be 1280 pixels wide in a 1920 display. Everything including the gui isn't scaled properly so maybe that is a clue.

I haven't been able to track down the problem yet.


lrusak:
hmm, this actually seems to be working ok. I think it could be an issue with GBM.

I can reset the video size by pressing `\` twice. So I think something isn't getting reset properly. Would be nice if someone could test this on another platform.

xbmc/cores/VideoPlayer/VideoRenderers/FrameBufferObject.cpp
```diff
@@ -53,10 +53,14 @@ CFrameBufferObject::CFrameBufferObject()
 
 bool CFrameBufferObject::IsSupported()
 {
+#if HAS_GLES >= 2
```
FernetMenta: no need for ifdef here, ask windowing

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoFilterShader.cpp
```diff
@@ -75,12 +76,70 @@ BaseVideoFilterShader::BaseVideoFilterShader()
     "gl_FragColor.a = gl_Color.a;"
     "}";
   PixelShader()->SetSource(shaderp);
+#elif HAS_GLES >= 2
```
FernetMenta: time to cleanup the ifdef hell here instead of making it worse

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoFilterShader.cpp
```diff
@@ -101,19 +160,25 @@ ConvolutionFilterShader::ConvolutionFilterShader(ESCALINGMETHOD method, bool str
     if (m_floattex)
       m_internalformat = GL_RGBA16F_ARB;
     else
-#endif
       m_internalformat = GL_RGBA;
+#elif HAS_GLES >= 2
```
FernetMenta: the ifdef here is a hack too

xbmc/cores/VideoPlayer/VideoRenderers/FrameBufferObject.cpp
lrusak: unfortunately there is no `GL_EXT_framebuffer_object` in GLES so I'm not sure what to do.

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoFilterShader.cpp
```diff
@@ -152,6 +217,8 @@ ConvolutionFilterShader::~ConvolutionFilterShader()
 
 void ConvolutionFilterShader::OnCompiledAndLinked()
 {
+  BaseVideoFilterShader::OnCompiledAndLinked();
```
FernetMenta: hack, this is just an empty method for i.e. opengl

xbmc/cores/VideoPlayer/VideoRenderers/FrameBufferObject.cpp
FernetMenta: the method takes a string as argument.

pkerling:
@lrusak just cherry-picked this for wayland (pkerling/xbmc@1da938569), same problem - so it's not gbm. whole UI gets drawn at the video resolution when I switch to a HQ scaler. Switching back does not fix it.

xbmc/cores/VideoPlayer/VideoRenderers/FrameBufferObject.cpp
koying: FBO's are core to gles 2.0, no extensions required.

FernetMenta: then return true in windowing if this one is queried.

koying: Mmm... Returning true to a non-existing extension by if'ing an hardcoded value?
You *really* hate IFDEF's ;)

FernetMenta:
btw: you get a much better performance if you scale the yuv image first, then do conversion to rgb.

xbmc/cores/VideoPlayer/VideoRenderers/FrameBufferObject.cpp
FernetMenta: so why do you call IsSupported on the fbo if you know it is supported?

koying: https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/VideoRenderers/FrameBufferObject.cpp#L65

FernetMenta: ahh. this check is nonsense and should be removed. opengl can test for support before calling initialize() and log a reasonable message other than "error". gles does not need to test for support.

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGLES.cpp
```diff
@@ -873,7 +873,7 @@ void CLinuxRendererGLES::RenderToFBO(int index, int field, bool weave /*= false*
       return;
     }
     //if (!m_fbo.fbo.CreateAndBindToTexture(GL_TEXTURE_2D, m_sourceWidth, m_sourceHeight, GL_RGBA))
-    if (!m_fbo.fbo.CreateAndBindToTexture(GL_TEXTURE_2D, m_sourceWidth, m_sourceHeight, GL_RGBA, GL_SHORT))
+    if (!m_fbo.fbo.CreateAndBindToTexture(GL_TEXTURE_2D, m_sourceWidth, m_sourceHeight, GL_RGBA16_EXT, GL_SHORT))
```
koying: Looks like it requires gles3.1: https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_norm16.txt

koying:
@lrusak You miss https://github.com/xbmc/xbmc/pull/12357/files#diff-41e0ceb3de6f45669f617deae0dbeb0cR1638 for whatever reason.
That's the cause of the "corner" issue.

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGLES.cpp
lrusak: hmm, this is included in `GLES2/gl2ext.h` from mesa. If it is a problem I can drop it.

koying: Well just check the jenkins failures ;)
Didn't dig deep, though. Just googled the symbol, and saw it's 3.1

lrusak:
@koying thanks, that indeed fixes the issue.

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
```diff
@@ -1109,6 +1109,12 @@ void CLinuxRendererGL::RenderToFBO(int index, int field, bool weave /*= false*/)
 
   if (!m_fbo.fbo.IsValid())
   {
+    if (!g_Windowing.IsExtSupported("GL_EXT_framebuffer_object"))
```
FernetMenta: just leave the fbo untouched if not supported. trying to initialize on platforms that don't support it makes no sense

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoFilterShader.h
```diff
@@ -36,12 +36,20 @@ namespace Shaders {
...
+    virtual void OnCompiledAndLinked();
+    virtual bool OnEnabled();
+    virtual void Free() { CGLSLShaderProgram::Free(); }
     virtual void SetSourceTexture(GLint ytex) { m_sourceTexUnit = ytex; }
     virtual void SetWidth(int w) { m_width  = w; m_stepX = w>0?1.0f/w:0; }
     virtual void SetHeight(int h) { m_height = h; m_stepY = h>0?1.0f/h:0; }
     virtual void SetNonLinStretch(float stretch) { m_stretch = stretch; }
     virtual bool GetTextureFilter(GLint& filter) { return false; }
+#if HAS_GLES >= 2
+    virtual GLint GetVertexLoc() { return m_hVertex; }
```
FernetMenta: maybe you shoud create a new class for gles that inherits from this one.

lrusak:
Updated, this now works properly with all scalers. This should be tested with GL also as it changes the format used for the 6x6 convolutions to use one with an alpha channel.

FernetMenta:
before testing with openGL you should address my comments. It won't have a change to get in with those ifdefs.

lrusak:
@FernetMenta yes of course. This is still very much a WIP. I'd like to clean it up as much as possible.

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGLES.cpp
```diff
@@ -803,11 +855,262 @@ void CLinuxRendererGLES::RenderSinglePass(int index, int field)
...
+  pYUVShader->SetBlack(CMediaSettings::GetInstance().GetCurrentVideoSettings().m_Brightness * 0.01f - 0.5f);
+  pYUVShader->SetContrast(CMediaSettings::GetInstance().GetCurrentVideoSettings().m_Contrast * 0.02f);
+  pYUVShader->SetWidth(planes[0].texwidth);
+  pYUVShader->SetHeight(planes[0].texheight);
+  pYUVShader->SetNonLinStretch(1.0);
+  if (field == FIELD_TOP)
+    pYUVShader->SetField(1);
+  else if(field == FIELD_BOT)
+    pYUVShader->SetField(0);
+
```
koying: You need `pYUVShader->SetConvertFullColorRange(!g_Windowing.UseLimitedColor());` here
The YUV shader default to limited those days...

lrusak:
@FernetMenta I've updated to remove much of the ifdefs. I hope that is better!

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGLES.cpp
lrusak: :+1:

lrusak:
jenkins build this please

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoFilterShader.h
```diff
@@ -57,11 +57,41 @@ namespace Shaders {
     GLint m_hStretch = 0;
   };
 
+  class BaseVideoFilterShaderGLES : public BaseVideoFilterShaderGL
```
FernetMenta: this is not a good idea. I consider OpenGL reference and don't want ports like GLES to break when the reference changes.

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoFilterShader.h
```diff
@@ -57,11 +57,41 @@ namespace Shaders {
...
+    GLint m_hModel;
+    GLint m_hAlpha;
+
+    GLfloat *m_proj;
+    GLfloat *m_model;
+    GLfloat  m_alpha;
+  };
+
+#if HAS_GLES >= 2
+  using BaseVideoFilterShader = BaseVideoFilterShaderGLES;
```
FernetMenta: I don't like this. That's kind of the same hack the fist GLES rendering fo vaapi had shown.

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoFilterShader.cpp
```diff
@@ -77,49 +77,116 @@ BaseVideoFilterShader::BaseVideoFilterShader()
   PixelShader()->SetSource(shaderp);
 }
 
+BaseVideoFilterShaderGLES::BaseVideoFilterShaderGLES()
```
FernetMenta: the diff between OpenGL and GLES gets too big for keeping both implementations in the same file

lrusak:
I'm not going to split the file. I think that is counterproductive. If OpenGL breaks OpenGLES then it will get fixed. If it was split then OpenGLES will be forgotten about.

I'm not happy about the VAAPI change but it was the only way forward. The `RendererVAAPIGL.cpp` and `RendererVAAPIGLES.cpp` are the exact same, there is no reason for them to be split other than the fact that that's how you wanted them to be.

In my opinion the path forward is to make the OpenGL code compatible with OpenGLES. This should be done by only using OpenGLES compatible calls.

FernetMenta:
Without splitting the file this won't get merged. No discussion about this. You can do by yourself or wait until I'll find time to do so. The latter will delay this PR.

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
```diff
@@ -1272,35 +1279,48 @@ void CLinuxRendererGL::RenderFromFBO()
...
 
-  glBegin(GL_QUADS);
+  GLubyte idx[4] = {0, 1, 3, 2};        //determines order of triangle strip
+  GLfloat vert[4][3];
+  GLfloat tex[4][2];
 
-  glMultiTexCoord2fARB(GL_TEXTURE0, 0.0f, 0.0f);
-  glVertex4f(m_rotatedDestCoords[0].x, m_rotatedDestCoords[0].y, 0, 1.0f );
+  GLint vertLoc = m_pVideoFilterShader->GetVertexLoc();
+  GLint loc     = m_pVideoFilterShader->GetcoordLoc();
```
FernetMenta: no vertical alignment

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
```diff
@@ -1272,35 +1279,48 @@ void CLinuxRendererGL::RenderFromFBO()
...
+  // Setup vertex position values
+  for(int i = 0; i < 4; i++)
+  {
+    vert[i][0] = m_rotatedDestCoords[i].x;
+    vert[i][1] = m_rotatedDestCoords[i].y;
+    vert[i][2] = 0.0f;// set z to 0
+  }
+
+  // Setup texture coordinates
+  tex[0][0] = tex[3][0] = 0.0f;
```
FernetMenta: put every assignment on a separate line

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoFilterShader.cpp
```diff
@@ -56,70 +56,106 @@ BaseVideoFilterShader::BaseVideoFilterShader()
 
   m_stretch = 0.0f;
 
+  m_hVertex = -1;
+  m_hcoord = -1;
+  m_hProj   = -1;
```
FernetMenta: no verticl alignment

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoFilterShader.cpp
```diff
@@ -56,70 +56,106 @@ BaseVideoFilterShader::BaseVideoFilterShader()
 
   m_stretch = 0.0f;
 
+  m_hVertex = -1;
```
FernetMenta: shozld be done in header, not in constructor (C++11 style)

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoFilterShader.cpp
```diff
@@ -56,70 +56,106 @@ BaseVideoFilterShader::BaseVideoFilterShader()
...
-    "}";
+      " attribute vec4 m_attrpos;"
+      " attribute vec2 m_attrcord;"
+      " varying vec2 cord;"
+      " uniform mat4 m_proj;"
+      " uniform mat4 m_model;"
+
+      " void main ()"
+      " {"
+      "   mat4 mvp    = m_proj * m_model;"
```
FernetMenta: no vertial alignment

lrusak:
Pushed some fixes so gl works like gles for the videofiltershader. I have tested and it is working.

@FernetMenta I addressed some of your concerns and will fix the rest soon.

lrusak:
Rebased, should make for an easier review

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
```diff
@@ -1248,6 +1254,7 @@ void CLinuxRendererGL::RenderFromFBO()
 {
   glEnable(GL_TEXTURE_2D);
   glActiveTextureARB(GL_TEXTURE0);
+  glBindTexture(GL_TEXTURE_2D, m_fbo.fbo.Texture());
```
FernetMenta: that looks wrong. we render from fbo that is loaded into texture units. what is your intent?

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
```diff
@@ -1278,35 +1280,52 @@ void CLinuxRendererGL::RenderFromFBO()
...
-    glTexEnvf(GL_TEXTURE_ENV, GL_TEXTURE_ENV_MODE, GL_MODULATE);
   }
 
   VerifyGLState();
 
   float imgwidth = m_fbo.width / m_sourceWidth;
   float imgheight = m_fbo.height / m_sourceHeight;
 
-  glBegin(GL_QUADS);
+  GLubyte idx[4] = {0, 1, 3, 2};        //determines order of triangle strip
```
FernetMenta: keep GL_QUADS or provide performance measures that justifiy the change

FernetMenta:
if you want to speed this up, drop all OpenGL related changes and focus only on what the title of this PR says.

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoFilterShader.h
```diff
@@ -55,13 +60,23 @@ namespace Shaders {
...
+    GLfloat *m_model;
+    GLfloat  m_alpha;
   };
 
   class ConvolutionFilterShader : public BaseVideoFilterShader
   {
   public:
     ConvolutionFilterShader(ESCALINGMETHOD method, bool stretch, GLSLOutput *output=NULL);
-    ~ConvolutionFilterShader() override;
+    virtual ~ConvolutionFilterShader() override;
```
garbear: clang's sanitizer will strip leading virtuals, so the next time someone does this dropping virtual will make their lives easier

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
lrusak: It is present in the GLES renderer however it doesn't seem to effect anything with or without it so I removed it now.

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
lrusak: I am not sure how to measure the performance here. I can only provide some links:

https://stackoverflow.com/questions/14555446/opengl-is-it-more-efficient-to-use-gl-quads-or-gl-triangles
https://stackoverflow.com/questions/6644099/what-is-so-bad-about-gl-quads

lrusak:
The changes to GL are only so I don't have to ifdef the hell out of everything.

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoFilterShader.h
lrusak: thanks, I don't know how that ended up there in the first place :)

FernetMenta:
this won't get enywhere if you touch GL. I insist to keep GL_QUADS as it is used in the other render methods. Create a separate videofilter for GLES as I already suggested.

lrusak:
```
lukas@lukas-macbook-pro ~/Documents/git/xbmc$ find xbmc/ -type f | xargs grep GL_QUADS 
xbmc/pictures/SlideShowPicture.cpp:  glBegin(GL_QUADS);
xbmc/guilib/GUIFontTTFGL.cpp:  glDrawArrays(GL_QUADS, 0, m_vertex.size());
xbmc/guilib/GUITextureGL.cpp:  glBegin(GL_QUADS);
xbmc/guilib/GUITextureGL.cpp:  glBegin(GL_QUADS);
xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp:  glBegin(GL_QUADS);
xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp:  glBegin(GL_QUADS);
xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp:  glBegin(GL_QUADS);
xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp:  glBegin(GL_QUADS);
xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp:  glBegin(GL_QUADS);
xbmc/cores/VideoPlayer/VideoRenderers/OverlayRendererGL.cpp:  glDrawArrays(GL_QUADS, 0, m_count * 4);
xbmc/cores/VideoPlayer/VideoRenderers/OverlayRendererGL.cpp:  glBegin(GL_QUADS);
```

I really would rather avoid code duplication again. I mean look at the vaapi example. http://sprunge.us/DMCD

FernetMenta:
this is the way to go, I want to keep this separately. some duplicate lines of code it not an issue. different maintainers, different files.

lrusak:
Updated to conform to @FernetMenta 's requests.

FernetMenta (approved):

lrusak:
jenkins build this please

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/VideoFilterShaderGLES.cpp
```diff
@@ -0,0 +1,302 @@
...
+  }
+  else if (m_method == VS_SCALINGMETHOD_SPLINE36 ||
+           m_method == VS_SCALINGMETHOD_LANCZOS3)
+  {
+    shadername = "convolution-6x6.glsl";
+  }
+
+  if (m_floattex)
+  {
+    m_internalformat = GL_RGBA16F_EXT;
```
Rechi: fails on RBPI beacause `GL_RGBA16F_EXT` isn't declared

FernetMenta: RBPI builds should not include this file

lrusak: why not?

FernetMenta: it only uses its own renderer

lrusak:
I'm assuming android, rbpi, aml, and imx don't use `LinuxRendererGLES` can anyone confirm?

FernetMenta:
android does use it: https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererMediaCodec.h#L29

koying:
> I'm assuming android, rbpi, aml, and imx don't use LinuxRendererGLES can anyone confirm?

Well, all GLES platforms use it in ffmpeg/software mode, really (besides maybe rpi)

lrusak:
>Well, all GLES platforms use it in ffmpeg/software mode, really (besides maybe rpi)

You're right, if I disable mediacodec on android I can use the HQ scaler methods.

I've updated to not build only on RPi

xbmc/cores/VideoPlayer/VideoRenderers/CMakeLists.txt
```diff
@@ -39,10 +39,7 @@ if(OPENGL_FOUND)
                       FrameBufferObject.h)
 endif()
 
-if(OPENGLES_FOUND AND (NOT CORE_PLATFORM_NAME STREQUAL rbpi AND
-                       NOT CORE_PLATFORM_NAME STREQUAL aml AND
-                       NOT CORE_PLATFORM_NAME STREQUAL android AND
-                       NOT CORE_PLATFORM_NAME STREQUAL imx))
+if(OPENGLES_FOUND AND NOT CORE_PLATFORM_NAME STREQUAL rbpi)
```
FernetMenta: actually this does not depend on whether GLES was found or not. it depends on if the platform depends of GLES renderer or not. means the logic here is not correct.

lrusak: Ok I revised it to just depend on which platforms need it

xbmc/cores/VideoPlayer/VideoRenderers/CMakeLists.txt
```diff
@@ -39,7 +39,12 @@ if(OPENGL_FOUND)
                       FrameBufferObject.h)
 endif()
 
-if(OPENGLES_FOUND)
+if(CORE_PLATFORM_NAME STREQUAL android OR
+   CORE_PLATFORM_NAME STREQUAL ios OR
+   CORE_PLATFORM_NAME STREQUAL gbm OR
+   CORE_PLATFORM_NAME STREQUAL imx OR
+   CORE_PLATFORM_NAME STREQUAL mir OR
+   CORE_PLATFORM_NAME STREQUAL wayland)
```
FernetMenta: wayland can use OpenGL or GLES

lrusak: I really don't know what you want. In this case I should leave the OPENGLES_FOUND statement,

FernetMenta: I just point out (once again) that our build system does not fit our requirements. The current issues you encounter are a clear indicator for this. 

Ideally a platform maintainer should be able to include platform dependent code in the build. The platform maintainer is the only one how really knows that.

lrusak:
Updated with the best possible solution that our build system provides.

lrusak:
Rebased and squashed. Would be nice if I could get an ok from other GLES platform maintainers. ping @koying @Memphiz @peak3d

Memphiz:
I have not the slightest amount of time for testing this currently. But given the fact that the last test definitely didn't break anything, and I am pretty sure this is still the case in the current state of this PR - you have my OK ;)

lrusak:
jenkins build this please

lrusak:
jenkins build this please

Rechi:
jenkins build this please

xbmc/cores/VideoPlayer/VideoRenderers/CMakeLists.txt
```diff
@@ -39,7 +39,12 @@ if(OPENGL_FOUND)
                       FrameBufferObject.h)
 endif()
 
-if(OPENGLES_FOUND)
+if(OPENGLES_FOUND AND (CORE_PLATFORM_NAME STREQUAL android OR
+                       CORE_PLATFORM_NAME STREQUAL ios OR
+                       CORE_PLATFORM_NAME STREQUAL gbm OR
+                       CORE_PLATFORM_NAME STREQUAL imx OR
+                       CORE_PLATFORM_NAME STREQUAL mir OR
+                       CORE_PLATFORM_NAME STREQUAL wayland))
```
Rechi: use `CORE_PLATFORM_NAME_LC` instead of `CORE_PLATFORM_NAME`
see comment at https://github.com/xbmc/xbmc/blob/4dfc7a8/cmake/scripts/common/Platform.cmake#L13-L16

lrusak: ah right, thanks

lrusak:
I don't see any reason why rbpi should be failing, it should not even be building that code

lrusak:
squshed, rbpi built successfully. Ready to merge :+1:

Memphiz:
Thx for getting it through :)

PR #12475, opened 2017-07-12
[VideoPlayer] WinRenderer: re-work video rendering

afedchin:
* This adds LUT and dithering support for windows,
* Re-factor render buffers
* Re-factor rendering methods.

@FernetMenta please review in part of work with buffers


xbmc/cores/VideoPlayer/VideoRenderers/ColorManager.h
```diff
@@ -24,7 +24,12 @@
 #include "lcms2.h"
 #endif
 
-#include <string>
```
Rechi: causes build errors

afedchin: oops. thanks

xbmc/video/dialogs/GUIDialogCMSSettings.cpp
```diff
@@ -20,7 +20,7 @@
 
 #include "system.h"
 
-#ifdef HAS_GL
+#if defined(HAS_GL) || defined(HAS_DX)
```
FernetMenta: we should do this with cmake, only include where needed.

xbmc/cores/VideoPlayer/VideoRenderers/WinRenderBuffer.cpp
```diff
@@ -0,0 +1,647 @@
...
+
+#define PLANE_Y 0
+#define PLANE_U 1
+#define PLANE_V 2
+#define PLANE_UV 1
+#define PLANE_D3D11 0
+
+CRenderBuffer::CRenderBuffer()
+  : loaded(false)
+  , frameIdx(0)
```
FernetMenta: ctor initialization is hard to maintain. much better is c++11 syle by assignment in header
your code, your decision

xbmc/cores/VideoPlayer/VideoRenderers/WinRenderer.cpp
```diff
@@ -911,38 +979,28 @@ bool CWinRenderer::RenderCapture(CRenderCapture* capture)
 //********************************************************************************************************
 // YV12 Texture creation, deletion, copying + clearing
 //********************************************************************************************************
-void CWinRenderer::DeleteYV12Texture(int index)
+void CWinRenderer::DeleteRenderBuffer(int index)
 {
   CSingleLock lock(g_graphicsContext);
   ReleaseBuffer(index);
-  SAFE_DELETE(m_VideoBuffers[index]);
+  m_renderBuffers[index].Release();
```
FernetMenta: that is exaclty what line 985 does, isn't it?

FernetMenta (approved):
this is a rather big change and I don't claim to understand the details. I looked at the buffer handling as requested. looks good

xbmc/cores/VideoPlayer/VideoRenderers/WinRenderer.cpp
afedchin: it's not the same. `ReleaseBuffer` releases video buffer only. `Release` method releases render buffer resources. perhaps need to change name for better understanding.

xbmc/video/dialogs/GUIDialogCMSSettings.cpp
afedchin: yes, but I'm not original author so I left this as is

FernetMenta: the define was here before cmake. now things can be improved when they are touched anyway.

afedchin:
@FernetMenta please review https://github.com/xbmc/xbmc/pull/12475/commits/f98e9ef4f5223482af71e35798e236bf4053b4a9 because you are the author of an idea of surface context

FernetMenta:
@afedchin yes, that makes sense. An alternative for MarkRender/MarkDecode is to grab a reference on the ffmpeg buffer in CDecoder::Decode(). Then ffmpeg won't return the view. See VTB for an example:
https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDCodecs/Video/VTB.cpp#L236

afedchin:
@FernetMenta thanks. with this approach dxva pool will look much simpler

afedchin:
build error is unrelated.

MartijnKaijser:
All green :)

jchapple:
When using this with an Intel NUC7i7BN on a non-HDR TV to decode 2160 24hz HDR content, using a generic `bt2020to709-4000to300-0.15blacklevel-0offset` 3DLUT file, the resulting video image is completely distorted, green and noisy.

The same content with the same 3DLUT file works well for HDR->SDR on a different rig with a GTX 970.

PR #12477, opened 2017-07-12
 [depends] samba-gplv3: fix large offset

Rechi:
## Description
Android NDK doesn't support `LARGE_OFFSET` and therefore `off_t` has only a width of 32 bit. This Replaces every occurrence of `off_t` with `off64_t` (64 bit with) for android.

## Motivation and Context
Kodi on android shows wrong filesize for samba shares if size is over 2^32 bytes because of the limited off_t width.
I missed some `off_t` at the samba bump.

## How Has This Been Tested?
confirmed working by @da-anda

## Types of change

## Checklist:


da-anda:
as mentioned, I can confirm this fixes the issues I noticed. Not only the file sizes had been displayed wrong, but also playback of files >  2^32 bytes was not possible, because anything beyond that point could not be read by ffmpeg/videoplayer, and thus playback got aborted.

koying (approved):

wsnipex (approved):

Rechi:
osx failure not related as this only touches android

PR #12478, opened 2017-07-12
Websocket connection header is case-insensitive

jeremyvisser:
Cherry-picking commit 34286426ce23509bb7215b625f62728ffa81f8c5 from master into Krypton.

This is a relatively benign bugfix which will ease the pain for users with reverse proxies attempting to forward websockets.


Rechi:
backports #11267

Rechi:
only unsupported platforms for v17 failed

PR #12479, opened 2017-07-12
Support more yuv420p bitwidths for direct rendering on Linux

yol:
Add support for yuv420p9, yuv420p12, and yuv420p14 to the linux OpenGL renderer on X11.

## Description
Kodi on Linux only supports yuv420p8, yuv420p10, and yuv410p16. If the video is encoded e.g. with yuv420p12, ffmpeg will automatically perform the color space in software, which costs CPU.
The ffmpeg log commit is related because it makes it easier to see what is actually going on and if the right format hits the screen. Can make a seperate PR if needed though.

## Motivation and Context
Some HEVC encodes already use yuv420p12. Support for the other bitwidths was added because it's very easy to support anyway.

## How Has This Been Tested?
Run-tested on X11 and Wayland.

## Types of change

## Checklist:


FernetMenta (approved):

FernetMenta:
jenkins errors not related

xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecFFmpeg.cpp
```diff
@@ -1146,6 +1146,13 @@ int CDVDVideoCodecFFmpeg::FilterOpen(const std::string& filters, bool scale)
     return result;
   }
 
+  char* graphDump = avfilter_graph_dump(m_pFilterGraph, nullptr);
+  if (graphDump)
```
FernetMenta: @pkerling this should only be dumped if component logging for video is turned on. the filter my get drained before ffmpeg and in this case it gets reopened a couple of times before eof. -> results in log spam

pkerling: sure, easy fix

PR #12480, opened 2017-07-12
VideoPlayer: set some process infos for the case get_format gets not …

FernetMenta:
see title

credits to @pkerling for bring this up


pkerling:
Can confirm this fixes the problem I saw. Thanks!

PR #12482, opened 2017-07-12
[PVR] Fix CPVRTimerInfoTag::GetEpgInfoTag for timer rules.

ksooo:
The method did not always work for timer rules and returned sometimes epg tags not related to the rule.

Runtime-tested with latest master on macOS.

@Jalle19 mind taking a look?

### DIFF
--- a/xbmc/pvr/timers/PVRTimerInfoTag.cpp
+++ b/xbmc/pvr/timers/PVRTimerInfoTag.cpp
@@ -998,9 +998,26 @@ CPVREpgInfoTagPtr CPVRTimerInfoTag::GetEpgInfoTag(bool bCreate /* = true */) con
             if (startTime > 0 && endTime > 0)
             {
               // if no epg uid present, try to find a tag according to timer's start/end time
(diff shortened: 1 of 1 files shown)


xbmc/pvr/timers/PVRTimerInfoTag.cpp
```diff
@@ -998,9 +998,26 @@ CPVREpgInfoTagPtr CPVRTimerInfoTag::GetEpgInfoTag(bool bCreate /* = true */) con
...
+                  const CPVRTimerInfoTagPtr timer = epgTag->Timer();
+                  if (timer && (timer->GetTimerRuleId() == m_iClientIndex))
+                  {
+                    bTagMatches = true;
+                  }
+                }
+
+                if (bTagMatches)
+                {
+                  m_epgTag = epgTag;
```
Jalle19: Previously `m_epgTag` would be set to `null` if no match was found, now it will not get set at all, potentially getting stuck with whatever value it had before. Is this intentional?

ksooo: All good. The whole block is only executed if m_epgTag is null. So, setting it previously to zero in that block was a nop.

PR #12490, opened 2017-07-13
[swig] Partially revert 71151fe2b9. Causes problems with swig < 3.0

hudokkow:
## Description
Partially revert 71151fe2b9

14.04 (trusty) ships swig 2.0.11 but swig's [C++11 support was only introduced in 3.0](http://www.swig.org/Doc3.0/CPlusPlus11.html#CPlusPlus11_explicit_overrides_final).
## Motivation and Context
Some team members are still using 14.04 as daily work/build env.

## How Has This Been Tested?
Not at all. No 14.04 here.

## Types of change

## Alternatives:
Package and ship 3.0.x in `xbmc-ppa-build-depends` or `xbmc-nightly`. @wsnipex?
Use a backports ppa, e.g. https://launchpad.net/~nschloe/+archive/ubuntu/swig-backports

@peak3d, I have no way to test unless I set up a new build env. Can you take it for a ride?

### DIFF
--- a/xbmc/interfaces/legacy/Addon.h
+++ b/xbmc/interfaces/legacy/Addon.h
@@ -80,7 +80,8 @@ namespace XBMCAddon
 
     public:
       Addon(const char* id = NULL);
-      ~Addon() override;
+      //! @todo Switch to 'override' usage once 14.04 (Trusty) hits EOL. swig <3.0 doesn't understand C++11
+      virtual ~Addon();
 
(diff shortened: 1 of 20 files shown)


hudokkow:
This is now working.

Tested with 14.04 with default swig (2.0.11).

peak3d:
can look earliest monday, but I think there is no need to retest as you already have done it.

hudokkow:
Works for me if jenkins is happy. Thanks for your patience. ;)

PR #12491, opened 2017-07-13
VideoPlayer: fixes

FernetMenta:
fix thump extraction for non 420p formats
fix setting processInfo for audio

/cc @pkerling @peak3d


FernetMenta:
jenkins error unrelated

pkerling:
thumbnails work again, thanks

PR #12493, opened 2017-07-13
RetroPlayer: Remove dependency on IPlayer interface

garbear:
This change removes a layer from the creation of saved states.

## Description

Previously, the autosave subsystem would call out to the IPlayer interface, which would call back in to the internal playback API, IGameClientPlayback:

* Autosave -> IPlayer -> IGameClientPlayback -> Playback API

Now, we call the internal playback API directly:

* Autosave -> IGameClientPlayback -> Playback API

## Motivation and Context
Comment in #12470


FernetMenta (approved):
thanks @garbear this was what I meant

PR #12500, opened 2017-07-14
VideoPlayer: drop FlipPage from RMs public interface

FernetMenta:


FernetMenta:
@popcornmix I think this requires adaptions in omx player

@garbear not sure if you have correct pts in videopicture. please check

xbmc/cores/RetroPlayer/RetroPlayerVideo.cpp
```diff
@@ -208,14 +208,10 @@ void CRetroPlayerVideo::SendPicture(VideoPicture& picture)
 {
   std::atomic_bool bAbortOutput(false); //! @todo
 
-  int index = m_renderManager.AddVideoPicture(picture);
+  if (!m_renderManager.AddVideoPicture(picture, bAbortOutput, VS_INTERLACEMETHOD_NONE, false))
   if (index < 0)
```
popcornmix: This doesn't compile - no index defined any more.

FernetMenta: updated

popcornmix:
@FernetMenta any advice on what is required by omxplayer now?

I used to call Flip (but not AddVideoPicture) to update pts value in renderer.

Now I don't really want to call AddVideoPicture as there is no picture to display.

If I don't call either, then I don't get any subs rendered.

FernetMenta:
@popcornmix actually nothing has changed. Both methods were called in a sequence. pts value argument of Flip was same as picture->pts of AddVideopicture

FernetMenta:
@popcornmix how is the status here?

popcornmix:
omxplayer has never called AddVideoPicture. It has no pictures to add - video rendering occurs on the gpu with the arm having no knowledge of individual frames being renderered.

omplayer does call FlipPage:
https://github.com/xbmc/xbmc/blob/master/xbmc/cores/omxplayer/OMXPlayerVideo.cpp#L313
which is required for subtitles to work.

I'm not sure what is now required. Should I generate a fake picture to pass to AddVideoPicture that gets discarded by renderer? It seems inefficient.

FernetMenta:
Call AddVIdeoPicture with a picture that has required parameters. You renderer can ignore the picture or whatever is appropriate for it.
At a later stage Renderer::FipPage will die as well becuase it is obsolete with buffering in renderer.

FernetMenta:
jenkins build this please

PR #12504, opened 2017-07-15
exit GIL when calling into GUI

FernetMenta:
GIL was never unlocked because the addon class clears tls

### DIFF
--- a/xbmc/interfaces/legacy/AddonUtils.cpp
+++ b/xbmc/interfaces/legacy/AddonUtils.cpp
@@ -30,21 +30,25 @@
 
 namespace XBMCAddonUtils
 {
-  GuiLock::GuiLock()
+  GuiLock::GuiLock(XBMCAddon::LanguageHook* languageHook, bool offScreen)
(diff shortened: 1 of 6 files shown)


FernetMenta:
@marcelveldt confirmed that this fixes deadlock issues he was observing with spotify addon.

MartijnKaijser:
@FernetMenta not sure backporting this one is easy as i had a lot of conflict when i tried.

FernetMenta:
@MartijnKaijser conflics are because of LOCKGUIIF, right? Resovling this should be straightforward. @marcelveldt wanted to backport this, iirc

MartijnKaijser:
@marcelveldt could you please make a PR soon so we can get 17.4 RC out the door?

marcelveldt:
@MartijnKaijser seems to be pretty straightforward indeed, I'll submit the PR later today.

marcelveldt:
@MartijnKaijser @FernetMenta 
https://github.com/xbmc/xbmc/pull/12555

PR #12506, opened 2017-07-15
GLES: allow using limited color range

lrusak:
This seems to do what is intended. I'm not sure if anything else is needed.

It would be nice to extend this so that hardware drivers can switch the colorspace also. For example intel graphics on linux allows switching between auto, full, and limited modes.


system/settings/settings.xml
```diff
@@ -2192,12 +2192,6 @@
           <control type="spinner" format="string" />
         </setting>
         <setting id="videoscreen.limitedrange" type="boolean" label="36042" help="36359">
-          <requirement>
-            <or>
-              <condition>HAS_GL</condition>
-              <condition>HAS_DX</condition>
-            </or>
-          </requirement>
```
MartijnKaijser: this will make it visible on all platforms

lrusak: afaik we have GL, GLES, and DX. So why make it conditional on all three?

FernetMenta:
Limited range for video only makes sense if the GUI can be rendered in limited range too. This is currently not the case for GLES. As a result black bars won't have the same black laevel as the background.

system/settings/settings.xml
FernetMenta: we have a lot of proprietary video renderers that do not support limited range.

lrusak:
@FernetMenta can you expand upon your black bars comment? Why can't the gui be rendered in limited range?

FernetMenta:
Currently all is rendered in full range. The config required is Kodi = full range, dirver = full range, TV = full range.
For limited range, that means head/foot room, Kodi needs to render in limited range, driver has to passthrough full range, and tv needs to be set to limited range.
In full range colour 0 is reference black. If you set your tv to limited, you'll get btb for a value of 0. Hence you need to render reference black with a value of 16.
We have implemented limited range rendering for the GUI for DX and OpenGL but not for GLES.

Note that rendering for pictures has to be adapted too.

lrusak:
>We have implemented limited range rendering for the GUI for DX and OpenGL but not for GLES.
Is that not what I'm trying to accomplish in this PR?

I can flick the limited range switch in kodi and clearly see that something is changing.

I've looked through the kodi but cannot see any other references to limited or full color range.

FernetMenta:
The limited range switch has impact on those parts:

https://github.com/xbmc/xbmc/search?p=1&q=UseLimitedColor&type=&utf8=%E2%9C%93
https://github.com/xbmc/xbmc/search?utf8=%E2%9C%93&q=limitedrange&type=

I don't see that it would have any effect on GUI rendering for GLES.

lrusak:
the only relevant places that I don't touch are `GUIFontTTFGL.cpp` and `SlideShowPicture.cpp`

FernetMenta:
sorry, somehow i missed your change in GUITextureGLES.cpp. gui is fine then

system/settings/settings.xml
lrusak: ah ok, so RPi, amlogic, etc that don't use LinuxRendererGLES will need to be excluded then.

FernetMenta: you could use a complex condition that allows dynamic change

lrusak: What is the best way to handle this? I can't see a way to depend on a not defined condition

FernetMenta: Some settings of AE can serve as an example or this one: https://github.com/FernetMenta/kodi-agile/blob/master/system/settings/settings.xml#L200

Memphiz:
@lrusak i tried this on IOS. While i see the change in color range when looking at the UI i fail to see anything changing during video playback. I simply start a movie and change the setting via the ios remote app. Is there something missing on that PR or is it my test case?

lrusak:
@Memphiz thanks for testing. I'll have to look into this again. It could be something missing in the PR. I'm not 100% sure.

lrusak:
Updated, this now seems to work. I'm not video color expert though.

@Memphiz if you would be so kind as to test this again. One this is that the `YUV2RGBProgressiveShader` is only configured once, so if you change the color range you need to stop and start a video again.

I'm using:
```
curl --data-binary '{ "jsonrpc": "2.0", "method": "Settings.SetSettingValue", "params": {"setting": "videoscreen.limitedrange", "value": false}, "id": 1 }' -H 'content-type: application/json;' -u xbmc:xbmc http://localhost:8080/jsonrpc
```

```
curl --data-binary '{ "jsonrpc": "2.0", "method": "Settings.SetSettingValue", "params": {"setting": "videoscreen.limitedrange", "value": true}, "id": 1 }' -H 'content-type: application/json;' -u xbmc:xbmc http://localhost:8080/jsonrpc
```

Rechi:
@lrusak this needs a rebase

Memphiz:
And now i finally got it ... i mixed up your PRs ... my comment:

"I have not the slightest amount of time for testing this currently. But given the fact that the last test definitely didn't break anything, and I am pretty sure this is still the case in the current state of this PR - you have my OK ;)" ... was meant for this very PR - but i accidently posted it on this (already merged) one: https://github.com/xbmc/xbmc/pull/12474

Well ... i guess users will figure out if 12474 really works on iOS or not ...

a1rwulf:
@lrusak What's the status here?
Anythink I could do to make it hit mainline?

a1rwulf:
@lrusak It doesn't compile anymore, i think this https://github.com/a1rwulf/xbmc/commit/28cd02b6c9905b3e398ce6c081369f8df34ae83a is missing.

system/settings/settings.xml
a1rwulf: @FernetMenta  @lrusak something like this maybe: https://github.com/a1rwulf/xbmc/commit/3de3b71f524f622f90d6d15d4d7db059dce86c47

a1rwulf:
@lrusak You can pick 28cd02b6c9905b3e398ce6c081369f8df34ae83a as well as b8e5ad88740ad8400c75ff158056d0b971ee19de and squash.
Then we can move forward.

a1rwulf:
Sorry @lrusak seems you picked an older version of my commit, can you have a look again at my branch:
https://github.com/a1rwulf/xbmc/commits/opengles-limited

You'll need the 2 top commits.
Next time I'll just PR into your branch to make it easier.
Thanks for your understanding.

a1rwulf:
@FernetMenta Can you have another look? All comments should be addressed now.
 @fritsch You had some issues initially (with pictures maybe?), if you find time your feedback is very welcome.

FernetMenta:
looks good imo

a1rwulf:
I did another runtime test.
IMHO it's good to go.

xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/YUV2RGBShaderGLES.h
```diff
@@ -67,6 +70,7 @@ namespace Shaders {
 
     void SetBlack(float black) override { m_black    = black; }
     void SetContrast(float contrast) override { m_contrast = contrast; }
+    void SetConvertFullColorRange(bool convertFullRange) { m_convertFullRange = convertFullRange; }
```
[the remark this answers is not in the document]
lrusak: thanks

lrusak:
jenkins build this please

lrusak:
jenkins build this please

a1rwulf:
@lrusak I think we should merge this :)

PR #12508, opened 2017-07-15
[modernize] Prefer 'nullptr' usage

hudokkow:
## Description
Modernize code and prefer `nullptr`over `NULL` usage.

## Motivation and Context
Modernize code.

## How Has This Been Tested?
Build/brief runtime test on 14.04, 16.10 and 17.04

## Types of change

- [X] Modernize code

Done automagically with clang-tidy.


xbmc/cores/RetroPlayer/PixelConverter.cpp
```diff
@@ -230,9 +230,9 @@ bool CPixelConverter::Decode(const uint8_t* pData, unsigned int size)
 
   const int stride = size / m_height;
 
-  uint8_t* src[] =       { dataMutable,         0,                   0,                   0 };
+  uint8_t* src[] =       { dataMutable,         nullptr,                   nullptr,                   nullptr };
```
garbear: Can you adjust the spacing here too?

hudokkow: Yep, will do after jenkins blessing. ;)

Rechi:
@hudokkow this needs a rebase

MartijnKaijser:
@hudokkow ping :)

Razzeee:
Rebase and move it in quickly to not have to rebase again? :)

hudokkow:
I'm sure rebase will be a nightmare by now. I'll try to find some time this week.

Razzeee:
Might be easier to redo then?

Rechi:
@hudokkow this needs a rebase

hudokkow:
@jenkins4kodi, I'm busy with other stuff. Get a life and shut up!

Rechi:
@hudokkow can this be merged?

Rechi:
@hudokkow with your latest change a massive amount of `conversion-null` warnings (assigning `NULL` to int) show up in jenkins.

hudokkow:
Weird. This is automagically done by clang-tidy. It should just work(™).
I'll see what's going on later today.

xbmc/addons/PVRClient.cpp
```diff
@@ -136,7 +136,7 @@ void CPVRClient::ResetProperties(int iClientId /* = PVR_INVALID_CLIENT_ID */)
   m_timertypes.clear();
   m_clientCapabilities.clear();
 
-  m_struct = {{0}};
+  m_struct = {{nullptr}};
```
ksooo: Are you sure this change is correct???

fritsch:
Please remove the AESinkPULSE changes. I looked after those myself. It needs great care here as we interfer with C API.

fritsch:
I see that you change a lot of ptrs in ffmpeg, which is a C API. Normally when compiled with a c++ compiler  NULL is defined nullptr and there should not be a problem at all.

I wonder if we need to take more care when passing "nullptr" into C API calls. As you know NULL in C is defined differently - and depending on the compiler might be something else.

@notspiff @Paxxi @Rechi  what's your take on that before we are running into a nightmare? :-)

Rechi:
Discussed it with @fritsch and we came to the following conclusion, replacing all `NULL` with `nullptr` doesn't work for C API calls.

Paxxi:
When I've toyed with this I've just run clang-tidy -modernize-use-nullptr on xbmc/*.cpp and xbmc/*.h and haven't noticed any issues afterwards. Passing in nullptr should still be safer than passing in NULL as it can be redefined depending on order of include files and any incorrect replacement would be a compile error.

clang-tidy will replace incorrect usages of NULL with 0 where it finds it passed to a non-pointer variable such as a uintptr_t etc.

I do not argue for using sed to do the work as that may certainly do some unexpected things.

PR #12509, opened 2017-07-15
[cmake][linux] Fix CodeCoverage.cmake path

hudokkow:
See title.

My bad! Comes from when I moved cmake to root folder.


hudokkow:
Stopped jenkins. It has no way to test this.

PR #12510, opened 2017-07-16
VideoPlayer: drop pvr SwitchChannel

FernetMenta:
This is the start of a bigger work product with the goal of cutting dependency from VideoPlayer to PVR. All code directly related to PVR will go away from VP.

Finally VP will play pvr channels through inputstream API.


xbmc/cores/VideoPlayer/DVDMessage.h
```diff
@@ -271,6 +266,28 @@ class CDVDMsgPlayerSetSpeed : public CDVDMsg
...
+public:
+  struct FileParams
+  {
+    CFileItem m_item;
+    CPlayerOptions m_options;
+  };
+
+  CDVDMsgOpenFile(FileParams params)
+  : CDVDMsg(PLAYER_OPENFILE)
+  , m_params(params)
```
garbear: `m_params(std::move(params))` assuming that both params have move constructors. If not, then passing by reference `(const FileParams &params)` is the next best thing

FernetMenta: thanks, updated

MartijnKaijser:
Unrelated Linux fail

FernetMenta:
the scope of this PR is now bigger. pvr channel switch code has become obsolete and I removed it.

FernetMenta:
@ksooo missing is channel entry timout. imo a useless feature. if you wnat to keep it, we have to find some mechanism where to implement the timer

ksooo:
Personally, I don't care about the channel switch delay feature, but I know our users cried every time I broke it accidentally (and I did multiple times ;-)...

More important for me is, that the channel preview feature is still functioning. According to the code changes in this PR it seems you took care of this, didn't you.

BTW: Really good work you're doing with removing pvr specific stuff from vp. Much appreciated.

opdenkamp:
Yeah you did, and I will complain when you break it again ;)
Nice work going on here. Really should find time to look at it again at some point soon :)

FernetMenta:
Thanks fo feedback. My last comment was not up-to-date. The last commit re-implements delayed channel entry. No worries :)
A positive side effect of this PR is that we don't need pvr specific load/store of video settings any more. Every channel switch goes the normal way via application.

ksooo:
👍

MartijnKaijser:
Good to go?

ksooo:
No objections from my end.

FernetMenta:
ok, if there are issues, I'll fix them

ksooo:
Some fallout:

0) Activate "confirm channel switch with okay"
1) tune a channel
2) hit up or down key
=> info for currently playing channel should appear, but info for next / prev channel appears

or
0) Activate "confirm channel switch with okay"
1) tune a channel
2) hit up or down some times to travel through channel info
=> try to bring up the info for the currently playing channel. You will not be able to.

ksooo:
Problem could be around here: https://github.com/xbmc/xbmc/pull/12510/files#diff-819d3671c423a1cd31e0279c3eda6ee1R956

FernetMenta:
yes, I interpreted the old code not correctly. It should not change m_currentFile if preview is off.

PR #12511, opened 2017-07-16
[win32] fix make-mingwlibs.bat

Montellese:
I recently had to re-do my windows setup and therefore had to run `make-mingwlibs.bat` but it just didn't work. I got several errors in the msys2 terminal about a git remote not being found and it failed to download the proper ffmpeg tarball and was therefore unable to build ffmpeg. So I started looking into it and figured out that the root of the problems was the definition of `BASE_URL` in `tools/depends/target/ffmpeg/FFMPEG-VERSION` which is parsed by `tools/buildsteps/windows/buildhelpers.sh` and then used in a `git ls-remote` command to figure out the proper git revision/hash to download. Since the `BASE_URL` for ffmpeg has the `/archive` at the end it's not a valid git repository anymore and `git ls-remote` fails and therefore the whole download fails.

I checked other version files like `tools/depends/target/libdvdcss/DVDCSS-VERSION` or `tools/depends/target/libdvdnav/DVDNAV-VERSION` and they all use a `BASE_URL` without the `/archive` path in the URL.

With the first commit from this PR everything works perfectly fine again.


FernetMenta (approved):
looks good imo but @wsnipex should check

wsnipex:
also needs adjustment here: https://github.com/xbmc/xbmc/blob/master/tools/depends/target/ffmpeg/autobuild.sh#L119

Rechi:
and also at https://github.com/xbmc/xbmc/blob/85b06c0/cmake/modules/FindFFMPEG.cmake#L227

Montellese:
@wsnipex @Rechi: thanks, already noticed that I missed those after having put up the PR but didn't have time to fix it yet. Should be all done now.

Paxxi:
Actually @Rechi knows this best, if he says it's good then merge

Rechi (approved):
Thanks for the fix.
Already some people thought other errors are caused by the not found error (which isn't a fatal error).

PR #12513, opened 2017-07-17
Fix several settings that broke after the last few ServiceManager refactorings

garbear:
This fixes several settings that have broke over the last few weeks:

* Enable/disable mouse
* Peripherals dialog
* Peripheral drivers dialog
* Controller configuration
* Test rumble
* CEC language locale
* Add-on autoupdates

## How Has This Been Tested?
Enable/disable mouse and controller configuration guaranteed to be working. Haven't tested the add-on autoupdates setting.

## Types of change

### DIFF
--- a/xbmc/addons/RepositoryUpdater.cpp
+++ b/xbmc/addons/RepositoryUpdater.cpp
@@ -45,7 +45,12 @@ CRepositoryUpdater::CRepositoryUpdater(CAddonMgr& addonMgr) :
   m_timer(this),
   m_doneEvent(true),
   m_addonMgr(addonMgr)
-{}
+{
+  // Register settings
+  std::set<std::string> settingSet;
+  settingSet.insert(CSettings::SETTING_ADDONS_AUTOUPDATES);
+  CServiceBroker::GetSettings().RegisterCallback(this, settingSet);
+}
 
 void CRepositoryUpdater::Start()
 {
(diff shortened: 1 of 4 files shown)


Montellese (approved):
This looks good and is finally moving in the direction that I planned from the beginning. We could even go as far as passing the `CSettings` instance as a paramter to `CRepositoryUpdater`, `CInputManager` and `CPeripherals` to reduce the dependency on `CServiceBroker`.

lrusak (approved):

garbear:
> We could even go as far as passing the `CSettings` instance as a paramter to `CRepositoryUpdater`, `CInputManager` and `CPeripherals` to reduce the dependency on `CServiceBroker`.

I thought of this, but settings are brought up *really* early so there's very little in our codebase that can interact with settings before it's ready.

Everything depends on settings, so I can see passing it as a parameter going down a deep rabbit hole.

garbear:
jenkins build this please

da-anda:
the CEC settings dialog still has not settings labels. Is this something you could look into as well please? I tested this PR as part of latest Milhouse build.

garbear:
@da-anda like this?

peripheral settings

That is caused by the error fixed in #12518. The current version of the PR doesn't fix this, but a previous version did that was modified per some suggestions. You can bring up this issue over there if you want.

garbear:
jenkins build this please

PR #12514, opened 2017-07-17
Remove autotools files from gitignore

yol:
Remove autotools files from gitignore

## Motivation and Context
Files from autotools are still in gitignore. They shouldn't be there so people can clean up old cruft with `git clean`.

## Types of change

## Checklist:


pkerling:
Maybe there's more that can be removed, input appreciated.

.gitignore
```diff
@@ -18,9 +18,6 @@ testMain
 .vscode/
 
 # generated files etc
-config.cache
-config.status
-config.log
```
wsnipex: those 3 still exist in tools/depends

pkerling: So would adding `/tools/depends/**/config.cache` etc. be OK?

wsnipex: sure

pkerling: done

.gitignore
```diff
@@ -133,9 +109,26 @@ cmake_install.cmake
...
+/addons/kodi.binary.instance.audioencoder/addon.xml
+/addons/kodi.binary.instance.game/addon.xml
+/addons/kodi.binary.instance.imagedecoder/addon.xml
+/addons/kodi.binary.instance.inputstream/addon.xml
+/addons/kodi.binary.instance.peripheral/addon.xml
+/addons/kodi.binary.instance.pvr/addon.xml
+/addons/kodi.binary.instance.screensaver/addon.xml
+/addons/kodi.binary.instance.vfs/addon.xml
+/addons/kodi.binary.instance.videocodec/addon.xml
+/addons/kodi.binary.instance.visualization/addon.xml
```
wsnipex: does it make sense to use a wildcard here?
i.e. /addons/kodi.binary.instance.*/addon.xml

pkerling: Why are those needed at all? Aren't they supposed to go to the build folder anyway?

pkerling: friendly ping :-)

Rechi: that's for people building in-source
as we warn about such builds at https://github.com/xbmc/xbmc/blob/d790e54/CMakeLists.txt#L14-L16 imo you can remove it completely from .gitignore

@garbear do you build in-source?

garbear: On windows I build within a build/ subfolder, but these are generated anyway

Rechi: If you don't build in-source, they are generated in your build subfolder which is already completely ignored.

garbear: Oh right, I symlink /build/addons to /addons to avoid having to recompile binary add-ons (a big pain with 60 game add-ons) whenever I remove and recreate my build folder

notspiff: would you accept a fix for the addon macros to do the same? I hate the write to source tree there just as much..

hudokkow: Yes please! 😀

AlwinEsch:
From my side OK

Razzeee:
@pkerling 
Do you mind if we squash this down to one commit on merge?

pkerling:
@Razzeee no, I think that's a good idea

PR #12517, opened 2017-07-17
Valgrind

a1rwulf:
Valgrind complained about those lines.

## Description
First change uses delete[] for an array that was allocated with new[].
Afaik this needs to be done according to the standard, although most platform
can handle it implicitly.

According to our docs StringUtils::Format must get .c_str and cannot get a std::string.
Not sure if this is right.

@Paxxi @Montellese ?

## How Has This Been Tested?
Couldn't find any regressions when having this commits.
I started this investigation because json-rpc and webserver crash sometimes on my machine and
I hoped the crashes would be related.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/interfaces/json-rpc/JSONServiceDescription.cpp
+++ b/xbmc/interfaces/json-rpc/JSONServiceDescription.cpp
@@ -768,7 +768,7 @@ JSONRPC_STATUS JSONSchemaTypeDefinition::Check(const CVariant &value, CVariant &
       // allowed there is no need to check every element
       if (value.size() < items.size() || (value.size() != items.size() && additionalItems.size() == 0))
       {
-        CLog::Log(LOGDEBUG, "JSONRPC: One of the array elements does not match in type {0}", name.c_str());
+        CLog::Log(LOGDEBUG, "JSONRPC: One of the array elements does not match in type %s", name.c_str());
         errorMessage = StringUtils::Format("{0} array elements expected but {1} received", items.size(), value.size());
         errorData["message"] = errorMessage.c_str();
         return InvalidParams;
--- a/xbmc/network/websocket/WebSocket.cpp
+++ b/xbmc/network/websocket/WebSocket.cpp
@@ -242,7 +242,7 @@ CWebSocketFrame::~CWebSocketFrame()
 
   if (m_free && m_data != NULL)
   {
-    delete m_data;
+    delete[] m_data;
     m_data = NULL;
   }
 }


xbmc/interfaces/json-rpc/JSONServiceDescription.cpp
```diff
@@ -768,7 +768,7 @@ JSONRPC_STATUS JSONSchemaTypeDefinition::Check(const CVariant &value, CVariant &
       // allowed there is no need to check every element
       if (value.size() < items.size() || (value.size() != items.size() && additionalItems.size() == 0))
       {
-        CLog::Log(LOGDEBUG, "JSONRPC: One of the array elements does not match in type {0}", name.c_str());
```
Montellese: This is the new formatting style from `{fmt}` so it's correct. We could even drop the `c_str()` as `{fmt}` is type-safe and can handle `std::string`.

xbmc/interfaces/json-rpc/JSONServiceDescription.cpp
```diff
@@ -1424,7 +1424,7 @@ bool CJSONServiceDescription::prepareDescription(std::string &description, CVari
   }
 
   if (description.at(0) != '{')
-    description = StringUtils::Format("{{{:s}}}", description);
+    description = StringUtils::Format("{{{:s}}}", description.c_str());
```
Montellese: See the comment above. The new formatting library / implementation (`{fmt}`) is type-safe and can handle std::string.

Montellese (changes requested):
The changes to the websocket implementation look correct. The changes to `CJSONServiceDescription` would be correct with the old implementation but are not necessary with the new one.

xbmc/interfaces/json-rpc/JSONServiceDescription.cpp
pkerling: Doesn't `CLog::Log` use regular `vsnprintf`?

Montellese: You're right, I didn't look close enough. Since the PR description mentioned `StringUtils::Format()` I assumed this was a `StringUtils::Format()` call as well. In that case this change is correct.

a1rwulf: Mhm, looks like it was ported to use StringUtils::Format internally (unless I oversee something):
https://github.com/xbmc/xbmc/blame/master/xbmc/utils/log.cpp

a1rwulf: However for both methods, valgrind complained about memory access on already freed bytes and adding c_str fixes it.
Do you think this is a false positive?

xbmc/interfaces/json-rpc/JSONServiceDescription.cpp
a1rwulf: Is this commet wrong then?
https://github.com/xbmc/xbmc/blob/master/xbmc/utils/StringUtils.h#L52

pkerling: I think the comment is wrong. fmt (at least version 4?) should be able to handle pretty much everything that has an `operator<<`. You can test the line without the `c_str()`, it will work.

xbmc/network/websocket/WebSocket.cpp
```diff
@@ -242,7 +242,7 @@ CWebSocketFrame::~CWebSocketFrame()
 
   if (m_free && m_data != NULL)
   {
-    delete m_data;
+    delete[] m_data;
```
Paxxi: Correct but would be better to change it to a unique_ptr

a1rwulf: If you don't mind, I'd like to keep it as is.
m_data has a double role in this file, in one case it's a plain const char* that get's a parameter assigned and is not freed (see m_free flag) - from a quick look I can't get why it's done like that, but I guess there's a reason.
In other cases it's a variable sized array on the heap and needs to be freed.

xbmc/interfaces/json-rpc/JSONServiceDescription.cpp
a1rwulf: I've dropped the second change and left the change to the log output.

xbmc/network/websocket/WebSocket.cpp
a1rwulf: @Paxxi what do you think? do you see a better solution without the need of touching too much existing code?

a1rwulf:
@Paxxi @Montellese anything blocking this from being merged?

Paxxi:
Go ahead

a1rwulf:
jenkins build and merge

PR #12518, opened 2017-07-18
Fix peripherals started before strings are loaded

garbear:
This fixes the following error:

screenshot000

## How Has This Been Tested?
The labels show up now.

## Types of change

### DIFF
--- a/xbmc/ServiceManager.cpp
+++ b/xbmc/ServiceManager.cpp
@@ -107,9 +107,6 @@ bool CServiceManager::InitStageTwo(const CAppParamParser &params)
   m_inputManager->InitializeInputs();
 
   m_peripherals.reset(new PERIPHERALS::CPeripherals(*m_announcementManager));
-  m_peripherals->Initialise();
-
-  m_gameServices.reset(new GAME::CGameServices(*m_gameControllerManager, *m_peripherals));
 
   init_level = 2;
   return true;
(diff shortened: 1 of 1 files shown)


tamland:
Why is this? Is something caching it? If the labels are translated when they are displayed this won't matter as they will get the correct one once the strings are loaded.

garbear:
The labels are cached, but translating at runtime would work too.

garbear:
Updated with @tamland's suggestion

garbear:
Some users found more places that this update didn't address

peripheral settings

I've restored the original solution to address this and possibly other locations in peripherals that rely on language strings early on.

garbear:
It's worth noting that initializing peripherals in stage 3 was what originally happened in #11824. I can't say if the move to stage 2 has other negative side effects besides strings breaking.

FernetMenta:
stage 3 is for components that require ui. this looks more like a work around but does not address the root cause.

garbear:
A bunch of stuff happens between stages 2 and 3, and peripherals could depend on any number of it. I don't know the reason peripherals were originally brought up near stage 3, but restoring this behavior seems like the least intrusive change.

quthla:
Will CEC be fixed maybe?

FernetMenta:
@garbear still a long way to go with our globals, isn't? could you place a comment in the code with this info?

garbear:
@FernetMenta added a comment, is that what you had in mind?

FernetMenta:
ok, thanks

garbear:
Jenkins build this please

tamland:
> I've restored the original solution to address this and possibly other locations in peripherals that rely on language strings early on.

Still a really bad idea though. Unless you register callback for all these things, it breaks language switching and updating. If string translation is simply kept to the gui layer (and not cached a second time), it just works.

garbear:
True, language switching will still be broken. I'll address this in a future PR. My main argument is that Peripherals doesn't need to be brought up so early, and indeed the change has already broken some stuff in the past, including causing a segfault in tests. Once this change is in I'll start to decouple peripherals from the GUI.

Jenkins build this please

PR #12519, opened 2017-07-18
[info] check if m_refreshCounter is zero

BigNoid:
This is an attempt to fix the problem mentioned by me in #12459

@notspiff I am not sure this is the correct way. It fixes my problem in any case.

### DIFF
--- a/xbmc/interfaces/info/InfoBool.h
+++ b/xbmc/interfaces/info/InfoBool.h
@@ -47,7 +47,7 @@ public:
   {
     if (item && m_listItemDependent)
       Update(item);
-    else if (m_refreshCounter != m_parentRefreshCounter)
+    else if (m_refreshCounter != m_parentRefreshCounter || m_refreshCounter == 0)
     {
       Update(NULL);
       m_refreshCounter = m_parentRefreshCounter;


akva2:
what is the value of m_parentRefreshCounter when you get the problem ?

akva2:
if i understand the code correctly, it's 0. and then i do not understand why this makes a difference, since Update(NULL) does absolutely nothing. and then we set them equal (0) again. so the whole changed should be a NOP.

BigNoid:
It's 0 indeed for the unevaluated expressions. If I add ```Update(NULL);``` [here](https://github.com/xbmc/xbmc/blob/master/xbmc/interfaces/info/InfoExpression.h#L52) I get a crash in ```InfoExpression::Update``` but maybe I misunderstood what you meant.

akva2:
that was what i wanted tested yes.

i guess if it does the job, it's not naughty (what was naughty was old code relying on undefined behavior). just wanted to try to understand it, but i guess this has regressed more than this one condition so better fix it.

BigNoid:
Okay, thx. Jenkins build this please

PR #12522, opened 2017-07-18
AddonSettings: Protect against nullptr

popcornmix:
I hit a crash with a null newSetting pointer.
No obvious issue after protecting the access.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/settings/AddonSettings.cpp
+++ b/xbmc/addons/settings/AddonSettings.cpp
@@ -313,7 +313,12 @@ bool CAddonSettings::Load(const CXBMCTinyXML& doc)
     }
 
     // try to load the old setting value
-    if (!newSetting->FromString(setting.second))
+    if (!newSetting)
+    {
+      CLog::Log(LOGERROR, "CAddonSettings[%s]: had null newSetting for value \"%s\" for setting %s",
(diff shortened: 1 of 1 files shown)


Montellese:
Could you add a log message in case `newSetting` is `nullptr`? Because in that case something is either wrong with the legacy parser or with the add-on's settings.

xbmc/addons/settings/AddonSettings.cpp
```diff
@@ -309,7 +309,7 @@ bool CAddonSettings::Load(const CXBMCTinyXML& doc)
     }
 
     // try to load the old setting value
-    if (!newSetting->FromString(setting.second))
+    if (newSetting && !newSetting->FromString(setting.second))
```
AlwinEsch: Can a empty "newSetting" come in normal condition?
If no maybe to use:
```cpp
if (newSetting == nullptr || !newSetting->FromString(setting.second))
```
to have a log message about them?

Razzeee:
@popcornmix can you address what  @Montellese asked for so we can merge this?

MartijnKaijser:
@popcornmix ping

MartijnKaijser:
Nudge

popcornmix:
Updated - @Montellese does this look okay?

Montellese (approved):

PR #12526, opened 2017-07-19
[pvr] drop channel guide button from info dialog + add channel guide to context menu

xhaggi:
First, this PR drops the channel guide button (introduced by myself) from the guide info dialog, because normally you open the pvr guide info dialog from within the channel guide dialog not the other way around. In that case it'll end-up in an unsightly UI workflow.

Second, I added a new entry "Channel guide" to the context menu for all channel items with existing EPG data. You can now reach the channel guide from there.

## Motivation and Context

Unsightly UI workflow.

## How Has This Been Tested?

Runtime tested on MacOS

### DIFF
--- a/addons/skin.estuary/xml/DialogPVRChannelGuide.xml
+++ b/addons/skin.estuary/xml/DialogPVRChannelGuide.xml
@@ -3,6 +3,13 @@
 	<defaultcontrol always="true">11</defaultcontrol>
 	<controls>
 		<control type="group">
+			<animation effect="fade" start="0" end="100" time="100">WindowOpen</animation>
+			<animation effect="fade" start="100" end="0" time="100">WindowClose</animation>
+			<animation effect="fade" start="100" end="75" time="0" condition="true">Conditional</animation>
+			<visible>!Window.IsActive(fullscreenvideo)</visible>
+			<include>ColoredBackgroundImages</include>
+		</control>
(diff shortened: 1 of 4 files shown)


ksooo:
You should implement the new context menu entry using the new context menu system (See PVRContextMenus.cpp).  Although there are still a few leftovers of the old system in pvr code we should not introduce new "old way" entries.

xhaggi:
Thank you for pointing me to the right place 👍 Will move it over.

xhaggi:
@ksooo updated

xhaggi:
I've done some tests and it works on homescreen + channel window.

xhaggi:
I've changed the condition in IsVisible so the context entry is only shown if epg data exists.

xhaggi:
@phil65 or @ronie I've changed estuary to show colored background images (same as it's done for pvr guide info dialog) if the channel guide is used outside fullscreen. Could one of you taking a look.

ksooo:
I can only speak for the non-Estuary changes, which are looking good.

xhaggi:
@ksooo thx for reviewing

phil65:
Skin change looks fine.

phil65 (approved):

xhaggi:
jenkins build this please

xhaggi:
Build errors are unrelated.

PR #12528, opened 2017-07-20
CHG: simplify emu_msvcrt FILE* handling

koying:
Context:
The latest Android NDK "opacifies" the FILE struct (making it `char[nn]`) which left me with 2 options:
1) Add an additional ifdef and find another hack
2) Do something else

This is 2). It takes advantage of the fact that the FILE* are completely ours, and, being supposidely opaque, can be anything.
So, rather than keeping track of an actual, platform-dependent, FILE, we return a "reinterpret_cast"ed pointer to our own structure.

This allows the code to be platform/compiler-independent (removing the ifdefs)

Limitation:
If the client lib takes a copy of the actual FILE struct, this will crash badly, as their definition and ours will be different, but there is no reason to do such thing.

Tested by me on Android arm32.
AFAIK, the only actual remaining way to test is to extract a DVD on the file system, and read the IFO.

### DIFF
--- a/xbmc/cores/DllLoader/exports/util/EmuFileWrapper.cpp
+++ b/xbmc/cores/DllLoader/exports/util/EmuFileWrapper.cpp
@@ -27,28 +27,11 @@ CEmuFileWrapper g_emuFileWrapper;
 namespace
 {
 
-#if defined(TARGET_WINDOWS) && (_MSC_VER >= 1900)
-constexpr kodi_iobuf* FileDescriptor(FILE& f)
-{
-  return static_cast<kodi_iobuf*>(f._Placeholder);
-}
-
-constexpr bool isValidFilePtr(FILE* f)
-{
-  return (f != nullptr && f->_Placeholder != nullptr);
-}
-
-#else
-constexpr FILE* FileDescriptor(FILE& f)
-{
-  return &f;
-}
-
 constexpr bool isValidFilePtr(FILE* f)
 {
   return (f != nullptr);
 }
-#endif
+
 }
 CEmuFileWrapper::CEmuFileWrapper()
 {
@@ -57,10 +40,7 @@ CEmuFileWrapper::CEmuFileWrapper()
   {
     memset(&m_files[i], 0, sizeof(EmuFileObject));
     m_files[i].used = false;
-#if defined(TARGET_WINDOWS) && (_MSC_VER >= 1900)
-    m_files[i].file_emu._Placeholder = new kodi_iobuf();
-#endif
-    FileDescriptor(m_files[i].file_emu)->_file = -1;
+    m_files[i].fd = -1;
   }
 }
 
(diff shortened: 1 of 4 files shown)


koying:
For info, the issue this PR workarounds blocks the usage of the latest Android NDK, which in turn blocks implemeting Samba 4.5 (thus SMB2/3).

So feedback welcome ;)

Memphiz:
I currently feel that I neither have any insight on the area of code that is changed here - nor do I grasp the impact :/

koying:
NP ;)
Just a test with this PR on a "expanded" DVD ISO on your respective platforms will do.

The emulation layer is bound to disappear anyway, just not yet...

Memphiz:
I won't have time to test anything like that until monday if i am lucky :(

Voyager1:
I need to take a deeper look into this... First off, DVD ISOs are no longer using the emulation layer as I recently changed this to use the libdvd streaming API. So the real test would be to use a VOB/IFO structure DVD. Secondly, there are other files than just our own FILE structures that need emulation, like the standard streams (stdout/stderr...)

koying:
> so the real test would be to use a VOB/IFO structure DVD

Yep, mentioned it in the comments.

> Secondly, there are other files than just our own FILE structures that need emulation, like the standard streams

Not sure what you exactly mean by that, I must say. Nothing is related to standard streams in the current code, as far as I can see.

Thanks to both for looking into it, anyway :)

Voyager1:
> Nothing is related to standard streams

not in our current code but in libdvd/css itself e.g. log output

koying:
At first sight, STD* shouldn't be an issue, as we only use `CEmuFileWrapper` for streams that are opened via the emulation layer, so ones where we really "owns" the FILE *. All others would simply be ignored.

Tests would tell, ofc,

Voyager1:
My understanding is that hooking calls into our emulation layer for every stream you open. So if libdvd (or any other addon) wants to create/open/delete etc. any file, whether it's a kodi-owned or not, it comes in our emulation layer. Therefore you cannot simply assume that any FILE* is emulated. So the reinterpret_cast to EmuFileObject* is wrong.

koying:
> Therefore you cannot simply assume that any FILE* is emulated

Well yes, we can, that's the whole point. *All* FILE are Kodi owned, as they go thru our emulation. Or not, but they cannot do both, ie have mixed FILE definitions.
If we assume the client (eg libdvd) only handles FILE *pointers*, we can thus return whatever we want, as FILE is opaque.

If it was not the case, overriding fields in the *opaque* FILE structure as we do now would most probably lead to unpredictable results as well.

Re the standard streams, they seem protected everywhere by `IS_STD_STREAM`, so I suspect they wouldn't be an issue.

Voyager1 (commented):
> If we assume the client (eg libdvd) only handles FILE pointers

I agree that the code is correct IF we can assume this. I think we need to take a closer look and do sufficient testing.

xbmc/cores/DllLoader/exports/util/EmuFileWrapper.cpp
```diff
@@ -264,7 +235,8 @@ bool CEmuFileWrapper::StreamIsEmulatedFile(FILE* stream)
 {
   if (isValidFilePtr(stream))
   {
-    return DescriptorIsEmulatedFile(FileDescriptor(*stream)->_file);
+    EmuFileObject* obj = reinterpret_cast<EmuFileObject*>(stream);
```
koying: The whole point of this PR is that it *never* is a real FILE ;)
If this assumption is proved false, obviously everything crumbles...

Voyager1:
hi @koying - I've just done some testing with libdvd and a VOB/IFO dvd and it looks fine. I also checked the existing code for the redirection of stderr/stdout to CLog and that's also properly guarded.

koying:
Cool, thanks.

Memphiz:
I tried this on ios now. I've extracted an iso and opened the ifo file in file mode which successfully started the dvd/menu ... the dvd folder was shared via nfs from my nas. I hope i got the essentials for this test covered that way?

koying:
Yep, thanks.
I guess we can say this one is good to go?

Memphiz (approved):

Voyager1 (approved):

Paxxi (approved):

PR #12531, opened 2017-07-20
[addons] show update all entry also if only one addon update is available

xhaggi:
See title.

## Motivation and Context

It's really annoying that you need to open the addon info dialog to update the addon if only one addon is listed under available updates.

## How Has This Been Tested?
Runtime tested on MacOS

## Types of change

### DIFF
--- a/xbmc/filesystem/AddonsDirectory.cpp
+++ b/xbmc/filesystem/AddonsDirectory.cpp
@@ -509,7 +509,7 @@ static void OutdatedAddons(const CURL& path, CFileItemList &items)
   VECADDONS addons = CAddonMgr::GetInstance().GetAvailableUpdates();
   CAddonsDirectory::GenerateAddonListing(path, addons, items, g_localizeStrings.Get(24043));
 
-  if (items.Size() > 1)
+  if (!items.IsEmpty())
   {
     CFileItemPtr item(new CFileItem("addons://update_all/", false));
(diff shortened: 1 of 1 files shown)


xhaggi:
@MartijnKaijser mind taking a look too because the changes to the code are trivial. We need to decide if that makes sense.

xbmc/filesystem/AddonsDirectory.cpp
```diff
@@ -509,7 +509,7 @@ static void OutdatedAddons(const CURL& path, CFileItemList &items)
   VECADDONS addons = CAddonMgr::GetInstance().GetAvailableUpdates();
   CAddonsDirectory::GenerateAddonListing(path, addons, items, g_localizeStrings.Get(24043));
 
-  if (items.Size() > 1)
+  if (items.Size() >= 1)
```
AlwinEsch: Maybe to use then a:
```cpp
  if (!items.IsEmpty())
```
?

AlwinEsch:
This change make also sense for me. Better to become also with only one informed about.

xbmc/filesystem/AddonsDirectory.cpp
xhaggi: Thanks, updated + force-pushed

xhaggi:
Any objections? If not I'll shove it in.

PR #12534, opened 2017-07-20
FIX: Database debug spew on demand only

koying:
Be easy on db logging by default


MartijnKaijser:
Unrelated Linux fail

mkortstiege (changes requested):
```LOGDEBUG``` output created in ```CVideoDatabase::CleanDatabase``` should be added as well. Other than that it looks good to me.

peak3d:
Any reason why this is not already merged?

mkortstiege (approved):

PR #12535, opened 2017-07-20
[pvr] change setting continue last channel to yes and always

xhaggi:
This PR change the setting "Continue last channel" and drop the values "foreground" and "background" in favor of "Yes", which means it will only playback the last viewed channel if it was playing on app quit, and "Always" which always continue the last viewed channel.

## Motivation and Context

Not many users differ between foreground and background, but want Kodi to playback TV on startup no matter if they played a channel on app quit. They use Kodi as TV receiver replacement (me too). 

## How Has This Been Tested?

Runtime tested on MacOS

## Types of change

### DIFF
--- a/system/settings/settings.xml
+++ b/system/settings/settings.xml
@@ -1386,9 +1386,9 @@
           <default>0</default> <!-- CONTINUE_LAST_CHANNEL_OFF -->
           <constraints>
             <options>
-              <option label="106">0</option>   <!-- CONTINUE_LAST_CHANNEL_OFF -->
-              <option label="19190">1</option> <!-- CONTINUE_LAST_CHANNEL_IN_BACKGROUND -->
-              <option label="19288">2</option> <!-- CONTINUE_LAST_CHANNEL_IN_FOREGROUND -->
+              <option label="106">0</option> <!-- CONTINUE_LAST_CHANNEL_OFF -->
+              <option label="107">1</option> <!-- CONTINUE_LAST_CHANNEL_ON -->
+              <option label="34122">2</option> <!-- CONTINUE_LAST_CHANNEL_ALWAYS -->
             </options>
           </constraints>
           <control type="list" format="string" />
--- a/xbmc/pvr/PVRManager.h
+++ b/xbmc/pvr/PVRManager.h
@@ -53,9 +53,9 @@ namespace PVR
 
   enum ContinueLastChannelOnStartup
   {
-    CONTINUE_LAST_CHANNEL_OFF  = 0,
-    CONTINUE_LAST_CHANNEL_IN_BACKGROUND,
-    CONTINUE_LAST_CHANNEL_IN_FOREGROUND
+    CONTINUE_LAST_CHANNEL_OFF = 0,
+    CONTINUE_LAST_CHANNEL_ON,       // continue with the last viewed channel if it was playing on app quit
+    CONTINUE_LAST_CHANNEL_ALWAYS    // continue with the last viewed channel no matter if it was playing on app quit.
   };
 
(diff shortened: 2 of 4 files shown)


xhaggi:
I've added a second commit which re-use the fullscreen playback setting for continue last channel to decide if the channel will be played in fullscreen or background.

ksooo:
I still think that the better approach is to drop this setting completely, in favour of a new settings value for the existing setting "kodi startup window".

xhaggi:
Sure we talked about that but this will not fix the issue, I have, after you added the 'only if playing on app quit' stuff. We also need to differ that there or how do we proceed here?

ksooo:
Well, I have no strong feelings about this. I can also live with your approach.

system/settings/settings.xml
```diff
@@ -1386,9 +1386,9 @@
           <default>0</default> <!-- CONTINUE_LAST_CHANNEL_OFF -->
           <constraints>
             <options>
-              <option label="106">0</option>   <!-- CONTINUE_LAST_CHANNEL_OFF -->
-              <option label="19190">1</option> <!-- CONTINUE_LAST_CHANNEL_IN_BACKGROUND -->
-              <option label="19288">2</option> <!-- CONTINUE_LAST_CHANNEL_IN_FOREGROUND -->
```
ksooo: Are these two strings still used after you removed then here?

xbmc/pvr/PVRManager.h
```diff
@@ -54,8 +54,8 @@ namespace PVR
   enum ContinueLastChannelOnStartup
   {
     CONTINUE_LAST_CHANNEL_OFF  = 0,
-    CONTINUE_LAST_CHANNEL_IN_BACKGROUND,
-    CONTINUE_LAST_CHANNEL_IN_FOREGROUND
+    CONTINUE_LAST_CHANNEL_ON,
+    CONTINUE_LAST_CHANNEL_ALWAYS
```
ksooo: Maybe comments explaining the meaning of the values  would  be good.

xhaggi: Sure can c/p the stuff from the `string.po` where it is explained.

xhaggi:
BTW if we really want to move the TV setting over we need to change the way the 'startup window' setting works. ATM everything is designed to handle a window ID as setting value see `CSkinInfo::LoadStartupWindows` and `CSkinInfo::GetStartWindow()`. Hacking the TV stuff in here sounds wrong.

system/settings/settings.xml
xhaggi: No. They can go.

ksooo:
I see it the other way around. Looking at it from the functional perspective, handling it as "startup window" is definitely the right place. But yeah, to not "hack it in", the current startup window handling implementation needs larger refactoring/rewrite. Your approach is far less effort, but not the best we could achieve function-wise.

xhaggi:
Pushed a fixup which address your comments. If okay by you, I'll squash everything.

> I see it the other way around. Looking at it from the functional perspective, handling it as "startup window" is definitely the right place. But yeah, to not "hack it in", the current startup window handling implementation needs larger refactoring/rewrite.

I'm with you, it makes totally sense, but as you said it results in more effort. If I found some more spare time, not some hours after work per week, I could look at it.

xhaggi:
@ksooo I'll merge this in tomorrow and we do another round to achieve what we're talked about. But it needs an option to override the default, which should be: Playback the content I've viewed before app quit, because I want Kodi to startup with TV no matter which content was played last.

xhaggi:
follow-up #12548

PR #12536, opened 2017-07-20
[xbmc] Addon settings fixes

Paxxi:
Fix a parsing issue for new setting format where settingId would be the element name instead of the id
Fix a conversion issue where settings without category was skipped over.

This should solve log spam of creating settings on the fly for valid addons and properly save/load addons after converted to the new setting format.

@peak3d for testing. 

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/settings/AddonSettings.cpp
+++ b/xbmc/addons/settings/AddonSettings.cpp
@@ -249,7 +249,9 @@ bool CAddonSettings::Load(const CXBMCTinyXML& doc)
       auto settingId = categoryId;
       if (!settingId.empty())
         settingId += ".";
-      settingId += setting->ValueStr();
+      auto id = setting->ToElement()->Attribute("id");
+      if (id)
+        settingId += id;
 
       // parse the setting value
       std::string settingValue;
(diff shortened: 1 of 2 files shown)


xbmc/addons/settings/AddonSettings.cpp
```diff
@@ -249,7 +249,9 @@ bool CAddonSettings::Load(const CXBMCTinyXML& doc)
       auto settingId = categoryId;
       if (!settingId.empty())
         settingId += ".";
-      settingId += setting->ValueStr();
+      auto id = setting->ToElement()->Attribute("id");
+      if (id)
+        settingId += id;
```
garbear: Maybe this should be 

```c++
auto id = setting->ToElement()->Attribute("id");
if (id)
{
  if (!settingId.empty())
    settingId += ".";
  settingId += id;
}
```

to avoid the dangling "."?

xbmc/addons/settings/AddonSettings.cpp
```diff
@@ -435,39 +437,11 @@ bool CAddonSettings::ParseSettingVersion(const CXBMCTinyXML& doc, uint32_t& vers
...
-    categoryId += 1;
-
-    // try to get the category's label and fall back to "General"
-    int categoryLabel = 128;
-    ParseOldLabel(categoryElement, g_localizeStrings.Get(categoryLabel), categoryLabel);
-    category->SetLabel(categoryLabel);
-
     // build a vector of settings from the same category
     std::vector<std::shared_ptr<const CSetting>> categorySettings;
+    std::set<std::string> settingIds;
```
peak3d: Better place the definition on top of the methode and call clear()
Avoids many malloc calls

peak3d (changes requested):
Looks good exept the one minor thing

xbmc/addons/settings/AddonSettings.cpp
Paxxi: don't think it's required, only check id so we don't crash on a nullptr. It should never return null unless xml is completely broken and we don't really handle that anywhere I think.

garbear: agreed

xbmc/addons/settings/AddonSettings.cpp
Paxxi: That means moving it outside the method where it's used. Too ugly for an unknown gain, likely not a hot path so I doubt the extra allocations would be noticable

peak3d: The sum of those "minor" things make it big.
I'm really not a friend of consuming CPU power for nothing.

Malloc is not noticable? Will be most possible the worst thing on stl types.
Why does STL implementation does not free the resourcen on clear() ? Performance!

MartijnKaijser:
@paxxi anything left or need updating?

xbmc/addons/settings/AddonSettings.cpp
Paxxi: Optimize where it's proven that we're slow. Complicating code for optimizations we don't know are required is the reason our code base is fairly messy and weird. 

Had the variable/loop been in the same method i would have made the change but that's not the case now.

Paxxi:
Good to go @martijn. Was waiting on me looking into peaks comment but no changes are warranted there

PR #12537, opened 2017-07-20
[bluray][simple menu] Search for a user provided list of playlists.

ace20022:


xbmc/filesystem/BlurayDirectory.cpp
```diff
@@ -291,4 +296,57 @@ std::string CBlurayDirectory::HexToString(const uint8_t *buf, int count)
...
+
+    CRegExp pl(true);
+    if (!pl.RegComp("(\\d+)"))
+    {
+      file.Close();
+      return userTitles;
+    }
+
+    uint8_t maxLines = 100;
+    while ((maxLines > 0) && file.ReadString(buffer, 1024))
```
Rechi: maxLines > 0 is always true as maxLines is unsigned

xbmc/filesystem/BlurayDirectory.cpp
```diff
@@ -291,4 +296,57 @@ std::string CBlurayDirectory::HexToString(const uint8_t *buf, int count)
...
+    {
+      maxLines--;
+      if (StringUtils::StartsWithNoCase(buffer, "playlists"))
+      {
+        int pos = 0;
```
Rechi: `unsigned int pos = 0;`

xbmc/filesystem/BlurayDirectory.cpp
```diff
@@ -291,4 +296,57 @@ std::string CBlurayDirectory::HexToString(const uint8_t *buf, int count)
...
+        while ((pos = pl.RegFind(buffer, static_cast<uint32_t>(pos))) >= 0)
```
Rechi: remove the cast

Rechi (commented):
regex handling looks correct
just a note: leading zeros are also counted for digit length

ace20022:
@koying ```stoi``` doesn't work for android currently. Do you know of a simple alternative/solution?

koying:
Just not in the NDK, yet: https://github.com/android-ndk/ndk/issues/82
Good ol' `strtol`is your friend, I guess

xbmc/filesystem/BlurayDirectory.cpp
ace20022: it's ```>``` not ```>=```

xbmc/filesystem/BlurayDirectory.cpp
ace20022: doesn't work, since ```pl.RegFind``` might returns -1, leading to an infinite loop.

xbmc/filesystem/BlurayDirectory.cpp
Rechi: sorry for the noise, you are right

xbmc/filesystem/BlurayDirectory.cpp
Rechi: at least replace `uint32_t` with `unsigned int`, which is what RegFind signature uses

ace20022:
Yes, that worked. Thx both of you

PR #12538, opened 2017-07-20
[cmake] Exclude *.xml.include files from install/package generation

hudokkow:
## Description
Skip foo.xml.include files when packaging add-ons

## Motivation and Context
@garbear complained

## How Has This Been Tested?
Zero tests. I'm sure some regex match is possible. Just want to put it here so garbear can test

## Types of change

### DIFF
--- a/cmake/scripts/common/AddonHelpers.cmake
+++ b/cmake/scripts/common/AddonHelpers.cmake
@@ -230,7 +230,9 @@ macro (build_addon target prefix libs)
     set(CPACK_COMPONENTS_IGNORE_GROUPS 1)
     list(APPEND CPACK_COMPONENTS_ALL ${target}-${${prefix}_VERSION})
     # Pack files together to create an archive
-    install(DIRECTORY ${target} DESTINATION ./ COMPONENT ${target}-${${prefix}_VERSION} PATTERN "*.xml.in" EXCLUDE)
+    install(DIRECTORY ${target} DESTINATION ./
+                                COMPONENT ${target}-${${prefix}_VERSION}
+                                REGEX ".+\.xml\.in(clude)?$" EXCLUDE)
     if(WIN32)
       if(NOT CPACK_PACKAGE_DIRECTORY)
         # determine the temporary path
@@ -323,7 +325,8 @@ macro (build_addon target prefix libs)
     if (${prefix}_CUSTOM_BINARY)
       install(FILES ${LIBRARY_LOCATION} DESTINATION ${CMAKE_INSTALL_LIBDIR}/addons/${target} RENAME ${LIBRARY_FILENAME})
     endif()
-    install(DIRECTORY ${target} DESTINATION ${CMAKE_INSTALL_DATADIR}/addons PATTERN "*.xml.in" EXCLUDE)
+    install(DIRECTORY ${target} DESTINATION ${CMAKE_INSTALL_DATADIR}/addons
+                                REGEX ".+\.xml\.in(clude)?$" EXCLUDE)
     if(${prefix}_CUSTOM_DATA)
       install(DIRECTORY ${${prefix}_CUSTOM_DATA} DESTINATION ${CMAKE_INSTALL_DATADIR}/addons/${target}/resources)
     endif()


garbear:
To expand on this: I wanted to generate a file at cmake-time, but the add-on didn't have a way to pass substitution variables to the build system. I created my own substitution routine, but had to name the file `settings.xml.include` because `settings.xml.in` was incorrectly processed by Kodi's build system.

hudokkow:
Hmm, then we need to see what's missing in add-ons cmake code and expand functionality.

Razzeee:
So is this obslete then or will you change it to that functionality?

hudokkow:
Need to revisit but not enough hours in the day.

AlwinEsch:
Can be also in and make is more safe if more addons are coming and someone use a *.xml.include.

hudokkow:
To clarify, @garbear's problem with file generation at build time is independent of this.

This can go in but I would prefer a regex match.
Played 2m with an online generator and came up with `\.[xml\.in]+`

Yeah I know, crappy regex capabilities. Who's the resident regex guru?

AlwinEsch:
If you find him tell me ;) I'm also struggling with regex around and with very little success.

Rechi:
@hudokkow question is do you only want to combine the two separate patterns (`*.xml.in` and `*.xml.include`) into one?
`\.[xml\.in]+` doesn't look correct in any case.

cmake/scripts/common/AddonHelpers.cmake
```diff
@@ -322,7 +325,9 @@ macro (build_addon target prefix libs)
     if (${prefix}_CUSTOM_BINARY)
       install(FILES ${LIBRARY_LOCATION} DESTINATION ${CMAKE_INSTALL_LIBDIR}/addons/${target} RENAME ${LIBRARY_FILENAME})
     endif()
-    install(DIRECTORY ${target} DESTINATION ${CMAKE_INSTALL_DATADIR}/addons PATTERN "*.xml.in" EXCLUDE)
+    install(DIRECTORY ${target} DESTINATION ${CMAKE_INSTALL_DATADIR}/addons
+                                PATTERN "*.xml.in" EXCLUDE)
```
xhaggi: `)`  at the end?

Razzeee:
I would do something like this `.+\.xml\.(include|in){1}` but I don't know which regex engine cmake is using

Rechi:
working regex example is `install(DIRECTORY ${target} DESTINATION ./ COMPONENT ${target}-${${prefix}_VERSION} REGEX ".+\.xml\.in(clude)?$" EXCLUDE)`
It will exclude the same file which are excluded with the current solution.

hudokkow:
Thanks for the help. Updated with @Rechi's suggestion.
Oddly, every online regex tester I could find refused to match with this regex.

> file1.xml.in
file2.xml.include
file3.xml

Completely untested.

Razzeee:
@hudokkow 
you might have to enable multiline matching for `$` to work in most online testers

Rechi (approved):

hudokkow:
yeah, @Rechi already pointed that in slack. All good!

Rechi:
Jenkins failure is unrelated, as this code path wasn't executed.

MilhouseVH:
Getting this build failure with pvr.mythtv and pvr.zattoo (possibly others):
```
-- MYTHTV_VERSION=5.4.0
-- Added usage definition: ADDON_GLOBAL_VERSION_MAIN_USED
-- Added usage definition: ADDON_GLOBAL_VERSION_GUI_USED
-- Added usage definition: ADDON_INSTANCE_VERSION_PVR_USED
CMake Error at /home/neil/projects/LibreELEC.tv/build.LibreELEC-Generic.x86_64-9.0-devel/toolchain/x86_64-libreelec-linux-gnu/sysroot/usr/share/kodi/cmake/AddonHelpers.cmake:328 (install):
  Syntax error in cmake code at

    /home/neil/projects/LibreELEC.tv/build.LibreELEC-Generic.x86_64-9.0-devel/toolchain/x86_64-libreelec-linux-gnu/sysroot/usr/share/kodi/cmake/AddonHelpers.cmake:329

  when parsing string

    .+\.xml\.in(clude)?$

  Invalid escape sequence \.
  CMakeLists.txt:54 (build_addon)


```

cmake is 3.7.2 on Linux.

More complete log: http://sprunge.us/PeGG

Rechi:
@MilhouseVH can confirm issue for pvr.mythtv (and others) and I've a fix. Will create a PR after testing it.

PR #12540, opened 2017-07-20
fix build without bluray

notspiff:
No BlurayDirectory for you!


ace20022:
@FernetMenta this should be enough, right?

FernetMenta:
I can't test this on OSX. Have no Linux system here atm

ace20022:
@garbear could you test it please?

Rechi (approved):
compiled for osx and windows with disabled (and enabled) bluray support

PR #12541, opened 2017-07-21
[VideoPlayer] WinRenderBuffer: fix possible memory access violation.

afedchin:
see title


peak3d:
Fixes the segfaults on all my win32 devices when using AddonVideoCodec streams 👍

PR #12543, opened 2017-07-21
Add static analysis tests

notspiff:
I did this for work and I figured I might as well offer it up here as well.

This adds static analysis tests using clang-check and cppcheck to the buildsystem.
You can execute these using ctest -C cppcheck|clang-check|analyze

To have informative pass/failure states shell scripts need to sit between the invocation and the testing tool. This sadly makes it tied to linux or osx.

Integration on kodi side is very quick so I expect refinements need to be done. But first let's see if there is interest..


Paxxi:
Looks like a good idea to me. Visual Studio has similar passing by /analyze so we can probably add it on windows that way and they might catch different issues

notspiff:
yup.

Rechi:
@notspiff bluray build commit has to be dropped

notspiff:
would have been autor.dropped on pull, but anyways, rebased.

wsnipex:
nice! Mind adding it to https://github.com/xbmc/xbmc/blob/master/tools/buildsteps/linux64/run-tests and  https://github.com/xbmc/xbmc/blob/master/tools/buildsteps/osx64/run-tests ?

notspiff:
i do not think that is a good idea. it easily takes 4h to execute.

wsnipex:
heh, I kinda underestimated that :)

MartijnKaijser:
If we could dump the results somewhere we can run it once a week during the night

notspiff:
uploading the ctest results to jenkins rather than 'dumping them somewhere' sounds like a better idea to me ;)

wsnipex:
good to go?

PR #12548, opened 2017-07-22
[settings] introduce new setting 'Startup action' and add 'Play TV/Radio'

xhaggi:
This PR is a follow-up for #12535. It introduces a new setting `Perform on startup` under `Interface > Startup`. It should be used to unify the various actions Kodi can perform on startup, at the moment only tow 'Play TV' and 'Play Radio'. But this can be extended by something @ksooo has in mind, 'Resume last played content'.

I've also changed the way the startup action 'Play TV/Radio' is performed. It always start playback of tv or radio no matter if a channel was last viewed or not. It will fallback if no last viewed channel exists to the first channel in first TV or Radio group. (Can be verified if reset the TV database and restart Kodi). I've also dropped the 'last played on app quit' stuff from the `PlayChannelOnStartup()` method, because this confusing the user if sometimes Live TV starts and sometimes not. I don't drop the implementation itself. If we later add a 'Resume' startup action it would be reused there.

## Motivation and Context

Not many users differ between foreground and background, but want Kodi to playback TV on startup no matter if they played a channel on app quit. They use Kodi as TV receiver replacement (me too).

## How Has This Been Tested?

Runtime tested on MacOS

screenshot006
screenshot007

## Types of change


addons/resource.language.en_gb/resources/strings.po
```diff
@@ -2294,7 +2294,12 @@ msgctxt "#507"
 msgid "Sort by: Usage"
 msgstr ""
 
-#empty strings from id 508 to 509
+#: system/settings/settings.xml
+msgctxt "#508"
```
MartijnKaijser: Please only use high string id numbers

xhaggi: sure will change it after reviewed by @ksooo and someone else

xhaggi: pushed a fixup for that

ksooo:
If the setting is named "playback live tv", why does it start playback of radio channels? Either it needs another name or the implementation should be changed to ignore radio channels.

xhaggi:
The setting is named `Startup action` and the option is named `Play Live TV` .. IIRC we call it Live TV for both tv and radio or PVR. Let me know what you prefer. Dropping support for radio is not a solution because `GetLastPlayedChannel` returns the last played tv or radio channel and that should not be changed here.

ksooo:
Live TV is not radio. We took really care to separate TV from radio everywhere in Kodi's UI. We should stay clean with our terminology. Maybe we can name the action "Play TV/Radio".

xhaggi:
Sure, thought the same. Done.

da-anda:
any objections to split TV from radio? If we split these in the UI AND channel groups/numbers they should also be separate actions IMO. Would also have the benefit that even if you listened to radio last, you end up in TV mode on startup if you use Kodi as settop box replacement.

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -20593,3 +20580,16 @@ msgstr ""
 msgctxt "#39024"
 msgid "Dependencies"
 msgstr ""
+
+#empty strings from id 39025 to 39099
+
+#: system/settings/settings.xml
```
da-anda: mind adding a contextual help string here?
`#. Label of a setting to specify an action that should be triggered on startup of Kodi`

xhaggi: Sure

xhaggi:
>any objections to split TV from radio?

No, but do we need a third option for both? For me only TV is really necessary. But let me know if we need both like now too.

xhaggi:
BTW should we also add a new sub category `Startup` and move both startup window and startup action setting from `Other` to it.

jjd-uk:
@xhaggi I also thought of that after seeing this PR. Considering moving RSS option and renaming **Other** to **Startup**.

xhaggi:
Np, but in which subcategory should we move the RSS setting?

jjd-uk:
What I was thinking was moving RSS to **Skin** category since it does impact look of skin.

xhaggi:
Sure, will address all at the evening

addons/resource.language.en_gb/resources/strings.po
xhaggi: Done

xhaggi:
Addressed everything. The option `Play TV/Radio` is split into `Play TV` and `Play Radio`. I've moved the RSS setting to category `Skin` and renamed the category `Other` to `Startup`.

jjd-uk:
I didn't actually mean for you to do it, I was going to look at doing it as a follow up to this PR, however you've save me the trouble :)

xhaggi:
:) .. it was really easy because I'm on it.

xbmc/pvr/PVRGUIActions.cpp
```diff
@@ -1198,30 +1198,36 @@ namespace PVR
...
+    {
+      group = groups->GetLastPlayedGroup(item->GetPVRChannelInfoTag()->ChannelID());
+    }
+    else
     {
-      CLog::Log(LOGNOTICE, "PVRGUIActions - %s - continue playback on channel '%s'", __FUNCTION__, channel->ChannelName().c_str());
-      CServiceBroker::GetPVRManager().SetPlayingGroup(CServiceBroker::GetPVRManager().ChannelGroups()->GetLastPlayedGroup(channel->ChannelID()));
-      return SwitchToChannel(item, true, iPlayMode == CONTINUE_LAST_CHANNEL_IN_FOREGROUND);
+      // fallback to first channel
+      PVR_CHANNEL_GROUP_SORTED_MEMBERS channels(group->GetMembers());
```
Paxxi: using `auto channels = group->getMembers();` feels a bit cleaner

xbmc/pvr/PVRGUIActions.cpp
```diff
@@ -1198,30 +1198,36 @@ namespace PVR
...
+      if (channels.empty())
+        return false;
+
+      item.reset(new CFileItem(channels.front().channel));
```
Paxxi: using `item = std::make_shared<CFileItem>(channels.front().channel);` is prefered way but we're quite bad at making use of it.

Paxxi (commented):
C++ wise it looks sane to me.

Nitpicks I'd use auto a lot more here and possibly use an enum class instead of plain enum.

Nothing that requires changes though.

xhaggi:
Thanks @paxxi will address your changes

xhaggi:
Rebased and @Paxxi addressed your comments.

xbmc/pvr/PVRGUIActions.cpp
```diff
@@ -1198,30 +1198,36 @@ namespace PVR
...
-  bool CPVRGUIActions::ContinueLastPlayedChannel() const
+  bool CPVRGUIActions::PlayChannelOnStartup() const
   {
-    const CFileItemPtr item(CServiceBroker::GetPVRManager().ChannelGroups()->GetLastPlayedChannel());
-    const CPVRChannelPtr channel(item ? item->GetPVRChannelInfoTag() : CPVRChannelPtr());
-    bool bWasPlaying = false;
-    if (channel)
-    {
-      // Obtain previous 'was playing on last app quit' flag and reset it, then.
-      channel->SetWasPlayingOnLastQuit(false, bWasPlaying);
```
ksooo: Please do not remove this call. The channel that was playing on app exit must be reset on Kodi startup. Otherwise we easily can end up with multiple channels flagged as "playing on app quit" in TV database.

Since you do not need the value of the flag, please insert a call `channel->SetWasPlayingOnLastQuit(false);`.

ksooo (changes requested):

xbmc/pvr/PVRGUIActions.cpp
xhaggi: As we don't use the feature at the moment we should remove the calls to `SaveLastPlayedChannel` in `CPVRManager::Deinit()` and `CPVRManager::OnSleep()`. I'll comment out the lines in a separate commit.

BTW if we re-add the feature we should reset each channel flagged as `played on last app quit` not only one. That way it never comes to an illegal state you described.

xhaggi:
@ksooo added a separate commit which comment out `SaveLastPlayedChannel` and add some useful comments.

ksooo:
> added a separate commit which comment out SaveLastPlayedChannel and add some useful comments.

 Please, no. Commenting out something that is not used sooner or later leads to breakage. Either save and reset the flag (even if we do not use it at the moment) or remove the whole "save/reset playing channel" code if you think we will never again need it.

xhaggi:
Dropped the whole implementation of 'was playing on app quit'.

ksooo (approved):

ksooo:
Good to go, imo.

xhaggi:
jenkins build this please

xbmc/pvr/PVRDatabase.cpp
```diff
@@ -143,6 +143,9 @@ void CPVRDatabase::UpdateTables(int iVersion)
 
   if (iVersion < 30)
     m_pDS->exec("ALTER TABLE channels ADD bWasPlayingOnQuit bool");
+
+  if (iVersion < 31)
+    m_pDS->exec("ALTER TABLE channels DROP bWasPlayingOnQuit");
```
xhaggi: There is no support for dropping columns in SQL lite. Sorry for that, I thought it is. Kodi will report the following error. We'll provide a fix.

```
10:02:07.370 T:140472773244672  NOTICE: Old database found - updating from version 30 to 31
10:02:07.382 T:140472773244672  NOTICE: Attempting to update the database TV31 from version 30 to 31
10:02:07.383 T:140472773244672   ERROR: SQL: [TV31.db] SQL error or missing database
                                            Query: ALTER TABLE channels DROP bWasPlayingOnQuit
10:02:07.383 T:140472773244672   ERROR: Exception updating database TV31 from version 30 to 31
10:02:07.383 T:140472773244672   ERROR: Error updating database TV31 from version 30 to 31
```

bam80:
Unfortunately, this new change seem breaks Play TV on wakup from suspend feature, which was essential part of the functionality in Kodi 17.
Please, bring it back. The feature is practically useless without it.

PR #12550, opened 2017-07-22
VideoPlayer, PVR: remove deprecated streamURL related functions

FernetMenta:
Drops streamURL as announced more than a year ago:
https://forum.kodi.tv/showthread.php?tid=259830

This is a necessary step in mirgrating pvr streaming relevant functions to inputstream API.


FernetMenta:
objections?

ksooo:
Generally no objections, but could you please also remove the related definitions from the pvr api headers and bump the api? So it should become very clear that stream urls will no longer work.

FernetMenta:
> but could you please also remove the related definitions from the pvr api headers and bump the api

yes, that would be the next step. I wanted to do this separately.

PR #12552, opened 2017-07-23
PVR: bump api to 5.4.0 - remove ChannelSwitch and streamURL

FernetMenta:
see title

PVR addons that did not use streamURL only need a recompile and version bump. Others are broken anyway


FernetMenta:
API change announced: https://forum.kodi.tv/showthread.php?tid=97764&pid=2621261#pid2621261

I would merge this next weekend if no objections.

ksooo:
No objections from my side.

margro:
I'm against this pull without a proper alternative. See https://forum.kodi.tv/showthread.php?tid=259830&pid=2621445#pid2621445 for my problems with this.

FernetMenta:
I announced this more than a year ago. Plenty of time for suggesting alternatives. This did not happen but still time until v18 will be released. Don't expext others to do all all the work.

margro:
The announcement one year ago was not really clear to me. Removing certain functions is ok, but I expected an alternative way of letting Kodi play an URL. Earlier posts from others in that thread had similar questions.
So far I haven't seen a way for sending an URL to the Kodi player. That is where my surprise comes from.

FernetMenta:
> So far I haven't seen a way for sending an URL to the Kodi player.

Why don't you make suggestions for this? Join forces with @rbuehlma who has similar but slightly different requirements. Start discussion of possible solutions in the forum, feel free to submit PRs for reviews. I am happy to assist.

ksooo:
I suggest to merge this now, as the add-ons are already broken functionally. Currently, thx loaf just fine, but playing channels us broken. Merging this PR makes the add-ons also technically incompatible (they won't load anymore), which at least takes us to a consistent  state.

FernetMenta:
Don't you think addon authors of not broken addons should have the time to prepare for the change.

ksooo:
For the not broken add-ons the change should be a no-brainer. Remove SwitchChannel implementation and occurrences of PVR_CHANNEL::strStreamURL" struct member, right? Or do I miss something?

FernetMenta:
Right, see my PR to hts.

ksooo:
Exactly, that's where my judgment is coming from. So, merge it?

rbuehlma:
With this change all the addons using strStreamURL will not work anymore. In my understanding, only the addons providing the stream by their own will still work.
If the plan is to also remove the stream related functions from the PVR API, won't that break all the remaining addons?

ksooo:
The add-ons are already broken, as the respective Kodi core functionality already got removed. So, not merging this PR will not fix anything function-wise. Merging this PR just makes it also technically consistent. See my previous comment from today.

ksooo:
A well, and this will not be the last API change for Leia, yes. And yes, add-ons may break some more time until Leia gets released.

FernetMenta:
I am ok with merging it now

ksooo:
> I am ok with merging it now

 Do it. 😈

MartijnKaijser:
better mark them "broken" with the API bump then let them half function.

da-anda:
@FernetMenta, to guide the PVR add-on authors, do you have a rough concept what these add-ons could do to work again? Would you suggest that they implement a custom demuxer, wait until binary add-on can register to multiple extension points and then go the inputstream route or something completely different? You have the most insight atm, so some advice for them might be good and not just "a go figure out on your own". I just would like to avoid that these devs investigate time in a possible solution that might get obsolete until Leia final when PVR is running via inputstream.

FernetMenta:
@da-anda see https://github.com/xbmc/xbmc/pull/12552#issuecomment-317274083

so far I haven't seen much interest of addon devs to work on this.

all: please stop commenting on this pr, we have a development section in the forum

PR #12555, opened 2017-07-23
[backport] exit GIL when calling into GUI

marcelveldt:
## Description
backport from https://github.com/xbmc/xbmc/pull/12504
fixes GUI lock issues with the Python GIL

## Motivation and Context
Fixes the issue where Kodi would lockup or even completely crash when a python add-on itself is serving content to Kodi (like e.g. an audio stream streaming through an internal webserver) and another or the same python add-on is reading a property from the Kodi GUI at the exact same time.
Issue is confirmed fixed with the original PR by FernetMenta and as it was not that hard to back port, I consider it useful in the Krypton 17.4 release too.

## How Has This Been Tested?
Tested on my dev machine (Windows x64). Confirm that Python interaction still works as intended and that the original issue is now fixed.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/interfaces/legacy/AddonUtils.h
+++ b/xbmc/interfaces/legacy/AddonUtils.h
@@ -31,6 +31,7 @@
 
 #include "threads/SingleLock.h"
 
+#include <memory>
 #include <vector>
 
 #ifdef TARGET_WINDOWS
@@ -52,11 +53,11 @@ namespace XBMCAddonUtils
   class GuiLock
   {
   public:
-    GuiLock();
+    GuiLock(XBMCAddon::LanguageHook* languageHook);
     ~GuiLock();
 
   protected:
-    XBMCAddon::LanguageHook* languageHook;
+    XBMCAddon::LanguageHook* m_languageHook = nullptr;
   };
 
   class InvertSingleLockGuard
@@ -66,8 +67,6 @@ namespace XBMCAddonUtils
     InvertSingleLockGuard(CSingleLock& _lock) : lock(_lock) { lock.Leave(); }
     ~InvertSingleLockGuard() { lock.Enter(); }
   };
-
-#define LOCKGUI XBMCAddonUtils::GuiLock __gl
 
   /*
    * Looks in references.xml for image name
--- a/xbmc/interfaces/legacy/Control.cpp
+++ b/xbmc/interfaces/legacy/Control.cpp
@@ -150,7 +150,7 @@ namespace XBMCAddon
     {
       if (!pGUIControl) return NULL;
 
-      LOCKGUI;
+      XBMCAddonUtils::GuiLock lock(languageHook);
       return ((CGUITextBox*) pGUIControl)->GetDescription();
     }
 
@@ -241,7 +241,7 @@ namespace XBMCAddon
 
       if (pGUIControl)
       {
-        LOCKGUI;
+        XBMCAddonUtils::GuiLock lock(languageHook);
         ((CGUIButtonControl*)pGUIControl)->PythonSetLabel(strFont, strText, textColor, shadowColor, focusedColor);
         ((CGUIButtonControl*)pGUIControl)->SetLabel2(strText2);
         ((CGUIButtonControl*)pGUIControl)->PythonSetDisabledColor(disabledColor);
@@ -254,7 +254,7 @@ namespace XBMCAddon
 
       if (pGUIControl)
       {
-        LOCKGUI;
+        XBMCAddonUtils::GuiLock lock(languageHook);
         ((CGUIButtonControl*)pGUIControl)->PythonSetDisabledColor(disabledColor);
       }
     }
@@ -263,7 +263,7 @@ namespace XBMCAddon
     {
       if (!pGUIControl) return NULL;
 
-      LOCKGUI;
+      XBMCAddonUtils::GuiLock lock(languageHook);
       return ((CGUIButtonControl*) pGUIControl)->GetLabel();
     }
 
@@ -271,7 +271,7 @@ namespace XBMCAddon
     {
       if (!pGUIControl) return NULL;
 
-      LOCKGUI;
+      XBMCAddonUtils::GuiLock lock(languageHook);
       return ((CGUIButtonControl*) pGUIControl)->GetLabel2();
     }
 
@@ -331,7 +331,7 @@ namespace XBMCAddon
     {
       strFileName = imageFilename;
 
-      LOCKGUI;
+      XBMCAddonUtils::GuiLock lock(languageHook);
       if (pGUIControl)
         ((CGUIImage*)pGUIControl)->SetFileName(strFileName, false, useCache);
     }
@@ -341,7 +341,7 @@ namespace XBMCAddon
       if (cColorDiffuse) sscanf(cColorDiffuse, "%x", &colorDiffuse);
       else colorDiffuse = 0;
 
-      LOCKGUI;
+      XBMCAddonUtils::GuiLock lock(languageHook);
       if (pGUIControl)
         ((CGUIImage *)pGUIControl)->SetColorDiffuse(colorDiffuse);
     }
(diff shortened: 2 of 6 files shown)


xbmc/interfaces/legacy/AddonUtils.h
```diff
@@ -52,11 +53,12 @@ namespace XBMCAddonUtils
   class GuiLock
   {
   public:
-    GuiLock();
+    GuiLock(XBMCAddon::LanguageHook* languageHook, bool offScreen);
```
FernetMenta: offscreen was introduced in v18. I won't backport this.

xbmc/interfaces/legacy/ListItem.h
```diff
@@ -83,13 +83,15 @@ namespace XBMCAddon
     public:
 #if !defined SWIG && !defined DOXYGEN_SHOULD_SKIP_THIS
       CFileItemPtr item;
+      bool m_offscreen;
```
FernetMenta: drop this, no offscreen updates in v17

xbmc/interfaces/legacy/AddonUtils.h
marcelveldt: OK, will do. I thought it was more work to remove it from the code than to leave it in but I can understand that v17 has a feature freeze. I'll update the PR

marcelveldt:
@FernetMenta the offscreen feature is now removed, I squashed the commit.

xbmc/interfaces/legacy/Control.cpp
```diff
@@ -150,7 +150,7 @@ namespace XBMCAddon
     {
       if (!pGUIControl) return NULL;
 
-      LOCKGUI;
+      XBMCAddonUtils::GuiLock(languageHook, false);
```
FernetMenta: there is no 2nd parameter anymore, right?

MartijnKaijser:
I think you forgot to remove some parts

xbmc/interfaces/legacy/Control.cpp
marcelveldt: huh? strange... It's gone on my end... I'll force push the commit again

marcelveldt:
@MartijnKaijser @FernetMenta repushed to git and all fine now.

FernetMenta:
thanks, looks ok ow

MartijnKaijser:
Still build issues

marcelveldt:
it's indeed failing to build on my end too, I seem to be missing something very obvious but it looks okay. @FernetMenta do you spot my rookie mistake here ?

xbmc/interfaces/legacy/Control.cpp
```diff
@@ -150,7 +150,7 @@ namespace XBMCAddon
     {
       if (!pGUIControl) return NULL;
 
-      LOCKGUI;
+      XBMCAddonUtils::GuiLock(languageHook);
```
Rechi: shouldn't this be `m_languageHook`?

marcelveldt: According to the original PR not:
https://github.com/xbmc/xbmc/pull/12504/files#diff-c7ffaf248f1f66ae2b5477774ff399a3
m_languageHook is a private var inside the class

marcelveldt: It is strange though, all I did was remove the offScreen part, which was really straight forward. So either I made a very small typo which I am obviously overlooking here or there is another place in the code where the constructor for the GuiLock is defined but that doesn't make sense to me.

Rechi: You are right `languageHook` is correct.
Looked at it again and replacing `XBMCAddonUtils::GuiLock(languageHook);` with `XBMCAddonUtils::GuiLock::GuiLock(languageHook);` everywhere worked for me on windows.

marcelveldt: That does seem to do the trick indeed but I don't understand why as it's not needed in the master branch (original PR). Before the removal of the offScreen argument it was building fine.
What do to with it? Update the PR with your proposed change and leave it like that ? I'm not exactly a C++ hero (but learning slowly) so I want to understand what's happening.

marcelveldt: @FernetMenta any idea why this happens ? If I place back the second (bool) offscreen argument again  it does compile just fine. When I leave it out, the compiler complains about a missing default constructor. Adding an empty default constructor which takes no parameters makes the code compile again but Kodi will freeze on execution as the thread guard is bypassed. The solution as reached out by @Rechi with explicitly defining the classname does the trick but I don't understand why. If it's no big deal, I can change it and this PR is done, if it's unsure maybe better to close the PR and wait for v18 ? It's not that big of a deal. Users face an incidental lockup of Kodi caused by Python adding which will be fixed in v18.

a1rwulf: Would be interesting.
Seems like a weird namespace related issue to me.
If you retrieve the languageHook via the getter method `GetLanguageHook` it compiles too.

marcelveldt: Yes, correct, just did that as a test. Weird stuff.

FernetMenta: @marcelveldt I don't know. I don't even remember last time I compiled v17.

LipkeGu: i noticed that a couple of python addons has this problem that gets kodi to lockup or crashing!

marcelveldt: Yes, that's indeed what this PR is all about...

marcelveldt: @Rechi are you okay if I change the PR to your suggestion by explicitly defining the namespace+classname ? It does build indeed fine with that change and I did some testing and did not experience the lockup issue so far.

Rechi: yes, please do it

marcelveldt:
@Rechi I've updated the PR with the suggested change of explicitly defining the namespace + GuiLock classname. It builds without issues now. Did some extended testing with impossible python scenarios and it doesn't break things. I did notice the lockup once but it didn't crash Kodi nor the player, instead it just continued after the freeze which is a much better behaviour than the previous crashing behaviour. I assume the freeze is just caused by the guilock kicking in on python's request.

This was just once after really stress testing the python implementation. Without the fix the fix I could reproduce the crash/freeze within seconds.

Thanks everyone for participating

MartijnKaijser:
Doesn't compile

marcelveldt:
Hmmm, it does indeed compile on Windows with the namespace fix but now it doesn't on all other platforms. Wtf, mysterious stuff. 

With nobody seeing a resolution I'd suggest to just close this PR and list the freeze issue as a known issue with v17. It's not happening that often, it's more like a corner case  when 2 python addons request a lock at the exact same time. It's fixed in v18 but somehow that code can't be ported to v17.

a1rwulf:
Have you tried to use
`XBMCAddonUtils::GuiLock(GetLanguageHook());`
instead of
`XBMCAddonUtils::GuiLock::GuiLock(languageHook);`

GetLanguageHook is an inline Getter in AddonClass.h.
This makes it compile on Linux for me.

marcelveldt:
@a1rwulf doesn't that reinitialise the languagehook at every time the guiclock is called ?

FernetMenta:
add variables https://github.com/xbmc/xbmc/pull/12590

I messed this up using find/replace

a1rwulf:
@marcelveldt I don't think so (however I haven't looked too close to the issue tbh)
`inline LanguageHook* GetLanguageHook() { return languageHook; }`
For me this looks like you just get the plain pointer without any additional work done.
Which should make no difference to what you have, namely using the protected member.
Maybe anyone can confirm my theory?

marcelveldt:
The missing variable name as pointed out by FernetMenta seems to be the issue. Still strange why the v18 code compiles just fine and it's giving issues in the v17 branch but glad it's discovered. I'll merge the changes, do some tests and update the PR

a1rwulf:
Perfect - thanks.

marcelveldt:
building fine now. Thanks @FernetMenta

PR #12562, opened 2017-07-24
[cmake] FindBluray.cmake: fix version check

wsnipex:
properly check libbluray version

## Motivation and Context
reported by @notspiff 
currently the version check is only done using pkg-config. If pkg-config fails (e.g. because of too low version), we still do the manual search without any further version checks.
The effect is that libbluray <=0.9.2 is accepted, even though the find rule tries to enforce 0.9.3

## How Has This Been Tested?
not really

## Types of change

### DIFF
--- a/cmake/modules/FindBluray.cmake
+++ b/cmake/modules/FindBluray.cmake
@@ -14,14 +14,21 @@
 #
 #   Bluray::Bluray   - The libbluray library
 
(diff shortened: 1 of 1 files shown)


notspiff:
ack.

hudokkow (approved):

hudokkow:
lgtm. thx to both!

PR #12565, opened 2017-07-24
[fix] coverity issues

Rechi:
## Description
Fix a few coverity issues

## Motivation and Context
reduce coverity issues

## How Has This Been Tested?
build for windows and osx

## Types of change

## Checklist:

### DIFF
--- a/xbmc/Util.cpp
+++ b/xbmc/Util.cpp
@@ -2352,14 +2352,14 @@ bool CUtil::ValidatePort(int port)
 
 int CUtil::GetRandomNumber()
 {
-#ifdef TARGET_WINDOWS
+#if !defined(TARGET_WINDOWS)
+  return rand_r(&s_randomSeed);
+#else
   unsigned int number;
   if (rand_s(&number) == 0)
     return (int)number;
-#else
-  return rand_r(&s_randomSeed);
-#endif
 
   return rand();
+#endif
 }
 
--- a/xbmc/threads/Thread.h
+++ b/xbmc/threads/Thread.h
@@ -124,7 +124,7 @@ private:
   // -----------------------------------------------------------------------------------
 
   ThreadIdentifier m_ThreadId;
-  ThreadOpaque m_ThreadOpaque;
+  ThreadOpaque m_ThreadOpaque = {};
   bool m_bAutoDelete;
   CEvent m_StopEvent;
   CEvent m_TermEvent;
--- a/xbmc/video/Teletext.cpp
+++ b/xbmc/video/Teletext.cpp
@@ -32,21 +32,11 @@
 #include "filesystem/SpecialProtocol.h"
 #include "guilib/GraphicContext.h"
 
-#define SDL_memset4(dst, val, len)		\
-do {						\
-	uint32_t _count = (len);		\
-	uint32_t _n = (_count + 3) / 4;		\
-	uint32_t *_p = static_cast<uint32_t *>(dst);	\
-	uint32_t _val = (val);			\
-	if (len == 0) break;			\
-        switch (_count % 4) {			\
-        case 0: do {    *_p++ = _val;		\
-        case 3:         *_p++ = _val;		\
-        case 2:         *_p++ = _val;		\
-        case 1:         *_p++ = _val;		\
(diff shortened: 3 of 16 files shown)


xbmc/cores/AudioEngine/Utils/AEUtil.cpp
```diff
@@ -554,12 +554,7 @@ AVSampleFormat CAEUtil::GetAVSampleFormat(AEDataFormat format)
     case AEDataFormat::AE_FMT_RAW:
       return AV_SAMPLE_FMT_U8;
     default:
-    {
-      if (AE_IS_PLANAR(format))
```
FernetMenta: dropping this looks wrong on first glance. please elaborate

Rechi: `AE_IS_PLANAR = AE_FMT_U8P|AE_FMT_S16NEP|AE_FMT_S32NEP|AE_FMT_S24NE4P|AE_FMT_S24NE4MSBP|AE_FMT_S24NE3P|AE_FMT_DOUBLEP|AE_FMT_FLOATP` and all those are already handled explicitly above.

FernetMenta: if you argue like this, you need to remove entire default but this is not what we want to achive with the default case.
coverity is wrong here.

Rechi: what should I do, drop this change?

xbmc/Util.cpp
```diff
@@ -2352,14 +2352,14 @@ bool CUtil::ValidatePort(int port)
 
 int CUtil::GetRandomNumber()
```
Paxxi: This method should be removed or switched to using a proper C11 rand

xbmc/threads/Thread.h
```diff
@@ -124,7 +124,7 @@ class CThread
   // -----------------------------------------------------------------------------------
 
   ThreadIdentifier m_ThreadId;
-  ThreadOpaque m_ThreadOpaque;
+  ThreadOpaque m_ThreadOpaque = { 0 };
```
Paxxi: This isn't entirely correct. On windows this is a HANDLE which is a pointer type

xbmc/video/Teletext.cpp
```diff
@@ -32,21 +32,20 @@
...
-	}					\
-} while(0)
+static inline void SDL_memset4(void* dst, uint32_t val, size_t len)
+{
+  size_t _n = (len + 3) / 4;
+  uint32_t *_p = static_cast<uint32_t *>(dst);
+  if (len == 0)
+    return;
+  switch (len % 4) {
+    case 0: do { *_p++ = val;
```
Paxxi: This is completely broken. What is an unterminated do clause doing here?

Paxxi (changes requested):
Sdl memset needs fixing rest looks good

xbmc/cores/AudioEngine/Utils/AEUtil.cpp
FernetMenta: yes, please

xbmc/video/Teletext.cpp
Rechi: `} while ( --_n );` is a few lines below
I just replaced the define with a function so behaviour hasn't changed.
The define is an exact copy from libsdl.

garbear: This is known as Duff's device: https://en.wikipedia.org/wiki/Duff%27s_device

notspiff:
@Rechi please consider if you want to do something about
```
1037: [/home/akva/kode/kodi-cmake/xbmc/threads/platform/win/ThreadLocal.h:45]: (warning) Class ThreadLocal is not safe, destructor throws exception
```

xbmc/video/Teletext.cpp
Paxxi: Thanks garbear. Leave loop unrolling to the compiler I say. This code looks horrible

Paxxi (approved):

PR #12566, opened 2017-07-24
[bluray] Implement OnNext and OnPrevious.

ace20022:


da-anda:
while I can now seek chapters using pageUp/Down (action up/down is catched by bluray), the chapter index in Kodi doesn't update correctly. In my case it's always showing chapter 0/21. Is this something you can fix?

PR #12573, opened 2017-07-25
[PVR] Fix channel preview OSD after #12510

ksooo:
Fixes some fallout caused by #12510, superseeds #12572.

This has been runtime-tested on macOS, latest Kodi master.


FernetMenta (approved):
thanks for fixing it

PR #12574, opened 2017-07-25
Joysticks: fix easter egg

lrusak:
not sure if @garbear has other plans for this but it was lost in the last major refactor.

This makes it work again.

shhhhhhhhhh

### DIFF
--- a/xbmc/input/joysticks/keymaps/KeymapHandler.h
+++ b/xbmc/input/joysticks/keymaps/KeymapHandler.h
@@ -19,6 +19,7 @@
  */
 #pragma once
 
+#include "input/joysticks/IButtonSequence.h"
 #include "input/joysticks/IInputHandler.h"
 #include "input/joysticks/interfaces/IKeymapHandler.h"
 #include "input/joysticks/JoystickTypes.h"
@@ -62,6 +63,10 @@ namespace JOYSTICK
     virtual bool OnButtonMotion(const FeatureName& feature, float magnitude, unsigned int motionTimeMs) override;
     virtual bool OnAnalogStickMotion(const FeatureName& feature, float x, float y, unsigned int motionTimeMs) override;
     virtual bool OnAccelerometerMotion(const FeatureName& feature, float x, float y, float z) override;
+
+  protected:
+    // Keep track of cheat code presses
+    std::unique_ptr<IButtonSequence> m_easterEgg;
 
   private:
     // Analog stick helper functions
--- a/xbmc/games/ports/Port.h
+++ b/xbmc/games/ports/Port.h
@@ -63,6 +63,7 @@ namespace GAME
     virtual void SetWindowID(int windowId) override { }
     virtual int GetFallthrough(int windowId) const override { return -1; }
     virtual bool UseGlobalFallthrough() const override { return false; }
+    virtual bool UseEasterEgg() const override { return false; }
 
   private:
     // Construction parameters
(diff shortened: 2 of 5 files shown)


garbear:
Thanks! There's only one change I want to make.

The new system added a level of complexity because I wanted game hotkeys (defined in joystick.xml's `<FullscreenGame>` tag) to work in the game OSD. To solve this, I added a "keymap environment" parameter. Outside of games, the environment's job is to translate the key based on the active window ID. Inside games, I register an additional keymapper, and use the environment parameter to force `<FullscreenGame>` keys.

We only need the top-level keymapper to monitor for easter eggs, so I'll add a variable to the keymap environment to prevent easter egg monitoring in game OSD menus. PR is coming your way.

xbmc/input/joysticks/keymaps/KeymapHandler.h
```diff
@@ -19,6 +19,7 @@
  */
 #pragma once
 
+#include "input/joysticks/IButtonSequence.h"
```
garbear: You can remove this include and instead forward-declare `IButtonSequence`

xbmc/input/joysticks/keymaps/KeymapHandler.h
```diff
@@ -63,6 +64,14 @@ namespace JOYSTICK
     virtual bool OnAnalogStickMotion(const FeatureName& feature, float x, float y, unsigned int motionTimeMs) override;
     virtual bool OnAccelerometerMotion(const FeatureName& feature, float x, float y, float z) override;
 
+  protected:
+    /*!
+     * \brief Keep track of cheat code presses
+     *
+     * Child classes should initialize this with the appropriate controller ID.
```
garbear: This comment doesn't apply to the new system

xbmc/input/joysticks/keymaps/KeymapHandler.h
lrusak: without it I get
```
xbmc/input/joysticks/keymaps/KeymapHandler.cpp: In constructor 'KODI::JOYSTICK::CKeymapHandler::CKeymapHandler(IActionListener*, const IKeymap*)':
xbmc/input/joysticks/keymaps/KeymapHandler.cpp:48:59: error: no matching function for call to 'std::unique_ptr<IButtonSequence>::reset(KODI::JOYSTICK::CJoystickEasterEgg*)'
   m_easterEgg.reset(new CJoystickEasterEgg(ControllerID()));
```

garbear: Try moving the `CKeymapHandler` destructor to the .cpp file:

```
CKeymapHandler::~CKeymapHandler() = default;
```

lrusak: nope

garbear: ok, just leave the header then

lrusak:
rebased and squashed

garbear:
jenkins build this please

lrusak:
jenkins build this please

da-anda:
is it still worth having the easter egg feature? Isn't that cat out of the bag already?

garbear:
jenkins build this please

garbear:
> is it still worth having the easter egg feature? Isn't that cat out of the bag already?

Do you have any idea how much technical debt I take on the second games are enabled by default? If we hit the switch today I throw away easily the next six months of my life, likely more. I'm not saying I didn't sign up for that, just that I'm not ready to commitment to this yet.

lrusak:
Not only that but we have to solve the infrastructure issue before we can go full on

PR #12578, opened 2017-07-26
joystick.xml: Fix errors due to unimplemented save actions

garbear:
These were causing errors in the log on startup. These will be restored when the new savestate manager is implemented. (on my radar after the player manager)

Reported here: https://trac.kodi.tv/ticket/17541


garbear:
Just tested, and no save-related errors show up in the log. Jenkins isn't needed here.

PR #12579, opened 2017-07-27
DVDDemuxFFmpeg: Discard repair track from GoPro recordings

popcornmix:
GoPro video files always stutter at the start (e.g. for 20 seconds) and then resume fine.
https://forum.kodi.tv/showthread.php?tid=259292
https://trac.kodi.tv/ticket/17371
https://trac.kodi.tv/ticket/17373

MrMC had a fix and in testing that resolves the reported issue.
https://forum.kodi.tv/showthread.php?tid=259292&pid=2622828#pid2622828

## Description
Video (mp4) from GoPro cameras can have a 'meta' track used for a file repair containing
'fdsc' data, this is also called the SOS track. Ignore it or we will glitch bad on startup.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
@@ -1443,6 +1443,10 @@ CDemuxStream* CDVDDemuxFFmpeg::AddStream(int streamIdx)
       {
         stream = new CDemuxStream();
         stream->type = STREAM_DATA;
+        // Video (mp4) from GoPro cameras can had a 'meta' track used for a file repair containing
+        //  'fdsc' data, this is also called the SOS track. Ignore it or we will glitch bad on startup.
+        if (pStream->codec->codec_tag == MKTAG('f','d','s','c'))
+          pStream->discard = AVDISCARD_ALL;
         break;
       }
     case AVMEDIA_TYPE_SUBTITLE:


xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
```diff
@@ -1443,6 +1443,10 @@ CDemuxStream* CDVDDemuxFFmpeg::AddStream(int streamIdx)
       {
         stream = new CDemuxStream();
         stream->type = STREAM_DATA;
+        // Video (mp4) from GoPro cameras can had a 'meta' track used for a file repair containing
+        //  'fdsc' data, this is also called the SOS track. Ignore it or we will glitch bad on startup.
+        if (pStream->codec->codec_tag == MKTAG('f','d','s','c'))
```
FernetMenta: adding a discareded stream is not really what you want to achieve, right? better not adding it to the map at all

popcornmix: @FernetMenta do you mean this?
```
        stream->type = STREAM_NONE;
```

FernetMenta: nope, AddStream adds a stream to a map of available streams for the demuxer that is queried by player. two lines above a new CDemuxStream object is created that you try to invalidate. I mean don't create it for this case.

popcornmix: Can you explain exactly what is desired here?
Returning nullptr from this function causes playback to fail 
```
17:03:30.360 T:1700598560   ERROR: CDVDDemuxFFmpeg::AddStream - internal error, stream is null
```

FernetMenta: do not allocate a packet and return empty (bReturnEmpty) in this case

popcornmix:
Superceded by https://github.com/xbmc/xbmc/pull/13095

PR #12585, opened 2017-07-28
add dynamic URLs to fileItems

FernetMenta:
Kodi identifies an item by its path. Hence path must never be changed once a FileItem was created. This adds a dynamic path to fileIIems that is allowed to be changed.

Required i.e. by PVR

### DIFF
--- a/xbmc/FileItem.cpp
+++ b/xbmc/FileItem.cpp
@@ -362,6 +362,7 @@ const CFileItem& CFileItem::operator=(const CFileItem& item)
   m_bLabelPreformatted=item.m_bLabelPreformatted;
   FreeMemory();
   m_strPath = item.GetPath();
+  m_strDynPath = item.GetDynPath();
   m_bIsParentFolder = item.m_bIsParentFolder;
   m_iDriveType = item.m_iDriveType;
   m_bIsShareOrDrive = item.m_bIsShareOrDrive;
@@ -486,6 +487,7 @@ void CFileItem::Reset()
   m_strDVDLabel.clear();
   m_strTitle.clear();
   m_strPath.clear();
+  m_strDynPath.clear();
   m_dateTime.Reset();
   m_strLockCode.clear();
   m_mimetype.clear();
@@ -510,6 +512,7 @@ void CFileItem::Reset()
   SetInvalid();
 }
 
+// do not archive dynamic path
 void CFileItem::Archive(CArchive& ar)
 {
   CGUIListItem::Archive(ar);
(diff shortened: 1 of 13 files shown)


ksooo (approved):
As discussed in Slack, I think this is the right way to go.

da-anda:
sorry for being late to the game, but I have a related suggestion. About a 1,5 years ago I tried to cleanup some of the cFileItem mess we have in regards to infodialogs etc. (wanted to provide context menu options within the dialog) When those get opened, Kodi is first creating a clone of the cFileItem and replacing it's VFS path (like library path) with the real file path. That way context menus can no longer be triggered from within these dialogs. Back then, I tried to add a "getPlaybackUrl" method to cFileItems, but got lost at all the places where I would have had to change the function calls.
Now, this PR basically does a similar thing, only that it's not really meant to also be used internally from the naming. What do you think about renaming these methods to "set/getPlaybackUrl" (or alike) and also use this internally whenever we need the playable url? This would finally allow to clean up the infodialogs and get rid of all the "set art", "refresh" and other buttons, and replace them with a single "manage" button that triggers the context menu options for this item.

FernetMenta:
I prefer the name DynamicURL because it best describes what is does. FileItem's path must never be changed because it serves as id for the item. DynamicURL can be changed.
What do you mean by used internally?

da-anda:
to me "DynamicURL" describes it's type but not the intended purpose. But
that's probably just a matter of taste.

With not used internally I mean that to me the name implies that this URL
will be set from the outside (add-ons etc) to override the default URL that
holds the VFS path. But Kodi should also use it by itself instead of
overwriting/changing the default path, which, as you said, should be
immutable. Unfortunately Kodi currently changes the default path
(identifier) on several places (like in the GUI layer when showing video
info dialog IIRC).

Am 29.07.2017 7:48 vorm. schrieb "Rainer Hochecker" <
notifications@github.com>:

da-anda:
that's the branch I was playing around with if you'd like to have a look on the places I had to change https://github.com/da-anda/xbmc/commits/infodialog-context (see 9728662aab44ef830c6d3026a50175e236c80533 for the path changes)

FernetMenta:
Your code allows to change the type of a FileItem after it was created, which is not a good idea. Doing this would allow to a picture or epg item to become a video. This is wrong and would create something way out of control. 
CFileItem already violates all principles of good software design. We need to move this in opposite direction. First get rid of those crappy IsYX() methods.

da-anda:
why does my code (branch) allow to change the type of a FileItem? The snippet quoted above is actual code used in Kodi which I tried to get rid of with my changes. I also tried to get rid of the excessive usage of the "isFoo" methods only to determine the playback URL by resolving the playback URL of DB items already when constructing the cFileItem.

FernetMenta:
> why does my code (branch) allow to change the type of a FileItem?

you use getPlayablePath for decision how to handle the item, track its state, and db operations. 
for example a pvr channel item has to stay a channel with unchanged id (path) for its entire lifetime.  the dynamic URL must not change the way this item is treated by GUI or by database.

this comment: https://github.com/xbmc/xbmc/blob/master/xbmc/FileItem.cpp#L1698
is very valid. id and type of a fileItem define how the item is handled. this must not dynamically change.

da-anda:
ok, might have gone too far with replacing the function calls, but that's
why I stopped, because I lost overview where I might need which path. But
wouldn't you in general prefer if the core wouldn't have to check for
videoinfo and what not in tons of places to obtain the playable URL/path?
That was the main intention of my attempt and it imo would be good to
specify which method should be used for this. Your newly added property
seemed perfect, which is why I started this discussion.

Am 29.07.2017 10:46 vorm. schrieb "Rainer Hochecker" <
notifications@github.com>:

FernetMenta:
> ok, might have gone too far with replacing the function calls, but that's
why I stopped, because I lost overview where I might need which path

good indicator that this approach was not ideal

> But wouldn't you in general prefer if the core wouldn't have to check for
videoinfo and what not in tons of places to obtain the playable URL/path?

we have to do what the linked comment says. remove SetURL from the public interface and fix the code that breaks. Core should not bother with all those different types. Look at how inputstream and videobuffers work. The component that handles the typ does know hoe to handle it.

PR #12586, opened 2017-07-29
pass resume script param when launching addons

faush01:
Recreate pull request to try to fix build errors
Original PR: https://github.com/xbmc/xbmc/pull/12181

## Types of change

## Checklist:

### DIFF
--- a/xbmc/interfaces/legacy/ListItem.cpp
+++ b/xbmc/interfaces/legacy/ListItem.cpp
@@ -482,6 +482,8 @@ namespace XBMCAddon
             videotag.SetTrailer(value);
           else if (key == "path")
             videotag.SetPath(value);
+          else if (key == "filenameandpath")
+            videotag.SetFileNameAndPath(value);
           else if (key == "date")
           {
             if (value.length() == 10)
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -2918,8 +2918,9 @@ bool CApplication::PlayMedia(const CFileItem& item, const std::string &player, i
   //If item is a plugin, expand out now and run ourselves again
   if (item.IsPlugin())
   {
+    bool resume = item.m_lStartOffset == STARTOFFSET_RESUME;
     CFileItem item_new(item);
-    if (XFILE::CPluginDirectory::GetPluginResult(item.GetPath(), item_new))
+    if (XFILE::CPluginDirectory::GetPluginResult(item.GetPath(), item_new, resume))
       return PlayMedia(item_new, player, iPlaylist);
     return false;
   }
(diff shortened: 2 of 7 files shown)


faush01:
ok, this looks to have fixed the build issues.
I dont know enough about the build system to know what happened but it looks like this has resolved it.
if we were just waiting for a build to work then we are done.

da-anda:
FYI, you don't need to recreate PRs in order to update them. All you have to do is to force push to the github branch you created the initial PR from.

faush01:
Yep know that, if you have a look at the original you can see the builds were failing with a weird git error, I could not get the builds on the other PR working.
If you look at the changes for the PR and the other one you will see that the files changed are exactly the same but this PR works and the other one fails consistently with the some error about GIT not finding a tag or something.
This one is build, YAY,  so we should be good to go to merge :-)

faush01:
can we merge this now?

xbmc/interfaces/legacy/ListItem.cpp
```diff
@@ -482,6 +482,8 @@ namespace XBMCAddon
             videotag.SetTrailer(value);
           else if (key == "path")
             videotag.SetPath(value);
+          else if (key == "filenameandpath")
```
da-anda: I don't know all about our inner workings, but this change doesn't seem related, is it?

da-anda:
I know too little about how we deal with resume points internally etc, but wouldn't it in general make more sense to pass a timestamp of the resume position to the scripts/add-ons than just a boolean? Not all scripts might know the resume point. I know that in your case the remote service keeps track of it accross devices (not naming it), but when we add new stuff, we should think about a more generic useage.
But as said - I know too little about our internals and this is suggestion purely based on your PRs changes.

Maybe @FernetMenta know a little more about resume point handling? Don't know whom else to ping about it.

xbmc/interfaces/legacy/ListItem.cpp
faush01: its to fix a problem with addons not being able to set the filenameandpath of the videotag.
this in turn fixes some handeling of items when working out if they are folders or not later on when an item is played from the info screen.

faush01:
I dont have access to the resume point with the minimal changes I have made to the interfaces, just if the item is resuming or not.
The addon should know what the resume point is, it was the one that set the resume point in the ListItem it created for display.

xbmc/interfaces/legacy/ListItem.cpp
Razzeee: Btw how does it differ from "path"? https://github.com/faush01/xbmc/blob/3c82c4fbfdff48a4190a0e7cda2bdde20f338911/xbmc/interfaces/legacy/ListItem.h#L634
I would think it holds the same info? (Which isn't your fault, just checking for docs sake)

PR #12589, opened 2017-07-30
fixed: store created stream in member

notspiff:
I think this is what is intended.


FernetMenta:
that was done by  @AchimTuran and is imo long outdated and not used in current state

notspiff:
So then we keep it broken?

AchimTuran:
Thx for the fix. 

@fernetmenta
Can we merge this?

AchimTuran:
@notspiff
I know you already provided this fix some time ago and I also had a fix. But somehow they never went in. Thank you again for having an eye on it and for providing this fix again.

PR #12591, opened 2017-07-30
FileItem, fix dynamic path

FernetMenta:
I accidentally followed the bad example already there. A copy constructor is a perfect example where using getters does not make sense.


ksooo (approved):
Hehe.

FernetMenta:
win64 build error unrelated

PR #12594, opened 2017-07-31
[win32] fix halt playback after resolution change

afedchin:
see title


afedchin:
win32 error is unrelated, win64 is fine

PR #12595, opened 2017-07-31
[win32] cleanup and fix some includes 

afedchin:
this is first phase in preparation to include uwp changes into mainline.

this fixes:
- DX::DeviceResources contains errors
- implementation of ProcessInfoWin10
- invalid includes of windowing (include implementation header instead of factory header)
- d3d9 includes isn't used actually
- remove HAS_DX ifdef from windows related files.


MartijnKaijser:
unrelated fails

PR #12596, opened 2017-07-31
[addons] change binary callbacks

AlwinEsch:
## Description
----------------------
1. Add binary addon callback to get used skin name id

This add a callback who also present on python and becomes needed
on e.g. Web browser addon to know used skin name.

The change is more general and included to a new request.

----------------------
2. allow binary addon to use label id with every id

This change the id usage of strings on new binary addons to override
a kodi string id with a own for him itself.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/interfaces/General.cpp
+++ b/xbmc/addons/interfaces/General.cpp
@@ -35,6 +35,7 @@
 #ifdef TARGET_POSIX
 #include "linux/XMemUtils.h"
 #endif
+#include "settings/Settings.h"
 #include "utils/CharsetConverter.h"
 #include "utils/log.h"
 #include "utils/LangCodeExpander.h"
@@ -64,6 +65,7 @@ void Interface_General::Init(AddonGlobalInterface* addonInterface)
   addonInterface->toKodi->kodi->get_region = get_region;
   addonInterface->toKodi->kodi->get_free_mem = get_free_mem;
   addonInterface->toKodi->kodi->get_global_idle_time = get_global_idle_time;
+  addonInterface->toKodi->kodi->get_current_skin_id = get_current_skin_id;
   addonInterface->toKodi->kodi->kodi_version = kodi_version;
 }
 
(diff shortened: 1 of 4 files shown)


xbmc/addons/interfaces/General.cpp
```diff
@@ -413,6 +412,18 @@ int Interface_General::get_global_idle_time(void* kodiBase)
...
+char* Interface_General::get_current_skin_name(void* kodiBase)
+{
+  CAddonDll* addon = static_cast<CAddonDll*>(kodiBase);
+  if (addon == nullptr)
+  {
+    CLog::Log(LOGERROR, "Interface_General::%s - invalid data (addon='%p')", __FUNCTION__, addon);
+    return nullptr;
+  }
+
+  return strdup(CServiceBroker::GetSettings().GetString(CSettings::SETTING_LOOKANDFEEL_SKIN).c_str());
```
tamland: The setting stores the skin id, not the name. I'm guessing you actually want is the id; if so the methods should say 'id'

AlwinEsch: Is changed to use id instead of name

AlwinEsch:
jenkins build this please

AlwinEsch:
jenkins build this please

AlwinEsch:
jenkins build this please

AlwinEsch:
jenkins build this please

ksooo:
You know that I always get puzzled about this version number stuff, but imo this change breaks all old style binary add-ons, like pvr add-ons. @AlwinEsch

@MilhouseVH not sure whether I'm actually right, be in case I am, you are informed...

ksooo:
Okay, you did not modify the function tables, right? Thus, all should be fine, but I would be happy if you could confirm this.

AlwinEsch:
PVR still use the old and not affected by them. Also is with a add of callback function (if added to the end) still backward compatible.

ksooo:
Sure, thanks.

PR #12597, opened 2017-07-31
[Music]Fix change of information provider for albums and artists 

DaveTBlake:
A follow up from #12120,  the improvements to scraping additional artist and album information, that fixes flaws in how changes to artist and album information providers are selected and applied. 

Music DB schema changes include:
- Remove albuminfosong table, as storing scraped tracks is deprecated.
- Adding the release group MBID, scrapedMBID flag and lastScraped to album and artist views. This fixes loss of release group MBID when refreshing with a scraper that does not return release group id e.g. TADB scraper or from NFO files, and needed for JSON API exposure of properties.
- Replacing the  content table with (more accurately named) infosetting table and store idSetting in artist and album tables. 

The "Change Info provider" context menu action  was truely broken, especially on saying saying yes to the subsequent "Do you want to refresh information for all items within this path?".  For example:
- doing so for an album it used the _song_ ids as if they were album ids and refreshed the info for those.
- doing it on "*all albums" set the _artist_ scraper, and again used _song_ ids.
- doing it on a parent folder could hang.
- from genres node, a path and settings were stored, but these settings were not applied when adding artists, or albums with songs of that genre, or "Query info for all".
Also "content" (if anyone did use this feature) was not cleaned up when orphaned.

The user access to set/clear these settings was also so limited as to be useless. 
- If a user had set scraper settings for an artist or album, then there was no way to reset them to default. The only way to know they were non-default was to look at the settings for each item at a time, and then they could not be reset, only set something else specifically.
- There was no practical way to set the setting for more than one artist or album  at a time (genre did not work, and anyway what or those with more than one genre). 
- The only way **to set the default settings** themselves was to go to System > Settings > Media > Music and change the addon used, then go to System >addons > My Addons > Information Providers > Album (or artist) and change the addon settings.
I suspect that all users have done in the past is to set the default settings, and then apply them to everything uniformly.

This PR addresses both aspects, to do this there is a **UX/UI change**
- "Change information provider"  is only on the context menu when the current item is an album or artist (from any "artists" or "albums" node or smart playlist)
- The action is followed by submenu offering to set information provider for 
a) current item, 
b) all items on node, filtered by criteria
c) default (for all).
This is followed by the `CGUIDialogContentSetting` dialog to select the info provider addon to use and change settings. Then a comfirmation message is displayed before the setting change is applied. Finally the user is asked if they want to query for additional information using these settings immediately (otherwise they can use "Query Info For all" later).

This gives the facility to set information provider for multiple artists or albums, and to reset to default, without cluttering ordinary music library use.

### DIFF
--- a/xbmc/filesystem/MusicDatabaseDirectory.cpp
+++ b/xbmc/filesystem/MusicDatabaseDirectory.cpp
@@ -121,7 +121,9 @@ void CMusicDatabaseDirectory::ClearDirectoryCache(const std::string& strDirector
 
 bool CMusicDatabaseDirectory::IsAllItem(const std::string& strDirectory)
 {
-  if (StringUtils::EndsWith(strDirectory, "/-1/"))
+  //Last query parameter, ignoring any appended options, is -1
+  CURL url(strDirectory);
+  if (StringUtils::EndsWith(url.GetWithoutOptions(), "/-1/"))
     return true;
   return false;
 }
--- a/xbmc/music/MusicDatabase.h
+++ b/xbmc/music/MusicDatabase.h
@@ -50,8 +50,7 @@ namespace dbiplus
 #define ERROR_DATABASE    315
 #define ERROR_REORG_SONGS   319
 #define ERROR_REORG_ARTIST   321
-#define ERROR_REORG_GENRE   323
-#define ERROR_REORG_ROLE   324
+#define ERROR_REORG_OTHER   323
 #define ERROR_REORG_PATH   325
 #define ERROR_REORG_ALBUM   327
 #define ERROR_WRITING_CHANGES  329
@@ -213,7 +212,7 @@ public:
   */
   bool AddAlbum(CAlbum& album);
 
-  /*! \brief Update an album and all its nested entities (artists, songs, infoSongs, etc)
+  /*! \brief Update an album and all its nested entities (artists, songs etc)
    \param album the album to update
    \return true or false
    */
@@ -261,8 +260,6 @@ public:
                    bool bScrapedMBID);
   bool ClearAlbumLastScrapedTime(int idAlbum);
   bool HasAlbumBeenScraped(int idAlbum);
-  bool HasScrapedAlbumMBID(int idArtist);
-  int  AddAlbumInfoSong(int idAlbum, const CSong& song);
 
   /////////////////////////////////////////////////
   // Audiobook
@@ -307,7 +304,6 @@ public:
   void SetTranslateBlankArtist(bool translate) { m_translateBlankArtist = translate; }
   bool HasArtistBeenScraped(int idArtist);
   bool ClearArtistLastScrapedTime(int idArtist);
-  bool HasScrapedArtistMBID(int idArtist);
   int  AddArtistDiscography(int idArtist, const std::string& strAlbum, const std::string& strYear);
   bool DeleteArtistDiscography(int idArtist);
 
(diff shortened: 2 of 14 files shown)


Rechi:
jenkins build this please

DaveTBlake:
Test build available as **9095c13**, so for win32 in particular 
http://mirrors.kodi.tv/test-builds/windows/win32/KodiSetup-20170801-9095c13-HEAD-x86.exe

DaveTBlake:
Some discussion on Slack if "Set for this artist", "Set for all artists shown", and "Set default information provider" are acceptale on a menu. That latter is long enough to need scrolling on Estuary, and translations will probably be longer.

Meanwhile getting some test feedback on the forum.

DaveTBlake:
New test build available as **ec166d8**, for win32 in particular  http://mirrors.kodi.tv/test-builds/windows/win32/KodiSetup-20170809-a130a8e139-InfoSettings-x86.exe

scott967:
Created about a dozen test cases, also did some free-play testing.  All tests passed (DB state consistent with info provider changes made via the UI).

DaveTBlake:
This change has received thorough testing, so I would like to merge it. Any team member care to review it?

xbmc/filesystem/MusicDatabaseDirectory.cpp
```diff
@@ -121,7 +121,8 @@ void CMusicDatabaseDirectory::ClearDirectoryCache(const std::string& strDirector
 
 bool CMusicDatabaseDirectory::IsAllItem(const std::string& strDirectory)
 {
-  if (StringUtils::EndsWith(strDirectory, "/-1/"))
+  //Last query parameter, ignoring any appended options is -1
+  if (StringUtils::EndsWith(strDirectory.substr(0, strDirectory.find("?")), "/-1/"))
```
Paxxi: this will blow up if there's no ? in the string

xbmc/music/MusicDatabase.h
```diff
@@ -429,8 +425,9 @@ class CMusicDatabase : public CDatabase
   /////////////////////////////////////////////////
   // Scraper
   /////////////////////////////////////////////////
-  bool SetScraperForPath(const std::string& strPath, const ADDON::ScraperPtr& info);
-  bool GetScraperForPath(const std::string& strPath, ADDON::ScraperPtr& info, const ADDON::TYPE &type);
+  bool SetScraper(int id, const CONTENT_TYPE &content, const ADDON::ScraperPtr& scraper);
```
Paxxi: should take a `ADDON::ScraperPtr scraper` avoid references to std::shared_ptr

xbmc/music/windows/GUIWindowMusicNav.cpp
```diff
@@ -253,6 +254,138 @@ std::string CGUIWindowMusicNav::GetQuickpathName(const std::string& strPath) con
   }
 }
 
+bool CGUIWindowMusicNav::ManageInfoProvider(const CFileItemPtr &item)
```
Paxxi: parameter should be either `const CFileItem& item` or `CFileItemPtr item` taking a const ref to std::shared_ptr should be avoided as it's easy to misunderstand the different reference semantics.

Paxxi (changes requested):
some nitpicks and the find("?") needs a fix

xbmc/filesystem/MusicDatabaseDirectory.cpp
DaveTBlake: Indeed! Using a CURL method to strip any options is probably better

xbmc/music/MusicDatabase.h
DaveTBlake: Changed SetScraper and SetScraperAll, however left GetScraper using `ADDON::ScraperPtr& scraper` as otherwise it doesn't work. I am reusing historic code so perhaps that needs to change in some other way?

DaveTBlake:
@Paxxi have fixed your comments, thanks for those. Could you approve, please check you are happy with `CMusicDatabase::GetScraper`

Paxxi (approved):
looks good to me

PR #12598, opened 2017-07-31
IPlayer: rework player times

FernetMenta:
This cleans up the current mess regarding player times and generalises timeshift. Those times are available now:

1) startTime
For a typical video this will be zero. For live TV, this is a reference time in units of time_t (UTC) from which time elapsed starts. Ideally this would be the start of the tv show but can be any other time as well.

2) current time of playback
advanced time in ms since startTime

3) maxTime
maximun time in ms player can skip forward. for a typical video this will be totalLength. For live tv without timeshift this is zero, and fot live tv with timeshift this will be buffer ahead

3) minTime
zero for typical video. with timeshift this is the time in ms player can go back. this can be before startTime

### DIFF
--- a/xbmc/ApplicationPlayer.cpp
+++ b/xbmc/ApplicationPlayer.cpp
@@ -312,7 +312,7 @@ void CApplicationPlayer::SeekTimeRelative(int64_t iTime)
     // use relative seeking if implemented by player
     if (!player->SeekTimeRelative(iTime))
     {
-      int64_t abstime = player->GetTime() + iTime;
+      int64_t abstime = GetTime() + iTime;
       player->SeekTime(abstime);
     }
   }
@@ -331,7 +331,46 @@ int64_t CApplicationPlayer::GetTime() const
 {
   std::shared_ptr<IPlayer> player = GetInternal();
   if (player)
-    return player->GetTime();
+    return CDataCacheCore::GetInstance().GetPlayTime();
+  else
+    return 0;
+}
+
+int64_t CApplicationPlayer::GetMinTime() const
+{
+  std::shared_ptr<IPlayer> player = GetInternal();
+  if (player)
+    return CDataCacheCore::GetInstance().GetMinTime();
+  else
+    return 0;
+}
+
+int64_t CApplicationPlayer::GetMaxTime() const
+{
+  std::shared_ptr<IPlayer> player = GetInternal();
+  if (player)
+    return CDataCacheCore::GetInstance().GetMaxTime();
+  else
+    return 0;
+}
+
+time_t CApplicationPlayer::GetStartTime() const
+{
+  std::shared_ptr<IPlayer> player = GetInternal();
+  if (player)
+    return CDataCacheCore::GetInstance().GetStartTime();
+  else
+    return 0;
+}
+
+int64_t CApplicationPlayer::GetTotalTime() const
+{
+  std::shared_ptr<IPlayer> player = GetInternal();
+  if (player)
+  {
+    int64_t total = CDataCacheCore::GetInstance().GetMaxTime() - CDataCacheCore::GetInstance().GetMinTime();
+    return total;
+  }
   else
     return 0;
 }
@@ -446,20 +485,17 @@ std::string CApplicationPlayer::GetRadioText(unsigned int line)
     return "";
 }
 
-int64_t CApplicationPlayer::GetTotalTime() const
-{
-  std::shared_ptr<IPlayer> player = GetInternal();
-  if (player)
-    return player->GetTotalTime();
-  else
-    return 0;
-}
-
 float CApplicationPlayer::GetPercentage() const
 {
   std::shared_ptr<IPlayer> player = GetInternal();
   if (player)
-    return player->GetPercentage();
+  {
+    int64_t iTotalTime = GetTotalTime();
+
+    if (!iTotalTime)
+      return 0.0f;
+    return GetTime() * 100 / static_cast<float>(iTotalTime);
+  }
   else
     return 0.0;
 }
--- a/xbmc/addons/PVRClient.h
+++ b/xbmc/addons/PVRClient.h
@@ -786,6 +786,11 @@ namespace PVR
     bool IsRealTimeStream() const;
 
     /*!
+     * @brief Get Stream times (will be moved to inputstream)
+     */
+    bool GetStreamTimes(PVR_STREAM_TIMES *times);
+
+    /*!
      * @brief reads the client's properties
      */
     bool GetAddonProperties(void);
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
@@ -630,6 +630,11 @@ extern "C"
   void OnPowerSavingDeactivated();
 
   /*!
+   * Get stream times. Intermediate, will be moved to inputstream
+   */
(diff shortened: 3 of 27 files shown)


FernetMenta:
@garbear could you please adapt RP? playing times are not polled anymore but have to be updated by players. Maybe you need something like 48d3755

xbmc/cores/DataCacheCore.h
```diff
@@ -117,4 +122,12 @@ class CDataCacheCore
     float m_tempo;
     float m_speed;
   } m_stateInfo;
+
+  struct STimeInfo
+  {
+    time_t m_startTime;
+    int64_t m_time;
```
garbear: Why are these allowed to be negative? I noticed that a lot of the checks for negative have been dropped. Is this intended to allow negative times?

garbear:
The PR description does a good job explaining the different parameters. Can you add these as comments in the code so they're available for future devs?

xbmc/cores/DataCacheCore.h
FernetMenta: yes, this is by intent. if for example the starttime of a stream can't be determined because epg is missing but there is a buffer that allows to rewind to a position before the stream was started.

FernetMenta:
> Can you add these as comments in the code

@garbear what part of the code do you have in mind?

garbear:
@FernetMenta https://github.com/FernetMenta/xbmc/pull/428

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
```diff
@@ -630,6 +630,11 @@ extern "C"
   void OnPowerSavingDeactivated();
 
   /*!
+   * Get stream times. Intermediate, will be moved to inputstream
+   */
+  bool GetStreamTimes(PVR_STREAM_TIMES &times);
```
ksooo: Two points here:
1) May I suggest to change the return value to PVR_ERROR. Despite from being consistent with existing API design, this gives you the chance to distinguish between some error occured due to some reason versus "not implemented". The latter could help to implement migration: first, call GetStreamTimes and if it returns NOT_IMPLEMENTED, call the deprectated counterpart functions. As an alternative, you could introduce an addon capability. This has the benefit that there is no need for the "call the new function, check for not implemented, call old functions in this case" cycle whenever the times are needed. Could make sense if the times are requested very often.

2) As this is a C-API you must not use C++ language constructs (here: pass by reference). I know we do this at other places and Alwin is fixing this for the reworked APIs, but for new functions we should do it right in the first place. Use a pointer to implement the out param, please (PVR_STREAM_TIMES *times).

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h
```diff
@@ -572,6 +572,17 @@ extern "C" {
...
+  /*!
+   * @brief times of playing stream
+   */
+  typedef struct PVR_STREAM_TIMES
+  {
+    time_t startTime;
+    int64_t ptsStart;
+    int64_t ptsBegin;
+    int64_t ptsEnd;
+  } ATTRIBUTE_PACKED PVR_STREAM_TIMES;
```
ksooo: Please add documentation for the struct fields.

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h
```diff
@@ -679,6 +690,7 @@ extern "C" {
     void (__cdecl* OnSystemWake)(void);
     void (__cdecl* OnPowerSavingActivated)(void);
     void (__cdecl* OnPowerSavingDeactivated)(void);
+    bool (__cdecl* GetStreamTimes)(PVR_STREAM_TIMES&);
```
ksooo: `PVR_STREAM_TIMES*`, please. see my other comment.

ksooo (commented):
Only looked at the PVR Addon API changes and have some change requests / suggestions. 

And please don't forget to bump the PVR API version number. This is currently missing in this PR.

xbmc/ApplicationPlayer.cpp
```diff
@@ -446,20 +485,17 @@ std::string CApplicationPlayer::GetRadioText(unsigned int line)
...
 {
   std::shared_ptr<IPlayer> player = GetInternal();
   if (player)
-    return player->GetPercentage();
+  {
+    int64_t iTotalTime = GetTotalTime();
+
+    if (!iTotalTime)
+      return 0.0f;
+    return GetTime() * 100 / (float)iTotalTime;
```
ksooo: c++ style cast?

xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamPVRManager.cpp
```diff
@@ -215,6 +215,24 @@ int CDVDInputStreamPVRManager::GetTime()
...
 }
 
+bool CDVDInputStreamPVRManager::GetTimes(Times &times)
+{
+  if (m_isRecording)
+    return false;
+
+  PVR_STREAM_TIMES streamTimes;
+  bool ret = CServiceBroker::GetPVRManager().Clients()->GetStreamTimes(streamTimes);
+  if (ret)
```
ksooo: This would be the place for implementing migration - old functions vs. new getstreamtimes functions, right?

FernetMenta: nope, the old functions are complete separate and queried via interface IDisplayTime. The new ones via ITimes. ITimes has preference and IDisplayTime is fallback until it gets removed.

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
FernetMenta: ok, will do

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h
FernetMenta: sure

xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamPVRManager.cpp
ksooo: ok, thanks.

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
ksooo: thanks

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_types.h
ksooo: thanks :-)

xbmc/addons/PVRClient.h
```diff
@@ -786,6 +786,11 @@ namespace PVR
     bool IsRealTimeStream() const;
 
     /*!
+     * @brief Get Stream times (will be moved to inoutstream)
```
ksooo: inoutstream? i guess you mean inputstream

FernetMenta: updated

PR #12599, opened 2017-07-31
[videoplayer] Fix CDVDFactoryInputStream::CreateInputStream fileitem …

ksooo:
… dynpath fallout.

Stumbled over this while working on https://github.com/kodi-pvr/pvr.demo/pull/50. Fixes tv/radio channel playback using `GetLiveStreamURL`

This has been runtime-tested on macOS, latest Kodi master.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDInputStreams/DVDFactoryInputStream.cpp
+++ b/xbmc/cores/VideoPlayer/DVDInputStreams/DVDFactoryInputStream.cpp
@@ -48,7 +48,7 @@ CDVDInputStream* CDVDFactoryInputStream::CreateInputStream(IVideoPlayer* pPlayer
 {
   using namespace ADDON;
 
-  std::string file = fileitem.GetPath();
+  std::string file = fileitem.GetDynPath();
   if (scanforextaudio)
   {
     // find any available external audio tracks


garbear:
Does RetroPlayer need similar adaptations? I read through the dynpath PR but couldn't identify the exact use cases for DynPath from the code.

FernetMenta (approved):

FernetMenta:
@garbear DynPath separates the playback url from id any type of a fileitem. It does not allow a pvr channel to become a picture item but it allows to change the playback path which is currently tied to the id. Consider a video stored in folder A. All settings are stored with the path. You can't move it to folder B with changing its id.

ksooo:
Detected some more fallout, for instance non-working channel up/down, up/down key presses while playing PVR channel via addon provided dynpath. Working on it.

garbear:
@FernetMenta Just to be sure I understand, DynPath is for e.g. a multi-file video whose path can change? This doesn't concern RetroPlayer currently, but in the future, RetroPlayer might support changing a PlayStation disk. Is that a sample use case of DynPath?

FernetMenta:
> DynPath is for e.g. a multi-file video whose path can change?

hmm, not sure. currently the id of an item is its path. this does not even allow to move the item to a different location.

da-anda:
@garbear I understood it this way: The path of a cFileItem is it's identity. In case of library items, it the VFS path pointing to something like videodb://movie/1 while the DynPath could/would be the path within the local filesystem. Same for python file items, the path would point to addon://foo/bar?baz (identifier) while DynPath would be the resolved stream URL. Is that about right @FernetMenta ?

> It does not allow a pvr channel to become a picture item but it allows to change the playback path which is currently tied to the id

But in theory anything getting in touch with the cFileItem could change the playback path of a PVR channel to a picture URL, couldn't it?

FernetMenta:
> But in theory anything getting in touch with the cFileItem could change the playback path of a PVR channel to a picture URL by changing the DynPath, couldn't it?

this will most likely reslult in a not playable item. I see no issue with this

PR #12600, opened 2017-08-01
[VideoPlayer] Never skip the same commercial twice in a row

b-pass:
## Description

Kodi should never skip the same commercial twice in a row.  Both when seeking backwards into a commercial and also when seeking backwards right before the commercial (or to the exact frame where the commercial starts).

To do this, we keep track of the last commercial skipped and we refuse to skip it again.  This allows us to seek back into the middle of the commercial and also to seek back slightly before the commercial.

## Motivation and Context

Commercial skipping is mostly only useful when watching a program linearly.  Once the user starts skipping around, it is usually because there is a problem with the commercial boundaries.  If the user skips back to the start of a commercial break, Kodi needs to *not* try to skip the commercial again.  This causes the user to fight with Kodi, skipping back to the beginning of the commercial and having kodi skip to the end. 

This code already tries to avoid skipping the same commercial twice in a row.  It does this by only skipping if we are playing forwards from a point before the commercial to a point inside the commercial.  That is good, but what if we skip back to the beginning of the commercial (such as with `SeekScene()` and the Next/Prev Scene buttons)?  In that case, Kodi will skip the commercial again, but this is almost surely not what we want.

The old DVDPlayer code in Jarvis and early Krypton builds accomplished this by keeping track of the last skipped commercial break  and basically refusing to skip breaks earlier in the show entirely (it keeps track in m_EdlAutoSkipMarkers, [see here](https://github.com/xbmc/xbmc/blob/Jarvis/xbmc/cores/dvdplayer/DVDPlayer.cpp#L2236)).  This pull is trying to do a similar thing to get closer to that old behavior.

## How Has This Been Tested?

Tested this with a commercial flagged recording in a variety of skip scenarios (normal watching, skipping backwards into the break, and skipping back to the start of the break).

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/Edl.cpp
+++ b/xbmc/cores/VideoPlayer/Edl.cpp
@@ -47,7 +47,7 @@ void CEdl::Clear()
   m_vecCuts.clear();
   m_vecSceneMarkers.clear();
   m_iTotalCutTime = 0;
-  m_lastCheckASSTime = 0;
+  m_lastCutTime = 0;
 }
 
 bool CEdl::ReadEditDecisionLists(const std::string& strMovie, const float fFrameRate, const int iHeight)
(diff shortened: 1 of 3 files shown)


FernetMenta:
thanks

PR #12602, opened 2017-08-01
[pvr] fix database migration and drop bWasPlayingOnQuit from create table

xhaggi:
1. Remove column `bWasPlayingOnQuit` from create table after #12548 (leftover)
2. Remove add column `bWasPlayingOnQuit` migration for version 30
3. Remove drop column `bWasPlayingOnQuit` migration for version 31 because SQLlite does not support dropping of columns.

We leave the column in because in SQLlite you need to rename the old table, create a new table without the column, copy the data over to the new table and drop the old table. The syntax to rename a table in SQLlite and MySQL also differ. Really annoying.

Take a look at this paste and you know what I mean.
https://pastebin.com/tbHXx4M3

## How Has This Been Tested?

Runtime tested on MacOS

## Types of change


ksooo (approved):
Looks good to me.

PR #12604, opened 2017-08-01
update requirements

newEntera:
## Types of change

## Checklist:

### DIFF
--- a/docs/README.raspberrypi
+++ b/docs/README.raspberrypi
@@ -23,7 +23,8 @@ The following commands build for newer Raspberry Pi 2 generation. In order to
 build for the first Raspberry Pi, the commands have to be adapted to use
 `--with-platform=raspberry-pi` instead of `--with-platform=raspberry-pi2`.
 
-    $ sudo apt-get install git autoconf curl g++ zlib1g-dev libcurl4-openssl-dev gawk gperf libtool autopoint swig default-jre
+    $ sudo apt-get install git autoconf curl g++ zlib1g-dev \
+    libcurl4-openssl-dev gawk gperf libtool autopoint swig default-jre zip unzip cmake build-essential
 
     $ git clone https://github.com/raspberrypi/tools
     $ sudo cp -r tools/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian-x64 /opt


Rechi:
@newEntera please explain why you think zip, unzip, cmake and build-essential are required

newEntera:
@Rechi That's too complicated for me to answer. Simply make of raspberry tools asks about it and I install it and I update the requirements here. Is that an valid answer ? Do I pass the exam :) ?

bh:
What a stupid answer for an legitimate question.

newEntera:
Thank you for insulting me.
Legitimate question is why do you think it is a legitimate ?
 Is it legitimate to have README instructions that more steps that are not described and I put them to make it easier for me when I come back here after 5 months If I have to be selfish. But I do it for the community to make it easier to the rest of people like me.

MartijnKaijser:
@newEntera Sorry but you were asked a question by a team member and your response was to thumbs down his question and make a smart ass remark. Goodbye

Jalle19:
How could `cmake` not be a requirement? Don't we use cmake for everything?

Rechi:
@Jalle19 the change is for the section which describes building with depends.
Cmake gets compiled from source in depends and the system cmake would never be used.

Jalle19:
@Rechi okay
@newEntera what about the other dependencies, at what point did you actually need them?

newEntera:
@Rechi 
Following xbmc/docs/README.raspberrypi cmake does not get compiled from source. Or please show me what I have missed if so, or I'm wrong and it's not included in $PATH ?
@Jalle19 following strictly xbmc/docs/README.raspberrypi on 16.04 is zip,unzip is what I have missed to succeed. It was at the point of raspberry tools. Sure out of kodi's scope but this is what you need if you want to make a build environment to compile for raspberry pi. Maybe the build environment is created other way for build for Ubuntu and Rechi is mistaken by that cmake is build from source?
While build-essential is included in 16.04 Desktop, it is not included in 16.04 minimal or 16.04 server which I used.

Rechi:
https://github.com/xbmc/xbmc/blob/4bd402f/docs/README.raspberrypi#L40-L50 builds depends (including cmake)

PR #12607, opened 2017-08-01
[PVR] Fix and refactor PVR internal 'playing' states after removal of…

ksooo:
… VideoPlayer's 'other stream hack'."

Hopefully the last fallout from #12550. Fixes Live TV OSD for playing "dynpath" channels and other glitches.

This has been runtime tested on macOS, latest Kodi master.

@FernetMenta as discussed on Slack...

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -3501,6 +3501,8 @@ void CApplication::OnPlayBackEnded()
   CDarwinUtils::EnableOSScreenSaver(true);
 #endif
 
+  CServiceBroker::GetPVRManager().OnPlaybackEnded(m_itemCurrentFile);
+
   CVariant data(CVariant::VariantTypeObject);
   data["end"] = true;
   CAnnouncementManager::GetInstance().Announce(Player, "xbmc", "OnStop", m_itemCurrentFile, data);
@@ -3528,6 +3530,8 @@ void CApplication::OnPlayBackStarted()
   if (m_pPlayer->IsPlayingVideo())
     CDarwinUtils::EnableOSScreenSaver(false);
 #endif
+
+  CServiceBroker::GetPVRManager().OnPlaybackStarted(m_itemCurrentFile);
 
   CGUIMessage msg(GUI_MSG_PLAYBACK_STARTED, 0, 0);
   g_windowManager.SendThreadMessage(msg);
@@ -3567,6 +3571,8 @@ void CApplication::OnPlayBackStopped()
 #elif defined(TARGET_DARWIN_IOS)
   CDarwinUtils::EnableOSScreenSaver(true);
 #endif
+
+  CServiceBroker::GetPVRManager().OnPlaybackStopped(m_itemCurrentFile);
 
   CVariant data(CVariant::VariantTypeObject);
   data["end"] = false;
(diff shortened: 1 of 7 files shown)


xbmc/addons/PVRClient.h
```diff
@@ -723,9 +723,41 @@ namespace PVR
     bool IsPlayingEncryptedChannel(void) const;
     bool IsPlayingRecording(void) const;
     bool IsPlaying(void) const;
-    CPVRRecordingPtr GetPlayingRecording(void) const;
+
+    /*!
+     * @brief Set the channel that is currently playing.
+     * @param channel The channel that is currently playing.
+     */
+    void SetPlayingChannel(const CPVRChannelPtr channel);
```
FernetMenta: future PVR addons can play multiple channels at the same time. if an addon implementes inputstream, it knows the playing channels anyway. if it doesn't, I see not need for it to know.

ksooo: Sure, but for now, to make add-ons without an item demuxer working again, we need this. I'm aware that this code might go away soon, yes.

ksooo: My goal was to come up with a quick and (for what we have currently) not dirty fix.

ksooo: Without the other pieces for new pvr streaming being in place I'm unable to do this "right" nowadays.

xbmc/Application.cpp
```diff
@@ -3501,6 +3501,8 @@ void CApplication::OnPlayBackEnded()
   CDarwinUtils::EnableOSScreenSaver(true);
 #endif
 
+  CServiceBroker::GetPVRManager().OnPlaybackEnded(m_itemCurrentFile);
```
FernetMenta: the other direction that PVR updates currentFileItem in application should not be required anymore: https://github.com/xbmc/xbmc/search?utf8=%E2%9C%93&q=setcurrentfileitem&type=

xbmc/addons/PVRClient.h
FernetMenta: ok

xbmc/Application.cpp
ksooo: Absolutely, but not part/goal of this PR. This could be one of the next cleanup steps.

FernetMenta (approved):

xbmc/Application.cpp
```diff
@@ -3568,6 +3572,8 @@ void CApplication::OnPlayBackStopped()
   CDarwinUtils::EnableOSScreenSaver(true);
 #endif
 
+  CServiceBroker::GetPVRManager().OnPlaybackStopped(m_itemCurrentFile);
+
```
da-anda: wouldn't it be cleaner to register PVRmanger and all the other components above to the respective announcement from CAnnouncementManager below? (same with onStart, etc). Would get rid of all the ifdeffery. Just a general question for some future refactoring that came to my mind when looking at this PRs changes.

da-anda: some other signal from cApplicationMessenger then?

FernetMenta: ApplicationMessenger is for messages in direction to the main thread.
This code is good as is. It is supposed to be snychronous, not to be hooked in to some undefined messaging anarchy

PR #12608, opened 2017-08-01
Controller models and other improvements

garbear:
This adds the ability to define multiple controller models in a single controller add-on.

## Description
In the upcoming player manager, it is possible to choose which device is plugged into each port. This expands on that idea by allowing the user to choose specific models.

A new parameter, "models", is added to the controller layout:

```xml
<layout label="30000" image="layout_proline.png" mask="mask_proline.png" models="models.xml">
```

This links to a models XML that can override layout parameters:

```xml
<models>
  <model name="proline">
    <layout label="30009"/>
  </model>
  <model name="gamepad">
    <layout label="30010" icon="icon_gamepad.png" image="layout_gamepad.png"/>
  </model>
</models>
```

In this example, the default Atari 7800 model is the Atari Proline Joystick:

layout_proline

The models.xml file above overrides the name shown in the Controller Dialog, "Atari 7800", with the model name "Atari Proline Joystick". It then adds a second model, with label "Atari 7800 Gamepad" and an alternative image:

layout_gamepad

## Motivation and Context
Some controllers identify themselves to the game console, so that the game console is aware of which controller is plugged into the port. This gives the user control over that.

An alternative was to add separate controller add-ons for each model, but then the Controller Dialog would show multiple controllers with the same button layout. With this solution, the Controller Dialog degrades to using a single layout, while in the upcoming Player Manager the multiple models are exposed.

## How Has This Been Tested?
Multiple models aren't available in the GUI yet, but I've set a breakpoint and verified that the models are loaded correctly.

Because the models degrade to the single layout in the controller dialog, this change is fully backwards compatible with Krypton.

## Types of change

### DIFF
--- a/xbmc/games/addons/GameClient.cpp
+++ b/xbmc/games/addons/GameClient.cpp
@@ -35,7 +35,6 @@
 #include "games/addons/playback/GameClientRealtimePlayback.h"
 #include "games/addons/playback/GameClientReversiblePlayback.h"
 #include "games/controllers/Controller.h"
-#include "games/controllers/ControllerLayout.h"
 #include "games/ports/PortManager.h"
 #include "games/GameServices.h"
 #include "guilib/GUIWindowManager.h"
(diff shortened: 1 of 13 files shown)


garbear:
jenkins build this please

PR #12612, opened 2017-08-02
Allow video scrapers/NFO files/library import to add any artwork type…

rmrector:
… for items

## Description
Scrapers (along with friends NFO files and library importing from single file XML export) can set any type of artwork (clearlogo, landscape, etc) for items in the video library.

This is such a simple loop change that doesn't disturb the beast that is scrapers, I so very much wish I had found it a couple of years ago when I first went looking for it.

## Motivation and Context
The video library has long supported freely named artwork for items in most other places, now scrapers can set this artwork from the beginning.

## How Has This Been Tested?

By hand, I modified the new demo Python scrapers and a classic scraper to set some extra artwork, and modified a couple of NFO files to match, then ensuring the results are what I expect. I also refreshed a few items with a regular scraper and ensured they still set their expected artwork.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/video/VideoInfoScanner.cpp
+++ b/xbmc/video/VideoInfoScanner.cpp
@@ -1422,25 +1422,6 @@ namespace VIDEO
       }
     }
 
-    // find online art
-    for (std::vector<std::string>::const_iterator i = artTypes.begin(); i != artTypes.end(); ++i)
-    {
-      if (art.find(*i) == art.end())
-      {
-        std::string image = GetImage(pItem, false, bApplyToDir, *i);
-        if (!image.empty())
-          art.insert(std::make_pair(*i, image));
-      }
-    }
-
-    // use the first piece of online art as the first art type if no thumb type is available yet
-    if (art.empty() && lookForThumb)
-    {
-      std::string image = GetImage(pItem, false, bApplyToDir, "thumb");
-      if (!image.empty())
-        art.insert(std::make_pair(artTypes.front(), image));
-    }
-
     // get & save fanart image (treated separately due to it being stored in m_fanart)
     bool isEpisode = (content == CONTENT_TVSHOWS && !pItem->m_bIsFolder);
     if (!isEpisode && art.find("fanart") == art.end())
@@ -1448,6 +1429,22 @@ namespace VIDEO
       std::string fanart = GetFanart(pItem, useLocal);
       if (!fanart.empty())
         art.insert(std::make_pair("fanart", fanart));
+    }
+
+    // add online art
+    for (const auto& i : pItem->GetVideoInfoTag()->m_strPictureURL.m_url)
+    {
+      if (i.m_type != CScraperUrl::URLTYPES::URL_TYPE_GENERAL)
+        continue;
+      std::string aspect = i.m_aspect;
+      if (aspect.empty())
+        // temporary support for XML music video scrapers that share music album scraper bits
+        aspect = content == CONTENT_MUSICVIDEOS ? "poster" : "thumb";
+      if (art.find(aspect) != art.end())
+        continue;
+      std::string image = GetImage(i, pItem->GetPath());
+      if (!image.empty())
+        art.insert(std::make_pair(aspect, image));
     }
 
     for (CGUIListItem::ArtMap::const_iterator i = art.begin(); i != art.end(); ++i)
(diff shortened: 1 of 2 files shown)


rmrector:
The build failure for Win-64 looks like it was caused by something else.

hudokkow:
Sorry, this is not my territory.
jenkins failure indeed seems unrelated. How the hell do I remove my name from review request list?

rmrector:
Anyone have any thoughts on this? This is kind of important, a lot of folks like extended artwork and this has been a big hole in artwork support for awhile now (relying on an add-on seems unnecessary, scrapers can add some artwork so they should be able to add all of it).

One hole is left, Kodi doesn't import extended artwork from image files next to media. I've got Python code that identifies these images just fine, but for some reason I just cannot translate it to Kodi core (C++ noob, gah!).

zag2me:
@DaveTBlake was looking at imports and exports recently

DaveTBlake:
For music, scraping of video is different, but I am interested in how we can handle art better in core rather than leave it to addons. There is also https://github.com/xbmc/xbmc/pull/11755, which could be of interest.

rmrector:
@phate89 You've been hard at work on bigger parts of the scrapers, what do you think of this? This code sits up a level higher, as a filter on the data from the scraper before it makes it into the database, and is still useful to your open PR #12661.

@MartijnKaijser With all your work on Artwork Downloader, what do you think of scrapers being able to add more artwork?

xbmc/video/VideoInfoScanner.cpp
```diff
@@ -1397,6 +1378,18 @@ namespace VIDEO
         art.insert(std::make_pair("fanart", fanart));
     }
 
+    // add online art
+    std::vector<CScraperUrl::SUrlEntry>& scraperArt = pItem->GetVideoInfoTag()->m_strPictureURL.m_url;
+    for (auto i = scraperArt.begin(); i != scraperArt.end(); ++i)
```
phate89: You could use a range-based for like this `for (const auto& a : pItem->GetVideoInfoTag()->m_strPictureURL.m_url)`

phate89:
I'm not an expert of artwork code but it seems fine and also simpler. Why did you move the loop after fanart gathering?

rmrector:
I moved the loop because it can also provide a "fanart" image, but the original fanart section still offers an extra feature, the rarely-used fanart colors for skins, so they should take priority.

I just added a bit that covers what the original `art.empty()` check did for XML scrapers: it considers artwork with an empty aspect to be "thumb", which is a reasonable assumption and matches scraper behavior for episodes, or "poster" for music videos, which is a quick hack to support the XML music video scrapers that share scraper bits with music album scrapers.

zag2me:
Does this work for music scrapers also?

phate89:
@rmrector I really don't know anything but I always tought that fanart was only in `m_fanart`..
We explicitely drop it from `arTypes` here https://github.com/rmrector/xbmc/blob/f489c22bd92aa374c17b0de62ccaef946c57236f/xbmc/video/VideoInfoScanner.cpp#L1338
But I think it make sense to be able to provide all artwork from one place.. maybe @xhaggi know better
@zag2me these are only video changes. No idea if the same could be done in music part

rmrector:
@zag2me The music library doesn't support the same wide selection of artwork as the video library in the first place, so it can't be extended this simply and is outside the scope of this PR.

@phate89 This new code doesn't use artTypes when processing scraper artwork (artTypes is what held back the original code), but accepts the first image of each art type/"aspect" the scraper provides, so if the scraper sets an image aspect to "fanart" in `addAvailableThumb` then it will show up here. I agree that all artwork should work the same, so I didn't want to add an exception to ignore "fanart" here, and I don't want to disturb XML scrapers, so I kept the existing m_fanart as well. If it weren't for that rarely-used feature of fanart, the colors (I like the idea, but I think they should be attached to the TV show and occasionally season rather than each fanart), I'd say Python scrapers shouldn't even have a `setAvailableFanart`. The only web service that has that info is TheTVDB, and their new v2 API (v1 will disappear soon) doesn't seem to provide the color anymore (nor their beta website), so maybe it should even go away anyway?

I've also push another change to handle season artwork, which is in another section of code. One thing about season artwork is that it doesn't work from Python scrapers in master (with notspiff's initial Python scraper), but does look to be addressed in your PR. I haven't given it a test yet, but I intend to soon; the scraper process has some goofiness with artwork, and I think the Python scraper interface can smooth some of it out.

xbmc/video/VideoInfoScanner.cpp
```diff
@@ -1463,24 +1460,17 @@ namespace VIDEO
...
-    if (useLocal)
-      thumb = CVideoThumbLoader::GetLocalArt(*pItem, type, bApplyToDir);
-
-    if (thumb.empty())
+    std::string thumb = CScraperUrl::GetThumbURL(image);
+    if (!thumb.empty())
     {
-      thumb = CScraperUrl::GetThumbURL(pItem->GetVideoInfoTag()->m_strPictureURL.GetFirstThumb(type));
-      if (!thumb.empty())
+      if (thumb.find("http://") == std::string::npos &&
```
notspiff: please update for the new world of https everywhere. and use StringUtils::StartsWith

notspiff (changes requested):
fix the very minors and i will merge this. sorry for the long wait!

xbmc/video/VideoInfoScanner.cpp
rmrector: The find in the line below catches https:// and every other string that has a protocol, this line is unnecessary altogether. I can remove this line and combine the two nested ifs, but I don't see a "Find"-like method in the StringUtils. Is there one somewhere?

notspiff: StringUtils::Contains(..) but really using find for stuff anywhere in string is fine by me. It was the anchoring to front that was most important in my request and startswith conveys that a bit quicker to the reader of code.

rmrector:
@notspiff ping.

notspiff (approved):

notspiff:
jenkins build this please

notspiff:
@xhaggi i will do a follow up with the cosmetics, let's get this in now.

olympia:
Cool!
Would someone explain to me (who is unable to read the code) what to change / how to use this in the scrapers?

PR #12614, opened 2017-08-03
[PVR] Always check whether a timer has an associated channel. It might be null.

ksooo:
The first commit fixes a crash described here: https://github.com/janbar/pvr.mythtv/issues/77

The second and third commit are just some code cosmetics.

@janbar fyi


ksooo:
@Jalle19 for review?

Jalle19 (approved):

ksooo:
Thanks. Superfluous space removed.

janbar:
+1

PR #12615, opened 2017-08-03
[PVR] Guide window: Fix initialization after skin reload (trac#17546).

ksooo:
Fixes https://trac.kodi.tv/ticket/17546

This change has been runtime-tested on macOS, latest kodi master.

@Jalle19 for review?

### DIFF
--- a/xbmc/pvr/windows/GUIWindowPVRGuide.cpp
+++ b/xbmc/pvr/windows/GUIWindowPVRGuide.cpp
@@ -74,7 +74,7 @@ void CGUIWindowPVRGuideBase::Init()
     epgGridContainer->GoToNow();
   }
 
-  if (!m_refreshTimelineItemsThread)
+  if (epgGridContainer && !epgGridContainer->HasData())
   {
     CSingleLock lock(m_critSection);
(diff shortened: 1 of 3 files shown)


xbmc/pvr/windows/GUIWindowPVRGuide.cpp
```diff
@@ -74,7 +74,7 @@ void CGUIWindowPVRGuideBase::Init()
     epgGridContainer->GoToNow();
   }
 
-  if (!m_refreshTimelineItemsThread)
```
Jalle19: So this check is no longer needed? Can the whole member be removed?

ksooo: Yes, the check is replaced by something that not only kicks in during very first open of the window, but also after the window controls have been replaced when doing a skin reload. Much cleaner now - and fixing the bug. 

The member is used at other places in this file.

Jalle19: +1

xhaggi:
looks good, but was not expressly asked ;)

PR #12617, opened 2017-08-03
Assorted touch fixes

yol:
I discovered issues with the touch subsystem while doing Wayland implementation tests (pkerling/xbmc#43), some of which are fixed by this PR

## Description
* Inertial scroll
  * Math does not add up (70b930fa) resulting in abrupt stop
  * Scrolling is not actually "inertial", i.e. if user starts touch, scrolls around a bit, then holds the finger for one second and releases, inertial scrolling will still kick in despite there not being any movement - very annoying (1d1c69f)
* Gesture detection
  * When windowing does not provide size, detection is way too sensitive and often turns taps into pan gestures (843e2b41)
  * Multi-pointer taps miss the check for whether the pointer moved even when size is provided (3d83da4) and so also frequently get turned into gestures
* Gestures are not aborted on window switch, resulting in unusable settings windows (because list control does not release exclusive mouse mode) when two-finger swipe gesture is used to go back to the previous window (f51f7bbd)

Rest of the commits are cleanup and C++11.

## Motivation and Context
Improve touch experience

## How Has This Been Tested?
Run-tested on Linux/Wayland

Android testing would be much appreciated. I can do it if someone tells me how I can get the resulting apk from jenkins?

## Types of change

## Checklist:

### DIFF
--- a/xbmc/input/linux/LinuxInputDevices.h
+++ b/xbmc/input/linux/LinuxInputDevices.h
@@ -26,8 +26,7 @@
 #include "windowing/XBMC_events.h"
 #include "input/XBMC_keyboard.h"
 #include "threads/SingleLock.h"
-#include "input/touch/ITouchInputHandler.h"
-#include "input/touch/generic/IGenericTouchGestureDetector.h"
+#include "input/touch/generic/GenericTouchInputHandler.h"
 #include "threads/Thread.h"
 
 struct KeymapEntry
@@ -87,9 +86,9 @@ private:
   bool m_bUnplugged;
   std::deque<XBMC_Event> m_equeue;
   int m_mt_currentSlot;
-  int m_mt_x[TOUCH_MAX_POINTERS];
-  int m_mt_y[TOUCH_MAX_POINTERS];
-  TouchInput m_mt_event[TOUCH_MAX_POINTERS];
+  int m_mt_x[CGenericTouchInputHandler::MAX_POINTERS];
+  int m_mt_y[CGenericTouchInputHandler::MAX_POINTERS];
+  TouchInput m_mt_event[CGenericTouchInputHandler::MAX_POINTERS];
 };
 
 class CLinuxInputDevices
--- a/xbmc/input/touch/generic/GenericTouchActionHandler.cpp
+++ b/xbmc/input/touch/generic/GenericTouchActionHandler.cpp
@@ -161,19 +161,6 @@ int CGenericTouchActionHandler::QuerySupportedGestures(float x, float y)
   return result;
 }
 
-void CGenericTouchActionHandler::touch(uint8_t type, uint8_t button, uint16_t x, uint16_t y)
-{
-  XBMC_Event newEvent;
-  memset(&newEvent, 0, sizeof(newEvent));
-  
-  newEvent.type = type;
-  newEvent.button.button = button;
-  newEvent.button.x = x;
-  newEvent.button.y = y;
-  
-  CWinEvents::MessagePush(&newEvent);
-}
-
 void CGenericTouchActionHandler::sendEvent(int actionId, float x, float y, float x2 /* = 0.0f */, float y2 /* = 0.0f */, int pointers /* = 1 */)
 {
   XBMC_Event newEvent;
--- a/xbmc/input/touch/generic/GenericTouchActionHandler.cpp
+++ b/xbmc/input/touch/generic/GenericTouchActionHandler.cpp
@@ -91,14 +91,14 @@ bool CGenericTouchActionHandler::OnTouchGestureStart(float x, float y)
 
 bool CGenericTouchActionHandler::OnTouchGesturePan(float x, float y, float offsetX, float offsetY, float velocityX, float velocityY)
 {
-  sendEvent(ACTION_GESTURE_PAN, x, y, offsetX, offsetY);
+  sendEvent(ACTION_GESTURE_PAN, x, y, offsetX, offsetY, velocityX, velocityY);
 
   return true;
 }
 
 bool CGenericTouchActionHandler::OnTouchGestureEnd(float x, float y, float offsetX, float offsetY, float velocityX, float velocityY)
 {
-  sendEvent(ACTION_GESTURE_END, velocityX, velocityY, x, y);
+  sendEvent(ACTION_GESTURE_END, velocityX, velocityY, x, y, offsetX, offsetY);
 
   return true;
 }
@@ -108,7 +108,7 @@ void CGenericTouchActionHandler::OnTap(float x, float y, int32_t pointers /* = 1
   if (pointers <= 0 || pointers > 10)
     return;
 
-  sendEvent(ACTION_TOUCH_TAP, x, y, 0.0f, 0.0f, pointers);
+  sendEvent(ACTION_TOUCH_TAP, x, y, 0.0f, 0.0f, 0.0f, 0.0f, pointers);
 }
 
 void CGenericTouchActionHandler::OnLongPress(float x, float y, int32_t pointers /* = 1 */)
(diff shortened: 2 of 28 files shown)


MartijnKaijser:
http://jenkins.kodi.tv/job/BuildMulti-All/821/  builds will end up in test-builds

xbmc/input/touch/generic/GenericTouchActionHandler.cpp
```diff
@@ -179,8 +182,8 @@ void CGenericTouchActionHandler::focusControl(float x, float y)
 {
   XBMC_Event newEvent{XBMC_SETFOCUS};
 
-  newEvent.focus.x = (uint16_t)x;
-  newEvent.focus.y = (uint16_t)y;
+  newEvent.focus.x = static_cast<int> (std::round(x));
```
notspiff: This is not equivalent.

pkerling: Yes. I think that it should have been rounded in the first place.

xbmc/input/touch/generic/GenericTouchInputHandler.cpp
```diff
@@ -28,7 +28,13 @@
 #include "threads/SingleLock.h"
 #include "utils/log.h"
 
-#define TOUCH_HOLD_TIMEOUT  1000
+namespace
+{
+
```
Montellese: Please remove the additional blank lines.

Montellese (approved):
Looks good apart from the minor. I can't comment on the inertial scrolling, that's @Memphiz' baby.

Memphiz:
Tested on ios - seems to work fine still. I am fine with removing the decrease of deacceleration - feels good enough without imo. I added it in the first place to mimic what apple does as good as possible (the 0.9 decrease factor was determined by stomach feeling ...

Memphiz (approved):

xbmc/input/linux/LinuxInputDevices.h
```diff
@@ -87,9 +86,9 @@ class CLinuxInputDevice
   bool m_bUnplugged;
   std::deque<XBMC_Event> m_equeue;
   int m_mt_currentSlot;
-  int m_mt_x[TOUCH_MAX_POINTERS];
-  int m_mt_y[TOUCH_MAX_POINTERS];
-  TouchInput m_mt_event[TOUCH_MAX_POINTERS];
+  int m_mt_x[CGenericTouchInputHandler::MAX_POINTERS];
```
Paxxi: If we're modernizing it would make sense to make these `std::aray` but might be out of scope for this PR.

xbmc/input/touch/generic/GenericTouchActionHandler.cpp
```diff
@@ -161,43 +166,28 @@ int CGenericTouchActionHandler::QuerySupportedGestures(float x, float y)
...
 void CGenericTouchActionHandler::focusControl(float x, float y)
 {
-  XBMC_Event newEvent;
-  memset(&newEvent, 0, sizeof(newEvent));
+  XBMC_Event newEvent{XBMC_SETFOCUS};
 
-  newEvent.type = XBMC_SETFOCUS;
-  newEvent.focus.x = (uint16_t)x;
-  newEvent.focus.y = (uint16_t)y;
+  newEvent.focus.x = static_cast<int> (std::round(x));
```
Paxxi: why is this changed from uint16_t to int?

xbmc/input/touch/generic/GenericTouchInputHandler.cpp
```diff
@@ -361,44 +363,40 @@ void CGenericTouchInputHandler::triggerDetectors(TouchInput event, int32_t point
...
       break;
     }
 
     default:
       return;
   }
 
-  std::set<IGenericTouchGestureDetector*> finishedDetectors;
-  for (std::set<IGenericTouchGestureDetector*>::const_iterator detector = m_detectors.begin(); detector != m_detectors.end(); ++detector)
+  for (auto it = m_detectors.begin(); it != m_detectors.end(); )
```
Paxxi: might as well switch this to `std::remove_if` when changing it.

Paxxi (commented):
I trust @Montellese review on the functional changes, added a few nitpicks that are optional.

Can try and test the changes on Windows over the weekend.

xbmc/input/touch/generic/GenericTouchActionHandler.cpp
pkerling: XBMC_SetFocusEvent has `int`s, why cut it to 16 bit? https://github.com/xbmc/xbmc/blob/master/xbmc/windowing/XBMC_events.h#L106

xbmc/input/linux/LinuxInputDevices.h
pkerling: I have to stop somewhere or this never ends ;-)

xbmc/input/touch/generic/GenericTouchInputHandler.cpp
pkerling: `remove_if` does not work with `std::set` because it is ordered and `remove_if` needs to reorder elements. I see no strong reason why `m_detectors` would need to be a `std::set` though. Not like you can try to push two identical `unique_ptr`s anyway. Should we make it a `std::vector`?

xbmc/input/touch/generic/GenericTouchInputHandler.cpp
pkerling: done

MartijnKaijser:
It feels like it improves touch quite a lot. Hitting buttons is sometimes still hit and miss but haven't tested extensively yet. Selecting in listings is still quite difficult to get the correct item selected but that could be a skinning engine issue. The highlighted item while scrolling also seems a bit flaky.

fritsch:
One thing at a time, please. If this one brings us forward, what it does,
squash together and we are good to go.

With the signed integer take care no one uses the center in the middle :-)

Am 05.08.2017 8:50 vorm. schrieb "Martijn Kaijser" <notifications@github.com
>:

It feels like it improves touch quite a lot. Hitting buttons is sometimes
still hit and miss but haven't tested extensively yet. Selecting in
listings is still quite difficult to get the correct item selected but that
could be a skinning engine issue. The highlighted item while scrolling also
seems a bit flaky.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/xbmc/xbmc/pull/12617#issuecomment-320419887>, or mute
the thread

MartijnKaijser:
@fritsch i was pointing out that one of the issues is still kinda there. So that is still that "one thing at a time".

pkerling:
I've identified potential problems with the gesture abort on window switch:

* I handled the gesture events in `CGUIWindowManager::OnApplicationMessage` - but the inertial scrolling handler might modify the events afterwards (drop some), so I think handling this in `CGUIWindowManager::OnAction` instead would be better. I'll add a fixup.
* What about other events that cause the focus and exclusive mouse control of a control to be broken no matter what it says - like dialogs (I get unresponsive dialogs from time to time, but haven't been able to pinpoint the problem)? Should we send gesture abort and inhibit likewise when `RegisterDialog` is called? Always or only under some circumstances, like only for modal dialogs or only when the topmost dialog changes?
  * Speaking of order: Some functions in window manager sort the dialogs based on their render order, like `RenderPass`. But most do not. `OnAction` for example doesn't do it. Won't that cause a mismatch between the topmost dialog on screen and the one that receives input?

MartijnKaijser:
@Memphiz could you try again on iOS if all still works? Android has improved a lot with this. @Paxxi if you could try on Windows?

Memphiz:
Is there a test build alteady?

MartijnKaijser:
Building now

pkerling:
squashed

Memphiz:
Behaves fine on iOS imo.

pkerling:
forgot to reorder the ARM32 fixed to before the std::round usage, so once again...

PR #12618, opened 2017-08-03
Player: drop unused option

FernetMenta:


ksooo:
Pleaes do not merge without my approval. I have no idea what the consequences of removing the call to CApplication::SetCurrentFileItem has. I want to understand first why the call is there and more important, in case something break I want a fix first.

FernetMenta:
sure. as far as I understand this was required for channel switching because this happened without knowledge of application. that changed and as a result this should ne be needed anymore.

ksooo:
Okay, can you please remove the commit removing CApplication::SetCurrentFileItem? I will come up with a PR doing this, plus some more cleanup in theaffected pvr code area. I want to avoid merge conflicts. Thanks.

ksooo:
Second commit is superseeded by #12621

FernetMenta:
2nd commit dropped, build error unrelated

PR #12620, opened 2017-08-03
[windowtranslator] windowname is std::string

mkortstiege:
This fixes an issue reported on the forums (https://forum.kodi.tv/showthread.php?tid=319304) where the DebugInfo fails to display the actual translated window name.

### DIFF
--- a/xbmc/input/WindowTranslator.cpp
+++ b/xbmc/input/WindowTranslator.cpp
@@ -244,12 +244,12 @@ int CWindowTranslator::GetFallbackWindow(int windowId)
   return -1;
 }
 
-std::map<int, const char*> CWindowTranslator::CreateReverseWindowMapping()
+std::map<int, std::string> CWindowTranslator::CreateReverseWindowMapping()
 {
-  std::map<WindowID, const char*> reverseWindowMapping;
+  std::map<WindowID, WindowName> reverseWindowMapping;
 
   for (auto itMapping : WindowMapping)
-    reverseWindowMapping.insert(std::make_pair(itMapping.second, itMapping.first.c_str()));
+    reverseWindowMapping.insert(std::make_pair(itMapping.second, itMapping.first));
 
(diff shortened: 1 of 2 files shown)


garbear:
You could also change `using WindowName = std::string;` to `using WindowName = const char*;` so save some allocations. Normally I'd take your approach, but it is over 100 strings...

MartijnKaijser:
If it's changing a lot so be it ;) Better do it correct if we can

xhaggi:
I would drop both `using WindowID` and `using WindowName` because you can't re-use it in all places. And what @garbear that use `const char*` instead of `std::string`.

xhaggi:
BTW as we use `TranslateWindow` a lot, should we use `std::unordered_map` instead.

garbear:
> I would drop both using WindowID and using WindowName because you can't re-use it in all places.

These aren't meant to be reused, they're aliases for code clarity

xhaggi:
@garbear without it's still understandable. We could add a comment to make it 100% clear for new devs.

mkortstiege:
@garbear, i tried to make it const char but it's not working as intended. Comparing the pointer values wont work. We might try to strcmp it in order to find the actual window name or go the std::string route.

garbear:
> @garbear, i tried to make it const char but it's not working as intended. Comparing the pointer values wont work. We might try to strcmp it in order to find the actual window name or go the std::string route.

Ah, now I remember, I used std::string to avoid strcmp. I would probably switch to strcmp for performance reasons. It's your call.

mkortstiege:
I'd like to avoid it as well.

garbear:
The current code will cause over 100 allocations per call to TranslateWindow(). Maybe we can use std::string, but generate a static struct and use that to translate windows?

mkortstiege:
Handled in https://github.com/xbmc/xbmc/pull/12627

PR #12621, opened 2017-08-03
[PVR] Cleanup PVR Manager's 'update current file' logics

ksooo:
... and remove some now unused stuff (with commit 2 and 3).

This change has ben runtime-tested on macOS, latest kodi master.

@FernetMenta mind taking a look?

### DIFF
--- a/xbmc/pvr/PVRManager.cpp
+++ b/xbmc/pvr/PVRManager.cpp
@@ -33,7 +33,6 @@
 #include "input/Key.h"
 #include "messaging/ApplicationMessenger.h"
 #include "messaging/helpers/DialogHelper.h"
-#include "music/tags/MusicInfoTag.h"
 #include "network/Network.h"
 #include "PlayListPlayer.h"
 #include "pvr/addons/PVRClients.h"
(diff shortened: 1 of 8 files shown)


xbmc/pvr/PVRManager.cpp
```diff
@@ -861,85 +860,17 @@ void CPVRManager::UpdateCurrentChannel(void)
...
 
-  CPVRChannelPtr channelTag(item.GetPVRChannelInfoTag());
-  CPVREpgInfoTagPtr epgTagNow(channelTag->GetEPGNow());
-
-  if (channelTag->IsRadio())
-  {
-    CMusicInfoTag* musictag = item.GetMusicInfoTag();
-    if (musictag)
-    {
-      musictag->SetTitle(epgTagNow
```
FernetMenta: I thought this updates current epg. Where is it updated if not here?

ksooo: This is dead code since you removed the call to CPVRManager::UpdateItem by no longer calling CInputStreamPVRMananager::UpdateItem.

ksooo: I have not found any misbehavior due to this, so far.

FernetMenta (approved):

xbmc/pvr/PVRManager.cpp
FernetMenta: thanks

PR #12622, opened 2017-08-03
[DVD] fix: libdvd's stream API doesn't handle incomplete reads well

Voyager1:
as title

## Description
streaming API in libdvd doesn't handle incomplete reads very well, as identified by a user who was trying to read DVD iso over an ftp connection.

## How Has This Been Tested?

tested using an ISO via an ftp connection

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamNavigator.cpp
+++ b/xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamNavigator.cpp
@@ -1589,7 +1589,7 @@ DVDNavVideoStreamInfo CDVDInputStreamNavigator::GetVideoStreamInfo()
 int dvd_inputstreamnavigator_cb_seek(void * p_stream, uint64_t i_pos)
 {
   CDVDInputStreamFile *lpstream = reinterpret_cast<CDVDInputStreamFile*>(p_stream);
-  if (lpstream->Seek(i_pos, 0) >= 0)
+  if (lpstream->Seek(i_pos, SEEK_SET) >= 0)
     return 0;
   else
     return -1;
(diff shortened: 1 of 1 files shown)


MilhouseVH:
Thanks, I'll include this in tonight's test build and get some feedback.

xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamNavigator.cpp
```diff
@@ -1598,7 +1598,24 @@ int dvd_inputstreamnavigator_cb_seek(void * p_stream, uint64_t i_pos)
...
+  while (i_ret < i_read)
+  {
+    int i_r;
+    i_r = lpstream->Read(reinterpret_cast<uint8_t *>(buffer) + i_ret, i_read - i_ret);
+    if (i_r < 0)
+    {
+      CLog::Log(LOGERROR,"read error dvd_inputstreamnavigator_cb_read");
+      return i_r;
+    }
+    if (i_r == 0)
```
ace20022: not sure, but shouldn't it be ```i_read - i_ret == 0``` ?

Voyager1: @ace20022 - forgot to mention that you will not always read "i_read" bytes. Sometimes at the end of a file you will read an incomplete block and i_r will be 0 before you read i_read bytes... The problem with FTP streams was apparently that one read operation only read e.g. 1600 bytes instead of 2048 and you needed to try again as long as you could read more.

Paxxi (approved):

Voyager1:
jenkins build this please

xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamNavigator.cpp
ace20022: okay, thx for the clarification!

ace20022 (approved):

Voyager1:
@MilhouseVH - have you received any feedback yet?

MilhouseVH:
@Voyager1 none so far. Unfortunately the user hasn't been back on the forum since reporting it, however I've just sent him a PM.

MilhouseVH:
@Voyager1 Still no feedback I'm afraid - have followed it up with a forum email. I assume it now works, but if not then take your time to fix it... :)

Voyager1:
@MilhouseVH - thanks. At least in my tests I reproduced the issue and was able to confirm that the fix remediated it.

PR #12623, opened 2017-08-04
Navigationfix when closing addons

a1rwulf:
This adds a new message to trigger PREVIOUS_WINDOW from a python addon.
When calling "close" from an addon we now trigger the new message instead of ACTIVATE_WINDOW.

## Motivation and Context
Scenario:
* You started up kodi and went into the tvshow section.
* You have access to an addon there and open it.
* The addon is not a dialog window but a normal window.
* When you hit back to close the addon, you are in the tvshow section again.
* When you hit back again, you assume to be taken to the home screen again, instead
  you are in the folder view of the MyVideoNav.xml

This happens because the ACTIVATE_WINDOW initializes the window without a start directory.
I can not think about any case where you want to not go "back" when closing an addon.

Maybe anyone can judge if this change is correct!?

## How Has This Been Tested?
Tested with kodi master and a windowed addon.

## Types of change
- [ x ] Bug fix (non-breaking change which fixes an issue)

@Paxxi Do I need any more documentation when I add new Appmessenger messages?
@xhaggi You've been in the reviewer suggestions ;)

### DIFF
--- a/xbmc/guilib/GUIWindowManager.cpp
+++ b/xbmc/guilib/GUIWindowManager.cpp
@@ -897,6 +897,12 @@ void CGUIWindowManager::OnApplicationMessage(ThreadMessage* pMsg)
   }
   break;
 
+  case TMSG_GUI_PREVIOUS_WINDOW:
+  {
+    PreviousWindow();
+  }
+  break;
+
   case TMSG_GUI_ADDON_DIALOG:
   {
     if (pMsg->lpVoid)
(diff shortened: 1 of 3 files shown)


xhaggi (approved):
IMO looks good and makes totally sense because addon windows behaves different and `Window::close()` should be the same as `CGUIWindow::OnBack()`.

```
bool CGUIWindow::OnBack(int actionID)
{
  g_windowManager.PreviousWindow();
  return true;
}
```

a1rwulf:
Cool.
Just in case anyone wonders:
If your addon is a Dialog close is handled by WindowDialogMixin.cpp which sends some other special message that ends up to be a OnBack later on, see:
https://github.com/xbmc/xbmc/blob/master/xbmc/interfaces/legacy/WindowDialogMixin.cpp#L45

myself-69 (approved):

PR #12625, opened 2017-08-04
[Music]Fix missing "* Item Folder" entry when browsing for artist fanart

DaveTBlake:
Fix missing "* Item Folder" entry when browsing for artist fanart from Artist Info Dialog


DaveTBlake:
OSX-64 fail nothing to do with this trvial change.

DaveTBlake:
OSX-64 fail nothing to do with this trvial change.

PR #12627, opened 2017-08-04
[windowtranslator] Restore const char* type

garbear:
This is an alternative fix for #12620 that replaces the map indexed by `std::string` with a set indexed by `const char*` using the `strcmp` comparator.

## Motivation and Context
Using const char* fixes the bug while avoiding over a hundred allocations on startup.

## How Has This Been Tested?
Confirmed working on slack.

## Types of change

### DIFF
--- a/xbmc/input/WindowTranslator.cpp
+++ b/xbmc/input/WindowTranslator.cpp
@@ -24,15 +24,10 @@
 #include "utils/StringUtils.h"
 
 #include <algorithm>
-#include <map>
+#include <cstring>
 #include <stdlib.h>
 
-namespace
-{
-using WindowName = std::string;
-using WindowID = int;
-
-static const std::map<WindowName, WindowID> WindowMapping =
+const CWindowTranslator::WindowMapByName CWindowTranslator::WindowMappingByName =
 {
     { "home"                     , WINDOW_HOME },
     { "programs"                 , WINDOW_PROGRAMS },
@@ -153,6 +148,8 @@ static const std::map<WindowName, WindowID> WindowMapping =
     { "gamevideosettings"        , WINDOW_DIALOG_GAME_VIDEO_SETTINGS },
 };
 
+namespace
+{
 struct FallbackWindowMapping
 {
   int origin;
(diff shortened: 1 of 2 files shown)


xhaggi:
You could use a map with a comparison functor, no need to make it more complicated or why you use a set now?

xbmc/input/WindowTranslator.cpp
```diff
@@ -25,14 +25,29 @@
...
+  const char *windowName;
+  int windowId;
+};
 
-static const std::map<WindowName, WindowID> WindowMapping =
+struct WindowMapCompare
+{
+  bool operator()(const WindowMapItem &lhs, const WindowMapItem &rhs) const
+  {
+    return strcmp(lhs.windowName, rhs.windowName) < 0;
```
xhaggi: Please use `std::strcmp`

garbear:
I forgot that map also supported comparator, but using a struct has the advantage of named fields, and others commented on my use of named types instead of named fields, so I prefer this method.

xhaggi:
I would stay with the map because we use it for the swapped/reversed one too. another improvement here would be to switch over to an `unordered_map`

garbear:
With `unordered_map` we would lose the O(logn) lookup time, as unordered_maps can't be indexed by const char* and also don't support comparators.

garbear:
With `std::map`:
```c++
reverseWindowMapping.insert(std::make_pair(itMapping.second, itMapping.first));
```
with `std::set`:
```c++
reverseWindowMapping.insert(std::make_pair(itMapping.windowId, itMapping.windowName));
```

much clearer

xhaggi:
If so please drop the map and also use a set with a different comparator for the swapped one.

```
using WindowIDMap = std::set<WindowMapItem, WindowIDCompare>;
...
WindowIDMap reverseWindowMapping;
reverseWindowMapping.insert(WindowMapping.begin(), WindowMapping.end());
```

garbear:
Good idea, updated

xbmc/input/WindowTranslator.h
```diff
@@ -52,5 +52,26 @@ class CWindowTranslator
...
+  {
+    bool operator()(const WindowMapItem &lhs, const WindowMapItem &rhs) const;
+  };
+
+  using WindowMapByName = std::set<WindowMapItem, WindowNameCompare>;
+  using WindowMapByID = std::set<WindowMapItem, WindowIDCompare>;
+
+  static WindowMapByID CreateReverseWindowMapping();
+
+  static const WindowMapByName WindowMapping;
```
xhaggi: Minor, but could we adjust the name to `WindowMappingByName` and `CreateWindowMappingByID`

garbear: done

xbmc/input/WindowTranslator.cpp
```diff
@@ -244,12 +252,11 @@ int CWindowTranslator::GetFallbackWindow(int windowId)
   return -1;
 }
 
-std::map<int, const char*> CWindowTranslator::CreateReverseWindowMapping()
+CWindowTranslator::WindowMapByID CWindowTranslator::CreateWindowMappingByID()
 {
-  std::map<WindowID, const char*> reverseWindowMapping;
+  WindowMapByID reverseWindowMapping;
```
xhaggi: Same here please `windowMappingByID`

xhaggi (approved):

xhaggi:
thank you for your patience and your time 👍

garbear:
thanks for the suggestions, they've led to much better code :+1: 

jenkins build this please

PR #12629, opened 2017-08-04
[PVR] Fix channel preview OSD (now behaves like in Krypton again).

ksooo:
@da-anda you reported a functional change in behavior on Slack. This PR restores the behavior as we had it in Krypton. up/down action now again ignores "channel switch delay" setting if "confirm channel switch with ok" setting is active.

This change has been runtime -tested on macOS, latest kodi master.

@FernetMenta mind taking a look.

### DIFF
--- a/xbmc/pvr/PVRManager.h
+++ b/xbmc/pvr/PVRManager.h
@@ -71,6 +71,12 @@ namespace PVR
     bool m_bStopped;
   };
 
+  enum class ChannelSwitchMode
+  {
+    NO_SWITCH,
+    DELAYED_SWITCH
+  };
+
   class CPVRManager : private CThread, public Observable, public ANNOUNCEMENT::IAnnouncer
   {
   public:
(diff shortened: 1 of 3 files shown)


da-anda:
thanks

FernetMenta (approved):
Maybe we should also fix this in settings. Having a delay with enabled confirmation by OK makes no sense imo.

xhaggi:
+1 what @FernetMenta said, it's possible to configure that this setting depends on the ok setting.

xbmc/pvr/PVRManager.h
```diff
@@ -71,6 +71,12 @@ namespace PVR
     bool m_bStopped;
   };
 
+  enum ChannelSwitchMode
```
xhaggi: If we don't need to convert the enum values to int we should use an enum class instead.

ksooo: Good point. Will change.

ksooo:
> Maybe we should also fix this in settings. Having a delay with enabled confirmation by OK makes no sense imo.

 Well, it makes sense. "Channel switch with OK" only affects up/down, but not ch+/ch- (pgup/pgdn). So, if you activate "Channel switch with OK", "switch delay" value is still of use when pressing ch+/ch-.

ksooo:
@xhaggi ChannelSwitchMode is now an enum class.

da-anda:
well, the direct channel up/down buttons/actions are unaffected by the
"confirm with ok" setting. For those the delay might still be needed,
doesn't it? But refactoring the settings is something for a separate PR.

Am 05.08.2017 9:39 vorm. schrieb "Sascha Woo" <notifications@github.com>:

ksooo:
> well, the direct channel up/down buttons/actions are unaffected by the
"confirm with ok" setting. For those the delay might still be needed,

What I said. :-)

xhaggi:
@ksooo thx

da-anda:
@ksooo, sorry, hadn't got the mail notification of your comment yet when I
wrote that reply via mail

Am 05.08.2017 3:52 nachm. schrieb "Sascha Woo" <notifications@github.com>:

PR #12630, opened 2017-08-04
[ADDON] Transfer ContentLookup properly

peak3d:
## Description
Currently calling listItem.setContentLookup from python addon does nothing, because it's value is not copied from Addon ListItem -> kodi ListItem in PluginDirectory. This PR does the copy step.

## Motivation and Context
Some Addons fail if Mime-Type is checked (HEAD request)

## Types of change


Rechi:
@peak3d backport needed?

peak3d:
Let me change the Platformdefs here as well because its the same issue, after that we can / should backport

PR #12631, opened 2017-08-05
[cleanup] silence compiler warnings

Rechi:
## Description
Silence compiler warnings.

## Motivation and Context
I don't like those warnings.

## How Has This Been Tested?
compiled for android, osx and windows

## Types of change

## Checklist:

### DIFF
--- a/xbmc/windowing/android/AndroidUtils.cpp
+++ b/xbmc/windowing/android/AndroidUtils.cpp
@@ -28,6 +28,9 @@
 #include <androidjni/System.h>
 
 #include <EGL/egl.h>
+
+#include <cmath>
+
 #include "AndroidUtils.h"
 
 #include "guilib/gui3d.h"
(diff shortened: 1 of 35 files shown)


xbmc/windowing/android/AndroidUtils.cpp
```diff
@@ -251,7 +251,7 @@ bool CAndroidUtils::SetNativeResolution(const RESOLUTION_INFO &res)
     CXBMCApp::SetDisplayMode(atoi(res.strId.c_str()));
     s_res_cur_displayMode = res;
   }
-  else if (abs(currentRefreshRate() - res.fRefreshRate) > 0.0001)
+  else if (std::abs(currentRefreshRate() - res.fRefreshRate) > 0.0001)
```
garbear: std::fabs? also missing `#include <cmath>` I think

Rechi: Doesn't `std::abs` have a float version (http://en.cppreference.com/w/cpp/numeric/math/fabs)?
Should I include `cmath` in this file also it currently compiles?

garbear: OK, `std::abs` is fine. You should include cmath in case another include is removed that breaks the code.

Rechi:
any objections?

FernetMenta (approved):

AchimTuran:
I'm fine with your changes too. Thank you for reducing the amount of warnings.

AchimTuran (approved):

PR #12635, opened 2017-08-05
VDPAU: fix potential crash

ZhouyangJia:
This patch fixes a potential bug that may cause a crash.
If "XOpenDisplay" failed to open XServer, "dl_vdp_device_create_x11" would cause a crash since "m_display" is null.
I think it's unsafe to assume that the library functions would be correct. It would be better if we could handle the errors properly.

## Types of change

## Checklist:


FernetMenta (approved):
Have you ever got this error? Just curious.

ZhouyangJia:
I write a bug detecting tool and find that the function XOpenDisplay frequently fails in many other projects (google "XOpenDisplay failed"). But it is clearly not properly handled here.
So I report this bug.

fritsch:
@ZhouyangJia thx very much.

PR #12637, opened 2017-08-05
Build libusb without -Werror

yol:
## Description
libusb has hardcoded `-Werror` CFLAG, but fails to compile on newer compilers (gcc-7.1) because of it.

## Motivation and Context
depends build should work

Also, libusb-0.1 is dead since 2006 or so. Alternatives: upgrade to libusb-1.0 and fix Kodi to use the new API, or build both libusb-1.0 and libusb-compat (for 0.1 API)

## How Has This Been Tested?
build on linux x86_64

## Types of change

## Checklist:


Rechi:
WIN-32 error unrelated
got the same error with gcc 7.1 and it is fixed with this PR

PR #12639, opened 2017-08-05
Game OSD: New dialog for video settings

garbear:
This adds a new dialog for selecting video effects in realtime.

## Description

I added nice icons, and moved the settings to a new dialog based on `Custom_1101_SettingsList.xml` used in the Video OSD.

Both "Video filters" and "View mode" use the same XML, `GameOSDVideoSelect.xml`. As the user scrolls, the game in the background is updated to use the focused scaling mode/view mode.

Currently, "Video filters" show a list of scaling modes. In the future, these will be replaced with shader effects.

View mode allows the user to select how the game is stretched to fill the display.

As part of the GSoC shader project, RetroPlayer is getting a new renderer that will allow previews to be drawn. For now, we just show a static image.

## Motivation and Context
We need a live fullscreen preview to show off our awesome shader work for GSoC.

## How Has This Been Tested?
Test builds are available in my test builds thread: https://forum.kodi.tv/showthread.php?tid=173361

## Types of change


garbear:
I've uploaded some test builds to my test builds thread: https://forum.kodi.tv/showthread.php?tid=173361

garbear:
pushed a commit that removes the dialogs' dependence on IPlayer, instead RetroPlayer registers itself with the dialogs

garbear:
I've pushed a new update. This adds a `gamewindow` control, similar to `videowindow`, and allows the rendering properties to be set. Now, the thumbnails in the dialog are rendered according to their effect.

screen shot 2017-08-11 at 1 58 52 pm

garbear:
Thanks to feedback from Rantanplan-1, I was able to combine GameOSDVideoSelect into DialogSelect in the latest commit.

DaveTBlake:
Forgive me if this is OT, but I just noticed that <defaultgamesettings>are in guisettings.xml in the old tag format, not the new <setting id="blah.blah" default="true">false</setting> format.  This is true for some other settings too, not just game. Should there be some update to the settings save/load implementation?

garbear:
Are you saying that the MediaSettings, including VideoSettings and the new GameSettings, are saved using the old format? If so, these should eventually be transitioned to the new format. Out of scope of this PR though, right?

DaveTBlake:
I'm sure out of scope :) 
I just happened to be looking at MediaSettings, and using the game settings for hints on how to do things, then had doubts. Hence raising it to make you aware and/or find out what the plan is (if there is one).

garbear:
If MediaSettings are saved using the old format, then a patch moving them to the new one would be welcome.

jenkins build this please

garbear:
Added a manager for game control rendering settings. jenkins build this please

garbear:
jenkins build this please

garbear:
unrelated errors on linux 64, rpi and ios

jenkins build this please

garbear:
Linux 64, rpi and ios 64 succeeded

gade01:
Isn't GameOSDVideoSelect.xml merged into DialogSelect.xml and should be removed?

garbear:
Looks like it got accidentally re-added after being removed. #12720

PR #12641, opened 2017-08-06
AESinkALSA: Return false if we don't know about our output channels

fritsch:
## Description
A sink is not allowed to return an unusable channel map. Therefore a check was introduced to find at least one channel that is not AE_CH_UNKNOWN1. Else we return false on Initialization.

## Motivation and Context
Lrusak gets issues with ffmpeg's resampler cannot init, cause of an empty destination layout, caused by ALSA sink returning UNKNOWN1, UNKNOWN1 map when it is queried while the display is off on refreshrate switching

## How Has This Been Tested?
Was not runtime tested. Let's wait on @lrusak for testing. Kodi should not crash anymore as we will fallback on the NULL sink first.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/AudioEngine/Sinks/AESinkALSA.cpp
+++ b/xbmc/cores/AudioEngine/Sinks/AESinkALSA.cpp
@@ -618,6 +618,11 @@ bool CAESinkALSA::Initialize(AEAudioFormat &format, std::string &device)
   }
   // adjust format to the configuration we got
   format.m_channelLayout = GetChannelLayout(format, outconfig.channels);
+  // we might end up with an unusable channel layout that contains only UNKNOWN
+  // channels, let's do a sanity check.
+  if (!format.m_channelLayout.IsLayoutValid())
+    return false;
+
   format.m_sampleRate = outconfig.sampleRate;
   format.m_frames = outconfig.periodSize;
   format.m_frameSize = outconfig.frameSize;
--- a/xbmc/cores/AudioEngine/Utils/AEChannelInfo.cpp
+++ b/xbmc/cores/AudioEngine/Utils/AEChannelInfo.cpp
@@ -257,6 +257,30 @@ CAEChannelInfo::operator std::string() const
   return s;
 }
 
+bool CAEChannelInfo::IsChannelValid(const unsigned int pos)
+{
+  assert(pos < m_channelCount);
+  bool isValid = false;
+  if (m_channels[pos] > AE_CH_NULL && m_channels[pos] < AE_CH_UNKNOWN1)
+    isValid = true;
+
+  return isValid;
+}
+
+bool CAEChannelInfo::IsLayoutValid()
+{
+  if (m_channelCount == 0)
+    return false;
+
+  for (unsigned int i = 0; i < m_channelCount; ++i)
+  {
+    // we need at least one valid channel
+    if (IsChannelValid(i))
+      return true;
+  }
+  return false;
+}
+
 const char* CAEChannelInfo::GetChName(const enum AEChannel ch)
 {
   assert(ch >= 0 && ch < AE_CH_MAX);
--- a/xbmc/cores/AudioEngine/Utils/AEChannelInfo.h
+++ b/xbmc/cores/AudioEngine/Utils/AEChannelInfo.h
@@ -51,6 +51,8 @@ public:
   inline unsigned int Count() const { return m_channelCount; }
   static const char* GetChName(const enum AEChannel ch);
   bool HasChannel(const enum AEChannel ch) const;
+  bool IsChannelValid(const unsigned int pos);
+  bool IsLayoutValid();
   bool ContainsChannels(const CAEChannelInfo& rhs) const;
   void ReplaceChannel(const enum AEChannel from, const enum AEChannel to);
   int BestMatch(const std::vector<CAEChannelInfo>& dsts, int* score = NULL) const;


fritsch:
@lrusak please test
@anssih for information as we discussed about it on slack

lrusak:
Seems to work, I tried playing the problematic file a bunch of times in a row and didn't crash.
http://sprunge.us/KVYY

fritsch:
Not sure I like what I see. It once tries to open the Analog output instead. In any case we will need this register / unregister thingy.

fritsch:
win64 build failure unrelated.

xbmc/cores/AudioEngine/Sinks/AESinkALSA.cpp
```diff
@@ -618,6 +618,18 @@ bool CAESinkALSA::Initialize(AEAudioFormat &format, std::string &device)
   }
   // adjust format to the configuration we got
   format.m_channelLayout = GetChannelLayout(format, outconfig.channels);
+  // we might end up with an unusable channel layout that only contains
+  // UNKNOWN1 channels, let's do a sanity check
+  bool isSaneLayout = false;
+  for (unsigned int i = 0; i < format.m_channelLayout.Count(); ++i)
+  {
+    isSaneLayout = (format.m_channelLayout[i] != AE_CH_UNKNOWN1);
```
FernetMenta: Unknown is only used for padding and must not precede any valid channel

fritsch: Therefore I check for only UNKNOWN1 channels. This code exits directly when we find an channel that is not UNKNOWN1.

FernetMenta: UNKNOWN1, UNKNOWN2 is no valid layout, UNKNOWN1, FL, FR is not valid either.

fritsch: Okay, so basically the check can be simplified. If UNKNOWN1 is the first channel, we directly return false?

FernetMenta: I don't see why the test should only test UNKNOWN1. Any unknown preceding a valid channel is error. Having no valid channel in the map is error too

fritsch: I don't see a code path in ALSA that might add anything else besides UNKNOWN1 as first channel - do you see one?

FernetMenta: ALSA codepath is not what counts here. Important is why we have something like UNKNOWN and how it can be used. We should always test against the spec and not against the current code that may change.

fritsch: That means I shall test for UNKNOWN1 to UNKNOWN64 as a first channel?

FernetMenta: >= AE_CH_UNKNOWN1 && < AE_CH_MAX

fritsch: Code wise it's not done quite well. Implementing an IsUnknown should be done in AEChannelData, but this one is only used in private API. Adding it to AEChannelInfo somehow sucks - but let's try that.

fritsch: @FernetMenta we don't have a ``` >= operator ``` I solved it differently, but it still is bad having the totally unrelated ```AE_CH_BROC``` in the code

xbmc/cores/AudioEngine/Sinks/AESinkALSA.cpp
```diff
@@ -618,6 +618,16 @@ bool CAESinkALSA::Initialize(AEAudioFormat &format, std::string &device)
   }
   // adjust format to the configuration we got
   format.m_channelLayout = GetChannelLayout(format, outconfig.channels);
+  // we might end up with an unusable channel layout starts with UNKNOWN
+  // channel, let's do a sanity check. UNKNOWN channels are only allowed for
+  // padding.
+  bool isSaneLayout = true;
+  if ((format.m_channelLayout.Count() > 0) && format.m_channelLayout.IsChannelUnknown(0));
```
FernetMenta: testing for IsChannelValid would seem more reasonalbe to me.
it should not only test index 0 but against the spec. FL, FL, UNKNOWN, BL, BR ist not a valid map

fritsch: Done

FernetMenta (approved):

anssih:
I think this breaks some HDMI ALSA channel layouts.

Looking at the HDMI kernel code, most of the time any empty channels in the middle seem to be reported as SND_CHMAP_UNKNOWN instead of SND_CHMAP_NA (I guess this is a driver bug, I can add it to my endless todo list), which then gets mapped to UNKNOWN1 on kodi, and those layouts are no longer considered valid (I think they worked at least in the past).

anssih:
Actually, scratch that, they (in-the-middle empty channels) are correctly reported as SND_CHMAP_NA by the kernel, but the ALSA sink maps that to UNKNOWN1, so there is an issue.

anssih:
So I guess either the "no UNKNOWNx in the middle of map" rule needs to be relaxed or we'll need to add AE_CH_EMPTY (or similar) for this purpose and map SND_CHMAP_NA to it.

fritsch:
@anssih I ammended two commits. They are plain ignored by the ffmpeg mapping code. So also not shipped empty. That will change the channeloutmap from e.g. FL,FR,EMPTY,EMPTY,SL,SR to FL,FR,SL,SR - not sure if we want that or should do that somehow else. That code addition would just do the same as it was handled before. UNKNOWNs were also ignored in the middle.

FernetMenta:
sorry, seems I was wrong with what I said. Haven't looked into this code for a while. ActiveAE requires at least one valid channel. Unknown in the middle is allowed. AE mutes channels in output map that are not in input.

fritsch:
@FernetMenta no problem. Will change the IsLayoutValid accordingly.

xbmc/cores/AudioEngine/Sinks/AESinkALSA.cpp
```diff
@@ -618,6 +618,12 @@ bool CAESinkALSA::Initialize(AEAudioFormat &format, std::string &device)
   }
   // adjust format to the configuration we got
   format.m_channelLayout = GetChannelLayout(format, outconfig.channels);
+  // we might end up with an unusable channel layout starts with UNKNOWN
+  // channel, let's do a sanity check. UNKNOWN channels are only allowed for
+  // padding. Also gaps with invalid channels are not allowed.
```
anssih: I think this comment needs updating now.

fritsch:
Updated.

lrusak:
Seems to work ok, http://sprunge.us/JHPM

fritsch:
jenkins build and merge

Rechi:
jenkins build this please

PR #12642, opened 2017-08-06
Fix crash when VAAPI is not available at runtime

yol:
## Description
Check vaInitialize return value in interop and fix related memleak

## Motivation and Context
Return value must be checked or VAAPI will segfault in vaCreateSurfaces if init failed e.g. due to missing driver

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/VAAPI.cpp
@@ -190,6 +190,7 @@ bool CVAAPIContext::CreateContext()
   int major_version, minor_version;
   if (!CheckSuccess(vaInitialize(m_display, &major_version, &minor_version)))
   {
+    vaTerminate(m_display);
     m_display = NULL;
     return false;
   }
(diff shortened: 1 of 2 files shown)


FernetMenta:
thanks. no need to compile test this on all platforms.

vindicatorr:
THANK YOU!
I just updated kodi from git and my issue(s) were finally fixed and I'm thinking this was the key here.
Essentially, given time, kodi was consuming 2MB of memory every second until linux completely locked up.
Also, cpu usage was higher (35-40%)than it was before (~5%) the bug was introduced (like a year ago?). And this cpu change was enough for me to look at the git log to see what VAAPI changes had occurred.
Thing is, I'd look at kodi's temp log and see VAAPI mentioned, but nothing really looked out of place. But whatever... IT WORKS NOW! I'm seeing great memory use and low cpu usage.
THANKS AGAIN!

pkerling:
Hi, as the memory leak was only on failed VAAPI initialization, it is very unlikely that the continuously increasing memory consumption you experienced was fixed by this PR. But anyway, good to see that your issues are resolved :-)

PR #12643, opened 2017-08-06
Remove dependency on zip always and on unzip for everything but Android

yol:
zip is never used and unzip only when building for Android.

## Motivation and Context
Should not require deps that are not used

## How Has This Been Tested?
depends build on linux

## Types of change

## Checklist:


koying:
What is unzip used for, only for droid?

Rechi:
@koying https://github.com/xbmc/xbmc/blob/dffd306/tools/depends/target/pythonmodule-pil/Makefile#L54

koying:
Fine by me as soon as it actually works ;)
```
checking for unzip... no
configure: error: "Missing program: unzip"
```

tools/depends/configure.ac
```diff
@@ -559,6 +549,11 @@ if test "$platform_os" == "android"; then
     AC_MSG_ERROR("Missing program: dx")
   fi
 
+  AC_PATH_PROG(UNZIP,unzip,"no",$build_tools_path)
```
Rechi: `AC_CHECK_PROG(HAVE_UNZIP,unzip,"yes","no",)`

pkerling: If that helps. Everything else uses `AC_PATH_PROG` though.

MartijnKaijser:
@koying can you try again?

koying:
The error message came directly from jenkins build this please ;)

MartijnKaijser:
Looks like it works now

koying (approved):

candrews:
The README's in https://github.com/xbmc/xbmc/blob/master/docs should also be updated to remove these dependencies.

Rechi:
@pkerling please squash the commits

pkerling:
@Rechi done

Rechi:
looks like drawin needs `unzip` (https://github.com/xbmc/xbmc/blob/dc52d70/tools/darwin/Support/Codesign.command#L63) and `zip` (https://github.com/xbmc/xbmc/blob/dc52d70/tools/darwin/Support/Codesign.command#L66)

pkerling:
> looks like drawin needs unzip (https://github.com/xbmc/xbmc/blob/dc52d70/tools/darwin/Support/Codesign.command#L63) and zip (https://github.com/xbmc/xbmc/blob/dc52d70/tools/darwin/Support/Codesign.command#L66)

Since this is not in depends and also needed when building without, shouldn't CMake check for it?

Rechi:
Darwin can't be build without depends.
@wsnipex any comment about where the check should be?

wsnipex:
I'd keep it in depends.

pkerling:
@wsnipex @Rechi OK now?

Rechi (approved):

PR #12644, opened 2017-08-06
Fix path autoconf-pre-depends looks in for m4

yol:
The pre-build-deps folder is under tools/depends so one .. is missing

## Motivation and Context
pre-depends build fails because autoconf does not find m4

## How Has This Been Tested?
depends build on linux

## Types of change

## Checklist:


Rechi (approved):
tested on linux and osx and now it works also if m4 is missing

PR #12648, opened 2017-08-08
[addons] save skin settings after they get changed

xhaggi:
Skin settings are not saved immediately after changing them, only on switching skins and a clean shutdown. To not lose the changes if Kodi crashes or something else happens the settings will now be saved after they get changed. This is done by a handler which do a delayed (atm 500ms) call to save the settings. Each update to a skin setting will trigger the handler but results in an unique save call within the delay period of 500ms.

@Montellese and @MartijnKaijser mind taking a look.

## Motivation and Context

Feature request: https://forum.kodi.tv/showthread.php?tid=319552

## How Has This Been Tested?

Runtime tested on MacOS

## Types of change

### DIFF
--- a/xbmc/addons/Skin.cpp
+++ b/xbmc/addons/Skin.cpp
@@ -34,6 +34,7 @@
 #include "messaging/helpers/DialogHelper.h"
 #include "settings/Settings.h"
 #include "settings/lib/Setting.h"
+#include "threads/Timer.h"
 #include "utils/log.h"
 #include "utils/StringUtils.h"
 #include "utils/URIUtils.h"
@@ -55,6 +56,22 @@ std::shared_ptr<ADDON::CSkinInfo> g_SkinInfo;
 
 namespace ADDON
 {
+
+class CSkinSettingUpdateHandler : private ITimerCallback
+{
+public:
+  CSkinSettingUpdateHandler(CAddon& addon)
+  : m_addon(addon), m_timer(this) {}
+  ~CSkinSettingUpdateHandler() override = default;
+
+  void OnTimeout() override;
+  void TriggerSave();
+private:
+  static constexpr int DELAY = 500;
+
+  CAddon &m_addon;
+  CTimer m_timer;
+};
 
 bool CSkinSetting::Serialize(TiXmlElement* parent) const
 {
@@ -185,6 +202,17 @@ std::unique_ptr<CSkinInfo> CSkinInfo::FromExtension(CAddonInfo addonInfo, const 
 
 CSkinInfo::CSkinInfo(
     CAddonInfo addonInfo,
+    const RESOLUTION_INFO& resolution /* = RESOLUTION_INFO() */)
+    : CAddon(std::move(addonInfo)),
+      m_defaultRes(resolution),
+      m_effectsSlowDown(1.f),
+      m_debugging(false)
+  {
+    m_settingsUpdateHandler.reset(new CSkinSettingUpdateHandler(*this));
+  }
+
+CSkinInfo::CSkinInfo(
+    CAddonInfo addonInfo,
     const RESOLUTION_INFO& resolution,
     const std::vector<RESOLUTION_INFO>& resolutions,
     float effectsSlowDown,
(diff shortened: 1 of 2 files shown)


MartijnKaijser:
See the comment from Montellese on the old PR that skins can change values all the time so you would still loose those changes.

xhaggi:
No, but now.

>And IIRC we already discussed the fix from MrMC and it was rejected because it could introduce serious performance issues when a lot of skin settings change in a short time.

We could introduce a job that will be triggered if settings are changed and save it only once.

xhaggi:
If you think we don't need that close the PR please.

Montellese:
IMO saving skin settings after the user has left the skin settings window makes perfect sense because the user manually changed those settings. Saving automatically changed skin settings does not seem as important though. Having a job that saves the settings periodically sounds reasonable though.

BTW the same problem applies to automatically changed (GUI) settings. If you e.g. change from fullscreen to windowed mode and then Kodi crashes it will start in fullscreen mode because the change of the (GUI) setting has not been automatically saved. Maybe we need something similar (i.e. periodic save) there as well.

I'll review this during the weekend when I have more time. Thanks for looking into it.

Jalle19:
I've seen skin settings disappear after reboots so any solution to this would be welcome.

xbmc/settings/windows/GUIWindowSkinSettings.h
```diff
@@ -0,0 +1,31 @@
...
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "guilib/GUIWindow.h"
+
+class CGUIWindowSkinSettings : public CGUIWindow
+{
+public:
+  CGUIWindowSkinSettings(void);
```
Montellese: No need for the `void` parameter in the ctor and dtor IMO.

xbmc/settings/windows/GUIWindowSkinSettings.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "GUIWindowSkinSettings.h"
+
+#include "addons/Skin.h"
+#include "guilib/WindowIDs.h"
+
+
+CGUIWindowSkinSettings::CGUIWindowSkinSettings(void)
+: CGUIWindow(WINDOW_SKIN_SETTINGS, "SkinSettings.xml")
+{
+  m_loadType = KEEP_IN_MEMORY;
```
Montellese: What is the default load type in `CGUIWindow`? If it's not `KEEP_IN_MEMORY` this might change behaviour.

Montellese (approved):
Looks good for me apart from the minor.

xhaggi:
@Montellese I've pushed a new commit which introduces a handler to do a delayed (500ms) call to save the settings after they get changed. If that looks good we could drop the first commit, because the skin settings will be saved immediately.

xbmc/addons/Skin.cpp
```diff
@@ -774,4 +812,17 @@ bool CSkinInfo::SettingsToXML(CXBMCTinyXML &doc) const
...
 
+void CSkinSettingUpdateHandler::OnTimeout()
+{
+  m_addon.SaveSettings();
+}
+
+void CSkinSettingUpdateHandler::TriggerSave()
+{
+  if (m_timer.IsRunning())
+    m_timer.Restart();
```
Montellese: This could result in the settings never being saved if a skin changes some setting at least once every 499ms.

Montellese (commented):
IMO the second approach is more flexible and should cover more cases. I'm not sure about the best delay though. 500ms would still mean writing the XML file twice every second in worst case.

xbmc/addons/Skin.cpp
xhaggi: Not sure if that is ever the case, but theoretically you're right.

xhaggi:
I would say, let's start with this and improve it if we got some feedback. I'll drop the first commit tomorrow.

xhaggi:
Dropped the first commit and updated the description of the PR.

MartijnKaijser:
Ignore win64 fail.

xhaggi:
@Montellese any objections?

Montellese:
Nope should be good to go.

xhaggi:
Should we backport it to v17?

MartijnKaijser:
@montellese your opinion on backporting this?

Montellese:
I don't have a strong opinion on this since I only ever change skin settings when I decide to use another skin and after that I usually never touch them again.

MartijnKaijser:
@xhaggi if you want to backport go ahead.

PR #12650, opened 2017-08-09
[CEC] add hibernate on TV standby

Rechi:
## Description
Resubmit #11436
Added hibernate mode on TV standby

## Motivation and Context
Hibernate is using less power than suspend mode

## Types of change

## Checklist:

### DIFF
--- a/system/peripherals.xml
+++ b/system/peripherals.xml
@@ -16,7 +16,7 @@
     <setting key="send_inactive_source" type="bool" value="1" label="36025" order="5" />
     <setting key="cec_standby_screensaver" type="bool" value="0" label="36009" order="6" />
     <setting key="cec_wake_screensaver" type="bool" value="1" label="36010" order="7" />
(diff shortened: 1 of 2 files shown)


Jalle19:
@Rechi so the minimum required libcec version doesn't have to be increased?

Rechi:
I've looked at https://github.com/Pulse-Eight/libcec/compare/libcec-4.0.0...libcec-4.0.1 and there wasn't any API change or change related to standby handling.
For me the version handling doesn't look optimal, but I know far tool little about what libcec expects and the way it was changed in the original PR also isn't optimal, so I skipped it.

Jalle19:
Okay, then it's probably fine like this

popcornmix:
On Pi the hibernate setting makes no sense (much like suspend).
Ideally shutdown options would be hidden from CEC settings when not available (as they are from shutdown menu), but that hasn't been done for suspend in CEC settings, so I wouldn't say that was a blocker for this PR.

PR #12651, opened 2017-08-09
DemuxClient: Force intialization of stream after OpenStream

peak3d:
## Description
Current Implementation requests a sream from addon after OpenStream is called.
If codec has not changed, values wich are evaluated from DemuxClient during packet analysis are not properly reinitialized.
This PR forces (re)initialization of all stream properties after OpenStream.

Example:
- Demuxclient requests on initialization an audio stream
- Demuxer does have codec, but not samplerate information, samplerate = 0
- DemuxClient calls OpenStream and requests changes (GetStream)
- Demuxer has now all necessary information, and returns correct samlerate
- DemuxClient does not set this new samplerate because codec has not changed.

## Motivation and Context
- solves Segfault in AudioSink at startup
- reduces initial streamchanges on video startup

## How Has This Been Tested?
strm file:
#KODIPROP:inputstreamaddon=inputstream.adaptive
#KODIPROP:inputstream.adaptive.manifest_type=hls
https://tungsten.aaplimg.com/VOD/bipbop_adv_fmp4_example/master.m3u8

inputstream.adaptive >= 2.0.7 available / activated

## Types of change

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -3729,11 +3729,6 @@ void CApplication::UpdateFileState()
            streamdetails if total duration > 15m (Should yield more correct info) */
         if (!(m_progressTrackingItem->IsDiscImage() || m_progressTrackingItem->IsDVDFile()) || m_pPlayer->GetTotalTime() > 15*60*1000)
         {
-          CStreamDetails details;
-          // Update with stream details from player, if any
-          if (m_pPlayer->GetStreamDetails(details))
-            m_progressTrackingItem->GetVideoInfoTag()->m_streamDetails = details;
-
           if (m_progressTrackingItem->IsStack())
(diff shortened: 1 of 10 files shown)


FernetMenta:
please elaborate on 2nd commit message.

peak3d:
Sure: While testing around with DemuxStreamChanges I arrived an segfault here:

https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/DVDFileInfo.cpp#L394

pDemux->GetStreams() returns an array of CDemuxStream* (and not std::shared_ptr)
Situation has been that this code was called from main thread (IIRC) but one of the streams was recreated during this time (and invalidates one of the demux streams).

I started changing all those places to std::shared_ptr but stopped after realizing how many files / changes are required.

FernetMenta (approved):

PR #12652, opened 2017-08-09
[guilib] multiimage don't fade out current image while new directory …

rmrector:
…is loading

## Description

The `multiimage` control loads some directories (from the filesystem when the directory isn't in an XBT, from plugins, maybe some other situations) in the background as it can take a bit of time, don't fade out the current image during that time.

## Motivation and Context

A slight delay in the artwork switch is much more pleasant than fading to transparent letting the background show through then fading the new image in after the directory has loaded. Jarvis also behaved this way, the fadeout was an unexpected side effect of #9715 introduced early in Krypton.

## How Has This Been Tested?

By hand, checked that the existing images doesn't fade out when changed to a slow loading directory and that changing to a single artwork is just as speedy as always.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/guilib/GUIMultiImage.cpp
+++ b/xbmc/guilib/GUIMultiImage.cpp
@@ -139,7 +139,7 @@ void CGUIMultiImage::Process(unsigned int currentTime, CDirtyRegionList &dirtyre
       }
     }
   }
-  else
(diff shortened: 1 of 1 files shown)


xhaggi:
Looks good

PR #12653, opened 2017-08-10
 [PVR] Clean the mess. Refactor channel navigation/selection, incl. channel preview.

ksooo:
Formerly, the channel preview implementation was scattered over a bunch of files, making it a huge mess. This PR tries to clean this up by encapsulating the implementation in its own class. At least I am able to understand now what the feature exactly does and how. ;-)

There are no functional changes for this PR, just a small bug fix fpr toggling the info display.

This has been carefully tested on macOS and Linux, latest Kodi master. 

@FernetMenta @xhaggi for review?

### DIFF
--- a/xbmc/music/windows/GUIWindowVisualisation.cpp
+++ b/xbmc/music/windows/GUIWindowVisualisation.cpp
@@ -29,6 +29,7 @@
 #include "guilib/GUIWindowManager.h"
 #include "input/InputManager.h"
 #include "input/Key.h"
+#include "pvr/PVRGUIActions.h"
 #include "pvr/PVRManager.h"
 #include "settings/AdvancedSettings.h"
 #include "settings/Settings.h"
@@ -50,13 +51,14 @@ CGUIWindowVisualisation::CGUIWindowVisualisation(void)
 bool CGUIWindowVisualisation::OnAction(const CAction &action)
 {
   if (CServiceBroker::GetSettings().GetBool(CSettings::SETTING_PVRPLAYBACK_CONFIRMCHANNELSWITCH) &&
-      g_infoManager.IsPlayerChannelPreviewActive() &&
-      (action.GetID() == ACTION_SELECT_ITEM || CServiceBroker::GetInputManager().GetGlobalAction(action.GetButtonCode()).GetID() == ACTION_SELECT_ITEM))
+      CServiceBroker::GetPVRManager().GUIActions()->GetChannelNavigator().IsPreview() &&
+      (action.GetID() == ACTION_SELECT_ITEM ||
+       CServiceBroker::GetInputManager().GetGlobalAction(action.GetButtonCode()).GetID() == ACTION_SELECT_ITEM))
   {
     // If confirm channel switch is active, channel preview is currently shown
     // and the button that caused this action matches (global) action "Select" (OK)
     // switch to the channel currently displayed within the preview.
-    CServiceBroker::GetPVRManager().ChannelPreviewSelect();
+    CServiceBroker::GetPVRManager().GUIActions()->GetChannelNavigator().SwitchToCurrentChannel();
     return true;
   }
 
--- a/xbmc/pvr/PVRActionListener.cpp
+++ b/xbmc/pvr/PVRActionListener.cpp
@@ -58,6 +58,15 @@ CPVRActionListener::~CPVRActionListener()
   g_application.UnregisterActionListener(this);
 }
 
+ChannelSwitchMode CPVRActionListener::GetChannelSwitchMode(int iAction)
+{
+  if ((iAction == ACTION_MOVE_UP || iAction == ACTION_MOVE_DOWN) &&
+      CServiceBroker::GetSettings().GetBool(CSettings::SETTING_PVRPLAYBACK_CONFIRMCHANNELSWITCH))
+    return ChannelSwitchMode::NO_SWITCH;
+
+  return ChannelSwitchMode::INSTANT_OR_DELAYED_SWITCH;
+}
+
 bool CPVRActionListener::OnAction(const CAction &action)
 {
   bool bIsJumpSMS = false;
@@ -111,9 +120,10 @@ bool CPVRActionListener::OnAction(const CAction &action)
     case REMOTE_8:
     case REMOTE_9:
     {
-      if (g_application.CurrentFileItem().IsLiveTV() &&
-          (g_windowManager.IsWindowActive(WINDOW_FULLSCREEN_VIDEO) ||
-           g_windowManager.IsWindowActive(WINDOW_VISUALISATION)))
+      if (!bIsPlayingPVR)
+        return false;
+
+      if (g_windowManager.IsWindowActive(WINDOW_FULLSCREEN_VIDEO) || g_windowManager.IsWindowActive(WINDOW_VISUALISATION))
       {
         // do not consume action if a python modal is the top most dialog
         // as a python modal can't return that it consumed the action.
@@ -124,10 +134,17 @@ bool CPVRActionListener::OnAction(const CAction &action)
         CServiceBroker::GetPVRManager().GUIActions()->GetChannelNumberInputHandler().AppendChannelNumberDigit(iRemote - REMOTE_0);
         return true;
       }
+      return false;
     }
-    break;
+    case ACTION_SHOW_INFO:
+    {
+      if (!bIsPlayingPVR)
+        return false;
 
-    // channel switching
+      CServiceBroker::GetPVRManager().GUIActions()->GetChannelNavigator().ToggleInfo();
+      return true;
+    }
+
     case ACTION_MOVE_UP:
     case ACTION_NEXT_ITEM:
     case ACTION_CHANNEL_UP:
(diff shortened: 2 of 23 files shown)


xbmc/music/windows/GUIWindowVisualisation.cpp
```diff
@@ -56,7 +57,7 @@ bool CGUIWindowVisualisation::OnAction(const CAction &action)
     // If confirm channel switch is active, channel preview is currently shown
     // and the button that caused this action matches (global) action "Select" (OK)
     // switch to the channel currently displayed within the preview.
-    CServiceBroker::GetPVRManager().ChannelPreviewSelect();
+    CServiceBroker::GetPVRManager().GUIActions()->GetChannelPreview().SelectChannel();
```
FernetMenta: would be great if we could cut the dependecy to PVR here. but this is work for another day

ksooo: Makes perfect sense. I found no good solution so far. But this is definitely on my radar for another day (PR).

xbmc/pvr/PVRManager.cpp
```diff
@@ -793,27 +771,22 @@ void CPVRManager::OnPlaybackStarted(const CFileItemPtr item)
...
-    }
+      const CPVRChannelPtr channel(item->GetPVRChannelInfoTag());
 
-    {
-      CSingleLock lock(m_critSection);
-      m_currentFile.reset(new CFileItem(*item));
-    }
+      m_addons->SetPlayingChannel(channel);
+
+      m_guiActions->GetChannelPreview().SetPlayingChannel(channel);
```
FernetMenta: CChannelPreview does a lot more than only preview. I would expect that if I have disabled preview, this class won't be required but it is used in normal code path here. Maybe find a moe appropriate name for it

xbmc/video/windows/GUIWindowFullScreen.cpp
```diff
@@ -31,6 +31,7 @@
 #include "video/dialogs/GUIDialogAudioSubtitleSettings.h"
 #include "guilib/GUIWindowManager.h"
 #include "input/Key.h"
+#include "pvr/PVRGUIActions.h"
```
FernetMenta: dependency to PVRManager is already not ideal. Adding an additional pvr dependency should be avoided.

xbmc/pvr/PVRManager.cpp
ksooo: Good point. Was thinking about a better name as well. I will try harder...

xbmc/video/windows/GUIWindowFullScreen.cpp
ksooo: Same "problem" as with GUIWindowVisualisation.cpp

As I said, I will get rid of the PVR dependency in another PR. IMO, larger changes are required for this.

Compared to the fat PVR Manager, PVR GUI Actions is tiny (very few includes needed e.g.)

xbmc/pvr/PVRManager.h
```diff
@@ -526,11 +518,6 @@ namespace PVR
     void SetPlayingGroup(const CPVRChannelPtr &channel);
 
     /*!
-     * @brief Save the currently playing channel as last played channel
-     */
-    void SaveLastPlayedChannel() const;
```
xhaggi: Thx for spotting ;)

xbmc/pvr/PVRGUIChannelPreview.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    if (bPlayingTV || bPlayingRadio)
+    {
+      const CPVRChannelGroupPtr group = CServiceBroker::GetPVRManager().GetPlayingGroup(bPlayingRadio);
+      if (group)
+      {
+        CSingleLock lock(m_critSection);
+        const CFileItemPtr item = bNext
+          ? group->GetByChannelUp(m_currentChannel)
+          : group->GetByChannelDown(m_currentChannel);;
```
xhaggi: Something for a rainy day: we should rename the methods to `GetPreviousNextChannel(m_currentChannel)` and `GetNextChannel(m_currentChannel)` .. ByChannelUp sounds odd and you don't really know what it mean, down from channel 1 to channel 2 or down from number 2 to number 1.

xbmc/pvr/PVRManager.cpp
ksooo: I will rename it to `CPVRGUIChannelNavigator`.

xbmc/pvr/PVRGUIChannelPreview.cpp
ksooo: Can do this in an additional commit in this PR, if you like.

xbmc/pvr/PVRGUIChannelPreview.cpp
```diff
@@ -0,0 +1,208 @@
...
+      }
+    }
+    return CPVRChannelPtr();
+  }
+
+  void CPVRGUIChannelPreview::PreviewChannel(const CPVRChannelPtr channel, ChannelSwitchMode eSwitchMode)
+  {
+    if (!g_infoManager.GetShowInfo() && eSwitchMode == ChannelSwitchMode::NO_SWITCH)
+    {
+      // just show info for current channel on first info activation.
```
xhaggi: we should adjust the comment, please! ATM I thought it's triggered `on first info activation` but that is not the case. We show the info for the current channel on the first up/down (`ACTION_MOVE_UP` / `ACTION_MOVE_DOWN`) action.

xbmc/pvr/PVRGUIChannelPreview.cpp
xhaggi: Would be good, thx.

xhaggi:
Apart from the minors +1

xhaggi (dismissed):

FernetMenta (approved):

ksooo:
@xhaggi @FernetMenta thanks for the review. I addressed all things you annotated, except the dependency from GUIFullscreenWindow and GUIMusicVisualizationWindow to PVR, which was not introduced by this PR. We agreed, to address this later.

xbmc/pvr/PVRGUIActions.h
```diff
@@ -323,6 +324,12 @@ namespace PVR
      */
     CPVRChannelNumberInputHandler &GetChannelNumberInputHandler();
 
+    /*!
+     * @brief Get the channel preview handler.
```
xhaggi: navigator instead of preview?

ksooo: Ha! Thanks.

ksooo:
jenkins build this please

ksooo:
Dependency of video fullscreen and visualization windows to pvr is now also gone. :-)

system/keymaps/keyboard.xml
```diff
@@ -661,8 +661,8 @@
       <right>StepForward</right>
       <up>Up</up>
       <down>Down</down>
-      <return>OSD</return>
-      <enter>OSD</enter>
+      <return>Select</return>
+      <enter>Select</enter>
```
xhaggi: Why do you change the mapping?

ksooo: Is this a problem?

xbmc/pvr/PVRActionListener.cpp
```diff
@@ -124,60 +134,66 @@ bool CPVRActionListener::OnAction(const CAction &action)
...
+          // switch to previewed channel
+          CServiceBroker::GetPVRManager().GUIActions()->GetChannelNavigator().SwitchToCurrentChannel();
+          return true;
+        }
+        else
+        {
+          // activate player controls
+          CGUIDialog *pOSD = g_windowManager.GetDialog(WINDOW_DIALOG_VIDEO_OSD);
+          if (pOSD && !pOSD->IsDialogRunning())
+            pOSD->Open();
```
xhaggi: Instead of handle the video osd here, we should ask for the global select action like we did before. I actually use select to open channels osd dialog and with your changes it's not possible to use the "confirm channel switch" feature anymore, because it never reach this `ACTION_SELECT_ITEM` case.

```
CServiceBroker::GetInputManager().GetGlobalAction(action.GetButtonCode()).GetID() == ACTION_SELECT_ITEM)
```

ksooo: Sorry, not sure waht you mean. This GetGlobalAction thing feels like a hack to me (although I introduced this myself, I know)...

ksooo: Could you please explain in detail what "I use select to open channels osd dialog" technically means?

ksooo: I think I now know what you mean. You have a custom keymap where you're overwriting the values for key "enter" and/or key "start" (and/or something similar in remote.xml whatever) , right?

Have to think about this scenario.

xhaggi: Yep, but in my case the keyboard.xml. `enter` and `return` is mapped to `Playlist`. That's the reason why we check the global action too.

xhaggi: BTW do be more clear. You can't map buttons to the player info because it's not a dialog or window. This info is only handled by an info bool `PLAYER_SHOWINFO`. Therefore that only way to get this working is asking for the global select action if we want the user to customize the buttons for the fullscreen/visualization windows.

If player info would be a dialog, this would be possible, but in that case STOP, PAUSE etc. should be delegated to the underlying fullscreen window if not handled.

ksooo: > If player info would be a dialog, this would be possible, but in that case STOP, PAUSE etc. should be delegated to the underlying fullscreen window if not handled.

 Yep. This is what I ment earlier with "to get rid of the dependecy of windowfullscreen and windowvisualization to PVR, larger changes are required". Too bad, I thought I had found an easy solution....

 I will remove this part from this PR, restoring the old implementation using getglobalaction, keeping the dependency.

xhaggi: IMO no need to revert everything. The handling could stay in `CPVRActionListener` + check for global action and you need to return `false` in `CGUIWindowVisualisation::OnAction` and `CGUIWindowFullScreen::OnAction` instead of calling the channel navigator.

ksooo:
> Dependency of video fullscreen and visualization windows to pvr is now also gone. :-)

 And it's again in. :-/ 

 Thanks @xhaggi for spotting that the approach I had chosen does not work. So, we will have to solve this indeed in another PR. No quick win possible here.

xhaggi (approved):
Ok lets do this in another PR.

ksooo:
jenkins errors are unrelated

da-anda:
in case you plan to work on the dedicated dialog for the OSD, I added a "fallback window" logic to our keymap handler, which means that you can specify a fallback window ID for the dialog, and if a custom button map is not found, it tries to resolve it from the fallback windows mapping. I added this for PVR fullscreen back then to get rid of the hardcoded keymaps back then.
So if you only define the "select" action for this dialog and specify PVRfullscreenX as fallback, their buttonmap will automatically be used for anythiing not specified in the dialogs keymap (no need to redefine play, pause, etc).

xhaggi:
Good to know ;)

PR #12654, opened 2017-08-10
Get stream url

rbuehlma:
This PR belongs to this discussion:
https://forum.kodi.tv/showthread.php?tid=318996&pid=2630102#pid2630102

It is intended as a replacement of strStreamURL

### DIFF
--- a/xbmc/addons/PVRClient.cpp
+++ b/xbmc/addons/PVRClient.cpp
@@ -1175,6 +1175,29 @@ bool CPVRClient::FillLiveStreamFileItem(const CPVRChannelPtr &channel, CFileItem
   return true;
 }
 
+bool CPVRClient::FillRecordingStreamFileItem(const CPVRRecordingPtr &recording, CFileItem *fileItem)
+{
+  if (!m_bReadyToUse)
+    return false;
+
+  PVR_RECORDING tag;
+  WriteClientRecordingInfo(*recording, tag);
+  PVR_FILE_ITEM_PROPERTY* properties = new PVR_FILE_ITEM_PROPERTY[PVR_ADDON_MAX_FILE_ITEM_PROPERTIES];
+  int propertyCount = PVR_ADDON_MAX_FILE_ITEM_PROPERTIES;
+  char url[PVR_ADDON_URL_STRING_LENGTH];
+  int urlLength = PVR_ADDON_URL_STRING_LENGTH-1;
+  PVR_ERROR err = m_struct.toAddon.GetRecordingStreamURL(tag, url, &urlLength, properties, &propertyCount);
+  if (err != PVR_ERROR_NO_ERROR) {
+    return false;
+  }
+  url[urlLength] = 0x00;
+  fileItem->SetDynPath(url);
+  for (int i = 0; i < propertyCount; i++) {
+    fileItem->SetProperty(properties[i].strName, properties[i].strValue);
+  }
+  return true;
+}
+
 PVR_ERROR CPVRClient::GetStreamProperties(PVR_STREAM_PROPERTIES *props)
 {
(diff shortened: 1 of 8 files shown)


xbmc/addons/PVRClient.cpp
```diff
@@ -1152,17 +1152,50 @@ bool CPVRClient::GetDescrambleInfo(PVR_DESCRAMBLE_INFO &descrambleinfo) const
   return false;
 }
 
-std::string CPVRClient::GetLiveStreamURL(const CPVRChannelPtr &channel)
+bool CPVRClient::FillLiveStreamFileItem(const CPVRChannelPtr &channel, CFileItem *fileItem)
```
FernetMenta: no need to pass channel anymore. it is a property of fileitem

xbmc/addons/PVRClient.cpp
```diff
@@ -1152,17 +1152,50 @@ bool CPVRClient::GetDescrambleInfo(PVR_DESCRAMBLE_INFO &descrambleinfo) const
...
   PVR_CHANNEL tag;
   WriteClientChannelInfo(channel, tag);
-  strReturn = m_struct.toAddon.GetLiveStreamURL(tag);
-  return strReturn;
+  PVR_FILE_ITEM_PROPERTY* properties = new PVR_FILE_ITEM_PROPERTY[PVR_ADDON_MAX_FILE_ITEM_PROPERTIES];
+  int propertyCount = PVR_ADDON_MAX_FILE_ITEM_PROPERTIES;
+  char url[PVR_ADDON_URL_STRING_LENGTH];
+  int urlLength = PVR_ADDON_URL_STRING_LENGTH-1;
+  PVR_ERROR err = m_struct.toAddon.GetLiveStreamURL(tag, url, &urlLength, properties, &propertyCount);
+  if (err != PVR_ERROR_NO_ERROR) {
```
FernetMenta: wrong coding style

xbmc/pvr/addons/PVRClients.h
```diff
@@ -282,11 +282,20 @@ namespace PVR
...
-     * @brief Get the stream URL for a given channel from the respective client.
-     * @param channel The channel to get the stream URL for.
-     * @return The stream URL or empty string if not available.
+     * @brief Fill the FileItem for a channel from the PVR backend.
+     * @param channel The channel to get the stream url for.
+     * @param fileItem The FileItem to be filled
+     * @return True if the url and and properties have been set
      */
-    std::string GetLiveStreamURL(const CPVRChannelPtr &channel);
+    bool FillLiveStreamFileItem(const CPVRChannelPtr &channel, CFileItem *fileItem);
```
FernetMenta: return value is never used

xbmc/addons/PVRClient.cpp
rbuehlma: Removed the parameter, also for recordings.

xbmc/addons/PVRClient.cpp
rbuehlma: Are you referring to the braces? I have corrected that.

xbmc/pvr/addons/PVRClients.h
rbuehlma: I know, in the current case it does not matter as the following code should be executed in both cases. I still think its a good idea to have this return parameter available.

Would you prefer if I remove it?

rbuehlma:
I added a commit. I will squash them all prior merging.

ksooo:
@rbuehlma Many thanks. Your idea is conceptionally fine, but I have so many annotations and wishes for changes to the current code, that it is less time consuming if I come up with a PR superceeding yours.

PR #12655, opened 2017-08-10
Patch for programs with empty audio/video streams

joaoppri:
## Description
Patch for programs with empty audio/video streams

## Motivation and Context
With this fix it will check if the program has a valid audio or video stream. 
Previous it would only check if it had streams, but not the type and if they were valid.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
@@ -534,6 +534,24 @@ bool CDVDDemuxFFmpeg::Open(CDVDInputStream* pInput, bool streaminfo, bool filein
       {
         nProgram = HLSSelectProgram();
       }
+      else
+      {
+        // skip programs without or empty audio/video streams
+        for (unsigned int i = 0; nProgram == UINT_MAX && i < m_pFormatContext->nb_programs; i++)
+        {
+          for (unsigned int j = 0; j < m_pFormatContext->programs[i]->nb_stream_indexes; j++)
+          {
+            int idx = m_pFormatContext->programs[i]->stream_index[j];
+            AVStream *st = m_pFormatContext->streams[idx];
+            if ((st->codec->codec_type == AVMEDIA_TYPE_VIDEO && st->codec_info_nb_frames > 0) ||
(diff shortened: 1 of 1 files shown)


xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxFFmpeg.cpp
```diff
@@ -1281,9 +1281,15 @@ void CDVDDemuxFFmpeg::CreateStreams(unsigned int program)
     // look for first non empty stream and discard nonselected programs
     for (unsigned int i = 0; i < m_pFormatContext->nb_programs; i++)
     {
-      if(m_program == UINT_MAX && m_pFormatContext->programs[i]->nb_stream_indexes > 0)
+      for (unsigned int j = 0; m_program == UINT_MAX && j < m_pFormatContext->programs[i]->nb_stream_indexes; j++)
       {
-        m_program = i;
+        int idx = m_pFormatContext->programs[i]->stream_index[j];
+        AVStream *st = m_pFormatContext->streams[idx];
+        if ((st->codec->codec_type == AVMEDIA_TYPE_VIDEO && st->codec_info_nb_frames > 0) ||
```
FernetMenta: we may not call av_find_stream_info and those parameres are not set. a program can only be selected using the info provided by pat/pmt or by property set on a fileitem

FernetMenta: you could do this in OpenDemuxer at a time we know that av_find_stream_info was called

joaoppri: @FernetMenta I committed new changes. thanks for the review. This new location to do the check should be safe for the av_find_stream_info.

joaoppri: @FernetMenta offtopic, I'm not a programmer, I know the basic and by examples. Could you tell me if it's possible a pvr addon, like the vu+, set the property "program" so then OpenDemuxer can get it with pInput->GetProperty("program")? Didn't find any example :(

FernetMenta: @joaoppri this API change https://github.com/xbmc/xbmc/pull/12660 will add this capability

PR #12656, opened 2017-08-10
keep extradata on stream-refresh if not given from demuxer

peak3d:
## Description

extradata extracted from stream by DemuxClient must not be removed if demuxer is requested for possible changed stream infos without providing extradata.

This issue is a side effect of: PR12651

## Motivation and Context
Live-TV with TVH does not provide video. Reason is that tvh demuxer does not provide codec extra-data.

## How Has This Been Tested?
Win10 / 32 TVH german channels

## Types of change

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxClient.cpp
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxClient.cpp
@@ -396,14 +396,6 @@ void CDVDDemuxClient::RequestStream(CDemuxStream *stream, std::map<int, std::sha
     streamAudio->iBlockAlign     = source->iBlockAlign;
     streamAudio->iBitRate        = source->iBitRate;
     streamAudio->iBitsPerSample  = source->iBitsPerSample;
-    if (source->ExtraSize > 0 && source->ExtraData)
-    {
-      delete[] streamAudio->ExtraData;
-      streamAudio->ExtraData = new uint8_t[source->ExtraSize];
-      streamAudio->ExtraSize = source->ExtraSize;
-      for (unsigned int j=0; j<source->ExtraSize; j++)
-        streamAudio->ExtraData[j] = source->ExtraData[j];
-    }
     streamAudio->m_parser_split = true;
     streamAudio->changes++;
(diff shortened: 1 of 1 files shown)


xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxClient.cpp
```diff
@@ -472,15 +456,6 @@ void CDVDDemuxClient::RequestStream(CDemuxStream *stream, std::map<int, std::sha
       if (streamSubtitle->m_parser)
         streamSubtitle->m_parser->flags |= PARSER_FLAG_COMPLETE_FRAMES;
     }
-
-    if (source->ExtraSize == 4)
```
FernetMenta: this check should not be dropped

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxClient.cpp
```diff
@@ -538,10 +513,27 @@ void CDVDDemuxClient::RequestStream(CDemuxStream *stream, std::map<int, std::sha
...
   {
     newStream->profile = stream->profile;
     newStream->level = stream->level;
+    extraSource = stream;
+  }
+  else if (oldStream != newStream)
+  {
+    newStream->profile = oldStream->profile;
+    newStream->level = oldStream->level;
+    extraSource = oldStream.get();
```
FernetMenta: this is kind of ugly and should not be required. if we create a new stream, we activate the parser that will split extradata. that will trigger a sream change event.

FernetMenta: that is even wrong because the old extradata may not match the new steeam's parameters

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxClient.cpp
peak3d: ok

da-anda:
@FernetMenta I suppose this one can be closed now that your PR is in?

PR #12658, opened 2017-08-11
[lang] remove language add-on migration code

tamland:
It's been 3 version since we migrated. No need for this anymore

### DIFF
--- a/xbmc/LangInfo.cpp
+++ b/xbmc/LangInfo.cpp
@@ -666,102 +666,23 @@ std::string CLangInfo::GetEnglishLanguageName(const std::string& locale /* = "" 
   return addon->Name();
 }
 
-bool CLangInfo::SetLanguage(const std::string &strLanguage /* = "" */, bool reloadServices /* = true */)
+bool CLangInfo::SetLanguage(std::string language /* = "" */, bool reloadServices /* = true */)
 {
-  bool fallback;
-  return SetLanguage(fallback, strLanguage, reloadServices);
-}
-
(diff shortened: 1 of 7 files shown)


MartijnKaijser (approved):

enen92:
Is this change for `strings.xml` or for the language folder layout (`resources/language/English/strings.po`)? or both?
Thanks

MartijnKaijser:
It's the handling of the transfer from when we had language files shipped with Kodi itself towards the Kodi language addons to find the matching one and download it. 
This shouldn't affect strings for addons itself.

xbmc/LangInfo.cpp
```diff
@@ -666,102 +666,23 @@ std::string CLangInfo::GetEnglishLanguageName(const std::string& locale /* = ""
...
-bool CLangInfo::SetLanguage(bool& fallback, const std::string &strLanguage /* = "" */, bool reloadServices /* = true */)
-{
-  fallback = false;
-
-  std::string language = strLanguage;
   if (language.empty())
-  {
     language = CServiceBroker::GetSettings().GetString(CSettings::SETTING_LOCALE_LANGUAGE);
 
-    if (language.empty())
```
Montellese: Theoretically this could still happen.

Montellese (approved):
Looks ok.

xbmc/LangInfo.cpp
MartijnKaijser: Should we still keep this for that theoretical case?

tamland: If it's empty it will print the below warning and fallback to default as it should. No need to have a special case for empty

PR #12659, opened 2017-08-11
[addons] log version of installed addons

tamland:


MilhouseVH:
Tested this on an RPi3 and works fine.

However changing the log level from INFO to NOTICE when reporting installed addon id would make it much easier to spot the cr@p that has been installed when someone is posting non-debug logs while seeking support.

MartijnKaijser:
Thx. 
Is there a reason you remove the repository log line?

tamland:
Ok, changed it to 'notice'

> Is there a reason you remove the repository log line?

No need to log it twice? Already included in the 'installed' log.

MilhouseVH:
Thanks for changing the NOTICE.

Just a thought, how difficult would it be to show enabled/disabled status, ie.:
```
21:18:50.229 T:1942881472  NOTICE: ADDON: script.module.html2text v2015.6.21 installed disabled
21:18:50.229 T:1942881472  NOTICE: ADDON: metadata.common.allmusic.com v3.1.1 installed enabled
```

And would it be possible to log the addons in identifier order? At the moment they're in a pretty random order, and I'm just thinking that it would make life easier looking for "plugin.video.exodus" (or whatever) if all the plugin addons were grouped together when sorted by id.

MartijnKaijser:
@tamland the things MilhouseVH proposed would indeed be nice if you had some spare moment.

tamland:
Due to global hell this is as much that can be done now (at least without elaborate workarounds, which I'm not going to add). The order is determined by cpluff, same as before.

MartijnKaijser:
understood. At least this is already a lot better :)

PR #12660, opened 2017-08-11
 [PVR][addons] PVR Addon API 5.6.0: Channel and recording URLs & properties

ksooo:
1. Replace API function `GetLiveStreamURL()` by `GetChannelStreamProperties()`
================================================================
Todo  for addon maintainers: Remove `GetLiveStreamURL` from your addon implementation. If your addon supports Live TV and/or Radio and it does implement live stream functions (`OpenLiveStream`, ...) (`bHandlesInputStream` set to false) you must implement the new API function. The new function basically does what `GetLiveStreamURL` did, but besides from the stream URL (as a new `property "streamurl"`) you can now also supply additional properties required to open the stream.

2. Add API function `GetRecordingStreamProperties()`
================================================================
Todo for addon maintainers: If your addon supports recordings and it does not implement recording stream functions (`OpenRecordedStream`, ...) you must implement the new API function. Functionality of the new function is similar to `GetChannelStreamProperties`.

3. Remove struct member `PVR_RECORDING::m_strStreamURL`
================================================================
Todo for addon maintainers: Remove references to `PVR_RECORDING::m_strStreamURL` from your addon implementation. Put the value you filled in there into the `property "streamurl"` in your implementation of the new API function `GetRecordingStreamProperties`.

Superseeds #12654 

Many thanks to @rbuehlma for the initial implementation which I shamelessly used as a base for this PR.

### DIFF
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
@@ -426,12 +426,24 @@ extern "C"
   PVR_ERROR GetDescrambleInfo(PVR_DESCRAMBLE_INFO* descrambleInfo);
 
   /*!
-   * Get the stream URL for a channel from the backend. Used by the MediaPortal add-on.
-   * @param channel The channel to get the stream URL for.
-   * @return The requested URL.
-   * @remarks Optional, and only used if bHandlesInputStream is set to true. Return NULL if this add-on won't provide this function.
+   * Get the stream properties for a channel from the backend.
+   * @param[in] channel The channel to get the stream properties for.
+   * @param[inout] properties in: an array for the properties to return, out: the properties required to play the stream.
+   * @param[inout] iPropertiesCount: in the size of the properties array, out: the number of properties returned.
+   * @return PVR_ERROR_NO_ERROR if the stream is available.
+   * @remarks Required if PVR_ADDON_CAPABILITIES::bSupportsTV or PVR_ADDON_CAPABILITIES::bSupportsRadio are set to true and PVR_ADDON_CAPABILITIES::bHandlesInputStream is set to false. In this case the implementation must fill the property PVR_STREAM_PROPERTY_STREAMURL with the URL Kodi should resolve to playback the channel. Return PVR_ERROR_NOT_IMPLEMENTED if this add-on won't provide this function.
    */
-  const char* GetLiveStreamURL(const PVR_CHANNEL& channel);
+  PVR_ERROR GetChannelStreamProperties(const PVR_CHANNEL* channel, PVR_NAMED_VALUE* properties, unsigned int* iPropertiesCount);
+
+  /*!
+   * Get the stream properties for a recording from the backend.
+   * @param[in] channel The recording to get the stream properties for.
+   * @param[inout] properties in: an array for the properties to return, out: the properties required to play the stream.
+   * @param[inout] iPropertiesCount: in the size of the properties array, out: the number of properties returned.
+   * @return PVR_ERROR_NO_ERROR if the stream is available.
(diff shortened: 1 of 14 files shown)


FernetMenta:
the descriptions soulds like all addons have to imlement this now?

rbuehlma:
Thanks for picking this up. I adapted the pvr.zattoo addon and it still works. Thanks.

rbuehlma:
> Need to revise this. What are the actual conditions for implementation?

Since the stream handling will be removed from the PVR API, and every Addon will need to provide an URL, I think every addon needs to implement these functions (if livetv and recordings are supported).

Or how else could a stream be played?

ksooo:
Well, today only certain add-ons must implement these functions, but my understanding is that at the long run (Leia final maybe) every add-on must implement  these functions. So, why not starting today to request proper implementation? 

@FernetMenta what do you think?

FernetMenta:
> but my understanding is that at the long run (Leia final maybe) every add-on must implement these functions

this is not correct. addons like vnsi and tvh, which implement their own inputstream api, do not need this. why should we ask them for an url for a channel and then feed this url back again.

FernetMenta:
on the other hand it does not harm if we require all addons to imlement this. at least they can set the props on the fileitem

ksooo:
Yes, I thought they would just return their static url (pvr://...) plus properties. How does this sound?

FernetMenta:
they can return whatever they want to. I won't return anything like "pvr://..." for vnsi because vnsi is not interested in this prefix. I will probably just return the channel number + a single prop which is vnsi for inputstream.

ksooo:
> they can return whatever they want to. I won't return anything like "pvr://..." for vnsi because vnsi is not interested in this prefix. I will probably just return the channel number + a single prop which is vnsi for inputstream.

 Okay, you are the insider for the streaming stuff. If you say that this is fine, so shall it be. :-)

FernetMenta (approved):

ksooo:
> I won't return anything like "pvr://..." for vnsi because vnsi is not interested in this prefix. I will probably just return the channel number + a single prop which is vnsi for inputstream.

 Thought about this a bit longer:

If we have use cases where a stream URL is meaningless, but properties are needed, why don't we make the stream URL a property itself? Means, we would drop the URL stuff (params, function name) from the two new API functions and returning the stream URL as a property named "streamurl", only if necessary for the addon to work only? This would coding make make simpler. So, we would end up with:

`PVR_ERROR GetChannelStreamProperties(const PVR_CHANNEL* channel, PVR_STREAM_URL_PROPERTY* properties, unsigned int* propertiesCount);`

and

`PVR_ERROR GetRecordingStreamProperties(const PVR_RECORDING* recording, PVR_STREAM_URL_PROPERTY* properties, unsigned int* propertiesCount);`

FernetMenta:
good idea

ksooo:
Okay, then I will update the code accordingly.

ksooo:
done.

xbmc/addons/kodi-addon-dev-kit/include/kodi/xbmc_pvr_dll.h
```diff
@@ -426,12 +426,24 @@ extern "C"
...
 
   /*!
-   * Get the stream URL for a channel from the backend. Used by the MediaPortal add-on.
-   * @param channel The channel to get the stream URL for.
-   * @return The requested URL.
-   * @remarks Optional, and only used if bHandlesInputStream is set to true. Return NULL if this add-on won't provide this function.
+   * Get the stream properties for a channel from the backend.
+   * @param[in] channel The channel to get the stream properties for.
+   * @param[inout] properties in: an array for the properties to return, out: the properties required to play the stream.
+   * @param[inout] propertiesCount: in the size of the properties array, out: the number of properties returned.
```
xhaggi: Could you please explain why it's necessary to pass the count in. Do you want that the addon could alter the max size to save some iterations in the fill methods?

ksooo: Will fix the typo. Thanks for spotting this.

Regarding the count parameter: This is C API. The addon has no chance to calculate the actual size of the passed array by itself. Alternative would be to always pass an array of a predefined size, which is less flexible.

xhaggi: Ok, thanks for explanation

ksooo:
jenkins errors are unrelated

PR #12661, opened 2017-08-11
Update themoviedb movie scraper to python

phate89:
Finally I managed to finish the first working copy of themoviedb python scraper. 
The themoviedb scraper is far from clean but I chose to create this pr to finally start the moving to python scrapers so every improvement is welcome. It's probably better to create a separate repo with it right?
I also have a tvdb scraper but current v2 api I have to use have several problems for us so it's on hold.
I used tmdb module to access tmdb api (a module based on requests).

## Description
This pr include changes to use our python api to set info to videoinfotag (first commit) and musicinfotag (second commit).

## Motivation and Context

Currently the addon gets the same results of the xml one (in theory) with some difference:
* Python scrapers sometimes match different items than the xml ones. This is because we sort in core the results provided only from xml. edit: This is a wanted change from notspiff to leave freedom to scraper devs
* It scrape more data that the old scraper left behind (for example the old scraper only got first studio and country, the python one adds all of them)
* It finally adds ratings and uniqueids with proper name 
* I added imdb top250 parsing so now it gets that info too

I'm not sure how it compares with old xml scrapers but I think old ones are still faster..

## How Has This Been Tested?
I tested them with several movies and compared the results directly in the db

## Types of change

## Checklist:


notspiff:
the idea was indeed to leave sorting to scrapers.

tamland:
Can you please split the addition of python modules into separate commits? It's impossible to find the relevant changes..

tamland:
Also there's a binary file in here (lcms2.dll)

phate89:
@tamland you're right I splitted all addons in separate commits and removed the dll file

Rechi:
TestScraperParser.General failed on all platforms

Razzeee:
Yes, the python stuff should get it's own repo. Maybe even on the org?

MartijnKaijser:
it's needs to be the same addon ID though because else there will be no auto conversion from xml to python scraper.

Razzeee:
Well then ignore me :)

phate89:
@mkortstiege I added the support for nfo urls in core and in the addon
@Rechi I removed the whole test because it doesn't make sense anymore since it's xml based and there's no way to do the same with python
@Razzeee the addon will probablyend in a repo but at least at the beginning I need one place to improve the addon without keeps both in sync. Maybe the alternative is to merge everything in core except the addon itself and develop it in another repo?

Razzeee:
Just do what you think is best. I just thoight I answere your question :)

Rechi:
@phate89 this needs a rebase

Rechi:
@phate89 this needs a rebase

MartijnKaijser:
@phate89 the idea is that we will include this once we branch of v18 so we have it early in v19 for regression testing.

anthonyryan1:
I'd like to bump this for consideration again.

By having the tmdb logic in a separate module from the scraper, it'll make it easier to share code and build more complex scrapers without reinventing the wheel. Also, these python scrapers feel much more maintainable to someone not familiar with the intricate XML replacement setup.

rmrector:
Most of the core changes from this were merged into Leia in #12857. The small music bit was left out as it will be better addressed with a larger reworking of Python scrapers for music. It also converted the demo scrapers to match the core changes rather than removing them.

The next small meaningful step planned is to move the Python add-on out to a separate git repo and _give it a new ID_. This needs use testing which can be done in Leia, and a new ID means it can be tested alongside the existing scrapers.

@phate89 Thanks for your work on this! You haven't been around much for awhile so I figure I'll handle this immediate plan, but if you want to jump in or have any thoughts to add I'd appreciate it.

I I'll also package it up and post it on the forums for wider use testing.

phate89:
I always wanted it to bring forward this pr but I don't have a lot of time.. feel free to take over..
if you need any help just ping me.. also since now the default tvshow scraper is tmdb will provide a library for that too (thetvdb scraper was  pretty worse because of crappy thetvdb api)

rmrector:
The last thing left from this PR is to split tmdbsimple out to a separate module, which will come in due course. Gonna close this.

Cheers!

PR #12664, opened 2017-08-12
Wayland windowing system

yol:
Add native Wayland windowing implementation as result of GSoC 2017 project from https://github.com/pkerling/xbmc/

## Description
This has been in the works for the last three months.

I have squashed the commits so that modifications and additions to the non-Wayland Kodi code are separate commits and that the Wayland windowing implementation itself is one commit. That implementation of course also has a history and is made up of a large number of commits (available via the linked GitHub repository) that were reviewed in feature chunks by @FernetMenta, but it would probably not be useful to keep it in mainline.

I realize that this is quite a lot of code to merge so I'm open to any suggestions that make it easier to handle.

Building with Wayland is fully integrated into the depends system and can be achieved by calling `tools/depends/configure` with `--enable-wayland`, so it should be easily integratable into Jenkins.

## Motivation and Context
Linux is transitioning to the Wayland protocol instead of X11 for the long term. To offer a competitive user experience on Linux and get all the shiny new features, Kodi must support Wayland natively instead of running via Xwayland X11 emulation.

## How Has This Been Tested?
Build and run test on linux64 with GL and GLES

It really looks just like on X :-)
(except for the window decorations which are better not shown off)

## Types of change

(This PR is mostly "new feature," but contains some commits that fix or improve core Kodi. I can list them individually if it helps.)

## Checklist:


fritsch:
Nice :-) I followed every issue and every issue review - really nice work. From my pov can go in every team if it does not break anything. Perhaps - only a suggest - it would be good to add "Wayland:" to the commits actually doing wayland features. This is especially useful when going just through a git log without seeing directly which file it changes.

But it's on your mentor to decide on this minor - only a suggestion.

pkerling:
iOS is OK now, Windows fail is unrelated

Regarding actually building the Wayland stuff, should I just copy linux64 to linux64wayland so that it can be added to jenkins?

xbmc/cores/VideoPlayer/Process/wayland/ProcessInfoWayland.h
```diff
@@ -1,7 +1,5 @@
-#pragma once
```
FernetMenta: git moved the deleted file. I guess this is not what you wanted to achieve

xbmc/guilib/GraphicContext.cpp
```diff
@@ -410,41 +411,138 @@ void CGraphicContext::SetVideoResolutionInternal(RESOLUTION res, bool forceUpdat
...
-  RESOLUTION_INFO info_org  = CDisplaySettings::GetInstance().GetResolutionInfo(res);
-
-  RESOLUTION_INFO info_mod = GetResInfo(res);
+  // FIXME Wayland windowing needs some way to "deny" resolution updates since what Kodi
+  // requests might not get actually set by the compositor.
+  // So in theory, m_iScreenWidth etc. would not need to be updated at all before the
+  // change is confirmed.
+  // But other windowing code expects these variables to be already set when
+  // SetFullScreen() is called, so set them anyway and remember the old values.
+  int origScreenWidth = m_iScreenWidth, origScreenHeight = m_iScreenHeight, origScreenId = m_iScreenId;
```
FernetMenta: put this on 3 separate lines

xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.h
```diff
@@ -196,6 +196,10 @@ class CRenderManager
   ERENDERSTATE m_renderState;
   CEvent m_stateEvent;
 
+  /// Display latency tweak value from AdvancedSettings for the current refresh rate
+  /// in seconds
```
FernetMenta: actually this is ms, right?

xbmc/settings/AdvancedSettings.cpp
```diff
@@ -375,11 +375,7 @@ void CAdvancedSettings::Initialize()
 
   m_enableMultimediaKeys = false;
 
-#if defined(TARGET_DARWIN_IOS)
```
FernetMenta: are you sure you want to change behavior for ios?

xbmc/utils/StringUtils.h
```diff
@@ -143,6 +143,69 @@ class StringUtils
...
+
+   \param d_first the beginning of the destination range
+   \param input Input string to be split
+   \param delimiter Delimiter to be used to split the input string
+   \param iMaxStrings (optional) Maximum number of splitted strings
+   \return output iterator to the element in the destination range, one past the last element
+   *       that was put there
+   */
+  template<typename OutputIt>
+  static OutputIt SplitTo(OutputIt d_first, const std::string& input, const std::string& delimiter, unsigned int iMaxStrings = 0)
```
FernetMenta: IMO we should not have those fat bodies in header files.

xbmc/windowing/XBMC_events.h
```diff
@@ -78,6 +80,10 @@ typedef struct XBMC_MoveEvent {
 	int y;		/* New y position */
 } XBMC_MoveEvent;
 
+struct XBMC_ModeChangeEvent {
```
FernetMenta: wrong format, we have brackets on next line

xbmc/windowing/XkbcommonKeymap.cpp
```diff
@@ -0,0 +1,351 @@
+/*
```
FernetMenta: this file is only used by wayland. hence it should be placed into wayland folder

FernetMenta:
Very good! Some minors then I think this is ready to go in.

xbmc/windowing/XkbcommonKeymap.cpp
pkerling: It really should be used by mir, too, that's why I put it into a common folder. I can move the file or commit the mir change.

xbmc/utils/StringUtils.h
pkerling: Having function bodies in headers is common C++ and for template functions there really is no alternative. Almost the whole STL is implemented like this.

xbmc/cores/VideoPlayer/Process/wayland/ProcessInfoWayland.h
pkerling: The `Keymap.h` removal is a separate commit in fact. Git just detects this as move when producing the diff for the whole history. There's nothing you can do about it as far as I know.

xbmc/guilib/GraphicContext.cpp
pkerling: done

xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.h
pkerling: Yes, very good catch! Forgot to change this line during the unit change.

xbmc/settings/AdvancedSettings.cpp
pkerling: This does not really change iOS behavior. The only place this is used is in DisplaySettings (see below) where the check now includes `g_Windowing.CanDoWindowed()`, which is false for iOS. The only difference is that you now cannot force Kodi to go windowed anyway by modifying `advancedsettings.xml`. As this makes zero sense on iOS in either case I don't think it has any negative impact.

xbmc/windowing/XBMC_events.h
pkerling: whole file is wrong format then, do you think it makes sense to only change this line?

xbmc/windowing/XkbcommonKeymap.cpp
FernetMenta: don't bother with mir, maybe it dies soon. for now it should be moved to wayland.

xbmc/windowing/XBMC_events.h
FernetMenta: sure it makes sense. the fact that someting is wrong is no justification to continue with doing it wrong.

xbmc/settings/AdvancedSettings.cpp
FernetMenta: ok

xbmc/windowing/XBMC_events.h
pkerling: No I think it's maybe justification to reformat the whole file :-)

xbmc/utils/StringUtils.h
FernetMenta: sure, but here we have a kind of ugly mixture. when I saw the method first, I did not see it was a template and could not find it in the body.
anyway, I have no strong feelings about this. If the team is ok with it, I am too

xbmc/windowing/XBMC_events.h
FernetMenta: you are not the maintainer of this file. if you want to reformat it, you should to submit a separate pr for it. cosmetic changes just inflate the diff

pkerling: ok

xbmc/windowing/XkbcommonKeymap.cpp
pkerling: ok

pkerling:
OK I think that was all of the minors

tools/depends/native/scons-native/Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+$(TARBALLS_LOCATION)/$(ARCHIVE):
+	cd $(TARBALLS_LOCATION); $(RETRIEVE_TOOL) $(RETRIEVE_TOOL_FLAGS) $(BASE_URL)/$(ARCHIVE)
+
+$(PLATFORM): $(TARBALLS_LOCATION)/$(ARCHIVE) $(DEPS)
+	-rm -rf $(PLATFORM)/*; mkdir -p $(PLATFORM)
+	cd $(PLATFORM); $(ARCHIVE_TOOL) $(ARCHIVE_TOOL_FLAGS) $(TARBALLS_LOCATION)/$(ARCHIVE)
+
+.installed-$(PLATFORM): $(PLATFORM)
+	cd $(PLATFORM); $(PREFIX)/bin/python setup.py install --prefix=$(PREFIX)
```
Rechi: does scons not have a build step?

tools/depends/target/wayland-protocols/Makefile
```diff
@@ -0,0 +1,47 @@
...
+CONFIGURE=./configure --prefix=$(PREFIX)
+
+CLEAN_FILES=$(ARCHIVE) $(PLATFORM)
+
+all: .installed-$(PLATFORM)
+
+$(TARBALLS_LOCATION)/$(ARCHIVE):
+	cd $(TARBALLS_LOCATION); $(RETRIEVE_TOOL) $(RETRIEVE_TOOL_FLAGS) $(BASE_URL)/$(ARCHIVE)
+
+$(PLATFORM): $(TARBALLS_LOCATION)/$(ARCHIVE) $(DEPS)
```
Rechi: something like https://github.com/xbmc/xbmc/blob/94faa07/tools/depends/target/crossguid/Makefile#L48-L52 is missing

tools/depends/target/waylandpp/Makefile
```diff
@@ -0,0 +1,54 @@
...
+SCONS=scons
+export CROSSCXX:=$(CXX)
+export CXX=$(CXX_FOR_BUILD)
+
+all: .installed-$(PLATFORM)
+
+$(TARBALLS_LOCATION)/$(ARCHIVE):
+	cd $(TARBALLS_LOCATION); $(RETRIEVE_TOOL) $(RETRIEVE_TOOL_FLAGS) $(BASE_URL)/$(ARCHIVE)
+
+$(PLATFORM): $(TARBALLS_LOCATION)/$(ARCHIVE) $(DEPS)
```
Rechi: something like https://github.com/xbmc/xbmc/blob/94faa07/tools/depends/target/crossguid/Makefile#L48-L52 is missing

tools/depends/configure.ac
```diff
@@ -89,6 +89,12 @@ AC_ARG_ENABLE([gplv3],
   [use_gplv3=$enableval],
   [use_gplv3=yes])
 
+AC_ARG_ENABLE([wayland],
+  [AS_HELP_STRING([--enable-wayland],
+  [build libraries needed for Wayland. default is no])],
+  [enable_wayland=$enableval],
+  [enable_wayland=no])
```
Rechi: IMO there should be a check if wayland can be used for the target platform.
OSX, ios, and android can't use it

tools/depends/native/scons-native/Makefile
pkerling: It's python setuptools, so it's debatable. I think you *can* call `setup.py build`, but not sure how useful that would be. tbh I just c&p-ed from https://github.com/xbmc/xbmc/blob/master/tools/depends/native/distribute-native/Makefile

tools/depends/target/wayland-protocols/Makefile
pkerling: added

tools/depends/target/waylandpp/Makefile
pkerling: added

tools/depends/configure.ac
pkerling: added

pkerling:
Just give me a ping if this is ready for squashing

Rechi:
please switch to the new style of component logging which was introduced at #12676

pkerling:
> please switch to the new style of component logging which was introduced at #12676

Done. Had to rebase anyway, squashed while at it.

FernetMenta (approved):

fritsch:
@FernetMenta  when do you push the button? Need to buy champagne before that I think :-)

FernetMenta:
if no objections, I will merge this this evening.

cmake/platform/linux/wayland.cmake
```diff
@@ -0,0 +1,17 @@
...
+set(PLATFORM_OPTIONAL_DEPS VAAPI)
+
+set(WAYLAND_RENDER_SYSTEM "" CACHE STRING "Render system to use with Wayland: \"gl\" or \"gles\"")
+
+if(WAYLAND_RENDER_SYSTEM STREQUAL "gl")
+  list(APPEND PLATFORM_REQUIRED_DEPS OpenGl)
+elseif(WAYLAND_RENDER_SYSTEM STREQUAL "gles")
+  list(APPEND PLATFORM_REQUIRED_DEPS OpenGLES)
+else()
+  message(SEND_ERROR "You need to decide whether you want to use GL- or GLES-based rendering in combination with the Wayland windowing system. Please set WAYLAND_RENDER_SYSTEM to either \"gl\" or \"gles\". For normal desktop systems, you will usually want to use \"gl\".")
```
wsnipex: Can we set sane defaults here? I.e gl on x86 and gles on arm

pkerling: @FernetMenta was against that and he convinced me. I also think now that it's better to not have a hardly predictable default that depends on CPU architecture and/or installed libraries.

wsnipex: I am pretty sure this will cause confusion. Having gl as default on x86 is hardly unpredictable

FernetMenta: it is, most wayland application most likely use GLES. this has absolutely nothing to do with processor arch.

pkerling: > I am pretty sure this will cause confusion. Having gl as default on x86 is hardly unpredictable

Hm, not sure why it would be confusing to have to choose. The message also explicitly states to try `gl` on normal systems (we can of course reword the message if it isn't clear enough). And if you're compiling for embedded you should know whether you want GLES or not.

wsnipex: How about making it dependent on which libs have been found/enabled? I.e if GL was enabled by the user, use them

FernetMenta: If someone builds Kodi, she/he should know what to build for and why. Otherwise refrain from building. Those kind of defaults are evil, you end up with an executable that may not fit your expectations.

wsnipex: In case it wasn't clear. We already have options for selecting gl or gles, those should be used instead of inventing another option that does the same thing.

pkerling: > We already have options for selecting gl or gles, those should be used instead of inventing another option that does the same thing.

I don't think there really is a strong precedent like you are implying. The only platform that at the moment even allows building with GL *or* GLES is mir, which is not used extensively. All other platforms are fixed to either GL or GLES.

pkerling: And the problem with the options we have (if you mean ENABLE_... cmake variables) is the following: Where should I add GL and GLES to in `wayland.cmake`? To `PLATFORM_REQUIRED_DEPS` or `PLATFORM_OPTIONAL_DEPS`? Neither fits in my opinion if you leave the decision up to ENABLE_...

wsnipex: I'd add them to optional, so both get searched for, but still retain the check here if one was enabled

pkerling: I can do that of course, but I think it's confusing because they are not in fact optional. You'd also have to disallow setting either of them to `AUTO` (since configuring must fail if it's not found). The current system isn't quite cut out for more complex dependencies like "you need either X or Y" :-/

wsnipex: True, those are valid concerns. Another option would be ENABLE_WAYLAND=OPENGL/OPENGLES

pkerling: I'm not sure I understand. You want to rename the variable from `WAYLAND_RENDER_SYSTEM` to `ENABLE_WAYLAND`?

wsnipex: I'm sorry, hard to keep an overview on a phone during holidays. I forgot that Wayland is enabled with core_platform name and not via enable_Wayland. So this doesn't make sense. Let s just keep what you have for now and see how it goes. We can always optimize later.

FernetMenta:
congratualtions and many thanks to @pkerling

MartijnKaijser:
Congrats and well done. Also for the other semi related fixes.

fritsch:
No need to stop helping and improving kodi for us! Thanks very much for this milestone work that pushes kodi to the next level into a great standard linux future.

paulpach:
This was exciting to watch.  This was the only reason why I don't have fedora in my HTPC,  A million thanks to @pkerling

da-anda:
congrats and thanks a lot pkerling

pkerling:
And there it goes 🎉  Thanks a lot to @FernetMenta for reviewing all of this stuff and giving valuable advice at every stage :-)

raneon:
Thank you so much pkerling!!! Highly appreciated :-)

PR #12546, opened 2017-07-21
RetroPlayer: Fix black screen on playback

garbear:
If the title of this PR looks familiar, it's because RetroPlayer broke yet again with the last VP update. This fixes the blank screen that shows when playing games.

## How Has This Been Tested?
Games play with video now.

## Types of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves existing functionality)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #12665, opened 2017-08-13
RetroPlayer fixes

garbear:
The primary fix is two missed calls to OnSpeedChange() from #12546, which can cause the game to play back when the menu is still open, or remain paused when the menu is closed.

This also fixes a compiler warning, refactors some GUI code outside of RetroPlayer, and refactors input scanning during gameplay to occur immediately before the frame instead of hard-coded at 120Hz.

## How Has This Been Tested?
Game speed no longer desynchronizes from OSD state.

## Types of change


xbmc/Application.cpp
```diff
@@ -3156,6 +3158,17 @@ PlayBackRet CApplication::PlayFile(CFileItem item, const std::string& player, bo
   if (item.GetMimeType().empty())
     item.FillInMimeType();
```
FernetMenta: I don't know what a game client is and many oth r devs don't know either. Application should not care about this. This is why we have the IPlayer interface and a player factory. Moving RP internals to the app layer is wrong.

garbear: OK, I'll find a better place for it

garbear:
Moved prompt into RetroPlayer, CApplication is now untouched

garbear:
jenkins build this please

PR #12670, opened 2017-08-13
[droid] Add visiblebehind and media session

koying:
Mediasession is needed for upcoming Google assistant integration


MartijnKaijser:
All seems to work as advertised. May need some refinement later on when there's more to test on.

PR #12673, opened 2017-08-14
[cleanup][libUPnP] remove platinum prebuild darwin frameworks

Rechi:
## Description
remove unused prebuild darwin frameworks

## Motivation and Context
reduces size of a shallow clone or zip/tar.gz download by about 15 MB

## How Has This Been Tested?
compiled for OSX and iOS

## Types of change

## Checklist:


Montellese:
Should be ok. Did you check if these prebuilt files also come with the source tarball provided by Platinum? If so we should add a patch file for this commit as well like we do for all other changes to the libUPnP directory.

Rechi:
@Montellese is https://github.com/plutinosoft/Platinum/releases upstream?
In the source code packages these prebuilt files are not present.

Montellese:
Back in the days I downloaded it from https://sourceforge.net/projects/platinum/files/platinum/. Maybe they've moved over to github by now but we are currently on version 1.0.5.13 which is only available on sourceforge and not on github.

MartijnKaijser:
So do we want to update libplatinum or move to the other one you worked on?

Montellese:
If someone wants to take a shot at it, updating libplatinum is the prefered approach right now because I don't have time to work on the ohnet based UPnP implementation. With all our patches to libplatinum in neat patch files it shouldn't be too hard to update.

Rechi:
superseded by 13244

PR #12674, opened 2017-08-14
Fix: check if we have to show simple menu only after we checked if we have a resume point

phate89:
## Description
Right now we skip bd simplemenu everytime we ask for a resume doesn't matter if there is. This pr changes that. It will skip simplemenu only if there's really a resumepoint set.

## Motivation and Context
Right now when we start "playdisc" builtin we ask to resume and there's a check in bd simplemenu that skips the menu if there's a resume request.
This means if you add a bd and start it with playdisc (from estuary main page for example) you won't ever get the bd simplemenu even if you set it and it will start to play directly the bluray. It's a problem that already exists in krypton but in leia it will even be worse because instead of loading the default playlist it will try to load the bd menu with different results (not all bd menus works great)

## How Has This Been Tested?
I tested with a couple of bluray and with resume points

## Types of change

## Checklist:

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -3207,14 +3207,6 @@ PlayBackRet CApplication::PlayFile(CFileItem item, const std::string& player, bo
     return PLAYBACK_FAIL;
   }
 
-  // a disc image might be Blu-Ray disc
-  if (item.IsBDFile() || item.IsDiscImage())
-  {
-    //check if we must show the simplified bd menu
-    if (!CGUIDialogSimpleMenu::ShowPlaySelection(const_cast<CFileItem&>(item)))
-      return PLAYBACK_CANCELED;
-  }
-
 #ifdef HAS_UPNP
   if (URIUtils::IsUPnP(item.GetPath()))
   {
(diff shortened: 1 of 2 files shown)


stefansaraev:
system/lcms2.dll should not be part of this pr :)

phate89:
mm it keeps popping up everytime not sure why.. Fixed it :)

ace20022 (changes requested):

xbmc/Application.cpp
```diff
@@ -3316,6 +3308,14 @@ PlayBackRet CApplication::PlayFile(CFileItem item, const std::string& player, bo
     }
   }
 
+  // a disc image might be Blu-Ray disc
+  if (!(options.startpercent > 0.0f || options.starttime > 0.0f) && (item.IsBDFile() || item.IsDiscImage()))
+  {
+	  //check if we must show the simplified bd menu
+	  if (!CGUIDialogSimpleMenu::ShowPlaySelection(const_cast<CFileItem&>(item)))
+		  return PLAYBACK_CANCELED;
+  }
```
phate89: mm locally was fine.. anyway fixed it!

ace20022 (approved):

ace20022:
Looks, good. Thanks for the pr!

phate89:
Do you think it's worth it to backport to krypton?

ace20022:
I don't have time to test it, but I trust you that it works in all cases. Therefore +1 for a backport.

PR #12676, opened 2017-08-15
[utils] CLog: add Log and LogFunction function with additional component parameter

Rechi:
## Description
Add additional `Log` and `LogFunction` functions to the `CLog` class which take a component parameter. The Functions check if component logging for that component is enabled and log in that case otherwise they do nothing.

## Motivation and Context
make component logging easier

## How Has This Been Tested?
compiled for windows

## Types of change

## Checklist:

### DIFF
--- a/xbmc/Util.cpp
+++ b/xbmc/Util.cpp
@@ -1923,7 +1923,7 @@ int CUtil::ScanArchiveForAssociatedItems(const std::string& strArchivePath,
                                          const std::vector<std::string>& item_exts,
                                          std::vector<std::string>& associatedFiles)
 {
-  CLog::LogFunction(LOGDEBUG, __FUNCTION__, "Scanning archive %s", CURL::GetRedacted(strArchivePath).c_str());
+  CLog::LogF(LOGDEBUG, "Scanning archive %s", CURL::GetRedacted(strArchivePath).c_str());
   int nItemsAdded = 0;
   CFileItemList ItemList;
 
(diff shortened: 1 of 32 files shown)


Montellese:
Makes sense IMO but right now it only compiles on windows.

popcornmix:
Makes sense to me too.
One issue I have with LogF is ``__FUNCTION__`` is not very informative. A lot of times this gives an ambiguous name like Create, Open, Close with no clues as to which component actually generated the message.

Most of the code I write uses:
```CLog::Log(LOGDEBUG, "%s::%s Destroying pool", CLASSNAME, __FUNCTION__);```
and I have to manually define ``CLASSNAME`` as it doesn't seem to be available from compiler.
This is a lot more useful information. I wonder if we could provide a standard way of achieving this?

It looks like gcc provides ``__PRETTY_FUNCTION__`` (with msvc ``__FUNCSIG__`` may be interesting, but I haven't investigated).

But  ``__PRETTY_FUNCTION__`` evaluates to something like:
``MMAL::CMMALPool::CMMALPool(const char*, bool, uint32_t, uint32_t, uint32_t, MMAL::MMALState)``
which is too verbose for logging, but truncating the string at the ``(`` might give something useful.

Rechi:
now it compiles on all platforms

afedchin (approved):

FernetMenta (approved):

MartijnKaijser:
@Rechi https://forum.kodi.tv/showthread.php?tid=298462&pid=2633192#pid2633192

PR #12678, opened 2017-08-16
[win32] Fix dxva renderer flickers at playback startup

afedchin:
see title


afedchin:
win32 failure unrelated

PR #12681, opened 2017-08-16
Mark some classes non-copyable + add default constructors

notspiff:
Correctness + cppcheck shut-up.

### DIFF
--- a/xbmc/ContextMenuItem.h
+++ b/xbmc/ContextMenuItem.h
@@ -62,6 +62,8 @@ private:
 class CContextMenuItem : public IContextMenuItem
 {
 public:
+  CContextMenuItem() = default;
+
   std::string GetLabel(const CFileItem& item) const  override { return m_label; }
   bool IsVisible(const CFileItem& item) const override ;
   bool IsParentOf(const CContextMenuItem& menuItem) const;
--- a/xbmc/ContextMenuManager.h
+++ b/xbmc/ContextMenuManager.h
@@ -47,8 +47,8 @@ public:
   ContextMenuView GetAddonItems(const CFileItem& item, const CContextMenuItem& root = MAIN) const;
 
 private:
-  CContextMenuManager(const CContextMenuManager&);
-  CContextMenuManager const& operator=(CContextMenuManager const&);
+  CContextMenuManager(const CContextMenuManager&) = delete;
+  CContextMenuManager& operator=(CContextMenuManager const&) = delete;
 
   bool IsVisible(
     const CContextMenuItem& menuItem,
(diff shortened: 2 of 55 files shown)


garbear:
There's a lot of places where non-copyableness is achieved by declaring protected constructors without a definition, e.g. `protected: Foo();` instead of `protected: Foo() = delete;`. Is there a way to detect and fix these? is the captain up to the challenge???

notspiff:
i've done the shit found by
```bash
#!/bin/bash

for file in `find . -name "*h"`
do
  lines1=`grep -z "private:.*operator=" $file | grep -v delete | wc -l`
  lines2=`grep -z "protected:.*operator=" $file | wc -l`

  if test $lines1 -gt 0
  then
    echo $file >> files-priv
  fi
  if test $lines2 -gt 0
  then
    echo $file >> files-prot
  fi
done

cat files-priv | sort > files-priv-s
cat files-prot| sort > files-prot-s

comm files-priv-s files-prot-s -3
rm -f files-priv files-priv-s
rm -f files-prot files-prot-s
```
that's the extent i will entertain your challenge. ie i will not hunt down stuff using private through no qualifier.

i had to resort to old-style default constructor for the Player class due to code generation (should fix build on jenkins).

notspiff:
i did spot some funkyness, disabling move constructors, prototyping and implementing through = default in the cpp file and even a place where shit was made private, then given a = default implementation in the cpp file. i chose not to touch these.

garbear (approved):

notspiff:
stpuid code generator :/ not worth the hassle, we can simply live with one more suppression in the cppcheck script.

notspiff:
kick the servant please

garbear:
jenkins build this please

notspiff:
all platforms has passed at some point, but please kick the bastard once more.

garbear:
jenkins build this please you loveable hunk of 1s and 0s

MartijnKaijser:
http://jenkins.kodi.tv/view/Helpers/job/BuildMulti-All/848/

Montellese:
IMO a constructor marked with `= delete` should always be `public` because you basically want to tell everyone that this object can't be constructed / copied i.e. it's part of the classes interface. Obviously before C++11 this could only be achieved by making the constructor(s) `protected` or `private`.

But this PR is fine with me either way, thanks for doing the monkey work.

MartijnKaijser:
Manual build succeeded

MartijnKaijser:
No objections and we can merge this?

Razzeee:
Building again for convinience

jenkins build this please

Razzeee:
Build fine and so far nobody objected to martijn, so I'm merging this now.

PR #12682, opened 2017-08-17
[Fix] Add .m4v to music tag loader factory as valid extension

DaveTBlake:
Add ".m4v" to tag loader factory as a valid extension so that .m4v files can be added to the music library. Taglib can handle that format of tag already, but it was omitted from the list of valid formats.

### DIFF
--- a/xbmc/music/tags/MusicInfoTagLoaderFactory.cpp
+++ b/xbmc/music/tags/MusicInfoTagLoaderFactory.cpp
@@ -80,6 +80,7 @@ IMusicInfoTagLoader* CMusicInfoTagLoaderFactory::CreateLoader(const CFileItem& i
       strExtension == "wma" || 
       strExtension == "flac" || 
       strExtension == "m4a" || strExtension == "mp4" || strExtension == "m4b" ||
+      strExtension == "m4v" ||
       strExtension == "mpc" || strExtension == "mpp" || strExtension == "mp+" ||
       strExtension == "ogg" || strExtension == "oga" || strExtension == "oggstream" ||
       strExtension == "opus" ||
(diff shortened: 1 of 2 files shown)


ace20022:
The commit message still says .m4a. m4v is a video container, why do you want to add these files to the music library?

DaveTBlake:
Oh poo, typo in commit too, I'll fix! Some users want to have music videos in the music library as music. MP4 files can be tagged (iTunes formt) and added, makes sense to do the same for .m4v

Discussed on forum https://forum.kodi.tv/showthread.php?tid=319796

MartijnKaijser:
Don't we have the music video library for that?iirc that should integrate with musiclibrary (might need some fixes)

DaveTBlake:
The music video library needs quite alot of work actually, waiting for an interested dev to take on that task. There is a non-team member looking at it, and I have offered to help where I can, but moving pictures is not my thing.  Anyway that really isn't the point here.

This is just about giving .m4v files the same functionality that Kodi already has for .mp4 and .m4a files. They are all video formats, but can be tagged (iTunes format) and Taglib can handle them, but .m4v was ommited from the list the factory sees as valid.

MartijnKaijser:
Maybe kick a test build for v17 so the user can test if it actually delivers

DaveTBlake:
>Maybe kick a test build for v17 so the user can test if it actually delivers

Have done so, but not sure how capable the main user asking for this is at using a test build. The reply on the forum was joy at a fix, not an offer to test. Maybe Scott will help out.

DaveTBlake:
User has tested this, well a Krypton build,  and it works.

PR #12684, opened 2017-08-17
[Fix]Add .m4v to music tag loader factory as a valid extension

DaveTBlake:
Add ".m4v" to tag loader factory as a valid extension so that .m4v files can be added to the music library. Taglib can handle that format of tag already, but it was omitted from the list of valid formats.

Discussed on forum https://forum.kodi.tv/showthread.php?tid=319796

Backport of #12682


DaveTBlake:
Test build on mirrors http://mirrors.kodi.tv/test-builds/windows/win32/KodiSetup-20170818-77a1adb799-HEAD-x86.exe

User testing feedback says that this is successful, so ready to be merged.

PR #12686, opened 2017-08-17
VideoPlayer - PVR: allow GetTimes for recordings

FernetMenta:
no reason why recordings should be excluded here


ksooo (approved):
Makes sense, although I don't know whether this change breaks something... ;-)

PR #12711, opened 2017-08-25
[PVR] Refactor and cleanup addon epg callback handling.

ksooo:
More structure, more object oriented and async epg state updates now better decoupled from pvr core (via a queue).

This change has been tested on macOS, latest  Kodi master.

@Jalle19 mind taking a look.

### DIFF
--- a/xbmc/pvr/epg/EpgContainer.cpp
+++ b/xbmc/pvr/epg/EpgContainer.cpp
@@ -31,6 +31,7 @@
 #include "settings/Settings.h"
 #include "settings/lib/Setting.h"
 #include "threads/SingleLock.h"
+#include "threads/SystemClock.h"
 #include "utils/log.h"
 
 #include "pvr/PVRManager.h"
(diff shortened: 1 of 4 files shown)


xbmc/pvr/epg/EpgContainer.cpp
```diff
@@ -40,7 +41,76 @@
 #include "pvr/recordings/PVRRecordings.h"
 #include "pvr/timers/PVRTimerInfoTag.h"
 
-using namespace PVR;
+namespace PVR
+{
+
+class CEpgUpdateRequest
```
Jalle19: Why not define this in the header file?

xbmc/pvr/epg/EpgContainer.cpp
```diff
@@ -40,7 +41,76 @@
...
+  if (!epg)
+  {
+    CLog::Log(LOGERROR, "PVR - %s - channel '%s' does not have an EPG! Unable to do the epg update!", __FUNCTION__, channel->ChannelName().c_str());
+    return;
+  }
+
+  epg->ForceUpdate();
+}
+
+class CEpgTagStateChange
```
Jalle19: Same thing here

xbmc/pvr/epg/EpgContainer.cpp
ksooo: Because it's an internal class and I always try to reduce type visibility to a minimum.

xbmc/pvr/epg/EpgContainer.cpp
ksooo: Same answer as before. ;-)

Jalle19 (approved):

PR #12713, opened 2017-08-25
don't use regex matching

tadly:
## Description

If the path within `${PROJECT_SOURCE_DIR}`contains special characters (e.g. `+`), the build will fail as they get interpreted by regex
## Types of change

## Checklist:

### DIFF
--- a/cmake/scripts/common/AddonHelpers.cmake
+++ b/cmake/scripts/common/AddonHelpers.cmake
@@ -52,7 +52,8 @@ macro (build_addon target prefix libs)
     # Read used headers from addon, needed to identitfy used kodi addon interface headers
     if(${prefix}_HEADERS)
       # Add the used header files defined with CMakeLists.txt from addon itself
-      if(${prefix}_HEADERS MATCHES ${PROJECT_SOURCE_DIR})
+      string(FIND "${${prefix}_HEADERS}" "${PROJECT_SOURCE_DIR}" position)
+      if(position GREATER -1)
         # include path name already complete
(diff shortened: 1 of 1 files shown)


Tadly:
Sorry it took this long.
I accidentally PR'ed against my own repo and forgot about it _face palms_

@Rechi you'll probably remember this :)

cmake/scripts/common/AddonHelpers.cmake
```diff
@@ -52,7 +52,7 @@ macro (build_addon target prefix libs)
     # Read used headers from addon, needed to identitfy used kodi addon interface headers
     if(${prefix}_HEADERS)
       # Add the used header files defined with CMakeLists.txt from addon itself
-      if(${prefix}_HEADERS MATCHES ${PROJECT_SOURCE_DIR})
+      if(${prefix}_HEADERS STREQUAL ${PROJECT_SOURCE_DIR})
```
Rechi: @AlwinEsch was your intention to use  `MATCHES` as exact string match or some kind of StartsWith, which cmake doesn't have?

AlwinEsch: No, was here during test the way who has worked :).
If "STREQUAL" can do the same it is also OK. Has not thought about circumstance with "+" in path.

Tadly:
Updated as of rechis request :)
Sorry it took so long

Rechi:
jenkins error unrelated, as this code isn't executed
build addons for windows and osx and everything still working

PR #12715, opened 2017-08-25
RetroPlayer: Add renderer for Windows, OpenGL, OpenGLES and MMAL

garbear:
This takes a significant step forward by separating the RetroPlayer and VideoPlayer renderers. The new RP renderer is pretty bare, and just contains enough code to render RGB images delivered from game add-ons.

## Motivation and Context
Based on https://github.com/garbear/xbmc/pull/84

## How Has This Been Tested?
Tested on Windows, Linux and OSX. The OpenGLES renderer hasn't been tested yet, but it uses mostly the same code as the OpenGL renderer.

## Types of change


garbear:
Rebased after the merge of #12665. Still TODO are a few remaining steps to fully separate RP and VP, including the separation of CProcessInfo.

xbmc/cores/RetroPlayer/process/RPProcessInfo.cpp
```diff
@@ -0,0 +1,121 @@
...
+    m_dataCache->SetVideoFps(0.0f);
+    m_dataCache->SetVideoDAR(1.0f);
+    m_dataCache->SetAudioDecoderName("");
+    m_dataCache->SetAudioChannels("");
+    m_dataCache->SetAudioSampleRate(0);
+    m_dataCache->SetAudioBitsPerSample(0);
+    m_dataCache->SetRenderClockSync(false);
+    m_dataCache->SetStateSeeking(false);
+    m_dataCache->SetSpeed(1.0f, 1.0f);
+    m_dataCache->SetGuiRender(true); //! @todo
```
garbear: @FernetMenta What is gui render? This isn't documented.

xbmc/cores/RetroPlayer/process/RPProcessInfo.cpp
```diff
@@ -0,0 +1,121 @@
...
+    m_dataCache->SetVideoRender(true); //! @todo
```
garbear: @FernetMenta What is video render? This isn't documented.

garbear:
jenkins build this please (to check GLES code)

popcornmix:
Just a warning that rendering through GLES will be significantly less efficient.

Prior to this PR on Pi we will directly render in the game's native bitmap format (e.g. RGB565, RGB888, RGBA32) to a separate layer without any copying required.
Now, then texture upload is effectively a memcpy to GPU memory, followed by a format conversion to a tiled texture format. There may also be an additional format conversion before texture upload to get the image into GL_BGRA format.

A few years ago all ARM platforms rendered video through GL. But performance sucked and now all ARM platforms render video directly on a separate video plane, mostly using zero copy.

I understand that for emulating 3D consoles you need GL rendering (although whether performance will  be adequate for this class of emulation is questionable). Also you may want it for shader based scanline effects. But I worry this scheme will rule out a class of emulation that previously worked acceptably but will no longer with the GL renderer.

It is possible we will want to support platform specific (e.g. MMAL) renderers for RP in the same way as we do for VP.

garbear:
Agreed, a MMAL renderer for RetroPlayer would be desirable. Do you have time to implement this? It's pretty much just a matter of copying the VP renderer and removing all the video stuff. If not, I can handle this eventually.

Using MMAL for rendering would preclude the use of shaders, right?

popcornmix:
I can help with MMAL renderer support when I get some time.
Ideally MMAL would be an optional renderer alongside GL.
If game uses GL or custom shaders then GL path is used otherwise use MMAL renderer.

I don't see a good way to use shaders alongside MMAL (you would lose the zero copy path so may as well render with GL as your pixels are already there).

fritsch:
Most ARMs (including IMX) offer some pseudo vendor optimized GL extensions to simulate a zero copy or at least a DMA copy to the 3D GPU. Still that would also less performant as "passthrough" rendering, but perhaps some custom shader / GL code could be supported that way?

eglCreateImage might be a candidate when doing it in a standard way?

xbmc/cores/RetroPlayer/rendering/VideoRenderers/RPWinRenderer.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+  m_format = format;
+  m_sourceWidth = width;
+  m_sourceHeight = height;
+  m_renderOrientation = orientation;
+
+  m_targetPixFormat = AV_PIX_FMT_BGRA;
+  m_targetDxFormat = DXGI_FORMAT_B8G8R8X8_UNORM;
+  if (g_Windowing.IsFormatSupport(DXGI_FORMAT_B8G8R8A8_UNORM, D3D11_USAGE_DYNAMIC))
+    m_targetDxFormat = DXGI_FORMAT_B8G8R8A8_UNORM;
```
afedchin: three lines above may be replaced with
`m_targetDxFormat = g_Windowing.GetBackBuffer()->GetFormat();`

garbear: Done

xbmc/cores/RetroPlayer/rendering/VideoRenderers/RPWinRenderer.cpp
```diff
@@ -69,7 +69,7 @@ bool CRPWinRenderer::Configure(AVPixelFormat format, unsigned int width, unsigne
   m_renderOrientation = orientation;
 
   m_targetPixFormat = AV_PIX_FMT_BGRA;
-  m_targetDxFormat = DXGI_FORMAT_B8G8R8X8_UNORM;
+  m_targetDxFormat = g_Windowing.GetBackBuffer()->GetFormat();
   if (g_Windowing.IsFormatSupport(DXGI_FORMAT_B8G8R8A8_UNORM, D3D11_USAGE_DYNAMIC))
```
afedchin: this (and line below) may be removed then

garbear: Thanks

garbear:
Updated with fixes from the last week

garbear:
Rebased after #12761, included recent changes from my shaders work and squashed. Just one bit of code in the GLES renderer needed before merging.

xbmc/cores/RetroPlayer/rendering/VideoRenderers/RPRendererOpenGLES.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+  GLint filter = (m_renderSettings.VideoSettings().GetScalingMethod() == VS_SCALINGMETHOD_NEAREST ? GL_NEAREST : GL_LINEAR);
+  glTexParameteri(m_textureTarget, GL_TEXTURE_MAG_FILTER, filter);
+  glTexParameteri(m_textureTarget, GL_TEXTURE_MIN_FILTER, filter);
+
+  //! @todo Translate OpenGL code to GLES
+  /*
+  glTexEnvf(GL_TEXTURE_ENV, GL_TEXTURE_ENV_MODE, GL_MODULATE);
+
+  glBegin(GL_QUADS);
```
garbear: How do I render an RGB texture with GLES?

FernetMenta: this is not implemented for GLES

garbear: Then what do I do? Convert to YUV and use Kodi's GLES YUV code?

FernetMenta: how have you done it in the past?

garbear: In the past I used VP's RenderManager. I converted to YUV and handed it off.

Now, with the new renderer, I'm trying to avoid YUV and render the RGB image directly.

FernetMenta: If the source is RGB, it should be rendered RGB. If renderer does not support this format there are two options: a) implement workaround in renderer, b) disabled RP on this platform
IMO full platform coverage is not a requirement for a feature. Missing parts can be implemented by platform maintainers at a later time.

xbmc/cores/RetroPlayer/process/RPProcessInfo.cpp
VelocityRa: Presumably it's for when rendering the GUI pass in VP/RM (`Render` is called with the `gui` parameter set to true).
Likewise, `SetVideoRender` below is probably for when rendering the output frame. But I don't know if you just need to set both to true once, or update them multiple times every frame while rendering.
As with anything from `CDataCacheCore`, updating it is optional.

xbmc/cores/RetroPlayer/rendering/GUIRenderTarget.h
```diff
@@ -0,0 +1,85 @@
...
+  // --- CGUIRenderFullScreen --------------------------------------------------
+
+  class CGUIRenderFullScreen : public CGUIRenderTarget
+  {
+  public:
+    CGUIRenderFullScreen(IRenderManager *renderManager, CGameWindowFullScreen &window);
+    ~CGUIRenderFullScreen() override = default;
+
+    void Render() override;
+    void RenderEx() override;
```
VelocityRa: What's this? How is it different from `Render`? Going by APIs like Win32, `MethodEx` would stand for "`MethodExtended`", meaning that it would provide you with more arguments to tweak, which isn't the case here.
This comment goes for every other definition of `RenderEx`.

xbmc/cores/RetroPlayer/rendering/VideoRenderers/RPRendererOpenGLES.cpp
VelocityRa: Looks like OpenGL ES supports at least some the formats we want (`RGB555`, `RGB565`): [Source in the spec]https://www.khronos.org/registry/OpenGL/specs/es/2.0/es_full_spec_2.0.pdf#page=127&zoom=auto]
@garbear If I recall correctly libretro cores also need BGRA8 support? If so we can probably use [OES_rgb8_rgba8](https://www.khronos.org/registry/OpenGL/extensions/OES/OES_rgb8_rgba8.txt) and swizzle our way into BGRA8 in the shader.

FernetMenta: rendering RBG is trivial, it's just not implemented by Kodi's GLES renderer

VelocityRa: "RGB" isn't very helpful, unless you mean RGB8 since that's the most common.
Anyway, the current renderer simply outputs in **BGRA8**. However, it relies on sws_scale to do the conversion from the possible _input_ formats taken from libretro cores.

My point above was about using those formats as the format _of the backbuffer_, so that we can achieve "zero copy" rendering and avoid sws_scale/software scaling altogether.
I looked into that for the D3D renderer, but these formats are only supported on DXGI 1.2+ devices unfortunately (no D3D 9), so we didn't go for it.

xbmc/cores/RetroPlayer/rendering/GUIRenderTarget.h
garbear: Render() is called on the GUI pass. RenderEx() is called on the video rendering pass. Eventually RetroPlayer will support h264 decoding for game streaming, so we'll need use of the video pass.

garbear:
@popcornmix I've started working on MMAL support. The buffers in RetroPlayer are very basic and can't handle things like buffer pools. I'll let you know when the new buffer architecture is complete so we can get MMAL rendering to work.

One of the features that flew into this PR before the last rebase was the ability to have multiple renderers at the same time. In the GSoC code, renderers are indexed by shader preset and are swapped around as needed. I don't see any reason we can't switch between MMAL and GLES depending on which features they support.

garbear:
@popcornmix I've added a MMAL renderer based on the VideoPlayer code. It needs some more work, but should serve as a good base for MMAL support.

Now that we have a buffer and buffer pool abstraction, I'll port the existing renderers to it so that we can achieve zero-copy support.

xbmc/cores/RetroPlayer/guicontrols/GUIGameControl.h
```diff
@@ -19,44 +19,64 @@
...
+  void UnregisterControl();
 
-  CGUIInfoLabel m_videoFilterInfo;
+  // GUI properties
+  CGUIInfoLabel m_scalingMethodInfo;
   CGUIInfoLabel m_viewModeInfo;
 
-  CGUIRenderSettings m_renderSettings;
+  // Rendering properties
+  bool m_bHasShaderPreset = false;
```
Rechi: unused private field

garbear: Good catch, this snuck its way in from https://github.com/garbear/xbmc/pull/86

Rechi:
@garbear this needs a rebase

garbear:
I've finished implementing render buffers for the remaining renderers. The code is based on Video Buffers, with a few improvements. It can handle multiple buffer pools. This means that we can use both MMAL and OpenGL renderers simultaneously in the video selection dialog from #12639.

@popcornmix I copied the MMAL code from VideoPlayer and haven't tested it yet. Can you test the renderer? If it needs more work I can set up an RPi debugging environment.

@koying @Memphiz @peak3d We need code for drawing a quad on OpenGLES. Can one of you help?

Memphiz:
Inhave zero skills in that regard

koying:
Sorry to sound negative, but I don't see in what an additional, specific, renderer to maintain, considering our ever-shrinking active development team, is an improvement vs. enriching the existing one, but hey...

@lrusak is the one most busy with GLES those days....

VelocityRa:
@koying I understand the maintenance concern, but there's obviously good reasons for all this.
https://github.com/garbear/xbmc/pull/84#issuecomment-313079642

**TLDR**: VP and RP have _very_ different goals and so do their renderers.

FernetMenta:
check out how picture renderer does it: https://github.com/xbmc/xbmc/blob/master/xbmc/pictures/SlideShowPicture.cpp#L917

koying:
@VelocityRa No doubt RP & VP have different goals, but a renderer is a renderer. Just creating a brand new one because the current one does not support RGB makes little sense.
Argument like "VP will advance to work without a gui" is irrelevant for a renderer.

Up to jarvis/dvdplayer, the renderer was actually a component of its own, before it got "phagocyted" by VP. The remaining arguments ("private" API's) are a direct result of this decision.
It might be wiser to reconsider the decision of making the renderer private to VP rather than duplicate.

VelocityRa:
@koying 
It's _definitely_ not just RGB. It's a bunch of stuff that VP doesn't have any use for, one of which is the shader preset rendering pipeline. It would be much much more difficult and in my opion messy, to just slap whatever RP needs to VP.

>It might be wiser to reconsider the decision of making the renderer private to VP rather than duplicate.

Oh you mean separate rendering from VP? Definitely. But it sounds like it would be significantly more difficult than making a "bare essentials" RP renderer. 
Existing VP code simply doesn't have to change _at all_ that way too (all the trouble that this would ensue is averted).

Furthermore - you said it youself - RP & VP have different goals; I think you'll find that the common code between them at this point is pretty insignificant to justify "combining" them that way.

xbmc/windowing/amlogic/WinSystemAmlogic.cpp
```diff
@@ -24,6 +24,9 @@
 #include <float.h>
 
 #include "ServiceBroker.h"
+#include "cores/RetroPlayer/process/RPProcessInfo.h"
+#include "cores/RetroPlayer/rendering/VideoRenderers/RPRendererOpenGLES.h"
+#include "cores/RetroPlayer/rendering/VideoRenderers/RPRendererOpenGLES.h"
```
peak3d: 2 identical includes here

peak3d:
I'm an 100% opponent for duplicting things.
Could be true that RenderManager must be duplicated, but memoryPool / Buffers are IMO abstract enough to reuse the current implementation

VelocityRa:
@peak3d Agreed.

garbear:
> I'm an 100% opponent for duplicting things.

My reason for writing a new renderer is partially due to the duplication in VP. For example, both the sysmem video buffers and the mmal video buffers use an identical pooling strategy. By copying the existing renderers, I've created a much smaller, more encapsulated renderer that can be refactored at a fraction of the cost, and by supporting less systems my task of testing is drastically easier.

I pushed a commit that encapsulates the two pooling strategies into a single base class. My same strategy could be applied to VP, but it would be much messier and touch far mode code due to the duplication.

Thanks for the sample code @FernetMenta, I think this is what I'm looking for.

garbear:
I wrote a new renderer based on the slideshow rendering code. It should work with OpenGLES. @MilhouseVH can you add this to your test builds?

MilhouseVH:
I can include it in x86 builds.

FernetMenta (changes requested):
please don't add deprecated OpenGL code (fixed pipeline)

garbear:
@MilhouseVH arm builds too?

@FernetMenta I've removed the OpenGL code that I got from the slideshow renderer. I'm confused, what parts of the pipeline were "fixed"? How is VP's renderer different?

MilhouseVH:
> @MilhouseVH arm builds too?

No, it conflicts with [newclock5](https://github.com/popcornmix/xbmc/commits/newclock5): http://sprunge.us/aCMD

Maybe @popcornmix can integrate this PR in newclock5.

MilhouseVH:
http://sprunge.us/SSHN

garbear:
@MilhouseVH try applying this PR to newclock5 now

MilhouseVH:
Thanks, this now builds for both x86.

For RPi (with newclock5), I had to revert https://github.com/xbmc/xbmc/pull/12715/commits/fd269e5804adcf3f2953f46d3126b075e5239adb and https://github.com/xbmc/xbmc/pull/12715/commits/940050cb1be27146d4e742830daca24e26e44899 before applying this PR on top of newclock5, and this now fails to build: http://sprunge.us/NGjJ

garbear:
I applied this PR to newclock5. There was only 1 line that needed to be changed to fix the build breakage you just got. I've pushed the rebased PR to: https://github.com/garbear/xbmc/commits/rp-renderer-pi. Shall I open a PR against newclock5, or can you just cherry-pick the commits?

MilhouseVH:
I'm not sure you need to open a PR against newclock5 as newclock5 already contains the changes you have in this PR ([1](https://github.com/popcornmix/xbmc/commit/f0ea860ccb05811fcef05150092f002883417a71#diff-73c84d39e60ce99125295665d6a923ee), [2](https://github.com/popcornmix/xbmc/commit/cc9b7c3769532d862cf73449efd83e445b40c35a#diff-73c84d39e60ce99125295665d6a923ee)), hence the two conflicts when applying this PR on top of newclock5.

I'm locally reverting the two commits from this PR that I mentioned previously in order to be able to apply this PR on top of newclock5, and that works for me. Assuming this PR goes in first then @popcornmix can just rebase the newclock5 branch as normal and drop his two commits.

I can cherry-pick https://github.com/garbear/xbmc/commit/af7aa576d4e8369e652971a5efce36d88fd45a92 for now. With it, this PR now builds for RPi (on top of newclock5), and x86 also builds so looking good. Not yet run-time tested, but will include in tonight's release if no obvious problems with video playback.

FernetMenta:
@garbear legacy OpenGL aka fixed function pipeline does not use shaders. That means that we have to migrate GUI rendering and picture rendering to profile 3.2. VP does already make use of shaders.

VelocityRa:
@FernetMenta 
>VP does already make use of shaders.

Yeah it does, but AFAIK using shaders doesn't mean that it's not fixed function. The OpenGL renderer (`CLinuxRendererGL`) is still fixed function right? [[1]](https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp#L563-L601), [[2]](https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp#L1026-L1055), etc. GLES isn't of course since it's not even available there.

FernetMenta:
@VelocityRa you are right, there are some parts still legacy. 
I am working on migration: https://github.com/FernetMenta/kodi-agile/commit/5a7d4d1a0a5164667c90a5581a7d40aa1faea54e
Then I will bump (OpenGL) shaders to >= 1.5

garbear:
I fixed the GLES renderer on the RPi. MMAL isn't ready yet, so it's been disabled. This PR is ready to go in now. @FernetMenta can you sign off on the two guilib changes?

xbmc/guilib/Texture.h
```diff
@@ -91,7 +91,7 @@ class CBaseTexture
 
   virtual void CreateTextureObject() = 0;
   virtual void DestroyTextureObject() = 0;
-  virtual void LoadToGPU() = 0;
+  virtual void LoadToGPU(bool bFreeSysMem = true) = 0;
```
FernetMenta: I don't like this default parameter hacks. From a design POV those have zero value. You only do this to avoid touching more code. Whenever I hit one of those hacks, I remove the default.

Why don't you add a new method for your purpose and factor out the common code into a separate protected method. Don't bother the rest of the app with a paramter is does not use.

garbear: Updated. The new commit adds a new method instead of using a default parameter in `LoadToGPU()`. How's that look?

garbear:
@FernetMenta sign off on the updated guilib commit?

garbear:
jenkins build this please

FernetMenta:
@garbear much better. thanks

garbear:
win32 error unrelated

PR #12723, opened 2017-08-27
FIX: [addons] clarify repo creation

koying:
This fixes the issue @MartijnKaijser had with his phone.
It's quite empiric, as it's not clear why it failed only in that situation, and not for the vast majority of others

1) Make it clear `SetLastChecked` creates the repo entry, and make it return the id.
The only reason I can see why this works and not the preveious code is some kind of race in `lastinsertid`. Placing it immediately after the `INSERT` seems to fix it

2) We were passing an int64 as idAddon to `PrepareSQL...%i`. That failed while debugging on win32 as well, so not sure how it works currently

### DIFF
--- a/xbmc/addons/AddonDatabase.cpp
+++ b/xbmc/addons/AddonDatabase.cpp
@@ -793,10 +793,9 @@ bool CAddonDatabase::UpdateRepositoryContent(const std::string& repository, cons
 
     DeleteRepository(repository);
 
-    if (!SetLastChecked(repository, version, CDateTime::GetCurrentDateTime().GetAsDBDateTime()))
+    int idRepo = SetLastChecked(repository, version, CDateTime::GetCurrentDateTime().GetAsDBDateTime());
+    if (idRepo < 0)
       return false;
-
-    int idRepo = static_cast<int>(m_pDS->lastinsertid());
     assert(idRepo > 0);
 
     m_pDB->start_transaction();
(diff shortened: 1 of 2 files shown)


tamland:
> The only reason I can see why this works and not the preveious code is some kind of race in lastinsertid. Placing it immediately after the INSERT seems to fix it

So it's not *really* fixed? Have you tried wrapping the lastinsertid call in a transaction?
Or just drop the whole thing and use a select to get the id instead?
Having the call to `SetLastChecked` inside `UpdateRepositoryContent` shouldn’t even be necessary; I don't know why it's there..

koying:
See the "empiric" part in the description ;)
From my understanding of sqlite3_last_insert_rowid, I don't think putting this specific call into a transaction would change anything, unless *all* calls are put in a transaction.

Then again, as it returns the last rowid whatever the thread, it probably won't fix a race issue.

> Having the call to SetLastChecked inside UpdateRepositoryContent shouldn’t even be necessary

Well, you surely know the code better than me. The repo is deleted, than recreated thru `SetLastChecked`, which is quite confusing, indeed ;)
As a matter of fact, the "UPDATE" part is never hit, currently.

Feel free to refactor at your own sauce, and ask @MartijnKaijser to test, if you so prefer...

tamland:
Yeah, that update content method is awful:\ Feel free to refactor you too. Or go with this. (Don't wait for me as I don't know when I'll have the time.)

koying:
Thanks.
We might envision thorough refactoring, but it's really not my part, and might even clash with @AlwinEsch addon work, so I'd merge this one as it is, and revisit later.

jenkins build this please

PR #12728, opened 2017-08-28
Enable shared cookie store

rbuehlma:
Since the removal of cookies.dat (http://github.com/xbmc/xbmc/pull/10611), the CURL implementation of Kodi does not preserve the cookies (over multiple curl sessions) any more which makes it impossible to use in addons where cookies are required (e.g. for identifying the session).

With this PR, I enable the share-interface of curl which allows sharing cookies over multiple curl sessions even in case of multithreading.

Thanks to @ksooo for his time to discuss the solution.

### DIFF
--- a/xbmc/filesystem/DllLibCurl.cpp
+++ b/xbmc/filesystem/DllLibCurl.cpp
@@ -63,6 +63,7 @@ using namespace XCURL;
 
 /* okey this is damn ugly. our dll loader doesn't allow for postload, preunload functions */
 static long g_curlReferences = 0;
+static CURLSH *g_curlShared = nullptr;
 #if(0)
 static unsigned int g_curlTimeout = 0;
 #endif
@@ -87,6 +88,28 @@ bool DllLibCurlGlobal::Load()
     CLog::Log(LOGERROR, "Error initializing libcurl");
     return false;
   }
+
+  g_curlShared = share_init();
+  CURLSHcode ret = share_setopt(g_curlShared, CURLSHOPT_LOCKFUNC, DllLibCurlGlobal::lock);
+  if (ret != CURLSHE_OK)
+  {
+    CLog::Log(LOGERROR, "Failed to set 'CURLSHOPT_LOCKFUNC': %s", share_strerror(ret));
+    return false;
+  }
+  ret = share_setopt(g_curlShared, CURLSHOPT_UNLOCKFUNC, DllLibCurlGlobal::unlock);
+  if (ret != CURLSHE_OK)
+  {
+    CLog::Log(LOGERROR, "Failed to set 'CURLSHOPT_UNLOCKFUNC': %s", share_strerror(ret));
+    return false;
+  }
+
+  ret = share_setopt(g_curlShared, CURLSHOPT_SHARE, CURL_LOCK_DATA_COOKIE);
+  if (ret != CURLSHE_OK)
+  {
+    CLog::Log(LOGERROR, "Failed to set 'CURLSHOPT_SHARE': %s", share_strerror(ret));
+    return false;
+  }
+
 
   /* check idle will clean up the last one */
   g_curlReferences = 2;
@@ -114,6 +137,8 @@ void DllLibCurlGlobal::Unload()
 
     // close libcurl
     global_cleanup();
+
+    share_cleanup(g_curlShared);
 
 #if defined(HAS_CURL_STATIC)
     // Cleanup ssl locking array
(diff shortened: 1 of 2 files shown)


ksooo:
@FernetMenta @kib do you have an opinion on this?

xbmc/filesystem/DllLibCurl.cpp
```diff
@@ -88,6 +89,25 @@ bool DllLibCurlGlobal::Load()
     return false;
   }
 
+  g_curlShared = share_init();
+  CURLSHcode ret = share_setopt(g_curlShared, CURLSHOPT_LOCKFUNC, DllLibCurlGlobal::lock);
+  if (ret != CURLSHE_OK) {
```
Rechi: curly bracket has to go into a separate line
same is true for all other places

xbmc/filesystem/DllLibCurl.cpp
```diff
@@ -174,6 +196,29 @@ void DllLibCurlGlobal::CheckIdle()
 #endif
 }
 
+CCriticalSection DllLibCurlGlobal::m_curlLock;
+
+void DllLibCurlGlobal::lock(CURL_HANDLE *handle, curl_lock_data data, curl_lock_access access, void *useptr ) {
+  (void)handle;
+  (void)access;
```
Rechi: why?

xbmc/filesystem/DllLibCurl.cpp
```diff
@@ -174,6 +196,29 @@ void DllLibCurlGlobal::CheckIdle()
...
+
+  m_curlLock.lock();
+}
+
+void DllLibCurlGlobal::unlock(CURL_HANDLE *handle, curl_lock_data data, void *useptr ) {
+  (void)handle;
+  (void)access;
```
Rechi: why?

Rechi (dismissed):
coding style had to be fixed, https://codedocs.xyz/xbmc/xbmc/code_guidelines.html#code_guidelines_2

xbmc/filesystem/DllLibCurl.cpp
rbuehlma: This was a C&P from a sample. Don't think it is required. -> Removed

rbuehlma:
Fixed code style

wsnipex:
jenkins build this please

da-anda:
I don't know how CURL and out implementation etc works, but might these shared cookies be a security issue? Like, could malicions add-ons read foreign add-on cookies with session IDs and hijack login sessions, or is there something like the "same origin" policiy in place, that browsers have?

rbuehlma:
Are you referring to XSS attacks? That would only work if you can trick CURL into opening a side-resource. As far as I know, CURL does not interpret the returned content, the addon does. So the "same origin" would need the be implemented in the addon from my point of view.

But yes, a malicious addon could probably use the session opened by a different addon but as they run as the same user and are not limited in their functionality, this is already possible now.

da-anda:
Not necessarily XSS, but a malicious add-on using my Amazon login session
to buy stuff via Amazon payments etc.. But if this already is an issue, no
objections to this PR I'm my end. Was just curious. Thanks.

Am 29.08.2017 10:55 vorm. schrieb "rbuehlma" <notifications@github.com>:

Are you referring to XSS attacks? That would only work if you can trick
CURL into opening a side-resource. As far as I know, CURL does not
interpret the returned content, the addon does. So the "same origin" would
need the be implemented in the addon from my point of view.

But yes, a malicious addon could probably use the session opened by a
different addon but as they run as the same user and are not limited in
their functionality, this is already possible now.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/xbmc/xbmc/pull/12728#issuecomment-325599934>, or mute
the thread

trummerjo:
@rbuehlma you should check this pull request: https://github.com/xbmc/xbmc/pull/12737
With this PR you should be able to get the cookie header from the requests, save it, and set the cookie header in further requests (not sure if kodi filters the cookie header in curl!)

rbuehlma:
Thanks for the info. Can we still merge this PR? In my opinion it is a bug if some requests reuse the cookies but not always.

wsnipex:
@arnova @FernetMenta  any comments?

FernetMenta:
shared cookies make no sense and can even interfere in a negative sense. curl is used by many different components and we don't want to share objects between them.
components that require to preserve data, need to do that by themselves

PR #12730, opened 2017-08-28
[PVR] epg grid container and model: cleanup and refactoring.

ksooo:
Just some non-functional cleanup.

@Jalle19 for review?


Jalle19 (commented):
I don't really understand these classes at all so it's up to you

PR #12731, opened 2017-08-28
 [depends] bump dbus to 1.11.16 (and actually build it)

Rechi:
## Description
build dbus for linux and bump it to version 1.11.16

## Motivation and Context
Dbus was never compiled by depends although it was added already 4 years ago.

## How Has This Been Tested?
compiled depends for linux

## Types of change

## Checklist:

### DIFF
--- a/tools/depends/target/Makefile
+++ b/tools/depends/target/Makefile
@@ -113,6 +113,7 @@ libdvdnav: libdvdread
 libdvdread: $(DVDREAD_DEPS)
 wayland: expat libffi
 waylandpp: wayland $(WAYLANDPP_DEPS)
+dbus: expat
 
 .installed-$(PLATFORM): $(DEPENDS)
 	touch $@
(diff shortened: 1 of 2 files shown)


wsnipex:
ok for me, but the question is, why do we need it?

Rechi:
don't know, but why is it in depends?
we can also remove it, but current state doesn't make sense (have it in depends, but never build it)

pkerling:
> ok for me, but the question is, why do we need it?

Jenkins currently builds with system DBus, which is probably not what is wanted?

wsnipex:
git history says: it was initially added with the whole depends system. Back then a whole lot didn't work. 
@pkerling right that is an argument, but it also shows we have a bug in our FindDBus module.
Looking at it, I wonder why....
>find_path(DBUS_ARCH_INCLUDE_DIR NAMES dbus/dbus-arch-deps.h
                                PATH_SUFFIXES dbus-1.0/include
                                PATHS ${PC_DBUS_LIBDIR}
                                      /usr/lib/${CMAKE_LIBRARY_ARCHITECTURE})

Rechi:
@wsnipex is this okay to get merged?
I can have a look at FindDBus.cmake if you tell me what you think is wrong, but would do that in a separate PR.

PR #12732, opened 2017-08-28
Default to not proxy localhost connections

GTechAlpha:
## Description

When using a proxy, localhost connections are directed through the proxy. This behavior prevents addons from using localhost and may affect other functionality. This PR changes the default behavior to not use proxy for locahost connections.

## Motivation and Context

Addons unable to use localhost when proxy has been set.

## How Has This Been Tested?

Functionality difference after change.

## Types of change

## Checklist:


wsnipex:
I'd much rather see a general no_proxy var that holds an exception list, with localhost being the default value.

GTechAlpha:
An exception list would be great but would preferably be accomplished using a mutable list in settings and the current API does not provide that capability - unless I have overlooked it. In the meantime, this solves a problem and could be improved later.

tamland:
why would anyone ever *enable* this setting?

GTechAlpha:
It seems very unlikely anybody would enable the setting so I removed it. I added the NO_PROXY env var for Python, etc.

PR #12735, opened 2017-08-28
[input] allow set of unicode in CAction

AlwinEsch:
## Description

This take also the unicode number of other CKey to construct.

This is required to get right character on addons in other countries.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/input/Action.cpp
+++ b/xbmc/input/Action.cpp
@@ -75,7 +75,7 @@ CAction::CAction(int actionID, const std::string &name, const CKey &key):
     m_amount[i] = 0;
   m_repeat = key.GetRepeat();
   m_buttonCode = key.GetButtonCode();
-  m_unicode = 0;
+  m_unicode = key.GetUnicode();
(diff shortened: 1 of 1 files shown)


xbmc/input/Action.cpp
```diff
@@ -75,7 +75,7 @@ CAction::CAction(int actionID, const std::string &name, const CKey &key):
     m_amount[i] = 0;
   m_repeat = key.GetRepeat();
   m_buttonCode = key.GetButtonCode();
-  m_unicode = 0;
+  m_unicode = key.GetUnicode();
```
AlwinEsch: @garbear was there a special reason why unicode was set to 0?

garbear: This code was c/p from the original code written 10 years ago. I suspect it's been wrong this whole time.

AlwinEsch: Then better bring this in? ;-)

garbear: Do it

PR #12737, opened 2017-08-29
Implement File::GetProperty to allow (binary) addons access to e.g. reponse headers

peak3d:
## Description
Provide Interface to get CURL Response Headers / CURL return code

## Motivation and Context
For manifest updates (Live content) I need the ETag Header field for requesting only new manifest data if content is changed.  If content is not changed (regarding ETag) I have to evaluate the return code.

Beside this Cookie header values are needed by zattoo pve addon.

## How Has This Been Tested?
Regression test using Win10 and several internet / local streams

## Types of change

### DIFF
--- a/xbmc/addons/interfaces/Filesystem.cpp
+++ b/xbmc/addons/interfaces/Filesystem.cpp
@@ -507,7 +507,7 @@ int Interface_Filesystem::get_file_chunk_size(void* kodiBase, void* file)
 const char* Interface_Filesystem::get_property(void* kodiBase, void* file, int type, const char *name)
 {
   CAddonDll* addon = static_cast<CAddonDll*>(kodiBase);
-  if (addon == nullptr || file == nullptr)
+  if (addon == nullptr || file == nullptr || name == nullptr)
   {
     CLog::Log(LOGERROR, "Interface_Filesystem::%s - invalid data (addon='%p', file='%p')", __FUNCTION__, addon, file);
     return "";
--- a/xbmc/addons/interfaces/Filesystem.cpp
+++ b/xbmc/addons/interfaces/Filesystem.cpp
@@ -504,13 +504,13 @@ int Interface_Filesystem::get_file_chunk_size(void* kodiBase, void* file)
   return static_cast<CFile*>(file)->GetChunkSize();
 }
 
-const char* Interface_Filesystem::get_property(void* kodiBase, void* file, int type, const char *name)
+char* Interface_Filesystem::get_property(void* kodiBase, void* file, int type, const char *name)
 {
   CAddonDll* addon = static_cast<CAddonDll*>(kodiBase);
   if (addon == nullptr || file == nullptr || name == nullptr)
   {
     CLog::Log(LOGERROR, "Interface_Filesystem::%s - invalid data (addon='%p', file='%p')", __FUNCTION__, addon, file);
-    return "";
+    return nullptr;
   }
 
   XFILE::FileProperty internalType;
@@ -533,9 +533,9 @@ const char* Interface_Filesystem::get_property(void* kodiBase, void* file, int t
     break;
   default:
     CLog::Log(LOGERROR, "Interface_Filesystem::%s - invalid data (addon='%p', file='%p')", __FUNCTION__, addon, file);
-    return "";
+    return nullptr;
   };
-  return static_cast<CFile*>(file)->GetProperty(internalType, name).c_str();
+  return strdup(static_cast<CFile*>(file)->GetProperty(internalType, name).c_str());
 }
 
 void* Interface_Filesystem::curl_create(void* kodiBase, const char* url)
--- a/xbmc/addons/interfaces/Filesystem.cpp
+++ b/xbmc/addons/interfaces/Filesystem.cpp
@@ -73,6 +73,7 @@ void Interface_Filesystem::Init(AddonGlobalInterface* addonInterface)
   addonInterface->toKodi->kodi_filesystem->get_file_download_speed = get_file_download_speed;
   addonInterface->toKodi->kodi_filesystem->close_file = close_file;
   addonInterface->toKodi->kodi_filesystem->get_file_chunk_size = get_file_chunk_size;
+  addonInterface->toKodi->kodi_filesystem->get_property = get_property;
 
   addonInterface->toKodi->kodi_filesystem->curl_create = curl_create;
   addonInterface->toKodi->kodi_filesystem->curl_add_option = curl_add_option;
@@ -501,6 +502,40 @@ int Interface_Filesystem::get_file_chunk_size(void* kodiBase, void* file)
   }
 
   return static_cast<CFile*>(file)->GetChunkSize();
+}
+
+const char* Interface_Filesystem::get_property(void* kodiBase, void* file, int type, const char *name)
+{
+  CAddonDll* addon = static_cast<CAddonDll*>(kodiBase);
+  if (addon == nullptr || file == nullptr)
+  {
+    CLog::Log(LOGERROR, "Interface_Filesystem::%s - invalid data (addon='%p', file='%p')", __FUNCTION__, addon, file);
+    return "";
+  }
+
+  XFILE::FileProperty internalType;
+  switch (type)
+  {
+  case ADDON_FILE_PROPERTY_RESPONSE_PROTOCOL:
+    internalType = XFILE::FILE_PROPERTY_RESPONSE_PROTOCOL;
+    break;
+  case ADDON_FILE_PROPERTY_RESPONSE_HEADER:
+    internalType = XFILE::FILE_PROPERTY_RESPONSE_HEADER;
+    break;
+  case ADDON_FILE_PROPERTY_CONTENT_TYPE:
+    internalType = XFILE::FILE_PROPERTY_CONTENT_TYPE;
+    break;
+  case ADDON_FILE_PROPERTY_CONTENT_CHARSET:
+    internalType = XFILE::FILE_PROPERTY_CONTENT_CHARSET;
+    break;
+  case ADDON_FILE_PROPERTY_MIME_TYPE:
+    internalType = XFILE::FILE_PROPERTY_MIME_TYPE;
+    break;
+  default:
+    CLog::Log(LOGERROR, "Interface_Filesystem::%s - invalid data (addon='%p', file='%p')", __FUNCTION__, addon, file);
+    return "";
+  };
+  return static_cast<CFile*>(file)->GetProperty(internalType, name).c_str();
 }
 
 void* Interface_Filesystem::curl_create(void* kodiBase, const char* url)
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
@@ -110,7 +110,7 @@ extern "C"
     double (*get_file_download_speed)(void* kodiBase, void* file);
     void (*close_file)(void* kodiBase, void* file);
     int (*get_file_chunk_size)(void* kodiBase, void* file);
-    const char* (*get_property)(void* kodiBase, void* file, int type, const char *name);
+    char* (*get_property)(void* kodiBase, void* file, int type, const char *name);
 
     void* (*curl_create)(void* kodiBase, const char* url);
     bool (*curl_add_option)(void* kodiBase, void* file, int type, const char* name, const char* value);
@@ -1547,7 +1547,15 @@ namespace vfs
         kodi::Log(ADDON_LOG_ERROR, "kodi::vfs::CURLCreate(...) needed to call before GetProperty!");
         return "";
       }
-      return ::kodi::addon::CAddonBase::m_interface->toKodi->kodi_filesystem->get_property(::kodi::addon::CAddonBase::m_interface->toKodi->kodiBase, m_file, type, name.c_str());
+      char *res(::kodi::addon::CAddonBase::m_interface->toKodi->kodi_filesystem->get_property(
+        ::kodi::addon::CAddonBase::m_interface->toKodi->kodiBase, m_file, type, name.c_str()));
+      if (res)
+      {
+        std::string strReturn(res);
+        ::kodi::addon::CAddonBase::m_interface->toKodi->free_string(::kodi::addon::CAddonBase::m_interface->toKodi->kodiBase, res);
+        return strReturn;
+      }
+      return "";
     }
     //--------------------------------------------------------------------------
 
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
@@ -110,6 +110,7 @@ extern "C"
     double (*get_file_download_speed)(void* kodiBase, void* file);
     void (*close_file)(void* kodiBase, void* file);
     int (*get_file_chunk_size)(void* kodiBase, void* file);
+    const char* (*get_property)(void* kodiBase, void* file, int type, const char *name);
 
     void* (*curl_create)(void* kodiBase, const char* url);
     bool (*curl_add_option)(void* kodiBase, void* file, int type, const char* name, const char* value);
@@ -192,6 +193,25 @@ typedef enum CURLOptiontype
   /// Add a Header
   ADDON_CURL_OPTION_HEADER
 } CURLOptiontype;
+//------------------------------------------------------------------------------
+
+//==============================================================================
+/// \ingroup cpp_kodi_vfs_Defs
+/// @brief Used CURL message types
+///
+typedef enum FilePropertyTypes
+{
+  /// Get protocol response line
+  ADDON_FILE_PROPERTY_RESPONSE_PROTOCOL,
+  /// Get a response header
+  ADDON_FILE_PROPERTY_RESPONSE_HEADER,
+  /// Get file content type
+  ADDON_FILE_PROPERTY_CONTENT_TYPE,
+  /// Get file content charset
+  ADDON_FILE_PROPERTY_CONTENT_CHARSET,
+  /// Get file mime type
+  ADDON_FILE_PROPERTY_MIME_TYPE
+} FilePropertyTypes;
 //------------------------------------------------------------------------------
 
 //============================================================================
@@ -1508,6 +1528,26 @@ namespace vfs
       if (!m_file)
         return -1;
       return ::kodi::addon::CAddonBase::m_interface->toKodi->kodi_filesystem->get_file_chunk_size(::kodi::addon::CAddonBase::m_interface->toKodi->kodiBase, m_file);
+    }
+    //--------------------------------------------------------------------------
+
+    //==========================================================================
+    ///
+    /// @ingroup cpp_kodi_vfs_CFile
+    /// @brief retrieve a file property
+    ///
+    /// @param[in] type         The type of the file property to retrieve the value for
+    /// @param[in] name         The name of a named property value (e.g. Header)
+    /// @return                 value of requested property, empty on failure / non-existance
+    ///
+    const std::string GetProperty(FilePropertyTypes type, const std::string &name) const
+    {
+      if (!m_file)
+      {
+        kodi::Log(ADDON_LOG_ERROR, "kodi::vfs::CURLCreate(...) needed to call before GetProperty!");
+        return "";
+      }
+      return ::kodi::addon::CAddonBase::m_interface->toKodi->kodi_filesystem->get_property(::kodi::addon::CAddonBase::m_interface->toKodi->kodiBase, m_file, type, name.c_str());
     }
     //--------------------------------------------------------------------------
 
--- a/xbmc/filesystem/CurlFile.cpp
+++ b/xbmc/filesystem/CurlFile.cpp
@@ -1925,7 +1925,7 @@ const std::string CCurlFile::GetProperty(XFILE::FileProperty type, const std::st
   case FILE_PROPERTY_MIME_TYPE:
     return m_state->m_httpheader.GetMimeType();
   default:
-    return StringUtils::Empty;
+    return "";
   }
 }
 
--- a/xbmc/filesystem/CurlFile.cpp
+++ b/xbmc/filesystem/CurlFile.cpp
@@ -1766,14 +1766,6 @@ void CCurlFile::SetRequestHeader(const std::string& header, long value)
   m_requestheaders[header] = StringUtils::Format("%ld", value);
 }
 
-std::string CCurlFile::GetServerReportedCharset(void)
-{
-  if (!m_state)
-    return "";
-
-  return m_state->m_httpheader.GetCharset();
-}
-
 std::string CCurlFile::GetURL(void)
 {
   return m_url;
@@ -1812,7 +1804,7 @@ bool CCurlFile::GetMimeType(const CURL &url, std::string &content, const std::st
     if (buffer.st_mode == _S_IFDIR)
       content = "x-directory/normal";
     else
-      content = file.GetMimeType();
+      content = file.GetProperty(XFILE::FILE_PROPERTY_MIME_TYPE);
     CLog::Log(LOGDEBUG, "CCurlFile::GetMimeType - %s -> %s", redactUrl.c_str(), content.c_str());
     return true;
   }
@@ -1834,7 +1826,7 @@ bool CCurlFile::GetContentType(const CURL &url, std::string &content, const std:
     if (buffer.st_mode == _S_IFDIR)
       content = "x-directory/normal";
     else
-      content = file.GetContent();
+      content = file.GetProperty(XFILE::FILE_PROPERTY_CONTENT_TYPE, "");
     CLog::Log(LOGDEBUG, "CCurlFile::GetContentType - %s -> %s", redactUrl.c_str(), content.c_str());
     return true;
   }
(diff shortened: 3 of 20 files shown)


xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
```diff
@@ -1266,6 +1266,44 @@ namespace vfs
...
+    ///
+    /// @param[in] name         The name of header to retrieve the value for
+    /// @return                 value of requested header, empty string if header name is not found
+    ///
+    std::string CURLGetHeader(const std::string &name)
+    {
+      if (!m_file)
+      {
+        kodi::Log(ADDON_LOG_ERROR, "kodi::vfs::CURLCreate(...) needed to call before!");
+        return false;
```
ksooo: return ““;

xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
```diff
@@ -1266,6 +1266,44 @@ namespace vfs
...
+    /// @brief retrieve a http return code
+    ///
+    /// @return      http return code (e.g. 200)
+    ///
+    int CURLGetReturnCode()
+    {
+      if (!m_file)
+      {
+        kodi::Log(ADDON_LOG_ERROR, "kodi::vfs::CURLCreate(...) needed to call before!");
+        return false;
```
ksooo: return -1; or whatever you want to return in error case. And please don't forget to document that special  value.

xbmc/addons/interfaces/Filesystem.cpp
```diff
@@ -564,5 +564,29 @@ bool Interface_Filesystem::curl_open(void* kodiBase, void* file, unsigned int fl
...
+}
+
+
+int Interface_Filesystem::curl_get_return_code(void* kodiBase, void* file)
+{
+  CAddonDll* addon = static_cast<CAddonDll*>(kodiBase);
+  if (addon == nullptr || file == nullptr)
+  {
+    CLog::Log(LOGERROR, "Interface_Filesystem::%s - invalid data (addon='%p', file='%p')", __FUNCTION__, addon, file);
+    return false;
```
ksooo: return -1; ?

xbmc/addons/interfaces/Filesystem.cpp
```diff
@@ -564,5 +564,29 @@ bool Interface_Filesystem::curl_open(void* kodiBase, void* file, unsigned int fl
   return static_cast<CFile*>(file)->CURLOpen(flags);
 }
 
+const char* Interface_Filesystem::curl_get_header(void* kodiBase, void* file, const char *name)
+{
+  CAddonDll* addon = static_cast<CAddonDll*>(kodiBase);
+  if (addon == nullptr || file == nullptr)
+  {
+    CLog::Log(LOGERROR, "Interface_Filesystem::%s - invalid data (addon='%p', file='%p')", __FUNCTION__, addon, file);
+    return false;
```
ksooo: return NULL;.  ?

xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
ksooo: in car the c api returns NULL, you need to check for this here. IIRC, std::string has no ctor taking a nullptr

ksooo (changes requested):
some copy/paste errors with return vales....

xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
```diff
@@ -1266,6 +1266,44 @@ namespace vfs
...
+    /// @return                 value of requested header, empty string if header name is not found
+    ///
+    std::string CURLGetHeader(const std::string &name)
+    {
+      if (!m_file)
+      {
+        kodi::Log(ADDON_LOG_ERROR, "kodi::vfs::CURLCreate(...) needed to call before!");
+        return false;
+      }
+      return ::kodi::addon::CAddonBase::m_interface->toKodi->kodi_filesystem->curl_get_header(::kodi::addon::CAddonBase::m_interface->toKodi->kodiBase, m_file, name);
```
ksooo: in case the c api returns NULL, you need to check for this here. IIRC, std::string has no ctor taking a nullptr

ksooo:
Not sure wheter we should include "response" in the respective function named to make clear that we are dealing with response headers here (not with request headers)?

peak3d:
Thx @ksooo for the fast feedback, yes, sure those places are wrong.

There is one part in this PR wich I want to be sure that it is agreed by you / @AlwinEsch:

Using CURL (CFile::m_curl) as a container for the response headers / response code.

I have seen 2 possibilities, Implementing GetHeader / GetReturnCode in IFile and overwrite it CCurlFile or the selected way: storing the return values in CURL.

The second way goes more in the direction CURL_SetOption which is already using CURL as the container to xchange the values. Beside this I don't want to mess up IFile and the other implementations with methods wich will never used there (e.g. standard files).

Pls. evaluate especially this part and let me know if it's Ok for you. Thx.

peak3d:
> Not sure wheter we should include "response" in the respective function named to make clear that we are dealing with response headers here (not with request headers)?

Yes, sounds reasonable

xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
```diff
@@ -1266,6 +1266,44 @@ namespace vfs
...
+    //--------------------------------------------------------------------------
+
+    //==========================================================================
+    ///
+    /// @ingroup cpp_kodi_vfs_CFile
+    /// @brief retrieve a http return code
+    ///
+    /// @return      http return code (e.g. 200)
+    ///
+    int CURLGetReturnCode()
```
ksooo: Http return code? You mean "status code", right? I would name it like this.

ksooo (commented):
I suggest to make the naming  consistent - return code vs. status code, response header vs. response value etc.

peak3d:
If we say that CURL is mainly thought for HTTP files I agree with changing namng conventions.
But if I'm not wrong CURL is a container for all types of files so I think it make sense to not use HTTP specific namings in CURL. Maybe in future someone will pass some return k/v pairs around for non HTTP responses and it would be sad if he has to create a new member because of naming conventions.

xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
peak3d: Sure, this should be changed

ksooo:
> But if I'm not wrong CURL is a container for all types of files so I think it make sense to not use HTTP specific namings in CURL.

 Very valid point.

xbmc/URL.h
```diff
@@ -190,6 +191,16 @@ class CURL
   void SetProtocolOption(const std::string &key, const std::string &value);
   void RemoveProtocolOption(const std::string &key);
 
+  const std::string &GetResponseValue(const std::string name)
```
ksooo: This method could (and should) be const.

xbmc/URL.h
```diff
@@ -190,6 +191,16 @@ class CURL
...
+  {
+    return m_responseValues[name];
+  }
+
+  int GetResponseCode()
```
ksooo: This method could (and should) be const.

xbmc/URL.h
```diff
@@ -204,4 +215,8 @@ class CURL
   std::string m_strProtocolOptions;
   CUrlOptions m_options;
   CUrlOptions m_protocolOptions;
+
+  //Response data
+  std::map<std::string, std::string> m_responseValues;
+  int m_responseCode = -1;
```
ksooo: Magic numbers are evil. I would introduce a constant with a speaking name for "invalid response code".

xbmc/addons/interfaces/Filesystem.cpp
```diff
@@ -564,5 +566,29 @@ bool Interface_Filesystem::curl_open(void* kodiBase, void* file, unsigned int fl
...
+}
+
+
+int Interface_Filesystem::curl_get_response_status_code(void* kodiBase, void* file)
+{
+  CAddonDll* addon = static_cast<CAddonDll*>(kodiBase);
+  if (addon == nullptr || file == nullptr)
+  {
+    CLog::Log(LOGERROR, "Interface_Filesystem::%s - invalid data (addon='%p', file='%p')", __FUNCTION__, addon, file);
+    return -1;
```
ksooo: Magic numbers "everywhere"... ;-)

xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
```diff
@@ -1266,6 +1268,44 @@ namespace vfs
...
+    /// @brief retrieve the response status code
+    ///
+    /// @return      response status code (e.g. 200 for HTTP requests)
+    ///
+    int CURLGetResponseStatusCode()
+    {
+      if (!m_file)
+      {
+        kodi::Log(ADDON_LOG_ERROR, "kodi::vfs::CURLCreate(...) needed to call before CURLGetResponseStatusCode!");
+        return -1;
```
ksooo: Magic numbers all over the place... As this is official API, we should be very clean here..

xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
```diff
@@ -1266,6 +1268,44 @@ namespace vfs
...
+    /// @return                 value of requested response header, empty on failure
+    ///
+    std::string CURLGetResponseHeader(const std::string &name)
+    {
+      if (!m_file)
+      {
+        kodi::Log(ADDON_LOG_ERROR, "kodi::vfs::CURLCreate(...) needed to call before CURLGetResponseHeader!");
+        return "";
+      }
+      return ::kodi::addon::CAddonBase::m_interface->toKodi->kodi_filesystem->curl_get_response_header(::kodi::addon::CAddonBase::m_interface->toKodi->kodiBase, m_file, name.c_str());
```
ksooo: If `Interface_Filesystem::curl_get_response_header`can return a nullptr, you should not pass this nullptr to the returned std::string. IIRC, std::string has no ctor handling nullptr correctly. According to c++11 standard, behavior is undefined in this case.

xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
```diff
@@ -1266,6 +1268,44 @@ namespace vfs
...
+    }
+    //--------------------------------------------------------------------------
+
+    //==========================================================================
+    ///
+    /// @ingroup cpp_kodi_vfs_CFile
+    /// @brief retrieve the response status code
+    ///
+    /// @return      response status code (e.g. 200 for HTTP requests)
```
ksooo: Please document what is returned in error case(s). No magic number, please. ;-)

xbmc/filesystem/File.cpp
```diff
@@ -254,6 +254,16 @@ bool CFile::CURLOpen(unsigned int flags)
   return Open(m_curl, flags);
 }
 
+const std::string &CFile::CURLGetResponseHeader(const std::string &name)
```
ksooo: This method could (and should) be const.

xbmc/filesystem/File.cpp
```diff
@@ -254,6 +254,16 @@ bool CFile::CURLOpen(unsigned int flags)
...
+{
+  return m_curl.GetResponseValue(name);
+}
+
+int CFile::CURLGetResponseStatusCode()
```
ksooo: This method could (and should) be const.

ksooo (commented):
Yeah, I'm a nitpicker when it comes to official APIs.

ksooo:
> There is one part in this PR wich I want to be sure that it is agreed by you / @AlwinEsch:
> Using CURL (CFile::m_curl) as a container for the response headers / response code.

 I do not like this approach as I cannot see how/why response headers are related to URLs. They are completely unrelated at that level of abstraction, imo. Also, you PR currently completely misses code for filling the container, right? I thinks this is by intention because this is WIP?

 I do also not like the other approach, using CFile as a container for the response headers, tbh. Response headers and status can only be bound to a certain request, not to a file. I can do many request on one file. So, what is the actual semantics of GetResponseHeader on a file? Is it actually "get the reponse header of the **last** request, what ever this was (because I have no possibility to obtain the 'question' for the given 'answer')"? Happy multithreading, then. Ugly. Not really working.

peak3d:
@ksooo there are 2 "I do not like" statements in your comment, but no single "I would like".
Can you pls. add the missing part? :-)

ksooo:
Well, somtimes one has no spontanious "I like" in mind, which should be no nogo for writing the "I don not likes". Honestly,

ksooo:
I would like an API like this (pseude code and laguage): 
`file = open-file, request = file->create-request, response = request->execute, header = response->getheader(x), file->close`.

But that might not work out well for a generic API covering several communication protocols.

ksooo:
Maybe the exiting API is just fine, don't know: `curl_create, curl_open (sends the http request) + curl_get_return_value (obtains the header value). Maybe a new curl_create is to be done for every http request anyway? Then, all would be fine as it is intened with this PR... and I would have learned something. Just wanted to raise my concerns, based on what I know.

peak3d:
Yes, CURL_XXX API is intended to do exactly what your pseudo code descripes.
Sure you can reuse a once CURL_CREATE created handle to make subsequent requests, but if you do, then you should know that response is always based on last "execute (==open) call.

Nevertheless I mislike the response in CCurl as well so I'll change it to a more abstract methode in CFile wich gets overriden from CCurlFile.

ksooo:
> then you should know that response is always based on last "execute (==open) call.

 And that's the bad part about it. Reminds me of global "getlasterror" shit. ;-) But if this fits our needs I will shut up now.

peak3d:
I hate GetLastError() as well, but we have a difference here: The file reference.
So its completely safe because you call it "on" the fileobject wich you have under control.

ksooo:
Still not convinced, but don't want to discuss this to detah. If it's just me and does what we need, go ahead.

AlwinEsch:
Mostly is it Ok for me.

peak3d:
Pls. wait for further looking, there will be a new commit later

xbmc/addons/interfaces/Filesystem.cpp
```diff
@@ -503,6 +504,40 @@ int Interface_Filesystem::get_file_chunk_size(void* kodiBase, void* file)
   return static_cast<CFile*>(file)->GetChunkSize();
 }
 
+const char* Interface_Filesystem::get_property(void* kodiBase, void* file, int type, const char *name)
+{
+  CAddonDll* addon = static_cast<CAddonDll*>(kodiBase);
+  if (addon == nullptr || file == nullptr)
```
ace20022: Can name be null to?

peak3d: No, because the Interface class methods all work with std::string
const char* is only used for the step dll <-> kodi

ksooo: I would add the check nevertheless. It does not cost to much computing power and makes the code more failsafe.

xbmc/addons/interfaces/Filesystem.cpp
```diff
@@ -503,6 +504,40 @@ int Interface_Filesystem::get_file_chunk_size(void* kodiBase, void* file)
...
+    internalType = XFILE::FILE_PROPERTY_CONTENT_CHARSET;
+    break;
+  case ADDON_FILE_PROPERTY_MIME_TYPE:
+    internalType = XFILE::FILE_PROPERTY_MIME_TYPE;
+    break;
+  default:
+    CLog::Log(LOGERROR, "Interface_Filesystem::%s - invalid data (addon='%p', file='%p')", __FUNCTION__, addon, file);
+    return "";
+  };
+  return static_cast<CFile*>(file)->GetProperty(internalType, name).c_str();
```
ksooo: This will most probably crash if a std::string for param 2 gets constructed and `name` is `nullptr`.

xbmc/filesystem/CurlFile.cpp
```diff
@@ -1918,6 +1910,25 @@ int CCurlFile::IoControl(EIoControl request, void* param)
...
+  case FILE_PROPERTY_RESPONSE_HEADER:
+    return m_state->m_httpheader.GetValue(name);
+  case FILE_PROPERTY_CONTENT_TYPE:
+    return m_state->m_httpheader.GetValue("content-type");
+  case FILE_PROPERTY_CONTENT_CHARSET:
+    return m_state->m_httpheader.GetCharset();
+  case FILE_PROPERTY_MIME_TYPE:
+    return m_state->m_httpheader.GetMimeType();
+  default:
+    return StringUtils::Empty;
```
ksooo: Why not simply `return "";`? `StringUtils::Empty` is mainly a trick if you have to return a reference to a string instead of an  instance. As this method is not called that often, I also cannot see any performance gains with `StringUtils::Empty`.

xbmc/filesystem/CurlFile.h
```diff
@@ -61,10 +61,8 @@ namespace XFILE
       bool ReadString(char *szLine, int iLineLength) override { return m_state->ReadString(szLine, iLineLength); }
       ssize_t Read(void* lpBuf, size_t uiBufSize) override { return m_state->Read(lpBuf, uiBufSize); }
       ssize_t Write(const void* lpBuf, size_t uiBufSize) override;
-      virtual std::string GetMimeType() { return m_state->m_httpheader.GetMimeType(); }
-      std::string GetContent() override { return m_state->m_httpheader.GetValue("content-type"); }
+      virtual const std::string GetProperty(XFILE::FileProperty type, const std::string &name = StringUtils::Empty) const override;
```
ksooo: `const std::string &name = ""`? Simpler to read and no new (!) dependency of this header file to "StringUtils.h" (which is currently missing, btw, if you want to stick with `const std::string &name = StringUtils::Empty`).

ksooo: ... and no need for `virtual` if you add `override`. I would remove the superflous `virtual`.

xbmc/filesystem/File.h
```diff
@@ -109,8 +110,7 @@ class CFile
   int64_t GetLength();
   void Close();
   int GetChunkSize();
-  std::string GetContentMimeType(void);
-  std::string GetContentCharset(void);
+  const std::string GetProperty(XFILE::FileProperty type, const std::string &name = StringUtils::Empty) const;
```
ksooo: See my other comment re. `StringUtils::Empty`. I would use `""` - also saves you the new dependency to `StringUtils.h`.

ksooo: ... and shouldn't you add `override` here?

xbmc/filesystem/FileCache.h
```diff
@@ -59,8 +59,7 @@ namespace XFILE
 
     IFile *GetFileImp();
 
-    std::string GetContent() override;
-    std::string GetContentCharset(void) override;
+    virtual const std::string GetProperty(XFILE::FileProperty type, const std::string &name = StringUtils::Empty) const override;
```
ksooo: No need for `virtual` here. No need for `StringUtils::Empty`. See my other comments.

xbmc/filesystem/IFile.h
```diff
@@ -129,8 +130,10 @@ class IFile
 
   virtual int IoControl(EIoControl request, void* param) { return -1; }
 
-  virtual std::string GetContent()                           { return "application/octet-stream"; }
-  virtual std::string GetContentCharset(void)                { return ""; }
+  virtual const std::string GetProperty(XFILE::FileProperty type, const std::string &name = StringUtils::Empty) const
+  {
+    return type == XFILE::FILE_PROPERTY_CONTENT_TYPE ? XFILE::DefaultContentType : StringUtils::Empty;
```
ksooo: No need for StringUtils::Empty. See my other comments.

xbmc/filesystem/IFileTypes.h
```diff
@@ -83,4 +84,15 @@ enum CURLOPTIONTYPE
...
+enum FileProperty
+{
+  FILE_PROPERTY_RESPONSE_PROTOCOL,          /**< Get response protocol line  */
+  FILE_PROPERTY_RESPONSE_HEADER,            /**< Get response Header value  */
+  FILE_PROPERTY_CONTENT_TYPE,               /**< Get file content-type  */
+  FILE_PROPERTY_CONTENT_CHARSET,            /**< Get file content charset  */
+  FILE_PROPERTY_MIME_TYPE                   /**< Get file mime type  */
+};
+
+/* default content-type for use in returns by ref */
```
ksooo: Always a good idea to define constants, but I woul remove the part re return by ref. It is a good thing also if not returning refs. And actually your code uses the constant only in a scenario where it is returned by value. ;-)

ksooo (changes requested):
I like the new approach. My comments are mostly nitpicking.

What is definitely missing is a bump of the filesystem API version! => https://github.com/xbmc/xbmc/blob/master/xbmc/addons/kodi-addon-dev-kit/include/kodi/versions.h#L69

xbmc/filesystem/File.h
peak3d: No override, its not derived from anything.

xbmc/filesystem/FileCache.h
peak3d: The removal of virtual is a matter of how peaple read foreign code. My eyes for example scan normally the beginning of member definitions, It could help some future devs scanning like me to get things faster together.

But, I don't want to spend much more time on this PR, so its. changed (PR will come soon)

ksooo: Sure, it's personal style. That's why I phrased this changes in almost all other places I wrote about this using "I would". ;-)

xbmc/filesystem/File.h
ksooo: Okay, overlooked that CFile is not derived from IFile. :-/

ksooo (approved):
Thanks says "die Nervensäge". ;-)

ksooo:
Oh well, please squash to a single commit before merge. :-)

peak3d:
jenkins build this please

peak3d:
jenkins build this please

peak3d:
jenkins build this please

peak3d:
jenkins build this please

ksooo:
> Oh well, please squash to a single commit before merge. :-)

 Usually we clean up our mess before merging. Makes it much easier to reproduce changes. :-(

Rechi:
@ksooo this was merged with githubs `squash and merge` feature

rbuehlma:
Thanks for this PR. But how is this meant to be used?

After I did a
`void* file = XBMC->CURLCreate(url.c_str());`
do I need to cast file into a CFile and then call GetProperty on it or what is the interface against a PVR-Addon?

ksooo:
> @ksooo this was merged with githubs squash and merge feature

 Argh.I should have known that @peak3d will not ignore our rules. Sorry, Marcus.

peak3d:
np

ksooo:
> do I need to cast file into a CFile and then call GetProperty

@rbuehlma no. I guess it's `XBMC->GetProperty(...)`.

peak3d:
@rbuehlma https://github.com/peak3d/inputstream.adaptive/blob/master/src/main.cpp#L247 is an example

ksooo:
>  @rbuehlma https://github.com/peak3d/inputstream.adaptive/blob/master/src/main.cpp#L247 is an example

 This is how it is done wit K18's new C++ API warapper, yes. `XBMC->` is the old, still working way, if I remember correctly. @AlwinEsch is this correct?

rbuehlma:
I now only changed the file handling to the "new style". As soon as I start calling a function from CFile I get this error while linking:
`CMakeFiles/pvr.zattoo.dir/src/Curl.cpp.o: In function `kodi::Log(AddonLog, char const*, ...)':
/media/space/zattoo/17/pvr.zattoo/build/build/depends/include/kodi/AddonBase.h:437: undefined reference to `kodi::addon::CAddonBase::m_interface'
/usr/bin/ld: CMakeFiles/pvr.zattoo.dir/src/Curl.cpp.o: relocation R_X86_64_PC32 against undefined hidden symbol `_ZN4kodi5addon10CAddonBase11m_interfaceE' can not be used when making a shared object
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
`
Do you have a clue what is going on here?

ksooo:
> I now only changed the file handling to the "new style".

 I'm sure whether this has to be "all or nothing approach" to work. @AlwinEsch knows.

peak3d:
@rbuehlma is your dev tree accessible anywhere in github?

ksooo:
There is a 17 in the oath of your error log. Ate you trying to mix 18 add-on devkit with 17 add-on sources? Won't work - at least not for new style c++ wrapper.

rbuehlma:
The 17 in the path is from earlier times where I had a workspace for every version. Now I'm switch the version using checkout. The code is from master.

@peak3d here is the code: https://github.com/rbuehlma/pvr.zattoo/tree/get_property but its not directly related to GetProperty as I can comment that line out and the problem still happens. It somehow seems to be related to the switch to C++ API. HEAD~1 builds without errors.

ksooo:
> It somehow seems to be related to the switch to C++ API

 Yeah, it definitely is related o this. I asked @AlwinEsch to support you here. He should know what's wrong.

peak3d:
IIRC the interface has to be updated to the new version because of pvr / inputstream dependencies in the future... So best way would be tu update the interface completely

rbuehlma:
I tried to port it to the c++ api but how do I call PVR functions there? Is there already a PVR addon using the C++ API? A Sample for PVR->TriggerTimerUpdate(); might already help :)

ksooo:
IIRC, providing the new c++ api wrapper for pvr is still on the to-do list of @AlwinEsch. I hope he will show up here soon too comment. Until then we're just doing wild/educated guessing.

peak3d:
Ok, if there is no response until noon, I'll add GetProperty into the "old" interface

rbuehlma:
Ok, great. There is no hurry for it.

Thanks for your help.

AlwinEsch:
Hi @rbuehlma yes PVR still use the old system and has no access to the new. If @peak3d can add them to the old interface can it be help for the Moment.

To change PVR to new Interface style is mostly done, but before must be the addon handling for parts with install and uninstall reworked, otherwise can it be never safe.

peak3d:
@rbuehlma https://github.com/peak3d/pvr.zattoo/tree/k18 contains a working pvr.zattoo based on the PR 12754.

rbuehlma:
Thank you, I will include that in my repo. I will also switch to the c++ api as far as possible.

ksooo:
> I will also switch to the c++ api as far as possible.

 Currently, this is not possible at all for PVR. @AlwinEsch has the rquired changes on his list, but it is not even PRed yet.

rbuehlma:
Yeah I know, but the addon API is already there and simplifies to addon code somewhat. The PVR part will follow when ready.

ksooo:
> Yeah I know, but the addon API is already there and simplifies to addon code somewhat. The PVR part will follow when ready.

 IIRC you cannot use the other (non-PVR) parts until some modifications to the PVR addon in general are done. Don't waste your time. Wait, until @AlwinEsch has finished the needed parts.

PR #12738, opened 2017-08-29
[addons] treat repository.xbmc.org as override repository

Rechi:
## Description
If an addonid is in the override repository, versions from other repos with the same addonid can't be installed regardless if those versions are higher, lower or equal.

## Motivation and Context
Block other repositories from overriding addons which `repository.xbmc.org` contains.
@MartijnKaijser FYI

## How Has This Been Tested?
Tested with `repository.ytplugin` which had the same version of the youtube addon, but version from youtube repository wasn't show, as expected.
Also used `repository.kodinerds` which had a newer version of `service.fussballpopup` (than `repository.xbmc.org`), but only the version from the xbmc repository was installed, even if going to install from repository -> kodi nerds and install from there. (as expected)

## Types of change

## Checklist:


akva2:
While i understand what you are trying to achieve it doesnt seem well thought true. this kills the ability to have testing repos for addons which is a major meh.

Rechi:
Testing repos can still exists, just append `.dev` to the addon id in that repo and it will work as expected.

garbear:
we have a `addons-manifest.xml`. Maybe the override repo can be defined there?

Rechi:
override repo has to compiled into source code, otherwise sooner or later an addon will change the file where it is defined
for rebrandings we may want to add the definition in `version.txt`, but compile the value into the executable, like we already do for `APP_NAME`, `VERSION_TAG`, ...

garbear:
+1 for version.txt

wsnipex:
I already see ppl hex editing kodi

echocoderxbmc:
Also, whats to stop someone forking a legit 3rd party addon then submitting it to org? The original dev then gets locked out of his addon id?

ronie:
> whats to stop someone forking a legit 3rd party addon then submitting it to org?

1. common decency
2. our repo rules which prohibit anyone from submitting someone else's code.
3. our review process

if all of the above fail, we'd simply remove the addon again from our repo, once notified by the original author.

echocoderxbmc:
Hi ronie.

So what if there is a third party addon called plugin.video.cardboardbox for example

This addon has been around for a few years and someone makes a new addon with the same id (different code etc but same code) and submits it to org.

I would assume the answer would be that nobody owns the rights to any plugin id?

jsergio123:
My real concern is would this one day lead to Kodi deciding to kill off any addons they wish by overriding their pulgin id's?

ronie:
> I would assume the answer would be that nobody owns the rights to any plugin id?
assuming no one notices the duplicate id before it gets accepted in our repo,
we would discuss changing the addon id with the person who submitted it to our repo once we're made aware of it.

ronie:
> My real concern is would this one day lead to Kodi deciding to kill off any addons they wish by overriding their pulgin id's?

this is the official repo we're talking about, not some crappy third party repo where it's common practice to overwrite / ruin eachothers addons.

we have no intent to ever lower ourselves to those practices.

tvaddonsco:
I definitely agree with you guys adding more security as to prevent all out chaos that will otherwise come, but have a few healthy concerns...

What about URLResolver? It has been maintained by different people outside of the official repo for years now, I'm assuming the official repo won't accept things like pairing with OpenLoad, Vidup, etc. Furthermore, it is generally updated weekly, which wouldn't be allowed in the official repo as far as I remember...

Also, I mean no disrespect, but less than a week ago an addon ID (plugin.video.newsy) was added to the official repo that was the same as one mhancoc7 had released over a month ago. He had no choice but to change the addon ID as to avoid pushing conflicting versions with the official repo. It also happened with plugin.video.ustvnow which we had maintained for years before someone else added it to the official repo, at which point we had no choice but to change the addon ID of our addon as well. See: https://forum.kodi.tv/showthread.php?tid=320402

- Eleazar Coding

ronie:
at the end of the day, if things can't be solved otherwise, the official kodi addon repo will 'prevail' above third party repos.

every addon author is encouraged to submit their legitimate addons to the official repo.
if they choose not to, i'm sorry, but it's their loss when things go wrong.

========
@ all please stop commenting on this PR, you can post all remaining questions here:
https://forum.kodi.tv/showthread.php?tid=320752

tamland:
Mind explaining what you're actually trying to achieve with this?

xbmc/addons/AddonManager.cpp
```diff
@@ -1190,7 +1192,19 @@ bool CAddonMgr::AddonsFromRepoXML(const CRepository::DirInfo& repo, const std::s
...
+          else
+          {
+            ADDON::VECADDONS addonsFind;
+            m_database.FindByAddonId(addon.get()->ID(), addonsFind);
+            if (std::find_if(addonsFind.begin(), addonsFind.end(), [](AddonPtr const& p) { return p.get()->Origin() == OVERRIDE_REPOSITORY; }) == addonsFind.end())
+              addons.push_back(std::move(addon)); 
+            else
+              CLog::Log(LOGNOTICE, "CAddonMgr: %s served by %s, skiping version from %s", addon.get()->ID().c_str(), OVERRIDE_REPOSITORY.c_str(), repoId.c_str());
+          }
+        }
```
tamland: This is all wrong anyway. Updates does not happen in any order so there is no guarantee anything will exist in the db. And it makes the function do different things depending on mutable global state; ugh

xbmc/addons/AddonManager.cpp
```diff
@@ -1190,7 +1192,19 @@ bool CAddonMgr::AddonsFromRepoXML(const CRepository::DirInfo& repo, const std::s
...
+          if (repoId.empty() || repoId == OVERRIDE_REPOSITORY)
+            addons.push_back(std::move(addon));
+          else
+          {
+            ADDON::VECADDONS addonsFind;
+            m_database.FindByAddonId(addon.get()->ID(), addonsFind);
+            if (std::find_if(addonsFind.begin(), addonsFind.end(), [](AddonPtr const& p) { return p.get()->Origin() == OVERRIDE_REPOSITORY; }) == addonsFind.end())
+              addons.push_back(std::move(addon)); 
+            else
+              CLog::Log(LOGNOTICE, "CAddonMgr: %s served by %s, skiping version from %s", addon.get()->ID().c_str(), OVERRIDE_REPOSITORY.c_str(), repoId.c_str());
```
ksooo: typo: "skiping" => "skipping"

MartijnKaijser:
@tamland To prevent non official repos overriding the add-ons we provide. This partly a security measure and counter measure towards possible harmful abuse of "trusted"  add-ons we currently provide which can be overridden by externals at any time. Discussion can be found here https://forum.kodi.tv/showthread.php?tid=320752  and in team internal.

xbmc/addons/AddonManager.cpp
Rechi: I would expect database already has information from override repo, if you install a 3rd party repo.
Also I know that it may compares with the data from previous sync from the override repo, but I don't expect it would be that tragic.

Rechi: @tamland just a general question, would you suggest me to not filter the addons inserted into the db, but filter the queries like I've done in variant 1?

tamland:
Let me be clear about this: this does in no way improve security or prevent non-official repos from overriding add-ons. Once a 3rd party repo is installed, it's over. It will have have full access to db and this check and any other origin check is easy to bypass.

I find this idea of trying to make one special repository the master of all other confusing and not a very good one. I'd much rather see cross-repository updates removed completely (and equally for all) or make it so that it required manual interaction to change the origin). Don't confuse this for security though. It will only "prevent" non-malicious repos and addons from taking over others.

da-anda:
and how would you do that @tamland? Track the originating repository in the DB, where any malicious add-on can simply change the value to it's own repo as well and replace any other add-on with spyware? As long as add-ons have direct write access to core databases and have full write access to everything it doesn't matter what we do. Building a 4m high wall to keep bad guys out doesn't help when there is a 5m high bridge over it.
Having the override repo compiled in would at least be one thing that couldn't be altered by malicious add-ons. But as it has been pointed out in the forum discussion, this might cause trouble for our add-on devs and beta testing.

TK-1One:
Bad Bad idea, i use the XBMCNerds (KodiNerds) Repo a lot, and often there are more actual Versions than in the official Repo. Therefor this Pull isnt whise and a Bad Idea.
Why the NerdsRepo? Nice Addons and almost all German ones on one Place.
I knew there is a Button to "hide foreign" Addons, but that simply never realy worked. And as long that is the case, i prevere to use a Repo with foreigen Addons rather than, or beside the official.
Regards

da-anda:
@TK-1One imagine you never would have heard of the Kodinerds repo and notice that some random german add-on is broken in Kodi repo. Months later you find out that over at Kodinerds, the add-on had been fixed for months. How would you feel about that? Add-ons that are in official repo should also be fixed in/for official repo and not get maintained in third party repos where only a subset of Kodi users benefits from it. But if they don't want to push fixes to official repo, they can stay in wonderland and simply have the add-on IDs changed.

chewitt:
The concept of repo priority has merit and provides much needed logic and order to the chaos seen when users install multiple repos - thumbs up on that. However, hard-coding 'god mode' for the Kodi repo over all others goes against the open spirit of Kodi and the projects' long-standing and neutral "what users do in private isn't our concern" policies. Code rip-off repos like tvarseholes and the larger issue of how to securely publish binary add-ons need to be addressed another way. This is not a good band-aid that advances security or user-experience so please implement a simple priority scheme and stop there; for now. Let's take this complex discussion to team channels and DevCon - I'd be happy to lead a session on this topic.

tyl0re:
I Understand the Problem, wouldnt it be Better, that the User can Descide "Priority Kodi.TV, Pririty to Version Number". If it Comes to the Solution now Suggested, i Would Request to Remove the Addons, or if its Not Possible: I will Change all IDS, an stop Maintain all Addons on Kodi.tv, so they will be deleted beacause its broken at some Point

@TK-1One  So , you dont have to worry, it will still work

phunkyfish:
@Rechi is this PR something you would still like to merge?

phunkyfish:
An alternative has been proposed here: https://github.com/xbmc/xbmc/pull/17677

PR #12740, opened 2017-08-29
[PVR][settings] Make epg.lingertime an official setting, remove the respective advanced setting.

ksooo:
Many users like to adjust "epg linger time" to zero, because for live tv, they are not interested in what was broadcasted in the past.

Others like to see many days back in their programme guide, as they are using a pvr addon that supports "replay" of tv shows (example: Zattoo TV).

Before this PR, those users had to deal with advanced settings (epg.lingertime) to make their programme guide fit their needs. This PR changes this by making "lingertime" an official setting (and removing the respective advanced setting).

screenshot000

BTW:  The new setting can be adjusted in the range from 0 to 31 days, default is 1 day (like the default for epg lingertime advanced setting was).

This change has been runtime.tested on macOS, latest Kodi master.

@xhaggi @Jalle19 mind taking a look at my changes.


xbmc/pvr/epg/Epg.cpp
```diff
@@ -145,8 +145,8 @@ void CPVREpg::Clear(void)
 
 void CPVREpg::Cleanup(void)
 {
-  CDateTime cleanupTime = CDateTime::GetCurrentDateTime().GetAsUTCDateTime() -
-      CDateTimeSpan(0, g_advancedSettings.m_iEpgLingerTime / 60, g_advancedSettings.m_iEpgLingerTime % 60, 0);
+  int iPastDays = CServiceBroker::GetSettings().GetInt(CSettings::SETTING_EPG_PAST_DAYSTODISPLAY);
+  const CDateTime cleanupTime = CDateTime::GetUTCDateTime() - CDateTimeSpan(iPastDays, 0, 0, 0);
```
xhaggi: could we unify initialization and use `cleanupTime(value)` like you do all over the place

xbmc/pvr/windows/GUIEPGGridContainerModel.cpp
```diff
@@ -325,7 +326,7 @@ void CGUIEPGGridContainerModel::FindChannelAndBlockIndex(int channelUid, unsigne
 
 unsigned int CGUIEPGGridContainerModel::GetGridStartPadding() const
 {
-  unsigned int iEpgLingerTime = g_advancedSettings.m_iEpgLingerTime;
+  unsigned int iEpgLingerTime = CServiceBroker::GetSettings().GetInt(CSettings::SETTING_EPG_PAST_DAYSTODISPLAY) * 24 * 60;
```
xhaggi: could we please rename the variable to `iPastMinutesToDisplay` or something similar to the setting name.

xhaggi (changes requested):
Apart from the minors, it looks good to me. We discussed that years ago, but decide (not me) to not introduce a real setting for it. IMO it makes sense for what you've said.

xbmc/pvr/epg/Epg.cpp
ksooo: For modern C++ it should be quite the other way round. "default construction + assignment" is more readable code and with modern compilers there is no need any more to use "copy construction". Formerly, I did this for performance reasons. Modern compilers will elide this - no need for such tricks anymore. So, my code evolves over time.  Changing all occurrences of this pattern definitely is for another day.

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -18090,11 +18094,14 @@ msgctxt "#36218"
 msgid "This category contains the electronic programming guide (EPG) settings."
 msgstr ""
 
-#empty string with id 36219
+#: system/settings/settings.xml
+msgctxt "#36219"
+msgid "Number of past days to show in the guide and import data for from backends."
```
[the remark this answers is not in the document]
ksooo: Will change it, yes.

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -18090,11 +18094,14 @@ msgctxt "#36218"
...
+msgstr ""
 
 #: system/settings/settings.xml
 msgctxt "#36220"
-msgid "Number of days to show in the guide and import data for from backends."
+msgid "Number of future days to show in the guide and import data for from backends."
```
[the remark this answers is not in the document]
ksooo: same

xbmc/pvr/windows/GUIEPGGridContainerModel.cpp
ksooo: sure

ksooo:
@xhaggi the changes you requested are in now.

xhaggi (approved):

xbmc/pvr/epg/Epg.cpp
xhaggi: Sure

xbmc/pvr/windows/GUIEPGGridContainerModel.cpp
```diff
@@ -325,10 +326,10 @@ void CGUIEPGGridContainerModel::FindChannelAndBlockIndex(int channelUid, unsigne
 
 unsigned int CGUIEPGGridContainerModel::GetGridStartPadding() const
 {
-  unsigned int iEpgLingerTime = g_advancedSettings.m_iEpgLingerTime;
+  unsigned int iPastDays = CServiceBroker::GetSettings().GetInt(CSettings::SETTING_EPG_PAST_DAYSTODISPLAY) * 24 * 60;
```
Jalle19: If this is `iPastDays`, why do you multiply by `24 * 60`?

xbmc/settings/AdvancedSettings.cpp
```diff
@@ -300,7 +300,6 @@ void CAdvancedSettings::Initialize()
   m_bVideoScannerIgnoreErrors = false;
   m_iVideoLibraryDateAdded = 1; // prefer mtime over ctime and current time
 
-  m_iEpgLingerTime = 60 * 24;           /* keep 24 hours by default */
```
Jalle19: I think this answers my previous question, but `iPastDays` is still a bit confusing if we're actually dealing with minutes

xbmc/pvr/windows/GUIEPGGridContainerModel.cpp
ksooo: Good catch. Will change this. Thanks.

ksooo:
@Jalle19 changed the misleading variable name, as requested by you.

xbmc/pvr/epg/EpgContainer.cpp
```diff
@@ -118,7 +118,9 @@ CPVREpgContainer::CPVREpgContainer(void) :
   m_settings({
     CSettings::SETTING_EPG_IGNOREDBFORCLIENT,
     CSettings::SETTING_EPG_EPGUPDATE,
-    CSettings::SETTING_EPG_DAYSTODISPLAY
+    CSettings::SETTING_EPG_FUTURE_DAYSTODISPLAY,
+    CSettings::SETTING_EPG_PAST_DAYSTODISPLAY,
+    CSettings::SETTING_EPG_PREVENTUPDATESWHILEPLAYINGTV
```
Jalle19: Where did this come from?

ksooo: + SETTING_EPG_DAYSTODISPLAY renamed to CSettings::SETTING_EPG_FUTURE_DAYSTODISPLAY
+ SETTING_EPG_PAST_DAYSTODISPLAY newly introduced
+ SETTING_EPG_PREVENTUPDATESWHILEPLAYINGTV strictly speaking unrelated, but noticed that it was forgotten to keep this value in local settings cache (and to obtain it from there - another unrelated change in this PR - but consistent) ;-)

PR #12747, opened 2017-08-30
FFMPEG: Fixes rbpi compilation

nsenica:
## Description
With the latest refactoring of platforms and systems, raspberry pi compilation was not getting the correct flags. This PR attempts to fix it.

## Motivation and Context
Fixing compiler flags for raspberry pi builds.
 
## How Has This Been Tested?
Compiling code.

## Types of change

## Checklist:

### DIFF
--- a/cmake/modules/FindFFMPEG.cmake
+++ b/cmake/modules/FindFFMPEG.cmake
@@ -250,6 +250,7 @@ if(NOT FFMPEG_FOUND)
                                  -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
                                  -DFFMPEG_VER=${FFMPEG_VER}
                                  -DCORE_SYSTEM_NAME=${CORE_SYSTEM_NAME}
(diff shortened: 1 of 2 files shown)


wsnipex (approved):
thanks, looks good

popcornmix:
It doesn't break my build, so if @wsnipex is happy then so am I.

Rechi:
I think the change for the CMakeLists.txt should be the same as https://github.com/xbmc/xbmc/commit/853c7f6403 for the Makefile, or is there actually a good reason for the different parts?

nsenica:
@Rechi , in practice both are equivalente, here we are just explicitly disabling those options, since it depends on the ffmpeg default config, it might be safer to explicitly doing so. If you find it important, I can add these switches to the Makefile.

About the --enable-pic, it is enabled a little down below at https://github.com/xbmc/xbmc/blob/master/tools/depends/target/ffmpeg/CMakeLists.txt#L66

wsnipex:
@rechi ok?

Rechi:
Fine with me.

PR #12748, opened 2017-08-30
[Fix] [Music] GetGenresByArtist for album artists

DaveTBlake:
GetGenresByArtist is not handling artists that have relationship with song is via album and album artist rather than song artist. Minor query change to return these values and correct the oversight.

Explanation: not all album artists are also song artists for songs on the album. For example Musicbrainz tagging guidelines use the guitarist as the release (album) artist but composer as the track/recording (song) artist on classical guitar music. So collating the song genres for an artist needs to consider the genre(s) of songs on albums by each artist.

### DIFF
--- a/xbmc/music/MusicDatabase.cpp
+++ b/xbmc/music/MusicDatabase.cpp
@@ -1782,17 +1782,34 @@ bool CMusicDatabase::GetGenresByArtist(int idArtist, CFileItem* item)
 {
   try
   {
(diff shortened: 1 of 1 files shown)


DaveTBlake:
Build of Krypton backport tested on both MySQL and SQLite databases, see #12750.
Build errors on 64 bit are not related to this PR.

I will merge this on Monday unless anyone has any objections before then.

DaveTBlake:
jenkins build this please

PR #12753, opened 2017-08-31
OpenGLES2 and EGL: Adapt library names for raspberry pi

nsenica:
## Description
In raspbian stretch they decided to remove standard libraries (libglesv2, libegl) from the firmware and include only the broadcom libraries (libbrcmglesv2, libbrcmegl).
When trying to compile kodi in stretch libglesv2 and libegl are not found in the firmware path, due to them being removed, so cmake resorts to the open source mesa drivers if installed.
In jessie both libraries are included in the firmware path, so there's no problem finding them.

To make it work in both systems seamlessly and without breaking the build for other systems, we're proposing this PR, which changes a bit how OpenGLES2 and EGL libraries are found, and will I'll explain in detail what is going on.

First, we replaced pkg_check_modules by pkg_search_module. According to the [documentation](https://cmake.org/cmake/help/v3.7/module/FindPkgConfig.html#command:pkg_search_module):

> Same as pkg_check_modules(), but instead it checks for given modules and uses the first working one.

This is exactly what we wanted, we can provide raspberry pi specific library first and if not found, try to find the generic one. In all other systems, the raspberry pi won't be found due to its specific name, so it won't change the behaviour.

The same reasoning was applied to the find_library, which uses the same reasoning above.

This will probably need a backport, after this is approved I'll provide it as well.

## Motivation and Context
Be able to compile against optimized libraries in raspbian in both jessie and stretch without breaking the build on other platforms/system.

## How Has This Been Tested?
Compiling and running kodi.

## Types of change

## Checklist:

### DIFF
--- a/cmake/modules/FindEGL.cmake
+++ b/cmake/modules/FindEGL.cmake
@@ -14,14 +14,18 @@
 #
 #   EGL::EGL   - The EGL library
 
+if(CORE_PLATFORM_NAME_LC STREQUAL rbpi)
+    set(_brcmprefix brcm)
+endif()
+
 if(PKG_CONFIG_FOUND)
-  pkg_check_modules(PC_EGL egl QUIET)
+  pkg_check_modules(PC_EGL ${_brcmprefix}egl QUIET)
 endif()
 
 find_path(EGL_INCLUDE_DIR EGL/egl.h
                           PATHS ${PC_EGL_INCLUDEDIR})
 
-find_library(EGL_LIBRARY NAMES EGL egl
+find_library(EGL_LIBRARY NAMES ${_brcmprefix}EGL egl
                          PATHS ${PC_EGL_LIBDIR})
 
 set(EGL_VERSION ${PC_EGL_VERSION})
--- a/cmake/modules/FindOpenGLES.cmake
+++ b/cmake/modules/FindOpenGLES.cmake
@@ -10,14 +10,18 @@
 # OPENGLES_LIBRARIES - the OpenGLES libraries
 # OPENGLES_DEFINITIONS - the OpenGLES definitions
 
+if(CORE_PLATFORM_NAME_LC STREQUAL rbpi)
+    set(_brcmprefix brcm)
+endif()
+
 if(PKG_CONFIG_FOUND)
-  pkg_check_modules(PC_OPENGLES glesv2 QUIET)
+  pkg_check_modules(PC_OPENGLES ${_brcmprefix}glesv2 QUIET)
 endif()
 
 if(NOT CORE_SYSTEM_NAME STREQUAL ios)
   find_path(OPENGLES_INCLUDE_DIR GLES2/gl2.h
                                  PATHS ${PC_OPENGLES_INCLUDEDIR})
-  find_library(OPENGLES_gl_LIBRARY NAMES GLESv2
+  find_library(OPENGLES_gl_LIBRARY NAMES ${_brcmprefix}GLESv2
                                    PATHS ${PC_OPENGLES_LIBDIR})
 else()
   find_library(OPENGLES_gl_LIBRARY NAMES OpenGLES


Rechi:
why have you added the specific broadcom libraries as first find lib instead of last?

wsnipex:
I don't like to have workarounds for distros "knowing it better".

nsenica:
@Rechi I think I explained that in the PR description. Broadcom libs are specific to raspbian, if they are not found in the system, cmake will try to find it by the next name we provided. This ensures that in raspbian it uses the correct library, but since it is not present on other systems it will try to find the next one. Did I answer your question ? 

@wsnipex Neither do I, but without this patch, kodi doesn't compile in raspbian stretch due to not having those libraries available.

I'm really looking forward to getting feedback on this and improving this PR, all comments and suggestions are welcome.

wsnipex:
ok, looks like this is not a raspbian change, but an upstream one:
https://github.com/raspberrypi/firmware/tree/master/opt/vc/lib
@popcornmix ping

nsenica:
yes, indeed, this is in the firmware that will reflect on raspbian downstream :) I wasn't clear about it on the details.

wsnipex:
pending comment from popcornmix, I'd prefer something like this:
>if(CORE_PLATFORM_NAME_LC STREQUAL rbpi)
  set(_bcrmprefix bcrm)
endif()
pkg_check_modules(EGL ${_bcrmprefix}egl)

nsenica:
Yes, that is another option :) 

However, I found it more confusing and less clean than mine, although it makes it easier to understand that there is an exception for rbpi. I'm torn here, I'd prefer a cleaner solution, but yours is clearer :)

popcornmix:
I can describe the background here.
Kodi uses the firmware GL libraries.
Historically they have had the standard mesa names (libEGL.so etc) but are installed in /opt/vc/lib.
Historically the mesa libs are software emulated without using the 3D hardware so are virtually useless.
Having the mesa libs installed generally breaks building Kodi (and other packages) if they are found first, so generally advice has been to ensure they aren't installed.

But the arm side GL mesa libs are now pretty good (after enabling the vc4 kms/fkms driver device tree overlay), and allows many standard GL packages to run well. At some point in the future the arm side libs will become the default and the firmware side libs will be deprecated.

To allow the two drivers to exist without conflicts we have renamed them to libbrcmEGL.so etc and applications like kodi should be linking with these.

At some point in the future Kodi will use the arm side GL driver and will just use the standard linux code for GL rendering (either GL or GLES should be possible but GLES will probably have better performance). Currently we don't a solution for HW accelerated video in this configuration but it is being worked on.

So this PR is desired for fixing the current problem with building on raspbian stretch. The desired behaviour is for the libbrcmEGL.so to be found even when libEGL.so also exists.
@wsnipex's suggestion looks like it achieves this behaviour.
I'm not certain of the behaviour of pkg_check_modules/findlibrary when there is more than one match in different paths - does the first named match always take priority, or will the first directory searched that matches any of the listed names be used?

hudokkow:
First found is used.

nsenica:
@popcornmix Thanks for the explanation, they're really insightful and sheds a light on the whole issue.

I think my changes really address this issue:
> The desired behaviour is for the libbrcmEGL.so to be found even when libEGL.so also exists.

The first found is the one that cmake will use. That's the whole purpose of this PR.

We can make it more clear to raspberry pi using @wsnipex's approach, or cleaner as my initial proposal. I'm equally fine with any of them, once agreed upon I'll update the PR as needed.

hudokkow:
Go with wsnipex approach please. Makes it clear that it's rpi specific

nsenica:
I'll try to make the changes today and will update this PR later on.

Thank you all for all the comments and feedback.

nsenica:
Commit updated :) Let's hope it goes well.

cmake/modules/FindEGL.cmake
```diff
@@ -14,14 +14,18 @@
 #
 #   EGL::EGL   - The EGL library
 
+if(CORE_PLATFORM_NAME_LC STREQUAL rbpi)
+    set(_brcmprefix brcm)
+endif()
+
 if(PKG_CONFIG_FOUND)
-  pkg_check_modules(PC_EGL egl QUIET)
+  pkg_check_modules(PC_EGL ${_brcmprefix}egl QUIET)
```
hudokkow: Missing space after `${_brcmprefix}` here and on all other changes below.

Rechi: @hudokkow that's correct, if rbpi search for `brcmegl` otherwise for `egl `

hudokkow: Doh, my bad! We're looking for `brcmegl`
lgtm.

Sorry for the noise.

wsnipex (approved):
looks good, thanks

PR #12754, opened 2017-08-31
GetFileProperty for "old" binary addon interface

peak3d:
## Description
Provide Interface to get CURL Response Headers / CURL return code

Addition to https://github.com/xbmc/xbmc/pull/12737

## Types of change

### DIFF
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/libXBMC_addon.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/libXBMC_addon.h
@@ -155,6 +155,7 @@ typedef struct CB_AddOn
   int (*GetFileChunkSize)(const void* addonData, void* file);
   bool (*FileExists)(const void* addonData, const char *strFileName, bool bUseCache);
   int (*StatFile)(const void* addonData, const char *strFileName, struct __stat64* buffer);
+  char *(*GetFileProperty)(const void* addonData, void* file, XFILE::FileProperty type, const char *name);
   bool (*DeleteFile)(const void* addonData, const char *strFileName);
   bool (*CanOpenDirectory)(const void* addonData, const char* strURL);
   bool (*CreateDirectory)(const void* addonData, const char *strPath);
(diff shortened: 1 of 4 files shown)


xbmc/addons/kodi-addon-dev-kit/include/kodi/libXBMC_addon.h
```diff
@@ -155,6 +155,7 @@ typedef struct CB_AddOn
   int (*GetFileChunkSize)(const void* addonData, void* file);
   bool (*FileExists)(const void* addonData, const char *strFileName, bool bUseCache);
   int (*StatFile)(const void* addonData, const char *strFileName, struct __stat64* buffer);
+  char *(*GetFileProperty)(const void* addonData, void* file, XFILE::FileProperty type, const char *name);
```
ksooo: The "type" parameter introduces a non Kodi devkit internal C++ type to the public C API , which is a nogo. It even has a namespace, which is a C++ language construct. Unfortunately, we already do this at other places in this file, but I think we can do better. Correct solution is to introduce a new enum for the type in this header file.

But as this API is deprecated  and supposed to die "soon", it's up to you to change this or not. Just saying.

ksooo:
@rbuehlma please don't forget to call XBMC-FreeString to free the memory allocated for the return value of GetFileProperty. Otherwise we would leak memory.

xbmc/addons/kodi-addon-dev-kit/include/kodi/libXBMC_addon.h
peak3d: I was a little bit surprised that we use XFILE things here, looking at the code IFileTypes.h is copied into the addon dev kit during binary addon installation step.

Sure, nice is different. but I thougth as well that this interface is a question of month befor it dies.

ksooo: Yeah it's a mess and good  that it will (hopefully) die soon.

PR #12755, opened 2017-09-01
[pvr] PVR Addon API 5.7.1: Support for passing MimeType for PVR streams

peak3d:
## Description
Allow setting Mime-Type for PVR streams + Disable Content-Lokup for known streams.

Todos for addon maintainers: none, as minimum pvr api version number did not change.

## Motivation and Context
Mime-Type content lookup takes some time (HEAD Request) and can be supressed if PVR addon knows the mime type. Speeds up channel switching.

Should not lead to version issues because PVR_VERSION_MIN is unchanged.

## How Has This Been Tested?
pvr.zattoo + mpeg DASH streams

## Types of change

### DIFF
--- a/xbmc/addons/PVRClient.cpp
+++ b/xbmc/addons/PVRClient.cpp
@@ -825,7 +825,11 @@ bool CPVRClient::FillEpgTagStreamFileItem(CFileItem &fileItem)
   {
     if (strncmp(properties[i].strName, PVR_STREAM_PROPERTY_STREAMURL, strlen(PVR_STREAM_PROPERTY_STREAMURL)) == 0)
       fileItem.SetDynPath(properties[i].strValue);
-
+    else if (strncmp(properties[i].strName, PVR_STREAM_PROPERTY_MIMETYPE, strlen(PVR_STREAM_PROPERTY_MIMETYPE)) == 0)
+    {
+      fileItem.SetMimeType(properties[i].strValue);
+      fileItem.SetContentLookup(false);
+    }
     fileItem.SetProperty(properties[i].strName, properties[i].strValue);
   }
 
@@ -1299,7 +1303,11 @@ bool CPVRClient::FillChannelStreamFileItem(CFileItem &fileItem)
   {
     if (strncmp(properties[i].strName, PVR_STREAM_PROPERTY_STREAMURL, strlen(PVR_STREAM_PROPERTY_STREAMURL)) == 0)
       fileItem.SetDynPath(properties[i].strValue);
-
+    else if (strncmp(properties[i].strName, PVR_STREAM_PROPERTY_MIMETYPE, strlen(PVR_STREAM_PROPERTY_MIMETYPE)) == 0)
+    {
+      fileItem.SetMimeType(properties[i].strValue);
+      fileItem.SetContentLookup(false);
+    }
     fileItem.SetProperty(properties[i].strName, properties[i].strValue);
   }
   return true;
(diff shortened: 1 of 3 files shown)


xbmc/addons/PVRClient.cpp
```diff
@@ -1299,8 +1299,13 @@ bool CPVRClient::FillChannelStreamFileItem(CFileItem &fileItem)
...
       fileItem.SetDynPath(properties[i].strValue);
-
-    fileItem.SetProperty(properties[i].strName, properties[i].strValue);
+    else if (strncmp(properties[i].strName, PVR_STREAM_PROPERTY_MIMETYPE, strlen(PVR_STREAM_PROPERTY_MIMETYPE)) == 0)
+    {
+      fileItem.SetMimeType(properties[i].strValue);
+      fileItem.SetContentLookup(false);
+    }
+    else
+      fileItem.SetProperty(properties[i].strName, properties[i].strValue);
```
ksooo: Why 'else'? I would also put the original property into the fileitem. That's what we do with "streamurl".

xbmc/addons/kodi-addon-dev-kit/include/kodi/versions.h
```diff
@@ -114,7 +114,7 @@
 #define ADDON_INSTANCE_VERSION_PERIPHERAL_DEPENDS     "addon-instance/Peripheral.h" \
                                                       "addon-instance/PeripheralUtils.h"
 
-#define ADDON_INSTANCE_VERSION_PVR                    "5.7.0"
+#define ADDON_INSTANCE_VERSION_PVR                    "5.7.1"
```
ksooo: Note: All PVR addons need to be recompiled @MilhouseVH

ksooo (commented):
Next time, please don't forget to label the PR "PVR Addon API change". And I always explain in detail in the description of such PRs what the changes mean exactly for the PVR Addon maintainers (assuming they read PRs flagged "PVR API change") - for instance, whether addon code must be adjusted and how or whether a recompile of the addon is sufficient.

xbmc/addons/PVRClient.cpp
peak3d: do you want to have streamurl in properties?

ksooo: Not sure whether not changing min version min actually leads to no incompatibilities. @AlwinEsch ?

xbmc/addons/kodi-addon-dev-kit/include/kodi/versions.h
ksooo: Not sure whether not changing min version min actually leads to no incompatibilities. Maybe I'm wrong on this. I always get puzzled about this min version stuff. @AlwinEsch ?

MartijnKaijser: It used to but i'm not sure about current state. Remember it used to be in addon.xml that made addon manager disable it?

ksooo: Just tested. Only increasing version number, but keeping min version number, like @peak3d did, is working just fine. Addon stays compatible, no need to recompile. All good. Sorry for the noise.

xbmc/addons/PVRClient.cpp
ksooo: And I see no reason against it. For me, it is a consistency thingy: Everything the addon sts as property can be found 1:1 at the FileItem.

ksooo (approved):

xbmc/addons/kodi-addon-dev-kit/include/kodi/versions.h
AlwinEsch: For a thing like them is not needed to increase min version. There is nothing changed in the "C" Interface, only a new identification string for properties added. All older compiled addons are still compatibel with the change here.

PR #12524, opened 2017-07-18
CPeripheralCecAdapter: Execute application operations from application thread

popcornmix:
These function are not designed to be called form other threads

See: https://github.com/xbmc/xbmc/commit/82fe52deb8c11df31fbdf8c298b28a5e1a1491aa#commitcomment-19794624

## Types of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [X] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #12758, opened 2017-09-01
[cec] Fix power messages handling.

ksooo:
This is a followup to #12524 which introduced the regression described here https://github.com/xbmc/xbmc/pull/12524#issuecomment-326426116

This change has been tested on Linux, latest Kod master.


da-anda:
using a new enum for the powerdown message will likely not catch all cases the actual power state messages are fired. Have a look at our Builtins which still fire the "regular" powerstate messages and thus won't trigger your adjusted logic. https://github.com/xbmc/xbmc/blob/master/xbmc/interfaces/builtins/SystemBuiltins.cpp
I'm not sure that all builtin actions are proxied through g_Application.ExecuteXBMCAction() which does the same PVR related check. I might probably be a better idea to either check for all individual power state ENUMs or adjust all related messages in core (and then get rid of the special treatment in ExecuteXBMCAction which might no longer be needed then). Up to you - I only noticed this and wanted to mention it.

ksooo:
If there is something still not correct it is neither caused by the PR that introduced this regression nor this fix. This fix restores the previous behavior wrt. power messages triggered by CEC. Everything else is for another PR.

ksooo:
jenkins error is unrelated.

PR #12767, opened 2017-09-04
Partial revert of imdb crashfix due to a introduced bug

lobermann:
As discussed with @DaVukovic and @koying in slack this reverts part of my previous imdb crash fix as it was causing a bug in that content can no longer be added manually to the videolibrary.

I would appreciate if also somebody else can also try this PR, so that we are really sure it fixes the issue :-)

### DIFF
--- a/xbmc/video/windows/GUIWindowVideoBase.cpp
+++ b/xbmc/video/windows/GUIWindowVideoBase.cpp
@@ -398,10 +398,6 @@ bool CGUIWindowVideoBase::ShowIMDB(CFileItemPtr item, const ScraperPtr &info2, b
     CGUIDialogOK::ShowAndGetInput(CVariant{13346}, CVariant{14057});
     return false;
   }
-  
-  // If the scraper failed above and no videoinfotag was created, return
-  if (!item->HasVideoInfoTag())
-    return false;
 
   bool listNeedsUpdating = false;
   // 3. Run a loop so that if we Refresh we re-run this block


koying (approved):
Thanks.
We probably want someone to confirm this fixes the issue, though

DaVukovic:
Probably I'm able to confirm the fix later in the evening. I will test it for sure, not later than tomorrow.

DaVukovic:
@koying @lobermann 

I tested it just now and manually adding files works at least for movies, which doesn't worked before. So at least that is fixed. I assume that everything else will work, too. 

I can do further tests tomorrow if needed. Just ping me in case.

MilhouseVH:
One negative report: https://forum.kodi.tv/showthread.php?tid=298461&pid=2643031#pid2643031

DaVukovic:
Probably not related because the user misses to mention that's the artwork which is missing. The scraping itself seems to work fine

See: https://forum.kodi.tv/showthread.php?tid=298461&pid=2643266#pid2643266

arnova:
Please elaborate how this fixes your issue. I don't get why the check for empty Videoinfotag caused a problem as an empty tag triggers a crash here previously, right?

koying:
Because the tag is almost guaranteed to be empty, here, and the `return`actually prevents it to be filled by the `RefreshItemModal`

The actual crash fix is, I assume, fixed by the other half of the reverted commit, which stays

Razzeee:
jenkins build this please

arnova:
@koying: Ah yes, now I see. Thanks for explaining it. +1 in that case.

MartijnKaijser:
Was this issue already fixed in Krypton branch?

koying:
The regression was introduced in 18.
I don't know if the original crash issue is present in 17 as well...

koying:
jenkins build and merge

brozikcz:
The regression is in 17 version too.

https://forum.kodi.tv/showthread.php?tid=321180

koying:
Ah, right. The regression was backported to 17.4

MartijnKaijser:
@lobermann care to create a fix for krypton?

a1rwulf:
afaik he is on vacation:
https://github.com/xbmc/xbmc/pull/12822

I did just checkout the Krypton branch and cherry-picked, I didn't even compile!

PR #12770, opened 2017-09-05
[PVR] CPVRClient: Refactor writing addon stream properties to file item.

ksooo:
Just a small code refactoring. No functionality should be changed by this.

@peak3d this effects and simplifies #12755 

@Jalle19 mind taking a look at the code change.

### DIFF
--- a/xbmc/addons/PVRClient.cpp
+++ b/xbmc/addons/PVRClient.cpp
@@ -636,7 +636,7 @@ PVR_ERROR CPVRClient::RenameChannel(const CPVRChannelPtr &channel)
   return retVal;
 }
 
-void CPVRClient::CallMenuHook(const PVR_MENUHOOK &hook, const CFileItem *item)
+void CPVRClient::CallMenuHook(const PVR_MENUHOOK &hook, const CFileItemPtr item)
 {
   if (!m_bReadyToUse)
     return;
@@ -1399,9 +1399,9 @@ bool CPVRClient::HasMenuHooks(PVR_MENUHOOK_CAT cat) const
   return bReturn;
 }
 
-PVR_MENUHOOKS *CPVRClient::GetMenuHooks(void)
+PVR_MENUHOOKS& CPVRClient::GetMenuHooks(void)
 {
-  return &m_menuhooks;
+  return m_menuhooks;
 }
 
 const char *CPVRClient::ToString(const PVR_ERROR error)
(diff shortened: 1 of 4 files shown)


peak3d:
@ksooo usuful refactoring :-) But how do I set mimetype now?

ksooo:
Not sure I understand what you mean. Just do "fileItem.SetFoo" in the newly added function, like you did before.

xbmc/addons/PVRClient.cpp
```diff
@@ -808,6 +808,17 @@ PVR_ERROR CPVRClient::IsPlayable(const CConstPVREpgInfoTagPtr &tag, bool &bIsPla
   return retVal;
 }
 
+void CPVRClient::WriteFileItemProperties(const PVR_NAMED_VALUE *properties, unsigned int iPropertyCount, CFileItem &fileItem)
```
xhaggi: Minor but `Set` fits better as `Write`

xhaggi: And put fileitem param as first?

xbmc/addons/PVRClient.h
```diff
@@ -905,6 +905,11 @@ namespace PVR
     bool LogError(const PVR_ERROR error, const char *strMethod) const;
 
     /*!
+     * @brief Write the given addon properties into given file item's properties.
```
xhaggi: Would be good to add docs for params too.

xbmc/addons/PVRClient.cpp
ksooo: Not really, as all CPVRClient methods copying data back and forth between add-on and Kodi are prefixed "Write". I actually started with "Set" and then changed my code  to fit  the wording already used here.

xhaggi: Ok don't know that

ksooo: Re second comment: no, as I ffollow source target order.

xbmc/addons/PVRClient.h
ksooo: I think the @brief explains it good enough, adding documentation for the parameters is no real value add here. And it is a private method anyway.

xhaggi: Ok I'll shut up now 😄

ksooo: Thanks anyway.

Jalle19 (approved):

PR #12773, opened 2017-09-05
Fix python3 incompatilities in build scripts

MilhouseVH:
After #12705 there were a couple of Python2-only code snippets that remained, which caused Linux builds to fail to install the Python scripts correctly. See https://github.com/xbmc/xbmc/pull/12705#issuecomment-326774773

## Description
Ensure code that is run by the Python3 interpreter is Python3 compatible.

## Motivation and Context
Kodi now installs the Python scripts into the correct location.

Not entirely sure about the Samba change, it may not be required but it probably doesn't do any harm as it will work with Python2 and Python3 (although I'm pretty sure Samba will not build with Python 3 without a _lot_ of work).

## How Has This Been Tested?
I've completed a Linux build, and the Python scripts are now installed correctly:

install_manifest.txt before: http://sprunge.us/STUY
install_manifest.txt after: http://sprunge.us/GLCE

## Types of change

## Checklist:

### DIFF
--- a/cmake/scripts/linux/Install.cmake
+++ b/cmake/scripts/linux/Install.cmake
@@ -178,7 +178,7 @@ install(FILES ${CMAKE_BINARY_DIR}/${CORE_BUILD_DIR}/scripts/${APP_NAME}Config.cm
         COMPONENT kodi-addon-dev)
 
 if(ENABLE_EVENTCLIENTS)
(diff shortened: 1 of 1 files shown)


razzeee (approved):

Rechi (approved):
cmake change is okay
I don't think the samba change is needed. If it is wanted it should be split into a separate patch as autotools41 is the patch which only reverts the upstream autotools removal.

MilhouseVH:
Dropped samba patch.

PR #12774, opened 2017-09-06
[PVR] Fix CPVRManager::OnPlaybackStarted to always clear playing flag…

ksooo:
…s for pvr components.

Fixes a problem described in the forum (=> https://forum.kodi.tv/showthread.php?tid=298462&pid=2642933#pid2642933), where PVR Manager was not aware that it was not playing anymore.


Jalle19 (approved):
Looks reasonable

PR #12775, opened 2017-09-06
[cmake] Add header-only directories

ksooo:
Followup to #12236 - same idea, same rational (see header only directories in Xcode).

@Rechi mind taking a look?


cmake/treedata/common/subdirs.txt
```diff
@@ -1,16 +1,23 @@
 xbmc                                            xbmc
 xbmc/addons                                     addons
 xbmc/addons/binary-addons                       addons_binary-addons
-xbmc/addons/interfaces                          addonsBinaryInterfaces
-xbmc/addons/interfaces/Addon                    addonCallbacks_Addon
+xbmc/addons/interfaces                          addons_interfaces
+xbmc/addons/interfaces/Addon                    addons_interfaces_addon
```
Rechi: any reason for those changes?

Rechi (approved):

cmake/treedata/common/subdirs.txt
ksooo: Cosmetics, to match the real directory names. They were renamed some time ago...

PR #12779, opened 2017-09-07
Allow the binary addon to set redirect limit

rbuehlma:
Sometimes it is preferred not to follow a redirect.

In my case, the pvr provider (teleboy.ch) sets a session cookie right after the login on the http 302 (redirect) response. If curl follows this redirect directly, the future response will not contain the cookie anymore which makes it impossible to retrieve the cookie from the addon (e.g. by using GetFileProperty)

### DIFF
--- a/xbmc/filesystem/CurlFile.cpp
+++ b/xbmc/filesystem/CurlFile.cpp
@@ -427,6 +427,7 @@ CCurlFile::CCurlFile()
   m_seekable = true;
   m_useOldHttpVersion = false;
   m_connecttimeout = 0;
+  m_redirectlimit = 5;
   m_lowspeedtime = 0;
   m_ftpauth = "";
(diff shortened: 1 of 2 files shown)


xbmc/filesystem/CurlFile.cpp
```diff
@@ -427,6 +427,7 @@ CCurlFile::CCurlFile()
   m_seekable = true;
   m_useOldHttpVersion = false;
   m_connecttimeout = 0;
+  m_redirectlimit = 5;
```
peak3d: Is 5 the default redirect limit of CURL ?
We should avoid changing the default behaviour.

rbuehlma: Before it was always set to 5 by kodi.

peak3d (approved):

PR #12780, opened 2017-09-07
Allow binary addon to get multiple values for same property name

rbuehlma:
This is an extension to https://github.com/xbmc/xbmc/pull/12737

There are some properties (probably only HTTP headers) which can have multiple values for the same key. (e.g. set-cookie can be sent multiple times by the server).
With GetFileProperties, it is possible to get a list of all these values.

@peak3d this function could replace File::GetProperty but as GetProperty is simpler to call and probably sufficient in most cases, I think its better to keep GetProperty. What do you think?

(sorry to bug again with "my cookies". but as far as I can see, this is the last remaining issue :-) )

### DIFF
--- a/xbmc/addons/interfaces/Addon/AddonCallbacksAddon.cpp
+++ b/xbmc/addons/interfaces/Addon/AddonCallbacksAddon.cpp
@@ -18,7 +18,7 @@
  *  <http://www.gnu.org/licenses/>.
  *
  */
-
+#include <vector>
 #include "Application.h"
 #include "addons/Addon.h"
 #include "addons/settings/AddonSettings.h"
@@ -60,6 +60,7 @@ CAddonCallbacksAddon::CAddonCallbacksAddon(CAddon* addon)
   m_callbacks->GetLocalizedString = GetLocalizedString;
   m_callbacks->GetDVDMenuLanguage = GetDVDMenuLanguage;
   m_callbacks->FreeString         = FreeString;
+  m_callbacks->FreeStringArray    = FreeStringArray;
 
   m_callbacks->OpenFile           = OpenFile;
   m_callbacks->OpenFileForWrite   = OpenFileForWrite;
@@ -76,7 +77,8 @@ CAddonCallbacksAddon::CAddonCallbacksAddon(CAddon* addon)
   m_callbacks->GetFileChunkSize   = GetFileChunkSize;
   m_callbacks->FileExists         = FileExists;
   m_callbacks->StatFile           = StatFile;
-  m_callbacks->GetFileProperty    = GetFileProperty;
+  m_callbacks->GetFilePropertyValue = GetFilePropertyValue;
+  m_callbacks->GetFilePropertyValues = GetFilePropertyValues;
   m_callbacks->DeleteFile         = DeleteFile;
 
   m_callbacks->CanOpenDirectory   = CanOpenDirectory;
@@ -315,6 +317,15 @@ void CAddonCallbacksAddon::FreeString(const void* addonData, char* str)
   free(str);
 }
 
+void CAddonCallbacksAddon::FreeStringArray(const void* addonData, char** arr, int numElements)
+{
+  for (int i = 0; i < numElements; ++i)
+  {
+    free(arr[i]);
+  }
+  free(arr);
+}
+
 void* CAddonCallbacksAddon::OpenFile(const void* addonData, const char* strFileName, unsigned int flags)
 {
   CAddonInterfaces* helper = (CAddonInterfaces*) addonData;
@@ -505,7 +516,7 @@ int CAddonCallbacksAddon::StatFile(const void* addonData, const char *strFileNam
   return CFile::Stat(strFileName, buffer);
 }
 
-char *CAddonCallbacksAddon::GetFileProperty(const void* addonData, void* file, XFILE::FileProperty type, const char *name)
+char *CAddonCallbacksAddon::GetFilePropertyValue(const void* addonData, void* file, XFILE::FileProperty type, const char *name)
 {
   CAddonInterfaces* helper = (CAddonInterfaces*)addonData;
   if (!helper)
@@ -513,8 +524,35 @@ char *CAddonCallbacksAddon::GetFileProperty(const void* addonData, void* file, X
 
   CFile* cfile = (CFile*)file;
   if (cfile)
-    return strdup(cfile->GetProperty(type, name).c_str());
+  {
+    std::vector<std::string> values = cfile->GetPropertyValues(type, name);
+    if (values.empty()) {
+      return nullptr;
+    }
+    return strdup(values[0].c_str());
+  }
   return nullptr;
+}
+
+char **CAddonCallbacksAddon::GetFilePropertyValues(const void* addonData, void* file, XFILE::FileProperty type, const char *name, int *numValues)
+{
+  CAddonInterfaces* helper = (CAddonInterfaces*)addonData;
+  if (!helper)
+    return nullptr;
+
+  CFile* cfile = static_cast<CFile*>(file);
+  if (!cfile)
+  {
+    return nullptr;
+  }
+  std::vector<std::string> values = cfile->GetPropertyValues(type, name);
+  *numValues = values.size();
+  char **ret = static_cast<char**>(malloc(sizeof(char*)*values.size()));
+  for (int i = 0; i < *numValues; ++i)
+  {
+    ret[i] = strdup(values[i].c_str());
+  }
+  return ret;
 }
 
 bool CAddonCallbacksAddon::DeleteFile(const void* addonData, const char *strFileName)
--- a/xbmc/addons/interfaces/Addon/AddonCallbacksAddon.h
+++ b/xbmc/addons/interfaces/Addon/AddonCallbacksAddon.h
@@ -62,6 +62,7 @@ public:
   static char* GetLocalizedString(const void* addonData, long dwCode);
   static char* GetDVDMenuLanguage(const void* addonData);
   static void FreeString(const void* addonData, char* str);
+  static void FreeStringArray(const void* addonData, char** arr, int numElements);
 
   // file operations
   static void* OpenFile(const void* addonData, const char* strFileName, unsigned int flags);
@@ -79,7 +80,8 @@ public:
   static int GetFileChunkSize(const void* addonData, void* file);
   static bool FileExists(const void* addonData, const char *strFileName, bool bUseCache);
   static int StatFile(const void* addonData, const char *strFileName, struct __stat64* buffer);
-  static char *GetFileProperty(const void* addonData, void* file, XFILE::FileProperty type, const char *name);
+  static char *GetFilePropertyValue(const void* addonData, void* file, XFILE::FileProperty type, const char *name);
+  static char **GetFilePropertyValues(const void* addonData, void* file, XFILE::FileProperty type, const char *name, int *numValues);
   static bool DeleteFile(const void* addonData, const char *strFileName);
   static bool CanOpenDirectory(const void* addonData, const char* strURL);
   static bool CreateDirectory(const void* addonData, const char *strPath);
--- a/xbmc/addons/interfaces/Filesystem.cpp
+++ b/xbmc/addons/interfaces/Filesystem.cpp
@@ -18,6 +18,8 @@
  *
  */
 
+#include <vector>
+
 #include "Filesystem.h"
 #include "addons/kodi-addon-dev-kit/include/kodi/Filesystem.h"
 
@@ -73,7 +75,7 @@ void Interface_Filesystem::Init(AddonGlobalInterface* addonInterface)
   addonInterface->toKodi->kodi_filesystem->get_file_download_speed = get_file_download_speed;
   addonInterface->toKodi->kodi_filesystem->close_file = close_file;
   addonInterface->toKodi->kodi_filesystem->get_file_chunk_size = get_file_chunk_size;
-  addonInterface->toKodi->kodi_filesystem->get_property = get_property;
+  addonInterface->toKodi->kodi_filesystem->get_property_values = get_property_values;
 
   addonInterface->toKodi->kodi_filesystem->curl_create = curl_create;
   addonInterface->toKodi->kodi_filesystem->curl_add_option = curl_add_option;
@@ -504,7 +506,7 @@ int Interface_Filesystem::get_file_chunk_size(void* kodiBase, void* file)
   return static_cast<CFile*>(file)->GetChunkSize();
 }
 
-char* Interface_Filesystem::get_property(void* kodiBase, void* file, int type, const char *name)
+char** Interface_Filesystem::get_property_values(void* kodiBase, void* file, int type, const char *name, int *numValues)
 {
   CAddonDll* addon = static_cast<CAddonDll*>(kodiBase);
   if (addon == nullptr || file == nullptr || name == nullptr)
@@ -535,7 +537,14 @@ char* Interface_Filesystem::get_property(void* kodiBase, void* file, int type, c
     CLog::Log(LOGERROR, "Interface_Filesystem::%s - invalid data (addon='%p', file='%p')", __FUNCTION__, addon, file);
     return nullptr;
   };
-  return strdup(static_cast<CFile*>(file)->GetProperty(internalType, name).c_str());
+  std::vector<std::string> values = static_cast<CFile*>(file)->GetPropertyValues(internalType, name);
+  *numValues = values.size();
+  char **ret = static_cast<char**>(malloc(sizeof(char*)*values.size()));
+  for (int i = 0; i < *numValues; ++i)
+  {
+    ret[i] = strdup(values[i].c_str());
+  }
+  return ret;
 }
 
 void* Interface_Filesystem::curl_create(void* kodiBase, const char* url)
--- a/xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
+++ b/xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
@@ -22,6 +22,7 @@
 #include "AddonBase.h"
 
 #include <map>
+#include <vector>
 
 #if !defined(_WIN32)
   #include <sys/stat.h>
@@ -110,7 +111,7 @@ extern "C"
     double (*get_file_download_speed)(void* kodiBase, void* file);
     void (*close_file)(void* kodiBase, void* file);
     int (*get_file_chunk_size)(void* kodiBase, void* file);
-    char* (*get_property)(void* kodiBase, void* file, int type, const char *name);
+    char** (*get_property_values)(void* kodiBase, void* file, int type, const char *name, int *numValues);
 
     void* (*curl_create)(void* kodiBase, const char* url);
     bool (*curl_add_option)(void* kodiBase, void* file, int type, const char* name, const char* value);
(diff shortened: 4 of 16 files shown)


xbmc/addons/kodi-addon-dev-kit/include/kodi/libXBMC_addon.h
```diff
@@ -478,6 +479,19 @@ namespace ADDON
     }
 
     /*!
+    * @brief Get multiple properties from an open file.
+    * @param file The file to the an properties for
```
ksooo: Seems this needs rephrasing!

xbmc/addons/kodi-addon-dev-kit/include/kodi/libXBMC_addon.h
```diff
@@ -478,6 +479,19 @@ namespace ADDON
...
+    * @param type type of the requested properties.
+    * @param name of the requested properties / can be null.
+    * @param numProperties number of properties returned.
```
ksooo: Please make the first letter after the parameter name capital - applies  to above three lines.

ksooo (changes requested):
Please also add implementation of the new functionality for "new style" API.

rbuehlma:
I have added the new style api and included the review outcome.

Yes, we could simply join the values somehow together into a single string but I do not consider that as a clean approach. What happens if the separator is used in the header value? We would need to encode and later decode the values.

I could replace the GetProperty functions with GetProperties in the Kodi internal code and on the interface simply return the first value of GetProperties if GetProperty is called. That would already reduce the code changes.

Montellese:
@peak3d: you're gonna have a hard time finding a character that you can use as a separator in the joined string that isn't allowed in the HTTP header value.

In our own webserver implementation I'm using an `std::multimap` to manage HTTP headers.

peak3d:
For passing such stuff (addon -> kodi and vice versa, see CCurlFile::GetCookies() we normally URLEncode the strings. Same for postdata, where we use b64 encoding.

Question is if we want to proceed with encoding, then its easy to select an separator (same as we do nowdays: &).

My approach without concatting would be pass an additional parameter through GetProperty (lastFetched, or a magic id) to tell GetProperty where to continue.

For me it feels as code duplicating, and tbh I mislike it.

rbuehlma:
This commit removes some code duplications.

xbmc/addons/kodi-addon-dev-kit/include/kodi/libXBMC_addon.h
```diff
@@ -468,8 +469,8 @@ namespace ADDON
     /*!
     * @brief Get a property from an open file.
     * @param file The file to get an property for
-    * @param type type of the requested property.
-    * @param name of the requested property / can be null.
+    * @param type Type of the requested property.
+    * @param name Of the requested property / can be null.
```
ksooo: This does not make sense. => "@param name Name of the ..."

xbmc/addons/kodi-addon-dev-kit/include/kodi/libXBMC_addon.h
```diff
@@ -478,6 +479,19 @@ namespace ADDON
     }
 
     /*!
+    * @brief Get multiple properties from an open file.
+    * @param file The file to get the properties for
+    * @param type Type of the requested properties.
+    * @param name Of the requested properties / can be null.
```
ksooo: This does not make sense. => "@param name Name of the ..."

ksooo:
I think the concat and base64 encode approach @peak3d suggested makes sense and will actually work just fine.

I do not like the "paging" approach. More complexity, more calls, no real gain, imo.

rbuehlma:
I agree that concat and encode would work. But we would need to do that also for a single value. The addon would then need to decode.

The code duplication in kodi would probably remain as we do not want to encode/decode in code internally but only for the api, right?

Apart from that, why do you think it is nicer to pass a list of values as a concatenated and encoded string instead of using an array?

ksooo:
> Apart from that, why do you think it is nicer to pass a list of values as a concatenated and encoded string instead of using an array?

  I think passing the values in an array/vector is the cleanest approach. My point was that I dislike paging and that I think encode/decode should work. ;-)

rbuehlma:
> I think passing the values in an array/vector is the cleanest approach. My point was that I dislike paging and that I think encode/decode should work. ;-)

Well, then I fully agree. Its ok for me to implement the base64 approach if required.  Who can/will decide here which way we should go?

ksooo:
> Who can/will decide here which way we should go?

 No hard feelings here. I can live with both the encode and the array approach.

peak3d:
vector is fine for me, code samples how to pass vectors can be found in VFS::OpenDirectory

ksooo:
> code samples how to pass vectors can be found in VFS::OpenDirectory

 it's fine like @rbuehlma currently implemented it, imo.

xbmc/filesystem/File.h
```diff
@@ -110,6 +111,7 @@ class CFile
   void Close();
   int GetChunkSize();
   const std::string GetProperty(XFILE::FileProperty type, const std::string &name = "") const;
+  const std::vector<std::string> GetProperties(XFILE::FileProperty type, const std::string &name = "") const;
```
Rechi: `... const override;`

peak3d:
I'm unhappy with the name "GetProperties" as we only retrieve one property (regardless if it produces multiple values). @rbuehlma can you pls. remove my GetProperty() and rename your GetProperty()

If I'm alone with the naming of the function I'm ok with GetProperties. But it sounds not correct at least for me.

ksooo:
> If I'm alone with the naming of the function I'm ok with GetProperties.

If it's single property name - multiple value, then I'd suggest `GetPropertyValues`.

peak3d:
+1 @ksooo  :-)

xbmc/filesystem/File.h
rbuehlma: Its not an override here. The compiler complains if I add "override".

rbuehlma:
Did the rename as proposed by @ksooo. Also did a squash.

xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
```diff
@@ -1547,15 +1548,45 @@ namespace vfs
...
-        std::string strReturn(res);
-        ::kodi::addon::CAddonBase::m_interface->toKodi->free_string(::kodi::addon::CAddonBase::m_interface->toKodi->kodiBase, res);
-        return strReturn;
+        std::vector<std::string> vecReturn;
+        for (int i = 0; i < numValues; i++)
+        {
+          vecReturn.push_back(res[i]);
+          ::kodi::addon::CAddonBase::m_interface->toKodi->free_string(::kodi::addon::CAddonBase::m_interface->toKodi->kodiBase, res[i]);
+        }
+        ::kodi::addon::CAddonBase::m_interface->toKodi->free_string(::kodi::addon::CAddonBase::m_interface->toKodi->kodiBase, (char*)res);
```
rbuehlma: Is it ok the way I free the string-array here? The contained elements are already freed above.

ksooo: Not sure. `free_string` is, as the name implies, for freeing strings (char*), not arbitrary heap allocated objects. @AlwinEsch ?

xbmc/addons/interfaces/Addon/AddonCallbacksAddon.cpp
```diff
@@ -513,10 +514,37 @@ char *CAddonCallbacksAddon::GetFileProperty(const void* addonData, void* file, X
 
   CFile* cfile = (CFile*)file;
   if (cfile)
-    return strdup(cfile->GetProperty(type, name).c_str());
+  {
+    std::vector<std::string> values = cfile->GetPropertyValues(type, name);
+    if (values.empty()) {
+      return nullptr;
+    }
+    strdup(values[0].c_str());
```
ksooo: `return` missing ???

xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
peak3d: Question is if freestring is capable of freeing : char **ret = new char*[*numValues];
I don't believe so, but just check freestring impl. and you know, but I think you have to use an malloc instead new[]

xbmc/addons/interfaces/Addon/AddonCallbacksAddon.cpp
```diff
@@ -513,10 +514,37 @@ char *CAddonCallbacksAddon::GetFileProperty(const void* addonData, void* file, X
...
   return nullptr;
 }
 
+char **CAddonCallbacksAddon::GetFilePropertyValues(const void* addonData, void* file, XFILE::FileProperty type, const char *name, int *numValues)
+{
+  CAddonInterfaces* helper = (CAddonInterfaces*)addonData;
+  if (!helper)
+    return nullptr;
+
+  CFile* cfile = (CFile*)file;
```
ksooo: Please use c++ style casts - here and everywhere in the code your PR introduces. Yes, we use c style casts in old code, but should use only c++ style casts in new code.

xbmc/addons/interfaces/Addon/AddonCallbacksAddon.cpp
```diff
@@ -513,10 +514,37 @@ char *CAddonCallbacksAddon::GetFileProperty(const void* addonData, void* file, X
...
+  if (!cfile)
+  {
+    return nullptr;
+  }
+  std::vector<std::string> values = cfile->GetPropertyValues(type, name);
+  *numValues = values.size();
+  char **ret = new char*[*numValues];
```
ksooo: If you allocate the memory with `new[]`you definitely cannot free it with `free_string`(which internally uses `free`)!!! Do we need another C API function to free arrays? @AlwinEsch any opinion on thias?

xbmc/addons/interfaces/Addon/AddonCallbacksAddon.cpp
```diff
@@ -513,10 +514,37 @@ char *CAddonCallbacksAddon::GetFileProperty(const void* addonData, void* file, X
...
+  for (int i = 0; i < *numValues; i++)
```
ksooo: I've learned that it is always good practice to use prefix operator++, not postfix - for performance reasons...

xbmc/addons/interfaces/Addon/AddonCallbacksAddon.h
```diff
@@ -80,6 +80,7 @@ class CAddonCallbacksAddon
   static bool FileExists(const void* addonData, const char *strFileName, bool bUseCache);
   static int StatFile(const void* addonData, const char *strFileName, struct __stat64* buffer);
   static char *GetFileProperty(const void* addonData, void* file, XFILE::FileProperty type, const char *name);
```
ksooo: For consistency reasons, I would rename all existing single name - single value methods to `GetFilePropertyValue`.

xbmc/addons/interfaces/Filesystem.cpp
```diff
@@ -535,7 +536,14 @@ char* Interface_Filesystem::get_property(void* kodiBase, void* file, int type, c
     CLog::Log(LOGERROR, "Interface_Filesystem::%s - invalid data (addon='%p', file='%p')", __FUNCTION__, addon, file);
     return nullptr;
   };
-  return strdup(static_cast<CFile*>(file)->GetProperty(internalType, name).c_str());
+  std::vector<std::string> values = static_cast<CFile*>(file)->GetPropertyValues(internalType, name);
+  *numValues = values.size();
+  char **ret = new char*[*numValues];
```
ksooo: Note: no `free_string`possible!

xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
```diff
@@ -1547,15 +1548,45 @@ namespace vfs
...
+      int numValues;
+      char **res(::kodi::addon::CAddonBase::m_interface->toKodi->kodi_filesystem->get_property_values(
+        ::kodi::addon::CAddonBase::m_interface->toKodi->kodiBase, m_file, type, name.c_str(), &numValues));
       if (res)
       {
-        std::string strReturn(res);
-        ::kodi::addon::CAddonBase::m_interface->toKodi->free_string(::kodi::addon::CAddonBase::m_interface->toKodi->kodiBase, res);
-        return strReturn;
+        std::vector<std::string> vecReturn;
+        for (int i = 0; i < numValues; i++)
```
ksooo: ++i

xbmc/addons/kodi-addon-dev-kit/include/kodi/libXBMC_addon.h
```diff
@@ -156,6 +156,7 @@ typedef struct CB_AddOn
   bool (*FileExists)(const void* addonData, const char *strFileName, bool bUseCache);
   int (*StatFile)(const void* addonData, const char *strFileName, struct __stat64* buffer);
   char *(*GetFileProperty)(const void* addonData, void* file, XFILE::FileProperty type, const char *name);
```
ksooo: GetFilePropertyValue

xbmc/filesystem/CurlFile.cpp
```diff
@@ -1929,6 +1929,21 @@ const std::string CCurlFile::GetProperty(XFILE::FileProperty type, const std::st
...
+{
+  if (type == FILE_PROPERTY_RESPONSE_HEADER)
+  {
+    return m_state->m_httpheader.GetValues(name);
+  }
+  std::vector<std::string> values;
+  std::string value = GetProperty(type, name);
+  if (!value.empty())
+  {
+    values.push_back(value);
```
ksooo: `emplace _back` is generally to prefer over `push_back`- here and everywhere in your PR.

xbmc/filesystem/CurlFile.h
```diff
@@ -62,6 +62,7 @@ namespace XFILE
       ssize_t Read(void* lpBuf, size_t uiBufSize) override { return m_state->Read(lpBuf, uiBufSize); }
       ssize_t Write(const void* lpBuf, size_t uiBufSize) override;
       const std::string GetProperty(XFILE::FileProperty type, const std::string &name = "") const override;
```
ksooo: GetPropertyValue

xbmc/filesystem/File.h
```diff
@@ -110,6 +111,7 @@ class CFile
   void Close();
   int GetChunkSize();
   const std::string GetProperty(XFILE::FileProperty type, const std::string &name = "") const;
```
ksooo: GetPropertyValue

xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
rbuehlma:
```
void CAddonDll::free_string(void* kodiBase, char* str)
{
  if (str)
    free(str);
}
```
I would say its technically ok.

xbmc/filesystem/FileCache.h
```diff
@@ -61,6 +61,12 @@ namespace XFILE
 
     const std::string GetProperty(XFILE::FileProperty type, const std::string &name = "") const override;
 
+    virtual const std::vector<std::string> GetPropertyValues(XFILE::FileProperty type, const std::string &name = "") const
+    {
+      std::vector<std::string> values;
+      return values;
```
ksooo: No need for a variable here. `return std::vector<std::string>();` is sufficient.

xbmc/filesystem/IFile.h
```diff
@@ -133,6 +134,17 @@ class IFile
...
     return type == XFILE::FILE_PROPERTY_CONTENT_TYPE ? "application/octet-stream" : "";
   };
+
+  virtual const std::vector<std::string> GetPropertyValues(XFILE::FileProperty type, const std::string &name = "") const
+  {
+    std::vector<std::string> values;
+    std::string value = GetProperty(type, name);
+    if (!value.empty())
+    {
+      values.push_back(value);
```
ksooo: emplace_back

xbmc/addons/interfaces/Addon/AddonCallbacksAddon.cpp
```diff
@@ -77,6 +77,7 @@ CAddonCallbacksAddon::CAddonCallbacksAddon(CAddon* addon)
   m_callbacks->FileExists         = FileExists;
   m_callbacks->StatFile           = StatFile;
   m_callbacks->GetFileProperty    = GetFileProperty;
```
ksooo: `GetFilePropertyValue` (see my other comment re the naming)

xbmc/addons/interfaces/Filesystem.cpp
```diff
@@ -18,6 +18,7 @@
  *
  */
 
+#include <vector>
```
ksooo: We usually put a blank line between system headers and "own" headers.

ksooo (changes requested):
not just nitpicking this time...

xbmc/addons/kodi-addon-dev-kit/include/kodi/Filesystem.h
ksooo: No it's not. new[] and free do not fit together!

rbuehlma: Oh, i see, free may only be used with malloc... Changed the allocation to malloc.

ksooo: array vs non-array, c++ vs libc.

ksooo: `malloc`is technically okay, then. 

But my clear -1 for using `free_string`to free the array - for **semantical** reasons. `free_string`'s semantics is to free a single string not an array and the fact that it internally uses `free` which does technically "fit" is nothing more than an implementation detail that can change anytime.

Sorry, but imo we need another API function`free_stringarray`, which does the whole job, freeing the array elements and the array itself.

xbmc/addons/interfaces/Addon/AddonCallbacksAddon.cpp
peak3d: GetFilePropertyValues (!!!)

peak3d:
@rbuehlma this is business as usual, maybe it heads up: If I PR something in @ksooo 's area it's the same.
But, I have learned that his comments give a good feeling if its finally done.

xbmc/addons/interfaces/Addon/AddonCallbacksAddon.cpp
ksooo: No. We have two API functions now. One for single name - single value => GetFilePropertyValue
One for single name - multi values => GetFilePropertyValues

ksooo:
> But, I have learned that his comments give a good feeling if its finally done.

:-)

xbmc/addons/interfaces/Addon/AddonCallbacksAddon.cpp
ksooo: No. We have two API functions now. One for single name - single value, one for single name - multi values. Please look at the full diff and you will see what I mean.

xbmc/addons/kodi-addon-dev-kit/include/kodi/libXBMC_addon.h
```diff
@@ -156,6 +156,7 @@ typedef struct CB_AddOn
...
+  char **(*GetFilePropertyValues)(const void* addonData, void* file, XFILE::FileProperty type, const char *name, int *numPorperties);
```
ksooo: Sorry, to mention this, but ofc this change (inserting a new function in the middle of the function table) makes each and every binary addon that still uses old-style API (like PVR addons) incompatible and therefore an API bump is required.

ksooo (dismissed):
Did I already say "API bump" in this PR?

peak3d:
1.) IIRC we have agreed to only have 1 API function: GetFilePropertyValues() and wanted to remove GetFileProperty instead.
2.) Position of API method should not change anything, they are dyloaded function by function (and not VMT based)

ksooo:
> 1.) IIRC we have agreed to only have 1 API function: GetFilePropertyValues() and wanted to remove GetFileProperty instead.

 Fine by me.

> 2.) Position of API method should not change anything, they are dyloaded function by function (and not VMT based)

 I admit that I always get puzzled about this stuff. So, I shut up at this point and trust in your expertise.

rbuehlma:
> @rbuehlma this is business as usual, maybe it heads up: If I PR something in @ksooo 's area it's the same.

I have no problem getting some feedback :) C/C++ is not really a language I'm used to. Hopefully I will improve.

I personally would prefer to keep GetProperty (aka GetPropertyValue) in the API as in over 90% of the time, only one value is expected and the call to it is much simpler.

I did implement the review outcome except the additional free function for arrays, shall I implement that as well?

I'm not sure about the API bump. Now that we rename GetProperty to GetPropertyValue, I guess a bump is required any way? How would I do that?

rbuehlma:
Added FreeStringArray

xbmc/filesystem/File.h
Rechi: you are right, override is missing in a different file, which overrides this function

xbmc/filesystem/FileCache.h
```diff
@@ -61,6 +61,11 @@ namespace XFILE
 
     const std::string GetProperty(XFILE::FileProperty type, const std::string &name = "") const override;
 
+    virtual const std::vector<std::string> GetPropertyValues(XFILE::FileProperty type, const std::string &name = "") const
```
Rechi: this is the function which needs the `override` specifier

peak3d:
@rbuehlma Ok, go ahead with keeping the previous one.
It#s still / espacially after this discussion / complexity of vector transport my deep opinion that a simple iterator logic would be much nicer / simpler / faster / transparent as this approach here.

rbuehlma:
Created PR against inputstream.adaptive:
https://github.com/peak3d/inputstream.adaptive/pull/67

Let me know when I should squash the commits of this PR.

Rechi:
If this is now the final state, please squash the commits.

ksooo:
> Now that we rename GetProperty to GetPropertyValue, I guess a bump is required any way? How would I do that?

https://github.com/xbmc/xbmc/blob/master/xbmc/addons/kodi-addon-dev-kit/include/kodi/versions.h#L45 => Increase micro version for both ADDON_GLOBAL_VERSION_MAIN and ADDON_GLOBAL_VERSION_MAIN_MIN

https://github.com/xbmc/xbmc/blob/master/xbmc/addons/kodi-addon-dev-kit/include/kodi/versions.h#L70 => Increase micro version for both ADDON_GLOBAL_VERSION_FILESYSTEM  and ADDON_GLOBAL_VERSION_FILESYSTEM _MIN

rbuehlma:
Squash and version bump done.

This change has been tested with pvr.zattoo, pvr.teleboy and inputstream.adaptive.

For pvr.teleboy to work, this PR is required as well: https://github.com/xbmc/xbmc/pull/12779 But I expect less iterations there :)

ksooo:
@MilhouseVH heads up! All binary addons need a rebuild if this PR gets merged.

ksooo (approved):
Looks good. Thanks for your patience.

cooley69:
Stop

ksooo:
What???

rbuehlma:
Amend without changes to trigger jenkins rebuild.

Rechi:
jenkins build this please

PR #12781, opened 2017-09-08
Fixed issues with Win64 build failing to build ffmpeg

nicklauslittle:
## Description

buildmpeg.sh was changed to match the value assigned to ARCH in other scripts/batch files (x86_64). It looks like it was just typoed initially.

The ffmpeg patch fixes an issue where the zlib includes (zconf.h:477) were trying to `#include <unistd.h>` when compiled with Visual Studio. I tracked this down to the ffmpeg configure script creating `#define HAVE_UNISTD_H 0` in config.h. This works fine for ffmpeg, since all its checks use `#if HAVE_UNISTD_H`, but does not work on zlib which uses `#ifdef HAVE_UNISTD_H` (zconf.h:436). The patch simply `#undef HAVE_UNISTD_H` if `HAVE_UNISTD_H` is set to `0`. This will cause `HAVE_UNISTD_H` checks to work with both zlib and ffmpeg.

## Motivation and Context

I wanted to check out the latest Kodi Alpha, and my build environment is Win64 using Visual Studio 2015. I followed the build instructions, but ended up running into errors. I tracked down the source of the different errors and came up with my best attempt at fixing them that worked with the existing build system.

## How Has This Been Tested?

I ran the build from a clean state (git clean -xdf) with only my additional commit and was able to build, run tests, and use Kodi.

## Types of change

## Checklist:

### DIFF
--- a/tools/buildsteps/windows/buildffmpeg.sh
+++ b/tools/buildsteps/windows/buildffmpeg.sh
@@ -19,7 +19,7 @@ do_getFFmpegConfig() {
     FFMPEG_OPTS_SHARED="$FFMPEG_BASE_OPTS $FFMPEG_DEFAULT_OPTS"
   fi
 
-  if [ "$ARCH" == "x86_x64" ]; then
+  if [ "$ARCH" == "x86_64" ]; then
     FFMPEG_TARGET_OS=mingw64
   elif [ "$ARCH" == "x86" ]; then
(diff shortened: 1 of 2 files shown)


Rechi (approved):

afedchin (approved):

Rechi:
jenkins build and merge

PR #12783, opened 2017-09-09
SPDIF: Make it possible to use 352 khz and 384 khz

fritsch:
I could not believe it first, but there are now SPDIF devices out there that accept 384 khz / 32 bit input. See for example the following device: http://www.ca.marantz.com/ca/Products/Pages/ProductDetails.aspx?CatId=HiFiComponents&SubCatId=0&ProductId=HDAMP1

A user on LibreELEC forum verified with ALSA's speaker-test that it works correctly. In order to support this usecase a minimal change to kodi (only Settings and Language) needs to happen.

Note: ALSA sink currently does not support 352 khz, but pulseaudio does. 384 khz both support.

LE Forum thread: https://forum.libreelec.tv/thread/9823-audio-output-device-sampling-rate/?pageNo=1

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -16383,8 +16383,20 @@ msgctxt "#34128"
 msgid "192.0"
 msgstr ""
 
+#. SPDIF max sampling rate
+#: system/settings/settings.xml
+msgctxt "#34129"
+msgid "352.8"
+msgstr ""
+
+#. SPDIF max sampling rate
+#: system/settings/settings.xml
+msgctxt "#34130"
+msgid "384.0"
+msgstr ""
+
 #empty strings from id 34129 to 34200
-#34129-34200 reserved for future use
+#34131-34200 reserved for future use
 
 #: xbmc/PlayListPlayer.cpp
 msgctxt "#34201"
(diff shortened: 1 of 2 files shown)


CiNcH83:
Specs of the Marantz say that 384kHz/32bit is only supported via USB though. Digital inputs (coax, optical) are limited to 192kHz/24bit which is the max for any S/PDIF transmitter I have seen so far.

fritsch:
Better: USB device, that tells it has a SPDIF out and accepts 384 / 32 bit ? Kodi handles it the way ALSA enumerates it and this is SPDIF.

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -16383,8 +16383,20 @@ msgctxt "#34128"
 msgid "192.0"
 msgstr ""
 
+#. SPDIF max sampling rate
+#: system/settings/settings.xml
+msgctxt "#34130"
```
vpeter4: Should be 34129 ?

fritsch: Most likely :-) Was early this morning

fritsch:
@vpeter4: thx - I fixed it up.

fritsch:
image
image

Working as it should. Nice.

PR #12785, opened 2017-09-10
AE: some fixes and limit of max error

FernetMenta:
- limits audio sync error to 1 sec
- fix snyc of planar audio format


FernetMenta:
win32 error unrelated

PR #12791, opened 2017-09-12
[videoplayer] CVideoPlayer::OpenFile: Increase the time player waits for open event

ksooo:
… from 2 secs to 5 secs. Should be sufficient for tuning channels on slow PVR backends.

This should fix trac17557: https://trac.kodi.tv/ticket/17557

@FernetMenta for review (we discussed this change on Slack already).
@MilhouseVH could you please include this PR in your next build? I will ping the bug reporters.

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -701,7 +701,7 @@ bool CVideoPlayer::OpenFile(const CFileItem& file, const CPlayerOptions &options
     params.m_item.SetMimeTypeForInternetFile();
     m_messenger.Put(new CDVDMsgOpenFile(params), 1);
 
-    if (m_openEvent.WaitMSec(2000))
+    if (m_openEvent.WaitMSec(5000))
     {
       if (!m_bStop && !m_bAbortRequest)
         return true;


FernetMenta:
this is only a verification of a theorie but should not be treated as a real fix. OpenFile is called by the main/gui thread. blocking any user interaction and gui updates while waiting for an addon to open a source results in bad user experience.

dvblogic:
@FernetMenta In real life it might take a while until stream starts coming from a tuner. For example, in case when motorized sat dish is used or decryption has to start on an encrypted channel. I have seen the cases when it has taken up to 12-15 seconds.

MilhouseVH:
@ksooo will do.

ksooo:
Very valid point @dvblogic

FernetMenta:
it can take as long as it wants unless open does not block and timeout like in this case.

FernetMenta:
to be more verbose: an input strream is supposed to open instantly without blocking. as long as it does not signal eof to demuxer, the demuxer keeps reading the stream regardless if there is valid data. if you have a motor dish or a thunderstorm between dish and satellite, you just get a black screen but video player will not terminate.

ksooo:
Superseeded by #12792.

@MilhouseVH can you please include #12792 instead of this PR in your next build?

PR #12792, opened 2017-09-12
VideoPlayer: bring up busy dialog if opening stream blocks too long

FernetMenta:
Opening a stream is not supposed to block. The busy dialog keeps the UI alive but finallly addons like dvblogic should fix their issues.

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -2954,8 +2954,6 @@ void CVideoPlayer::SetCaching(ECacheState state)
     m_VideoPlayerVideo->SetSpeed(DVD_PLAYSPEED_PAUSE);
     m_streamPlayerSpeed = DVD_PLAYSPEED_PAUSE;
 
-    m_pInputStream->ResetScanTimeout((unsigned int) CServiceBroker::GetSettings().GetInt(CSettings::SETTING_PVRPLAYBACK_SCANTIME) * 1000);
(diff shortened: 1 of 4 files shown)


xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -719,7 +723,11 @@ bool CVideoPlayer::OpenFile(const CFileItem& file, const CPlayerOptions &options
 
   m_openEvent.Reset();
   Create();
-  CGUIDialogBusy::WaitOnEvent(m_openEvent, g_advancedSettings.m_videoBusyDialogDelay_ms, false);
+  if (!CGUIDialogBusy::WaitOnEvent(m_openEvent, g_advancedSettings.m_videoBusyDialogDelay_ms, true))
```
ksooo: https://github.com/xbmc/xbmc/pull/12792/files#diff-d920d5ea15baa1b7ba5b617ae9b76489R704 uses hardcoded 2 secs for `CGUIDialogBusy::WaitOnEvent`, here you are using advanced setting value. Is there a reason for the different handling?

FernetMenta: yes, here the busy dialog is expected to come up after a short time. the other one is an exceptional path and as such not configurable.

ksooo: Okay.

ksooo (approved):

PR #12797, opened 2017-09-13
VideoPlayer: make OpenFile asynchron

FernetMenta:
Instead of blocking in OpenFile VideoPlayer always returns true and fires OnPlaybackError in case of error.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -3618,6 +3618,14 @@ void CApplication::OnPlayBackStopped()
   g_windowManager.SendThreadMessage(msg);
 }
 
+void CApplication::OnPlayBackError()
+{
+  //@todo Playlists can be continued by calling OnPlaybackEnded instead
+  // open error dialog
+  HELPERS::ShowOKDialogText(CVariant{16026}, CVariant{16027});
+  OnPlayBackStopped();
+}
+
 void CApplication::OnPlayBackPaused()
 {
 #ifdef HAS_PYTHON
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -2482,6 +2482,7 @@ void CVideoPlayer::HandleMessages()
       SAFE_DELETE(m_pCCDemuxer);
       SAFE_DELETE(m_pInputStream);
 
+      m_callback.OnPlayBackStopped();
       m_item = msg.GetItem();
       m_playerOptions = msg.GetOptions();
       Prepare();
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -72,7 +72,6 @@
 #include "utils/StreamUtils.h"
 #include "utils/Variant.h"
 #include "storage/MediaManager.h"
-#include "dialogs/GUIDialogBusy.h"
 #include "dialogs/GUIDialogKaiToast.h"
 #include "utils/StringUtils.h"
 #include "Util.h"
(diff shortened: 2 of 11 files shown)


xbmc/interfaces/legacy/Player.h
```diff
@@ -257,6 +257,20 @@ namespace XBMCAddon
 #ifdef DOXYGEN_SHOULD_USE_THIS
       ///
       /// \ingroup python_PlayerCB
+      /// @brief \python_func{ onPlayBackError() }
+      ///-----------------------------------------------------------------------
+      /// onPlayBackStopped method.
+      ///
+      /// Will be called when user stops Kodi playing a file.
```
wsnipex: c/p error

FernetMenta: thanks! I'll fix it

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -719,19 +710,10 @@ bool CVideoPlayer::OpenFile(const CFileItem& file, const CPlayerOptions &options
   m_item.SetMimeTypeForInternetFile();
 
   m_bAbortRequest = false;
+  m_error = false;
   m_renderManager.PreInit();
 
-  m_openEvent.Reset();
   Create();
-  if (!CGUIDialogBusy::WaitOnEvent(m_openEvent, g_advancedSettings.m_videoBusyDialogDelay_ms, true))
```
ksooo: Please remove m_videoBusyDialogDelay_ms from CAdvancedSettings as it is now unused.

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -694,23 +695,13 @@ bool CVideoPlayer::OpenFile(const CFileItem& file, const CPlayerOptions &options
   // this has to be changed so we won't have to close it.
```
ksooo: Is this comment still valid? CloseFile is now no longer called...

xbmc/cores/VideoPlayer/VideoPlayer.cpp
ksooo: Please remove `#include "dialogs/GUIDialogBusy.h" from this source file as you are no longer using `CGUIDialogBusy` here.

xbmc/Application.cpp
```diff
@@ -3618,6 +3618,14 @@ void CApplication::OnPlayBackStopped()
   g_windowManager.SendThreadMessage(msg);
 }
 
+void CApplication::OnPlayBackError()
+{
+  //@todo Playlists can be continued by calling OnPlaybackEnded instread
```
garbear: spelling: `instead`

FernetMenta: should I add another todo to correct he spelling of the first one? :)

FernetMenta: just kidding

xbmc/interfaces/legacy/Player.h
```diff
@@ -259,9 +259,9 @@ namespace XBMCAddon
       /// \ingroup python_PlayerCB
       /// @brief \python_func{ onPlayBackError() }
       ///-----------------------------------------------------------------------
-      /// onPlayBackStopped method.
+      /// onPlayBackError method.
       ///
-      /// Will be called when user stops Kodi playing a file.
+      /// Will be called when playback stops due to an erorr.
```
popcornmix: erorr -> error

FernetMenta:
updated
great that we could even kill another advanced setting. thanks for review

ksooo:
@MilhouseVH could you please include this PR in your next build, so that we can get feedback from the affected users?

MilhouseVH:
OK

ksooo:
We got very positive feedback from the testers.

ksooo (approved):

PR #12798, opened 2017-09-13
Make sure we don't overwrite path with an empty value. 

razzeee:
## Description

Fixes out of library bookmark deletion on "Set as watched"

While I'm not 100% happy with this approach, this seems to be what we do in this case, see for e.g. https://github.com/xbmc/xbmc/blob/53038bb8ea815d4f19d35324eb67b1bf0a799c6b/xbmc/interfaces/json-rpc/FileItemHandler.cpp#L277

## Motivation and Context

https://forum.kodi.tv/showthread.php?tid=316509

## How Has This Been Tested?

Tested on windows and out of library. I didn't test inside library, but I'm pretty sure it will still work.

## Types of change

## Checklist:


Razzeee:
jenkins build this

Rechi:
jenkins build this please

PR #12799, opened 2017-09-14
VideoPlayer: fix OnPlayBack callbacks

FernetMenta:
next I'll align PAPlayer

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -701,6 +701,8 @@ bool CVideoPlayer::OpenFile(const CFileItem& file, const CPlayerOptions &options
     return true;
   }
 
+  m_callback.OnPlayBackStarted();
+
   m_playerOptions = options;
   m_item = file;
   // Try to resolve the correct mime type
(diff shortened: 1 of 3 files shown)


ksooo (approved):
Nice cleanup.

ksooo:
I runtime tested the change with pvr.hts and pvr.zattoo. Works like a charm as far as I can tell.

@MilhouseVH could you add this PR to your next build, please?

xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -2475,14 +2475,17 @@ void CVideoPlayer::HandleMessages()
     {
       CDVDMsgOpenFile &msg(*static_cast<CDVDMsgOpenFile*>(pMsg));
 
+      CJobManager::GetInstance().Submit([this]() {
+        m_callback.OnPlayBackStarted();
```
Montellese: Was the removed call to `m_callback.OnPlayBackStopped()` a few lines down wrong or should this have been `m_callback.OnPlayBackStopped()` instead of `m_callback.OnPlayBackStarted()`?

ksooo: The call a few lines down was introduced yesterday and was wrong. Calling the started callback instead if stopped seems to be correct. Also moving this dime lines up seems okay to me. 

Additionally, I runtime tested the changes and could not find any suspicious change in behavior.

FernetMenta: the change here is correct. playback is not stopped if we call OpenFile while playing.

PR #12801, opened 2017-09-15
player state fixes

FernetMenta:
more player state fixes. this is still a mess but getting better.


FernetMenta:
build errors unrelated.

PR #12805, opened 2017-09-17
VideoPlayer: smooth motion

FernetMenta:
Implement smooth motion with example for OpenGL:

- instead of skipping a frame, renderer can combine 2 frames into a single one
- if we come close to the next frame but this is not due, renderer can combine the current frame and the next one. Playing 24fps material on a 60Hz display normally results in : 
a a a b b c c c d d 
now it should play like this:
a a a+b b b c c c+d d d

LinuxRendererGL uses blending for combining the frames. Maybe this can be improved.

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
@@ -143,7 +143,6 @@ CLinuxRendererGL::CLinuxRendererGL()
   m_fbo.width = 0.0;
   m_fbo.height = 0.0;
   m_NumYV12Buffers = 0;
-  m_iLastRenderBuffer = 0;
   m_bConfigured = false;
   m_bValidated = false;
   m_clearColour = 0.0f;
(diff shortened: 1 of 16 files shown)


xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
```diff
@@ -530,31 +529,16 @@ void CLinuxRendererGL::RenderUpdate(int index, bool clear, DWORD flags, DWORD al
...
-  }
-  else
+  if (index2 >= 0)
   {
-    if (!Render(flags, index) && clear)
-      ClearBackBuffer();
+    m_iYV12RenderBuffer = index;
+    glEnable(GL_BLEND);
+    glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
+    glColor4f(1.0f, 1.0f, 1.0f, 255.0f / 2.0f / 255.0f);
```
afedchin: the last param can be written `.5f` :)

FernetMenta: updated

Memphiz:
Is this always turned on then?

FernetMenta:
currently yes but it only engages in situations where irregulariteis occur. that is a frames gets skipped or we have to repeat an old frame for some reason.

PR #12806, opened 2017-09-18
LinuxRendererGLES: implement smooth motion

lrusak:
follow up to #12805 

I'm believe this is implemented correctly but I'm not sure I like the effect as it seems that is "blurs" the picture a bit.

looking for comments.

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGLES.cpp
+++ b/xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGLES.cpp
@@ -353,7 +353,10 @@ void CLinuxRendererGLES::Update()
 
 void CLinuxRendererGLES::RenderUpdate(int index, int index2, bool clear, unsigned int flags, unsigned int alpha)
 {
-  m_iYV12RenderBuffer = index;
+  if (index2 >= 0)
+    m_iYV12RenderBuffer = index2;
+  else
+    m_iYV12RenderBuffer = index;
 
   if (!m_bConfigured)
     return;
@@ -403,7 +406,21 @@ void CLinuxRendererGLES::RenderUpdate(int index, int index2, bool clear, unsigne
   if ((flags & RENDER_FLAG_TOP) && (flags & RENDER_FLAG_BOT))
     CLog::Log(LOGERROR, "GLES: Cannot render stipple!");
   else
-    Render(flags, index);
+  {
+    Render(flags, m_iYV12RenderBuffer);
+
+    if (index2 >= 0)
+    {
+      m_iYV12RenderBuffer = index;
+      glEnable(GL_BLEND);
+      glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
+      if (m_pYUVProgShader)
+        m_pYUVProgShader->SetAlpha(0.5f);
+      if (m_pYUVBobShader)
+        m_pYUVBobShader->SetAlpha(0.5f);
+      Render(flags, m_iYV12RenderBuffer);
+    }
+  }
 
   VerifyGLState();
   glEnable(GL_BLEND);


FernetMenta:
As I wrote on #12805 the method of combining two frames can be improved. Blending is probably not the best method

CiNcH83:
I wonder whether this isn't counterproductive for devices that apply FRC further downstream? At least with my Sony however, Motionflow is only being engaged when using MediaCodec Surface.

Just asking as there seems no option to control your smooth motion thingy.

lrusak:
This looks much better with #12809 added but there is still some noticeable blurring happening sometimes.

FernetMenta:
with fps matching refresh rate?

da-anda:
@lrusak as mentioned on Slack - mind picking this up again and maybe even work on the possible improvements Fernet was talking about? We could certainly also add a setting to enable/disable it if too many don't like the effect. But IMO we should get it in sooner than later to get more feedback from users until release of v18. And since we have a vivid userbase of PI users running nightlies, we should get good and fast feedback. Or would you prefer to get feedback upfront the merge by including it in Milhouse nightlies?

markist:
Hi, so if I would want to test this, should i build the current master, use 18 beta or something else?

madshi:
Hey there,

would love to see this feature in Kodi, especially if it will also work on the Vero 4K, which I've just bought for my secondary TV. Unfortunately my secondary TV, although accepting 24fps input, internally applies 3:2 pulldown. So I'm stuck with awful judder right now.

I'm not sure if the feature is currently implemented the same way as I did in madVR, though. Because in #12805 you wrote:

> it only engages in situations where irregulariteis occur.
> that is a frames gets skipped or we have to repeat an
> old frame for some reason.

That is certainly a way to do it, but it will not always be as smooth as it could/should be. Let me try to explain:

If we're extremely lucky, we will have a situation where the refresh rate is exactly 59.940Hz and the movie exactly 23.976fps. If that's the case, we could just ignore the VSync position completely, render frame A twice, then render (A+B)/2, then render B twice, etc. That should give nearly perfect results (well, some blur/ghosting, which is unavoidable with frame blending like this).

```
[60 display][60 display][60 display][60 display][60 display]
[         24 source          ][         24 source          ]
```

If you ask me, I would already be extremely happy with a very simple solution like this. However, if you want to support any frame rate and refresh rate combination, then things get more complicated, of course. In that situation madVR works by simulating how camera exposure works.

```
[refresh 1][refresh 2][refresh 3][refresh 4][refresh 5][refresh 6][refresh 7]
   [     video frame 1     ][     video frame 2     ][     video frame 3     ]
```

In this situation, ideally "refresh 2" should only render "frame1". Then "refresh 3" should mix something like 45% "frame 1" and 55% "frame 2". So basically the exact blend mix factor depends on how long the camera (the display refresh) is exposed to each video frame.

In an extreme case, it could be like this:

```
          [refresh 1][refresh 2][refresh 3][refresh 4][refresh 5][refresh 6]
[ frame 1  ][ frame 2  ][ frame 3  ][ frame 4  ][ frame 5  ][ frame 6  ][ frame 7  ]
```

You can see here that each of the 6 refreshes is exposed to 2 different frames, so all 6 refreshes need to be blended/mixed, although there's no repeated or dropped frame at all.

So the key thing is that if you blend only if "irregularities" occur, then your solution is quite different to what madVR does, and I believe your solution will be noticeably less smooth in the last example shown above.

I'm not sure if you really need to implement a full "any fps with any hz" solution, as outlined above. I mean it would be great if you did, but if it's a lot of work to get it all working nicely, then maybe it would make sense to just implement a very simple A, A, (A+B)/2, B, B upmix logic to make 24fps playback smooth on any display out there? I think 24@60 is the single most important situation where this type of "smooth motion" frame blending is useful. So if you could do just that, it would already be extremely helpful. Well, at least it would solve *my* judder trouble...  ;)

Let's talk about options:

From my experience I can say that some users are very sensitive to the little blur and ghosting which frame blending produces. It's rare, though, most users don't even notice the blur/ghosting at all. It might also depend on the display technology. In any case, since some users are sensitive to the artifacts, I'd recommend to add an option to turn it off, in any case.

If you implement a full blown "any fps with any hz" solution, it might have to be possible for the user to activate frame blending only for specific frame rate vs refresh rate situations. E.g. the user will certainly want to disable it for 24fps@24hz playback. Probably also for 59.940fps@60.000hz playback, because the negative aspects of the algo in that situation may outweight the positive aspects (it's user preference, though). You should also disable it for e.g. 30fps@60hz! I would say that the feature is most useful if the difference between frame rate and refresh rate is:

1) Bigger than a certain threshold (probably bigger than 24.000/23.976 = 1.001). And:
2) Not an even multiply (30fps@60hz). Threshold > 1.001 probably applies here, too.

If you decide to implement only the very simple A, A, (A+B)/2, B, B logic for 24@60 playback, then there should still be a user option to enable/disable this feature, for users who are sensitive to blur/ghosting. But more importantly: Some TVs (e.g. mine) claim to support 24fps playback, and they do, but they apply simple 3:2 pulldown and internally play at 60hz. So for such TVs it's important that Kodi has an option to actually *NOT* use 24hz output, but instead switch to 60hz output with frame blending. This should even work if auto refresh rate switching is turned on.

E.g. you could add an option named something like "my TV sucks at 24fps playback". If turned on, Kodi should internally disable 24hz output and never use it. And should auto switch to 60hz instead and use frame blending. Of course ideally Kodi should switch to 59.940hz for 23.976fps movies and to 60.000hz for 24.000fps movies.

What do you think?

FernetMenta:
Great input but you should not post it here on the GLES PR. Many hw decoders on GLES platforms are not even in exact sync with videoplayer's clock. In addition most of them bypass GL rendering and would need to implement proprietary blending if possible at all.
Reference platform is OpenGL.

madshi:
Ah ok. To be honest, I'm totally lost which part of Kodi is used for which platform. This GLES PR was the only frame blending Kodi reference I found, so I posted here.

Where should I post this instead? My personal interest is in getting this to work for Vero 4K. Trouble is, I can only provide information, I don't have the resources to do actual Kodi development.

PR #12808, opened 2017-09-18
VideoPlayer: frame advance

FernetMenta:
When playback is paused this allows advancing video by a given number of frames.

### DIFF
--- a/xbmc/interfaces/builtins/PlayerBuiltins.cpp
+++ b/xbmc/interfaces/builtins/PlayerBuiltins.cpp
@@ -137,6 +137,20 @@ static int PlayerControl(const std::vector<std::string>& params)
   {
     g_application.StopPlaying();
   }
+  else if (StringUtils::StartsWithNoCase(params[0], "frameadvance"))
+  {
+    std::string strFrames;
+    if (params[0].size() == 12)
+      CLog::Log(LOGERROR, "PlayerControl(frameadvance(n)) called with no argument");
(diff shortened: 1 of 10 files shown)


xbmc/interfaces/builtins/PlayerBuiltins.cpp
```diff
@@ -137,6 +137,20 @@ static int PlayerControl(const std::vector<std::string>& params)
...
+    std::string strFrames;
+    if (params[0].size() == 12)
+      CLog::Log(LOGERROR, "PlayerControl(frameadvance(n)) called with no argument");
+    else if (params[0].size() < 15) // arg must be at least "(N)"
+      CLog::Log(LOGERROR, "PlayerControl(frameadvance(n)) called with invalid argument: \"%s\"", params[0].substr(13).c_str());
+    else
+
+    strFrames = params[0].substr(13);
+    StringUtils::TrimRight(strFrames, ")");
+    float frames = (float) atof(strFrames.c_str());
```
ace20022: atof returns a double casted to float and FrameAdvance wants an int... Also the old cast style is used and you said that negatives doesn't work -- check it here?

FernetMenta: I did not spend much time on this, just c/p from another section. I wanted to give users someting to play with. There's more work to be done from an UI perspective, i.e. not displaying the OSD when paused and using this.

PR #12811, opened 2017-09-19
GLX: fixups

fritsch:
Fixes: http://sprunge.us/VIZA - root cause the visuals did not match.

### DIFF
--- a/xbmc/windowing/X11/GLContextGLX.cpp
+++ b/xbmc/windowing/X11/GLContextGLX.cpp
@@ -153,9 +153,7 @@ bool CGLContextGLX::IsSuitableVisual(XVisualInfo *vInfo)
     return false;
   if (glXGetConfig(m_dpy, vInfo, GLX_BLUE_SIZE, &value) || value < 8)
     return false;
-  if (glXGetConfig(m_dpy, vInfo, GLX_ALPHA_SIZE, &value) || value < 8)
-    return false;
-  if (glXGetConfig(m_dpy, vInfo, GLX_DEPTH_SIZE, &value) || value < 8)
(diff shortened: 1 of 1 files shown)


fritsch:
Removed two commits. The above code is enought to make it work again.

fritsch:
jenkins build this please

xbmc/windowing/X11/GLContextGLX.cpp
```diff
@@ -153,10 +153,6 @@ bool CGLContextGLX::IsSuitableVisual(XVisualInfo *vInfo)
     return false;
   if (glXGetConfig(m_dpy, vInfo, GLX_BLUE_SIZE, &value) || value < 8)
     return false;
-  if (glXGetConfig(m_dpy, vInfo, GLX_ALPHA_SIZE, &value) || value < 8)
-    return false;
-  if (glXGetConfig(m_dpy, vInfo, GLX_DEPTH_SIZE, &value) || value < 8)
```
FernetMenta: why do you drop depth check?

fritsch: Was done by mistake. 8 is enough? For EGL we set it to 16 on certain platforms and to 24 on others. What do we really need?

FernetMenta: it should be 24, aligned with this: https://github.com/xbmc/xbmc/blob/master/xbmc/windowing/X11/GLContextEGL.cpp#L284

fritsch: Thx. made the check for 24

PR #12812, opened 2017-09-19
Open EPG instead of channel window

rbuehlma:
This PR harmonizes the behaviour of the "Guide" button.
Currently, if I push the button while
- I am on the home screen, the EPG is opened
- I am watching a channel, the Channel View is opened
- I am watching a video, nothing happens

I would prefer if always the EPG is shown, even if I am already watching TV.

### DIFF
--- a/system/keymaps/remote.xml
+++ b/system/keymaps/remote.xml
@@ -196,7 +196,7 @@
       <select>OSD</select>
       <title>PlayerProcessInfo</title>
       <info>Info</info>
-      <guide>ActivateWindow(PVRChannelGuide)</guide>
+      <guide>ActivateWindow(TVGuide)</guide>
       <teletext>ActivateWindow(Teletext)</teletext>
       <subtitle>NextSubtitle</subtitle>
       <star>NextSubtitle</star>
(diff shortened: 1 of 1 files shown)


ksooo:
No hard feelings about that. :-)

ksooo:
@xhaggi what do you think?

xhaggi:
Opening the guide window while playing tv or radio makes sense, but would do this only for tv and radio fullscreen window. But have no strong opinion because I use a custom keyboard.xml for ages. So up to you.

xhaggi:
BTW it would be good if we change it in all mapping xmls not only for remotes.

rbuehlma:
> but would do this only for tv and radio fullscreen window

I thought about this as well. But at least on my remote the guide button is a separate one without any other functions attached. Even when watching a video, you might want to open the guide to switch to tv or simple see what is running.

> BTW it would be good if we change it in all mapping xmls not only for remotes.

I checked the other files and keyboard.xml uses <e> which is already always mapped to the EPG. The other mappings currently never open the EPG.

Jalle19 (approved):

xhaggi (approved):
It's fine by me.

zag2me:
Big +1 on this, user interfaces should be consistent.

rbuehlma:
Thank you for merging :) This was the last commit of my customized Kodi build :)

Jalle19:
Thanks for contributing!

PR #12817, opened 2017-09-21
VideoPlayer: send cb notifications on exit via a job

FernetMenta:
Works around a deadlock situation caused by one of Kodi's biggest design flaws. The issue has been around forever but changes like #10893 made it even worse. GUI interactions or usage of AppMessenger require the main thread. When the mainthread originates a synchronous call of actions and one of those actions use AppMessenger, we deadlock:

Example:
- user presses stop playback
- mainthread deletes player object
- delete is synchronous, it has to wait for threads to exit

If one of those threads sends a synchronous message to AppMessenger (SendMsg), the app is dead.

### DIFF
--- a/xbmc/cores/VideoPlayer/VideoPlayer.cpp
+++ b/xbmc/cores/VideoPlayer/VideoPlayer.cpp
@@ -2455,12 +2455,17 @@ void CVideoPlayer::OnExit()
   CFFmpegLog::ClearLogLevel();
   m_bStop = true;
 
-  if (m_error)
-    m_callback.OnPlayBackError();
-  else if (m_bAbortRequest)
-    m_callback.OnPlayBackStopped();
-  else
-    m_callback.OnPlayBackEnded();
+  IPlayerCallback *cb = &m_callback;
+  bool error = m_error;
+  bool abort = m_bAbortRequest;
+  CJobManager::GetInstance().Submit([=]() {
+    if (error)
+      cb->OnPlayBackError();
+    else if (abort)
+      cb->OnPlayBackStopped();
+    else
+      cb->OnPlayBackEnded();
+  }, CJob::PRIORITY_NORMAL);
 }
 
 void CVideoPlayer::HandleMessages()


xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -2455,12 +2455,14 @@ void CVideoPlayer::OnExit()
   CFFmpegLog::ClearLogLevel();
   m_bStop = true;
 
-  if (m_error)
-    m_callback.OnPlayBackError();
-  else if (m_bAbortRequest)
-    m_callback.OnPlayBackStopped();
-  else
-    m_callback.OnPlayBackEnded();
+  CJobManager::GetInstance().Submit([this]() {
```
garbear: Is `this` guaranteed to exist until the job is complete? Or should we only capture the callback, given that the callback is CApplication which has an appropriate lifetime?

FernetMenta: good question. I think we should use copes and no refs passing to the lambda

FernetMenta: updated

nsenica:
@FernetMenta Any chance that this change can be backported to Krypton ? I have a hunch that this would solve problems like the ones decribed in this topic: https://forum.kodi.tv/showthread.php?tid=323394

Since this is not a straightforward change I'd prefer it submited by someone who knows what's going on. Also, I can validate if this really fixes those issues before merging it.

Thanks.

MartijnKaijser:
We're not gonna backport.

nsenica:
Would you mind explaining the reasoning for not backporting ?

MartijnKaijser:
Krypton is end of life.

nsenica:
I find your statement quite disturbing when v17 is the current stable version and v18 doesn't even have a release date set yet.

I'm also not requesting a feature to be backported, I'm requested a deadlock fix to be backported.

MartijnKaijser:
Then i expect you to do complete regression testing on everything that touches this and give a 100% guarantee to every user that it will not cause any issues what so ever for any one out there.

Also you are free to use the software or don't. Additionally we have no obligations to any one to do backporting or even do fixes.
That's the last i'll say about this.

nsenica:
@MartijnKaijser First of all, I'm not demanding anything, I use kodi and at times contribute to kodi as well, so don't take me for a demanding user who doesn't understand anything about how things are done nor understands the risks of backporting code.

I would have been happy to hear, as your first answer that this is a sensitive part of kodi that would require a lot of testing not to cause issues and that was the reason for to supporting a backport. I would have totally understand it and for me the issue was resolved.

Also, even if you (team) have a release date set in mind, it's not public yet, so as far as the public is concerned v17 is the stable version and v18 the alpha version as it is stated in version.txt.

So don't get me wrong, I was only trying to help to fix a problem that I and a couple of other users experience in v17 and I think that this PR could solve it. If it is too troublesome I'm ok with that, I just don't understand where all that the aggressiveness comes from.

MartijnKaijser:
I'm not aggressive. I'm just being very clear and that's the end of it.

PR #12823, opened 2017-09-22
[PVR] Fix channel OSD dialog channel preselection.

ksooo:
This fixes a problem reported in the forum: https://forum.kodi.tv/showthread.php?tid=298462&pid=2649281#pid2649281 ff

As channel switching is asynchronous, the currently playing channel group cannot be obtained safely directly after triggering the channel switch. If channel switching takes some time, there will be not yet a new playing channel  group. Furthermore, resetting `m_group` in `CGUIDialogPVRChannelsOSD::GotoChannel` is not needed at all, as the new channel will always be from the same group as the old channel. 

OTOH, `m_group` must be set when an actual channel group switch occurs in `CGUIDialogPVRChannelsOSD::OnAction`. This was missing.

@xhaggi for review?

@MilhouseVH this needs to be runtime tested by the bug reporters. Can you please include this PR in your next build?

### DIFF
--- a/xbmc/pvr/dialogs/GUIDialogPVRChannelsOSD.cpp
+++ b/xbmc/pvr/dialogs/GUIDialogPVRChannelsOSD.cpp
@@ -123,16 +123,9 @@ void CGUIDialogPVRChannelsOSD::OnDeinitWindow(int nextWindowID)
 {
   if (m_group)
   {
-    if (m_group != GetPlayingGroup())
-    {
-      CGUIWindowPVRBase::SetSelectedItemPath(CServiceBroker::GetPVRManager().IsPlayingRadio(), GetLastSelectedItemPath(m_group->GroupID()));
-      CServiceBroker::GetPVRManager().SetPlayingGroup(m_group);
-    }
-    else
-    {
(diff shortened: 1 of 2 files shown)


ksooo:
Fix has been confirmed working by the bug reporter.

xbmc/pvr/dialogs/GUIDialogPVRChannelsOSD.cpp
```diff
@@ -123,16 +123,9 @@ void CGUIDialogPVRChannelsOSD::OnDeinitWindow(int nextWindowID)
...
-      CServiceBroker::GetPVRManager().SetPlayingGroup(m_group);
-    }
-    else
-    {
-      CGUIWindowPVRBase::SetSelectedItemPath(CServiceBroker::GetPVRManager().IsPlayingRadio(), m_viewControl.GetSelectedItemPath());
-    }
+    CGUIWindowPVRBase::SetSelectedItemPath(m_group->IsRadio(), m_viewControl.GetSelectedItemPath());
 
+    // next OnInitWindow will set the group which is then selceted
     m_group.reset();
```
xhaggi: Sorry for the late response, on holiday atm. Minor typo in this line. Rest looks good.

ksooo: oops.

PR #12825, opened 2017-09-23
[addons] fix addon install if not present before

AlwinEsch:
On last changes comes the problem that a addon install who was not present
before has not worked.

This comes from a try to disable a addon who is not present and brought a
false.

With them becomes before checked that a addon is present, otherwise is this
step ignored.

Related to https://github.com/xbmc/xbmc/pull/12473#pullrequestreview-64744902

## Types of change

## Checklist:

### DIFF
--- a/xbmc/addons/AddonInstaller.cpp
+++ b/xbmc/addons/AddonInstaller.cpp
@@ -614,7 +614,9 @@ bool CAddonInstallJob::DoWork()
   if (!Install(installFrom, m_repo))
     return false;
 
-  if (!CAddonMgr::GetInstance().LoadAddon(m_addon->ID()))
+  // Load new installed and if successed replace defined m_addon here with new one
+  if (!CAddonMgr::GetInstance().LoadAddon(m_addon->ID()) ||
+      !CAddonMgr::GetInstance().GetAddon(m_addon->ID(), m_addon))
   {
     CLog::Log(LOGERROR, "CAddonInstallJob[%s]: failed to reload addon", m_addon->ID().c_str());
     return false;
--- a/xbmc/addons/AddonInstaller.cpp
+++ b/xbmc/addons/AddonInstaller.cpp
@@ -604,7 +604,7 @@ bool CAddonInstallJob::DoWork()
   // run any pre-install functions
   ADDON::OnPreInstall(m_addon);
 
-  if (!CAddonMgr::GetInstance().UnloadAddon(m_addon))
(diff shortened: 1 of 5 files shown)


AlwinEsch:
@tamland is this fix OK for you, that we bring it in asap?

tamland:
This check needs to be move inside `UnloadAddon` (I.e it should return true if the addon isn't installed). Otherwise there are race conditions

AlwinEsch:
Is changed.

Think to replace the `CAddonMgr::GetInstance()` with `CServiceBroker::GetAddonMgr()` on all places on another request

stefansaraev:
works for me. thanks @AlwinEsch @tamland

AlwinEsch:
Install test has worked, but there seems still a fault with start of a new installed addon.

If I have installed YouTube and start it are the strings from them not visible. Also brings it a Python error after select of a field. 

If Kodi is restarted does it work, there seems the loading of new installed addon not work!

stefansaraev:
yep same behaviour here. I thought it was unrelated but indeed completely reverting #12473 makes those strings work again...

AlwinEsch:
Have added a second commit who fix the fault with not usable addon after install.

The fault itself was coming after the change of the function value `AddonPtr& addon` to `const std::string& addonId` on the LoadAddon (old ReloadAddon).

If all commits are reverted and only this with change to std::string is done, comes the fault.

The function has changed the addon class on CAddonInstaller.

xbmc/addons/AddonInstaller.cpp
```diff
@@ -614,12 +614,15 @@ bool CAddonInstallJob::DoWork()
   if (!Install(installFrom, m_repo))
     return false;
 
-  if (!CAddonMgr::GetInstance().LoadAddon(m_addon->ID()))
+  AddonPtr addon = CAddonMgr::GetInstance().LoadAddon(m_addon->ID());
```
tamland: Please don't misuse null to mean 'error'. That's what true/false is for. If AddonInstaller needs an instance of the new addon it can call CAddonMgr::GetAddon afterwards

AlwinEsch: I change it, thought to hold the `CAddonMgr::GetAddon` calls smaller.

AlwinEsch:
Is changed to use GetAddon

tamland (approved):
Thanks! Merge when ready

PR #12829, opened 2017-09-24
VideoPlayer: expose stereo mdoe to dataCache

FernetMenta:
@da-anda @popcornmix  StereoscopicManager is broken. In order to fix it it needs to pull stereoMode from DataCacheCore

### DIFF
--- a/xbmc/cores/DataCacheCore.cpp
+++ b/xbmc/cores/DataCacheCore.cpp
@@ -116,6 +116,20 @@ std::string CDataCacheCore::GetVideoPixelFormat()
   return m_playerVideoInfo.pixFormat;
 }
 
+void CDataCacheCore::SetVideoStereoMode(std::string mode)
+{
+  CSingleLock lock(m_videoPlayerSection);
+
+  m_playerVideoInfo.stereoMode = mode;
+}
+
+std::string CDataCacheCore::GetVideoStereoMode()
+{
+  CSingleLock lock(m_videoPlayerSection);
+
+  return m_playerVideoInfo.stereoMode;
+}
+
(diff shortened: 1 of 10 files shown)


FernetMenta:
@popcornmix couls you please adapt OMXPlayer

da-anda:
thanks for the headsup. @popcornmix as I haven't touched any C++ in likely over a year, would you mind adjusting stereoscopic manager accordingly?

@FernetMenta as the stream info might not be available when the "playback started" message is fired by core (which stereoscopic manager listens too to injects it's logic), is there by now a better way to be notified about stream changes?

FernetMenta:
@da-anda StereoscopicManager has even been wrong in this regard. OnPlaybackStarted was fired when a new file was opened but not on every stream change.

da-anda:
well, it was the best non-intrusive way I could find back then. Are there better suited messages/signals around now? I stopped following the development over a year ago, when my worker thread got hold of a lock which is preventing my fun thread from filling the sparetime buffer, thus the buffer drained. Seems I have to refactor cMyLife application ;)

FernetMenta:
@da-anda currently you can only poll this info. What does StereoscopicsManager do when it detects a change?

da-anda:
sorry for not responding earlier. IIRC, up until now, StereoscopicsManager checked on playback start of a file if it is a stereoscopic file or not, and depending on this information and the currently active rendering mode (disabled, monoscopic, some stereoscopic mode) and the users preferences (always switch between regular mode and stereoscopic on mixed playback // keep current mode // always use mode of video // ask using a dialog) it will set the desired stereoscopic mode in the graphicsContext. So it's mostly a class to handle user preferences and configure related components accordingly

PR #12832, opened 2017-09-24
[ios/tvos] - added latest apple devices to allow retina support

Memphiz:
As the topic says. This adds some new apple devices to our scale factor/retina list.

### DIFF
--- a/xbmc/platform/darwin/DarwinUtils.mm
+++ b/xbmc/platform/darwin/DarwinUtils.mm
@@ -66,6 +66,8 @@ enum iosPlatform
   iPadMiniGSMCDMA,
   iPadMiniWIFI,
   AppleTV2,
+  AppleTV4,
+  AppleTV4K,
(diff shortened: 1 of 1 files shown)


xbmc/platform/darwin/DarwinUtils.mm
```diff
@@ -66,6 +66,8 @@
   iPadMiniGSMCDMA,
   iPadMiniWIFI,
   AppleTV2,
+  AppleTV4,
+  AppleTV4K,
   iPhone4,            //from here on list devices with retina support (e.x. mainscreen scale == 2.0)
```
MrMC: AppleTV4K will report [[UIScreen mainScreen] nativeScale] == 2 when running under 4K display.

Memphiz: Thx for the info ... I find this odd though as UHD could just be treated as usual native resolution. I treat it as scale 1 for now - as 4K support needs some more tweaks anyway I guess and this is master branch anyway ...

PR #12834, opened 2017-09-24
[music] SortByTrackNumber as default sort for "songs"

DaveTBlake:
The most natural default sort order for viewing songs is track number. But on a clean installation "songs" nodes are currently sorted by file name, and this catches a number of users out. Many get very confused because the name my include a track number, so the odd sorting only shows up when they come to a multiple disc album e.g. have two track 01s next to each other.

A trivial change @jjd-uk  requested, and makes sense to me.

### DIFF
--- a/xbmc/view/ViewStateSettings.cpp
+++ b/xbmc/view/ViewStateSettings.cpp
@@ -47,7 +47,7 @@ CViewStateSettings::CViewStateSettings()
 {
   AddViewState("musicnavartists");
   AddViewState("musicnavalbums");
-  AddViewState("musicnavsongs");
+  AddViewState("musicnavsongs", DEFAULT_VIEW_LIST, SortByTrackNumber);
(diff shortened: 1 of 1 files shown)


Razzeee:
Just one question, the label for each row is still how you want it in this case?

jjd-uk:
Seems to operate as I'd expect sorting by track where there is one then falling back to name sorting where there isn't.

So +1 from me.

jjd-uk (approved):
Seems to operate as I'd expect sorting by track where there is one then falling back to name sorting where there isn't.

So +1 from me.

DaveTBlake:
>Just one question, the label for each row is still how you want it in this case?
Yes @Razzeee I think so

razzeee (approved):

Razzeee:
jenkins build and merge

PR #12835, opened 2017-09-25
X11: fix getting visual

FernetMenta:
fix segfault on NVidia systems


ksooo:
I runtime tested it the other day and comforting it working.

PR #12837, opened 2017-09-25
cleanup internal nfs vfs in favour of vfs.nfs

lrusak:
This is to go with https://github.com/notspiff/vfs.nfs/pull/1

I've tested this and it works fine. The only thing that I think needs adding is a dialog when someone selects nfs from the drop down and vfs.nfs isn't install it notifies them or prompts to install it. If anyone has ideas about the best way to accomplish that (including the other vfs add-ons when they are ready) that would be great.


lrusak:
All green baby! :1st_place_medal:

Rechi:
https://github.com/xbmc/xbmc/blob/45874ebd3e57a897b1435d9bcfecbce258eb9926/project/BuildDependencies/scripts/0_package.target-win32.list#L28 https://github.com/xbmc/xbmc/blob/dd0dc92289281098eb28187eb134242ca1f3716f/project/BuildDependencies/scripts/0_package.target-x64.list#L26
are also obsolete and have to be removed

Razzeee:
At the moment we have to deliver all binary addons with our build. So for now, it should only prompt to activate the addon and not to install it (that's for later).

lrusak:
@Rechi done

@Razzeee sounds reasonable :smile:

wsnipex:
note that the addons needs packaging for our PPA

lrusak:
@xhaggi can you do the debian packaging for vfs.nfs? I see that you did it for vfs.rar

xhaggi:
@lrusak sure, will create a PR for it.

lrusak:
I think this is ready to go, Please take one last look

Also depends on, https://github.com/xbmc/repo-binary-addons/pull/94

Rechi (dismissed):
vfs.nfs won't compile in the current state on windows.

lrusak:
@Rechi can you look into it?

lrusak:
What's the best way to test this on windows? Can I push a test commit that moves the binary add-on repo to my repo so I can test with jenkins?

xhaggi:
I've add appveyor ci config to vfs.nfs repo which should do a windows build, but @notspiff needs to register the repo on appveyor.

notspiff:
isn't it better that ci is consolidated under one kodi account?

MartijnKaijser:
@notspiff we can but then we need to move/fork the addons

notspiff:
yes, i realized, hence the edit. of course, if you want to move the add-ons to your account, that is in general all good with me, but then i will not take maintenance any longer.

notspiff:
@xhaggi seems a bit suboptimal that the appveyor job depends on kodi-binary-addons. in this case, it bites us bad since it's not yet in there.

xhaggi:
@notspiff how should we fix it? All other binary addons use the same ci config.

notspiff:
using ADDON_SRC_DIR, like in the build instructions should do no?

Rechi:
If you set `ADDON_SRC_DIR` it still checks if the addon is in the binary repository.

notspiff:
it should check against that directory i thought, so you'd generate the text file on-the-fly in there.

Rechi:
no just uses the source from that directory, but still does the check of valid addon id and platform

notspiff:
ADDONS_DEFINITIONS_DIR was the one i was on about.

xhaggi:
`ADDON_DEFINITION_DIR` but where to point it to, relative to `KODI/xbmc/cmake/addons`?

notspiff:
you *generate* a dir with vfs.nfs/vfs.nfs.txt before kicking off the cmake command, pointing it to that dir.

xhaggi:
done for both vfs.nfs and vfs.rar.

https://github.com/notspiff/vfs.nfs/pull/7
https://github.com/notspiff/vfs.rar/pull/21

xhaggi:
@Rechi could you please take a look at the appveyor build error for vfs.nfs https://ci.appveyor.com/project/notspiff/vfs-nfs/build/master-14fc2c0-2017-09-30

lrusak:
@Rechi is this all good for Windows now?

Rechi:
@lrusak first question is, have you overlooked https://github.com/lrusak/xbmc/pull/5? My point is either we do it correctly (remove all special nfs cases from core) or don't move it to a binary addon.
`vfs.nfs` builds correctly for Windows (32 & 64 bit) but someone has to runtime test it.

Rechi:
@lrusak please squash the commits
I'll trigger a test build afterwards.

lrusak:
@Rechi should I push a temp commit to change the addon-repo to https://github.com/lrusak/repo-binary-addons/tree/vfs ???

Rechi:
@lrusak yes please

lrusak:
@Rechi did we ever get a windows test build?

Rechi:
Yes I triggered a build for all platforms shortly after you pushed the temp commit, the build name contains `20171007-3bd7377`.

CvH:
`KodiSetup-20171007-3bd7377394-HEAD-x64.exe` works for my nfs shares.
But even if the nfs addon is disabled I have an NFS source available at the source chooser. 
This is not intuitiv as nothing points you to install/activate some addon.

Rechi:
@lrusak this needs a rebase

Memphiz:
did anyone try libnfs-2.0.0? there was timeout handling added (currently if no nfs server is present - trying to connect to it freezes indefinitly).

lrusak:
We use 2.0.0 in LibreELEC so any recent master build will include it.

xhaggi:
@lrusak needs rebase + remove the temp commit because vfs.rar is already part of the repo.

xbmc/network/GUIDialogNetworkSetup.cpp
```diff
@@ -219,9 +219,6 @@ void CGUIDialogNetworkSetup::InitializeSettings()
          { true,  true,  true,  true, false,  80,   "rss", 20304}};
 
  m_protocols.insert(m_protocols.end(), defaults.begin(), defaults.end());
-#ifdef HAS_FILESYSTEM_NFS
-  m_protocols.emplace_back(Protocol{true, false, false, false, true, 0, "nfs", 20259});
```
lrusak: is this correct?

xbmc/network/ZeroconfBrowser.cpp
```diff
@@ -60,9 +60,7 @@ CZeroconfBrowser::CZeroconfBrowser():mp_crit_sec(new CCriticalSection),m_started
 #endif
   AddServiceType("_ftp._tcp.");
   AddServiceType("_webdav._tcp.");
-#ifdef HAS_FILESYSTEM_NFS
-  AddServiceType("_nfs._tcp.");  
-#endif// HAS_FILESYSTEM_NFS
+  AddServiceType("_nfs._tcp.");
```
lrusak: is this correct?

xbmc/network/GUIDialogNetworkSetup.cpp
notspiff: Yes but i have to finalize my work first. add-on shoukd carry the info and then we would append addon info here

xbmc/network/ZeroconfBrowser.cpp
notspiff: This needs to be addonized as well

xbmc/network/GUIDialogNetworkSetup.cpp
notspiff: see https://github.com/xbmc/xbmc/pull/13038

xbmc/network/ZeroconfBrowser.cpp
notspiff: see https://github.com/xbmc/xbmc/pull/13037

xbmc/Util.cpp
```diff
@@ -1500,15 +1500,13 @@ bool CUtil::MakeShortenPath(std::string StrInput, std::string& StrOutput, size_t
 
 bool CUtil::SupportsWriteFileOperations(const std::string& strPath)
 {
-  // currently only hd, smb, nfs and dav support delete and rename
+  // currently only hd, smb and dav support delete and rename
```
notspiff: this removes functionality.

xbmc/FileItem.cpp
```diff
@@ -2610,7 +2605,6 @@ void CFileItemList::StackFolders()
       // 1. rars and zips may be on slow sources? is this supposed to be allowed?
       if( !item->IsRemote()
         || item->IsSmb()
-        || item->IsNfs()
```
notspiff: this removes functionality.

xbmc/cores/VideoPlayer/Edl.cpp
```diff
@@ -99,8 +99,7 @@ bool CEdl::ReadEditDecisionLists(const std::string& strMovie, const float fFrame
    * network share.
    */
   if ((URIUtils::IsHD(strMovie)  || 
-       URIUtils::IsSmb(strMovie) || 
-       URIUtils::IsNfs(strMovie))         &&
+       URIUtils::IsSmb(strMovie))         &&
```
notspiff: this removes functionality.

xbmc/storage/MediaManager.cpp
```diff
@@ -214,11 +214,9 @@ void CMediaManager::GetNetworkLocations(VECSOURCES &locations, bool autolocation
     locations.push_back(share);
 #endif
 
-#ifdef HAS_FILESYSTEM_NFS
     share.strPath = "nfs://";
```
notspiff: this cannot be unconditionally added, it should be unconditionally removed.

notspiff (changes requested):
In general the IsNfs based code should be kept. we do not want to remove stuff such as write support, we only want to drop the dependency from mainline.

Rechi:
As we move nfs to an addon it isn't correct to do some checks for it in core (`IsNfs`).
The slow sources, `CreateDirectory`, `SupportsWriteFileOperations` and can handle Edl decisions have to be handled by the specific addon used for the given path.

notspiff:
That is what i meant.. I realize my formiulation could be read as that.

xbmc/storage/MediaManager.cpp
notspiff: see https://github.com/xbmc/xbmc/pull/13070

xbmc/Util.cpp
notspiff: see https://github.com/xbmc/xbmc/pull/13071

xbmc/FileItem.cpp
notspiff: on closer inspection this can safely be removed. we do not want to encourage nfs over wan and nfs will be handled by the IsOnLAN check.

xbmc/cores/VideoPlayer/Edl.cpp
notspiff: see https://github.com/xbmc/xbmc/pull/13072

candrews:
Just curious on the status of this one, I'm interested in seeing it in Kodi 18

lrusak:
superseded by #14188

PR #12839, opened 2017-09-26
[addons] remove the CAddonMgr::GetInstance() usage

AlwinEsch:
## Description
Do this direct with 'CServiceBroker::GetAddonMgr()'.

Also becomes the class copy construction prevent changed to C++11 who generates a compile fault instead of a linker fault.

## Types of change

## Checklist:


garbear:
Changes look good. Only desire of mine is to have "ServiceBroker.h" be included near the end to respect alphabetization and putting folders before files.

AlwinEsch:
Alphabetization was for me with them on base folder first :)

I think to do this on other request, is more to do with this.

ksooo:
@AlwinEsch just in case: Please do not reshuffle the includes for PVR files. I recently reordered them  in a consistent way and would like them to keep this order. Thanks.

PR #12841, opened 2017-09-27
OpenGL: move away from fixed function pipeline

FernetMenta:
This is first iteration of upgrading OpenGL to >= 3.2. I moved all shaders used by OpenGL to a separate folder with GLSL version. This allows keeping backwards compatibility on systems that don't support GLSL 1.5 that comes with OpenGL 3.2

OSX: define environment variable KODI_GL_PROFILE_3_2 to get OpenGL core profile 3.2


notspiff:
i tested this, it seems to work fine on linux/nvidia legacy drivers. i'm surprised that it doesn't seem to break add-ons using fixed pipeline functionality. i thought they were exclusive starting in gl3.2 ?

FernetMenta:
GL 3.2 dropped fixed pipeline completely. If you request profile 3.2 (on Intel) by setting MESA_GL_VERSION_OVERRIDE to 3.2, you get a black sceen even with this PR. GL 3.0 is last version that supports fixed. I don't know how nvidia drivers manage compatibility.

MilhouseVH:
> i tested this, it seems to work fine on linux/nvidia legacy drivers
ION2 is now using only software decode (vc-1, h264 etc.) with this PR - presumably this is expected?

FernetMenta:
vdpau should work again with added commit

MilhouseVH:
Thanks, hardware acceleration is working again with ION2.

xbmc/cores/VideoPlayer/VideoRenderers/LinuxRendererGL.cpp
```diff
@@ -558,61 +557,140 @@ void CLinuxRendererGL::ClearBackBuffer()
...
-  glColor4f(m_clearColour, m_clearColour, m_clearColour, 1.0f);
   glDisable(GL_BLEND);
-  glBegin(GL_QUADS);
+
+  struct Svertex
+  {
+    float x,y,z;
+  };
+  Svertex vertices[24];
+  Svertex vertex;
```
Rechi: unused variable

FernetMenta:
next step done. on OSX I can play video with OpenGL core profile 3.2 by defining environment variable 3.2.

Rechi:
Not sure if this the following is expected in the current state.
I only have a green background with enabled hardware decoding instead of seeing the video on OSX.
After disabled hardware decoding the video is displayed correctly.

FernetMenta:
right, I haven't looked into hw accels so far. works with latest update

Memphiz:
What is the reason for not enabling gl3.2 on OS X in general? I guess there is no guarantee that everything running OS X 10.8 (our current deployment target) supports it?

FernetMenta:
Addons (visualizations) need to adapt to 3.2 too. If you are using a viz that still uses fixed function pipeline, you can run Kodi in legacy OpenGL.

FernetMenta:
jenkins build this please

MilhouseVH:
Latest version of this PR crashes on ION2:

http://sprunge.us/BScR
http://sprunge.us/PWPj

Let me know if you need a debug-enabled crashlog.

FernetMenta:
@MilhouseVH updated, could you please try again

MilhouseVH:
Thanks, that's working fine again.

xbmc/guilib/MatrixGLES.cpp
```diff
@@ -326,10 +326,7 @@ void CMatrixGL::PrintMatrix(void)
 
 void CMatrixGLStack::Load()
 {
-#ifdef HAS_GL
-  glMatrixMode(m_type);
```
Rechi: `m_type` is now unused at least on osx

MilhouseVH:
On both a NUC (Skylake i5) and ION2, with this PR there's some sort of alpha/blending issue in Estuary:

WIthout this PR:
s1

With this PR:
s2

xbmc/guilib/MatrixGLES.cpp
FernetMenta: This entire file is a hack: MatrixGLES and used by OpenGL. Load() is empty now and can be removed but this would reslts in more changes in GLES only code. I think it is best to do this in a separate PR.

FernetMenta:
@MilhouseVH should be fixed now

FernetMenta:
jenkins build this please

MilhouseVH:
Thanks, that looks good now on both the NUC and ION2.

MilhouseVH:
> Addons (visualizations) need to adapt to 3.2 too. If you are using a viz that still uses fixed function pipeline, you can run Kodi in legacy OpenGL.

Perhaps this is the reason why the "Dim" screensaver is crashing kodi on the ION2 (NUC is OK)? Or maybe it's something else...

Crashlog: https://pastebin.com/raw/8cttCMaq

Rechi:
jenkins build this please

FernetMenta:
stupid jenkins builds outdated sources all the time

FernetMenta:
@MilhouseVH I just forgot another method to migrate. Could you please try again.

Rechi:
somehow jenkins now wasn't even triggered, trying again
jenkins build this please

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

MilhouseVH:
> @MilhouseVH I just forgot another method to migrate. Could you please try again.

Thanks, latest PR is OK on ION2 now.

FernetMenta:
jenkins build this please

FernetMenta:
@garbear would be graet if you could adapt RP to OpenGL >= 3.2 in the near future. I think it is the only component left that still uses legacy OpenGL.

stefansaraev:
fyi. with this pr amlogic is broken - black screen (no gui at all) after initial splash screen. gui is back to normal once screensaver (dim) is activated few minutes later and everything works then.

FernetMenta:
strange. are other GLES sysystems affected too?

da-anda:
I suppose some of the stereoscopic rendering modes that currently use the GL stipple feature need to be dropped/converted to shaders as well? Out of memory line interlaced 3D mode and checkerboard use the stipple feature of OpenGL

popcornmix:
Same problem on Pi after this PR.
All text is blank until the screensaver (dim) kicks in.
Then all is fine after waking screesnaver.

lrusak:
Same with android. tested on oct 6th, nightly build

FernetMenta:
is the issue a black screen or no fonts?

popcornmix:
For me it's missing fonts. I still see the blue-ish rectangles that make up the Estuary skin.

popcornmix:
Screenshots and logs in this forum post:
https://forum.kodi.tv/showthread.php?tid=298461&pid=2652767#pid2652767

stefansaraev:
for me, it is black screen (using confluence) - home window not rendered at all. but some widgets are visible:

- progress bar while importing epg ftom tvheadend is visible at startup
- osd is partialy visible (controls ok but no art and text) after playing video via kodi-send: https://i.imgur.com/e3g3YJE.png vs https://i.imgur.com/ZB1CQTY.png

poor man's bisect revealed fb00118 is the first bad commit

FernetMenta:
weird, there is nothing on the GLES path but this single line of code:
https://github.com/FernetMenta/xbmc/blob/c578f71f0c63f74a44a067da12524a3589626dd7/xbmc/guilib/GUIFontTTFGL.cpp#L108

can you try to comment this line and try again?

popcornmix:
Commenting out ``glActiveTexture(GL_TEXTURE0);`` didn't help.

FernetMenta:
this should fix it: https://github.com/xbmc/xbmc/pull/12890

Guilouz:
With this PR I have issue on my Intel NUC NUC5i5RYH, tested with Stock Estuary and my Estuary MOD

See my post here : https://forum.kodi.tv/showthread.php?tid=298462&pid=2653307#pid2653307

MilhouseVH:
As reported by @Guilouz there continues to be some sort of alpha/transparency issue.

Build #1004 on Skylake NUC (does not include this PR):

s1

Build #1005 on Skylake NUC (does include this PR):

s2

In addition to the transparency difference, as can be seen in the screenshots above the GUI appears to be darker *with* this PR suggesting maybe a difference in the way limited colour range is being handled?

xrandr --verbose from #1004 (limited colour range enabled in Kodi, corresponding to first screenshot): http://sprunge.us/VeeF
xrandr --verbose from #1005 (limited colour range enabled in Kodi, corresponding to second screenshot): http://sprunge.us/EfiS

Display settings from #1005:
s3

FernetMenta:
@MilhouseVH could you please attach screenshots taken with stock kodi. I would like to reproduce.

FernetMenta:
@MilhouseVH do oyu observe the alpha issue only with limited colour or also with full range?

MilhouseVH:
> @MilhouseVH do oyu observe the alpha issue only with limited colour or also with full range?

Yes, transparency only seems to be an issue when "Use limited colour range (16-235)" is ENABLED.

With limited colour range DISABLED there is no transparency issue - the dialogue is opaque.

I'm currently building Ubuntu in an effort to find a simple way to reproduce.

FernetMenta:
I think I see the error

MilhouseVH:
Yes, I can reproduce on Ubuntu with Estuary Mod skin but only when enabling limited colour range.

stefansaraev:
@FernetMenta do you remember my issue with the renderloop deadlocks after starting playback and switching to fullscreen too fast, that I reported to you maybe 6 months ago? it was magicaly fixed by an unrelated rework (I dont remember the exact PR) done by you, but it's now back! :)

here's what I used as a workaround in confluence to keep the renderloop busy: http://sprunge.us/JGDI

xbmc/windowing/osx/WinSystemOSX.mm
```diff
@@ -1252,18 +1262,32 @@ static void DisplayReconfigured(CGDirectDisplayID display,
...
     NSOpenGLPFAScreenMask, (NSOpenGLPixelFormatAttribute)CGDisplayIDToOpenGLDisplayMask(displayID),
     (NSOpenGLPixelFormatAttribute)0
   };
 
-  NSOpenGLPixelFormat* pixFmt = [[NSOpenGLPixelFormat alloc] initWithAttributes:fsattrs];
+  NSOpenGLPixelFormat* pixFmt = [[NSOpenGLPixelFormat alloc] initWithAttributes:fsattrs_gl3];
+  if (!getenv("KODI_GL_PROFILE_3_2"))
+    pixFmt = [[NSOpenGLPixelFormat alloc] initWithAttributes:fsattrs];
+
+  pixFmt = [[NSOpenGLPixelFormat alloc] initWithAttributes:fsattrs];
```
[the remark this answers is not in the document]
FernetMenta: good catch, thanks a lot!
fixed here: https://github.com/xbmc/xbmc/pull/12899/commits/325ab1cbcbaf677c87cb84208a973e1fa697a675

FernetMenta:
@stefansaraev yes, I remember it. It was triggered by calling swapBuffers without having rendered anything.
Are you saying this PR brought it back?

stefansaraev:
> Are you saying this PR brought it back?

yep. 100% reproducible on amlogic s905 (wetek play 2). I can't test on other gles platform.

FernetMenta:
the problem does only show when playing videos, right? is hq scaling involved?

stefansaraev:
the problem is only visible here when playing live tv (there is a noticeable delay in playback start so I can switch to fullscreen before actual playback) with amlogic hw decoders enabled.  I can't not reproduce with any video from my library.

I also can't reproduce with software decoding (playback starts faster when hw decoder is disabled).

changing videoplayer.hqscalers value to 0% does not make a difference.

FernetMenta:
This PR was not supposed to change anything on the GLES path. It changed FBO which is used by GLES too. This is why I asked if hq scaling makes a difference (but aml renderer won'tuse this anyway, right?)
I don't see any GLES related changes here. Would be graet if you could manage it to bisect.

stefansaraev:
will do, but not today, I'll open a trac ticket tomorrow :)

stefansaraev:
@FernetMenta I was wrong. it is not this PR. 

tried to bisect but gave up after few hours rebuilding and testing. seems the bug was re-introduced somewhere here: https://github.com/xbmc/xbmc/compare/16b9b7cc4d98f846996fcd6af0db46a2bc18b7cd...6f90513683589d4162f655262160a145fb423a11

I was stuck with some pvr api / binary addons nonsense, other bugs came into play and am not going to continue bisecting. I'd keep a workaround in confluence instead. sorry for the noise :)

FernetMenta:
can you try setting this to false: https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/Process/ProcessInfo.cpp#L55

stefansaraev:
yes. that helps. now I have black screen instead of video (audio is playing fine), but the deadlock is gone. I can get out of fullscreen (to confluence home screen by pressing "back" on the remote) and then picture appears immediately.

EDIT: can't we just render black the moment playback is started? that would solve the issue.

FernetMenta:
slack?

stefansaraev:
I dont have access to kodi slack. only LibreELEC's but I dont see you there.

da-anda:
@stefansaraev the Kodi IRC channels (#kodi, #kodi-dev, ...) are bridged to Slack if that helps, so even if IRC doesn't show certain members online, they might still read your messages on Slack). We can also add you as a guest account on Slack if it helps you tracking down bugs or in any other way

PR #12842, opened 2017-09-27
[utils] add file extension provider to enclose extension handling (AS + binary addons)

xhaggi:
This PR adds a new `FileExtensionProvider` to enclose extension providing from `AdvancedSettings` and binary addons like `vfs`, `encoder`.

@notspiff mind taking a look too.

## Motivation and Context

If vfs.rar is not installed or disabled Kodi should not support `.rar` files. They should not to be scanned by our info scanners, because it results in various issues if we only add `file.rar` to the library instead of `rar://path-to-rar/file.mkv` (dupe entries, not able to playback etc.)

## How Has This Been Tested?

Runtime tested on MacOS.

## Types of change

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -93,6 +93,7 @@
 #include "settings/SkinSettings.h"
 #include "guilib/LocalizeStrings.h"
 #include "utils/CPUInfo.h"
+#include "utils/FileExtensionProvider.h"
 #include "utils/log.h"
 #include "utils/SeekHandler.h"
 #include "ServiceBroker.h"
@@ -2545,7 +2546,7 @@ void CApplication::OnApplicationMessage(ThreadMessage* pMsg)
       else
         pathToUrl = URIUtils::CreateArchivePath("rar", CURL(pMsg->strParam), "");
 
-      CUtil::GetRecursiveListing(pathToUrl.Get(), items, g_advancedSettings.GetPictureExtensions(), XFILE::DIR_FLAG_NO_FILE_DIRS);
+      CUtil::GetRecursiveListing(pathToUrl.Get(), items, CServiceBroker::GetFileExtensionProvider().GetPictureExtensions(), XFILE::DIR_FLAG_NO_FILE_DIRS);
       if (items.Size() > 0)
       {
         pSlideShow->Reset();
@@ -2580,7 +2581,7 @@ void CApplication::OnApplicationMessage(ThreadMessage* pMsg)
 
     CFileItemList items;
     std::string strPath = pMsg->strParam;
-    std::string extensions = g_advancedSettings.GetPictureExtensions();
+    std::string extensions = CServiceBroker::GetFileExtensionProvider().GetPictureExtensions();
     if (pMsg->param1)
       extensions += "|.tbn";
     CUtil::GetRecursiveListing(strPath, items, extensions);
--- a/xbmc/ServiceManager.cpp
+++ b/xbmc/ServiceManager.cpp
@@ -39,6 +39,7 @@
 #include "interfaces/python/XBPython.h"
 #include "pvr/PVRManager.h"
 #include "settings/Settings.h"
+#include "utils/FileExtensionProvider.h"
 
 using namespace KODI;
 
@@ -111,6 +112,8 @@ bool CServiceManager::InitStageTwo(const CAppParamParser &params)
 
   m_gameRenderManager.reset(new RETRO::CGUIGameRenderManager);
 
+  m_fileExtensionProvider.reset(new CFileExtensionProvider());
+
   init_level = 2;
   return true;
 }
@@ -173,6 +176,7 @@ void CServiceManager::DeinitStageThree()
 
 void CServiceManager::DeinitStageTwo()
 {
+  m_fileExtensionProvider.reset();
   m_gameRenderManager.reset();
   m_peripherals.reset();
   m_inputManager.reset();
@@ -311,6 +315,11 @@ CFavouritesService& CServiceManager::GetFavouritesService()
 CInputManager& CServiceManager::GetInputManager()
 {
   return *m_inputManager;
+}
+
+CFileExtensionProvider& CServiceManager::GetFileExtensionProvider()
+{
+  return *m_fileExtensionProvider;
 }
 
 // deleters for unique_ptr
--- a/xbmc/settings/AdvancedSettings.cpp
+++ b/xbmc/settings/AdvancedSettings.cpp
@@ -25,11 +25,6 @@
 #include <string>
 #include <vector>
 
-#include "addons/AudioDecoder.h"
-#include "addons/BinaryAddonCache.h"
-#include "addons/IAddon.h"
-#include "addons/ImageDecoder.h"
-#include "addons/binary-addons/BinaryAddonBase.h"
 #include "Application.h"
 #include "ServiceBroker.h"
 #include "filesystem/File.h"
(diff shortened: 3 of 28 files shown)


notspiff:
no time right now, but .001 is also used on vcd's so i'd be a bit careful with removing that one (though vcd isn't exactly state of the art these days).

xhaggi:
Okay thanks, will re-add .001 if that is used by vcd.

xbmc/Application.cpp
```diff
@@ -1211,6 +1211,9 @@ bool CApplication::Initialize()
   CServiceBroker::GetRepositoryUpdater().Start();
   CServiceBroker::GetServiceAddons().Start();
 
+  CAddonMgr::GetInstance().Events().Subscribe(this, &CApplication::OnAddonEvent);
```
AlwinEsch: A `CServiceBroker::GetAddonMgr().Events().Unsubscribe(this);` is missing.

xbmc/settings/AdvancedSettings.cpp
```diff
@@ -1427,29 +1431,62 @@ void CAdvancedSettings::setExtraLogLevel(const std::vector<CVariant> &components
...
+  for (auto const type : ADDON_TYPES)
   {
-    result += '|';
-    result += addonInfo->Type(ADDON_IMAGEDECODER)->GetValue("@extension").asString();
+    std::vector<std::string> extensions;
+    BinaryAddonBaseList addonInfos;
+    CServiceBroker::GetBinaryAddonManager().GetAddonInfos(addonInfos, true, type);
+    for (auto addonInfo : addonInfos)
+    {
+      std::string ext = addonInfo->Type(type)->GetValue(ADDON_VFS ? "@extensions" : "@extension").asString();
```
AlwinEsch: there is by `ADDON_VFS ? ...` a `... == type` missing

AlwinEsch: Maybe possible to change the `@extension` to `@extensions` on the types?

xbmc/Application.cpp
AlwinEsch: Can `Subscribe` be done from CAdvancedSettings itself?

AlwinEsch (changes requested):

xbmc/Application.cpp
xhaggi: As is init before app is init, tried before, so no.

xbmc/settings/AdvancedSettings.cpp
xhaggi: Ahh thx will address it asap

xbmc/Application.cpp
xhaggi: We could expose a method that do subscribe/unsubscribe but need to call it in `CApplication` let me know what you prefer.

xbmc/settings/AdvancedSettings.cpp
xhaggi: >Maybe possible to change the @extension to @extensions on the types?

Not sure why this differs for various types. But would be good to unify it.

notspiff: one of these things that happens by accident. sync it up!

notspiff:
looks good to me, apart from what has been discussed.

xbmc/Application.cpp
tamland: I prefer this to be removed from both CApplication and CAdvancedSettings, and rather made a separate 'service' that holds the extension info. It can then read extensions from both AS and addons. Others that needs the combined list reads from it. 

URIUtils::RemoveExtension should be moved here too to get rid of the dependency on globals.

xhaggi: This will end up in a massive refactor because we use AS's `GetXXXExtensions` everywhere.

tamland: It's just a matter of replacing `g_advancedSettings.GetXXXExtensions` with a call to another singleton. If that require too much refactoring, just have `GetXXXExtensions` forward the calls as a temp solution; that's not an issue.

The important part is to start replacing these ugly mutable globals with things that have a clear lifetime.

xhaggi: Sure. I'm a bit lazy, because my spare time is limited ;)

xhaggi:
@tamland here we go, started from scratch and introduced a new `ExtensionManager` as you prefer, which enclode the magic for both AS + binary addon extensions.

xhaggi:
BTW ignore the first commits, only the last is relevant, because this PR depends on #12843.

xbmc/utils/ExtensionManager.cpp
```diff
@@ -0,0 +1,142 @@
...
+    SetAddonExtensions(type);
+  }
+}
+
+void CExtensionManager::SetAddonExtensions(const TYPE& type)
+{
+  std::vector<std::string> extensions;
+  BinaryAddonBaseList addonInfos;
+  CServiceBroker::GetBinaryAddonManager().GetAddonInfos(addonInfos, true, type);
+  for (auto addonInfo : addonInfos)
```
AlwinEsch: You can use a `for (const auto& addonInfo : addonInfos)` to prevent a class copy

AlwinEsch:
Only the small thing the others looks very good for me and approved from my side.

xbmc/utils/ExtensionManager.cpp
xhaggi: sure

xhaggi: done

xbmc/utils/ExtensionManager.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+  m_addonExtensions.insert(make_pair(type, StringUtils::Join(extensions, "|")));
+}
+
+void CExtensionManager::OnAddonEvent(const AddonEvent& event)
+{
+  if (typeid(event) == typeid(AddonEvents::Disabled) ||
+      typeid(event) == typeid(AddonEvents::Unload) ||
+      typeid(event) == typeid(AddonEvents::Enabled) ||
+      typeid(event) == typeid(AddonEvents::Load))
```
tamland: As mentioned in the other pr: don't use Load/Unload. This is for something else entirely

xhaggi: ok removed load/unload

garbear:
LGPL?

xhaggi:
@garbear what do you mean?

notspiff:
License header in new file..

xhaggi:
Ahh c/p from another file. Will change it.

MartijnKaijser:
Might need to "fix" that other file?

xhaggi:
Puuh not sure anymore from which one I've copied the header ;)

Rechi:
`git grep 'GNU Lesser General' -- xbmc` lists 19 files.

xhaggi:
Suee sure but would go out of scope.of this PR

ksooo:
@xhaggi I would use "FilenameExtension" for naming here, not just "Extension". The latter is imo too generic. There can be other things that have extensions. Just my 2 ct, though.

xhaggi:
@ksoo sure. Thought about a good class name, but end up with this one :) .. Not sure if we really need the `Manager` suffix .. I would go with `FilenameExtensionService`.

garbear:
Or `FileExtensionService` for brevity

xhaggi:
@garbear sure .. what you think about `FileExtensionProvider`

garbear:
If the choice is between `manager`, `service` and `provider`, I think I like manager the best.

Thanks for this, it looks like something I could relocate my game extension stuff to. BTW, for game extensions I actually use extensions from remote non-downloaded add-ons. This allows for "just-in-time installation", where Kodi can install the appropriate game client without any effort on the user's part.

Also, HAHA I think the conflict is my fault ;)

xhaggi:
Updated. Addressed the header and @garbear I decided to use `FileExtensionProvider` because you can't manage anything. It only provides various filename extensions.

xhaggi:
@alanwww1 @tamland mind taking a look again.

xbmc/ServiceManager.cpp
```diff
@@ -188,6 +191,7 @@ void CServiceManager::DeinitStageTwo()
   m_binaryAddonManager.reset();
   m_addonMgr.reset();
   m_Platform.reset();
+  m_fileExtensionProvider.reset();
```
garbear: This needs to go at the beginning of the function. Deinit in the opposite order as init to fix cross dependency issues

xhaggi: Thx, will change the order

xhaggi:
any objections?

jenkins build this please

AlwinEsch (approved):

xhaggi:
jenkins build this please again

xhaggi:
jenkins build this please

PR #12843, opened 2017-09-27
[addons] binary cache update is not triggered on enable/disable of addons

xhaggi:
Enabling/disabling of VFS addons does not trigger update of binary addon caches. This PR will fix the issue.

@mkortstiege mind taking a look.

## How Has This Been Tested?

Runtime tested on MacOS

## Types of change

### DIFF
--- a/xbmc/addons/AddonEvents.h
+++ b/xbmc/addons/AddonEvents.h
@@ -25,27 +25,35 @@ namespace ADDON
 {
   struct AddonEvent
   {
+    std::string id;
+    explicit AddonEvent(std::string id) : id(std::move(id)) {};
     virtual ~AddonEvent() = default;
   };
 
   namespace AddonEvents
   {
+    /**
+     * Emitted after the add-on has been enabled.
+     */
     struct Enabled : AddonEvent
     {
-      std::string id;
-      explicit Enabled(std::string id) : id(std::move(id)) {}
+      explicit Enabled(std::string id) : AddonEvent(std::move(id)) {}
     };
 
+    /**
+     * Emitted after the add-on has been disabled.
+     */
     struct Disabled : AddonEvent
     {
-      std::string id;
-      explicit Disabled(std::string id) : id(std::move(id)) {}
+      explicit Disabled(std::string id) : AddonEvent(std::move(id)) {}
     };
 
+    /**
+     * Emitted after the add-on's metadata has been changed.
+     */
     struct MetadataChanged : AddonEvent
     {
-      std::string id;
-      explicit MetadataChanged(std::string id) : id(std::move(id)) {}
+      explicit MetadataChanged(std::string id) : AddonEvent(std::move(id)) {}
     };
 
     /**
@@ -54,8 +62,7 @@ namespace ADDON
      */
     struct ReInstalled: AddonEvent
     {
-      std::string id;
-      explicit ReInstalled(std::string id) : id(std::move(id)) {}
+      explicit ReInstalled(std::string id) : AddonEvent(std::move(id)) {}
     };
 
     /**
@@ -63,25 +70,23 @@ namespace ADDON
      */
     struct UnInstalled : AddonEvent
     {
-      std::string id;
-      explicit UnInstalled(std::string id) : id(std::move(id)) {}
+      explicit UnInstalled(std::string id) : AddonEvent(std::move(id)) {}
     };
 
     /**
-     * @deprecated Use Enabled, ReInstalled and UnInstalled instead.
+     * Emitted after the add-on has been loaded.
      */
-    struct InstalledChanged : AddonEvent {};
-
     struct Load : AddonEvent
     {
-      std::string id;
-      explicit Load(std::string id) : id(std::move(id)) {}
+      explicit Load(std::string id) : AddonEvent(std::move(id)) {}
     };
 
+    /**
+     * Emitted after the add-on has been unloaded.
+     */
     struct Unload : AddonEvent
     {
-      std::string id;
-      explicit Unload(std::string id) : id(std::move(id)) {}
+      explicit Unload(std::string id) : AddonEvent(std::move(id)) {}
     };
-  };
-};
+  }
+}
--- a/xbmc/addons/AddonManager.cpp
+++ b/xbmc/addons/AddonManager.cpp
@@ -654,6 +654,12 @@ bool CAddonMgr::GetAddon(const std::string &str, AddonPtr &addon, const TYPE &ty
   return false;
 }
 
+bool CAddonMgr::HasType(const std::string &id, const TYPE &type)
+{
+  AddonPtr addon;
+  return GetAddon(id, addon, type, false);
+}
+
 bool CAddonMgr::FindAddons()
 {
   bool result = false;
--- a/xbmc/addons/AddonManager.cpp
+++ b/xbmc/addons/AddonManager.cpp
@@ -692,8 +692,6 @@ bool CAddonMgr::FindAddons()
     tmp.clear();
     m_database.GetBlacklisted(tmp);
     m_updateBlacklist = std::move(tmp);
-
-    m_events.Publish(AddonEvents::InstalledChanged());
   }
 
   return result;
@@ -758,7 +756,6 @@ bool CAddonMgr::LoadAddon(const std::string& addonId)
   }
 
   m_events.Publish(AddonEvents::ReInstalled(addon->ID()));
-  m_events.Publish(AddonEvents::InstalledChanged());
   CLog::Log(LOGDEBUG, "CAddonMgr: %s successfully loaded", addon->ID().c_str());
   return true;
 }
@@ -768,7 +765,6 @@ void CAddonMgr::OnPostUnInstall(const std::string& id)
   CSingleLock lock(m_critSection);
   m_disabled.erase(id);
   m_updateBlacklist.erase(id);
-  m_events.Publish(AddonEvents::InstalledChanged());
   m_events.Publish(AddonEvents::UnInstalled(id));
 }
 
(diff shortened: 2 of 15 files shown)


AlwinEsch (commented):
Can be maybe complicated if e.g. VFS addon do something.

The ReInstalled and UnInstalled becomes called after the addon is changed or deleted.

Here where I change for PVR:
https://github.com/AlwinEsch/kodi/pull/35/commits/82b0d7aabbf629debaa2c539deae629d9e208296

xhaggi:
Not sure what you mean with do something. IMO if an addon becomes disabled or enabled the cache should be updated to reflect the current state. ATM if you enabled the installed vfs.rar addon you need to restart Kodi to use the addon. That's not very nice and fixed by this changes.

AlwinEsch:
Thats complete true, I mean if a vfs addon e.g. unzip is unpacking a zip file and becomes deinstalled. The CVFSAddonCache becomes informed after the addon is deleted.

xhaggi:
Ok let me know what I have to change and I'll do that.

AlwinEsch:
Have looked, I think you can use your way (a VFS addon is not always running) and implement also something by use of https://github.com/xbmc/xbmc/blob/master/xbmc/addons/Addon.h#L57. To check before the addon is changed.

Have currently not a direct idea how this `IsInUse` can be used from VFS, maybe you?

xhaggi:
Not really. We could go with this and take an eye on it. Let me know if we could merge it in.

AlwinEsch:
Ok from my side

xbmc/addons/VFSEntry.cpp
```diff
@@ -70,8 +70,13 @@ VFSEntryPtr CVFSAddonCache::GetAddonInstance(const std::string& strId, TYPE type
 
 void CVFSAddonCache::OnEvent(const AddonEvent& event)
 {
-  if (typeid(event) == typeid(AddonEvents::InstalledChanged))
+  if (typeid(event) == typeid(AddonEvents::Enabled) ||
+      typeid(event) == typeid(AddonEvents::Disabled) ||
+      typeid(event) == typeid(AddonEvents::ReInstalled) ||
+      typeid(event) == typeid(AddonEvents::UnInstalled))
+  {
     Update();
```
AlwinEsch: Only maybe add a:
```cpp
  ADDON::AddonPtr addon;
  if (CServiceBroker::GetAddonMgr().GetAddon(enableEvent->id, addon, ADDON_VFS, false))
    Update();
```
to prevent not wanted update if e.g. YouTube addon is changed.

xhaggi: Sure will add it.

xhaggi:
@AlwinEsch pushed some improvements after our discussion.

The first commit move the addon id to `AddonEvent` struct so that we don't not need to cast to get the addon id out of an event. It also drops the special `InstalledChanged` event in favor of a new event `Initilized` which is only triggered in `FindAddons`. This will reduce the events triggered in case of `UnInstalled` + `ReInstalled` where you also publish the `InstalledChanged` before. I've adjust all places where `InstalledChanged` was used.

The second commit adds a convenient method to check if an addon has a given type.

Hope this will make things a bit easier.

AlwinEsch:
From my side perfect, to have addon id defined on the base structure makes it much more easy to identify the addon type.

Thanks a lot!

xbmc/addons/VFSEntry.cpp
xhaggi: This won't work for event `UnInstalled` because `GetAddon()` will always return false. We need to update the cache in that case.

xhaggi:
> From my side perfect, to have addon id defined on the base structure makes it much more easy to identify the addon type.

Cool. Force-pushed everything. Mind take a look again and finally sign-off ;)

xbmc/addons/AddonEvents.h
```diff
@@ -25,27 +25,34 @@ namespace ADDON
...
+    explicit AddonEvent(std::string id) : id(std::move(id)) {};
     virtual ~AddonEvent() = default;
   };
 
   namespace AddonEvents
   {
+    /**
+     * Emitted after initilizing add-ons.
+     */
+    struct Initilized : AddonEvent
```
AlwinEsch: A small typo should be "Initialized"

xbmc/addons/AddonManager.cpp
```diff
@@ -654,6 +654,12 @@ bool CAddonMgr::GetAddon(const std::string &str, AddonPtr &addon, const TYPE &ty
   return false;
 }
 
+bool CAddonMgr::HasType(const std::string &id, const TYPE &type)
+{
+  AddonPtr addon;
+  return GetAddon(id, addon, type, false);
+}
+
 bool CAddonMgr::FindAddons()
```
AlwinEsch: The FindAddons is also called on `bool CAddonMgr::LoadAddon(const std::string& addonId)`. Maybe add a `bool CAddonMgr::FindAddons(bool initialFind)` or prevent the `m_events.Publish(AddonEvents::Initilized());` from this function complete.

xbmc/addons/VFSEntry.cpp
```diff
@@ -70,8 +70,18 @@ VFSEntryPtr CVFSAddonCache::GetAddonInstance(const std::string& strId, TYPE type
 
 void CVFSAddonCache::OnEvent(const AddonEvent& event)
 {
-  if (typeid(event) == typeid(AddonEvents::InstalledChanged))
+  if (typeid(event) == typeid(AddonEvents::Initilized) ||
+      typeid(event) == typeid(AddonEvents::UnInstalled))
```
AlwinEsch: The `AddonEvents::UnInstalled` also to the check below.

xbmc/peripherals/bus/virtual/PeripheralBusAddon.cpp
```diff
@@ -332,9 +332,11 @@ void CPeripheralBusAddon::GetDirectory(const std::string &strPath, CFileItemList
...
 void CPeripheralBusAddon::OnEvent(const ADDON::AddonEvent& event)
 {
-  if (typeid(event) == typeid(ADDON::AddonEvents::Enabled) ||
+  if (typeid(event) == typeid(ADDON::AddonEvents::Initilized) ||
+      typeid(event) == typeid(ADDON::AddonEvents::Enabled) ||
       typeid(event) == typeid(ADDON::AddonEvents::Disabled) ||
-      typeid(event) == typeid(ADDON::AddonEvents::InstalledChanged))
+      typeid(event) == typeid(ADDON::AddonEvents::ReInstalled) ||
+      typeid(event) == typeid(ADDON::AddonEvents::UnInstalled))
     UpdateAddons();
```
AlwinEsch: Add also a `if (CServiceBroker::GetAddonMgr().HasType(event.id, ADDON_PERIPHERALDLL))` check.

xbmc/addons/AddonEvents.h
xhaggi: Oh thx, will address it.

xbmc/addons/AddonManager.cpp
xhaggi: Then we should move `m_events.Publish(AddonEvents::Initilized());` to `Init`

xbmc/addons/VFSEntry.cpp
xhaggi: As I said `UnInstalled` can't be used in conjunction with `HasType` because the addon is gone

xbmc/peripherals/bus/virtual/PeripheralBusAddon.cpp
xhaggi: Sure, will add the same here

tamland:
>It also drops the special InstalledChanged event in favor of a new event Initilized which is only triggered in FindAddons. This will reduce the events triggered in case of UnInstalled + ReInstalled where you also publish InstalledChanged before. I've adjust all places where InstalledChanged was used.

Don't. It is deprecated for a reason: you do not need it. This event is just a old quick solution. If you listen to the new Enabled, ReInstalled, and UnInstalled, it's redundant.

xhaggi:
@tamland I noticed that and there was one place (`FindAddons`) where `InstalledChanged` is published that does not fit to any of the events `Enabled`, `ReInstalled` and `UnInstalled`. If we don't need an event to be published in that case I'll drop the `Initialized` event.

xhaggi:
@AlwinEsch merged your changes into the branch. Runtime tested and works for me too.
@tamland we dropped `InstalledChanged` now.

xhaggi:
jenkins build this please

AlwinEsch (approved):

xbmc/addons/AddonEvents.h
```diff
@@ -54,34 +62,31 @@ namespace ADDON
...
-
     struct Load : AddonEvent
     {
-      std::string id;
-      explicit Load(std::string id) : id(std::move(id)) {}
+      explicit Load(std::string id) : AddonEvent(std::move(id)) {}
     };
 
+    /**
+     * Emitted after the add-on has been unloaded.
```
tamland: This is wrong. Both Load and Unload are before (that's the entire point). Actually, they should not even be part of AddonEvent. That's just more stuff from #12473 that wasn't completed.

tamland:
Come on guys. Adding Load/Unload to the existing event handler together with the async ones is not how this was supposed to be used. I tried to demonstrate show how to it safely in https://github.com/xbmc/xbmc/commit/8bd7213d2ec081717f3c3f4fc0ed43eea843e019. Neither CPeripheralBusAddon, CVFSAddonCache, CBinaryAddonManager is 'fixed'. It now just contains race conditions instead, so TBC. when you notice binary addons randomly failing to be detected.

xhaggi:
@tamland ok will push a PR that removes the use of Load/Unload in CPeripheralBusAddon, CVFSAddonCache, CBinaryAddonManager

PR #10850, opened 2016-11-03
[guilib] Change System.HasModalDialog to System.HasActiveModalDialog …

BigNoid:
…and add System.HasVisibleModalDialog

Needs Estuary to be updated. That commit will be added when v18 is open for business.
## Motivation and Context
```System.HasModalDialog``` ignored the window close animations. Similar to ```Window.IsVisible``` and ```Window.IsActive``` we can now choose to use a visible condition that either respects or ignores the window close animations of modal dialogs.

## How Has This Been Tested?
Add a control with either visibility check and open/close a modal dialog. With ```System.HasActiveModalDialog``` the control will be instantly visible on closing. With ```System.HasVisibleModalDialog``` only when the modal dialog has finished the animation.

## Types of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [X] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [X] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #12845, opened 2017-09-27
[cosmetics] update skin docs for PR10850

ronie:
things changed, docs didn't. let's fix.

ref: https://github.com/xbmc/xbmc/pull/10850


BigNoid:
Thx!

PR #12846, opened 2017-09-28
[addons] don't use event load/unload in conjunction with async events

xhaggi:
Follow-up after #12843 

@tamland ping

### DIFF
--- a/xbmc/addons/BinaryAddonCache.cpp
+++ b/xbmc/addons/BinaryAddonCache.cpp
@@ -101,14 +101,9 @@ AddonPtr CBinaryAddonCache::GetAddonInstance(const std::string& strId, TYPE type
 
 void CBinaryAddonCache::OnEvent(const AddonEvent& event)
 {
-  if (typeid(event) == typeid(AddonEvents::ReInstalled) ||
-      typeid(event) == typeid(AddonEvents::UnInstalled))
-  {
-    Update();
-  }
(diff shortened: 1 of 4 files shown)


AlwinEsch (commented):
Problem is always that there the parts becomes informed after the addon is changed.

Currently do PVR another way, but want also to have all addon types standardized in the way how changes are handled. For PVR most critical.

But related to every addon who does work at this moment (Python or Binary). It's mostly true that a addon file is cached by system, but Kodi does never know what a addon do.

As long a addon is a in Kodi included code and not a independent exe, can it bring always a crash during change of a running one!

AlwinEsch:
I see no alternative for minimum one blocked call for something (disable or like now for unload)

Currently with @tamland's thing becomes a blocked call used on unload and load.
During test seen calls:
Disable addon: Calls AddonEvents::Disabled only
Uninstall addon: Calls AddonEvents::Unload only (no Disabled)
Enable addon: Calls AddonEvents::Enabled only
Install addon: Calls AddonEvents::Load only (no Enabled)

The for me very bad thing is that with `OnEvent(const ADDON::AddonEvent& event)` every registered place becomes informed, e.g. by a Python script or Skin addon change also the VFS binary addon.

In case of 20 registered OnEvent parts and 100 changed addons is this horrible.

Rechi:
@xhaggi this needs a rebase

tamland:
jenkins build this pleas

tamland:
jenkins build this please

PR #12847, opened 2017-09-28
[droid] network fixes

koying:
Fixes:
1) Crash when disconnecting from airplay
It's due to the limitation on the number of zeroconf discoverers, that still needs to be workarounded somehow

2) Hostname is always "localhost"
Now uses the revert DNS of the IP of the the default interface
Hostname appears at the start of the log

Ref: https://forum.kodi.tv/showthread.php?tid=319539


MartijnKaijser:
Still seems to work fine for me so I guess it's all good

PR #12855, opened 2017-09-29
[Droid] Assorted fixes

koying:
Fixes https://forum.kodi.tv/showthread.php?tid=321657


peak3d (approved):
Thx, looks good (wondering that this ERROR log message was really still present)

PR #12857, opened 2017-09-30
Use our api to set info on python scrapers

phate89:
This is a splitup from https://github.com/xbmc/xbmc/pull/12661
I replaced the use of properties with our functions we already use to set info in plugins.
I also added nfo url support that @notspiff original implementation was missing.
I'm not sure if it's worth it in music part since there are only few functions available.
@ronie @DaveTBlake any opinion?

## Motivation and Context
We can use functions we are more familiar with.

## How Has This Been Tested?
Tested with my themoviedb python scraper

## Types of change

## Checklist:

### DIFF
--- a/addons/metadata.demo.movies/demo.py
+++ b/addons/metadata.demo.movies/demo.py
@@ -112,5 +112,10 @@ elif action == 'getartwork':
         liz.setProperty('video.fanart2.preview', '/home/akva/Pictures/hawaii-shirt.png')
         liz.setProperty('video.fanart2.dim', '1080')
         xbmcplugin.setResolvedUrl(handle=int(sys.argv[1]), succeeded=True, listitem=liz)
+elif action == 'nfourl':
+    nfo=urllib.unquote_plus(params["nfo"])
+    print 'Find url from nfo file'
+    liz=xbmcgui.ListItem('Demo movie 1', offscreen=True)
+    xbmcplugin.addDirectoryItem(handle=int(sys.argv[1]), url="/path/to/movie1", listitem=liz, isFolder=True)
 
 xbmcplugin.endOfDirectory(int(sys.argv[1]))
--- a/addons/metadata.demo.movies/demo.py
+++ b/addons/metadata.demo.movies/demo.py
@@ -46,71 +46,48 @@ elif action == 'getdetails':
     url=urllib.unquote_plus(params["url"])
     if url == '/path/to/movie':
         liz=xbmcgui.ListItem('Demo movie 1', offscreen=True)
-        liz.setProperty('video.original_title', 'Demo måvie 1')
-        liz.setProperty('video.sort_title', '2')
-        liz.setProperty('video.ratings', '1')
-        liz.setProperty('video.rating1.value', '5')
-        liz.setProperty('video.rating1.votes', '100')
-        liz.setProperty('video.user_rating', '5')
-        liz.setProperty('video.top250', '3')
-        liz.setProperty('video.unique_id', '123')
-        liz.setProperty('video.imdb_id', '456')
-        liz.setProperty('video.plot_outline', 'Outline yo')
-        liz.setProperty('video.plot', 'Plot yo')
-        liz.setProperty('video.tag_line', 'Tag yo')
-        liz.setProperty('video.duration_minutes', '110')
-        liz.setProperty('video.mpaa', 'T')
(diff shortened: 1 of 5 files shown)


DaveTBlake:
>I'm not sure if it's worth it in music part since there are only few functions available.

TBH I'm not sure it helps either. For music scraping from NFO, or online, is about getting additional data, hence most of it does not have an equivalent field in CMusicInfoTag. When there is more data to scrape (e.g. I hope to add gender, type and disambiguation to artist) it won't get added to CMusicInfoTag. 

The demo album and artist scrapers are also not how music scraping will work (I have worked on the scrapers you started @ronie but got distracted).

So thank you for considering music @phate89 but I would probably drop that part of it.

addons/metadata.demo.tv/demo.py
```diff
@@ -46,156 +46,146 @@ def get_params():
...
+                     'tag': ['Porn', 'Umomz']
+                    })
+        liz.setRating("imdb", 9, 100000, True )
+        liz.setRating("tvdb", 8.9, 1000)
+        liz.setUniqueIDs({ 'tvdb': '3894', 'imdb' : 'tt384940' }, 'tvdb')
+        liz.addSeason(1, 'Horrible')
+        liz.addSeason(2, 'Crap')
+        liz.setCast([{'name': 'spiff', 'role': 'himself', 'thumbnail': '/home/akva/Pictures/fish.jpg', 'order': 2},
+                    {'name': 'monkey', 'role': 'orange', 'thumbnail': '/home/akva/Pictures/coffee.jpg', 'order': 1}])
+        liz.addAvailableThumb('DefaultBackFanart.png', '1.78')
```
rmrector: Like the TV show itself above, the second parameter to `addAvailableThumb` should be an art type, generally 'thumb', or could even something wild like 'clearlogo'.

rmrector (changes requested):
Ohh, these Python scrapers are looking wonderful.

There are a couple of small changes needed, and a suggestion to smooth out a part of the API for artwork.

MartijnKaijser:
@phate89 @notspiff can we throw this one in and start testing?

addons/metadata.demo.movies/demo.py
```diff
@@ -46,71 +46,53 @@ def get_params():
...
+        liz.setRating("themoviedb", 8.9, 1000)
+        liz.setUniqueIDs({ 'imdb': 'tt8938399', 'tmdb' : '9837493' }, 'imdb')
+        liz.setCast([{'name': 'spiff', 'role': 'himself', 'thumbnail': '/home/akva/Pictures/fish.jpg', 'order': 2},
+                    {'name': 'monkey', 'role': 'orange', 'thumbnail': '/home/akva/Pictures/coffee.jpg', 'order': 1}])
+        liz.addAvailableThumb('DefaultBackFanart.png', 'banner')
+        liz.addAvailableThumb('/home/akva/Pictures/hawaii-shirt.png', 'poster')
+        liz.setAvailableFanart([{'image': 'DefaultBackFanart.png', 'preview': 'DefaultBackFanart.png'}, 
+                                {'image': '/home/akva/Pictures/hawaii-shirt.png', 'preview': '/home/akva/Pictures/hawaii-shirt.png'}])
         xbmcplugin.setResolvedUrl(handle=int(sys.argv[1]), succeeded=True, listitem=liz)
 elif action == 'getartwork':
```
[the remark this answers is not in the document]
phate89: The names are the same we use in the xml scraper.. That's probably the reason spiff used those..

phate89:
thanks @rmrector for the review..I fixed a couple of your points..
About naming I agree with your naming considerations but the problem is that it's all implemented like it was for xml scraper. addAvailableThumb and setAvailableFanart are really different and they are implemented very differently in kodi core so I preferred not to touch anything and replicate them on python

addons/metadata.demo.tv/demo.py
rmrector: The aspect in the line above and below still need adjusted, but otherwise this PR looks good.

phate89:
I rebased and renamed some argument like @rmrector suggested. It should be good to go

rmrector:
Looks good, and I definitely appreciate that you were able to rename that guy. Thanks!

anthonyryan1:
I just thought I'd throw out a vote of support for this. I just finished writing a python scraper (not yet published) and this is a significant improvement.

PR #12858, opened 2017-09-30
ADD: [amcs] restart codec when surface is lost and found

koying:
Fixes issue when, eg, playback is interrupted by chromecast, and surface is lost

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAndroidMediaCodec.cpp
@@ -77,7 +77,8 @@ enum MEDIACODEC_STATES
   MEDIACODEC_STATE_FLUSHED,
   MEDIACODEC_STATE_RUNNING,
   MEDIACODEC_STATE_ENDOFSTREAM,
-  MEDIACODEC_STATE_ERROR
+  MEDIACODEC_STATE_ERROR,
+  MEDIACODEC_STATE_STOPPED
 };
 
 static bool IsBlacklisted(const std::string &name)
@@ -602,14 +603,6 @@ bool CDVDVideoCodecAndroidMediaCodec::Open(CDVDStreamInfo &hints, CDVDCodecOptio
       CLog::Log(LOGERROR, "CDVDVideoCodecAndroidMediaCodec: VideoView creation failed!!");
       goto FAIL;
     }
-
-    m_jnivideosurface = m_jnivideoview->getSurface();
-    if (!m_jnivideosurface)
-    {
-      CLog::Log(LOGERROR, "CDVDVideoCodecAndroidMediaCodec: VideoView getSurface failed!!");
-      goto FAIL;
-    }
-    m_surface = ANativeWindow_fromSurface(xbmc_jnienv(), m_jnivideosurface.get_raw());
   }
 
   m_codec = nullptr;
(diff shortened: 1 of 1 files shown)


CiNcH83:
Can you provide a test-build?

MartijnKaijser:
It's on the server

CiNcH83:
Resume after focus loss seems to be fixed. Tried Chromecast, the "open with" dialog when connecting an USB storage and switching to another app.

What people still have to keep in mind is that when doing some Chromecast or switching app, a running PVR only goes into pause, which means that it keeps fully running, writing data to some timeshift buffer, which might be desired or not. In the latter case, people have to stop playback manually.

koying:
Yeah. Might want to do stop altogether when PVR is involved.
That will avoid headaches ;)

@MartijnKaijser Do you think just stopping is ok, or is it worth looking for advanced solutions.

MartijnKaijser:
I'm OK with stopping. Maybe later we can look for a better solution.

MartijnKaijser:
Maybe we can already get this merged?

koying:
@CiNcH83 Continuing from 12866
Yep, it works if the codec is completely deleted on my Sony (nothing really Sony specific, the issue would be the same on all single decoder devices)

It's tricky to restart it though, as the shutdown of the codec from, e.g. ChromeCast, takes a while, and surely longer than when we get the signal that the surface is available again.

I confirmed it working with a semi-hack, but I have to think about a better way, if possible

CiNcH83:
I retested on Sony Android TV with Kodi being paused in the background. I tested with file (H.264 720p50) from USB (to eliminate potential PVR and network problems). Here is what I tried during file playback...

(1) Initiate Chromecast playback of some YouTube clip:
Video plays at reduced resolution, starting in slow motion, constantly interrupting showing the "LOADING..." OSD, as if it needs to rebuffer which can't be the case though as Kodi in the background is neither consuming any CPU time nor network traffic.
When I stop playback in Kodi before I start Chromecast playback, everything is perfectly fine. Full resolution and smooth playback.

(2) Pressing HOME > open Netflix:
I tried a network bandwidth test in the Netflix settings which yielded the maximum of my broadband connection. So network seems to work perfectly fine while Kodi is paused in the background.
When I now try to play some Netflix movie or show, buffering is very slow, always stopping at exactly 24%. When I now go back HOME, the Kodi 'Now Playing' card has a Netflix symbol to the lower right, while it still mentions the file which is currently being paused in Kodi. When I enter it, I see the still image of the paused file, but I can't do anything. Can't bring up Kodi controls, go back, whatever. After several seconds, Netflix stops working (Android says "Unfortunately, Netflix has stopped"), and I can fully use Kodi again.

So pausing Kodi in the background does not work at all for me, neither for PVR nor for file. Might be due to the slow Sony TV.

[Here](https://www.youtube.com/watch?v=vt7iUUet_g4) is a video of the Netflix incident (2). Netflix crashes at the very end of the video.

So like @FernetMenta wrote, I would also do a stop. Things seem to go wrong as soon as another app tries to do some video playback. If I think about it more, it might very well be a video decoder resource shortage, with the second video (Chromecast, Netflix) having to use SW. A SHIELD can probably cope with that. DRM video might require HW video decoding though.

CiNcH83:
BTW. According to Sony MediaTek MT5890 specs, the SoC can handle up to 4 AVC or HEVC 1080p60 streams. As for 4K and VP9 generally, it can only do 1 stream.

I only tested YouTube which is VP9 most probably and Netflix in 4K. I should probably try some Netflix 1080p. It is still a problem for the mentioned cases where only 1 stream is supported..

koying:
Will already merge this one, to unblock the cases where:
1) multiple decoders are readily available
2) focus is lost to a non-video app

PR #12859, opened 2017-09-30
FIX: [voicerecog] add alternative keys

koying:
Add more alternatives to trigger voice recog, now that the "search" button is no more trappable on Shield


koying:
@MartijnKaijser @garbear Who do you think I should ping to get this approved / commented on?
"VoiceRecognizer" is (currently) no-op on all platforms besides android.

MartijnKaijser:
I'll trust you that it works and if it's a noop on other platforms. 👍

PR #12860, opened 2017-09-30
Updated hungarian.xml frequently used buttons add to keyboard layout

Wolftronics-SBC:
Updated hungarian.xml frequently used buttons add to first keyboard layout


ronie:
thx!

PR #12866, opened 2017-10-01
 FIX: [droid] stop instead of pause for non-pausable streams when losing focus 

koying:


xbmc/cores/VideoPlayer/VideoPlayer.cpp
```diff
@@ -3014,6 +3014,14 @@ bool CVideoPlayer::HasRDS() const
   return m_CurrentRadioRDS.id >= 0;
 }
 
+bool CVideoPlayer::HasPVR() const
+{
+  if(m_pInputStream && m_pInputStream->IsStreamType(DVDSTREAM_TYPE_PVRMANAGER))
```
ksooo: Not 100% sure but I think this also is true for PVR recordings, not only Live TV.

xbmc/cores/VideoPlayer/VideoPlayer.h
```diff
@@ -305,6 +305,7 @@ class CVideoPlayer : public IPlayer, public CThread, public IVideoPlayer, public
   bool HasVideo() const override;
   bool HasAudio() const override;
   bool HasRDS() const override;
+  bool HasPVR() const override;
```
FernetMenta: I am in the process of cleaning out every PVR (and similar) dependency from VP. VP just plays a stream. It does not care if this is PVR or any other source.

maybe CanPause is what you are looking for?

koying:
Ok, using CanPause(), now.
Not exactly what I had in mind, but it'll do

CiNcH83:
I don't think that this is the intended behavior.

CVideoPlayer->(CDVDInputStreamPVRManager)m_pInputStream->CanPause asks the PVR add-on whether it can pause stream (CanPauseStream()) which will return true if it supports timeshifting. So we are in the same situation as before. Live stream will be paused instead of stopped when losing the focus.

koying:
Well, there is no.way to detect that a stream is PVR (live or timeshifted) and my solution was rejected, so it's the best I can do.

Anyway, on 2nd though, it's probably best to *not* lose the timeshifting buffet in this situation

FernetMenta:
for what reason do you want it to stop or pause?

CiNcH83:
Let me quote myself:

> What people still have to keep in mind is that when doing some Chromecast or switching app, a running PVR only goes into pause, which means that it keeps fully running, writing data to some timeshift buffer, which might be desired or not. In the latter case, people have to stop playback manually.

It is how people use Android. They for example have Kodi PVR running, then decide to do some Chromecast or switch app. In both cases PVR will keep running, going into Timeshift mode, unless you manually stop. So you will have to teach people to stop playback manually first if they don't want their precious flash memory to be hammered with Timeshift data. This behavior might be desired though. You can resume playback later on from where you just left off. So leaving it the way it is also has its advantage. The user can decide.

However, I experienced severe issues yesterday when PVR and Chromecast were running in parallel, causing issues on my Sony as both were downloading data in parallel (as PVR timeshifting for DVBViewer Client is handled at client side, not at server-side). Chromecast was buffering all the time as if network speed was not enough.

FernetMenta:
Maybe a better approach is to stop always and do "resume from last" if possible when coming back. There is a whole bunch of cases where pause is not appropriate, apart from PVR streams.

fritsch:
See: https://github.com/xbmc/xbmc/pull/12871

CiNcH83:
I retested on Sony Android TV with Kodi being paused in the background. I tested with file (H.264 720p50) from USB (to eliminate potential PVR and network problems). Here is what I tried during file playback...

(1) Initiate Chromecast playback of some YouTube clip:
Video plays at reduced resolution, starting in slow motion, constantly interrupting showing the "LOADING..." OSD, as if it needs to rebuffer which can't be the case though as Kodi in the background is neither consuming any CPU time nor network traffic.
When I stop playback before I start Chromecast playback, everything is perfectly fine. Full resolution and smooth playback.

(2) Pressing HOME > open Netflix:
I tried a network bandwidth test in the Netflix settings which yielded the maximum of my broadband connection. So network seems to work perfectly fine while Kodi is paused in the background.
When I now try to play a file, buffering is very slow and stops at some point in time (at 20-30%). Nothing more happens. When I now go back HOME, the Kodi 'Now Playing' card has a Netflix symbol to the lower right, while it still mentions the file which is currently being paused in Kodi. When I enter it, I see the still image of the paused file, but I can't do anything. Can't bring up Kodi controls, go back, whatever. After several seconds, Netflix stops working (Android says "Unfortunately, Netflix has stopped"), and I can fully use Kodi again.

So pausing Kodi in the background does not work at all for me. Might be due to the slow Sony TV, probably a RAM shortage.

CiNcH83:
[Here](https://www.youtube.com/watch?v=vt7iUUet_g4) is a video of the Netflix incident (2). Netflix crashes at the very end of the video.

So like @FernetMenta wrote, I would also do a stop. Things seem to go wrong as soon as another app tries to do some video playback. If I think about it more, it might very well be a video decoder resource shortage, with the second video (Chromecast, Netflix) having to use SW. A SHIELD can probably cope with that. DRM video might require HW video decoding though.

koying:
Yep, I suspect it's because the MediaCodec is not properly closed in Kodi, and the Sony has only 1 IIRC (vs. more on the Shield).
I'll have a look.

koying:
@CiNcH83 Actually, can you retest with https://github.com/xbmc/xbmc/pull/12858 (Not on a PVR stream, obviously) ?
Test build are available.

If it's still failing, please c/p your observations on that PR, as it's more related to that one, I suspect

CiNcH83:
I did my tests with exactly this build. So what I wrote above applies.

koying:
Will merge this one as it is, as it's an improvement anyway

PR #12869, opened 2017-10-02
Special protocol migration

noobSaibot1:
Migrating from using static map to special paths stored in https://github.com/xbmc/xbmc/blob/master/xbmc/filesystem/SpecialProtocol.cpp and calculated in
https://github.com/xbmc/xbmc/blob/master/xbmc/Application.cpp to On-the-fly path calculations in "SpecialProtocol.cpp" instead.
## Description

Instead of using GetPath() in "SpecialProtocol.h" specific Static getters were implemented, means that the static path map in "SpecialProtocol.cpp" isn't needed for path calculations anymore.

## Motivation and Context

The aim is to solve the early destruction of the static path map (m_pathMap) while Kodi is exiting when the stored paths still needed by other code.

## How Has This Been Tested?

Ran the kodi-test project main and passed all tests, Ran Kodi to check for errors. Printed some logs to make sure the path translation (TranslatePath() in SpecialProtocol.h) is the same via GetPath() and new/replacement getters .

VS 2017 debugger (Windows 10), kodi-test main (which runs all tests)

## Types of change

- [ x] Improvement (non-breaking change which improves existing functionality)

## Checklist:

- x[ ] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [x ] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [x ] All new and existing tests passed

### DIFF
--- a/xbmc/filesystem/SpecialProtocol.cpp
+++ b/xbmc/filesystem/SpecialProtocol.cpp
@@ -35,6 +35,71 @@
 #include "utils/StringUtils.h"
 #endif
 
+#ifdef TARGET_WINDOWS
+#include "platform/win32/WIN32Util.h"
+#endif
+
+
+#ifdef TARGET_WINDOWS
+//Getters
+std::string CSpecialProtocol::GetHomePath() 
+{
+  return CWIN32Util::GetProfilePath();
+}
+
+std::string CSpecialProtocol::GetHomePath(const std::string &fileName) 
+{
+  return URIUtils::AddFileToFolder(GetHomePath(), fileName);
+}
+
+std::string CSpecialProtocol::GetTmpPath() 
+{
+  return URIUtils::AddFileToFolder(CWIN32Util::GetProfilePath(), "cache");
+}
+
+std::string CSpecialProtocol::GetTmpPath(const std::string &fileName) 
+{
+  return URIUtils::AddFileToFolder(GetTmpPath(), fileName);
+}
+
+
+std::string CSpecialProtocol::GetXBMCBinAddonPath() 
+{
+  std::string str = GetXBMCPath("/addons");
+  return str;
+}
+
+std::string CSpecialProtocol::GetProfilePath()
+{
+  return CProfilesManager::GetInstance().GetProfileUserDataFolder();
+}
+
+std::string CSpecialProtocol::GetProfilePath(const std::string &fileName)
+{
+  return URIUtils::AddFileToFolder(GetProfilePath(), fileName);
+}
+
+std::string CSpecialProtocol::GetMasterProfilePath() 
+{
+  return URIUtils::AddFileToFolder(CWIN32Util::GetProfilePath(), "userdata");
+}
+
+std::string CSpecialProtocol::GetMasterProfilePath(const std::string &fileName)
+{
+  return URIUtils::AddFileToFolder(GetMasterProfilePath(), fileName);
+}
+
+std::string CSpecialProtocol::GetXBMCPath()
+{
+  return CUtil::GetHomePath();
+}
+
+std::string CSpecialProtocol::GetXBMCPath(const std::string &fileName) 
+{
+  return URIUtils::AddFileToFolder(GetXBMCPath(), fileName);
+}
+#endif
+
 void CSpecialProtocol::SetProfilePath(const std::string &dir)
 {
   SetPath("profile", dir);
@@ -130,7 +195,6 @@ std::string CSpecialProtocol::TranslatePath(const CURL &url)
   }
 
   std::string FullFileName = url.GetFileName();
-
   std::string translatedPath;
   std::string FileName;
   std::string RootDir;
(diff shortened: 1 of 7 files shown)


Rechi:
@noobSaibot1 please remove the merge commit.
If you want the latest changes from master rebase instead of merge.

Rechi:
You have to completely drop the merge (and revert) commit instead of reverting it.

xbmc/filesystem/SpecialProtocol.cpp
```diff
@@ -263,16 +359,16 @@ std::string CSpecialProtocol::TranslatePathConvertCase(const std::string& path)
...
+  CLog::Log(LOGNOTICE, "special://masterprofile/ is mapped to: %s", GetMasterProfilePath().c_str());
 #if defined(TARGET_POSIX)
   CLog::Log(LOGNOTICE, "special://envhome/ is mapped to: %s", GetPath("envhome").c_str());
 #endif
-  CLog::Log(LOGNOTICE, "special://home/ is mapped to: %s", GetPath("home").c_str());
-  CLog::Log(LOGNOTICE, "special://temp/ is mapped to: %s", GetPath("temp").c_str());
-  CLog::Log(LOGNOTICE, "special://logpath/ is mapped to: %s", GetPath("logpath").c_str());
+  CLog::Log(LOGNOTICE, "special://home/ is mapped to: %s", GetHomePath().c_str());
+  CLog::Log(LOGNOTICE, "special://temp/ is mapped to: %s", GetTmpPath().c_str());
+  CLog::Log(LOGNOTICE, "special://logpath/ is mapped to: %s", GetHomePath().c_str());
```
wsnipex: all those methods are undefined on posix platforms.

noobSaibot1: I've fixed that issue. now i'm passing some of the Jenkins OS checks.
I'm seeing in the logs (which seems to be the problem):
"FATAL: Could not checkout 6d2ebe732a5d7a9babb51c6522040b9e232cef47"

Rechi:
jenkins build this please

garbear:
Why hasn't the commented code been dropped? that's what history is for

hudokkow:
@garbear this is WIP afaik

hudokkow:
jenkins build this please

Razzeee:
jenkins build this please

hudokkow:
jenkins build this please

wsnipex:
in case you missed my comment on irc, https://github.com/xbmc/xbmc/blob/master/xbmc/test/TestUtils.cpp#L114 gets set to CMAKE_INSTALL_PREFIX, which causes some tests to fail when kodi was compiled with a non empty CMAKE_INSTALL_PREFIX.

I couldn't reproduce the error from jenkins.
From a quick runtime test, kodi builds and runs fine on linux.

Rechi:
wrong revision was build
jenkins build this please

Rechi:
wrong revision was build
jenkins build this please

Rechi:
@noobSaibot1 the reason why the testuite faile on jenkins but not on your/my local machine is that there `%AppData%\Kodi` exists.

MartijnKaijser:
Can we go forward with this?

Paxxi:
Sorry about this. Will try to go through this over the holidays so we can get it merged.

xbmc/filesystem/SpecialProtocol.cpp
```diff
@@ -28,12 +28,402 @@
...
+{
+  return URIUtils::AddFileToFolder(GetXBMCBinPath(), fileName);
+}
+
+
+std::string CSpecialProtocol::GetXBMCBinAddonPath()
+{
+  std::string path;
+#ifdef TARGET_WINDOWS
+  path = GetXBMCPath("/addons");
```
Paxxi: you don't need a slash here `URIUtils::AddFileToFolder` will make sure a slash is added.

xbmc/filesystem/SpecialProtocol.cpp
```diff
@@ -28,12 +28,402 @@
...
+
+std::string CSpecialProtocol::GetXBMCAltBinAddonPath()
+{
+  std::string path;
+#if defined(TARGET_POSIX)
+  /* REVIEW:
+  if (getenv("KODI_BINADDON_PATH")) // don't see why to do this check. what to do if false?
+  */
+  path = getenv("KODI_BINADDON_PATH");
+  if (!g_application.PlatformDirectoriesEnabled())
```
Paxxi: same here

xbmc/filesystem/SpecialProtocol.cpp
```diff
@@ -28,12 +28,402 @@
...
+  return URIUtils::AddFileToFolder(GetXBMCBinAddonPath(), fileName);
+}
+
+
+std::string CSpecialProtocol::GetXBMCAltBinAddonPath()
+{
+  std::string path;
+#if defined(TARGET_POSIX)
+  /* REVIEW:
+  if (getenv("KODI_BINADDON_PATH")) // don't see why to do this check. what to do if false?
```
Paxxi: yeah this check seems redundant, and can be removed.

xbmc/filesystem/SpecialProtocol.cpp
```diff
@@ -28,12 +28,402 @@
...
+    userHome = "/root";
+
+  if (g_application.PlatformDirectoriesEnabled())
+  {
+    std::string strTempPath = userHome;
+    std::string appName = CCompileInfo::GetAppName();
+    // REVIEW: Make a get function for dotLowerAppName calc
+    std::string dotLowerAppName = "." + appName;
+    StringUtils::ToLower(dotLowerAppName);
+    strTempPath = URIUtils::AddFileToFolder(strTempPath, dotLowerAppName + "/temp");
```
Paxxi: only use "temp", the slash is redundant and you can call this as 
`strTempPath = URIUtils::AddFileToFolder(strTempPath, dotLowerAppName, "temp");` it takes an "unlimited" amount of arguments.

xbmc/filesystem/SpecialProtocol.cpp
```diff
@@ -28,12 +28,402 @@
...
+    if (getenv(envAppTemp))
+      strTempPath = getenv(envAppTemp);
+    path = strTempPath;
+  }
+  else
+  {
+    std::string strTempPath = GetXBMCPath();
+    strTempPath = URIUtils::AddFileToFolder(strTempPath, "portable_data/temp");
```
Paxxi: `strTempPath = URIUtils::AddFileToFolder(strTempPath, "portable_data", "temp");`

xbmc/filesystem/SpecialProtocol.cpp
```diff
@@ -28,12 +28,402 @@
...
+#endif
+
+#ifdef TARGET_ANDROID
+#include "platform/android/activity/XBMCApp.h"
+#include <androidjni/ApplicationInfo.h>
+#include "platform/android/bionic_supplement/bionic_supplement.h"
+#endif
+
+using namespace XFILE;
+
```
Paxxi: You can stick internal helper functions here like this
```
namespace
{
std::string GetLowerAppName()
{
  std::string appName = CCompileInfo::GetAppName();
  std::string dotLowerAppName = "." + appName;
  StringUtils::ToLower(dotLowerAppName);
  return dotLowerAppName;
}
}
```

instead of having them inside the class. This will hide them inside this file only and is a nice way to hide helper functions.

Paxxi (changes requested):
I think this looks good.

I added some comments that are mostly nitpicks but I'd like to see addressed.

You also need to clean up the added logging/fprintf stuff before it's ready for merging.

I'm working on a solution for the test case failures so you don't need to worry about that.

Great work! and again sorry for going dark and letting this idle.

xbmc/filesystem/SpecialProtocol.cpp
garbear: And even better is to move hard-coded strings to the top of the .cpp files using #defines or constants

Paxxi: Moving the constants is a good idea as well. I think this can go in even without that as it's his/her first PR and some cleanup can be done in a second pr. Now that I think about it, same goes for my comment about the helper method. It can be refactored later.

I'm feeling bad about this taking so long. I need to go over my previous comments and see what what affects functionality and what's more cosmetic that can be bumped.

xbmc/filesystem/SpecialProtocol.cpp
Paxxi: I was wrong about this. It returns null if the environment variable doesn't exist and assigning that to a string will cause issues so the check should stay.

noobSaibot1:
Hi.
I was not available for a while.
I will look at all the remarks soon and start to implement changes needed.
Thanks for the remarks.

Paxxi:
@noobSaibot1 how's it going? Is there anything I can help you with?
I just want to make it clear that as long as it's working we can merge it. Any comments about style, duplicate code, improvements etc we can solve later. I think the important part now is to get something merged so you get a "win" and then if you want to you can continue working on cleanup in a second PR or we can help out.

phunkyfish:

PR #12870, opened 2017-10-02
fixed: fix multitrack support for audio decoders

notspiff:
Somebody has confused tracks and tags, as well as enforced that tags must be supported for tracks to work. Fix this.

### DIFF
--- a/xbmc/addons/AudioDecoder.cpp
+++ b/xbmc/addons/AudioDecoder.cpp
@@ -31,6 +31,7 @@ CAudioDecoder::CAudioDecoder(const BinaryAddonBasePtr& addonInfo)
 {
   m_CodecName = addonInfo->Type(ADDON_AUDIODECODER)->GetValue("@name").asString();
   m_strExt = m_CodecName + "stream";
+  m_hasTags = addonInfo->Type(ADDON_AUDIODECODER)->GetValue("@tags").asBoolean();
   m_struct = {{ 0 }};
 }
 
@@ -116,10 +117,18 @@ int CAudioDecoder::GetTrackCount(const std::string& strPath)
 
   int result = m_struct.toAddon.track_count(&m_struct, strPath.c_str());
 
-  if (result > 1 && !Load(strPath, XFILE::CMusicFileDirectory::m_tag, nullptr))
-    return 0;
+  if (result > 1)
+  {
+    if (m_hasTags)
+    {
+      if (!Load(strPath, XFILE::CMusicFileDirectory::m_tag, nullptr))
+        return 0;
+    }
+    else
+      XFILE::CMusicFileDirectory::m_tag.SetTitle(CURL(strPath).GetFileNameWithoutPath());
+    XFILE::CMusicFileDirectory::m_tag.SetLoaded(true);
+  }
 
-  XFILE::CMusicFileDirectory::m_tag.SetLoaded(true);
   return result;
 }
 
--- a/xbmc/utils/FileExtensionProvider.cpp
+++ b/xbmc/utils/FileExtensionProvider.cpp
@@ -86,12 +86,33 @@ std::string CFileExtensionProvider::GetSubtitleExtensions() const
 std::string CFileExtensionProvider::GetVideoExtensions() const
 {
   std::string extensions(m_advancedSettings->m_videoExtensions);
-  extensions += '|' + GetAddonExtensions(ADDON_VFS);
+  if (!extensions.empty())
+    extensions += '|';
+  extensions += GetAddonExtensions(ADDON_VFS);
+
+  return extensions;
+}
+
+std::string CFileExtensionProvider::GetFileFolderExtensions() const
+{
+  std::string extensions(GetAddonFileFolderExtensions(ADDON_VFS));
+  if (!extensions.empty())
+    extensions += '|';
+  extensions += GetAddonFileFolderExtensions(ADDON_AUDIODECODER);
 
   return extensions;
 }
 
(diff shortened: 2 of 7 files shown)


xbmc/addons/AudioDecoder.cpp
```diff
@@ -116,10 +117,18 @@ int CAudioDecoder::GetTrackCount(const std::string& strPath)
...
   int result = m_struct.toAddon.track_count(&m_struct, strPath.c_str());
 
-  if (result > 1 && !Load(strPath, XFILE::CMusicFileDirectory::m_tag, nullptr))
-    return 0;
+  if (result > 1)
+  {
+    if (m_hasTags)
+    {
+      if (!Load(strPath, XFILE::CMusicFileDirectory::m_tag, nullptr))
+      return 0;
```
xhaggi: wrong indentation

notspiff:
fixed.

xhaggi:
Thx

notspiff:
appended some more fixes;
1) let the file extension provider handle file folder extensions
2) a stupid infinite loop due to the fact that add-on id matches a file folder extension. this leads to an infinite loop (e.g. audiodecoder.wsr and .wsr files)

notspiff:
Shall i invent a fake TYPE entry then? Or make the map a vector of strings? Neither appealed to me..

notspiff:
As in i need to encode the bool in addition to the type. Existence of an addon is not enough

xhaggi:
Sure I meant you should store it in condition of the boolean value into a new string or vector of string member, `m_addonFileFolderExtension` and return it in the getter.

xhaggi:
something like

```c++
for (const auto& addonInfo : addonInfos)
{
  std::string ext = addonInfo->Type(type)->GetValue(ADDON_VFS == type ? "@extensions" : "@extension").asString();
  if (!ext.empty())
  {
    extensions.push_back(ext);

    std::string info = ADDON_VFS == type ? "@filedirectories" : "@tracks";
    if (addonInfo->Type(type)->GetValue(info).asBoolean())
      fileFolderExtensions.push_back(ext);
  }
}

m_addonFileFolderExtensions = StringUtils::Join(fileFolderExtensions, "|");
```

notspiff:
ah, i remember why i didn't do that. because of the the event, it only updates for a single type.

xhaggi:
We could use the same map structure and combine each type in the getter

notspiff:
right. if you think it's worth it, i will do another map.

notspiff:
now with more map.

xhaggi:
Yes please, we don't update the addons very often at runtime and if it's done it will only effect the related types.

notspiff:
sure i meant in terms of code effort ;)

xbmc/utils/FileExtensionProvider.cpp
```diff
@@ -123,6 +145,28 @@ void CFileExtensionProvider::SetAddonExtensions(const TYPE& type)
   m_addonExtensions.insert(make_pair(type, StringUtils::Join(extensions, "|")));
 }
 
+void CFileExtensionProvider::SetAddonFileFolderExtensions(const TYPE& type)
```
xhaggi: Thought we could combine both in `SetAddonExtensions` and add the conditions there, but its fine too.

notspiff: sure. different rules in different projects, my colleagues would hate on me for doing two things in one method.

xbmc/utils/FileExtensionProvider.cpp
```diff
@@ -111,16 +132,30 @@ void CFileExtensionProvider::SetAddonExtensions()
...
+    std::string info = ADDON_VFS == type ? "@extensions" : "@extension";
+    std::string ext = addonInfo->Type(type)->GetValue(info).asString();
     if (!ext.empty())
       extensions.push_back(ext);
+    if (type == ADDON_VFS || type == ADDON_AUDIODECODER)
+    {
+      std::string info2 = ADDON_VFS == type ? "@filedirectories" : "@tracks";
+      if (addonInfo->Type(type)->GetValue(info2).asBoolean())
+      {
+        ext = addonInfo->Type(type)->GetValue(info).asString();
```
xhaggi: No need to get this value again, it's already done on line 141

xbmc/utils/FileExtensionProvider.cpp
```diff
@@ -111,16 +132,30 @@ void CFileExtensionProvider::SetAddonExtensions()
...
   BinaryAddonBaseList addonInfos;
   CServiceBroker::GetBinaryAddonManager().GetAddonInfos(addonInfos, true, type);
   for (const auto& addonInfo : addonInfos)
   {
-    std::string ext = addonInfo->Type(type)->GetValue(ADDON_VFS == type ? "@extensions" : "@extension").asString();
+    std::string info = ADDON_VFS == type ? "@extensions" : "@extension";
+    std::string ext = addonInfo->Type(type)->GetValue(info).asString();
     if (!ext.empty())
       extensions.push_back(ext);
+    if (type == ADDON_VFS || type == ADDON_AUDIODECODER)
```
xhaggi: Enclose the code below into `if (!ext.empty())`, we can skip it if ext is empty.

xhaggi (changes requested):

notspiff:
updated.

Rechi:
jenkins build this please

xhaggi:
@notspiff thx

xhaggi (approved):

notspiff:
approved for a week, no conflicts, is a bug fix. what's holding this back?

xhaggi:
Merge it in

PR #12871, opened 2017-10-02
XBMCApp: Use STOP if PAUSE is not supported

fritsch:
Test for: #12866

### DIFF
--- a/xbmc/platform/android/activity/XBMCApp.cpp
+++ b/xbmc/platform/android/activity/XBMCApp.cpp
@@ -258,7 +258,12 @@ void CXBMCApp::onStop()
   android_printf("%s: ", __PRETTY_FUNCTION__);
 
   if ((m_playback_state & PLAYBACK_STATE_PLAYING) && (m_playback_state & PLAYBACK_STATE_VIDEO) && !m_hasReqVisible)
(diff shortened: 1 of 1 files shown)


fritsch:
Testbuild pending, will have: androidstop in name.

koying:
Nope, I don't want to stop instead of pause.
Or do I miss something?

fritsch:
It's for:
onVisibleBehindCanceled
onStop
onAudioFocusChange

only? 

Do I miss something?

koying:
Not following you.
Currently, when Kodi loses the focus, the current playback is paused.

I want to keep that.

The issue solved with the other PR is that, obvisouly, you cannot pause a non-pausable stream.
Furthermore, there was a reflection on whether timeshifted live streams should be stopped or paused.

Pro for stop: avoid to forget about the timeshift being filled up
Con for stop: When we stop, the existing timeshift buffer is lost forever.

Anyway, as there is no way to find out if the player is playing a timeshift, the "stop for all pvr" option is not possible

fritsch:
Okay - that means, we keep the PAUSE as it is now?

koying:
Yep

fritsch:
Can't we just use CanPause on the applications player interface instead? (Added this code)

fritsch:
The updated original PR does exactly the same, while querying player only once. Sorry for the noise.

PR #12874, opened 2017-10-02
AESinkPULSE: Allow pulseaudio 11 new option remixing-use-all-sink-channels=no

fritsch:
See: https://www.freedesktop.org/wiki/Software/PulseAudio/Notes/11.0/

I implemented the self-mapping workaround in earlier version, cause there was basically no other possibility to output stereo as stereo when pulseaudio was in multi-channel mode.

This change brings the choice back to the user.

### DIFF
--- a/xbmc/cores/AudioEngine/Sinks/AESinkPULSE.cpp
+++ b/xbmc/cores/AudioEngine/Sinks/AESinkPULSE.cpp
@@ -570,6 +570,14 @@ bool CAESinkPULSE::Initialize(AEAudioFormat &format, std::string &device)
   }
   else
   {
+    // version 11 got the new option "remixing-use-all-sink-channels" which
+    // can be set to "no". Therefore we give the choice back to user and let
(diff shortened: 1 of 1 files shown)


fritsch:
@Jalle19 this might be something for you, too. As one of your critics was this "it always uses all channels" .

fritsch:
More information: https://bugs.freedesktop.org/show_bug.cgi?id=94563

Jalle19:
Epic, with this I would be able to stop using pasuspender and ALSA.

fritsch:
jenkins build this please

lrusak:
@MilhouseVH can you include this?

lrusak (approved):

fritsch:
@lrusak do you already ship PA11? If yes, you also need to alter the server configuration with the mentioned setting. As you only use it for Stereo BT - it won't have any effect there I assume.

lrusak:
@fritsch yes we do. I'll have a look at editing the default server config

fritsch:
jenkins build this please

PR #12878, opened 2017-10-02
language file - update window names

ronie:
the System.CurrentWindow infolabel didn't work properly (see: https://trac.kodi.tv/ticket/17398)
to fix this, the language file needs to be in sync with WindowIDs.h

- removed outdated window names
- added missing window names
- moved two strings that got in the way

fixes trac #17398

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -4702,6 +4702,8 @@ msgstr ""
 
 #empty strings from id 2105 to 9999
 
+# ID's from 10000 to 10999 reserved for window names
+
 #: xbmc/guilib/WindowIDs.h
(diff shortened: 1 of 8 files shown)


ronie:
@jeroenpardon for review

jeroenpardon:
@ronie thanks for picking this up! Should I be able to just drop this in my kodi install>addons>resource.language.en_gb ? If I do, labels go missing and I get things like this: https://i.imgur.com/4kN1bpF.png

It's probably my method causing this, but I don't have a build environment set up at the moment. Window names are showing though 👍

phil65 (approved):

BigNoid (approved):
This makes using System.CurrentWindow as header a lot better, thx!

PR #11114, opened 2016-12-11
[gui] remove old code for viewtype parsing

phil65:
Removes the old-school way of adding viewtypes to mediawindows (without `<views>`tag). I´m quite sure this isnt used anymore at all.
@ronie

PR #12879, opened 2017-10-03
fix MusicPlaylistEditor

ronie:
the removal of the automatic view detection code in https://github.com/xbmc/xbmc/pull/11114 broke the MusicPlaylistEditor window.

this can easily be fixed by adding an explicit view id definition to the skin xml file.

@phil65


DaveTBlake:
Thanks for this @ronie despite being music I had not noticed, then again I don't use playlists much.

phil65:
Interesting, I didnt know that MyMusicPlaylistEditor.xml is a MediaWindow. Seems to be a weird choice for that imo.
Thx for fixing it!

PR #12884, opened 2017-10-05
Controller dialog: Resize layouts

garbear:
Currently, the controller dialog uses 16:9 images. This doesn't work well for controllers that are held vertically, such as the Atari 5600. As a result, the controllers would appear really tiny. This PR changes the controller dialog to use square layouts to better allow for vertical controllers.

All controller add-ons have been updated. The update add-ons have been submitted in https://github.com/xbmc/repo-resources/pull/33, which should be merged with this PR.

## Motivation and Context
Upcoming player window will use square layouts, and I realized how much better that sounds than rectangles with an arbitrary (and varying) aspect ratio.

## How Has This Been Tested?
Tested in controller dialog and add-on browser.

Before:

screenshot001

After:

screenshot000

## Types of change


ronie:
looks good to me

garbear:
Jenkins not needed. I plan to merge in a few hours.

PR #12888, opened 2017-10-06
Update strings.po

RoccoMathijn:
Fix typo


MartijnKaijser:
Sorry but these changes are lost on next Transifex update on they need to be fixed there

PR #12889, opened 2017-10-06
Fix seek on left with some SNES controllers

garbear:
This fixes a problem where the StepBack command was sent during gameplay when the left direction was pressed on a SNES dpad. Two analog axes are used instead of four buttons, and because axis events are sent very frame, axis events after the first were ignored by the game. This caused them to fall through to the GUI action.

Normally this wouldn't be fatal, because no GUI actions are defined for a FullscreenGame window. However, a second bug was causing FullscreenVideo actions to be mapped to FullscreenGame also.

## How Has This Been Tested?
Tested by pressing Left with a SNES controller on windows.

## Types of change


garbear:
jenkins build this please

PR #12890, opened 2017-10-07
GLES: fix accidentally changed code for fonts

FernetMenta:
see title


MilhouseVH:
This fixes the missing fonts on RPi2, thanks.

PR #6406, opened 2015-02-11
video library: queue database write processes

Montellese:
This is an RFC that I came up with while working on the media importing. Right now we disallow marking a video library item as watched/unwatched when a library update is already in progress and we disallow scanning another video source while we're already updating the library. That makes sense in general because video library updating is done in a huge database transaction so nothing else can write to the database anyway. But IMO it's not user friendly and also forces us to remember and handle all kinds of special circumstances when interacting with the library.

This approach adds a `CVideoLibraryQueue` and the following `CVideoLibraryJob` implementations:
- `CVideoLibraryMarkWatchedJob` is a refactored `CMarkWatchedJob` and can mark a video item as watched/unwatched
- `CVideoLibraryCleaningJob` starts cleaning the video library (for specific paths)
- `CVideoLibraryScanningJob` starts updating the video library (for specific paths)

`CVideoLibraryQueue` is a job queue of the lowest priority and only allows a single job being processed at any time. That way we can queue as many tasks that want to write to the video library/database as we want without having to think about potential special cases.

For the sake of cleaning up I have also moved the logic to start a library update / cleaning from `CApplication` into `CVideoLibraryQueue`.
Ideally we would refactor `CVideoInfoScanner` which doesn't run in its own custom thread anymore completely into `CVideoLibraryScanningJob` but I haven't done that yet because I wanted to know if the concept is working at all and get some feedback before going any further.

PR #12891, opened 2017-10-07
[Music] Export improvements and introducing the Artist Info Folder

DaveTBlake:
This addresses a number of issues, both fixes and improvements, related to local artist information and artwork, and the music library export facility in general.  All interrelated so had to be done at once.

It is just one of several phased changes that intend to make use of locally held artist and album data more robust and functional, and hence reduce the number of times that Kodi requests _the same music data_ from the online data providers. If we do not make more efficient use of these free online resources then with the millions of Kodi users we run the risk over over loading their servers and either having access restricted or being blocked completely. 

Feature/Change Summary
- Support for local art and nfo files for **all** artists via new  **Artist Information Folder**  and numerous fixes to better match art and nfo with correct artist
- Export done asynchronously (not lock UI), with new dialog (rather than series of yes/no dialogs), offering export of _all_ artists and allowing greater control of what exported to where.
______________________________________________________

The major enhancement at the core of this work is the concept of an **Artist Information Folder**. Until now all local artist art and nfo file processing has depended on the artist having a unique folder that contained all music by that artist and _only_ that artist. Unfortunately that is oversimpilifcation, and music often does not fit in that structure. Examples of things that don't fit include:
- artists with music on more than one music source e.g. my folder verses the kids folder
- artists with collaboration albums e.g. classical music listing composer, conductor, orchestra as album artists, or in pop music such as "Riding with the King" by Eric Clapton & BB King.
- song artists that don't have album just feature on compliations and as guest artists on albums by others
- other artists e.g. composers, producers, DJ mixers, those invloved with the music in some way but not the credited main artist

User's efforts to arrange their music in a way that compies can produce some unexpected results, that are difficult to pin down. Since the rest of the music library is agnostic about the folder layout and naming used (using embedded metadata to identify the music), it is also an necessesary imposition.

So from this PR on users will be able to identify an artist info folder - a location where all local artist art and nfo can be found. Backwards complatible, so art and nfo located next to the music files will still be found, but from v18 onwards they will be able to put it in one clear place. This single location can also be used by addons for other artist artwork, extra fanart etc. Also, unlike addons like cdArt and Skin Helper Service, Kodi has a view of all the artists and albums in the library so can handle the situation of duplicate artist names and folder naming with ease.

A number of issues that cause wrong artist and album art and nfo files to be used have also been fixed:
- Prevent export to XML writing data for more than one artist into NFO file.
- Check that an artist or album path is found before exporting that item to an NFO file. Check GetArtistPath is successful, and clear path and log when fails, before attempting to process NFO file for artist.
- Fix import to handle artists with same name by checking for match using mbid first.
- Check when getting album path that it is unique for that album and does not contain songs fro other albums.
- When getting local album artist art during initial scan to library: only look in the folder immediately above the album folder (when it contains only 1 album), and only for the first album artist if the album is a collaboration. Avoid repeatedly processing the same artist by maintaining a set.

Also a flaw meant that art set via import from xml was not fully applied, and users had to then individually  visit each album or artist manually to actually see the art. This is fixed.

**Improve music library export facility**
First thing to note is that the music lib export facility is _not_ a backup facility, and was never intended as one. Unlike video only parts of the music library are exported, and importing the resulting output will _not_ populate an empty library. That is done by scanning embedded metatdata from the music files themselves. Instead it is a means of outputing the additional artist and album data and artwork (that may have been scraped from varied online sources at different times) in a way that can be manipulated and edited by the user, and then either reloaded or transferred to another installation. It provides the main way in which online data can be captured once locally and then reused rather than being fetched from the server again.

Export of artist data was limited to just those album artists with a unique folder containing all music below it, that had scraped online data. This PR adds support for export of info and art for *all* artists in library. Having increased the data that _could_ be exported, then it is helpful to be able to select what is exported and where it is put, hence a UI change

- Replace series of yes/no responses with a dialog box and store previous values (using CServiceBroker) for use next time.
Image of Export Lib Dialog
image
- Single file output has unique (timestamped) file name.
- Separate file (nfo) output can be to a selected location.
- Support export of info and art for *all* artists in library, (not just scraped album artists with a unique folder containing all music below it)
- Add options for a) what items are exported, b) exporting items that have not been scraped, c) to export just artwork. 

Export used to take place on the same thread as the  UI, hence a delay during export (with DB access  or file writing) would cause the UI to lock. This PR moves this to be done  asynchronously.  Looking to future improvements separating DB processing from UI, it creates a music lib job queue  equivalent to that created for video in #6406 by Montellese.  

Finally there are also some minor changes to the UI when **adding a music source**
- On adding a new music source to the library, before scanning prompt optionally enter Artist Info Folder setting. Locate folder directly, not go to settings,  as in middle of adding source.
- When scan is initiated from file view context menu the option to do full rescan of tags (allowing changed system settings and tag processing to be applied regardless of files themselves being unchanged) is offered. Unnecessary when adding a new music source to the library, or from normal "Update Library" facility. This replaces the <promptfulltagscan> advanced setting added in #11557.


xbmc/interfaces/builtins/LibraryBuiltins.cpp
```diff
@@ -160,18 +162,89 @@ static int ExportLibrary(const std::vector<std::string>& params)
...
+*           params[3,...] = "unscraped" to include unscraped items
+*           params[3,...] = "overwrite" to overwrite exitsing files.
+*           params[3,...] = "artwork" to include images such as thumbs and fanart.
+*           params[3,...] = "skipnfo" to not include nfo files (just art).
+*           params[3,...] = "ablums" to include albums.
+*           params[3,...] = "albumartists" to include album artists.
+*           params[3,...] = "songartists" to include song artists.
+*           params[3,...] = "otherartists" to include other artists.
+*/
+static int ExportLibrary2(const std::vector<std::string>& params)
```
Rechi: (non-void) function misses a return at the end

xbmc/music/infoscanner/MusicInfoScanner.cpp
```diff
@@ -1617,17 +1756,19 @@ bool CMusicInfoScanner::ResolveMusicBrainz(const std::string &strMusicBrainzID,
...
+    level = 3; //Don't go up more than 2 levels of folders
 
   // check thumb
-  std::string strFolder;
   std::string thumb;
   if (!artist.strPath.empty())
   {
     strFolder = artist.strPath;
-    for (int i = 0; i < 3 && thumb.empty(); ++i)
+    for (int i = 0; i < level && thumb.empty(); ++i)
```
Rechi: comparison between signed and unsigned integer

xbmc/music/infoscanner/MusicInfoScanner.cpp
```diff
@@ -1647,7 +1788,7 @@ std::map<std::string, std::string> CMusicInfoScanner::GetArtistArtwork(const CAr
   if (!artist.strPath.empty())
   {
     strFolder = artist.strPath;
-    for (int i = 0; i < 3 && fanart.empty(); ++i)
+    for (int i = 0; i < level && fanart.empty(); ++i)
```
Rechi: comparison between signed and unsigned integer

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -1808,9 +1815,23 @@ bool CGUIDialogVideoInfo::ManageVideoItemArtwork(const CFileItemPtr &item, const
       noneitem->SetIconImage("DefaultVideo.png");
   }
   else
-  {
-    std::string strThumb = URIUtils::AddFileToFolder(artistPath, "folder.jpg");
-    if (XFILE::CFile::Exists(strThumb))
+  {    
+    std::string strThumb;
+    bool existsThumb;
```
Rechi: may be used uninitialized

Rechi (dismissed):
please fix those compiler warnings

xbmc/settings/dialogs/GUIDialogLibExportSettings.cpp
```diff
@@ -0,0 +1,376 @@
...
+#include "utils/URIUtils.h"
+#include "filesystem/Directory.h"
+
+#define SETTING_MUSIC_EXPORT_FILETYPE "musiclibrary.exportfiletype"
+#define SETTING_MUSIC_EXPORT_FOLDER "musiclibrary.exportfolder"
+#define SETTING_MUSIC_EXPORT_ITEMS "musiclibrary.exportitems"
+#define SETTING_MUSIC_EXPORT_UNSCRAPED "musiclibrary.exportunscraped"
+#define SETTING_MUSIC_EXPORT_OVERWRITE "musiclibrary.exportoverwrite"
+#define SETTING_MUSIC_EXPORT_ARTWORK "musiclibrary.exportartwork"
+#define SETTING_MUSIC_EXPORT_SKIPNFO "musiclibrary.exportskipnfo"
```
xhaggi: Please use `const` instead of `#define`

xbmc/interfaces/builtins/LibraryBuiltins.cpp
Rechi: thanks for fixing
As this function only returns `-1` and `0` you could switch from int return type to bool.

xbmc/settings/dialogs/GUIDialogLibExportSettings.cpp
DaveTBlake: @xhaggi I used `#define` to be consistent with other dialogs e.g. `CGUIDialogAudioSubtitleSettings`, `CGUIDialogNetworkSetup`, `GUIDialogPVRTimerSetting`, `CGUIDialogVideoSettings` etc. I could not find anywhere that used `const` in a similar context. Maybe all need to change, or is `#define` OK?

xhaggi: We should prefer const over define for newer code. Same with cast and c++ style cast.

xbmc/music/MusicDatabase.cpp
```diff
@@ -6232,204 +6487,331 @@ std::string CMusicDatabase::GetItemById(const std::string &itemType, int id)
...
   }
   catch (...)
   {
     CLog::Log(LOGERROR, "%s failed", __FUNCTION__);
     iFailCount++;
   }
 
-  if (progress)
-    progress->Close();
+  if (progressDialog != NULL)
```
xhaggi: `nullptr` is preferred over `NULL`, but `if (progressDialog)` should be enough here and everywhere.

xbmc/video/dialogs/GUIDialogVideoInfo.cpp
```diff
@@ -1703,15 +1703,22 @@ bool CGUIDialogVideoInfo::ManageVideoItemArtwork(const CFileItemPtr &item, const
...
+  std::string artistOldPath;
   std::string artType = "thumb";
   if (type == MediaTypeArtist)
   {
     CMusicDatabase musicdb;
     if (musicdb.Open())
-    {
-      idArtist = musicdb.GetArtistByName(item->GetLabel());
-      if (idArtist >= 0 && musicdb.GetArtistPath(idArtist, artistPath))
+    {
```
xhaggi: trailing whitespace

xbmc/settings/dialogs/GUIDialogLibExportSettings.cpp
```diff
@@ -0,0 +1,376 @@
...
+  if (settingControl != NULL && settingControl->GetControl() != NULL)
+    SET_CONTROL_FOCUS(settingControl->GetID(), 0);
+}
+
+int CGUIDialogLibExportSettings::GetExportItemsFromSetting(SettingConstPtr setting)
+{
+  std::shared_ptr<const CSettingList> settingList = std::static_pointer_cast<const CSettingList>(setting);
+  if (settingList->GetElementType() != SettingType::Integer)
+  {
+    CLog::Log(LOGERROR, "CGUIDialogLibExportSettings::GetItemsFromSetting - wrong items element type");
```
xhaggi: `CGUIDialogLibExportSettings::%s` and `__FUNCTION__`

xbmc/settings/dialogs/GUIDialogLibExportSettings.cpp
```diff
@@ -0,0 +1,376 @@
...
+    return 0;
+  }
+  int exportitems = 0;
+  std::vector<CVariant> list = CSettingUtils::GetList(settingList);
+  for (const auto &value : list)
+  {
+    if (!value.isInteger())
+    {
+      CLog::Log(LOGERROR, "CGUIDialogLibExportSettings::GetItemsFromSetting - wrong items value type");
```
xhaggi: same here

xhaggi (dismissed):

xbmc/settings/dialogs/GUIDialogLibExportSettings.h
```diff
@@ -0,0 +1,66 @@
...
+*  <http://www.gnu.org/licenses/>.
+*
+*/
+
+#include <map>
+
+#include "settings/dialogs/GUIDialogSettingsManualBase.h"
+#include "settings/LibExportSettings.h"
+
+class CGUIDialogLibExportSettings : public CGUIDialogSettingsManualBase
```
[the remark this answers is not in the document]
DaveTBlake: Plan is for video use too in future. But this PR is big enough so leaving video for separate one.

xbmc/music/MusicDatabase.cpp
```diff
@@ -1399,6 +1403,16 @@ bool CMusicDatabase::GetArtistExists(int idArtist)
   return false;
 }
 
+int CMusicDatabase::GetLastArtist()
+{
+  std::string strSQL = "SELECT MAX(idArtist) FROM artist LIMIT 1";
```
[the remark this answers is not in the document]
DaveTBlake: I have a vague memory that there was a reason for LIMIT either in SQLite or one of the MySQL derivatives. leaving this while I check

xbmc/music/MusicDatabase.cpp
```diff
@@ -5375,56 +5424,193 @@ bool CMusicDatabase::GetArtistPath(int idArtist, std::string &basePath)
...
+  if (!strAlbumPath.empty())
+  {  
+    // Get last folder from full path
+    strFolder = strAlbumPath;
+    // Remove end slash
+    strFolder.erase(strFolder.size() - 1, 1);
+    //Find previous slash
+    size_t iPosSlash = strFolder.find_last_of("/\\");
+    folderfound = (iPosSlash != std::string::npos);
+    if (folderfound)
```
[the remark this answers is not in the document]
DaveTBlake: Always a utility I miss, thank you for this

xbmc/music/MusicDatabase.cpp
```diff
@@ -5375,56 +5424,193 @@ bool CMusicDatabase::GetArtistPath(int idArtist, std::string &basePath)
...
+      // Check paths to see folder name derived this way is unique for the (first) albumartist.
+      // Could have different albums on different path with same first album artist and folder name
+      // or duplicate albums from separate music files on different paths
+      // At least one will have mbid, so append it to start of mbid to folder.
+      std::string strSQL = PrepareSQL("SELECT DISTINCT album_artist.idAlbum FROM album_artist "
+        "JOIN song ON album_artist.idAlbum = song.idAlbum "
+        "JOIN path on path.idPath = song.idPath "
+        "WHERE album_artist.iOrder = 0 "
+        "AND album_artist.idArtist = %ld "
+        "AND path.strPath LIKE '%%%s%%'",
```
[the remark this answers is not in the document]
DaveTBlake: Yes, the effect of PrepareSQL and need for double up to get "%" in final query

xbmc/settings/dialogs/GUIDialogLibExportSettings.cpp
DaveTBlake: These other dialogs were all visited recently (#12251), but `#define` not replaced? Sorry to be quibbling, but this just seems an inconsistent change to make.

xbmc/settings/LibExportSettings.h
```diff
@@ -0,0 +1,66 @@
...
+  ELIBEXPORT_ALBUMS = 0x0010,
+  ELIBEXPORT_ALBUMARTISTS = 0x0020,
+  ELIBEXPORT_SONGARTISTS = 0x0040,
+  ELIBEXPORT_OTHERARTISTS = 0x0080,
+  ELIBEXPORT_ARTWORK = 0x0100,
+  ELIBEXPORT_NFOFILES = 0x0200,
+  ELIBEXPORT_ACTORTHUMBS = 0x0400
+};
+
+class CLibExportSettings
```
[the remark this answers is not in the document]
DaveTBlake: I think a class, there are methods with potential for more once extended to video export too.

xbmc/settings/dialogs/GUIDialogLibExportSettings.cpp
DaveTBlake: I assume what you are suggesting is replace them with

xbmc/music/MusicDatabase.cpp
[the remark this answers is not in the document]
DaveTBlake: OK, seems I misremembered and was thinking about situations in MySQL where ORDER BY DESC LIMIT 1 is quicker than MAX()

xbmc/settings/dialogs/GUIDialogLibExportSettings.cpp
xhaggi: After a second look you should move them to `CSettings` h/cpp

xbmc/music/MusicDatabase.cpp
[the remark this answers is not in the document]
xhaggi: Ok

xbmc/settings/dialogs/GUIDialogLibExportSettings.cpp
DaveTBlake: That makes some sense. I have to say that it has been far from obvious how a dialog that remembers previous values should be best implemented. I have tried to follow existing examples, with varied success.

xbmc/settings/LibExportSettings.h
```diff
@@ -0,0 +1,66 @@
...
+{
+public:
+  CLibExportSettings();
+  ~CLibExportSettings() = default;
+
+  bool operator!=(const CLibExportSettings &right) const;
+  bool GetExportItem(ELIBEXPORTOPTIONS item) const;
+  std::vector<int> GetExportItems() const;
+  bool GetSinglefile() const;
+  bool GetToLibFolders() const;
```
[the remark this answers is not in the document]
DaveTBlake: Probably should be called `IsItemExported()`
I started intending to have class with private attributes, but ended up leaving attributes public for convenience. My idea is that the way the list of items to be exported is held  should be encapsulated in the class, and routines that use the class as a parameter  don't need to know.

xbmc/settings/dialogs/GUIDialogLibExportSettings.cpp
xhaggi: Adding the export settings to to media settings -> music -> export should not be an issue if you asked me. We do that for various other settings too.

xbmc/settings/LibExportSettings.h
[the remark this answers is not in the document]
xhaggi: That's the reason why I ask if we really need a class, a struct should be used for a plain-object with no class functionality.

xbmc/settings/dialogs/GUIDialogLibExportSettings.cpp
DaveTBlake: Media settings are a bit of a jumble - some handled by CServiceBroker, others written by CMediaSettings directly to XML. The previous export criteria only want displaying when you are about to export, so don't belong on the media settings screen itself. Coming up on a dialog is correct. But I will follow the PVR settings (most recent addition I think) and switch to `const` in CSettings as you suggest.

xbmc/settings/LibExportSettings.h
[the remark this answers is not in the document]
DaveTBlake: I think there is some (admitedly minimal) class functionality, I just need to make a better job of it. Thanks for taking the time to look at this and all the corrections.

xbmc/settings/dialogs/GUIDialogLibExportSettings.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+using namespace ADDON;
+using namespace KODI::MESSAGING;
+
+using KODI::MESSAGING::HELPERS::DialogResponse;
+
+CGUIDialogLibExportSettings::CGUIDialogLibExportSettings()
+  : CGUIDialogSettingsManualBase(WINDOW_DIALOG_LIBEXPORT_SETTINGS, "DialogSettings.xml")
+{
+  m_destinationChecked = false;
```
garbear: preferred c++11 style is initialize in the header

xbmc/settings/LibExportSettings.h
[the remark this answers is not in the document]
garbear: I agree this should be a class. The member variables should be made private to improve encapsulation.

xbmc/settings/LibExportSettings.cpp
```diff
@@ -0,0 +1,96 @@
+/*
+ *      Copyright (C) 2005-2015 Team KODI
```
garbear: New files should carry a 2017 date

xbmc/settings/LibExportSettings.h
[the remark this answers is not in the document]
DaveTBlake: Have improved encapsulation

DaveTBlake:
Thanks @Rechi, @xhaggi and @garbear for the compiler warnings and C++ style corrections, and for taking the time to review this.

The overall design has been discussed on the forum https://forum.kodi.tv/showthread.php?tid=319284 and pre-PR test builds and been user tested.  I will set off another test build to ensure that in improving code quality I have not broken anything.

xbmc/interfaces/builtins/LibraryBuiltins.cpp
Rechi: @DaveTBlake have you overlooked the above comment, with maybe switching return type to bool?

DaveTBlake: Not overlooked. I have a bigger question over if I can make a breaking change to the builtin interface for `ExportLibrary()` rather than have `ExportLibrary2()`at all, so this routine would go. Would change return type then potentially

DaveTBlake:
Test build 328feae8f7 available on the mirrors 
http://mirrors.kodi.tv/test-builds/windows/win32/KodiSetup-20171010-328feae8f7-HEAD-x86.exe

zag2me:
Tested but I don't like the new dialogue asking for location on a first scan, seems an unnecessary step(can't it in default userdata?). Will update the forum thread with other suggestions.

DaveTBlake:
Fix bug in getting album folder name that was introduced by code tidy-up

Rechi:
@DaveTBlake this needs a rebase

xbmc/interfaces/builtins/LibraryBuiltins.cpp
notspiff: Nothing crucial depends on the builtin. Just change imo we never promised a stable api. The return tyoe is however not easily changeable due to the function ptr table usage.

xbmc/settings/dialogs/GUIDialogLibExportSettings.h
```diff
@@ -0,0 +1,66 @@
+#pragma once
+
+/*
+*      Copyright (C) 2005-2014 Team XBMC
```
garbear: Looks like Visual Studio screwed up the copyright header alignment

notspiff (commented):
Great work, just a few small nits. And a typo which cause real shit. Great to see this improved upon again.

xbmc/guilib/WindowIDs.h
```diff
@@ -82,6 +82,7 @@
 #define WINDOW_DIALOG_PROFILE_SETTINGS    10130
 #define WINDOW_DIALOG_LOCK_SETTINGS       10131
 #define WINDOW_DIALOG_CONTENT_SETTINGS    10132
+#define WINDOW_DIALOG_LIBEXPORT_SETTINGS  10133
```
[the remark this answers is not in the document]
DaveTBlake: Has not been used since before Dharma, so whenever it was used it was a long time ago and not in Git. Therefore I think this is fine but I will change if required.

xbmc/music/jobs/MusicLibraryExportJob.cpp
```diff
@@ -0,0 +1,53 @@
...
+bool CMusicLibraryExportJob::operator==(const CJob* job) const
+{
+  if (strcmp(job->GetType(), GetType()) != 0)
+    return false;
+
+  const CMusicLibraryExportJob* exportJob = dynamic_cast<const CMusicLibraryExportJob*>(job);
+  if (exportJob == NULL)
+    return false;
+
+  return !(m_settings != exportJob->m_settings);
```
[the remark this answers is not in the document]
DaveTBlake: Negations because the settings class only has a != operator

DaveTBlake:
Rebased, and minor corrections made. Thanks for review @notspiff  good to have your input.

I'm hoping to merge this soonest to get it (and hopefully some additional things that depend on this that I am waiting to PR) into alpha , if that seems reasonable to you all. I will squish first if we are there.

xbmc/guilib/WindowIDs.h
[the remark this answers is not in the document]
notspiff: nah. then it's fine. just worried about random leftover xml files in skins, but dharma is more than sufficiently back in time.

notspiff:
this was explicitly mentioned as something we wanted in before alpha at devcon, so imho squash and merge if the issues zag ran into in forum is addressed. code is sane and any leftover bugs can be rooted out during alphas/betas.

DaveTBlake:
Have squished prior to merge and letting jenkins build. @xhaggi I dismissed review as actions complete and I think you are happy, hope that is OK.

xhaggi (approved):
Sure, I trust you

xbmc/music/jobs/MusicLibraryExportJob.cpp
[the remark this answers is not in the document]
garbear: Better to add a `==` operator than play negation games, but no big deal

DaveTBlake:
I believe this has solved Trac issues:
https://trac.kodi.tv/ticket/15294
https://trac.kodi.tv/ticket/16644
But will wait a while for user feedback before closing them

PR #12895, opened 2017-10-08
[PVR] Fix CPVRManager ctor to instanciate timers, recordings and chan…

ksooo:
…nelgroups. Otherwise crashes while pvr client not connected to backend may occur.

This change has been runtime tested on macOS, latest kodi master.

@FernetMenta you reported the problem and we discussed the cause and solution on Slack.


Rechi:
test suite crashes on all platforms

ksooo:
> test suite crashes on all platforms

definitely not related to this change

ksooo:
jenkins build this please

Rechi:
On Windows the test suite crash is definitely related to this change.
Just tested it local with latest master with and without this PR. With the PR applied it crashes, without it doesn't crash.

ksooo:
> With the PR applied it crashes, without it doesn't crash.

Any insights where it crashes?

ksooo:
Third commit fixes the crash in teardown of testsuite, which actually was caused by this PR. It actually revealed an old bug.

FernetMenta (approved):

PR #12899, opened 2017-10-09
OpenGL fixes

FernetMenta:
- fix restore state block (rebind to VAO)
- fix limited colour range


FernetMenta:
@MilhouseVH this should fix limited range issues

MilhouseVH:
Tested on Ubuntu (with Estuary Mod skin) and LibreELEC - looks good. Thanks.

PR #12104, opened 2017-05-16
[cmake][addons][windows] mingw: fix x64

Rechi:
## Description
Differentiate between windows 32 and 64 bit builds at the binary addon mingw environment.

## Motivation and Context
Build `game.libretro.2048` for windows x64.
@garbear FYI

## How Has This Been Tested?
compiled `game.libretro.2048` for windows win32 and x64

## Types of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the [Code guidelines](https://codedocs.xyz/xbmc/xbmc/code_guidelines.html) of this project 
- [ ] My change requires a change to the documentation, either Doxygen or wiki
- [ ] I have updated the documentation accordingly
- [x] I have read the [CONTRIBUTING](https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md) document
- [ ] I have added tests to cover my change
- [ ] All new and existing tests passed

PR #12901, opened 2017-10-09
blind attempt at adjusting add-on mingw support to new paths

notspiff:
ran into this while trying to setup appveyor for game add-ons. this has not been adjusted for new paths or something something.


Rechi:
@notspiff everything is correct in that part of code

notspiff:
okay, then. hard to guess what's up when all i get is 'make.exe not found' :)

Rechi:
#12104 was the fix for adding windows x64 support.
Before that `game.libretro.2048` failed to build on x64.

notspiff:
job for game.libretro.2048 fails precisely in the same way so something is not happy on appveyor. there's some registry crap going on and msys2 warns you have to restart the shell (probably for PATH entries). meh.

PR #12903, opened 2017-10-09
[cmake] adjust json-rpc build method

lrusak:
This allows kodi to be built with ninja

ninja fails because cmake calls another instance of cmake without the `-GNinja` export command. So the addon.xml never gets generated because ninja doesn't know about it. Then the install of the addon.xml fails because it was never generated.

### DIFF
--- a/cmake/scripts/common/GenerateVersionedFiles.cmake
+++ b/cmake/scripts/common/GenerateVersionedFiles.cmake
@@ -11,14 +11,12 @@ function(generate_versioned_file _SRC _DEST)
   file(WRITE ${CMAKE_BINARY_DIR}/${_DEST} "${file_content}")
 endfunction()
 
+# workaround for xbmc.json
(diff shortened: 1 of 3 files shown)


wsnipex:
Looks like this breaks rebuilding jsonrpc when the schema is changed.

lrusak:
@wsnipex you mean if the version.txt is changed then it won't be detected by cmake and recompiled?

wsnipex:
I mean https://github.com/xbmc/xbmc/blob/master/xbmc/interfaces/json-rpc/schema/CMakeLists.txt#L1-L5 and addon.xml.in.

lrusak:
Any other ideas? I'm not sure the best way to do this.

wsnipex:
did you try simpy passing the generator to the cmake command?

lrusak:
Yea I did. I tried using `-G${CMAKE_GENERATOR} but that didn't work.

The command shows up in the ninja.build file but it doesn't seem to run. I'm not sure why.

PR #12907, opened 2017-10-10
RetroPlayer: Fix black screen on OpenGL

garbear:
After #12841, RetroPlayer on OpenGL systems shows a black screen. My solution here is to switch to using a GUI texture and slideshow code like we do for OpenGLES. Long term, Kodi should get first-class OpenGL(ES) renderers, but I won't be able to work on this until after DevCon.

## How Has This Been Tested?
Tested on Linux and OSX.

## Types of change

### DIFF
--- a/xbmc/cores/RetroPlayer/rendering/RenderContext.cpp
+++ b/xbmc/cores/RetroPlayer/rendering/RenderContext.cpp
@@ -139,16 +139,16 @@ int CRenderContext::GUIShaderGetUniCol()
   return -1;
 }
 
-#if defined(HAS_DX)
 CGUIShaderDX* CRenderContext::GetGUIShader()
 {
+#if defined(HAS_DX)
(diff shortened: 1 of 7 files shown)


FernetMenta:
> Long term, Kodi should get a first-class OpenGL(ES) renderer

a combined OpenGL / GLES won't happen. I did intentionally split those two and will continue this path by splitting those completely. Having a dependency just makes it harder to upgrade i.e. OpenGL or do some other maintenance work without the need of placing crappy ifdefs into the code.

garbear:
Changed to read "first-class OpenGL(ES) renderers", as I intend to separate these as well when we move away from the GUI texture approach.

FernetMenta:
btw different components like vp, rp, gui, pictures, or addons have different requirements how to render their objects. keeping rendering of those components separate from each other is the right approach. in the conetxt of OpenGl each component will get its own vertex array object.

garbear:
jenkins build this please

PR #12915, opened 2017-10-11
VideoPlayer: render gui layer on first frame

FernetMenta:
Do not render gui layer as long as we don't have video.


FernetMenta:
confirmed working by @stefansaraev 
OSX build error unrelated

PR #12917, opened 2017-10-12
VideoPlayer: move addon interface to dedicated folder

FernetMenta:
see title.

@hudokkow would be great if you could change addon-bindings that it accepts folders too.


ksooo:
Not sure whether the files you moved to `xbmc/cores/VideoPlayer/Interface/Addon` better should go into `xbmc/addons/kodi-addon-dev-kit/include/kodi`, where all other public addon API stuff resides?

FernetMenta:
IMO it does not. It's confusing and wrong that a local component has to pull headers from a messy folder that is more a build kit than a dev kit. Components are managed best in keeping the files together under a single folder.

hudokkow:
In the past we resisted the urge to use globbing whenever we can because cmake fails to see a changed file as "new" and won't pick the file again during a rebuild.

In other words, build may fail and lead to dev frustration because there is no way for the dev to know this is standard cmake behavior.
At least that's what I recall. Might be wrong. Haven't touched cmake stuff in a while.

Anyway, don't know if we want to change the current behavior but ping @Rechi, @wsnipex and @fetzerch for more opinions.

ksooo:
> IMO it does not. It's confusing and wrong that a local component has to pull headers from a messy folder that is more a build kit than a dev kit. Components are managed best in keeping the files together under a single folder.

 Good point. Will think about it a bit more and maybe move PVR API headers to `xbmc/pvr/interface/addon`, templating what you did here.

FernetMenta:
If this is not possible/wanted, I ask for a machanism that allows binding withing the boundaries of a single conmponent. This is Interface/Addon in this case.

hudokkow:
I'm not sure I follow. You mean another component like `kodi-addon-dev`?
In this case, a new component named `kodi-interface-addon`?

FernetMenta:
Currently we have a single addon-bindings for all addons. I don't want to touch global files when adding a file to VP interface. I want VP to have its own addon-bindings. Global addon-dev should just pull the info from components that have extension points.

wsnipex:
it is doable, but I fail to see the difference it makes which file you have to touch.

FernetMenta:
globals are evil regardless if they are in c code or in the build system. assume you want to add a header shared with addons. currently you have to edit a global addon-bindings.mk and some install script. that is very uncommon to developers and error prone because it gets forgotton. Even if you you know there is some global file, it is hard to locate.
Having a folder belonging to the components and a file inside this folder to specify what headers to be exposed is much better.

FernetMenta:
jenkins build this please

wsnipex:
addon-bindings.mk is a leftover from autotools, we can easily migrate this to cmake and/or reuse it in the install scripts. This doesn't necessarily need to be a single file, reading several is ok as well. The format is not really important either, I'd suggest using cmake syntax for uniformity.

What we must not do is glob files, as hudokkow already explained. So no reading directories.

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

FernetMenta:
build failures unrelated.

PR #12918, opened 2017-10-12
[amlogic] Fix RetroPlayer on AMLogic hardware

garbear:
This fixes the following error on AMLogic:

```
ERROR: Failed to create RetroPlayer - no process info registered
```

The problem is that I used the VP registration as a template, and if no ProcessInfo is registered for VP it uses a default one.

RP, on the other hand, has no default ProcessInfo, so it must be registered explicitly.

## Motivation and Context
Reported in the LE slack.

## How Has This Been Tested?
Untested.

## Types of change


Raybuntu:
@garbear Thanks for the fix, I've this PR and it's working again in my builds.

garbear:
Jenkins build this please

PR #12923, opened 2017-10-13
windowing/gbm: allow scanning /dev/dri/card[0-9] to find the correct device

lrusak:
previously I just hardcoded /dev/dri/card0

This scans through /dev/dri/card[0-9] and tries to open it with a set of modules.

This also allows you to have multiple cards but will only continue if one has a connector/encoder.

I wish libdrm had a helper here but this is how they do it also
https://cgit.freedesktop.org/mesa/drm/tree/tests/util/kms.c#n153


tomlohave:
The right card is chosen with this patch (card1) on cuboxi. Thanks

PR #12927, opened 2017-10-15
Show watched instead of collection icon for moviesets

fetzerch:
## Description
I'd like to improve the movie views in Estuary. Currently when moviesets are enabled it's not possible to see if the movieset has been watched already or not:

screenshot001
(all movies in the selected set have been watched)

This change drops the collection icon in favor of the watched state:
screenshot000

I'm not sure if that's the best solution from the user's perspective. Maybe it's possible to show the collection icon if the set is not watched and only show the watched check when it's watched.

## Motivation and Context
Be able to see if a movieset has been fully watched or contains unwatched movies.

## How Has This Been Tested?
See screenshots above.

## Types of change

## Checklist:

### DIFF
--- a/addons/skin.estuary/xml/Variables.xml
+++ b/addons/skin.estuary/xml/Variables.xml
@@ -314,14 +314,12 @@
 		<value condition="ListItem.IsRecording">windows/pvr/record.png</value>
 		<value condition="ListItem.IsPlaying">overlays/watched/OverlayPlaying-List.png</value>
 		<value condition="ListItem.IsResumable">overlays/watched/resume.png</value>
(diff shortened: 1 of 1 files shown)


phil65:
No strong opinion, but I think this is a step backwards since sets cannot get recognized easily in lists anymore.
In addition, this PR would have to remove the set image since it isnt used anymore anywhere after this change.

ksooo:
> but I think this is a step backwards since sets cannot get recognized easily in lists anymore.

I'm of the same opinion. This is a step back.

phunkyfish:

PR #12928, opened 2017-10-15
[PVR] Fix PVR startup / shutdown.

ksooo:
Couple of fixes to prevent crashes / deadlocks while starting / stopping PVR Manager (Kodi startup/shutdown, Profile switch, PVR client addon enable/disable/install/uninstall).

- Make startup and shutdown actions "symmetric".
- Prevent startup if another startup is already in progress (we had that already for stop, btw)
- Make `CPVRClients::UpdateAddons` threadsafe
- On addon manager enable/disable event, restart PVR Mananger asynchronously
- Fix a bug with addon manager event subscription, that resulted in one additional subscriber per PVR Manager restart  
- some logging improvements for PVR Manager startup and shutdown

I carefully tested this on macOS, latest Kodi master, especially together with another WIP of mine - improvements for multiple enabled PVr addons (which increases complexity and deadlock/crash risk).

### DIFF
--- a/xbmc/pvr/addons/PVRClients.cpp
+++ b/xbmc/pvr/addons/PVRClients.cpp
@@ -53,6 +53,8 @@ CPVRClients::CPVRClients(void) :
     m_bIsPlayingRecording(false),
     m_bIsPlayingEpgTag(false)
 {
+  CServiceBroker::GetAddonMgr().RegisterAddonMgrCallback(ADDON_PVRDLL, this);
+  CServiceBroker::GetAddonMgr().Events().Subscribe(this, &CPVRClients::OnAddonEvent);
 }
 
 CPVRClients::~CPVRClients(void)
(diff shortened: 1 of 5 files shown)


xbmc/pvr/addons/PVRClients.cpp
```diff
@@ -1078,8 +1092,10 @@ void CPVRClients::UpdateAddons(void)
         CLog::Log(LOGERROR, "%s - failed to create add-on %s, status = %d", __FUNCTION__, addon->Name().c_str(), status);
         if (status == ADDON_STATUS_PERMANENT_FAILURE)
         {
+          lock.Leave();
```
FernetMenta: CSingleExit

ksooo: Right. Thanks.

ksooo:
jenkins build this please

ksooo:
jenkins build this please

MartijnKaijser:
Seems test suite doesn't like it.
Jenkins build this please

ksooo:
Yeah, will investigate and fix later.

ksooo:
jenkins build this please

PR #12929, opened 2017-10-15
RetroPlayer: Cache/pre-compile output shaders

VelocityRa:
## Description

* Prevents stutering induced from the compilation of output shaders, when switching between them in the Video Filters OSD.

* Also fixes a bug where the shaders would compile continously in the Video Filters OSD.

I initially started working on an `RPWinOutputShaderManager` that would hold the output shaders and map them to scaling methods to make `RPWinRenderer` usage of shaders simpler, but it was getting a bit out of hand so I just did the logic in the renderer.

We won't have many other supported scaling types for this anyway, plus the output shader probably won't be used very much when [video shaders](https://github.com/garbear/xbmc/pull/86) are merged :)

## Motivation and Context

UI stuttering should be avoided when possible and in this case it was pretty trivial to do.

Also fixes the aforementioned bug. Sorry about that, I'll test stuff better next time.

## How Has This Been Tested?

Tested by loading a game, going into the filters osd and switching between the 2 scaling options.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/cores/RetroPlayer/rendering/VideoRenderers/RPWinRenderer.cpp
+++ b/xbmc/cores/RetroPlayer/rendering/VideoRenderers/RPWinRenderer.cpp
@@ -20,6 +20,7 @@
 
 #include "RPWinRenderer.h"
 #include "cores/RetroPlayer/rendering/RenderContext.h"
+#include "cores/RetroPlayer/rendering/RenderTranslator.h"
 #include "cores/RetroPlayer/rendering/RenderVideoSettings.h"
 #include "cores/RetroPlayer/rendering/VideoShaders/windows/RPWinOutputShader.h"
 #include "guilib/D3DResource.h"
@@ -169,12 +170,14 @@ AVPixelFormat CWinRenderBuffer::GetPixFormat(DXGI_FORMAT dxFormat)
 
 // --- CWinRenderBufferPool ----------------------------------------------------
 
+CWinRenderBufferPool::CWinRenderBufferPool()
+{
+  CompileOutputShaders();
+}
+
 bool CWinRenderBufferPool::IsCompatible(const CRenderVideoSettings &renderSettings) const
 {
-  if (!CRPWinRenderer::SupportsScalingMethod(renderSettings.GetScalingMethod()))
-    return false;
-
-  return true;
+  return GetShader(renderSettings.GetScalingMethod()) != nullptr;
 }
 
 IRenderBuffer *CWinRenderBufferPool::CreateRenderBuffer(void *header /* = nullptr */)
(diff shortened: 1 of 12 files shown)


xbmc/cores/RetroPlayer/rendering/VideoRenderers/RPWinRenderer.h
```diff
@@ -123,12 +123,11 @@ namespace RETRO
...
   private:
-    void CompileOutputShader(ESCALINGMETHOD scalingMethod);
-    void HandleScalingChange();
+    void CompileOutputShaders();
     void Render(CD3DTexture *target);
 
-    std::unique_ptr<CRPWinOutputShader> m_outputShader;
-    ESCALINGMETHOD m_prevScalingMethod = VS_SCALINGMETHOD_AUTO;
+    std::unique_ptr<CRPWinOutputShader> m_outputShaderLinear;
+    std::unique_ptr<CRPWinOutputShader> m_outputShaderNearest;
```
garbear: How about a map from ESCALINGMETHOD to CRPWinOutputShader? That way we don't have to add a member variable for every scaling method.

xbmc/cores/RetroPlayer/rendering/VideoRenderers/RPWinRenderer.cpp
```diff
@@ -269,19 +260,29 @@ bool CRPWinRenderer::SupportsScalingMethod(ESCALINGMETHOD method)
...
-        m_context.UseLimitedColor() ? 1 : 0);
+      // Pick appropriate output shader depending on the scaling method of the renderer
+      ESCALINGMETHOD scalingMethod = m_renderSettings.VideoSettings().GetScalingMethod();
+      CRPWinOutputShader* outputShader;
+
+      if (scalingMethod == VS_SCALINGMETHOD_NEAREST)
+        outputShader = m_outputShaderNearest.get();
+      else
+        // default to linear interpolation
+        outputShader = m_outputShaderLinear.get();
```
garbear: Might as well default to `CRPProcessInfo::GetDefaultScalingMethod()`

xbmc/cores/RetroPlayer/rendering/VideoRenderers/RPWinRenderer.h
VelocityRa: Yeah I was trying to avoid that and go for the solution with less code, but I'll try it.

garbear: More code is better than less code if it's scalable and maintainable

garbear:
@VelocityRa https://github.com/VelocityRa/xbmc/pull/25

garbear:
jenkins build this please

PR #12930, opened 2017-10-16
[buildtools][windows] don't fail if patch was already applied

Rechi:
## Description
Skip a patch if it was already applied in a previous run

## Motivation and Context
With cd670240 the script exits if applying a  patch fails.
Patch also returns a non zero exit status for already applied patches (even if invoked with the `--forward`  / `-N` parameter).
Therefore check if the patch was already applied and skip it in that case.

## How Has This Been Tested?
run make-mingwlibs more then once

## Types of change

## Checklist:

### DIFF
--- a/tools/buildsteps/windows/buildhelpers.sh
+++ b/tools/buildsteps/windows/buildhelpers.sh
@@ -129,7 +129,12 @@ do_download() {
   for patch in ${patches[@]}; do
     echo "Applying patch ${patch}"
     if [[ -f $patch ]]; then
-      patch -d $LOCALSRCDIR -i $patch -N -r - || exit $?
+      patch -d $LOCALSRCDIR --dry-run --reverse --force -i $patch 2>&1 > /dev/null
+      if [ $? == 0 ]; then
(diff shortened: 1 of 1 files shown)


afedchin (approved):

tools/buildsteps/windows/buildhelpers.sh
```diff
@@ -129,7 +129,12 @@ do_download() {
   for patch in ${patches[@]}; do
     echo "Applying patch ${patch}"
     if [[ -f $patch ]]; then
-      patch -d $LOCALSRCDIR -i $patch -N -r - || exit $?
+      patch -d $LOCALSRCDIR --dry-run --reverse --force -i $patch 2>&1 > /dev/null
+      if [ $? == 0 ]; then
```
wsnipex: [ $? -eq 0

Rechi:
This solution may also fail if patches touch the same lines of a file.
I revert https://github.com/xbmc/xbmc/commit/cd6702402057f2d7b602fb6ef1cf030dff9da9a5 until a better solution is found.

PR #12934, opened 2017-10-18
Websocket: fix handshake parser

mglae:
Correct websocket connection header check after #11267 to allow other tokens beside "Upgrade" in the "Connection" header field.

## Description

Implements strtstr() like parsing again and adds case insensitivity by converting the sent string.

## Motivation and Context

Firefox i.e. sends `Connection: keep-alive, Upgrade` 

Fixes Chorus2 "Lost Websocket Connection" error as reported [here](https://forum.kodi.tv/showthread.php?tid=308456) and [here](https://github.com/xbmc/chorus2/issues/274) and let the [PlayThis Plugin](https://github.com/anxdpanic/plugin.video.playthis) work again.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/network/websocket/WebSocketV13.cpp
+++ b/xbmc/network/websocket/WebSocketV13.cpp
@@ -101,8 +101,9 @@ bool CWebSocketV13::Handshake(const char* data, size_t length, std::string &resp
   }
 
   // There must be a "Connection" header with the value "Upgrade"
-  value = header.getValue(WS_HEADER_CONNECTION_LC);
-  if (value == NULL || strnicmp(value, WS_HEADER_UPGRADE, strlen(WS_HEADER_UPGRADE)) != 0)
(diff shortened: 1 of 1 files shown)


Rechi:
With this change `Connection: Upgradee` header would also be accepted, which we don't want.

mglae:
Without this change too.

WebSocketV13.cpp may be improved by replacing all the strnicmp() and find() functions causing this errors, but I doubt this regression fix to be backported is the right place to do it.

yarda:
AWXi with Firefox 56 is also affected by this regression - when the playback is started user gets several: "Failed to connect to websocket" errors and the UI is frozen for cca. 20 seconds until it autoswitches to the HTTP polling.

Rechi:
@mglae can you replace this with https://github.com/xbmc/xbmc/commit/34762e89473703aa53e640e106f6ea320d035065 or should I do it myself?

mglae:
@Rechi please do it

PR #12935, opened 2017-10-18
Websocket: fix handshake parser

mglae:
Correct websocket connection header check after #11267 to allow other tokens beside "Upgrade" in the "Connection" header field.

Backport of #12934


yarda:
It's a regression e.g. breaking remote control from recent Firefox and several other browsers, please backport.

PR #12937, opened 2017-10-19
Update NEON support for cortex-a53

candrews:
Update the NEON configuration for cortex-a53 to use that CPU's crypto capabilities and specifically optimize for that CPU

## Description

This change optimizes the NEON support for the cortex-a53 (used by the Raspberry Pi 3).

The change was inspired by https://bugs.gentoo.org/633764#c1

The result of this change is more optimized code generated for the cortex-a53, particularly taking advantage of that CPU's crypto capabilities resulting in (slightly) improved performance. See https://github.com/RetroPie/RetroPie-Setup/issues/1385 (not that this PR does not use `-funsafe-math-optimizations` because it can cause problems - but the performance improvements of `-mcpu=cortex-a53 -mfpu=crypto-neon-fp-armv8` are still relevant).

## Motivation and Context

Improve performance/optimization for users of the cortex-a53.

## How Has This Been Tested?

I did not test this change as I don't have this hardware.

## Types of change

## Checklist:


asavah:
`-mfpu=crypto-neon-fp-armv8` I would not use this as is, while it may benefit the pi3 it could potentially break other platforms, crypto is _optional_ in armv8-a (http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0500e/CJHDEBAF.html), and there are cortex-a53 SoCs which don't have crypto extensions at all, eg. amlogic S905 
I'm not even sure that rpi3 SoC has the crypto instructions (aes, sha*). Maybe @popcornmix can confirm/deny this fact.

stefansaraev:
@asavah for the record, CPU is a hack in the buildsystem intended for RPi only (and should die in fire). it does not affect any other platforms. nobody should use -DWITH_CPU on non-rpi platform.

wsnipex:
looks ok, but I'd like to have this tested before merge. @MilhouseVH could you include this in your nightlies?

MilhouseVH:
LibreELEC applies [this logic](https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/mediacenter/kodi/package.mk#L177-L181) when building Kodi:
```
if [ "$TARGET_ARCH" = "x86_64" ]; then
  KODI_ARCH="-DWITH_CPU=$TARGET_ARCH"
else
  KODI_ARCH="-DWITH_ARCH=$TARGET_ARCH"
fi
```
So for RPi/RPi2 (`TARGET_ARCH=arm`) we specify `-DWITH_ARCH=arm` in which case I don't think this change is likely to have any effect on Kodi when building LibreELEC, and my testing won't prove anything I'm afraid.

As I recall, we have to use `-DWITH_CPU=x86_64` because of soname naming issues (precise details long forgotten).

koenkooi:
And a drive-by comment: -mfloat-abi=hard is a distro setting, not something specific to a CPU. It's the de-facto standard nowadays, so you get away with it most of the times, but e.g. WebOS uses -mfloat-abi=softfp (not to be confused with 'soft').

candrews:
So what are the next steps on this one?

* Do we keep or remove `-mfloat-abi=hard`?
* What is necessary to either merge this or reject it?

wsnipex:
We should make this RBPI specific once again, Then we can keep -mfloat-abi=hard and the optimized neon flags. As has been pointed out, those flags do not work all ARM SOCs and distros.

wsnipex:
superseded by https://github.com/xbmc/xbmc/pull/13638

PR #12940, opened 2017-10-21
Fix MacOSX 10.13 SDK build

lobermann:
Fix the build on MacOS when using the 10.13 SDK Version with XCode 9 but still using 10.12 MacOS

## Description
When updating XCode to the new version it also updated to the SDK Version 10.13, but the OS itself is still on 10.12 . After working on this issue for a day with @Rechi and @FernetMenta this is what I had to change to make the build work again.

What was fixed:

Error: `configure: error: error in configure of --with-sdk=10.13`
Fixed by adding SDK Version 10.13 Support

Error: libgcrypt: `The CPU Jitter random number generator must not be compiled with optimizations. See documentation. Use the compiler switch -O0 for compiling jitterentropy-base.c.`
Most likely caused by the recent update of libgcrypt. Fixed by patching the Makefile to also replace the -Og flag with -O0

Error:
````
dyld: lazy symbol binding failed: Symbol not found: _utimensat
  Referenced from: /Users/Shared/xbmc-depends/x86_64-darwin16.7.0-native/bin/cmake
  Expected in: /usr/lib/libSystem.B.dylib
````
Fixed build issue by updating CMake to 3.9.4, patching it to address the issues on macos and adding additional parameters for the native toolchain build.

I am not sure which of those changes we want to get into master right now. 
The SDK Part is no problem.
The libgcrypt should also not cause any issues on other systems, as the patch is only executed on MacOS. But I can only test it on my mac.
The CMake update worked by just increasing the version number and built without issues. So I assume it is OK for other systems also. I think Windows is also using those dependencies, or? Needs to be tested. Also, the CMake tar.gz needs to be put onto the mirror if merged.

## Types of change

## Checklist:

### DIFF
--- a/tools/depends/configure.ac
+++ b/tools/depends/configure.ac
@@ -322,6 +322,7 @@ case $host in
           10.10);;
           10.11);;
           10.12);;
+          10.13);;
           *)
             AC_MSG_ERROR(error in configure of --with-sdk=$use_sdk)
         esac
--- a/tools/depends/native/cmake-native/01-osx-utimensat.diff
+++ b/tools/depends/native/cmake-native/01-osx-utimensat.diff
@@ -1,0 +1,35 @@
+--- a/Source/kwsys/kwsysPlatformTestsCXX.cxx
++++ b/Source/kwsys/kwsysPlatformTestsCXX.cxx
+@@ -265,6 +265,12 @@ int main()
+ #ifdef TEST_KWSYS_CXX_HAS_UTIMENSAT
+ #include <fcntl.h>
+ #include <sys/stat.h>
++#if defined(__APPLE__)
++#include <AvailabilityMacros.h>
++#if MAC_OS_X_VERSION_MIN_REQUIRED < 101300
++#error "utimensat not available on macOS < 10.13"
++#endif
++#endif
+ int main()
+ {
+   struct timespec times[2] = { { 0, UTIME_OMIT }, { 0, UTIME_NOW } };
+--- a/Utilities/cmlibarchive/libarchive/archive_platform.h
++++ b/Utilities/cmlibarchive/libarchive/archive_platform.h
+@@ -52,6 +52,17 @@
+ #error Oops: No config.h and no pre-built configuration in archive_platform.h.
+ #endif
+ 
++/* On macOS check for some symbols based on the deployment target version.  */
++#if defined(__APPLE__)
++# undef HAVE_FUTIMENS
++# undef HAVE_UTIMENSAT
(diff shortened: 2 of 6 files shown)


hudokkow:
Do we really need the latest cmake RC? I would prefer the latest stable release.

lobermann:
@hudokkow unfortunately yes, as the issue is only fixed in 3.10, and not in 3.9.4 ... and trying to patch 3.9.4 did not work. Not sure when the first official release of 3.10 will be.

hudokkow:
OK. Thanks.

FernetMenta:
I don't thinnk we should bump cmake to a non-released version. Maybe there are not even 10 people out there who run XCode 9 on OSX 10.12 and want to build Kodi. Those can easily patch cmake to rc while all others use a stable release.

MrMC:
https://github.com/MrMC/mrmc/commit/0cab5053593299cde210e22bbe75f69e4587c63d

lobermann:
thanks to the patch from @Rechi it works now with 3.9.4.

Rechi (approved):

Rechi:
@wsnipex cmake version bump is okay for you?

lobermann:
@FernetMenta I think you have done this already, or? Can you share how to use the system cmake?

FernetMenta:
@lobermann I hacked Makefile.include after having run configure. I think cmake-native should be made configurable to this list: https://github.com/xbmc/xbmc/blob/master/tools/depends/native/Makefile#L7

wsnipex:
to use cmake from system, you'd have to add a configure switch and cmake check to depends/configure.ac and fixup https://github.com/xbmc/xbmc/blob/master/tools/depends/Makefile.include.in#L67 to use it

Memphiz:
Damn - i didn't realise that this PR is up. I worked on that during devcon (using xcode9 for osx and ios) and figured the same stuff (and a bit more) out again. Not sure how this will influence now ... which PR do we want to use then? :)

Memphiz:
My approach is btw to use a deploymenttarget to the natvie build which matches the running osx version. This ensures that native binaries that are built with newer SDKs are able to run on the host.

Memphiz:
Also the cmake bump is not needed - the patch is enough ...

Memphiz:
i will merge this PR with mine - because this one is to osx specific and leaves out ios to much. I do a new PR with the merged state ...

Memphiz:
Please continue here:

https://github.com/xbmc/xbmc/pull/12969

PR #8865, opened 2016-01-14
merge 'ok', 'yesno' & 'progress' dialogs xml files

ronie:
merge another set of dialog xml's into a single file:
- DialogOK.xml
- DialogProgress.xml
- DialogYesNo.xml

new xml file is DialogConfirm.xml (working title).
i'm open to other/better names for the xml file.

PR #12943, opened 2017-10-22
[GUI]Add optional 3rd button to YesNoDialog 

DaveTBlake:
The potential for a 3rd button on YesNoDialog was suggested and included in refactoring of various dialogs into  DialogConfirm.xml in #8865. This makes the 3rd button fully availalble to calling routines.

It adds an overload of `CGUIDialogYesNo::ShowAndGetInput` that enables and labels the 3rd button.

~~It modifies the interface to `HELPERS::ShowYesNoDialogText` and `ShowYesNoDialogLines` to do the same. Since the parameters are all variants I have checked every call to these routines and adjusted those that need it to allow for having inserted a parameter.~~
It adds convenience method `HELPERS::ShowYesNoCustomDialog` to invoke a 3 btton dialog, so avoiding a empty parameter or changes to existing calls.

@ronie, @phil65 and @Paxxi perhaps you could check this.

I have a use for the 3rd button in a separate PR,  and have tested it using that.

### DIFF
--- a/xbmc/dialogs/GUIDialogYesNo.cpp
+++ b/xbmc/dialogs/GUIDialogYesNo.cpp
@@ -28,6 +28,7 @@ CGUIDialogYesNo::CGUIDialogYesNo(int overrideId /* = -1 */)
 {
   m_bConfirmed = false;
   m_bCanceled = false;
+  m_bCustom = false;
 }
 
 CGUIDialogYesNo::~CGUIDialogYesNo() = default;
@@ -54,6 +55,13 @@ bool CGUIDialogYesNo::OnMessage(CGUIMessage& message)
           Close();
           return true;
         }
+        if (iControl == CONTROL_CUSTOM_BUTTON)
+        {
+          m_bConfirmed = false;
+          m_bCustom = true;
+          Close();
+          return true;
+        }
       }
     }
     break;
@@ -65,12 +73,16 @@ bool CGUIDialogYesNo::OnBack(int actionID)
 {
   m_bCanceled = true;
   m_bConfirmed = false;
+  m_bCustom = false;
   return CGUIDialogBoxBase::OnBack(actionID);
 }
 
 void CGUIDialogYesNo::OnInitWindow()
 {
-  SET_CONTROL_HIDDEN(CONTROL_CUSTOM_BUTTON);
+  if (!m_strChoices[2].empty())
+    SET_CONTROL_VISIBLE(CONTROL_CUSTOM_BUTTON);
+  else
+    SET_CONTROL_HIDDEN(CONTROL_CUSTOM_BUTTON);
   SET_CONTROL_HIDDEN(CONTROL_PROGRESS_BAR);
   SET_CONTROL_FOCUS(CONTROL_NO_BUTTON, 0);
 
@@ -102,6 +114,7 @@ bool CGUIDialogYesNo::ShowAndGetInput(CVariant heading, CVariant line0, CVariant
     dialog->SetAutoClose(autoCloseTime);
   dialog->SetChoice(0, !noLabel.empty() ? noLabel : 106);
   dialog->SetChoice(1, !yesLabel.empty() ? yesLabel : 107);
+  dialog->SetChoice(2, "");
   dialog->m_bCanceled = false;
   dialog->Open();
 
@@ -117,6 +130,14 @@ bool CGUIDialogYesNo::ShowAndGetInput(CVariant heading, CVariant text)
 
 bool CGUIDialogYesNo::ShowAndGetInput(CVariant heading, CVariant text, bool &bCanceled, CVariant noLabel /* = "" */, CVariant yesLabel /* = "" */, unsigned int autoCloseTime)
 {
+  int result = ShowAndGetInput(heading, text, noLabel, yesLabel, "", autoCloseTime);
+
+  bCanceled = result == -1;
+  return result == 1;
+}
+
+int CGUIDialogYesNo::ShowAndGetInput(CVariant heading, CVariant text, CVariant noLabel, CVariant yesLabel, CVariant customLabel, unsigned int autoCloseTime)
+{
   CGUIDialogYesNo *dialog = g_windowManager.GetWindow<CGUIDialogYesNo>(WINDOW_DIALOG_YES_NO);
   if (!dialog)
     return false;
@@ -126,12 +147,21 @@ bool CGUIDialogYesNo::ShowAndGetInput(CVariant heading, CVariant text, bool &bCa
   if (autoCloseTime)
     dialog->SetAutoClose(autoCloseTime);
   dialog->m_bCanceled = false;
+  dialog->m_bCustom = false;
   dialog->SetChoice(0, !noLabel.empty() ? noLabel : 106);
   dialog->SetChoice(1, !yesLabel.empty() ? yesLabel : 107);
+  dialog->SetChoice(2, customLabel);  // Button only visible when label is not empty
+
   dialog->Open();
 
-  bCanceled = dialog->m_bCanceled;
-  return (dialog->IsConfirmed()) ? true : false;
+  if (dialog->m_bCanceled)
+    return -1;
+  else if (dialog->m_bCustom)
+    return 2;
+  else if (dialog->IsConfirmed())
+    return 1;
+  else
+    return 0;
 }
 
 int CGUIDialogYesNo::ShowAndGetInput(const KODI::MESSAGING::HELPERS::DialogYesNoMessage& options)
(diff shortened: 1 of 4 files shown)


xbmc/dialogs/GUIDialogYesNo.cpp
```diff
@@ -65,12 +74,16 @@ bool CGUIDialogYesNo::OnBack(int actionID)
...
   m_bCanceled = true;
   m_bConfirmed = false;
+  m_bCustomClicked = false;
   return CGUIDialogBoxBase::OnBack(actionID);
 }
 
 void CGUIDialogYesNo::OnInitWindow()
 {
-  SET_CONTROL_HIDDEN(CONTROL_CUSTOM_BUTTON);
+  if (m_bCustomEnabled)
```
Razzeee: Maybe named a little bad? Enabled is something that can be disabled for me. This seems to set visibility, so shouldn't it be `m_bCustomVisible`?

xbmc/dialogs/GUIDialogYesNo.cpp
```diff
@@ -134,6 +147,35 @@ bool CGUIDialogYesNo::ShowAndGetInput(CVariant heading, CVariant text, bool &bCa
   return (dialog->IsConfirmed()) ? true : false;
 }
 
+int CGUIDialogYesNo::ShowAndGetInput(CVariant heading, CVariant text, CVariant noLabel, CVariant yesLabel, CVariant customLabel, unsigned int autoCloseTime)
+{
```
Razzeee: Just a note, this duplicates code. Yes, the other implementations also do this. I would like to look at it when it's merged and try to combine it. Just a heads up.

xbmc/dialogs/GUIDialogYesNo.h
```diff
@@ -100,6 +100,16 @@ class CGUIDialogYesNo :
    */
   static bool ShowAndGetInput(CVariant heading, CVariant text, bool &bCanceled, CVariant noLabel, CVariant yesLabel, unsigned int autoCloseTime);
 
+  /*! \brief Show a yes-no dialog with 3rd custom button, then wait for user to dismiss it.
+  \param heading Localized label id or string for the heading of the dialog
+  \param text Localized label id or string for the dialog message
+  \param noLabel Localized label id or string for the no button
+  \param yesLabel Localized label id or string for the yes button
+  \param autoCloseTime Time in ms before the dialog becomes automatically closed
```
Razzeee: Might want to add `customLabel` to the docs here :)

razzeee (commented):
Just some minors

DaveTBlake:
Thanks @Razzeee, fixed the minors including removing at least the duplication I added

Razzeee:
Much appreciated

Rechi:
wrong revision was build
jenkins build this please

mkortstiege (approved):

xbmc/dialogs/GUIDialogYesNo.cpp
```diff
@@ -148,9 +182,15 @@ int CGUIDialogYesNo::ShowAndGetInput(const KODI::MESSAGING::HELPERS::DialogYesNo
     SetChoice(0, options.noLabel);
   if (!options.yesLabel.isNull())
     SetChoice(1, options.yesLabel);
+  if (!options.customLabel.isNull())
+  {
+    m_bCustomVisible = true;
+    SetChoice(2, options.customLabel);
```
xhaggi: we should reset choise 2 to the default at the top too.

xbmc/pvr/PVRManager.cpp
```diff
@@ -962,6 +962,7 @@ bool CPVRManager::CanSystemPowerdown(bool bAskUser /*= true*/) const
                                        CVariant{text},
                                        CVariant{222}, // "Shutdown anyway",
                                        CVariant{19696}, // "Cancel"
+                                       CVariant{ "" }, // No 3rd button
```
xhaggi: why not use a convenience method without the custom button label parameter. In most cases we don't need a custom button.

xbmc/dialogs/GUIDialogYesNo.cpp
```diff
@@ -65,12 +74,16 @@ bool CGUIDialogYesNo::OnBack(int actionID)
...
   m_bCanceled = true;
   m_bConfirmed = false;
+  m_bCustomClicked = false;
   return CGUIDialogBoxBase::OnBack(actionID);
 }
 
 void CGUIDialogYesNo::OnInitWindow()
 {
-  SET_CONTROL_HIDDEN(CONTROL_CUSTOM_BUTTON);
+  if (m_bCustomVisible)
```
xhaggi: if we rely on `m_strChoices[2]` there is no need to hold a bool for the visibility.

```c++
if (m_strChoices[2].empty())
  SET_CONTROL_HIDDEN(CONTROL_CUSTOM_BUTTON);
else
  SET_CONTROL_VISIBLE(CONTROL_CUSTOM_BUTTON);
```

xbmc/dialogs/GUIDialogYesNo.cpp
```diff
@@ -171,5 +217,7 @@ int CGUIDialogYesNo::GetDefaultLabelID(int controlId) const
     return 106;
   else if (controlId == CONTROL_YES_BUTTON)
     return 107;
+  else if (controlId == CONTROL_CUSTOM_BUTTON)
+    return 636;
```
xhaggi: the default should be empty

xbmc/pvr/PVRManager.cpp
DaveTBlake: Because the parameters are variants or int I can not see how to differentiate between
`ShowAndGetInput(522, 20340, 106, 107, 1000)` as giving 
`heading, text, noLabel, yesLabel, autoCloseTime` or
`heading, text, noLabel, yesLabel, customLabel`
Yes we generally use `CVariant{19696}` as a parameter, but `19696` works too.

There are only a few calls that use auto close, and so not too inconvenient to have to give a blank custom button

DaveTBlake: Or did you mean a new function in DialogHelper.cpp rather than an overlay?

Razzeee: I counted 3 or 4 times, that you need autoclose and have to explicitly call it. So I did not mention it.

DaveTBlake: So would you prefer me to leave the parameters to `ShowYesNoDialogText`  and `ShowYesNoDialogLines` unchanged, and add new convenience methods `ShowYesNoOtherDialogText` and `ShowYesNoOtherDialogLines` ?

DaveTBlake: @Paxxi thinks it makes sense to have a new convenience method too, so going with that

xbmc/messaging/helpers/DialogHelper.h
```diff
@@ -74,6 +76,27 @@ DialogResponse ShowYesNoDialogText(CVariant heading, CVariant text, CVariant noL
...
+\param[in]  customLabel       The text to display on the optional 3rd custom button
+                              defaults to empty and button not shown
+\param[in]  autoCloseTimeout  The time before the dialog closes
+                              defaults to 0 show indefinitely
+\return -1 on cancelled, 0 on no, 1 on yes and 2 on 3rd custom response
+\sa ShowYesNoDialogLines
+\sa CGUIDialogYesNo::ShowAndGetInput
+\sa DialogYesNoMessage
+*/
+DialogResponse ShowYesNoOtherDialogText(CVariant heading, CVariant text, CVariant noLabel = CVariant(), CVariant yesLabel = CVariant(),
```
xhaggi: I would suggest `ShowYesNoCustomDialog`. Using the dialog with lines should be deprecated IMO.

xhaggi:
@DaveTBlake thanks for addressing all my comments. Looks way better now.

DaveTBlake:
@xhaggi I agree it looks better now, thank you for your help.

Rechi:
wrong revision was build
jenkins build this please

DaveTBlake:
Think we are good to merge this, so let's see if jenkins can get the right build today.
jenkins build this please

xbmc/messaging/helpers/DialogHelper.h
```diff
@@ -74,6 +76,27 @@ DialogResponse ShowYesNoDialogText(CVariant heading, CVariant text, CVariant noL
                                    CVariant yesLabel = CVariant(), uint32_t autoCloseTimeout = 0);
 
 /*!
+\brief This is a helper method to send a threadmessage to open a Yes/No dialog box with a cutom button
```
Throne3d: (Typo: 'cutom'?)

garbear: and wrap comments at 80 chars please

ksooo: > and wrap comments at 80 chars please

 Is this part of our coding guidelines? I'm all in for it, but nobody seems to do this...

PR #12945, opened 2017-10-22
drop global CAudioSettings

FernetMenta:
Having global audio dsp settings is wrong and can't work. CApplication does not know when audio player switches to the next song. x-fade makes it even more complex. Since DSP is half removed anyway, this change does not harm.


FernetMenta:
@AchimTuran agree?

AchimTuran:
Yes

PR #12946, opened 2017-10-22
Reenable advanced settings forcevaapienabled

lethalwp:
modified:   xbmc/windowing/X11/WinSystemX11GLContext.cpp

Kodi has an advancedsetting option "forcevaapienabled" in 17.x. It got disabled in kodi 18-git because of code-move.

Reenable the option for people that "know what they are doing"

This code is no bugfix nor really a new feature.
It is for people (like me) that saw options in the documentation and don't understand why nothing shows in the debug log files. Once enabled, they can check how things work with their hardware.

### DIFF
--- a/xbmc/windowing/X11/WinSystemX11GLContext.cpp
+++ b/xbmc/windowing/X11/WinSystemX11GLContext.cpp
@@ -32,6 +32,7 @@
 #include <vector>
 #include "Application.h"
 #include "VideoSyncDRM.h"
+#include "settings/AdvancedSettings.h"
 
 #ifdef HAS_GLX
 #include "VideoSyncGLX.h"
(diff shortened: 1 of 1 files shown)


Rechi:
this needs some log warning

the option was originally added at https://github.com/xbmc/xbmc/commit/bfd9b82db1 and also had a log warning

FernetMenta:
no advancedsettings hacks

Razzeee:
Hey @lethalwp thanks bringing this to our attention, so should we update the documentation? I think you're referring to the wiki in this case?

lethalwp:
@Razzeee 
You can find the documentation at http://kodi.wiki/view/advancedsettings.xml

<forcevaapienabled>true</forcevaapienabled> | VAAPI is only tested on intel hardware. This flag helps wrapper developers and fglrx users to easily test current VAAPI state on their hardware. Default is false. Please report crashes directly to your driver vendor.
---

Maybe the correct fix would be to remove the "intel gpu" check. Since Amd should have working vaapi in a couple of months (4?)

FernetMenta:
I updated the wiki.

> Maybe the correct fix would be to remove the "intel gpu" check

if the vaapi check fails on intel, EGL is still to be used. on AMD it should switch to GLX

PR #12955, opened 2017-10-25
[Estuary] Change default button text & icon for tvshows in videoinfo …

jeroenpardon:
…dialog

## Description
This changes the button text and associated icon of the default focused button in the video info dialog when a tv show or season is selected.

## Motivation and Context
When opening the video info dialog for tv shows or seasons the button currently says "Play" while the action performed by pressing the button is browsing the tv show or season at hand, nothing is being played.

## How Has This Been Tested?
Tested on custom estuary install

## Types of change

## Checklist:

### DIFF
--- a/addons/skin.estuary/xml/DialogVideoInfo.xml
+++ b/addons/skin.estuary/xml/DialogVideoInfo.xml
@@ -382,10 +382,12 @@
 					<align>center</align>
 					<orientation>horizontal</orientation>
 					<scrolltime tween="quadratic">200</scrolltime>
-					<include content="InfoDialogButton">
+					<include content="InfoDialogToggleButton">
 						<param name="id" value="8" />
-						<param name="icon" value="icons/infodialogs/play.png" />
-						<param name="label" value="$LOCALIZE[208]" />
+						<param name="icon_on" value="icons/filemanager.png" />
+						<param name="icon_off" value="icons/infodialogs/play.png" />
+						<param name="selected" value="String.IsEqual(ListItem.DBType,tvshow) | String.IsEqual(ListItem.DBType,season)" />
+						<param name="label" value="$VAR[VideoInfoPlayButtonLabelVar]" />
 					</include>
 					<include content="InfoDialogButton">
 						<param name="id" value="11" />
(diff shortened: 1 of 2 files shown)


Rechi:
@jeroenpardon this needs a rebase

Rechi:
@phil65 @ronie ping

addons/skin.estuary/xml/DialogVideoInfo.xml
```diff
@@ -382,10 +382,16 @@
 					<align>center</align>
 					<orientation>horizontal</orientation>
 					<scrolltime tween="quadratic">200</scrolltime>
-					<include content="InfoDialogButton">
+					<include content="InfoDialogButton" condition="String.IsEqual(ListItem.DBType,tvshow) | String.IsEqual(ListItem.DBType,season)">
+						<param name="id" value="8" />
+						<param name="icon" value="icons/filemanager.png" />
+						<param name="label" value="$VAR[VideoInfoPlayLabelVar]" />
```
phil65: Why using a $VAR[] here when you already check the condition for the include?

jeroenpardon: The intention was to set both the button's label and the button's icon using var's so that only one control / include was needed. Unfortunately the icon path could not be set through a var in a param no matter what I tried. I believe I came across something about limitations of var's in a param on the forum at some point, but couldn't find it. 
Ideally, that shortcoming would be solved and the include can be dropped here.

phil65: I am not sure if you understood correctly :)
The first button you created is only included for tvshows and seasons. Why do you also check that again in a new VAR instead of directly assigning label 1024 to that button?

jeroenpardon: I understood. Ideally though only one control include would be needed. But as this currently doesn't seem possible (see my previous comment), I've removed the var.

phil65: i think it is possible by using the include InfoDialogToggleButton. Did you try that?

MartijnKaijser:
@phil65 @ronie your button if ok

ronie:
up to @phil65, not sure if he wants to pursue the togglebutton alternative

Rechi:
@phil65 @ronie ping

phil65:
@jeroenpardon this is what i meant: https://github.com/phil65/skin.estuary/commit/5016b670d0ca3a8547b7f67699093f314f9ddcc0

HitcherUK:
What's the deal with this? Needs adding imo.

ronie:
your wish is my command ;-)

HitcherUK:
Thanks. I would have done it myself but wasn't sure if we were waiting on Phil's approval.

PR #12959, opened 2017-10-25
[PVR] Fixes/Improvements for multiple enable PVR client addons

ksooo:
Fixes:
- do not delete data from not (yet) ready clients from database. Led to effects like watched state of channels lost, channel sort order lost, etc pp
- fix visibility of context menu items "add timer" and "record" in context of multiple enabled clients.
 
Improvements:

- Introduce priority for pvr clients. makes it possible to sort channels client-by-client (something that was impossible before or achieved via hacks like IPTV simple client's 'start channel numbers at' feature).
screenshot004
screenshot005
- show settings depending on number of enabled clients only if they make sense (i.e. "use backend channel numbers").

Cosmetics:

- some code cleanup along the way

All changes have been runtime tested carefully on macOS and Linux, latest Kodi master

@Jalle19 for review?


ksooo:
@Jalle19 would be cool if you could find some time to review this, thx.

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -18371,7 +18379,10 @@ msgctxt "#36209"
 msgid "Delete the databases for channels and guide and reimport the data from the backend afterwards."
 msgstr ""
 
-#empty string with id 36210
+#: system/settings/settings.xml
+msgctxt "#36210"
+msgid "Open the client priorities dialog, which allows modifying the priority for the enabled PVR clients, for instance to order channels client by client, according to your personal preferences. Lower value means higher priority"
```
Jalle19: Could be `Opens the client priorities dialog which allows modifying the priority for enabled PVR clients according to your personal preferences, for instance to order channels client by client. Lower value means higher priority.`

xbmc/addons/PVRClient.h
```diff
@@ -847,6 +847,18 @@ namespace PVR
     void OnPowerSavingDeactivated();
 
     /*!
+     * @brief Get the priority of this client. Higher value means lower priority.
```
Jalle19: Personally I would have "higher value means higher priority", IMO it's much more logical

xbmc/pvr/PVRDatabase.cpp
```diff
@@ -148,6 +158,67 @@ void CPVRDatabase::UpdateTables(int iVersion)
...
+
+  return ExecuteQuery(strQuery);
+}
+
+bool CPVRDatabase::Delete(const CPVRClient &client)
+{
+  if (client.GetID() == PVR_INVALID_CLIENT_ID)
+    return false;
+
+  CLog::Log(LOGDEBUG, "PVR - %s - deleting client '%s' from the database", __FUNCTION__, client.ID().c_str());
```
Jalle19: Could be `%i` and `client.GetID()`, right?

xbmc/pvr/PVRGUIActions.cpp
```diff
@@ -1251,8 +1251,8 @@ namespace PVR
     if (!CServiceBroker::GetPVRManager().IsStarted() || IsRunningChannelScan())
       return false;
 
-    PVR_CLIENT scanClient;
-    std::vector<PVR_CLIENT> possibleScanClients = CServiceBroker::GetPVRManager().Clients()->GetClientsSupportingChannelScan();
+    CPVRClientPtr scanClient;
+    std::vector<CPVRClientPtr> possibleScanClients = CServiceBroker::GetPVRManager().Clients()->GetClientsSupportingChannelScan();
```
Jalle19: Does any addon actually support channel scanning?

xbmc/pvr/PVRGUIActions.cpp
```diff
@@ -1502,10 +1502,14 @@ namespace PVR
 
     if (!bResetEPGOnly)
     {
-      pvrDatabase->DeleteChannelGroups();
+      pvrDatabase->DeleteClients();
```
Jalle19: Is it correct to delete clients before channel groups?

xbmc/pvr/channels/PVRChannelGroup.cpp
```diff
@@ -644,6 +676,26 @@ bool CPVRChannelGroup::AddAndUpdateChannels(const CPVRChannelGroup &channels, bo
   return bReturn;
 }
 
+bool CPVRChannelGroup::IsMissingChannelsFromClient(int iClientId) const
+{
+  for (int iFailedClientID : m_failedClientsForChannels)
```
Jalle19: Could perhaps be shortened to something like `return std::find(bla.begin(), bla.end(), iClientId) != bla.end()`

xbmc/pvr/channels/PVRChannelGroup.cpp
```diff
@@ -644,6 +676,26 @@ bool CPVRChannelGroup::AddAndUpdateChannels(const CPVRChannelGroup &channels, bo
...
+  {
+    if (iFailedClientID == iClientId)
+      return true;
+  }
+  return false;
+}
+
+bool CPVRChannelGroup::IsMissingChannelGroupMembersFromClient(int iClientId) const
+{
+  for (auto iFailedClientID : m_failedClientsForChannelGroupMembers)
```
Jalle19: Could perhaps be shortened to something like return std::find(bla.begin(), bla.end(), iClientId) != bla.end()

xbmc/addons/PVRClient.h
ksooo: Good point. Will change this.

xbmc/pvr/PVRDatabase.cpp
ksooo: ID() returns something like "pvr.hts", compared to GetID() which returns a numeric value calculated from ID() (a hash value). I thought for the log entry the string is of much more value.

xbmc/pvr/PVRGUIActions.cpp
ksooo: No, not at all, it's just a few afaik. But this is not related to this PR. Or did I break something here?

xbmc/pvr/PVRGUIActions.cpp
ksooo: Well, technically it's okay but you might be right, because it feels wrong from the logical point of view. Will change the order. ;-)

xbmc/pvr/channels/PVRChannelGroup.cpp
ksooo: yeah

xbmc/pvr/channels/PVRChannelGroup.cpp
ksooo: yeah yeah yeah...

addons/resource.language.en_gb/resources/strings.po
ksooo: 👍

ksooo: "Open" is okay, btw. All of those help texts are using direct speech.

ksooo:
@Jalle19 I changed most the stuff you requested.  Anything else?

xbmc/pvr/PVRDatabase.cpp
Jalle19: Ah, thought they were the same

xbmc/pvr/PVRGUIActions.cpp
Jalle19: No, I just thought of it when I saw it here

PR #12961, opened 2017-10-26
Update customcontroller.Harmony.xml

glorious1:
Corrected typos and removed some redundant assignments

## Description
Corrected typos and removed redundancies as follows (original line numbers)
42             <button id="102">Down</button>
Although it's just a comment, 'minus' should probably be 'down'.

47            <button id="107">Fullscreen</button>
I don't know if action names are case sensitive, but elsewhere it is 'FullScreen'.

108            <button id="185">xbmc.ActivateWindow(SystemSettings)</button>
Should 'xbmc.' be dropped?

114            <button id="16">ActivateWindow(PlayerControls)</button>
Button id should be 106

There are quite a few button assignments that are the same between Global and other windows.  If I understand correctly, they don't need to be specified in other windows if they are Global.  Not an error really, but ....  Some example lines are 157, 196, 231, 304-314.

In FullScreenVideo, lines 158-159, those actions are already available in Global, but through different buttons (F10 and F11).

The section ScreenCalibration, lines 235-242, is duplicated in lines 243-249, except that one button is missing in the copy (button id 165).

## Types of change

## Checklist:

### DIFF
--- a/system/keymaps/customcontroller.Harmony.xml
+++ b/system/keymaps/customcontroller.Harmony.xml
@@ -39,12 +39,12 @@
   <global>
     <customcontroller name="Harmony">
       <!-- up       	-->      <button id="101">Up</button>
-      <!-- minus      	-->      <button id="102">Down</button>
+      <!-- down      	-->      <button id="102">Down</button>
       <!-- left       	-->      <button id="103">Left</button>
       <!-- right      	-->      <button id="104">Right</button>
       <!-- OK		-->      <button id="105">Select</button>
       <!-- menu       	-->      <button id="106">ContextMenu</button>
-      <!-- hold play  	-->      <button id="107">Fullscreen</button>
+      <!-- hold play  	-->      <button id="107">FullScreen</button>
       <!-- hold menu  	-->      <button id="108">ContextMenu</button>
       <!-- hold right 	-->      <button id="109">Right</button>
       <!-- hold left  	-->      <button id="110">Left</button>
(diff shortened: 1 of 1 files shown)


system/keymaps/customcontroller.Harmony.xml
```diff
@@ -154,9 +154,6 @@
       <!-- right      	-->      <button id="104">StepForward</button>
       <!-- menu       	-->      <button id="106">OSD</button>
       <!-- Prev		-->      <button id="132">SmallStepBack</button>
-      <!-- Info  	-->      <button id="131">Info</button>
-      <!-- F7		-->      <button id="195">NextSubtitle</button>
```
Memphiz: In global section the ids 195 and 194 are mapped differently. You can't remove them here bevause you'd loose the subtitle features that are mapped here

system/keymaps/customcontroller.Harmony.xml
```diff
@@ -240,13 +235,6 @@
       <!-- Guide  	-->      <button id="165">NextResolution</button>
     </customcontroller>
   </ScreenCalibration>
-  <ScreenCalibration>
-    <customcontroller name="Harmony">
-      <!-- OK		-->      <button id="105">NextCalibration</button>
```
Memphiz: Same here - the global mapping of those buttons is not usefull for the screen calobration. You can't delete them here.

Memphiz (changes requested):
Beside my comments it is fine after requested changes were made

Memphiz:
Thx for digging into github for this. I know its a scary place at the beginning but this is i nice way to get your contribution into Kodi :)

system/keymaps/customcontroller.Harmony.xml
glorious1: I don't get why there needs to be two ScreenCalibration sections, one after the other, mapping almost the same.  But again, there's a lot I don't understand.

system/keymaps/customcontroller.Harmony.xml
glorious1: But in global section the subtitle features are mapped to 174-175, which should be still available in the FullScreenVideo section.  I guess I don't understand how it works.

Memphiz:
Hold on - about to board a plane - will have a closer look tomorrow - maybe i misread ...

system/keymaps/customcontroller.Harmony.xml
Memphiz: I can't tell which buttons 194 and 195 are refering to. I just think that they were mapped for a reason in the first place.

system/keymaps/customcontroller.Harmony.xml
Memphiz: you are right with this change - i didn't see the second section in the diff ...

Memphiz (approved):

Memphiz:
fine as is now that i looked into the original file. Thx for the pull request.

PR #12962, opened 2017-10-26
FIX: Only remove options for actual URL in URIUtils::Split (fixes #17627)

koying:
@mkortstiege @Razzeee 

Quick fix for the issue we discussed about.
Untested, so if you could make it tested, somehow...

### DIFF
--- a/xbmc/utils/URIUtils.cpp
+++ b/xbmc/utils/URIUtils.cpp
@@ -218,16 +218,19 @@ void URIUtils::Split(const std::string& strFileNameAndPath,
   // everything to the right of the directory separator
   strFileName = strFileNameAndPath.substr(i+1);
 
-  // ignore options
-  i = strFileName.size() - 1;
-  while (i > 0)
+  // if actual uri, ignore options
+  if (IsURL(strFileNameAndPath))
   {
-    char ch = strFileName[i];
-    if (ch == '?') break;
-    else i--;
+    i = strFileName.size() - 1;
+    while (i > 0)
+    {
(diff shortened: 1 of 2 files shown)


Rechi (dismissed):
`TestURIUtils.Split` fails on all platforms

mkortstiege (approved):
Actual change looks good to me.

Razzeee:
@koying can you try https://github.com/Razzeee/xbmc/commit/940379937f93894a0bc9a5f45475bce5c9591b87

koying:
@Rechi https://github.com/xbmc/xbmc/blob/8ddffac5c558ef679c191ebc0207e84397a7ace9/xbmc/utils/test/TestURIUtils.cpp#L127

The test case is basically what this PR is trying to "solve".
AFAIK, plain filenames are *not* URI, so something like "/path/to/movie.avi?showinfo=true" is actually invalid.

I'd remove the testcase, or replace by "file:///path/to/movie.avi?showinfo=true", but your thoughts are welcome

koying:
Added an additional testcase for the specific one we're trying to solve

razzeee (approved):

koying:
jenkins build this please

koying:
@Rechi Test cases ok. Good for you?

Rechi:
wrong revision was build
jenkins build this please

MartijnKaijser:
@koying  you want this backported or risk too high?

koying:
Actually, the code introducing the issue is not in K, so irrelevant

PR #12963, opened 2017-10-26
[JSON RPC][Music]Extra artist data with JSON API update

DaveTBlake:
Handling for some additional artist data that Musicbrainz provides:  type (person, group, orchestra, choir etc.),  gender, and disambiguation comments that help to differentiate artists with the same name. The music db version is bumped to 68.

Info label,  media filter and smartplaylist support is added, and the data is included in import from and export it to xml and NFO files. This is ahead of a scraper update, so until that happends the initial source of this data for testing is from NFO files, or import of a suitably edited xml file.

Updates the JSON API to include these, and the artist sort name (#11738) and album release groupd id data (#12120) that was added to the music DB in previous PRs. 

@Montellese if you could check the JSON part - addition of properties at the end to be non-breaking changes, with minor version bump. That can be split into a separate PR if necessary, however including it ensures it does not get forgotten.


xbmc/interfaces/json-rpc/schema/methods.json
```diff
@@ -974,7 +974,12 @@
       { "name": "genre", "type": [ "null", { "$ref": "Array.String", "required": true } ], "default": null },
       { "name": "died", "$ref": "Optional.String" },
       { "name": "disbanded", "$ref": "Optional.String" },
-      { "name": "yearsactive", "type": [ "null", { "$ref": "Array.String", "required": true } ], "default": null }
+      { "name": "yearsactive", "type": [ "null", { "$ref": "Array.String", "required": true } ], "default": null },
+      { "name": "musicbrainzartistid", "$ref": "Optional.String" },
+      { "name": "sortname", "$ref": "Optional.String" },
+      { "name": "type", "$ref": "Optional.String" },
+      { "name": "gender", "$ref": "Optional.String" },
```
Montellese: Are `type` and `gender` freely choosable or are there specifically supported values? In case of the latter you could turn these into enums (same below).

xbmc/music/MusicDatabase.cpp
```diff
@@ -124,6 +124,7 @@ void CMusicDatabase::CreateTables()
   m_pDS->exec("CREATE TABLE artist ( idArtist integer primary key, "
               " strArtist varchar(256), strMusicBrainzArtistID text, "
               " strSortName text, "
+              " strType text, strGender text, strDisambiguation text, "
```
Montellese: Shouldn't these be added at the very end of the list of columns? Because otherwise you will end up with a different column layout for newly created tables vs. updated tables which will screw up positional indexing on SELECT queries.

xbmc/music/tags/MusicInfoTag.cpp
```diff
@@ -813,10 +813,10 @@ void CMusicInfoTag::Serialize(CVariant& value) const
 
   value["displayartist"] = GetArtistString();
   value["displayalbumartist"] = GetAlbumArtistString();
-  value["artistsort"] = GetArtistSort();
+  value["sortartist"] = GetArtistSort();
   value["album"] = m_strAlbum;
   value["albumartist"] = m_albumArtist;
-  value["albumartistsort"] = m_strAlbumArtistSort;
+  value["sortalbumartist"] = m_strAlbumArtistSort;
```
Montellese: These two changes look backwards incompatible. Or have these properties never been exposed to JSON-RPC?

xbmc/utils/DatabaseUtils.cpp
```diff
@@ -109,6 +109,9 @@ std::string DatabaseUtils::GetField(Field field, const MediaType &mediaType, Dat
     if (field == FieldId) return "artistview.idArtist";
     else if (field == FieldArtistSort) return "artistview.strSortName";
     else if (field == FieldArtist) return "artistview.strArtist";
+    else if (field == FieldArtistType) return "artistview.strType";
+    else if (field == FieldGender) return "artistview.strGender";
+    else if (field == FieldDisambiguation) return "artistview.Disambiguation";
```
Montellese: `artistview.strDisambiguation`

Montellese (changes requested):
Looks like there are a few oopsies in there. But the changes to JSON-RPC in general are ok.

xbmc/interfaces/json-rpc/schema/methods.json
DaveTBlake: Like much of the music data `type` and `gender` are not codified and could conian any string, hence no enums.

xbmc/music/MusicDatabase.cpp
DaveTBlake: There is no positinal indexing or access used on the music db because of such a vulnerability. Changing column layout is not an issue for the db in the way it is fro JSON itself

xbmc/music/tags/MusicInfoTag.cpp
DaveTBlake: Although I tried to anticipate JSON use in the original PR that added this field, it has never been exposed. Hance switching to more consistent naming

xbmc/utils/DatabaseUtils.cpp
DaveTBlake: Opps, somehow lost in rebase. Thanks

DaveTBlake:
Thanks for making time to check this @Montellese , hope that you feel can hit approve/dismiss changes now.

DaveTBlake:
@Montellese I have adjusted Audio.Details.Media having realised that there are some historic errors with the type that I can not correct without removing fields (thus a breaking change). Even though can fixing something already broken be a breaking change??  I have added descriptions instead.

In **Audio.Details.Media**.
a) "musicbrainzalbumid", like other album data, is only returned by `GetAlbums()` and `GetAlbumDetails()`. It is always black for `GetSongs()` and `GetSongDetails()`. I think it should be a field in  Audio.Details.Albums instead,  and not enumerated in Audio.Fields.Song
b)"musicbrainzalbumartistid" is returned by `GetAlbums()`, `GetAlbumDetails()`, `GetSongs()` and `GetSongDetails()` as a **string array** not a single string as defined in the type. The array is correct as it can have multiple values - albums can have more than one album artist.

I am happy to make type changes instead, so please advise.

DaveTBlake:
@notspiff an example of adding new (non-tag based) music data fields, maybe of interest given discussions of CFileItem etc.

xbmc/music/tags/MusicInfoTag.cpp
Montellese: Personally I'd prefer if the database layout looks the same for a newly created database and an updated database even if there is no positional indexing right now.

xbmc/music/MusicDatabase.cpp
Montellese: Personally I'd prefer if the database layout looks the same for a newly created database and an updated database even if there is no positional indexing right now.

Montellese:
@DaveTBlake if the API definition does not match the implementation fixing it is not a breaking change because it wasn't working properly anyway.

IMO if there are fields currently defined in the API that are not returned keep them in but write a description that informs the user that it's not used / deprecated.

DaveTBlake:
@Montellese I have amended **Audio.Details.Media** as suggested - corrected the type for "musicbrainzalbumartistid" to match what is (correctly) returned. The deprecation of "musicbrainzalbumid" for GetSongs and GetSongDetails is in the type description.

DaveTBlake:
Rebased for version.txt format change (that happened without bump?)

xbmc/music/MusicDatabase.cpp
DaveTBlake: That has not happened before so already have differences. There are potentially more changes to come, so I could do an migration step that makes column order same for all.

Montellese: Right I haven't done much in the music database. Feel free to do whatever you prefer.

Montellese (approved):
Looks good for me.

PR #12964, opened 2017-10-27
VideoPlayer: add OnAVChange callback

FernetMenta:
This event can be used by clients interested in the state of a/v streams. On this event they can query CDataCacheCore for current state of streams.

@da-anda this can be used by i.e. StereoScopicManager


da-anda:
thanks @FernetMenta - trying to fix 3D stuff during devcon using this new event

PR #12966, opened 2017-10-27
Bump to v17.6

MartijnKaijser:


kekePower:
@MartijnKaijser Should configure.ac also be updated for 17.6?

PR #12967, opened 2017-10-27
[PVR] Fix epg and pvr database reset (was completely broken).

ksooo:
I guess I broke this some months ago during pvr gui code refactoring (not sure though).

Fix has been runtime-tested on macOS, latest Kodi master.

@Jalle19 might want to take a look at the code change.

### DIFF
--- a/xbmc/pvr/PVRGUIActions.cpp
+++ b/xbmc/pvr/PVRGUIActions.cpp
@@ -1463,8 +1463,6 @@ namespace PVR
 
     CDateTime::ResetTimezoneBias();
 
-    CServiceBroker::GetPVRManager().EpgContainer().Stop();
-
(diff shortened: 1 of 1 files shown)


xbmc/pvr/PVRGUIActions.cpp
```diff
@@ -1482,32 +1480,43 @@ namespace PVR
     pDlgProgress->SetPercentage(10);
     pDlgProgress->Progress();
 
-    /* reset the EPG pointers */
-    const CPVRDatabasePtr database(CServiceBroker::GetPVRManager().GetTVDatabase());
```
Jalle19: Is this the same database as for radio?

ksooo: Yep. There is only one for both tv and radio.

PR #12969, opened 2017-10-28
Add Xcode9 support for iOS 11 SDK and macOS 13 SDK

Memphiz:
This is a follow up to https://github.com/xbmc/xbmc/pull/12940 which also includes xcode9 support for compiling against iOS 11 SDK

--- original Text from lobermann----
Fix the build on MacOS when using the 10.13 SDK Version with XCode 9 but still using 10.12 MacOS

## Description
When updating XCode to the new version it also updated to the SDK Version 10.13, but the OS itself is still on 10.12 . After working on this issue for a day with @Rechi and @FernetMenta this is what I had to change to make the build work again.

What was fixed:

Error: `configure: error: error in configure of --with-sdk=10.13`
Fixed by adding SDK Version 10.13 Support

Error: libgcrypt: `The CPU Jitter random number generator must not be compiled with optimizations. See documentation. Use the compiler switch -O0 for compiling jitterentropy-base.c.`
Most likely caused by the recent update of libgcrypt. Fixed by patching the Makefile to also replace the -Og flag with -O0

Error:
````
dyld: lazy symbol binding failed: Symbol not found: _utimensat
  Referenced from: /Users/Shared/xbmc-depends/x86_64-darwin16.7.0-native/bin/cmake
  Expected in: /usr/lib/libSystem.B.dylib
````
Fixed build issue by updating CMake to 3.9.4, patching it to address the issues on macos and adding additional parameters for the native toolchain build.

I am not sure which of those changes we want to get into master right now. 
The SDK Part is no problem.
The libgcrypt should also not cause any issues on other systems, as the patch is only executed on MacOS. But I can only test it on my mac.
The CMake update worked by just increasing the version number and built without issues. So I assume it is OK for other systems also. I think Windows is also using those dependencies, or? Needs to be tested. Also, the CMake tar.gz needs to be put onto the mirror if merged.

## Types of change

## Checklist:

### DIFF
--- a/tools/depends/target/samba-gplv3/Makefile
+++ b/tools/depends/target/samba-gplv3/Makefile
@@ -1,5 +1,5 @@
 include ../../Makefile.include
-DEPS= ../../Makefile.include Makefile autotools41.patch configureEndian.patch samba_android.patch samba_off64_t.patch no_fork_and_exec.patch
+DEPS= ../../Makefile.include Makefile autotools41.patch configureEndian.patch samba_android.patch samba_off64_t.patch no_fork_and_exec.patch crt_extensions.patch
 
 # lib name, version
 LIBNAME=samba
@@ -57,8 +57,9 @@ ifeq ($(OS),android)
 	cd $(PLATFORM); patch -p0 < ../samba_android.patch
 	cd $(PLATFORM); patch -p1 < ../samba_off64_t.patch
 endif
-ifeq ($(TARGET_PLATFORM),appletvos)
+ifeq ($(OS), ios)
 	cd $(PLATFORM); patch -p0 < ../no_fork_and_exec.patch
+	cd $(PLATFORM); patch -p0 < ../crt_extensions.patch
 endif
 ifeq (darwin, $(findstring darwin, $(HOST)))
 	sed -ie "s|ifndef HAVE_CLOCK_GETTIME|if !defined(HAVE_CLOCK_GETTIME) \&\& !defined(CLOCK_REALTIME)|" "$(PLATFORM)/lib/replace/system/time.h"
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

**Comment by Memphiz** [comment]
I will manually verify this PR on jenkins (which also runs 10.12 and xcode9) before merging it. @lobermann please try it out too if you can.

**Comment by lobermann** [comment]
@Memphiz tested and works. +1

**Comment by Memphiz** [comment]
thx - lets see if jenkins with older sdk and other platforms is still fine - jenkins build this please

**Comment by Memphiz** [comment]
once again - jenkins build this please

**Comment by Memphiz** [comment]
jenkins build this please for god sake

**Comment by Memphiz** [comment]
@MartijnKaijser problem is - jenkins still thinks that this PR has the no-jenkins tag attached ...

**Comment by Memphiz** [comment]
kicked a manually build: http://jenkins.kodi.tv/view/Helpers/job/BuildMulti-All/941/

**Comment by Memphiz** [comment]
everything built fine - win32 build error is unrelated (couldn't remove build dir). Added @Rechi as reviewer (as he was on the initial PR).

File: tools/depends/target/samba-gplv3/Makefile [hunk 1]
```diff
@@ -57,8 +57,9 @@ ifeq ($(OS),android)
 	cd $(PLATFORM); patch -p0 < ../samba_android.patch
 	cd $(PLATFORM); patch -p1 < ../samba_off64_t.patch
 endif
-ifeq ($(TARGET_PLATFORM),appletvos)
+ifeq ($(OS), ios)
```
Rechi: no_fork_and_exec patch is now applied for ios.
Why is this wanted?

Memphiz: Because iOS 11 SDK doesn't allow it anymore like tvOS did in the first place

Memphiz: Oh and tvOS is a subset of iOS - it's applied to both now

Rechi (approved):

PR #12970, opened 2017-10-28
[cmake][windows] avoid linkage against libkodi

AchimTuran:
My local Windows failed because Visual Studio tried to link against the target libkodi, which is not present for Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -310,7 +310,12 @@ else()
 endif()
 add_dependencies(${APP_NAME_LC} ${APP_NAME_LC}-libraries export-files pack-skins)
 whole_archive(_MAIN_LIBRARIES ${core_DEPENDS})
-target_link_libraries(${APP_NAME_LC} ${_MAIN_LIBRARIES} lib${APP_NAME_LC} ${DEPLIBS})
+
+if(NOT WIN32)
+  target_link_libraries(${APP_NAME_LC} ${_MAIN_LIBRARIES} lib${APP_NAME_LC} ${DEPLIBS})
+else()
+  target_link_libraries(${APP_NAME_LC} ${_MAIN_LIBRARIES} ${DEPLIBS})
+endif()
 unset(_MAIN_LIBRARIES)
 
 if(WIN32)


AchimTuran:
jenkins build this please

Paxxi:
@Rechi is the current champ of cmake on Windows so I think he should have final say.

I haven't seen this cause any issues so to me this feels like a solution to an old/dirty/broken workspace rather than something that should be changed in cmake.

AchimTuran:
I did a clean build and tried to rebuild. During linking a got the same error.

Paxxi:
try to wipe the workspace and do a fresh clone or `git clean -xfd && git clean -Xfd` and download deps and such again

notspiff:
Windows doesnt do the lib prefix. Just link to the target and not the accidentially working lib prefixed target name.

AchimTuran:
@paxxi same issue with a fresh cloned repository.

Rechi (changes requested):
I don't know what happened on your system, but that can't work.
The xbmc target you see in Visual Studio is internally called `lib${APP_NAME_LC}` and it has for sure to be linked against `${APP_NAME_LC}`.

AchimTuran:
Thank you guys for helping to repair my Windows build. The issue was that CMake took the wrong Windows SDK...
You can set the Toolset Version with the following cmake call:
`cmake -G"Visual Studio 14 Win64" -DCMAKE_SYSTEM_VERSION="<specific Windows SDK version>"`

AchimTuran:
Sorry the PR isn't correct and will never work, which is the reason I will close it now.

Montellese:
Having to specify the Windows SDK version manually sucks. On my Win7 machine it picks up 8.1 by default. Is there any way we can support both the 8.1 and the 10 version with some ifdefery?

Rechi:
run cmake with `-DCMAKE_SYSTEM_VERSION="10.0.xxxxx.0"` (`xxxxx` is the SDK 10 version you have installed)

Montellese:
@Rechi I realize that but it's cumbersome because I have to change it when(ever) I update the SDK

Rechi:
maybe it is possible to specify a more general version number (`10.0`), but I don't know

Paxxi:
Never had to specify an sdk version except for building uwp, cmake have automatically picked the newest available. Curious what's different on your systems.

AchimTuran:
Yes that's super strange. On my machine it always uses the oldest one...

PR #12973, opened 2017-10-30
changed: drop enumeration and hardcoding for supported protocols

notspiff:
instead use a descriptor struct for supported protocols

this is a preparatory step to allow reading this info for vfs add-ons

## Description
Drop enumeration and hardcoded if statements. instead encode properties of protocols in a descriptor struct. this will allow reading the same info from vfs add-ons so stuff only shows up if available.

i think this is better than having an installation query at this time since we cannot actually perform that installation.

## How Has This Been Tested?
It gives the same entries as before.

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -21292,3 +21292,8 @@ msgstr ""
 msgctxt "#39102"
 msgid "Play Radio"
 msgstr ""
+
+#: xbmc/network/GUIDialogNetworkSetup.cpp
+msgctxt "#39103"
+msgid "Unable to configure network location. Invalid path given."
+msgstr ""
--- a/xbmc/network/GUIDialogNetworkSetup.cpp
+++ b/xbmc/network/GUIDialogNetworkSetup.cpp
@@ -22,7 +22,13 @@
 
 #include <utility>
 
+#include "addons/AddonManager.h"
+#include "addons/VFSEntry.h"
+#include "addons/binary-addons/BinaryAddonBase.h"
+#include "ServiceBroker.h"
+
 #include "dialogs/GUIDialogFileBrowser.h"
+#include "messaging/helpers/DialogOKHelper.h"
 #include "guilib/GUIEditControl.h"
 #include "guilib/GUIWindowManager.h"
 #include "guilib/LocalizeStrings.h"
@@ -31,6 +37,11 @@
 #include "utils/log.h"
 #include "utils/StringUtils.h"
 #include "utils/URIUtils.h"
+
+
+using namespace ADDON;
+using namespace KODI::MESSAGING;
+
 
 #define CONTROL_OK              28
 #define CONTROL_CANCEL          29
@@ -46,7 +57,7 @@
 CGUIDialogNetworkSetup::CGUIDialogNetworkSetup(void)
     : CGUIDialogSettingsManualBase(WINDOW_DIALOG_NETWORK_SETUP, "DialogSettings.xml")
 {
-  m_protocol = NET_PROTOCOL_SMB;
+  m_protocol = 0;
   m_confirmed = false;
   m_loadType = KEEP_IN_MEMORY;
 }
@@ -131,7 +142,12 @@ bool CGUIDialogNetworkSetup::ShowAndGetNetworkAddress(std::string &path)
   CGUIDialogNetworkSetup *dialog = g_windowManager.GetWindow<CGUIDialogNetworkSetup>(WINDOW_DIALOG_NETWORK_SETUP);
   if (!dialog) return false;
   dialog->Initialize();
-  dialog->SetPath(path);
+  if (!dialog->SetPath(path))
+  {
+    HELPERS::ShowOKDialogText(CVariant{ 10218 }, CVariant{ 39103 });
+    return false;
+  }
+
   dialog->Open();
   path = dialog->ConstructPath();
   return dialog->IsConfirmed();
@@ -191,20 +207,23 @@ void CGUIDialogNetworkSetup::InitializeSettings()
   // Add our protocols
   TranslatableIntegerSettingOptions labels;
 #ifdef HAS_FILESYSTEM_SMB
-  labels.push_back(std::make_pair(20171, NET_PROTOCOL_SMB));
+  m_protocols.emplace_back(Protocol{true, true, true, false, true, 0, "smb", 20171});
 #endif
-  labels.push_back(std::make_pair(20301, NET_PROTOCOL_HTTPS));
-  labels.push_back(std::make_pair(20300, NET_PROTOCOL_HTTP));
-  labels.push_back(std::make_pair(20254, NET_PROTOCOL_DAVS));
-  labels.push_back(std::make_pair(20253, NET_PROTOCOL_DAV));
-  labels.push_back(std::make_pair(20173, NET_PROTOCOL_FTP));
-  labels.push_back(std::make_pair(20175, NET_PROTOCOL_UPNP));
-  labels.push_back(std::make_pair(20304, NET_PROTOCOL_RSS));
+  const std::vector<Protocol> defaults =
+        {{ true,  true,  true,  true, false, 443, "https", 20301},
+         { true,  true,  true,  true, false,  80,  "http", 20300},
+         { true,  true,  true,  true, false, 443,  "davs", 20254},
+         { true,  true,  true,  true, false,  80,   "dav", 20253},
+         { true,  true,  true,  true, false,  21,   "ftp", 20173},
+         {false, false, false, false,  true,   0,  "upnp", 20175},
+         { true,  true,  true,  true, false,  80,   "rss", 20304}};
+
+ m_protocols.insert(m_protocols.end(), defaults.begin(), defaults.end());
 #ifdef HAS_FILESYSTEM_NFS
-  labels.push_back(std::make_pair(20259, NET_PROTOCOL_NFS));
+  m_protocols.emplace_back(Protocol{true, false, false, false, true, 0, "nfs", 20259});
 #endif
 #ifdef HAS_FILESYSTEM_SFTP
-  labels.push_back(std::make_pair(20260, NET_PROTOCOL_SFTP));
+  m_protocols.emplace_back(Protocol{true, true, true, true, false, 22, "sftp", 20260});
 #endif
 
   AddSpinner(group, SETTING_PROTOCOL, 1008, SettingLevel::Basic, m_protocol, labels);
@@ -262,20 +281,9 @@ void CGUIDialogNetworkSetup::OnProtocolChange()
     CGUIMessage msg(GUI_MSG_ITEM_SELECTED, GetID(), settingControl->GetID());
     if (!OnMessage(msg))
       return;
-    m_protocol = (NET_PROTOCOL)msg.GetParam1();
+    m_protocol = msg.GetParam1();
     // set defaults for the port
-    if (m_protocol == NET_PROTOCOL_FTP)
-      m_port = "21";
-    else if (m_protocol == NET_PROTOCOL_HTTP || 
-       m_protocol == NET_PROTOCOL_RSS || 
-       m_protocol == NET_PROTOCOL_DAV)
-      m_port = "80";
-    else if (m_protocol == NET_PROTOCOL_HTTPS || m_protocol == NET_PROTOCOL_DAVS)
-      m_port = "443";
-    else if (m_protocol == NET_PROTOCOL_SFTP)
-      m_port = "22";
-    else
-      m_port = "0";
+    m_port = StringUtils::Format("%i", m_protocols[m_protocol].defaultPort);
 
     UpdateButtons();
   }
@@ -289,7 +297,7 @@ void CGUIDialogNetworkSetup::UpdateButtons()
   {
     int addressControlID = addressControl->GetID();
     SET_CONTROL_LABEL2(addressControlID, m_server);
-    if (m_protocol == NET_PROTOCOL_SMB)
+    if (m_protocols[m_protocol].type == "smb")
     {
       SET_CONTROL_LABEL(addressControlID, 1010);  // Server name
     }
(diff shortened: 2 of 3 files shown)


xbmc/network/GUIDialogNetworkSetup.cpp
```diff
@@ -191,21 +200,22 @@ void CGUIDialogNetworkSetup::InitializeSettings()
...
+         { true,  true,  true,  true, false,  80,   "dav", 20253},
+         { true,  true,  true,  true, false,  21,   "ftp", 20173},
+         {false, false, false, false,  true,   0,  "upnp", 20175},
+         { true,  true,  true,  true, false,  80,   "rss", 20304}};
+
+ m_protocols.insert(m_protocols.end(), defaults.begin(), defaults.end());
 #ifdef HAS_FILESYSTEM_NFS
-  labels.push_back(std::make_pair(20259, NET_PROTOCOL_NFS));
-#endif
-#ifdef HAS_FILESYSTEM_SFTP
```
stefansaraev: internal sftp is optional. was this ifdef removed by mistake?

notspiff: yes. bad backport (did it on top of sftp removal branch, and forgot to adjust it back).

notspiff: remedied.

xbmc/network/GUIDialogNetworkSetup.h
```diff
@@ -25,17 +25,19 @@
...
-                      NET_PROTOCOL_DAVS,
-                      NET_PROTOCOL_UPNP,
-                      NET_PROTOCOL_RSS,
-                      NET_PROTOCOL_SFTP,
-                      NET_PROTOCOL_NFS};
+  //! \brief A structure encapsulating properties of a supported protocol.
+  struct Protocol
+  {
+    bool paths;       //!< Protocol has path in addition to server name
+    bool usernames;   //!< Protocol uses logins
```
xhaggi: Should we use singular for `paths` and `usernames` like you did for `password`, `port` etc?
Not sure if it makes sense to be more clear in case of the property names e.g. prefix with `support` .. `supportPath`, `supportUsername`, `supportBrowsing` etc.

xbmc/network/GUIDialogNetworkSetup.cpp
```diff
@@ -191,20 +200,23 @@ void CGUIDialogNetworkSetup::InitializeSettings()
...
-  labels.push_back(std::make_pair(20173, NET_PROTOCOL_FTP));
-  labels.push_back(std::make_pair(20175, NET_PROTOCOL_UPNP));
-  labels.push_back(std::make_pair(20304, NET_PROTOCOL_RSS));
+  const std::vector<Protocol> defaults =
+        {{ true,  true,  true,  true, false, 443, "https", 20301},
+         { true,  true,  true,  true, false,  80,  "http", 20300},
+         { true,  true,  true,  true, false, 443,  "davs", 20254},
+         { true,  true,  true,  true, false,  80,   "dav", 20253},
+         { true,  true,  true,  true, false,  21,   "ftp", 20173},
+         {false, false, false, false,  true,   0,  "upnp", 20175},
```
xhaggi: missing space

notspiff: do not agree. trailing space before } either.

xbmc/network/GUIDialogNetworkSetup.cpp
```diff
@@ -191,20 +200,23 @@ void CGUIDialogNetworkSetup::InitializeSettings()
...
+         { true,  true,  true,  true, false,  80,   "rss", 20304}};
```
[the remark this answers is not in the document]
notspiff: i think it improves readability of these tables to have the commas aligned.

xbmc/network/GUIDialogNetworkSetup.cpp
```diff
@@ -371,90 +357,51 @@ void CGUIDialogNetworkSetup::UpdateButtons()
...
-    m_protocol = NET_PROTOCOL_UPNP;
-  else if (url.IsProtocol("rss"))
-    m_protocol = NET_PROTOCOL_RSS;
-  else if (url.IsProtocol("nfs"))
-    m_protocol = NET_PROTOCOL_NFS;
-  else if (url.IsProtocol("sftp") || url.IsProtocol("ssh"))
-    m_protocol = NET_PROTOCOL_SFTP;
-  else
-    m_protocol = NET_PROTOCOL_SMB;  // default to smb
+  m_protocol = 0;
```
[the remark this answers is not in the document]
notspiff: the old code blindly used smb no matter as default value, which would point to .. nothing if smb was not enabled. if you want to old behavior of the spinner showing label 0 = 'Programs' i can add it but that doesn't sound very productive ;)

xbmc/network/GUIDialogNetworkSetup.h
notspiff: done.

xbmc/network/GUIDialogNetworkSetup.h
```diff
@@ -25,17 +25,19 @@
...
+    bool password;    //!< Protocol supports passwords
+    bool port;        //!< Protocol support port customization
```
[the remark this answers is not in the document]
notspiff: done.

xbmc/network/GUIDialogNetworkSetup.h
```diff
@@ -25,17 +25,19 @@
...
+    bool browseable;  //!< Protocol support server browsing
```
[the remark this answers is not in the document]
notspiff: done.

xbmc/network/GUIDialogNetworkSetup.h
```diff
@@ -25,17 +25,19 @@
...
+    int defaultport;  //!< Default port to use for protocol
```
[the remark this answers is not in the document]
notspiff: done.

xbmc/network/GUIDialogNetworkSetup.cpp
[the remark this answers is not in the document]
xhaggi: okay, your choice, for me ltr is better to read

xbmc/network/GUIDialogNetworkSetup.cpp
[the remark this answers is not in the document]
xhaggi: Okay only want to know if that is desired.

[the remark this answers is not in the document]
notspiff: i don't know what would be a better approach - maintaining a bug certainly won't be. i can look for some protocol that is always there, but that would give perceived differences when smb is enabled.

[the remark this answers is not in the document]
xhaggi: Calling this method with a path that does not contain a valid protocol should throw an error or shouldn't set the path. Before and now it suppose the first is the right one, but that's not the case.

xbmc/network/GUIDialogNetworkSetup.cpp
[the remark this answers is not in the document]
Paxxi: We don't really have anything about this in the coding guidelines so it's up to personal taste or match previous code if in same file

xbmc/network/GUIDialogNetworkSetup.h
```diff
@@ -25,17 +25,19 @@
...
+  struct Protocol
+  {
+    bool supportPath;      //!< Protocol has path in addition to server name
+    bool supportUsername;  //!< Protocol uses logins
+    bool supportPassword;  //!< Protocol supports passwords
+    bool supportPort;      //!< Protocol supports port customization
+    bool supportBrowsing;  //!< Protocol supports server browsing
+    int defaultPort;       //!< Default port to use for protocol
+    std::string type;      //!< URL type for protocol
+    int label;             //!< String ID to use as label in dialog
```
Paxxi: This could be reordered to avoid some padding. Not really something we've cared about but it's been in the back of my mind recently to look into

notspiff: Why would we care about padding for this struct?

xbmc/network/GUIDialogNetworkSetup.cpp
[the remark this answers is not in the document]
notspiff: no path set, made function bool and return to caller. but there is no sane way to do exceptions here and it would be rather alien to other code in kodi.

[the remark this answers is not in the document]
xhaggi: okay cool thanks. Instead of just log an error should we inform the user about the issue (ok dialog + user-friendly message)?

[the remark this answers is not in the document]
notspiff: not sure how friendly i want to be to peeps who have messed up the xml by hand ;)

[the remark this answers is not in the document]
xhaggi: okay leave it as it is, it's definitely better as before. IMO we can merge if all is fine.

xhaggi (approved):

xbmc/network/GUIDialogNetworkSetup.cpp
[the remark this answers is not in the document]
notspiff: dialog added. sorry whomever have string additions in the queue.

xbmc/network/GUIDialogNetworkSetup.cpp
```diff
@@ -131,7 +142,8 @@ bool CGUIDialogNetworkSetup::ShowAndGetNetworkAddress(std::string &path)
   CGUIDialogNetworkSetup *dialog = g_windowManager.GetWindow<CGUIDialogNetworkSetup>(WINDOW_DIALOG_NETWORK_SETUP);
   if (!dialog) return false;
   dialog->Initialize();
-  dialog->SetPath(path);
+  if (!dialog->SetPath(path))
+    HELPERS::ShowOKDialogText(CVariant{ 10218 }, CVariant{ 39103 });
   dialog->Open();
```
xhaggi: should we prevent opening the network dialog?

notspiff: yeh that's better. changed.

xbmc/network/GUIDialogNetworkSetup.cpp
```diff
@@ -131,7 +142,12 @@ bool CGUIDialogNetworkSetup::ShowAndGetNetworkAddress(std::string &path)
   CGUIDialogNetworkSetup *dialog = g_windowManager.GetWindow<CGUIDialogNetworkSetup>(WINDOW_DIALOG_NETWORK_SETUP);
   if (!dialog) return false;
   dialog->Initialize();
-  dialog->SetPath(path);
+  if (!dialog->SetPath(path))
```
xhaggi: meh, there is a call to `ShowAndGetNetworkAddress` in `CGUIDialogFileBrowser::OnAddNetworkLocation` where an empty path is given to this method. This should open a blank dialog to add a new network location https://github.com/xbmc/xbmc/blob/master/xbmc/dialogs/GUIDialogFileBrowser.cpp#L897

We should return true in `SetPath` if the given path is empty (early return).

addons/resource.language.en_gb/resources/strings.po
```diff
@@ -21292,3 +21292,8 @@ msgstr ""
 msgctxt "#39102"
 msgid "Play Radio"
 msgstr ""
+
+#: xbmc/network/GUIDialogNetworkSetup.cpp
+msgctxt "#39103"
+msgid "Invalid path given. Edit cancelled."
+msgstr ""
```
xhaggi: `Unable to configure network location. Invalid path given.`

xbmc/network/GUIDialogNetworkSetup.cpp
notspiff: i'm sure the call for empty was to wipe out previous data, so i did it a bit differently.

addons/resource.language.en_gb/resources/strings.po
notspiff: changed.

xbmc/network/GUIDialogNetworkSetup.cpp
```diff
@@ -371,94 +366,69 @@ void CGUIDialogNetworkSetup::UpdateButtons()
...
+  if (m_protocol == -1)
+  {
+    if (!path.empty())
+      CLog::Log(LOGERROR, "__PRETTY_FUNCTION__: Asked to initialize for unknown path %s", path.c_str());
+    m_username.clear();
+    m_password.clear();
+    m_port.clear();
+    m_server.clear();
+    m_path.clear();
+    m_protocol = 0;
```
xhaggi: we could add a new method `Reset()` like we did in other dialogs IIRC and call it here and in case the path is empty. This way we could call it earlier and do not need to loop over protocols etc.

notspiff: the nanosecs that takes we will lose a million time over having this discussion ;)

xhaggi: sure sure, but it's handy if we need to somewhere else. You could leave it as it is if it's worth the time

notspiff: done.

xhaggi: Thx  you are fast ;)

xbmc/network/GUIDialogNetworkSetup.cpp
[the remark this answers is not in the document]
Montellese: I've moved all my strings in my media importing work to 40000+ ;-)

xhaggi:
Any objections? If not merge it in @notspiff

Rechi:
wrong revision was build
jenkins build this please

xbmc/network/GUIDialogNetworkSetup.cpp
```diff
@@ -191,20 +207,23 @@ void CGUIDialogNetworkSetup::InitializeSettings()
...
 #ifdef HAS_FILESYSTEM_NFS
-  labels.push_back(std::make_pair(20259, NET_PROTOCOL_NFS));
+  m_protocols.emplace_back(Protocol{true, false, false, false, true, 0, "nfs", 20259});
 #endif
 #ifdef HAS_FILESYSTEM_SFTP
-  labels.push_back(std::make_pair(20260, NET_PROTOCOL_SFTP));
+  m_protocols.emplace_back(Protocol{true, true, true, true, false, 22, "sftp", 20260});
 #endif
 
   AddSpinner(group, SETTING_PROTOCOL, 1008, SettingLevel::Basic, m_protocol, labels);
```
Rechi: @notspiff `labels` now doesn't contain any elements and therefore the Protocol spinner doesn't show up

notspiff:
Sorry must have fallen out in one of the many rewrites. Will fix asap

7. nov. 2017 18:49 skrev "Rechi" <notifications@github.com>:

PR #12974, opened 2017-10-30
FIX: [droid] pick toolchain gcc at depends configure time

koying:
There is a bug since forever that the depends configure uses the host gcc rather than the target one:
```
onfigure:2158: checking for arm-linux-androideabi-g++
configure:2188: result: no
configure:2158: checking for arm-linux-androideabi-c++
configure:2188: result: no
configure:2158: checking for arm-linux-androideabi-gpp
configure:2188: result: no
configure:2158: checking for arm-linux-androideabi-aCC
configure:2188: result: no
configure:2158: checking for arm-linux-androideabi-CC
configure:2188: result: no
configure:2158: checking for arm-linux-androideabi-cxx
configure:2188: result: no
configure:2158: checking for arm-linux-androideabi-cc++
configure:2188: result: no
configure:2158: checking for arm-linux-androideabi-cl.exe
configure:2188: result: no
configure:2158: checking for arm-linux-androideabi-FCC
configure:2188: result: no
configure:2158: checking for arm-linux-androideabi-KCC
configure:2188: result: no
configure:2158: checking for arm-linux-androideabi-RCC
configure:2188: result: no
configure:2158: checking for arm-linux-androideabi-xlC_r
configure:2188: result: no
configure:2158: checking for arm-linux-androideabi-xlC
configure:2188: result: no
configure:2202: checking for g++
configure:2218: found /usr/bin/g++
configure:2229: result: g++
configure:2256: checking for C++ compiler version
configure:2265: g++ --version >&5
g++ (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609
```

It now bites because gcc6(?) has c++11 builtin, so the proper "--std=" is not added when building on such systems

gcc5:
```
configure:2659: checking whether g++ supports C++11 features by default
[...]
configure:2700: result: no
configure:2709: checking whether g++ supports C++11 features with -std=gnu++11
configure:2745: g++ -c -g -O2 -std=gnu++11  conftest.cpp >&5
configure:2745: $? = 0
configure:2754: result: yes
```

I don't have gcc6, but garbear has, and the result is "yes", so "std=gnu++11" is not added and builds fail with the 4.9 of the toolchain

Thanks to @wsnipex for the final solution

### DIFF
--- a/tools/depends/configure.ac
+++ b/tools/depends/configure.ac
@@ -14,9 +14,6 @@ passed_cxxflags="$CXXFLAGS"
 
 m4_include([m4/xbmc_arch.m4])
 m4_include([m4/ax_cxx_compile_stdcxx_11.m4])
-AX_CXX_COMPILE_STDCXX_11(,[mandatory])
-c11_flags=$(echo "$CFLAGS" | sed 's/-O@<:@123@:>@//g;s/-g //g;s/ //g')
-cxx11_flags=$(echo "$CXXFLAGS" | sed 's/-O@<:@123@:>@//g;s/-g //g;s/ //g')
 
 # check for not same cpu value
 AC_DEFUN([MC_CHECK_NOT_CPU],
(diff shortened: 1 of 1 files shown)


Rechi:
adding the toolchain path to the PATH env variable should be done in configure.ac

koying:
Suggestions (and code) welcome, i'm not an autotools guru ;)

wsnipex:
please check why https://github.com/xbmc/xbmc/blob/master/tools/depends/configure.ac#L480-L489
doesn't work. And we use toolchain not the ndk-path directly

koying:
@wsnipex Right about toolchain, ofc. Blind code

Do you know which macro hosts the `configure:2158: checking for arm-linux-androideabi-*` ?
Most certainly doesn't use the variables you reference, as it happens before

wsnipex:
I guess m4/ax_cxx_compile_stdcxx_11.m4 might be the culprit here

Rechi:
CXX isn't set to the correct compiler in m4/ax_cxx_compile_stdcxx_11.m4

koying:
ax_cxx_compile_stdcxx_11.m4 uses the detected compiler to test for c++11, but I don't think it's the one doing the detection, is it?

wsnipex:
yes it is, if CXX is not set

wsnipex (approved):

Rechi (approved):

Rechi:
wrong revision was build
jenkins build this please

Rechi:
WIN-32 error unrelated as this only touches posix depends.

PR #12983, opened 2017-10-31
Some CFileItem cleaning

notspiff:
I was asked to assist with CFileItem restructuring on slack. Everything from performance to desktop search to inheritance issues was mentioned, but nobody could really tell me what they wanted done.
Until a clear idea is formulated for what this restructuring entails, I chose to at least reduce the amount of code that has to be restructured. By making CFileItem::IsType() being able to check for either extensions or protocols, we can get rid of a lot of trivial member functions.

I expect some breakage in platform code, I'm sure jenkins will yell.

### DIFF
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -2973,7 +2973,7 @@ bool CApplication::PlayMedia(CFileItem& item, const std::string &player, int iPl
       }
     }
   }
-  else if (item.IsPVR())
+  else if (item.IsType("pvr://"))
   {
     return CServiceBroker::GetPVRManager().GUIActions()->PlayMedia(CFileItemPtr(new CFileItem(item)));
   }
--- a/xbmc/Application.cpp
+++ b/xbmc/Application.cpp
@@ -3275,7 +3275,7 @@ void CApplication::PlaybackCleanup()
   }
 
   // DVD ejected while playing in vis ?
-  if (!m_appPlayer.IsPlayingAudio() && (m_itemCurrentFile->IsCDDA() || m_itemCurrentFile->IsOnDVD()) && !g_mediaManager.IsDiscInDrive() && CServiceBroker::GetGUI()->GetWindowManager().GetActiveWindow() == WINDOW_VISUALISATION)
+  if (!m_appPlayer.IsPlayingAudio() && (m_itemCurrentFile->IsType("cdda://") || m_itemCurrentFile->IsOnDVD()) && !g_mediaManager.IsDiscInDrive() && CServiceBroker::GetGUI()->GetWindowManager().GetActiveWindow() == WINDOW_VISUALISATION)
   {
     // yes, disable vis
     m_ServiceManager->GetSettings().Save();    // save vis settings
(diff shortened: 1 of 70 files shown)


fritsch:
Thanks very much. In deed the first step is to unclutter it, which makes upcoming restructuring way more easy. It does not improve anything architecture wise, but that we also discussed before hand.

@Paxxi this was the first step we discussed yesterday, right?

hudokkow:
I thought you were busy this week?... ;-)

notspiff:
you never met a procrastinator before? ;) this was just left-handed scripted stuff, little brain involved. aka sed'n'pray.

DaveTBlake:
Can someone explain why having many trivial member functions is a problem in itself? 

I think CFileItem seems a poor OO design, but the `IsXXXX()` member functions just show how many different kinds of thing are stuffed into this generic bucket.  In fact it codifies what those things can be in a clear way.

Why is replacing them with single function with a string parameter better? 
I could have `IsType("blah")`, and it would compile even though "blah" is meaningless. More realistically I could have a typo like `IsType("muiscdb:/") and it would work always returning false. In comparison `IsMusicDb()` is either a known function, or won't compile.

I just don't see how this is a step forwards. Why not leave the long list of trivial but codifying member functions, that show clearly the many fucntions this item performs,  until we have a plan of how to replace the generic bucket with at least an OO bucket.

notspiff:
If you are asking me: this is indeed a step backwards. exactly the point i wanted somebody to pick up on. You can have these small members which some find messy, you can try to build some hierarchy or you use strings and shift to runtime risks. A hierarchy will mostly yield lots of decoration code and the same mess, instead of IsXXX members you now cast like it is no tomorrow.

DaveTBlake:
> If you are asking me: this is indeed a step backwards

Yes, I feel is just hides the issue. I don't see how reducing the code in this way really helps at all.  Instead I think we need some fundamental design discussion first. Actually formulate a clear idea for what  restructuring we need.

fritsch:
This was discussed on IRC right before the work started. The idea was to do that step so that later refactoring / shuffeling can be done easier. Also this change here directly shows who calls those methods to help identifying the future architecture we want to go to.

I'd say let's keep it open as is - as it helps tracking its usage.

fritsch:
@notspiff  fully agree and already said that yesterday. The architecture flaw that produced a CFileItem starts in other layers.

Paxxi:
Removing the methods is mainly to reduce cognitive overhead my minimizing the amount of work required, not necessarily a good step in itself but that's often the case when trying to refactor incrementally without breaking.

I was toying with an idea of how to make it uniform from a ui perspective. We currently use one or two columns of info. 
We could basically provide an interface getMainText, getSecondaryText and letting the item provide reasonable values. Might require some way to toggle the display.
I don't think this is a good idea but thought I'd mention it. 

Another possible way would be to use adaptors that could contain some logic for item types. These would act as an interface between the CFileItem and the ui view. Thinking the view would have one adaptor and not one per item. This would require a decent grouping of items that can be displayed together. 

I can probably expand on the above ideas but writing this on my phone

Rechi:
@notspiff this needs a rebase

Voyager1:
I don't like these repeated string constants... At least, couldn't we use some defined constants to avoid bugs due to typos?

PR #12987, opened 2017-11-01
[Music]Information Provider dialog 

DaveTBlake:
Add a new dialog for selection of both artist and album scraper addons, and their settings.

Information Provider  Dialog

The workflow when adding a music source is improved by making the default music scraper settings readily available, from a "settings" button on the "Add to library" prompt.  This way it avoids cluttering the UI or adding unnecessary complexity for beginners. 

Add To Library prompt

Previously if a user wanted to change the music scraping settings (that scanning was about to use) they had to leave the music library, go to System>Media>Music to pick scrapers and then to each addon separately to change any settings. There are few scrapers to choose from, and settings are generally static but much better to be able to do this from where you are adding a music source.

The same dialog is also used to provide the UI for the facility to change information provider for artists and albums separately. This replaces the ugly sub-menu added by #12597 with a select list on the dialog itself.

Change Info Provider (Artists)

This is based on a standard dialog so no skin changes needed.
I have had an early test build up and had positive feedback from @karellen and @jjd-uk


jjd-uk (approved):
+1 Having tested this it's IMHO it's a big improvement to the workflow for adding music to DB. The old way was a right pain in the arse if anything needed changing.

Can't comment on the actual code however.

zag2me:
This is a nice UI improvement! definitely +1

And I thought you were a database guy... :)

xbmc/music/windows/GUIWindowMusicNav.cpp
```diff
@@ -272,47 +271,30 @@ bool CGUIWindowMusicNav::ManageInfoProvider(const CFileItemPtr item)
...
+      scraper = std::dynamic_pointer_cast<ADDON::CScraper>(defaultScraper);
     }
   }
 
   // Set Information provider and settings
-  if (CGUIDialogContentSettings::Show(scraper, content))
+  int applyto = CGUIDialogInfoProviderSettings::Show(scraper);
+  if (applyto >= 0)
   {
+    bool result;
```
Rechi: variable is sometimes used uninitialized

xbmc/music/dialogs/GUIDialogInfoProviderSettings.cpp
```diff
@@ -0,0 +1,470 @@
...
+
+#include <map>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <limits.h>
+
+#include "GUIDialogInfoProviderSettings.h"
```
notspiff: please include first to avoid potential lacking includes in header.

xbmc/music/dialogs/GUIDialogInfoProviderSettings.cpp
```diff
@@ -0,0 +1,470 @@
...
+}
+
+void CGUIDialogInfoProviderSettings::OnInitWindow()
+{
+  CGUIDialogSettingsManualBase::OnInitWindow();
+}
+
+void CGUIDialogInfoProviderSettings::OnSettingChanged(std::shared_ptr<const CSetting> setting)
+{
+  if (setting == NULL)
```
notspiff: nullptr

xbmc/music/dialogs/GUIDialogInfoProviderSettings.cpp
```diff
@@ -0,0 +1,470 @@
...
+  {
+    m_applyToItems = std::static_pointer_cast<const CSettingInt>(setting)->GetValue();
+    SetupView();
+    SetFocus(SETTING_APPLYTOITEMS);
+  }
+}
+
+void CGUIDialogInfoProviderSettings::OnSettingAction(std::shared_ptr<const CSetting> setting)
+{
+  if (setting == NULL)
```
notspiff: nullptr

notspiff (changes requested):
Just a couple of knee-jerks. The lesson I learn here is that if you just put off things like this for approximately 10 years, somebody eventually cleans up your mess! :D

DaveTBlake:
Thanks for code review @notspiff. I tend to copy exitsting code, and as styles evolve that is not always a good thing.

notspiff (approved):

PR #12990, opened 2017-11-01
[linux] remove imx platform

lrusak:
as discussed at devcon.

No maintainer has come forward for this broken method.

DRM/V4L2 is the future and we should concentrate on that instead.


fritsch:
If someone wonders: https://schd.ws/hosted_files/osseu17/e2/zero-copy_video_streaming.pdf here is an example of what will be upcoming. Pure zero copy with dma_buf from decoder to rendering.

lrusak:
Can we put a deadline on this? Say a week from now it is slated for merging?

notspiff:
why wait?

lrusak:
Some people may have strong opinions about this?

notspiff:
code doesn't even build right? those strong opinions would be worthless as i see it. but anyways, not trying to make a thing out of it, just doesn't make sense to wait from my understanding.

fritsch:
Push it. It's git, it forgets nothing.

Am 03.11.2017 15:32 schrieb "Arne Morten Kvarving" <notifications@github.com

lrusak:
jenkins build this please

PR #12995, opened 2017-11-03
Make Player.HasVideo exclusive for video

garbear:
Currently, `Player.HasVideo` will return true if a game is being played. This causes a bug in the FullscreenGame window where the `Fullscreen` action attempts to open FullscreenVideo instead of leaving FullscreenGame for the GUI.

## Motivation and Context
Technically, games have video. I relied on this technicality in the original RetroPlayer PR to reduce the amount of code. Just imagine trying to rebase these hundreds of lines of changes spread throughout Kodi every month...

Thanks to @AchimTuran for reporting this bug during DevCon.

## How Has This Been Tested?
The bug causing the following screenshot no longer occurs on OSX.

The logic refactors will need another set of eyes. When I split FullscreenGame and FullscreenVideo, a small bug snuck in, so it's possible I missed something again.

screen shot 2017-11-02 at 11 30 51 pm

## Types of change

### DIFF
--- a/xbmc/ApplicationPlayer.cpp
+++ b/xbmc/ApplicationPlayer.cpp
@@ -252,12 +252,12 @@ bool CApplicationPlayer::IsPausedPlayback()
 
 bool CApplicationPlayer::IsPlayingAudio() const
 {
-  return (IsPlaying() && !HasVideo() && HasAudio());
+  return (IsPlaying() && !HasVideo() && !HasGame() && HasAudio());
 }
 
 bool CApplicationPlayer::IsPlayingVideo() const
 {
-  return (IsPlaying() && HasVideo());
+  return (IsPlaying() && !HasGame() && HasVideo());
 }
 
 bool CApplicationPlayer::IsPlayingGame() const
@@ -778,7 +778,7 @@ void CApplicationPlayer::SetPlaySpeed(float speed)
   if (!player)
     return;
 
-  if (!IsPlayingAudio() && !IsPlayingVideo())
+  if (!IsPlaying())
     return ;
 
   SetSpeed(speed);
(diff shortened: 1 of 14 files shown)


garbear:
ping @Montellese for the JSON-RPC change. Do we want this?

xbmc/ApplicationPlayer.cpp
```diff
@@ -224,12 +224,12 @@ bool CApplicationPlayer::IsPausedPlayback()
 
 bool CApplicationPlayer::IsPlayingAudio() const
 {
-  return (IsPlaying() && !HasVideo() && HasAudio());
+  return (IsPlaying() && !HasVideo() && !HasGame() && HasAudio());
```
FernetMenta: this is a perfect indicator for a weak design.

FernetMenta:
> Make Player.HasVideo exclusive for video

you changed 14 files for this intent. imo this change made the already weak design even worse.
instead of trying to make clients aware of player's state, we should have players (or ApplicationPlayer) control what should happen.

garbear:
How would this look?

FernetMenta:
let's take CApplication::CheckScreenSaverAndDPMS as an example. Currently this method has to be aware of all possible reasons that may inhibit screen saver or dpms. CApplicationPlayer knows player state. It can place an inhibitor for not going into DPMS without CApplication having to know the reason.

CApplication::OnAction
this entire condition checking should go away. players know when to handle a command. if a component handles the action, it signals back to app.

...

garbear:
Where's a good place to start for moving state checks into the player?

FernetMenta:
why don't you start with the two examples I gave?

PR #12998, opened 2017-11-03
[win32][cmake][fix] Fix detection of FXC for win10 sdk

Paxxi:
Win 10 SDK uses versioned subdirectories so this never worked but fell back to the Windows 8/8.1 SDK.

The trailing .0 in the path is there because the ProductVersion in registry doesn't include it but all released SDKs so far use it in the folder names.

## Types of change

## Checklist:

### DIFF
--- a/cmake/modules/FindD3DX11Effects.cmake
+++ b/cmake/modules/FindD3DX11Effects.cmake
@@ -23,6 +23,7 @@ endif()
 find_program(FXC fxc
              PATHS
                "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Wow6432Node\\Microsoft\\Microsoft SDKs\\Windows\\v10.0;InstallationFolder]/bin/x86"
+               "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Wow6432Node\\Microsoft\\Microsoft SDKs\\Windows\\v10.0;InstallationFolder]/bin/[HKEY_LOCAL_MACHINE\\SOFTWARE\\Wow6432Node\\Microsoft\\Microsoft SDKs\\Windows\\v10.0;ProductVersion].0/x86"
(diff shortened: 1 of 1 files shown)


Rechi:
On my system I have the `fcx.exe` at the following places
```C:\Program Files (x86)\Windows Kits\10\bin\arm64\fxc.exe
C:\Program Files (x86)\Windows Kits\10\bin\x64\fxc.exe
C:\Program Files (x86)\Windows Kits\10\bin\x86\fxc.exe
C:\Program Files (x86)\Windows Kits\8.1\bin\x64\fxc.exe
C:\Program Files (x86)\Windows Kits\8.1\bin\x86\fxc.exe
```
So it isn't located where it is now seeked.

Paxxi:
image
image

Is what it looks like on my system but they're available in the versioned path.

cmake/modules/FindD3DX11Effects.cmake
```diff
@@ -22,7 +22,7 @@ endif()
 
 find_program(FXC fxc
              PATHS
-               "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Wow6432Node\\Microsoft\\Microsoft SDKs\\Windows\\v10.0;InstallationFolder]/bin/x86"
+               "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Wow6432Node\\Microsoft\\Microsoft SDKs\\Windows\\v10.0;InstallationFolder]/bin/[HKEY_LOCAL_MACHINE\\SOFTWARE\\Wow6432Node\\Microsoft\\Microsoft SDKs\\Windows\\v10.0;ProductVersion].0/x86"
```
Rechi: just keep the line as it was and add an additional which matches your path

Paxxi:
sounds like a plan

hudokkow:
Just to add to @Rechi's input, my jenkins node also has

`C:\Program Files (x86)\Windows Kits\10\bin\arm64\fxc.exe`

PR #13001, opened 2017-11-03
[fix][websocket] Connection header must include not match Upgrade

Rechi:
## Description
RFC 6455 4.1. Client Requirements
The request MUST contain a |Connection| header field whose value MUST include the "Upgrade" token.

## Motivation and Context
newer Firefox versions send `Connection: keep-alive, Upgrade`

## How Has This Been Tested?
Tested with Firefox 56.

## Types of change

## Checklist:

### DIFF
--- a/xbmc/network/websocket/WebSocketV13.cpp
+++ b/xbmc/network/websocket/WebSocketV13.cpp
@@ -18,6 +18,7 @@
  *
  */
 
+#include <algorithm>
 #include <string>
 #include <sstream>
 
@@ -102,7 +103,10 @@ bool CWebSocketV13::Handshake(const char* data, size_t length, std::string &resp
 
   // There must be a "Connection" header with the value "Upgrade"
   value = header.getValue(WS_HEADER_CONNECTION_LC);
-  if (value == NULL || strnicmp(value, WS_HEADER_UPGRADE, strlen(WS_HEADER_UPGRADE)) != 0)
+  std::vector<std::string> elements;
+  if (value != nullptr)
+    elements = StringUtils::Split(value, ",");
+  if (elements.empty() || !std::any_of(elements.begin(), elements.end(), [](std::string& elem) { return StringUtils::EqualsNoCase(StringUtils::Trim(elem), WS_HEADER_UPGRADE); }))
   {
     CLog::Log(LOGINFO, "WebSocket [RFC6455]: invalid \"%s\" received", WS_HEADER_CONNECTION_LC);
     return true;


xbmc/network/websocket/WebSocketV13.cpp
```diff
@@ -102,7 +104,11 @@ bool CWebSocketV13::Handshake(const char* data, size_t length, std::string &resp
 
   // There must be a "Connection" header with the value "Upgrade"
   value = header.getValue(WS_HEADER_CONNECTION_LC);
-  if (value == NULL || strnicmp(value, WS_HEADER_UPGRADE, strlen(WS_HEADER_UPGRADE)) != 0)
+  if (value == nullptr)
+    elements = std::vector<std::string>();
+  else
+    elements = StringUtils::Split(value, ",");
+  if (elements.empty() || !std::any_of(elements.cbegin(), elements.cend(), [](std::string elem) { return StringUtils::EqualsNoCase(StringUtils::Trim(elem), WS_HEADER_UPGRADE); }))
```
garbear: can you pass `elem` by reference or use move constructors?

Rechi: pass by reference doesn't work (StringUtils::Trim doesn't accept const)
`... StringUtils::Trim(std::move(elem)) ...` is the move syntax you mean?

garbear: yup

MartijnKaijser:
@Montellese if you could have a brief look? If all good we can put it in v17 and do last update.

xbmc/network/websocket/WebSocketV13.cpp
```diff
@@ -102,7 +104,11 @@ bool CWebSocketV13::Handshake(const char* data, size_t length, std::string &resp
 
   // There must be a "Connection" header with the value "Upgrade"
   value = header.getValue(WS_HEADER_CONNECTION_LC);
-  if (value == NULL || strnicmp(value, WS_HEADER_UPGRADE, strlen(WS_HEADER_UPGRADE)) != 0)
+  if (value == nullptr)
+    elements = std::vector<std::string>();
```
Montellese: Why is this necessary? `elements` has been declared and initialized as `std::vector<std::string>` already.

Montellese (changes requested):
Looks good apart from the unnecessary re-initialization of `elements`.

xbmc/network/websocket/WebSocketV13.cpp
Rechi: just wanted to make it safe for the future, but I'll move elements declaration down here

Rechi:
@Montellese removed the  unnecessary re-initialization of `elements`

Montellese (approved):

PR #13008, opened 2017-11-05
[win10] temporary controller workaround until proper joystick support

koying:
Temporary workaround for gamepad on xbox based upon @afedchin & @Paxxi suggestions

For reference, unless we want to put this in until @garbear has time

### DIFF
--- a/xbmc/windowing/win10/WinEventsWin10.cpp
+++ b/xbmc/windowing/win10/WinEventsWin10.cpp
@@ -388,17 +388,28 @@ VirtualKey keyDown = VirtualKey::None;
 
 void CWinEventsWin10::OnKeyDown(CoreWindow^ sender, KeyEventArgs^ args)
 {
-  if (!args->KeyStatus.IsExtendedKey)
+  int vk = static_cast<unsigned>(args->VirtualKey);
+  if ((vk == 0x08) // VK_BACK
+    || (vk == 0x09) // VK_TAB
+    || (vk == 0x0C) // VK_CLEAR
+    || (vk == 0x0D) // VK_RETURN
+    || (vk == 0x1B) // VK_ESCAPE
+    || (vk == 0x20) // VK_SPACE
+    || (vk >= 0x30 && vk <= 0x39) // numeric keys
+    || (vk >= 0x41 && vk <= 0x5A) // alphabetic keys
+    || (vk >= 0x60 && vk <= 0x6F) // keypad keys
+    || (vk >= 0x92 && vk <= 0x96) // OEM specific
+    || (vk >= 0xBA && vk <= 0xC0) // OEM specific
+    || (vk >= 0xDB && vk <= 0xDF) // OEM specific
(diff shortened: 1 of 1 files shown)


Paxxi (approved):
Looks good to me for now. The whole thing will need a rewrite in the future anyway

koying:
For the record, I tried afedchin's solution and I had issues with plain keyboard mappings as well.
As far as I could read, OnCharacterReceived is the only way to properly get the unicode of a keypress

afedchin:
what's wrong with https://github.com/afedchin/xbmc/commit/d64a84e76f84a17e7fe57c40da118ece74756774 ?

afedchin:
did you try with fixed peripherals.joystick?

koying:
> what's wrong with afedchin/xbmc@d64a84e ?

When using keyboard on PC, some of the alphanum were showing numbers, on the left of my azerty keyboard, a bit like if they were translated to numpad.

Didn't try with addon, by my droid philosophy is that when the gamepad is the primary input method, it should always work, joystick addon or not ;)
Addons are... addons. There shouldn't be such thing as a "required" addon.

Montellese:
@koying that understanding sounds incompatible with what @garbear has done with the `peripheral.joysticks` add-on. AFAIK no controller (like Xbox 360, Xbox One, ...) works without that add-on on other platform (unless they act like a keyboard) and the add-on has to be part of the kodi installation (just like the default skin add-on or the english language add-on etc.)

koying:
@Montellese Well, pretty sure it works w/o the addon on droid (at least as far as I remember; I even remember some discussions with garbear to get it to work that way), and, indeed, it's due to the fact that they basically act as keyboards (or are emulated as such) by the OS.

It's exactly the same situation on UWP.

Anyway, see https://github.com/xbmc/xbmc/pull/13008/files?w=1 for the actual diff w/o eol issue.
Controller working is just a side effect, really. 

The 
```
-  if (!args->KeyStatus.IsExtendedKey)
-  {
-    // fix F1-F24 keys
-    if (args->VirtualKey < VirtualKey::F1 || args->VirtualKey > VirtualKey::F24)
-    {
```
seems pretty bogus, as discussed

afedchin:
> When using keyboard on PC, some of the alphanum were showing numbers, on the left of my azerty keyboard, a bit like if they were translated to numpad.

dont' understand what's difference. the same events handled by one method instead of three, and as I can understand CoreDispatcher event are little more low level

koying:
I think the difference is that `OnAcceleratorKeyActivated` only deals with keycodes, while the current implem actually get unicode chars in `OnCharacterReceived`

afedchin:
I've tested with my russian keyboard and all work perfect.

afedchin:
I've tested Kodi on XB1 with peripheral.joystik and gamepad is working well. no need to extra support it inside keyevents.

koying:
Superseded by https://github.com/xbmc/xbmc/pull/13024

PR #13010, opened 2017-11-05
VideoPlayer: ffmpeg - drop deprecated merging of side data

fritsch:
Fixes playback of certain TrueHD files.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxUtils.cpp
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxUtils.cpp
@@ -36,13 +36,17 @@ void CDVDDemuxUtils::FreeDemuxPacket(DemuxPacket* pPacket)
 {
   if (pPacket)
   {
-    try {
-      if (pPacket->pData) _aligned_free(pPacket->pData);
-      delete pPacket;
+    if (pPacket->pData)
+      _aligned_free(pPacket->pData);
+    if (pPacket->iSideDataElems)
+    {
+      AVPacket avPkt;
+      av_init_packet(&avPkt);
+      avPkt.side_data = static_cast<AVPacketSideData*>(pPacket->pSideData);
+      avPkt.side_data_elems = pPacket->iSideDataElems;
+      av_packet_free_side_data(&avPkt);
     }
-    catch(...) {
-      CLog::Log(LOGERROR, "%s - Exception thrown while freeing packet", __FUNCTION__);
-    }
+    delete pPacket;
   }
 }
 
@@ -81,4 +85,13 @@ DemuxPacket* CDVDDemuxUtils::AllocateDemuxPacket(unsigned int iDataSize, unsigne
   if (ret && encryptedSubsampleCount > 0)
     ret->cryptoInfo = std::shared_ptr<DemuxCryptoInfo>(new DemuxCryptoInfo(encryptedSubsampleCount));
   return ret;
+}
+
+void CDVDDemuxUtils::StoreSideData(DemuxPacket *pkt, AVPacket *src)
+{
+  AVPacket avPkt;
+  av_init_packet(&avPkt);
+  av_packet_copy_props(&avPkt, src);
+  pkt->pSideData = avPkt.side_data;
+  pkt->iSideDataElems = avPkt.side_data_elems;
 }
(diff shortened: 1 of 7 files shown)


fritsch:
Thanks @FernetMenta 

@MartijnKaijser for you to test. Works fine for me with the samples we checked yesterday.

fritsch:
Wait for merging, there is a "minor" segfault involved currently: http://paste.ubuntu.com/25896241/

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxUtils.cpp
```diff
@@ -36,13 +36,17 @@ void CDVDDemuxUtils::FreeDemuxPacket(DemuxPacket* pPacket)
...
-      CLog::Log(LOGERROR, "%s - Exception thrown while freeing packet", __FUNCTION__);
+    if (pPacket->pData)
+      _aligned_free(pPacket->pData);
+    if (pPacket->iSideDataElems)
+    {
+      AVPacket avPkt;
+      av_init_packet(&avPkt);
+      avPkt.side_data = reinterpret_cast<AVPacketSideData*>(pPacket->pSideData);
+      avPkt.side_data_elems = pPacket->iSideDataElems;
+      av_packet_free_side_data(&avPkt);
```
fritsch: Don't we have to set pPacket->iSideDatalElems to 0 too?

fritsch: and src ptr might be null - which directly segfaults: https://ffmpeg.org/doxygen/3.1/avpacket_8c_source.html#l00531

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxUtils.cpp
```diff
@@ -82,3 +86,12 @@ DemuxPacket* CDVDDemuxUtils::AllocateDemuxPacket(unsigned int iDataSize, unsigne
     ret->cryptoInfo = std::shared_ptr<DemuxCryptoInfo>(new DemuxCryptoInfo(encryptedSubsampleCount));
   return ret;
 }
+
+void CDVDDemuxUtils::StoreSideData(DemuxPacket *pkt, AVPacket *src)
+{
+  AVPacket avPkt;
+  av_init_packet(&avPkt);
+  av_packet_copy_props(&avPkt, src);
```
fritsch: src might be null, which directly crashes: https://ffmpeg.org/doxygen/3.1/avpacket_8c_source.html#l00531

xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxUtils.cpp
fritsch: I think the reinterpret_cast is wrong here and we end up with a dangeling ptr.

fritsch:
Ha :-) I missed to rebuild vnsi ... all good.

fritsch:
jenkins build this please

fritsch:
I changed the reinterpret_cast to static_cast

FernetMenta (approved):

FernetMenta:
The ugly thing is that pvr and inputstream addons expose DemuxPacket via their own APIs. Both versions should be bumped.

MartijnKaijser:
@fritsch ping

fritsch:
As a simple rebuild is enough - I hessitate a bit to bump version as we are mid in development.

@FernetMenta what do you think is the right way forward?

FernetMenta:
I don't understand your concerns. We have bumped versions always for things like that.

fritsch:
My concerns are: API is not changed, but ABI - we want to trigger a rebuild, nothing else.

FernetMenta:
API is changed because the header included by an addon is changed. This is per definition an API change that should be reflected by a version bump.

fritsch:
I think what I realy don't like is that we "govern" a hidden API here :-) - but never the less, bumped min version and current version of inputstream and pvr.

fritsch:
@FernetMenta good to go - anyone else I need to inform?

FernetMenta:
good to go

ksooo:
@MilhouseVH please be aware that this PR requires a recompile of all PVR and inputstream addons.

MilhouseVH:
@ksooo thanks - I rebuild all the addons I include in the release every night. Those I don't include...oh well! :)

wsnipex:
just FYI. ALL addons that need to be rebuilt, should have their version bumped, otherwise no PPA updates..

fritsch:
That's the reason this PR stood here for more than 4 weeks - seems it did not work as intended. Do we have a list? Most likely everything inputstream and pvr related that use the packet.

PR #13013, opened 2017-11-05
[PVR] Fix handling of epg tag attributes 'cast', 'writer', 'director' and 'genre'.

ksooo:
EPG tag attributes 'cast', 'writer', 'director' and 'genre' each can contain multiple persons/genres. This was not handled correctly so far. Genres was handled half-way right, the others completey wrong.

@Jalle19 mind taking a look.


Jalle19 (approved):

ksooo:
Jenkins errors are unrelated.

DaveTBlake:
@ksooo does this not need JSON API changes too? Switching serialized items from single to array etc. probably means a breaking change to JSON

ksooo:
> Switching serialized items from single to array

I did not change external representation. Only internal representation changed from single string to vector. External rep is still single string.

DaveTBlake:
The change just caught my eye as I have been looking at fixing some places in music where API expects a string and is getting an array, or vica-versa. Also adding where data exists but is not exposed at all.  

I am not familiar with PVR, but I know how easy it is to overlook the JSON aspects of changes to tag handling. Might be worth a review when you have time.

PR #13016, opened 2017-11-06
osx: fix high cpu usage when obscured

FernetMenta:
see title

### DIFF
--- a/xbmc/windowing/osx/WinSystemOSXGL.mm
+++ b/xbmc/windowing/osx/WinSystemOSXGL.mm
@@ -38,6 +38,11 @@ void CWinSystemOSXGL::PresentRenderImpl(bool rendered)
   if (rendered)
     FlushBuffer();
 
+  // FlushBuffer does not block if window is obscured
+  // in this case we need to throttle the render loop
+  if (IsObscured())
+    usleep(10000);
+
   if (m_delayDispReset && m_dispResetTimer.IsTimePast())
   {
     m_delayDispReset = false;


Memphiz:
Oha ... didn't expect this sort of workaround from you tbh - but am totally fine with is. But it could need a comment on why we sleep there ...

FernetMenta:
this is no workaround at all. the architecture delegates throteling of the render/main loop to windowing. if, for what reason ever, swapBuffers does not block, windowing has to take action.

Rechi:
Is this worth a backport (applies cleanly) or is it to risky?

FernetMenta:
I don't think it is risky.

FernetMenta:
comment added

Memphiz:
Thx - sorry somehow i was confused because the last time we had this cpu usage problem it was in the dirty regions area of things. Guess this throttle not working is something in the osx gl implementation then (were swapbuffers is a nop or something when kodi is offscreen maybe).

FernetMenta:
@Memphiz the method IsObscured was actually implemented for this purpose: https://github.com/xbmc/xbmc/commit/de396863c328ddf9fa66e8eb548c1d541acfc338

It got lost because the old code violated a basic OO principle: information hiding. OSX promoted its private function to CApplication.

Memphiz:
Yes i know that method very well - i had to add exceptions for all those full screen alpha translucent apps that are meant to decrease brightness (which resulted in frame limiter kicking in). So in case people are reporting stuttering ui rendering - we need to have those apps in mind aswell... ;)

Rechi:
jenkins build this please

Rechi:
WIN-UWP-ARM failure is unrelated

FernetMenta:
build error unrelated

MartijnKaijser:
@Memphiz you want this backported or risk too high?

Memphiz:
No risk immo @fernetmenta what do you think?

FernetMenta:
I think risk is low. Note that I did not test it on v17

PR #13019, opened 2017-11-07
[python-api] InfoTagVideo additional methods

regagain:
Exposed more information from InfoTagVideo to the python API.

## Description
Added the methods _getArtist()_, _getAlbum()_, _getTrack()_ and _getDuration()_ to _InfoTagVideo_.

## Motivation and Context
Mostly useful for musicvideos. For example, thanks to these methods, musicvideos could be scrobbled to Libre.fm/Last.fm or ListenBrainz.

## How Has This Been Tested?
* Added debug prints so I could see that the information was retrieved properly when playing videos.
* Made a POC from [service.scrobbler.lastfm](https://github.com/XBMC-Addons/service.scrobbler.lastfm/) to scrobble musicvideos.
* My test environment is just my laptop, which is running Linux & Kodi HEAD.
* Since this only retrieves existing information, I don't think it should impact other areas.

## Types of change

## Checklist:
- [X] I have updated the documentation accordingly


xbmc/interfaces/legacy/InfoTagVideo.h
```diff
@@ -565,6 +565,78 @@ namespace XBMCAddon
...
+#ifdef DOXYGEN_SHOULD_USE_THIS
+      ///
+      /// \ingroup python_InfoTagVideo
+      /// @brief \python_func{ getArtist() }
+      ///-----------------------------------------------------------------------
+      /// To get the artist name (for musicvideos)
+      ///
+      /// @return [string] Artist name
+      ///
+      ///
```
ronie: can you add a 'new function' comment to each of the new methods?

```/// @python_v18 New function added.```

**Comment by ronie** [comment]
thx! code looks good to me.

File: xbmc/interfaces/legacy/InfoTagVideo.cpp [hunk 2]
```diff
@@ -176,5 +176,25 @@ namespace XBMCAddon
     {
       return infoTag->m_strTrailer;
     }
+
+    String InfoTagVideo::getArtist()
+    {
+      return StringUtils::Join(infoTag->m_artist, g_advancedSettings.m_videoItemSeparator);
```
tamland: APIs shouldn't depend on user settings. Just return the original string list such that the consumers can format it as they please

xbmc/interfaces/legacy/InfoTagVideo.h
```diff
@@ -565,6 +565,78 @@ namespace XBMCAddon
...
+      /// @return [unsigned int] Duration
+      ///
+      ///
+      ///-----------------------------------------------------------------------
+      ///
+      ///
+      getDuration();
+#else
+      int getDuration();
```
ace20022: unsigned int

xbmc/interfaces/legacy/InfoTagVideo.cpp
```diff
@@ -176,5 +176,25 @@ namespace XBMCAddon
...
+    {
+      return infoTag->m_strAlbum;
+    }
+
+    int InfoTagVideo::getTrack()
+    {
+      return infoTag->m_iTrack;
+    }
+
+    int InfoTagVideo::getDuration()
```
ace20022: unsigned int

xbmc/interfaces/legacy/InfoTagVideo.h
regagain: Done, see 411388e84029726b6a62317703d794c1e7d74dfe

xbmc/interfaces/legacy/InfoTagVideo.cpp
regagain: Makes sense! This is done, see c1ebe41efa8c2f1aaaea7a7bd06fad57dbb1b25b.

Note: _getDirector()_, _getWritingCredits()_ and _getGenre()_ also do this. I dare not change them for fear of breaking existing functionality (this being said I'd love to, but I feel I would need some guidance).

xbmc/interfaces/legacy/InfoTagVideo.cpp
regagain: Thanks! See e62a54e7917724d880021102d2fd49c70d5a9d04.
I also had to add support for unsigned int in PythonSwig.cpp.template, see 61e31eef0c2557e4dbe3064157fbdbfc677f4eac.

Rechi:
Can you please squash/reorder the commits a bit
- Comments should go into the same commit where the function is added.
- changing the return type for `getDuration` should be added to the function addition commit
- same for `getArtist` function changes
- PythonSwig change should be before the commit which needs the change

You might also want adjust the email address of your commits to the one you use for github that your commits get associated with your github user.

If you need assistance just ask.

xbmc/interfaces/legacy/InfoTagVideo.cpp
phil65: Perhaps you could add new API methods getDirectors(), getWriters() and getGenres() which return a proper list, and we deprecate the old ones then? (this should be done in a separate PR though)
Or we delay bw-incompatible  API changes to the point when we switch to Py3?

regagain:
Done!

xbmc/interfaces/legacy/InfoTagVideo.cpp
Rechi: No we don't want to break Python API during the switch to Py3.

Rechi:
wrong revision was build
jenkins build this please

Rechi:
@ronie good to go?

ronie:
yes sure, go ahead :-)

PR #11455, opened 2017-01-15
fixed: make news available when browsing add-on repositories

notspiff:
News was only filled for local add-ons, not from repository listings. Remedy this (i.e. put it into addons.db)

Ref http://forum.kodi.tv/showthread.php?tid=303941

PR #13021, opened 2017-11-07
Fix missing add-on description for repo add-ons

garbear:
This must have broke when we added the News field to the add-on database.

## Motivation and Context
Gotta have those sexy descriptions for the game controller museum.

## How Has This Been Tested?
See screenshots.

Before:
screenshot000

After:
screenshot001

## Types of change

### DIFF
--- a/xbmc/addons/AddonDatabase.cpp
+++ b/xbmc/addons/AddonDatabase.cpp
@@ -647,7 +647,7 @@ bool CAddonDatabase::GetAddon(int id, AddonPtr &addon)
     builder.SetVersion(AddonVersion(m_pDS2->fv(3).get_asString()));
     builder.SetName(m_pDS2->fv(4).get_asString());
     builder.SetSummary(m_pDS2->fv(5).get_asString());
-    builder.SetDescription(m_pDS2->fv(6).get_asString());
+    builder.SetDescription(m_pDS2->fv("description").get_asString());
     DeserializeMetadata(m_pDS2->fv(1).get_asString(), builder);
     addon = builder.Build();
     return addon != nullptr;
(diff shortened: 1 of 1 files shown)


xbmc/addons/AddonDatabase.cpp
```diff
@@ -732,7 +732,7 @@ bool CAddonDatabase::GetRepositoryContent(const std::string& id, VECADDONS& addo
       builder.SetVersion(version);
       builder.SetName(m_pDS->fv(4).get_asString());
       builder.SetSummary(m_pDS->fv(5).get_asString());
-      builder.SetDescription(m_pDS->fv(6).get_asString());
+      builder.SetDescription(m_pDS->fv("description").get_asString());
```
xhaggi: if we use named columns here, should we do the same for all the other ones?

garbear: Yea. The alternative is to explicitly list the fields in the SELECT statement above so that the field order matches the ints. I defer this decision to someone else.

xhaggi: I would prefer to name the columns on the getters, so we can leave `SELECT *`.

garbear: Then should I update the other ints to use column names?

xhaggi: Would be nice if you had the time

xhaggi (commented):
Great thanks

garbear:
@xhaggi updated. I'll test by browsing the add-on manager later. I also found another SELECT * statement for the installed table, what's the best way to test this?

tamland:
As mentioned in #11455 the schema is broken and should be repaired

garbear:
So can this be merged? I don't want to spend a day refactoring a system I didn't write.

tamland:
It's a db migration to revert it, not a day of refactoring. But I don't want to waste my time on this mess either so workaround it is.

PR #13023, opened 2017-11-08
[PVR] Fix CPVR(Internal)ChannelGroup::Update to respect error code fr…

ksooo:
…om LoadFromClients. No need to update anything if LoadFromClients failed.

Runtime tested on macOS 10.13, latest Kodi master.

@Jalle19 good to go?


Jalle19 (approved):
What if LoadFromClients fails halfway through? Now we get nothing instead of something. Still looks good though.

ksooo:
> What if LoadFromClients fails halfway through? Now we get nothing instead of something.

 Good point. Have to think once again about this. Maybe, 'something' actually works. Have to check whether there is some semi-cute 'I can delete a group if it is not provided by the update set' logic kicks in. Maybe this works with my recent changes to detect and respect failed clients...

ksooo:
Yep, partial update should work nowadays. Closing this PR...

PR #13026, opened 2017-11-09
added: ability to only clean a given video library content type

notspiff:
Allow CleanLibrary(movies|tvshows|musicvideos) in addition. This will only clean paths with the given content type. requested by klojum.

@xhaggi @paxxi (nit-)pick it to pieces :)


xbmc/Application.h
```diff
@@ -255,8 +255,9 @@ class CApplication : public CXBApplicationEx, public IPlayerCallback, public IMs
   /*!
    \brief Starts a video library cleanup.
    \param userInitiated Whether the action was initiated by the user (either via GUI or any other method) or not.  It is meant to hide or show dialogs.
+   \param content Content type to clean, blank for everything
    */
-  void StartVideoCleanup(bool userInitiated = true);
+  void StartVideoCleanup(bool userInitiated = true, const std::string& content="");
```
ksooo: Would you mind putting a blank around `=`? Don't know whether this is part of our coding guidelines, but this is how we usually format our code.

notspiff: not at all. added.

ksooo: man, you are pretty fast!

xbmc/interfaces/builtins/LibraryBuiltins.cpp
```diff
@@ -48,10 +48,16 @@ static int CleanLibrary(const std::vector<std::string>& params)
...
-  if (!params.size() || StringUtils::EqualsNoCase(params[0], "video"))
+  if (!params.size() || StringUtils::EqualsNoCase(params[0], "video")
+                     || StringUtils::EqualsNoCase(params[0], "movies")
+                     || StringUtils::EqualsNoCase(params[0], "tvshows")
+                     || StringUtils::EqualsNoCase(params[0], "musicvideos"))
   {
     if (!g_application.IsVideoScanning())
-      g_application.StartVideoCleanup(userInitiated);
+    {
+      std::string type = (params.empty() || params[0] == "video") ? "" : params[0];
```
ksooo: could be `const std::string ...`

xbmc/Application.cpp
```diff
@@ -5010,12 +5010,34 @@ void CApplication::StopMusicScan()
...
+    {
+      for (const std::string& path : content_paths)
+      {
+        if (db.GetContentForPath(path) == content)
+        {
+          paths.insert(db.GetPathId(path));
+          std::vector<std::pair<int,std::string>> sub;
+          if (db.GetSubPaths(path, sub))
+          {
+            for (auto& it : sub)
```
ksooo: could be `const auto& ...

xbmc/interfaces/builtins/LibraryBuiltins.cpp
notspiff: done.

xbmc/Application.cpp
notspiff: done.

xbmc/interfaces/builtins/LibraryBuiltins.cpp
ksooo: thx

xbmc/Application.cpp
ksooo: Thanks

ksooo:
👍

MartijnKaijser:
One side note is that currently when we hit clean library we also try every plugin:// path if it's still valid which takes a really long time if you use a lot plugins. Maybe you can think of a way to do this in a nice way.
https://trac.kodi.tv/ticket/15950

notspiff:
there's not much we can do about it, either
1) ignore them
2) rip em all out on clean.

MartijnKaijser:
It's nice to have watch status in plugins if we go to several official ones so i'd go for ignore in that case. Haven't tried lately but it used to block everything else so if it as low prio job then it doesn't really matter in that regard that we try to clean it. It does hit the content provider each time though that we check for the file.

notspiff:
then it's something different, plugin:// paths would not hit any providers, it would just fail and remove them. i think what you mean is http(s) et al. that's a bit tricker..

Montellese:
How does a stat operation on a `plugin://` path work? Does it call the plugin and wait for a response and does the plugin load the resource? If so that obviously is rather slow (even just calling the plugin). What could be done is to simply check if the respective plugin / add-on the file belongs to is installed. If it is keep the file reference if it isn't remove the file reference.

Either way having links to plugin files in the database is not the same as having links to library files in the database because the latter still appear in your library whereas the former don't. They are just there to enrich the plugin experience.

notspiff:
on playback, we do a callback to the plugin asking it to resolve the virtual plugin:// path to the real playback url. this is how you handle dynamic urls. we store settings based on the plugin:// url since that is supposed to be static.

DaveTBlake:
Update to the Doxygen comments needed here https://github.com/notspiff/xbmc/blob/41eba302b9403b3bf47bea8d10c7cc9d6a44a39f/xbmc/interfaces/builtins/LibraryBuiltins.cpp#L311
so that interface change gets documented.

Also update to JSON API could be considered as it exposes this routine
https://github.com/xbmc/xbmc/blob/master/xbmc/interfaces/json-rpc/VideoLibrary.cpp#L911
modifying  https://github.com/xbmc/xbmc/blob/master/xbmc/interfaces/json-rpc/schema/methods.json#L1754

notspiff:
done and done. i cba to test the json-rpc update.

notspiff:
rpc tested. that made me realize a minor issue which i have rectified - if user has no content of the requested type we would clean the entire library. now we just return.

wsnipex:
jenkins build this please

DaveTBlake:
@notspiff  with JSON API change (and fix) the version also needs a bump (not sure if minor or just revision I always get confused about that, maybe @Montellese will say)

notspiff:
this is a non-breaking change so we can just bump patch.

notspiff:
jenkins build this please

xbmc/interfaces/json-rpc/schema/methods.json
```diff
@@ -1753,11 +1753,12 @@
...
   "VideoLibrary.Clean": {
     "type": "method",
-    "description": "Cleans the video library from non-existent items",
+    "description": "Cleans the video library for non-existent items",
     "transport": "Response",
     "permission": "RemoveData",
     "params": [
-      { "name": "showdialogs", "type": "boolean", "default": true, "description": "Whether or not to show the progress bar or any other GUI dialog" }
+      { "name": "showdialogs", "type": "boolean", "default": true, "description": "Whether or not to show the progress bar or any other GUI dialog" },
+      { "name": "content", "type": "string", "default": false, "enum": [ "videos", "movies", "tvshows", "musicvideos" ], "description": "Content type to clean for" }
```
DaveTBlake: This has default = false, but that is not one of the enum values

xbmc/interfaces/json-rpc/schema/version.txt
```diff
@@ -1 +1 @@
-JSONRPC_VERSION 8.4.0
+JSONRPC_VERSION 8.4.1
```
DaveTBlake: I think this should be 8.5.0, as it is an API addition

notspiff: you bump 
- major on complete redesign
- minor on API breakage
- patch on everything else

so then you apply other rules than the rest of the world

xbmc/interfaces/json-rpc/schema/methods.json
notspiff: ah,  through default was for making an argument optional, remedied.

xbmc/interfaces/json-rpc/schema/version.txt
DaveTBlake: Sorry, I don't think that sounds right

notspiff: i cba to fight over this, you will have your 8.5.0. but you *ARE* wrong.

DaveTBlake: For future see https://github.com/notspiff/xbmc/blob/1fdfd4d4b5c9a02a4fec94b3eb6a9576a421d316/xbmc/interfaces/json-rpc/schema/methods.json#L33

notspiff: you do apply other rules than the rest of the world.

DaveTBlake: Entirely possible, they are not my rules. Sorry I didn't spot immediately and we had this tennis

notspiff: nw. i should have spotted it as well, assumptions are the mother of all evil. and i should stop using 'you' and start using 'we'. habits take some time to change..

xbmc/Application.cpp
```diff
@@ -5010,12 +5010,36 @@ void CApplication::StopMusicScan()
...
+                                     const std::string& content /* = "" */)
 {
   if (userInitiated && CVideoLibraryQueue::GetInstance().IsRunning())
     return;
 
   std::set<int> paths;
+  if (!content.empty())
+  {
+    CVideoDatabase db;
+    std::set<std::string> content_paths;
```
xhaggi: camel-case `contentPaths`

xbmc/Application.cpp
```diff
@@ -5010,12 +5010,36 @@ void CApplication::StopMusicScan()
...
+    if (db.Open() && db.GetPaths(content_paths))
+    {
+      for (const std::string& path : content_paths)
+      {
+        if (db.GetContentForPath(path) == content)
+        {
+          paths.insert(db.GetPathId(path));
+          std::vector<std::pair<int,std::string>> sub;
```
xhaggi: space between type1 and type2 -> `std::pair<int, std::string>`

xbmc/interfaces/builtins/LibraryBuiltins.cpp
```diff
@@ -48,10 +48,16 @@ static int CleanLibrary(const std::vector<std::string>& params)
...
-  if (!params.size() || StringUtils::EqualsNoCase(params[0], "video"))
+  if (!params.size() || StringUtils::EqualsNoCase(params[0], "video")
+                     || StringUtils::EqualsNoCase(params[0], "movies")
+                     || StringUtils::EqualsNoCase(params[0], "tvshows")
+                     || StringUtils::EqualsNoCase(params[0], "musicvideos"))
   {
     if (!g_application.IsVideoScanning())
-      g_application.StartVideoCleanup(userInitiated);
+    {
+      const std::string type = (params.empty() || params[0] == "video") ? "" : params[0];
```
xhaggi: rename `type` to `content`

xbmc/interfaces/builtins/LibraryBuiltins.cpp
```diff
@@ -48,10 +48,16 @@ static int CleanLibrary(const std::vector<std::string>& params)
   bool userInitiated = true;
```
xhaggi: As you use `param[0]` very often should we assign `param[0]` to a local variable e.g. `std::string content = param[0]`

xhaggi (commented):
Some minors .. everything else looks good.

xbmc/Application.cpp
notspiff: done.

xbmc/Application.cpp
notspiff: done.

xbmc/interfaces/builtins/LibraryBuiltins.cpp
notspiff: how many strings named content do you want ;) in any case this violates my ODR rule. param[0] tells me much more than some random string named content. for instance, i immediately know it is a parameter to the built-in that is externally provided through the generic context. if it was put into a temporary i would have to read more code to understand its origin. so i disagree and won't do this unless you insist.

xbmc/interfaces/builtins/LibraryBuiltins.cpp
notspiff: done.

xbmc/Application.cpp
```diff
@@ -5010,12 +5010,36 @@ void CApplication::StopMusicScan()
...
 
-void CApplication::StartVideoCleanup(bool userInitiated /* = true */)
+void CApplication::StartVideoCleanup(bool userInitiated /* = true */,
+                                     const std::string& content /* = "" */)
 {
   if (userInitiated && CVideoLibraryQueue::GetInstance().IsRunning())
     return;
 
   std::set<int> paths;
+  if (!content.empty())
```
Paxxi: Reverse check to avoid some nesting, same for next if

xbmc/interfaces/json-rpc/VideoLibrary.cpp
```diff
@@ -910,7 +910,12 @@ JSONRPC_STATUS CVideoLibrary::Export(const std::string &method, ITransportLayer
 
 JSONRPC_STATUS CVideoLibrary::Clean(const std::string &method, ITransportLayer *transport, IClient *client, const CVariant &parameterObject, CVariant &result)
 {
-  std::string cmd = StringUtils::Format("cleanlibrary(video, %s)", parameterObject["showdialogs"].asBoolean() ? "true" : "false");
+  std::string cmd;
+  if (parameterObject["content"].empty())
+    cmd = StringUtils::Format("cleanlibrary(video, %s)", parameterObject["showdialogs"].asBoolean() ? "true" : "false");
```
Paxxi: use new format string `{}` or positional `{0}`

xbmc/interfaces/json-rpc/VideoLibrary.cpp
```diff
@@ -910,7 +910,12 @@ JSONRPC_STATUS CVideoLibrary::Export(const std::string &method, ITransportLayer
...
+  else
+    cmd = StringUtils::Format("cleanlibrary(%s, %s)", parameterObject["content"].asString().c_str(), parameterObject["showdialogs"].asBoolean() ? "true" : "false");
```
Paxxi: same here and no need to call `c_str()` `StringUtils::Format` uses `fmt` so it handles complex types fine

Paxxi (commented):
just some minor nitpicking 😄 

The format strings should be fixed, the nesting is just a preference so do as you like with it.

xbmc/Application.cpp
notspiff: i don't get it.

xbmc/interfaces/json-rpc/VideoLibrary.cpp
notspiff: done. the new-fangled shit offers absolutely nothing here, but..

xbmc/interfaces/json-rpc/VideoLibrary.cpp
notspiff: done.

xbmc/interfaces/builtins/LibraryBuiltins.cpp
xhaggi: Okay leave it

xhaggi (approved):

xhaggi:
Thx for addressing my nit-picking

xbmc/Application.cpp
Paxxi: if you do `if (content.empty) return paths;` and same with next if there's too levels less of nesting/indentation and easier to read. But as I said it's a matter of taste so I'm just mentioning it

notspiff: I an not returning, it i am sending the array into the db clean after the if nest.

xbmc/interfaces/json-rpc/VideoLibrary.cpp
Paxxi: it's slightly more performant because of how it's implemented. The new format strings are tested first and if no replacements are done it tries to fall back on old formats https://github.com/xbmc/xbmc/blob/master/xbmc/utils/StringUtils.h#L67

xbmc/Application.cpp
Paxxi: right, my example above should've just been `if (paths.empty()) return;` and `if (!(db.Open() && db.GetPaths(contentPaths)) return;`

notspiff: Which still would be wrong. If no content is specified we clean all content type and in that case we pass an empty vector.

Paxxi: oh right. I see it now. Sorry about that.

notspiff:
this has been heavily scrutinized and has waited 10 days for a review of a version number. i think that's enough patience ?

Rechi:
wrong revision was build
jenkins build this please

xbmc/interfaces/builtins/LibraryBuiltins.cpp
```diff
@@ -62,6 +68,8 @@ static int CleanLibrary(const std::vector<std::string>& params)
     else
       CLog::Log(LOGERROR, "CleanLibrary is not possible while scanning for media info");
   }
+  else
+    CLog::Log(LOGERROR, "Unknown content type '{0}' passed to CleanLibrary, ignoring", params[0]);
```
Rechi: `params[0].c_str()`

notspiff: i was told to use libfmt and not pass by POD in an earlier review comment.

Rechi: not for `CLog::Log`, it doesn't use libmft and therefore doesn't compile (see jenkins log)

notspiff: alrightythen.

notspiff:
jenkins build this please

xbmc/interfaces/builtins/LibraryBuiltins.cpp
Rechi: still isn't fixed, you have done something wrong during updating the PR

Rechi: `{0}` also has to be replaced with `%s`

akva2: a fix had been squashed wrongly (into the json commit) which reverted my fix once i moved the commit in history. fixed now.

notspiff: fixed now.

notspiff:
jenkins build this please

notspiff:
i cannot get jenkins to run for this. grml

notspiff:
only failure is wayland due to warning thing. merging.

PR #13030, opened 2017-11-09
[music]Fix sort names for artists that are only album artists

DaveTBlake:
Artist sort name  was not always being correctly retrieved from the ALBUMARTISTSORT tag for some album artists. When the music file did not have Musicbrainz id tags and the artists are only album artists (not the song artist given in the ARTIST tag) the sort name was not being used.

This fix has been tested by users (test build from my repo) and solves the issue that has existed since the feature was added by #11738

Discussed here https://forum.kodi.tv/showthread.php?tid=323013


notspiff:
looks fine.

PR #13035, opened 2017-11-10
Game client dialog: Append "Saved" if emulator has a savestate

garbear:
This adds a "Saved" label next to the game client that has a savestate.

## Description
Until we get the Saved Game Manager, only one save per game is possible. Saves are not cross-emulator compatible, so starting another emulator will erase the save. Before this happens Kodi will inform you what emulator you should use, but it would be much better to indicate this in the dialog before the trial-and-error attempts occur.

## Motivation and Context
A single savestate per game isn't ideal, but until we have a Saved Game Manager it's what we're stuck with. The change here should make this limitation slightly less painful.

screen shot 2017-11-02 at 6 45 11 pm

## Types of change

### DIFF
--- a/xbmc/games/dialogs/GUIDialogSelectGameClient.cpp
+++ b/xbmc/games/dialogs/GUIDialogSelectGameClient.cpp
@@ -70,6 +70,8 @@ std::string CGUIDialogSelectGameClient::ShowAndGetGameClient(const std::string &
     {
       CFileItemPtr item(XFILE::CAddonsDirectory::FileItemFromAddon(candidate, candidate->ID()));
       item->SetLabel2(g_localizeStrings.Get(35257)); // "Installed"
+      if (item->GetPath() == selected)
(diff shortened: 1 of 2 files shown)


garbear:
jenkins build this please

xbmc/games/dialogs/GUIDialogSelectGameClient.cpp
```diff
@@ -70,6 +70,8 @@ std::string CGUIDialogSelectGameClient::ShowAndGetGameClient(const std::string &
     {
       CFileItemPtr item(XFILE::CAddonsDirectory::FileItemFromAddon(candidate, candidate->ID()));
       item->SetLabel2(g_localizeStrings.Get(35257)); // "Installed"
+      if (item->GetPath() == selected)
```
Razzeee: This if reads so weird, it might just be worth to renamed "selected" to something more helpful, like "hasSavestate"

garbear: Renamed

razzeee (approved):

PR #13040, opened 2017-11-11
[PVR] PVR client cleanup and improvement

ksooo:
First commit is a larger cleanup and routes all calls to the client addon through a single proxy method. So, common tasks before after the client addon call can be held in one place . The diff is large, thus not easy to read. Take a look at a good example snippet, to get the idea: https://github.com/xbmc/xbmc/pull/13040/commits/2a62ed20196c48f700c00599af3314fc2c7351e7#diff-d63034db2a0e41f8c4bf14542b168cd4L569

Second commit improves pvr manager shutdown performance. I've seen cases before where the client addon could not reach the backend (network problems, misconfiguration, ...) and pvr manager shutdown took ages, because it had to wait for the timeouts for a bunch of backend calls ongoing while pvr manager should be shut down - pvr gui info thread updating backend data cache and calling several client addon methods in a row is one of the candidates causing this.  

This changes are runtime-tested on macOS, latest Kodi master.

@FernetMenta  for review?

### DIFF
--- a/xbmc/addons/PVRClient.cpp
+++ b/xbmc/addons/PVRClient.cpp
@@ -120,6 +120,7 @@ void CPVRClient::ResetProperties(int iClientId /* = PVR_INVALID_CLIENT_ID */)
   m_strUserPath           = CSpecialProtocol::TranslatePath(Profile());
   m_strClientPath         = CSpecialProtocol::TranslatePath(Path());
   m_bReadyToUse           = false;
+  m_bBlockAddonCalls      = false;
   m_connectionState       = PVR_CONNECTION_STATE_UNKNOWN;
   m_prevConnectionState   = PVR_CONNECTION_STATE_UNKNOWN;
   m_ignoreClient          = false;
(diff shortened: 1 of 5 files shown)


ksooo:
jenkins build this please

FernetMenta (approved):
looks sane imo. but I think this is going to be obsolete after we split pvr and streaming api. not sure about the status of @AlwinEsch 's work

ksooo:
I don't think this will be obsolete, as it makes sense as long as you have a PVR API with more than one function, which will be still the case, even after we rip off inputstream functions.

ksooo:
I'm not aware that Alwin is working on the API split, btw.

FernetMenta:
@ksooo before the API can be split the PVR addon has to be migrated to the new design. That I hoped @AlwinEsch was working on.

ksooo:
> before the API can be split the PVR addon has to be migrated to the new design. That I hoped @AlwinEsch was working on.

 Ah, okay.

PR #13042, opened 2017-11-12
[Music]Fix album genre inconsistencies

DaveTBlake:
Fixing the inconsistent way album genre data is handled.

In Musicdb (since Eden) the album table has field `strGenres`, and there is a link table `album_genre` resolving a many to many relationship between album and genre. One could think that `strGenres` is a denormalised copy of what is in `album_genre`, but that is not always true:
1. Both are initially populated using genre values of the songs of the album. However while `strGenres` is set to a concatenation of the genre values of either the first song (if music files have Musicbrainz id tags) or last song (no mbids), `album_genre` is filled with a union of genre values from _all_ the songs.
2. Scraping can then update the `strGenres` string, but `album_genre` and `genre` tables are not updated.
3. The `strGenres` string could also be updated using JSON API (although I don't know of any app that does), but again `album_genre` and `genre` tables are not updated.

So what is in `strGenres` and `album_genre` can diverge wildly. This has gone largely unnoticed by users because:
a) All filtering by genre, and values listed on the Genres node, is based on song genre (queried from the `song_genre` link table). The value in `strGenres` is simply used as a display string.
b) Most songs on an album will all have the same genre(s),
c) Overwriting with scraped genre(s) from NFO files or online sources is reasonably rare.

Point a) is significant. Since both all filtering by genre and the Genres node, are based on querying the `song_genre`  table, the data in the `album_genre` table is actually **unused by Kodi**. There is no  reason to codify album genre values hence the `album_genre` table can be removed from the schema. 

This makes album genre similar to artist genre (or album mood, style and theme)  - a discriptive string that happens to have the name "genre" but can be completely different from the song genre values of related songs. [Note I had originally thought that both artist genre and album genres should be codified, but there simply is no need to associate ids with these strings, much like mood, style and theme, since we are not using them to navigate the library]

Continuing to initially fill the album genre from song genre values, in the same way we fill year and label from song,  makes sense to me rather than leave it blank until scraping. As a user I want to see (song) genres when listing an album, after all that could have been the filter criteria, but it is a perhaps "dirty" way to achieve that. As a user I also want my albums to have genres from the way I tagged my music files, not have to either manaually edit NFO files to match, or put-up with the genre mess that comes from online sources. Allocation of genre to music is a very personal and subjective thing.  But I would rather fix the data inconsistency first, then have a separate discussion about how and when these genre fields get populated. 

This PR also fixes the genre aspects of the long outstanding bug in SetSongDetails()  https://trac.kodi.tv/ticket/15247 reported in 2014. However this method still has other faults - updating artists is not done correctly, and it has album fields that are beyond the scope of updating a song and should be removed.

**JSON API** also needs updating because is exposes  `strGenres` and the union of song genre ids as if they are matching pairs when they are not. This is covered in a separate PR.

### DIFF
--- a/xbmc/music/MusicDatabase.cpp
+++ b/xbmc/music/MusicDatabase.cpp
@@ -165,8 +165,6 @@ void CMusicDatabase::CreateTables()
 
   CLog::Log(LOGINFO, "create album_artist table");
   m_pDS->exec("CREATE TABLE album_artist (idArtist integer, idAlbum integer, iOrder integer, strArtist text)");
-  CLog::Log(LOGINFO, "create album_genre table");
-  m_pDS->exec("CREATE TABLE album_genre (idGenre integer, idAlbum integer, iOrder integer)");
 
   CLog::Log(LOGINFO, "create genre table");
   m_pDS->exec("CREATE TABLE genre (idGenre integer primary key, strGenre varchar(256))");
@@ -217,9 +215,6 @@ void CMusicDatabase::CreateAnalytics()
   m_pDS->exec("CREATE UNIQUE INDEX idxAlbumArtist_1 ON album_artist ( idAlbum, idArtist )");
   m_pDS->exec("CREATE UNIQUE INDEX idxAlbumArtist_2 ON album_artist ( idArtist, idAlbum )");
 
-  m_pDS->exec("CREATE UNIQUE INDEX idxAlbumGenre_1 ON album_genre ( idAlbum, idGenre )");
-  m_pDS->exec("CREATE UNIQUE INDEX idxAlbumGenre_2 ON album_genre ( idGenre, idAlbum )");
-
   m_pDS->exec("CREATE INDEX idxGenre ON genre(strGenre(255))");
 
   m_pDS->exec("CREATE INDEX idxArtist ON artist(strArtist(255))");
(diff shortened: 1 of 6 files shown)


xbmc/music/MusicDatabase.cpp
```diff
@@ -1633,33 +1625,43 @@ bool CMusicDatabase::DeleteAlbumArtistsByAlbum(int idAlbum)
...
     return true;
-  
-  std::string strSQL;
-  strSQL=PrepareSQL("replace into album_genre (idGenre, idAlbum, iOrder) values(%i,%i,%i)",
-                    idGenre, idAlbum, iOrder);
-  return ExecuteQuery(strSQL);
+  }
+  catch (...)
+  {
+    CLog::Log(LOGERROR, "%s(%i) %s failed", __FUNCTION__, idSong, strSQL);
```
Rechi: `strSQL.c_str()`

DaveTBlake:
Just to confirm that from a user view point nothing changes. The album genre that they see continues to be populated in the same way.

All the change is internal: dropping data that was not actually being used, was unlikely to be used in the future and was not being maintained consistently if anyone had tried to use it.  Sorry @Razzeee and @jjd-uk if my first post could have been clearer on that.

DaveTBlake:
Discussion on Slack has been posivite about this cleanup, so intending to merge later today unless anyone objects.

PR #13043, opened 2017-11-12
start separating render loop from main thread

FernetMenta:
Motivation for starting work on this:
Allow clients to dynamically hook into render loop. This is requirered i.e. when kodi has multiple instances of VideoPlayer. (pip)
As a side producs this killed another ugly global: CWinEvents. Instead of having this global, WinEvents is now woned by Windowing.


Rechi:
wrong revision was build
jenkins build this please

xbmc/guilib/DispResource.h
```diff
@@ -27,5 +27,11 @@ class IDispResource
   virtual void OnLostDisplay() {};
   virtual void OnResetDisplay() {};
   virtual void OnAppFocusChange(bool focus) {};
+  virtual void FrameMove() {};
 };
 
+class IRenderLoop
```
garbear: Can you add a virtual destructor?

xbmc/windowing/WinSystem.cpp
```diff
@@ -273,3 +274,27 @@ KODI::WINDOWING::COSScreenSaverManager* CWinSystemBase::GetOSScreenSaver()
...
+void CWinSystemBase::RegisterRenderLoop(IRenderLoop *client)
+{
+  CSingleLock lock(m_renderLoopSection);
+  m_renderLoopClients.push_back(client);
+}
+
+void CWinSystemBase::UnegisterRenderLoop(IRenderLoop *client)
+{
+  CSingleLock lock(m_renderLoopSection);
+  std::vector<IRenderLoop*>::iterator i = find(m_renderLoopClients.begin(), m_renderLoopClients.end(), client);
```
garbear: You can use `auto` to make the end of the line better readable on smaller screens

FernetMenta: done

xbmc/windowing/wayland/WinSystemWayland.cpp
```diff
@@ -833,15 +833,15 @@ void CWinSystemWayland::SetResolutionInternal(CSizeInt size, std::int32_t scale,
         XBMC_Event msg{XBMC_MODECHANGE};
         msg.mode.res = RES_WINDOW;
         SetWindowResolution(sizes.bufferSize.Width(), sizes.bufferSize.Height());
-        CWinEvents::MessagePush(&msg);
+        dynamic_cast<CWinEventsWayland*>(m_winEvents)->MessagePush(&msg);
```
garbear: dynamic_cast can return `nullptr`, i suggest changing to static_cast

xbmc/guilib/DispResource.h
FernetMenta: objects of this type are not supposed to be deleted.

garbear: what about RetroPlayer? players are destroyed when the file is closed

xbmc/windowing/wayland/WinSystemWayland.cpp
FernetMenta: you can't use static_cast for casting interfaces. have you ever tried a static cast on an object that was derived from multiple base classes?

xbmc/guilib/DispResource.h
FernetMenta: players are not destroyed using the type IRenderLoop

xbmc/windowing/wayland/WinSystemWayland.cpp
garbear: both static and dynamic cast hit the vtable, regardless of the cast type. Casting to wayland events outside of wayland is a programmer error, so there's no need for RTTI to cause a segfault in this case

xbmc/guilib/DispResource.h
garbear: ok that makes sense. putting this in the documentation would help.

xbmc/windowing/wayland/WinSystemWayland.cpp
FernetMenta: I know that static cast fails in some cases because I already fell into that trap. 

class A : public B, public C
when casting from i.e. B to C you either have to go through A, when using static_cast, or you use dynamic_cast.

anyway, a better solution for wayland would be to maintain a member with correct type. I leave this to @pkerling

FernetMenta:
jenkins build this please

garbear (approved):
It would be nice if you could add the point above to a doxy comment for the class. other than that changes approved.

xbmc/input/touch/generic/GenericTouchActionHandler.cpp
```diff
@@ -179,7 +179,7 @@ void CGenericTouchActionHandler::sendEvent(int actionId, float x, float y, float
   newEvent.touch.y3 = y3;
   newEvent.touch.pointers = pointers;
 
-  CWinEvents::MessagePush(&newEvent);
+  g_application.OnEvent(newEvent);
```
pkerling: Is it OK to call CApplication::OnEvent from any thread?

xbmc/windowing/CMakeLists.txt
```diff
@@ -9,7 +8,7 @@ set(HEADERS OSScreenSaver.h
             XBMC_events.h
             VideoSync.h)
 
-if(CORE_SYSTEM_NAME STREQUAL linux AND NOT MIR_FOUND)
+if(CORE_SYSTEM_NAME_LC STREQUAL rbpi OR CORE_SYSTEM_NAME_LC STREQUAL gbm)
```
pkerling: I think you want `CORE_PLATFORM_NAME` here

xbmc/windowing/WinSystem.h
```diff
@@ -148,9 +150,13 @@ class CWinSystemBase
   // text input interface
   virtual void EnableTextInput(bool bEnable) {}
   virtual bool IsTextInputEnabled() { return false; }
-
   virtual std::string GetClipboardText(void);
 
+  // render loop
+  void RegisterRenderLoop(IRenderLoop *client);
+  void UnegisterRenderLoop(IRenderLoop *client);
```
pkerling: typo: `Unegister`

xbmc/windowing/CMakeLists.txt
Rechi: more precisely `CORE_PLATFORM_NAME_LC`

xbmc/windowing/wayland/WinSystemWayland.cpp
pkerling: Do we even still need a common `CWinEvents` at all? Shouldn't it be an implementation detail of the windowing implementation in question how it wants to treat the event stuff? Then the cast also would not be necessary.

xbmc/input/touch/generic/GenericTouchActionHandler.cpp
FernetMenta: it has to be ok. Ideally OnEvent should not block the caller, in particular it should not hijack the render thread for synchronous processing of the event. something for one of the next iterations

xbmc/windowing/WinSystem.h
FernetMenta: thanks

xbmc/windowing/wayland/WinSystemWayland.cpp
FernetMenta: In this iteration we still poll MessagePump through the interface. Later when windowing will own the render thread this can be dropped and windowing can send the events to application directly.

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

xbmc/input/touch/generic/GenericTouchActionHandler.cpp
pkerling: Then touch on Wayland might break in this iteration, because this will absolutely be called from the Wayland event handling thread. If you are aware of it then it's fine.

xbmc/windowing/wayland/WinSystemWayland.cpp
pkerling: Can we leave it like this for this iteration then and add a FIXME? I don't think it makes much sense to add a workaround now if it will go away in any case.

If the NULL pointer thing is still a concern, you can do `dynamic_cast<CWinEventsWayland&>(*m_winEvents).MessagePush` instead.

xbmc/input/touch/generic/GenericTouchActionHandler.cpp
FernetMenta: if an event required the main thread to be processed, it was already broken in terms of design.

xbmc/windowing/wayland/WinSystemWayland.cpp
FernetMenta: ok

FernetMenta:
jenkins build this please

xbmc/input/touch/generic/GenericTouchActionHandler.cpp
pkerling: Might be, but we still need to do something about it now that `OnEvent` is not only called from the main thread. I'm not sure what the solution is. Do we have to make sure that every code path through `OnEvent` is thread-safe?

FernetMenta: the current design is as follows: if something requires to be run on the main thread, it has to use ApplicationMessanger. That has been the original purpose of that class.
now we have to distinguish between the main thread and the render thread. ApplicationMessenger is for messages in direction of the main tread. Windowing will encapsulate the renderthread and the interface will hide the details. That means it will take care what calls it will pass to the render thread internally.

FernetMenta: btw: there is another design flaw associated with CGenericTouchInputHandler: it is one of those ugly global singletons. Kill the global and embed an instance into Wayland. Then you can handle those events internally.

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

FernetMenta:
UWP still seems to build an old version
jenkins build this please

FernetMenta:
jenkins build this please

Rechi:
@FernetMenta PRs with merge conflicts can't be build.

FernetMenta:
jenkins build this please

FernetMenta:
@Rechi @afedchin could you please provide a fix for UWP. I have no clue why all those methods a static.

afedchin:
>  I have no clue why all those methods a static.

because the code is more readable. a delegate must be
+ lambda
+ static function
+ pointer-to-member (ref class only)
+ std::function

static was preferable in this case because `CWinSystemWin10` isn't a ref class but lambdas cause a capture the `this` pointer
see https://github.com/FernetMenta/xbmc/pull/436

FernetMenta:
jenkins build this please

xbmc/guilib/DispResource.h
```diff
@@ -20,12 +20,21 @@
 
 #pragma once
 
+// interface for registering into windowing
+// to get notified about display events
+// interface only, does not control lifetime of the object
 class IDispResource
 {
 public:
-  virtual ~IDispResource() = default;
```
ksooo: Why have you deleted the dtor? If you want instances derived from this interface to get deleted correctly, even if you handle them using a reference to the interface (not the derived class), you need the virtual dtor.

IFace * i = new CDerivedFromIFace;
delete i; // will not call ~CDerivedFromIFace, ~IFace until you define virtual ~IFace()

xbmc/guilib/DispResource.h
ksooo: See my comment re. removal of virtual ~IDispResource() some lines above...

xbmc/guilib/DispResource.h
FernetMenta: objects of this type are not supposed to be deleted. by having removed the dtor I stressed this. see my comment in the header

pkerling: I think it would be a good idea to make the destructor `protected` in this case so `delete i` in the example of @ksooo will not compile even if done by accident at some point.
See also https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rc-dtor-virtual

ksooo: Even if you remove  the dtor the compiler will still generate  a non-virtual one, if needed. So, this does not help to prevent object destruction.

FernetMenta: imo it is useless to prevent for something that can't happen anyway. nothing is completely foolproof and if you don't know how to use an interface and for what purpose, just don't use it. 
this is a callback interface used to get events. the holder of the interface does not control the lifetime of the object.

ksooo: Sorry, I don't get it. Aren't there any non-abstract classes derived from this interface? If so, there will be instances of these subclasses. And these instances will be created and necessarily destructed at some point in time, even if it is on application exit. So, how can it be that an instance is not supposed to be destructed?

Is the whole point that somebody getting a pointer to the interface must not / shall not destruct the instance behind the interface? If the answer is yes, then I would make the dtor protected.

FernetMenta: Yes, the holder of this interface is not supposed to either construct nor delete an object. Everbody taking a high level view on the design will see this. I won't make a brick jump off the wall if it is used incorrectly.

ksooo: It's all up to you.😄 I just wanted to understand your intention.

MartijnKaijser:
@koying @peak3d this needs some extra android attention

MartijnKaijser:
Current PR works OK on android. Does it still need extra attention from android devs?

FernetMenta:
@MartijnKaijser thanks. No urgent task for this PR but if the design flaws don't get attention, Android will either break or slow down work of separating the render thread. 
XBMC_events are not supposed to be sent in derection of windowing.

MartijnKaijser:
@peak3d or @koying could you look into this ?

FernetMenta:
jenkins build this please

PR #13047, opened 2017-11-13
kill g_Windowing, another nasty global

FernetMenta:
another chapter in the endless story of kill ugly globals

@pkerling this will eventually enable running X11 and Wayland from the same binary.


pkerling:
Nice, very much looking forward to this!

xbmc/cores/VideoPlayer/VideoRenderers/OverlayRendererUtil.cpp
```diff
@@ -24,7 +24,7 @@
 #include "cores/VideoPlayer/DVDCodecs/Overlay/DVDOverlayImage.h"
 #include "cores/VideoPlayer/DVDCodecs/Overlay/DVDOverlaySpu.h"
 #include "cores/VideoPlayer/DVDCodecs/Overlay/DVDOverlaySSA.h"
-#include "windowing/WindowingFactory.h"
+#include "windowing/WinSystem.h"
```
notspiff: No other changes so have to ask; is this include necessary?

xbmc/ServiceManager.cpp
```diff
@@ -322,6 +323,17 @@ CFileExtensionProvider& CServiceManager::GetFileExtensionProvider()
...
 }
 
+CWinSystemBase &CServiceManager::GetWinSystem()
+{
+  return *m_winSystem.get();
+}
+
+void CServiceManager::SetWinSystem(std::unique_ptr<CWinSystemBase> &winSystem)
+{
+  m_winSystem.reset();
```
xhaggi: Necessary?

xbmc/cores/VideoPlayer/VideoRenderers/OverlayRendererUtil.cpp
FernetMenta: probably not, thanks

xbmc/ServiceManager.cpp
FernetMenta: left over from debugging, thanks

garbear:
awesome. thanks for making the RetroPlayer changes

xbmc/Application.cpp
```diff
@@ -768,7 +775,9 @@ bool CApplication::InitWindow(RESOLUTION res)
 
 bool CApplication::DestroyWindow()
 {
-  return g_Windowing.DestroyWindow();
+  return CServiceBroker::GetWinSystem().DestroyWindow();
+  std::unique_ptr<CWinSystemBase> winSystem;
```
pkerling: dead code?

FernetMenta: oops

FernetMenta:
@afedchin would be great if you could adapt the Windows ports to this change

afedchin:
It will be not easy and I haven't time on this weekend, so may be at monday.

FernetMenta:
@afedchin thanks!
@peak3d @lrusak @popcornmix could you adapt your platforms too?

lrusak:
@FernetMenta please pick https://github.com/lrusak/xbmc/commit/0e9e87679987e72175a355b67e55f23df35593d3

FernetMenta:
You are a silly bot that knows nothing. This pr is waiting for some contributions

lrusak:
Which platforms are we missing?
- [ ] Windows
- [x] OSX
- [x] iOS
- [x] X11
- [ ] Wayland
- [ ] Mir
- [x] GBM
- [ ] Android
- [x] Amlogic
- [ ] RPi

FernetMenta:
Windows, RPi, Android, (Mir) are missing.

pkerling:
What's the stance on Mir? Is it maintained?

FernetMenta:
> What's the stance on Mir? Is it maintained?
@BrandonSchaefer ^^

BrandonSchaefer:
Right now Mir supports Wayland clients. So the only reason to maintain a Mir backend currently is to wait for the Wayland support in Mir to get as good as the Mir client support. So let me test if the current Wayland core support in Mir is enough to possibly remove the Mir backend.

For this fix though it would be pretty easy to fix up the Mir parts (I can get a diff soon-ish if testing the Wayland core bits isnt good enough yet, also yay no more g_window globa!). 

Also more info on Mir and its current direction:
https://community.ubuntu.com/c/mir

popcornmix:
@FernetMenta I have most of pi stuff fixed up. OMXImage is the trickier part:
https://github.com/xbmc/xbmc/blob/master/xbmc/cores/omxplayer/OMXImage.cpp#L339

Can you suggest the desired path to get egl display/context info in the new world?

FernetMenta:
@popcornmix the fastest method is to typecast the genering interface to the specific one of your implementation. A better way is to pass required info to the instance when you create it.

xbmc/input/MouseStat.cpp
```diff
@@ -231,7 +234,9 @@ void CMouseStat::SetActive(bool active /*=true*/)
   // 1. The mouse is active (it has been moved) AND
   // 2. The XBMC mouse is disabled in settings AND
   // 3. XBMC is not in fullscreen.
-  g_Windowing.ShowOSMouse(m_mouseState.active && !IsEnabled() && !g_Windowing.IsFullScreen());
+  CWinSystemBase &winSystem = CServiceBroker::GetWinSystem();
+  if (&winSystem)
```
Rechi: reference cannot be bound to dereferenced null pointer in well-defined C++ code; pointer may be assumed to always convert to true [-Wundefined-bool-conversion]

I think you already have to handle the null pointer case in `CWinSystemBase &CServiceManager::GetWinSystem()` if it can even happen.

FernetMenta: this is a left-over and will disappear soon when I'll move MouseStat from InputManager to WinSystem

FernetMenta:
@afedchin 
  C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\agile.h(19): fatal error C1189: #error:  agile.h can only be used with /ZW (compiling source file C:\jenkins_slave\workspace\WIN-64\xbmc\guilib\D3DResource.cpp)

xbmc/rendering/dx/RenderContext.h
```diff
@@ -0,0 +1,69 @@
...
+#pragma once
+
+#if defined(TARGET_WINDOWS_DESKTOP)
+#include "windowing/windows/WinSystemWin32DX.h"
+#elif defined(TARGET_WINDOWS_STORE)
+#include "windowing/win10/WinSystemWin10DX.h"
+#endif
+#include "xbmc/ServiceBroker.h"
+
+#include <agile.h>
```
afedchin: please move this inside `#elif defined(TARGET_WINDOWS_STORE)`
this will fix build error on win64/win32

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

FernetMenta:
jenkins build this please

garbear:
@FernetMenta small improvement for you: https://github.com/FernetMenta/xbmc/pull/440

xbmc/windowing/X11/WinEventsX11.cpp
```diff
@@ -18,18 +18,17 @@
 *  http://www.gnu.org/copyleft/gpl.html
 *
 */
+#pragma once
```
Rechi: in a cpp file?

FernetMenta: oops :)

FernetMenta:
@popcornmix do you plan to submit PI adaptations in the near future?

lrusak:
Do we want to wait on Mir?

FernetMenta:
I am not aware that we have any/many alpha testers on Mir. Hence moving forward has higher prio. Patches for Mir can be submitted at any time later.

FernetMenta:
Linux-64 test suite error in ThreadLocal unrelated to this PR. Test runs fine on my local system.

PR #13049, opened 2017-11-14
Fix Back button in Android TV Remote Control app

garbear:
The Back button shows up as a joystick button, which is mapped to the B button in https://github.com/xbmc/peripheral.joystick/pull/129 (and backported to Krypton).

This PR makes the B button behave like the Backspace button on a keyboard, which is how the button is interpreted when peripheral.joystick is disabled. Now, the Back button should work the same whether it's operating as a joystick or keyboard.

This required two steps: changing joystick.xml to match keyboard.xml and fixing a joystick bug. The bug caused the FullscreenVideo window to close immediately upon reopening from the Home screen.

## Motivation and Context
Reported here: https://forum.kodi.tv/showthread.php?tid=308480&pid=2667452#pid2667452

## How Has This Been Tested?
* Tested on Shield TV with Android TV Remote Control app
* Tested on Linux with Xbox 360 controller

Android TV Remote Control:
android tv remote control

## Types of change

### DIFF
--- a/system/keymaps/joystick.xml
+++ b/system/keymaps/joystick.xml
@@ -85,7 +85,8 @@
   <FullscreenVideo>
     <joystick profile="game.controller.default">
       <a>Pause</a>
-      <b>Stop</b>
+      <b>Fullscreen</b>
+      <b holdtime="500">Stop</b>
       <x>OSD</x>
       <y>FullScreen</y>
       <start>Info</start>
--- a/system/keymaps/joystick.xml
+++ b/system/keymaps/joystick.xml
@@ -70,6 +70,12 @@
       <rightstick direction="down">VolumeDown</rightstick>
     </joystick>
   </global>
+  <Home>
+    <joystick profile="game.controller.default">
+      <b>Back</b>
+      <b holdtime="500">ActivateWindow(ShutdownMenu)</b>
+    </joystick>
+  </Home>
   <FileManager>
     <joystick profile="game.controller.default">
       <rightbumper>Highlight</rightbumper>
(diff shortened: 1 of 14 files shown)


garbear:
jenkins build this please

bb10:
I think this is very much a breaking change. Please see: https://github.com/xbmc/xbmc/commit/09a8dae7e9384278b446e464202059803c22c3e8#commitcomment-31427168

bb10:
That would be a great improvement, yes.

garbear:
@bb10 I raised your idea on the forum thread: https://forum.kodi.tv/showthread.php?tid=308480&pid=2795177#pid2795177

Just to see what others think. I'm for switching the two like you proposed.

bb10:
Any update on this? Would be a shame if this made it into the final v18 release.

garbear:
Can you open a PR with the ideas suggested in the thread? I can't until this weekend, so to hit v18 it should be opened today.

bb10:
I hope I did this right, it's my first PR.
https://github.com/xbmc/xbmc/pull/15267 is a conservative change, just reversing the B button actions.
https://github.com/xbmc/xbmc/pull/15268 is a more radical change, which has a bunch of back button additions and changes and also reverses the B button actions like the former.

I did not feel comfortable changing the keyboard.xml atm. I guess we could just change it and see if anyone complains and revert it in a point release if anyone does?

The problem is that CiNcH somehow expects B and backspace to behave the same, but they never have and any changes to them will be breaking for any keyboard and joystick users.

Let me know what you think. 😃

